PR #30246, opened 2018-12-03
Create a `define_editor` command to customize `@edit` and friends.

haberdashPI:
This is my initial proposal to create a `define_editor` command to make it possible to customize how `@edit` tells an editor about line numbers. I described this in #29979. As implemented, one can define a new editor, like so, in `~/.julia/config/startup.jl`

```julia
define_editor(r"\bemacs\b.*(-nw|--no-window-system)",background=false) do cmd,path,line
    `$cmd +$line $path`
end
```

If the pattern matches the value of `$EDITOR` the `do` function body will be used to issue the editor command. At present the function is not exported, though maybe it would be better if it were?

I recognize there might a little bit more work to clean this up (docs? tests?). However, before I do any clean up, I wanted to check that there weren't any other changes I should make to the interface or implementation.

PR #34107, opened 2019-12-15
InteractiveUtils: make editors pluggable via `define_editor`

StefanKarpinski:
This function makes it possible to properly define how to open a file and pass a specific line number for a given editor using `@edit` and friends.

I meant to do a few minor edits on top of https://github.com/JuliaLang/julia/pull/30246 but I ended up doing a pretty significant refactor/rewrite. With apologies to @haberdashPI both for taking so long to review and for rewriting his work. I've squashed all of @haberdashPI's commits into a single commit and then put my refactor after that. Instead of using custom types, the editor stack is now just a vector of callback function taking `(cmd, path, line)`, each of which can return `true` to indicate that they have handled editing the file or `false` to decline handling it. The `edit` command now just loops through all the editor callbacks and returns when one returns true. This simplifies the logic a bit (all the complexity is pushed into `define_editor`, which defines a fancy closure that does all this) but allows a general editing hook to do literally anything it wants, so the hooking mechanism is more general.

Oh, I also unexported `define_editor`: let's not make this an official API just yet; it may need some tweaks, especially to how we match editors, so consider it experimental for now.

### DIFF
--- a/stdlib/InteractiveUtils/docs/src/index.md
+++ b/stdlib/InteractiveUtils/docs/src/index.md
@@ -9,6 +9,7 @@ InteractiveUtils.subtypes
 InteractiveUtils.edit(::AbstractString, ::Integer)
 InteractiveUtils.edit(::Any)
 InteractiveUtils.@edit
+InteractiveUtils.define_editor
 InteractiveUtils.less(::AbstractString)
 InteractiveUtils.less(::Any)
 InteractiveUtils.@less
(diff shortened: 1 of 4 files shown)


StefanKarpinski:
I'd like to get this into 1.4 since I promised @haberdashPI that I would review for 1.4 and his original PR has been open for over a year.

haberdashPI:
Thanks @StefanKarpinski, much appreciated! Glad to see this make it in.

PR #34082, opened 2019-12-12
Bump dependencies to fix ppc64le page size issues

staticfloat:
This PR is to see if we can't get PPC64le bootstrapping again.

Dependencies bumped so far:

- [X] GMP
- [X] MPFR
- [X] MbedTLS
- [X] libssh2
- [X] libgit2
- [X] libCURL

PR #34108, opened 2019-12-15
Revert "Bump dependencies to fix ppc64le page size issues"

StefanKarpinski:
Reverts JuliaLang/julia#34082. Since it seems to be causing download hash failures.


giordano:
I think reverting only the change to `SUITESPARSE_BB_URL_BASE` in [`deps/suitesparse.mk`](https://github.com/JuliaLang/julia/commit/cfb6e79146c963286115cb76781aaa525f7c99fc#diff-a631d2e400d40fcebd84976af4c2ff91) might be sufficient.  I was also almost about to delete the tag referenced by that change, as it's wrong.

giordano:
Ok, I think something went very wrong with the registration of a new tag in SparseSuite.

StefanKarpinski:
I'd rather revert this whole thing and get back to working CI; someone can unrevert it along with the rest of the fix once it's ready. In the meantime, this is causing all CI runs to fail.

giordano:
> I'd rather revert this whole thing and get back to working CI

Yes, that's probably safer now.

vchuravy:
Reverted it for now, good way of finding a race condition in the system...

PR #34110, opened 2019-12-15
Abstract doc

matbesancon:
[skip ci]


matbesancon:
I got tricked while re-implementing a dictionary, implementing the iteration interface as an iterator of `Tuple{K,V}`. It all goes fine and collect is the only function that errored out because eltype(D) == Pair{K, V} is assumed here for any AbstractDict.

base/essentials.jl
```diff
@@ -19,6 +19,7 @@ abstract type AbstractSet{T} end
 
 Supertype for dictionary-like types with keys of type `K` and values of type `V`.
 [`Dict`](@ref), [`IdDict`](@ref) and other types are subtypes of this.
+An `AbstractDict{K,V}` should be an iterator of `Pair{K, V}`.
```
JeffBezanson:
```suggestion
An `AbstractDict{K, V}` should be an iterator of `Pair{K, V}`.
```

PR #33465, opened 2019-10-03
Add winsomely that converts args to a command line

musm:
From https://github.com/JuliaLang/julia/pull/30614    adressing @vtnash 's comments to move the function to a more appropriate location.

~~However see my comment there https://github.com/JuliaLang/julia/pull/30614/files/f3d3cf36c73b646b2d65a91fa2289e22d64b23f5#diff-06a75b3b926c52ae4594356ce1eb5405~~

~~Hopefully I can get some inputs on how to test the special meta-char processing by cmd and also my concerns in the above comment.~~

The code  implements the  inverse of `CommandLineToArgvW`.

**References** 

https://blogs.msdn.microsoft.com/twistylittlepassagesallalike/2011/04/23/everyone-quotes-command-line-arguments-the-wrong-way/

https://devblogs.microsoft.com/oldnewthing/?p=12833 

https://docs.microsoft.com/en-us/cpp/c-language/parsing-c-command-line-arguments?redirectedfrom=MSDN&view=vs-2019

Issue #29580, opened 2018-10-10
raw string repr incorrect


vtjnash:
Various string macros don't repr correctly. I thought we had an issue for this, but can't find it.
```
julia> r"\\\\"
r"\\"

julia> s"\n"
s"\\n"
```

PR #33474, opened 2019-10-04
REPL shell mode for Windows

musm:
Previous Attempts: https://github.com/JuliaLang/julia/pull/31490 https://github.com/JuliaLang/julia/pull/23703 https://github.com/JuliaLang/julia/pull/21294 https://github.com/JuliaLang/julia/pull/20776 https://github.com/JuliaLang/julia/pull/11478 https://github.com/JuliaLang/julia/pull/33427 https://github.com/JuliaLang/julia/pull/11478

Some of these approaches tried to implement special Windows rule in the `shell_escape` mechanism, I think some also tried to pass 'raw' strings to the shell, all with varying degrees of success.

Fixes https://github.com/JuliaLang/julia/issues/6316 https://github.com/JuliaLang/julia/issues/23597

**References**

Powershell:

- https://www.red-gate.com/simple-talk/wp-content/uploads/imported/2264-PS%20Quoting%20Wallchart_1_0_1.pdf
- https://stackoverflow.com/questions/44981177/run-powershell-custom-function-in-new-window/44986341#44986341
- https://ss64.com/ps/syntax-esc.html

CMD:

- https://www.robvanderwoude.com/escapechars.php
- https://ss64.com/nt/syntax-esc.html

PR #34111, opened 2019-12-15
Refactor shell_escape_winsomely() and add escaping function for CMD.EXE syntax

mgkuhn:
Refactored shell_escape_winsomely() into two separate functions escape_microsoft_c_args() and escape_raw_string(), such that the latter becomes separately available (needed for #29580) and to make it easier to understand.
As requested by @vtjnash in #33465.
Also added Windows `CMD.EXE` escaping function shell_escape_wincmd() as requested by @vtjnash to be separated out from #30614 into base/shell.jl (for SSHManager to be able to prepare Windows command lines)


mgkuhn:
When testing this patch, I had big problems making `julia-sysimg-ji` and eventually, after lots of trial and error, in the absence of any meaningful error message, I had to uncomment in `shell.jl` some lines containing the expressions `r"[ \t\"]"` or `Cmd`. Otherwise otherwise `make` mysteriously failed with:
```
$ make
    JULIA usr/lib/julia/sys.ji
[...]
shell.jl
error during bootstrap:

jl_undefined_var_error at julia/src/rtutils.c:130
jl_eval_global_var at julia/src/interpreter.c:393
getproperty at ./Base.jl:26
[...]
main at /local/bigscratch/mgk25/w.github/julia/ui/repl.c:217
__libc_start_main at /lib/x86_64-linux-gnu/libc.so.6 (unknown line)
_start at /local/bigscratch/mgk25/w.github/julia/usr/bin/julia (unknown line)

*** This error might be fixed by running `make clean`. If the error persists, try `make cleanall`. ***
sysimage.mk:67: recipe for target '/local/bigscratch/mgk25/w.github/julia/usr/lib/julia/sys.ji' failed
Makefile:81: recipe for target 'julia-sysimg-ji' failed
```

(`make clean` did not help, but removing `r"[ \t\"]"` and `Cmd` did.)

Any idea what is going on there?

vchuravy:
`regex.jl` is loaded after `shell.jl`, see https://github.com/JuliaLang/julia/blob/a6bf74a2f9d69f31060d6294ab0e59bf61ce26d8/base/Base.jl#L177-L178

base/shell.jl
```diff
@@ -255,60 +255,91 @@ shell_escape_posixly(args::AbstractString...) =
...
+```
+"""
+function shell_escape_wincmd(io, s::AbstractString)
+    # https://stackoverflow.com/a/4095133/1990689
+    quoted = false
+    for c in s
+        if c == '"'
+            quoted = !quoted
+        else
+            if !quoted && c in ( '(', ')', '!', '^', '<', '>', '&', '|' )
```
vtjnash:
```suggestion
            if !quoted && c in ( '%', '(', ')', '!', '^', '<', '>', '&', '|' )
```

(actually, `(` is not special, as that stack overflow link points out, but I think it'd just confused folks not to include it anyways)

base/shell.jl
```diff
@@ -255,60 +255,91 @@ shell_escape_posixly(args::AbstractString...) =
...
+    shell_escape_wincmd(s::AbstractString)
+    shell_escape_wincmd(io, s::AbstractString)
+
+The unexported `shell_escape_wincmd` function escapes the meta
+characters `()!^<>&|` processed by the Windows `cmd.exe` shell: it
+places a `^` in front of any metacharacter that follows an even number
+of quotation marks on the command line.
+
+The percent sign (`%`) is not escaped, therefore shell variable
+references (like `%USER%`) will still be substituted by `cmd.exe`.
```
vtjnash: This is not right: this is what `shell_escape_wincmd` exists to prevent.
```suggestion
```

base/shell.jl
```diff
@@ -255,60 +255,91 @@ shell_escape_posixly(args::AbstractString...) =
...
+"""
+    shell_escape_wincmd(s::AbstractString)
+    shell_escape_wincmd(io, s::AbstractString)
+
+The unexported `shell_escape_wincmd` function escapes the meta
+characters `()!^<>&|` processed by the Windows `cmd.exe` shell: it
+places a `^` in front of any metacharacter that follows an even number
+of quotation marks on the command line.
+
+The percent sign (`%`) is not escaped, therefore shell variable
```
vtjnash:
```suggestion
```

base/shell.jl
```diff
@@ -255,60 +255,91 @@ shell_escape_posixly(args::AbstractString...) =
...
-                    write(io, '\\')
-                end
-                backslashes = 0
-                write(io, c)
+"""
+    shell_escape_wincmd(s::AbstractString)
+    shell_escape_wincmd(io, s::AbstractString)
+
+The unexported `shell_escape_wincmd` function escapes the meta
+characters `()!^<>&|` processed by the Windows `cmd.exe` shell: it
```
vtjnash:
```suggestion
characters `%()!^<>&|` processed by the Windows `cmd.exe` shell: it
```

base/shell.jl
```diff
@@ -255,60 +255,91 @@ shell_escape_posixly(args::AbstractString...) =
...
+See also: [`escape_microsoft_c_args`](@ref), [`shell_escape_posixly`](@ref)
+
+# Example
+```jldoctest
+julia> println(shell_escape_wincmd(escape_microsoft_c_args("^\\\"^\\", "^ C")))
+"^\\\\\\"^^\\\\" "^^ C"
+```
+"""
+function shell_escape_wincmd(io, s::AbstractString)
+    # https://stackoverflow.com/a/4095133/1990689
```
vtjnash:
```suggestion
    # https://stackoverflow.com/a/4095133/1990689
    any(c -> c in ('\r', '\n'), s) && throw(ArgumentError("Encountered unsupported character by CMD."))
```

from: https://github.com/JuliaLang/julia/blob/09c1d61ae2ff77f53c0362a2a3cc31a2d5d95e3c/base/shell.jl#L316-L321

base/shell.jl
```diff
@@ -255,60 +255,91 @@ shell_escape_posixly(args::AbstractString...) =
...
-                write(io, c)
+"""
+    shell_escape_wincmd(s::AbstractString)
+    shell_escape_wincmd(io, s::AbstractString)
+
+The unexported `shell_escape_wincmd` function escapes the meta
+characters `()!^<>&|` processed by the Windows `cmd.exe` shell: it
+places a `^` in front of any metacharacter that follows an even number
+of quotation marks on the command line.
+
```
vtjnash:
```suggestion
The result is safe to pass as an argument to a command call being processed by `CMD.exe /c` and will be received verbatim by the target application (or this function will fail with an ArgumentError).
This function may be useful in concert with the `windows_verbatim` flag to [`Cmd`](@ref) when constructing process pipelines.

!warning
    The argument parsing done by CMD when calling batch files (either inside `.bat` files or as arguments to them) is not compatible with the output of this function.

!important
    Due to a peculiar behavior of the CMD parser/interpreter, each command after a literal `|` character
    (indicating a command pipeline) must have `shell_escape_wincmd` applied twice since it will be parsed twice by CMD. For example:
    ```
    to_print = "All for 1 & 1 for all!"
    run(Cmd(Cmd(["cmd /c \"break | echo \$(shell_escape_wincmd(shell_escape_wincmd(to_print)))"]), windows_verbatim=true))
    ```
```
(adopting some of the advisory text from #33474):https://github.com/JuliaLang/julia/blob/09c1d61ae2ff77f53c0362a2a3cc31a2d5d95e3c/base/shell.jl#L324-L345

base/shell.jl
```diff
@@ -255,60 +255,91 @@ shell_escape_posixly(args::AbstractString...) =
...
+            write(io, arg)
+        end
+    end
+end
+escape_microsoft_c_args(args::AbstractString...) =
+    sprint(escape_microsoft_c_args, args...;
+           sizehint = (sum(sizeof.(args)) + 3*length(args)))
+
+# alias for an earlier implementation (to be removed)
 shell_escape_winsomely(args::AbstractString...) =
```
musm: Might as well just get rid of these here  `shell_escape_winsomely`

base/shell.jl
```diff
@@ -255,60 +255,91 @@ shell_escape_posixly(args::AbstractString...) =
...
+function escape_microsoft_c_args(io::IO, args::AbstractString...)
+    # http://daviddeley.com/autohotkey/parameters/parameters.htm#WINCRULES
+    first = true
+    for arg in args
+        if first
+            first = false
+        else
+            write(io, ' ')  # separator
+        end
+        if isempty(arg) || occursin(r"[ \t\"]", arg)
```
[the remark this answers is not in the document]
mgkuhn: Depends on the length. On my computer, if `length(arg)<27` then there are about 50 ns setup penalty associated with regex use:
```
julia> using BenchmarkTools
julia> arg = 'a'^1; @btime occursin(r"[ \t\"]", $arg) ; @btime any(c -> c in (' ', '\t', '"'), $arg)
  56.275 ns (0 allocations: 0 bytes)
  7.988 ns (0 allocations: 0 bytes)
```
For `length(arg)==27` there is no difference:
```
julia> arg = 'a'^27; @btime occursin(r"[ \t\"]", $arg) ; @btime any(c -> c in (' ', '\t', '"'), $arg)
  88.455 ns (0 allocations: 0 bytes)
  89.344 ns (0 allocations: 0 bytes)
```
For `length(arg)>27` regex is faster, asymptotically twice as fast:
```
julia> arg = 'a'^1000; @btime occursin(r"[ \t\"]", $arg) ; @btime any(c -> c in (' ', '\t', '"'), $arg)
  1.380 μs (0 allocations: 0 bytes)
  2.673 μs (0 allocations: 0 bytes)
false
```
So overall, I still prefer using a regular expression here (and I suspect they might be beneficially used in some of the other shell escaping functions as well).

base/shell.jl
mgkuhn: It's not so easy! The problem is that CMD.EXE performs variable substitution at the very beginning, _before_ processing the ^ and " escaping characters. See the following experiment to demonstrate this:
```
C:\Users\mgk25>echo %USER%            
dc\mgk25                     
                             
C:\Users\mgk25>echo ^%USER% 
dc\mgk25                     
                             
C:\Users\mgk25>echo ^%USER^% 
%USER%                       
                             
C:\Users\mgk25>set USER^^=foo

C:\Users\mgk25>echo ^%USER^%  
foo                          

```
So if you try to escape % with ^%, you may in fact just be adding a ^ to the variable name that will be substituted. Yes, I know its crazy, but Windows always was an acquired taste ...

I believe it is impossible for `shell_escape_wincmd` to escape % effectively. I can reformulate the documentation to clarify that passing through % was not our choice, but was done to clarify that escaping it would not have been effective anyway.

base/shell.jl
```diff
@@ -255,60 +255,91 @@ shell_escape_posixly(args::AbstractString...) =
...
+passed to many Windows command-line applications.
+
+Microsoft Windows passes the entire command line as a single string to
+the application (unlike POSIX systems, where the shell splits the
+command line into a list of arguments). Many Windows API applications
+(including julia.exe), use the conventions of the [Microsoft C/C++
+runtime](https://docs.microsoft.com/en-us/cpp/c-language/parsing-c-command-line-arguments)
+to split that command line into a list of strings.
+
+This function implements the inverse of such a C runtime command-line
```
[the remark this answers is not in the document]
mgkuhn: ok

base/shell.jl
vtjnash: Oh, wow. Yeah, I completely missed that. (I doubt I'm alone too, since it means 8.10 in http://daviddeley.com/autohotkey/parameters/parameters.htm#WINCRULES is similarly wrong). It's very unexpected to realize that `%` is _not_ a cmd shell metacharacter, so `^%` is no different from say `^A`. It's instead a half separate pass of the parser.

Following the someone more detailed guide at https://stackoverflow.com/questions/4094699/how-does-the-windows-command-interpreter-cmd-exe-parse-scripts/7970912#7970912, I realized there's a workaround for rules that is valid for your experiment above. That is, an almost-working escape sequence for `%` is `%^\n`, as can be seen in the below example:

```
julia> run(pipeline(`cmd /Q /K set PROMPT=\?`, stdin=IOBuffer("echo %^\nOS%^\n\n\n")))

?More? More? More? %OS%


?Process(`cmd /Q /K set 'PROMPT=?'`, ProcessExited(0))
```

Note that passing this to `/C` fails though because of rule 1.05: "the remainder of the line from the <LF> onward" is dropped (ignored). Note that we are gaining an extra newline on the output too, since we're running afoul of the quoting rules for `\n` (it's `^\n\n`) to handle the trailing `%`—there would need to be an exception to the rule for when `%` is at the end of the line already.

Also, perhaps we should mention that the user is advised to normally pass the S flag as such: `/S /C " $the $command $line "` to avoid losing their final `"` mark in `$line` contents under very rare circumstances.

mgkuhn:
@vtjnash I just committed a rewrite of shell_escape_wincmd() that now passes through unescaped double-quotes only in pairs, to ensure that the quote flag in CMD.EXE remains cleared afterwards. This is essential when using shell_escape_wincmd() multiple times in a line to build up a more complex command line.

base/shell.jl
mgkuhn: This is starting to get rather esoteric. I actually want to pass through % unescaped, because my use case is building a command line to be sent with ssh to _another_ Windows server, where it can be perfectly useful to have variables such as %APPDATA% substituted remotely (as I don't have local access to them). Also, in the ssh use case, I am not calling CMD.EXE, so I can't pass any flags to it, the remote sshd is doing that.

If someone really needs a variant of this function that also goes to such great lengths to escape `%`, then I would be happy to draft a separate function `shell_escape_wincmd2` that has that property. However, the algorithm used will be completely different, because in order to escape every `%` with your hack, I can no longer pass through pairs of `"` unescaped, as they too might contain a `%`. Escaping every `%` means also escaping every `"`, and I'd rather not do the latter as it makes all the most common cases look very odd.

base/shell.jl
mgkuhn: Done in latest commit, although using regular expressions (benchmarks asymptotically faster) and also including `\0`.

base/shell.jl
mgkuhn: I remain confused about the exact meta character nature of `(` and `)`, so prefer to keep both escaped for now.

base/shell.jl
```diff
@@ -255,60 +255,107 @@ shell_escape_posixly(args::AbstractString...) =
...
+            write(io, arg)
+        end
+    end
+end
+escape_microsoft_c_args(args::AbstractString...) =
+    sprint(escape_microsoft_c_args, args...;
+           sizehint = (sum(sizeof.(args)) + 3*length(args)))
+
+# alias for an earlier implementation (to be removed)
 shell_escape_winsomely(args::AbstractString...) =
```
musm: I think we should just get rid of `shell_escape_winsomely` and `print_shell_escaped_winsomely` in favor of the implementations here

base/shell.jl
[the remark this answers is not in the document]
musm: Isn't it more common to have `arg < 27` than long arguments with `arg>27` .

base/shell.jl
mgkuhn: done

base/shell.jl
mgkuhn: I'm afraid I don't quite understand yet that suggested addition and I am not really familiar yet with all the behind-the-scenes processing that may go on during `run(Cmd(Cmd(["cmd /c \"...` on a Windows machine. My own use case and expertise is preparing command lines that get sent via ssh to a Windows server running OpenSSH sshd (e.g., in Distributed), where no libuv processing of command-line arguments passed to Windows processes is involved. I couldn't reproduce yet myself the effect referred to in this suggestion, and therefore would be more comfortable if we left this to a separate PR.

base/shell.jl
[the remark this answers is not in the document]
mgkuhn: We are talking a maximum of ~50 ns benchmark overhead here on my 8-year-old PC (the time in which a bit travels 10 metres on a cable), which is many orders of magnitude less than the overhead of invoking a new process.

base/shell.jl
vtjnash: Yeah, that's really abysmal behavior/design by CMD since it means you can't easily have one script call another (say, on another machine). I guess we could also attempt to hack it with `ENV["PCTESC"]="%"; replace(cmd, "%" => "%PCTESC%")`, but that seems pretty brittle too.

base/shell.jl
vtjnash: > all the behind-the-scenes processing

vtjnash: > all the behind-the-scenes processing

There is none. That's the point of using that syntax and flag: to disable all helpers and get the raw string without libuv processing or such.

Though I would perhaps further clarify here that it may be best passed as `CMD.exe /S /C " prog args "` (with the '/S' and surrounding " pair).

base/shell.jl
```diff
@@ -255,60 +255,101 @@ shell_escape_posixly(args::AbstractString...) =
...
+"a^^\\"^o\\"^^u^\\""
+```
+"""
+function shell_escape_wincmd(io::IO, s::AbstractString)
+    # https://stackoverflow.com/a/4095133/1990689
+    occursin(r"[\r\n\0]", s) &&
+        throw(ArgumentError("control character unsupported by CMD.EXE"))
+    i = 1
+    len = ncodeunits(s)
+    if len > 0 && codeunit(s,1) == UInt8('@')
```
vtjnash: I think this assumes that `s::String` (e.g. that the code unit is ascii-compatible bytes)?
```suggestion
    if len > 0 && first(s) == '@'
```

base/shell.jl
```diff
@@ -255,60 +255,101 @@ shell_escape_posixly(args::AbstractString...) =
...
                 write(io, c)
             end
         end
-        # escape all backslashes, letting the terminating double quote we add below to then be interpreted as a special char
-        quotes && (backslashes *= 2)
-        for j = 1:backslashes
-            write(io, '\\')
-        end
-        quotes && write(io, '"')
+        i += 1
```
vtjnash: This would assume that the UTF-8 string is ASCII

base/shell.jl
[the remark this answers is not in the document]
vtjnash: Either is fine with me.

base/shell.jl
mgkuhn: Fixed in the latest commit. I no loner try to bypass the UTF-8 decoder and now use the AbstractString API properly (at a 30%-60% performance penalty).

base/shell.jl
mgkuhn: Fixed.

mgkuhn:
Any chance this could still go into 1.4, such that the old `shell_escape_winsomely()` doesn't end up in a release?

base/shell.jl
```diff
@@ -255,60 +255,101 @@ shell_escape_posixly(args::AbstractString...) =
...
+    if len > 0 && s[1] == '@'
+        write(io, '^')
+    end
+    while i <= len
+        c = s[i]
+        if c == '"' && (j = findnext('"', s, nextind(s,i))) !== nothing
+            write(io, SubString(s,i,j))
+            i = j
+        else
+            if c in ('"', '(', ')', '!', '^', '<', '>', '&', '|')
```
vtjnash: I just realized that `cmd /?` emits this list, but that I don't know where this list comes from:
```
The special characters that require quotes are:
     <space>
     &()[]{}^=;!'+,`~
```

mgkuhn: I hadn't spotted that list before, but it also seems wrong, at least for my use case of command invocation. For example: typing into the command line `julia -e "print(\"hi\")">b` or `julia -e "print(\">\")"`quite clearly demonstrates that any `>` in a `CMD.EXE` command-line argument needs to be escaped to not trigger stdout redirect by `CMD.EXE`. But the `>` character is not listed in that last line of `cmd /?`.

Reading the text before, it may well be meant in the context of the “completion” function, i.e. it may be a list of characters that the authors of the completion function worried about, and that function may work differently in different contexts. That may not be the same list of characters that we need to escape in command invocation, i.e. when sending strings such as `set var=value&&pushd directory&&julia -e "print(\"...\")"`. When you ask for filename completion after having typed a filename that starts with `>`, you may want to redirect rather than have the `>` escaped.

vtjnash: Yes, but note that the list is intentionally incomplete with respect to our usage since `\/:*?<>|` (plus newline) can't appear in the result of the completion, so any of those might also be candidates for this list. The question then is where some of the others come from??

mgkuhn: The [Jan Erik / Dave Benham description of the CMD.EXE parser](https://stackoverflow.com/questions/4094699/how-does-the-windows-command-interpreter-cmd-exe-parse-scripts/4095133#4095133) does say that under certain circumstances in Phase 7 if the quote flag is off, then when testing if a command token is an internal command, it “break[s] the command token before the first occurrence of `+` `/` `[` `]` `<space>` `<tab>` `,` `;` or `=`”, which probably accounts for the remaining characters in that list. I'm not claiming to fully understand yet all of this, but I doubt it is relevant to our use cases. These characters seem just used as token delimiters, not as special characters that trigger substitutions of their own in regular command-line arguments.

[Note that my original approach (which did not factor Microsoft C ARGV quoting and CMD.EXE quoting into separate functions) would have been able to use quotes more aggressively (i.e., quote something for the C library to unquote because CMD.EXE might interpret it as a meta character, not because the C library might), and thereby guarantee far more easily that none of these contexts are ever reached. But I think we are fine. Can you construct any use case where my proposed current implementation is not already doing the right thing?]

mgkuhn:
@vtjnash I would quite like to merge it such that we can then proceed with #30614 to make Distributed's SSHManager compatible with `cmd.exe`. We can always iron out later obscure special cases that might be needed to synthesize complex batch scripts.

I'm quite confident that the existing code is transparent for all printable characters (as long as they do not contain existing environment variable names between percent signs). All I wanted is to be confident that arbitrary single-line Julia source code will pass through `cmd.exe` unharmed, which was my original motivation here (to be able to eventually replace `Distributed`-specific command-line options such as `julia --worker --bind-to ...` with just generic `julia -e "..."` remote invocations, such that `Distributed` could eventually become just another regular package that does not rely on any special options hard-coded into Julia's command-line parser.

mgkuhn:
A rebased version of this PR is now at #38352.

PR #34112, opened 2019-12-16
Update the custom type broadcasting example

fjarri:
Without this method, the following code fails:
```
a = ArrayAndChar([1, 2, 3, 4], 'x')
tp = UInt64
func(x) = convert(tp, x)
d = func.(a)
```

Also, if someone knows more about broadcasting: 
- is it now a complete definition, or should some other methods be added?
- why, if one writes `func` as `func(x) = convert(UInt64, x)`, `Extruded` does not appear in the broadcast expression?

### DIFF
--- a/doc/src/manual/interfaces.md
+++ b/doc/src/manual/interfaces.md
@@ -565,6 +565,7 @@ end
 find_aac(bc::Base.Broadcast.Broadcasted) = find_aac(bc.args)
 find_aac(args::Tuple) = find_aac(find_aac(args[1]), Base.tail(args))
 find_aac(x) = x
+find_aac(x::Base.Broadcast.Extruded) = x.x
 find_aac(::Tuple{}) = nothing
 find_aac(a::ArrayAndChar, rest) = a
 find_aac(::Any, rest) = find_aac(rest)
--- a/doc/src/manual/interfaces.md
+++ b/doc/src/manual/interfaces.md
@@ -570,7 +570,7 @@ find_aac(::Tuple{}) = nothing
 find_aac(a::ArrayAndChar, rest) = a
 find_aac(::Any, rest) = find_aac(rest)
 # output
-find_aac (generic function with 6 methods)
+find_aac (generic function with 7 methods)
 ```
 
 From these definitions, one obtains the following behavior:


mbauman:
Hm, this is unfortunate.  It'd be nice to figure out a way in which we don't need to expose this implementation detail.

> why, if one writes func as func(x) = convert(UInt64, x), Extruded does not appear in the broadcast expression?

The way you've originally written it, there's a type instability.  This means that Broadcast goes through a secondary code path that's more robust to instabilities.

mbauman:
Just as a bit more info here — the problem is that `ArrayAndChar` leans upon the builtin `copy` implementation.  In cases that are type-stable, we can just allocate the output right away up-front — this is where the `similar` (and thus `find_aac`) kicks in.

In cases that are not type-stable, we need to be able to call `similar` many times in the _midst)_ of running through the broadcast.  As an optimization (in both cases), we `preprocess` the broadcasted object to do a bit of work ahead of time.  One part of that optimization is the `Extruded` thing.

https://github.com/JuliaLang/julia/blob/074974ad0b8c47f79f07254a940cd4054bd9549a/base/broadcast.jl#L841-L849

Perhaps we should have defaulted to a no-op `preprocess` for non-default-style `Broadcasted`s.  :-\

fjarri:
Just a thought, perhaps the whole `find_aac()` can be made the default behavior, so that someone who wants an array-like broadcasting doesn't even have to define it? This way the reference to `Extruded` will be hidden.

MilesCranmer:
Could this be merged? I got caught up in the same issue mentioned here:
https://discourse.julialang.org/t/defining-broadcast-for-custom-types-the-example-in-the-docs-fails/32291. I think if the docs had included this example I would have saved a few hours today. I am sure others have ran into this too; the interfaces page is not complete.

MilesCranmer:
Yeah it's just docs. @mbauman can you merge this?

doc/src/manual/interfaces.md
```diff
@@ -565,11 +565,12 @@ end
 find_aac(bc::Base.Broadcast.Broadcasted) = find_aac(bc.args)
 find_aac(args::Tuple) = find_aac(find_aac(args[1]), Base.tail(args))
 find_aac(x) = x
+find_aac(x::Base.Broadcast.Extruded) = x.x
```
mbauman, line 568: Should this recurse?  That is, should it actually be this?  I'm honestly not sure.

```suggestion
find_aac(x::Base.Broadcast.Extruded) = find_aac(x.x)
```

MilesCranmer: Might as well. It looks like I actually did this in my library https://github.com/SymbolicML/DynamicQuantities.jl/blob/9625403f76385603a3acd0fa1119fada3912a5ba/src/arrays.jl#L285

DilumAluthge:

@fjarri Are you still interested in working on this PR?

fjarri:
I think it's better if someone else takes over, I haven't been following Julia for years now, and I'm not sure what kind of things have been happening with broadcasting rules.

PR #34119, opened 2019-12-17
lower `x for x in ...` to `identity` function in generator exprs

JeffBezanson:
`x for x in y if f(x)` can be a convenient way to write filter iterators, but the generator part (`x for`) is unnecessary in this case. We could consider removing the generator entirely, but for now it helps to pass the `identity` function to the generator instead of creating a new `x->x` function. That reduces the number of new function types, which makes compilation more efficient. It also allows some functions specialized on `typeof(identity)` to work. For example it fixes this strange case:

```
julia> sum(identity(x) for x in Int[])
0

julia> sum(x for x in Int[])
```


andyferris (approved):
Yes! I’ve wanted this one before. LGTM.

StefanKarpinski:
Seems unobjectionable. They asked for some function of the form `x->x` and `identity` is one such.

andyferris:
As a general question (beyond this PR), would `Fix1` and `Fix2` be welcome for the same reason of introspectability?

StefanKarpinski:
Seems similar, although a bit harder to detect. I guess if you have
```
f(x, y) for y in v
f(x, y) for x in v
```
then you could replace that with `Fix1` and `Fix2`. I would worry a bit more about people relying on that and it breaking if detection of that pattern doesn't work out for some reason.

PR #34120, opened 2019-12-17
Add scaffolding for notarization on macOS

staticfloat:
This adds some Makefile and `bash` tools to aid in notarizing macOS apps.  None of this is run automatically; the process is still too manual to be done automatically, but this at least makes it very easy for a human to supervise.

### DIFF
--- a/contrib/mac/app/.gitignore
+++ b/contrib/mac/app/.gitignore
@@ -1,3 +1,4 @@
 julia/
 dmg/
 *.dmg
+notarize-*.xml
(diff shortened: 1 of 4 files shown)


contrib/mac/app/.gitignore
```diff
@@ -1,3 +1,4 @@
 julia/
 dmg/
 *.dmg
+notarize-*.xml
```
StefanKarpinski: PSA: unless a git ignore pattern starts with a `/` it applies anywhere beneath the directory containing the git ignore file. If you want to ignore just a specific top-level file with a given name, start the pattern with a `/`.

ViralBShah:
Apple should have at least provided an automated process for this.

PR #33886, opened 2019-11-18
layout optimization internal changes (support pointers inlining/unboxing into parents/codegen) [disabled]

vtjnash:
This includes fairly general support for pointers inside immutable structs appearing in first class aggregates (FCA) throughout the system (e.g. gc, codegen, layout). Currently it's also pretty widely enabled, although we may want to disable it for the initial merge, then slowly enable larger pieces of it as we gain understanding of the effects on the ecosystem.

This replaces https://github.com/JuliaLang/julia/pull/18632. It solves the "fundamental problems" described in that PR with the following rules (these need to be added to the docs and NEWs):

- If a struct declaration is self-referential (aka `references_name`), and thus couldn't qualify for `isbits`, it also doesn't qualify for inlining. This solves the problem with cycles.
- It a struct declaration can be incompletely initialized (n_initialized != n_fields), it doesn't quality for inlining. This solves the problem of tracking #undef. This may be relaxed in the future by adding an extra byte to track this state, although I'm not particularly keen on doing the extra work to support this edge case.
- Note that `Union{T, S}` will not inline ("union-split") if either T or S contains a pointer (wouldn't be legal to codegen this, and somewhat hard to describe to the gc)

I'm interested in folks thoughts on how they'd like to see this land. Should I merge the support for each component in a separate commit (without users until all is done), or just land this as one big commit but disabled? One PR or several? And specific pieces you'd like to see land first?

@nanosoldier `runbenchmarks(ALL, vs=":master")`

PR #34127, opened 2019-12-17
Replace erroneous inbounds function annotation with inline

ararslan:
See https://github.com/JuliaLang/julia/pull/33886#discussion_r358975658.


JeffBezanson:
I have some other small updates to that commit that I'll group with this one.

PR #34128, opened 2019-12-17
some small tweaks to the layout optimization PR

JeffBezanson:
- replace inbounds with inline where intended
- rename jl_justbits


vtjnash (approved):

vtjnash:
Win32 segfault is a bug that was introduced by 987a3d675a7—I'll add a fix here.

Issue #33956, opened 2019-11-26
`realpath` errors for shared network drives on Julia 1.3 on windows


ckingdon95:
I get the following error:
```
julia> realpath("C:/Users")
"C:\\Users"
julia> realpath("M:/")
ERROR: IOError: realpath: permission denied (EACCES)
Stacktrace:
 [1] uv_error at .\libuv.jl:97 [inlined]
 [2] realpath(::String) at .\path.jl:374
 [3] top-level scope at REPL[8]:1
```
This seems like this could be related to a previous issue on 1.2: https://github.com/JuliaLang/julia/issues/33127 

This results in an error when trying to activate an environment on a network drive:
```
pkg> activate "M:/test/"
ERROR: IOError: realpath: permission denied (EACCES)
Stacktrace:
 [1] uv_error at .\libuv.jl:97 [inlined]
 [2] realpath(::String) at .\path.jl:374
 [3] safe_realpath(::String) at D:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.3\Pkg\src\Types.jl:505
 [4] find_project_file(::Nothing) at D:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.3\Pkg\src\Types.jl:248
 [5] Pkg.Types.EnvCache(::Nothing) at D:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.3\Pkg\src\Types.jl:336 (repeats 2 times)
 [6] _activate_dep(::String) at D:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.3\Pkg\src\API.jl:795
 [7] #activate#122(::Bool, ::typeof(Pkg.API.activate), ::String) at D:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.3\Pkg\src\API.jl:817
 [8] activate at D:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.3\Pkg\src\API.jl:809 [inlined]
 [9] do_activate!(::Array{String,1}, ::Dict{Symbol,Any}) at D:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.3\Pkg\src\REPLMode.jl:454
[...]

PR #34132, opened 2019-12-18
Fix calls to `realpath` for shared network drives

corakingdon:
see https://github.com/JuliaLang/julia/issues/33956 for discussion

### DIFF
--- a/base/loading.jl
+++ b/base/loading.jl
@@ -117,7 +117,14 @@ end
 
 const ns_dummy_uuid = UUID("fe0723d6-3a44-4c41-8065-ee0f42c8ceab")
 
-dummy_uuid(project_file::String) = uuid5(ns_dummy_uuid, realpath(project_file))
+function dummy_uuid(project_file::String)
+    try
+        project_path = realpath(project_file)
+    catch
+        project_path = project_file
+    end
+    return uuid5(ns_dummy_uuid, project_path)
+end
 
 ## package path slugs: turning UUID + SHA1 into a pair of 4-byte "slugs" ##
 
--- a/base/loading.jl
+++ b/base/loading.jl
@@ -118,10 +118,10 @@ end
 const ns_dummy_uuid = UUID("fe0723d6-3a44-4c41-8065-ee0f42c8ceab")
 
 function dummy_uuid(project_file::String)
-    try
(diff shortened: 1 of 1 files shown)


StefanKarpinski (approved):
Looks good to me. Let us know when you've tested with the CI-built binary.

fredrikekre:
Answering https://github.com/JuliaLang/julia/issues/33956#issuecomment-567095365 here:

> Thanks @fredrikekre! where on the PR can I find the option to download?

https://s3.amazonaws.com/julialangnightlies/assert_pretesting/winnt/x64/1.4/julia-fe14bfcf42-win64.exe

ckingdon95:
Thank you! for future reference though, where did that link come from?

KristofferC:
Click Details on the `buildbot/tester_win64` (https://build.julialang.org/#/builders/32/builds/6845)

Then look at

> download_url | `"https://s3.amazonaws.com/julialangnightlies/assert_pretesting/winnt/x64/1.4/julia-fe14bfcf42-win64.exe"`

ckingdon95:
@StefanKarpinski in combination with the changes made to Pkg, this seems to work fully on my network drive! I'm wondering though if this is a strange work around, and if it would be better to change the `realpath` definition instead, since it's a julia function? thoughts?https://github.com/JuliaLang/julia/blob/8707744ae5f778c005d6c29342fe01cb654527af/base/path.jl#L409-L434

StefanKarpinski:
It's true that the `realpath` function is kind of problematic, but it's a very low level function and I don't think we want our `realpath` to deviate too much from the system one, which has this behavior. Honestly, in this case we maybe shouldn't be calling `realpath` at all. In any case, improving/changing `realpath` is a much bigger discussion and since this fixes your issue, let's go with it. We can merge the two fixes independently and then once they're both included in master, the issue will be fixed. Another question is whether these are backportable fixes or not. I guess I don't really see why not, but it would be good to get some other opinions on backporting.

StefanKarpinski:
I've approved but, @ckingdon95, I need you to make this no longer a draft before I can merge.

StefanKarpinski:
I've opened https://github.com/JuliaLang/julia/issues/34135 to discuss improving `realpath` or offering a better replacement for it.

PR #34136, opened 2019-12-18
Restore compatibility for extending fld/cld

### LINKED ISSUE #34134: Compatibility for overriding `cld`/`fld`
The implementation on current master (see below) claims to provide compatibility for overriding `cld`/`fld`, but actually doesn’t (at least in the case of [HalfIntegers.jl](https://github.com/sostock/HalfIntegers.jl), which is currently [broken on master](https://travis-ci.com/sostock/HalfIntegers.jl/jobs/267409184#L236) for this reason).

https://github.com/JuliaLang/julia/blob/8707744ae5f778c005d6c29342fe01cb654527af/base/div.jl#L223-L233

HalfIntegers.jl defines `fld(x::T, y::T) where T<:HalfInteger`. Due to the changes on master, this method isn’t called anymore when `fld` is called with two different types of `HalfInteger`s. In order for the above code to fulfill its purpose, the first two and last two methods above could be replaced with
```julia
fld(x::Real, y::Real) = fld(promote(x,y)...)
cld(x::Real, y::Real) = cld(promote(x,y)...)
fld(x::T, y::T) where {T<:Real} = div(x, y, RoundDown)
cld(x::T, y::T) where {T<:Real} = div(x, y, RoundUp)
```

sostock:
Fixes #34134.

### DIFF
--- a/base/div.jl
+++ b/base/div.jl
@@ -223,14 +223,14 @@ rem(a::UInt128, b::UInt128, ::typeof(RoundToZero)) = rem(a, b)
 # These are kept for compatibility with external packages overriding fld/cld.
 # In 2.0, packages should extend div(a,b,r) instead, in which case, these can
 # be removed.
-fld(x::Real, y::Real) = div(promote(x,y)..., RoundDown)
-cld(x::Real, y::Real) = div(promote(x,y)..., RoundUp)
+fld(x::Real, y::Real) = fld(promote(x,y)...)
+cld(x::Real, y::Real) = cld(promote(x,y)...)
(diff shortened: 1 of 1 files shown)


Keno (changes requested):
See comment in the linked issue. With this definition we get a dispatch loop rather than a MethodError due to the reverse fallback elsewhere. Packages running into problems here should additionally provide the three argument `div` method.

sostock:
Ok, if compatibility cannot be achieved across all packages, I will update my package and mark older versions as incompatible with Julia 1.4.

PR #34137, opened 2019-12-18
some fixes for structs with pointers inlined into arrays

JeffBezanson:


vtjnash:
Added the bugfixes from here to #34128. We may want to do the changes here to be more robust against changes to the jl_lineinfonode_type layout, but that should be safe to defer to a later improvement.

PR #33119, opened 2019-08-30
Adds interpolation syntax to `Threads.@spawn`, to evaluate arguments immediately

NHDaly:
Adds $-interpolation syntax to `@async` and `Threads.@spawn`, to evaluate arguments immediately.

Add the ability to evaluate some parts of a `@spawn`/`@async`
immediately, in the current thread context.

This prevents variables being "boxed" in order to capture them in the
closure, exactly the same as wrapping them in a let-block locally.

For example, `$x` expands like this:
```julia
julia> @macroexpand @async $x + 2
quote
    #= task.jl:361 =#
    let var"##454" = x
        #= task.jl:362 =#
        local var"#9#task" = Base.Task((()->begin
                            #= task.jl:358 =#
                            var"##454" + 2
                        end))
        #= task.jl:363 =#
        if $(Expr(:islocal, Symbol("##sync#95")))
            #= task.jl:364 =#
            Base.push!(var"##sync#95", var"#9#task")
        end
        #= task.jl:366 =#
        Base.schedule(var"#9#task")
        #= task.jl:367 =#
        var"#9#task"
    end
end
```

------------------------------------

Fixes #30896

Add the ability to evaluate some parts of a `@spawn` immediately, in the current thread context.

~Adds a new macro, `Threads.@spawncall`, which evaluates its arguments immediately, then `@spawn`s the closure with the argument values.~

~This acts more like a "parallel function call", where the arguments are evaluated immediately, in the current thread, and the function is then invoked in any available thread with the stored argument values.~

EDIT: Updated to add `$`-interpolation syntax support to `@spawn` to allow you to manually specify which parts of the expression to evaluate early.

This prevents variables being "boxed" when capturing them in the closure.

~This mirrors the behavior of golangs `go` command, which requires it be used with a function call.~

Fixes #30896

PR #34142, opened 2019-12-19
Fix race condition in test introduced in #33119 (oops).

### LINKED ISSUE #34141: Race condition in new threading test
As seen in https://build.julialang.org/#/builders/22/builds/7621/steps/2/logs/stdio.

@vtjnash says the test is bad.

NHDaly:
Introduce synchronization (via a `Channel()`) to force spawned tasks to
run after the local variables are updated, showcasing the problem and
the solution with `$`-interpolation.

Fixes race condition in test introduced in #33119.

Fixes https://github.com/JuliaLang/julia/issues/34141.

### DIFF
--- a/test/threads_exec.jl
+++ b/test/threads_exec.jl
@@ -781,19 +781,3 @@ end
     @test fetch(@async :($($a))) == a
     @test fetch(@async "$($a)") == "$a"
 end
-
-# errors inside @threads
-function _atthreads_with_error(a, err)
-    Threads.@threads for i in eachindex(a)
-        if err
-            error("failed")
-        end
-        a[i] = Threads.threadid()
-    end
-    a
-end
-@test_throws TaskFailedException _atthreads_with_error(zeros(nthreads()), true)
-let a = zeros(nthreads())
-    _atthreads_with_error(a, false)
-    @test a == [1:nthreads();]
-end
--- a/test/threads_exec.jl
+++ b/test/threads_exec.jl
@@ -705,9 +705,10 @@ catch ex
 end
 
 @testset "@spawn interpolation" begin
-    # Issue #30896: evaluating argumentss immediately
+    # Issue #30896: evaluating arguments immediately
     begin
         outs = zeros(5)
+        # Use interpolation to fill outs with the values of `i`
         @sync begin
             local i = 1
             while i <= 5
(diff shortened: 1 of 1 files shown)


test/threads_exec.jl
```diff
@@ -718,40 +719,50 @@ end
...
+            a = 2  # update `a` after spawning, before `t` runs
+            put!(ch, nothing)  # now run t
+            @test fetch(t) == Expr(:call, :+, 1, 2, :a)
+        end
+
+        # Test the difference between different levels of interpolation
+        # Without interpolation, each spawned task sees the last value of `i` (6);
+        # with interpolation, each spawned task has the value of `i` at time of `@spawn`.
+        let
+           oneinterp  = Vector{Any}(undef, 5)
```
ararslan: Should be 4 spaces instead of 3 for this block

test/threads_exec.jl
```diff
@@ -770,19 +781,3 @@ end
     @test fetch(@async :($($a))) == a
     @test fetch(@async "$($a)") == "$a"
 end
-
-# errors inside @threads
```
JeffBezanson: These probably should not be deleted?

vtjnash (approved):
Looks right. Was there a merge conflict at the bottom though? Deleting that other test doesn't seem related.

NHDaly:
Ah sorry, the extra test at the bottom was incorrectly deleted via a bad
merge conflict when I was rebasing and squashing for #33119. (Shouldn't be
working remotely from this tiny laptop.)

This PR removes the duplicated tests.

Sorry, should have called that out in the description.

vchuravy:
Merged locally as 608567f8566363536042e7f73d84b8670fc910d4 to fixup the indent and squash the PR.

KristofferC:
Did that fix the wrongly deleted test?

NHDaly:
Thanks a bunch, @vchuravy! <3

@KristofferC, no, sorry, it wasn't wrongly deleted. It was correctly deleted, because #33119 wrongly _duplicated_ it due to a bad merge. Added here:
https://github.com/JuliaLang/julia/pull/33119/files#diff-d94a81ec91d1b6e2ba2cdb900b30cb3dR774
Duplicated from here:
https://github.com/JuliaLang/julia/pull/33119/files#diff-d94a81ec91d1b6e2ba2cdb900b30cb3dR684

vtjnash:
I appreciate the parallelism of how this was fixing two race conditions.

NHDaly:
😂

PR #34148, opened 2019-12-19
[WIP]  Trying to fix spawn-interpolation for macrocalls

NHDaly:
Expand macrocalls when parsing async early, before recursing to
interpolate `$` variables. This fixes inner macrocalls that themselves
operate on `$`, by giving them the chance to use up the `$` _if they
use them_, before allowing `@async` / `@spawn` to do the intepolation
themselves.

But this still doesn't seem to be working correctly, since "outer
macrocalls" still aren't working, e.g.:
```julia
julia> @eval fetch(@async 2+$$x)
ERROR: syntax: "$" expression outside quote
```

### DIFF
--- a/base/task.jl
+++ b/base/task.jl
@@ -353,7 +353,7 @@ isolating the aysnchronous code from changes to the variable's value in the curr
     Interpolating values via `\$` is available as of Julia 1.4.
 """
 macro async(expr)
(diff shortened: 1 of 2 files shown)


JeffBezanson:
I'll put up a simpler fix.

PR #34149, opened 2019-12-19
make @timed return NamedTuple (# 34147)

### LINKED ISSUE #34147: Return named tuple from `@timed`
Can we return a named tuple from `@timed` macro?
```julia
julia> value, time, bytes, gctime, gcdiff = @timed rand(10^6);
```
is error prone and hard to remember.

jw3126:
fix #34147


base/util.jl
```diff
@@ -276,21 +276,21 @@ See also [`@time`](@ref), [`@timev`](@ref), [`@elapsed`](@ref), and
 [`@allocated`](@ref).
 
 ```julia-repl
-julia> val, t, bytes, gctime, memallocs = @timed rand(10^6);
+julia> o = @timed rand(10^6);
```
StefanKarpinski: Maybe call this `t` instead? Calling it `o` is a bit odd.

base/util.jl
```diff
@@ -276,21 +276,21 @@ See also [`@time`](@ref), [`@timev`](@ref), [`@elapsed`](@ref), and
...
-julia> bytes
+julia> o.bytes
 8000256
 
-julia> gctime
+julia> o.gctime
 0.0055765
 
-julia> fieldnames(typeof(memallocs))
+julia> fieldnames(typeof(o.gcdiff))
```
JeffBezanson: How about `gcstats` as the name of this field?

StefanKarpinski: Right, the rainfall about using a named tuple is that now the names are significant.

jw3126:
I think the CI fail is unrelated.

StefanKarpinski:
Would it be sacrelidge to name these `gc_time` and `gc_stats`?

JeffBezanson:
Yes :stuck_out_tongue:

jw3126:
Can somebody rerun CI?

jw3126:
Merge?

StefanKarpinski:
Seems good but I think everyone is taking a break until the new year.

base/util.jl
```diff
@@ -276,23 +276,26 @@ See also [`@time`](@ref), [`@timev`](@ref), [`@elapsed`](@ref), and
...
-julia> fieldnames(typeof(memallocs))
+julia> propertynames(stats.gcstats)
 (:allocd, :malloc, :realloc, :poolalloc, :bigalloc, :freecall, :total_time, :pause, :full_sweep)
 
-julia> memallocs.total_time
+julia> stats.gcstats.total_time
 5576500
 ```
+
+!!! compat "Julia 1.4"
```
JeffBezanson:
```suggestion
!!! compat "Julia 1.5"
```

base/util.jl
```diff
@@ -276,23 +276,26 @@ See also [`@time`](@ref), [`@timev`](@ref), [`@elapsed`](@ref), and
...
+    The return type of this macro was changed from `Tuple` to `NamedTuple` in Julia 1.4
```
JeffBezanson:
```suggestion
    The return type of this macro was changed from `Tuple` to `NamedTuple` in Julia 1.5.
```

JeffBezanson (approved):

PR #34154, opened 2019-12-19
fix #34138, interpolation in async/spawn should ignore macros

### LINKED ISSUE #34138: Interpolation syntax in Threads.@spawn breaks $-aware macros
It seems interpolation syntax for `@spawn` #33119 is a breaking change:

```julia
julia> x = [reshape(1:4, 2, 2);]
2×2 Array{Int64,2}:
 1  3
 2  4

julia> @. exp(x)
2×2 Array{Float64,2}:
 2.71828  20.0855
 7.38906  54.5982

julia> @. $exp(x)
2×2 Array{Float64,2}:
 51.969   112.105
 74.7366  164.074

julia> fetch(Threads.@spawn @. $exp(x))
2×2 Array{Float64,2}:
 2.71828  20.0855
 7.38906  54.5982

julia> VERSION
v"1.4.0-DEV.634"
```

In Julia 1.3

```julia
julia> fetch(Threads.@spawn @. $exp(x))
2×2 Array{Float64,2}:
 51.969   112.105
 74.7366  164.074
```

But maybe it's OK to have breaking changes as `@spawn` is explicitly marked as experimental?

Stopping recursing into `macrocall` in the AST manipulation of `@spawn` can "fix" this problem.  However, it's a bit limiting approach if you consider something like `@spawn h(@spawn g($(f(x))))`.

One approach may be to repeat `$` to "escape" `$`; e.g., `@spawn @. $$exp(x)` if you want to do `@. $exp(x)` in another task. (This is of course breaking.)

JeffBezanson:
fix #34138

### DIFF
--- a/base/task.jl
+++ b/base/task.jl
@@ -377,7 +377,7 @@ function _lift_one_interp!(e)
 end
 _lift_one_interp_helper(v, _, _) = v
 function _lift_one_interp_helper(expr::Expr, in_quote_context, letargs)
-    if expr.head == :$
+    if expr.head === :$
         if in_quote_context  # This $ is simply interpolating out of the quote
             # Now, we're out of the quote, so any _further_ $ is ours.
             in_quote_context = false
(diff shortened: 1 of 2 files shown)


NHDaly:
Thanks for the PR, Jeff. Too bad about having to simplify it, but I think it makes sense, yeah.

Can you please also add a test for outer macrocalls too, like `x = 2;  @test @eval(fetch(@async 2+$x)) == 4`, like I had here:
https://github.com/JuliaLang/julia/pull/34148/files#diff-d94a81ec91d1b6e2ba2cdb900b30cb3dR796

That should work correctly now, too, but it's good to test that as well. :)

--------------

Sorry your build is failing; its because of the nondeterminism in the tests I added (boy, I really botched this PR; teach me not to have code merged in while on vacation). That should be fixed now if you merge in the latest master to get the fix from here:
https://github.com/JuliaLang/julia/pull/34142

Thanks!

NHDaly (approved):

tkf:
Does it make sense to include something like

```julia
t1, t2 = fetch(@span ((@span $(current_task())), current_task()))
@test t1 === t2
```

?

PR #34161, opened 2019-12-20
Decrease build time by 15 minutes

JesterOrNot:
Decrease build time by 15 minutes Simplify Contributor Onloading and bring vscode like editing in browser with Gitpod
image
try it out
https://gitpod.io/#https://github.com/JesterOrNot/julia

### DIFF
--- a/.gitpod.Dockerfile
+++ b/.gitpod.Dockerfile
@@ -1,0 +1,12 @@
+FROM gitpod/workspace-full
+
+USER gitpod
+
+# Install custom tools, runtime, etc. using apt-get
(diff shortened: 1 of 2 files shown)


KristofferC:
Could you explain what the intention of this PR is? What build time is decreased? How is this used? Why should these files be in the julia repo?

JesterOrNot:
@KristofferC for building the source code this simplifies contributor onloading by installing the dependencies for julia and providing an editor so people can contribute to Julia

ViralBShah:
I think this can be provided in a separate repo perhaps in the JuliaEditorSupport org. It shouldn't be part of the Julia project files here.

PR #31464, opened 2019-03-24
Accept both Cvoid and Ptr{Void} return type in ccall(:memcpy)

Keno:
POSIX specifies that memcpy returns its first argument, which is
a bit of a useless feature (e.g. the llvm intrinsic just returns
`void`). Nevertheless, since we're intercepting the ccall here, we
should allow it. For convenience, still allow the Cvoid return though.

Fixes #31073

PR #34165, opened 2019-12-20
Correct signature of memmove

Keno:
wasm was complaining about the incorrect Cvoid rather than Ptr{Void}
in the :memmove calls in ryu, but while we're here also change the
third argument to Csize_t, since that's what the standard says it is.

### DIFF
--- a/base/array.jl
+++ b/base/array.jl
@@ -244,7 +244,7 @@ segfault your program, in the same manner as C.
 function unsafe_copyto!(dest::Ptr{T}, src::Ptr{T}, n) where T
     # Do not use this to copy data between pointer arrays.
     # It can't be made safe no matter how carefully you checked.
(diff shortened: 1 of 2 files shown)


vtjnash:
Wasn't the decision in https://github.com/JuliaLang/julia/pull/31464 to commit to detecting this pair of functions as a special case in the backend and make this signature mismatch valid?

quinnj (approved):

Keno:
> Wasn't the decision in #31464 to commit to detecting this pair of functions as a special case in the backend and make this signature mismatch valid?

We certainly decided that for `memcpy` which we explicitly intercept in ccall.cpp. I'm fine with extending that to `memmove` also, though of course that would mean adding an appropriate special case in ccall.cpp.

PR #34174, opened 2019-12-21
Fix reference to concurrency_violation

Keno:
This file is in `Threads`, but `concurrency_violation` is in `Base`.

cc @NHDaly


Keno:
Side note: Can we have a test that exercises this error path without breaking the system?

vtjnash:
I thought we did—it's just `@test_throws ErrorException("concurrency violation detected") unlock(SpinLock())`

JeffBezanson:
`unlock(::SpinLock)` doesn't do any checks currently. It needs to be really fast. Would checking `islocked` be basically 0 cost?

Issue #34115, opened 2019-12-16
possible wrong use of time()


nico202:
https://github.com/JuliaLang/julia/blob/ba4f858cd0618e209b81a1800005662769cdc1c1/stdlib/Distributed/src/cluster.jl#L1027

Hi! I was trying to add support to [SOURCE_DATE_EPOCH](https://reproducible-builds.org/specs/source-date-epoch/). I patched the file /src/support/timefuncs.c, compiled and everything went well. All tests were successful, but they never ended. I saw in julia src many uses of time() like the one above.

However, time() is a wrapper for `gettimeofday`. According to the man page
> The time returned by gettimeofday() is affected by discontinuous jumps in the system time (e.g., if the system administrator manually changes the system time). If you need a monotonically increasing clock, see clock_gettime(2))

there are jumps. Shouldn't julia use in those situations `time_ns()` (wrapper for libuv `uv_hrtime` that has [no drift](http://docs.libuv.org/en/v1.x/misc.html#c.uv_hrtime)?)

@StefanKarpinski this follows the other time-related error #34056 

Thanks, Nicolò

Keno:
Yep, I think you're correct.

nico202:
Couple more (wrong) uses I could find, but sure there are more:
https://github.com/JuliaLang/julia/blob/220681f21fc5aaf83b7c3aee0a2b8d6186805b2c/contrib/generate_precompile.jl#L182
https://github.com/JuliaLang/julia/blob/4f19dc27cc3290f4044e2a50190c5ffec920b0b8/stdlib/Profile/test/runtests.jl#L21
https://github.com/JuliaLang/julia/blob/b580e34b10d01a245ccab88a2d4fa87906540e00/stdlib/Sockets/test/runtests.jl#L386

nico202:
A bit OT, but can somebody with a better understanding of the code caching help me find how julia adds the timestamp to the cache file? I found the place where it is _read_ here:
https://github.com/JuliaLang/julia/blob/3608c84e6093594fe86923339fc315231492484c/base/loading.jl#L1292

But I think it is stored in some C code, right? I suppose it's here
https://github.com/JuliaLang/julia/blob/1e39c69a550c9be945db2ccf1ab6ded02c404ec4/src/dump.c#L1336
But I have difficulties finding back where it is read. It does not seems to be a `time()` call, I'm a bit lost

JeffBezanson:
[...]

PR #34181, opened 2019-12-22
Fix the use of time() in measuring performances (replace with time_ns())

nico202:
time() is a wrapper for gettimeofday, which is affected by by discontinuous
jumps in the system time (e.g., if the system administrator manually changes the
system time), see `man gettimeofday`.  On the other hand, time_ns() is a wrapper
for libuv uv_hrtime that has no drift, whose manual page says "The primary use
is for measuring performance between intervals.".

Tentative fix for bug #34115

### DIFF
--- a/stdlib/Distributed/src/cluster.jl
+++ b/stdlib/Distributed/src/cluster.jl
@@ -293,12 +293,11 @@ Base.showerror(io::IO, e::LaunchWorkerError) = print(io, e.msg)
 # The master process uses this to connect to the worker and subsequently
 # setup a all-to-all network.
 function read_worker_host_port(io::IO)
-    t0 = time()
+    t0 = time_ns()
 
     # Wait at most for JULIA_WORKER_TIMEOUT seconds to read host:port
     # info from the worker
-    timeout = worker_timeout()
-
+    timeout = worker_timeout() * 1e9
     # We expect the first line to contain the host:port string. However, as
     # the worker may be launched via ssh or a cluster manager like SLURM,
     # ignore any informational / warning lines printed by the launch command.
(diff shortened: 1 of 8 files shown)


nico202:
I'm not on a computer where I can ran tests right now, and I possibly missed some use

test/file.jl
```diff
@@ -382,7 +382,7 @@ if Sys.iswindows()
 else
     @test filesize(dir) > 0
 end
-nowtime = time()
+nowtime = time_ns()
```
JeffBezanson: This doesn't look correct, since this needs to be in the same units as file `mtime`.

JeffBezanson:
Thanks, I agree we should change most of these.

JeffBezanson (approved):

nico202:
Thanks! Tests are passing (I'm 90% into it, should end in a few minutes). I'll rebase it

nico202:
Here we go :)

nico202:
Is windows failure my fault?

stdlib/Distributed/src/cluster.jl
```diff
@@ -1023,10 +1022,10 @@ function _rmprocs(pids, waitfor)
             end
         end
 
-        start = time()
-        while (time() - start) < waitfor
+        start = time_ns()
+        while (time_ns() - start) < waitfor
```
JeffBezanson:
```suggestion
        while (time_ns() - start) < waitfor*1e9
```

nico202:
If everything is fine I can squash again. Arm 64 error seems unrelated, right?

KristofferC:
The commits can be squashed on merge so no need to do it manually.

nico202:
@KristofferC fine! Didn't know, thanks

PR #34182, opened 2019-12-22
RFC: enable completions for identifiers following ! operator

aviatesk:
### Rationale

Currently `!hoge|` doesn't suggestion any completion, but this case is better to be completed as with `hoge|`. (where `|` represents a current position)
Those cases needs to be special-cased, since `!` can be a part of a valid identifier name but atst the parser separates the left-most `!` from the preceding identifier.
So I implemented the corresponding logic in `REPLCompletions` module.

### Results

REPL now completes e.g. `!isnothi` to `!isnothing`.
Tests are also added.

### DIFF
--- a/stdlib/REPL/src/REPLCompletions.jl
+++ b/stdlib/REPL/src/REPLCompletions.jl
@@ -612,7 +612,9 @@ function completions(string, pos, context_module=Main)::Completions
     inc_tag==:string && return String[], 0:-1, false
     if inc_tag === :other && should_method_complete(partial)
         frange, method_name_end = find_start_brace(partial)
-        ex = Meta.parse(partial[frange] * ")", raise=false, depwarn=false)
+        # strip preceding ! operator
+        s = replace(partial[frange], r"\!+([^=\(]+)" => s"\1")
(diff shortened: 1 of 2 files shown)


stdlib/REPL/Project.toml
```diff
@@ -7,8 +7,8 @@ Markdown = "d6f4376e-aef5-505a-96c1-9c027394607a"
 Sockets = "6462fe0b-24de-5631-8697-dd941f90decc"
 
 [extras]
-Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
```
JeffBezanson: Probably best not to include this?

JeffBezanson:
Nice, this seems very useful.

stdlib/REPL/Project.toml
aviatesk: well, I included this because `(REPL) pkg> resolve` automatically made this change, but I'm totally okay not to include this :)
Maybe this should be done in a separate PR or somewhere.

PR #34191, opened 2019-12-24
in(p::Pair, a::AbstractDict, valcmp) bug sort.  swapped valcmp arguments

questors:
AsbstractDict 'in'  valcmp arguments are swapped compared to the ImmutableDict version.

There are tests for ImmutableDict, but not AbstractDict.    PR makes Abstract Dict behave the same as ImmutableDict  and adds same tests to verify

### DIFF
--- a/base/abstractdict.jl
+++ b/base/abstractdict.jl
@@ -19,7 +19,7 @@ haskey(d::AbstractDict, k) = in(k, keys(d))
 function in(p::Pair, a::AbstractDict, valcmp=(==))
     v = get(a,p[1],secret_table_token)
     if v !== secret_table_token
-        return valcmp(v, p[2])
+        return valcmp(p[2], v)
     end
     return false
 end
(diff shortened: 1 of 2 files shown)


Keno:
Is the argument order here documented anywhere? If not, it probably should be.

JeffBezanson:
I suspect we only intended this to be an equality comparison. If we want to expand it, the argument order should probably be equivalent to passing `==(p[2])` as a predicate; in other words `valcmp(v, p[2])`, changing ImmutableDict instead of AbstractDict.

StefanKarpinski:
If I'm reading @JeffBezanson's comment right, we should not change this comparison order.

JeffBezanson:
I think we should just remove the `valcmp` argument. It's a bit strange, and only used for implementing `==` and `isequal` on whole Dicts, so I think it should be implemented entirely in those functions instead of being passed to `in`.

StefanKarpinski:
More generally, the `valcmp` option is super dubious: instead one should just write `d[key] == val` and use whatever comparison function one wants. It's entirely unclear to me why this API was added in the first place. Some attempt to avoid multiple key lookups? Would be better addressed by an official token API.

PR #34194, opened 2019-12-25
Update arrays.md and performance-tips.md

yuxi-liu-wired:
some stylistic changes

### DIFF
--- a/doc/src/manual/performance-tips.md
+++ b/doc/src/manual/performance-tips.md
@@ -1443,7 +1443,7 @@ Interpreting the output of [`@code_warntype`](@ref), like that of its cousins [`
 Your code is being presented in form that has been heavily digested on its way to generating
 compiled machine code. Most of the expressions are annotated by a type, indicated by the `::T`
 (where `T` might be [`Float64`](@ref), for example). The most important characteristic of [`@code_warntype`](@ref)
-is that non-concrete types are displayed in red; since this document is written in Markdown, which has no color, 
+is that non-concrete types are displayed in red; since this document is written in Markdown, which has no color,
 in this document, red text is denoted by uppercase.
 
 At the top, the inferred return type of the function is shown as `Body::Float64`.
--- a/doc/src/manual/performance-tips.md
+++ b/doc/src/manual/performance-tips.md
@@ -611,8 +611,8 @@ number (the result of [`/`](@ref) operator). This makes it more difficult for th
 optimize the body of the loop. There are several possible fixes:
 
   * Initialize `x` with `x = 1.0`
-  * Declare the type of `x`: `x::Float64 = 1`
-  * Use an explicit conversion: `x = oneunit(Float64)`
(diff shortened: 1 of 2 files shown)


doc/src/manual/performance-tips.md
```diff
@@ -1428,7 +1428,7 @@ julia> @code_warntype f(3.2)
 Variables
   #self#::Core.Compiler.Const(f, false)
   x::Float64
-  y::Union{Float64, Int64}
+  y::UNION{FLOAT64, INT64}
```
vchuravy: I guess that by turning this all uppercase you wanted to emulate the fact that in the terminal this is printed red?
Personally I would find it more surprising to see the different capitalisation. Maybe we could actually add the color here?

doc/src/manual/performance-tips.md
```diff
@@ -1443,8 +1443,8 @@ Interpreting the output of [`@code_warntype`](@ref), like that of its cousins [`
 Your code is being presented in form that has been heavily digested on its way to generating
 compiled machine code. Most of the expressions are annotated by a type, indicated by the `::T`
 (where `T` might be [`Float64`](@ref), for example). The most important characteristic of [`@code_warntype`](@ref)
-is that non-concrete types are displayed in red; in the above example, such output is shown in
-uppercase.
+is that non-concrete types are displayed in red; since this document is written in Markdown, which has no color, 
+in this document, red text is denoted by uppercase.
```
vchuravy: Ah I saw this to late.

doc/src/manual/performance-tips.md
Keno: I think we also do apply this effect in the REPL if color is not supported.

Keno: Of course, if we could render the color in the docs that'd be even better.

KristofferC: > I think we also do apply this effect in the REPL if color is not supported.

Yes

yuxiliu1995: A bit later in the doc, there was capitalization, so I made the doc consistent in this regard.

doc/src/manual/performance-tips.md
```diff
@@ -1443,8 +1443,8 @@ Interpreting the output of [`@code_warntype`](@ref), like that of its cousins [`
 Your code is being presented in form that has been heavily digested on its way to generating
 compiled machine code. Most of the expressions are annotated by a type, indicated by the `::T`
 (where `T` might be [`Float64`](@ref), for example). The most important characteristic of [`@code_warntype`](@ref)
-is that non-concrete types are displayed in red; in the above example, such output is shown in
-uppercase.
+is that non-concrete types are displayed in red; since this document is written in Markdown, which has no color,
```
mbauman: This is the trailing whitespace that's causing the red ❌.

```suggestion
is that non-concrete types are displayed in red; since this document is written in Markdown, which has no color,
```

Issue #33248, opened 2019-09-13
Julep: Structured Concurrency


c42f:
In julia 1.3 the threadsafe runtime for truly parallel tasks [has arrived](https://julialang.org/blog/2019/07/multithreading) which will greatly increase the interest in concurrent computation in Julia's core numerical and technical computing community. For the moment, `Threads.@spawn` is experimental, but now is a good time to think about APIs for expressing concurrent computation in a safe and composable way.

There's been a lot of recent excitement and hard work on libraries to support so-called *structured concurrency*, with the claim that this is fundamentally a better model for composable concurrent computation.

I'm convinced that people are onto something here and that the overall ideas are an excellent fit for the Julia language. However, there's a lot of prior art to absorb so I have started a document with the aim to do two things:
* Survey existing ideas and implementations
* Explore how these ideas fit within the Julia language and runtime.

Here is a link to WIP version of [**the document**](https://github.com/c42f/Juleps/blob/cjf/structured-concurrency/StructuredConcurrency.md).

Let's try discussing this here, (non-substantive procedural comments can go to https://github.com/JuliaLang/Juleps/pull/56).

tkf:
I'm copying some comments I made in https://github.com/c42f/Juleps

### Discuss "black box rule" and robust cancellation separately?

I think it might be better to emphasize that structured concurrency has a few sub-concepts.

My understanding is that what njs calls [_"black box rule"_](https://vorpus.org/blog/notes-on-structured-concurrency-or-go-statement-considered-harmful/#what-happened-to-goto) is a sub-concept of structured concurrency and it's more fundamental than robust cancellation.  So, I'd suggest to introduce "black box rule" and cancellation as different sub-concepts from the get-go.  I think this is particularly important in Julia because many compute-intensive task don't need cancellation and people may not want to have runtime and conceptual overhead of robust cancellation.  An alternative term may be [_call tree_](http://250bpm.com/blog:124).

One counter-example to my argument that "black box rule" is more fundamental is Go's `errgroup` and `context`.  `errgroup` is the implementation of the "black box rule" while `context` is the implementation of robust cancellation.  `errgroup` is implemented on top of `context` and it may contradict with my argument that "black box rule" (`errgroup`) is more fundamental than robust cancellation (`errgroup`).  But I think the fact that they are implemented in separate packages in Go at least backs up my argument that those are different concepts.

So, maybe it is a good idea to organize the sections like this?

* Structured Concurrency
    * Background terminology
[...]

PR #34198, opened 2019-12-25
A version of the sync macro that throws earlier

Keno:
I've been looking at what causes deadlocks in our test suite in an effort to cut down on the number of failed tests on CI that result in hangs (since those are hard to diagnose and resolve). I found that by playing with various resource limits, it is easy to create hangs in the test suite. The reason we get a hang rather than a more easily diagnosable error is two fold. We either:
1) Aren't watching for the error (e.g. a socket remote end closing)
2) We aren't propagating the error to the top level

A very common situation for case 2) is that the test is wrapped in `@sync` which doesn't return until all tasks have finished or error'ed. However, in many cases one of the tasks produces data for the others, so if that task errors, the remaining tasks will wait forever. This PR aims to address that situation by introducing a new `@syncany` macro that immediately rethrows any errors thrown by a contained task rather than waiting for all of them to finish. The implementation isn't super performant (it allocates a new task per object being waited on), but should be sufficient for use in the test suite. A better implementation would create a new scheduler object that can be inserted into multiple wait queues.

Example usage of the new macro:

```
@sync begin
    @async error("Hello")
    @async sleep(1000)
end # Waits 1000s

@syncany begin
    @async error("Hello")
    @async sleep(1000)
end # Throws immediately
```

The macro doesn't do any sort of cleanup for the tasks that do not finish, and just lets them run. In the future, we may want to automatically cancel those tasks, but that seemed like a bigger design problem than the simple thing that I wanted (something that propagates error messages more readily, so we see them in the logs).

### DIFF
--- a/base/task.jl
+++ b/base/task.jl
@@ -318,32 +318,6 @@ function sync_end(refs)
     nothing
 end
 
-function syncany_end(refs)
-    local c_ex
-    defined = false
-    t = current_task()
-    cond = Threads.Condition()
-    lock(cond)
-    nremaining = length(refs)
-    for r in refs
-        schedule(Task(()->begin
-            try
-                wait(r)
-                lock(cond)
-                nremaining -= 1
-                nremaining == 0 && notify(cond)
-                unlock(cond)
-            catch e
-                lock(cond)
-                notify(cond, e; error=true)
-                unlock(cond)
-            end
-        end))
-    end
-    wait(cond)
-    unlock(cond)
-end
-
 const sync_varname = gensym(:sync)
 
 """
@@ -359,28 +333,6 @@ macro sync(block)
         let $var = Any[]
             v = $(esc(block))
             sync_end($var)
-            v
-        end
-    end
-end
-
-"""
-    @syncany
-
-Wait until all lexically-enclosed uses of `@async`, `@spawn`, `@spawnat` and `@distributed`
-are complete, or at least one of them has errored. The first exception is immediately
-rethrown. It is the responsibility of the user to cancel any still-running operations
-during error handling.
-
-!!! Note
-    This interface is experimental and subject to change or removal without notice.
-"""
-macro syncany(block)
-    var = esc(sync_varname)
-    quote
-        let $var = Any[]
-            v = $(esc(block))
-            syncany_end($var)
             v
         end
     end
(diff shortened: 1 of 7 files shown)


jonas-schulze:
Couldn’t this be the new default behavior?

base/task.jl
```diff
@@ -338,6 +364,28 @@ macro sync(block)
...
+"""
+    @syncany
+
+Wait until all lexically-enclosed uses of `@async`, `@spawn`, `@spawnat` and `@distributed`
+are complete, or at least one of them has errored. The first exception is immediately
+rethrown. It is the responsibility of the user to cancel any still-running operations
+during error handling.
+
+!!! Note
+    This interface is experimental and subject to change or removal without notice.
```
tkf: :+1: for marking this experimental. Otherwise, implementing structured concurrency #33248 becomes much harder.

vtjnash:
Maybe we should call this `Base.Future.@async` (with the idea that once optimized, it should be the default)? The `any` name seems odd (because it is waiting for `all`, it’s just early error).

Keno:
You mean `Base.Future.@sync`? I'd be fine with that.

tkf:
Isn't it strange to put an experimental API in `Future`?  I think it makes harder to fix the bigger design problem.

c42f:
Letting tasks drop out the bottom seems like a pretty big change to the semantics of `@sync` so I'm not sure it makes sense to declare it the `Future` without trying it out for a while. I'd also be happier not to export it for the moment if it's experimental.

As you say, it would ideally interact with some kind of cancellation. That's hard but it feels a lot more consistent with the way `@sync` works right now.

It seems handy though and could be a big practical improvement for the test suite. For the moment, could we just name it clearly and not export it?  How about `@sync_failfast` or something equally ugly but explicit?

Keno:
From triage: let's call this `Base.Experimental.@sync`

tkf:
`Base.Experimental.@sync` doesn't sound like a stop-gap measure. Shouldn't the macro name (`@sync`) clarify that:

* this macro has undesirable property (tasks leak),
* it is included temporary only because a real solution is impossible to implement with the current infrastructure, and
* it will be removed or at least will be discouraged to use in the future?

Keno:
Let's merge this and see if it helps CI. Being in experimental makes it explicitly unsupported, so we can change or remove it if we feel like it at any point.

PR #34199, opened 2019-12-26
Allow passing NTuple{N,T} as Ref{T}/Ptr{T} to ccall

simonbyrne:
This makes it convenient when working with APIs that use small arrays.

Based on discussion on Slack with @vtjnash and [Discourse comment](https://discourse.julialang.org/t/ccall-with-pointer-argument-to-array-of-custom-type-via-an-ntuple-of-immutables/1048/2) by @yuyichao

### DIFF
--- a/base/refpointer.jl
+++ b/base/refpointer.jl
@@ -126,6 +126,16 @@ cconvert(::Type{Ref{P}}, a::Array{<:Ptr}) where {P<:Ptr} = a
 cconvert(::Type{Ptr{P}}, a::Array) where {P<:Union{Ptr,Cwstring,Cstring}} = Ref{P}(a)
 cconvert(::Type{Ref{P}}, a::Array) where {P<:Union{Ptr,Cwstring,Cstring}} = Ref{P}(a)
 
+# pass NTuple{N,T} as Ptr{T}/Ref{T}
+cconvert(::Type{Ref{T}}, t::NTuple{N,T}) where {N,T} = Ref{NTuple{N,T}}(t)
+cconvert(::Type{Ref{T}}, r::Ref{NTuple{N,T}}) where {N,T} = r
+unsafe_convert(::Type{Ref{T}}, r::Ref{NTuple{N,T}}) where {N,T} =
+    convert(Ptr{T}, unsafe_convert(Ptr{NTuple{N,T}}, r))
+unsafe_convert(::Type{Ptr{T}}, r::Ref{NTuple{N,T}}) where {N,T} =
+    convert(Ptr{T}, unsafe_convert(Ptr{NTuple{N,T}}, r))
+unsafe_convert(::Type{Ptr{T}}, r::Ptr{NTuple{N,T}}) where {N,T} =
+    convert(Ptr{T}, r)
+
 ###
 
 getindex(b::RefArray) = b.x[b.i]
--- a/NEWS.md
+++ b/NEWS.md
@@ -29,6 +29,8 @@ New library features
 --------------------
 * Function composition now works also on one argument `∘(f) = f` (#34251)
 
+* `Ref{NTuple{N,T}}` can be passed to `Ptr{T}`/`Ref{T}` `ccall` signatures ([#34199])
+
 
 Standard library changes
 ------------------------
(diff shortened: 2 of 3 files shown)


base/refpointer.jl
```diff
@@ -126,6 +126,17 @@ cconvert(::Type{Ref{P}}, a::Array{<:Ptr}) where {P<:Ptr} = a
 cconvert(::Type{Ptr{P}}, a::Array) where {P<:Union{Ptr,Cwstring,Cstring}} = Ref{P}(a)
 cconvert(::Type{Ref{P}}, a::Array) where {P<:Union{Ptr,Cwstring,Cstring}} = Ref{P}(a)
 
+
+cconvert(::Type{Ref{T}}, t::NTuple{N,T}) where {N,T} = Ref{NTuple{N,T}}(t)
+cconvert(::Type{Ptr{T}}, t::NTuple{N,T}) where {N,T} = Ref{NTuple{N,T}}(t)
```
vtjnash:
```suggestion
```

vtjnash (commented):
This looks nice—a bit like how we handle Array objects.

base/refpointer.jl
vtjnash: I’m not sure we want this conversion, if it seems to make sense semantically to convert a Tuple to a Ptr

simonbyrne: Fair enough, can remove that.

JeffBezanson (approved):

base/refpointer.jl
JeffBezanson: I guess it makes as much sense as converting an Array to a Ptr?

yuyichao: Not really since the `Tuple` doesn't own the memory.

I would be fine to allow converting `Ref` of `NTuple` to `Ptr` of the `eltype` if it is not already allowed.

vtjnash: Yeah, it’s not fundamentally invalid to define this, but I think Ref is just generally closer to the right concept so there’s no reason to also define this. IIRC, cconvert to Ptr is generically defined as the identity function (intentionally letting it fall through to the unsafe_convert rules) so the latter case Yichao mentioned may already work. Would be good to see if it does and add tests for it.

simonbyrne:
NEWS item added (First Post!)

base/refpointer.jl
```diff
@@ -126,6 +126,16 @@ cconvert(::Type{Ref{P}}, a::Array{<:Ptr}) where {P<:Ptr} = a
...
 cconvert(::Type{Ref{P}}, a::Array) where {P<:Union{Ptr,Cwstring,Cstring}} = Ref{P}(a)
 
+# pass NTuple{N,T} as Ptr{T}/Ref{T}
+cconvert(::Type{Ref{T}}, t::NTuple{N,T}) where {N,T} = Ref{NTuple{N,T}}(t)
+cconvert(::Type{Ref{T}}, r::Ref{NTuple{N,T}}) where {N,T} = r
+unsafe_convert(::Type{Ref{T}}, r::Ref{NTuple{N,T}}) where {N,T} =
+    convert(Ptr{T}, unsafe_convert(Ptr{NTuple{N,T}}, r))
+unsafe_convert(::Type{Ptr{T}}, r::Ref{NTuple{N,T}}) where {N,T} =
+    convert(Ptr{T}, unsafe_convert(Ptr{NTuple{N,T}}, r))
+unsafe_convert(::Type{Ptr{T}}, r::Ptr{NTuple{N,T}}) where {N,T} =
```
JeffBezanson: I believe this method is redundant? (There is already an unsafe_convert method to go between pointer types.)

base/refpointer.jl
```diff
@@ -133,8 +133,6 @@ unsafe_convert(::Type{Ref{T}}, r::Ref{NTuple{N,T}}) where {N,T} =
     convert(Ptr{T}, unsafe_convert(Ptr{NTuple{N,T}}, r))
 unsafe_convert(::Type{Ptr{T}}, r::Ref{NTuple{N,T}}) where {N,T} =
     convert(Ptr{T}, unsafe_convert(Ptr{NTuple{N,T}}, r))
-unsafe_convert(::Type{Ptr{T}}, r::Ptr{NTuple{N,T}}) where {N,T} =
```
JeffBezanson: Oops, I guess this is needed after all to resolve an ambiguity.

simonbyrne: reverted

Issue #33859, opened 2019-11-15
error/warning for unary -,+ in void context?


JeffBezanson:
I recently saw a case where code like this:
```
x = f(z) + g(z) - h(z)
y = ...
```
was reformatted to:
```
x = f(z) + g(z)
         - h(z)
y = ...
```

Of course, that silently changes the `x` value computed, and the value of `- h(z)` is just ignored. While rare, it's totally understandable how a change like this could be made accidentally. It makes me wonder if we should actually make this an error. Very few cases are affected; this can basically only happen with `-` and `+`, and it's hard to see why you would want to call those with one argument in void context.

stevengj:
I've defined `+(::String)` to file an issue for the most recently loaded package, so adding a warning here would [break my workflow](https://xkcd.com/1172/).

StefanKarpinski:
I realize that's not a serious comment, but that wouldn't actually be in void context, so this change wouldn't break your workflow 😁

stevengj:
Actually it would be in a void context—the point of the joke was that this would only affect someone who has perversely defined a unwary +/- method with a side effect.

DanielDoehring:
I recently lost half a day in debugging a statement of the form 
```
a = b 
   + c
```

which is not the desired 
`a = b +c `

I can understand that a leading `-` is taken without complaints, but what about throwing at least a warning if there are statements beginning with an `+` ?
In my opinion, there are extremely few cases (not talking about the representation of internal functions like `sum` for one element) where a user would use `+` with only one operator.

PR #34200, opened 2019-12-26
allow ± and ∓ as unary operators

simeonschaub:
This came up on slack, that it would be nice to have `±` as an allowed unary operator. This PR also adds `∓`. <s>further plus-like operators, which looked enough like `+` and `-` to make sense in my mind, but I can also remove the weirder ones, if so desired</s>. Probably should have tests, but I'm not quite sure what to test here exactly.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,10 @@ New language features
 * `⨟` is now parsed as a binary operator with times precedence. It can be entered in the REPL
   with `\bbsemi` followed by <kbd>TAB</kbd> ([#34722]).
 
+* `±` and `∓` are now unary operators as well, like `+` or `-`. Attention has to be paid in
+  macros and matrix constructors, which are whitespace sensitive, because expressions like
+  `[a ±b]` now get parsed as `[a ±(b)]` instead of `[±(a, b)]`. ([#34200])
+
 Language changes
 ----------------
 
--- a/src/julia-parser.scm
+++ b/src/julia-parser.scm
@@ -102,12 +102,13 @@
                                       0))
 
 (define unary-ops (append! '(|<:| |>:|)
-                           (add-dots '(+ - ! ~ ¬ √ ∛ ∜ ⋆))))
+                           (add-dots '(+ - ! ~ ¬ √ ∛ ∜ ⋆ ± ∓))))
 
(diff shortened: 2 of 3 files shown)


JeffBezanson:
Will have to decide how to transition the parsing of e.g. `[1 ±2]`.

tkf:
I suppose merging this PR would make #33859 more breaking?

JeffBezanson:
Strictly speaking yes, but these operators are used much less often than + and -.

stevengj:
I'm skeptical that all of these operators should be unary.  `±` and `∓`, certainly, but is there any example in the literature of the other operators being unary?

JeffBezanson:
Probably should not include ≏ "difference between", which doesn't make sense for 1 argument.

JeffBezanson:
Let's cut this down to just `±` and `∓`, which are obviously useful, and run PkgEval to see if anything breaks.

simeonschaub:
@JeffBezanson That seems like a good plan. Adding further operators can always be discussed later, so it makes sense to keep this more minimal. Could someone trigger PkgEval?

StefanKarpinski:

StefanKarpinski:
I think I did that right? 🤷‍♂

nanosoldier:
[Your package evaluation job](https://github.com/JuliaLang/julia/pull/34200#issuecomment-581567145) has completed - possible new issues were detected. A full report can be found [here](https://github.com/JuliaCI/NanosoldierReports/blob/da6eb2fd97d86cc99fb4933984c67730f1f7c3b2/pkgeval/by_hash/4b6d267_vs_2c43528/report.md). cc @maleadt

StefanKarpinski:
None of the failures seem related to this change (a lot of random tests with now seed being set—tsk, tsk) and a couple of segfaults, which are troubling but unlikely to be due to this.

simeonschaub:
What is the current state on this? Is this acceptable minor breakage for a next release, or do we need to deprecate this syntax in whitespace sensitive contexts for a intermediary release?

stevengj:
Seems fine for a minor release to me.

StefanKarpinski:
Just needs a rebase and merge after CI passes.

PR #34202, opened 2019-12-26
Return the first argument in `put!`

tkf:
`push!` usually returns the first argument ("container")

```julia
julia> push!([], 1)
1-element Array{Any,1}:
 1

julia> push!(Dict(), :a => 1)
Dict{Any,Any} with 1 entry:
  :a => 1

julia> push!(Set(), 1)
Set{Any} with 1 element:
  1
```

but not with channels

```julia
julia> push!(Channel(Inf), 1)
1
```

This PR fixes this inconsistency by tweaking `put!` such that `put!(c, _) === c`.  This automatically fixes `push!` as `push!` is just an alias of `put!`:

https://github.com/JuliaLang/julia/blob/2835347ff051f85997220dfe0dc152249763c6f4/base/channels.jl#L351

This also makes `put!(::Channel, _)` consistent with `put!(::RemoteChannel, _)`

https://github.com/JuliaLang/julia/blob/2835347ff051f85997220dfe0dc152249763c6f4/stdlib/Distributed/src/remotecall.jl#L584-L591

### DIFF
--- a/base/channels.jl
+++ b/base/channels.jl
@@ -304,14 +304,19 @@ Append an item `v` to the channel `c`. Blocks if the channel is full.
 
 For unbuffered channels, blocks until a [`take!`](@ref) is performed by a different
 task.
+Return the first argument.
 
 !!! compat "Julia 1.1"
     `v` now gets converted to the channel's type with [`convert`](@ref) as `put!` is called.
+
+!!! compat "Julia 1.4"
+    `put!` now returns the channel `c`; older versions of Julia returned `v`.
 """
 function put!(c::Channel{T}, v) where T
     check_channel_state(c)
     v = convert(T, v)
-    return isbuffered(c) ? put_buffered(c, v) : put_unbuffered(c, v)
+    isbuffered(c) ? put_buffered(c, v) : put_unbuffered(c, v)
+    return c
 end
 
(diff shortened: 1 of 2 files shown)


StefanKarpinski:
Perhaps that should be the distinction between push and put for channels?

tkf:
Sounds reasonable to me.  But in this case, ideally, I think `put!(::RemoteChannel, v)` should then be re-defined to return `v` and the vararg form `put!(::RemoteChannel, args...)` should be deprecated (or maybe return `args[end]`?) for consistency.  This is technically more breaking as the docstring for `put!(::RemoteChannel, args...)` explicitly says "Return the first argument."  Perhaps this has to be postponed until 2.0?

Maybe the least breaking option is to just define

```julia
push!(c::AbstractChannel, v) = (put!(c, v); c)
```

and leave `put!` methods as-is.  This would let people use `Channel` and `RemoteChannel` with a consistent API.

TotalVerb:
`put!` for a `Future` also returns the first argument, which is very useful for e.g. creating an immediately resolved future `put!(Future(), x)`. So for consistency, I'd also prefer `put!` to return the channel when the first argument is a channel. It is not too bad to do `v = value; put!(c, v); v` in the cases where the value is actually needed.

PR #34208, opened 2019-12-28
Fix codegen for getfield of homogeneous tuples

### LINKED ISSUE #34206: Malformed view created by ntuple on master
```
function mre()
  a = [44]
  b = ntuple(_ -> view(a, :), 1)[1]
  @assert pointer(a) == pointer(b)
end
```
The assertion in the above function fails on
```
julia> versioninfo()
Julia Version 1.4.0-DEV.667
Commit acb7bd93fb* (2019-12-27 21:20 UTC)
Platform Info:
  WORD_SIZE: 64
  LIBM: libopenlibm
```
Works fine on 1.3 and it seems like this issue was introduced by the recent layout changes.

### LINKED ISSUE #34207: Unreachable reached
TOMLs attached. This occurs on nightly, but not on 1.0–1.3. On Linux "Unreachable reached" is printed, on macOS, it simply segfaults:

```julia
❯ /Applications/Julia-1.4.app/Contents/Resources/julia/bin/julia
               _
  (_)     | (_) (_)    |
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.4.0-DEV.667 (2019-12-27)
 _/ |\__'_|_|_|\__'_|  |  Commit acb7bd93fb (0 days old master)
|__/                   |

(@v1.4) pkg> activate .
Activating environment at `/private/tmp/crash/Project.toml`
```
```julia
julia> using AtomicLevels, HalfIntegers

julia> function crash(a, b)
           ao,bo = Orbital(:k, a.orb.ℓ),Orbital(:k, b.orb.ℓ)
           i,j = ao ≤ bo ? (1,2) : (2,1)
           uv = (ao,bo)[[i,j]]
           m′m = (abs(a.m[1]), abs(b.m[1]))
           # Intermittent segfault on next line
           m′,m = uv[1] < uv[2] ? m′m[[i,j]] : reverse(minmax(m′m...))
       end
crash (generic function with 1 method)

julia> o = SpinOrbital(o"ks", 0, half(1))
ks₀α

julia> crash(o, o)

in expression starting at REPL[5]:1
_platform_strcmp at /usr/lib/system/libsystem_platform.dylib (unknown line)
Allocations: 30720411 (Pool: 30719296; Big: 1115); GC: 6
zsh: segmentation fault  /Applications/Julia-1.4.app/Contents/Resources/julia/bin/julia
```
[TOMLs.zip](https://github.com/JuliaLang/julia/files/4007026/TOMLs.zip)

Keno:
This fastpath wasn't checking whether or not the element type
was inline allocated or not.

Fixes #34206
Fixes #34207

### DIFF
--- a/test/core.jl
+++ b/test/core.jl
@@ -7140,3 +7140,10 @@ end
 struct SplatBadIterate; end
 Base.iterate(s::SplatBadIterate, args...) = ()
 @test_throws BoundsError (SplatBadIterate()...,)
+
+# Issue #34206/34207
+function mre34206(a)
+    b = ntuple(_ -> view(a, :), 1)[1]
+    b.offset1
+end
+@test mre34206([44]) == 0
--- a/test/core.jl
+++ b/test/core.jl
@@ -7143,7 +7143,12 @@ Base.iterate(s::SplatBadIterate, args...) = ()
 
 # Issue #34206/34207
 function mre34206(a)
-    b = ntuple(_ -> view(a, :), 1)[1]
+function mre34206(a, n)
+    va = view(a, :)
+    b = ntuple(_ -> va, n)::Tuple{Vararg{typeof(va)}}
+    return b[1].offset1
+end
+@test mre34206([44], 1) == 0
     b.offset1
 end
 @test mre34206([44]) == 0
--- a/src/codegen.cpp
+++ b/src/codegen.cpp
@@ -2892,20 +2892,22 @@ static bool emit_builtin_call(jl_codectx_t &ctx, jl_cgval_t *ret, jl_value_t *f,
                         jl_value_t *jt = jl_tparam0(utt);
                         if (jl_is_vararg_type(jt))
                             jt = jl_unwrap_vararg(jt);
-                        Value *vidx = emit_unbox(ctx, T_size, fld, (jl_value_t*)jl_long_type);
-                        // This is not necessary for correctness, but allows to omit
-                        // the extra code for getting the length of the tuple
-                        jl_value_t *boundscheck = (nargs == 3 ? argv[3].constant : jl_true);
-                        if (!bounds_check_enabled(ctx, boundscheck)) {
-                            vidx = ctx.builder.CreateSub(vidx, ConstantInt::get(T_size, 1));
-                        } else {
-                            vidx = emit_bounds_check(ctx, obj, (jl_value_t*)obj.typ, vidx,
-                                emit_datatype_nfields(ctx, emit_typeof_boxed(ctx, obj)),
-                                jl_true);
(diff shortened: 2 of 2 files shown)


vtjnash:
Much thanks for this! I think this is perhaps just a design issue with `typed_load`? It's rather awkward that all callers now avoid calling it unless they're fairly confident that they won't accidentally hit the `isboxed` case. Seems like that should have been an argument to the function instead?

test/core.jl
```diff
@@ -7140,3 +7140,10 @@ end
 struct SplatBadIterate; end
 Base.iterate(s::SplatBadIterate, args...) = ()
 @test_throws BoundsError (SplatBadIterate()...,)
+
+# Issue #34206/34207
+function mre34206(a)
+    b = ntuple(_ -> view(a, :), 1)[1]
```
vtjnash:
```suggestion
function mre34206(a, n)
    va = view(a, :)
    b = ntuple(_ -> va, n)::Tuple{Vararg{typeof(va)}}
    return b[1].offset1
end
@test mre34206([44], 1) == 0
```

The test was relying on constant propagation heuristics being disabled because we decide not to inline that ntuple call. I think this rewrite should give us greater confidence that changes to inference would find it very difficult to cause this to get any better or worse type information than we need it to have to trigger this codegen path.

JeffBezanson:
Yes, it would be nice if we could preserve this fast path for the case where the element type is known *not* to be inlinealloc.

test/core.jl
```diff
@@ -7143,7 +7143,12 @@ Base.iterate(s::SplatBadIterate, args...) = ()
 
 # Issue #34206/34207
 function mre34206(a)
-    b = ntuple(_ -> view(a, :), 1)[1]
+function mre34206(a, n)
```
cafaxo: Seems like something went wrong here?

JeffBezanson:
Should `julia_type_to_llvm` be checking inlinealloc instead of `jl_is_concrete_immutable`?

JeffBezanson:
How about something like this:
```diff
--- a/src/cgutils.cpp
+++ b/src/cgutils.cpp
@@ -1321,6 +1321,10 @@ static jl_cgval_t typed_load(jl_codectx_t &ctx, Value *ptr, Value *idx_0
     Type *elty = julia_type_to_llvm(jltype, &isboxed);
     if (type_is_ghost(elty))
         return ghostValue(jltype);
+    if (!isboxed && !jl_datatype_isinlinealloc(jltype)) {
+        elty = T_prjlvalue;
+        isboxed = true;
+    }
     Type *ptrty = PointerType::get(elty, ptr->getType()->getPointerAddressSpace());
```

PR #34209, opened 2019-12-28
SparseArrays: specialize zero(...) so result has nnz(...) = 0

### LINKED ISSUE #31835: zero(::AbstractSparseArray) full of structural nonzeros set to zero
An obvious implementation for `zero(a::AbstractSparseArray)` would be

```julia
zero(a::AbstractSparseArray) = spzeros(eltype(a), size(a)...)
```

However, in reality there's no specialization, and (the abstract fallback uses `fill!`:

https://github.com/JuliaLang/julia/blob/93c9ae488d48d14214873d683d8f039313bdbd4d/base/abstractarray.jl#L903

In particular, that means that it will be filled with non-structural zeros.

The reason why I haven't sent the obvious pull request is that this change is backwards incompatible where it relates to object identity. Namely, a side-effect of using `fill!` is that for mutable objects, every zero has the same identity. Example:

```
julia> using SparseArrays

julia> y = zero(BigInt[1,2,3]); y[1] === y[2]
true

julia> y = zero(sparse(BigInt[1,2,3])); y[1] === y[2]
true

julia> Base.zero(a::AbstractSparseArray) = spzeros(eltype(a), size(a)...)

julia> y = zero(sparse(BigInt[1,2,3])); y[1] === y[2]
false
```
What do you think? Can we make this change in a minor release? If so, I'll gladly send in a pull request.

tkluck:
This pull request fixes #31835.

Before this commit, the result of
```julia
zero(x::AbstractSparseArray)
```
is filled with stored zeros exactly where `x` has a stored value. That is a
wasteful artifact of the default fallback implementation for `AbstractArray`.
After this commit, we return a sparse array of the same dimensions as `x`
without any stored values.

This change is backwards incompatible where it relates to object identity.
Before this commit, for mutable objects, every stored zero has the same
identity. Example:

    julia> using SparseArrays

    julia> y = zero(BigInt[1,2,3]); y[1] === y[2]
    true

    julia> y = zero(sparse(BigInt[1,2,3])); y[1] === y[2]
    true

    julia> Base.zero(a::AbstractSparseArray) = spzeros(eltype(a), size(a)...)

    julia> y = zero(sparse(BigInt[1,2,3])); y[1] === y[2]
    false

But this behaviour should be classified as a performance bug and can therefore
be fixed in a minor (but not a patch) release.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -80,6 +80,10 @@ Standard library changes
 
 #### SparseArrays
 
+* The return value of `zero(x::AbstractSparseArray)` has no stored zeros anymore ([#31835]).
+  Previously, it would have stored zeros wherever `x` had them. This makes the operation
+  constant time instead of `O(<number of stored values>)`.
+
 #### Dates
 
 #### Statistics
--- a/stdlib/SparseArrays/src/SparseArrays.jl
+++ b/stdlib/SparseArrays/src/SparseArrays.jl
@@ -10,7 +10,7 @@ using Base: ReshapedArray, promote_op, setindex_shape_check, to_shape, tail,
 using Base.Sort: Forward
 using LinearAlgebra
 
-import Base: +, -, *, \, /, &, |, xor, ==
+import Base: +, -, *, \, /, &, |, xor, ==, zero
 import LinearAlgebra: mul!, ldiv!, rdiv!, cholesky, adjoint!, diag, eigen, dot,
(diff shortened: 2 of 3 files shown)


ViralBShah (approved):

tkluck:
Thanks for approving @ViralBShah !

Just had a look at the CI failure. The only failure is `buildbot/tester_win32` and nothing in the logs seems related to this change. Should be good to merge.

PR #34211, opened 2019-12-28
Add hpmv! to BLAS in stdlib/LinearAlgebra

iolaszlo:


stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -823,6 +824,77 @@ Only the [`ul`](@ref stdlib-blas-uplo) triangle of `A` is used.
...
+        throw(DimensionMismatch("Packed Hermitean Matrix A has size smaller than length(x) =  $(N)."))
+    end
+    GC.@preserve x y AP hpmv!(uplo, N, convert(T, α), AP, pointer(x), BlasInt(stride(x, 1)), convert(T, β), pointer(y), BlasInt(stride(y, 1)))
+    y
+end
+
+"""
+    hpmv!(uplo, α, AP, x, β, y)
+
+Update vector `y` as `α*AP*x + β*y` where `AP` is a a packed Hermitian matrix.
```
dkarrasch: remove one "a"

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -823,6 +824,77 @@ Only the [`ul`](@ref stdlib-blas-uplo) triangle of `A` is used.
...
+The storage layout for `AP` is described the reference BLAS module, level-2 BLAS at
```
dkarrasch: described IN the reference

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -823,6 +824,77 @@ Only the [`ul`](@ref stdlib-blas-uplo) triangle of `A` is used.
...
+function hpmv!(uplo::AbstractChar,
+               α::Number, AP::Union{DenseArray{T}, AbstractVector{T}}, x::Union{DenseArray{T}, AbstractVector{T}},
+               β::Number, y::Union{DenseArray{T}, AbstractVector{T}}) where T<:BlasFloat
+    require_one_based_indexing(AP, x, y)
+    N = length(x)
+    if N != length(y)
+        throw(DimensionMismatch("x has length $(N), but y has length $(length(y))"))
+    end
+    if length(AP) < Int64(N*(N+1)/2)
+        throw(DimensionMismatch("Packed Hermitean Matrix A has size smaller than length(x) =  $(N)."))
```
dkarrasch: "Hermitian matrix"

dkarrasch (changes requested):
Just a couple of typos, otherwise LGTM.

stdlib/LinearAlgebra/src/blas.jl
iolaszlo: Removed  one "a".

stdlib/LinearAlgebra/src/blas.jl
iolaszlo: Added "in".

stdlib/LinearAlgebra/src/blas.jl
iolaszlo: Changed to "Hermitian matrix".

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -823,6 +824,77 @@ Only the [`ul`](@ref stdlib-blas-uplo) triangle of `A` is used.
...
+                  β,
+                  y,
+                  incy)
+        end
+    end
+end
+
+function hpmv!(uplo::AbstractChar,
+               α::Number, AP::Union{DenseArray{T}, AbstractVector{T}}, x::Union{DenseArray{T}, AbstractVector{T}},
+               β::Number, y::Union{DenseArray{T}, AbstractVector{T}}) where T<:BlasFloat
```
dkarrasch: I'm not sure but it seems to me that it should be `{T<:BlasComplex}`? `AP`, `x` and `y` are not converted in this method. Or are they converted somewhere implicitly? Also, the type parameters (after `where`) should be wrapped by curly brackets.

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -823,6 +824,77 @@ Only the [`ul`](@ref stdlib-blas-uplo) triangle of `A` is used.
...
+    GC.@preserve x y AP hpmv!(uplo, N, convert(T, α), AP, pointer(x), BlasInt(stride(x, 1)), convert(T, β), pointer(y), BlasInt(stride(y, 1)))
+    y
+end
+
+"""
+    hpmv!(uplo, α, AP, x, β, y)
+
+Update vector `y` as `α*AP*x + β*y` where `AP` is a packed Hermitian matrix.
+The storage layout for `AP` is described in the reference BLAS module, level-2 BLAS at
+<http://www.netlib.org/lapack/explore-html/>.
```
dkarrasch: I think we should mention here that `AP`, `x` and `y` must be complex.

stdlib/LinearAlgebra/test/blas.jl
```diff
@@ -200,6 +201,48 @@ Random.seed!(100)
             @test_throws DimensionMismatch BLAS.trmm('R','U','N','N',one(elty),triu(Cnn),Cnm)
         end
 
+        #hpmv!
+        @testset "hpmv!" begin
+            if elty == ComplexF32 || elty == ComplexF64
```
dkarrasch: Nitpicking, but maybe you should stick a little closer to the style in this test file. The comment line is probably obsolete, and the testset should be set up only if `elty` is complex, similar to `her2k!` in line 26/27:
```julia
if elty in (ComplexF32, ComplexF64)
    @testset "hpmv!" begin
```
Finally, your tests are inside the `let`-block started on line 55/56, within which `n = 10` is used for vector lengths/matrix sizes. You could replace your loop over `N in 2:5` by a single test with `n = 10` instead of `N in 2:5`, or move your tests outside of the block.

stdlib/LinearAlgebra/test/blas.jl
```diff
@@ -200,6 +201,48 @@ Random.seed!(100)
...
+                # Both matrix dimensions N coincide, as we have Hermitian matrices.
+                for N in 2:5
+                    # Define the inputs and outputs of hpmv!, y = α*A*x+β*y
+                    α = rand(elty) + elty(im) * rand(elty)
```
dkarrasch: There is a complex random number generator, so `α = rand(elty)` is sufficient to have complex scalar numbers and achieve the same goal that you trying to achieve here. Same with the matrices below:
```julia
M = rand(elty, N, N) + elty(im)*rand(elty, N ,N)
```
generates two complex(!) `N` by `N` matrices, multiplies the second with `im` and then adds the two.

dkarrasch (changes requested):

stdlib/LinearAlgebra/src/blas.jl
iolaszlo: Thanks for pointing this out. I agree also that  ```T```  should be```T <: BlasComplex``` and will apply this change  in order to enforce the correct input type. I haven't added any explicit type conversions for arrays and there are no implicit type conversions during the ```ccall``` as its  array inputs are of type Ptr{eltype} whose eltype can only be ComplexF32 and ComplexF64. Thus for example for real ```Float64``` array inputs we get a method error as expected:
```
 MethodError: no method matching Ptr{Array{Float64,2}}(::Array{Float64,1})
```

stdlib/LinearAlgebra/src/blas.jl
iolaszlo: Added the specs for the input types in the documentation.

stdlib/LinearAlgebra/test/blas.jl
iolaszlo: Moved the ```if``` statement outside of the ```@testset```. Use predefined ```n``` as the matrix dimension.

stdlib/LinearAlgebra/test/blas.jl
iolaszlo: Corrected the unnecessarily complicated input/output generation.

dkarrasch (approved):
Neat. Let's try to get another review before merging it.

iolaszlo:
> Neat. Let's try to get another review before merging it.

Thank you for taking the time to review my commits.

iolaszlo:
Could someone please retrigger the win64 and win32 builds, as their failure doesn't seem to be related to the current change. Thank you.

ViralBShah:
Retriggered win64. Win32 had been a bit unhappy lately.

iolaszlo:
> Retriggered win64. Win32 has been a bit unhappy lately.

Thank you, the win64 build succeeded. Win32 still seems unhappy as it gives a timeout currently. :)

dkarrasch:
@iolaszlo In case you're not yet familiar with the conventions here, the "needs news" label means that you should write an entry in `NEWS.md` and announce the new feature in the `LinearAlgebra` section.

iolaszlo:
> @iolaszlo In case you're not yet familiar with the conventions here, the "needs news" label means that you should write an entry in `NEWS.md` and announce the new feature in the `LinearAlgebra` section.

Thank you for the information. I just added the update in `NEWS.md` and rebased the commit onto the most current master branch.

JeffBezanson:
Thanks, great first PR!

StefanKarpinski:
Anyone else to review? @andreasnoack or @jiahao care to take a look?

ViralBShah:
This is good to merge.

iolaszlo:
Thank you guys for the feedback and the quick responses.

PR #34212, opened 2019-12-28
runtests: Print currently running tests when asked

Keno:
With this patch, pressing '?' while the test are running will print
which tests are still running. Of course, you could just try to look
through the log to try to figure out which tests have started but not
yet finished, but that's pretty annoying to do. Being able to just ask
is much easier for the impatient developer.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -155,6 +155,7 @@ cd(@__DIR__) do
     try
         # Monitor stdin and kill this task on ^C
         # but don't do this on Windows, because it may deadlock in the kernel
+        running_tests = Dict{String, DateTime}()
         if !Sys.iswindows() && isa(stdin, Base.TTY)
             t = current_task()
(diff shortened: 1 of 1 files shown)


test/runtests.jl
```diff
@@ -155,16 +155,21 @@ cd(@__DIR__) do
...
                     REPL.Terminals.raw!(term, true)
                     while true
-                        if read(term, Char) == '\x3'
+                        c = read(term, Char)
+                        if c == '\x3'
                             Base.throwto(t, InterruptException())
                             break
+                        elseif c == '?'
+                            println("Currently running: ", join(running_tests, ", "))
+                            break
```
vtjnash:
```suggestion
```

vtjnash (approved):
Sounds like a nice idea! Should we add start time (or time elapsed since starting) too? (i.e., by making the Set a Dict)

PR #34217, opened 2019-12-30
Improve printing of BoundsError with colon

nalimilan:
Avoid exposing implementation details like `Base.Slice(Base.OneTo(n))` and print `:` instead of `Colon()`.

Before:
```julia
julia> rand(2, 2)[1:10, :]
ERROR: BoundsError: attempt to access 2×2 Array{Float64,2} at index [1:10, Base.Slice(Base.OneTo(2))]
```

After:
```julia
julia> rand(2, 2)[1:10, :]
ERROR: BoundsError: attempt to access 2×2 Array{Float64,2} at index [1:10, 1:2]
```

A more "radical" approach would be to implement `print` for `Slice`, `OneTo` and `Colon` so that it avoids details, and call it from `showerror`. What do you think?

### DIFF
--- a/base/errorshow.jl
+++ b/base/errorshow.jl
@@ -110,6 +110,11 @@ function show_error_hints(io, ex, args...)
     end
 end
 
+index_string(x::Any) = string(x)
+index_string(x::Slice) = index_string(x.indices)
+index_string(x::OneTo) = "1:" * string(x.stop)
+index_string(x::Colon) = ":"
+
 function showerror(io::IO, ex::BoundsError)
     print(io, "BoundsError")
     if isdefined(ex, :a)
@@ -121,7 +126,7 @@ function showerror(io::IO, ex::BoundsError)
             if isa(ex.i, AbstractRange)
                 print(io, ex.i)
             else
-                join(io, ex.i, ", ")
+                join(io, index_string.(ex.i), ", ")
             end
             print(io, ']')
         end
--- a/base/errorshow.jl
+++ b/base/errorshow.jl
@@ -110,10 +110,12 @@ function show_error_hints(io, ex, args...)
     end
 end
 
-index_string(x::Any) = string(x)
-index_string(x::Slice) = index_string(x.indices)
-index_string(x::OneTo) = "1:" * string(x.stop)
-index_string(x::Colon) = ":"
+show_index(io::IO, x::Any) = show(io, x)
+show_index(io::IO, x::Slice) = show_index(io, x.indices)
+show_index(io::IO, x::LogicalIndex) = show_index(io, x.mask)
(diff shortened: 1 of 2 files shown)


bkamins:
Also I would consider checking in `showerror` for `BoundsError` if its `i` fields is an `AbstractString` and if so print it verbatim as passed instead of `"at index [some_index_here]"`. We already have a special case for `AbstractRange`. This would allow more flexibility in printing custom error messages that would be more informative for the users.

JeffBezanson:
We could certainly use `:` for `show(:)`. `1:n` might also make sense as the `:compact` output of `OneTo(n)`?

nalimilan:
> We could certainly use `:` for `show(:)`. `1:n` might also make sense as the `:compact` output of `OneTo(n)`?

I thought about `:compact`, but AFAIK it's only documented to be used to print fewer digits (`?IOContext` says "Boolean specifying that small values should be printed more compactly"). `?print` says "The representation used by print includes minimal formatting and tries to avoid Julia-specific details."  `:typeinfo` could also be used since the details we want to remove are all related to type information (but we would have to abuse it slightly by pretending the type is already available from `showerror`, which isn't the case).

nalimilan:
Bump.

StefanKarpinski:
Who should review, @nalimilan?

bkamins:
@nalimilan apart from rebasing I think it would be good to test against all cases you define in `index_string`.

nalimilan:
I've done that, plus a small improvement to avoid allocating a temporary string.

base/errorshow.jl
```diff
@@ -110,6 +110,13 @@ function show_error_hints(io, ex, args...)
     end
 end
 
+show_index(io::IO, x::Any) = show(io, x)
+show_index(io::IO, x::Slice) = show_index(io, x.indices)
+show_index(io::IO, x::LogicalIndex) = show_index(io, x.mask)
```
bkamins, line 115: how would that handle the case when `x.mask` is very long? Do we show then it compactly? (I guess we do but I just want to be sure)

nalimilan: Yes I get this:
```julia
julia> x[10, fill(true, 1000)]
ERROR: BoundsError: attempt to access 2×2 Array{Float64,2} at index [10, Bool[1, 1, 1, 1, 1, 1, 1, 1, 1, 1  …  1, 1, 1, 1, 1, 1, 1, 1, 1, 1]]
```

bkamins (approved):

nalimilan:
@JeffBezanson Good to merge?

nalimilan:
I'll merge in a few days if nobody objects.

PR #33205, opened 2019-09-09
Allow more general eltypes in sparse array multiplication

pablosanjose:
Fixes #33169

Replaces `one(T)` by `1` in five-argument `mul!` calls with `SparseArray` types, where `T` was before the target eltype. This allows eltypes that don't define `one` by relying on the fact that `1` is always a multiplicative identity for `AbtractArrays` as far as I can tell ~(even for `Bool` eltypes)~ EDIT: changed `1, 0` to `true, false`.

With this we can in particular use `StaticArray`s as eltypes of `SparseMatrixCSC`. It does not allow `Array` eltypes, since they don't define `zero`, which is used in `mul!`. Solving the issue also for eltypes that don't define `zero` is of much broader scope than the present PR due to the pervasive assumption throughout SparseArrays that `zero(eltype(A))` for `A::AbstractSparseArray` is defined.

As a consequence of allowing general (e.g. matrix-like) eltypes in all product combinations involving a sparse vector or matrices (see tests) a range of changes throughout SparseArrays were needed, including in `copy(::AbstractSparseMatrix)`, `halfperm!`, `spmatmul` and `_At_or_Ac_mul_B`.

This fix does not address a possible version of #33169 without any sparse factors (i.e. all changes here are strictly confined to the SparseArrays lib)

(This is my first PR to Julia stdlib. It might be a good idea to show this to Nanosoldier, and to have some careful review if possible)

PR #34219, opened 2019-12-30
edit and update NEWS for 1.4

JeffBezanson:
I added a section for "new library features", for fully-non-breaking extensions to functions (e.g. new keyword arguments). That seems worth distinguishing from entirely new functions, and from changes in the behavior of currently-supported forms.


ericphanson:
The sparse arrays change [#33205] doesn’t look like it belongs under Random.

ararslan:
I think we should just re-branch 1.4 after merging this.

JeffBezanson:
@ericphanson Good catch, thanks.

PR #34223, opened 2019-12-30
fix #33954, recursion through field types in `is_derived_type`

### LINKED ISSUE #33954: Compiler throws internal stack overflow error when plotting with Makie on Julia master
MWE:

```julia
julia> versioninfo()
Julia Version 1.4.0-DEV.519
Commit 76ebc419f0 (2019-11-24 03:08 UTC)
Platform Info:
  CPU: Intel(R) Core(TM) i7-6820HQ CPU @ 2.70GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
Environment:
  JULIA_PATH = /home/scheme/build/julia/usr

julia> begin
       using Makie

        x = rand(10)
        y = rand(10)
        colors = rand(10)
        scene = scatter(x, y, color = colors)
        end
StackOverflowError()
getindex at ./essentials.jl:583
iterate at ./essentials.jl:602 [inlined]
iterate at ./essentials.jl:602 [inlined]
is_derived_type at ./compiler/typelimits.jl:68
is_derived_type at ./compiler/typelimits.jl:58
is_derived_type at ./compiler/typelimits.jl:71
is_derived_type at ./compiler/typelimits.jl:58
...
```

JeffBezanson:
I'm not fully sure how best to address this. But, I assume looking through field types in `is_derived_type` can't be that important, since we don't do it for any non-bits types. So I just tried something simple and limited the depth of recursion that can happen via that path.

fix #33954

### DIFF
--- a/src/datatype.c
+++ b/src/datatype.c
@@ -306,6 +306,8 @@ static int references_name(jl_value_t *p, jl_typename_t *name) JL_NOTSAFEPOINT
     if (jl_is_datatype(p)) {
         if (((jl_datatype_t*)p)->name == name)
             return 1;
+        if (((jl_datatype_t*)p)->layout && jl_datatype_nfields(p) == 0)
+            return 0;
         size_t i, l = jl_nparams(p);
         for (i = 0; i < l; i++) {
             if (references_name(jl_tparam(p, i), name))
@@ -401,6 +403,7 @@ void jl_compute_field_offsets(jl_datatype_t *st)
             jl_value_t *fld = jl_svecref(w->types, i);
             if (references_name(fld, w->name)) {
                 isinlinealloc = 0;
+                isbitstype = 0;
                 break;
             }
         }
--- a/base/compiler/typelimits.jl
+++ b/base/compiler/typelimits.jl
@@ -57,21 +57,6 @@ function is_derived_type(@nospecialize(t), @nospecialize(c), mindepth::Int)
         for p in cP
             is_derived_type(t, p, mindepth) && return true
         end
-        if isconcretetype(c) && isbitstype(c)
-            # see if it was extracted from a fieldtype
-            # however, only look through types that can be inlined
-            # to ensure monotonicity of derivation
-            # since we know that for immutable, concrete, bits types,
-            # the field types must have been constructed prior to the type,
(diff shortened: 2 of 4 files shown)


src/datatype.c
```diff
@@ -523,6 +525,7 @@ void jl_compute_field_offsets(jl_datatype_t *st)
         //else if (st->layout->fielddesc_type != 0) // GC only implements support for this
         //    isinlinealloc = 0;
         isinlinealloc = 0;
+        isbitstype = 0;
```
vtjnash: we just asserted this was zero on the previous line

vtjnash:
> can't be that important

I agree. Let's just delete it then—I don't think it's even correct as written (except perhaps for the trivial case of maxdepth=1). I think it was supposed to permit cases like this:
```f(x) = x isa X ? f(x.x) : other```
Where you're recursing into yourself over the fields of an object, but they happen to all be trivial things like `Ptr{T}` or `Dual{Complex{T}}`. However, in that case, this function (is_derived_type) would only be allowed to look at the fields of the declared type object, and should not be permitted to look inside the fields of any parameters.

File: src/datatype.c (hunk 1 above)
Comment (JeffBezanson) [reply src/datatype.c]: Yes I have a fix for this.

File: src/datatype.c [hunk 2]
```diff
@@ -306,6 +306,8 @@ static int references_name(jl_value_t *p, jl_typename_t *name) JL_NOTSAFEPOINT
     if (jl_is_datatype(p)) {
         if (((jl_datatype_t*)p)->name == name)
             return 1;
+        if (((jl_datatype_t*)p)->layout && (jl_datatype_nfields(p) == 0 || ((jl_datatype_t*)p)->isbitstype))
```
vtjnash: I’m not sure we’ve always computed layout yet. In theory, this can eventually be changed to search the declared type structure (wrapper) for a use of the TypeVar as the direct field type (or as a Union) and cached on the TypeName

src/datatype.c
vtjnash: It seemed like there might be a case above where we did set not-inlinealloc (from references-name), but don’t simultaneously also set not-isbitstype

vtjnash (approved):

Issue #34157, opened 2019-12-20
searchsorted corner case


jw3126:
Today I hit the following corner case of `searchsorted` and friends:
```julia
julia> searchsorted(1:2.0, -Inf) # works
1:0

julia> searchsorted([1,2], -Inf) # works
1:0

julia> searchsorted(1:2, -Inf) # throws
ERROR: InexactError: trunc(Int64, Inf)
```

inkydragon:
https://github.com/JuliaLang/julia/blob/3af96bcefcb0aec559d4652fd75e1bc3fcfe99da/base/sort.jl#L254-L261

PR #34224, opened 2019-12-31
fix searchsorted corner case #34157

jw3126:


timholy:
@nanosoldier `runbenchmarks(ALL, vs = @8b57f642c45b6158db0dce145e1c1aa8c1ba425c)`

timholy:
Hmm, is it not possible to trigger nanosoldier on merged PRs?

timholy:
If this wasn't tested on the Jan 2 run (and it was merged close to the boundary between days, so I'm not certain), then the nanosoldier run seems promising: https://github.com/JuliaCI/BaseBenchmarkReports/blob/5d9e24c56e4ad9f013a362f3f904debfd88a4af2/daily_2020_01_03/report.md

I was worried because this is such a performance hotspot. Nice work @jw3126!

PR #34225, opened 2019-12-31
Fix CI hangs on win32

Keno:
After much debugging, it turns out that OpenBlas is sometimes hanging
on exit due to bad assumptions in its thread management. The upstream
discussion is at https://github.com/xianyi/OpenBLAS/pull/2350, but this
should fix our frequent win32 CI failures in the meantime.


Keno:
Looks like the BB OpenBLAS was upgrade to 0.3.7, but we never imported that into Julia? 0.3.7 needs a rebased patch. What's the right way to proceed here? @staticfloat

JeffBezanson:
Bump. Should we merge and backport this, then separately upgrade openblas to 0.3.7 and replace the patch?

vtjnash (approved):

Keno:
@staticfloat what's the status of the 0.3.7 binaries that have this patch?

staticfloat:
They are a part of https://github.com/JuliaLang/julia/pull/34242

staticfloat:
I'm building you a 0.3.5 version with your patch so that we can land this on 1.3 and 1.4 without needing to change the rest of the BB binaries.  It's going to require some manual renaming of tarballs, but should be done pretty soon.  https://github.com/JuliaPackaging/Yggdrasil/pull/356

ViralBShah:
Should we do a quick 1.3.2 with this?

Keno:
We should include it if we do 1.3.2, but not worth it to do a release just for this.

ViralBShah:
People have reported hangs in their windows CI on 1.3, which I thought this might fix.

Keno:
I doubt it. It's pretty hard to trigger. The only reason Julia managed to is that we spawn hundreds of julia processes in short succession and immediately kill them.

ViralBShah:
Ok. It may be a different win32 issue then. Could you take a quick look at https://github.com/Circuitscape/Omniscape.jl/issues/13

Keno:
Somebody will have to debug that. I don't think it looks like this issue because this issue causes julia to hang on exit (unless of course it's creating additional processes while it goes along). My suggestion would be to use AppVeyor's rdp feature and try to attach a debugger or generate a minidump.

ViralBShah:
It's only happening on Travis Windows, IIUC.

staticfloat:
We do need a 1.3.2 with a new Pkg anyway, so it should be on our radar.

PR #33936, opened 2019-11-25
More Dict and Set examples

aminya:
Ready to get merged...

Issue #29527, opened 2018-10-05
Improving the relationship between devdocs and code


c42f:
I've been working on the julia internals on and off when I get the time, for example, over at https://github.com/JuliaLang/julia/pull/28878.

One thing I've been finding is that I don't always get the benefit of the devdocs, as I tend to read the code and code comments as a first resort, and don't always find some of the great tidbits of information available in the devdocs. I've been finding that the relationship between devdocs and the code itself strangely disjointed. It's hard to put my finger on what the problem is exactly, but I thought I'd open an issue to see what other people think. Perhaps it's just me and my overuse of `grep` (actually, `ag`) to discover things in the `src` and `base` directories.

Personally I like the code itself and inline contextual comments to tell the story of how things are, rather than relying on documentation which is elsewhere. For example, I love @yuyichao's documentation of the GC, and I love @Keno's documentation of the static analyzer annotations. These two are some of the better documented parts of the internals, but having the static analyzer documentation in `doc/src/devdoc/gc-sa.md` meant I found it only by accident when browsing rather than just finding it in the source at `src/support/analyzer_annotations.h`.

This leaves me wondering how we can make the documentation of the internals a little more uniform and accessible. A very simple approach would be to link from the code comments to the devdocs where applicable. This would certainly help, but I feel some of the markdown devdocs could be embedded in — or next to — the code which would make working with the code itself (using simple tools like grep) simpler and more contextual. If we want to extract these and insert them into the generated devdocs that would be possible with an automated extraction tool, in the same way that we document packages.

Issue #14946, opened 2016-02-05
Use typealiases in method printing?


jiahao:
I'd like to request some speculative brain cells spent on how we print method signatures to the REPL.

Currently, we expand out all typealiases. While this policy has the advantage of producing unambiguous output, the composition of type aliases through type parameters leads to combinatorially long string representations.

A particularly egregious example in Base is `eigfact!`:

``` jl
julia> methods(eigfact!)

#12 methods for generic function "eigfact!":

...

eigfact!{T<:Union{Complex{Float32},Complex{Float64}}}(A::Union{DenseArray{T,2},SubArray{T,2,A<:DenseArray{T<:Any,N<:Any},I<:Tuple{Vararg{Union{Colon,Int64,Range{Int64}}}},LD<:Any}}) at linalg/eigen.jl:50

eigfact!{T<:Union{Float32,Float64}}(A::Union{DenseArray{T,2},SubArray{T,2,A<:DenseArray{T<:Any,N<:Any},I<:Tuple{Vararg{Union{Colon,Int64,Range{Int64}}}},LD<:Any}}, B::Union{DenseArray{T,2},SubArray{T,2,A<:DenseArray{T<:Any,N<:Any},I<:Tuple{Vararg{Union{Colon,Int64,Range{Int64}}}},LD<:Any}}) at linalg/eigen.jl:121

eigfact!{T<:Union{Complex{Float32},Complex{Float64}}}(A::Union{DenseArray{T,2},SubArray{T,2,A<:DenseArray{T<:Any,N<:Any},I<:Tuple{Vararg{Union{Colon,Int64,Range{Int64}}}},LD<:Any}}, B::Union{DenseArray{T,2},SubArray{T,2,A<:DenseArray{T<:Any,N<:Any},I<:Tuple{Vararg{Union{Colon,Int64,Range{Int64}}}},LD<:Any}}) at linalg/eigen.jl:142

...
```

I estimate that only three people in the world can read the first method signature and immediately recognize that the original calling signature was

```
eigfact!{T<:BlasComplex}(A::StridedMatrix{T})
```

and it would take an eagle eye to realize that the first method takes one argument whereas the other two take two inputs each.

It would be much nicer to print something like the original method signature instead of feeling overwhelmed by C++ expression template-like verbal diarrhea. It's clear, though, that the additional bookkeeping of typealiases and matching of typealiases to method signatures is a hard problem.

carnaval:
[...]

PR #34226, opened 2019-12-31
Support brief and extended docs (closes #25930)

### LINKED ISSUE #25930: Brief and extensive docs with `?` and `??`
Most methods have short documentation, but there are a few exceptions. It might be nice to have `??foo` mean something like "print only the first few lines of help for `foo`," basically for experienced users to remind themselves what arguments should be supplied in what order. EDIT: or the opposite, long docs with `??` and short docs with `?`.

Inspired by some extremely thorough (and useful) documentation in https://github.com/JuliaImages/ImageFiltering.jl/pull/54.

timholy:
Demo:
```julia
julia> """
           f()
           
       Do something cool with `f`.

       # Extended help

       The mathematics behind `f` is, like, totally intense: `∅`.
       """
       f() = nothing
f

help?> f
search: f fd for fma fld fld1 fill fdio frexp foldr foldl flush floor float

  f()

  Do something cool with f.

  ────────────────────────────────────────────────────────────────────────────

Extended help is available with `??`

help?> ?f
search: f fd for fma fld fld1 fill fdio frexp foldr foldl flush floor float

  f()

  Do something cool with f.

  Extended help
  ≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡

  The mathematics behind f is, like, totally intense: ∅.

julia> 
```

I've checked that this doesn't seem to break Juno. I've not tested jupyter.

### DIFF
--- a/stdlib/Markdown/src/Markdown.jl
+++ b/stdlib/Markdown/src/Markdown.jl
@@ -60,4 +60,30 @@ import Base.Docs: catdoc
 
 catdoc(md::MD...) = MD(md...)
 
+function trimdocs(md::MD, brief::Bool)
+    brief || return md
+    md, trimmed = _trimdocs(md, brief)
+    if trimmed
+        push!(md.content, Message("Extended help is available with `??`", (color=Base.info_color(), bold=true)))
+    end
+    return md
+end
+
+function _trimdocs(md::MD, brief::Bool)
+    content, trimmed = [], false
+    for c in md.content
+        if isa(c, Header{1}) && isa(c.text, AbstractArray) &&
+                                lowercase(c.text[1]) == "extended help"
+            trimmed = true
+            break
+        end
+        c, trm = _trimdocs(c, brief)
+        trimmed |= trm
+        push!(content, c)
+    end
+    return MD(content, md.meta), trimmed
+end
+
+_trimdocs(md, brief::Bool) = md, false
+
 end
--- a/stdlib/Markdown/src/Markdown.jl
+++ b/stdlib/Markdown/src/Markdown.jl
@@ -60,30 +60,4 @@ import Base.Docs: catdoc
 
 catdoc(md::MD...) = MD(md...)
 
-function trimdocs(md::MD, brief::Bool)
-    brief || return md
-    md, trimmed = _trimdocs(md, brief)
-    if trimmed
-        push!(md.content, Message("Extended help is available with `??`", (color=Base.info_color(), bold=true)))
-    end
-    return md
-end
-
-function _trimdocs(md::MD, brief::Bool)
-    content, trimmed = [], false
-    for c in md.content
-        if isa(c, Header{1}) && isa(c.text, AbstractArray) &&
-                                lowercase(c.text[1]) == "extended help"
-            trimmed = true
-            break
-        end
-        c, trm = _trimdocs(c, brief)
-        trimmed |= trm
-        push!(content, c)
-    end
-    return MD(content, md.meta), trimmed
-end
-
-_trimdocs(md, brief::Bool) = md, false
-
 end
--- a/stdlib/Markdown/src/parse/parse.jl
+++ b/stdlib/Markdown/src/parse/parse.jl
@@ -8,13 +8,6 @@ mutable struct MD
         new(content, meta)
 end
 
-# For direct messages to the terminal
-struct Message
-    msg    # AbstractString
-    fmt    # keywords to `printstyled`
-end
-Message(msg) = Message(msg, ())
-
 MD(xs...) = MD(vcat(xs...))
 
 function MD(cfg::Config, xs...)
--- a/stdlib/Markdown/src/parse/parse.jl
+++ b/stdlib/Markdown/src/parse/parse.jl
@@ -8,6 +8,13 @@ mutable struct MD
         new(content, meta)
 end
 
+# For direct messages to the terminal
+struct Message
+    msg    # AbstractString
+    fmt    # keywords to `printstyled`
+end
+Message(msg) = Message(msg, ())
+
 MD(xs...) = MD(vcat(xs...))
 
 function MD(cfg::Config, xs...)
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,6 +12,14 @@ Language changes
   where it used to be incorrectly allowed. This is because `NTuple` refers only to homogeneous
   tuples (this meaning has not changed) ([#34272]).
 
+* In docstrings, a level-1 markdown header "Extended help" is now
+  interpreted as a marker dividing "brief help" from "extended help."
+  The REPL help mode only shows the brief help (the content before the
+  "Extended help" header) by default; prepend the expression with '?'
+  (in addition to the one that enters the help mode) to see the full
+  docstring. ([#25903])
+
+
 Multi-threading changes
 -----------------------
 
--- a/doc/src/manual/documentation.md
+++ b/doc/src/manual/documentation.md
@@ -197,6 +197,10 @@ As in the example above, we recommend following some simple conventions when wri
    rather than users, explaining e.g. which functions should be overridden and which
    functions automatically use appropriate fallbacks. Such details are best kept separate
    from the main description of the function's behavior.
+5. For long docstrings, consider splitting the documentation with an
+   `# Extended help` header. The typical help-mode will show only the
+   material above the header; you can access the full help by adding a '?'
+   at the beginning of the expression (i.e., "??foo" rather than "?foo").
 
 ## Accessing Documentation
 
--- a/stdlib/Markdown/src/render/terminal/render.jl
+++ b/stdlib/Markdown/src/render/terminal/render.jl
@@ -116,6 +116,10 @@ function term(io::IO, br::HorizontalRule, columns)
    print(io, ' '^margin, '─'^(columns - 2margin))
 end
 
+function term(io::IO, msg::Message, columns)
+    printstyled(io, msg.msg; msg.fmt...)
+end
+
 term(io::IO, x, _) = show(io, MIME"text/plain"(), x)
 
 # Inline Content
--- a/stdlib/Markdown/src/render/terminal/render.jl
+++ b/stdlib/Markdown/src/render/terminal/render.jl
@@ -116,10 +116,6 @@ function term(io::IO, br::HorizontalRule, columns)
    print(io, ' '^margin, '─'^(columns - 2margin))
 end
 
-function term(io::IO, msg::Message, columns)
-    printstyled(io, msg.msg; msg.fmt...)
-end
-
 term(io::IO, x, _) = show(io, MIME"text/plain"(), x)
 
 # Inline Content
--- a/stdlib/Markdown/test/runtests.jl
+++ b/stdlib/Markdown/test/runtests.jl
@@ -1131,25 +1131,3 @@ let m = Markdown.parse("---"), io = IOBuffer()
     show(io, "text/latex", m)
     @test String(take!(io)) == "\\rule{\\textwidth}{1pt}\n"
 end
-
-# Brief and extended docs (issue #25930)
-let text =
-        """
-            brief_extended()
-
-        Short docs
-
-        # Extended help
-
-        Long docs
-        """,
-        md = Markdown.parse(text)
-    @test md == Markdown.trimdocs(md, false)
-    @test !isa(md.content[end], Markdown.Message)
-    mdbrief = Markdown.trimdocs(md, true)
-    @test length(mdbrief.content) == 3
-    @test isa(mdbrief.content[1], Markdown.Code)
-    @test isa(mdbrief.content[2], Markdown.Paragraph)
-    @test isa(mdbrief.content[3], Markdown.Message)
-    @test occursin("??", mdbrief.content[3].msg)
-end
--- a/stdlib/Markdown/test/runtests.jl
+++ b/stdlib/Markdown/test/runtests.jl
@@ -1131,3 +1131,25 @@ let m = Markdown.parse("---"), io = IOBuffer()
     show(io, "text/latex", m)
     @test String(take!(io)) == "\\rule{\\textwidth}{1pt}\n"
 end
+
+# Brief and extended docs (issue #25930)
+let text =
+        """
+            brief_extended()
+
+        Short docs
+
+        # Extended help
+
+        Long docs
+        """,
+        md = Markdown.parse(text)
+    @test md == Markdown.trimdocs(md, false)
+    @test !isa(md.content[end], Markdown.Message)
+    mdbrief = Markdown.trimdocs(md, true)
+    @test length(mdbrief.content) == 3
+    @test isa(mdbrief.content[1], Markdown.Code)
+    @test isa(mdbrief.content[2], Markdown.Paragraph)
+    @test isa(mdbrief.content[3], Markdown.Message)
+    @test occursin("??", mdbrief.content[3].msg)
+end
--- a/stdlib/REPL/src/docview.jl
+++ b/stdlib/REPL/src/docview.jl
@@ -19,12 +19,16 @@ using InteractiveUtils: subtypes
 helpmode(io::IO, line::AbstractString) = :($REPL.insert_hlines($io, $(REPL._helpmode(io, line))))
 helpmode(line::AbstractString) = helpmode(stdout, line)
 
+const extended_help_on = Ref{Any}(nothing)
+
 function _helpmode(io::IO, line::AbstractString)
     line = strip(line)
     if startswith(line, '?')
         line = line[2:end]
+        extended_help_on[] = line
         brief = false
     else
+        extended_help_on[] = nothing
         brief = true
     end
     x = Meta.parse(line, raise = false, depwarn = false)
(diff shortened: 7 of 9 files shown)


timholy:
I've added the docs and NEWS, so this is ready for any feedback.

johnnychen94:
A limitation of this feature is that REPL users still can't read those lengthy extended docstrings. I imagine some paging utils in doc mode would be more useful.

timholy:
> A limitation of this feature is that REPL users still can't read those lengthy extended docstrings. I imagine some paging utils (e.g., more) in doc mode would be more useful.

I think that's mostly an orthogonal PR. Currently help is displayed via the `show` method for markdown objects. One might have to modify the REPL-loop itself for help mode to be able to switch to using a pager.

johnnychen94:
Oh, what I intended to say is once there was a pager, the feature proposed in this PR would be useless. IPython uses `??` to get the source codes, I suggest we reserve this "keyword" for that usage.

timholy:
That's an idea worth exploring. Would we use the pager to handle multiple methods? I.e., if you do `?findall`, rather than seeing the horizontal rule divider you just get separate pages, perhaps with an "index" at the beginning to choose the method you want? I've noticed a bit of a readability problem when there are many methods, even when their help is short, so a pager might fix this.

A downside of the pager is that it might be slightly irritating to have to exit the pager each time you do `?function_with_long_help` and all you're looking for is a quick API summary.

>  IPython uses ?? to get the source codes, I suggest we reserve this "keyword" for that usage.

We already have `@less` and `@edit` for this. To me it seems that IPython's usage of this is just a way of giving up, "I can't imagine providing any better help than just showing you the source." Personally, I don't think that's a good direction to go in, and that it would be better to say, "ok, you want more help, let me direct you to this more detailed document."

johnnychen94:
> Would we use the pager to handle multiple methods?

I think that makes reading docstrings easier.

> A downside of the pager is that it might be slightly irritating to have to exit the pager each time you do ?function_with_long_help and all you're looking for is a quick API summary.

Is something like `?function_with_long_help |> pager` achievable?

timholy:
> Is something like ?function_with_long_help |> pager achievable?

Currently that retrieves the documentation for `|>`. I see the point, but how do you know you "want" the pager until you know whether the documentation will be long? It seems that's something the documentation *author* needs to figure out, and that the user doesn't (yet) have enough information.

timholy:
If someone comes along later and adds pager support, will we regret anything here? I think the real issue is whether we're happy with `# Extended help` being given special meaning.

Would be nice to get a code review, maybe @mortenpi? But if no one has time I'll merge this in the next few days.

stdlib/Markdown/src/parse/parse.jl
```diff
@@ -8,6 +8,13 @@ mutable struct MD
         new(content, meta)
 end
 
+# For direct messages to the terminal
+struct Message
```
mortenpi: Could we (internally) document here that this is only used by the REPL code and the Markdown parser will never put this into the AST? Then we don't have to update Documenter and other users of the Markdown module to support this node type.

Arguably, the `Message` and the `trimdocs` code could all live in the `REPL` module. Or do you think we should make use of this in Documenter as well?

stdlib/Markdown/src/Markdown.jl
```diff
@@ -60,4 +60,30 @@ import Base.Docs: catdoc
...
+
+function _trimdocs(md::MD, brief::Bool)
+    content, trimmed = [], false
+    for c in md.content
+        if isa(c, Header{1}) && isa(c.text, AbstractArray) &&
+                                lowercase(c.text[1]) == "extended help"
+            trimmed = true
+            break
+        end
+        c, trm = _trimdocs(c, brief)
```
mortenpi: Is this recursion here because sometimes the docstrings get double wrapped in `MD`?

mortenpi (approved):

timholy (commented):
I've moved all the functionality to the REPL. I agree, I think this makes more sense.

fredrikekre:
This is great! Some thoughts:

1. For the case with two methods, but just one of them have extended docs:
   ```
       foo(x)
   
   Short help for `foo(x)`.
   
   # Extended help
   
   Extended help for `foo(x)`.
   """
   foo(x) = x
   
   """
       foo(x, y)
   
   Short help for `foo(x, y)`.
   """
   foo(x, y) = x + y
   ```
   the `Extended help is available with ??` is shown below all docstrings:
   ```
   help?> foo
   
     foo(x)
   
     Short help for foo(x).
   
     ───────────────────────────────
   
     foo(x, y)
   
     Short help for foo(x, y).
   
     ───────────────────────────────
   
   Extended help is available with `??`
   ```
    Should it be attached just to the methods that actually has extended docs? I.e. something like:
   ```
   help?> foo
   
     foo(x)
   
     Short help for foo(x).


     Extended help is available with `??`
     ───────────────────────────────
   
     foo(x, y)
   
     Short help for foo(x, y).
   ```


2. Can we make `??` (literally) work? That way you dont have to type in `?functionname` again (or up arrow, home, ?, enter). Should be pretty easy I think by just saving if the last show docstring had extended docs, and if so also save the method. It also matches what is printed, that extended docs are available with `??`. (Alternatively we could print `Extended docs is available with `??foo`.)

3. Should it be called `Extended documentation` (`Extended docs`?) instead?

4. Since this is now included in the REPL code, it would be nice to change the prompt to `help??>`.


Doesn't have to be adressed in this PR, just some thoughts I had while trying it out.

timholy:
> `Extended help is available with ??` is shown below all docstrings

That was deliberate, in case there was a function with `n` methods and all of them have extended help I was worried it would be annoying to show it each time. It's actually easier to do the thing you suggested, *unless* we also follow your suggestion about "...available with `??foo`" in which case we'd have to come up with a suggestion for invoking each one specifically. That could become awkward if it turns out the type you'd need to specify for `foo(::T)` spanned many lines (e.g., #14946). Since I think being more explicit about how to get longer help is the more important of these suggestions, I decided to leave it so that the message is printed at the end of however many methods are shown.

> Can we make ?? (literally) work?

I like that idea, but a cautionary note:
```julia
help?> ?
search:

  Welcome to Julia 1.3.2-pre.0. The full manual is available at

  https://docs.julialang.org/

  as well as many great tutorials and learning resources:

  https://julialang.org/learning/

  For help on a specific function or macro, type ? followed by its name, e.g.
  ?cos, or ?@time, and press enter. Type ; to enter shell mode, ] to enter
  package mode.
```

I'm a little uneasy giving this special meaning just because it happened to follow a case where there was extended help available.

> Alternatively we could print `Extended docs is available with ??foo`

I went with that suggestion, good idea.

> Should it be called Extended documentation (Extended docs?) instead?

Good idea, I've allowed all three.

> it would be nice to change the prompt to help??>.

I worry this is more subtle than it sounds. What happens when you hit Ctrl-C at the beginning of the line---does it drop back to non-extended help or the julia prompt? (If we just changed the prompt it would drop back to Julia mode.) What happens when you're paging back through your history, does it remember to access it in extended vs non-extended mode? Really I think the only way to do it right is to add a whole new REPL mode. By requiring the `?` at the beginning of the line we get almost all the benefits for one extra character.

fredrikekre:
> > Extended help is available with ?? is shown below all docstrings
> 
> That was deliberate ....

Yea, and I agree the current approach is less spammy.

> > Can we make ?? (literally) work?
> 
> I like that idea, but a cautionary note: ...

But that is just a default fallback, you get the same from just "Enter" at the empty help prompt.

fredrikekre (approved):

c42f:
I love this and I've already started using it after I discovered it a week ago.

There's already a strong convention in Julia documentation that the first part of the short help is *API reference* and that is followed by `# Examples` which are *tutorial content*. These are two very different types of documentation which serve the needs of different users.

The following short blog post is the best and most clarifying thing I've ever read about the categories of documentation: https://www.divio.com/blog/documentation/

What do people think about extending this with more detailed guidelines about the content before and after `# Extended Help`? I also think it's worth considering whether the existing convention of having `# Examples` should be part of this system.

timholy:
Thanks for posting that fantastic article. I'm all in favor of formalizing this a bit more. In the context of the blog post, it seems that my vision of `# Extended help` (which was prompted by docstrings like the one linked in #25930) was mostly about the "explanation" category. The typical Julia convention regarding docstrings seems to start with the "reference" material. In simple cases this seems appropriate, but for methods that have a lot of arguments I could see merit in having a brief summary appear at the top and the full story in `# Extended help` under a `# Reference` section.

In the context of docstrings for individual methods (I'm specifically *not* talking about whole package ecosystems), the line between "tutorial" and "how to" often seems blurry. I think this is specifically because of the ethic that functions like those in Base should generally do one specific small thing, and so one or two examples pretty much shows you what you need to know. When functions start taking a dozen arguments, then "tutorial" and "how to" take different forms.

c42f:
Yes, docstrings tend to start with a terse reference section, and that's often all I want to see. Sometimes I look up the help for, say, `CartesianIndices` (just to pick on one example) and there's all this great detail and examples, but the usage summary which really matters to me has already scrolled off the top of the terminal!

For sure there's unlikely to be something recognizable as a full tutorial or how-to in a docstring. Those tend to need a bigger story which will thread together use of many different functions. Though I wonder whether there's some value in being able to extract the example sections from docstrings to insert into a tutorial in the main docs. It might be tricky because it increases the coupling between the separate docs and the docstrings, but it would make for a super consistent set of examples between related functions.

Anyway, so you're imagining that we could have subsections under extended help? Something like the following perhaps?

```
<terse API reference>

<short examples>

# Extended help
## Reference

<Reference material for large APIs like CSV.FIle>

## Explanation

<Algorithm description, design rational, academic references>

## Examples

<More extended tutorial-like examples>
```

Say we had all this structure; what could the system do with it by default to really brings a lot of value for users?

timholy:
The `??` is clearly a bit too limited to extract those subsections. What about `?r?` (reference), `?d?` (design=explanation), and `?x?` (examples)? So for example `?r?CSV.File` would give the full details of the API. `??` would continue to give all of the subsections.

StefanKarpinski:
I really feel like "short" and "long" modes are all that anyone will remember. Need a quick refresher? Use `?`. Need to see all the gory details? Use `??` and get the full docs in a pager.

tkf:
> For sure there's unlikely to be something recognizable as a full tutorial or how-to in a docstring. Those tend to need a bigger story which will thread together use of many different functions.

I think it'd be better to keep the main focus of docstrings to API references and link materials of the other three quadrants just before `# Extended help`. As you said, tutorials and how-to guides are hard to do with a single function or type. Also, the inclusion of examples does not necessarily mean it's a tutorial or how-to guide. IIRC he was mentioning in the talk that examples for clarifying API is a good fit for API reference too (and I agree).  I think a better approach would be to make REPL `help?>` somehow aware of `docs/` so that you can read the tutorials, how-to guides, and discussions in it. (It's a bit tangential but this way you can write tutorials and how-to guides in markdown. I think it's a better file format for writing non-trivial texts.)

c42f:
> I really feel like "short" and "long" modes are all that anyone will remember.

Agreed; so far been unable to think of a compelling use for extra structure in the docstrings, at least from a REPL point of view.

I'm starting to think that the simple answer is that docstrings should be *reference documentation only*. This is natural for their location interspersed with the source code, where developers only want to see a terse but accurate and complete description.

Aha! And on this point, I finally know why the content of https://github.com/JuliaLang/julia/pull/33936 bothers me. It's tutorial content, but we don't have a place for tutorials.

> Also, the inclusion of examples does not necessarily mean it's a tutorial or how-to guide.

Very true.

> I think a better approach would be to make REPL help?> somehow aware of docs/ so that you can read the tutorials, how-to guides, and discussions in it.

Great idea! If we could allow easy cross references from the docstrings to the tutorial and explanation sections of the manual, I think we would solve our problem. One thing I've worried about in the past was that the source code is not linked to the explanation sections of the Julia manual. So anyone who either
* reads the source only
* looks at docstrings only with `help>` mode

is missing out on a great deal of important content.

tkf:
> So anyone who either
> 
> * reads the source only
> * looks at docstrings only with `help>` mode
> 
> is missing out on a great deal of important content.

I think linking tutorials etc. from the docstring somewhat helps the second kind of readers.  Ideally we can use `TerminalMenus` or something to quickly jump to the relevant section. (Though at this point it feels like we are re-inventing web browser.)  But not sure how to help the first kind of readers. Maybe put URL to the relevant section in the documentation?

c42f:
> But not sure how to help the first kind of readers. Maybe put URL to the relevant section in the documentation?

I feel like just the reference in the docstring is enough as that sits next to the source code. It doesn't need to be super convenient to access, but merely a reminder that there's more content to look for. (Personally I often read the source before looking at the docs, but I assume we don't need to optimize for this case :-) )

Baking in URLs isn't ideal as they refer to a particular hosting location. On that note: it would be cool if `@ref`s in docstrings could resolve to https://pkg.julialang.org/docs/... and especially if there was a way for `@ref`s of dev'd packages to resolve to localhost.

tkf:
> Baking in URLs isn't ideal as they refer to a particular hosting location.

For docstrings, I totally agree. I was thinking more about lengthy inline comments that explain something about surrounding code blocks. It may be useful to link to a discussion/explanation section of the documentation. It would be certainly nice if this link is also re-locatable. But I think it's OK in this case because the audience of the inline comments is very narrow. (Alternatively it makes sense to use relative file paths.)

c42f:
> It may be useful to link to a discussion/explanation section of the documentation.

Agreed. Actually I made an issue about this very thing a while ago, but it didn't generate much discussion and I kind of forgot about it https://github.com/JuliaLang/julia/issues/29527.

StefanKarpinski:
Maybe we could have some way of indicating in a docstring which sections of the docstring should be included in the short versus long docs?

c42f:
> which sections of the docstring should be included in the short versus long docs

Do you mean something like `# Extended docs`, but instead of being just two parts we'd allow long vs short to be interspersed?

One option would be to extend the title convention: for example any heading starting with `Extended` could go into the long docs. This would be super simple and it's appealing to have a pure markdown approach.

Another option would be to consider `extended` a kind of metadata which you don't want to render directly. We already have this kind of thing in the [Documenter `@id` tag](https://juliadocs.github.io/Documenter.jl/stable/man/syntax/index.html#Duplicate-Headers-1). Adding a `@class` annotation like `# (Some Heading)[@class Extended]` would seem to be a natural extension?

StefanKarpinski:
Yes, that's what I meant. It does seem like having them interspersed in the long version might be better and the convention of the name starting with "Extended" is pretty clear.

nalimilan:
I had completely missed this discussion. Very nice! Looking at changes that are needed to use this in DataFrames (https://github.com/JuliaData/DataFrames.jl/pull/2250), I wonder whether we shouldn't treat content appearing after *any* heading as extended help.

In DataFrames, 90% of the time the PR just adds `# Extended help` right above `# Arguments` or `# Examples`. Doing this automatically would ensure some consistency and enable the feature automatically for most packages

timholy:
A lot of docstrings look like [this](https://github.com/JuliaLang/julia/blob/84dd4dda5fb4ba64965d09f1bdfbdda0e51557c1/base/strings/search.jl#L89-L103), I worry that without the example it will become much less clear.

rfourquet:
> I worry that without the example it will become much less clear.

I agree, and even sometimes examples are part of the specification (but that might be a documentation bug...), e.g. in which order `pushfirst(v, 1, 2, 3)` inserts `1, 2, 3` is explicited only in the example.

nalimilan:
I disagree. The short description should be improved if it's not clear/detailed enough to explain the general behavior of the function without looking at examples (even if examples are great to help understanding). In the `findfirst` example, we should arguably mention that `nothing` is returned when no occurrence is found, rather than merely saying it *in a comment* in the examples section.

mgkuhn:
English may be the common language for open-source packages, but I'd expect lots of in-house packages to have docstrings in other languages. So baking in a special meaning for the English phrase “Extended help” seems less than ideal. (Also, I prefer not to refer to documentation as “help”: to me “help” is what you offer to people who haven't read the documentation.) It would be nice to be able to start the extended docstring with arbitrary headings, such as “Details”, “Examples”, “Application hints”, “Further information”, “Gory details”, “Full specification”, “Background”, “Compatibility concerns”, “Performance tips”, etc.

@nalimilan's suggestion to simply treat the first heading as the start of extended help sounds simple and practical. That convention works well e.g. for Wikipedia/Mediawiki, where the convention is that there should be a self-contained short [lead section](https://en.wikipedia.org/wiki/Wikipedia:Manual_of_Style/Lead_section) of the article before the first heading. If that cuts off examples, perhaps that's a good thing if it encourages authors to make the first section so clear that “Examples” sections really becomes what they should be: optional additional information.

timholy:
Your point is a good one but I disagree that we can suppress the current `# Examples` section of many docstrings without significant loss of clarity, or even that it's possible to write such a clear description that you don't need it.

We might need some other syntactic idea here, but at the moment nothing comes to mind.

mgkuhn:
Many Markdown applications allow the use of HTML tags in Markdown-formatted text, so HTML could be a source of suitable syntax for an invisible division marker.

One good example is the “[more tag](https://wordpress.com/support/wordpress-editor/blocks/more-block/2/)” that WordPress uses to separate the introductory part of a blog post (which goes in the front page) from the full blog post:
```
<!--more-->
```
or with an added continuation text:
```
<!--more Click here for full details-->
```
That's a stylized SGML/HTML comment section.

Other ideas from HTML/SGML syntax would be adding an element such as `<more/>` or `<more>Click here for full details</more>`, or an entity reference such as `&more;`.

Also:
In any generated HTML, you want the separation point to be addressable, such that a “Click here for full details” link points the reader at exactly the line where the text continues (like WordPress does). That means the formatter will have to add an `id` attribute to the next HTML element, as in
```
<h2 id="more123">Extended help<h2>
```
However, we can't have the docstring author insert an `id` like that manually: it needs to be unique on a HTML page that may collect several docstrings, so `Documenter.jl` needs to put that in.

PR #24431, opened 2017-11-01
Optimize GC frame push pop insertion

yuyichao:
Try to only insert GC frame push and pop into control flow path that needs the GC frame
while make sure each path has at most one push and one pop.

This is the improved version of https://github.com/JuliaLang/julia/pull/11508, finally.............

PR #34227, opened 2019-12-31
lock: optimize re-entrant performance

vtjnash:
```
julia> r = ReentrantLock(); lock(r); unlock(r);
julia> f(r) = @time for i = 1:10^7; lock(r); unlock(r); end;

before:
julia> f(r)
  0.271201 seconds
julia> lock(r); f(r); unlock(r)
  0.274864 seconds

after:
julia> f(r)
  0.280015 seconds
julia> lock(r); f(r); unlock(r)
  0.149524 seconds
```

This also may need #24431 to show better improvement. Both are also being harmed by incorrect PhiC node placement (will open an issue about this).

### DIFF
--- a/base/lock.jl
+++ b/base/lock.jl
@@ -40,13 +40,14 @@ Each successful `trylock` must be matched by an [`unlock`](@ref).
 """
 function trylock(rl::ReentrantLock)
     t = current_task()
+    if t === rl.locked_by
+        rl.reentrancy_cnt += 1
+        return true
+    end
     lock(rl.cond_wait)
     if rl.reentrancy_cnt == 0
         rl.locked_by = t
         rl.reentrancy_cnt = 1
-        got = true
-    elseif t === notnothing(rl.locked_by)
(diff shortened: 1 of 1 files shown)


JeffBezanson (approved):

yuyichao:
I think this is currently undefined behavior.

vtjnash:
What part?

yuyichao:
The part that's accessing shared field without lock. The read (and the corresponding store) of the `locked_by` field must be atomic (relaxed order).

vtjnash:
`unlock` usually should contain the memory barrier to make that unnecessary

yuyichao:
No the barrier always works in pairs. The unlock can have all the barrier you can possibly implement and it won't make the unprotected access in lock any safer.

vtjnash:
> The read (and the corresponding store) of the locked_by field must be atomic (relaxed order).

Now implemented in #36507

PR #32653, opened 2019-07-22
add CartesianIndex() for Ref getindex and setindex!

bkamins:
See https://discourse.julialang.org/t/indexing-with-cartesianindex/26615/1 for a discussion of the issue.

CC @mbauman

PR #34228, opened 2019-12-31
Support broadcast assignment to Ref objects

jmert:
This is a pretty straight-forward extension of adding `CartesianIndex()` access in #32653 — I was motivated by trying to develop code which exploits broadcasting to operate on anything from scalars to multidimensional arrays using `CartesianIndices` indexing. This doesn't work for a `Ref`-wrapped scalar without these definitions, though.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -23,6 +23,7 @@ New library functions
 
 New library features
 --------------------
+- `Ref` objects are now valid assignment locations in broadcast expressions ([#34228])
 
 
 Standard library changes
--- a/NEWS.md
+++ b/NEWS.md
@@ -23,6 +23,8 @@ New library functions
 
 New library features
 --------------------
+- `RefArray` and `RefValue` objects now accept index
+  `CartesianIndices(())` in `getindex` and `setindex!` ([#34228])
 
 
 Standard library changes
--- a/NEWS.md
+++ b/NEWS.md
@@ -23,8 +23,6 @@ New library functions
 
 New library features
 --------------------
-- `RefArray` and `RefValue` objects now accept index
-  `CartesianIndices(())` in `getindex` and `setindex!` ([#34228])
 
 
 Standard library changes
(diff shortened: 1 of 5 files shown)


mbauman:
If the analogy is that `Ref`s are like 0-dimensional arrays, then shouldn't this just return a `Ref` right back at you?

```Julia
julia> A = fill(1)
0-dimensional Array{Int64,0}:
1

julia> A[CartesianIndices(())]
0-dimensional Array{Int64,0}:
1
```

Or perhaps more to the point, if we support this, I think we should support all `AbstractArray{CartesianIndex{0},0}`s.  But I'd be very hesitant to do either.

jmert:
Yes, it seems you're right — dropping the `Ref` wrapper isn't what I want. With `getindex` unwrapping, I hit a new error in broadcasting, and it's the same as if the code assumes the left hand side is a `Ref`:

```julia
julia> g1!(x) = x[] .= 1.0
g1! (generic function with 1 method)

julia> g1!(Ref(1.0))
ERROR: MethodError: no method matching copyto!(::Float64, ::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{0},Tuple{},typeof(identity),Tuple{Float64}})
...
```

Maybe I should poke at the broadcasting behavior instead?

jmert:
It looks like if I instead define:
```julia
@eval Base begin
    maybeview(A::Ref, ::CartesianIndex{0}) = A
    maybeview(A::Ref, ::CartesianIndices{0,Tuple{}}) = A
end
@eval Base.Broadcast begin
    @inline function copyto!(dest::Ref, bc::Broadcasted{<:AbstractArrayStyle{0}})
        dest[] = materialize(bc)
    end
end
```
I can get the behavior I wanted. Is this a better idea? I suppose the alternative is to replicate a `Ref`-like struct and add similar definitions to it instead.

PR #28328, opened 2018-07-28
Add a callback so packages can correct line numbers for "moving code"

timholy:
One of the most longstanding [annoyances with Revise](https://github.com/timholy/Revise.jl/issues/51) is that heavy editing of source files can cause line numbers to become wildly inaccurate. Rather than recompile all the code just to update the line numbers, this PR allows one to register a callback to correct the line numbers before showing the stacktrace to the user. Together with a [heavily-redesigned Revise](https://github.com/timholy/Revise.jl/pull/116), initial local testing suggests this strategy is effective.

While the necessary changes to Revise were extensive, this PR is pretty modest.

PR #34235, opened 2020-01-02
Profile: add line number correction (Revise)

timholy:
This ensures that Profiling will report the current line number in interactively-modified code.

It looks like #28328 failed to include a test, so I also added that here.


StefanKarpinski:
Who would be good to review?

timholy:
Probably @vtjnash, who has done a lot with the profiler lately.

vtjnash (approved):

Issue #34216, opened 2019-12-30
Default RNGs in threads slowdown performance in Julia 1.3


eisthf:
Default RNGs in threads slowdown performance in Julia 1.3.

Here is my environment:
> julia> versioninfo(verbose=true)
> Julia Version 1.3.0
> Commit 46ce4d7933 (2019-11-26 06:09 UTC)
> Platform Info:
>   OS: Windows (x86_64-w64-mingw32)      
>       Microsoft Windows [Version 10.0.18362.535]
>   CPU: Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz:
>               speed         user         nice          sys         idle          irq
>        #1  4008 MHz     633296            0       811953     12463687       195718  ticks
>        #2  4008 MHz     423671            0       328500     13156765         6218  ticks
>        #3  4008 MHz     935390            0       638296     12335250         6562  ticks
>        #4  4008 MHz     436546            0       558328     12913859         2953  ticks
>        #5  4008 MHz     772937            0       654812     12480984         5468  ticks
>        #6  4008 MHz     499781            0       503546     12905406         3000  ticks
>        #7  4008 MHz     721312            0       661609     12525812         3750  ticks
>        #8  4008 MHz     578609            0       338343     12991781         2156  ticks
> 
>   Memory: 15.953498840332031 GB (8580.15625 MB free)
>   Uptime: 13908.0 sec
>   Load Avg:  0.0  0.0  0.0
>   WORD_SIZE: 64
>   LIBM: libopenlibm
>   LLVM: libLLVM-6.0.1 (ORCJIT, skylake)
> Environment:
>   JULIA_EDITOR = "C:\Users\Earth\AppData\Local\atom\app-1.42.0\atom.exe"  -a
>   JULIA_NUM_THREADS = 8

In Julia 1.3 each thread has its own default global RNG. Calling rand() in each thread is safe. So the following code that calculates pi returns correct result:

```
function par_pi(n::Int)
    hits = zeros(Int, nthreads())
    @threads for i in 1:n
        x, y = rand(), rand()
        hits[threadid()] += (x^2 + y^2 <= 1)
    end

    4.0 * sum(hits) / n
end
@btime par_pi(10_000_000) #102.228 ms (59 allocations: 7.00 KiB)
```

But the above code runs 4x slower than the following version that uses manually allocated RNGs:

```
const threadsRNG = [MersenneTwister() for i in 1:nthreads()]

function par_pi2(n::Int)
[...]

PR #34237, opened 2020-01-02
outline length assert in default_rng(tid)

KristofferC:
Ref https://github.com/JuliaLang/julia/issues/34216

Benchmark

```jl
# export JULIA_NUM_THREADS=4
using BenchmarkTools, Random, Base.Threads

function par_pi(n::Int)
    hits = zeros(Int, nthreads())
    @threads for i in 1:n
        x, y = rand(), rand()
        hits[threadid()] += (x^2 + y^2 <= 1)
    end

    4.0 * sum(hits) / n
end
@btime par_pi(10_000_000)
```

master:

```
168.895 ms (24 allocations: 2.83 KiB)
```

PR:

```
 83.366 ms (24 allocations: 2.83 KiB)
```

### DIFF
--- a/stdlib/Random/src/RNGs.jl
+++ b/stdlib/Random/src/RNGs.jl
@@ -295,7 +295,7 @@ seed!(seed::Union{Integer,Vector{UInt32}}) = seed!(default_rng(), seed)
 const THREAD_RNGs = MersenneTwister[]
 @inline default_rng() = default_rng(Threads.threadid())
 @noinline function default_rng(tid::Int)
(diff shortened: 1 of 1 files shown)


vtjnash (approved):

vtjnash:
Good find! I thought I had already beat on this as much as was possible.

jebej:
Is the outlining trick for fixing performance when something throws an exception something that will be fixed? It seems unsustainable to do this in every single case.

Keno:
It's a micro-optimization that doesn't usually matter. That said, it would be nice to do this automatically eventually.

PR #34241, opened 2020-01-02
[GCLowering] correctly handle GC.preserve for demoted objects

vtjnash:


Keno (approved):
This looks fine to me assuming NumberAll deals with FCAs correctly, which I assume it does otherwise we'd have bigger problems.

vtjnash:
Yeah, in fact it should handle all pointers (including the scalar case) but I haven't gotten a round to deleting the `Number` set of functions

PR #34242, opened 2020-01-02
BinaryBuilder dependencies bump

staticfloat:
This PR contains a number of BinaryBuilder-related improvements:

* Upgrade the BB tarball naming methodology from the flawed `-gccX` tags to the much more explicit `-libgfortranX` tags that BB has been spitting out for over half a year now.  This necessitates changing `Openlibm`, `OpenBLAS` and `LLVM` at the least, as all of these dependencies were marked as having a dependency on GCC version/libgfortran version.
* Upgrade `OpenLibm` to a new release that happily removes the incorrect dependence on `libgfortran`.
* Upgrade `OpenBLAS` to `v0.3.7`, including the new threading patches from @Keno.
* Re-name the `LLVM` checksums to honor the upgraded naming convention.


staticfloat:
And because I often screw this up, here is a build with no BB dependencies: https://build.julialang.org/#/builders/147/builds/7

staticfloat:
Hmmm, win32 has a signal 11 (is that still segfault on windows?) when just trying to run Julia to print out version info.  That's not good.

Keno:
I'm seeing even the shell segfault: https://build.julialang.org/#/builders/31/builds/8176/steps/3/logs/stdio. Did the BinaryBuilder download overwrite something?

JeffBezanson:
This is great but it's not clear to me whether it's backportable --- maybe we should merge #34225 for now? Or perhaps only on release-1.4?

staticfloat:
> Did the BinaryBuilder download overwrite something?

Unsure; I looked into the logs and it seems that a lot of the builders have been segfaulting.  I've rebooted them to see if that fixes it, if not I'll inspect the bots a bit, and if I can't figure it out quickly, I'll just rebuild them.

staticfloat:
To play it safe, I think I should build a special 0.3.5 release of OpenBLAS with Keno's patches, integrate those into https://github.com/JuliaLang/julia/pull/34225, land those on 1.3/1.4, and this can go into `master`.

Issue #32414, opened 2019-06-25
Wrong GEP emission for tuple containing VecElement


chriselrod:
```julia
julia> @noinline function foo(a)
           v = ntuple(Val(8)) do w Core.VecElement(Float64(10w)) end
           a, (v, (a,(1e6,1e9)))
       end
foo (generic function with 1 method)

julia> foo(-35.0)
(10.0, ((VecElement{Float64}(30.0), VecElement{Float64}(40.0), VecElement{Float64}(50.0), VecElement{Float64}(60.0), VecElement{Float64}(70.0), VecElement{Float64}(80.0), VecElement{Float64}(-35.0), VecElement{Float64}(1.0e6)), (1.0e9, (1.2353556789225747e-307, 6.9513596813479e-310))))

julia> versioninfo()
Julia Version 1.3.0-DEV.428
Commit 863e03fbb1* (2019-06-19 15:07 UTC)
Platform Info:
  OS: Linux (x86_64-generic-linux)
  CPU: Intel(R) Core(TM) i9-7900X CPU @ 3.30GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-8.0.0 (ORCJIT, skylake)
```
The answer should be:
```julia
(-35.0, ((VecElement{Float64}(10.0), VecElement{Float64}(20.0), VecElement{Float64}(30.0), VecElement{Float64}(40.0), VecElement{Float64}(50.0), VecElement{Float64}(60.0), VecElement{Float64}(70.0), VecElement{Float64}(80.0), VecElement{Float64}(-35.0), VecElement{Float64}(1.0e6)), (-35.0, (1.0e6, 1.0e9))))
```
That is, the incorrect answer we received is shifted to the left by 16 bytes, and there appears to be an 8-byte gap between the first float and the following vector.

If we replace the tuple of `VecElement`s with a standard tuple, that is what we get:
```julia
julia> @noinline function bar(a)
           v = ntuple(Val(8)) do w Float64(10w) end
           a, (v, (a,(1e6,1e9)))
       end
bar (generic function with 1 method)

julia> bar(-35.0)
(-35.0, ((10.0, 20.0, 30.0, 40.0, 50.0, 60.0, 70.0, 80.0), (-35.0, (1.0e6, 1.0e9))))
```

This silently corrupts data -- not the easiest problem to track down!

[...]

PR #34245, opened 2020-01-03
codegen: on alignment incongruence, replace assert by jl_error

tkluck:
It is possible that LLVM and Julia disagree on the offsets of struct
fields. This has been observed in #32414 in cases like

    Tuple{Float64, NTuple{8, VecElement{Float64}}}

where LLVM wants to align the `vec` on 64 bytes (adding 56 bytes
of padding after the first `Float64`) but Julia refusing to
align like that because it can't guarantee that on the heap:

https://github.com/JuliaLang/julia/blob/c7e9d9f8ade647582c6635c8c9a587f2360af404/src/datatype.c#L453

There is no easy solution this I can see, but since this can be
triggered from pure Julia, the least we can do is not abort the program.

For this reason, this commit replaces the assert by jl_error.
To be able to give a sort-of meaningful error message, we return
a sentinel value from `convert_struct_offset` and generate
an error message in the caller.

### DIFF
--- a/src/cgutils.cpp
+++ b/src/cgutils.cpp
@@ -511,7 +511,10 @@ static unsigned convert_struct_offset(Type *lty, unsigned byte_offset)
     const DataLayout &DL = jl_data_layout;
     const StructLayout *SL = DL.getStructLayout(cast<StructType>(lty));
     unsigned idx = SL->getElementContainingOffset(byte_offset);
-    assert(SL->getElementOffset(idx) == byte_offset);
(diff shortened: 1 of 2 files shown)


tkluck:
@ararslan thanks for having a look and requesting a review. I just pushed a simple test per your request.

tkluck:
Closing as there is now a complete fix in the form of https://github.com/JuliaLang/julia/pull/34473

Issue #28192, opened 2018-07-19
Coverage misses lines


iamed2:
For this file: https://github.com/invenia/Memento.jl/blob/master/src/stdlib.jl

The coverage report is:
```
        - if VERSION > v"0.7.0-DEV.2980"
        -     import Base.CoreLogging:
        -         AbstractLogger,
        -         handle_message,
        -         min_enabled_level,
        -         shouldlog,
        -         global_logger,
        -         Debug
        - 
        -     const LEVEL_MAP = Dict{AbstractString}
        -     struct CoreLogger <: AbstractLogger
        -     end
        - 
        -     min_enabled_level(logger::CoreLogger) = Debug
        4     shouldlog(logger::CoreLogger, arags...) = true
        - 
        -     function handle_message(::CoreLogger, cl_level, msg, mod, group, id, filepath, line; kwargs...)
        2         logger = getlogger(mod)
        -         level = lowercase(string(cl_level))
        2         log(logger, logger.record(logger.name, level, getlevels(logger)[level], msg))
        -     end
        - 
        -     function substitute!()
        1         global_logger(CoreLogger())
        -         notice(getlogger(@__MODULE__), "Substituting global logging with Memento")
        -     end
        - else
        -     function substitute!()
        -         warn(
        -             getlogger(@__MODULE__),
        -             "Global logging substitution is not support for julia $VERSION"
        -         )
        -     end
        - end
        - 
```

Notice that `level = lowercase(string(cl_level))` is between two covered lines but somehow isn't reported as covered.

vtjnash:
Now is perhaps a good time to update LineInfoNode to track line-number ranges in the parser (but might be hard still for the optimizer to handle updating this information). And see also https://github.com/JuliaLang/julia/pull/11802/files

Perhaps we should inject this information before optimization (like we are doing at the llvm level), as this'll problem will only get worse as our optimizer gets better. We could more-easily do it then at the basic-block level. This would be more likely to require recompiling the system image to switch the flag though.

maxbennedich:
We are seeing the same thing starting in Julia 0.7. Every Julia 0.7 and 1.0 project with test coverage I've seen suffer from the same thing. I've tried with `--inline=no` but it's still incorrect for us.

sbromberger:
I don't know whether this is related, but LightGraphs just dropped to from 99.8% to 66% codecov (and the reported coverage gaps don't make sense).

matteoacrossi:
[...]

PR #34254, opened 2020-01-03
fix codecov accuracy for optimized and dead statements

vtjnash:
As people have noted (e.g https://github.com/JuliaLang/julia/issues/28192), Keno made the optimizer much too good. Rather than making the optimizer worse, we instead here preprocess the IR to inject a weak compiler barrier at each line change. This ensures we preserve the line number record, even if we drop all other side-effects at that point.

### DIFF
--- a/base/compiler/ssair/driver.jl
+++ b/base/compiler/ssair/driver.jl
@@ -49,20 +49,37 @@ function normalize(@nospecialize(stmt), meta::Vector{Any})
     return stmt
 end
 
-function just_construct_ssa(ci::CodeInfo, code::Vector{Any}, nargs::Int, sv::OptimizationState)
+function convert_to_ircode(ci::CodeInfo, code::Vector{Any}, coverage::Bool, nargs::Int, sv::OptimizationState)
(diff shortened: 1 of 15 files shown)


base/compiler/ssair/driver.jl
```diff
@@ -49,28 +49,45 @@ function normalize(@nospecialize(stmt), meta::Vector{Any})
...
     idx = 1
     oldidx = 1
     changemap = fill(0, length(code))
+    labelmap = coverage ? fill(0, length(code)) : changemap
+    prevloc = zero(eltype(ci.codelocs))
     while idx <= length(code)
+        codeloc = ci.codelocs[idx]
+        if coverage && codeloc != prevloc && codeloc != 0
+            # insert a side-effect instruction before the current instruction in the same basic block
+            insert!(code, idx, Expr(:code_coverage_marker))
```
JeffBezanson: `:code_coverage_effect`?

vtjnash: :sob:

PR #34255, opened 2020-01-03
handle lack of uv_os_get_passwd

vtjnash:
Ran into this problem on an unusual machine configuration (sssd PAM running x86 binaries on an amd64 OS) where Julia was failing to run the cmdlinearg tests (which I needed to run to check my PR). Turned out that getpwuid_r was unavailable for the current user. Working around that by checking return value and ensuring HOME is set in these tests. Note, Julia may still fail other tests (notably, the `path` test for the homedir functionality), and is broken at startup with the following error:
```
$ env -u HOME ./usr/bin/julia
fatal: error thrown and no exception handler available.
#<null>
uv_error at ./libuv.jl:97 [inlined]
uv_error at ./libuv.jl:96 [inlined]
homedir at ./path.jl:75
append_default_depot_path! at ./initdefs.jl:78
init_depot_path at ./initdefs.jl:100
__init__ at ./Base.jl:411
jfptr___init___13626 at /data/vtjnash/julia/build-x86/usr/lib/julia/sys.so (unknown line)
_jl_invoke at /data/vtjnash/julia/src/gf.c:2144 [inlined]
jl_apply_generic at /data/vtjnash/julia/src/gf.c:2328
jl_apply at /data/vtjnash/julia/src/julia.h:1692 [inlined]
jl_module_run_initializer at /data/vtjnash/julia/src/toplevel.c:74
_julia_init at /data/vtjnash/julia/src/init.c:800
julia_init__threading at /data/vtjnash/julia/src/task.c:248
main at /data/vtjnash/julia/ui/repl.c:211
__libc_start_main at /lib/i386-linux-gnu/libc.so.6 (unknown line)
_start at ./usr/bin/julia (unknown line)
```

### DIFF
--- a/src/runtime_ccall.cpp
+++ b/src/runtime_ccall.cpp
@@ -136,18 +136,21 @@ std::string jl_format_filename(StringRef output_pattern)
     for (auto c : output_pattern) {
         if (special) {
             if (!got_pwd && (c == 'i' || c == 'd' || c == 'u')) {
-                uv_os_get_passwd(&pwd);
-                got_pwd = true;
+                int r = uv_os_get_passwd(&pwd);
+                if (r == 0)
+                    got_pwd = true;
             }
             switch (c) {
             case 'p':
                 outfile << jl_getpid();
                 break;
             case 'd':
-                outfile << pwd.homedir;
+                if (got_pwd)
+                    outfile << pwd.homedir;
                 break;
             case 'i':
-                outfile << pwd.uid;
+                if (got_pwd)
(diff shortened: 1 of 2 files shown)


Keno:
This isn't really a kernel bug. The kernel doesn't handle passwd at all - that's all in libc. The issue on the machine in question is that the 32bit version of sssd isn't installed. The libc provides an interface for identity providers to provide `.so` plugins that then handle the specific passwd request. Since the 32bit version of sssd isn't installed, those requests fail.

PR #34218, opened 2019-12-30
Set master version to 1.5

ararslan:
VERSION has been set to 1.5.0-DEV, the 1.4 NEWS items have been moved to HISTORY, and NEWS has been cleared out for 1.5 items.

Please don't merge anything to master until this is merged so that versions don't get messed up, since 1.4 has been branched.

PR #34260, opened 2020-01-05
Use correct version for master docs

### LINKED ISSUE #34259: Documentation issue: 1.4-dev page showing 1.5 version
image

mortenpi:
Ideally, it should have been updated in #34218. Fix #34259.

cc @fredrikekre


vtjnash:
Can you add this to the release checklist also in the toplevel Makefile?

mortenpi:
I'd say that this is good to go.

Also, just for the record, this could probably be automated by taking these values from the `VERSION` file. But I'd leave that as up for grabs for someone else at the moment.

PR #33741, opened 2019-11-01
RFC: for the release, change tempfile cleanup default back to false

vtjnash:
This seems to be exacerbating Windows CI problems such as #31810 and failing its own test. Disable it by default for this release (e.g. reverts 75cb00604e), so we can get this into the next RC, and explore fixes on master.

PR #34273, opened 2020-01-06
Bump OpenBLAS BB release to include win32 threading patch

staticfloat:
This adds OpenBLAS v0.3.5 BB tarballs to #34225, as discussed in https://github.com/JuliaLang/julia/pull/34225#issuecomment-570731201


Keno:
Good to go?

vchuravy:
The win64 failures look relevant since it is LinearAlgebra that is failing.

Keno:
Missing the Skylake avx patch? @staticfloat

staticfloat:
@Keno does this win64 failure look like openblas is hanging on exit?

Keno:
No that looks like an unrelated tempfile error we've seen before

vtjnash:
Ref https://github.com/JuliaLang/julia/pull/33741

staticfloat:
I'm going to re-roll the CI dice a few times to see if we can get a green checkmark then.

ViralBShah:
Got the green checkmark

PR #34277, opened 2020-01-06
Add xrefs and some missing example headers

kshyatt:


stdlib/LinearAlgebra/src/lq.jl
```diff
@@ -67,15 +67,15 @@ LQPackedQ(factors::AbstractMatrix{T}, τ::Vector{T}) where {T} = LQPackedQ{T,typ
 """
     lq!(A) -> LQ
 
-Compute the LQ factorization of `A`, using the input
+    Compute the [`LQ`](@ref) factorization of `A`, using the input
```
dkarrasch: The 4-spaces shift is probably an accident.

dkarrasch (commented):
Just one typo, otherwise LGTM. I'm not sure the xref to `StridedMatrix` is really helpful, and whether it has an entry in the docs in the first place.

stdlib/LinearAlgebra/src/lq.jl
kshyatt: Thanks!

kshyatt:
I checked, `StridedMatrix` does have a doc entry but it's pretty short.

dkarrasch (approved):

StefanKarpinski:
This good to merge? Go for it when you're ready, @kshyatt.

PR #34280, opened 2020-01-06
Some missing xrefs for Threads

kshyatt:


base/threadingconstructs.jl
```diff
@@ -75,7 +75,7 @@ end
 """
     Threads.@threads
 
-A macro to parallelize a for-loop to run with multiple threads. This spawns `nthreads()`
+A macro to parallelize a for-loop to run with multiple threads. This spawns [`nthreads()`](@ref)
```
dkarrasch: Is this going to work without the `Threads` specifier? Because `nthreads` is not exported. Not sure if this is tested on the buildbots...

kshyatt: `make docs` worked locally...

PR #34284, opened 2020-01-06
add a compatibility fallback for `divrem` and `fldmod`

JeffBezanson:
Fixes test failures in ~AbstractNumbers.jl~ and CurrenciesBase/FixedDecimals.


JeffBezanson:
Ok, that didn't work since some non-number types like Periods also implement this. I guess we'll have to sacrifice `applicable`? (AbstractNumbers was using it to determine which functions to test on which types.)

PR #34289, opened 2020-01-06
add some compatibility shims for internal functions changed in 1.4

JeffBezanson:


vtjnash (approved):
LGTM. Can you mentioned "deprecated" in a comment somewhere near each so we can find the cleanup to address in e.g. v2.0?

PR #33526, opened 2019-10-11
Transducer as an optimization: map, filter and flatten

tkf:
This PR implements very minimal (stateless) transducers in Base and use it internally in `foldl` (`mapfoldl`).  My intention is to make transducer-based code _completely an implementation detail_ and so that code using iterators can just become faster.  Here is a benchmark:

```julia
using BenchmarkTools
xs = [abs(x) < 1 ? x : missing for x in randn(1000)]
@btime foldl(+, (x for x in $xs if x !== missing))
```

* Before (a68237f9c9805d492fc2d483bf63b8ecba647e8f):  1.484 μs (9 allocations: 144 bytes)
* After (fb89009b655261ef45b521df40f7def83b5bd482): 692.771 ns (2 allocations: 32 bytes)
* `@btime sum(skipmissing($xs))`: 881.146 ns (5 allocations: 80 bytes)

This benchmark shows that transducer-based optimization can yield 2x speed up for a simple filter.  Also, it's "faster" than `skipmissing` although it does pair-wise summation so this is not really a fair comparison.  However, transducer-based optimization can also be applied to `reduce` which then would eliminate the need for special `mapreduce_impl` implementation (ref #27743).  Furthermore, similar speed up may happen for _any_ type-based filtering, not just `missing`.

(By the way, specialized `mapreduce_impl` for `skipmissing` seems to be required due to the requirement for hard-coding `=== missing` according to this comment https://github.com/JuliaLang/julia/issues/27681#issuecomment-399054971 by @Keno.  However, this seems to be not applied to transducers.  Why is this the case?  This is puzzling me for a while.)

I think I can optimize it more but I think this is a good starting point where code is still simple and so easy to review.  <strike>Other "stateless" iterator transforms like `Flatten` should be easy to support.</strike>

### tl;dr

* Transducers make `foldl` combined with iterator comprehensions (`Iterator.filter` and `Generator`) faster.
* <strike>Full comprehension support is easy (i.e., adding `Flatten`).</strike> (Edit: Flatten is supported in this PR now)
* Some `skipmissing`-specific code can be removed (maybe).

What do people think about this?  Is it worth using simple transducers in Julia Base?

### Implementation

Quoting the docstring, the central idea is implemented in `_xfadjoint` which (roughly speaking) converts iterators to transducers:

>     _xfadjoint(op, itr) -> op′, itr′
>
[...]

PR #34293, opened 2020-01-07
Improve foldl with tail-call function-barrier

tkf:
This PR improves `foldl` for type-based filtering by using the function-barrier at tail-call positions.  This improves the performance of the benchmark (equivalent to what) I mentioned in https://github.com/JuliaLang/julia/pull/33526

```julia
using BenchmarkTools
xs = [abs(x) < 1 ? x : missing for x in randn(1000)];
@btime sum(x for x in $xs if x !== missing)
```

* Before (589a6d891c89d0e6b03da4a520dfe0318e0b9d56): 1.222 μs (0 allocations: 0 bytes)
* After (14839b2e7ae1ec37d4f3166e46a2d10c304d0b13): 680.954 ns (0 allocations: 0 bytes)

This benchmark is included in https://github.com/JuliaCI/BaseBenchmarks.jl/pull/254

To be honest, I am not 100% sure this is the best approach.  Maybe this is too much of a strain to the compiler?  Should we care about generated code size?  Does x1.8 speedup justify that?  As some versions of Julia compiler was able to generate a comparable code, is it better to tweak the compiler to recover the performance?

### Implementation notes

As I discussed [in this discourse thread](https://discourse.julialang.org/t/tail-call-optimization-and-function-barrier-based-accumulation-in-loops/25831), using this trick in generic `foldl` is harder than in implementations of `collect`-like functions.  This is because it is easy to have stack overflow if you don't have some kind of monotonicity in the way the accumulator changes its type.  This is why I have the counter at type-level.  At the moment the type can be refined only up to three times (`counter = Val(3)`) which is arbitrary but I think conservative.

Aside: it would be great if Julia compiler generates a finite state machine from the tail calls as Guy Steele argued in the talk I linked in the discourse thread.

### DIFF
--- a/base/reduce.jl
+++ b/base/reduce.jl
@@ -47,18 +47,29 @@ function foldl_impl(op::OP, nt, itr) where {OP}
     return v
 end
 
-function _foldl_impl(op::OP, init, itr) where {OP}
-    # Unroll the while loop once; if init is known, the call to op may
-    # be evaluated at compile time
+function _foldl_impl(op::OP, init::T, itr) where {OP, T}
     y = iterate(itr)
     y === nothing && return init
     v = op(init, y[1])
-    while true
-        y = iterate(itr, y[2])
-        y === nothing && break
-        v = op(v, y[1])
+    if v isa T
+        return _foldl_impl(op, v, itr, y[2])
+    else
+        return _foldl_impl(op, v, itr, y[2])
     end
-    return v
+end
+
+_dec(::Val{n}) where {n} = Val(n - 1)
+
+@inline function _foldl_impl(op::OP, acc::T, itr, state, counter = Val(3)) where {OP,T}
+    while true
+        y = iterate(itr, state)
+        y === nothing && return acc
+        x, state = y
+        result = op(acc, x)
+        counter === Val(0) || result isa T ||
+            return _foldl_impl(op, result, itr, state, _dec(counter))
+        acc = result
+    end
 end
 
 struct _InitialValue end


base/reduce.jl
```diff
@@ -47,18 +47,29 @@ function foldl_impl(op::OP, nt, itr) where {OP}
...
-    # Unroll the while loop once; if init is known, the call to op may
-    # be evaluated at compile time
+function _foldl_impl(op::OP, init::T, itr) where {OP, T}
     y = iterate(itr)
     y === nothing && return init
     v = op(init, y[1])
+    if v isa T
+        return _foldl_impl(op, v, itr, y[2])
+    else
+        return _foldl_impl(op, v, itr, y[2])
```
ararslan: Isn't this exactly the same as what's done in the other branch of the conditional? :thinking:

tkf: Yes, but it helps the compiler.  Removing this branch introduces an allocation and makes the code a bit slower:

```julia
julia> @eval Base function _foldl_impl(op::OP, init::T, itr) where {OP, T}
           y = iterate(itr)
           y === nothing && return init
           v = op(init, y[1])
           return _foldl_impl(op, v, itr, y[2])
       end
_foldl_impl (generic function with 4 methods)

julia> @btime sum(x for x in $xs if x !== missing)
  780.859 ns (1 allocation: 16 bytes)
```

But the speedup is not as drastic as I felt while I was playing with it. So I'm OK with removing this micro-optimization.

tkf: Actually, as this "no-op `if` branch" gets rid of a constant-time work, the difference becomes large if you consider shorter arrays:

```julia
julia> xs = [abs(x) < 1 ? x : missing for x in randn(10)];

julia> @btime sum(x for x in $xs if x !== missing)
  8.405 ns (0 allocations: 0 bytes)
1.5696312630017393

julia> @eval Base function _foldl_impl(op::OP, init::T, itr) where {OP, T}
           y = iterate(itr)
           y === nothing && return init
           v = op(init, y[1])
           return _foldl_impl(op, v, itr, y[2])
       end
_foldl_impl (generic function with 4 methods)

julia> @btime sum(x for x in $xs if x !== missing)
  29.120 ns (1 allocation: 16 bytes)
1.5696312630017393
```

ararslan: That is incredibly bizarre. :flushed:

tkf: I thought I was just doing union splitting manually.

adienes:
while I'm sure there is room for improvement in `foldl`, the diff here rebased on nightly appears to no longer constitute a performance improvement

PR #34295, opened 2020-01-07
use ssh multiplexing to create SSH workers with tunneling

s417-lama:
The current implementation of `Distributed.addprocs` with `tunnel=true` had problems:
- If a multiplexing process had been already launched, the creation of new ssh workers hung.
- It required two separate connections, which needed two times of authentication.

These problems can be solved by using ssh multiplexing.
I modified the code to use ssh multiplexing when `tunnel` is true, and it has some advantages to the current one:

- The ssh commands got much simpler
    - by using control commands (`-O` options) of ssh
- Reuse a multiplexing session if it has already been established outside of Julia.
    - which reduces overhead and avoids authentication at each run

### DIFF
--- a/doc/src/manual/parallel-computing.md
+++ b/doc/src/manual/parallel-computing.md
@@ -1700,13 +1700,12 @@ requirements for the inbuilt `LocalManager` and `SSHManager`:
     via `sshflags`, for example ```sshflags=`-i <keyfile>` ```.
 
     Internally, SSH multiplexing is used to create a tunnel between the master and workers. If you have
-    been configured SSH multiplexing on your own and the connection has already been established,
-    forwarding is set by using the existing connection (`-O forward` option in ssh). This is beneficial
-    if your servers require password authentication; you can avoid authentication in Julia by logging
-    in to the server ahead of `addprocs`. Even if you do not explicitly configure SSH multiplexing,
-    Julia will use multiplexing to reduce the number of authentication. The control socket will be
-    located at `~/.ssh/julia-%r@%h:%p` during the session unless the existing multiplexing connection
-    is used.
+    configured SSH multiplexing on your own and the connection has already been established, forwarding
+    is set by using the existing connection (`-O forward` option in ssh). This is beneficial if your
+    servers require password authentication; you can avoid authentication in Julia by logging in to the
+    server ahead of `addprocs`. Even if you do not explicitly configure SSH multiplexing, Julia will
+    use multiplexing to reduce the number of authentication. The control socket will be located at
+    `~/.ssh/julia-%r@%h:%p` during the session unless the existing multiplexing connection is used.
 
     In an all-to-all topology (the default), all workers connect to each other via plain TCP sockets.
     The security policy on the cluster nodes must thus ensure free connectivity between workers for
--- a/doc/src/manual/parallel-computing.md
+++ b/doc/src/manual/parallel-computing.md
@@ -1699,6 +1699,15 @@ requirements for the inbuilt `LocalManager` and `SSHManager`:
     authenticated via public key infrastructure (PKI). Authentication credentials can be supplied
     via `sshflags`, for example ```sshflags=`-i <keyfile>` ```.
 
+    Internally, SSH multiplexing is used to create a tunnel between the master and workers. If you have
+    been configured SSH multiplexing on your own and the connection has already been established,
+    forwarding is set by using the existing connection (`-O forward` option in ssh). This is beneficial
+    if your servers require password authentication; you can avoid authentication in Julia by logging
+    in to the server ahead of `addprocs`. Even if you do not explicitly configure SSH multiplexing,
+    Julia will use multiplexing to reduce the number of authentication. The control socket will be
+    located at `~/.ssh/julia-%r@%h:%p` during the session unless the existing multiplexing connection
+    is used.
+
     In an all-to-all topology (the default), all workers connect to each other via plain TCP sockets.
     The security policy on the cluster nodes must thus ensure free connectivity between workers for
     the ephemeral port range (varies by OS).
--- a/doc/src/manual/parallel-computing.md
+++ b/doc/src/manual/parallel-computing.md
@@ -1699,20 +1699,22 @@ requirements for the inbuilt `LocalManager` and `SSHManager`:
     authenticated via public key infrastructure (PKI). Authentication credentials can be supplied
     via `sshflags`, for example ```sshflags=`-i <keyfile>` ```.
 
-    Internally, SSH multiplexing is used to create a tunnel between the master and workers. If you have
-    configured SSH multiplexing on your own and the connection has already been established, forwarding
-    is set by using the existing connection (`-O forward` option in ssh). This is beneficial if your
-    servers require password authentication; you can avoid authentication in Julia by logging in to the
-    server ahead of `addprocs`. Even if you do not explicitly configure SSH multiplexing, Julia will
-    use multiplexing to reduce the number of authentication. The control socket will be located at
-    `~/.ssh/julia-%r@%h:%p` during the session unless the existing multiplexing connection is used.
-
     In an all-to-all topology (the default), all workers connect to each other via plain TCP sockets.
     The security policy on the cluster nodes must thus ensure free connectivity between workers for
     the ephemeral port range (varies by OS).
 
     Securing and encrypting all worker-worker traffic (via SSH) or encrypting individual messages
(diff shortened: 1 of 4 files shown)


ViralBShah:
Maybe @tanmaykm might be able to review this as well.

stdlib/Distributed/src/managers.jl
```diff
@@ -178,6 +179,14 @@ function launch_on_machine(manager::SSHManager, machine, cnt, params, launched,
     end
     sshflags = `$(params[:sshflags]) $portopt`
 
+    # First it checks if ssh multiplexing has been already enabled and the master process is running.
+    # If it's already running, later ssh sessions also use the same ssh multiplexing session.
+    # If not, it creates a new ssh multiplexing session to open an ssh tunnel by default.
+    if tunnel && !success(`ssh $sshflags -O check $host`)
+        controlpath = "~/.ssh/julia-%r@%h:%p"
```
tanmaykm: ControlPath format needs to be documented. The ssh configuration required for this to work should be documented as well.

stdlib/Distributed/src/managers.jl
```diff
@@ -264,17 +273,11 @@ Return a port number `localport` such that `localhost:localport` connects to `ho
...
-    # `sleep 60` command specifies that an alloted time of 60 seconds is allowed to start the
-    # remote julia process and establish the network connections specified by the process topology.
-    # If no connections are made within 60 seconds, ssh will exit and an error will be printed on the
-    # process that launched the remote process.
-    ssh = `ssh -T -a -x -o ExitOnForwardFailure=yes`
+    # It assumes that an ssh multiplexing session has been already started by the remote worker.
     while cnt > 0
         localport = next_tunnel_port()
-        if success(detach(`$ssh -f $sshflags $user@$host -L $localport:$bind_addr:$port sleep 60`))
+        if success(`ssh $sshflags -O forward -L $localport:$bind_addr:$port $user@$host`)
```
tanmaykm: Would `-O forward` still work when tunneling without ssh multiplexing?

tanmaykm (changes requested):
SSH multiplexing would be useful. However it is not clear to me whether with this tunneling will always require ssh multiplexing to be configured manually. It should fallback to use non multiplexed connections if multiplexing is not configured IMO.

s417-lama:
Thank you for reviewing it, @tanmaykm !

> SSH multiplexing would be useful. However it is not clear to me whether with this tunneling will always require ssh multiplexing to be configured manually. It should fallback to use non multiplexed connections if multiplexing is not configured IMO.

I need to clarify it first. In this PR,
- If ssh multiplexing is set in `ssh_config` and the master process already exists, julia just reuses that session.
- If ssh multiplexing is not set in `ssh_config`, julia sets the multiplexing configuration dynamically as following (L.187 in `manager.jl`).
``` 
sshflags = `$sshflags -o ControlMaster=auto -o ControlPath=$controlpath -o ControlPersist=no`
```

Then it will use multiplexing even if it is not explicitly configured in config, and I don't think fallback to non-multiplexing mode is necessary.

tanmaykm:
Thanks @s417-lama for clarifying. That looks good!

I would recommend adding a test to ensure existence of the control socket and subsequent deletion [here](https://github.com/JuliaLang/julia/blob/c330f8d0d59a63dd161edd34e785a1085d5c7084/stdlib/Distributed/test/distributed_exec.jl#L721). Maybe something like:

```
    @test issocket(joinpath(homedir(), ".ssh", "julia-$(ENV["USER"])@localhost:22"))
    test_n_remove_pids(new_pids)
    @test :ok == timedwait(()->!issocket(joinpath(homedir(), ".ssh", "julia-$(ENV["USER"])@localhost:22")), 10; pollint=0.5)
```

stdlib/Distributed/src/managers.jl
tanmaykm: I think multiplexed tunneling should be mentioned in this doc section: https://github.com/JuliaLang/julia/blob/master/doc/src/manual/parallel-computing.md#network-requirements-for-localmanager-and-sshmanager.

stdlib/Distributed/src/managers.jl
```diff
@@ -516,8 +520,21 @@ end
 
 
 function connect_to_worker(host::AbstractString, bind_addr::AbstractString, port::Integer, tunnel_user::AbstractString, sshflags)
```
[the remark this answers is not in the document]
s417-lama: Right. I renamed it in 44c3c2d3b099a94bf999bcc2a942f1b080427730

s417-lama:
@tanmaykm Thank you for your kind guide! I added tests in ce7135778e7806595eb3e6a04bce17ba40181755

stdlib/Distributed/src/managers.jl
s417-lama: I added doc about it in ce7135778e7806595eb3e6a04bce17ba40181755

doc/src/manual/parallel-computing.md
```diff
@@ -1699,6 +1699,15 @@ requirements for the inbuilt `LocalManager` and `SSHManager`:
     authenticated via public key infrastructure (PKI). Authentication credentials can be supplied
     via `sshflags`, for example ```sshflags=`-i <keyfile>` ```.
 
+    Internally, SSH multiplexing is used to create a tunnel between the master and workers. If you have
+    been configured SSH multiplexing on your own and the connection has already been established,
```
tanmaykm: Typo: should probably be "have configured" instead of "have been configured"?

tanmaykm:
Thanks @s417-lama. I am not sure why win32 build failed, I'll restart that.

I have requested a minor change in the doc.

There remain some concerns about whether there would be any impact of having a single connection and hence single TCP buffer and window across multiple workers. It could finally limit the usable bandwidth. Also discussed this offline with @amitmurthy.

I feel it may be useful to have a way not to multiplex in some cases. For example we could probably multiplex only if `-tunnel=multiplex` is specified. Thoughts?

StefanKarpinski:
Win32 issue shouldn't be related, but I've restarted it and hopefully we'll get that green checkmark.

doc/src/manual/parallel-computing.md
s417-lama: Right. Fixed in 79cf3d74e58a755a4aaae9d6ccbbcc6d51a46393

s417-lama:
> Thanks @s417-lama. I am not sure why win32 build failed, I'll restart that.
> 
> I have requested a minor change in the doc.
> 
> There remain some concerns about whether there would be any impact of having a single connection and hence single TCP buffer and window across multiple workers. It could finally limit the usable bandwidth. Also discussed this offline with @amitmurthy.
> 
> I feel it may be useful to have a way not to multiplex in some cases. For example we could probably multiplex only if `-tunnel=multiplex` is specified. Thoughts?

Oh, I didn't think of bandwidth issue! I have no reason to decline it.

There would be several options to add `multiplex` option, like:
- Just add `:multiplex` option to `tunnel`, like `tunnel :: :multiplex | true | false`. Some might feel weird about this type.
- Make `tunnel` string type, like `tunnel :: :multiplex | :on | :off`. It's a breaking change, though.
- Have another parameter `multiplex :: true | false`

Which option do you think is better?

tanmaykm:
:+1: Having another parameter specifying `multiplex=true|false` seems preferable to me.

s417-lama:
Got it! I added `:multiplex = true | false` option. And also fixed test and docs.
(in commit a886c7d21f8b44dddbf1ddf18e73f0daa71e50c4)

tanmaykm (approved):

tanmaykm:
Looks good to me. I have restarted tests for linuxaarch64, they seem to have failed for unrelated reasons. Hoping they will pass this time.
Thanks @s417-lama!

StefanKarpinski:
No dice but the failure seems unlikely to be related... anyone familiar with this failure?

KristofferC:
Yes, it happens because our CI doesn't run on merge commits (https://github.com/JuliaCI/julia-buildbot/issues/126) and this PR was made before some fixes to aarch64 were merged. PR needs to be rebased if we want that to pass.

StefanKarpinski:
Ok, doesn't seem relevant, so rebase or not, this seems good to go.

tanmaykm:
maybe this can be merged now?

StefanKarpinski:
You call! Merge if it seems in good shape to you. Seems like a cool and useful feature.

tanmaykm:
:+1: merging now.

PR #34297, opened 2020-01-07
nicer constructor for ImmutableDict

simeonschaub:
It's currently quite cumbersome to construct ImmutableDicts with multiple entries. This PR allows constructing ImmutableDicts by pairs in the same way as regular Dicts. If people approve, I can add some docs.


JeffBezanson (approved):
Seems reasonable to me.

PR #34299, opened 2020-01-07
Fix trailing whitespace in svd docs introduced in cec4c3

fredrikekre:


andreasnoack (approved):

andreasnoack:
Thanks. Hadn't realized that whitespace is a separate builder and assumed 16/17 was sufficently close to 100%.

KristofferC:
No need to create a merge commit here imo.

PR #34300, opened 2020-01-07
Sockets: fix getipaddr/getipaddrs documentation, add islinklocaladdr function

mgkuhn:
Fix `getipaddr`/`getipaddrs` documentation:
    
- resolved IPv4/IPv6 confusion, e.g. `getipaddrs()` does _not_ return only IPv4

- more realistic example addresses (not just non-routeable IPv6 link-local addresses)

- example addresses taken from address ranges reserved for documentation https://en.wikipedia.org/wiki/Reserved_IP_addresses

Add `Sockets.islinklocaladdr`:

IPv6 interfaces always have an fe80::/10 link-local address configured, which is also reported by getipaddrs(). However, typical application servers are not reachable under their link-local address, therefore this convenience function helps users of getipaddrs() to filter out such addresses with `filter(!islinklocaladdr, getipaddrs())`.
    
This commit also adds a "See also" hint to make potential users of `getaddrs` aware that in applications where a server process has been started remotely via ssh the address found in `split(ENV["SSH_CONNECTION"], ' ')[3]` can be a preferable alternative source for an IP address under which this server is reachable, namely the very IP address that ssh has already used successfully to get here.

### DIFF
--- a/stdlib/Sockets/src/addrinfo.jl
+++ b/stdlib/Sockets/src/addrinfo.jl
@@ -241,6 +241,9 @@ addresses are available.
 Get an IP address of the local machine of the specified type. Throws if no
 addresses of the specified type are available.
 
+This function is a backwards-compatibility wrapper around [`getipaddrs`](@ref).
+New applications should use [`getipaddrs`](@ref) instead.
+
 # Examples
 ```julia-repl
 julia> getipaddr()
@@ -249,6 +252,8 @@ ip"192.168.1.28"
 julia> getipaddr(IPv6)
 ip"fe80::9731:35af:e1c5:6e49"
 ```
+
+See also: [`getipaddrs`](@ref)
 """
 function getipaddr(addr_type::Type{T}) where T<:IPAddr
     addrs = getipaddrs(addr_type)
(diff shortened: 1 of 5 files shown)


stdlib/Sockets/src/addrinfo.jl
```diff
@@ -319,3 +328,29 @@ function getipaddrs(addr_type::Type{T}=IPAddr; loopback::Bool=false) where T<:IP
...
+the address blocks `169.254.0.0/16` or `fe80::/10`.
+
+# Example
+```julia
+filter(!islinklocaladdr, getipaddrs())
+```
+"""
+function islinklocaladdr(addr::IPv4)
+    # RFC 3927
+    return addr.host &
```
Keno: This could use parens around the & I think. Looks odd otherwise.

mgkuhn: OK, parentheses now added in rebased commit b62e10c.

mgkuhn:
@Keno Ready to be merged? (the buildbot/tester_win32 failure seems unrelated to this PR and I don't have permission to press the Rebuild button on https://build.julialang.org/#/builders/65/builds/6777: "Cannot rebuild: you need to have role 'JuliaLang'", how does one get this?)

JeffBezanson (approved):

PR #34306, opened 2020-01-07
RFC: optimizer: add Instruction/InstructionStream abstraction

vtjnash:
This is an attempt to broaden and refine the generality of the AnySSAValue abstraction to a concrete Instruction struct. This lets us make aggregate statements about a node as:
```
result[nidx][] = ir.stmts[idx]
```
instead of writing it out individually as we do now:
```
result_types[nidx] = ir.types[idx]
result_ltable[nidx] = ir.lines[idx]
result_flags[nidx] = ir.flags[idx]
```

It also means we would want to (over time) alter the API from doing conjoined operations on the new index and the old fissioned values:
```
insert_node!(ircode::IRCode, newidx, @nospecialize ir.inst[idx], @nospecialize ir.types[idx], ir.lines[idx])
```
into separate operations on the index and the content:
```
newnode = insert_node!(ircode, newidx)
newnode[] = ir.stmts[idx]
```
such that in the future manual tracking of `nidx` goes away almost entirely, and the first example just reads as `result[] = old` (while nidx is still available as a field of result). But that is mostly left as future work here.

(and yes, this is nearly precisely opposite of LLVM, which first makes the clone, then inserts it—but is also more alike to LLVM that the operations are now more separable)
(when I first wrote this ~6 months ago, I checked that the compiler seemed to generally optimize this away, since we don't yet have stack-allocated immutables. I haven't looked into that again after rebasing this PR over more than 1000 commits before making this PR however to make sure something hadn't changed.)

### DIFF
--- a/base/compiler/ssair/ir.jl
+++ b/base/compiler/ssair/ir.jl
@@ -53,7 +53,6 @@ struct CFG
     index::Vector{Int} # map from instruction => basic-block number
                        # TODO: make this O(1) instead of O(log(n_blocks))?
 end
-copy(c::CFG) = CFG(BasicBlock[copy(b) for b in c.blocks], copy(c.index))
 
 function block_for_inst(index::Vector{Int}, inst::Int)
     return searchsortedfirst(index, inst, lt=(<=))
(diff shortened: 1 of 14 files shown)


vtjnash:
I've heard some offline comments that this may be good to go, so should I merge this?

base/compiler/ssair/ir.jl
```diff
@@ -53,7 +53,7 @@ struct CFG
     index::Vector{Int} # map from instruction => basic-block number
                        # TODO: make this O(1) instead of O(log(n_blocks))?
 end
-copy(c::CFG) = CFG(BasicBlock[copy(b) for b in c.blocks], copy(c.index))
+# copy(c::CFG) = CFG(BasicBlock[copy(b) for b in c.blocks], copy(c.index)) # unused
```
JeffBezanson: Delete?

base/compiler/ssair/slot2ssa.jl
```diff
@@ -492,18 +496,28 @@ function domsort_ssa!(ir::IRCode, domtree::DomTree)
         end
         new_bbs[new_bb] = BasicBlock(inst_range, new_preds, new_succs)
     end
-    result_stmts = Any[renumber_ssa!(result_stmts[i], inst_rename, true) for i in 1:length(result_stmts)]
+    #@assert bb_start_off == length(result)
+    #for i = 1:length(result)
+    #    @assert isassigned(result.types, i) && isassigned(result.stmts, i)
+    #end
```
JeffBezanson: Delete?

JeffBezanson (approved):

Keno:
Shall we get this landed?

Issue #24645, opened 2017-11-18
Rename issparse?


garrison:
... aka Bikeshed and/or improve documentation for `issparse`.  The docstring for `issparse` says

> `issparse(S)`
>
> Returns `true` if `S` is sparse, and `false` otherwise.

but this is ambiguous.  What, precisely, does it mean for `S` (presumably an array) to be sparse?  Here are two (very different) possible definitions:

1. The array has a density of nonzero elements less than some parameter `x` (`x` could have some default value, but this would be rather arbitrary).
2. The array is represented by a some "sparse" data type.

It is the second option that is implemented by `issparse`.  In a way this makes sense, as the function does not take such a parameter `x` (representing the threshold density of nonzero elements).  But at a minimum this should be more clearly explained in the documentation.  (I'm also unable to explain why `issparse(::Diagonal)` and related types return `false` -- perhaps it's a bug, or perhaps I do not understand fully what this function is meant to do.)

On the other hand, the current behavior makes `issparse` different from many other functions I can think of, such as `istril`, `istriu`, `issymmetric`, `ishermitian`, and `isreal`.  Each of these considers the *value* itself rather than the datatype; for instance, `isreal(0im) == true`.

Perhaps a better name for this function, then, would be `issparsetype`?  I can think of downsides to this name (you do not pass a type to it).  Another idea would be `issparsestorage`, but I don't know if I like this either.

So I ask:

1. Do people agree that `issparse` is a poor name?  If it is, can we come up with anything better?
2. What should `issparse(::Diagonal)`, `issparse(::Tridiagonal)`, etc. return?
3. Is this function even useful?  I just searched discourse, and it's never been mentioned there.  In Base,
 it is used only in the concatenation functions in `sparsevector.jl`.  I have not grepped the package ecosystem, but I wonder how frequently it is used there.

andreasnoack:
The methods are very old and most likely inherited from Matlab. In Julia, I think it makes more sense to check if the type is a `SparseMatrixCSC` or `AbstractSparseMatrix` but there is the complication that a this would be false for e.g. `LowerTriangular{SparseMatrixCSC}` for which `issparse` is true.

oxinabox:
`issparse` should be a property of the type only.
Like a trait.

as @andreasnoack  said it can't be just replaced with:

`x isa AbstractSparseMatrix` as it also depends on type parameters for e.g. `LowerTriangular`.


[...]

PR #34308, opened 2020-01-08
Make issparse handle nested wrappers

oxinabox:
Follow up to: https://github.com/JuliaLang/julia/pull/34266

Handles all the existing cases, plus:
 - arbitary nesting of wrappers
 - `SubArrays`
 - `Diagonal`

It all constant folds out:
```
julia> z = adjoint(transpose((1 + im) .*sprandn(5,5, 0.5)));

julia> @code_llvm issparse(z)

;  @ REPL[18]:2 within `issparse'
define i8 @julia_issparse_17287(%jl_value_t addrspace(10)* nonnull align 8 dereferenceable(8)) {
top:
;  @ REPL[18]:4 within `issparse'
  ret i8 1
}

julia> q = adjoint(transpose((1 + im) .*randn(5,5)));

julia> @code_llvm issparse(q)

;  @ REPL[18]:2 within `issparse'
define i8 @julia_issparse_17303(%jl_value_t addrspace(10)* nonnull align 8 dereferenceable(8)) {
top:
;  @ REPL[18]:4 within `issparse'
  ret i8 0
}
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -117,7 +117,7 @@ Standard library changes
 
 * new `sizehint!(::SparseMatrixCSC, ::Integer)` method ([#30676]).
 * `cholesky()` now fully preserves the user-specified permutation. ([#40560])
-
+* `issparse` now applies consistently to all wrapper arrays, including nested, by checking `issparse` on the wrapped parent array ([#37644]).
 
 #### Dates
 
--- a/stdlib/SparseArrays/src/abstractsparse.jl
+++ b/stdlib/SparseArrays/src/abstractsparse.jl
@@ -50,17 +50,19 @@ julia> issparse(Array(sv))
 false
 ```
 """
-issparse(A::AbstractArray) = false
+function issparse(A::AbstractArray)
+    # Handle wrapper arrays: sparse if it is wrapping a sparse array.
+    # This gets compiled away during specialization.
+    p = parent(A)
+    if p === A
+        # have reached top of wrapping without finding a sparse array, assume it is not.
+        return false
+    else
+        return issparse(p)
+    end
+end
+issparse(A::DenseArray) = false
 issparse(S::AbstractSparseArray) = true
-issparse(S::LinearAlgebra.Adjoint{<:Any,<:AbstractSparseArray}) = true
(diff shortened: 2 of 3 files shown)


fredrikekre:
Perhaps OT, but can we just delete `issparse` instead? It does not seem very useful, what do you do with it? Outside of tests it is only used in three places: https://github.com/JuliaLang/julia/blob/3608c84e6093594fe86923339fc315231492484c/stdlib/SparseArrays/src/sparsevector.jl#L1040 https://github.com/JuliaLang/julia/blob/3608c84e6093594fe86923339fc315231492484c/stdlib/SparseArrays/src/sparsevector.jl#L1045 and https://github.com/JuliaLang/julia/blob/3608c84e6093594fe86923339fc315231492484c/stdlib/SparseArrays/src/sparsevector.jl#L1049

(See also https://github.com/JuliaLang/julia/issues/24645)

oxinabox:
I agree with the sentiment.
a trait without a set of rules about what it promises is not good for much.

But since it is a breaking change, we can't delete it.

So we might as well make it work well.

That may or may not include making it work for these cases.

Perhaps before doing this we should decide what it means.

It probably should mean something about the cost of various operations.

oxinabox:
This errors because:
 -  `Diagonal(rand(4,4))` works fine and just cuts out the diagonal, ignoring nondiagonal elements.
 - `Diagonal(LowerTriangular(rand(4,4)))` errors because the matrix has zeros on the non-diagonal

Which seems really silly and inconsistent to me.

ViralBShah:
I thought we can't delete APIs that are in 1.0.

oxinabox:
Closing in favour of https://github.com/JuliaLang/julia/pull/37644
Since that one doesn't need to be rebased.

oxinabox:
Revisiting this. Since I apparently decided to open a second PR having forgotten i made this one.
#37644 

> What does it mean to be issparse?
> I haven't seen a defination of an interface that must be furfilled.
> 
> Possibly all people want this for is to know if they have some algorthm that really won't run great on a sparse array, then they could issue a warning or run collect to make it dense.

dkarrasch:
The tests pass locally for me, so this should be good to go after a rebase.

oxinabox:
Done

PR #34313, opened 2020-01-08
Support LLVM 10 changes

vchuravy:


src/jitlayers.cpp
```diff
@@ -460,18 +460,16 @@ void JuliaOJIT::addModule(std::unique_ptr<Module> M)
...
-        ++I;
-        if (F->isDeclaration()) {
-            if (F->use_empty())
-                F->eraseFromParent();
-            else if (!((isa<Function>(F) && isIntrinsicFunction(cast<Function>(F))) ||
-                       findUnmangledSymbol(F->getName()) ||
+    for (GlobalObject &F : M->global_objects()) {
+        if (F.isDeclaration()) {
+            if (F.use_empty())
+                F.eraseFromParent();
```
vtjnash: this doesn't seem safe

chriselrod:
I don't know where the correct place to bring this up is, but this is a change in LLVM10 I am not excited about:
> -mprefer-vector-width=256 is now the default behavior skylake-avx512 and later Intel CPUs. This tries to limit the use of 512-bit registers which can cause a decrease in CPU frequency on these CPUs. This can be re-enabled by passing -mprefer-vector-width=512 to clang or passing -mattr=-prefer-256-bit to llc.
[LLVM 10 Release Notes](https://llvm.org/docs/ReleaseNotes.html#changes-to-the-x86-target)

Any idea what impact this would have on code using LLVM intrinsics with 512 bits? What if those intrinsics used the `fast` flag?

It would be nice to be able to specify `-mprefer-vector-width=512`. I don't necessarily want to use `llvmcall` for all my optimized code. I suspect the average use case for Julia tends to have more number crunching on longer vectors than most other languages.

vchuravy:
Good to merge?

vtjnash (approved):

vchuravy:
> I don't know where the correct place to bring this up is, but this is a change in LLVM10 I am not excited about:

Not really, but as a short commentary: One thing that is important is to make sure that benchmarks are reflective of real-world applications and if your application compute heavy I can see that avx512 is beneficial.  On the other hand you have reports like [1-3] which shows that there is a non-obvious and real cost with potentially having spurious avx512 instructions. These effects are not really visible when executing an avx512 benchmark.

In the end it doesn't matter much since you can configure this by setting `JULIA_CPU_TARGET="native,-prefer-256-bit"` and if you can convince us with the right benchmark data, we can look into making that the default (at least on something like O3)

[1] https://travisdowns.github.io/blog/2020/01/17/avxfreq1.html
[2] https://lemire.me/blog/2018/09/07/avx-512-when-and-how-to-use-these-new-instructions/
[3] https://blog.cloudflare.com/on-the-dangers-of-intels-frequency-scaling/

PR #33224, opened 2019-09-11
Clean up updated hypot function

musm:
@cfborges Looks like you forgot to delete the existing code   (ref https://github.com/JuliaLang/julia/pull/31922)

PR #34316, opened 2020-01-08
add `hypot` fallback for 2 Numbers of the same type

JeffBezanson:
#33224 broke `hypot` with Unitful.jl. There used to be a method for this. The weird thing is, in the current state it works if you pass 3 arguments to `hypot` but not 2, so we should probably add this.

### DIFF
--- a/base/math.jl
+++ b/base/math.jl
@@ -613,6 +613,7 @@ julia> hypot(3, 4im)
 hypot(x::Number, y::Number) = hypot(promote(x, y)...)
 hypot(x::Complex, y::Complex) = hypot(abs(x), abs(y))
 hypot(x::T, y::T) where {T<:Real} = hypot(float(x), float(y))
+hypot(x::T, y::T) where {T<:Number} = (z = y/x; abs(x) * sqrt(one(z) + z*z))
 function hypot(x::T, y::T) where T<:AbstractFloat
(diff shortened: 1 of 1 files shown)


KristofferC:
Seems like this is causing some subtyping crash when creating the sysimage?

base/math.jl
```diff
@@ -613,6 +613,7 @@ julia> hypot(3, 4im)
 hypot(x::Number, y::Number) = hypot(promote(x, y)...)
 hypot(x::Complex, y::Complex) = hypot(abs(x), abs(y))
 hypot(x::T, y::T) where {T<:Real} = hypot(float(x), float(y))
+hypot(x::T, y::T) where {T<:Number} = sqrt(abs2(x) + abs2(y))
```
giordano: Doesn't this overflow/underflow, which `hypot` should _not_ do?

ViralBShah: At least use something like `abs(x) * sqrt (1 + (y/x)^2)`.

ViralBShah:
Probably not worth holding up the PR.

jebej:
Why not just revert that PR and keep the old method for non-AbstractFloats?

JeffBezanson:
There already is a method for two Reals and two Complexes. So this would only apply to non-Real, non-Complex Numbers, so potentially somewhat exotic types that might not support all operations.

PR #34320, opened 2020-01-08
Add spmv! to BLAS in stdlib/LinearAlgebra

iolaszlo:


stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -964,6 +965,74 @@ Return the updated `y`.
...
+                  β,
+                  y,
+                  incy)
+        end
+    end
+end
+
+function spmv!(uplo::AbstractChar,
+               α::Number, AP::Union{DenseArray{T}, AbstractVector{T}}, x::Union{DenseArray{T}, AbstractVector{T}},
+               β::Number, y::Union{DenseArray{T}, AbstractVector{T}}) where {T <: BlasReal}
```
dkarrasch: Now that I see similar code again as in the `hpmv` PR, I started wondering why `AP`, `x` and `y` can be of type `DenseArray{T}`. For instance, in `sbmv!` the arguments `x` and `y` must be of type `AbstractVector{$elty}`. Same thing for `AP`: if the BLAS method can handle anything besides `AbstractVector`s, this should be tested then, or this signature (and `hpmv`'s narrowed).

dkarrasch (changes requested):
This looks great. Just one comment on the allowed signature, with relevance for the previous `hpmv` PR.

stdlib/LinearAlgebra/src/blas.jl
ViralBShah: AbstractVector would be too broad for BLAS.

dkarrasch: Could be, but most if not any functions in `blas.jl` have `AbstractMatrix` and `AbstractVector` in their signatures... which may need revision. But the case here is even a bit more particular because these multiplication methods seem to require some packed storage, essentially the columns of a triangle `vec`ed. That's why I was even more confused about the rather general `DenseArray` here.

iolaszlo: > Now that I see similar code again as in the `hpmv` PR, I started wondering why `AP`, `x` and `y` can be of type `DenseArray{T}`. For instance, in `sbmv!` the arguments `x` and `y` must be of type `AbstractVector{$elty}`. Same thing for `AP`: if the BLAS method can handle anything besides `AbstractVector`s, this should be tested then, or this signature (and `hpmv`'s narrowed).

Thank you, that's quite an interesting point.

I would say, without claiming to know the context and reasons which might decide
whether to use `DenseArray` or not in the long run, that the approach depends on which existing
`blas.jl` function implementation one prefers to take as a reference.  It's
true that `sbmv!` does not contain `DenseArray` in its header, yet for example
`axpby!` does so and I couldn't find any obvious references to `DenseArray` in the
tests of `axpby!`.

Regarding `DenseArray` I collected some info from looking at the julia source code.
It seems to be some sort of inbuilt abstract type, used to subtype
concrete implementations with guarantees of memory contiguity beyond(?) that of
usual numeric arrays. The latter usual numeric arrays seem to be already valid
inputs to wrapped BLAS calls.

Beyond `DenseArray`'s implementation in `base` and references to it in various
function headers (for example in `blas.jl`) I found only two examples where
`DenseArray` is subtyped:

```
stdlib/SharedArrays/src/SharedArrays.jl: mutable struct SharedArray{T,N} <: DenseArray{T,N}
stdlib/Random/RNGs.jl: struct UnsafeView{T} <: DenseArray{T,1}
```

Thus I am not sure if the two options of removing `DenseArray` completely from `hpmv!`
and `spmv!` or writing extra tests are feasible, since `DenseArray` is referenced in other
BLAS function headers as well without explicit tests written and the newly added functions don't seem to
contradict this pattern. As `DenseArray` is abstract, a solution could be to create
a custom type within the tests that is a subtype of `DenseArray` but
the context and scope of doing that seems a bit misterious to me at the moment.

So any remarks, clarifications, ideas would be appreciated.

dkarrasch: I was reading `DenseArray` more like "`DenseMatrix`", so my concern was more that you're not enforcing the packed format. I'm not sure, but maybe you could specify the `N` parameter as `1`? Like restricting to "`DenseVector`s"? That would resolve my main concern... unless it is possible to pass a two-dimensional array for `A`, and if that was the case I think we should test the method with an array (I was referring to this situation earlier when I was suggesting to add tests).

iolaszlo: > I was reading `DenseArray` more like "`DenseMatrix`", so my concern was more that you're not enforcing the packed format. I'm not sure, but maybe you could specify the `N` parameter as `1`? Like restricting to "`DenseVector`s"? That would resolve my main concern... unless it is possible to pass a two-dimensional array for `A`, and if that was the case I think we should test the method with an array (I was referring to this situation earlier when I was suggesting to add tests).

Yes, your concern about not enforcing dimensionality of 1 seems valid to me, as opposed to to the cases of `dot*`, `axpy!`, `axpby!` and `nrm2` which all use `DenseArray` and have vector inputs, whereas for hpmv! and spmv! we have matrices dressed up as vectors. I am not sure though whether `DenseArray{T <: BlasReal}`  is already silently implied to be `DenseArray{T,1}` or not. The existing syntax for the other functions with truly vectorial inputs seems suggestive toward this possibility, yet then again why not change those function headers to `DenseVector` as well?

iolaszlo: > .... But the case here is even a bit more particular because these multiplication methods seem to require some packed storage, essentially the columns of a triangle `vec`ed....

I was considering to adopt such a conversion earlier but in my opinion requiring this would be be a compromise in speed and flexibility in some cases. I would speculate that the typical use case would be that the user is aware of BLAS and is granted to use it in a performant way. To serve the latter, there is a short function header (for example in `hpmv!`) serving as a convenience function and a longer header with a one-to-one mapping to the BLAS API if optimization is required.

dkarrasch: In BLAS ...

> `dot*`, `axpy!`, `axpby!` and `nrm2`

... all work for both vectors and matrices, and actually arbitrary arrays, by interpreting them as lists of numbers ("vectors"). I guess the same is true for `hpmv` and `spmv`, it's just that they are interpreted according to the packed format. So I came to realize that the dimension of the arrays is all the same to BLAS: you could provide 3-dimensional arrays, and then BLAS decides how to interpet them, as if `vec`ed, or packed, or whatever.

There is no need to provide a function that `vec`s the columns of one of the triangles. For the "full" matrix input, we have `hemv` and `symv`, even though here also only one of the triangles is read from.

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -964,6 +965,74 @@ Return the updated `y`.
...
+    GC.@preserve x y AP spmv!(uplo, N, convert(T, α), AP, pointer(x), BlasInt(stride(x, 1)), convert(T, β), pointer(y), BlasInt(stride(y, 1)))
+    y
+end
+
+"""
+    spmv!(uplo, α, AP, x, β, y)
+
+Update vector `y` as `α*AP*x + β*y` where `AP` is a packed symmetric matrix.
+The storage layout for `AP` is described in the reference BLAS module, level-2 BLAS at
+<http://www.netlib.org/lapack/explore-html/>.
```
dkarrasch: This may be a bit confusing and too brief. I suggest to be a bit more explicit here, to avoid confusion.
```
Update vector `y` as `α*A*x + β*y`, where `A` is a symmetric matrix provided in packed format `AP`.
With `uplo = 'U'`, the array AP must contain the upper triangular part of the symmetric matrix
packed sequentially, column by column, so that `AP[1]` contains `A[1, 1]`,
`AP[2]` and `AP[3]` contain `A[1, 2]` and `A[2, 2]` respectively, and so on.
With `uplo = 'L'`, the array AP must contain the lower triangular part of the symmetric matrix
packed sequentially, column by column, so that `AP[1]` contains `A[1, 1]`,
`AP[2]` and `AP[3]` contain `A[2, 1]` and `A[3, 1]` respectively, and so on.
```

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -964,6 +965,74 @@ Return the updated `y`.
...
+
+The scalar inputs `α` and `β` shall be non-complex numbers.
```
dkarrasch: Rewording suggestion:
The scalar inputs `α` and `β` must be either `Float32` or `Float64`.

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -964,6 +965,74 @@ Return the updated `y`.
...
+
+The array inputs `x`, `y` and `AP` must be real one-dimensional julia arrays of the
+same type that is either `Float32` or `Float64`.
```
dkarrasch: I don't think this is required. `x` and `y` are addressed by linear indexing, and otherwise BLAS doesn't seem to care about the formal dimensionality of them. I'd think we can remove this.

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -964,6 +965,74 @@ Return the updated `y`.
...
+               β::Number, y::Union{DenseArray{T}, AbstractVector{T}}) where {T <: BlasReal}
+    require_one_based_indexing(AP, x, y)
+    N = length(x)
+    if N != length(y)
+        throw(DimensionMismatch("x has length $(N), but y has length $(length(y))"))
+    end
+    if length(AP) < Int64(N*(N+1)/2)
+        throw(DimensionMismatch("Packed symmetric matrix A has size smaller than length(x) = $(N)."))
+    end
+    GC.@preserve x y AP spmv!(uplo, N, convert(T, α), AP, pointer(x), BlasInt(stride(x, 1)), convert(T, β), pointer(y), BlasInt(stride(y, 1)))
```
dkarrasch: Do we need to preserve `AP` explicitly? I guess it's referenced directly and not in danger of being gc'ed.

stdlib/LinearAlgebra/test/blas.jl
```diff
@@ -241,6 +241,46 @@ Random.seed!(100)
...
+
+                y_result_julia = α*A*x+β*y
+
+                # Create lower triangular packing of A
+                AP = typeof(A[1,1])[]
+                for j in 1:n
+                    for i in j:n
+                        push!(AP, A[i,j])
+                    end
+                end
```
dkarrasch: I'd suggest to make the test a bit "fancier". 😄  Have `A = Symmetric(M)`, and use `M` in the loop to build `AP`, so never really build a symmetric matrix in memory... just a suggestion.

dkarrasch (changes requested):
I think my previous concern on "enforcing 1-dimensionality" was void, so that's resolved. I'd suggest though to explain the packed format explicitly, since the link to the LAPACK page is rather indirect, one has to search on their webpage to get to the specific page. This might carry over to the `hpmv` code.

stdlib/LinearAlgebra/src/blas.jl
iolaszlo: > In BLAS ...
> 
> > `dot*`, `axpy!`, `axpby!` and `nrm2`
> 
> ... all work for both vectors and matrices, and actually arbitrary arrays, by interpreting them as lists of numbers ("vectors"). I guess the same is true for `hpmv` and `spmv`, it's just that they are interpreted according to the packed format. So I came to realize that the dimension of the arrays is all the same to BLAS: you could provide 3-dimensional arrays, and then BLAS decides how to interpet them, as if `vec`ed, or packed, or whatever.
> 
> There is no need to provide a function that `vec`s the columns of one of the triangles. For the "full" matrix input, we have `hemv` and `symv`, even though here also only one of the triangles is read from.

Thank you for the final insight. I think the whole conversation was quite helpful to form a picture of what is going on.

stdlib/LinearAlgebra/src/blas.jl
iolaszlo: Currently the convert routines called in the short version of the function permit any non-complex number. The long function header is more restrictive and currently undocumented as are all  the other lowest-level-longest-header `blas.jl` functions. So if I would apply this docstring change I would be stating an untrue claim about the short-header function namely that it is actually more restrictive than it is reality. In the long run it would make sense to write separate docstrings for all function headers, but maybe this could be part of a separate commit/pull request?

stdlib/LinearAlgebra/src/blas.jl
iolaszlo: Maybe I'm too cautious here due to lack of deep GC knowledge, so looking at the other @GC.preserve statements in `blas.jl` I do not see any crucial difference and reason to drop AP from the preserve statement. Could you maybe explain this a bit more or point out the the thing that I'm missing?

dkarrasch: IIUC, then only those are listed which are wrapped by a `pointer(...)`. You could double-check if that's the case, but that was my understanding.

iolaszlo: > IIUC, then only those are listed which are wrapped by a `pointer(...)`. You could double-check if that's the case, but that was my understanding.

So if I were to drop the `pointer(...)` around x and y as well then the whole @GC.preserve might be unnecessary? If true this might be tempting to apply to have a more clean API. :)

iolaszlo: Possibly following here a pattern similar to that in the  `symv!` implementation strengthens your reading so removing the `@GC.preserve` alltogether might good to go. If you are also okay with it I will apply this modification as well.

iolaszlo: On the other hand maybe the following is a non-issue but wherever we have a `@GC.preserve` we also have Union{DenseArray,...} or Union{Unitrange,...} or Union{.., ...}. Maybe these are unrelated but they do appear together so the removal of `@GC.preserve` would  induce to suggest to remove `pointer(...)` and `GC.preserve` from all those functions. I don't know whether this is wise as people might have put it this way for a reason unknown to me at the moment.

stdlib/LinearAlgebra/src/blas.jl
iolaszlo: I am not sure what would happen if someone threw say `Integer` vectors at the input but I am going to look into it.

dkarrasch: True, I was actually referring to the "one-dimensional". So indeed, we should leave this sentence, but remove "one-dimensional". Suggestion:
```julia
The inputs `x`, `y` and `AP` must all be  of `Float32` or `Float64` type.
```

stdlib/LinearAlgebra/src/blas.jl
dkarrasch: You're right. Then maybe
```julia
The scalar inputs α and β must be real.
```
And you could enforce that actually in the signature by requiring `Real`. My point was mainly that "non-complex" sounds fairly generic.

stdlib/LinearAlgebra/src/blas.jl
dkarrasch: As I said in some other comment, this could be `α::Real` and `β::Real`, then you're on the safe side.

stdlib/LinearAlgebra/src/blas.jl
dkarrasch: Probably there is no harm in keeping `AP` here, so unless someone comes by and helps in resolving this aspect, this should not keep us from merging this.

dkarrasch (commented):
I'd suggest we work on this until `spmv!` is perfect, and then edit `hpmv!` accordingly in another PR. I believe there are only mild adoptions necessary, like the restriction to `Real`, and some small work on the docstrings, so there is an end in sight. We can raise the `GC.gc` issue in the next PR again, unless someone comes by and helps out with expertise here.

stdlib/LinearAlgebra/src/blas.jl
iolaszlo: I am planning other similar pull requests. Would it be okay  in your opinion to ask someone within this comment thread  whether to use the `@GC.preserve` to ease further development?

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -964,6 +965,82 @@ Return the updated `y`.
 """
 sbmv!
 
+### spmv!, (SP) symmetric packed matrix-vector operation defined as y := alpha*A*x + beta*y.
+for (fname, elty) in ((:dspmv_, :Float64),
+                      (:sspmv_, :Float32))
+    @eval begin
+        function spmv!(uplo::AbstractChar,
+                       n::BlasInt,
```
dkarrasch: I guess we kind of figured out in #34360 that this should be `Integer`.

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -964,6 +965,82 @@ Return the updated `y`.
...
+               β::Real, y::Union{DenseArray{T}, AbstractVector{T}}) where {T <: BlasReal}
+    require_one_based_indexing(AP, x, y)
+    N = length(x)
+    if N != length(y)
+        throw(DimensionMismatch("x has length $(N), but y has length $(length(y))"))
+    end
+    if length(AP) < Int64(N*(N+1)/2)
+        throw(DimensionMismatch("Packed symmetric matrix A has size smaller than length(x) = $(N)."))
+    end
+    GC.@preserve x y AP spmv!(uplo, BlasInt(N), convert(T, α), AP, pointer(x), BlasInt(stride(x, 1)), convert(T, β), pointer(y), BlasInt(stride(y, 1)))
```
dkarrasch: ... and here no `BlasInt(N)` etc. conversions

stdlib/LinearAlgebra/test/blas.jl
```diff
@@ -241,6 +241,46 @@ Random.seed!(100)
...
         end
 
+        # spmv!
+        if elty in (Float32, Float64)
+            @testset "spmv!" begin
+                # Both matrix dimensions n coincide, as we have symmetric matrices.
+                # Define the inputs and outputs of spmv!, y = α*A*x+β*y
+                α = rand(elty)
+                M = rand(elty, n, n)
+                A = Symmetric(M)
```
dkarrasch: I would write
```julia
AL = Symmetric(M, :L)
AU = Symmetric(M, :U)
```
...

stdlib/LinearAlgebra/test/blas.jl
```diff
@@ -241,6 +241,46 @@ Random.seed!(100)
...
+                x = rand(elty, n)
+                β = rand(elty)
+                y = rand(elty, n)
+
+                y_result_julia = α*A*x+β*y
```
dkarrasch: then
```julia
y_result_julia = α*AL*x+β*y
```
...

stdlib/LinearAlgebra/test/blas.jl
```diff
@@ -241,6 +241,46 @@ Random.seed!(100)
...
+
+                # Create lower triangular packing of A
+                AP = typeof(A[1,1])[]
+                for j in 1:n
+                    for i in j:n
+                        push!(AP, A[i,j])
```
dkarrasch: ...
```julia
push!(AP, M[i,j])
```
and `AP = typeof(M[1,1])[]` (even though that doesn't matter at all, just for clarity where we take types from)...

stdlib/LinearAlgebra/test/blas.jl
```diff
@@ -241,6 +241,46 @@ Random.seed!(100)
...
+                    end
+                end
+
+                y_result_blas_lower = copy(y)
+                BLAS.spmv!('L', α, AP, x, β, y_result_blas_lower)
+                @test y_result_julia≈y_result_blas_lower
+
+                # Create upper triangular packing of A
+                AP = typeof(A[1,1])[]
```
dkarrasch: ... and analogous changes here, with a recomputed
```julia
y_result_julia = α*AU*x+β*y
```

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -964,6 +965,82 @@ Return the updated `y`.
 """
 sbmv!
 
+### spmv!, (SP) symmetric packed matrix-vector operation defined as y := alpha*A*x + beta*y.
+for (fname, elty) in ((:dspmv_, :Float64),
+                      (:sspmv_, :Float32))
+    @eval begin
+        function spmv!(uplo::AbstractChar,
```
dkarrasch: Could you add here the original BLAS signature here for completeness. You have it also in `hpmv!`.

dkarrasch (changes requested):
This is shaping up nicely, just some minor polishing and this may serve as the role model for a minor `hpmv!` "fix".

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -964,6 +965,82 @@ Return the updated `y`.
...
+respectively, and so on.
+
+With `uplo = 'L'`, the array AP must contain the lower triangular part of the
+symmetric matrix packed sequentially, column by column, so that `AP[1]`
+contains `A[1, 1]`, `AP[2]` and `AP[3]` contain `A[2, 1]` and `A[3, 1]`
+respectively, and so on.
+
+The scalar inputs `α` and `β` must be real.
+
+The array inputs `x`, `y` and `AP` must all be vectors of `Float32` or `Float64`.
```
dkarrasch: One more quick comment: I don't think `AP` must be a vector necessarily, we do allow `DenseArray`s in the signature. For instance, the following works
```julia
julia> using LinearAlgebra

julia> using LinearAlgebra.BLAS

julia> AP = rand(ComplexF64, 10, 10);

julia> x = rand(ComplexF64, 10);

julia> y = rand(ComplexF64, 10);

julia> α = rand()
0.3159974975521751

julia> β = rand()
0.44211744434461364

julia> hpmv!('U', α, AP, x, β, y)
```
It grabs as much as needed from `AP`, and goes through `AP` by linear indexing, ignoring the back part.

stdlib/LinearAlgebra/src/blas.jl
iolaszlo: @dkarrasch

dkarrasch (approved):

dkarrasch:
@yuyichao Could you please help us and take a look at the `GC.@preserve` line. We are unsure whether `AP` needs to be listed there. It would be great if you could also drop a line as to what needs to be protected from garbage collection.

@iolaszlo Could you maybe rebase your branch on current master. That might help to resolve some of the (otherwise unrelated) test failures.

dkarrasch:
It would be great to get this PR in, and to learn about how to use `GC.@preserve` correctly for future PRs. Anyone care to advise us? @yuyichao @andreasnoack @ViralBShah ? The remaining question is which variables need to be preserved from gc.

stdlib/LinearAlgebra/src/blas.jl
andreasnoack: It's better just to pass the arrays, see https://github.com/JuliaLang/julia/pull/34783.

ViralBShah:
+1 to get the PR in, but we need someone else to chime in on `GC.@preserve`

dkarrasch (approved):
Thanks for your stamina @iolaszlo! 😃

dkarrasch:
On win32, the `LinearAlgebra/blas` tests pass, macos64 is a download failure.

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -964,6 +965,90 @@ Return the updated `y`.
...
+                  uplo,
+                  n,
+                  α,
+                  AP,
+                  x,
+                  incx,
+                  β,
+                  y,
+                  incy)
+        end
```
andreasnoack: You'd have to add `return y` here.

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -964,6 +965,90 @@ Return the updated `y`.
...
+    require_one_based_indexing(AP, x, y)
+    N = length(x)
+    if N != length(y)
+        throw(DimensionMismatch("x has length $(N), but y has length $(length(y))"))
+    end
+    if length(AP) < Int64(N*(N+1)/2)
+        throw(DimensionMismatch("Packed symmetric matrix A has size smaller than length(x) = $(N)."))
+    end
+    spmv!(uplo, N, convert(T, α), AP, x, stride(x, 1), convert(T, β), y, stride(y, 1))
+    y
```
andreasnoack: ...then you can leave it out here. I prefer explicit `return spmv!...` but it's not required.

stdlib/LinearAlgebra/test/blas.jl
```diff
@@ -241,6 +241,50 @@ Random.seed!(100)
...
+                AP = typeof(M[1,1])[]
+                for j in 1:n
+                    for i in 1:j
+                        push!(AP, AU[i,j])
+                    end
+                end
+
+                y_result_blas_upper = copy(y)
+                BLAS.spmv!('U', α, AP, x, β, y_result_blas_upper)
+                @test y_result_julia_upper≈y_result_blas_upper
```
andreasnoack:
```suggestion
                @test y_result_julia_upper ≈ y_result_blas_upper
```

stdlib/LinearAlgebra/test/blas.jl
```diff
@@ -241,6 +241,50 @@ Random.seed!(100)
...
+                        push!(AP, AL[i,j])
+                    end
+                end
+
+                y_result_blas_lower = copy(y)
+                BLAS.spmv!('L', α, AP, x, β, y_result_blas_lower)
+                @test y_result_julia_lower≈y_result_blas_lower
+
+
+                y_result_julia_upper = α*AU*x+β*y
```
andreasnoack:
```suggestion
                y_result_julia_upper = α*AU*x + β*y
```

stdlib/LinearAlgebra/test/blas.jl
```diff
@@ -241,6 +241,50 @@ Random.seed!(100)
...
+                # Define the inputs and outputs of spmv!, y = α*A*x+β*y
+                α = rand(elty)
+                M = rand(elty, n, n)
+                AL = Symmetric(M, :L)
+                AU = Symmetric(M, :U)
+                x = rand(elty, n)
+                β = rand(elty)
+                y = rand(elty, n)
+
+                y_result_julia_lower = α*AL*x+β*y
```
andreasnoack:
```suggestion
                y_result_julia_lower = α*AL*x + β*y
```

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -964,6 +965,90 @@ Return the updated `y`.
...
+
+function spmv!(uplo::AbstractChar,
+               α::Real, AP::Union{DenseArray{T}, AbstractVector{T}}, x::Union{DenseArray{T}, AbstractVector{T}},
+               β::Real, y::Union{DenseArray{T}, AbstractVector{T}}) where {T <: BlasReal}
+    require_one_based_indexing(AP, x, y)
+    N = length(x)
+    if N != length(y)
+        throw(DimensionMismatch("x has length $(N), but y has length $(length(y))"))
+    end
+    if length(AP) < Int64(N*(N+1)/2)
```
andreasnoack:
```suggestion
    if 2*length(AP) < N*(N + 1)
```

andreasnoack (commented):
Just a few minor comments, but looks good to me other than that.

andreasnoack (approved):

dkarrasch (approved):

dkarrasch:
CI failures are in Distributed and Sockets, respectively. I'd say unrelated.

iolaszlo:
@dkarrasch Thank you for the constructive review process.

PR #34321, opened 2020-01-09
Update suitesparse download url

ViralBShah:
Download SuiteSparse from github. The tests do not check this, but I've verified it locally. Only relevant if you build with USE_BINARYBUILDER=0.


ViralBShah:
I believe the test failures are not relevant. Would be good if @staticfloat could give it a quick look - and merge.

staticfloat:
Too bad the source hashes changed.  LGTM.

PR #34324, opened 2020-01-09
Build all the targets we build in SuiteSparse artifacts:

ViralBShah:
https://github.com/JuliaPackaging/Yggdrasil/blob/master/S/SuiteSparse%405/build_tarballs.jl

The artifacts SuiteSparse build does quite a bit more. This at least builds all the same libraries, but does not link against metis (which is only used if found, but keeps the same API). Essentially this should not be used for anything except for development.

### DIFF
--- a/deps/suitesparse.mk
+++ b/deps/suitesparse.mk
@@ -16,8 +16,8 @@ CHOLMOD_CONFIG += -DNPARTITION
 
 ifneq ($(USE_BINARYBUILDER_SUITESPARSE), 1)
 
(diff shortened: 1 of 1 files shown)


nalimilan:
Did you voluntarily avoid adding these new libraries to other places like these?
https://github.com/JuliaLang/julia/blob/a90a4472dbbcce6fe414eafed0bebd88a5b75a9a/deps/suitesparse.mk#L113
https://github.com/JuliaLang/julia/blob/a90a4472dbbcce6fe414eafed0bebd88a5b75a9a/base/Makefile#L197

ViralBShah:
Probably more an oversight.

PR #34327, opened 2020-01-09
[WIP] Update openlibm to 0.7. 

ViralBShah:
Use artifact binaries. Disable gcc platform expansion.

Perhaps someone knows why we do gcc platform expansion for openlibm. @staticfloat ? openlibm should be like many of the simple C libraries we ship (like pcre or dsfmt) where we do not do gcc platform expansion.


staticfloat:
@ViralBShah I think https://github.com/JuliaLang/julia/pull/34242 will help answer your questions.  It's a more comprehensive dependencies bump, and includes the elimination of the OpenLibm -> libgfortran dependency.

ViralBShah:
Ah, I didn't realize it also included the openlibm version bump. Thanks. At least I learnt a bit more about the inner workings of the various build systems as a result of this.

PR #34328, opened 2020-01-09
Sysinfo kilobytes -> Bytes change

### LINKED ISSUE #34294: Sys.free_memory()/total_memory() documentation incorrect (returns bytes, not kbytes)
Both Sys.free_memory() and Sys.total_memory() return free memory measured in bytes, not kilobytes as the documentation claims.

This can be verified in the documentation for libuv, eg. http://docs.libuv.org/en/v1.x/misc.html#c.uv_get_free_memory

Both documentation should be amended to read bytes.

Chetank99:
Fixes: https://github.com/JuliaLang/julia/issues/34294

### DIFF
--- a/base/sysinfo.jl
+++ b/base/sysinfo.jl
@@ -242,14 +242,14 @@ end
 """
     Sys.free_memory()
 
-Get the total free memory in RAM in kilobytes.
+Get the total free memory in RAM in bytes.
 """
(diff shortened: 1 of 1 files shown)


musm (approved):

StefanKarpinski (approved):

JeffBezanson:
Thanks!

Issue #29333, opened 2018-09-23
interpret call f(; x, y) as f(; x=x, y=y)


KristofferC:
Sorry if this is a dup but I couldn't find one.

I find it quite common that one calls functions with keyword arguments like

```jl
f(; x=x, y=y, this_was_a_bit_long=this_was_a_bit_long)
```

My proposal is to write this as

```jl
f(;x, y, this_was_a_bit_long)
```

and to parse that to the first example.

Since

```jl
f(; x, y)
```

is already a syntax error, I think this should be possible?

yuyichao:
This is implementable as a macro if you really need it and it doesn't enabling doing anything that couldn't be done while making the syntax much more confusing (even more so than the significance of `;` in the function definition).

It'll also give `f(; :a=>b)` and `x = :a=>b; f(; x)` completely different meanings.

KristofferC:
> This is implementable as a macro 

Of course, the whole point of introducing this as syntax is to avoid a macro.

> It'll also give `f(; :a=>b)` and `x = :a=>b; f(; x)` completely different meanings.

`f(a=b)` and `d = a=b; f(d)` already have completely different meanings. I don't think people find that very confusing.

yuyichao:
> `f(a=b)` and `x = a=b; f(x)`

For start, `=` is not an operator so that's very different.

And it did cause issues with macros in corner cases.

ararslan:
My biggest concern over this is that really it shouldn't matter what name you've given to a variable, but in this case you have to be careful if you rename anything, because it can subtly change behavior when using keyword arguments. Instead of just being able to rename your variable, you have to comb through all potential uses of it in this context, and if you miss one you get an unsupported keyword argument error (if you're lucky) or a different result entirely (if you aren't).

yuyichao:
> Renaming the variable a to some other name might give you a completely different behavior (e.g by calling it s).

This isn't a defence at all. It's all about the "subtleness" of the change and more precisely, the localness of the effect.

Of course you can't rename a variable that override another one without any observable effect. You don't even need that `let`, `a = 1; b = 2` will obviously be different if you rename `b` to `a`. However, the effect is caused by something that is **in the same scope** (by which I include parents and children ones). You can easily infer locally whether the override can happen or not. This is also why macro hygiene is so important, you can't expect the user to avoid naming conflict with name introduced in a invisible way by the macro author.

JeffBezanson:
This is already implemented here, commented out:
[...]

PR #34331, opened 2020-01-09
add implicit named tuple and keyword argument names

JeffBezanson:
This implements #29333.

I went ahead and added dot syntax as well, i.e. `a.b` implies `b=a.b`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,6 +11,11 @@ New language features
   macros and matrix constructors, which are whitespace sensitive, because expressions like
   `[a ±b]` now get parsed as `[a ±(b)]` instead of `[±(a, b)]`. ([#34200])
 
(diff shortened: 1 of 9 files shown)


ararslan:
Based on the doctest failure, it seems this inadvertently made unprovided required kwargs produce an `UndefVarError` instead of the expected `UndefKeywordError`.

tpapp:
I think the dot syntax part, ie `plot(x, y; options.width)` being equivalent to `plot(x, y; width=options.width)`, is a bit too much for now. I think would be better to just experiment with the feature as described in #29333 first.

JeffBezanson:
I understand wanting to be cautious here, but I'm pretty sure languages tend to have either both of these features or neither. Also the `a.b` case should be less controversial, since `b` is already implicitly quoted, so there is no worry about e.g. renaming variables changing the meaning of a program.

PR #34339, opened 2020-01-10
Cache lookup of matching methods between inference and inlining

yhls:
This reduces time spent in `METHOD_MATCH` by roughly 40%, as measured with `#define ENABLE_TIMINGS` in options.h. I found that, during inference, attempting to look up methods in the cache gave slightly better performance than only adding to it then.

We save ~2% of the time in compiling the sysimage, for which `METHOD_MATCH` currently takes ~5% of the time. We save about 7% of the time for `julia -e "using Plots; using DataFrames; display(plot(rand(10), rand(10)))"`, which currently spends ~20% of the time in `INFERENCE` (which includes inlining) and ~15% in `METHOD_MATCH`.

### DIFF
--- a/base/compiler/abstractinterpretation.jl
+++ b/base/compiler/abstractinterpretation.jl
@@ -42,12 +42,22 @@ function abstract_call_gf_by_type(@nospecialize(f), argtypes::Vector{Any}, @nosp
         splitsigs = switchtupleunion(atype)
         applicable = Any[]
         for sig_n in splitsigs
-            xapplicable = _methods_by_ftype(sig_n, max_methods, sv.params.world, min_valid, max_valid)
+            (xapplicable, min_valid[1], max_valid[1]) =
+                get!(sv.matching_methods_cache, sig_n) do
+                    ms = _methods_by_ftype(sig_n, max_methods, sv.params.world,
+                                           min_valid, max_valid)
+                    return (ms, min_valid[1], max_valid[1])
+                end
             xapplicable === false && return Any
             append!(applicable, xapplicable)
         end
     else
-        applicable = _methods_by_ftype(atype, max_methods, sv.params.world, min_valid, max_valid)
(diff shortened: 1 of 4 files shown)


base/compiler/abstractinterpretation.jl
```diff
@@ -42,12 +42,24 @@ function abstract_call_gf_by_type(@nospecialize(f), argtypes::Vector{Any}, @nosp
         splitsigs = switchtupleunion(atype)
         applicable = Any[]
         for sig_n in splitsigs
-            xapplicable = _methods_by_ftype(sig_n, max_methods, sv.params.world, min_valid, max_valid)
+            (xapplicable, min_valid[1], max_valid[1]) =
+                get(sv.matching_methods_cache, sig_n) do
```
JeffBezanson: This can use `get!`; then the explicit assignment into the cache below is not needed.

yhls: Oh, thanks--amended.

base/compiler/inferencestate.jl
```diff
@@ -41,6 +41,10 @@ mutable struct InferenceState
     inferred::Bool
     dont_work_on_me::Bool
 
+    # cached results of calling `_methods_by_ftype`, including `min_valid` and
+    # `max_valid`, to be used in inlining
+    matching_methods_cache::IdDict{DataType, Tuple{Any, UInt, UInt}}
```
JeffBezanson:
```suggestion
    matching_methods_cache::IdDict{Any, Tuple{Any, UInt, UInt}}
```
@vtjnash what do you think?

JeffBezanson (approved):

yhls:
By latest measurements, we save about 12% of the time for `julia -e "using Plots; using DataFrames; display(plot(rand(10), rand(10)))"` overall, although the `display` part is only 2% faster.

Issue #25233, opened 2017-12-21
Reserve some syntax for a tee pipe


davidanthoff:
R's magrittr has the tee pipe ``%T>%.`` The difference with the normal pipe ``%>%`` is that an expression like ``a %T>% b`` returns ``a``, not ``b``. This can be super useful in data processing pipelines a la dplyr (and what I'm currently building out with the standalone operators in [Query.jl](https://github.com/davidanthoff/Query.jl)).

It would be great if julia could at some point get something like that as well. One option would be ``|>>``. It would kind of signal that the stuff on the left hand side is going to be moved one more over in a pipe, i.e. something like ``a |>> b |> c`` to me looks fairly reasonable in that it signals that the return value of ``a`` would "jump" over ``b``. But I'm sure there are other ideas as well, so I'm not super wedded to this specific syntax.

It is not clear to me whether something like ``|>>`` could be introduced in julia 1.x, or whether that would be breaking. When I do ``a |>>b`` I get an error ``ERROR: syntax: ">" is not a unary operator``. I think that means one could define ``|>>`` in a julia 1.x without breaking anything, but is that right?

If that is correct, great. But if not, i.e. if this would be breaking, is there a chance that we could make some syntax illegal for julia 1.0, with the plan to later enable this kind of tee pipe in some 1.x version?

davidanthoff:
@StefanKarpinski mentioned somewhere that we should bump issues that we would like the core team to look at before the alpha. Here we go, BUMP.

Most likely that answer here is just that this could be done in 1.x if there is agreement, and we can move on for now. But I'm not sure, so would be great if someone with a better understanding could look at this.

StefanKarpinski:
There are a few cases to check and they're all errors currently:
```jl
julia> |>>
ERROR: syntax: "|>" is not a unary operator

julia> 1 |>>
ERROR: MethodError: no method matching >(::Int64)
Closest candidates are:
  >(::Any, ::Any) at operators.jl:283
  >(::BigFloat, ::BigFloat) at mpfr.jl:698
Stacktrace:
 [1] |>(::Int64, ::typeof(>)) at ./operators.jl:774
 [2] top-level scope

julia> |>> 2
ERROR: syntax: "|>" is not a unary operator

julia> 1 |>> 2
ERROR: syntax: ">" is not a unary operator
```

davidanthoff:
Cool, thanks!

davidanthoff:
[...]

PR #34340, opened 2020-01-10
RFC: Add new pipe-like operator syntaxes `</`, `|>>`, etc.

tkf:
This patch adds pipe-like operators such as `</` and `|>>` (as requested in #25233) etc.  All combinations of "heads" `<`, `<<`, `>`, and `>>` and "tails" `|`, `/` and `\` are supported.

To increase the probability of this PR to be merged, I'd like to avoid implementing any interface in `Base` on top of these syntaxes here.  But the motivating example is to have the opposite composition operator

```julia
/>(args...) = ∘(reverse(args)...)
```

so that one can have

```julia
1 |> sin |> cos |> tan == 1 |> (sin /> cos /> tan)
```

This is useful when you have an interface where using successive `|>` is a natural way to call functions (e.g., Query.jl, Transducers.jl).  Then, the pipeline part can be factored out by using the opposite composition operator `/>`.  For symmetry, it may be useful to define `</` to be the usual composition operator.  This would give us an ASCII alias of `∘` #33573. (Note: the associativity of `</` and `∘` are different. But composition is associative so I think it's OK.) See discussion started at https://github.com/tkf/Transducers.jl/issues/67#issuecomment-572744009 for more explanations on the motivation.

Another motivating usecase of a pipe-like operator `|>>` is discussed in #25233 and #34344.

cc @davidanthoff Maybe you are interested in new operators like this too?


davidanthoff:
Hm, to be honest, not clear to me that this is a high priority thing? At least from the Query.jl side of things I haven't missed it :) But I might be missing something deeper here.

I think if it gets added, it should have an implementation with it, otherwise I don't see how one can add consistent functionality to it without type piracy?

tkf:
You can use operator names overlapping with `Base` without any piracy

```julia
julia> module MyModule
       *(args...) = foldr(Base.:*, args)
       end
Main.MyModule

julia> MyModule.:* === Base.:*
false
```

But yes, I guess I'll open a PR to add `Base./>` after this PR.  I just want to minimize the patch to get this merged as fast as possible.

tkf:
In https://github.com/JuliaLang/julia/issues/25233#issuecomment-376641367, @StefanKarpinski mentioned that operators like this can be added even though they are technically breaking due to

```juila
julia> :(_ />)
:(_ / (>))

julia> VERSION
v"1.3.1"
```

JeffBezanson:
`|>>` is maybe ok, but I find the pipes with the slashes pretty confusing.

tkf:
Why is it confusing? Do you mean it doesn't have `|` while it is listed in `prec-pipe`? Would it be OK if they are moved to other list (e.g., `prec-times`)? My motivation is to have some variants of `|>` (or more precisely anything usable as an ASCII opposite composition operator).

JeffBezanson:
They're confusing because they don't really visually convey to me what they do, nor are they used by any other system that I'm aware of.

StefanKarpinski:
This seems very syntax-greedy. We often find ourselves in the situation of needing syntax for something and every last bit of available possible syntax is valuable. Why is having this many pipelike operators a good way to spend syntax? Some of them I can see, but all of these? Really? You can't use functions or one of the billion Unicode operators that Julia supports?

Some other points from triage: other systems don't seem to support most of these—especially the slash and backslash variants—so this seems to be totally unprecedented operator territory. They're also just visually confusing to parse. This is veering deep in the Perl ASCII salad territory. When someone sees `\>>` why is "oh, that's a single operator" be obvious?

davidanthoff:
https://github.com/JuliaLang/julia/pull/34344 is a PR that only adds the `|>>` case, which is the part that I would like to see (mostly for Query.jl).

tkf:
@StefanKarpinski

> Why is having this many pipelike operators a good way to spend syntax? Some of them I can see, but all of these?

My primary motivation was to add one or two operators (say `/>`).  So I'm willing to reduce the number of operators in this PR if it increases the chance for it to be merged.

But since the parser allows `<|` even though `Base` does not have the definition of it, I thought to follow the same "convention" (though maybe I misread the intention) and define other operators for symmetry.  By symmetry, I mean that once `/>` is introduced, it's strange that `x />` is a syntax error while `x </` is not.  Another reasoning was to reduce future syntax breakage in case other combinations are needed; adding `</` has the effect of making `x </` a syntax error.

> When someone sees `\>>` why is "oh, that's a single operator" be obvious?

But you don't see infix operators in isolation.  If you see `x />> y />> z`, it's hard to _not_ think that's an infix operator.

> You can't use functions

As [Dijkstra also noted](https://www.cs.utexas.edu/users/EWD/transcriptions/EWD13xx/EWD1300.html), infix operator is great because you don't have to put parentheses.  This is a strong indication that the operation is associative.  Usual function call does not indicate this algebraic property.

> or one of the billion Unicode operators that Julia supports?

Yes, it's great that Julia provides Unicode as a fallback option for situation like this.  But I do hope that fundamental operations such as composition and opposite composition can be done with ASCII characters.

---

@JeffBezanson

> They're confusing because they don't really visually convey to me what they do

I think `/>` and `</` still convey a sense of directionality.  Arguably this nuance is missing in more standard (and unfortunately non-ASCII) notations `∘` and `∘ᵒᵖ`.  I think `/>` and `</` also visually similar to `|>` and `<|`.  Also, by appearing at the infix position, they visually indicate that the operation is likely associative.

Other than that, I agree that it's impossible to infer the semantics of the operation. But I think it can be said for any less famous infix operators in programming.

> nor are they used by any other system that I'm aware of.

I know similar operators [`<<<` and `>>>`](https://hackage.haskell.org/package/base-4.12.0.0/docs/Control-Arrow.html#v:-62--62--62-) are defined in a Haskell library.  Unfortunately, `>>>` is already defined in Julia.  (Though I guess "Haskell has it" is not a convincing argument...)

---

If the concern is that `/>` does not look like a pipe operator, would it be better to introduce something like `|->`, `|+>`, `|*>`, or `|%>`?  Operator `|!>` may make sense for #34344 to indicate that the operation is mutation or side-effect.

yurivish:
I don’t have an opinion on this issue, but I thought I would mention that the operators with slashes read strongly as HTML/XML to me.

StefanKarpinski:
> If you see `x />> y />> z`, it's hard to _not_ think that's an infix operator.

Sure. But when I see `x/>>y/>>z` I don't really see an infix operator. It's also entirely possible that there's some other syntax involving `/` and `>` that would currently be possible to add without breaking any code which the addition of this infix operator would prevent. It's possible that we could add these, but it's just not so clear-cut. There seems to be a common feeling that syntax is free and that there's no cost to adding new syntaxes, but there is a very real opportunity cost.

tkf:
> There seems to be a common feeling that syntax is free and that there's no cost to adding new syntaxes, but there is a very real opportunity cost.

I think I understand that ASCII syntaxes are real estates that should not be wasted and tweaking syntaxes in general is a one-way street.  That's actually why I suggested adding all the combinations.  Right now, `a/>>` is parsed as `a / (>>)` which is wasting the valuable ASCII syntax.

But, admittedly, this is not exactly the best "fix" in the sense we can be even more cautious.  For example, maybe it's better to make consecutive operators without spaces between them a syntax error?  Even if we don't use `/>` for pipe operator, it may be useful for something else in the future (e.g., for XML/HTML DSL like @yurivish just mentioned).

Also, I still would like to add one rightward pipe operator (to be used for opposite the composition operator) and, ideally, another leftward pipe operator that is the flip version of it.  How about `|->` (and `<-|`)?

c42f:
This is pretty interesting. If we want a stronger visual analogy to piping which includes the `|` char, perhaps `<<|` and `|>>` would be good for composition rather than other uses.

With `|>>` for reverse composition we have

```
data |> (f |>> g |>> h)  ==
data |>  f |>  g |>  h
```

which has a nice symmetry.  (There's also `||>` and `<||` as potential options?)

It might be good to emphasize again the mathematical point of view which unifies transducers and iterator transforms into a single category with left and right actions https://github.com/tkf/Transducers.jl/issues/67#issuecomment-572595547. That seems like an important enough observation to justify some kind of syntax eventually.

A couple of other thoughts — somehow I find it troubling that `∘` vs `<<|` would have different associativity; unrelatedly I very much like the suggestion of `|!>` as side-effectful piping.

PR #24990, opened 2017-12-08
RFC: curry underscore arguments to create anonymous functions

stevengj:
This PR addresses JuliaLang/julia#554, JuliaLang/julia#5571, and JuliaLang/julia#22710 by "currying" underscores in function calls like `f(_,y)` into anonymous function expressions `x -> f(x,y)`.  (Note that `_.foo` works and turns into `x -> x.foo` since it is equivalent to a `getfield` call, and `_[i]` works and turns into `x -> x[i]`, since it is equivalent to a `getindex(_,i)` call.)

This will help us get rid of functions like `equalto` (#23812) or `occursin` (#24967), is useful for "destructuring" as discussed in JuliaLang/julia#22710, and should generally be convenient in lots of cases to avoid having to explicitly do `x -> f(x,y)`.

Some simplifying design decisions that I made:

* The currying is "tight", i.e. it only converts the immediately surrounding function call into a lambda, [as suggested by](https://github.com/JuliaLang/julia/issues/22710#issuecomment-314259250) @JeffBezanson (and [as in Scala](https://www.scala-lang.org/files/archive/spec/2.11/06-expressions.html#placeholder-syntax-for-anonymous-functions)).   So, e.g. `f(g(_,y))` is equivalent to `f(x -> g(x,y))`.   (Note that something like `find(!(_ in c), y)` will work fine, because the `!` operator works on functions; you can also use `_ ∉ c`.)  Any other rule seems hard to make comprehensible and consistent.

* ~~Only a single underscore is allowed.  `f(_,_)` throws an error: this case seems ambiguous to me (do you want `x -> f(x,x)` or `x,y -> f(x,y)`?), so it seemed better to punt on this for now.  We can always add a meaning for multiple underscores later.~~ Similar to Scala, multiple underscores are converted into multiple arguments in the order they appear.  e.g. `f(_,y,_)` is equivalent to `(x,z) -> f(x,y,z)`.  See [rationale below](https://github.com/JuliaLang/julia/pull/24990#issuecomment-350879561).

The implementation is pretty trivial.  If people are in favor, I will add
- [x] Tests
- [x] Documentation
- [ ] Fix interaction with broadcasting `f.(x, _)`
- [ ] Fix interaction with keyword arguments `f(x; y=_)`
- [ ] Support chained comparisons, e.g. `3 ≤ _ ≤ 10`, since they parse as a single expression?

PR #34344, opened 2020-01-11
RFC: Add support for tee pipe |>> syntax and function

### LINKED ISSUE #25233: Reserve some syntax for a tee pipe

davidanthoff:
Fixes #25233.

This allows for tee piping, i.e. it works like the standard pipe operator `|>`, but returns the left side argument. This can be super useful in typical data science workflows, for example R's magrittr adds that as `%T>%` in R. I would love to have this for [Query.jl](https://github.com/queryverse/Query.jl).

I have _absolutely_ no idea what I'm doing with the parser changes. @tkf suggested that change, and I just copied it :) I have no idea what precedence this ends up with etc., the whole code in the scm file is entirely unclear to me. So someone who does understand that needs to review carefully, if this idea was generally approved.


tkf:
Oops, I was just adding `|>>` and `<<|` in https://github.com/JuliaLang/julia/pull/34340/commits/e5a0a8c08be6562a27ac5f300bb54a27c2f8bb1c as I thought it might be nice to review the syntax part in one go.

base/operators.jl
```diff
@@ -822,6 +822,17 @@ julia> [1:5;] |> x->x.^2 |> sum |> inv
 """
 |>(x, f) = f(x)
 
+"""
+    |>>(x, f)
+
+Applies a function to the preceding argument, and returns that argument.
+This allows for easy tee function chaining.
```
tkf: Is "tee function chaining" a popular terminology for this?

Also, I guess it'd be nice to have some motivating examples as doctests. Is it for mutating functions like `pop!` and side-effects like `display`?

davidanthoff: > Is "tee function chaining" a popular terminology for this?

I have no idea :) I think "tee" is used for this, but if the general idea here gets accepted, we should revisit the terminology.

The main use case I've found is file saving in queryverse:
```julia
df |> @filter(...) |>> save("foo.csv") |> @vlplot(...) |>> save("foo.pdf")
```
is just something I'd like a lot :) The main use case I have is side-effects. Your display example is another one.

davidanthoff:
> Oops, I was just adding |>> and <<| in e5a0a8c as I thought it might be nice to review the syntax part in one go.

Cool, thanks! I'll leave this one open here for feedback, and if it gets the nod we can still figure out whether we should just merge the combined PR.

base/operators.jl
tkf: FYI, Python has [`itertools.tee`](https://docs.python.org/3/library/itertools.html#itertools.tee) which has a different meaning (especially when the input is a "stateful" iterator).  That's why I wondered if it has a standard meaning. But clarifying the terminology after this is accepted makes sense.

I think `save` example is appealing. I guess an objection may be that it can already be done if `FileIO.save` returns the second argument.  But I suppose there are cases where this is not possible.

nickrobinson251: there is also https://en.wikipedia.org/wiki/Tee_(command)

davidanthoff: Yes, that is the one where I got the name from!

tkf: The `tee` command is closer to Python's `itertools.tee` than `|>>`.

I guess the question is if it is "allowed" to specialize `|>>(x, f)` in case `x` is some kind of "lazy" object (e.g., iterators) to minimize some overheads.  For example, in principle, it is possible to run filtering in `@filter(...) |>> save("foo.csv") |> @vlplot(...)` only once.  It would then be closer to the `tee` command.

Roger-luo:
I came across from discourse: https://discourse.julialang.org/t/parse-pipe-operator-for-side-effect-calls/38857 I also pro the idea of in-place pipe operator `|!>` since we would like to emphasis the side effect due to non-cloning theorem in our API.

twolodzko:
There's an interesting comment about different usage of the possible `|>>` operator: https://github.com/JuliaLang/julia/pull/24990#issuecomment-605364343

The proposed usage is to make `|>>` an operator that passes input as a second argument to the function. So if

```julia
foo(a=1, b=2) = (a, b)

3 |> foo  # => (2, 2)
3 |> foo  # => (1, 3)
```

Similar operators `->` and `->>` [are available in Clojure ](https://stackoverflow.com/questions/19318977/difference-between-arrow-and-double-arrow-macros-in-clojure), so I'd say this syntax is more intuitive for the Clojure's operator, rather than the tee pipe.

max-sixty:
Does this need to start as a _language_ change? Could this be a function `tee`, like:

```
df |> @filter(...) |> tee(save("foo.csv")) |> @vlplot(...) |> tee(save("foo.pdf"))
```
...assuming `save` works the same as the original example.

where the function is the same as the one in the PR:
```
function tee(x, f)
    f(x)
    return x
end
```

max-sixty:
Does this need to start as a _language_ change? Could this be a function `tee`, like:

```julia
df |> @filter(...) |> tee(save("foo.csv")) |> @vlplot(...) |> tee(save("foo.pdf"))
```

Separately — I'm unclear on the signature of `save` is in the original example — please could you clarify?

davidanthoff:
Yes, one can do that without new syntax, for example https://github.com/queryverse/Queryverse.jl had a `@tee` macro for a long time. This PR/issue is about adding a nicer syntax for that.

StefanKarpinski:
It seems fine to have some more pipelike operators that people can use for whatever they want. Any concerns, @JeffBezanson, Syntax Czar?

LilithHafner:
The default answer for new syntax is no, unless it's a slam dunk.

PR #34345, opened 2020-01-11
Tty stdout color

### LINKED ISSUE #34317: enable color output whenever stdout is a TTY

karan0299:
Fixes #34317 . Made a check if STDOUT is a TTY and color is not set explicitly by the user , then set global variable `have_color` to true.
In my opinion , there should be separate `have_color` variable for STDOUT and STDERR, since we can have separate TTY for STDOUT and STDERR


jpsamaroo:
Seems like you based this off of the wrong branch (see all the unrelated commits); you will probably need to rebase your commit on top of master and then force-push to fix it.

karan0299:
I am sorry , I am closing this pull request due to huge number of merge conflicts. I have open a new one 
that fixes the given issue

Issue #30126, opened 2018-11-23
`f() = 2^5` not compiled all the way to result


jakobnissen:
Have a look at:
```
julia> f() = 2^5; @code_native f()
    .text
; Function f {
; Location: none:1
; Function literal_pow; {
; Location: none
; Function macro expansion; {
; Location: none
; Function ^; {
; Location: none:1
    pushq   %rax
    movabsq $power_by_squaring, %rax
    movl    $2, %edi
    movl    $5, %esi
    callq   *%rax
;}}}
    popq    %rcx
    retq
    nopl    (%rax)
;}
```

This behaviour is weird. Since the result is completely fixed, shouldn't it compile all the way down to "return this predefined number" instead of requiring a new function call?
To make things weirder, the following functions *do* compile down to "return this number":
`m() = 2^8; n() = 3^4; g() = 2.0^5; h() = 2^5.0, k() = 1<<5` 
In practice, this means that running `k()` is unnecessarily ~3x faster than `f()`. Is this a bug that should be fixed?

### Environment:
```
julia> versioninfo()
Julia Version 1.0.0
Commit 5d4eaca0c9 (2018-08-08 20:58 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Core(TM) i5-4210H CPU @ 2.90GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.0 (ORCJIT, haswell)
Environment:
  JULIA_EDITOR = atom -a
  JULIA_NUM_THREADS = 2
```

ranocha:
Is there any update? On my system, I get nice results of `literal_pow` for powers two and three, but bad results starting at four.
```julia
julia> versioninfo()
Julia Version 1.5.0-beta1.0
Commit 6443f6c95a (2020-05-28 17:42 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Core(TM) i7-8700K CPU @ 3.70GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-9.0.1 (ORCJIT, skylake)
Environment:
  JULIA_NUM_THREADS = 1

julia> foo(x) = x^4
foo (generic function with 1 method)

julia> @code_llvm foo(5.0)

;  @ REPL[24]:1 within `foo'
define double @julia_foo_3022(double) {
top:
[...]

PR #34346, opened 2020-01-11
@inline_meta instead of @inline @generated in literal_pow

tkoolen:
Happened to come across this. It seems to be the only such case in Base.


tkoolen:
This does *not* appear to help with https://github.com/JuliaLang/julia/issues/30126 by the way.

aviatesk:
Close in favor of https://github.com/JuliaLang/julia/pull/40408.

PR #34347, opened 2020-01-11
Enable color output whenever stdout is TTY

### LINKED ISSUE #34317: enable color output whenever stdout is a TTY

karan0299:
Fixes #34317 . Made a check if STDOUT is a TTY and color is not set explicitly by the user , then set global variable have_color to true.
In my opinion , there should be separate have_color variable for STDOUT and STDERR, since we can have separate TTY for STDOUT and STDERR

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -51,9 +51,8 @@ Language changes
   used to yield the string " a\nb", since the single space before `b` set the indent level.
   Now the result is "a\n b", since the space before `b` is no longer considered to occur
   at the start of a line. The old behavior is considered a bug ([#35001]).
-  
-* Color now defaults to on when stdout and stderr are TTYs ([#34347])
 
+* Color now defaults to on when stdout and stderr are TTYs ([#34347])
 
 Multi-threading changes
 -----------------------
--- a/NEWS.md
+++ b/NEWS.md
@@ -51,6 +51,8 @@ Language changes
   used to yield the string " a\nb", since the single space before `b` set the indent level.
   Now the result is "a\n b", since the space before `b` is no longer considered to occur
   at the start of a line. The old behavior is considered a bug ([#35001]).
+  
+* Color now defaults to on when stdout and stderr are TTYs ([#34347])
 
 
 Multi-threading changes
--- a/base/Base.jl
+++ b/base/Base.jl
@@ -241,6 +241,7 @@ include("filesystem.jl")
 using .Filesystem
 include("cmd.jl")
 include("process.jl")
+include("ttyhascolor.jl")
 include("grisu/grisu.jl")
 include("secretbuffer.jl")
 
--- a/base/client.jl
+++ b/base/client.jl
@@ -493,7 +493,7 @@ function _start()
         invokelatest(display_error, catch_stack())
         exit(1)
     end
-    if is_interactive && have_color
+    if is_interactive && have_color === true
         print(color_normal)
     end
 end
--- a/base/client.jl
+++ b/base/client.jl
@@ -112,7 +112,7 @@ function eval_user_input(errio, @nospecialize(ast), show_value::Bool)
     have_color = get(stdout, :color, false)
     while true
         try
-            if have_color === true
+            if have_color
                 print(color_normal)
             end
             if lasterr !== nothing
@@ -124,7 +124,7 @@ function eval_user_input(errio, @nospecialize(ast), show_value::Bool)
                 value = Core.eval(Main, ast)
                 ccall(:jl_set_global, Cvoid, (Any, Any, Any), Main, :ans, value)
                 if !(value === nothing) && show_value
-                    if have_color === true
+                    if have_color
                         print(answer_color())
                     end
                     try
--- a/base/client.jl
+++ b/base/client.jl
@@ -112,7 +112,7 @@ function eval_user_input(errio, @nospecialize(ast), show_value::Bool)
     have_color = get(stdout, :color, false)
     while true
         try
-            if have_color
+            if have_color === true
                 print(color_normal)
             end
             if lasterr !== nothing
@@ -124,7 +124,7 @@ function eval_user_input(errio, @nospecialize(ast), show_value::Bool)
                 value = Core.eval(Main, ast)
                 ccall(:jl_set_global, Cvoid, (Any, Any, Any), Main, :ans, value)
                 if !(value === nothing) && show_value
-                    if have_color
+                    if have_color === true
                         print(answer_color())
                     end
                     try
@@ -493,7 +493,7 @@ function _start()
         invokelatest(display_error, catch_stack())
         exit(1)
     end
-    if is_interactive && have_color
+    if is_interactive && have_color === true
         print(color_normal)
     end
 end
--- a/base/client.jl
+++ b/base/client.jl
@@ -109,6 +109,7 @@ display_error(er, bt=nothing) = display_error(stderr, er, bt)
 function eval_user_input(errio, @nospecialize(ast), show_value::Bool)
     errcount = 0
     lasterr = nothing
+    have_color = get(stdout, :color, false)
     while true
         try
             if have_color
@@ -210,19 +211,7 @@ end
 
 # call include() on a file, ignoring if not found
 include_ifexists(mod::Module, path::AbstractString) = isfile(path) && include(mod, path)
-function ttyhascolor()
-    term_type = get(ENV, "TERM","")
-    startswith(term_type, "xterm") && return true
-    try
-        @static if Sys.KERNEL === :FreeBSD
-            return success("tput AF 0")
-        else
-            return success("tput setaf 0")
-        end
-    catch
-        return false
-    end
-end
+
 function exec_options(opts)
     if !isempty(ARGS)
         idxs = findall(x -> x == "--", ARGS)
@@ -232,13 +221,8 @@ function exec_options(opts)
     startup               = (opts.startupfile != 2)
     history_file          = (opts.historyfile != 0)
     color_set             = (opts.color != 0) # --color!=auto
-    global have_color     = (opts.color == 1) # --color=on
+    global have_color     = color_set ? (opts.color == 1) : nothing # --color=on
     global is_interactive = (opts.isinteractive != 0)
-
-    #check if stdout is TTY and color is not set manually
-    if !color_set && isa(stdout, TTY) && isa(stderr, TTY) && ttyhascolor()
-        global have_color = true
-    end
 
     # pre-process command line argument list
     arg_is_program = !isempty(ARGS)
--- a/base/client.jl
+++ b/base/client.jl
@@ -3,7 +3,7 @@
 ## client.jl - frontend handling command line options, environment setup,
 ##             and REPL
 
-have_color = false
+have_color = nothing
 default_color_warn = :yellow
 default_color_error = :light_red
 default_color_info = :cyan
@@ -210,7 +210,19 @@ end
 
 # call include() on a file, ignoring if not found
 include_ifexists(mod::Module, path::AbstractString) = isfile(path) && include(mod, path)
-
+function ttyhascolor()
+    term_type = get(ENV, "TERM","")
+    startswith(term_type, "xterm") && return true
+    try
+        @static if Sys.KERNEL === :FreeBSD
+            return success("tput AF 0")
+        else
+            return success("tput setaf 0")
+        end
+    catch
+        return false
+    end
+end
 function exec_options(opts)
     if !isempty(ARGS)
         idxs = findall(x -> x == "--", ARGS)
@@ -224,7 +236,7 @@ function exec_options(opts)
     global is_interactive = (opts.isinteractive != 0)
 
     #check if stdout is TTY and color is not set manually
-    if isa(stdout, TTY) && !color_set
+    if !color_set && isa(stdout, TTY) && isa(stderr, TTY) && ttyhascolor()
         global have_color = true
     end
 
--- a/base/client.jl
+++ b/base/client.jl
@@ -223,6 +223,11 @@ function exec_options(opts)
     global have_color     = (opts.color == 1) # --color=on
     global is_interactive = (opts.isinteractive != 0)
 
+    #check if stdout is TTY and color is not set manually
+    if isa(stdout, TTY) && !color_set
+        global have_color = true
+    end
+
     # pre-process command line argument list
     arg_is_program = !isempty(ARGS)
     repl = !arg_is_program
--- a/base/loading.jl
+++ b/base/loading.jl
@@ -1169,10 +1169,11 @@ function create_expr_cache(input::String, output::String, concrete_deps::typeof(
             eval(Meta.parse(code))
         end
         """
+
     io = open(pipeline(`$(julia_cmd()) -O0
                        --output-ji $output --output-incremental=yes
                        --startup-file=no --history-file=no --warn-overwrite=yes
-                       --color=$(have_color ? "yes" : "no")
+                       --color=$(have_color === true ? "yes" : "no")
                        --eval $code_object`, stderr=stderr),
               "w", stdout)
     in = io.in
--- a/base/stream.jl
+++ b/base/stream.jl
@@ -485,29 +485,6 @@ function displaysize(io::TTY)
     return h, w
 end
 
-function ttyhascolor()
-    term_type = get(ENV, "TERM","")
-    startswith(term_type, "xterm") && return true
-    try
-        @static if Sys.KERNEL === :FreeBSD
-            return success("tput AF 0")
-        else
-            return success("tput setaf 0")
-        end
-    catch
-        return false
-    end
-end
-function get_have_color()
-    global have_color
-    have_color === nothing && (have_color = ttyhascolor())
-    return have_color::Bool
-end
-in(key_value::Pair{Symbol,Bool}, ::TTY) = key_value.first === :color && key_value.second === get_have_color()
-haskey(::TTY, key::Symbol) = key === :color
-getindex(::TTY, key::Symbol) = key === :color ? get_have_color() : throw(KeyError(key))
-get(::TTY, key::Symbol, default) = key === :color ? get_have_color() : default
-
 ### Libuv callbacks ###
 
 ## BUFFER ##
--- a/base/stream.jl
+++ b/base/stream.jl
@@ -485,10 +485,28 @@ function displaysize(io::TTY)
     return h, w
 end
 
-in(key_value::Pair{Symbol,Bool}, ::TTY) = key_value.first === :color && key_value.second === have_color
+function ttyhascolor()
+    term_type = get(ENV, "TERM","")
+    startswith(term_type, "xterm") && return true
+    try
+        @static if Sys.KERNEL === :FreeBSD
+            return success("tput AF 0")
+        else
+            return success("tput setaf 0")
+        end
+    catch
+        return false
+    end
+end
+function get_have_color()
+    global have_color
+    have_color === nothing && (have_color = ttyhascolor())
+    return have_color::Bool
+end
+in(key_value::Pair{Symbol,Bool}, ::TTY) = key_value.first === :color && key_value.second === get_have_color()
 haskey(::TTY, key::Symbol) = key === :color
-getindex(::TTY, key::Symbol) = key === :color ? have_color : throw(KeyError(key))
-get(::TTY, key::Symbol, default) = key === :color ? have_color : default
+getindex(::TTY, key::Symbol) = key === :color ? get_have_color() : throw(KeyError(key))
+get(::TTY, key::Symbol, default) = key === :color ? get_have_color() : default
 
 ### Libuv callbacks ###
 
--- a/base/ttyhascolor.jl
+++ b/base/ttyhascolor.jl
@@ -1,0 +1,24 @@
+function ttyhascolor()
+    term_type = get(ENV, "TERM","")
+    startswith(term_type, "xterm") && return true
+    try
+        @static if Sys.KERNEL === :FreeBSD
+            return success(`tput AF 0`)
+        else
+            return success(`tput setaf 0`)
+        end
+    catch e
+        return false
+    end
+end
+function get_have_color()
+    global have_color
+    have_color === nothing && (have_color = ttyhascolor())
+    return have_color
+end
+in(key_value::Pair{Symbol,Bool}, ::TTY) = key_value.first === :color && key_value.second === get_have_color()
+haskey(::TTY, key::Symbol) = key === :color
+getindex(::TTY, key::Symbol) = key === :color ? get_have_color() : throw(KeyError(key))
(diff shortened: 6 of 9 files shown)


base/client.jl
```diff
@@ -223,6 +223,11 @@ function exec_options(opts)
     global have_color     = (opts.color == 1) # --color=on
     global is_interactive = (opts.isinteractive != 0)
 
+    #check if stdout is TTY and color is not set manually
+    if isa(stdout, Base.TTY) && color_set == false
```
ararslan:
```suggestion
    if isa(stdout, TTY) && !color_set
```
There's no need to qualify bindings inside of a module with the module's name. Also `x == false` is typically written `!x`.

JeffBezanson:
Welcome, and thanks for working on this!

> In my opinion , there should be separate have_color variable for STDOUT and STDERR

That's a good point. For now, the easy thing would be to only enable color if stdout and stderr are both TTYs.

Looking at the code in more detail, we should really reduce use of the `have_color` global and instead keep a value per TTY object. Within julia the right way to check for color is `get(iostream, :color, false)`, but for TTYs that is implemented using the global variable. Instead, it should be a field of the object that is set automatically and overridden for stdout/stderr by the `--color` command line option. What do you think?

karan0299:
@JeffBezanson 
Yeah , This is quite a better approach . There is a mutable struct TTY in base/stream.jl . I think if we  add field color in that struct this initially set true and can be overridden by `--color ` as  you have already mentioned. This would solve our problem I think. Will try to get a PR on this.

vtjnash:
> Looking at the code in more detail, we should really reduce use of the have_color global and instead keep a value per TTY object

We have both because `stdout` could be assigned to any object. So there's a `have_color` global that reflects mostly just reflects the `--color` flag and applies to any IO object type. And then `TTY` also is supposed to have some auto-detection capabilities.

JeffBezanson:
It looks to me like `printstyled` uses `get(io, :color, false)`, and only TTYs implement `:color` (by returning the global `have_color`). I guess `--color` can continue to determine the default color setting for all TTYs, but ideally when `--color` is not specified each TTY would set its `:color` flag by autodetection (by running `tput` redirected to that object?)

karan0299:
@JeffBezanson For now i have implemented code to enable color when stdout and stderr are both TTY's . What should be the next step ?

karan0299:
@JeffBezanson Any updates on this one?

JeffBezanson:
This seems OK for now. @ararslan requested a NEWS entry, so just add a line to NEWS.md saying that color now defaults to on when stdout and stderr are TTYs.

karan0299:
I have made an entry in News.md with the purpose of this pr

NEWS.md
```diff
@@ -32,6 +32,8 @@ Language changes
   (in addition to the one that enters the help mode) to see the full
   docstring. ([#25930])
 
+* Color nows defaults to on when stdout and stderr are TTYs ([#34347])
```
JeffBezanson:
```suggestion
* Color now defaults to on when stdout and stderr are TTYs ([#34347])
```

JeffBezanson:
Thinking about it some more, this probably needs more discussion. In interactive mode, we try to check whether the terminal supports color. If checking is important, then it doesn't make much sense to skip the check in non-interactive mode. Before, we probably wanted to avoid the check to save startup time. I think there are 3 options:

1. Keep things as they are, default to color=no so we don't need any checks.
2. Add the `hascolor` check from stdlib/REPL/src/Terminals.jl to this PR; don't worry about the startup time cost.
3. Use this PR as-is, and remove the `hascolor` check from the REPL. Just assume color is available.

StefanKarpinski:
Idea from discussion on triage: delay checking of the output stream (by calling tput, etc.) until we actually do something that would be different based on if there's color or not. So, for example, `println("Hello, world")` doesn't care if there's color or not, whereas logging does care, so the color check would happen the first time a log message is printed and then get cached after that.

JeffBezanson:
Could be done this way:
1. Make the global `have_color` 3-state: true / false / nothing (nothing means not yet set)
2. The first time the TTY code queries the global, if it's `nothing` then call the detection function, which will have to be refactored a bit so you don't need a `TTYTerminal` object.

karan0299:
@JeffBezanson  I have understood your point but while implementing , I am not sure how I can do this without `TTYTerminal`  object.
Thanks.

tkf:
> 1. Make the global `have_color` 3-state

Do you mean for a "quick fix"?  It would be really nice if this is a per-instance property.  I have [multiple](https://github.com/tkf/TmuxDisplays.jl) [packages](https://github.com/tkf/JuliaCLI.jl) opening non-stdio TTYs and relying on a global state is [rather ugly](https://github.com/tkf/JuliaCLI.jl/blob/ef0870fe5aaebb3bc5c0107159d90686e37918da/src/evalutils.jl#L53-L62) for such usages.

karan0299:
I have the implementation to remove Dependency of `TTYTerminal` object from hascolor function
`function hascolor()`
 `       term_type = get(ENV, "TERM", @static Sys.iswindows() ? "" : "dumb")`
  `      startswith(term_type, "xterm") && return true`
  `      try `
   `        @static if Sys.KERNEL === :FreeBSD  `
 `               return success('tput AF 0')     `
   `         else`
   `             return success('tput setaf 0')  `
  `          end`
  `      catch`
    `        return false`
    `end
`end`

@JeffBezanson  please verify.

JeffBezanson:
Yes looks OK to me.

karan0299:
@JeffBezanson Please review this PR.

kmsquire:
Hi @karan0299, thanks for working on this PR. Most developers are pretty busy, and as contributors, we should try to be respectful of this fact in how we write messages. 

You probably didn’t realize it, but to me (at least), the message to Jeff comes across as a little demanding.

It would be a little better to write something like “When you get the chance, this PR is ready for re-review. Thanks!” or “I’ve made some small changes-can you please review it again?”  

Cheers!

karan0299:
Right, I am sorry but I didn't meant to demanding.
As you have pointed out, I felt the same way, again I am extremely sorry.
I will make sure this doesnt happen again.

karan0299:
@JeffBezanson any updates on this ?
Thanks!

karan0299:
@JeffBezanson any updates on this PR?
Thanks!

base/client.jl
```diff
@@ -211,6 +211,20 @@ end
 # call include() on a file, ignoring if not found
 include_ifexists(mod::Module, path::AbstractString) = isfile(path) && include(mod, path)
 
+function hascolorTTY()
+    term_type = get(ENV, "TERM", @static Sys.iswindows() ? "" : "dumb")
+    startswith(term_type, "xterm") && return true
+    try
```
ararslan: There's trailing whitespace on this and several following lines, which is causing the whitespace buildbot check to fail.

base/client.jl
```diff
@@ -211,6 +211,20 @@ end
 # call include() on a file, ignoring if not found
 include_ifexists(mod::Module, path::AbstractString) = isfile(path) && include(mod, path)
 
+function hascolorTTY()
+    term_type = get(ENV, "TERM", @static Sys.iswindows() ? "" : "dumb")
```
ararslan: Seems like the Windows check isn't doing anything meaningful here; perhaps this could just be
```suggestion
    term_type = get(ENV, "TERM", "")
```
?

base/client.jl
```diff
@@ -211,6 +211,20 @@ end
 # call include() on a file, ignoring if not found
 include_ifexists(mod::Module, path::AbstractString) = isfile(path) && include(mod, path)
 
+function hascolorTTY()
```
ararslan:
```suggestion
function ttyhascolor()
```
Minor nitpick, but this would be a more conventional name for this function.

base/client.jl
```diff
@@ -210,19 +210,36 @@ end
...
+end
 function exec_options(opts)
     if !isempty(ARGS)
         idxs = findall(x -> x == "--", ARGS)
         length(idxs) > 0 && deleteat!(ARGS, idxs[1])
     end
     quiet                 = (opts.quiet != 0)
     startup               = (opts.startupfile != 2)
-    history_file          = (opts.historyfile != 0)
+    history_file          = (opits.historyfile != 0)
```
ararslan:
```suggestion
    history_file          = (opts.historyfile != 0)
```
Seems to just have been a mistake. This is what's causing the current CI failures.

ararslan (approved):
LGTM if CI agrees.

StefanKarpinski:
Would be good to get this over the line at last. @JeffBezanson, perhaps you could take a last look through today and merge if it looks good to you?

base/client.jl
```diff
@@ -210,7 +210,19 @@ end
 
 # call include() on a file, ignoring if not found
 include_ifexists(mod::Module, path::AbstractString) = isfile(path) && include(mod, path)
-
+function ttyhascolor()
+    term_type = get(ENV, "TERM","")
+    startswith(term_type, "xterm") && return true
+    try
+        @static if Sys.KERNEL === :FreeBSD
+            return success("tput AF 0")
```
JeffBezanson:
```suggestion
            return success(`tput AF 0`)
```

base/client.jl
```diff
@@ -210,7 +210,19 @@ end
...
+        else
+            return success("tput setaf 0")
```
JeffBezanson:
```suggestion
            return success(`tput setaf 0`)
```

base/client.jl
ararslan: lol I didn't even notice this; the fact that it works at all is because this is inside of a `try` :smile:

base/client.jl
```diff
@@ -223,6 +235,11 @@ function exec_options(opts)
     global have_color     = (opts.color == 1) # --color=on
     global is_interactive = (opts.isinteractive != 0)
 
+    #check if stdout is TTY and color is not set manually
+    if !color_set && isa(stdout, TTY) && isa(stderr, TTY) && ttyhascolor()
```
JeffBezanson: This is not the design described above. It said to set this the first time the variable is used, which is in stream.jl.

karan0299:
I have changed the design as requested. I have a doubt: I have changed the 4 functions - `get`, `in`, `getindex` and `haskey` to non-inline functions. Does that cause some overhead? Are these functions right functions to change? There seems to be a type instabilty inside introduced because of this thing. Am I right about this?
I will squash the commits if this is right. Thanks!

StefanKarpinski:
If I'm reading this right now, this appears to add dict-like behavior to TTY objects, which seems rather fishy to me. Can you give a little more explanation of where this is intended to go?

base/client.jl
karan0299: This is causing checks(CI) to fail.

karan0299:
I am not sure where the  `global have_color` would be used first in stream.jl. So I added a check wherever color property of TTY is queried in each of the functions `in`,`haskey`,`getIndex` and `index`(https://github.com/JuliaLang/julia/blob/master/base/stream.jl#L488)
if the `have_color ` is not set , call is made to detection function  and change the value accordingly.
This is the design as Jeff suggested above https://github.com/JuliaLang/julia/pull/34347#issuecomment-580420151
So far dict is like behaviour is concerned , I think it is already in use currently for `:color` property
as in for `hasKey` and `getIndex` functions(https://github.com/JuliaLang/julia/blob/master/base/stream.jl#L490)

base/stream.jl
```diff
@@ -484,11 +484,39 @@ function displaysize(io::TTY)
...
+    end
+end
+function in(key_value::Pair{Symbol,Bool}, ::TTY)
+    global have_color
+    have_color == nothing && (have_color = ttyhascolor())
+    return key_value.first === :color && key_value.second === have_color
+end
+function haskey(::TTY, key::Symbol)
+    global have_color
+    have_color == nothing && (have_color = ttyhascolor())
```
JeffBezanson: Don't repeat this code. You can move it to its own function, then replace uses of `have_color` with `get_have_color()`:
```
function get_have_color()
    global have_color
    have_color === nothing && (have_color = ttyhascolor())
    return have_color::Bool
end
```

karan0299: Yes, this looks quite better , thanks!

karan0299:
When this PR is reviewed and looks fine to reviewer , I will rebase on master. Thanks!!

StefanKarpinski:
This version looks pretty good to me. The slew of CI failures is concerning though.

karan0299:
I try to figure out problem with CI. Thanks!!

karan0299:
@StefanKarpinski , I have checked thoroughly , I dont think the failure in test are due to changes in this PR.
If you could look into this , it would be really helpful. Thanks!!

StefanKarpinski:
It's pretty unlikely that all of the tests fail due to something unrelated to this PR. Have you run tests locally?

stdlib/REPL/src/REPL.jl
```diff
@@ -1168,17 +1168,17 @@ function run_frontend(repl::StreamREPL, backend::REPLBackendRef)
     dopushdisplay = !in(d,Base.Multimedia.displays)
     dopushdisplay && pushdisplay(d)
     while !eof(repl.stream)
-        if have_color
+        if have_color === true
```
JeffBezanson: Instead I would set `have_color = get(repl.stream, :color, false)` at the top of this function.

base/client.jl
```diff
@@ -111,7 +111,7 @@ function eval_user_input(errio, @nospecialize(ast), show_value::Bool)
     lasterr = nothing
     while true
         try
-            if have_color
+            if have_color === true
```
JeffBezanson: `get(stdout, :color, false)`

stdlib/REPL/src/REPL.jl
karan0299: Right

karan0299:
> It's pretty unlikely that all of the tests fail due to something unrelated to this PR. Have you run tests locally?

Yes , I tried running tests locally but same results.
**Also I ran tests locally on master branch and tests failed for master too**

mkitti (changes requested):
I add created a PR against your branch to resolve the outstanding issues.

base/stream.jl
```diff
@@ -485,10 +485,28 @@ function displaysize(io::TTY)
...
 
-in(key_value::Pair{Symbol,Bool}, ::TTY) = key_value.first === :color && key_value.second === have_color
+function ttyhascolor()
+    term_type = get(ENV, "TERM","")
+    startswith(term_type, "xterm") && return true
+    try
+        @static if Sys.KERNEL === :FreeBSD
+            return success("tput AF 0")
+        else
+            return success("tput setaf 0")
```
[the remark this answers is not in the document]
mkitti: See https://github.com/karan0299/julia/pull/1 OR https://github.com/karan0299/julia/compare/TTY-STDOUT-color...mkitti:TTY-STDOUT-color

base/ttyhascolor.jl
```diff
@@ -0,0 +1,24 @@
+function ttyhascolor()
+    term_type = get(ENV, "TERM","")
+    startswith(term_type, "xterm") && return true
+    try
+        @static if Sys.KERNEL === :FreeBSD
+            return success(`tput AF 0`)
+        else
+            return success(`tput setaf 0`)
```
mkitti: What happens here on Microsoft Windows?

mkitti:
Looks like all the CI tests are now passing.

karan0299:
Thank you @mkitti for the great help!!

base/ttyhascolor.jl
JeffBezanson: The original code in the repl just returns `true` on windows.

base/ttyhascolor.jl
```diff
@@ -0,0 +1,24 @@
...
+        end
+    catch e
+        return false
+    end
+end
+function get_have_color()
+    global have_color
+    have_color === nothing && (have_color = ttyhascolor())
+    return have_color
```
JeffBezanson:
```suggestion
    return have_color::Bool
```

stdlib/REPL/test/repl.jl
```diff
@@ -51,7 +51,10 @@ function fake_repl(@nospecialize(f); options::REPL.Options=REPL.Options(confirm_
     repl.options = options
 
     hard_kill = kill_timer(900) # Your debugging session starts now. You have 15 minutes. Go.
+    @eval Base old_have_color = have_color
+    @eval Base have_color = false
```
JeffBezanson: Is it possible to remove all use of `Base.have_color` in the REPL so that this is not necessary?

base/ttyhascolor.jl
JeffBezanson: It would also be much better not to have two copies of this code; with a few more change it should be possible to remove the one in Terminals.jl.

mkitti: In some circumstances, `tput` may exist on Windows but not all cases (Git Bash Shell, Cygwin, WSL[2]). I suggest that if we catch an exception from `success` meaning that `tput` does not exist then we return `true` if on Windows.

stdlib/REPL/test/repl.jl
```diff
@@ -51,7 +51,10 @@ function fake_repl(@nospecialize(f); options::REPL.Options=REPL.Options(confirm_
     repl.options = options
 
     hard_kill = kill_timer(900) # Your debugging session starts now. You have 15 minutes. Go.
+    # @eval Base old_have_color = have_color
+    # @eval Base have_color = false
```
mkitti: To pass this test we have to disable colors. If not this way, then we need another mechanism.

stdlib/REPL/test/repl.jl
mkitti: We need a mechanism to disable colors. Perhaps we need to be able to `set`

stdlib/REPL/test/repl.jl
mkitti: @karan0299 I think you will have to implement a `set(::TTY, key::Symbol, value::Bool)` then we can disable color for `fake_repl`

mkitti:
I think the commits 95c8ea3 59586de in #35240 may be useful to get this over the line.

base/ttyhascolor.jl
mkitti: I have a candidate to address the redundancy. See 95c8ea3 pending merge to this PR.

stdlib/REPL/test/repl.jl
mkitti: @karan0299 This was tricky and required adding `hascolor` to the concrete REPL types. That done all references to `Base.have_color` are removed from the REPL package: 59586de .

Feel free to merge from https://github.com/mkitti/julia/tree/TTY-STDOUT-color

karan0299:
@mkitti yeah I think this is a concrete approach to settle the problem

stdlib/REPL/test/repl.jl
karan0299: I’m so grateful for your help. Thank you very much.

base/ttyhascolor.jl
```diff
@@ -0,0 +1,30 @@
...
+end
+in(key_value::Pair{Symbol,Bool}, ::TTY) = key_value.first === :color && key_value.second === get_have_color()
+haskey(::TTY, key::Symbol) = key === :color
+getindex(::TTY, key::Symbol) = key === :color ? get_have_color() : throw(KeyError(key))
+get(::TTY, key::Symbol, default) = key === :color ? get_have_color() : default
+<<<<<<< HEAD
+
+
+=======
+>>>>>>> 59586ded0da55b383c01b5d6e3538cd9680d79a3
```
mkitti: Remove lines 26 - 30
```suggestion
<<<<<<< HEAD


=======
>>>>>>> 59586ded0da55b383c01b5d6e3538cd9680d79a3
```

mkitti:
@karan0299 You still have some merge conflict marks at the end of `base/ttyhascolor.jl`. Take those out.

base/Base.jl
```diff
@@ -241,6 +241,7 @@ include("filesystem.jl")
 using .Filesystem
 include("cmd.jl")
 include("process.jl")
+include("ttyhascolor.jl")
```
JeffBezanson: Does this need to be its own file? Otherwise it seems like it makes sense to keep it with the implementation of TTY.

stdlib/REPL/test/repl.jl
JeffBezanson: Starting to look good! So much nicer without the global. Can this now be done by passing `false` to the REPL constructors above?

base/Base.jl
karan0299: I think ,we can shift this to stdlib/REPL/src/Terminal.jl

mkitti: @JeffBezanson The code in ttyhascolor.jl needs to come after process.jl due to the use of `success`. stream.jl is included before process.jl, and we were getting compilation errors because of this.

JeffBezanson: Ok, I figured it was something like that.

stdlib/REPL/test/repl.jl
mkitti: @JeffBezanson 

For `LineEditREPL` you can now just pass in `hascolor` as I do for the test above. The constructor previously ignored it and always set it to `true`.

`BasicREPL` does not have a struct field to store `hascolor`. It purely depends on `hascolor(r.terminal)`.
Similarly, `StreamREPL` now purely depends on `hascolor(r.stream)`.

I did not implement `hascolor(r::AbstractREPL)`. Should we? It would probably be `hascolor(r::AbstractREPL) = false`

I think there is a PR for another day where one eliminates the redundancy of `hascolor` between `REPL.Options` and the `LineEditREPL`. We could also consider how `REPL.GlobalOptions` interacts with the `BasicREPL` and `StreamREPL`.

JeffBezanson: The tests are passing so it seems fine now. Can we remove the commented-out lines?

mkitti: @karan0299 Removing those comments is all you. I already removed them in my branch.

base/loading.jl
```diff
@@ -1169,10 +1169,11 @@ function create_expr_cache(input::String, output::String, concrete_deps::typeof(
             eval(Meta.parse(code))
         end
         """
+
     io = open(pipeline(`$(julia_cmd()) -O0
                        --output-ji $output --output-incremental=yes
                        --startup-file=no --history-file=no --warn-overwrite=yes
-                       --color=$(have_color ? "yes" : "no")
+                       --color=$(have_color === true ? "yes" : "no")
```
JeffBezanson:
```suggestion
                       --color=$(have_color === nothing ? "auto" : have_color ? "yes" : "no")
```

mkitti:
I'm not sure why the one test failed for building on Mac. Looks like some kind of stochastic error. Could someone restart that test to see?

karan0299:
All CI tests passed

JeffBezanson:
Thanks for your persistence on this one. Ended up being more complicated than I would have thought! Glad to have this finally merged :)

PR #34350, opened 2020-01-11
nonsquare (underdetermined) LQ solves

### LINKED ISSUE #689: lq(A) \ b unnecessarily requires square A
As described in the [LAPACK documentation](https://www.netlib.org/lapack/lug/node41.html), the LQ factorization can be used to compute a minimum-norm solution for (full-rank) underdetermined systems (more columns than rows).  However `lq(A) \ b` currently calls `checksquare`, and the `ldiv!` routine also doesn't handle the case where the output solution has more rows than the input right-hand-side ([discourse](https://discourse.julialang.org/t/solution-of-underdetermined-systems-using-lapack/33000)).

Should be straightforward to update the code to eliminate this restriction, since it doesn't require any new solver.

### LINKED ISSUE #34348: lq(A) \ b unnecessarily requires square A
As described in the [LAPACK documentation](https://www.netlib.org/lapack/lug/node41.html), the LQ factorization can be used to compute a minimum-norm solution for (full-rank) underdetermined systems (more columns than rows).  However `lq(A) \ b` currently calls `checksquare`, and the `ldiv!` routine also doesn't handle the case where the output solution has more rows than the input right-hand-side ([discourse](https://discourse.julialang.org/t/solution-of-underdetermined-systems-using-lapack/33000)).

Should be straightforward to update the code to eliminate this restriction, since it doesn't require any new solver.

stevengj:
Fixes JuliaLang/LinearAlgebra.jl#689.

Seems to be about 2x faster than pivoted QR for a large underdetermined solve, with similar accuracy (assuming full row rank):
```jl
julia> A = rand(1000,1500); b = rand(size(A,1));

julia> @btime lq($A) \ $b;
  101.955 ms (14 allocations: 19.37 MiB)

julia> @btime qr($A, Val{true}()) \ $b;
  250.584 ms (8046 allocations: 31.98 MiB)

julia> x = lq(A) \ b; norm(x - qr(A, Val{true}()) \ b) / norm(x)
3.5357944054772334e-15
```

### DIFF
--- a/stdlib/LinearAlgebra/src/lq.jl
+++ b/stdlib/LinearAlgebra/src/lq.jl
@@ -81,7 +81,7 @@ orthogonal/unitary component via `S.Q`, such that `A ≈ S.L*S.Q`.
 Iterating the decomposition produces the components `S.L` and `S.Q`.
 
 The LQ decomposition is the QR decomposition of `transpose(A)`, and it is useful
-in order to compute the minimum-norm solution `lq(A) \\ b` to an overdetermined
+in order to compute the minimum-norm solution `lq(A) \\ b` to an underdetermined
 system of equations (`A` has more columns than rows, but has full row rank).
 
 # Examples
@@ -310,7 +310,7 @@ _rightappdimmismatch(rowsorcols) =
 function (\)(A::LQ{TA},B::StridedVecOrMat{TB}) where {TA,TB}
     S = promote_type(TA,TB)
     m, n = size(A)
-    m ≤ n || throw(DimensionMismatch("LQ solver does not support underdetermined systems (more rows than columns)"))
+    m ≤ n || throw(DimensionMismatch("LQ solver does not support overdetermined systems (more rows than columns)"))
     m == size(B,1) || throw(DimensionMismatch("Both inputs should have the same number of rows"))
     AA = Factorization{S}(A)
     X = _zeros(S, B, n)
--- a/stdlib/LinearAlgebra/src/lq.jl
+++ b/stdlib/LinearAlgebra/src/lq.jl
@@ -80,7 +80,9 @@ orthogonal/unitary component via `S.Q`, such that `A ≈ S.L*S.Q`.
 
 Iterating the decomposition produces the components `S.L` and `S.Q`.
 
-The LQ decomposition is the QR decomposition of `transpose(A)`.
+The LQ decomposition is the QR decomposition of `transpose(A)`, and it is useful
+in order to compute the minimum-norm solution `lq(A) \\ b` to an overdetermined
+system of equations (`A` has more columns than rows, but has full row rank).
 
 # Examples
 ```jldoctest
@@ -174,11 +176,13 @@ end
 
 
 ## Multiplication by LQ
-lmul!(A::LQ, B::StridedVecOrMat) =
-    lmul!(LowerTriangular(A.L), lmul!(A.Q, B))
+function lmul!(A::LQ, B::StridedVecOrMat)
+    lmul!(LowerTriangular(A.L), view(lmul!(A.Q, B), 1:size(A,1), axes(B,2)))
+    return B
+end
 function *(A::LQ{TA}, B::StridedVecOrMat{TB}) where {TA,TB}
     TAB = promote_type(TA, TB)
-    lmul!(Factorization{TAB}(A), copy_oftype(B, TAB))
+    _cut_B(lmul!(Factorization{TAB}(A), copy_oftype(B, TAB)), 1:size(A,1))
 end
 
 ## Multiplication by Q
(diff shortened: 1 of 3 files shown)


stdlib/LinearAlgebra/src/lq.jl
```diff
@@ -80,7 +80,9 @@ orthogonal/unitary component via `S.Q`, such that `A ≈ S.L*S.Q`.
 
 Iterating the decomposition produces the components `S.L` and `S.Q`.
 
-The LQ decomposition is the QR decomposition of `transpose(A)`.
+The LQ decomposition is the QR decomposition of `transpose(A)`, and it is useful
+in order to compute the minimum-norm solution `lq(A) \\ b` to an overdetermined
```
dkarrasch: under-determined?

stdlib/LinearAlgebra/src/lq.jl
```diff
@@ -303,36 +307,33 @@ _rightappdimmismatch(rowsorcols) =
...
-    x = ldiv!(AA, copy_oftype(b, S))
-    return x
-end
-function (\)(A::LQ{TA},B::StridedMatrix{TB}) where {TA,TB}
+function (\)(A::LQ{TA},B::StridedVecOrMat{TB}) where {TA,TB}
     S = promote_type(TA,TB)
-    m = checksquare(A)
-    m == size(B,1) || throw(DimensionMismatch("left hand side has $m rows, but right hand side has $(size(B,1)) rows"))
+    m, n = size(A)
+    m ≤ n || throw(DimensionMismatch("LQ solver does not support underdetermined systems (more rows than columns)"))
```
dkarrasch: over-determined?

stdlib/LinearAlgebra/src/lq.jl
```diff
@@ -303,36 +307,33 @@ _rightappdimmismatch(rowsorcols) =
...
                            isa(B, AbstractVector) ? (size(F,2),) : (size(F,2), size(B,2)))
 end
 
 
 function ldiv!(A::LQ{T}, B::StridedVecOrMat{T}) where T
-    lmul!(adjoint(A.Q), ldiv!(LowerTriangular(A.L),B))
-    return B
+    require_one_based_indexing(B)
+    ldiv!(LowerTriangular(A.L), view(B, 1:size(A,1), axes(B,2)))
+    return lmul!(adjoint(A.Q), B)
```
dkarrasch: This seems to error when I checked manually. BTW, are we testing `ldiv!` for non-square LQ? I could only find `\` tests.

dkarrasch (commented):
One of us is getting over- and under-determined wrong, perhaps worth a second thought. The other issue I couldn't see resolved is `ldiv!`, which promises solve in-place of `B`, but must necessarily widen/stretch `B` to fit the longer `X` in `A*X = B`. It's not easy to track the changes in the test, but I couldn't find tests for `ldiv!`.

stdlib/LinearAlgebra/src/lq.jl
stevengj: `\` calls `ldiv!`

stevengj:
For non-square problems with more cols n than rows m, `ldiv!` only looks at the first `m` rows of its input.

stdlib/LinearAlgebra/src/lq.jl
dkarrasch: Yes, but `\` allocates a correctly sized vector/matrix `X` that has more rows than `B`. I guess my question is: can you do `ldiv!(lq(rand(2,5)), rand(2))`? It must return a vector of length 5, but I don't see that `ldiv!` allocates such a vector, or `append!`s some trailing zeros to the input `B`.

dkarrasch: I'm getting
```julia
julia> ldiv!(lq(rand(2,5)), rand(2))
ERROR: DimensionMismatch("for a left-sided multiplication, the first dimension of C, 2, must equal the second dimension of A, 5")
Stacktrace:
 [1] ormlq!(::Char, ::Char, ::Array{Float64,2}, ::Array{Float64,1}, ::Array{Float64,1}) at /Users/karrasch/Documents/julia/usr/share/julia/stdlib/v1.5/LinearAlgebra/src/lapack.jl:2715
 [2] lmul! at /Users/karrasch/Documents/julia/usr/share/julia/stdlib/v1.5/LinearAlgebra/src/lq.jl:198 [inlined]
 [3] ldiv!(::LQ{Float64,Array{Float64,2}}, ::Array{Float64,1}) at /Users/karrasch/Documents/julia/usr/share/julia/stdlib/v1.5/LinearAlgebra/src/lq.jl:336
 [4] top-level scope at REPL[2]:1

julia> lq(rand(2,5)) \ rand(2)
5-element Array{Float64,1}:
 -0.21105452790704704
  0.41169392192670684
 -0.4022006494025896
  0.44360732676056863
  0.2111926086486643
```
with this PR. So we may need to append zeros to the rhs within `ldiv!`. Which leaves open what to do in the matrix rhs case: have a 3-arg `ldiv!` and solve columnwise?

stevengj: > I guess my question is: can you do `ldiv!(lq(rand(2,5)), rand(2))`? 

No, *you* need to append 3 rows of padding to the right-hand side so that it holds space for the solution.  It won't allocate, because the whole point of `ldiv!` is to work in-place.

For example:
```
julia> A = rand(2,5);

julia> A \ [3,4]
5-element Array{Float64,1}:
  1.798302830371654
  0.1864736948656177
  2.353194756515655
  1.5853030268731967
 -0.6713831865960345

julia> lq(A) \ [3,4]
5-element Array{Float64,1}:
  1.7983028303716546
  0.18647369486561782
  2.353194756515656
  1.5853030268731967
 -0.6713831865960342

julia> ldiv!(lq(A), [3.0,4,0,0,0])
5-element Array{Float64,1}:
  1.7983028303716546
  0.18647369486561782
  2.353194756515656
  1.5853030268731967
 -0.6713831865960342

julia> ldiv!(qr(A, Val(true)), [3.0,4,0,0,0])
5-element Array{Float64,1}:
  1.798302830371654
  0.1864736948656177
  2.353194756515655
  1.5853030268731967
 -0.6713831865960345
```
Note the last line — this is **not a new behavior**.  `ldiv!` behaves the same way for pivoted QR when solving underdetermined systems.

dkarrasch: Thanks. I wasn't aware of that bevavior, I just saw that `\` does that for the caller.

dkarrasch (approved):

stevengj:
win32 CI failure seems to be an unrelated `Error in testset SharedArrays`

stdlib/LinearAlgebra/test/lq.jl
```diff
@@ -5,48 +5,42 @@ module TestLQ
...
 
-areal = randn(n,n)/2
-aimg  = randn(n,n)/2
-a2real = randn(n,n)/2
-a2img  = randn(n,n)/2
-breal = randn(n,2)/2
-bimg  = randn(n,2)/2
+asquare = randn(ComplexF64, m, m) / 2
+awide = randn(ComplexF64, m, m+3) / 2
+bcomplex = randn(ComplexF64, m, 2)/2
```
StefanKarpinski:
```suggestion
bcomplex = randn(ComplexF64, m, 2) / 2
```

stevengj:
win32 failure is now another unrelated glitch `cp: cannot stat 'dist-extras/7z.*': No such file or directory`.

Should be good to merge?

stevengj:
I wonder if we should change the default `A \ b` polyalgorithm to use LQ for undetermined problems ("wide" A)?  It still seems to be twice as fast on Julia 1.11.

dkarrasch:
Currently, we only distinguish square from non-square, right? Sounds like a good idea.

stevengj:
Currently, we don't have row-pivoted `lq`, so using `lq(A) \ b` as-is might be problematic for badly conditioned "wide" `A` (i.e. nearly dependent rows).

However, we could equivalently use pivoted QR on $A^T$ via `qr(A', ColumnNorm())' \ b`.  That seems to be the fastest option of all, actually:
```jl
julia> A = rand(100, 1000); b = rand(100);

julia> @btime $A \ $b; @btime qr($A, ColumnNorm()) \ $b; @btime lq($A) \ $b; @btime qr($A', ColumnNorm())' \ $b;
  6.802 ms (42 allocations: 1.89 MiB)
  6.851 ms (42 allocations: 1.89 MiB)
  3.847 ms (21 allocations: 926.41 KiB)
  2.454 ms (29 allocations: 1008.34 KiB)
```

PR #34033, opened 2019-12-06
add `Iterators.accumulate`

JeffBezanson:
I was surprised we didn't have this. Even IterTools.jl doesn't seem to have it. Nice to complete the set  with map and filter.

PR #34352, opened 2020-01-12
Add Iterators.map

tkf:
As @JeffBezanson mentioned in https://github.com/JuliaLang/julia/pull/34033#issuecomment-562671137, it makes sense to have `Iterators.map`.  This PR adds it (with some tests).


NEWS.md
```diff
@@ -19,7 +19,7 @@ Build system changes
 
 New library functions
 ---------------------
-
+* `Iterators.map` is added ([#34352]).
```
StefanKarpinski: Might want to indicate what it does.

base/iterators.jl
```diff
@@ -22,7 +22,27 @@ import .Base:
     getindex, setindex!, get, iterate,
     popfirst!, isdone, peek
 
-export enumerate, zip, rest, countfrom, take, drop, takewhile, dropwhile, cycle, repeated, product, flatten, partition
+export enumerate, zip, rest, countfrom, take, drop, takewhile, dropwhile, cycle, repeated, product, flatten, partition, map
```
JeffBezanson: It should not be exported, since it will conflict with Base.map.

StefanKarpinski: I'm a little sad that we didn't stick with the convention that the capitalized type name is the lazy construct while the lowercase function name is the eager construct. IIRC, we had agreed on that around 0.2 era but then didn't end up sticking to it.

tkf:
I addressed the points in the review.  Now `Iterators.map` is not exported and NEWS.md has a bit more description.

bramtayl:
Can we have generators lower to `Iterators.map` instead?

tkf:
That sounds like a great idea. Same holds for `Iterators.filter`, `Iterators.flatten`, and `Iterators.product`. Can you open a new issue to track the idea?

bramtayl:
Sure #34368

NEWS.md
```diff
@@ -19,7 +19,8 @@ Build system changes
 
 New library functions
 ---------------------
-
+* `Iterators.map` is added. It provides another syntax `Iterators.map(f, iterators...)`
+  for wiring `(f(args...) for args in zip(iterators...))` ([#34352]).
```
StefanKarpinski:
```suggestion
  for writing `(f(args...) for args in zip(iterators...))`, i.e. a lazy `map` ([#34352]).
```

StefanKarpinski:
Seems to have trailing whitespace.

tkf:
buildbot/whitespace_linux32 (https://build.julialang.org/#/builders/150/builds/4226) seems to fail due to network failure (?):

```
fatal: git fetch-pack: expected ACK/NAK, got 'ERR upload-pack: not our ref 7ad6d9c178229d71962daa2913aee8b21865c175'
fatal: The remote end hung up unexpectedly
```

oxinabox:
This has been in IterTools.jl forever as `imap`
I feel like adding every function from IterTools.jl into `Base.Iterators` is a kind of bad path.

Julia since the 0.7 release (and even before) has pushed towards functionality should live in packages, not that standard library.

In particular, `Base.Iterators`  provides the minimal set of iterators to implement the rest of the standard libraries.

tkf:
> In particular, `Base.Iterators` provides the minimal set of iterators to implement the rest of the standard libraries.

#34033 doesn't seem to imply this.

oxinabox:
As I said, I would have called #34033 out if I had have seen it before it was merged.

bramtayl:
I support not adding unnecessary stuff to Base.

This issue isn't about adding a new iterator, just an interface to one that already exists (namely, Base.Generator).

tkf:
Yes, that's a good point. It'd be nice to have a public and functional interface for constructing `Base.Generator`.

test/iterators.jl
```diff
@@ -67,6 +67,13 @@ end
 @test length(zip(1:3,product(1:7,cycle(1:3)))) == 3
 @test length(zip(1:3,product(1:7,cycle(1:3)),8)) == 1
 
+# map
+# ----
+@testset "Iterators.map" begin
+    @test collect(Iterators.map(string, 1:3)::Base.Generator) == map(string, 1:3)
+    @test collect(Iterators.map(tuple, 1:3, 4:6)::Base.Generator) == map(tuple, 1:3, 4:6)
+end
```
stev47, line 75: it might be good to test that `Iterators.map` can be applied to infinite iterables such as `Iterators.cycle(1:3)` and has the expected behaviour for the first n iterates.

base/iterators.jl
```diff
@@ -24,6 +24,26 @@ import .Base:
...
 
+"""
+    Iterators.map(f, iterators...)
+
+Create a lazy mapping.  This is another syntax for writing
+`(f(args...) for args in zip(iterators...))`.
+
+!!! compat "Julia 1.5"
+    This function requires at least Julia 1.5.
+
```
stev47, line 35: Maybe cross-reference with the docstring of `Base.Generator`

stev47 (commented):
I support this change, since it essentially just adds an alias that is very intuitive.

base/iterators.jl
tkf: I don't think `Base.Generator` is included in the documentation.  This is partially why I want to add `Iterators.map`.

test/iterators.jl
tkf: What is the reason why testing infinite iterators is interesting here?  I'm testing these two invocations since `Base.Generator` has different methods for them. So, I think it's somewhat meaningful to make sure we can hit these methods.  Other than that, I don't think we need to add tests specially for `Iterators.map` and they can be done in `Base.Generator`.  I think it makes sense to rename direct uses of `Base.Generator` in Base/stdlib.  But I think it'd be better to do this in a separate PR to make this PR more "atomic".

stev47:
> What is the reason why testing infinite iterators is interesting here?

It tests the laziness of `Iterators.map` since non-lazy `map` can't handle
infinite iterators.

> I don't think we need to add tests specially for `Iterators.map` and they can
> be done in `Base.Generator`. I think it makes sense to rename direct uses of
> `Base.Generator` in Base/stdlib. But I think it'd be better to do this
> in a separate PR to make this PR more "atomic".

I agree.

stev47:
> I don't think `Base.Generator` is included in the documentation. This is
> partially why I want to add `Iterators.map`.

It has a docstring on Julia 1.4.1 at least.

tkf:
> It tests the laziness of `Iterators.map` since non-lazy `map` can't handle infinite iterators.

Existing tests already check that `Iterators.map` returns `Base.Generated`.

> It has a docstring on Julia 1.4.1 at least.

Having a docstring does not mean it's a public API. See #35715

stev47:
> Having a docstring does not mean it's a public API. See #35715

Thank you for clarifying, but even internal documentation can benefit from
cross-referencing. Mentioning the alias `Iterators.map` in the docstring of
`Base.Generator` does not seem like any harm to me.

tkf:
There is a discussion on the extension of `Iterators.map` #34368. Documenting what it returns limits such extension.  In general, I think it's better to document how a certain object behaves rather than what it actually is.

tkf:
I resolved the merge conflict and CI is all green now. Can someone merge this?

(ping @StefanKarpinski @JeffBezanson)

tkf:
(ping @StefanKarpinski @JeffBezanson)

andyferris:
This is exciting!

I'd note that the NEWS entry *seems* to imply that `(f(x) for x in iter)` would lower to `Iterators.map(f, iter)`, but that doesn't appear to be implemented here. (I realize there is the discusison at #34368).

andyferris:
Hmm... or maybe I'm just reading that sentence wrong?

tkf:
My intention was that both `(f(x) for x in iter)` and `Iterators.map(f, iter)` _evaluate_ to the same thing.  But yeah, I think lowering the comprehensions to `Iterators.map` etc. would be great.

PR #34354, opened 2020-01-12
fix missing type prefix for nested containers

### LINKED ISSUE #34343: missing type prefix in display of some nested containers
Some types (e.g. Vector and Dict) are able to rely on their element values to imply the element types in printing. When there are no elements, we can handle it for the outer container but not nested ones. For example Vector:
```
julia> show([[1],Int[]])
[[1], []]
```
or Dict:
```
julia> show([Dict(1=>1),Dict{Int,Int}()])
[Dict(1 => 1), Dict()]
```

karan0299:
Fixes #34343 . Removed the check if context already knows about the type and element type, since it was preventing from printing the type of nested container , for e.g.
`julia> show([[1],Int[]])`
`[[1], []]`
Since the context is already known for Int[] , type is not shown.

### DIFF
--- a/base/arrayshow.jl
+++ b/base/arrayshow.jl
@@ -491,19 +491,17 @@ function typeinfo_prefix(io::IO, X)
         typeinfo = Any
     end
 
-    # what the context already knows about the eltype of X:
-    eltype_ctx = typeinfo_eltype(typeinfo)
     eltype_X = eltype(X)
 
     if X isa AbstractDict
-        if eltype_X == eltype_ctx || (!isempty(X) && typeinfo_implicit(keytype(X)) && typeinfo_implicit(valtype(X)))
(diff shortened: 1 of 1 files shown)


base/arrayshow.jl
```diff
@@ -427,6 +427,7 @@ end
 
 ### 0-dimensional arrays (#31481)
 show_zero_dim(io::IO, X::BitArray{0}) = print(io, "BitArray(", Int(X[]), ")")
+print("hi")
```
KristofferC: :)

karan0299: I am sorry, I was checking the flow , correcting it right now

JeffBezanson:
It looks like the fix will need to be a bit more subtle. Any ideas @rfourquet? Maybe if a container decides not to print a type prefix, it should then not set `:typeinfo`?

rfourquet:
Sorry @JeffBezanson , I missed your ping. I have an alternate fix at ##34660.

Issue #2626, opened 2013-03-20
a, b... = [1,2,3]


StefanKarpinski:
This would be a very nice syntax for taking head and rest. Likewise `a..., b = [1,2,3]` might be good for slurping the initial elements into `a` and the tail element into `b`.

StefanKarpinski:
I found myself wanting this syntax yet again today. I think we should consider this.

kmsquire:
(Match.jl has this)

StefanKarpinski:
There are some design issues here, primarily: what should the type of `b` (or `a`) be – array, tuple or iterator? Since we haven't addressed this, it seems best to bump this to 1.0.

HarrisonGrodin:
Here's an interesting example that begins to address the issue.

```julia
julia> a, b, c = countfrom()
Base.Count{Int64}(1,1)

julia> a, b, c
(1,2,3)
```

This works as expected, but what if "splatting" was used on one of the variables?

In the case that `c` was "splatted", it would make sense to return an iterator with `start` returning the "current" iterator value and `next`/`done` matching that of the existing iterator.
However, what if `b` was "splatted"? Would the code run indefinitely? An iterator could be useful if the last variable is "splatted", but it could become more confusing otherwise.

In the case that an iterator is not the right choice, though, the question of mutability (tuple vs. array) definitely seems to be worth debating.

StefanKarpinski:
Yes, it does seem that returning a `Rest` iterator might be necessary in the general case.

eveydee:
This is a feature I have missed a few times. FWIW, Python seems to do the equivalent of `collect` on the splatted unpacking. I don't like this.

Could the return type be delegated to the iterator? I. e., translate
```julia
a, b, c..., d, e = f
->
a = first(f); b = second(f);
c = slurp(f, 3, end-2);
d = secondlast(f); e = last(f)
```

StefanKarpinski:
I think the issue is for iterable things that are not infinite but have indefinite length and/or can only be iterated in order – you can't get the last item until you've collected the ones before it.

iamed2:
Is this not a 1.x possibility? This is currently invalid syntax.

mlhetland:
[...]

PR #34355, opened 2020-01-12
Make RegexMatch iterable

oxinabox:
Look at the tests.
Isn't this a nice feature?

In general its weird to define `getindex` without matching `iterate`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -47,6 +47,7 @@ Standard library changes
   `keep` that are to be kept as they are. ([#38597]).
 * `getindex` can now be used on `NamedTuple`s with multiple values ([#38878])
 * `keys(::RegexMatch)` is now defined to return the capture's keys, by name if named, or by index if not ([#37299]).
+* `RegexMatch` now iterate to give their captures. ([#34355]).
 
 #### Package Manager
 
--- a/base/regex.jl
+++ b/base/regex.jl
@@ -166,7 +166,7 @@ function show(io::IO, m::RegexMatch)
         for (i, capture_name) in enumerate(capture_keys)
             print(io, capture_name, "=")
             show(io, m.captures[i])
-            if i < length(m.captures)
+            if i < length(m)
                 print(io, ", ")
             end
         end
@@ -189,6 +189,10 @@ function haskey(m::RegexMatch, name::Symbol)
     return idx > 0
 end
 haskey(m::RegexMatch, name::AbstractString) = haskey(m, Symbol(name))
+
+iterate(m::RegexMatch, args...) = iterate(m.captures, args...)
+length(m::RegexMatch) = length(m.captures)
+eltype(m::RegexMatch) = eltype(m.captures)
 
 function occursin(r::Regex, s::AbstractString; offset::Integer=0)
     compile(r)
--- a/test/regex.jl
+++ b/test/regex.jl
@@ -167,6 +167,24 @@
         @test r"this|that"^2 == r"(?:this|that){2}"
     end
 
+    @testset "iterate" begin
+        m = match(r"(.) test (.+)", "a test 123")
+        @test first(m) == "a"
+        @test collect(m) == ["a", "123"]
+        for (i, capture) in enumerate(m)
+            i == 1 && @test capture == "a"
+            i == 2 && @test capture == "123"
+        end
+    end
+
+    @testset "Destructuring dispatch" begin
+        handle(::Nothing) = "not found"
+        handle((capture,)::RegexMatch) = "found $capture"
+
+        @test handle(match(r"a (\d)", "xyz")) == "not found"
+        @test handle(match(r"a (\d)", "a 1")) == "found 1"
+    end
+
     # Test that PCRE throws the correct kind of error
     # TODO: Uncomment this once the corresponding change has propagated to CI
     #@test_throws ErrorException Base.PCRE.info(C_NULL, Base.PCRE.INFO_NAMECOUNT, UInt32)


oxinabox:
Needs News.md.

May warrent mentioning in Docs?

DilumAluthge:
I think you'll also need to define:
```julia
length(m::RegexMatch) = length(m.captures)
```

tkf:
Also `eltype(::Type{RegexMatch})`?

Reading the test code, it looks like the plan is to treat `RegexMatch` like a `Dict`. If so, maybe defining `keys`, `values`, `keytype`, `valtype` and `pairs` makes sense?

DilumAluthge:
I thought the idea was to treat `RegexMatch` like an `AbstractVector`. After all, we are basically just forwarding the relevant methods (`iterate`, `length`, etc.) onto the `captures` field of `RegexMatch`, and `captures` is a `Vector`.

tkf:
I was just looking at this new test

https://github.com/JuliaLang/julia/blob/e6794e11473513c6e32183cc02a5217b2dcd271e/test/regex.jl#L139-L147

I think `Dict` like interface makes sense since we already have `getindex(::RegexMatch, ::Symbol)` for named capture.

oxinabox:
This PR makes it iterable.
Making it Dict-like feels like a possible follow up PR.
Seems more contentious potentially.
And getting a smaller PR reviewed and in is much easier.
Otherwise get bogged down in arguing about one feature.

The bit this adds is much less debatable.

base/regex.jl
```diff
@@ -168,6 +168,8 @@ function getindex(m::RegexMatch, name::Symbol)
 end
 getindex(m::RegexMatch, name::AbstractString) = m[Symbol(name)]
 
+iterate(m::RegexMatch) = iterate(m.captures)
```
oxinabox:
```suggestion
iterate(m::RegexMatch, args...) = iterate(m.captures, args...)
```

StefanKarpinski:
This makes sense to be although I can't say I've encountered the motivation to iterate over the captures of a single regex match. Can you give a sense of what kind of code needs this? Allowing indexing by index also seems like a fine addition to me. Why not, after all?

oxinabox:
> I can't say I've encountered the motivation to iterate over the captures of a single regex match. Can you give a sense of what kind of code needs this?

I often want to write code that looks like this test:
https://github.com/JuliaLang/julia/pull/34355/files#diff-f172897babd48052d6f286028f613cb2R149-R155
where I dispatch to `f(::Nothing)` or `f((capture1, capture2)::RegexMatch)`
It basically just means I can get rid of the `RegexMatch` object and get down to the capture using destructuring.

> Allowing indexing by index also seems like a fine addition to me. 

Indexing by index? We've always allowed that.

test/regex.jl
```diff
@@ -136,6 +136,24 @@
         @test r"this|that"^2 == r"(?:this|that){2}"
     end
 
+    @testset "iterate" begin
+        m = match(r"(.) test (.+)", "a test 123")
+        @test first(m) == "a"
+        @test collect(m) == ["a", "123"]
+        @test for (i, capture) in m
```
StefanKarpinski: I'm confused here: does `m` iterate just the capture values or does it iterate `(index, capture)` pairs?

StefanKarpinski:
> Indexing by index? We've always allowed that.

Then what is the proposed indexing by?

test/regex.jl
oxinabox:
```suggestion
        @test for (i, capture) in enumerate(m)
```

oxinabox:
>> Indexing by index? We've always allowed that.
> Then what is the proposed indexing by?

We are not adding indexing at all?
Only `iterate`

test/regex.jl
oxinabox: Iterates capture values, I just forgot the `enumerate`.
This is what i get for not running tests locally 😂

StefanKarpinski:
> I think `Dict` like interface makes sense

I'm asking about that.

StefanKarpinski:
There's a bit of thinking to be done here: if a RegexMatch object is map from indices or symbols to captures, then shouldn't it iterate like that dictionary? If you want a vector of captures, you can already get that by accessing `m.captures` which is just that.

oxinabox:
>> I think Dict like interface makes sense
>  I'm asking about that.

Ok, so you are talking to @tkf  fair enough.

From my point of view:
I don't think iterating as `(capturename, capturevalue)` makes sense.
It makes my motivating example of destructuring in a function call ugly.

I would have do to
`f((_, capture1), (_, capture2))::RegexMatch)`
Rather than  `f((capture1, capture2)::RegexMatch)`

While I very often want the capture values to be passed to a function, I have never wanted the capture names. I always know those since i named it.

StefanKarpinski:
Why not just pass the `.captures` field to the function?

oxinabox:
>   Why not just pass the .captures field to the function?

Issue with that is can't access `(nothing).captures` on the nonmatch.

Dispatching into `f(::Nothing)`, and `f((capture1, capture2)::RegexMatch)` gives an elegant way to solve that, while also giving them into variables.

I have a bunch of code that does;
```
function f(m::RegexMatch)
    capture1, capture2 = m.captures
    ...
end
f(::Nothing)=...
```

and I think the destructuring is much cleaner

StefanKarpinski:
> Issue with that is can't access `(nothing).captures` on the nonmatch.

I see. That makes sense. I guess I always just write out my `m !== nothing` check explicitly. I think we need to figure out what the data model of a RegexMatch is, lest we end up in the situation that DataFrames of yore did, where it was fundamentally confused about whether a DataFrame was: (a) a two-dimensional collection, (b) an integer-indexed collection of rows, or (c) a name-indexed collection of columns. So what is RegexMatch? (a) a struct that contains a collection of captures, (b) a collection of captures itself, or (b) a map from names to captures?

tkf:
I think a challenge for the Dict-like route is that it's not super clear what keys part should be for mixed named and unnamed captures like `m = match(r"(?P<first>.) (.) (.) (?P<last>.)", "1 2 3 4")`.  Should `keys(m)` be `[:first, 2, 3, :last]`? `[1, 2, 3, 4]`? `[1, 2, 3, 4, :first, :last]`? `[1, 2, 3, 4, :first, :last, "first", "last"]`?  There is also type-stability problem for output like `[:first, 2, 3, :last]`.  So maybe Array-like route (a collection of captures itself) is the way to go?

oxinabox:
I think it is a collection of captures that is indexed by position (int) and optionally by name.
It's a bit like NamedTuple in that way.
NamedTuple is more Tuple-like than Dict-like.

StefanKarpinski:
Ok, that makes sense. I just wanted to clarify the data model. Does this complete the implementation of that data model?

oxinabox:
> Does this complete the implementation of that data model?

`Tuple`, and `NamedTuple`, and `Vector` all overload `keys`.
But `keys` is contentious.

oxinabox:
To return to this PR.
I think we should do this.
I am going to make a follow up PR to add `keys` because that is a different question.

NEWS.md
```diff
@@ -84,6 +84,7 @@ Standard library changes
 * `push!(c::Channel, v)` now returns channel `c`. Previously, it returned the pushed value `v` ([#34202]).
 * `RegexMatch` objects can now be probed for whether a named capture group exists within it through `haskey()` ([#36717]).
 * A new standard library `TOML` has been added for parsing and printing [TOML files](https://toml.io) ([#37034]).
+* `RegexMatch` now iterate to give their captures. This means you can write `foo(::Noting)` and `foo((cap1, cap2)::RegexMatch)` to dispatch on if `match` found anything or not ([#34355]).
```
omus, line 87:
```suggestion
* `RegexMatch` now iterate to give their captures. This means you can write `foo(::Nothing)` and `foo((cap1, cap2)::RegexMatch)` to dispatch on if `match` found anything or not ([#34355]).
```

omus: This brings up the thought of having `foo((cap1, cap2)::typeof(regex))` to allow for dispatch on multiple regular expressions (this would require parameterizing `RegexMatch`)

omus:
Something that isn't great about
`foo((cap1, cap2)::RegexMatch)` is that it's quite fragile. You need to only call `foo` with the results of a specific regular expression match or encounter an exception when the number of captures differs. 
For example having conditional captures will break this pattern.

oxinabox:
I don't see that this needs to solve every problem to be a good idea.
I think 90% or regexes i have ever written have exactly 1 captures and probably another 5% exactly 2.
I don't know that i have ever used conditional captures.
I think this pattern captures (pun not intended) a pretty sunbstantial number of use-cases.
And its not like we are talking away the current way of doing things.

Further for conditional capture won't you want to write multiple different methods:
```julia
foo((cap1,)::RegexMatch) =...
foo((cap1, cap2)::RegexMatch) = ...
foo((cap1, cap2, cap3)::RegexMatch) = ...
```

This will also play nice with:
https://github.com/JuliaLang/julia/issues/2626#issuecomment-586740630
then we will be able to have:
`foo((cap1, caps...)::RegexMatch))`

NEWS.md
oxinabox: You mean like if we have `RegexMatch{r"This is (\d) regex"i}`?
That would be super cute, i can see that that would be handy for things like parsing a string with a list of regexs that might match.
After this is merged can open an issue about it.

KristofferC: Why did you mark this as resolved without fixing it?

KristofferC:
I must say the use-case shown in the NEWS looks extremely forced / unnatural to me. The code in https://github.com/JuliaLang/julia/pull/34355#issuecomment-573816926 looks perfectly fine and is easy to understand while the NEWS one is not.

NEWS.md
oxinabox: > Why did you mark this as resolved without fixing it?

Not indentional.
GitHub automatically marks it as resolved when you click apply suggestion.
Top comment has a suggestion to fix a typo.

omus:
> For example having conditional captures will break this pattern.

I'll need to make a correction on my part. I thought that alternative capture groups `(?|...)` (also known as [branch reset groups](https://www.regular-expressions.info/branchreset.html)) could result in a variable amount of capture groups being returned. This is not the case and I have not found an example where a single regular expression can return a variable amount of captured groups.

> I think it (a `RegexMatch`) is a collection of captures that is indexed by position (int) and optionally by name.

This statement does seem to be correct to me. When mixing named and positional capture groups the groups are always in the order in which they were defined:
```julia
julia> match(r"((?<b>b)|(?<a>a)(\n)?)", "a")
RegexMatch("a", 1="a", b=nothing, a="a", 4=nothing)
```

> foo((cap1,)::RegexMatch) =...
> foo((cap1, cap2)::RegexMatch) = ...
> foo((cap1, cap2, cap3)::RegexMatch) = ...

These all currently define the same method so `foo` is currently only compatible with a specific regular expression. That said, if this is a pattern we want to embrace we could parameterize `RegexMatch` by the capture groups (e.g. `RegexMatch{(1,:b,:a,4)}`) so that you could write separate methods to handle results from several regular expressions.

oxinabox:
> These all currently define the same method so foo is currently only compatible with a specific regular expression

Oh, good point.

tkf:
> As a bit of an aside: AFAIK `RegexMatch` is the only object in Base that as its API recommends accessing a field of the result

FYI I think another example is

> `Process`: Wait for a process or process chain to exit. **The `exitcode` field of a process can be used to determine success or failure.**
>
> --- https://docs.julialang.org/en/v1.6-dev/base/parallel/#Base.wait

I don't think we need to "fix" these cases, though.

oxinabox:
I continue to think we should do this.
What is blocking making a decision on this?
Reiterating the three reasons we should do this:

1. Matching other things that have `getindex` defined, this should have `iterate` defined.
2. Access to the collection as an iterable without needing to access a field (which we rarely do)
3. Destructuing dispatch on if matched or not.

---

@StefanKarpinski  asked:
>  Does this complete the implementation of that data model?

Yes, combined with  #37300 (merged) this implements the complete data model to act like a (partially) named-tuple.

@KristofferC  said:
> I must say the use-case shown in the NEWS looks pretty forced / unnatural to me.

Should I remove that example from the NEWS.md, and just simply say that regex matches now iterate to give the captures?
It is the thing I perstonally would do and what initially motivated this, but it is just one of 3 motivating reasons for this change.

oxinabox:
>  Should I remove that example from the NEWS.md, and just simply say that regex matches now iterate to give the captures?

I have now done this

test/regex.jl
```diff
@@ -166,6 +166,24 @@
         @test r"this|that"^2 == r"(?:this|that){2}"
     end
 
+    @testset "iterate" begin
+        m = match(r"(.) test (.+)", "a test 123")
+        @test first(m) == "a"
+        @test collect(m) == ["a", "123"]
+        @test for (i, capture) in enumerate(m)
```
oxinabox, line 173:
```suggestion
        for (i, capture) in enumerate(m)
```

oxinabox:
bump

JeffBezanson (approved):

PR #34367, opened 2020-01-13
Fixed keyword calls to code_macros with keyword arguments

### LINKED ISSUE #33163: Keyword args to code_ macros don't work with keyword calls
```jl
julia> foo(x; y) = x + y

julia> @code_typed foo(3);

julia> @code_typed foo(3, y=2);

julia> @code_typed optimize=true foo(3);

julia> @code_typed optimize=true foo(3; y=2);
ERROR: syntax: invalid syntax optimize = true
Stacktrace:
```

I wouldn't expect the last one to error.

Liozou:
Fixes #33163.

I added a small test just in case.

### DIFF
--- a/stdlib/InteractiveUtils/src/macros.jl
+++ b/stdlib/InteractiveUtils/src/macros.jl
@@ -6,18 +6,20 @@ import Base: typesof, insert!
 
 separate_kwargs(args...; kwargs...) = (args, kwargs.data)
 
-function gen_call_with_extracted_types(__module__, fcn, ex0)
+function gen_call_with_extracted_types(__module__, fcn, ex0, kws=Expr[])
     if isa(ex0, Expr)
         if any(a->(Meta.isexpr(a, :kw) || Meta.isexpr(a, :parameters)), ex0.args)
             return quote
                 local arg1 = $(esc(ex0.args[1]))
                 local args, kwargs = $separate_kwargs($(map(esc, ex0.args[2:end])...))
                 $(fcn)(Core.kwfunc(arg1),
-                       Tuple{typeof(kwargs), Core.Typeof(arg1), map(Core.Typeof, args)...})
+                       Tuple{typeof(kwargs), Core.Typeof(arg1), map(Core.Typeof, args)...};
+                       $(kws...))
             end
         elseif ex0.head === :call
             return Expr(:call, fcn, esc(ex0.args[1]),
-                        Expr(:call, typesof, map(esc, ex0.args[2:end])...))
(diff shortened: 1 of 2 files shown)


stdlib/InteractiveUtils/test/runtests.jl
```diff
@@ -272,6 +272,11 @@ let _true = Ref(true), f, g, h
     @test_throws ErrorException h()
 end
 
+# Issue #33163
+A33163(x; y) = x + y
+@test try @code_typed A33163(1, y=2); true; catch; false; end
```
Keno: The try/catch shouldn't be necessary. The test will fail if an error is thrown. Generally the pattern is just to check the return type or something in the test expr

stdlib/InteractiveUtils/src/macros.jl
```diff
@@ -106,12 +106,27 @@ function gen_call_with_extracted_types_and_kwargs(__module__, fcn, ex0)
         end
     end
     thecall = gen_call_with_extracted_types(__module__, fcn, arg)
+    if (thecall.head === :call && thecall.args[1] === fcn)
```
Keno: This seems a bit awkward and reliant on the exact order of things. Can we invert this by passing in the list of kwargs into the previous call or at least returning the position pointer therefrom?

stdlib/InteractiveUtils/src/macros.jl
```diff
@@ -106,12 +106,27 @@ function gen_call_with_extracted_types_and_kwargs(__module__, fcn, ex0)
...
+    elseif (thecall.head === :block && length(thecall.args) >= 6
+                                    && thecall.args[6].head === :call
+                                    && thecall.args[6].args[1] === fcn)
+        # This case corresponds to the presence of a keyword argument within
+        # the arg, which results in thecall having a different structure.
+        kwarg_position = thecall.args[6].args
+    elseif thecall.head === :call && thecall.args[1] === :error
+        # Any error from gen_call_with_extracted_types is propagated
+        return thecall
+    else
```
Keno: When does this branch happen? Add a test that exercises it?

Liozou:
Thanks for the review @Keno!
I updated the PR to make `gen_call_with_extracted_types` responsible for passing on the keyword arguments. The tests now check that the keyword arguments are taken into account (and will incidentally fail if the macro call errors as you pointed out).

Keno:
> The tests now check that the keyword arguments are taken into account (and will incidentally fail if the macro call errors as you pointed out).

Perfect, thanks.

PR #34373, opened 2020-01-14
Emphasise that closing is not about deallocation

cmcaine:
Maybe this is just something that I felt I needed to check and it shouldn't be in the docs? I thought I'd pitch the idea anyway.

### DIFF
--- a/base/channels.jl
+++ b/base/channels.jl
@@ -175,6 +175,8 @@ Close a channel. An exception (optionally given by `excp`), is thrown by:
 
 * [`put!`](@ref) on a closed channel.
 * [`take!`](@ref) and [`fetch`](@ref) on an empty, closed channel.
+
+Note: it is not necessary to close a channel for it to be deallocated.
 """
 function close(c::Channel, excp::Exception=closed_exception())
     lock(c)


KristofferC:
Is the concept of when things are "deallocated" mentioned anywhere else (except in e.g. C interop)? Since you don't have to deallocate anything else manually, is there really a need to special case the docs for this specific type?  Personally, I feel that it perhaps raises more questions than it answers.

jonas-schulze:
I stumbled upon this issue a while ago when building a pipeline using `RemoteChannels`. One part in the [Parallel Computing](https://docs.julialang.org/en/v1/manual/parallel-computing/#Channels-and-RemoteChannels-1) (search for "proper") docs made me wonder, whether I had to close a (remote) channel for it to be `gc`ed properly:

> `kill(manager, pid, config)` is called to remove a worker from the cluster. On the master process, the corresponding `IO` objects must be closed by the implementation to ensure proper cleanup. [...]

Since I am using a `RemoteChannel` to send information to a worker process, I thought of it as an `IO`-kind-of object. Just to be on the safe side, I closed all of them manually at the end of my routines. I still don't know whether that was a necessary thing to do.

But I don't know what would be the right place to clarify/document this potential issue.

vtjnash:
I don't think this is quite right or neededd

PR #34375, opened 2020-01-14
Added Docs For filter and getindex in string

Akshat-mehrotra:


Akshat-mehrotra:
What is going wrong? I've just forked the repo and added docs not changing the code at all. How could the tests fail!

Akshat-mehrotra:
according to CONTRIBUTING.md,
I had to make some edits in the strings.md too. However, I am unsure that did I do it as required or not.

base/strings/string.jl
```diff
@@ -206,6 +206,23 @@ function iterate_continued(s::String, i::Int, u::UInt32)
...
+
+# Examples
+
+```jldoctest
+julia> getindex("abcdef", 5)
+'e': ASCII/Unicode U+0065 (category Ll: Letter, lowercase)
+
+julia> getindex("Julia Is The Best Programming Language", 8)
+'s': ASCII/Unicode U+0073 (category Ll: Letter, lowercase)
+
```
logankilpatrick: I think this line needs to be removed.

base/strings/string.jl
```diff
@@ -239,6 +256,25 @@ function getindex_continued(s::String, i::Int, u::UInt32)
     return reinterpret(Char, u)
 end
 
+"""
+    getindex(s::String, r::UnitRange{<:Integer})
+
+Returns a `String` derived from `s` which is situated at indexes from [`Base.first`](@ref) to [`Base.last`](@ref) in `s`.
+
+# Examples
+
```
logankilpatrick: remove this line

Akshat-mehrotra:
Removed

vtjnash:
Thank you @Akshat-mehrotra for the first-time contribution. However, I think this should be documented as the interface, such that these specific usages are merely implementations and don't require special mention. I've opened https://github.com/JuliaLang/julia/issues/40542 to discuss how we could improve the current printing to better clarify its role as an interface.

PR #34377, opened 2020-01-14
Eliminate erroneous quoting on `${PATCHELF}`

staticfloat:
This only manifests on platforms like `aarch64` and `ppc64le` because we
force extra arguments to `PATCHELF` in these cases which didn't interact
well with these well-meaning quotes.


Keno (approved):

staticfloat:
This is necessary, but not sufficient, to get AArch64 passing completely.  Something else is going wrong; I'm debugging.

staticfloat:
Despite the fact that I want to eliminate the `contrib/fixup*` scripts entirely, I think I've fixed these for AArch64, so if this goes green on the `tester_linuxaarch64` builder (thanks to @maleadt for fixing the AArch64 tests!) then we're good to merge.

PR #33864, opened 2019-11-15
RFC: bring back v0.6 scope rules in the REPL

JeffBezanson:
After thinking about it off and on for quite a while, @StefanKarpinski and I more-or-less decided that the best way to change the REPL scope situation (if at all) is just to bring back what v0.6 did. That's what SoftGlobalScope and IJulia already do (albeit somewhat approximately; implementing it internally it's much easier to get it 100%), and seems less disruptive than introducing a *third* behavior.

This needs to be finished up but is ready to try. Please give it a whirl.

I'm not sure what the best interface to it is; it seemed easiest just to drop a special expression in the AST itself.

fixes #28789

PR #34383, opened 2020-01-15
WIP: warn when an implicit local at the toplevel shadows a global

JeffBezanson:
This is a variation of #33864 that prints a warning (in all contexts) whenever, for example, IJulia would interpret an expression differently than normal julia. I'm just putting this up for now in case anybody would like to experiment with it --- try it out at the repl, run package tests, etc. to see what happens.

### DIFF
--- a/src/ast.c
+++ b/src/ast.c
@@ -131,11 +131,33 @@ static jl_value_t *scm_to_julia(fl_context_t *fl_ctx, value_t e, jl_module_t *mo
 static value_t julia_to_scm(fl_context_t *fl_ctx, jl_value_t *v);
 static jl_value_t *jl_expand_macros(jl_value_t *expr, jl_module_t *inmodule, struct macroctx_stack *macroctx, int onelevel);
 
+value_t fl_defined_julia_global(fl_context_t *fl_ctx, value_t *args, uint32_t nargs)
+{
+    // tells whether a var is defined in and *by* the current module
+    argcount(fl_ctx, "defined-julia-global", nargs, 1);
+    (void)tosymbol(fl_ctx, args[0], "defined-julia-global");
+    jl_ast_context_t *ctx = jl_ast_ctx(fl_ctx);
+    jl_sym_t *var = jl_symbol(symbol_name(fl_ctx, args[0]));
+    jl_binding_t *b =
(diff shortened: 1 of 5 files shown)


yuyichao:
I assume the way to suppress the warning if the behavior is intended is to declare it as local explicitly?

JeffBezanson:
Yes, or to rename it.

StefanKarpinski:
Love it! Can I remove the "DO NOT MERGE" label and merge this? 😁

src/julia-syntax.scm
```diff
@@ -2580,27 +2628,51 @@
...
            ,(resolve-scopes- (cadddr e) scope (method-expr-static-parameters e))))
         (else
+         (if (and (eq? (car e) '=) (symbol? (cadr e))
+                  scope (null? (lam:vars (scope:lam scope)))
+                  (warn-var?! (cadr e) scope))
+             (let* ((v    (cadr e))
+                    (loc  (extract-line-file loc))
+                    (line (if (= (car loc) 0) (julia-current-line) (car loc)))
+                    (file (if (eq? (cadr loc) 'none) (julia-current-file) (cadr loc))))
+               (julia-logmsg
```
vtjnash: Lowering probably should not be doing arbitrary IO. I think this needs to inject a `(:warn)` expr, so that the warning definitely gets deferred for runtime.

src/ast.c
```diff
@@ -131,13 +131,35 @@ static jl_value_t *scm_to_julia(fl_context_t *fl_ctx, value_t e, jl_module_t *mo
...
 
+value_t fl_defined_julia_global(fl_context_t *fl_ctx, value_t *args, uint32_t nargs)
+{
+    // tells whether a var is defined in and *by* the current module
+    argcount(fl_ctx, "defined-julia-global", nargs, 1);
+    (void)tosymbol(fl_ctx, args[0], "defined-julia-global");
+    jl_ast_context_t *ctx = jl_ast_ctx(fl_ctx);
+    jl_sym_t *var = jl_symbol(symbol_name(fl_ctx, args[0]));
+    jl_binding_t *b =
+        (jl_binding_t*)ptrhash_get(&ctx->module->bindings, var);
```
vtjnash: Needs thread-safety

Issue #30901, opened 2019-01-30
Printing of time-related types in containers is inconsistent


tkf:
I think #30817 changed `show(io, "text/plain", ::Array{Day})` etc.  On cb7a5691a51b6bbc0e670f601d6269d893340e44 (and in Julia 1.0 and 1.1) it was:

```julia
julia> Day.(1:3)
3-element Array{Day,1}:
 1 day
 2 days
 3 days
```

But now (c0e9182fa94e2b96106e8e4c14cb36dc9d35cb40) it's:

```julia
julia> Day.(1:3)
3-element Array{Day,1}:
 Day(1)
 Day(2)
 Day(3)
```

Furthermore, confusingly (to me), `Matrix{Day}` still prints human readable form:

```julia
julia> [Day(1) Day(2)]
1×2 Array{Day,2}:
 1 day  2 days
```

However, @omus and others approved the following behavior in #30200:

> ```
> julia> fill(d, 2)
> 2-element Array{Date,1}:
>  Date(2018, 12, 7)
>  Date(2018, 12, 7)
> 
> julia> fill(d, 2, 2)
> 2×2 Array{Date,2}:
>  2018-12-07  2018-12-07
>  2018-12-07  2018-12-07
> ```

So maybe the current behavior of `Vector{Day}` and `Matrix{Day}` is intentional?  (But, to be honest, I was confused why `ndims` of an array changes how elements are printed.)

Another confusing part is that `show(io, ::Array{Day})` still shows human-readable form.  Shouldn't it closer to `repr` than 3-argument `show`?

```julia
julia> show(Day.(1:2))
Day[1 day, 2 days]
```

Note that the output of `Dict{Day,Day}` and `Vector{Pair{Day,Day}}` are not changed.  It has been (and is):

```julia
julia> Dict(Day(1) => Day(2))
Dict{Day,Day} with 1 entry:
  1 day => 2 days

julia> [Day(1) => Day(2)]
1-element Array{Pair{Day,Day},1}:
 1 day => 2 days
```

What is the intended printing behavior of `Day`, `Date`, and alike in the containers?

See also: #30683

JeffBezanson:
We should remove the branch that switches to `1 day` based on the `:compact` property. We can optionally add `typeinfo` support as well.

nalimilan:
@tkf Would you make a pull request?

tkf:
I think I can make a PR, but I need to understand the desired output.

[...]

PR #34387, opened 2020-01-15
Make Date/Time Printing Consistent

fchorney:
A lot of discussion has been had on this subject #30901, and I'm hoping this can finally close that chapter.

Just a quick note: I have not modified any of the tests yet, as I'd like confirmation on the output formats before I do all the test changing work.

Changes
- Array element printing to use 3-arg show with compact set to true. 
- Periods (Day, Minute, etc), Date, and DateTime types have had a bit of an overhaul on their print /show/etc methods.

We're now following the convention that `show(io, x)` (and thus `repr(x)`) output  their parsable output e.g. "Day(1)".
and that `print(x)` (and thus `string(x)`) and `show(io,"test/plain", x)` (as is use by REPL display) show the the human readable version e.g. "1 day". You can still see the human readable version in the REPL, using print, string, and accessing specific array elements but other than that you will see the parsable output.

Here is various output of the different date/time types and various different output commands.

```julia
julia> using Dates

# PERIOD

julia> d = Day(1)
1 day

julia> show(d)
Day(1)
julia> show([d])
[Day(1)]
julia> show([d d])
[Day(1) Day(1)]
julia> show(IOContext(stdout, :compact=>true), [d, d])
[Day(1), Day(1)]
julia> show(IOContext(stdout, :compact=>false), [d, d])
[Day(1), Day(1)]
julia> [d, d][1]
1 day

julia> fill(d, 2)
2-element Array{Day,1}:
 1 day
 1 day

julia> fill(d, 2, 2)
2×2 Array{Day,2}:
 1 day  1 day
 1 day  1 day

julia> print(d)
1 day
julia> string(d)
"1 day"

julia> repr(d)
"Day(1)"

julia> fill(d => d, 2)
2-element Array{Pair{Day,Day},1}:
 Day(1)=>Day(1)
 Day(1)=>Day(1)

julia> fill(d => d, 2, 3)
2×3 Array{Pair{Day,Day},2}:
 Day(1)=>Day(1)  Day(1)=>Day(1)  Day(1)=>Day(1)
 Day(1)=>Day(1)  Day(1)=>Day(1)  Day(1)=>Day(1)

# DATE

julia> d = Date("2012-01-01")
2012-01-01

julia> show(d)
Date("2012-01-01")
julia> show([d])
[Date("2012-01-01")]
julia> show([d d])
[Date("2012-01-01") Date("2012-01-01")]
julia> show(IOContext(stdout, :compact=>true), [d, d])
[Date("2012-01-01"), Date("2012-01-01")]
julia> show(IOContext(stdout, :compact=>false), [d, d])
[Date("2012-01-01"), Date("2012-01-01")]
julia> [d, d][1]
2012-01-01

julia> fill(d, 2)
2-element Array{Date,1}:
 2012-01-01
 2012-01-01

julia> fill(d, 2, 2)
2×2 Array{Date,2}:
 2012-01-01  2012-01-01
 2012-01-01  2012-01-01

julia> print(d)
2012-01-01
julia> string(d)
"2012-01-01"

julia> repr(d)
"Date(\"2012-01-01\")"

julia> fill(d => d, 2)
2-element Array{Pair{Date,Date},1}:
 Date("2012-01-01")=>Date("2012-01-01")
 Date("2012-01-01")=>Date("2012-01-01")

julia> fill(d => d, 2, 3)
2×3 Array{Pair{Date,Date},2}:
 Date("2012-01-01")=>Date("2012-01-01")  Date("2012-01-01")=>Date("2012-01-01")  Date("2012-01-01")=>Date("2012-01-01")
 Date("2012-01-01")=>Date("2012-01-01")  Date("2012-01-01")=>Date("2012-01-01")  Date("2012-01-01")=>Date("2012-01-01")

# DATETIME

julia> d = DateTime("2019-01-01T01:01:01.123")
2019-01-01T01:01:01.123

julia> show(d)
DateTime("2019-01-01T01:01:01.123")
julia> show([d])
[DateTime("2019-01-01T01:01:01.123")]
julia> show([d d])
[DateTime("2019-01-01T01:01:01.123") DateTime("2019-01-01T01:01:01.123")]
julia> show(IOContext(stdout, :compact=>true), [d, d])
[DateTime("2019-01-01T01:01:01.123"), DateTime("2019-01-01T01:01:01.123")]
julia> show(IOContext(stdout, :compact=>false), [d, d])
[DateTime("2019-01-01T01:01:01.123"), DateTime("2019-01-01T01:01:01.123")]
julia> [d, d][1]
2019-01-01T01:01:01.123

julia> fill(d, 2)
2-element Array{DateTime,1}:
 2019-01-01T01:01:01.123
 2019-01-01T01:01:01.123

julia> fill(d, 2, 2)
2×2 Array{DateTime,2}:
 2019-01-01T01:01:01.123  2019-01-01T01:01:01.123
 2019-01-01T01:01:01.123  2019-01-01T01:01:01.123

julia> print(d)
2019-01-01T01:01:01.123
julia> string(d)
"2019-01-01T01:01:01.123"

julia> repr(d)
"DateTime(\"2019-01-01T01:01:01.123\")"

julia> fill(d => d, 2)
2-element Array{Pair{DateTime,DateTime},1}:
 DateTime("2019-01-01T01:01:01.123")=>DateTime("2019-01-01T01:01:01.123")
 DateTime("2019-01-01T01:01:01.123")=>DateTime("2019-01-01T01:01:01.123")

julia> fill(d => d, 2, 3)
2×3 Array{Pair{DateTime,DateTime},2}:
 DateTime("2019-01-01T01:01:01.123")=>DateTime("2019-01-01T01:01:01.123")  …  DateTime("2019-01-01T01:01:01.123")=>DateTime("2019-01-01T01:01:01.123")
 DateTime("2019-01-01T01:01:01.123")=>DateTime("2019-01-01T01:01:01.123")     DateTime("2019-01-01T01:01:01.123")=>DateTime("2019-01-01T01:01:01.123")
```


mattBrzezinski:
My only comment would be to squash commits and update the message.

💬

oxinabox (approved):

mattBrzezinski (approved):

StefanKarpinski:
Marking for triage review. Hopefully this is the one!

oxinabox:
Here is a side-by-side of current vs this PR
https://gist.github.com/oxinabox/e98743f50a14915e2fdf8754cf23c91c/revisions?diff=split#diff-699ad01b092e1cf321f46c709404b864

stdlib/Dates/src/io.jl
```diff
@@ -578,33 +578,32 @@ function format(dt::TimeType, f::AbstractString; locale::Locale=ENGLISH)
...
-    return "$yy-$mm-$dd"
+    print(io, "$yy-$mm-$dd")
+end
+
+for date_type in (:Date, :DateTime)
+    # Human readable output (i.e. "2012-01-01")
+    @eval Base.show(io::IO, ::MIME"text/plain", dt::$date_type) = print(io, dt)
+    # Parsable output (i.e. Date("2012-01-01"))
+    @eval Base.show(io::IO, dt::$date_type) = print(
+        io, $date_type, "(\"", sprint(print, dt, context=io), "\")"
```
JeffBezanson:
```suggestion
        io, typeof(dt), "(\"", dt, "\")"
```

JeffBezanson:
This raises a couple design questions:

- What do we do in array printing if the IO context already specifies `:compact => false`? Force it to true, or go back to calling 2-argument show for array elements?

- Many types will now need to handle `:compact` in their 3-argument show methods. Basically, for e.g. arrays and dicts you want to call 2-argument show if compact is set. Should showing array elements just implement that logic: i.e. when an array element is an AbstractArray, AbstractDict, etc. call 2-arg show instead?

stdlib/Dates/src/io.jl
```diff
@@ -578,33 +578,30 @@ function format(dt::TimeType, f::AbstractString; locale::Locale=ENGLISH)
...
-
-function Base.string(dt::DateTime)
+function Base.print(io::IO, dt::DateTime)
     if millisecond(dt) == 0
-        format(dt, dateformat"YYYY-mm-dd\THH:MM:SS", 24)
+        str = format(dt, dateformat"YYYY-mm-dd\THH:MM:SS", 24)
     else
-        format(dt, dateformat"YYYY-mm-dd\THH:MM:SS.s", 26)
+        str = format(dt, dateformat"YYYY-mm-dd\THH:MM:SS.s", 26)
     end
```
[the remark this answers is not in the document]
fchorney: oh right. The python programmer in me keeps forgetting that this is a thing you can do in Julia.

JeffBezanson:
Triage is :+1: on this output.

A good idea from @vtjnash : first call 3-arg show (with compact defaulted to true, but leave it alone if already set), then check whether its output contains line breaks. If it does, try 2-arg show instead. I think that neatly handles both of my questions.

base/arrayshow.jl
```diff
@@ -101,7 +101,21 @@ function print_matrix_row(io::IO,
         if isassigned(X,Int(i),Int(j)) # isassigned accepts only `Int` indices
             x = X[i,j]
             a = alignment(io, x)
-            sx = sprint(show, x, context=io, sizehint=0)
+
+            # If compact is set, keep the value, else set to true
+            compact = get(io, :compact, nothing)
+            if compact == nothing
```
omus:
```suggestion
            if compact === nothing
```

base/math.jl
```diff
@@ -51,7 +51,7 @@ are promoted to a common type.
 ```jldoctest
 julia> clamp.([pi, 1.0, big(10.)], 2., 9.)
 3-element Array{BigFloat,1}:
- 3.141592653589793238462643383279502884197169399375105820974944592307816406286198
```
omus: Do we want to special case this?

fchorney:
So a few things.
1. Guess I can't fix the doctest for `Statistics.jl` as that's an external package now(?)
2. There seems to be some sort of stack overflow error now with printing circular dependency arrays which I haven't been able to track down yet. The failing test line is here: https://github.com/JuliaLang/julia/blob/a25879d5cc969912cf89f5f0f06a431cb2e6cb4f/test/show.jl#L1153

base/arrayshow.jl
```diff
@@ -101,7 +101,21 @@ function print_matrix_row(io::IO,
...
             x = X[i,j]
             a = alignment(io, x)
-            sx = sprint(show, x, context=io, sizehint=0)
+
+            # If compact is set, keep the value, else set to true
+            compact = get(io, :compact, nothing)
+            if compact === nothing
+                compact = true
+            end
+            context = IOContext(io, :compact=>compact)
```
JeffBezanson: I think this can be removed and we can just keep the `:compact` logic in `show(io::IO, ::MIME"text/plain", X::AbstractArray)`.

JeffBezanson:
> 2\. There seems to be some sort of stack overflow error now with printing circular dependency arrays

Ah, I guess this is because 3-argument show was never recursive before, and we only handle cycles for vectors. `show_vector` calls `show_delim_array` which calls `show_circular`. To fix this we probably need to add calls to `show_circular` in the more general N-d array printing code.

base/arrayshow.jl
fchorney: Just to double check, are you suggesting I move that logic to `show(io::IO, ::MIME"text/plain", X::AbstractArray)`? or just deleting this logic and keeping the logic that is currently in that function:

```julia
    if !haskey(io, :compact) && length(axes(X, 2)) > 1
        io = IOContext(io, :compact => true)
    end
```

JeffBezanson: Keep what's currently in that function.

JeffBezanson (approved):

test/show.jl
```diff
@@ -1150,7 +1150,7 @@ let x = [], y = [], z = Base.ImmutableDict(x => y)
     push!(x, y)
     push!(y, x)
     push!(y, z)
-    @test replstr(x) == "1-element Array{Any,1}:\n Any[Any[Any[#= circular reference @-2 =#]], Base.ImmutableDict{Array{Any,1},Array{Any,1}}([Any[#= circular reference @-3 =#]] => [#= circular reference @-2 =#])]"
+    @test replstr(x) == "1-element Array{Any,1}:\n Any[Any[#= circular reference @-2 =#], Base.ImmutableDict{Array{Any,1},Array{Any,1}}([#= circular reference @-3 =#] => [#= circular reference @-2 =#])]"
```
fchorney: so, I had to change this test for how I added in the circular check for 3 arg array show. Does this seem acceptable, or is there perhaps somewhere else I should have added the check?

JeffBezanson: This looks more correct to me than the previous output.

test/show.jl
```diff
@@ -1443,7 +1443,7 @@ end
     # left-alignment when not using the "=>" symbol
     @test replstr(Any[Pair{Integer,Int64}(1, 2), Pair{Integer,Int64}(33, 4)]) ==
         "2-element Array{Any,1}:\n Pair{Integer,Int64}(1, 2)\n Pair{Integer,Int64}(33, 4)"
-end
+    end
```
omus:
```suggestion
end
```

test/show.jl
```diff
@@ -1922,7 +1922,7 @@ Z = Array{Float64}(undef,0,0)
     @test showstr(vec_undefined) == "Any[#undef, #undef]"
     @test showstr(vec_initialisers) == "[$undef, $undef]"
     @test replstr(vec_undefined) == "2-element Array{Any,1}:\n #undef\n #undef"
-    @test replstr(vec_initialisers) == "2-element Array{UndefInitializer,1}:\n $undef\n $undef"
+    @test replstr(vec_initialisers) == "2-element Array{UndefInitializer,1}:\n UndefInitializer(): array initializer with undefined values\n UndefInitializer(): array initializer with undefined values"
```
omus: The compact form would be better in this case. I'm. It sure how that can be done without impacting the 3-arg show used on the REPL

JeffBezanson: Agreed; after this is merged we can add some handling for `:compact` in the 3-arg show methods for UndefInitializer and Char and maybe some others.

rfourquet:
I think this broke the printing of some arrays:
```julia
julia> [[1]]
 [1]

julia> [Int[]]
 0-element Array{Int64,1}
```

StefanKarpinski:
Better open that as an issue and mark it for the appropriate milestone.

Moelf:
Agree that's hella confusing

PR #34388, opened 2020-01-15
Add test for setindex with a matrix of bools

kshyatt:
Trying to test [this](https://codecov.io/gh/julialang/julia/src/master/stdlib/SparseArrays/src/sparsematrix.jl#L2798) - a call to `@which` confirms that this test should call the method.

### DIFF
--- a/stdlib/SparseArrays/test/sparse.jl
+++ b/stdlib/SparseArrays/test/sparse.jl
@@ -1138,6 +1138,18 @@ end
         S[I] .= J
         @test sum(S) == (sumS1 - sumS2 + sum(J))
     end
+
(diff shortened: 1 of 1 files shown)


dkarrasch (commented):
Cool. I'd suggest to include testing all other entries as well.
```julia
A = spzeros(10, 10)
# now all your definitions and tests
@test nnz(A) == 2
```
With that, I'd say you can merge directly.

kshyatt:
Heck yeck, got [most](https://codecov.io/gh/julialang/julia/src/master/stdlib/SparseArrays/src/sparsematrix.jl#L2798) of it

PR #34390, opened 2020-01-15
some basic views documentation

stevengj:
Added a little basic discussion at the beginning of the "views" section of the standard-library documentation, based on [this discourse post](https://discourse.julialang.org/t/could-you-explain-what-are-views/17535/3?u=stevengj) which people seem to have found helpful.

### DIFF
--- a/doc/src/base/arrays.md
+++ b/doc/src/base/arrays.md
@@ -104,6 +104,12 @@ Base.checkindex
 
 ## Views (SubArrays and other view types)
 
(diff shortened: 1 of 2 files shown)


dkarrasch (commented):
Maybe break the lines after 82(?) characters?

Indeed, this addition will be much appreciated.

stevengj:
@dkarrasch, I changed it to a hard wrap, thanks.

Issue #19981, opened 2017-01-11
should allocation tracking ignore compilation by default?


StefanKarpinski:
When using `--track-allocation` it's almost impossible to coax it into not putting a bogus looking large allocation on the first line of each function, which presumably is due to compilation kicking in when the function first runs. Can we turn off allocation tracking when we enter the compiler and then turn it back on once we've left compilation? Perhaps `--track-allocation=user` should exclude compilation time and `--track-allocation=all` should include it?

vchuravy:
> The recommended procedure is to force compilation by executing all the commands you want to analyze, then call `Profile.clear_malloc_data()` to reset all allocation counters. Finally, execute the desired commands and quit Julia to trigger the generation of the .mem files.

Maybe we should make the above more prominent in http://docs.julialang.org/en/stable/manual/profile/#memory-allocation-analysis?

StefanKarpinski:
Wouldn't it be better (if possible) to just do the right thing by default?

PR #34391, opened 2020-01-15
track-allocation: charge entry cost to caller

vtjnash:
This effectively makes two related changes:
1. Completely removes one-time costs of runtime functions (aka "warmup costs") from being charged to the user's code. This was already being partly subtracted in some cases (depending on whether we hit a malloc record point) resulting in both over- and under- reporting of the user's allocations. However, theoretically these costs are implementation details that the user doesn't specifically care about, and should thus be charged against the internal allocation budget instead. (This implements #19981 and should make `Profile.clear_malloc_data()` less necessary, and just get the desired answer on the first pass). In the future, we can continue to sharpen this if people report cases I've missed. We can also even create a bin to put these in so that they continue to be reported, but just charged against the correct cause.
2. Adjust where entry point costs are charged against. Previously we charged the callee for any setup costs incurred by the caller. That sometimes wasn't too bad, but other times would result in very large numbers being added the the first instruction in the user's function. Now, we'll charge those against the caller, so you won't see that happen anymore. Instead those costs should now show up on the line of code that actually incurred them.

### DIFF
--- a/src/codegen.cpp
+++ b/src/codegen.cpp
@@ -322,6 +322,7 @@ Function *juliapersonality_func;
 #endif
 #endif
 static Function *diff_gc_total_bytes_func;
+static Function *sync_gc_total_bytes_func;
 static Function *jlarray_data_owner_func;
 static GlobalVariable *jlgetworld_global;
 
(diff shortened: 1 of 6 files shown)


KristofferC:
Would it be better to let this go into 1.5. It doesn't seem that urgent?

PR #34392, opened 2020-01-15
coverage: fix name of the new code_coverage_effect Expr head [NFCI]

vtjnash:


KristofferC:
Removing backport label since original PR doesn't have it.

PR #34394, opened 2020-01-15
alternative fix for mul!

### LINKED ISSUE #684: mul! performance regression on master

daviehh:
Currently, the `MulAddMul` is implemented as a callable structure for calculating a short-circuit version of `x * a + y * b` depending on whether a is 1 or b is 0. The short-circuit is implemented with "[value types](https://docs.julialang.org/en/v1/manual/types/#"Value-types"-1)" on a pair of Booleans (ais1, bis0). This may lead to a performance regression on master, possibly due to constant propagation, details at https://github.com/JuliaLang/LinearAlgebra.jl/issues/684

This is an alternative way to fix https://github.com/JuliaLang/LinearAlgebra.jl/issues/684 by implementing the short-circuit callable structure `MulAddMul` differently, where the short-circuiting is done in the function body rather than depending on the dispatch system for the value type (ais1, bis0). The 3-argument `mul!` here might be a bit slower compared with 1.3.1 (~7-10 ns on my testings), not sure how to improve, but this might be less brittle than constant propagation?

test script:
```
using LinearAlgebra
using BenchmarkTools

ndim = 3
al = .5;
bt = .7im;

@btime mul!($ou, $m1, $m2);
@btime mul!($ou, $m1, $m2, $al, $bt);
@btime mul!($ou, $m1, $m2, .5, .7im);
```

on 1.3.1:
Julia Version 1.3.1
Commit 2d5741174c* (2019-12-30 21:36 UTC)

>   35.637 ns (0 allocations: 0 bytes)
>   350.521 ns (3 allocations: 112 bytes)
>   54.422 ns (0 allocations: 0 bytes)

on master:
Julia Version 1.5.0-DEV.71
Commit 15d693b0ec (2020-01-15 18:13 UTC)

>   397.612 ns (1 allocation: 16 bytes)
>   369.288 ns (3 allocations: 80 bytes)
>   371.888 ns (3 allocations: 80 bytes)

this PR:
Julia Version 1.5.0-DEV.72
Commit 20f2720891 (2020-01-15 20:16 UTC)

>   45.862 ns (0 allocations: 0 bytes)
>   59.554 ns (0 allocations: 0 bytes)
>   61.749 ns (0 allocations: 0 bytes)

### DIFF
--- a/stdlib/LinearAlgebra/src/generic.jl
+++ b/stdlib/LinearAlgebra/src/generic.jl
@@ -20,22 +20,22 @@ julia> MulAddMul(12, 34)(56, 78) == 56 * 12 + 78 * 34
 true
 ```
 """
-struct MulAddMul{ais1, bis0, TA, TB}
+struct MulAddMul{TA, TB}
     alpha::TA
     beta::TB
 end
 
-MulAddMul(alpha::TA, beta::TB) where {TA, TB} =
-    MulAddMul{isone(alpha), iszero(beta), TA, TB}(alpha, beta)
-
 MulAddMul() = MulAddMul(true, false)
 
-@inline (::MulAddMul{true})(x) = x
-@inline (p::MulAddMul{false})(x) = x * p.alpha
-@inline (::MulAddMul{true, true})(x, _) = x
-@inline (p::MulAddMul{false, true})(x, _) = x * p.alpha
-@inline (p::MulAddMul{true, false})(x, y) = x + y * p.beta
-@inline (p::MulAddMul{false, false})(x, y) = x * p.alpha + y * p.beta
+@inline (p::MulAddMul)(x) = isone(p.alpha) ? x : x * p.alpha
+@inline function (p::MulAddMul)(x, y)
+    x_mul_a = isone(p.alpha) ? x : x * p.alpha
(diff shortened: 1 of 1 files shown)


dkarrasch:
I have wanted to move short-circuiting to the runtime code, but couldn't figure out how to do that with minimal invasion. This is a very elegant way to do that.

I wonder one thing, though: we take care of whether `a` is one, and avoid multiplication if it is. We don't do that with `b`, and haven't done that earlier. Is that because this is a special, yet not default case? Would it make sense/be worth it to add this logic here? It would add a second layer of branching, and I guess the one layer you added is responsible for the little slowdown.

Anyway, I think the fact that this handles 5-arg mul! nicely is a great achievement.

daviehh:
closed in favor of https://github.com/JuliaLang/julia/pull/34384

PR #34400, opened 2020-01-16
RFC: Add `Lockable{T, L<:AbstractLock}` struct

DilumAluthge:
I find myself using this pattern a lot, so I'm curious if it would be worth adding to `Base`.

The basic idea is that there is some object that should only ever be accessed when its corresponding lock is held, so it makes sense to package the object and its lock together into a single object.

Here is the code that this PR adds to `base/lock.jl`:

```julia
struct Lockable{T, L<:AbstractLock}
    value::T
    lock::L
end

Lockable(value) = Lockable(value, ReentrantLock())

function lock(f, l::Lockable)
    lock(l.lock) do
        f(l.value)
    end
end
```

## Example usage:

```julia
julia> function foo(x::Lockable)
           lock(x) do value
               push!(value, value[end-1] + value[end])
           end
       end

julia> bar = Lockable([1, 1], ReentrantLock())

julia> foo(bar)
3-element Array{Int64,1}:
 1
 1
 2

julia> foo(bar)
4-element Array{Int64,1}:
 1
 1
 2
 3

julia> foo(bar)
5-element Array{Int64,1}:
 1
 1
 2
 3
 5
```

### DIFF
--- a/base/lock.jl
+++ b/base/lock.jl
@@ -210,17 +210,6 @@ function lock(f, l::Lockable)
     end
 end
 
-function trylock(f, l::Lockable)
-    if trylock(l.lock)
-        try
-            return f(l.value)
-        finally
-            unlock(l.lock)
-        end
-    end
-    return false
-end
-
 # implement the rest of the Lock interface on Lockable
 islocked(l::Lockable) = islocked(l.lock)
 lock(l::Lockable) = lock(l.lock)
--- a/base/lock.jl
+++ b/base/lock.jl
@@ -164,6 +164,52 @@ function lock(f, l::AbstractLock)
     end
 end
 
+"""
+Lockable(value, lock = ReentrantLock())
+
+Creates a `Lockable` object that wraps `value` and
+associates it with the provided `lock`.
+
+!!! compat "Julia 1.5"
+    Requires at least Julia 1.5.
+"""
+struct Lockable{T, L<:AbstractLock}
+    value::T
+    lock::L
+end
+
+"""
+
+Lockable(value)
+
+Creates a `Lockable` object that wraps `value` and
+associates it with a newly created `ReentrantLock`.
+
+!!! compat "Julia 1.5"
+    Requires at least Julia 1.5.
+"""
+Lockable(value) = Lockable(value, ReentrantLock())
+
+"""
+    lock(f::Function, l::Lockable)
+
+Acquire the lock associated with `l`, execute `f` with the lock held,
+and release the lock when `f` returns. `f` will receive one positional
+argument: the value wrapped by `l`. If the lock is already locked by a
+different task/thread, wait for it to become available.
+
+When this function returns, the `lock` has been released, so the caller should
+not attempt to `unlock` it.
+
+!!! compat "Julia 1.5"
+    Requires at least Julia 1.5.
+"""
+function lock(f, l::Lockable)
+    lock(l.lock) do
+        f(l.value)
+    end
+end
+
 function trylock(f, l::AbstractLock)
     if trylock(l)
         try
--- a/base/lock.jl
+++ b/base/lock.jl
@@ -210,6 +210,23 @@ function lock(f, l::Lockable)
     end
 end
 
+function trylock(f, l::Lockable)
+    if trylock(l.lock)
+        try
+            return f(l.value)
+        finally
+            unlock(l.lock)
+        end
+    end
+    return false
+end
+
+# implement the rest of the Lock interface on Lockable
+islocked(l::Lockable) = islocked(l.lock)
+lock(l::Lockable) = lock(l.lock)
+trylock(l::Lockable) = trylock(l.lock)
+unlock(l::Lockable) = unlock(l.lock)
+
 function trylock(f, l::AbstractLock)
     if trylock(l)
         try
--- a/NEWS.md
+++ b/NEWS.md
@@ -36,7 +36,7 @@ Language changes
 
 Multi-threading changes
 -----------------------
-
+* There is a new struct `Lockable{T, L<:AbstractLock}` that makes it easy to bundle a resource and its lock together.
 
 Build system changes
 --------------------
--- a/base/exports.jl
+++ b/base/exports.jl
@@ -656,6 +656,7 @@ export
     istaskfailed,
     lock,
     notify,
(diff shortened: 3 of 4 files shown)


TsurHerman:
Maybe change the name to Lockable?

base/lock.jl
```diff
@@ -164,6 +164,17 @@ function lock(f, l::AbstractLock)
     end
 end
 
+struct Locked{L<:AbstractLock, T}
+    lock::L
+    value::T
+end
+
+function lock(f::Callable, l::Locked)
```
tkf:
```suggestion
function lock(f, l::Locked)
```

Since any Julia object can be a callable, I don't think `f` should be restricted to `Function`s and `Type`s.  Similar discussions: https://github.com/JuliaLang/julia/pull/34296 https://github.com/JuliaLang/julia/pull/30382

base/lock.jl
```diff
@@ -164,6 +164,17 @@ function lock(f, l::AbstractLock)
     end
 end
 
+struct Locked{L<:AbstractLock, T}
```
tkf: Does it make sense to use `Locked{T, L<:AbstractLock}`? I'm guessing that dispatching on `Locked{<:MyType}` is more common than `Locked{<:LockType}`.

DilumAluthge:
> Maybe change the name to Lockable?

Done!

base/lock.jl
DilumAluthge: I agree. I have changed it to `Lockable{T, L<:AbstractLock}`.

base/lock.jl
DilumAluthge: I've removed the `::Callable` restriction.

DilumAluthge:
I have added docstrings, compat annotations, tests, and a `NEWS.md` entry. I've also applied all of the suggestions from code review.

DilumAluthge:
Bump.

What did triage think of this PR?

base/lock.jl
```diff
@@ -164,6 +164,52 @@ function lock(f, l::AbstractLock)
...
+not attempt to `unlock` it.
+
+!!! compat "Julia 1.5"
+    Requires at least Julia 1.5.
+"""
+function lock(f, l::Lockable)
+    lock(l.lock) do
+        f(l.value)
+    end
+end
```
quinnj: While this is probably the most useful pattern, it probably makes sense to implement the full `Lock` interface, right? i.e. plain `lock(l::Lockable)`, `unlock(l::Lockable)`, `trylock`, `islocked`, etc.

DilumAluthge: Yeah that would not be difficult.

Should `Lockable` subtype `AbstractLock`? Right now, it does not.

quinnj: I'd say yes.

tkf: I'd suggest not to. Currently, `lock(f, l::AbstractLock)` works with `f` without an argument:

https://github.com/JuliaLang/julia/blob/ef0e0f2923fbbf6f88d01d9a8f97d62c0c33bdbd/base/lock.jl#L158-L165

I think it'd be confusing the arity of `f` in `lock(f, l::AbstractLock)` depends on the exact `typeof(l)`.

DilumAluthge: Yeah that's a good point.

So I've implemented the rest of the Lock interface. But I did not make `Lockable`a subtype of `AbstractLock`.

DilumAluthge: @quinnj @tkf Are you satisfied that this is resolved?

base/lock.jl
```diff
@@ -164,6 +164,69 @@ function lock(f, l::AbstractLock)
...
+function trylock(f, l::Lockable)
+    if trylock(l.lock)
+        try
+            return f(l.value)
+        finally
+            unlock(l.lock)
+        end
+    end
+    return false
+end
```
tkf: I'm unsure about this as it's not easy for the caller to decide if `f` is called or not reliably.  I think it might be better to just not define `trylock`.

Having said that, I'd suggest something like

```julia
function trylock(f, l::Lockable)
    if trylock(l.lock)
        try
            return Some(f(l.value))
        finally
            unlock(l.lock)
        end
    end
    return nothing
end
```

if we were to define this.

Maybe @quinnj also has some ideas about the implementation?

DilumAluthge: I’d rather just not define `trylock`.

DilumAluthge: Frankly, no one should be using a `Lockable` except in the `lock(lockable) do object` context. After all, that’s the whole point of bundling the lock and the object together.

DilumAluthge: Alright, I no longer define the method `trylock(f, l::Lockable)`.

@tkf Can this be resolved?

base/lock.jl
DilumAluthge: @quinnj @tkf I'm going to mark this as resolved. I have implemented the Lock interface, except for the `trylock` function per @tkf's recommendations.

Reopen this if I am mistaken.

base/lock.jl
tkf: I don't see the resolve button. Maybe because I did it as a comment? If you (or anyone) see the resolve button please go ahead and do it.

DilumAluthge:
The more I think about this, the more I think this doesn't need to be in Base.

I think it is a useful design pattern, but it is easy to put this in a package instead of Base or a stdlib.

quinnj:
I've come across the need several times lately for something like this; does it live in a package somewhere?

DilumAluthge:
Not that I'm aware of. Want to take the code and submit it somewhere?

If there is some kind of "threading utilities" package, it could go there.

quinnj:
I mean, for my use case, I realized I only needed:
```julia
struct Lockable{T}
    x::T
    lock::ReentrantLock
end

Lockable(x::T) where {T} = Lockable{T}(x, ReentrantLock())

Base.lock(x::Lockable) = lock(x.lock)
Base.unlock(x::Lockable) = unlock(x.lock)
```
So I just put it in my package directly. But yeah, sounds like a good fit for a thread utils package.

quinnj:
Well, I was planning on registering this new package soon anyway, so now it includes `Lockable`.

quinnj:
https://github.com/JuliaServices/WorkerUtilities.jl

Issue #24569, opened 2017-11-11
issues with the semantics of top-level expressions


JeffBezanson:
Top-level expressions are special because they are able to cause certain side-effects that strongly affect program behavior and that ordinary expressions can't cause. These are:

1. Binding resolution (`global`, `const`, `import`, `using`)
2. Method definition

Fortunately, that's all. (Type definitions could be on this list, but the only tricky part of their behavior is assigning a global name, which is equivalent to item 1.)

The most desirable behavior for any statements not inside methods is "interpreter-like": all statements should see all effects of all previous statements. Unfortunately this is at odds with compiling fast code, which would like to assume a fixed world state (i.e. world counter). This tension has led to several bugs and tricky design decisions, such as:

Code behaving differently based on whether it gets compiled: #2586 #24566
Top-level code breaking inference's fixed-world assumption: #24316
Problems with binding effects: #18933 #22984 #12010

Three general kinds of solutions are possible: (1) Change top-level semantics to enable optimizations, (2) make optimizations sound with respect to top-level semantics, or (3) don't optimize top-level expressions. (1) is likely to be unpopular, since it would lead to things like:

```
f(x) = 1
begin
    f(x::Int) = 2
    f(1)  # gives 1
end
```

due to inference's usual fixed-world assumption. That leads us to (2). But it wouldn't be able to optimize very much. Consider a case like

```
for i = 1:100
    include("file$i.jl")
    f(1)
end
```

where we can't reasonably assume anything about what `f(1)` does.

That brings us to (3). In practice, it is basically never useful to infer and compile a top-level expression. The only exceptions are casual benchmarking (`@time`), and maybe an occasional long-running loop in the REPL. So to compromise, for a long time we've been compiling top-level expressions if they have loops. But that's a rather blunt instrument, since compiling loops that e.g. call `eval` is not worthwhile, and in other cases is unsound (#24316).

I'm not sure what to do, but overall I think we should optimize top-level code much less often (which might improve load times as well). One idea is to add a `@compile` or `@fast` macro that turns off top-level semantics, basically equivalent to wrapping the code in a 0-arg function and calling it.

Thoughts?

StefanKarpinski:
Would it be possible to compile loops only if they don't call anything that does one of the above special actions or results in eval being called? I suspect that requires trying some amount of compilation and bailing out if those conditions are encountered, which is a bit suboptimal.

yuyichao:
[...]

PR #34405, opened 2020-01-16
remove code forcing testsets to be compiled

### LINKED ISSUE #34252: Functions behaves strangely different inside and outside `@testset`
MWE from PkgEval error in ExactPredicates.jl (https://github.com/maleadt/BasePkgEvalReports/blob/0bb12607c245153c8e7215788e580d43d7d2d51a/pkgeval-9bd498d_vs_2e6715c/logs/ExactPredicates/1.4.0-DEV-082f446f52.log#L64-L81)

Running

```jl
using Test
using ExactPredicates
p2(x, y=0.0) = (float(x), float(y))
p3(x, y=0, z=0) = (float(x), float(y), float(z))
```

and then running the code below *twice*(!!)

```jl
@testset "pertubations" begin
    a = p2(2.0)
    b = p2(1.0, 1.0)
    c = p2(1.0, - 1.0)
    q = p2(prevfloat(0.0))
    @test incircle(a, b, c, q) == -1
end
```

makes it fail with

```jl
pertubations: Test Failed at REPL[23]:6
  Expression: incircle(a, b, c, q) == -1
   Evaluated: 0 == -1
```

If we move the definitions of `a`, `b`, `c`, `q` outside the `@testset`, the `@test`  never fails even after executing it many times. 

The package does some pretty heavy metaprogramming https://github.com/lairez/ExactPredicates.jl/blob/216010eb619e71b51f24020e71902382411d0d81/src/Codegen.jl#L371-L478 but not sure if that is relevant.

This doesn't happen on 1.3.1.

JeffBezanson:
This was added quite a while ago, and by now we have fixed interpreter backtraces so it shouldn't be needed. Removing it should cut down compile time and make tests complete a bit faster. By luck, it also happens to fix #34252 (if the package tests change, e.g. a loop is added, the problem could come back).


cafaxo:
Did this actually fix #34252? It is not clear from your description. Github might have closed #34252 automatically.

KristofferC:
I think it fixes the particular instance but the general issue: https://github.com/JuliaLang/julia/issues/24569 is still open.

PR #34412, opened 2020-01-16
fix #33882, `step(::StepRangeLen)` should not usually convert

### LINKED ISSUE #33882: step(::StepRangeLen{T}) converts step to T
https://github.com/JuliaLang/julia/blob/05a80e1e82ac976bd3323f10e1139d86b3954cbe/base/range.jl#L503

StepRangeLen does not ask that `S` be convertable to `T`, but requires it in `step`.  

It seems like this was done intentionally for Floats but doesn't account for cases where the step can't (or shouldn't) be converted to the type of the start. This would come up more often for `AbstractTime`, but `AbstractTime` has the exact combination of traits to prevent it from being implicitly created by `range`.

```
julia> StepRangeLen(DateTime(2018,1,1), Hour(1), 2)
2018-01-01T00:00:00:0001-01-01T01:00:00:2018-01-01T01:00:00
```
That's not a correct value for the step.

In the case of https://github.com/invenia/Intervals.jl/issues/78, it's not even possible to convert the step to the type of the start, so errors are thrown.

Not sure what the right resolution here is but I don't think we should leave it as is.

JeffBezanson:
fix #33882

### DIFF
--- a/base/range.jl
+++ b/base/range.jl
@@ -500,7 +500,8 @@ julia> step(range(2.5, stop=10.9, length=85))
 """
 step(r::StepRange) = r.step
 step(r::AbstractUnitRange{T}) where{T} = oneunit(T) - zero(T)
-step(r::StepRangeLen{T}) where {T} = T(r.step)
+step(r::StepRangeLen) = r.step
+step(r::StepRangeLen{T}) where {T<:AbstractFloat} = T(r.step)
 step(r::LinRange) = (last(r)-first(r))/r.lendiv
 
 step_hp(r::StepRangeLen) = r.step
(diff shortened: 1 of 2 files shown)


base/range.jl
```diff
@@ -500,7 +500,8 @@ julia> step(range(2.5, stop=10.9, length=85))
 """
 step(r::StepRange) = r.step
 step(r::AbstractUnitRange{T}) where{T} = oneunit(T) - zero(T)
-step(r::StepRangeLen{T}) where {T} = T(r.step)
+step(r::StepRangeLen) = r.step
+step(r::StepRangeLen{T}) where {T<:AbstractFloat} = T(r.step)
```
timholy: What happens if you're even more selective here and restrict it to `StepRangeLen{T,TwicePrecision{T},TwicePrecision{T}}`? Any problems?

timholy (approved):

PR #25862, opened 2018-02-02
add NEWS for mutable struct singletons change (#25854)

JeffBezanson:
[ci skip]

PR #25854, opened 2018-02-01
Make empty mutable struct distinct

Keno:
Fixes #17149

PR #34416, opened 2020-01-17
Fix #34414: remove legacy doc content [skip ci]

### LINKED ISSUE #34414: [Doc] `===` is not a valid way to test singletons
In the [types](https://docs.julialang.org/en/v1/manual/types/) doc page, it says
> Immutable composite types with no fields are singletons; there can be only one instance of such types:
```julia
julia> struct NoFields
       end

julia> NoFields() === NoFields()
true
```

Nonetheless, `===` is not a valid way to test singletons. In [the doc page](https://docs.julialang.org/en/v1/base/base/#Core.:===) of `===`, it says
> First the types of x and y are compared. If those are identical, mutable objects are compared by address in memory and immutable objects (such as numbers) are compared by contents at the bit level.

Therefore, `===` is the same as `==` for immutable types, and it does not compare the addresses. Indeed, I tried another immutable type but *with fields*; the result is the same.

Then, I tested it for *mutable* types, and there was no chance:
```julia
julia> mutable struct NoFields
       end

julia> NoFields() == NoFields()
false

julia> NoFields() === NoFields()
false
```

In conclusion, the doc is wrong and misleading.

WenjieZ:
Remove the paragraphs saying "Immutable composite types with no fields are singletons".

Previously, `MutableNoFields() === MutableNoFields()`, but this was changed in #25854. When that got changed, the existing docs was tweaked to say that the struct had to be immutable(#25862) . However, there is no longer anything specific about it having no fields. The docs about singletons can thus just be removed since they are not special anymore.

### DIFF
--- a/doc/src/manual/types.md
+++ b/doc/src/manual/types.md
@@ -384,19 +384,6 @@ to point to different objects.
 Where required, mutable composite objects can be declared with the keyword [`mutable struct`](@ref), to be
 discussed in the next section.
 
-Immutable composite types with no fields are singletons; there can be only one instance of such types:
-
-```jldoctest
-julia> struct NoFields
-       end
-
-julia> NoFields() === NoFields()
-true
-```
-
-The [`===`](@ref) function confirms that the "two" constructed instances of `NoFields` are actually one
-and the same. Singleton types are described in further detail [below](@ref man-singleton-types).
-
 There is much more to say about how instances of composite types are created, but that discussion
 depends on both [Parametric Types](@ref) and on [Methods](@ref), and is sufficiently important
 to be addressed in its own section: [Constructors](@ref man-constructors).


yuyichao (changes requested):
See comment in the issue. This part is still correct (use of `===`), relevant (it's defining singleton) and should not be removed. It can be improved by mentioning the consistency with the immutable semantics and such change will be welcome.

ViralBShah:
Please reopen if @yuyichao's suggestion is updated into this PR.

WenjieZ:
Sorry for my delayed reply. I am working on COVID-19, and thus this PR is not on my priority list. I would be happy if someone else is willing to take care of this PR.

ViralBShah:
@JeffBezanson Perhaps this is a quick one for you?

StefanKarpinski:
https://github.com/JuliaLang/julia/pull/35938

PR #34419, opened 2020-01-17
add supertypes(T) function

### LINKED ISSUE #34362: supertypes(T) function?
When trying to understand (or explain) a type `T`, I often find myself typing `supertype(T)`, `supertype(supertype(T))`, `supertype(supertype(supertype(T)))`, etcetera in order to see the type hierarchy.  This is tedious and annoying.  It would be nice to have a function `supertypes(T)` that just returned a tuple of `T` and all its supertypes, e.g.

```jl
julia> supertypes(Int)
(Int64, Signed, Integer, Real, Number, Any)
```

Sample implementation:
```jl
function supertypes(T::Type)
    S = supertype(T)
    return S === T ? (T,) : (T, supertypes(S)...)
end
```
It could be made to execute completely statically (e.g. is adding `@pure` enough?).

stevengj:
Closes #34362, adding this function to InteractiveUtils along with `subtypes`.

Unlike `subtypes`, returns a tuple instead of an array, just in case we ever decide that we want this function to work statically (since it can).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -32,6 +32,7 @@ New library features
 Standard library changes
 ------------------------
 * The `@timed` macro now returns a `NamedTuple` ([#34149])
+* New `supertypes(T)` function returns a tuple of all supertypes of `T` ([#34419]).
 
 #### LinearAlgebra
 * The BLAS submodule now supports the level-2 BLAS subroutine `hpmv!` ([#34211]).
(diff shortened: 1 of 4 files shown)


JeffBezanson:
One reason to discourage code from depending on supertypes is that they are not part of API stability guarantees --- we are allowed to insert new abstract types in the hierarchy in 1.x. So enabling compile-time tricks with them might not be for the best.

stevengj:
@JeffBezanson, I'm not sure I understand your point.  You can't insert new abstract types into the hierarchy in a given binary *after compilation*, so compile-time tricks are perfectly safe, no?  Besides, the `supertype` function itself is already `@_pure_meta`, so the horse is out of the barn.

JeffBezanson:
It's safe, just subject to breakage. But ¯\_(ツ)_/¯

PR #34421, opened 2020-01-17
Add Filesystem Docs

Akshat-mehrotra:


Keno:
The examples here are all windows specific, we should have examples for Unix operating systems also. @mortenpi @fredrikekre do we have OS specific examples for doc examples/doctests? In addition there's a bunch of uses of `julia-repl`, with non-REPL examples.

Akshat-mehrotra:
Made it non-windows specific and replaced `julia-repl` with `jldoctest`.

logankilpatrick:
Nice work @Akshat-mehrotra

mortenpi:
> The examples here are all windows specific, we should have examples for Unix operating systems also. @mortenpi @fredrikekre do we have OS specific examples for doc examples/doctests?

Just for the record, at the moment we don't have that. But we could add an argument to `jldoctest` for that if we need this feature.

ViralBShah:
A lot of this documentation is now available. We could possibly use some of the examples and  jldoctest sections in here, but best to start a new PR.

PR #28982, opened 2018-08-31
original text will lead to fail

KM-200:
new users are bound to be frustrated by a non-working "hello world" again. quelle surprise.

PR #34422, opened 2020-01-17
Small tweak to embedding.md

Keno:
Adds -rpath to the link line which was suggested in #28982 and removes the line about JULIA_BINDIR, which isn't necessary anymore, since we autodetect the directory. Also remove the reference to libstdc++. If you built julia yourself, it's definitely compatible with the system one. If not, we now build julia binary distributions with wide compatibility with different libstdc++ versions, so the system one should be fine.

### DIFF
--- a/doc/src/manual/embedding.md
+++ b/doc/src/manual/embedding.md
@@ -41,11 +41,8 @@ and link against `libjulia`. For instance, when Julia is installed to `$JULIA_DI
 the above test program `test.c` with `gcc` using:
 
 ```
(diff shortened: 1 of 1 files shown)


vtjnash (approved):
Yeah, the previous suggestions were kinda bad. Though should we suggest `$ORIGIN/../lib`?

Keno:
> Though should we suggest `$ORIGIN/../lib`?

That doesn't work, because `$ORIGIN` is the location of the executable, which can be anywhere. I think this is fine as is. People who want to do more fancy things will need to know what rpath is.

PR #29370, opened 2018-09-26
Minor English fixes/improvements

mark-summerfield:
There is mention of IJulia; ought there also be mention of Jupyter?

I left in the humour ('sold separately', 'secret sauce'), but think it is distracting and ought to go. (Of course, jokes can work in docs, but these I find rather lame.)

In one place this file has `f(x, y=1)` and in another `f(x, y = 1)`; I think a consistent style would be better.

In Tables, is the default (i.e., no `:`s) really right-align?

PR #29286, opened 2018-09-20
Minor English improvements

mark-summerfield:

PR #34424, opened 2020-01-17
Roll up stalled out doc PRs

Keno:
This rolls up and rebases doc PRs #29370 and #29286 which were reviewed and basically ready to go, but had stalled out.


Keno:
Added a bunch more PRs that were in the same boat.

PR #22235, opened 2017-06-06
Converting between integer types

arghhhh:
Julia currently has two ways of converting between integer types.  
You can construct one integer type from another and this is range checked eg:
```
julia> Int8(1000)
ERROR: InexactError()
```
The other way is to use the modulo operator `%`:
```
julia> 1000 % Int8
-24
```
This never fails and just chops the MSBs to give you the answer modulo style - and this is very fast (a single LLVM instruction).

This pull request adds a third method - where you also want the conversion to always succeed, but you want the closest representable value. This is essentially the `clamp` function followed by the `%` operator  - so I've extended the functionality of  `clamp` rather than make up a new name. eg:
```
julia> clamp( 260, UInt8 )
0xff
```
For completeness, I have also added the equivalent in-place function `clamp!`  though it cannot change the type.

PR #34426, opened 2020-01-17
Add clamp(x, T)

Keno:
This is #22235, but with the reversed argument order.
```
julia> clamp( 260, UInt8 )
0xff
```

Co-authored-by: "arghhhh <david.hossack@gmail.com>"

### DIFF
--- a/base/math.jl
+++ b/base/math.jl
@@ -69,6 +69,22 @@ clamp(x::X, lo::L, hi::H) where {X,L,H} =
                   convert(promote_type(X,L,H), x)))
 
 """
+    clamp(x, T)::T
+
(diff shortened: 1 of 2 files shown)


JeffBezanson:
A difficult argument order situation, but I think I agree. My explanation would be that functions primarily about converting between types (convert, trunc, ...) put the type first, but other functions where a type stands in for numbers, ranges, etc. leave the argument in its natural position (clamp, %, ...).

Keno:
Yes, I think the genericness argument wins here. E.g. if we wanted to clamp to some interval `I` in a package that method should be `clamp(x, I)`. Merging.

PR #34427, opened 2020-01-18
Add `isdisjoint`

Keno:
This was added in #13192, but the PR became stale. We've also changed
a bit how we write these kinds of algorithms, so make use of the tools
we have (e.g. `fastin`).

### DIFF
--- a/base/abstractset.jl
+++ b/base/abstractset.jl
@@ -260,6 +260,8 @@ true
 """
 issubset, ⊆, ⊇
 
+const FASTIN_SET_THRESHOLD = 70
+
 function issubset(l, r)
     if haslength(r) && (isa(l, AbstractSet) || !hasfastin(r))
         rlen = length(r) # conditions above make this length computed only when needed
@@ -269,7 +271,7 @@ function issubset(l, r)
         end
         # when `in` would be too slow and r is big enough, convert it to a Set
         # this threshold was empirically determined (cf. #26198)
-        if !hasfastin(r) && rlen > 70
+        if !hasfastin(r) && rlen > FASTIN_SET_THRESHOLD
             return issubset(l, Set(r))
         end
     end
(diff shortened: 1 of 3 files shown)


base/abstractset.jl
```diff
@@ -375,6 +377,27 @@ function issetequal(l, r)
...
+    isdisjoint(v1, v2) -> Bool
+
+Returns whether the collections `v1` and `v2` are disjoint, i.e. whether
+their intersection is empty.
+"""
+function isdisjoint(l, r)
+    function _isdisjoint(l, r)
+        hasfastin(r) && return !any(in(r), l)
+        hasfastin(l) && return !any(in(l), r)
+        haslength(r) && length(r) < FASTIN_SET_THRESHOLD &&
```
rfourquet: Just a note, in `issubset` the conversion to `Set` can be done only if `haslength(r)`, here you made the opposite choice... Was it for a specific reason? (I didn't think much about it, but indeed converting to `Set` when length is unknown seems like a better choice to me; what about updating `issubset` then?).

base/abstractset.jl
```diff
@@ -375,6 +377,27 @@ function issetequal(l, r)
...
+            return !any(in(r), l)
+        return !any(in(Set(r)), l)
+    end
+    if haslength(l) && haslength(r) && lenght(r) < length(l)
```
rfourquet: Sometimes `lenght` is not constant time (e.g. `BitSet`), and in some cases it might be computed twice here. Not a big deal at all (I don't have any example where this is an actual problem (this requires `haslengh`, `!hasfastin` and non-fast `length`)), but if this can be easily re-arranged to not compute the length twice, why not!

test/sets.jl
```diff
@@ -301,13 +301,15 @@ end
...
 
-@testset "issubset, symdiff" begin
+@testset "issubset, symdiff, isdisjoint" begin
     for S in (Set, BitSet, Vector)
         for (l,r) in ((S([1,2]),     S([3,4])),
                       (S([5,6,7,8]), S([7,8,9])),
                       (S([1,2]),     S([3,4])),
                       (S([5,6,7,8]), S([7,8,9])),
                       (S([1,2,3]),   S()),
+                      (Set(),        Set()),
```
rfourquet: Should be `S()`. Or does it not work properly with `BitSet` and `Vector`?

base/abstractset.jl
Keno: Yes, converting to `Set` in general seemed like a good idea to me, with the exception of a fast path when the thing we're iterating over happens to be short.

base/abstractset.jl
Keno: I think we generally assume that `length` is constant time. Otherwise these checks are unlikely to be profitable. I don't think it's worth the extra complication. We're already doing a quadratic operation here.

test/sets.jl
Keno: Yep, sorry

rfourquet (approved):

Issue #34302, opened 2020-01-07
Distributed: colons in numeric IPv6 address break format for specifying port number


mgkuhn:
`function launch_on_machine` in [managers.jl](https://github.com/JuliaLang/julia/blob/master/stdlib/Distributed/src/managers.jl) currently parses the address of a machine as
```julia
    machine_def = split(machine_bind[1], ':')
    # if this machine def has a port number, add the port information to the ssh flags
    if length(machine_def) > 2
        throw(ArgumentError("invalid machine definition format string: invalid port format \"$machine_def\""))
    end
    host = machine_def[1]
    portopt = ``
    if length(machine_def) == 2
        portstr = machine_def[2]
        if !all(isdigit, portstr) || (p = parse(Int,portstr); p < 1 || p > 65535)
            msg = "invalid machine definition format string: invalid port format \"$machine_def\""
            throw(ArgumentError(msg))
        end
        portopt = ` -p $(machine_def[2]) `
    end
```

So if the machine string contains a `:` character, the function assumes that this separates a host address from a port number. However, numeric IPv6 addresses also contain lots of colons!

The URL syntax had the same problem, and there [RFC 2732](https://tools.ietf.org/html/rfc2732) introduced the (somewhat cumbersome) convention/fix that in order to provide both a numeric IPv6 address and a port number separated by a colon, the former has to be enclosed in square brackets, as in `http://[2001:db8::1]:80/index.html`.

How do we want to fix this in `Distributed`:
1. Support the RFC 2732 square-bracket notation? This would be fully backwards compatible and is well known from URLs.
2. Switch to a different separator character (`Distributed` already uses `#` when reporting the bind address and port back to the caller)?

This affects the `machine_spec` parameter of [`addprocs`](https://docs.julialang.org/en/v1/stdlib/Distributed/#Distributed.addprocs), which can hold a colon-separated port number after both the `host` and `bind_addr` field.

JeffBezanson:
Supporting RFC 2732 would be good :+1:

mgkuhn:
RFC 2732 syntax is definitely the way to go: OpenSSH also uses it now in several command-line options. I have now implemented support for that syntax.

[...]

PR #34430, opened 2020-01-18
#34302 add ipv6 square bracket support to launch_on_machine

ssikdar1:
PR for #34302 to let launch_on_machine handle ipv6 address by adding support for the square bracket notation.

Note: this does not address the ipv6 socket related problems reported in the issue.

- [X] add a function to handle both ipv4 and ipv6 host strings
- [X] Add a few tests. 
- [X] Add a NEWS item. 

Testing: 
I added a unit test julia file. It includes `managers.jl` directly and tests some simple use cases.
`127.0.0.1` , `127.0.0.1:80`, `[2001:db8::1]`, `[2001:db8::1]:443`

I also ran `$ ./julia stdlib/Distributed/test/runtests.jl ` with no issues.

### DIFF
--- a/test/enums.jl
+++ b/test/enums.jl
@@ -10,6 +10,7 @@ using Main.MacroCalls
 
 @test_throws MethodError convert(Enum, 1.0)
 
+@test_throws ArgumentError("invalid type expression for enum 1 + 1") @macrocall(@enum 1 + 1 2)
 @test_throws ArgumentError("no arguments given for Enum Foo") @macrocall(@enum Foo)
 @test_throws ArgumentError("invalid base type for Enum Foo2, Foo2::Float64=::Float64; base type must be an integer primitive type") @macrocall(@enum Foo2::Float64 apple=1.)
 
@@ -89,6 +90,7 @@ end
                                   2
                               end""") @macrocall(@enum Test2  x ? 1 : 2)
 @test_throws ArgumentError("invalid argument for Enum Test22: 1 = 2") @macrocall(@enum Test22 1=2)
+@test_throws ArgumentError("invalid name for Enum TestEnum; \"1 + 1\" is not a valid identifier") var"@enum"(LineNumberNode(@__LINE__), @__MODULE__, :TestEnum, Symbol("1 + 1"))
 
 # other Integer types of enum members
 @enum Test3::UInt8 _one_Test3=0x01 _two_Test3=0x02 _three_Test3=0x03
@@ -167,3 +169,34 @@ end
 @test Int(haggis) == 4
 
 @test (Vector{Fruit}(undef, 3) .= apple) == [apple, apple, apple]
+
+# long, discongruous
+@enum Alphabet begin
+    alphabet_a
+    alphabet_b
+    alphabet_c
+    alphabet_d
+    alphabet_e
+    alphabet_f
+    alphabet_g
+    alphabet_h
+    alphabet_i
+    alphabet_j
+    alphabet_k
+    alphabet_l
+    alphabet_m
+    alphabet_n
+    alphabet_o
+    alphabet_p
+    alphabet_q
+    alphabet_r
+    alphabet_s
+    alphabet_t
+    alphabet_u
+    alphabet_v
+    alphabet_w
+    alphabet_x
+    alphabet_y
+    alphabet_z = 26
+end
+@test alphabet_z == Alphabet(26)


Keno:
Please rebase this branch to only have the relevant commit.

ssikdar1:
@Keno  sorry about that! branch rebased

stdlib/Distributed/src/managers.jl
```diff
@@ -145,7 +145,32 @@ end
 show(io::IO, manager::SSHManager) = println(io, "SSHManager(machines=", manager.machines, ")")
 
 
-function launch_on_machine(manager::SSHManager, machine, cnt, params, launched, launch_ntfy::Condition)
+function parse_machine(machine::AbstractString)
+    hoststr = ``
+    portstr = ``
```
StefanKarpinski: Why return empty command objects here?

ssikdar1: changed to empty strings in adafbd0

StefanKarpinski:
Going a bit further, since you're introducing the `parse_machine` function, I would suggest that it should parse the port number all the way to validated port integer or `nothing` in the parse function rather than right afterwards. I was also going to suggest parsing the host part to an `IPAddr` (`IPv4` or `IPv6` instance) but it could also be a hostname, so there would be a lot of potential return types, which is not actually a big problem, but then it gets immediately turned back into a string and passing through the `IPAddr` type could lead to differences in formatting which might just confuse matters. So probably best to push the port validation into the `parse_machine` function but return the host as a string.

stdlib/Distributed/src/managers.jl
```diff
@@ -161,15 +186,9 @@ function launch_on_machine(manager::SSHManager, machine, cnt, params, launched,
...
-    # if this machine def has a port number, add the port information to the ssh flags
-    if length(machine_def) > 2
-        throw(ArgumentError("invalid machine definition format string: invalid port format \"$machine_def\""))
-    end
-    host = machine_def[1]
+    host, portstr = parse_machine(machine_bind[1])
     portopt = ``
-    if length(machine_def) == 2
-        portstr = machine_def[2]
+    if portstr != ``
```
StefanKarpinski: This should be `isempty(portstr)` now.

stdlib/Distributed/src/managers.jl
```diff
@@ -145,7 +145,32 @@ end
...
+    if length(machine_def) > 2
+        throw(ArgumentError("invalid machine definition format string: invalid port format \"$machine_def\""))
+    end
+
+    if length(machine_def) == 2
+        portstr = machine_def[2]
+    end
+    (hoststr, portstr)
+end
+
```
StefanKarpinski: Drop the extra newlines here.

stdlib/Distributed/src/managers.jl
ssikdar1: Addressed in 87affe1

stdlib/Distributed/src/managers.jl
ssikdar1: deleted newlines in 8b9bbb1 , now theres only one new line between the two functions

stdlib/Distributed/test/managers.jl
```diff
@@ -0,0 +1,20 @@
+using Test
+using Distributed
+using Base: Semaphore
+using Sockets
+
+const BASE_TEST_PATH = joinpath(Sys.BINDIR, "..", "share", "julia", "stdlib", "v1.5", "Distributed", "src")
+@eval Main include(joinpath($(BASE_TEST_PATH), "managers.jl"))
```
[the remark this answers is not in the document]
fredrikekre: You should be able to just import the function directly:
```
using Distributed: parse_machine
```

ssikdar1:
> So probably best to push the port validation into the `parse_machine` function but return the host as a string.

8b9bbb1 : Pushed the port validation to inside `parse_machine` , `parse_machine`  still returns host as a string. Added some unit tests to test the port edge cases

stdlib/Distributed/src/managers.jl
```diff
@@ -145,7 +145,34 @@ end
...
+    end
+
+    if length(machine_def) == 2
+        portstr = machine_def[2]
+        if !all(isdigit, portstr) || (p = parse(Int,portstr); p < 1 || p > 65535)
+            msg = "invalid machine definition format string: invalid port format \"$machine_def\""
+            throw(ArgumentError(msg))
+        end
+    end
+    (hoststr, portstr)
```
StefanKarpinski: Return the actual port number here or `nothing` if there is no port number supplied.

ssikdar1: changed to return the actual parsed port number or nothing and tests updated in aa7a0b6

stdlib/Distributed/src/managers.jl
```diff
@@ -161,20 +196,10 @@ function launch_on_machine(manager::SSHManager, machine, cnt, params, launched,
...
-    if length(machine_def) == 2
-        portstr = machine_def[2]
-        if !all(isdigit, portstr) || (p = parse(Int,portstr); p < 1 || p > 65535)
-            msg = "invalid machine definition format string: invalid port format \"$machine_def\""
-            throw(ArgumentError(msg))
-        end
-        portopt = ` -p $(machine_def[2]) `
+    if portnum != nothing
+        portopt = ` -p $(portnum) `
     end
```
StefanKarpinski: I would write all of this as:
```jl
portopt = portnum === nothing ? `` : `-p $portnum`
```

StefanKarpinski:
One simplifying comment but otherwise looks good!

StefanKarpinski:
Ok. This is looking great. Now all you need is a NEWS entry! It should probably go in the standard library changes section under a "Distributed" subheading.

StefanKarpinski (approved):

StefanKarpinski:
This is good to go when CI passes. Whoever merges, please squash!

stdlib/Distributed/test/managers.jl
```diff
@@ -0,0 +1,19 @@
+using Test
```
Keno: This test doesn't run unless it's added to runtests.jl

Keno (changes requested):

StefanKarpinski:
Did you mean to close this? We're so close to being done here...

ssikdar1:
sorry I had hard reset my fork to the upstream master, I must have closed the issue then by accident

StefanKarpinski:
I can't reopen it either. Guess you'll need to make a new PR.

PR #34431, opened 2020-01-18
Update parallel-computing.md

OkonSamuel:
I got a different result on my system when i ran the following codeblock
```julia-repl
julia> i = Threads.Atomic{Int}(0);

julia> ids = zeros(4);

julia> old_is = zeros(4);

julia> Threads.@threads for id in 1:4
           old_is[id] = Threads.atomic_add!(i, id)
           ids[id] = id
       end
```
The result gotten in the julia manual is
``` julia-repl
julia> old_is
4-element Array{Float64,1}:
 0.0
 1.0
 7.0
 3.0
````
While on my system i got
```julia-repl
julia> ids
4-element Array{Float64,1}:
 0.0
 1.0
 3.0
 6.0
````

### DIFF
--- a/doc/src/manual/parallel-computing.md
+++ b/doc/src/manual/parallel-computing.md
@@ -341,8 +341,8 @@ julia> old_is
 4-element Array{Float64,1}:
  0.0
  1.0
- 7.0
  3.0
+ 6.0
 
 julia> ids
 4-element Array{Float64,1}:


Keno (changes requested):
Either result is fine, the output depends on the thread execution order. Of course that makes it a bit of a confusing example in this place. @kshyatt @vchuravy Looks like you wrote this section. What was the intention behind this example.

vchuravy:
Maybe we should remove that example and use the one afterwards since it is independent of order of execution. I believe that particular one is from @kshyatt so let's wait for her to chime in.

OkonSamuel:
Ok

ViralBShah:
How about just mention for this example that the results will be printed in order of execution, and that it may different every time the program is run?

vtjnash:
I don't think we need any updates, since the point is to make people avoid this unless they understand race conditions. Although I can include changes in https://github.com/JuliaLang/julia/pull/37847 (where I'm already editing this) if something is desired.

PR #34433, opened 2020-01-18
fix #34415, remove affinity setting code

### LINKED ISSUE #34415: Julia seems to ignore process affinity settings on Linux
Julia (tested with v1.0 and v1.3) currently seems to ignore (or escape from) process affinity settings on Linux. This is a problem in HPC computing contexts (but also in other scenarios), e.g. when trying to run one Julia process per NUMA domain.

Let's say we have a machine with 16 virtual cores (8 * 2):

```
$ numactl -s
policy: default
preferred node: current
physcpubind: 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 
[...]
```

Now we use `numactl` to pin a process to CPUs 1 and 2, that process being a `numactl -s` to check that the limitation is actually in effect:

```
$ numactl --physcpubind=1,2 -- numactl -s
policy: default
preferred node: current
physcpubind: 1 2 
[...]
```

Now with `numactl` running `numactl -s` via julia:

```
$ numactl --physcpubind=1,2 -- julia -E 'run(`numactl -s`)'
policy: default
preferred node: current
physcpubind: 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 
[...]
```

Within Julia, the pinning to CPUs 1 and 2 is gone. I see the same when running multi-threaded Julia code (Julia multi-threading or BLAS via Julia): All CPUs are used, instead of just the selected ones.

JeffBezanson:
This was removed before and added back apparently by mistake.

fix #34415

### DIFF
--- a/src/init.c
+++ b/src/init.c
@@ -698,18 +698,6 @@ void _julia_init(JL_IMAGE_SEARCH rel)
     }
 #endif
 
-#if defined(__linux__)
-    int ncores = jl_cpu_threads();
-    if (ncores > 1) {
-        cpu_set_t cpumask;
-        CPU_ZERO(&cpumask);
-        for(int i=0; i < ncores; i++) {
-            CPU_SET(i, &cpumask);
-        }
-        sched_setaffinity(0, sizeof(cpu_set_t), &cpumask);
-    }
-#endif
-
     if ((jl_options.outputo || jl_options.outputbc) &&
         (jl_options.code_coverage || jl_options.malloc_log)) {
         jl_error("cannot generate code-coverage or track allocation information while generating a .o or .bc output file");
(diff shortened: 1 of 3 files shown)


Keno:
Test please :). This already regressed once.

JeffBezanson:
Is it ok to use `taskset` to test this? (Linux only though I assume?)

KristofferC:
Seems like it works. Gonna merge this so I can get it into the 1.4 branch and merge that when CI passes.

mloubout:
Doesn't work on 1.3.1
```
This issue is still happenning

```
$ numactl --physcpubind=1,2 -- julia -E 'run(`numactl -s`)'
policy: default
preferred node: current
physcpubind: 0 1 2 3 4 5 6 7 
cpubind: 0 
nodebind: 0 
membind: 0 
Process(`numactl -s`, ProcessExited(0))
```

Keno:
Looks like 1.4 was the first version with this fix?

mloubout:
It looks fixed in 1.4 indeed thanks

PR #34435, opened 2020-01-19
Improve error message for character literal

### LINKED ISSUE #29982: ERROR: syntax: Invalid character literal is ambiguous
I've been using Julia for a month now and switch between many languages. Two of the other langauges i use quite a bit are SQL and python, languages that support single-quote strings. Obviously i know strings are defined using `"` double-quotes in Julia, but sometimes I slip up and define a string using `'` rather than `"`. It would be nice in these situations to return a slightly more useful error than the current one.

```julia
'this is intended to be string' #< throws ERROR: syntax: Invalid character literal
```

**improvement**
```julia
'this is intended to be string' # ERROR: syntax: char length greater than 1 (or something)
```

0x005c:
Modified the error message emitted when character literal have 2 or more characters.
fixes #29982

### DIFF
--- a/src/julia-parser.scm
+++ b/src/julia-parser.scm
@@ -2264,9 +2264,12 @@
                                                   b))
                                   (loop (read-char (ts:port s))))))
                      (let ((str (tostr #f b)))
-                       (if (= (string-length str) 1)
-                           (string.char str 0)
-                           (error "character literal contains multiple characters")))))))
+                       (let ((len (string-length str)))
+                         (if (= len 1)
+                             (string.char str 0)
+                             (if (= len 0)
+                                 (error "invalid empty character literal")
+                                 (error "character literal contains multiple characters")))))))))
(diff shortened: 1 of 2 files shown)


StefanKarpinski:
This is good but `''` would also report "character literal contains multiple characters" which isn't accurate. So to fix this, check for zero characters and report "invalid empty character literal" separately. You'll also want to write tests to make sure you're doing this correctly. I.e. that `''` throws the "empty literal" error while `'abc'` throws the "multiple characters" error.

0x005c:
@StefanKarpinski Thank you.
I fixed it, and added tests. Could you please check my changes?

StefanKarpinski (approved):
LGTM but @JeffBezanson should take a look.

StefanKarpinski:
Nice work! Thanks for the improvement.

PR #34436, opened 2020-01-19
Add docs for several string functions.

kimttfung:
Added docs for `iterate`, `length`, and `isascii`.


logankilpatrick (approved):
Nice work @fungktt

StefanKarpinski:
Thank you for the contribution! However, all of these specific functionalities are already covered by the docstrings for more generic signatures:
```
help?> length
search: length

  length(collection) -> Integer

  Return the number of elements in the collection.

  Use lastindex to get the last valid index of an indexable collection.

  Examples
  ≡≡≡≡≡≡≡≡≡≡

  julia> length(1:5)
  5

  julia> length([1, 2, 3, 4])
  4

  julia> length([1 2; 3 4])
  4

  ───────────────────────────────────────────────────────────────────────────────────────────────────

  length(A::AbstractArray)

  Return the number of elements in the array, defaults to prod(size(A)).

  Examples
  ≡≡≡≡≡≡≡≡≡≡

  julia> length([1, 2, 3, 4])
  4

  julia> length([1 2; 3 4])
  4

  ───────────────────────────────────────────────────────────────────────────────────────────────────

  length(s::AbstractString) -> Int
  length(s::AbstractString, i::Integer, j::Integer) -> Int

  The number of characters in string s from indices i through j. This is computed as the number of
  code unit indices from i to j which are valid character indices. With only a single string
  argument, this computes the number of characters in the entire string. With i and j arguments it
  computes the number of indices between i and j inclusive that are valid indices in the string s. In
  addition to in-bounds values, i may take the out-of-bounds value ncodeunits(s) + 1 and j may take
  the out-of-bounds value 0.

  See also: isvalid, ncodeunits, lastindex, thisind, nextind, prevind

  Examples
  ≡≡≡≡≡≡≡≡≡≡

  julia> length("jμΛIα")
  5

help?> iterate
search: iterate InteractiveUtils isinteractive Iterators

  iterate(iter [, state]) -> Union{Nothing, Tuple{Any, Any}}

  Advance the iterator to obtain the next element. If no elements remain, nothing should be returned.
  Otherwise, a 2-tuple of the next element and the new iteration state should be returned.

  ───────────────────────────────────────────────────────────────────────────────────────────────────

  iterate(s::AbstractString, i::Integer) -> Union{Tuple{<:AbstractChar, Int}, Nothing}

  Return a tuple of the character in s at index i with the index of the start of the following
  character in s. This is the key method that allows strings to be iterated, yielding a sequences of
  characters. If i is out of bounds in s then a bounds error is raised. The iterate function, as part
  of the iteration protocol may assume that i is the start of a character in s.

  See also: getindex, checkbounds

help?> isascii
search: isascii isabstracttype

  isascii(c::Union{AbstractChar,AbstractString}) -> Bool

  Test whether a character belongs to the ASCII character set, or whether this is true for all
  elements of a string.

  Examples
  ≡≡≡≡≡≡≡≡≡≡

  julia> isascii('a')
  true

  julia> isascii('α')
  false

  julia> isascii("abc")
  true

  julia> isascii("αβγ")
  false
```
Note that they all have entries for `AbstractString` of which `String` is a subtype.

PR #34438, opened 2020-01-19
Document the preferred use of show over display

JeffFessler:
One-line addition to documentation to help prevent future questions like this:
https://discourse.julialang.org/t/what-function-display-does-ijulia-call-at-the-end-of-each-cell/33554

The difference between `show` and `display` can be overlooked by newcomers (like me)...

### DIFF
--- a/base/multimedia.jl
+++ b/base/multimedia.jl
@@ -315,6 +315,8 @@ a `MethodError` if this type is not supported by either the display(s) or by `x`
 variants, one can also supply the "raw" data in the requested MIME type by passing
 `x::AbstractString` (for MIME types with text-based storage, such as text/html or
 application/postscript) or `x::Vector{UInt8}` (for binary MIME types).
+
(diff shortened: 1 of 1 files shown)


base/multimedia.jl
```diff
@@ -315,6 +315,8 @@ a `MethodError` if this type is not supported by either the display(s) or by `x`
 variants, one can also supply the "raw" data in the requested MIME type by passing
 `x::AbstractString` (for MIME types with text-based storage, such as text/html or
 application/postscript) or `x::Vector{UInt8}` (for binary MIME types).
+
+To customize how instances of a type are displayed, overload [`show`](@ref) rather than `display`.
```
stevengj: Maybe add
```
as explained in the manual section on [custom pretty-printing](@id man-custom-pretty-printing).
```

JeffFessler: done, thanks

stevengj (approved):

PR #34439, opened 2020-01-19
Blas hpmv minor updates

iolaszlo:
@dkarrasch


dkarrasch:
Once #34783 is merged, this will need a rebase.

dkarrasch (approved):
LGTM. The `GC.@preserve` change is already included in #34783, so once that is merged and this is rebased, the change in that line is going to disappear from the change log.

Issue #34288, opened 2020-01-06
codegen assertion in BitIntegers


JeffBezanson:
With #34287 running `]test BitIntegers`:
```
julia: /home/jeff/src/julia/src/cgutils.cpp:514: unsigned int convert_struct_offset(llvm::Type*, unsigned int): Assertion `SL->getElementOffset(idx) == byte_offset' failed.

signal (6): Aborted
in expression starting at /home/jeff/.julia/packages/BitIntegers/4QmBA/test/runtests.jl:333
gsignal at /lib/x86_64-linux-gnu/libc.so.6 (unknown line)
abort at /lib/x86_64-linux-gnu/libc.so.6 (unknown line)
unknown function (ip: 0x7f5b6a6b2bf5)
__assert_fail at /lib/x86_64-linux-gnu/libc.so.6 (unknown line)
convert_struct_offset at /home/jeff/src/julia/src/cgutils.cpp:514
convert_struct_offset at /home/jeff/src/julia/src/cgutils.cpp:527 [inlined]
convert_struct_offset at /home/jeff/src/julia/src/cgutils.cpp:520 [inlined]
emit_struct_gep at /home/jeff/src/julia/src/cgutils.cpp:525
emit_getfield_knownidx at /home/jeff/src/julia/src/cgutils.cpp:1672
emit_builtin_call at /home/jeff/src/julia/src/codegen.cpp:2872
```

vtjnash:
Duplicate of https://github.com/JuliaLang/julia/issues/32414

rfourquet:
The fix for the duplicate doesn't seem to fix this issue, so re-opening. Sorry if I messed my testing and got this wrong.

rfourquet:
Minimal repro:
```
using BitIntegers
BitIntegers.@define_integers 24
Base.sub_with_overflow(typemin(UInt24)+UInt24(3), UInt24(3)) == (typemin(UInt24), false)
```

rfourquet:
The minimal repro above seems to have been fixed in 1.6. But I found another "Internal error" in today's master (d6c092d6d7):
```julia
julia> using BitIntegers

julia> BitIntegers.@define_integers 24; x = UInt24(0);

julia> x
Internal error: encountered unexpected error in runtime:
MethodError(f=Main.UInt24, args=(0,), world=0x0000000000001133)
jl_method_error_bare at /local/src/julia/src/gf.c:1818
jl_method_error at /local/src/julia/src/gf.c:1836
[...]

PR #34452, opened 2020-01-20
[release-1.4] Set VERSION to 1.4.0-rc1

ararslan:


fredrikekre (approved):

KristofferC (approved):

rfourquet:
#32414 is on the 1.4 milestone, is it planned to be fixed after RC1 ?

KristofferC:
Maybe? I don't understand why it is on the milestone though.

rfourquet:
> I don't understand why it is on the milestone though.

Probably because it was identified as the cause of the failure of BitIntegers (#34288) (I already fixed the other problem with this package on 1.4 related to `signed`, mentioned there: https://github.com/JuliaLang/julia/pull/34238#issuecomment-571297077)

KristofferC:
Previously, it just gave garbage though, right? We just added the assert to catch it in assert builds.

rfourquet:
> Previously, it just gave garbage though, right?

After investigation, you seem to be right: I was testing `@test f(x) == y`, `f(x)` was actually not equal to `y` but the `==` was still returning `true`, now it segfaults because of the assertion. I guess this still should not segfault though.

KristofferC (changes requested):
Let's get https://github.com/JuliaLang/julia/pull/34461 in as well

KristofferC:
> I guess this still should not segfault though.

Ideally, all bugs should be fixed but this is at least an improvement over status quo?

KristofferC (approved):
Should be good to go now

ararslan:
CI seems not particularly happy

KristofferC:
All green now.

Issue #13294, opened 2015-09-24
Profiler can deadlock on OS X


Keno:
If we happen to stop a thread while it is trying to register a dynamic fde, the profiler thread will dead lock trying to acquire a lock held by the blocked thread.

vtjnash:
I started some work to address this on my threading branch

Keno:
Can we move that work onto master?

yuyichao:
@vtjnash Is this fixed? (Or is this still not due to one of your todo's in `signals-mach.c`?)

vtjnash:
no, i think keno wanted us to add a guard around calls to `__register_frame` that stops us from trying to compute a backtrace then

PR #34454, opened 2020-01-21
Fix deadlock in windows profiler

Keno:
The deadlock here happens when the main thread is trying to register
a new JIT object while the profiler is triggered. The main thread
will acquire the objectmap lock and then get suspended. Then the profiler
thread will attempt to stackwalk, but in order to do that effectively,
it needs to look at the object map to find the unwind info for the function
on the stack. We don't have the same problem on linux because the main
thread runs the backtracing. On linux and mac the registration process
works differently and on linux at least the backtracing happens on a
different thread so we may not have this problem. However, #13294 says
that it is also a problem on OS X. We should keep an eye out for it.
For the moment, just try to fix this by terminating the stack unwind
when we fail to acquire the lock. That's not ideal, because it reduces
the quality of the profiler info, but only in situations where we would
previous deadlock. This entire code needs a rewrite, but for now, I'm
just hoping to get CI to stopping deadlocking on us.

### DIFF
--- a/src/debuginfo.cpp
+++ b/src/debuginfo.cpp
@@ -432,6 +432,14 @@ public:
         uv_rwlock_rdlock(&threadsafe);
         return objectmap;
     }
+
+    Optional<std::map<size_t, ObjectInfo, revcomp>*> trygetObjectMap()
+    {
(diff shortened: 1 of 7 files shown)


Keno:
F'ing hell. How many of these do we have?

Issue #34455, opened 2020-01-21
Parallelize LinearAlgebra/addmul test


Keno:
This test does a big loop over various combinations of matrix types: https://github.com/JuliaLang/julia/blob/master/stdlib/LinearAlgebra/test/addmul.jl#L134 and tries to add them together. This test takes the longest of any of the tests in the system. That loop should be fairly easy to distribute across multiple workers. Since each test runs with a different argument combination, I don't think we'd even lose anything by not caching common specializations between workers.

tkf:
I think we can get even more speedup by launching the workers with `--compile=min` flag.

Keno:
Ah, neat idea. We can't do that in general, but it would make sense if we're launching the workers specifically for this test.

andreasnoack:
> I think we can get even more speedup by launching the workers with --compile=min flag.

If this become possible, we could probably reduce CI time for LinearAlgebra by a lot. E.g. almost all of the time in `triangular.jl` is spent on compilation.

ViralBShah:
Do we still need to do something here?

PR #34456, opened 2020-01-21
Shuffle LinearAlgebra tests to the front of the test queue

Keno:
They take much longer than the rest of the tests, so if they're
run at the end, they often determine the length of the entire test
run. Instead, try running them at the start while there's still
plenty of work to be done. Hopefully this will reduce overall duration
of CI tests. If this doesn't work out as planned, we can try a more
fine grained strategy to load balancing (e.g. by checking in a table
of approximate runtimes and sorting according to that).

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -63,6 +63,13 @@ move_to_node1("stress")
 # since it starts a lot of workers and can easily exceed the maximum memory
 limited_worker_rss && move_to_node1("Distributed")
 
+# Shuffle LinearAlgebra tests to the front, because they take a while, so we might
+# as well get them all started early.
+linalg_test_ids = findall(x->occursin("LinearAlgebra", x), tests)
+linalg_tests = tests[linalg_test_ids]
+deleteat!(tests, linalg_test_ids)
+prepend!(tests, linalg_tests)
+
 import LinearAlgebra
 cd(@__DIR__) do
     n = 1


tkf:
Sorry that this test dominates the CI time (I wrote this)... It is a super brute force test but it's really useful for catching bugs.

FYI, you can reduce this number to make it shorter, but at the cost of the bug to be discovered much later than the actual PR touching the relevant code

https://github.com/JuliaLang/julia/blob/159de94dc812e4d1d89d1badf0c0d19b5f21d387/stdlib/LinearAlgebra/test/addmul.jl#L122-L123

Keno:
It's not just addmul, some of the others also take a bunch of time.

tkf:
Ah, OK. I mixed it with #34455.

Keno:
I mean, they're related, but if this works (maybe while simultaneously bumping the number of cores allotted to testing to make sure that all the LinearAlgebra tests start at one, then the other one is less urgent.

JeffBezanson:
Looks like there are ambiguities in Furlongs.jl that are test-order-dependent (only errors if Furlongs is loaded before the ambiguity tests). I'll try to fix them.

Keno:
This made the aarch64 bot 10 minutes faster. I think it's worth doing once the ambiguity error is fixed.

JeffBezanson:
This can be rebased and merged now.

dkarrasch:
Could this be rebased to see whether it resolves the time-outs we see frequently?

mbauman:
We used to do this, but I guess we lost it in the stdlib shakeup.

PR #34467, opened 2020-01-21
Updated the documentation for Range function

### LINKED ISSUE #33798: Possible documentation error: range function docstring at julia/base/range.jl
The second paragraph states

> If `length` and `stop` are provided and `step` is not, the step size will be computed
automatically such that there are `length` linearly spaced elements in the range (a [`LinRange`](@ref)).

but in this setting I get the following
```
julia> typeof(range(1, length=10,stop=100))
```
A `StepRangeLen` instead of a `LinRange`.

kamalojasv181:
Closes #33798

### DIFF
--- a/Make.inc
+++ b/Make.inc
@@ -1408,3 +1408,4 @@ print-%:
 # Literal values that are hard to use in Makefiles otherwise:
 COMMA:=,
 SPACE:=$(eval) $(eval)
+USE_BINARYBUILDER_LLVM=0
--- a/base/range.jl
+++ b/base/range.jl
@@ -53,11 +53,16 @@ Given a starting value, construct a range either by length or from `start` to `s
 optionally with a given step (defaults to 1, a [`UnitRange`](@ref)).
 One of `length` or `stop` is required.  If `length`, `stop`, and `step` are all specified, they must agree.
 
-If `length` and `stop` are provided and `step` is not, the step size will be computed
-automatically such that there are `length` linearly spaced elements in the range (a [`LinRange`](@ref)).
+In most cases if `step` and `stop` are provided and `length` is not, the overall range length will be computed automatically such that the elements are `step` spaced (a [`StepRange`](@ref)).
 
-If `step` and `stop` are provided and `length` is not, the overall range length will be computed
-automatically such that the elements are `step` spaced (a [`StepRange`](@ref)).
+In case the elements are not simple numeric, elements are spaced using (a [`LinRange`](@ref))
+# Example
+
+```jldoctest
+julia> range([1,2], [3,4], length=10)
+10-element LinRange{Array{Int64,1}}:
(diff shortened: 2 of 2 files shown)


Make.inc
```diff
@@ -1408,3 +1408,4 @@ print-%:
 # Literal values that are hard to use in Makefiles otherwise:
 COMMA:=,
 SPACE:=$(eval) $(eval)
+USE_BINARYBUILDER_LLVM=0
```
StefanKarpinski: Revert this change.

base/range.jl
```diff
@@ -53,11 +53,16 @@ Given a starting value, construct a range either by length or from `start` to `s
 optionally with a given step (defaults to 1, a [`UnitRange`](@ref)).
 One of `length` or `stop` is required.  If `length`, `stop`, and `step` are all specified, they must agree.
 
-If `length` and `stop` are provided and `step` is not, the step size will be computed
-automatically such that there are `length` linearly spaced elements in the range (a [`LinRange`](@ref)).
+In most cases if `step` and `stop` are provided and `length` is not, the overall range length will be computed automatically such that the elements are `step` spaced (a [`StepRange`](@ref)).
```
```suggestion
In most cases if `step` and `stop` are provided and `length` is not, the overall range length will
be computed automatically such that the elements are `step` spaced (a [`StepRange`](@ref)).
```

StefanKarpinski (changes requested):
A couple of changes required.

base/range.jl
```diff
@@ -53,11 +53,16 @@ Given a starting value, construct a range either by length or from `start` to `s
...
 
-If `step` and `stop` are provided and `length` is not, the overall range length will be computed
-automatically such that the elements are `step` spaced (a [`StepRange`](@ref)).
+In case the elements are not simple numeric, elements are spaced using (a [`LinRange`](@ref))
+# Example
```
dkarrasch: The last sentence should probably be finished with a period, and also the brackets look strange. Between the main docstring and the `# Example` header you could also introduce an empty line, though I'm not sure that makes a difference in the end.

base/range.jl
stevengj: The original documentation was describing the case with both `length` and `stop`, and we should continue to document that case:
```
If numeric `length` and `stop` values provided and `step` is not, the return value is a [`StepRangeLen`](@ref)
with a computed step.  However, for non-numeric values of `start` and `stop` (such as arrays), the `range` function instead returns a  [`LinRange`](@ref) object.  (Both `StepRangeLen` and `LinRange` correspond to equally spaced points up to roundoff errors, but deal with floating-point rounding slightly differently.)
```

kamalojasv181:
https://github.com/JuliaLang/julia/pull/34475 created a new PR

stevengj:
@kamalojasv181, in the future, please don't create multiple PRs for the same issue.  Just update your existing PR.

PR #34469, opened 2020-01-21
another soft scope PkgEval run

JeffBezanson:
This time excluding `let` (and testsets).

@nanosoldier `runtests(ALL, vs = ":master")`

PR #34472, opened 2020-01-21
Distributed, SHA tests: fix soft scope shadowing

StefanKarpinski:
Fix the "soft scope violations" found in https://github.com/JuliaLang/julia/pull/34469. Good change regardless.


KristofferC:
Isn't this the wrong way of fixing it? In other words, shouldn't the usage of e.g. the global `npids` rather be eliminated or "constrained"?

StefanKarpinski:
This was the minimal change to eliminate the warnings/errors. IMO the "right" way to fix this would be to change these test files to use `@testset`. But that seemed like a lot of unnecessary noise for the moment.

PR #34475, opened 2020-01-22
Revised Range Documentation

### LINKED ISSUE #33798: Possible documentation error: range function docstring at julia/base/range.jl

kamalojasv181:
Closes #33798

### DIFF
--- a/base/range.jl
+++ b/base/range.jl
@@ -53,9 +53,13 @@ Given a starting value, construct a range either by length or from `start` to `s
 optionally with a given step (defaults to 1, a [`UnitRange`](@ref)).
 One of `length` or `stop` is required.  If `length`, `stop`, and `step` are all specified, they must agree.
 
-In most cases if `step` and `stop` are provided and `length` is not, the overall range length will be computed automatically such that the elements are `step` spaced (a [`StepRange`](@ref)).
+If numeric `length` and `stop` values are provided but `step` is not specified, the return value is
+a [`StepRangeLen`](@ref) with a computed step.  However, for non-numeric values of
+`start` and `stop` (such as arrays), the `range` function instead returns a
+[`LinRange`](@ref) object.  (Both `StepRangeLen` and `LinRange` correspond
+to equally spaced points up to roundoff errors, but deal with floating-point rounding
+slightly differently.)
 
-In case the elements are not simple numeric, elements are spaced using (a [`LinRange`](@ref))
 # Example
 
 ```jldoctest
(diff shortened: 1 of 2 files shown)


base/range.jl
```diff
@@ -53,11 +53,20 @@ Given a starting value, construct a range either by length or from `start` to `s
 optionally with a given step (defaults to 1, a [`UnitRange`](@ref)).
 One of `length` or `stop` is required.  If `length`, `stop`, and `step` are all specified, they must agree.
 
-If `length` and `stop` are provided and `step` is not, the step size will be computed
-automatically such that there are `length` linearly spaced elements in the range (a [`LinRange`](@ref)).
```
timholy: To me it would be clearer to keep the

> the step size will be computed automatically such that there are `length` linearly spaced elements in the range

part because it's important and applies to everything. Then say something like

```julia
"""
...in a range. The returned type depends on the inputs, a [`StepRangeLen`](@ref) for floating-point inputs and [`LinRange`](@ref) otherwise.
"""
``` 

because that's effectively a detail that most people won't need to worry about. (Since the docs for `StepRangeLen` mention error-free arithmetic, I think we're covered as far as "purpose" goes.)

timholy:
Yes, deleting the info about return type seems like an OK solution to the problem too. Any other thoughts?

kamalojasv181:
Oh I thought you meant doing that in the first place. Anyways I think it would be great to do so for now.

timholy:
Well, there don't seem to be any objections to just deleting the info about the return type. If you can clean up this branch with the extraneous stuff then I think we could merge it.

kamalojasv181:
Sorry for the trouble. Now we are back at where you want. It is ready to be merged.

base/range.jl
```diff
@@ -54,10 +54,7 @@ optionally with a given step (defaults to 1, a [`UnitRange`](@ref)).
 One of `length` or `stop` is required.  If `length`, `stop`, and `step` are all specified, they must agree.
 
 If `length` and `stop` are provided and `step` is not, the step size will be computed
-automatically such that there are `length` linearly spaced elements in the range (a [`LinRange`](@ref)).
-
-If `step` and `stop` are provided and `length` is not, the overall range length will be computed
```
timholy: Hmm, maybe I didn't notice this before or maybe it changed, but now you've deleted explanation of what happens if you don't provide `length`. This does need to be explained...

kamalojasv181: Yes, It was me who did not add it earlier as well. Edited it now. Thanks

timholy:
Thanks @kamalojasv181!

PR #34480, opened 2020-01-22
fix method ambiguities in testhelpers/Furlongs.jl

JeffBezanson:
This fixes failures in the ambiguities test if Furlongs.jl happens to be loaded before it. I see that these constructors were designed to only allow making one Furlong from another if the exponent `p` matches, but that leaves a bunch of method space un-covered, making the ambiguities harder to resolve. Hopefully that does not matter so much.

### DIFF
--- a/test/testhelpers/Furlongs.jl
+++ b/test/testhelpers/Furlongs.jl
@@ -11,12 +11,13 @@ export Furlong
 struct Furlong{p,T<:Number} <: Number
     val::T
     Furlong{p,T}(v::Number) where {p,T} = new(v)
-    Furlong{p,T}(x::Furlong{p}) where {p,T} = new(x.val)
 end
 Furlong(x::T) where {T<:Number} = Furlong{1,T}(x)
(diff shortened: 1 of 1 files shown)


timholy (approved):

Keno:
> Hopefully that does not matter so much.

Maybe at least check it with an assertion?

timholy:
My rule on these things is that `contruction ≠ conversion`. I expect the error from `convert(T, x)` because conversion implies equivalence, but not with construction. After all, you can only build a furlong with `Furlong{p}(3.0)`, and `3.0` is definitely not equivalent to `3.0Furlongs`. (If they behaved the same you couldn't even construct one unless you had some pre-existing furlongs lying around somewhere.)

Keno:
Alright, I don't really care too much for this particular instance, so maybe let's just get this merged, but I still don't really like that method. If this were a real package, I'd complain ;).

JeffBezanson:
Added an assert.

PR #34483, opened 2020-01-23
Merge pull request #34483 from JuliaLang/sf/python_search

staticfloat:
Distributions, more and more, are starting to drop `python2` completely as it's officially EOL'ed now.  Some distributions don't like that `python` can mean either `python2` or `python3`, so they provide only `python3` by default.  For things like `normalize_triplet.py`, we don't really care what version of python is being used, we just want any of them.  So we set `PYTHON` as the result of looking around a bit for the best choice.


KristofferC:
Does this need to be backported to the 1.4 branch? (cf https://build.julialang.org/#/builders/64/builds/8940).

staticfloat:
Yes, I think this is good for backporting. I've also manually patched the environments to have a `python -> python3` symlink, so you can just re-run those jobs for now (which I have done for you [here](https://build.julialang.org/#/builders/64/builds/8945)).

KristofferC:
Can skip the merge commit for PRs like this.

staticfloat:
You mean by using "rebase" when I merge?

StefanKarpinski:
When there's only one commit, just squash it and the result is no merge commit.

Issue #33418, opened 2019-09-30
Provide a clean way to add custom LogLevels


ufechner7:
On discourse a workaround for this feature was provided, but it is not future proof, uses type piracy and cannot be used well in any module. See: xxx
My feature request: Add a function
```
register_loglevel(<LogLevel>, <msg>)
```
Example usage:
```
using Logging
const Trace = LogLevel(500)
register_loglevel(Trace, "Trace")
```

ufechner7:
Two questions:
1. Does it make sense to add this feature to the module Logging or is there a better place?
2. Is my proposed Syntax possible?

Seelengrab:
I don't think much of a change is needed, actually 🤔 

Changing the `LogLevel` [struct](https://github.com/JuliaLang/julia/blob/d0365753fb57f57bcea097c05de8c07aef706561/base/logging.jl#L99-L101) from 

```julia
struct LogLevel
    level::Int32
end
```

to 

```julia
struct LogLevel
    level::Int32
    name::String
end
```

would allow the `show` function to print the user defined name instead of the default `LogLevel(level)`. 

As for the macro, `@logmsg Trace msg` seems sufficient to me, as it would also print the correct name with the change to the struct and be compatible with other custom loggers. I'm not sure macro generation is a good idea, since the core macro already exists.

ufechner7:
Would that be a braking change or could it be added to Julia 1.4?

Seelengrab:
I'm not a core developer, but I think changing the printing as I proposed with a fallback constructor to `LogLevel(level, "LogLevel($level")` for `LogLevel(level)` shouldn't be breaking -- everything visible from the outside should still work as it used to.

ufechner7:
Now we just need a pull request...

c42f:
Good feature request!  I agree with @Seelengrab's assessment: there's some relatively minor fairly backward compatible changes which could make this work better, and having a macro frontend is orthogonal to the internal changes required to `CoreLogging`.

I'm not entirely sure that changing `LogLevel` itself is the right direction; rather I think we could allow custom user-defined types as log levels.

> The real question is if it should be incorporated into a general logging rework

Yes, I do have some thoughts about this. In particular it would be nice to resolve the following design problems:

1. What are log levels really, and what is the interface that logging backends should expect from them?
2. How can logging backends be composable such that they can "downgrade" the importance of messages coming from specific library, without loosing the original log level label?
[...]

PR #34485, opened 2020-01-23
Fix a bug in the `Base.CoreLogging.env_override_minlevel` function

### LINKED ISSUE #34484: JULIA_DEBUG=logging doesn't work
The example in https://github.com/JuliaLang/julia/blob/c819cde083a8161aebd7bec626527dbec47fed44/stdlib/Logging/docs/src/index.md#environment-variables doesn't work (using `JULIA_DEBUG=all` does work).

It would be good to document what values are valid, and also add it to https://github.com/JuliaLang/julia/blob/master/doc/src/manual/environment-variables.md

DilumAluthge:
Fixes #34484

### DIFF
--- a/base/logging.jl
+++ b/base/logging.jl
@@ -441,7 +441,7 @@ global function env_override_minlevel(group, _module)
     if !(:all in _debug_groups_exclude) && (:all in _debug_groups_include || !isempty(_debug_groups_exclude))
         if isempty(_debug_groups_exclude)
             return true
-        elseif isa(group, Symbol) && group in _debug_groups_exclude
+        elseif (isa(group, Symbol) && group in _debug_groups_exclude) || (isa(group, AbstractString) && Symbol(group) in _debug_groups_exclude)
             return false
         elseif isa(_module, Module) && (nameof(_module) in _debug_groups_exclude || nameof(Base.moduleroot(_module)) in _debug_groups_exclude)
             return false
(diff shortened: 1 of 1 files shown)


tkf:
I think it's clear that `_group` expects a symbol

https://github.com/JuliaLang/julia/blob/07a16d695feeba66db2a44c640675e7974ae870b/base/logging.jl#L163-L164

Perhaps fix the macro instead?

https://github.com/JuliaLang/julia/blob/07a16d695feeba66db2a44c640675e7974ae870b/base/logging.jl#L294-L297

DilumAluthge:
Unfortunately I do not know how to fix the macro :(

Want to take a stab at it?

bramtayl:
```
QuoteNode(splitext(basename(file))[1])
```

This line looks suspect: I'm pretty sure it should be

```
QuoteNode(Symbol(splitext(basename(file))[1]))
```

bramtayl:
Might be nice to have a function

```
default_group(file) = Symbol(splitext(basename(file))[1])
```
because that code appears twice

c42f:
Yes it looks like @bramtayl is correct: this is quite an old bug (see b0075cb). The correct fix would indeed be to use `QuoteNode(Symbol(splitext(...)))` so that it's interpolated into the AST as a symbol literal rather than a string and undo the other changes.

(Side note: Interestingly, this brings into question how many people are really using this default value for something important, and suggests we might be able to change it in the future without too much disruption (cf. discussion at https://github.com/JuliaLang/julia/issues/33418#issuecomment-545405073).)

tkf:
I opened #34509

DilumAluthge:
Closing in favor of #34509 :)

PR #34499, opened 2020-01-24
Doc: broadcast fusion is not a scope construct…

mbauman:
…but generators are.


ararslan (approved):
Incredible branch name

JeffBezanson:
Will be fixed by #33864

PR #34502, opened 2020-01-24
add some missing timeouts in Distributed

### LINKED ISSUE #34486: ClusterManager hangs when worker dies after connection is made but before setup is done
ClusterManager [waits](https://github.com/JuliaLang/julia/blob/master/stdlib/Distributed/src/cluster.jl#L658) for a JoinCompleteMsg while holding a [worker_lock](https://github.com/JuliaLang/julia/blob/master/stdlib/Distributed/src/cluster.jl#L439). If the worker dies in the middle of the setup the JoinCompleteMsg is never received. Then the wait is never notified, the lock is not released, and the whole cluster is not able to add new workers any more.

Not sure how to test or fix it properly. As a workaround: instead of plain `wait(rr_ntfy_join)` have a wait with a timeout:
```
timeout = Ref(false);
@async (sleep(10); timeout[] = true; notify(rr_ntfy_join))
wait(rr_ntfy_join)
if timeout[]
     error("timeout waiting for JoinCompleMsg from worker")
end
```

JeffBezanson:
This will hopefully fix some of the intermittent hangs in mac CI. @Keno believes we are running out of file descriptors there, which I agree is a likely cause. That was causing an exception at an unexpected point, leaving some processes waiting forever. This should turn it into a hard fail. Then we just need to stop leaking descriptors, or get more :)

fixes #34486 (by converting the hang into an exception)

### DIFF
--- a/stdlib/Distributed/src/cluster.jl
+++ b/stdlib/Distributed/src/cluster.jl
@@ -569,6 +569,7 @@ end
 function create_worker(manager, wconfig)
     # only node 1 can add new nodes, since nobody else has the full list of address:port
     @assert LPROC.id == 1
+    timeout = worker_timeout()
 
     # initiate a connect. Does not wait for connection completion in case of TCP.
     w = Worker()
@@ -633,9 +634,14 @@ function create_worker(manager, wconfig)
             (notnothing(x.config.ident) in something(wconfig.connect_idents, []))
 
         wlist = filter(filterfunc, PGRP.workers)
+        waittime = 0
         while wconfig.connect_idents !== nothing &&
               length(wlist) < length(wconfig.connect_idents)
+            if waittime >= timeout
+                error("peer workers did not connect within $timeout seconds")
+            end
             sleep(1.0)
+            waittime += 1
             wlist = filter(filterfunc, PGRP.workers)
         end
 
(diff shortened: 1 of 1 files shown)


StefanKarpinski:
When we run out of file descriptors, I wonder if it would be too clever to try doing a full gc and then trying again...

JeffBezanson:
That sounds pretty good to me actually; potentially many places might need to be modified to do the error check and retry though. We could start by trying that in the specific places in Distributed that are failing.

stdlib/Distributed/src/cluster.jl
```diff
@@ -655,7 +661,13 @@ function create_worker(manager, wconfig)
     send_msg_now(w, MsgHeader(RRID(0,0), ntfy_oid), join_message)
 
     @async manage(w.manager, w.id, w.config, :register)
+    # wait for rr_ntfy_join with timeout
+    timedout = false
+    @async (sleep(worker_timeout()); timedout = true; put!(rr_ntfy_join, 1))
```
vtjnash:
```suggestion
    @async (sleep($timeout); timedout = true; put!(rr_ntfy_join, 1))
```

vtjnash (approved):

PR #34503, opened 2020-01-24
fix bug in printing of block expressions in ref

simeonschaub:
This one really caught me by surprise while debugging another part of `show_unquoted`


simeonschaub:
Seems like this was introduced by #33946

stevengj (approved):

PR #34504, opened 2020-01-24
fix printing of LineNumberNode in block within ref

simeonschaub:
This was brought up [on discourse](https://discourse.julialang.org/t/julia-v1-4-0-rc1-is-now-available/33715/13?u=simeonschaub). #34503 is needed for the tests to pass, so this currently also contains these commits.


JeffBezanson:
Thanks!

I think `@weak_test_repr` will work here.

simeonschaub:
You're right, probably not necessary here.

PR #34505, opened 2020-01-24
allow @foo{...} macro calls

### LINKED ISSUE #34498: allow @foo{...} macro calls
I noticed that you cannot invoke a macro as `@foo{...}` — you have to do `@foo {...}`.   **Can we remove the requirement of a space before the brace?** This would be useful for macros to construct parameterized types, for example.

*Background:* In [a recent discourse discussion](https://discourse.julialang.org/t/anonymous-nested-struct/33723/9?u=stevengj), I was looking at implementing a macro `@NamedTuple` that allows you to write `@NamedTuple{raw::Vector{Float64}, value::Int}` as a synonym for `NamedTuple{(:raw, :value),Tuple{Array{Float64,1},Int64}}`, via the implementation:
```
macro NamedTuple(ex)
    Meta.isexpr(ex, :braces) || error("@NamedTuple expects {...}")
    all(e -> Meta.isexpr(e, :(::)), ex.args) || error("@NamedTuple a sequence of name::type expressions")
    vars = [QuoteNode(e.args[1]) for e in ex.args]
    types = [e.args[2] for e in ex.args]
    return :(NamedTuple{($(vars...),), Tuple{$(types...)}})
end
```
This works with `@NamedTuple {raw::Vector{Float64}, value::Int}`, but omitting the space gives
```jl
julia> @NamedTuple{raw::Vector{Float64}, value::Int}
ERROR: syntax: invalid macro usage "@(NamedTuple{raw::Vector{Float64}, value::Int})"
```
Since this is currently a syntax error (that is, parsing fails), supporting it would be non-breaking.

stevengj:
Closes #34498.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,7 +3,7 @@ Julia v1.5 Release Notes
 
 New language features
 ---------------------
-
+* Macro calls `@foo {...}` can now also be written `@foo{...}` (without the space) ([#34498]).
 
 Language changes
 ----------------
--- a/src/julia-parser.scm
+++ b/src/julia-parser.scm
@@ -2397,6 +2397,10 @@
          `(macrocall ,(macroify-name (cadr call))
                      ,startloc
                      ,@(cddr call)))
+        ((and (pair? call) (eq? (car call) 'curly))
+         `(macrocall ,(macroify-name (cadr call))
+                     ,startloc
+                     (braces ,@(cddr call))))
         ((and (pair? call) (eq? (car call) 'do))
          `(do ,(macroify-call s (cadr call) startloc) ,(caddr call)))
         (else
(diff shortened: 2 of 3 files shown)


stevengj:
Unrelated windows temp-file errors in CI.

stevengj:
Hmm, I'm getting
```jl
julia> :(@foo{bar}(baz))
ERROR: syntax: invalid macro usage "@(foo{bar})"
```
I think we probably want `@foo{bar}(baz)` to parse as `(@foo{bar})(baz)`.  I'll look into it.

JeffBezanson:
This still doesn't seem to be fully general, e.g. `@m{a}{b}` doesn't work. I think instead `parse-call-chain` could be modified ala 6c4e30e3677c404c1cbec8086acd49ed9152a786, which made `@a[i]` work.

stevengj:
What should `@m{a}{b}` parse to?

stevengj:
As @JeffBezanson, suggested, I moved the implementation to the `parse-call-chain` function, which indeed seems to make it much simpler and more general.

PR #34418, opened 2020-01-17
Update to LLVM 9

vchuravy:
We are still missing a patch [1] which I rebased recently,
but let's see how CI takes to this.

### TODO:
- [x] patch for D66657
- [x] patch for D71495
- [x] patch for D50010

[1] https://reviews.llvm.org/D50010

PR #34506, opened 2020-01-24
Move `realpath()` earlier in `cmdlineargs` test

staticfloat:
This should hopefully fix issues such as https://github.com/JuliaLang/julia/pull/34418#issuecomment-578275655

### DIFF
--- a/stdlib/LibGit2/test/libgit2.jl
+++ b/stdlib/LibGit2/test/libgit2.jl
@@ -600,6 +600,7 @@ end
 end
 
 mktempdir() do dir
+    dir = realpath(dir)
     # test parameters
     repo_url = "https://github.com/JuliaLang/Example.jl"
     cache_repo = joinpath(dir, "Example")
--- a/stdlib/REPL/test/replcompletions.jl
+++ b/stdlib/REPL/test/replcompletions.jl
@@ -804,7 +804,7 @@ end
 if Sys.iswindows()
     tmp = tempname()
     touch(tmp)
-    path = dirname(tmp)
+    path = realpath(dirname(tmp))
     file = basename(tmp)
     temp_name = basename(path)
     cd(path) do
--- a/test/cmdlineargs.jl
+++ b/test/cmdlineargs.jl
@@ -416,6 +416,9 @@ let exename = `$(Base.julia_cmd()) --startup-file=no`
 
     # test the program name remains constant
     mktempdir() do dir
+        # dir can be case-incorrect sometimes
+        dir = realpath(dir)
+
         a = joinpath(dir, "a.jl")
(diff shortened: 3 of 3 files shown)


stdlib/LibGit2/test/libgit2.jl
```diff
@@ -600,6 +600,7 @@ end
 end
 
 mktempdir() do dir
+    dir = realpath(dir)
```
stevengj: If this is needed on Windows, shouldn’t it be part of the mktempdir implementation?

staticfloat: I think probably yes.

stevengj:
I'm still confused about the real source of the bug here.  Why do we need the `realpath` result?

Currently, `mktempdir()`, `mktemp()`, and `tempname()` all use `tempdir()` as their default parent directly.  The `tempdir()` function gets its path from a user environment variable (`TMP` etc.), and hence it may be a path that does not yet exist or a path spelled with the "wrong" case (on Windows or Mac).

I just checked, and none of `mktempdir()`, `mktemp()`, or `tempname()` currently return the `realpath` result on Windows — all of them preserve the case that was passed (and on my system return the "short" 8.3 path since `tempdir()` is a short path).   Similarly, on MacOS (also case-insensitive, case-preserving), if I set `ENV["TMPDIR"]=uppercase(tempdir())` then all of these functions return the uppercase path.

I see three options:

1. Continue to do what we do now — preserve the user environment's specified path as-is in all returned paths.   Any caller that needs the `realpath` (such as these tests, apparently) should call this manually.

2. Change `tempdir()` to return the `realpath()` if the path exists.  This seems potentially dangerous, however — if they current path is a symbolic link, but the symbolic link is changed to a different target after the `tempdir()` call

3. Keep `tempdir()` as-is, but change `mktemp` etcetera to return the `realpath`, which should be safer once the file is created.  (If someone renames the file or rearranges symbolic links after the `mktemp` call, we already have potential problems when you try to delete the file.)

KristofferC:
Bump, all Windows CI currently fails so it would be good to do something here.

stevengj:
I think it's fine to merge this for now and debate whether we want to call `realpath` by default for `mktemp` etc. later.

vtjnash:
These tests are probably demonstrating bad practice and should be using `samefile` instead of `==`, but they're only tests.

staticfloat:
I think Jameson is correct; I think the reason the `realpath()` is needed has more to do with the tests trying to ensure that two things are the same file than that `mktempdir()` is doing something fundamentally wrong.  That being said, it does seem reasonable to me that if a standard library function returns you a pathname to a file, it would be the most "canonical" way to refer to that file, which `realpath()` seems to be to me.

PR #34509, opened 2020-01-25
Always use Symbol as _group value of log records

### LINKED ISSUE #34484: JULIA_DEBUG=logging doesn't work

tkf:
fix #34484
close #34485

### DIFF
--- a/base/logging.jl
+++ b/base/logging.jl
@@ -248,6 +248,8 @@ function log_record_id(_module, level, message, log_kws)
     end
 end
 
+default_group(file) = Symbol(splitext(basename(file))[1])
+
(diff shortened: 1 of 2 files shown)


c42f:
Looks great to me :+1:

PR #34512, opened 2020-01-25
Fix condskeel definition

saschatimme:
Skeel's condition number is a relative condition number, however the definition of `condskeel(A,x)` missed to divide by the norm of `x`. The definition is for example in Higham "Accuracy and stability of numerical algorithms" eq. (7.13) (p. 123).


dkarrasch (approved):
Nice catch! Is this tested somewhere? Maybe you could add a test to make sure we have the right scaling, something like `@test condskeel(A, x) <= condskeel(A)` for some random `A` and `x = 3randn(10)`?

saschatimme:
So far this was never tested directly. I added a slightly different test to make sure that the scaling is applied. Let me know if I should also add your suggested one.

PR #34520, opened 2020-01-26
Missing bool tests

bramtayl:
[Uncovered](https://codecov.io/gh/JuliaLang/julia/src/master/base/bool.jl#L6)
[Uncovered](https://codecov.io/gh/JuliaLang/julia/src/master/base/bool.jl#L7)
[Uncovered](https://codecov.io/gh/JuliaLang/julia/src/master/base/bool.jl#L83)

### DIFF
--- a/test/numbers.jl
+++ b/test/numbers.jl
@@ -64,10 +64,10 @@ end
     @test iszero(false) && !iszero(true)
     @test isone(true) && !isone(false)
 
-    @test !typemin(Bool)
-    @test typemax(Bool)
-    @test !abs(false)
-    @test abs(true)
+    @test typemin(Bool) == false
+    @test typemax(Bool) == true
+    @test abs(false) == false
+    @test abs(true) == true
 end
 @testset "basic arithmetic" begin
     @test 2 + 3 == 5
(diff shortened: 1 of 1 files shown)


dkarrasch (approved):

test/numbers.jl
```diff
@@ -63,6 +63,11 @@ end
 
     @test iszero(false) && !iszero(true)
     @test isone(true) && !isone(false)
+
+    @test !typemin(Bool)
```
StefanKarpinski:
```suggestion
    @test typemin(Bool) === false
```

test/numbers.jl
```diff
@@ -63,6 +63,11 @@ end
...
+    @test typemax(Bool)
```
StefanKarpinski:
```suggestion
    @test typemax(Bool) === true
```

test/numbers.jl
```diff
@@ -63,6 +63,11 @@ end
...
+    @test !abs(false)
```
StefanKarpinski:
```suggestion
    @test abs(false) === false
```

test/numbers.jl
```diff
@@ -63,6 +63,11 @@ end
...
+    @test abs(true)
```
StefanKarpinski:
```suggestion
    @test abs(true) === true
```

StefanKarpinski:
Seems clearer to spell out which boolean value each test returns.

bramtayl:
Ok that makes sense

PR #34528, opened 2020-01-27
Line 57 update

Mohit-Nain:
much can be replaced with some , to make it sound positive .


ViralBShah:
`much` is appropriate here.

ViralBShah:
Please use `[ci skip]`.

PR #34533, opened 2020-01-27
Missing trait tests

bramtayl:
[Uncovered](https://codecov.io/gh/JuliaLang/julia/src/master/base/traits.jl#L9)
[Uncovered](https://codecov.io/gh/JuliaLang/julia/src/master/base/traits.jl#L19)
[Uncovered](https://codecov.io/gh/JuliaLang/julia/src/master/base/traits.jl#L59)

### DIFF
--- a/test/misc.jl
+++ b/test/misc.jl
@@ -804,3 +804,9 @@ end
 
     GC.safepoint()
 end
+
+@testset "traits" begin
+    @test Base.OrderStyle(1) isa Base.Ordered
+    @test Base.ArithmeticStyle(1) isa Base.ArithmeticWraps
+    @test Base.RangeStepStyle(1:2) isa Base.RangeStepRegular
+end


test/misc.jl
```diff
@@ -804,3 +804,9 @@ end
 
     GC.safepoint()
 end
+
+@testset "traits" begin
+    @test Base.OrderStyle(1) isa Base.Ordered
+    @test Base.ArithmeticStyle(1) isa Base.ArithmeticWraps
+    @test Base.RangeStepStyle(1:2) isa Base.RangeStepRegular
+end
```
JeffBezanson: These seem awfully specific to me, especially since these are not exported. If these are not covered, it must mean there is something higher level that's not being tested, or perhaps coverage is missing them due to always being inlined?

bramtayl: Agreed. The type version of the traits is tested, but just not the instance version. I guess the instance methods are offered for convenience? But since they aren't exported or used, maybe we could just delete them?

bramtayl: Does that make sense?

bramtayl:
Let me know if you want me rebase this

bramtayl:
Let me know if you want me rebase this

Issue #34482, opened 2020-01-22
Incorrect nothrow modeling for intrinsics


Keno:
```
julia> function foo()
           Base.not_int("ABC")
           1
       end
foo (generic function with 1 method)

julia> foo()
1

julia> Base.not_int("ABC")
ERROR: not_int: value is not a primitive type
Stacktrace:
 [1] top-level scope at REPL[42]:1
```
`foo` should have thrown, but didn't.

Keno:
Now that I think about it, that might also be a codegen issue, but either way it should be fixed.

PR #34539, opened 2020-01-27
Fix nothrow behavior for intrinsics

Keno:
This fixes the case in #34482 and adds some more robustness for similar
cases, though I wouldn't be surprised if there were more dragons hiding
here. Intrinsics are a bit special and from the start, we sort of expected
them to only ever be called correctly under pain of segfaults or other
undefined behavior. We've been gradually making these more robust,
but fundamentally, they were never intended to be used by users directly,
only through the type-validating wrappers in Base. What has changed
to cause the recent influx of issues in this area is that people now
like to do compiler transforms that happily recurse through these wrappers
and perform transforms that are not always legal on Intrinsics. This should
help catch a number of them, but this interface is still not very thoroughly
validated, and I would be surprised to see crashes or other errors stemming
from incorrect usage here.

### DIFF
--- a/base/compiler/ssair/inlining.jl
+++ b/base/compiler/ssair/inlining.jl
@@ -1169,8 +1169,7 @@ function early_inline_special_case(ir::IRCode, s::Signature, e::Expr, params::Pa
             val = etype.val
             is_inlineable_constant(val) || return nothing
             if isa(f, IntrinsicFunction)
-                if is_pure_intrinsic_infer(f) &&
-                    (intrinsic_nothrow(f) || intrinsic_nothrow(f, atypes[2:end]))
+                if is_pure_intrinsic_infer(f) && intrinsic_nothrow(f, atypes[2:end])
                     return quoted(val)
                 end
             elseif ispuretopfunction(f) || contains_is(_PURE_BUILTINS, f)
--- a/base/compiler/tfuncs.jl
+++ b/base/compiler/tfuncs.jl
@@ -403,20 +403,23 @@ add_tfunc(Core._typevar, 3, 3, typevar_tfunc, 100)
 add_tfunc(applicable, 1, INT_INF, (@nospecialize(f), args...)->Bool, 100)
 add_tfunc(Core.Intrinsics.arraylen, 1, 1, @nospecialize(x)->Int, 4)
 add_tfunc(arraysize, 2, 2, (@nospecialize(a), @nospecialize(d))->Int, 4)
+function pointer_eltype(@nospecialize(ptr))
+    a = widenconst(ptr)
+    if a <: Ptr
+        if isa(a,DataType) && isa(a.parameters[1],Type)
+            return a.parameters[1]
+        elseif isa(a,UnionAll) && !has_free_typevars(a)
+            unw = unwrap_unionall(a)
+            if isa(unw,DataType)
+                return rewrap_unionall(unw.parameters[1], a)
+            end
+        end
+    end
+    return Any
+end
 add_tfunc(pointerref, 3, 3,
           function (@nospecialize(a), @nospecialize(i), @nospecialize(align))
-              a = widenconst(a)
-              if a <: Ptr
-                  if isa(a,DataType) && isa(a.parameters[1],Type)
-                      return a.parameters[1]
-                  elseif isa(a,UnionAll) && !has_free_typevars(a)
-                      unw = unwrap_unionall(a)
-                      if isa(unw,DataType)
-                          return rewrap_unionall(unw.parameters[1], a)
-                      end
-                  end
-              end
-              return Any
+            return pointer_eltype(a)
           end, 4)
 add_tfunc(pointerset, 4, 4, (@nospecialize(a), @nospecialize(v), @nospecialize(i), @nospecialize(align)) -> a, 5)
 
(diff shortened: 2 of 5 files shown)


base/compiler/tfuncs.jl
```diff
@@ -1393,15 +1393,26 @@ function builtin_tfunction(@nospecialize(f), argtypes::Array{Any,1},
...
+        return false
+    end
+    tf = T_IFUNC[iidx]
+    tf = tf::Tuple{Int, Int, Any}
+    if !(tf[1] <= length(argtypes) <= tf[2])
+        # wrong # of args
+        return false
+    end
+    for i = 1:length(argtypes)
+        # Intrinsics are defined only on primitive types
```
vtjnash: Some intrinsics (arrayref, bitcast) are defined for other types

Keno: Ah, forgot about those - thanks.

src/intrinsics.cpp
```diff
@@ -895,8 +895,12 @@ static jl_cgval_t emit_intrinsic(jl_codectx_t &ctx, intrinsic f, jl_value_t **ar
     // return emit_runtime_call(ctx, f, argv, nargs);
 
     switch (f) {
-    case arraylen:
-        return mark_julia_type(ctx, emit_arraylen(ctx, argv[0]), false, jl_long_type);
+    case arraylen: {
+        const jl_cgval_t &x = argv[0];
+        if (!jl_is_datatype(x.typ) || ((jl_datatype_t*)x.typ)->name != jl_array_typename)
```
vtjnash: `unwrap_unionall()`
```suggestion
        jl_value_t *typ = jl_unwrap_unionall(x.typ);
        if (!jl_is_datatype(typ) || ((jl_datatype_t*)typ)->name != jl_array_typename)
```

base/compiler/tfuncs.jl
```diff
@@ -1393,22 +1396,31 @@ function builtin_tfunction(@nospecialize(f), argtypes::Array{Any,1},
...
-        f === Intrinsics.checked_udiv_int ||
-        f === Intrinsics.checked_srem_int ||
-        f === Intrinsics.checked_urem_int ||
-        f === Intrinsics.cglobal
-    )
 
 function intrinsic_nothrow(f::IntrinsicFunction, argtypes::Array{Any, 1})
+    # First check that we have the correct number of arguments
+    iidx = Int(reinterpret(Int32, f::IntrinsicFunction)) + 1
+    if iidx < 0 || iidx > length(T_IFUNC)
```
vtjnash:
```suggestion
    if iidx < 1 || iidx > length(T_IFUNC)
```

base/compiler/tfuncs.jl
```diff
@@ -1393,22 +1396,31 @@ function builtin_tfunction(@nospecialize(f), argtypes::Array{Any,1},
...
     f === Intrinsics.cglobal && return false
+    # TODO: We can't know for sure, but the user should have a way to assert
+    # that it won't
+    f === Intrinsics.llvmcall && return false
     if f === Intrinsics.checked_udiv_int || f === Intrinsics.checked_urem_int || f === Intrinsics.checked_srem_int || f === Intrinsics.checked_sdiv_int
         # Nothrow as long as the second argument is guaranteed not to be zero
         isa(argtypes[2], Const) || return false
         den_val = argtypes[2].val
         den_val !== zero(typeof(den_val)) || return false
+        return true
```
vtjnash: Feel like these should have `isprimitivetype` tests added too? (and checked_sdiv_int below)

base/compiler/ssair/inlining.jl
```diff
@@ -1169,8 +1169,7 @@ function early_inline_special_case(ir::IRCode, s::Signature, e::Expr, params::Pa
             val = etype.val
             is_inlineable_constant(val) || return nothing
             if isa(f, IntrinsicFunction)
-                if is_pure_intrinsic_infer(f) &&
-                    (intrinsic_nothrow(f) || intrinsic_nothrow(f, atypes[2:end]))
+                if is_pure_intrinsic_infer(f) && intrinsic_nothrow(f, atypes[2:end])
```
vtjnash: Do we need to handle Vararg early in this function and bail?

base/compiler/tfuncs.jl
```diff
@@ -1422,6 +1434,34 @@ function intrinsic_nothrow(f::IntrinsicFunction, argtypes::Array{Any, 1})
...
+    end
+    if f === Intrinsics.bitcast
+        ty = instanceof_tfunc(argtypes[1])[1]
+        xty = widenconst(argtypes[2])
+        return isprimitivetype(ty) && isprimitivetype(xty) && ty.size === xty.size
+    end
+    if f in (Intrinsics.sext_int, Intrinsics.zext_int, Intrinsics.trunc_int,
+             Intrinsics.fptoui, Intrinsics.fptosi, Intrinsics.uitofp,
+             Intrinsics.sitofp, Intrinsics.fptrunc, Intrinsics.fpext)
+        ty = instanceof_tfunc(argtypes[1])[1]
```
vtjnash: IIUC, if the second return value is `false`, then `ty` may include the value `Union{}` (e.g. the call `Intrinsics.sext_int(Union{}, 1)`)

base/compiler/tfuncs.jl
```diff
@@ -1422,6 +1434,34 @@ function intrinsic_nothrow(f::IntrinsicFunction, argtypes::Array{Any, 1})
...
+        xty = widenconst(argtypes[2])
+        return isprimitivetype(ty) && isprimitivetype(xty)
```
vtjnash: Separately, we may need to also review what happens if the relation between `ty.size ≏ xty.size` is not what LLVM expects for these

base/compiler/tfuncs.jl
```diff
@@ -1422,6 +1434,34 @@ function intrinsic_nothrow(f::IntrinsicFunction, argtypes::Array{Any, 1})
...
+    end
+    # The remaining intrinsics are math/bits intrinsics. They work on all
```
vtjnash:
```suggestion
    # The remaining intrinsics are math/bits/comparison intrinsics. They work on all
```

base/compiler/tfuncs.jl
```diff
@@ -1422,6 +1434,34 @@ function intrinsic_nothrow(f::IntrinsicFunction, argtypes::Array{Any, 1})
...
+    # primitive types.
```
vtjnash: "...all primitive types of the same type" ?

vtjnash:
Looking good! Just thought of a few more cases we may want to handle, while we're at it.

base/compiler/tfuncs.jl
Keno: In theory yes, but I don't think we check this in the runtime implementation.

base/compiler/ssair/inlining.jl
Keno: No, vararg is handled earlier in inlining and it refuses to construct a `Signature` object for such cases.

base/compiler/tfuncs.jl
Keno: Thanks!

base/compiler/tfuncs.jl
Keno: Yes, but I'll leave that for future work.

base/compiler/tfuncs.jl
vtjnash: easy to check...
```
julia> Core.Intrinsics.add_int(Int64(1), UInt64(1))
ERROR: add_int: types of a and b must match
Stacktrace:
 [1] top-level scope at REPL[1]:1
```

Keno: Ah, I misread some code.

Keno:
Good to go?

vtjnash:
Sure. I was just realizing that the test we do is insufficient though:
```
julia> isprimitivetype(Ptr)
true
julia> Ptr === Ptr
true
julia> Ptr >: Ptr{UInt8} === Ptr{Int8} <: Ptr
false
```

Keno:
Are there cases where our intrinsics throw when they see the `Ptr` UnionAll? Anyway, let's get this in and we can refine in future commits.

Issue #34513, opened 2020-01-26
Internal error: encountered unexpected error in runtime: MethodError(...


ablaom:
I'm encountering this only in julia 1.0.5 and julia 1.1.1. No problems for 1.2, 1,3 and nightly.

See [this](https://travis-ci.com/alan-turing-institute/MLJTuning.jl/jobs/280125850#L325) travis test of my unregistered package MLJTuning, with all testing turning off except testing of `src/julia_bug.jl` where I've tried to isolate the issue as best as I can.

```julia
Internal error: encountered unexpected error in runtime:
325MethodError(f=typeof(Core.Compiler.fieldindex)(), args=(MLJTuning.DeterministicTunedModel{T, M, R} where R, :acceleration, false), world=0x0000000000000eb9)
326rec_backtrace at /buildworker/worker/package_linux64/build/src/stackwalk.c:94
327record_backtrace at /buildworker/worker/package_linux64/build/src/task.c:217 [inlined]
328jl_throw at /buildworker/worker/package_linux64/build/src/task.c:417
329jl_method_error_bare at /buildworker/worker/package_linux64/build/src/gf.c:1627
330jl_method_error at /buildworker/worker/package_linux64/build/src/gf.c:1645
331jl_apply_generic at /buildworker/worker/package_linux64/build/src/gf.c:2173
332_fieldtype_nothrow at ./compiler/tfuncs.jl:803
333_fieldtype_nothrow at ./compiler/tfuncs.jl:799
334jfptr__fieldtype_nothrow_2515.clone_1 at /home/travis/julia/lib/julia/sys.so (unknown line)
335jl_apply_generic at /buildworker/worker/package_linux64/build/src/gf.c:2197
336fieldtype_nothrow at ./compiler/tfuncs.jl:794
337_builtin_nothrow at ./compiler/tfuncs.jl:1168
338stmt_effect_free at ./compiler/ssair/queries.jl:34
339optimize at ./compiler/optimize.jl:178
340typeinf at ./compiler/typeinfer.jl:35
341abstract_call_method_with_const_args at ./compiler/abstractinterpretation.jl:198
342abstract_call_gf_by_type at ./compiler/abstractinterpretation.jl:107
343abstract_call at ./compiler/abstractinterpretation.jl:776
344abstract_eval_call at ./compiler/abstractinterpretation.jl:805
345abstract_eval at ./compiler/abstractinterpretation.jl:890
[...]

PR #34541, opened 2020-01-27
inference: make getfield_tfunc more robust for abstract PartialStruct

vtjnash:
We're currently careful never to make these. But good to be careful?

refs #34513

### DIFF
--- a/base/compiler/tfuncs.jl
+++ b/base/compiler/tfuncs.jl
@@ -826,17 +826,18 @@ function getfield_tfunc(@nospecialize(s00), @nospecialize(name))
             end
         end
         s = typeof(sv)
-    elseif isa(s, PartialStruct)
+    elseif isa(s00, PartialStruct)
+        s = widenconst(s00)
+        sty = unwrap_unionall(s)::DataType
         if isa(name, Const)
             nv = name.val
             if isa(nv, Symbol)
-                nv = fieldindex(widenconst(s), nv, false)
+                nv = fieldindex(sty, nv, false)
             end
-            if isa(nv, Int) && 1 <= nv <= length(s.fields)
-                return unwrapva(s.fields[nv])
+            if isa(nv, Int) && 1 <= nv <= length(s00.fields)
+                return unwrapva(s00.fields[nv])
             end
         end
-        s = widenconst(s)
     end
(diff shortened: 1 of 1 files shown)


base/compiler/tfuncs.jl
```diff
@@ -727,17 +727,21 @@ function getfield_tfunc(@nospecialize(s00), @nospecialize(name))
             end
         end
         s = typeof(sv)
-    elseif isa(s, PartialStruct)
+    elseif isa(s00, PartialStruct)
+        s01 = widenconst(s00)
+        s = unwrap_unionall(s01)::DataType
```
vtjnash:
```suggestion
        s = s01::DataType
```
currently this simplification would hold, but do we want to assume that?

vtjnash:
It turns out we might make abstract PartialStruct for Tuple (refs https://github.com/JuliaLang/julia/pull/39402), so reopening to add vararg handling to this also

martinholters:
So is it possible to construct a case that fails with the current code?

aviatesk (approved):
LGTM. AFAIU, `Vararg` handling was already added https://github.com/JuliaLang/julia/pull/37769.
Still needs rebase though.

base/compiler/tfuncs.jl
```diff
@@ -826,17 +826,18 @@ function getfield_tfunc(@nospecialize(s00), @nospecialize(name))
             end
         end
         s = typeof(sv)
-    elseif isa(s, PartialStruct)
+    elseif isa(s00, PartialStruct)
+        s = widenconst(s00)
+        sty = unwrap_unionall(s)::DataType
```
aviatesk, line 831: Just to make sure: is there any possibility where `s::DataType` doesn't hold ?
This check is good to have though, we may want to form `PartialStruct` for abstract types other than tuples in the future.

aviatesk (approved):

base/compiler/tfuncs.jl
vtjnash: Currently, no. That is why this PR has sat for so long untouched. But turns out to be fairly straightforward here to add (though I think the necessary type lattice changes may currently be difficult).

PR #34543, opened 2020-01-27
add `Threads.foreach` for convenient multithreaded Channel consumption

jrevels:
This is a spiritual successor to #34185, but is a bit more minimal. The motivation came about when I  realized #34185 was a wonky solution to a pattern that had started cropping up for me: I wanted a multithreaded `foreach(f, channel)` where I could communicate multiple results per `f` call by `put!`ing them individually to a closed-over `Channel` (so not a direct map). While #34185 could be hackily employed for this, I thought it was overkill so I ripped this out instead. Once I started using it, I realized I liked it a bit more.

Unlike #34185, it's only defined on `Channel`s, not arbitrary iterators. It could be extended to arbitrary iterators if we wanted via #34185's locking approach; I just thought it'd be better to keep things simple from the get-go. I could also loosen the dispatch constraint to `AbstractChannel` and just document the expectation of threadsafe iteration.

cc @tkf, who made a wonderful Transducers implementation of #34185

### DIFF
--- a/base/channels.jl
+++ b/base/channels.jl
@@ -474,13 +474,13 @@ end
 IteratorSize(::Type{<:Channel}) = SizeUnknown()
 
 """
-    threaded_foreach(f, channel::Channel; ntasks=Threads.nthreads(), async=true)
+    Threads.foreach(f, channel::Channel; ntasks=Threads.nthreads())
 
 Similar to `foreach(f, channel)`, but iteration over `channel` and calls to
 `f` are split across `ntasks` Tasks spawned by `Threads.@spawn`. This function
 will wait for all internally spawned tasks to complete before returning.
 """
-function threaded_foreach(f, channel::Channel; ntasks=Threads.nthreads())
+function Threads.foreach(f, channel::Channel; ntasks=Threads.nthreads())
     @sync for _ in 1:ntasks
         Threads.@spawn begin
             for item in channel
--- a/base/channels.jl
+++ b/base/channels.jl
@@ -472,3 +472,24 @@ function iterate(c::Channel, state=nothing)
 end
 
 IteratorSize(::Type{<:Channel}) = SizeUnknown()
+
+"""
+    threaded_foreach(f, channel::Channel; ntasks=Threads.nthreads(), async=true)
+
+Similar to `foreach(f, channel)`, but iteration over `channel` and calls to
+`f` are split across `ntasks` Tasks spawned by `Threads.@spawn`.
+
+If `async` is `false`, this function will `wait` for all internally spawned tasks
+to complete before returning.
+"""
+function threaded_foreach(f, channel::Channel; ntasks=Threads.nthreads(), async=true)
+    tasks = map(1:ntasks) do _
+        Threads.@spawn begin
+            for item in channel
+                wait(Threads.@spawn f(item))
+            end
+        end
+    end
+    async || foreach(wait, tasks)
+    return nothing
+end
--- a/base/channels.jl
+++ b/base/channels.jl
@@ -481,13 +481,20 @@ Similar to `foreach(f, channel)`, but iteration over `channel` and calls to
 will wait for all internally spawned tasks to complete before returning.
 """
 function Threads.foreach(f, channel::Channel; ntasks=Threads.nthreads())
-    tasks = map(1:ntasks) do _
-        Threads.@spawn begin
+    stop = Threads.Atomic{Bool}(false)
+    @sync for _ in 1:ntasks
+        Threads.@spawn try
             for item in channel
-                wait(Threads.@spawn f(item))
+                _waitspawn(f, item)
+                stop[] && break  # do this _after_ f(item) to avoid loosing `item`
             end
+        catch
+            stop[] = true
+            rethrow()
         end
     end
-    Base.sync_end(tasks)
     return nothing
 end
+
+# Avoid registering the task in outer @sync:
+_waitspawn(f, x) = wait(Threads.@spawn f(x))
--- a/base/channels.jl
+++ b/base/channels.jl
@@ -477,19 +477,16 @@ IteratorSize(::Type{<:Channel}) = SizeUnknown()
     threaded_foreach(f, channel::Channel; ntasks=Threads.nthreads(), async=true)
 
 Similar to `foreach(f, channel)`, but iteration over `channel` and calls to
-`f` are split across `ntasks` Tasks spawned by `Threads.@spawn`.
-
-If `async` is `false`, this function will `wait` for all internally spawned tasks
-to complete before returning.
+`f` are split across `ntasks` Tasks spawned by `Threads.@spawn`. This function
+will wait for all internally spawned tasks to complete before returning.
 """
-function threaded_foreach(f, channel::Channel; ntasks=Threads.nthreads(), async=true)
-    tasks = map(1:ntasks) do _
+function threaded_foreach(f, channel::Channel; ntasks=Threads.nthreads())
+    @sync for _ in 1:ntasks
         Threads.@spawn begin
             for item in channel
                 wait(Threads.@spawn f(item))
             end
         end
     end
-    async || foreach(wait, tasks)
     return nothing
 end
--- a/base/channels.jl
+++ b/base/channels.jl
@@ -481,12 +481,13 @@ Similar to `foreach(f, channel)`, but iteration over `channel` and calls to
 will wait for all internally spawned tasks to complete before returning.
 """
 function Threads.foreach(f, channel::Channel; ntasks=Threads.nthreads())
-    @sync for _ in 1:ntasks
+    tasks = map(1:ntasks) do _
         Threads.@spawn begin
             for item in channel
                 wait(Threads.@spawn f(item))
             end
         end
     end
+    Base.sync_end(tasks)
     return nothing
 end
--- a/base/channels.jl
+++ b/base/channels.jl
@@ -472,29 +472,3 @@ function iterate(c::Channel, state=nothing)
 end
 
 IteratorSize(::Type{<:Channel}) = SizeUnknown()
-
-"""
-    Threads.foreach(f, channel::Channel; ntasks=Threads.nthreads())
-
-Similar to `foreach(f, channel)`, but iteration over `channel` and calls to
-`f` are split across `ntasks` Tasks spawned by `Threads.@spawn`. This function
-will wait for all internally spawned tasks to complete before returning.
-"""
-function Threads.foreach(f, channel::Channel; ntasks=Threads.nthreads())
-    stop = Threads.Atomic{Bool}(false)
-    @sync for _ in 1:ntasks
-        Threads.@spawn try
-            for item in channel
-                _waitspawn(f, item)
-                stop[] && break  # do this _after_ f(item) to avoid loosing `item`
-            end
-        catch
-            stop[] = true
-            rethrow()
-        end
-    end
-    return nothing
-end
-
-# Avoid registering the task in outer @sync:
-_waitspawn(f, x) = wait(Threads.@spawn f(x))
--- a/base/threadfuncs.jl
+++ b/base/threadfuncs.jl
@@ -1,37 +1,32 @@
 # This file is a part of Julia. License is MIT: https://julialang.org/license
 
 """
-    Threads.foreach(f, channel::Channel; ntasks=Threads.nthreads(),
-                    schedule=:fair)
+    Threads.foreach(f, channel::Channel,
+                    schedule::Threads.AbstractSchedule=Threads.FairSchedule();
+                    ntasks=Threads.nthreads())
 
 Similar to `foreach(f, channel)`, but iteration over `channel` and calls to
 `f` are split across `ntasks` tasks spawned by `Threads.@spawn`. This function
 will wait for all internally spawned tasks to complete before returning.
 
-If `schedule` is `:fair`, `Threads.foreach` will attempt to spawn tasks in a
+If `schedule isa FairSchedule`, `Threads.foreach` will attempt to spawn tasks in a
 manner that enables Julia's scheduler to more freely load-balance work items across
 threads. This approach generally has higher per-item overhead, but may perform
-better than the `:static` schedule in concurrence with other multithreaded
-workloads.
+better than `StaticSchedule` in concurrence with other multithreaded workloads.
 
-If `schedule` is `:static`, `Threads.foreach` will spawn tasks in a manner that
-incurs lower per-item overhead than the `:fair` schedule, but is less amenable
+If `schedule isa StaticSchedule`, `Threads.foreach` will spawn tasks in a manner
+that incurs lower per-item overhead than `FairSchedule`, but is less amenable
 to load-balancing. This approach thus may be more suitable for fine-grained,
-uniform workloads, but may perform worse than the `:fair` schedule in
-concurrence with other multithreaded workloads.
+uniform workloads, but may perform worse than `FairSchedule` in concurrence
+with other multithreaded workloads.
 
 !!! compat "Julia 1.6"
     This function requires Julia 1.6 or later.
 """
-function Threads.foreach(f, channel::Channel; ntasks=Threads.nthreads(),
-                         schedule=:fair)
-    if schedule === :fair
-        apply = _waitspawn
-    elseif schedule === :static
-        apply = (f, x) -> f(x)
-    else
-        throw(ArgumentError("`schedule` must be either `:fair` or `:static`"))
-    end
+function Threads.foreach(f, channel::Channel,
+                         schedule::Threads.AbstractSchedule=Threads.FairSchedule();
+                         ntasks=Threads.nthreads())
+    apply = _apply_for_schedule(schedule)
     stop = Threads.Atomic{Bool}(false)
     @sync for _ in 1:ntasks
         Threads.@spawn try
@@ -52,5 +47,5 @@ function Threads.foreach(f, channel::Channel; ntasks=Threads.nthreads(),
     return nothing
 end
 
-# avoid registering the task within `Threads.foreach`'s outer `@sync`
-_waitspawn(f, x) = wait(Threads.@spawn f(x))
+_apply_for_schedule(::Threads.StaticSchedule) = (f, x) -> f(x)
+_apply_for_schedule(::Threads.FairSchedule) = (f, x) -> wait(Threads.@spawn f(x))
--- a/base/threadfuncs.jl
+++ b/base/threadfuncs.jl
@@ -34,7 +34,7 @@ function Threads.foreach(f, channel::Channel; ntasks=Threads.nthreads(),
     @sync for _ in 1:ntasks
         Threads.@spawn try
             for item in channel
-                apply(f, item)
+                $apply(f, item)
                 # do `stop[] && break` after `f(item)` to avoid losing `item`.
                 # this isn't super comprehensive since a task could still get
                 # stuck on `take!` at `for item in channel`. We should think
--- a/base/threadfuncs.jl
+++ b/base/threadfuncs.jl
@@ -1,0 +1,30 @@
+# This file is a part of Julia. License is MIT: https://julialang.org/license
+
+"""
+    Threads.foreach(f, channel::Channel; ntasks=Threads.nthreads())
+
+Similar to `foreach(f, channel)`, but iteration over `channel` and calls to
+`f` are split across `ntasks` Tasks spawned by `Threads.@spawn`. This function
+will wait for all internally spawned tasks to complete before returning.
+
+!!! compat "Julia 1.6"
+    This function requires Julia 1.6 or later.
+"""
+function Threads.foreach(f, channel::Channel; ntasks=Threads.nthreads())
+    stop = Threads.Atomic{Bool}(false)
+    @sync for _ in 1:ntasks
+        Threads.@spawn try
+            for item in channel
+                _waitspawn(f, item)
+                stop[] && break  # do this _after_ f(item) to avoid loosing `item`
+            end
+        catch
+            stop[] = true
+            rethrow()
+        end
+    end
+    return nothing
+end
+
+# Avoid registering the task in outer @sync:
+_waitspawn(f, x) = wait(Threads.@spawn f(x))
--- a/base/threadfuncs.jl
+++ b/base/threadfuncs.jl
@@ -1,18 +1,35 @@
 # This file is a part of Julia. License is MIT: https://julialang.org/license
 
 """
-    Threads.foreach(f, channel::Channel; ntasks=Threads.nthreads())
+    Threads.foreach(f, channel::Channel; ntasks=Threads.nthreads(),
+                    schedule=:fair)
 
 Similar to `foreach(f, channel)`, but iteration over `channel` and calls to
-`f` are split across `ntasks` Tasks spawned by `Threads.@spawn`. This function
+`f` are split across `ntasks` tasks spawned by `Threads.@spawn`. This function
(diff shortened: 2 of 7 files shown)


base/channels.jl
```diff
@@ -457,3 +457,24 @@ function iterate(c::Channel, state=nothing)
...
+"""
+function threaded_foreach(f, channel::Channel; ntasks=Threads.nthreads(), async=true)
+    tasks = map(1:ntasks) do _
+        Threads.@spawn begin
+            for item in channel
+                wait(Threads.@spawn f(item))
+            end
+        end
+    end
+    async || foreach(wait, tasks)
```
tkf: I highly appreciate if `async` option is removed.  There is discussion structured concurrency #33248 and adding "unstructured" construct such as `async=true` here makes implementing structured concurrency in Julia much harder.  You can still do `@async threaded_foreach(f, channel)` (preferably within a `@sync` block).

tkf:
> it's only defined on `Channel`s, not arbitrary iterators

It'd be really nice if we had `Channel(iterator)` overload but unfortunately we already have `Channel(callable)`.  I've been wondering if it makes sense to add a function (say) `into(::Type{T}, iterator) :: T` with methods like `into(::Type{Channel}, iterator)` and `into(::Type{<:AbstractVector}, iterator)`. Then, we can just do `threaded_foreach(f, into(Channel, iterator))`.

base/channels.jl
```diff
@@ -457,3 +457,24 @@ function iterate(c::Channel, state=nothing)
...
+"""
+    threaded_foreach(f, channel::Channel; ntasks=Threads.nthreads(), async=true)
+
+Similar to `foreach(f, channel)`, but iteration over `channel` and calls to
+`f` are split across `ntasks` Tasks spawned by `Threads.@spawn`.
+
+If `async` is `false`, this function will `wait` for all internally spawned tasks
+to complete before returning.
+"""
+function threaded_foreach(f, channel::Channel; ntasks=Threads.nthreads(), async=true)
```
tkf: Does it make sense to use the name `Threads.foreach`, like we have `Iterators.filter`?

base/channels.jl
```diff
@@ -457,3 +457,24 @@ function iterate(c::Channel, state=nothing)
...
+    tasks = map(1:ntasks) do _
+        Threads.@spawn begin
+            for item in channel
+                wait(Threads.@spawn f(item))
```
tkf: Maybe a bit too tangent, but FYI I noticed you can get a bit more speedup https://github.com/tkf/Transducers.jl/pull/123#issuecomment-579010628 if you collect the elements into a buffer `Vector` and then iterate over it. You may want to improve latency instead of throughput so I guess it's not always a better choice. But I wonder if it makes sense to add a "buffer size" option to it.

base/channels.jl
tkf: ref: @vchuravy suggested to use C++ -like policy-based interface https://github.com/JuliaLang/julia/pull/34185#issuecomment-568878123

jrevels: > Does it make sense to use the name Threads.foreach, like we have Iterators.filter?

That would be nice. 

One tricky detail (but not an actual problem) is that writing code that uses `Threads.@spawn` from within `Threads` is pretty wonky right now because the macro depends on functions that aren't defined until later (in `tasks.jl`) and `Channel`s also aren't defined until later. Clearly can churn things around to be defined in the right order but might involve some bikeshedding to untangle the dependencies

tkf: Ah, I didn't know that.  I guess one way to work around the problem is to just declare the "stub" function like this?

```julia
module Threads
function foreach end
...
end

# later
function Threads.foreach(f, channel::Channel)
    ...
end
```

jrevels: I pushed the above change, since I like `Threads.foreach` and the stub approach is way less hassle/churn than actually shuffling code :p 

Folks should, of course, feel free to continue bikeshedding the name.

jrevels:
Added some extremely basic tests - let me know if folks have other testing ideas and I'll pop 'em in.

base/channels.jl
jrevels: This indeed sounds like a useful option! I'd probably want to fiddle around with it in a follow-up PR, though.

jrevels:
> I've been wondering if it makes sense to add a function (say) into(::Type{T}, iterator) :: T with methods like into(::Type{Channel}, iterator) and into(::Type{<:AbstractVector}, iterator).

In this vein, I feel like I often want shorthand `Channel{T}(c -> foreach(i -> put!(i, c), iterator), ...)` it's not that long really, but the chaining usually makes my eyeballs do a double-scan :p 

It might be nice to be able to write e.g. `Channel{T}(from(iterator), ...)` for this, but `from(x) = c -> foreach(i -> put!(i, c), x)` is so short and easy to write that I almost hesitate to suggest adding it to Base...maybe something with a less generic/more descriptive name like `puteach(iterator)` wouldn't be bad though.

base/channels.jl
```diff
@@ -457,3 +457,21 @@ function iterate(c::Channel, state=nothing)
...
+
+Similar to `foreach(f, channel)`, but iteration over `channel` and calls to
+`f` are split across `ntasks` Tasks spawned by `Threads.@spawn`. This function
+will wait for all internally spawned tasks to complete before returning.
+"""
+function Threads.foreach(f, channel::Channel; ntasks=Threads.nthreads())
+    @sync for _ in 1:ntasks
+        Threads.@spawn begin
+            for item in channel
+                wait(Threads.@spawn f(item))
```
tkf: I just realized there is a memory leak and thread-unsafe `push!` (I think). With this code as-is, `Threads.@spawn f(item)` keeps `push!`ing the tasks to the task list of the outer `@sync`. I think we need to declare local lexical scope like this:

```suggestion
                @sync Threads.@spawn f(item)
```

tkf:
> It might be nice to be able to write e.g. `Channel{T}(from(iterator), ...)` for this, but `from(x) = c -> foreach(i -> put!(i, c), x)` is so short and easy to write that I almost hesitate to suggest adding it to Base

It looks like `from` is a nice useful one-linear but I agree it's not enough for adding it to Base.

base/channels.jl
tkf: ref https://github.com/JuliaLang/julia/issues/34666

jrevels: Ah, really good catch. Just got rid of the `@sync` and manually called `Base.sync_end` on the tasks for now.

jrevels:
> I suppose we should stop all the tasks immediately and propagate the error to the caller

Yeah, that'd probably be the right thing to do. 

> One question is if it is OK to close the given channel on error.

Eh, this would go against my expectation as a caller - is this generally desirable for a reason I might be missing? It makes sense to me if a use case treated "dropped" elements (i.e. elements removed from the channel but not fully consumed due to the error) as a corruption or something like that, but in that case, the caller should probably just wrap this in a try-catch and `close` the channel themselves if that's how they desire to handle the situation, right?

jrevels:
> is this generally desirable for a reason I might be missing

Oh, you probably mean to prevent other threads from taking objects before we get a chance to kill them? Hmm. Yeah, it would be ideal if there were a different mechanism for that IMO though I obviously haven't thought about this much 😅

tkf:
It was a genuine question as I don't know which one is better.

> the caller should probably just wrap this in a try-catch and `close` the channel themselves if that's how they desire to handle the situation, right?

Ah, I think this is a good point.  It's easy to mimic close-on-error behavior on top of non-closing `Threads.foreach`.  However, inserting a "shim" channel to opt-out close-on-error behavior takes more code.

jrevels:
Bump. @JeffBezanson thoughts on this PR (especially the error-handling discussion above?)

ararslan:
Re-bump. It would be really nice to be able to use this.

tkf:
(I'm thinking to add this overload to [`ThreadsX.foreach`](https://tkf.github.io/ThreadsX.jl/dev/#ThreadsX.foreach) but I feel bad to steal @jrevels's idea twice :smile:)

jrevels:
> (I'm thinking to add this overload to ThreadsX.foreach but I feel bad to steal @jrevels's idea twice 😄)

Go for it! 😁

jrevels:
@JeffBezanson / @vtjnash if either of you do get some time to looks at this, IMO it's essentially mergeable as-is pending approval.

The only remaining decision point AFAICT (discussed above) is whether we should/need to add some exception handling, i.e. to prevent threads from taking objects from the channel before we get a chance to kill them in the event of an error. I'm not sure we do - seems like the caller can handle a good chunk of that if they need to - but would love to hear your thoughts, especially if you have any ideas that would make exception handling here more ergonomic

base/channels.jl
```diff
@@ -457,3 +457,22 @@ function iterate(c::Channel, state=nothing)
...
+    tasks = map(1:ntasks) do _
+        Threads.@spawn begin
+            for item in channel
+                wait(Threads.@spawn f(item))
+            end
+        end
+    end
+    Base.sync_end(tasks)
+    return nothing
+end
```
tkf, line 478: Two suggestions:

* Use `@sync` instead of `sync_end`. It's recently changed to accept `Channel`. So current PR doesn't work anymore. I think it's better to use public interface even internally.
* Add a simple implementation of stop-all-on-error behavior.

```suggestion
function Threads.foreach(f, channel::Channel; ntasks=Threads.nthreads())
    stop = Threads.Atomic{Bool}(false)
    @sync for _ in 1:ntasks
        Threads.@spawn try
            for item in channel
                _waitspawn(f, item)
                stop[] && break  # do this _after_ f(item) to avoid loosing `item`
            end
        catch
            stop[] = true
            rethrow()
        end
    end
    return nothing
end

# Avoid registering the task in outer @sync:
_waitspawn(f, x) = wait(Threads.@spawn f(x))
```

jrevels: Thanks!

base/channels.jl
```diff
@@ -472,3 +472,29 @@ function iterate(c::Channel, state=nothing)
...
 
 IteratorSize(::Type{<:Channel}) = SizeUnknown()
+
+"""
+    Threads.foreach(f, channel::Channel; ntasks=Threads.nthreads())
+
+Similar to `foreach(f, channel)`, but iteration over `channel` and calls to
+`f` are split across `ntasks` Tasks spawned by `Threads.@spawn`. This function
+will wait for all internally spawned tasks to complete before returning.
+"""
```
ararslan, line 482:
```suggestion

!!! compat "Julia 1.6"
    This function requires Julia 1.6 or later.
"""
```

base/channels.jl
```diff
@@ -472,3 +472,29 @@ function iterate(c::Channel, state=nothing)
...
+function Threads.foreach(f, channel::Channel; ntasks=Threads.nthreads())
+    stop = Threads.Atomic{Bool}(false)
+    @sync for _ in 1:ntasks
```
ararslan, line 485: `@sync` isn't defined yet at the point this is loaded while building the sysimage, which is causing the builds to fail.

ararslan: `@sync` is defined in base/task.jl, which is included directly after base/channels.jl.

base/channels.jl
```diff
@@ -472,3 +472,29 @@ function iterate(c::Channel, state=nothing)
...
+        Threads.@spawn try
+            for item in channel
+                _waitspawn(f, item)
+                stop[] && break  # do this _after_ f(item) to avoid loosing `item`
```
tkf, line 489: Hmm... Actually, this is not enough because a task could be stuck on `take!` at `for item in channel`.  But I'm not sure how to terminate the tasks without closing the `channel` or loosing an `item`.  I think we need builtin "`select`" (as in Go) for this.

base/channels.jl
jrevels: I naively fixed this by introducing a new file (`threadfuncs.jl`) included after both `channels.jl` and `tasks.jl`. Hopefully isn't too controversial (though the name is bikesheddable, of course 😁).

tkf (approved):
LGTM! (But probably I shouldn't merge it myself as I suggest too many parts of the implementation.)

jrevels:
Thanks @tkf! 

I'll see if I can wrangle another set of eyes on this before merging...doesn't seem contentious, so normally I'd just do the whole "will merge myself in a few days if no other concerns" thing if it's just blocked on reviewer time, but I know there's been a lot of thought about Threads-related API messaging/stability over the past few weeks so I don't won't to blindside anybody 😅

base/threadfuncs.jl
```diff
@@ -0,0 +1,35 @@
...
+    Threads.foreach(f, channel::Channel; ntasks=Threads.nthreads())
+
+Similar to `foreach(f, channel)`, but iteration over `channel` and calls to
+`f` are split across `ntasks` Tasks spawned by `Threads.@spawn`. This function
+will wait for all internally spawned tasks to complete before returning.
+
+!!! compat "Julia 1.6"
+    This function requires Julia 1.6 or later.
+"""
+function Threads.foreach(f, channel::Channel; ntasks=Threads.nthreads())
```
tkf, line 13:
```suggestion
function Threads.foreach(f, channel::Channel; ntasks=Threads.nthreads(),
                         schedule=:default)
    apply = schedule === :static ? (f, x) -> f(x) : _waitspawn
```

(should be `scheduler`?)

base/threadfuncs.jl
```diff
@@ -0,0 +1,35 @@
...
+    stop = Threads.Atomic{Bool}(false)
+    @sync for _ in 1:ntasks
+        Threads.@spawn try
+            for item in channel
+                _waitspawn(f, item)
```
tkf, line 18:
```suggestion
                apply(f, item)
```

tkf (commented):
I think one quick way to implement `schedule` argument would be something like this.

Some characteristics I can think of:

* `:default`: Adequate when `f(item)` takes more than a few tens of micro seconds. Automatic load-balancing. Works well even there are other multi-threaded functions running.
* `:static`: Smaller overhead than `:default`. It doesn't handle thread migration (yet).

base/threadfuncs.jl
jrevels: I prefer `schedule`. 

Also how about `:fair` for the default `schedule` value vs. `:default`?

tkf: `:fair` sounds good to me.

jrevels:
if folks are curious, naively ran the following to check overhead (would be interested in seeing how it scales with threadcount/`ntasks` but don't really have time today to check):

```
julia> Threads.nthreads()
4

julia> k = 10000
10000

julia> foo(x) = nothing

julia> @benchmark Threads.foreach(foo, x; schedule=:fair) setup=(x=Channel(ch -> foreach(i -> put!(ch, i), 1:k), k)) evals=1
BenchmarkTools.Trial:
  memory estimate:  7.01 MiB
  allocs estimate:  89183
  --------------
  minimum time:     7.255 ms (0.00% GC)
  median time:      8.061 ms (0.00% GC)
  mean time:        9.511 ms (11.35% GC)
  maximum time:     71.131 ms (0.00% GC)
  --------------
  samples:          474
  evals/sample:     1

julia> @benchmark Threads.foreach(foo, x; schedule=:static) setup=(x=Channel(ch -> foreach(i -> put!(ch, i), 1:k), k)) evals=1
BenchmarkTools.Trial:
  memory estimate:  160.36 KiB
  allocs estimate:  10088
  --------------
  minimum time:     2.485 ms (0.00% GC)
  median time:      8.772 ms (0.00% GC)
  mean time:        8.796 ms (0.00% GC)
  maximum time:     13.226 ms (0.00% GC)
  --------------
  samples:          506
  evals/sample:     1
```

jrevels:
Added docs for `schedule`; still needs a test or two as well. In the meantime, how does this look @vchuravy? 

(note to onlookers: this change was motivated by a Slack convo between myself/vchuravy/tkf/vtjnash)

base/threadfuncs.jl
```diff
@@ -0,0 +1,54 @@
...
+If `schedule` is `:static`, `Threads.foreach` will simply spawn `ntasks` tasks
+such that each task is locked to the thread it was spawned on. This approach has
+lower scheduling overhead and thus may be more suitable for more uniform,
+fine-grained workloads.
+
+!!! compat "Julia 1.6"
+    This function requires Julia 1.6 or later.
+"""
+function Threads.foreach(f, channel::Channel; ntasks=Threads.nthreads(),
+                         schedule=:fair)
```
vchuravy, line 25: If the keyword is not constant propagated `apply` will be type-instable. So this relies on inlining...

Probably better to have schedule be a "trait" so that any instability is pushed to the caller

base/threadfuncs.jl
```diff
@@ -0,0 +1,54 @@
...
+                    schedule=:fair)
+
+Similar to `foreach(f, channel)`, but iteration over `channel` and calls to
+`f` are split across `ntasks` tasks spawned by `Threads.@spawn`. This function
+will wait for all internally spawned tasks to complete before returning.
+
+If `schedule` is `:fair`, `Threads.foreach` will spawn a new task for each work
+item, allowing Julia's scheduler to more freely load-balance work items across
+threads. Technically, in `:fair` mode, `Threads.foreach` may execute up to
+`2 * ntasks` tasks concurrently.
```
tkf, line 14: I think it'd be better to treat one-task-per-item as an implementation detail.  It may be possible to get this without the re-spawn trick at some point #35688.  Maybe just say `:fair` tries to load-balance but it comes with somewhat high-cost per item? I think it also makes sense to explain it as implementation detail to give the users some idea about the overhead.

base/threadfuncs.jl
```diff
@@ -0,0 +1,54 @@
...
+    if schedule === :fair
+        apply = _waitspawn
+    elseif schedule === :static
+        apply = (f, x) -> f(x)
+    else
+        throw(ArgumentError("`schedule` must be either `:fair` or `:static`"))
+    end
+    stop = Threads.Atomic{Bool}(false)
+    @sync for _ in 1:ntasks
+        Threads.@spawn try
```
vchuravy, line 35: This is prone to a pathological case where each task is first picked up by the same worker-thread and is subsequently pinned there. For both schedules it is important that these are not ending up on the same thread, but the `:static` schedule will degrade quicker in performance.

base/threadfuncs.jl
```diff
@@ -0,0 +1,54 @@
...
+            for item in channel
+                apply(f, item)
```
vchuravy, line 37:
```suggestion
                $apply(f, item)
```

Looking at:

```
function g(cond)
         if cond
           apply = Core._apply
         else
           apply = (f, x) -> f(x)
         end
         @sync for i in 1:2
           @async begin
             apply(print, $i)
           end
         end
       end
```

We are boxing `apply`, due to the conditional.

base/threadfuncs.jl
tkf, line 35: Yet another note for future improvement: I think, in principle, we can use `jl_set_task_tid` here. However, it'd freeze at `@sync` if there are long-running task on some threads without yields. Leaking a task for such a case could be an option but it would pile up in the task queue so I don't think it's a good solution. I think we need a task cancellation API for this, too.

base/threadfuncs.jl
tkf: I thought it won't matter match as `Task` needs dynamic dispatch anyway?

jrevels: Yeah, I figured dispatch/spawn overhead would be high enough here that it wouldn't matter, but could definitely be wrong. Easy to add some `FairSchedule`, `StaticSchedule`, etc. traits to `Threads` if it's indeed useful. API-wise, it seems like we should ensure there's parity with the `:static` argument to `@threads`, but could also just use a singleton to make them look the same to users e.g. `const static = StaticSchedule()`. Then make this function signature `Threads.foreach(f, channel::Channel, ::AbstractSchedule)` and call it like `Threads.foreach(f, channel, Threads.static)`...actually I kinda hate that now that I've written it out lol; seems more transparent/readable just to stick with `Threads.foreach(f, channel, Threads.StaticSchedule())`.

Anyway, could add those traits here if useful.

base/threadfuncs.jl
jrevels: Edited the docstring to focus more on the tradeoffs and less on implementation details 👍

base/threadfuncs.jl
jrevels: I guess for the static schedule, this per-item dispatch overhead could accumulate a bit over the course of the loop. I'll implement these traits, then.

jrevels: Schedule selection is now accomplished via traits instead of `Symbol`s

jrevels:
Docs/tests updated with `FairSchedule` vs. `StaticSchedule` selection, and additional suggestions from @vchuravy/@tkf. 

I'm not sure of a good way to cleanly help the scheduler avoid the pathological case @vchuravy mentioned; any workaround that involved manual task management seems like it'd be the kind of scheduler-sidestepping that would just leave the caller worse-off in practice, no? 

It'd be nice if we could express the intent here to the scheduler without manually forcing any given task assignment, and let it otherwise make good decisions for us. I'm not super familiar with the underlying task code, but does the `jl_set_task_tid` thing @tkf suggested essentially accomplish this if `sticky == false` e.g. does it set the task to `tid` but allow the scheduler to migrate it to a new thread if so desired? The scheduler isn't actually able to migrate tasks yet, though, right? 

Anyway, as long as this API seems solid and like it wouldn't need to break to enable future improvements, then I'm good with the state of this branch as is. Re-requested review from @vchuravy and @tkf (thanks to both for all your help/insight!)

test/threads_exec.jl
```diff
@@ -845,3 +845,32 @@ fib34666(x) =
...
+        g = i -> begin
+            iseven(i) && sleep(delay)
+            put!(ch, f(i))
+        end
+        Threads.foreach(g, x, schedule; ntasks=ntasks)
+    end
+    return y
+end
+
+@testset "Threads.foreach(f, ::Channel)" begin
```
tkf, line 861: I think it'd be nice to test the default keyword argument (e.g., a typo may can be introduced with some edits in the future). Maybe something like

```julia
ys = Channel() do ys
    Threads.foreach(Channel(xs -> foreach(i -> put!(xs, i), 1:3))) do x
        put!(ys, x)
    end
end
@test sort!(collect(ys)) == 1:3
```

tkf (commented):
LGTM other than the minor point about the test.

base/threads_overloads.jl
```diff
@@ -0,0 +1,51 @@
...
+that incurs lower per-item overhead than `FairSchedule`, but is less amenable
+to load-balancing. This approach thus may be more suitable for fine-grained,
+uniform workloads, but may perform worse than `FairSchedule` in concurrence
+with other multithreaded workloads.
+
+!!! compat "Julia 1.6"
+    This function requires Julia 1.6 or later.
+"""
+function Threads.foreach(f, channel::Channel,
+                         schedule::Threads.AbstractSchedule=Threads.FairSchedule();
```
tkf, line 27: Hmm... How about making `schedule` a keyword argument? I think we might want to have `Threads.foreach(f, arrays...)` at some point? It'd be a bit hard to implement robust dispatch when schedule is at the end of vararg function.

test/threads_exec.jl
jrevels: Done.

base/threads_overloads.jl
jrevels: Done.

tkf (approved):

jrevels:
Assuming no further feedback, it seems like `Threads.foreach` has iterated towards a fairly uncontroversial state by this point. I'll plan on merging Thursday morning-ish unless there are further concerns; would still very much welcome additional review if folks see something over the next couple days that jumps out at them. 🙂

Issue #34547, opened 2020-01-28
inconsistent handling of abstract types in NamedTuple


stevengj:
I noticed the following seeming inconsistency:
```jl
julia> (3,4) isa Tuple{Number,Number}
true

julia> (a=3,b=4) isa NamedTuple{(:a,:b), Tuple{Number,Number}}
false
```
You can use `<:Tuple` for the second argument for subtyping, but even that doesn't work as a constructor:
```jl
julia> (a=3,b=4) isa NamedTuple{(:a,:b), <:Tuple{Number,Number}}
true

julia> NamedTuple{(:a,:b), <:Tuple{Number,Number}}((3,4))
ERROR: MethodError: no method matching NamedTuple{(:a, :b),#s1} where #s1<:Tuple{Number,Number}(::Tuple{Int64,Int64})
Stacktrace:
 [1] top-level scope at REPL[9]:1
```
(Of course, ~~`Tuple{Number,Number}(3,4)` doesn't work~~ either.  Correction: `Tuple{Number,Number}((3,4))` is the correct spelling and works.)

This makes it a bit hard to use a `NamedTuple` declaration with abstract types for the fields.  Proposal:

1. `(a=3,b=4) isa NamedTuple{(:a,:b), Tuple{Number,Number}}` should return `true`, and in general it should work like `isa Tuple{...}`.

2. `Tuple{Number,Number}` and `NamedTuple{(:a,:b), Tuple{Number,Number}}` should also work as constructors, taking the concrete type from their argument types.

In theory these are breaking, but in practice I think it should qualify as a minor change.  `NamedTuple` with abstract types is basically useless right now, and (2) throws an exception, so it is hard to imagine any working code depending on the current behaviors.

quinnj:
Ref https://github.com/JuliaLang/julia/issues/32699, which is about making abstractly inferred NamedTuples useful

JeffBezanson:
Making `NamedTuple` covariant has been discussed before, and the current thinking is that it's probably a bad idea. In fact the opposite change, making tuples invariant, is on the table: #24614. For tabular data, we want a named tuple with a `Union{Int,Missing}` element type to be concrete.

The `Tuple` constructor already exists:
```
julia> Tuple{Number,Number}((1,2))
(1, 2)

julia> typeof(ans)
Tuple{Int64,Int64}
```

The NamedTuple constructor behaves as we need:
```
[...]

PR #34548, opened 2020-01-28
RFC: NamedTuple macro for easier type construction

stevengj:
Following #34498, #34505, and [a recent discourse discussion](https://discourse.julialang.org/t/anonymous-nested-struct/33723/9), this PR defines (and exports) a macro `@NamedTuple` that provides a more convenient `struct`-like syntax for defining `NamedTuple` types:
```jl
julia> @NamedTuple{a::Int, b::String}
NamedTuple{(:a, :b),Tuple{Int64,String}}

julia> @NamedTuple begin
           a::Int
           b::String
       end
NamedTuple{(:a, :b),Tuple{Int64,String}}
```
If a `::Type` is not specified, it is treated as `::Any` (similar to a `struct`).   ~~This will only be useful if #34547 is resolved, however.  (Alternatively, the PR could be changed to require a type specifier for all keys.)~~  Abstractly typed fields of `NamedTuple` work fine.

### DIFF
--- a/base/namedtuple.jl
+++ b/base/namedtuple.jl
@@ -337,18 +337,18 @@ If the `::Type` declaration is omitted, it is taken to be `Any`.   The `begin ..
 declarations to be split across multiple lines (similar to a `struct` declaration), but is otherwise
 equivalent.
 
-For example, the tuple `(a=3, b="hello")` has a type `NamedTuple{(:a, :b),Tuple{Int64,String}}`, which
+For example, the tuple `(a=3.1, b="hello")` has a type `NamedTuple{(:a, :b),Tuple{Float64,String}}`, which
 can also be declared via `@NamedTuple` as:
 
 ```jldoctest
-julia> @NamedTuple{a::Int, b::String}
-NamedTuple{(:a, :b),Tuple{Int64,String}}
+julia> @NamedTuple{a::Float64, b::String}
+NamedTuple{(:a, :b),Tuple{Float64,String}}
 
 julia> @NamedTuple begin
-           a::Int
+           a::Float64
            b::String
        end
-NamedTuple{(:a, :b),Tuple{Int64,String}}
+NamedTuple{(:a, :b),Tuple{Float64,String}}
 ```
 """
 macro NamedTuple(ex)
--- a/base/namedtuple.jl
+++ b/base/namedtuple.jl
@@ -358,6 +358,6 @@ macro NamedTuple(ex)
     all(e -> e isa Symbol || Meta.isexpr(e, :(::)), decls) ||
         throw(ArgumentError("@NamedTuple must contain a sequence of name or name::type expressions"))
     vars = [QuoteNode(e isa Symbol ? e : e.args[1]) for e in decls]
-    types = [e isa Symbol ? :Any : e.args[2] for e in decls]
+    types = [esc(e isa Symbol ? :Any : e.args[2]) for e in decls]
     return :(NamedTuple{($(vars...),), Tuple{$(types...)}})
 end
--- a/base/namedtuple.jl
+++ b/base/namedtuple.jl
@@ -17,6 +17,8 @@ using [`values`](@ref).
     Iteration over `NamedTuple`s produces the *values* without the names. (See example
     below.) To iterate over the name-value pairs, use the [`pairs`](@ref) function.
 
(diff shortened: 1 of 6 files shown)


StefanKarpinski:
Might be even better to just have real syntax for this...

stevengj:
> Might be even better to just have real syntax for this...

I don't think we would want `NamedTuple{a::Int, b::String}` (without the `@`) to work, since these aren't "really" the type parameters.    Or did you have something else in mind?

StefanKarpinski:
🤷‍♂ maybe the macro syntax is for the best.

base/namedtuple.jl
```diff
@@ -324,3 +326,38 @@ julia> Base.setindex(nt, "a", :a)
...
+```
+"""
+macro NamedTuple(ex)
+    Meta.isexpr(ex, :braces) || Meta.isexpr(ex, :block) ||
+        throw(ArgumentError("@NamedTuple expects {...} or begin...end"))
+    decls = filter(e -> !(e isa LineNumberNode), ex.args)
+    all(e -> e isa Symbol || Meta.isexpr(e, :(::)), decls) ||
+        throw(ArgumentError("@NamedTuple must contain a sequence of name or name::type expressions"))
+    vars = [QuoteNode(e isa Symbol ? e : e.args[1]) for e in decls]
+    types = [e isa Symbol ? :Any : e.args[2] for e in decls]
```
tkf: I guess you need `esc` to make it work for types defined outside `Base`?

```suggestion
    types = [esc(e isa Symbol ? :Any : e.args[2]) for e in decls]
```

tkf:
Is it too evil to use `@NamedTuple` in `show`?  Maybe at least for text/plain?

base/namedtuple.jl
stevengj: You don't need `esc` just to *read* external symbols, right?  Macro hygiene only rewrites variables that you assign to.

However, I guess we should put `esc` here out of paranoia that someone does something bananas like `NamedTuple{x::begin; someglobal += 1; Int}`. 😆

stevengj:
> Is it too evil to use `@NamedTuple` in `show`? 

I vote for too evil, at least for now.  We can reconsider later.

tkf:
Re https://github.com/JuliaLang/julia/pull/34548#discussion_r373283808 I'm seeing the effect of `esc`:

```julia
julia> module Demo1
       macro NamedTuple(ex)
           Meta.isexpr(ex, :braces) || Meta.isexpr(ex, :block) ||
               throw(ArgumentError("@NamedTuple expects {...} or begin...end"))
           decls = filter(e -> !(e isa LineNumberNode), ex.args)
           all(e -> e isa Symbol || Meta.isexpr(e, :(::)), decls) ||
               throw(ArgumentError("@NamedTuple must contain a sequence of name or name::type expressions"))
           vars = [QuoteNode(e isa Symbol ? e : e.args[1]) for e in decls]
           types = [esc(e isa Symbol ? :Any : e.args[2]) for e in decls]
           return :(NamedTuple{($(vars...),), Tuple{$(types...)}})
       end
       end  # module Demo1
Main.Demo1

julia> module Demo2
       macro NamedTuple(ex)
           Meta.isexpr(ex, :braces) || Meta.isexpr(ex, :block) ||
               throw(ArgumentError("@NamedTuple expects {...} or begin...end"))
           decls = filter(e -> !(e isa LineNumberNode), ex.args)
           all(e -> e isa Symbol || Meta.isexpr(e, :(::)), decls) ||
               throw(ArgumentError("@NamedTuple must contain a sequence of name or name::type expressions"))
           vars = [QuoteNode(e isa Symbol ? e : e.args[1]) for e in decls]
           types = [e isa Symbol ? :Any : e.args[2] for e in decls]
           return :(NamedTuple{($(vars...),), Tuple{$(types...)}})
       end
       end  # module Demo2
Main.Demo2

julia> struct Foo end

julia> Demo1.@NamedTuple{a::Foo}
NamedTuple{(:a,),Tuple{Foo}}

julia> Demo2.@NamedTuple{a::Foo}
ERROR: UndefVarError: Foo not defined
```

Maybe this'd be a nice test case?

stevengj:
win32 test failures seem unrelated (failure in cmdlineargs and downloading curl).

JeffBezanson:
:+1: from triage.

PR #34554, opened 2020-01-28
test: fix pool size assumption

vtjnash:
On 32-bit, this object falls into the 12-byte pool, which does not exist
on some platforms (such as Win32). Query for which pool this will land
into and test the track-allocation result accordingly.


vtjnash:
https://build.julialang.org/#/builders/65/builds/7445 seems to be just the usually httpbin test flakiness, restarting it

PR #34561, opened 2020-01-28
Merge pull request #34561 from JuliaLang/sf/bump_openblas_version

staticfloat:
This should fix FreeBSD issues.


staticfloat:
@KristofferC I've opened this PR directly against `release-1.4`, is that the correct way to target 1.4 without going through `master` first?  Also, the windows failures look like they're due to us needing to backport https://github.com/JuliaLang/julia/pull/34506.  I'll open a PR to do that first.

KristofferC:
>  I've opened this PR directly against release-1.4, is that the correct way to target 1.4 without going through master first?

I would prefer if everything went through the backports branch (because that is what we run PkgEval on before merging it into the release branch). You could push this directly to the backport branch or change the merge branch to it.

PR #34566, opened 2020-01-29
soft scope: fix doc tests

StefanKarpinski:
In the original PR, these were hidden by the fact that a stray leading space caused what was considered code and not to be messed up for the second half of this file. Then the PR that fixed it didn't have CI run (not sure why). This fixes the doc tests but ideally should be reverted once Documenter supports the new soft scope behavior correctly.


StefanKarpinski:
@staticfloat, any idea why the doctest buildbot isn't running?

staticfloat:
I'm not totally sure why yet, but the buildbots seem to be really swamped today.  It's taking a while for a runner to catch up to the queue and report status back to GitHub.

PR #34568, opened 2020-01-29
Mention that `string(message)` is used for formatting log messages

tkf:
As discussed in https://github.com/JunoLab/ProgressLogging.jl/pull/13#discussion_r371113550

cc @c42f

### DIFF
--- a/stdlib/Logging/docs/src/index.md
+++ b/stdlib/Logging/docs/src/index.md
@@ -73,7 +73,7 @@ automatically extracted. Let's examine the user-defined data first:
 * The *message*  is an object describing the event. By convention
   `AbstractString`s passed as messages are assumed to be in markdown format.
   Other types will be displayed using `print(io, obj)` or `string(obj)` for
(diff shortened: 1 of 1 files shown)


stdlib/Logging/docs/src/index.md
```diff
@@ -72,8 +72,9 @@ automatically extracted. Let's examine the user-defined data first:
 
 * The *message*  is an object describing the event. By convention
   `AbstractString`s passed as messages are assumed to be in markdown format.
-  Other types will be displayed using `show(io,mime,obj)` according to the
-  display capabilities of the installed logger.
+  Other types will be displayed using `print(io, obj)` or `string(obj)` for
+  text-based logger and possibly `show(io,mime,obj)` for other multimedia
```
c42f: To make this more grammatical:

```suggestion
  text-based output and possibly `show(io,mime,obj)` for other multimedia
```

c42f (approved):
Yes, I think this is correct for the reasons stated in the linked issue. And also because this is what we actually do in practice!

PR #34573, opened 2020-01-29
make package cache interit permission from source

### LINKED ISSUE #25971: Permission of precompiled files
Hello.
I installed julia packages with root privileges at /opt/julia by setting JULIA_PKGDIR=/opt/julia. Permission of the directory is 755.

I executed `using PackageName` as root after installing packages. However, default permissions of the files in /opt/julia/lib/v-0.x is 600. So I have to change these to 644 manually if I want to use packeges as non-root user. Permissions of other files that in /opt/julia/v-0.x is 644.

bicycle1885:
This fixes #25971.

### DIFF
--- a/test/precompile.jl
+++ b/test/precompile.jl
@@ -815,6 +815,8 @@ let
     finally
         rm(load_path, recursive=true)
         rm(load_cache_path, recursive=true)
+        filter!((≠)(load_path), LOAD_PATH)
+        filter!((≠)(load_cache_path), DEPOT_PATH)
     end
 end
 
--- a/test/precompile.jl
+++ b/test/precompile.jl
@@ -797,4 +797,25 @@ let
     end
 end
 
+# Issue #25971
+let
+    load_path = mktempdir()
+    load_cache_path = mktempdir()
+    try
+        pushfirst!(LOAD_PATH, load_path)
+        pushfirst!(DEPOT_PATH, load_cache_path)
+        sourcefile = joinpath(load_path, "Foo25971.jl")
+        write(sourcefile, "module Foo25971 end")
+        chmod(sourcefile, 0o666)
+        cachefile = Base.compilecache(Base.PkgId("Foo25971"))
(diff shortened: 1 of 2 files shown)


test/precompile.jl
```diff
@@ -797,4 +797,25 @@ let
...
+        write(sourcefile, "module Foo25971 end")
+        chmod(sourcefile, 0o666)
+        cachefile = Base.compilecache(Base.PkgId("Foo25971"))
+        @test filemode(sourcefile) == filemode(cachefile)
+        chmod(sourcefile, 0o600)
+        cachefile = Base.compilecache(Base.PkgId("Foo25971"))
+        @test filemode(sourcefile) == filemode(cachefile)
+    finally
+        rm(load_path, recursive=true)
+        rm(load_cache_path, recursive=true)
```
jonas-schulze: Even though CI is passing, should the updates to `LOAD_PATH` and `DEPOT_PATH` be reversed again?

```suggestion
        rm(load_cache_path, recursive=true)
        filter!(s -> s != load_path, LOAD_PATH)
        filter!(s -> s != load_cache_path, DEPOT_PATH)
```

StefanKarpinski: I should suggest using `popfirst!(LOAD_PATH)` instead.

jonas-schulze: I've thought about that, too, but using `filter!` would be more robust in case of a failure in between or even before the `pushfirst!`s, or should the tests be run concurrently some time in the future.

KristofferC: Another way is to `copy` before running the code and then `copy!` it back when done.

bicycle1885: Removed those paths as @jonas-schulze suggested. Thank you.

KristofferC: FWIW, `mktempdir` now automatically deletes the folder when the process exits.

StefanKarpinski: Still a best practice to delete when done using.

PR #34575, opened 2020-01-29
Use norm instead of abs in generic lu factorization

mschauer:
Use norm instead of abs in generic lu factorization as discussed in JuliaLang/LinearAlgebra.jl#693. The factorisation then works also with tensors represented by matrices with elements which are matrix like.


mschauer:
This makes sense, but the fact that it broke `generic` because `generic` had a type without `norm` or `iteratate` defined suggest that one should see how it affects the registered packages.

mschauer:
Bump.

PR #34581, opened 2020-01-29
[release-1.4] tests: Use `realpath()` before `==` path comparisons (#34506)

staticfloat:
This should fix CI issues due to windows buildbot environment changes.


staticfloat:
@KristofferC Is this correct?  If so, I'll merge this then rebase https://github.com/JuliaLang/julia/pull/34561 on top of it.

PR #34583, opened 2020-01-29
handle masked_gather in late-gc-lowering

### LINKED ISSUE #34569: sum(::Diagonal, dims=1) crashes Julia
While trying to debug #34553, I thought I'd test on an even simpler structured matrix, a `Diagonal`. Turns out to be a very bad idea. :-(

```julia
julia> versioninfo()
Julia Version 1.4.0-rc1.0
Commit b0c33b0cf5 (2020-01-23 17:23 UTC)
Platform Info:
  CPU: Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
Environment:
  JULIA_EDITOR = atom  -a
  JULIA_NUM_THREADS = 4

julia> using LinearAlgebra

julia> D = Diagonal(ones(1000));

julia> sum(Matrix(D), dims=1)
1×1000 Array{Float64,2}:
 1.0  1.0  1.0  1.0  1.0  1.0  1.0  1.0  …  1.0  1.0  1.0  1.0  1.0  1.0  1.0

julia> sum(D, dims=1)
  %wide.masked.gather297 = call <4 x double addrspace(13)*> @llvm.masked.gather.v4p13f64.v4p11p13f64(<4 x double addrspace(13)* addrspace(11)*> %broadcast.splat296, i32 8, <4 x i1> %102, <4 x double addrspace(13)*> undef), !dbg !193, !tbaa !114Unexpected generating operation for derived values
UNREACHABLE executed at /Users/sabae/buildbot/worker/package_macos64/build/src/llvm-late-gc-lowering.cpp:842!

signal (6): Abort trap: 6
__pthread_kill at /usr/lib/system/libsystem_kernel.dylib (unknown line)
Allocations: 26682699 (Pool: 26676228; Big: 6471); GC: 29
```
Also crashes on master, but works fine on Julia v1.3.1 and v1.0.5.

vchuravy:
fixes #34569

### DIFF
--- a/src/llvm-late-gc-lowering.cpp
+++ b/src/llvm-late-gc-lowering.cpp
@@ -474,21 +474,49 @@ static std::pair<Value*,int> FindBaseValue(const State &S, Value *V, bool UseCac
             CurrentV = EEI->getVectorOperand();
         }
         else if (auto LI = dyn_cast<LoadInst>(CurrentV)) {
-            if (auto PtrT = dyn_cast<PointerType>(LI->getType())) {
+            if (auto PtrT = dyn_cast<PointerType>(LI->getType()->getScalarType())) {
                 if (PtrT->getAddressSpace() == AddressSpace::Loaded) {
                     CurrentV = LI->getPointerOperand();
+                    fld_idx = -1;
                     if (!isSpecialPtr(CurrentV->getType())) {
-                        // Special case to bypass the check below.
                         // This could really be anything, but it's not loaded
                         // from a tracked pointer, so it doesn't matter what
-                        // it is.
-                        return std::make_pair(CurrentV, fld_idx);
+                        // it is--just pick something simple.
+                        CurrentV = ConstantPointerNull::get(Type::getInt8PtrTy(V->getContext()));
                     }
                     continue;
                 }
             }
             // In general a load terminates a walk
             break;
+        }
+        else if (auto II = dyn_cast<IntrinsicInst>(CurrentV)) {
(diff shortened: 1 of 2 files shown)


src/llvm-late-gc-lowering.cpp
```diff
@@ -828,6 +828,27 @@ std::vector<int> LateLowerGCFrame::NumberAllBase(State &S, Value *CurrentV) {
                 Numbers.push_back(BaseNumbers[i]);
         }
         assert(CountTrackedPointers(EVI->getType()).count == Numbers.size());
+    } else if (auto *II = dyn_cast<IntrinsicInst>(CurrentV)) {
+        switch (II->getIntrinsicID()) {
+            case Intrinsic::masked_gather: {
+                Numbers = NumberAll(S, II->getOperand(0));
```
vchuravy: I am simply forwarding the operand here, unsure if that is the right way of doing things, additionally should we handle
other masked intrinsics like `masked.scatter` and `masked.load`?

According to the Langref https://llvm.org/docs/LangRef.html#llvm-masked-gather-intrinsics there is also a passthrough
argument which would be a tracked pointer (or in this case it was undef).

vtjnash: No, it’s equivalent to a load, so they’re both a normal generating instruction (just like a regular CallInst)

KristofferC:
Bump, can this be merged?

PR #34587, opened 2020-01-30
Add Libc.mkfifo

tkf:
This PR adds a `mkfifo` wrapper `Libc.mkfifo` for Unix platforms.

Python has it in [`os` stdlib](https://docs.python.org/3/library/os.html#os.mkfifo) and it's useful sometimes.  I think it'd be nice if Julia has it in Base/stdlib too.


base/libc.jl
```diff
@@ -10,6 +10,9 @@ import Core.Intrinsics: bitcast
 
 export FILE, TmStruct, strftime, strptime, getpid, gethostname, free, malloc, calloc, realloc,
     errno, strerror, flush_cstdio, systemsleep, time, transcode
+if Sys.isunix()
```
vtjnash, line 13: I think this can be unconditional

vtjnash (approved):

base/libc.jl
tkf: I thought to make it consistent with

https://github.com/JuliaLang/julia/blob/4fa8055225dabd6619f6e70a10c1e36cb2282cc9/base/libc.jl#L16-L18

just below.  Also, it'd make `mkfi` <kbd>TAB</kbd> completes `mkfifo` in Windows while accessing to it throws:

```julia
julia> module Demo
       export foo
       end
Main.Demo

julia> using .Demo

julia> fo  # TAB
foldl   foldr    foo      for      foreach
julia> foo
ERROR: UndefVarError: foo not defined
```

So, I thought it's better to hide it in Windows.

vtjnash: It's defined though, just has no methods.

tkf: Ah, yes, you are right.  But what about the consistency with `GetLastError` and `FormatMessage`?  I'm OK with unconditional `export` but I just don't see a big reason to break the pattern.

StefanKarpinski:
Not sure why this is a release blocker?

base/libc.jl
vtjnash: I'm okay with those being different

base/libc.jl
```diff
@@ -400,4 +403,27 @@ Interface to the C `srand(seed)` function.
...
+    mkfifo(path::AbstractString, [mode::Integer]) -> path
+
+Make a FIFO special file (a named pipe) at `path`.  Return `path` as-is on success.
+
+!!! compat "Julia 1.5"
+    `mkfifo` requires at least Julia 1.5.
+"""
+function mkfifo end
+
+if Sys.isunix()
```
JeffBezanson, line 416: Maybe better to move this check inside the method and give a clearer "not supported on this platform" error instead of a MethodError?

tkf: I actually thought it's a nice pattern to not provide the method at all so that you can use `hasmethod` etc. to check if some API is implemented in a given platform. I thought it's nicer than each caller to encode such information in their code.

JeffBezanson: That's an interesting point, but I don't think `hasmethod` can be used consistently to check for platform support. For example, there could be cases where certain argument values aren't supported on a certain OS.

tkf: OK. It makes sense.

tkf:
@vtjnash @JeffBezanson Thanks for the review. I incorporated your comments.

StefanKarpinski:
Looking good... All that green! @JeffBezanson want to merge if you're happy here?

base/libc.jl
```diff
@@ -400,4 +400,27 @@ Interface to the C `srand(seed)` function.
 """
 srand(seed=floor(Int, time()) % Cuint) = ccall(:srand, Cvoid, (Cuint,), seed)
 
+"""
+    mkfifo(path::AbstractString, [mode::Integer]) -> path
+
+Make a FIFO special file (a named pipe) at `path`.  Return `path` as-is on success.
```
musm, line 406: Probably should also mention that this is a Unix only utiliy

musm:
Just needs a rebase and to change the error message to   ENOT_SUP: kernel does not support fifo

base/libc.jl
```diff
@@ -400,4 +400,27 @@ Interface to the C `srand(seed)` function.
...
+    mode::Integer = Base.S_IRUSR | Base.S_IWUSR | Base.S_IRGRP | Base.S_IWGRP |
+                    Base.S_IROTH | Base.S_IWOTH,
+)
+    @static if Sys.isunix()
+        # Default `mode` is compatible with `mkfifo` CLI in coreutils.
+        ret = ccall(:mkfifo, Cint, (Cstring, Base.Cmode_t), path, mode)
+        systemerror("mkfifo", ret == -1)
+        return path
+    else
+        error("not supported on this platform")
```
vtjnash, line 422:
```suggestion
        systemerror("mkfifo", ENOTSUP)
```

vtjnash (approved):

base/libc.jl
```diff
@@ -402,6 +402,31 @@ Interface to the C `srand(seed)` function.
...
+    mode::Integer = Base.S_IRUSR | Base.S_IWUSR | Base.S_IRGRP | Base.S_IWGRP |
+                    Base.S_IROTH | Base.S_IWOTH,
+)
+    @static if Sys.isunix()
+        # Default `mode` is compatible with `mkfifo` CLI in coreutils.
+        ret = ccall(:mkfifo, Cint, (Cstring, Base.Cmode_t), path, mode)
+        systemerror("mkfifo", ret == -1)
+        return path
+    else
+        systemerror("mkfifo", ENOTSUP)
```
tkf, line 426: BTW, I noticed that `ENOTSUP` doesn't exist in my local build. I guess it's OK as long as this is defined in Windows, though.

vtjnash: Looks like Linux defines EOPNOTSUP instead, and our regex isn't smart enough to pick up the value for ENOTSUP from that. But CI seems to be okay with this.

tkf: It seems Windows prints `ENOTSUP` as "Unknown error":

>     Error in testset file:
>     Test Failed at C:\buildbot\worker-tabularasa\tester_win64\build\share\julia\test\file.jl:1635
>       Expression: Libc.mkfifo(joinpath(pwd(), "dummy_path"))
>         Expected: "mkfifo: Operation not supported"
>          Message: "SystemError: mkfifo: Unknown error"
>
> --- https://build.julialang.org/#/builders/65/builds/5360

So, I suggest simply doing:

https://github.com/JuliaLang/julia/blob/02961cc2b25031cb83628342f4471adbc55c9cf1/base/libc.jl#L426-L428

Is it OK?

MagicMuscleMan:
As I just needed this functionality, I was going to reinvent the wheel when I stumbled across this PR.

I do not understand why it hasn't been merged. Is there a blocker in general? Otherwise, if we prepared that for 1.10 or 1.11, would it be mergeable?

vtjnash:
I just got lost because the merge me tag was removed. Putting that back now.

base/libc.jl
```diff
@@ -437,6 +437,33 @@ function srand(seed::Integer=_make_uint64_seed())
...
 end
 
+"""
+    mkfifo(path::AbstractString, [mode::Integer]) -> path
+
+Make a FIFO special file (a named pipe) at `path`.  Return `path` as-is on success.
+
+`mkfifo` is supported only in Unix platforms.
+
+!!! compat "Julia 1.8"
```
pthariensflame, line 447: Has this been corrected?  Obviously this merge was *a bit* late for 1.8.

IanButterworth: Good catch. Please PR to fix

PR #34588, opened 2020-01-30
Various grammar changes

Mohit-Nain:
Since sounds more appropriate than while in line 129

### DIFF
--- a/doc/src/manual/integers-and-floating-point-numbers.md
+++ b/doc/src/manual/integers-and-floating-point-numbers.md
@@ -99,8 +99,8 @@ julia> UInt
 UInt64
 ```
 
-Larger integer literals that cannot be represented using only 32 bits but can be represented in
-64 bits always create 64-bit integers, regardless of the system type:
+Larger integer literals that cannot be represented using only 32 bits, but can be represented in
+64 bits, always create 64-bit integers, regardless of the system type:
 
 ```jldoctest
 # 32-bit or 64-bit system:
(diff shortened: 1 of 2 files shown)


KristofferC:
Hm, I personally feel the "while" works well because it is sort of saying that even though there are no hard rules, there are still some conventions. But the conventions are not a direct consequence of the lack of rules... Maybe :)

doc/src/manual/integers-and-floating-point-numbers.md
```diff
@@ -99,16 +99,16 @@ julia> UInt
...
+64 bits, always create 64-bit integers, regardless of the system type:
 
 ```jldoctest
 # 32-bit or 64-bit system:
 julia> typeof(3000000000)
 Int64
 ```
 
-Unsigned integers are input and output using the `0x` prefix and hexadecimal (base 16) digits
+Unsigned integers are used as input and output using the `0x` prefix and hexadecimal (base 16) digits
```
rfourquet: I was understanding `input` and `output` as verbs, so the initial version looks better to me. Moreover, in this new version, "used" and "using" are used too close in the same sentence.

mcabbott: You could say something like "are typed in and printed out using the"...

StefanKarpinski: "typed" is pretty overloaded in the programming context. I think the original is quite clear.

KristofferC:
Has some issues that have been brought up in the comments. Feel free to open a new PR with these fixed.

PR #34593, opened 2020-01-30
Add some examples to `in` docstring

logankilpatrick:
Added a simple example to check if something isn't in a collection without using a special character.


logankilpatrick:
I decided to link the "not in" section and provide an example here because I knew that `in` existed and could find it in the docs, but when I typed in `not in` I got a bunch of unhelpful results in the docs. Hopefully, this will help some people.

PR #34595, opened 2020-01-30
mapexpr argument for include: transform each parsed expression

stevengj:
This PR adds an optional `mapexpr` ~~keyword~~ first argument to `include` and `include_string`, allowing you to specify a function that rewrites each expression after it is parsed.    I've needed this functionality several times, e.g. in SoftScope.jl and ChangePrecision.jl, and it is tricky implement manually in a way that preserves line-number information and other semantics of `include`.

Moreover, now that #33864 has merged, I would like to use the "official" `REPL.softscope!` function in IJulia instead of SoftScope.jl, or equivalently change `SoftScope.jl` to use this function on newer Julia releases.   With this PR, one can just pass `REPL.softscope!` to `include_string`.

To do:
- [x] Tests
- [x] Documentation
- [x] Benchmarking.  Invoking the identity transformation `x->x` adds a 2–3% overhead to `include_string` measured for `"nothing\n"^10000`,  but I eliminated this by optimizing the default case of `mapexpr === identity`.

### DIFF
--- a/base/Base.jl
+++ b/base/Base.jl
@@ -365,8 +365,8 @@ for m in methods(include)
 end
 # These functions are duplicated in client.jl/include(::String) for
 # nicer stacktraces. Modifications here have to be backported there
-include(mod::Module, path::AbstractString) = include(mod, convert(String, path))
-function include(mod::Module, _path::String; mapexpr::Function=identity)
+include(mod::Module, _path::AbstractString) = include(identity, mod, _path)
+function include(mapexpr::Function, mod::Module, _path::AbstractString)
     path, prev = _include_dependency(mod, _path)
     for callback in include_callbacks # to preserve order, must come before Core.include
         invokelatest(callback, mod, path)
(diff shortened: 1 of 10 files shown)


base/Base.jl
```diff
@@ -364,7 +364,7 @@ end
 # These functions are duplicated in client.jl/include(::String) for
 # nicer stacktraces. Modifications here have to be backported there
 include(mod::Module, path::AbstractString) = include(mod, convert(String, path))
-function include(mod::Module, _path::String)
+function include(mod::Module, _path::String; mapexpr::Function=identity)
```
StefanKarpinski: I realize this is a bit of a corner case to call it like this, but perhaps the function argument should be the optional first one?

stevengj: Sure, that might be nice.

stevengj:
@JeffBezanson, where is a module's `include(x)` function defined?   I need to add an `include(mapexpr, ex)` method there…

StefanKarpinski:
I like the mapping function as the first argument. What else would one pass there, after all?

stevengj:
If I do
```jl
julia> module Foo; end; methods(Foo.include)
# 1 method for generic function "include":
[1] include(x) in Main.Foo at REPL[1]:1
```
where is `Foo.include(x)` defined?  I'm having trouble tracking it down.

stevengj:
Rebased.  Is there anything else that is needed here?

StefanKarpinski:
I think only for @JeffBezanson to have a look. He was traveling last week so a bit occupied. He'll probably be able to take a look and merge if he approves this week.

PR #34596, opened 2020-01-30
add some helpful type information to Base and REPLCompletions

### LINKED ISSUE #34098: Loading CSV makes tab completion in the REPL extremely slow
Doing `using CSV` and then `foo("<TAB` in the REPL makes the REPL lag for ~10 seconds. Running with `--trace-compile` and timing the different precompile statement there is one that takes 10(!) seconds to run. MWE is:

```
julia> using CSV; using REPL; @time precompile(Tuple{typeof(REPL.REPLCompletions.completions), String, Int64})
 10.879856 seconds (25.56 M allocations: 1.508 GiB, 3.41% gc time)
```

Entering the Pkg REPL also takes like 20 seconds after using CSV.

JeffBezanson:
fixes #34098

### DIFF
--- a/base/client.jl
+++ b/base/client.jl
@@ -427,7 +427,7 @@ using ..Base
 # include(fname::AbstractString) = Main.Base.include(Main, fname)
 function include(fname::AbstractString)
     mod = Main
-    isa(fname, String) || (fname = Base.convert(String, fname))
+    isa(fname, String) || (fname = Base.convert(String, fname)::String)
     path, prev = Base._include_dependency(mod, fname)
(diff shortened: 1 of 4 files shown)


KristofferC:
Seems to indeed work well! Is there anything written somewhere about how one would identify what calls need some help with type information?

JeffBezanson:
I did come up with some sketchy tooling to do this. I'll try to pull it into some re-usable state and at least put it in a gist or PR.

KristofferC:
Good to go?

PR #34597, opened 2020-01-30
more specific type for `file` field of `LineNumberNode`

JeffBezanson:
Any reason not to do this?

I also see we need to rename `jl_void_type` to `jl_nothing_type`.


vtjnash (approved):
Probably not. I think many of these have just been un-typed out of simplicity.

PR #34601, opened 2020-01-31
Construct MulAddMul at gemm_wrapper! call sites

### LINKED ISSUE #684: mul! performance regression on master

### LINKED ISSUE #34013: mul! performance regression on master

tkf:
This is yet another PR to fix JuliaLang/LinearAlgebra.jl#684.  The idea is to move alpha and beta to the type domain as early as possible.  Maybe this is better than JuliaLang/julia#34384 as it relies less on inlining and constant propagation?

### DIFF
--- a/stdlib/LinearAlgebra/src/generic.jl
+++ b/stdlib/LinearAlgebra/src/generic.jl
@@ -28,7 +28,7 @@ end
 MulAddMul(alpha::TA, beta::TB) where {TA, TB} =
     MulAddMul{isone(alpha), iszero(beta), TA, TB}(alpha, beta)
 
-MulAddMul() = MulAddMul(true, false)
+MulAddMul() = MulAddMul{true,true}(true, false)
 
 @inline (::MulAddMul{true})(x) = x
 @inline (p::MulAddMul{false})(x) = x * p.alpha
--- a/stdlib/LinearAlgebra/src/generic.jl
+++ b/stdlib/LinearAlgebra/src/generic.jl
@@ -25,8 +25,21 @@ struct MulAddMul{ais1, bis0, TA, TB}
     beta::TB
 end
 
-MulAddMul(alpha::TA, beta::TB) where {TA, TB} =
-    MulAddMul{isone(alpha), iszero(beta), TA, TB}(alpha, beta)
+@inline function MulAddMul(alpha::TA, beta::TB) where {TA,TB}
+    if isone(alpha)
+        if iszero(beta)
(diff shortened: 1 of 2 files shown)


stdlib/LinearAlgebra/src/generic.jl
```diff
@@ -25,10 +25,23 @@ struct MulAddMul{ais1, bis0, TA, TB}
...
+        if iszero(beta)
+            return MulAddMul{false,true,TA,TB}(alpha, beta)
+        else
+            return MulAddMul{false,false,TA,TB}(alpha, beta)
+        end
+    end
+end
 
-MulAddMul() = MulAddMul(true, false)
+MulAddMul() = MulAddMul{true,true}(true, false)
```
KristofferC:
```suggestion
MulAddMul() = MulAddMul{true,true,Bool,Bool}(true, false)
```

maybe?

KristofferC:
Let's go with this one?

tkf:
Yeah, I think this is better.  The latest commits in #34384 started breaking the tests.

KristofferC:
Win32 CI hang happens on the other PRs as well.

StefanKarpinski:
The part of https://github.com/JuliaLang/julia/pull/34384 that moves the calls to matmul2x2! and matmul3x3! before all the heavywight checking that they don't require seems like an independently good change, however, since it should reduce overhead in those cases. Separating higher overhead call to BLAS into its own subroutine also seems like a good thing. Not sure why this was suddenly decided to be much better.

KristofferC:
Because it didn't rely on constant propagation, it seemed less brittle to compiler changes.

StefanKarpinski:
Sure, that's good. The other parts of that change are good anyway though, without the parts that force inlining and constant propagation—i.e. that version of `gemm_wrapper!` should be easier for the compiler to deal with.

PR #34605, opened 2020-01-31
Added usage example of `ncodeunits`.

logankilpatrick:
Added usage example of `ncodeunits`.

### DIFF
--- a/base/strings/basic.jl
+++ b/base/strings/basic.jl
@@ -55,11 +55,11 @@ code unit value when calling `codeunit(s,i)`.
 julia> ncodeunits("The Julia Language")
 18
 
-julia> ncodeunits("∫")
-3
+julia> ncodeunits("∫eˣ")
+6
 
-julia> ncodeunits("Ä")
-2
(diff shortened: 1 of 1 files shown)


StefanKarpinski (commented):
Might want to give an example where `ncodeunits` and `length` are not the same (i.e. non-ASCII).

logankilpatrick:
Done @StefanKarpinski

base/strings/basic.jl
```diff
@@ -50,6 +50,18 @@ access this string must satisfy `1 ≤ i ≤ ncodeunits(s)`. Not all such indic
...
+```jldoctest
+julia> ncodeunits("The Julia Language")
+18
+
+julia> ncodeunits("∫")
+3
+
+julia> ncodeunits("Ä")
+2
+```
```
stevengj: I would give examples with a multi-character string and a `Char`, instead of single-character strings.  e.g.
```jl
julia> ncodeunits("∫eˣ")
6

julia> ncodeunits('∫'), ncodeunits('e'), ncodeunits('ˣ')
(3, 1, 2)
```
```

stevengj (approved):

PR #34606, opened 2020-01-31
Add some examples for the codeunit docstring

logankilpatrick:
Added usage example for `codeunit()`.

### DIFF
--- a/base/strings/basic.jl
+++ b/base/strings/basic.jl
@@ -79,6 +79,17 @@ Return the code unit value in the string `s` at index `i`. Note that
 I.e. the value returned by `codeunit(s, i)` is of the type returned by
 `codeunit(s)`.
 
+```jldoctest
+julia> codeunit("Hello", 1)
+0x48
+
+julia> codeunit("Hello", 2)
+0x65
+
+julia> typeof(codeunit("Hello", 2))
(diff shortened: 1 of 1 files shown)


base/strings/basic.jl
```diff
@@ -79,6 +79,18 @@ Return the code unit value in the string `s` at index `i`. Note that
 I.e. the value returned by `codeunit(s, i)` is of the type returned by
 `codeunit(s)`.
 
+# Examples
+```jldoctest
+julia> codeunit("Hello", 1)
+0x48
+
+julia> codeunit("Hello", 2)
```
KristofferC: Maybe this can be removed since it does the same thing as the one above?

base/strings/basic.jl
```diff
@@ -79,6 +79,18 @@ Return the code unit value in the string `s` at index `i`. Note that
...
+0x65
+
+julia> typeof(codeunit("Hello", 2))
```
KristofferC: Could save the value in a variable and call typeof on that instead, which might make this line simpler.

logankilpatrick:
@KristofferC I had the two examples because I wanted to show that each letter returned a different value which I guess is implicit and not really needed.

PR #34607, opened 2020-01-31
Add an example to the codeunits docstring

logankilpatrick:
Added usage example for `codeunits()`

### DIFF
--- a/base/strings/basic.jl
+++ b/base/strings/basic.jl
@@ -715,5 +715,23 @@ unsafe_convert(::Type{Ptr{Int8}}, s::CodeUnits{UInt8}) = unsafe_convert(Ptr{Int8
 Obtain a vector-like object containing the code units of a string.
 Returns a `CodeUnits` wrapper by default, but `codeunits` may optionally be defined
 for new string types if necessary.
+
+```jldoctest
+julia> codeunits("Julia is fast")
+13-element Base.CodeUnits{UInt8,String}:
+ 0x4a
+ 0x75
+ 0x6c
+ 0x69
+ 0x61
+ 0x20
+ 0x69
+ 0x73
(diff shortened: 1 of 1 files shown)


base/strings/basic.jl
```diff
@@ -715,5 +715,24 @@ unsafe_convert(::Type{Ptr{Int8}}, s::CodeUnits{UInt8}) = unsafe_convert(Ptr{Int8
 Obtain a vector-like object containing the code units of a string.
 Returns a `CodeUnits` wrapper by default, but `codeunits` may optionally be defined
 for new string types if necessary.
+
+# Examples
+```jldoctest
+julia> codeunits("Julia is fast")
```
KristofferC: This becomes a bit long (e.g. in the REPL). Perhaps just Julia and maybe some Unicode character so that the length of the CodeUnits isn't exactly the same as the number of characters.

logankilpatrick:
Done @KristofferC

base/strings/basic.jl
StefanKarpinski: Perhaps
```jl
julia> length("¡Julia!")
7

julia> ncodeunits("¡Julia!")
8
```

base/strings/basic.jl
```diff
@@ -718,21 +718,19 @@ for new string types if necessary.
...
- 0x20
- 0x69
- 0x73
- 0x20
- 0x66
- 0x61
- 0x73
- 0x74
+
+julia> codeunits("∫")
```
KristofferC: I just meant adding one unicode to the first string (to keep it short), could be `Juλia` for example.

logankilpatrick: ah, got it. Updated.

PR #34610, opened 2020-01-31
indent formulae on terminal

bicycle1885:
I found LaTeX formulae are not indented in the same way as other components:
```
julia> using LinearAlgebra

help?> qr
search: qr qr! QR QRPivoted sqrt isqrt QuickSort PartialQuickSort

  qr(A, pivot=Val(false); blocksize) -> F

  Compute the QR factorization of the matrix A: an orthogonal (or unitary if A is complex-valued) matrix Q, and an upper
  triangular matrix R such that

A = Q R

  The returned object F stores the factorization in a packed format:
```

This fixes it and adds minor refactoring.

### DIFF
--- a/stdlib/Markdown/src/IPython/IPython.jl
+++ b/stdlib/Markdown/src/IPython/IPython.jl
@@ -30,6 +30,3 @@ latex(io::IO, tex::LaTeX) =
 
 latexinline(io::IO, tex::LaTeX) =
     print(io, '$', tex.formula, '$')
-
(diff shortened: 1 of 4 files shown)


KristofferC:
What is the reason we ident the text at all? Seems like a waste of two characters.

bicycle1885:
No idea. But indenting docs will make it easier to find the starting location on terminal, especially when they are long. If we had a pager, we don't need it.

bicycle1885:
FYI, this is how docs look like without indentation:
image

KristofferC:
Yeah, maybe the current way is better. Sorry for the tangent.

PR #34611, opened 2020-01-31
remove unused macros `__FILE_SYMBOL__` and `get!`

JeffBezanson:
I believe the first was only used in the deprecation code for the new iteration protocol. The second, IIRC, predates higher-order functions being fast and is no longer used anywhere.

### DIFF
--- a/base/dict.jl
+++ b/base/dict.jl
@@ -463,14 +463,6 @@ function get!(default::Callable, h::Dict{K,V}, key::K) where V where K
     return v
 end
 
-# NOTE: this macro is trivial, and should
-#       therefore not be exported as-is: it's for internal use only.
-macro get!(h, key0, default)
-    return quote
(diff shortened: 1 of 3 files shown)


aviatesk:
I guess this may be better to be deprecated instead of just removed ?
Some packages imports it and thus they can't even be precompiled.

Because I used those in my environment, I found at lease the packages below use it:
- MacroTools
- Lazy
- Requiresa
- CodeTools

(well, I already create patches to them :))

Thoughts ?

pfitzseb:
`@get!` is also used in a [few more packages](https://pkg.julialang.org/docs/search?q=%40get%21):
- CodeTools
- Dagger
- DataArrays
- DataFlow
- Flux
- JuliaFormatter
- Lazy
- MacroTools
- ModelSanitizer
- Requires
- WebIO
- Zygote

JeffBezanson:
Thanks for making all those PRs!

PR #34613, opened 2020-01-31
simple inference graph profiler

JeffBezanson:
This is the method I used in #34596.
Here is a gist with the analysis code and an example run:
https://gist.github.com/JeffBezanson/e381ea9cbe790d6d8f9bcf473a664f8c

Basically this is way to find which calls are leading to the most inference work, roughly speaking. It's not time-based and might not be helpful in every case, but in this case it gave a useful profile.

### DIFF
--- a/base/compiler/typeinfer.jl
+++ b/base/compiler/typeinfer.jl
@@ -453,8 +453,15 @@ function resolve_call_cycle!(linfo::MethodInstance, parent::InferenceState)
     return false
 end
 
+const inference_graph = IdDict{Any,IdSet{Any}}()
+const record_edges = Bool[false]
+
 # compute (and cache) an inferred AST and return the current best estimate of the result type
 function typeinf_edge(method::Method, @nospecialize(atypes), sparams::SimpleVector, caller::InferenceState)
+    if record_edges[1]
+        callees = get!(IdSet{Any}, inference_graph, caller.linfo.specTypes)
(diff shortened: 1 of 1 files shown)


base/compiler/typeinfer.jl
```diff
@@ -453,8 +453,15 @@ function resolve_call_cycle!(linfo::MethodInstance, parent::InferenceState)
     return false
 end
 
+const inference_graph = IdDict{Any,IdSet{Any}}()
+const record_edges = Ref(false)
```
KristofferC:
```suggestion
const record_edges = Bool[false]
```

maybe (for bootstrap)

StefanKarpinski: Is `Ref` not defined early enough? Maybe a comment to that effect since that was the first thing I wondered when seeing the current state of the PR.

vtjnash: `Ref` is available, but no constructors. Use `RefValue`

vtjnash:
I'm not sure if this is worth merging. On one hand, it was useful information for you. On the other hand, it's inactive code and might be unstable as an API over time (it's a very specific subset of information, and we will probably want more general result reports some day).

JeffBezanson:
I did not really intend to merge this; it's just here in case anybody wants to use this method again.

vtjnash:
I think we can effectively declare this present now

PR #34617, opened 2020-02-01
Threads tests

bramtayl:
[Uncovered](https://codecov.io/gh/JuliaLang/julia/src/master/base/threadingconstructs.jl#L86)
[Uncovered](https://codecov.io/gh/JuliaLang/julia/src/master/base/threadingconstructs.jl#L90)
[Uncovered](https://codecov.io/gh/JuliaLang/julia/src/master/base/threadingconstructs.jl#L99)

### DIFF
--- a/test/threads_exec.jl
+++ b/test/threads_exec.jl
@@ -7,6 +7,12 @@ using Base.Threads: SpinLock
 # for cfunction_closure
 include("testenv.jl")
 
+@testset "@threads syntax" begin
+    @test_throws ArgumentError("wrong number of arguments in @threads") var"@threads"(LineNumberNode(@__LINE__), @__MODULE__, 1, 2)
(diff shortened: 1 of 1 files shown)


bramtayl:
CI failure unrelated?

dkarrasch:
I noticed you include the entire the exact error message string in the `@test_throws`, but isn't this a bit overly specific? To me it looks like we're testing/"guaranteeing" too much.

bramtayl:
Sure I can see that. Happy to change it back to `ArgumentError`.

bramtayl:
On the other hand, spelling out the error message is a good way to test whether it is correct, especially because its easy to make interpolation errors, e.g. `"$typeof(x)"` instead of `"$(typeof(x))"`

dkarrasch:
It may depend, I agree. My comment was more of general kind, proposing to test the full error message sparsely, not always.

bramtayl:
Let me know if you want me rebase this

PR #34621, opened 2020-02-01
define generic get! for AbstractDict

### LINKED ISSUE #34100: `get!` does not work with ENV
MWE:
```julia
get!(ENV, "key", "val")

ERROR: MethodError: no method matching get!(::Base.EnvDict, ::String, ::String)
Closest candidates are:
  get!(::IdDict{K,V}, ::Any, ::Any) where {K, V} at abstractdict.jl:648
  get!(::Dict{K,V}, ::Any, ::Any) where {K, V} at dict.jl:420
  get!(::WeakKeyDict{K,V} where V, ::Any, ::Any) where K at weakkeydict.jl:99
```

bicycle1885:
This closes #34100.

### DIFF
--- a/base/abstractdict.jl
+++ b/base/abstractdict.jl
@@ -514,7 +514,7 @@ getindex(t::AbstractDict, k1, k2, ks...) = getindex(t, tuple(k1,k2,ks...))
 setindex!(t::AbstractDict, v, k1, k2, ks...) = setindex!(t, v, tuple(k1,k2,ks...))
 
 get!(t::AbstractDict, key, default) = get!(() -> default, t, key)
-function get!(default::Callable, t::AbstractDict{K}, key0) where K
+function get!(default::Callable, t::AbstractDict{K,V}, key0) where K where V
     key = convert(K, key0)
     if !isequal(key, key0)
         throw(ArgumentError("$(limitrepr(key0)) is not a valid key for type $K"))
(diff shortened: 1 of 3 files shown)


base/abstractdict.jl
```diff
@@ -513,6 +513,18 @@ end
 getindex(t::AbstractDict, k1, k2, ks...) = getindex(t, tuple(k1,k2,ks...))
 setindex!(t::AbstractDict, v, k1, k2, ks...) = setindex!(t, v, tuple(k1,k2,ks...))
 
+get!(t::AbstractDict, key, default) = get!(() -> default, t, key)
```
JeffBezanson: `Dict` has a copy of this method, which can now be removed.

base/abstractdict.jl
```diff
@@ -513,6 +513,18 @@ end
...
+function get!(default::Callable, t::AbstractDict{K,V}, key0) where K where V
+    key = convert(K, key0)
+    if !isequal(key, key0)
```
JeffBezanson: Ah, this is probably why we didn't have this generic definition before --- without knowing more about the type of dictionary, it's not clear which keys are valid. Maybe we should not do the check, and leave it to `setindex!`? Of course, that calls the `default` function even if the value won't be used, but if `setindex!` is going to throw an error I'm not sure it matters.

StefanKarpinski:
Triage agrees with Jeff's comment https://github.com/JuliaLang/julia/pull/34621#discussion_r374276766 and is otherwise cool with this.

JeffBezanson:
Yes this is ok if we remove all the `convert`s and checks.

bicycle1885:
@JeffBezanson Thank you for your comments. I've updated as you suggested.

PR #34622, opened 2020-02-01
remove unused @get! macro

bicycle1885:
I found this macro definition while working on #34621, which is never used in the current codebase.  Perhaps a leftover of refactoring? Since it is not exported and is annotated as internal-use only, I think we can safely remove it without breaking user's code.


fredrikekre:
#34611

bicycle1885:
:sob:

PR #34626, opened 2020-02-02
make REPL ast transforms instance-specific

stevengj:
As suggested [here](https://github.com/JuliaLang/julia/pull/33864#discussion_r371343045), this makes the REPL ast transformation list specific to a `REPLBackend` instance, though there is a global array of defaults for new instances.

This PR also changes `REPL.softscope!` to `REPL.sofscope`.  There doesn't seem to be a big advantage to having this function mutating, and the non-mutating version is easier to use safely.  (In particular, though passing the mutating version to the new `include` function in #34595 is safe because the AST expressions are temporary, that would require us to document this property of `include` which is bothersome.)

### DIFF
--- a/stdlib/REPL/src/REPL.jl
+++ b/stdlib/REPL/src/REPL.jl
@@ -66,20 +66,28 @@ mutable struct REPLBackend
     response_channel::Channel
     "flag indicating the state of this backend"
     in_eval::Bool
+    "transformation functions to apply before evaluating expressions"
+    ast_transforms::Vector{Any}
     "current backend task"
     backend_task::Task
 
-    REPLBackend(repl_channel, response_channel, in_eval) =
-        new(repl_channel, response_channel, in_eval)
+    REPLBackend(repl_channel, response_channel, in_eval, ast_transforms=copy(repl_ast_transforms)) =
+        new(repl_channel, response_channel, in_eval, ast_transforms)
 end
 
-function softscope!(ex)
+"""
+    softscope(ex)
+
+Return a modified version of the parsed expression `ex` that uses
+the REPL's "soft" scoping rules for global syntax blocks.
+"""
+function softscope(@nospecialize ex)
     if ex isa Expr
         h = ex.head
         if h === :toplevel
-            for i = 1:length(ex.args)
-                ex.args[i] = softscope!(ex.args[i])
-            end
+            ex′ = Expr(h)
+            map!(softscope, resize!(ex′.args, length(ex.args)), ex.args)
+            return ex′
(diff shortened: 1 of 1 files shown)


stdlib/REPL/src/REPL.jl
```diff
@@ -66,20 +66,26 @@ mutable struct REPLBackend
...
+the REPL's "soft" scoping rules for global syntax blocks.
+"""
+function softscope(@nospecialize ex)
     if ex isa Expr
         h = ex.head
         if h === :toplevel
-            for i = 1:length(ex.args)
-                ex.args[i] = softscope!(ex.args[i])
-            end
+            return Expr(h, Any[softscope(a) for a in ex.args])
```
JeffBezanson:
```
e = Expr(h)
e.args = Any[softscope(a) for a in ex.args]
return e
```
Somewhat strangely, I don't believe there is a standard non-varargs Expr constructor.

stdlib/REPL/src/REPL.jl
```diff
@@ -89,7 +97,7 @@ function softscope!(ex)
     return ex
 end
 
-const repl_ast_transforms = Any[softscope!]
+const repl_ast_transforms = Any[softscope] # defaults for new REPL backends
```
StefanKarpinski: It feels like this could be limited to `Function` of `Callable`. Not that it helps much, but if someone wants to pass an object that has its calling overloaded, one can always just wrap it in a closure.

stevengj: I thought about it, but it’s not clear what we would gain by tightening the signature here? Certainly not performance...

StefanKarpinski: Fair enough. Doesn't matter I guess.

fredrikekre:
> This PR also changes REPL.softscope! to REPL.sofscope.

Documenter has been patched to use `softscope!` so in order to get the doc builder to pass on this PR you will have to revert https://github.com/JuliaLang/julia/commit/2c43528f39dfc22d2b02a800c27dcd09b002cb9f

StefanKarpinski:
Maybe the least disruptive course of action would be:

1. make `softscope!` and alias for `softscope` for the time being
   - I doubt anyone was relying on the side effect, they were probably mapping it
2. change Documenter to call `softscope` instead of `softscope!`
3. bump the Documenter version that Base uses
4. delete the `softscope!` alias

stevengj:
What is the remaining doctest error?

KristofferC:
It has been fixed on master but this needs to be rebased to pick it up (CI don't run merge commits now).

stevengj:
Hooray, green tests!

PR #34631, opened 2020-02-02
fix docstrings of floatmin and floatmax

bicycle1885:
The current docstring of `floatmin` says:
>  The smallest in absolute value non-subnormal value representable by the given floating-point DataType T.

If we literally interpret this sentence, the returned value would be zero, because zero is not a subnormal number with the minimum absolute value. So I changed it as "the smallest positive normal number". 

Also, the style does not follow the standard guideline of docstrings. So I fixed it as well and made the docstrings of `floatmin` and `floatmax` more symmetric.

### DIFF
--- a/base/float.jl
+++ b/base/float.jl
@@ -750,17 +750,30 @@ end
 end
 
 """
-    floatmin(T)
+    floatmin([T])
 
-The smallest in absolute value non-subnormal value representable by the given
-floating-point DataType `T`.
+Return the smallest positive normal number representable by the floating-point
+type `T` (`T = Float64` by default).
+
+# Examples
+```jldoctest
+julia> floatmin(Float16)
+Float16(6.104e-5)
+
+julia> floatmin(Float32)
+1.1754944f-38
+
+julia> floatmin()
+2.2250738585072014e-308
+```
 """
 floatmin(x::T) where {T<:AbstractFloat} = floatmin(T)
 
 """
-    floatmax(T)
+    floatmax([T])
 
-The highest finite value representable by the given floating-point DataType `T`.
+Return the largest finite number representable by the floating-point type `T`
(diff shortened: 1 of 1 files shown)


dkarrasch (approved):
LGTM.

base/float.jl
```diff
@@ -750,17 +750,30 @@ end
 end
 
 """
-    floatmin(T)
+    floatmin([T])
 
-The smallest in absolute value non-subnormal value representable by the given
-floating-point DataType `T`.
+Return the smallest positive normal number representable by the floating-point
```
halleysfifthinc: Given that positive and negative zeros are possible for floats (IEEE at least), this language still doesn't exclude zero. Perhaps the a combination of the original language and your additions, with the clarification of "non-zero" would be better?

> Return the smallest positive, non-zero, and non-subnormal value representable by the given floating-point type `T` (`T = Float64` by default).

base/float.jl
```diff
@@ -750,17 +750,30 @@ end
 end
 
 """
-    floatmin(T)
+    floatmin([T])
```
JeffBezanson:
```suggestion
    floatmin(T = Float64)
```
Then it does not need to be mentioned separately below.

base/float.jl
bicycle1885: I say "positive *normal* number" instead of "positive number", which excludes the possibility of any zeros since zeros are not a normal number.

halleysfifthinc: That's a fair point, but that wording is also subtle enough that I think it could be easily missed.

bicycle1885: Hmm, I'd prefer simple and concise wording in this case. It is apparent that this function does not return a zero from examples. So, adding more words looks rather redundant.

PR #34633, opened 2020-02-02
Windows - guidance for shell mode in the REPL

reumle:
Shell mode behavior is currently very non intuitive on Windows. It essentially does nothing a basic user (that would be me..) understands by shell mode (list directory, show environment variables etc) . So I clarified, and showed a way to fire up the two most common shells, `cmd.exe` and `powershell`.

### DIFF
--- a/stdlib/REPL/docs/src/index.md
+++ b/stdlib/REPL/docs/src/index.md
@@ -135,7 +135,23 @@ julia> ; # upon typing ;, the prompt changes (in place) to: shell>
 shell> echo hello
 hello
 ```
+* Note for Windows users: Julia Shell mode does not expose windows shell commands.Hence, this will fail: 
+```julia-repl
+julia> ; # upon typing ;, the prompt changes (in place) to: shell>
+shell> dir
+ERROR: IOError: could not spawn `dir`: no such file or directory (ENOENT)
+Stacktrace!
+.......
+```
+However, you can get access to Power Shell like this:
+```julia-repl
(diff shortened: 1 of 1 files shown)


stdlib/REPL/docs/src/index.md
```diff
@@ -135,7 +135,23 @@ julia> ; # upon typing ;, the prompt changes (in place) to: shell>
 shell> echo hello
 hello
 ```
+* Note for Windows users: Julia Shell mode does not expose windows shell commands.Hence, this will fail:
```
StefanKarpinski:
```suggestion
!!! note
    For Windows users, Julia's shell mode does not expose windows shell commands.
    Hence, this will fail:
```

reumle: closing (?)  this pull request, in order to resubmit with your suggestions, and other small enhancements (failing test etc), in a proper branch this time (submission guidelines). Thanks for reviewing!

StefanKarpinski: You can push new commits to the existing pull request. It's much easier if you just commit the suggestions on the same branch.

reumle:
This now passes (10/12) tests (2 still running), and shows how to start `cmd.exe`, the second, legacy, windows shell. 

Small caveat, both shells are in french… I will try to get English texts… but it will need another commit..

reumle:
Not sure what to do about the 2 failing tests...

ViralBShah:
Thanks @reumle. I ran into these issues on Windows myself, so it is good to have these.

PR #34635, opened 2020-02-02
inserteach! of Vector into Vector

aminya:
## Changes
- inserteach! Vector into Vector at the given index
- inserteach! Vector into Vector for the given indices
- inserteach! Vector into Vector for the given range
<del>- inserteach! specialization for the items with the same type of Vector </del>

## What is remaining:
- Testing new algorithm for multi-index inserteach!

<del>

- Additional benchmark requested: https://github.com/JuliaLang/julia/pull/34635#issuecomment-589281419
- To avoid overlapping with `insert!` in some special cases, a new name for these methods should be chosen. I like <del> `insert_elements!`</del>  `inserteach!` (the most descriptive name IMO) <del>, and `thrust!` (compact name). </del> Let me know your suggestion. 

</del>

## Performance

Specialization is done for AbstractRanges, AbstractUnitRages, and AbstractBool.
To get the best algorithm for general AbstractVector, I tried 5 other alternatives, and what is given in the PR is the fastest:
<details>
  <summary>Spoiler</summary>

### Single Index
`inserteach!` is much faster than the others.
```julia
function inserteach!(a::Vector, index::Integer, items::AbstractVector)
    m = length(items)
    if m === 0
        return a
    end
    _growat!(a, index, m)  # does bound check
    @inbounds for k = index:index+m-1
        a[k] = items[k-index+1]
    end
    return a
end

function inserteach_singleindex_2!(a::Vector, index::Integer, items::AbstractVector)
    for (i, item) in enumerate(items)
        insert!(a, i-1+index, item)
    end
    return a
end

function inserteach_singleindex_3!(a::Vector, index::Integer, items::AbstractVector)
    sizehint!(a, length(items)+length(a))
    for (i, item) in enumerate(items)
        insert!(a, i-1+index, item)
    end
    return a
end
```
### Single-index result
```julia
a = [1, 2, 3, 4]
b = 3
c = [10, 9, 8]

using Test; @test inserteach!(copy(a), b, c) == inserteach_singleindex_2!(copy(a), b, c) == inserteach_singleindex_3!(copy(a), b ,c)

using BenchmarkTools
@btime inserteach!($copy(a), $b, $c)
  # 97.465 ns (2 allocations: 192 bytes)
@btime inserteach_singleindex_2!($copy(a), $b, $c)
  # 115.888 ns (2 allocations: 192 bytes)
@btime inserteach_singleindex_3!($copy(a), $b, $c)
  # 119.802 ns (2 allocations: 192 bytes)

vsize = 1000
d = rand(Float32, vsize)
e = 30
f = rand(Float32, vsize÷10)

@btime inserteach!($copy(d), $e, $f)
  # 880.000 ns (2 allocations: 11.94 KiB)
@btime inserteach_singleindex_2!($copy(d), $e, $f)
  # 2.678 μs (2 allocations: 11.94 KiB)
@btime inserteach_singleindex_3!($copy(d), $e, $f)
  # 2.862 μs (2 allocations: 11.94 KiB)

g = rand(vsize)
h = 30
i = rand(vsize÷10)
@btime inserteach!($copy(g), $h, $i)
  # 1.480 μs (2 allocations: 23.63 KiB)
@btime inserteach_singleindex_2!($copy(g), $h, $i)
  # 3.825 μs (2 allocations: 23.63 KiB)
@btime inserteach_singleindex_3!($copy(g), $h, $i)
  # 4.243 μs (2 allocations: 23.63 KiB)

```

### Multi-Index
For larger sizes `inserteach!` is a lot faster. `inserteach_multiindex7!` has a similar speed to `inserteach!` until a batch `growat!` is written in C

```julia
function inserteach!(a::Vector, indices::Union{AbstractVector, AbstractRange{<:Integer}}, items::AbstractVector)
    itemslen = length(items)
    indiceslen = length(indices)
    if itemslen !== indiceslen
        throw(DimensionMismatch("Length of the indices and items should be the same"))
    elseif itemslen === 0
        return a
    end
    if issorted(indices) # based on the tests overhead/advantage is low
        indices_sorted = indices
        items_sorted = items
    else
        # sorts so the final index is correct
        indices_sorted, items_sorted = sort_indices_items(indices, items)
    end
    # _growat!.(Ref(a), indices_sorted, 1) # does bound check
    for (index, item) in zip(indices_sorted, items_sorted)
         _growat!(a, index, 1)
        @inbounds a[index] = item
    end
    return a
end

function sort_indices_items(indices::AbstractVector, items)
    sort_indices = sortperm(indices)
    indices_sorted = indices[sort_indices]
    items_sorted = items[sort_indices]
    return indices_sorted, items_sorted
end
function inserteach_multiindex_2!(a::Vector{T}, indices::AbstractVector, items::AbstractVector) where T
# defines a new array, finds the 1:newlength indices that are not in the given indices (rest), then populates the new array based on that.
    itemslen = length(items)
    indiceslen = length(indices)
    if itemslen !== indiceslen
        throw(DimensionMismatch("Length of the indices and items should be the same"))
    elseif itemslen === 0
        return a
    end
    alen = length(a)
    v = Vector{T}(undef, alen+itemslen)
    # rest = filter(x -> !in(x, indices), 1:alen+itemslen)
    rest = .!in.(1:alen+itemslen, Ref(indices))
   @inbounds v[rest] = a
    v[indices] = items
    return v
end

function inserteach_multiindex_3!(a::Vector, indices::AbstractVector, items::AbstractVector)
# grows the array at the end for the length of indices, then finds the rest of the indices that are not part of given indices (rest), then populates the array based on the indices.
    itemslen = length(items)
    indiceslen = length(indices)
    if itemslen !== indiceslen
        throw(DimensionMismatch("Length of the indices and items should be the same"))
    elseif itemslen === 0
        return a
    end
    alen = length(a)
    _growat!(a, alen+1, indiceslen) # does bound check
    # rest = filter(x -> !in(x, indices), 1:alen+itemslen)
     rest = .!in.(1:alen+itemslen, Ref(indices))
    @inbounds begin
        a[rest] = a[1:alen]
        a[indices] = items
    end
    return a
end

function inserteach_multiindex_3_2!(a::Vector, indices::AbstractVector, items::AbstractVector)
# grows the array at the end for the length of indices, then finds the rest of the indices that are not part of given indices (rest), then populates the array based on the indices.
    itemslen = length(items)
    indiceslen = length(indices)
    if itemslen !== indiceslen
        throw(DimensionMismatch("Length of the indices and items should be the same"))
    elseif itemslen === 0
        return a
    end
    alen = length(a)
    b = copy(a)
    _growat!(a, alen+1, indiceslen) # does bound check
    # rest = filter(x -> !in(x, indices), 1:alen+itemslen)
     rest = .!in.(1:alen+itemslen, Ref(indices))
    @inbounds begin
        a[rest] = b
        a[indices] = items
    end
    return a
end

function inserteach_multiindex_4!(a::Vector{T}, indices::AbstractVector, items::AbstractVector) where T
# sorts the indices. Then in a loop populates the array based on the fact that an index of the new array is in the given indices. We need sort because `in` only gives true (not the index).
    itemslen = length(items)
    indiceslen = length(indices)
    if itemslen !== indiceslen
        throw(DimensionMismatch("Length of the indices and items should be the same"))
    elseif itemslen === 0
        return a
    end
    alen = length(a)
    v = Vector{T}(undef, alen+itemslen)
    if issorted(indices) # based on the tests overhead/advantage is low
        indices_sorted = indices
        items_sorted = items
    else
        # sorts so the final index is correct
        indices_sorted, items_sorted = sort_indices_items(indices, items)
    end
    kitem = 1
    ka=1
    @inbounds for k=1:alen+itemslen
        if k in indices_sorted
            v[k] = items_sorted[kitem]
            kitem+=1
        else
            v[k] = a[ka]
            ka+=1
        end
    end
    return v
end

function inserteach_multiindex_5!(a::Vector{T}, indices::AbstractVector, items::AbstractVector) where T
# In a loop populates the array based on the finded index of the new array is in the given indices.
    itemslen = length(items)
    indiceslen = length(indices)
    if itemslen !== indiceslen
        throw(DimensionMismatch("Length of the indices and items should be the same"))
    elseif itemslen === 0
        return a
    end
    alen = length(a)
    v = Vector{T}(undef, alen+itemslen)
    kitem = 1
    ka=1
    @inbounds for k=1:alen+itemslen
        kitem = findfirst(x -> x == k, indices)
        if isnothing(kitem)
            v[k] = a[ka]
            ka+=1
        else
            v[k] = items[kitem]
            kitem+=1
        end
    end
    return v
end

function inserteach_multiindex_6!(a::Vector{T}, indices::AbstractVector, items::AbstractVector) where T
# Similar to 6 but first checks `in` then uses `findfirst` only for the ones that already exist
    itemslen = length(items)
    indiceslen = length(indices)
    if itemslen !== indiceslen
        throw(DimensionMismatch("Length of the indices and items should be the same"))
    elseif itemslen === 0
        return a
    end
    alen = length(a)
    v = Vector{T}(undef, alen+itemslen)
    kitem = 1
    ka=1
    @inbounds for k=1:alen+itemslen
        if k in indices
            kitem = findfirst(x -> x == k, indices)
            v[k] = items[kitem]
            kitem+=1
        else
            v[k] = a[ka]
            ka+=1
        end
    end
    return v
end


function inserteach_multiindex_7!(a::Vector, indices::AbstractVector, items::AbstractVector)
# For loops around insert
    itemslen = length(items)
    indiceslen = length(indices)
    if itemslen !== indiceslen
        throw(DimensionMismatch("Length of the indices and items should be the same"))
    elseif itemslen === 0
        return a
    end
    if issorted(indices) # based on the tests overhead/advantage is low
        indices_sorted = indices
        items_sorted = items
    else
        # sorts so the final index is correct
        indices_sorted, items_sorted = sort_indices_items(indices, items)
    end
    for (index, item) in zip(indices_sorted, items_sorted)
        insert!(a, index, item)
    end
    return a
end

function inserteach_multiindex_8!(a::Vector, indices::AbstractVector, items::AbstractVector)
# For loops around insert with sizehint
    itemslen = length(items)
    indiceslen = length(indices)
    if itemslen !== indiceslen
        throw(DimensionMismatch("Length of the indices and items should be the same"))
    elseif itemslen === 0
        return a
    end
    if issorted(indices) # based on the tests overhead/advantage is low
        indices_sorted = indices
        items_sorted = items
    else
        # sorts so the final index is correct
        indices_sorted, items_sorted = sort_indices_items(indices, items)
    end
    sizehint!(a, length(items)+length(a))
    for (index, item) in zip(indices_sorted, items_sorted)
        insert!(a, index, item)
    end
    return a
end
```
### Multi-Index Result:
For small sizes inserteach5! and 6 are definitely faster
```julia
using BenchmarkTools

a = [1, 2, 3, 4]
b = collect(6:-2:2)
c = [10, 9, 8]

using Test;
@test inserteach_multiindex_8!(copy(a), b, c) == inserteach_multiindex_7!(copy(a), b, c) == inserteach_multiindex_6!(copy(a), b, c) == inserteach_multiindex_5!(copy(a), b, c) == inserteach_multiindex_4!(copy(a), b, c) == inserteach_multiindex_3!(copy(a), b, c) == inserteach_multiindex_2!(copy(a), b, c) == inserteach!(copy(a), b, c)

println("small set")

@btime inserteach!(copy($a), $b, $c)
  # 336.814 ns (8 allocations: 576 bytes)
@btime inserteach_multiindex_2!(copy($a), $b, $c)
  # 244.755 ns (5 allocations: 416 bytes)
  # 241.960 ns (11 allocations: 688 bytes)  # using filter instead of in
@btime inserteach_multiindex_3!(copy($a), $b, $c)
  # 272.378 ns (6 allocations: 464 bytes)
  # 257.534 ns (11 allocations: 720 bytes) # using filter instead of in
@btime inserteach_multiindex_3_2!(copy($a), $b, $c)
  # 272.810 ns (6 allocations: 464 bytes)
@btime inserteach_multiindex_4!(copy($a), $b, $c)
  # 321.397 ns (8 allocations: 640 bytes)
@btime inserteach_multiindex_5!(copy($a), $b, $c)
  # 85.020 ns (2 allocations: 256 bytes)
@btime inserteach_multiindex_6!(copy($a), $b, $c)
  # 88.223 ns (2 allocations: 256 bytes)
@btime inserteach_multiindex_7!(copy($a), $b, $c)
  # 337.900 ns (8 allocations: 576 bytes)
@btime inserteach_multiindex_8!(copy($a), $b, $c)
  # 348.843 ns (8 allocations: 576 bytes)

```
But for larger sizes inserteach! is a lot faster. `inserteach_multiindex7!` has a similar speed to `inserteach!` until a batch `growat!` is written in C
```julia
println("big set, with skips, falling, non-sorted")

vsize = 1000
d = rand(Float32, vsize)
e = collect(vsize:-10:1)
f = rand(Float32, vsize÷10)

@btime inserteach!(copy($d), $e, $f)
  # 5.480 μs (8 allocations: 14.22 KiB)
@btime inserteach_multiindex_2!(copy($d), $e, $f)
  # 55.400 μs (6 allocations: 12.97 KiB)
  # 58.800 μs (11 allocations: 33.55 KiB) # using filter instead of in
@btime inserteach_multiindex_3!(copy($d), $e, $f)
  # 57.999 μs (7 allocations: 20.47 KiB)
  # 59.501 μs (11 allocations: 48.41 KiB)  # using filter instead of in
@btime inserteach_multiindex_3_2!(copy($d), $e, $f)
  # 56.500 μs (7 allocations: 20.47 KiB)
@btime inserteach_multiindex_4!(copy($d), $e, $f)
  # 53.199 μs (8 allocations: 10.78 KiB)
@btime inserteach_multiindex_5!(copy($d), $e, $f)
  # 71.999 μs (2 allocations: 8.50 KiB)
@btime inserteach_multiindex_6!(copy($d), $e, $f)
  # 59.999 μs (2 allocations: 8.50 KiB)
@btime inserteach_multiindex_7!(copy($d), $e, $f)
  # 5.417 μs (8 allocations: 14.22 KiB)
@btime inserteach_multiindex_8!(copy($d), $e, $f)
  # 5.633 μs (8 allocations: 14.22 KiB)

```

For sorted array, it gets even faster:
```julia
println("big set, with skips, rising, sorted")

g = rand(vsize)
h = collect(1:10:vsize)
i = rand(vsize÷10)

@btime inserteach!(copy($g), $h, $i)
  # 7.100 μs (2 allocations: 23.63 KiB)
@btime inserteach_multiindex_2!(copy($g), $h, $i)
  # 55.700 μs (6 allocations: 21.16 KiB)
  # 58.699 μs (11 allocations: 33.55 KiB) # using filter instead of in
@btime inserteach_multiindex_3!(copy($g), $h, $i)
  # 58.201 μs (7 allocations: 36.03 KiB)
  # 59.700 μs (11 allocations: 48.41 KiB)  # using filter instead of in
@btime inserteach_multiindex_3_2!(copy($g), $h, $i)
  # 58.000 μs (7 allocations: 36.03 KiB)
@btime inserteach_multiindex_4!(copy($g), $h, $i)
  # 52.100 μs (2 allocations: 16.69 KiB)
@btime inserteach_multiindex_5!(copy($g), $h, $i)
  # 75.299 μs (2 allocations: 16.69 KiB)
@btime inserteach_multiindex_6!(copy($g), $h, $i)
  # 58.300 μs (2 allocations: 16.69 KiB)
@btime inserteach_multiindex_7!(copy($g), $h, $i)
  # 7.050 μs (2 allocations: 23.63 KiB)
@btime inserteach_multiindex_8!(copy($g), $h, $i)
  # 7.425 μs (2 allocations: 23.63 KiB)
```
</details>

## Application
It is an algorithm that can be used for a lot of different applications. I usually ended up writing my own `inserteach!`.

It allows running different algorithms with different goals independent of each other. Like doing some sort first. Then inserting some other array into our array, etc. 

As an example:
Imagine your Instagram feed being an array that its order is defined (sorted based on your interests). Then Instagram decides to insert ads for every 10 posts. You can do it using
```julia
sort!(posts) # based on interests
insert!(posts, 1:10:length(posts), ads)
```
or imagine you have a song (like an array) that you want to insert a drum crash every 4 bars. You can think of many other things. 

https://github.com/JuliaLang/julia/pull/34635#issuecomment-581754452 :
> From an algorithmic viewpoint, it's more efficient to do the batch insertion.
> 
> For arrays of length N and a number of (sorted!) insertion points M it's an O(N + M) algorithm to do the insertion in a batch and a single reallocation. This is opposed to an O(N*M) algorithm to iteratively insert! because you only need to move the existing elements once. (Not to mention insert! is mutating so iteratively inserting needs to keep track of how the insertion points change which is annoying.)

### DIFF
--- a/base/array.jl
+++ b/base/array.jl
@@ -1300,9 +1300,14 @@ function insert!(a::Array{T,1}, indices::AbstractVector, items::Array{T,1}) wher
     elseif itemslen === 0
         return a
     end
-    sortindices = sortperm(indices) # sorts so the final index is correct
-    indices_sorted = indices[sortindices]
-    items_sorted = items[sortindices]
+    if issorted(indices) # based on the tests overhead/advantage is low
+        indices_sorted = indices
+        items_sorted = items
+    else
+        sort_indices = sortperm(indices) # sorts so the final index is correct
+        indices_sorted = indices[sort_indices]
+        items_sorted = items[sort_indices]
+    end
     _growat!.(Ref(a), indices_sorted, 1) # does bound check
     @inbounds for (index, item) in zip(indices_sorted, items_sorted)
         a[index] = item
@@ -1319,11 +1324,16 @@ function insert!(a::Array{T,1}, indices::AbstractRange{<:Integer}, items::Array{
     elseif itemslen === 0
         return a
     end
-    indicessorted = sort(indices) # sorting them is faster than indexing into them
-    sortindices = sortperm(indices) # sorts so the final index is correct
-    items_sorted = items[sortindices]
-    _growat!.(Ref(a), indicessorted, 1) # does bound check
-    @inbounds for (index, item) in zip(indicessorted, items_sorted)
+    if issorted(indices) # based on the tests overhead/advantage is low
+        indices_sorted = indices
+        items_sorted = items
+    else
+        indices_sorted = sort(indices) # sorting them is faster than indexing into them
+        sort_indices = sortperm(indices) # sorts so the final index is correct
+        items_sorted = items[sort_indices]
+    end
+    _growat!.(Ref(a), indices_sorted, 1) # does bound check
+    @inbounds for (index, item) in zip(indices_sorted, items_sorted)
         a[index] = item
     end
     return a
--- a/base/array.jl
+++ b/base/array.jl
@@ -1234,14 +1234,14 @@ julia> inserteach!([1, 2, 3], 2, [8, 9])
  3
 ```
 """
-function inserteach!(a::Vector, i::Integer, items::AbstractVector)
+function inserteach!(a::Vector, index::Integer, items::AbstractVector)
     m = length(items)
     if m === 0
         return a
     end
-    _growat!(a, i, m)  # does bound check
-    @inbounds for k = i:i+m-1
-        a[k] = items[k-i+1]
+    _growat!(a, index, m)  # does bound check
+    @inbounds for k = index:index+m-1
+        a[k] = items[k-index+1]
     end
     return a
 end
@@ -1263,7 +1263,7 @@ julia> inserteach!([1, 2, 3], 2:3, [8, 9])
  3
 ```
 
-Following example inserts `10, 9, and 8` at index `6 ,4, and 2` respectively
+Following example inserts `10, 9, and 8` at index `6, 4, and 2` respectively
 ```jldoctest
 julia> inserteach!([1, 2, 3, 4, 5], 6:-2:2, [10, 9, 8])
 8-element Array{Int64,1}:
--- a/base/array.jl
+++ b/base/array.jl
@@ -1216,6 +1216,11 @@ function insert!(a::Array{T,1}, i::Integer, item) where T
     return a
 end
 
+function insert!(a::Array{T,1}, i::Integer, item::T) where T
+    _growat!(a, i, 1) # does bound check
+    @inbounds a[i] = item
+    return a
+end
 """
     deleteat!(a::Vector, i::Integer)
 
--- a/base/array.jl
+++ b/base/array.jl
@@ -1280,6 +1280,17 @@ julia> insert!([1, 2, 3, 4, 5], 6:-2:2, [10, 9, 8])
   4
   5
 ```
+
+It is possible to use boolean indices (each true shows insertion index):
+```jldoctest
+julia> insert!([1, 2, 3], [true, false, true, false, false], [10, 9])
+5-element Array{Int64,1}:
+ 10
+  1
+  9
+  2
+  3
+```
 """
 function insert!(a::Array{T1,1}, indices::T2, items::Array{T1,1}) where {T1} where {T2<:Union{Array{<:Integer,1}, OrdinalRange{<:Integer, <:Integer}}}
     itemslen = length(items)
@@ -1307,6 +1318,27 @@ function insert!(a::Array{T,1}, r::UnitRange{<:Integer}, items::Array{T,1}) wher
     else
         return insert!(a, first(r), items)
     end
+end
+
+function insert!(a::Array{T,1}, boolindices::AbstractVector{Bool}, items::Array{T,1}) where T
+    alen = length(a)
+    itemslen = length(items)
+    boolindiceslen = length(boolindices)
+    if alen + itemslen !== boolindiceslen
+        throw(ArgumentError("Length of the indices should be equal to length of the final vector"))
+    end
+    indices = findall(x -> x=== true, boolindices) # doesn't need sort afterwards
+    indiceslen = length(indices)
+    if itemslen !== indiceslen
+        throw(ArgumentError("Number of true indices should be the same as length of items"))
+    elseif itemslen === 0
+        return a
+    end
+    _growat!.(Ref(a), indices, 1) # does bound check
+    @inbounds for (index, item) in zip(indices, items)
+        a[index] = item
+    end
+    return a
 end
 
 """
--- a/base/array.jl
+++ b/base/array.jl
@@ -1247,44 +1247,51 @@ function inserteach!(a::Vector, index::Integer, items::AbstractVector)
 end
 
 """
-    inserteach!(a::Vector, range, items::Vector)
-    inserteach!(a::Vector, indices::Vector, items::Vector)
+    inserteach!(a::Vector, indices, items::Vector)
 
-Insert all of the elements of `items` into `a` for the given `range` or `indices`. `range` or `indices` are the resulting indices of elements of `items` in the `a`.
+Insert all of the elements of `items` into `a` for the given `indices`.
+
+`indices` shows the indices in the original `a`, and it can be a Vector, UnitRange, StepRange, or Vector{Bool}.
 
 # Examples
+Inserts `0` between each two elements:
 ```jldoctest
-julia> inserteach!([1, 2, 3], 2:3, [8, 9])
-5-element Array{Int64,1}:
- 1
- 8
- 9
- 2
- 3
-```
-
-Following example inserts `10, 9, and 8` at index `6, 4, and 2` respectively
-```jldoctest
-julia> inserteach!([1, 2, 3, 4, 5], 6:-2:2, [10, 9, 8])
-8-element Array{Int64,1}:
+julia> inserteach!([1, 2, 3, 4, 5, 6], 1:2:5, [0, 0, 0])
+9-element Array{Int64,1}:
+  0
   1
-  8
   2
-  9
+  0
   3
- 10
   4
+  0
   5
+  6
 ```
 
-It is possible to use boolean indices (each true shows insertion index):
+Inserts `10, 9, and 8` at
 ```jldoctest
-julia> inserteach!([1, 2, 3], [true, false, true, false, false], [10, 9])
+julia> inserteach!([1, 2, 3, 4, 5, 6], 1:2:5, [0, 0, 0])
+9-element Array{Int64,1}:
+  0
+  1
+  2
+  0
+  3
+  4
+  0
+  5
+  6
+```
+
+For `Bool` indices, each `true` shows insertion index for each item:
+```jldoctest
+julia> inserteach!([1, 2, 3], [false, false, true, true], [10, 9])
 5-element Array{Int64,1}:
  10
   1
-  9
   2
+  9
   3
 ```
 """
@@ -1303,10 +1310,7 @@ function inserteach!(a::Vector, indices::Union{AbstractVector, AbstractRange{<:I
         # sorts so the final index is correct
         indices_sorted, items_sorted = sort_indices_items(indices, items)
     end
-    for (index, item) in zip(indices_sorted, items_sorted)
-        _growat!(a, index, 1) # TODO write a batch growat in C for speed-up
-        @inbounds a[index] = item
-    end
+    _inserteach!(a, indices_sorted, items_sorted)
     return a
 end
 
@@ -1324,6 +1328,23 @@ function sort_indices_items(indices::AbstractRange{<:Integer}, items)
     return indices_sorted, items_sorted
 end
 
+function _inserteach!(a, indices, items)
+    # Insert items into `a` such that `items[j]` is placed between the elements
+    # `a[indices[j-1]:indices[j]]`.
+    read_idx = lastindex(a)
+    resize!(a, length(a) + length(items))
+    insert_idx = lastindex(a)
+    for (new_idx,newitem) in zip(Iterators.reverse(indices), Iterators.reverse(items))
+        while read_idx >= new_idx
+            a[insert_idx] = a[read_idx]
+            read_idx -= 1
+            insert_idx -= 1
+        end
+        a[insert_idx] = newitem
+        insert_idx -= 1
+    end
+end
+
 function inserteach!(a::Vector, r::AbstractUnitRange{<:Integer}, items::AbstractVector)
     ilen = length(items)
     rlen = length(r)
@@ -1338,20 +1359,14 @@ function inserteach!(a::Vector, boolindices::AbstractVector{Bool}, items::Abstra
     alen = length(a)
     itemslen = length(items)
     boolindiceslen = length(boolindices)
-    if alen + itemslen !== boolindiceslen
-        throw(ArgumentError("Length of the indices should be equal to length of the final vector"))
-    end
-    indices = findall(x -> x=== true, boolindices) # doesn't need sort afterwards
+    indices = findall(x -> x=== true, boolindices) # no sort needed afterwards
     indiceslen = length(indices)
     if itemslen !== indiceslen
         throw(ArgumentError("Number of true indices should be the same as length of items"))
     elseif itemslen === 0
         return a
     end
-    for (index, item) in zip(indices, items)
-        _growat!(a, index, 1) # TODO write a batch growat in C for speed-up
-        @inbounds a[index] = item
-    end
+    _inserteach!(a, indices, items)
     return a
 end
 
--- a/base/array.jl
+++ b/base/array.jl
@@ -1216,12 +1216,6 @@ function insert!(a::Array{T,1}, i::Integer, item) where T
     return a
 end
 
-function insert!(a::Array{T,1}, i::Integer, item::T) where T
-    _growat!(a, i, 1) # does bound check
-    @inbounds a[i] = item
-    return a
-end
-
 """
     insert!(a::Vector, index::Integer, items::Vector)
 
--- a/base/array.jl
+++ b/base/array.jl
@@ -1221,6 +1221,37 @@ function insert!(a::Array{T,1}, i::Integer, item::T) where T
     @inbounds a[i] = item
     return a
 end
+
+"""
+    insert!(a::Vector, index::Integer, items::Vector)
+
+Insert all of the elements of `items` into `a` at the given `index`. `index` is the index of first element of `items` in the resulting `a`.
+
+# Examples
+```jldoctest
+julia> insert!([6, 5, 4, 2, 1], 4, [8, 9, 9])
+8-element Array{Int64,1}:
+ 6
+ 5
+ 4
+ 8
+ 9
+ 9
+ 2
+ 1
+```
+"""
+function insert!(a::Array{T,1}, i::Integer, items::Array{T,1}) where {T}
+    m = length(items)
+    if m === 0
+        return a
+    end
+    _growat!(a, i, m)  # does bound check
+    @inbounds for k = i:i+m-1
+        a[k] = items[k-i+1]
+    end
+    return a
+end
 """
     deleteat!(a::Vector, i::Integer)
 
--- a/base/array.jl
+++ b/base/array.jl
@@ -1234,7 +1234,7 @@ julia> inserteach!([1, 2, 3], 2, [8, 9])
  3
 ```
 """
-function inserteach!(a::Array{T,1}, i::Integer, items::Array{T,1}) where {T}
+function inserteach!(a::Vector, i::Integer, items::AbstractVector)
     m = length(items)
     if m === 0
         return a
@@ -1288,7 +1288,7 @@ julia> inserteach!([1, 2, 3], [true, false, true, false, false], [10, 9])
   3
 ```
 """
-function inserteach!(a::Array{T,1}, indices::Union{AbstractVector, AbstractRange{<:Integer}}, items::Array{T,1}) where {T}
+function inserteach!(a::Vector, indices::Union{AbstractVector, AbstractRange{<:Integer}}, items::AbstractVector)
     itemslen = length(items)
     indiceslen = length(indices)
     if itemslen !== indiceslen
@@ -1324,7 +1324,7 @@ function sort_indices_items(indices::AbstractRange{<:Integer}, items)
     return indices_sorted, items_sorted
 end
 
-function inserteach!(a::Array{T,1}, r::AbstractUnitRange{<:Integer}, items::Array{T,1}) where T
+function inserteach!(a::Vector, r::AbstractUnitRange{<:Integer}, items::AbstractVector)
     ilen = length(items)
     rlen = length(r)
     if ilen !== rlen
@@ -1334,7 +1334,7 @@ function inserteach!(a::Array{T,1}, r::AbstractUnitRange{<:Integer}, items::Arra
     end
 end
 
-function inserteach!(a::Array{T,1}, boolindices::AbstractVector{Bool}, items::Array{T,1}) where T
+function inserteach!(a::Vector, boolindices::AbstractVector{Bool}, items::AbstractVector)
     alen = length(a)
     itemslen = length(items)
     boolindiceslen = length(boolindices)
--- a/base/array.jl
+++ b/base/array.jl
@@ -1288,7 +1288,7 @@ julia> inserteach!([1, 2, 3], [true, false, true, false, false], [10, 9])
   3
 ```
 """
-function inserteach!(a::Array{T,1}, indices::AbstractVector, items::Array{T,1}) where {T}
+function inserteach!(a::Array{T,1}, indices::Union{AbstractVector, AbstractRange{<:Integer}}, items::Array{T,1}) where {T}
     itemslen = length(items)
     indiceslen = length(indices)
     if itemslen !== indiceslen
@@ -1300,9 +1300,8 @@ function inserteach!(a::Array{T,1}, indices::AbstractVector, items::Array{T,1}) 
         indices_sorted = indices
         items_sorted = items
     else
-        sort_indices = sortperm(indices) # sorts so the final index is correct
-        indices_sorted = indices[sort_indices]
-        items_sorted = items[sort_indices]
+        # sorts so the final index is correct
+        indices_sorted, items_sorted = sort_indices_items(indices, items)
     end
     _growat!.(Ref(a), indices_sorted, 1) # does bound check
     @inbounds for (index, item) in zip(indices_sorted, items_sorted)
@@ -1311,28 +1310,18 @@ function inserteach!(a::Array{T,1}, indices::AbstractVector, items::Array{T,1}) 
     return a
 end
 
+function sort_indices_items(indices::AbstractVector, items)
+    sort_indices = sortperm(indices)
+    indices_sorted = indices[sort_indices]
+    items_sorted = items[sort_indices]
+    return indices_sorted, items_sorted
+end
 
-function inserteach!(a::Array{T,1}, indices::AbstractRange{<:Integer}, items::Array{T,1}) where {T}
-    itemslen = length(items)
-    indiceslen = length(indices)
-    if itemslen !== indiceslen
-        throw(DimensionMismatch("Length of the indices and items should be the same"))
-    elseif itemslen === 0
-        return a
-    end
-    if issorted(indices) # based on the tests overhead/advantage is low
-        indices_sorted = indices
-        items_sorted = items
-    else
-        indices_sorted = sort(indices) # sorting them is faster than indexing into them
-        sort_indices = sortperm(indices) # sorts so the final index is correct
-        items_sorted = items[sort_indices]
-    end
-    _growat!.(Ref(a), indices_sorted, 1) # does bound check
-    @inbounds for (index, item) in zip(indices_sorted, items_sorted)
-        a[index] = item
-    end
-    return a
+function sort_indices_items(indices::AbstractRange{<:Integer}, items)
+    indices_sorted = sort(indices) # sorting them is faster than indexing into them
+    sort_indices = sortperm(indices)
+    items_sorted = items[sort_indices]
+    return indices_sorted, items_sorted
 end
 
 function inserteach!(a::Array{T,1}, r::AbstractUnitRange{<:Integer}, items::Array{T,1}) where T
--- a/base/array.jl
+++ b/base/array.jl
@@ -1310,7 +1310,26 @@ function insert!(a::Array{T,1}, indices::AbstractVector, items::Array{T,1}) wher
     return a
 end
 
-function insert!(a::Array{T,1}, r::UnitRange{<:Integer}, items::Array{T,1}) where T
+
+function insert!(a::Array{T,1}, indices::AbstractRange{<:Integer}, items::Array{T,1}) where {T}
+    itemslen = length(items)
+    indiceslen = length(indices)
+    if itemslen !== indiceslen
+        throw(DimensionMismatch("Length of the indices and items should be the same"))
+    elseif itemslen === 0
+        return a
+    end
+    indicessorted = sort(indices) # sorting them is faster than indexing into them
+    sortindices = sortperm(indices) # sorts so the final index is correct
+    items_sorted = items[sortindices]
+    _growat!.(Ref(a), indicessorted, 1) # does bound check
+    @inbounds for (index, item) in zip(indicessorted, items_sorted)
+        a[index] = item
+    end
+    return a
+end
+
+function insert!(a::Array{T,1}, r::AbstractUnitRange{<:Integer}, items::Array{T,1}) where T
     ilen = length(items)
     rlen = length(r)
     if ilen !== rlen
--- a/base/array.jl
+++ b/base/array.jl
@@ -1229,16 +1229,13 @@ Insert all of the elements of `items` into `a` at the given `index`. `index` is 
 
 # Examples
 ```jldoctest
-julia> insert!([6, 5, 4, 2, 1], 4, [8, 9, 9])
-8-element Array{Int64,1}:
- 6
- 5
- 4
+julia> insert!([1, 2, 3], 2, [8, 9])
+5-element Array{Int64,1}:
+ 1
  8
  9
- 9
  2
- 1
+ 3
 ```
 """
 function insert!(a::Array{T,1}, i::Integer, items::Array{T,1}) where {T}
@@ -1261,14 +1258,13 @@ Insert all of the elements of `items` into `a` for the given `range` or `indices
 
 # Examples
 ```jldoctest
-julia> insert!([1, 2, 3, 4], 3:4, [8, 9])
-6-element Array{Int64,1}:
+julia> insert!([1, 2, 3], 2:3, [8, 9])
+5-element Array{Int64,1}:
  1
- 2
  8
  9
+ 2
  3
- 4
 ```
 
 Following example inserts `10, 9, and 8` at index `6 ,4, and 2` respectively
--- a/base/array.jl
+++ b/base/array.jl
@@ -1303,9 +1303,9 @@ function inserteach!(a::Vector, indices::Union{AbstractVector, AbstractRange{<:I
         # sorts so the final index is correct
         indices_sorted, items_sorted = sort_indices_items(indices, items)
     end
-    _growat!.(Ref(a), indices_sorted, 1) # does bound check
-    @inbounds for (index, item) in zip(indices_sorted, items_sorted)
-        a[index] = item
+    for (index, item) in zip(indices_sorted, items_sorted)
+        _growat!(a, index, 1) # TODO write a batch growat in C for speed-up
+        @inbounds a[index] = item
     end
     return a
 end
@@ -1348,9 +1348,9 @@ function inserteach!(a::Vector, boolindices::AbstractVector{Bool}, items::Abstra
     elseif itemslen === 0
         return a
     end
-    _growat!.(Ref(a), indices, 1) # does bound check
-    @inbounds for (index, item) in zip(indices, items)
-        a[index] = item
+    for (index, item) in zip(indices, items)
+        _growat!(a, index, 1) # TODO write a batch growat in C for speed-up
+        @inbounds a[index] = item
     end
     return a
 end
--- a/base/array.jl
+++ b/base/array.jl
@@ -1252,6 +1252,67 @@ function insert!(a::Array{T,1}, i::Integer, items::Array{T,1}) where {T}
     end
     return a
 end
+
+"""
+    insert!(a::Vector, range, items::Vector)
+    insert!(a::Vector, indices::Vector, items::Vector)
+
+Insert all of the elements of `items` into `a` for the given `range` or `indices`. `range` or `indices` are the resulting indices of elements of `items` in the `a`.
+
+# Examples
+```jldoctest
+julia> insert!([1, 2, 3, 4], 3:4, [8, 9])
+6-element Array{Int64,1}:
+ 1
+ 2
+ 8
+ 9
+ 3
+ 4
+```
+
+Following example inserts `10, 9, and 8` at index `6 ,4, and 2` respectively
+```jldoctest
+julia> insert!([1, 2, 3, 4, 5], 6:-2:2, [10, 9, 8])
+8-element Array{Int64,1}:
+  1
+  8
+  2
+  9
+  3
+ 10
+  4
+  5
+```
+"""
+function insert!(a::Array{T1,1}, indices::T2, items::Array{T1,1}) where {T1} where {T2<:Union{Array{<:Integer,1}, OrdinalRange{<:Integer, <:Integer}}}
+    itemslen = length(items)
+    indiceslen = length(indices)
+    if itemslen !== indiceslen
+        throw(DimensionMismatch("Length of the indices and items should be the same"))
+    elseif itemslen === 0
+        return a
+    end
+    sortindices = sortperm(indices) # sorts so the final index is correct
+    indices_sorted = indices[sortindices]
+    items_sorted = items[sortindices]
+    _growat!.(Ref(a), indices_sorted, 1) # does bound check
+    @inbounds for (index, item) in zip(indices_sorted, items_sorted)
+        a[index] = item
+    end
+    return a
+end
+
+function insert!(a::Array{T,1}, r::UnitRange{<:Integer}, items::Array{T,1}) where T
+    ilen = length(items)
+    rlen = length(r)
+    if ilen !== rlen
+        throw(DimensionMismatch("Length of the range and items should be the same"))
+    else
+        return insert!(a, first(r), items)
+    end
+end
+
 """
     deleteat!(a::Vector, i::Integer)
 
--- a/base/array.jl
+++ b/base/array.jl
@@ -1195,6 +1195,8 @@ end
 Insert an `item` into `a` at the given `index`. `index` is the index of `item` in
 the resulting `a`.
 
+See [`insert_elements!`](@ref) for inserting the elements of a vector into another vector.
+
 # Examples
 ```jldoctest
 julia> insert!([6, 5, 4, 2, 1], 4, 3)
@@ -1217,13 +1219,13 @@ function insert!(a::Array{T,1}, i::Integer, item) where T
 end
 
 """
-    insert!(a::Vector, index::Integer, items::Vector)
+    insert_elements!(a::Vector, index::Integer, items::Vector)
 
 Insert all of the elements of `items` into `a` at the given `index`. `index` is the index of first element of `items` in the resulting `a`.
 
 # Examples
 ```jldoctest
-julia> insert!([1, 2, 3], 2, [8, 9])
+julia> insert_elements!([1, 2, 3], 2, [8, 9])
 5-element Array{Int64,1}:
  1
  8
@@ -1232,7 +1234,7 @@ julia> insert!([1, 2, 3], 2, [8, 9])
  3
 ```
 """
-function insert!(a::Array{T,1}, i::Integer, items::Array{T,1}) where {T}
+function insert_elements!(a::Array{T,1}, i::Integer, items::Array{T,1}) where {T}
     m = length(items)
     if m === 0
         return a
@@ -1245,14 +1247,14 @@ function insert!(a::Array{T,1}, i::Integer, items::Array{T,1}) where {T}
 end
 
 """
-    insert!(a::Vector, range, items::Vector)
-    insert!(a::Vector, indices::Vector, items::Vector)
+    insert_elements!(a::Vector, range, items::Vector)
+    insert_elements!(a::Vector, indices::Vector, items::Vector)
 
 Insert all of the elements of `items` into `a` for the given `range` or `indices`. `range` or `indices` are the resulting indices of elements of `items` in the `a`.
 
 # Examples
 ```jldoctest
-julia> insert!([1, 2, 3], 2:3, [8, 9])
+julia> insert_elements!([1, 2, 3], 2:3, [8, 9])
 5-element Array{Int64,1}:
  1
  8
@@ -1263,7 +1265,7 @@ julia> insert!([1, 2, 3], 2:3, [8, 9])
 
 Following example inserts `10, 9, and 8` at index `6 ,4, and 2` respectively
 ```jldoctest
-julia> insert!([1, 2, 3, 4, 5], 6:-2:2, [10, 9, 8])
+julia> insert_elements!([1, 2, 3, 4, 5], 6:-2:2, [10, 9, 8])
 8-element Array{Int64,1}:
   1
   8
@@ -1277,7 +1279,7 @@ julia> insert!([1, 2, 3, 4, 5], 6:-2:2, [10, 9, 8])
 
 It is possible to use boolean indices (each true shows insertion index):
 ```jldoctest
-julia> insert!([1, 2, 3], [true, false, true, false, false], [10, 9])
+julia> insert_elements!([1, 2, 3], [true, false, true, false, false], [10, 9])
 5-element Array{Int64,1}:
  10
   1
@@ -1286,7 +1288,7 @@ julia> insert!([1, 2, 3], [true, false, true, false, false], [10, 9])
   3
 ```
 """
-function insert!(a::Array{T,1}, indices::AbstractVector, items::Array{T,1}) where {T}
+function insert_elements!(a::Array{T,1}, indices::AbstractVector, items::Array{T,1}) where {T}
     itemslen = length(items)
     indiceslen = length(indices)
     if itemslen !== indiceslen
@@ -1310,7 +1312,7 @@ function insert!(a::Array{T,1}, indices::AbstractVector, items::Array{T,1}) wher
 end
 
 
-function insert!(a::Array{T,1}, indices::AbstractRange{<:Integer}, items::Array{T,1}) where {T}
+function insert_elements!(a::Array{T,1}, indices::AbstractRange{<:Integer}, items::Array{T,1}) where {T}
     itemslen = length(items)
     indiceslen = length(indices)
     if itemslen !== indiceslen
@@ -1333,17 +1335,17 @@ function insert!(a::Array{T,1}, indices::AbstractRange{<:Integer}, items::Array{
     return a
 end
 
-function insert!(a::Array{T,1}, r::AbstractUnitRange{<:Integer}, items::Array{T,1}) where T
+function insert_elements!(a::Array{T,1}, r::AbstractUnitRange{<:Integer}, items::Array{T,1}) where T
     ilen = length(items)
     rlen = length(r)
     if ilen !== rlen
         throw(DimensionMismatch("Length of the range and items should be the same"))
     else
-        return insert!(a, first(r), items)
+        return insert_elements!(a, first(r), items)
     end
 end
 
-function insert!(a::Array{T,1}, boolindices::AbstractVector{Bool}, items::Array{T,1}) where T
+function insert_elements!(a::Array{T,1}, boolindices::AbstractVector{Bool}, items::Array{T,1}) where T
     alen = length(a)
     itemslen = length(items)
     boolindiceslen = length(boolindices)
--- a/base/array.jl
+++ b/base/array.jl
@@ -1195,7 +1195,7 @@ end
 Insert an `item` into `a` at the given `index`. `index` is the index of `item` in
 the resulting `a`.
 
-See [`insert_elements!`](@ref) for inserting the elements of a vector into another vector.
+See [`inserteach!`](@ref) for inserting the elements of a vector into another vector.
 
 # Examples
 ```jldoctest
@@ -1219,13 +1219,13 @@ function insert!(a::Array{T,1}, i::Integer, item) where T
 end
 
 """
-    insert_elements!(a::Vector, index::Integer, items::Vector)
+    inserteach!(a::Vector, index::Integer, items::Vector)
 
 Insert all of the elements of `items` into `a` at the given `index`. `index` is the index of first element of `items` in the resulting `a`.
 
 # Examples
 ```jldoctest
-julia> insert_elements!([1, 2, 3], 2, [8, 9])
+julia> inserteach!([1, 2, 3], 2, [8, 9])
 5-element Array{Int64,1}:
  1
  8
@@ -1234,7 +1234,7 @@ julia> insert_elements!([1, 2, 3], 2, [8, 9])
  3
 ```
 """
-function insert_elements!(a::Array{T,1}, i::Integer, items::Array{T,1}) where {T}
+function inserteach!(a::Array{T,1}, i::Integer, items::Array{T,1}) where {T}
     m = length(items)
     if m === 0
         return a
@@ -1247,14 +1247,14 @@ function insert_elements!(a::Array{T,1}, i::Integer, items::Array{T,1}) where {T
 end
 
 """
-    insert_elements!(a::Vector, range, items::Vector)
-    insert_elements!(a::Vector, indices::Vector, items::Vector)
+    inserteach!(a::Vector, range, items::Vector)
+    inserteach!(a::Vector, indices::Vector, items::Vector)
 
 Insert all of the elements of `items` into `a` for the given `range` or `indices`. `range` or `indices` are the resulting indices of elements of `items` in the `a`.
 
 # Examples
 ```jldoctest
-julia> insert_elements!([1, 2, 3], 2:3, [8, 9])
+julia> inserteach!([1, 2, 3], 2:3, [8, 9])
 5-element Array{Int64,1}:
  1
  8
@@ -1265,7 +1265,7 @@ julia> insert_elements!([1, 2, 3], 2:3, [8, 9])
 
 Following example inserts `10, 9, and 8` at index `6 ,4, and 2` respectively
 ```jldoctest
-julia> insert_elements!([1, 2, 3, 4, 5], 6:-2:2, [10, 9, 8])
+julia> inserteach!([1, 2, 3, 4, 5], 6:-2:2, [10, 9, 8])
 8-element Array{Int64,1}:
   1
   8
@@ -1279,7 +1279,7 @@ julia> insert_elements!([1, 2, 3, 4, 5], 6:-2:2, [10, 9, 8])
 
 It is possible to use boolean indices (each true shows insertion index):
 ```jldoctest
-julia> insert_elements!([1, 2, 3], [true, false, true, false, false], [10, 9])
+julia> inserteach!([1, 2, 3], [true, false, true, false, false], [10, 9])
 5-element Array{Int64,1}:
  10
   1
@@ -1288,7 +1288,7 @@ julia> insert_elements!([1, 2, 3], [true, false, true, false, false], [10, 9])
   3
 ```
 """
-function insert_elements!(a::Array{T,1}, indices::AbstractVector, items::Array{T,1}) where {T}
+function inserteach!(a::Array{T,1}, indices::AbstractVector, items::Array{T,1}) where {T}
     itemslen = length(items)
     indiceslen = length(indices)
     if itemslen !== indiceslen
@@ -1312,7 +1312,7 @@ function insert_elements!(a::Array{T,1}, indices::AbstractVector, items::Array{T
 end
 
 
-function insert_elements!(a::Array{T,1}, indices::AbstractRange{<:Integer}, items::Array{T,1}) where {T}
+function inserteach!(a::Array{T,1}, indices::AbstractRange{<:Integer}, items::Array{T,1}) where {T}
     itemslen = length(items)
     indiceslen = length(indices)
     if itemslen !== indiceslen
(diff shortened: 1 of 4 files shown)


aminya:
@mortenpi is this doctest error because of indentation?

mortenpi:
No, I'm pretty sure that you would get the same error if you'd run the code snippet (`insert!([1, 2, 3], [true, false, true, false, false], [10, 9])`) in the REPL.

aminya:
> 
> 
> No, I'm pretty sure that you would get the same error if you'd run the code snippet (`insert!([1, 2, 3], [true, false, true, false, false], [10, 9])`) in the REPL.

Aye, thanks!, the problem is that `Array{Bool,1} <: Array{<:Integer,1} ` and it triggers the wrong method

KristofferC:
Isn't this just `splice!`?

```
julia> a = [1, 2, 3];

julia> splice!(a, 2:1, [8, 9]);

julia> a
5-element Array{Int64,1}:
 1
 8
 9
 2
 3
```

aminya:
> 
> 
> Isn't this just `splice!`?
> 
> ```julia
> julia> a = [1, 2, 3];
> 
> julia> splice!(a, 2:1, [8, 9]);
> 
> julia> a
> 5-element Array{Int64,1}:
>  1
>  8
>  9
>  2
>  3
> ```

Only the method at https://github.com/JuliaLang/julia/pull/34635/files#diff-8cc03187983013adb308460f8365e1d0R1241 can be done by `splice!`.
It can't insert when the range is a vector itself (that may have jumps).

Using `splice!` for this is [misusing of its algorithm]((https://github.com/JuliaLang/julia/blob/8a084003bbd41fa59b00367f29dd48d9070d40ed/base/array.jl#L1435) since you should give some `range` that has a length of `0` (or less than the length of `items`, so the following gives the same answer no matter what `r` is:
```julia
julia> a=[1,2,3,4,5];

julia> splice!(a,4:-1,[9,9]); a
7-element Array{Int64,1}:
 1
 2
 3
 9
 9
 4
 5

julia> a=[1,2,3,4,5];

julia> splice!(a,4:2,[9,9]); a
7-element Array{Int64,1}:
 1
 2
 3
 9
 9
 4
 5
```

Even for this single method `splice!` is slower
```julia
const d = rand(1000)
const e = 10
const f = rand(100)
@btime insert!(copy(d), e, f)
 # 15.500 μs (3 allocations: 156.33 KiB)
@btime splice!(copy(d), e:e-1, f)
 # 16.400 μs (4 allocations: 156.41 KiB)
```

KristofferC:
> Using splice! for inserting is misusing of its algorithm 

Hm, it says in the docs for `splice!`:

> To insert replacement before an index `n` without removing any items, use `splice!(collection, n:n-1,
  replacement).`

so I am not sure it really is a "misuse".

StefanKarpinski:
I'm afraid this is not something that is ok. Consider this case:
```jl
julia> a = Any[1, [2, 3], 4, 5]
4-element Array{Any,1}:
 1
  [2, 3]
 4
 5

julia> insert!(a, 3, [6, 7])
5-element Array{Any,1}:
 1
  [2, 3]
  [6, 7]
 4
 5
```
This PR changes the current behavior of `insert!` in that case. Sometimes elements are arrays.

aminya:
> 
> 
> I'm afraid this is not something that we can do. Consider this case:
> 
> ```julia
> julia> a = Any[1, [2, 3], 4, 5]
> 4-element Array{Any,1}:
>  1
>   [2, 3]
>  4
>  5
> 
> julia> insert!(a, 3, [6, 7])
> 5-element Array{Any,1}:
>  1
>   [2, 3]
>   [6, 7]
>  4
>  5
> ```
> 
> This PR changes the current behavior of `insert!` in that case. Sometimes elements are arrays.

This is because of this method : 
https://github.com/JuliaLang/julia/pull/34635/files#diff-8cc03187983013adb308460f8365e1d0R1219

Do you know how to fix `T` to be exactly the same for both `items` and `collection`?
I don't want the collection to be of `eltype` `Any` and `items` be `Int64`. They should have the same types

yuyichao:
> Do you know how to fix T to be exactly the same for both items and collection?
I don't want the collection to be of eltype Any and items be Int64. They should have the same types

No that does not help. You can have T being a super type of Array{T}, e.g. when T is Any. It’s an true ambiguity that you cannot possibly fix.

aminya:
@StefanKarpinski I removed Insert! specialization for the items with the same type of Vector. It should throw `convert` error if the types are not the same.
However, for your example on a clean Julia installation the same behavior happens:
```julia
julia> insert!( Any[1, [2, 3], 4, 5], 3, [6, 7])
5-element Array{Any,1}:
 1
  [2, 3]
  [6, 7]
 4
 5

```

aminya:
> 
> 
> > Do you know how to fix T to be exactly the same for both items and collection?
> > I don't want the collection to be of eltype Any and items be Int64. They should have the same types
> 
> No that does not help. You can have T being a super type of Array{T}, e.g. when T is Any. It’s an true ambiguity that you cannot possibly fix. (In another word, you cannot fix the issue when someone want to insert an Array{Any} to another Array{Any}. Picking either behavior is wrong)

Based on what you say, the following call triggers the default insert behavior and doesn't trigger the new element insertion.
```julia
julia> insert!( [2,"2"], 3, [6, 7])
3-element Array{Any,1}:
 2
  "2"
  [6, 7]
```
But something like `insert_elements!` should trigger a method like.
```julia
julia> insert_elements!( [2,"2"], 3, [6, 7])
4-element Array{Any,1}:
 2
"2"
6
7
```
So do you think I should introduce a new function like `insert_elements!`?

When `range` or `indices` are not integers non-of these issues happen though.

yuyichao:
What I'm saying is that if you want to have something that splice an array into another, which seems to be what this PR is all about, it must not be `insert!`. (That or you find a different name for `insert!` of course but that'll be very breaking).

As for whether such a function should be added, what it should be called, whether it should/can be added as a method for `splice!`, etc. I currently have no opinion on.

yuyichao:
Also this may also apply to the cases where the index is an array. For insert that might as well means inserting the same element at multiple locations.

aminya:
> 
> 
> What I'm saying is that if you want to have something that splice an array into another, which seems to be what this PR is all about, it must not be `insert!`. (That or you find a different name for `insert!` of course but that'll be very breaking).
> 
> As for whether such a function should be added, what it should be called, whether it should/can be added as a method for `splice!`, etc. I currently have no opinion on.

The default behavior of `splice!` is to remove the elements (and only works for a single integer index and by hacking the given range). 

But introducing a new function like `insert_elements!` would be more suitable, and it will not break anything.

yuyichao:
> The default behavior of splice! is to remove the elements (and only works for a single integer index and by hacking the given range).

As I said, I saw your argument above and I don't want to argue about it since I don't have a strong opinion either way. I'm not suggesting you should have used `splice!` instead... I only strongly think none of the methods added in this PR should be methods of `insert!`....

----

(I do kind of doubt `insert_elements!` will be very popular for people with strong opinion on names but I don't really care about the name either....)

aminya:
> 
> 
> > The default behavior of splice! is to remove the elements (and only works for a single integer index and by hacking the given range).
> 
> As I said, I saw your argument above and I don't want to argue about it since I don't have a strong opinion either way. I'm not suggesting you should have used `splice!` instead... I only strongly think none of the methods added in this PR should be methods of `insert!`....
> 
> (I do kind of doubt `insert_elements!` will be very popular for people with strong opinion on names but I don't really care about the name either....)

I agree. 

Other than `insert_elements!`, which IMO is the most descriptive one, there are other names to choose from like `thrust!`. 
I am not concerned about the name either. I just want to see what people think about it.

StefanKarpinski:
I don't understand what the objection to doing these operations with `splice!` is. It is specifically intended for this.

c42f:
How about `inserteach!`? That would evoke naming consistently with various iteration-based APIs such as `foreach` and `eachslice`.

c42f:
> `splice!` is. It is specifically intended for this

`insert_elements!` seems to be catering for the uneasy middle ground between `splice!` and `insert!`? It's like `insert!` (and unlike `splice!`) in that it doesn't delete elements. It's like `splice!` and unlike `insert!` because it deals with multiple elements.

aminya:
> 
> 
> I don't understand what the objection to doing these operations with `splice!` is. It is specifically intended for this.

As I said `splice!` only can insert at one index.  It can't do when `indices` are a vector themselves.
For example: 
```julia
julia> insert_elements!([1, 2, 3, 4, 5], 6:-2:2, [10, 9, 8])
8-element Array{Int64,1}:
  1
  8
  2
  9
  3
 10
  4
  5
```

See my previous comment https://github.com/JuliaLang/julia/pull/34635#issuecomment-581333419

aminya:
> 
> 
> > `splice!` is. It is specifically intended for this
> 
> `insert_elements!` seems to be catering for the uneasy middle ground between `splice!` and `insert!`? It's like `insert!` (and unlike `splice!`) in that it doesn't delete elements. It's like `splice!` and unlike `insert!` because it deals with multiple elements.

Thanks for the suggestion! I like it. I am going to change the name to `inserteach!` now.

kmsquire:
Just curious, what is the use case for this version of insert?  Is there an issue where it was discussed?  Are there other languages that have this functionality?

aminya:
> Just curious, what is the use case for this version of insert? Is there an issue where it was discussed? Are there other languages that have this functionality?

It is an algorithm that can be used for a lot of different applications. I usually ended up adding my own.

It allows running different algorithms with different goals independent of each other. Like doing some sort first. Then inserting some other array into our array, etc. 

As an example:
Imagine your Instagram feed being an array that its order is defined (sorted based on your interests). Then Instagram decides to insert ads for every 10 posts. You can do it using
```julia
sort!(posts) # based on interests
insert!(posts, 1:5:length(posts), ads)
```
or imagine you have a song (like an array) that you want to insert a drum crash every 4 bars. I just made these up. You can think of many other things.

c42f:
> what is the use case for this version of insert

From an algorithmic viewpoint, it's more efficient to do batch insertion.

For arrays of length `N` and a number of (sorted!) insertion points `M` it's an `O(N + M)` algorithm to do the insertion in a batch and a single reallocation. This is opposed to an `O(N*M)` algorithm to iteratively `insert!` because you only need to move the existing elements once. (Not to mention `insert!` is mutating so iteratively inserting needs to keep track of how the insertion points change which is annoying.)

aminya:
For the reference, the current algorithm for the case that `range` is a vector uses broadcasting of `_growat!` which is a `ccall` to `jl_array_growat`: https://github.com/JuliaLang/julia/blob/3720edfc06ee8687b54ee7fbb908c9f03f4fc557/src/array.c#L735. 
One may be able to improve the speed further if they write a `_growatbatch!` in C instead.

In the OP, alternative algorithms that don't do this exist but they are all slower (now as of Julia 1.5).

c42f:
> uses broadcasting of `_growat!`

That sounds kind of iffy to me, especially because `broadcast` is mean to produce output but you're using it for side effects. Can't you just `resize!` the output array, do a reverse scan of the output moving/copying the existing and new elements to their final positions? You'd need to carefully keep track of the insertion point using the sorted list of insertion locations.

aminya:
> 
> 
> > uses broadcasting of `_growat!`
> 
> That sounds kind of iffy to me, especially because `broadcast` is mean to produce output but you're using it for side effects. Can't you just `resize!` the output array, do a reverse scan of the output moving/copying the existing and new elements to their final positions? You'd need to carefully keep track of the insertion point using the sorted list of insertion locations.

I think this is what `inserteach3!` in the OP does, however, it isn't fast enough.

StefanKarpinski:
> As I said `splice!` only can insert at one index. It can't do when `indices` are a vector themselves.

That would straightforward to fix. Unlike values, where you have to consider the possibility that a value can itself be a collection, when index is a collection, it's safe to dispatch on that. So adding these behaviors to `splice! is possible. There seem to be two additional behavior that would be wanted here:

* `splice!(a::Vector, r::StepRange, replacement)` to replace items at each index produced by `r` with the values produced by `replacement`;
* `splice!(a::Vector, rs::AbstractVector{<:Union{Integer,Range}}, replacement)` to replace items at each of the locations indicated by the collection of indices/ranges of indices in `rs`.

If the existing `splice!` is slower than your insert operation, then `splice!` should be made faster. I do worry with this potential API that it is getting into crazy vectorized programming territory, however. It seems like the main issue with just splicing stuff in a loop is the potential for repeated array resizing. That could be handled by just having an easy way to figure out the new size that the final array will have and calling `sizehint!` on the array before the splicing.

aminya:
> 
> 
> > As I said `splice!` only can insert at one index. It can't do when `indices` are a vector themselves.
> 
> That would straightforward to fix. Unlike values, where you have to consider the possibility that a value can itself be a collection, when index is a collection, it's safe to dispatch on that. So adding these behaviors to `splice! is possible. There seem to be two additional behavior that would be wanted here:
> 
>     * `splice!(a::Vector, r::StepRange, replacement)` to replace items at each index produced by `r` with the values produced by `replacement`;
> 
>     * `splice!(a::Vector, rs::AbstractVector{<:Union{Integer,Range}}, replacement)` to replace items at each of the locations indicated by the collection of indices/ranges of indices in `rs`.
> 
> 
> If the existing `splice!` is slower than your insert operation, then `splice!` should be made faster. I do worry with this potential API that it is getting into crazy vectorized programming territory, however. It seems like the main issue with just splicing stuff in a loop is the potential for repeated array resizing. That could be handled by just having an easy way to figure out the new size that the final array will have and calling `sizehint!` on the array before the splicing.

Using these methods for `splice` would be breaking. I don't want to break any code which is already running without any issues.

The behavior of `splice!` is different from `insert`. `splice` replaces items, but `insert` inserts. `splice` thinks of the given `range/index` as the old indices in the old array, but `insert` thinks of the given `rang/index` as the resulting indices in the final array.

- `splice!(a::Vector, index::Integer, [replacement]) -> item` removes the item at `index` and then replaces them with `replacement`. There is no way to tell `splice` not to remove that old item (this is expected behavior). Also regarding the function output, it returns the removed `item` and not the original `a`.

- `splice!(a::Vector, range, [replacement]) -> items` removes the items of given `range` and then replaces them with `replacement`. 
- `splice!(a::Vector, range, [replacement]) -> items` removes the items of given `range` and then replaces them with `replacement`. 

This method has a `bug` (IMO) that you can give a `range` that doesn't make sense. Like `4:2`. Then algorithm uses the `first` of this non-existent range, and thinks of it as an `index`. So no matter what the 2nd number of `range` is (like `4:2, 4:-2, 4:-1000000000, 4:-123456789, etc. sky is the limit), it will add the replacement at index 4.
In contrast `inserteach!` returns an error if non-sense `range` is given, and you can't remove.

Because `splice` thinks of the indices as the old original indices it will return error for such example:
```julia
 a = [1, 2, 3, 4]
 splice!(a, 2:5, [10,9,8,7])  # index 5 doesn't exist in the old array so it returns error
```

However, with that being said, I can make another PR to add some methods to `splice` so it can have capability of replacing at given indices (that are not necessarily `range`).

StefanKarpinski:
> Using these methods for `splice` would be breaking.

`splice!(v, 2:3:10, [4, 5, 6])` and `splice!(v, [2:3:10, 1], [4, 5, 6, 7])` are currently method errors, so I'm not sure how that would be breaking.

> There is no way to tell splice not to remove that old item (this is expected behavior)

Passing it an empty range at a given index does tell it that. So you could pass a collection of empty ranges at the locations you want to insert items. Still, I agree that's a somewhat awkward API.

> This method has a bug (IMO) that you can give a range that doesn't make sense.

It's very much intentional, not a bug. It is based on [Perl's `splice` function](https://perldoc.perl.org/functions/splice.html) which also behaves this way:
```perl
  DB<6> x @v = 1..10
0  1
1  2
2  3
3  4
4  5
5  6
6  7
7  8
8  9
9  10
  DB<7> x splice @v, 2, 0, 11, 12
  empty array
  DB<8> x @v
0  1
1  2
2  11
3  12
4  3
5  4
6  5
7  6
8  7
9  8
10  9
11  10
```
It is fairly common in Julia for APIs where a range of indices are passed to allow an empty range at a particular offset. It's quite useful in many cases and the correct behavior generally falls out of using this in combination with search methods that indicate the lack of elements at a particular location by returning the same kind of empty range. For example, suppose I have a sorted array and I want to replace any occurrence the value 5 with the pair of values 4.5 and 5.5:
```jl
julia> v = Float64[1, 2, 4, 5, 5, 7];

julia> splice!(v, searchsorted(v, 5), (4.5, 5.5))
2-element Array{Float64,1}:
 5.0
 5.0

julia> v
6-element Array{Float64,1}:
 1.0
 2.0
 4.0
 4.5
 5.5
 7.0
```
Great, it works. What if I try it with a value that's not present? It still works:
```jl
julia> splice!(v, searchsorted(v, 3), (2.5, 3.5))
0-element Array{Float64,1}

julia> v
8-element Array{Float64,1}:
 1.0
 2.0
 2.5
 3.5
 4.0
 4.5
 5.5
 7.0
```
It inserts the values 2.5, 3.5 at the location where 3 would have been in the sorted vector.

> So no matter what the 2nd number of `range` is (like `4:2, 4:-2, 4:-1000000000, 4:-123456789`, etc. the sky is the limit), it will add the replacement at index 4.

This is because `4:2` and `4:-2` are the same range: the empty range starting at 4:
```jl
julia> 4:2
4:3

julia> 4:-2
4:3
```

Given the awkwardness of doing this by passing `splice!` a bunch of empty index ranges to insert items at, I can see the argument for `inserteach!` and the name is not bad for that functionality. I'll make some comments about the implementation.

base/array.jl
```diff
@@ -1216,6 +1218,154 @@ function insert!(a::Array{T,1}, i::Integer, item) where T
...
+    ilen = length(items)
+    rlen = length(r)
+    if ilen !== rlen
+        throw(DimensionMismatch("Length of the range and items should be the same"))
+    else
+        return inserteach!(a, first(r), items)
+    end
+end
+
+function inserteach!(a::Array{T,1}, boolindices::AbstractVector{Bool}, items::Array{T,1}) where T
```
StefanKarpinski: Not sure we need this method. Seems like the user could call `inserteach!(a, findall(boolindices), items)` if they really need this.

base/array.jl
```diff
@@ -1216,6 +1218,154 @@ function insert!(a::Array{T,1}, i::Integer, item) where T
...
+    end
+    _growat!.(Ref(a), indices_sorted, 1) # does bound check
+    @inbounds for (index, item) in zip(indices_sorted, items_sorted)
+        a[index] = item
+    end
+    return a
+end
+
+
+function inserteach!(a::Array{T,1}, indices::AbstractRange{<:Integer}, items::Array{T,1}) where {T}
```
StefanKarpinski: This method should be combined with the previous one, factoring out the index/item sorting logic, which seems to be the only thing that differs, into a `indices, items = sort_indices_items(indices, items)` helper method.

base/array.jl
```diff
@@ -1216,6 +1218,154 @@ function insert!(a::Array{T,1}, i::Integer, item) where T
...
+julia> inserteach!([1, 2, 3], [true, false, true, false, false], [10, 9])
+5-element Array{Int64,1}:
+ 10
+  1
+  9
+  2
+  3
+```
+"""
+function inserteach!(a::Array{T,1}, indices::AbstractVector, items::Array{T,1}) where {T}
```
StefanKarpinski:
```suggestion
function inserteach!(a::Vector, indices::AbstractVector, items:: AbstractVector)
```

base/array.jl
StefanKarpinski:
```suggestion
function inserteach!(a::Vector, indices::AbstractRange{<:Integer}, items::AbstractVector)
```

base/array.jl
```diff
@@ -1216,6 +1218,154 @@ function insert!(a::Array{T,1}, i::Integer, item) where T
...
+julia> inserteach!([1, 2, 3], 2, [8, 9])
+5-element Array{Int64,1}:
+ 1
+ 8
+ 9
+ 2
+ 3
+```
+"""
+function inserteach!(a::Array{T,1}, i::Integer, items::Array{T,1}) where {T}
```
StefanKarpinski:
```suggestion
function inserteach!(a::Array, i::Integer, items::AbstractVector)
```

StefanKarpinski (changes requested):
This will also require generic definitions for when the first argument is an AbstractVector, even though that situation won't be optimized as well—if this is going to be a generic operation, it has to work generically. It might be good to make the index and item sorting code generic and then have a helper function for inserting a set of items at pre-sorted indices. Another thing to do would be to generalize the `items` argument to support anything collectable, although that does start to get a bit tricky.

StefanKarpinski:
General note on generic collections operations: it's tempting to try to make the element types match, but we generally don't do that, instead relying on convert machinery to either fail (if the item cannot be converted to the correct type) or do the conversion when assigning into a collection.

base/array.jl
aminya: It is more optimized since it doesn't check for the sort (no `issorted`).

aminya:
> 
> 
> General note on generic collections operations: it's tempting to try to make the element types match, but we generally don't do that, instead relying on convert machinery to either fail (if the item cannot be converted to the correct type) or do the conversion when assigning into a collection.

Do I need to update the docstrings too? 
It says `items::Vector`
https://github.com/JuliaLang/julia/pull/34635/files#diff-8cc03187983013adb308460f8365e1d0R1251

aminya:
We also need to write the functions for `BitVector`, but I am not that familiar with them now. Maybe I make a PR later for 
```julia
inserteach!(x::BitVector, indices, items::BitVector)
```

base/array.jl
```diff
@@ -1216,6 +1218,143 @@ function insert!(a::Array{T,1}, i::Integer, item) where T
...
+julia> inserteach!([1, 2, 3], 2:3, [8, 9])
+5-element Array{Int64,1}:
+ 1
+ 8
+ 9
+ 2
+ 3
+```
+
+Following example inserts `10, 9, and 8` at index `6 ,4, and 2` respectively
```
StefanKarpinski:
```suggestion
Following example inserts `10, 9, and 8` at index `6, 4, and 2` respectively
```

StefanKarpinski:
It would be good to add some corner-case tests when the range of indices and values to be inserted are length zero. That's the kind of corner case where otherwise-working code often fails.

base/array.jl
```diff
@@ -1216,6 +1218,143 @@ function insert!(a::Array{T,1}, i::Integer, item) where T
...
+julia> inserteach!([1, 2, 3], 2, [8, 9])
+5-element Array{Int64,1}:
+ 1
+ 8
+ 9
+ 2
+ 3
+```
+"""
+function inserteach!(a::Vector, i::Integer, items::AbstractVector)
```
JeffBezanson: This method does not seem consistent with the others --- in the other cases, the indices and items have the same length.

StefanKarpinski:
Some feedback from triage. At a high level, it's unclear that such a complex new API is really necessary. Can you provide some benchmarks showing the performance advantage of this over just doing repeated insertion? The things to compare with would be:

1. just doing the insertions repeatedly
2. resize storage with a single sizehint! call, then do the insertions repeatedly
3. using `inserteach!`

The reason this needs performance motivation is that without the performance advantage, there's no reason to stuff this complex functionality into vectorized operation.

If there is a clear performance justification, then the API needs to be considered carefully. Does the collection of indices and values have to be the same length? If not, then what's the rationale for that and how is it handled? It seems like this API should definitely not support boolean indices.

base/array.jl
aminya: Here the index is a single `Integer`. I renamed `i` to `index` for clarification.

JeffBezanson: The problem isn't the name of the argument, it's the behavior. It's this difference:
```
julia> a[1:1] = [1]
1-element Array{Int64,1}:
 1

julia> a[1] = [1]
ERROR: MethodError: Cannot `convert` an object of type Array{Int64,1} to an object of type Int64
```

Keno:
Another observation from triage. There's also broadcasted insert!, i.e. `insert!.(&a, 1, [4,5,6])` which behaves similar to this.

base/array.jl
aminya: These are two different methods, they are not meant to be the same:
- `function inserteach!(a::Vector, index::Integer, items::AbstractVector)`:
Inserts all the items at index `i`. 1st item becomes `i`th.

- `function inserteach!(a::Vector, indices::Vector, items::AbstractVector)`:
Inserts the items at the indices given. Indices now has a 1-1 association with items.

aminya:
> insert!.(&a, 1, [4,5,6])

I am not sure what you mean by `&a`, but if it is `Ref()` I should say no, 
```julia
function inserteach_singleindex_broadcast!(a::Vector, index::Integer, items::AbstractVector)
    insert!.(Ref(a), index, items)
end
using Test
a = [1, 2, 3, 4]; b = 3; c = [10, 9, 8]
@test inserteach!(a, b, c) == inserteach_singleindex_broadcast!(a, b, c) # fails
```

aminya:
> 
> 
> Some feedback from triage. At a high level, it's unclear that such a complex new API is really necessary. Can you provide some benchmarks showing the performance advantage of this over just doing repeated insertion? The things to compare with would be:
> 
>     1. just doing the insertions repeatedly
> 
>     2. resize storage with a single sizehint! call, then do the insertions repeatedly
> 
>     3. using `inserteach!`
> 
> 
> The reason this needs performance motivation is that without the performance advantage, there's no reason to stuff this complex functionality into vectorized operation.
> 
> If there is a clear performance justification, then the API needs to be considered carefully. Does the collection of indices and values have to be the same length? If not, then what's the rationale for that and how is it handled? It seems like this API should definitely not support boolean indices.

Benchmark is updated in OP:
## Performance

Specialization is done for AbstractRanges, AbstractUnitRages, and AbstractBool.
To get the best algorithm for general AbstractVector, I tried 5 other alternatives, and what is given in the PR is the fastest:
<details>
  <summary>Spoiler</summary>

### Single Index
`inserteach!` is much faster than the others.
```julia
function inserteach!(a::Vector, index::Integer, items::AbstractVector)
    m = length(items)
    if m === 0
        return a
    end
    _growat!(a, index, m)  # does bound check
    @inbounds for k = index:index+m-1
        a[k] = items[k-index+1]
    end
    return a
end

function inserteach_singleindex_2!(a::Vector, index::Integer, items::AbstractVector)
    for (i, item) in enumerate(items)
        insert!(a, i-1+index, item)
    end
    return a
end

function inserteach_singleindex_3!(a::Vector, index::Integer, items::AbstractVector)
    sizehint!(a, length(items)+length(a))
    for (i, item) in enumerate(items)
        insert!(a, i-1+index, item)
    end
    return a
end
```
### Single-index result
```julia
a = [1, 2, 3, 4]
b = 3
c = [10, 9, 8]

using Test; @test inserteach!(copy(a), b, c) == inserteach_singleindex_2!(copy(a), b, c) == inserteach_singleindex_3!(copy(a), b ,c)

using BenchmarkTools
@btime inserteach!($copy(a), $b, $c)
  # 97.465 ns (2 allocations: 192 bytes)
@btime inserteach_singleindex_2!($copy(a), $b, $c)
  # 115.888 ns (2 allocations: 192 bytes)
@btime inserteach_singleindex_3!($copy(a), $b, $c)
  # 119.802 ns (2 allocations: 192 bytes)

vsize = 1000
d = rand(Float32, vsize)
e = 30
f = rand(Float32, vsize÷10)

@btime inserteach!($copy(d), $e, $f)
  # 880.000 ns (2 allocations: 11.94 KiB)
@btime inserteach_singleindex_2!($copy(d), $e, $f)
  # 2.678 μs (2 allocations: 11.94 KiB)
@btime inserteach_singleindex_3!($copy(d), $e, $f)
  # 2.862 μs (2 allocations: 11.94 KiB)

g = rand(vsize)
h = 30
i = rand(vsize÷10)
@btime inserteach!($copy(g), $h, $i)
  # 1.480 μs (2 allocations: 23.63 KiB)
@btime inserteach_singleindex_2!($copy(g), $h, $i)
  # 3.825 μs (2 allocations: 23.63 KiB)
@btime inserteach_singleindex_3!($copy(g), $h, $i)
  # 4.243 μs (2 allocations: 23.63 KiB)

```

### Multi-Index
For larger sizes `inserteach!` is a lot faster. `inserteach_multiindex7!` has a similar speed to `inserteach!` until a batch `growat!` is written in C

```julia
function inserteach!(a::Vector, indices::Union{AbstractVector, AbstractRange{<:Integer}}, items::AbstractVector)
    itemslen = length(items)
    indiceslen = length(indices)
    if itemslen !== indiceslen
        throw(DimensionMismatch("Length of the indices and items should be the same"))
    elseif itemslen === 0
        return a
    end
    if issorted(indices) # based on the tests overhead/advantage is low
        indices_sorted = indices
        items_sorted = items
    else
        # sorts so the final index is correct
        indices_sorted, items_sorted = sort_indices_items(indices, items)
    end
    # _growat!.(Ref(a), indices_sorted, 1) # does bound check
    for (index, item) in zip(indices_sorted, items_sorted)
         _growat!(a, index, 1)
        @inbounds a[index] = item
    end
    return a
end

function sort_indices_items(indices::AbstractVector, items)
    sort_indices = sortperm(indices)
    indices_sorted = indices[sort_indices]
    items_sorted = items[sort_indices]
    return indices_sorted, items_sorted
end

function inserteach_multiindex_2!(a::Vector{T}, indices::AbstractVector, items::AbstractVector) where T
# defines a new array, finds the 1:newlength indices that are not in the given indices (rest), then populates the new array based on that.
    itemslen = length(items)
    indiceslen = length(indices)
    if itemslen !== indiceslen
        throw(DimensionMismatch("Length of the indices and items should be the same"))
    elseif itemslen === 0
        return a
    end
    alen = length(a)
    v = Vector{T}(undef, alen+itemslen)
    # rest = filter(x -> !in(x, indices), 1:alen+itemslen)
    rest = .!in.(1:alen+itemslen, Ref(indices))
   @inbounds  v[rest] .= a
    v[indices] .= items
    return v
end

function inserteach_multiindex_3!(a::Vector, indices::AbstractVector, items::AbstractVector)
# grows the array at the end for the length of indices, then finds the rest of the indices that are not part of given indices (rest), then populates the array based on the indices.
    itemslen = length(items)
    indiceslen = length(indices)
    if itemslen !== indiceslen
        throw(DimensionMismatch("Length of the indices and items should be the same"))
    elseif itemslen === 0
        return a
    end
    alen = length(a)
    _growat!(a, alen+1, indiceslen) # does bound check
    # rest = filter(x -> !in(x, indices), 1:alen+itemslen)
     rest = .!in.(1:alen+itemslen, Ref(indices))
    @inbounds begin
        a[rest] .= a[1:alen]
        a[indices] .= items
    end
    return a
end

function inserteach_multiindex_4!(a::Vector{T}, indices::AbstractVector, items::AbstractVector) where T
# sorts the indices. Then in a loop populates the array based on the fact that an index of the new array is in the given indices. We need sort because `in` only gives true (not the index).
    itemslen = length(items)
    indiceslen = length(indices)
    if itemslen !== indiceslen
        throw(DimensionMismatch("Length of the indices and items should be the same"))
    elseif itemslen === 0
        return a
    end
    alen = length(a)
    v = Vector{T}(undef, alen+itemslen)
    if issorted(indices) # based on the tests overhead/advantage is low
        indices_sorted = indices
        items_sorted = items
    else
        # sorts so the final index is correct
        indices_sorted, items_sorted = sort_indices_items(indices, items)
    end
    kitem = 1
    ka=1
    @inbounds for k=1:alen+itemslen
        if k in indices_sorted
            v[k] = items_sorted[kitem]
            kitem+=1
        else
            v[k] = a[ka]
            ka+=1
        end
    end
    return v
end

function inserteach_multiindex_5!(a::Vector{T}, indices::AbstractVector, items::AbstractVector) where T
# In a loop populates the array based on the finded index of the new array is in the given indices.
    itemslen = length(items)
    indiceslen = length(indices)
    if itemslen !== indiceslen
        throw(DimensionMismatch("Length of the indices and items should be the same"))
    elseif itemslen === 0
        return a
    end
    alen = length(a)
    v = Vector{T}(undef, alen+itemslen)
    kitem = 1
    ka=1
    @inbounds for k=1:alen+itemslen
        kitem = findfirst(x -> x == k, indices)
        if isnothing(kitem)
            v[k] = a[ka]
            ka+=1
        else
            v[k] = items[kitem]
            kitem+=1
        end
    end
    return v
end

function inserteach_multiindex_6!(a::Vector{T}, indices::AbstractVector, items::AbstractVector) where T
# Similar to 6 but first checks `in` then uses `findfirst` only for the ones that already exist
    itemslen = length(items)
    indiceslen = length(indices)
    if itemslen !== indiceslen
        throw(DimensionMismatch("Length of the indices and items should be the same"))
    elseif itemslen === 0
        return a
    end
    alen = length(a)
    v = Vector{T}(undef, alen+itemslen)
    kitem = 1
    ka=1
    @inbounds for k=1:alen+itemslen
        if k in indices
            kitem = findfirst(x -> x == k, indices)
            v[k] = items[kitem]
            kitem+=1
        else
            v[k] = a[ka]
            ka+=1
        end
    end
    return v
end


function inserteach_multiindex_7!(a::Vector, indices::AbstractVector, items::AbstractVector)
# For loops around insert
    itemslen = length(items)
    indiceslen = length(indices)
    if itemslen !== indiceslen
        throw(DimensionMismatch("Length of the indices and items should be the same"))
    elseif itemslen === 0
        return a
    end
    if issorted(indices) # based on the tests overhead/advantage is low
        indices_sorted = indices
        items_sorted = items
    else
        # sorts so the final index is correct
        indices_sorted, items_sorted = sort_indices_items(indices, items)
    end
    for (index, item) in zip(indices_sorted, items_sorted)
        insert!(a, index, item)
    end
    return a
end

function inserteach_multiindex_8!(a::Vector, indices::AbstractVector, items::AbstractVector)
# For loops around insert with sizehint
    itemslen = length(items)
    indiceslen = length(indices)
    if itemslen !== indiceslen
        throw(DimensionMismatch("Length of the indices and items should be the same"))
    elseif itemslen === 0
        return a
    end
    if issorted(indices) # based on the tests overhead/advantage is low
        indices_sorted = indices
        items_sorted = items
    else
        # sorts so the final index is correct
        indices_sorted, items_sorted = sort_indices_items(indices, items)
    end
    sizehint!(a, length(items)+length(a))
    for (index, item) in zip(indices_sorted, items_sorted)
        insert!(a, index, item)
    end
    return a
end
```
### Multi-Index Result:
For small sizes inserteach5! and 6 are definitely faster
```julia
using BenchmarkTools

a = [1, 2, 3, 4]
b = collect(6:-2:2)
c = [10, 9, 8]

using Test;
@test inserteach_multiindex_8!(copy(a), b, c) == inserteach_multiindex_7!(copy(a), b, c) == inserteach_multiindex_6!(copy(a), b, c) == inserteach_multiindex_5!(copy(a), b, c) == inserteach_multiindex_4!(copy(a), b, c) == inserteach_multiindex_3!(copy(a), b, c) == inserteach_multiindex_2!(copy(a), b, c) == inserteach!(copy(a), b, c)

println("small set")

@btime inserteach!(copy($a), $b, $c)
  # 341.860 ns (8 allocations: 576 bytes)
@btime inserteach_multiindex_2!(copy($a), $b, $c)
  # 315.443 ns (12 allocations: 688 bytes)
  # 241.960 ns (11 allocations: 688 bytes)  # using filter instead of in
@btime inserteach_multiindex_3!(copy($a), $b, $c)
  # 344.032 ns (12 allocations: 720 bytes)
  # 257.534 ns (11 allocations: 720 bytes) # using filter instead of in
@btime inserteach_multiindex_4!(copy($a), $b, $c)
  # 338.636 ns (8 allocations: 640 bytes)
@btime inserteach_multiindex_5!(copy($a), $b, $c)
  # 84.870 ns (2 allocations: 256 bytes)
@btime inserteach_multiindex_6!(copy($a), $b, $c)
  # 85.595 ns (2 allocations: 256 bytes)
@btime inserteach_multiindex_7!(copy($a), $b, $c)
  # 344.811 ns (8 allocations: 576 bytes)
@btime inserteach_multiindex_8!(copy($a), $b, $c)
  # 361.722 ns (8 allocations: 576 bytes)
```
But for larger sizes inserteach! is a lot faster. `inserteach_multiindex7!` has a similar speed to `inserteach!` until a batch `growat!` is written in C
```julia
println("big set, with skips, falling, non-sorted")

vsize = 1000
d = rand(Float32, vsize)
e = collect(vsize:-10:1)
f = rand(Float32, vsize÷10)

@btime @fastmath inserteach!(copy($d), $e, $f)
  # 5.533 μs (8 allocations: 14.22 KiB)
@btime inserteach_multiindex_2!(copy($d), $e, $f)
  # 56.799 μs (13 allocations: 21.06 KiB)
  # 58.800 μs (11 allocations: 33.55 KiB) # using filter instead of in
@btime inserteach_multiindex_3!(copy($d), $e, $f)
  # 55.900 μs (13 allocations: 28.55 KiB)
  # 59.501 μs (11 allocations: 48.41 KiB)  # using filter instead of in
@btime inserteach_multiindex_4!(copy($d), $e, $f)
  # 57.700 μs (8 allocations: 10.78 KiB)
@btime inserteach_multiindex_5!(copy($d), $e, $f)
  # 104.999 μs (2 allocations: 8.50 KiB)
@btime inserteach_multiindex_6!(copy($d), $e, $f)
  # 57.999 μs (2 allocations: 8.50 KiB)
@btime inserteach_multiindex_7!(copy($d), $e, $f)
  # 5.417 μs (8 allocations: 14.22 KiB)
@btime inserteach_multiindex_8!(copy($d), $e, $f)
  # 5.700 μs (8 allocations: 14.22 KiB)
```

For sorted array, it gets even faster:
```julia
println("big set, with skips, rising, sorted")

g = rand(vsize)
h = collect(1:10:vsize)
i = rand(vsize÷10)

@btime inserteach!(copy($g), $h, $i)
  # 7.100 μs (2 allocations: 23.63 KiB)
@btime inserteach_multiindex_2!(copy($g), $h, $i)
  # 57.299 μs (13 allocations: 29.25 KiB)
  # 58.699 μs (11 allocations: 33.55 KiB) # using filter instead of in
@btime inserteach_multiindex_3!(copy($g), $h, $i)
  # 56.901 μs (13 allocations: 44.11 KiB)
  # 59.700 μs (11 allocations: 48.41 KiB)  # using filter instead of in
@btime inserteach_multiindex_4!(copy($g), $h, $i)
  # 54.999 μs (2 allocations: 16.69 KiB)
@btime inserteach_multiindex_5!(copy($g), $h, $i)
  # 72.200 μs (2 allocations: 16.69 KiB)
@btime inserteach_multiindex_6!(copy($g), $h, $i)
  # 56.001 μs (2 allocations: 16.69 KiB)
@btime inserteach_multiindex_7!(copy($g), $h, $i)
  # 7.075 μs (2 allocations: 23.63 KiB)
@btime inserteach_multiindex_8!(copy($g), $h, $i)
  # 7.425 μs (2 allocations: 23.63 KiB)
```
</details>

So for single index `inserteach!` is much faster, but for multi-index, `inserteach_multiindex7!` has a similar speed to `inserteach!` until a batch `growat!` is written in C

c42f:
Here's an implementation `inserteach_via_resize!` using the approach I suggested in https://github.com/JuliaLang/julia/pull/34635#issuecomment-581768193.

For the benchmark at the bottom, I measure 20x faster than the current `inserteach!` in this PR. This comes from only touching memory once with a single scan through the elements.

Note that the semantics of this implementation are slightly different: the *indices refer to the insertion locations in the* **input array**, rather than the final locations of new elements in the output array. I believe this may make more sense because any computation you do to define location of the new elements is likely to be based on comparing to elements of the existing array and the existing indices. However the implementation technique should apply regardless of the semantic change.

<details>
<summary>Code</summary>

```julia
# Your existing code:
function sort_indices_items(indices::AbstractVector, items)
    sort_indices = sortperm(indices)
    indices_sorted = indices[sort_indices]
    items_sorted = items[sort_indices]
    return indices_sorted, items_sorted
end

function inserteach!(a::Vector, indices::Union{AbstractVector, AbstractRange{<:Integer}}, items::AbstractVector)
    itemslen = length(items)
    indiceslen = length(indices)
    if itemslen !== indiceslen
        throw(DimensionMismatch("Length of the indices and items should be the same"))
    elseif itemslen === 0
        return a
    end
    if issorted(indices) # based on the tests overhead/advantage is low
        indices_sorted = indices
        items_sorted = items
    else
        # sorts so the final index is correct
        indices_sorted, items_sorted = sort_indices_items(indices, items)
    end
    for (index, item) in zip(indices_sorted, items_sorted)
        Base._growat!(a, index, 1) # TODO write a batch growat in C for speed-up
        @inbounds a[index] = item
    end
    return a
end

# Alternative implementation:

# Insert items into `a` such that `items[j]` is placed between the elements
# `a[indices[j-1]:indices[j]]`.
# `indices` and `items` are assumed to be sorted (TODO: Ensuring reverse-sorted
# would be better.)
function _inserteach!(a, indices, items)
    read_idx = lastindex(a)
    resize!(a, length(a) + length(indices))
    insert_idx = lastindex(a)
    for (new_idx,newitem) in zip(reverse(indices), reverse(items))
        while read_idx >= new_idx
            a[insert_idx] = a[read_idx]
            read_idx -= 1
            insert_idx -= 1
        end
        a[insert_idx] = newitem
        insert_idx -= 1
    end
end

function inserteach_via_resize!(a::AbstractVector, indices::AbstractVector, items::AbstractVector)
    itemslen = length(items)
    if itemslen !== length(indices)
        throw(DimensionMismatch("Length of the indices and items should be the same"))
    elseif itemslen === 0
        return a
    end
    if issorted(indices)
        _inserteach!(a, indices, items)
    else
        indices, items = sort_indices_items(indices, items)
        _inserteach!(a, indices, items)
    end
    return a
end

using BenchmarkTools

@btime inserteach_via_resize!(A, indices, values) setup=begin
    A = rand(1000_000)
    indices = rand(1:1000_000, 500)
    values = rand(500)
end

@btime inserteach!(A, indices, values) setup=begin
    A = rand(1000_000)
    indices = rand(1:1000_000, 500)
    values = rand(500)
end
```
</details>

Benchmark results:

```
  1.378 ms (9 allocations: 7.65 MiB)
  26.342 ms (7 allocations: 7.64 MiB)
```

aminya:
> 
> 
> Here's an implementation `inserteach_via_resize!` using the approach I suggested in [#34635 (comment)](https://github.com/JuliaLang/julia/pull/34635#issuecomment-581768193).
> 
> For the benchmark at the bottom, I measure 20x faster than the current `inserteach!` in this PR. This comes from only touching memory once with a single scan through the elements.
> 
> Note that the semantics of this implementation are slightly different: the _indices refer to the insertion locations in the_ **input array**, rather than the final locations of new elements in the output array. I believe this may make more sense because any computation you do to define location of the new elements is likely to be based on comparing to elements of the existing array and the existing indices. However the implementation technique should apply regardless of the semantic change.

Thank you the algorithm! 👍 
I kind of agree with you regarding thinking about the old syntax instead of the new syntax, and if we do that it makes things a whole lot easier!

If we want to insert `0` between each two element of `a=[1,2,3,4,5,6,7]`, we can easily do `_inserteach!(a, 2:2:5, [0,0,0])` and get `[1,2,0,3,4,0,5,6,0,7]`.

But what about thinking of the new indices? Doesn't it have its applications as well?

c42f:
> But what about thinking of the new indices? Doesn't it have its applications as well?

I don't ever recall needing this but I can't remember specific cases where I've wanted bulk insert. I vaguely recall that the workflow is something like:
1. Iterate through the array and find some indices where things should be inserted.
2. Insert them in bulk

During step 1 you inherently only have access to the original indices.

StefanKarpinski:
> The API doesn't seem complex to me. It is way more intuitive than what `splice!` can do for single-index.

It's clear that you don't like the `splice!` API, but that's irrelevant here. The splice API is inherited from other languages where it has been working just fine in the same way for a long while. Dislike of the splice API does not justify adding another different API to do overlapping things.

Here are some of the issues with this API.

* It's a suspiciously vectorized API. We try to avoid vectorized APIs unless there's a strong reason for them. The behavior of `inserteach!` can be accomplished with repeated calls to `insert!`. Therefore the burden is on you to show that there's compelling reason that this API is needed when the pieces already exist for someone to implement this behavior themselves with much greater flexibility and clarity (see the next point on confusion about indexes).

* It's not entirely clear what the right meaning of indices is. The fact that you're having a back and forth with @c42f about whether the indices are for the original positions of the values or the new positions of the values is concerning because it means there's at least two equally reasonable interpretations and some people would assume one while others assume the other. For example, I agree with Chris that it makes much more sense that the indices be into the original array positions. I didn't even realize that what you were implementing was not that. That different people can have such different assumptions about what the indices mean is the problem: some people will assume one thing while others will assume the other thing; either way some people will get it wrong. If, on the other hand, someone just used repeated insertion, the behavior is clear: the index in each call obviously must be into the data as it is at the time the insertion occurs.

* It lacks motivation. On triage yesterday, we were having a hard time coming up with circumstances in which one would need this API. When adding a new API, it's helpful to know that someone actually needed it at some point. This may be incorrect, but it seems like your primary motivation may have been that you don't like `splice!` with an empty range to insert items and that you wanted to use `inserteach!(a::Vector, index::Integer, items::AbstractVector)` instead. The vectorized methods seem like they might be an afterthought added because "hey, why not?" (See above about preferring to avoid vectorized APIs unless absolutely necessary.) It would be good to know what the motivation for this is. What are you doing that you needed this? If you were just using it to insert at one position, then `splice!` can already do this. If you actually needed to use it for repeated insertion, was performance an issue? Lack of attention to performance in the original PR suggests not. Since potential performance advantage is the main thing that differentiates this API from repeated insertion, if performance wasn't an issue for you, then what was?

c42f:
@StefanKarpinski, I do believe an `inserteach!` API is worthwhile which is why I've been commenting on this PR. However I think the implementation here still needs quite a lot of work.

To respond to your concerns specifically:

> It's a suspiciously vectorized API

At a surface level yes, but it's completely warranted from an algorithmic complexity perspective:
* Inserting `M` things at random but sorted indices into a length `N` vector is `O(M + N)` in the algorithm I pasted above. This is clearly optimal. For unsorted indices, it's `O(M*log(M) + N)`.
* Naive repeated insertion is `O(M*N)` which is asymptotically worse; for large `N` really a lot worse.

> If, on the other hand, someone just used repeated insertion, the behavior is clear: the index in each call obviously must be into the data as it is at the time the insertion occurs.

It's clear what `insert!` does, but incremental `insert!` is not a good tool for batchwise insertion: the first call to `insert!` will invalidate any insertion locations you've previously computed.

> It lacks motivation

I've given two abstract reasons why `inserteach!` is desirable (asymptotically faster, and invalidation of indices). However I agree that there's a lack of *concrete* use cases where someone would be doing batch insertion in the first place.

I'm certain I've needed this before (in C++) but alas, it was too long ago to remember the exact circumstances. To give a rough gist: Imagine updating a large sorted vector `L` with a smaller number of insertions (from vector `J`, say). In that case you'd like to do binary search to find an insertion index for each element of `J` in `L`. Then do a single pass to splice all elements of `J` into `L`. The last step is exactly what you want `inserteach!` for. Without such an API you will be stuck with `O(M*N)` behavior which you definitely don't want.

@aminya it would help if you could provide a clear motivation for why you need to insert as a batch.

aminya:
> function _inserteach!(a, indices, items)
>     read_idx = lastindex(a)
>     resize!(a, length(a) + length(indices))
>     insert_idx = lastindex(a)
>     for (new_idx,newitem) in zip(reverse(indices), reverse(items))
>         while read_idx >= new_idx
>             a[insert_idx] = a[read_idx]
>             read_idx -= 1
>             insert_idx -= 1
>         end
>         a[insert_idx] = newitem
>         insert_idx -= 1
>     end
> end
> 

@c42f is this while-part the desired behavior? using this part of the algorithm:
```julia
inserteach!([6, 5, 4], 6:-1:5, [8, 9])  # [6, 5, 4, 9, 8]
```
which I thought we want to throw a `BoundsError` for.

aminya:
> @aminya it would help if you could provide a clear motivation for why you need to insert as a batch.

In the original post in the application section, I gave a couple of examples. Also here: https://github.com/JuliaLang/julia/pull/34635#issuecomment-581748390

Your example is also a very good one.

StefanKarpinski:
I buy the algorithmic argument, although in that case the implementation should be careful to do well in terms of asymptotic performance: grow the array only once and move items into the correct locations in a single pass. It seems like in all the examples the API is clearer of the indices refer to the original array.

aminya:
I updated the algorithm for multi-index API. 
Tests should be updated, but I want to decide about https://github.com/JuliaLang/julia/pull/34635#issuecomment-592909613 first.

c42f:
> grow the array only once and move items into the correct locations in a single pass.

Right, this is exactly what the proof of concept in `inserteach_via_resize!` does and consequently why it's 20x faster for N=1000_000, M=500 vs the code which was here.

c42f:
> which I thought we want to throw a `BoundsError` for because 6 is out of range

Indeed this is a bug and should be fixed! I think the last insertion index location should be no greater than `lastindex(a)+1`. By the way, I don't claim `inserteach_via_resize!` is complete and correct in all respects — it's merely a proof of concept to make sure the implementation suggested in https://github.com/JuliaLang/julia/pull/34635#issuecomment-581768193 was understood.

By the way, I think we need a stable sort for the indices to ensure that elements to be inserted are not reordered when they have the same index.

> Suppose we have an arpeggio repeated two times

Thanks for the extra usage examples. I see that you're very much coming at this from a API standpoint.

To further motivate inclusion in `Base` we should think more about what types of objects `inserteach!` applies to. I feel a non-mutating version `inserteach(a,inds,b)` could be seen as an API for sequences `a` (ie, iterables) rather than arrays as such. Then we should require consistency with indexing, for example that iteration should place `b[i]` immediately before `a[inds[i]]`. If we make the mutating version share these properties I think we have fairly general semantics even if we only implement it for `Vector`. I looked at clojure for some guidance because they have rich sequence APIs, but I only found https://clojuredocs.org/clojure.core/interleave and https://clojuredocs.org/clojure.core/interleave. I know Transducers.jl uses a lot of those ideas — maybe you have some thoughts on this @tkf?

tkf:
I don't remember seeing something like this in Clojure API when I looked into it. (Though I've never done any serious programming in Clojure. So maybe there are something relevant, e.g., outside the Clojure stdlib.)  I think it'd be straight forward to implement the lazy version for pre-sorted indices (either using iterators or transducers).  But exact mapping to the array indexing requires special casing offset arrays (which is not so bad, I guess).

It's hard to imagine the usecase of random indices, though. For example, I'd probably use `partition` for `StepRange`:

```julia
julia> using Base.Iterators: partition, flatten

julia> notes = repeat(["E2", "A3", "C3"], 2);

julia> roots = ["A2", "A2"];

julia> [x for (r, arp) in zip(roots, partition(notes, 3)) for x in flatten(((r,), arp))]
8-element Array{String,1}:
 "A2"
 "E2"
 "A3"
 "C3"
 "A2"
 "E2"
 "A3"
 "C3"
```

Though I can imagine that random indices can be useful for vectors; e.g., you want to bulk-insert elements into a sorted vectors.

c42f:
>  But exact mapping to the array indexing requires special casing offset arrays (which is not so bad, I guess).

I was imagining this could be done generically via `axes` or `eachindex` (which is how `firstindex` and `lastindex` work). Maybe I overlooked something though.

tkf:
If you use `eachindex`, or maybe `LinearIndices`, don't you instead need to "special case" non-array iterators?

```julia
_LinearIndices(x::AbstractArray) = LinearIndices(x)
_LinearIndices(::Any) = Iterators.countfrom()
```

I think this is a consequence of the fact that `eachindex`/`LinearIndices` already is special casing the offset arrays.

aminya:
> 
> 
> > which I thought we want to throw a `BoundsError` for because 6 is out of range
> 
> Indeed this is a bug and should be fixed! I think the last insertion index location should be no greater than `lastindex(a)+1`.

Bound checking here is more complex than this. We need to calculate `maximum(indices)`, with `length(a)+length(indices)`

c42f:
> We need to calculate `maximum(indices)`, with `length(a)+length(indices)`. Also `minimum(indices)>=1`

I don't think so. `indices` gets sorted, so the maximum and minimum are trivial. Also you're not comparing with `length(a)+length(indices)` if we take the convention that indices refers to insertion locations within the existing array `a`.

DilumAluthge:

@aminya If you're still interested in working on this PR, let me know, and I can re-open it.

Issue #34636, opened 2020-02-03
Overloading MethodErrors for certain method signatures


jkrumbiegel:
In response to this thread https://discourse.julialang.org/t/idea-julia-lite-or-juliette/34076/60 I think it could be valuable to add error messages for some errors that appear with high frequency, but which are not easily understood just from the `MethodError` they throw.

Example:
```julia
rand(5) + 3
ERROR: MethodError: no method matching +(::Array{Float64,1}, ::Int64)
Closest candidates are:
  +(::Any, ::Any, ::Any, ::Any...) at operators.jl:529
  +(::Complex{Bool}, ::Real) at complex.jl:297
  +(::Missing, ::Number) at missing.jl:115
  ...
Stacktrace:
 [1] top-level scope at REPL[123]:1
```

We know that this is something tons of people try when they start using the language coming over from Python, R, Matlab, etc. So it would be beneficial to teach them about dot broadcasting right here, in the error message. One could overload addition to remedy this:

```julia
function Base.:+(arr::AbstractArray{<:Number}, x::Number)
    error("""No method matching +(::$(typeof(arr)), ::$(typeof(x))).
    For element-wise addition, try dot-call broadcasting syntax: arr .+ x""")
```
This would result in:
```julia
julia> rand(5) + 3
ERROR: No method matching +(::Array{Float64,1}, ::Int64).
For element-wise addition, try dot-call broadcasting syntax: arr .+ x
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] +(::Array{Float64,1}, ::Int64) at ./REPL[1]:1
 [3] top-level scope at REPL[2]:1
```

I don't think actually overloading all these methods directly is the right way to go. But would it be possible to overload only the specific `MethodError` for this kind of problem? Are `MethodError`s typed in a way that one could dispatch on the type of error and give helpful messages in well-known cases like this, Array of Numbers plus / times / minus / etc. a scalar? I think this could be a highly effective mechanism in other cases as well.

timholy:
I think you've hit on the right approach. Here's how it can be fixed:

```julia
julia> a = [1,2,3];

julia> err = try
       a + 1
       catch e
       e
       end
MethodError(+, ([1, 2, 3], 1), 0x0000000000006899)

julia> @edit showerror(IOBuffer(), err)
```
As you'll discover, the `MethodError` captures the function and the arguments, so it should be reasonably straightforward.

jkrumbiegel:
[...]

PR #34642, opened 2020-02-03
Better error for elem-wise + and - of arrays and scalars

jkrumbiegel:
For addition and subtraction of arrays of numbers and a scalar, the error message suggests using broadcasting.

https://github.com/JuliaLang/julia/issues/34636

### DIFF
--- a/base/errorshow.jl
+++ b/base/errorshow.jl
@@ -265,6 +265,21 @@ function showerror(io::IO, ex::MethodError)
         end
         print(io, ")")
     end
+    # catch the two common cases of element-wise addition and subtraction
+    if f in (Base.:+, Base.:-) && length(arg_types_param) == 2
+        # we need one array of numbers and one number, in any order
+        if any(x -> x <: AbstractArray{<:Number}, arg_types_param) &&
+            any(x -> x <: Number, arg_types_param)
+
+            nouns = Dict(
+                Base.:+ => "addition",
+                Base.:- => "subtraction",
+            )
+            varnames = ("scalar", "array")
+            first, second = arg_types_param[1] <: Number ? varnames : reverse(varnames)
+            print(io, "\nFor element-wise $(nouns[f]), use broadcasting with dot syntax: $first .$f $second")
+        end
+    end
     if ft <: AbstractArray
         print(io, "\nUse square brackets [] for indexing an Array.")
     end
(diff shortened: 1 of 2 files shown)


jkrumbiegel:
looks like this in action
```julia
a = rand(5)

>>> 1 + a
ERROR: MethodError: no method matching +(::Int64, ::Array{Int64,1})
For element-wise addition, use broadcasting with dot syntax: scalar .+ array
Closest candidates are:
  +(::Any, ::Any, ::Any, ::Any...) at operators.jl:529
  +(::T, ::T) where T<:Union{Int128, Int16, Int32, Int64, Int8, UInt128, UInt16, UInt32, UInt64, UInt8} at int.jl:53
  +(::Integer, ::Integer) at int.jl:871
  ...
Stacktrace:
 [1] top-level scope at none:0
```

or
```julia
>>> a - 1//3
ERROR: MethodError: no method matching -(::Array{Int64,1}, ::Rational{Int64})
For element-wise subtraction, use broadcasting with dot syntax: array .- scalar
Closest candidates are:
  -(::Complex{Bool}, ::Real) at complex.jl:303
  -(::Missing, ::Number) at missing.jl:115
  -(::Complex, ::Real) at complex.jl:315
  ...
Stacktrace:
 [1] top-level scope at none:0
```

JeffBezanson (approved):

JeffBezanson:
Thanks, very nice first PR! This should be quite helpful for many new users.

StefanKarpinski:
It would be good to have a more general framework for pattern matching on errors and giving improved error messages, but this is a great addition for now.

KristofferC:
If possible, would be good with some kind of test that hits this code path and checks that the error message ends up as expected.

And in the future, we should make this extensible so maybe packages can add stuff as well, but for now, let's just get this in.

timholy:
Yes, this does need a test, but it's awesome.

I think you meant to link to #34636.

jkrumbiegel:
I could add tests, but how do you test errors? By triggering the error and comparing the error string to the expected one? And where would that test go, somewhere in https://github.com/JuliaLang/julia/blob/master/test/errorshow.jl I assume?

fredrikekre:
> I could add tests, but how do you test errors?

You can use `@test_throws` (from the `Test` stdlib).

timholy:
I wouldn't use `@test_throws`, consider modeling it after https://github.com/JuliaLang/julia/blob/3720edfc06ee8687b54ee7fbb908c9f03f4fc557/test/errorshow.jl#L345-L374

jkrumbiegel:
> In this specific case I wouldn't use @test_throws, consider modeling it after 

Hopefully I did this correctly, the tests ran fine on my local machine but I have zero experience with making changes in Base (well, one experience now)

timholy:
> but I have zero experience with making changes in Base (well, one experience now)

And So It Begins. :laughing: 

Thanks for a great contribution!

PR #32088, opened 2019-05-20
Prevent the GC task callback from segfaulting.

rbehrends:
It is possible when marking task objects in the GC for the associated
jl_ptls_t reference to be NULL. To determine whether a task is a root
task, we therefore also have to check if that reference is valid.

This is a fix for a bug in the old "GC Extensions" [changes added last year](https://github.com/JuliaLang/julia/pull/28368) that did not manifest until now, because it was triggered by a combination of unusual circumstances. It should be fairly straightforward, as it only adds a missing null pointer check.

PR #28368, opened 2018-07-31
GC Extensions

rbehrends:
This pull request implements the GC extension proposal described in [this Julia Enhancement proposal](https://github.com/JuliaLang/Juleps/blob/master/GcExtensions.md).

See the Julep above for a full description of the API and semantics.

PR #34644, opened 2020-02-03
Backport GC Extensions to Julia 1.0

fingolfin:
This PR backports PR #28368 and #32088 by @rbehrends to the `release-1.0` branch. The first PR contains extensions for the Julia GC to make used by the [OSCAR project](https://oscar.computeralgebra.de) (CC @wbhart @fieker) resp. the [GAP.jl](https://github.com/oscar-system/GAP.jl) package which integrates [GAP](https://www.gap-system.org) and Julia.

We realize that this might be controversial, and fully expect a discussion about this, but we wanted to at least try to get this in, as it would be really nice if people could use OSCAR with the Julia LTS releases.

I just discovered the rules for patch releases posted at https://discourse.julialang.org/t/proposed-release-process-and-schedule/15623, and those say that patch releases " ... should only contain bug fixes, low-risk performance improvements, and documentation updates". That does not fit with this PR (it adds an internal feature!), so perhaps that already is enough for you to reject it.... :-(. Anyway, I figured it can't hurt to at least try and ask...

To this end, note that this PR here is a super quick backport, which has ZERO testing right now -- of course that ought to change, but I before I waste time on that, I wanted to find out whether there is any chance at all for this being merged. If there is, we'll go over it carefully and test extensively. Also, I'd be happy to squash this down into a single commit, or do anything else you may require for backports.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -81,7 +81,7 @@ endif
 # headers are used for dependency tracking, while public headers will be part of the dist
 UV_HEADERS :=
 HEADERS := $(BUILDDIR)/julia_version.h $(wildcard $(SRCDIR)/support/*.h) $(addprefix $(SRCDIR)/,julia.h julia_assert.h julia_threads.h tls.h locks.h atomics.h julia_internal.h options.h timing.h)
-PUBLIC_HEADERS := $(BUILDDIR)/julia_version.h $(wildcard $(SRCDIR)/support/*.h) $(addprefix $(SRCDIR)/,julia.h julia_assert.h julia_threads.h tls.h locks.h atomics.h)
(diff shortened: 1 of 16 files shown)


JeffBezanson:
I believe 1.5 or 1.6 is likely to be the next LTS release, so it should not be too long. (Note though that this has not been formally decided yet.)

Strictly speaking, I have to agree with your assessment that this goes beyond what would normally be backported. It can be discussed though.

fingolfin:
@JeffBezanson ah I wasn't aware that there would be another 1.x LTS ! Of course if 1.5 or 1.6 were LTS, that would be perfectly fine for us as well (I think). Question is how likely that is, and when...

PR #34647, opened 2020-02-03
provide non-aliasing constructors for sparse arrays

### LINKED ISSUE #34630: Aliasing problem with `SparseMatrixCSC` and `SparseVector`
In `v1.2.0` and `v1.5.0` it is easy to create a corrupted sparse object unintentionally:
```
julia> B = sparse([1 1;0 0]);
julia> A = SparseMatrixCSC{Float64,Int}(B);
julia> A[2,1] = 1
julia> A
2×2 SparseMatrixCSC{Float64,Int64} with 3 stored entries:
  [1, 1]  =  1.0
  [2, 1]  =  1.0
  [1, 2]  =  1.0
julia> B
2×2 SparseMatrixCSC{Int64,Int64} with 3 stored entries:
  [1, 1]  =  1
  [2, 1]  =  1
  [1, 2]  =  #undef

julia> b = sparse([1;0]);
julia> a = SparseVector{Float64,Int}(b);
julia> a[2] = 1;
julia> a
2-element SparseVector{Float64,Int64} with 2 stored entries:
  [1]  =  1.0
  [2]  =  1.0
julia> b
2-element SparseVector{Int64,Int64} with 1 stored entry:
  [1]  =  1
  [2]  =  #undef
```
Reason seems to be that the constructors `SparseMatrixCSC{Tv,Ti}(S::SparseMatrixCSC)` and `SparseVector{Tv,Ti}(s:SparseVector)`
may produce objects, which share some but not all of their fields with the source objects.
That is for example the case, if `Ti == eltype(S.rowval)` and `Tv != eltype(S.nzval)`. That results in inconsistent objects when one of them is modified. For example:
```
julia> dump(b)
SparseVector{Int64,Int64}
  n: Int64 2
  nzind: Array{Int64}((2,)) [1, 2]
  nzval: Array{Int64}((1,)) [1]
```
where the sizes of the component vectors of `b` diverge.
IMO the constructors should be changed to not use `convert`, but make copies.

The issue surfaced in `v1.5` (but not in `v1.2`) when doing a calculation like follows, where the
corruption of `B` was rather unexpected:
```
julia> B = Symmetric(sparse([1 1;0 0]));
julia> B \ ones(2);
ERROR: MethodError: no method matching lu! ...

julia> B
2×2 Symmetric{Int64,SparseMatrixCSC{Int64,Int64}}:
   1     #undef
 #undef    0
```

KlausC:
This fixes #34630, which ~~does not touch~~ makes copy sparse constructors.

### DIFF
--- a/stdlib/SparseArrays/src/sparsevector.jl
+++ b/stdlib/SparseArrays/src/sparsevector.jl
@@ -427,7 +427,7 @@ SparseVector{Tv}(s::SparseVector{<:Any,Ti}) where {Tv,Ti} = SparseVector{Tv,Ti}(
 function SparseVector{Tv,Ti}(s::SparseVector) where {Tv,Ti}
     copyind = Vector{Ti}(nonzeroinds(s))
     copynz = Vector{Tv}(nonzeros(s))
-    SparseVector{Tv,Ti}(length(s::SparseVector), copyind, copynz)
+    SparseVector{Tv,Ti}(length(s), copyind, copynz)
 end
 
 # convert between different types of SparseVector
--- a/stdlib/SparseArrays/src/sparsevector.jl
+++ b/stdlib/SparseArrays/src/sparsevector.jl
@@ -432,6 +432,12 @@ SparseVector{Tv}(s::SparseVector{<:Any,Ti}) where {Tv,Ti} =
     SparseVector{Tv,Ti}(length(s::SparseVector), nonzeroinds(s), convert(Vector{Tv}, nonzeros(s)))
 
 convert(T::Type{<:SparseVector}, m::AbstractVector) = m isa T ? m : T(m)
+convert(::Type{SparseVector}, s::AbstractSparseVector) = convert(SparseVector{eltype(nonzeros(s)),eltype(nonzeroinds(s))}, s)
+convert(::Type{SparseVector{Tv}}, s::AbstractSparseVector) where Tv = convert(SparseVector{Tv,eltype(nonzeroinds(s))}, s)
+function convert(T::Type{SparseVector{Tv,Ti}}, s::AbstractSparseVector) where {Tv,Ti<:Integer}
+    s isa T && return s
+    T(length(s), Vector{Ti}(nonzeroinds(s)), Vector{Tv}(nonzeros(s)))
+end
 
 convert(T::Type{<:SparseVector}, m::AbstractSparseMatrixCSC) = T(m)
 convert(T::Type{<:AbstractSparseMatrixCSC}, v::SparseVector) = T(v)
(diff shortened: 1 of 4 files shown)


andreasnoack:
I'm not completely up-to-date on the conversations on the sparse matrix constructors but around 0.6/0.7, we discussed the memory behavior of constructors and convert methods when the input is of the same type and I believe the conclusion was the `convert` should avoid copying when possible whereas constructors should avoid aliasing.

KlausC:
> `convert` should avoid copying when possible whereas constructors should avoid aliasing

this implementation of `convert` avoids copying, if possible (that means the argument has already the target type). If that is not the case, a type conversion is necessary, which implies at least one of the 3 vector fields of `SparseMatrixCSC` or the 2 vector fields of `SparseVector` has to be copied, too.
The demand, that the usage of published methods does not surprisingly destroy the integrity of some of the read-only arguments of those methods leads to this solution.

I agree with you, that it would be better to make copies in the constructor `SparseMatrixCSC{Tv,Ti}(::AbstractSparseMatrx)`, but got backfire on this suggestion (see #34630).

KristofferC:
> I agree with you, that it would be better to make copies in the constructor SparseMatrixCSC{Tv,Ti}(::AbstractSparseMatrx), but got backfire on this suggestion (see #34630).

But the "constructor" for sparse matrices is `sparse`? You can't even do

```
julia> SparseMatrixCSC(rand(5))
ERROR: MethodError: no method matching SparseMatrixCSC(::Array{Float64,1})
```

KristofferC:
I do agree that it is pretty weird that the `SparseMatrixCSC` constructor is undocumented. Documenting it and making it the "first class" constructor would make sense to me.

Right now we have `SparseMatrixCSC(M::Matrix) = sparse(M)` but it should really be the opposite imo.

KristofferC:
Deciding between doing this for `convert` and `SparseMatrixCSC`, I would prefer to copy for `SparseMatrixCSC` for the reasons outlined in https://github.com/JuliaLang/julia/issues/34630#issuecomment-581869806 (consistency with `Array`).

KlausC:
> Deciding between doing this for `convert` and `SparseMatrixCSC`, I would prefer to copy for `SparseMatrixCSC`

Ok, that is fine. I will change this PR accordingly.

KristofferC (approved):

stdlib/SparseArrays/src/sparsevector.jl
```diff
@@ -421,18 +421,17 @@ SparseVector{Tv}(s::AbstractVector{Tv}) where {Tv} = SparseVector{Tv,Int}(s)
 
 SparseVector(s::AbstractVector{Tv}) where {Tv} = SparseVector{Tv,Int}(s)
 
+# copy-constructors
+SparseVector(s::SparseVector{Tv,Ti}) where {Tv,Ti} = SparseVector{Tv,Ti}(s)
+SparseVector{Tv}(s::SparseVector{<:Any,Ti}) where {Tv,Ti} = SparseVector{Tv,Ti}(s)
+function SparseVector{Tv,Ti}(s::SparseVector) where {Tv,Ti}
+    copyind = Vector{Ti}(nonzeroinds(s))
+    copynz = Vector{Tv}(nonzeros(s))
+    SparseVector{Tv,Ti}(length(s::SparseVector), copyind, copynz)
```
KristofferC: I know this was here before but could remove the `::SparseVector` here?

KlausC: you mean in the `length...` ?

KristofferC: Yes

KlausC: done

dkarrasch (approved):

Issue #34518, opened 2020-01-26
Rename isimmutable trait to ismutable


juliohm:
Based on recent discussions on slack, I noticed that the trait `isimmutable` is an exception to the rule compared to other `is*` traits in the language. For a concept `foo` we usually have `isfoo` as the trait, and `!isfoo` as the negation. However, with the concept `mutable`, we currently have `is(not)mutable` (a.k.a. `isimmutable`) implemented as the trait.

Would it be possible to deprecate `isimmutable` in favor of the more readable `ismutable` trait?

Keno:
I'm in favor. This naming is left over from when immutables where introduced with the `immutable` keyword, so it made a lot of sense to use the same terminology. Now that the keyword is `mutable`, switching over the predicate makes sense too. Of course, we can't actually change it until 2.0.

rapus95:
also, `!isimmutable` feels unnecessarily redundant as it holds a multiple negations

stevengj:
See also the discussion in #18168.

JeffBezanson:
Agree. In fact we can add `ismutable` now and silently deprecate `isimmutable`.

vtjnash:
I think the main reason for the existence of this name might be that they aren't necessarily exact opposites in the type domain (isimmutabletype). In particular, for Abstract and Union, we have that some of the instances might be mutable and others constant. (though also, we have now that `AbstractArray.body.body.mutable === false` which might be misleading in this example)

JeffreySarnoff:
`isimmutabletype` is not all there yet (https://github.com/JuliaLang/julia/issues/30210).
+1 to introducing `ismutable`.

ravibitsgoa:
It seems that this issue is closed now. Is it still open?
I am a beginner, looking for a "good first issue"! :)

KristofferC:
Yes seems to be fixed in #34518

PR #34652, opened 2020-02-04
#34518 - rename isimmutable to ismutable

ssikdar1:
PR for issue #34518. 
Created a `ismutable` in   `base/reflection.jl ` and added a `depwar` to `isimmutable`.

Result in REPL:
```
julia> isimmutable(1)
┌ Warning:  isimmutable is deprecated use ismutable instead 
│   caller = top-level scope at REPL[1]:1
└ @ Core REPL[1]:1
true

julia> ismutable(1)
false
```

Switched `isimmutable` to `ismutable` in other parts of the code. 

- [X] add `isimmutable`
- [X] replace `isimmutable` to `ismutable`
- [X] Add a few tests.
Didn't add any new tests but `test/reflection.jl` passed.
```
$ ./julia test/reflection.jl
Test Summary: |
issue #31687  | No tests
Test Summary:       | Pass  Total
methods with module |   11     11
```
- [X] Add a NEWS item.

### DIFF
--- a/base/deprecated.jl
+++ b/base/deprecated.jl
@@ -209,6 +209,6 @@ julia> isimmutable([1,2])
 false
 ```
 """
-isimmutable(@nospecialize(x)) = (@_pure_meta; !typeof(x).mutable)
+isimmutable(@nospecialize(x)) = !ismutable(x)
 
 # END 1.5 deprecations
--- a/base/deprecated.jl
+++ b/base/deprecated.jl
@@ -190,3 +190,24 @@ MPFR.BigFloat(x::Real, prec::Int, rounding::RoundingMode) = BigFloat(x, rounding
 end
 
 # END 1.3 deprecations
+
+# BEGIN 1.5 deprecations
+"""
+    isimmutable(v) -> Bool
+
+Return `true` iff value `v` is immutable.  See [Mutable Composite Types](@ref)
+for a discussion of immutability. Note that this function works on values, so if you give it
+a type, it will tell you that a value of `DataType` is mutable.
+
+# Examples
+```jldoctest
+julia> isimmutable(1)
+true
+
+julia> isimmutable([1,2])
+false
+```
+"""
+isimmutable(@nospecialize(x)) = (@_pure_meta; !typeof(x).mutable)
+
+# END 1.5 deprecations
--- a/base/deprecated.jl
+++ b/base/deprecated.jl
@@ -195,7 +195,7 @@ end
 """
     isimmutable(v) -> Bool
 !!! warning
-    Consider using `!ismutable(v)` instead, as `isimmutable(v)` will be replaced by `!ismutable(v)` in a future release.
+    Consider using `!ismutable(v)` instead, as `isimmutable(v)` will be replaced by `!ismutable(v)` in a future release. (Since Julia 1.5)
 Return `true` iff value `v` is immutable.  See [Mutable Composite Types](@ref)
 for a discussion of immutability. Note that this function works on values, so if you give it
 a type, it will tell you that a value of `DataType` is mutable.
--- a/base/deprecated.jl
+++ b/base/deprecated.jl
@@ -210,5 +210,6 @@ false
 ```
 """
 isimmutable(@nospecialize(x)) = !ismutable(x)
+export isimmutable
 
 # END 1.5 deprecations
--- a/base/deprecated.jl
+++ b/base/deprecated.jl
@@ -194,7 +194,8 @@ end
 # BEGIN 1.5 deprecations
 """
     isimmutable(v) -> Bool
-
+!!! warning
+    Consider using `!ismutable(v)` instead, as `isimmutable(v)` will be replaced by `!ismutable(v)` in a future release.
 Return `true` iff value `v` is immutable.  See [Mutable Composite Types](@ref)
 for a discussion of immutability. Note that this function works on values, so if you give it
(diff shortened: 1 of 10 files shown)


base/reflection.jl
```diff
@@ -417,7 +417,28 @@ julia> isimmutable([1,2])
 false
 ```
 """
-isimmutable(@nospecialize(x)) = (@_pure_meta; !typeof(x).mutable)
+function isimmutable(@nospecialize(x))
```
timholy: We generally don't start introducing "live" depwarns of exported functions in the middle of a stable release. Instead you should move this to the end of `base/deprecated.jl` but don't actually issue the depwarn. Most or all of the "live" depwarns there are for types or methods that were never exported in the first place and therefore not protected by the "no breaking changes" rule. Thus, those deprecations are not necessary but are there as a kindness. In contrast, this is part of Julia's public API.

ssikdar1: ok! moved `isimmutable` from `reflections.jl` to the bottom of `deprecated.jl` in 6b76c59

base/deprecated.jl
```diff
@@ -190,3 +190,24 @@ MPFR.BigFloat(x::Real, prec::Int, rounding::RoundingMode) = BigFloat(x, rounding
...
 
 # END 1.3 deprecations
+
+# BEGIN 1.5 deprecations
+"""
+    isimmutable(v) -> Bool
+
+Return `true` iff value `v` is immutable.  See [Mutable Composite Types](@ref)
+for a discussion of immutability. Note that this function works on values, so if you give it
+a type, it will tell you that a value of `DataType` is mutable.
```
timholy: Sorry I didn't mention this before, but perhaps it would be good to add
```
!!! warning
    Consider using `!ismutable(v)` instead, as `isimmutable(v)` will be replaced by `!ismutable(v)` in a future release.
```

Then the last step for a *truly* awesome contribution would be to add `ismutable` to https://github.com/JuliaLang/Compat.jl.

timholy (commented):
Looking good. One more ask and I would approve this for merging. (Oh, but please do add the NEWS and a test of `ismutable` directly.)

ssikdar1:
@timholy sure!

>> please do add the NEWS and a test of ismutable directly.
In the PR changed the tests in `test/reflection.jl` to use `ismutable` instead of `isimmuatable`
https://github.com/JuliaLang/julia/pull/34652/files#diff-2ea596942a7d251702840ffcaab1e763R107-R108

Is there somewhere else I should test `ismutable` ?  Should I revert this?

base/deprecated.jl
ssikdar1: Warning message added in a77d03a

ssikdar1: Compat PR https://github.com/JuliaLang/Compat.jl/pull/686

timholy (approved):

base/deprecated.jl
```diff
@@ -190,3 +190,25 @@ MPFR.BigFloat(x::Real, prec::Int, rounding::RoundingMode) = BigFloat(x, rounding
 end
 
 # END 1.3 deprecations
+
+# BEGIN 1.5 deprecations
+"""
+    isimmutable(v) -> Bool
+!!! warning
+    Consider using `!ismutable(v)` instead, as `isimmutable(v)` will be replaced by `!ismutable(v)` in a future release.
```
fredrikekre: Would be good if this mentioned that ismutable was introduced in Julia v1.5.

base/deprecated.jl
```diff
@@ -190,3 +190,25 @@ MPFR.BigFloat(x::Real, prec::Int, rounding::RoundingMode) = BigFloat(x, rounding
...
+# Examples
+```jldoctest
+julia> isimmutable(1)
+true
+
+julia> isimmutable([1,2])
+false
+```
+"""
+isimmutable(@nospecialize(x)) = (@_pure_meta; !typeof(x).mutable)
```
fredrikekre: Maybe this could now be implemented as
```suggestion
isimmutable(@nospecialize(x)) = !ismutable(x)
```
but perhaps not that important.

base/exports.jl
```diff
@@ -642,6 +642,7 @@ export
     isbits,
     isequal,
     isimmutable,
```
fredrikekre: You could move this export to where the function is defined in `deprecated.jl` such that we don't forget to remove it once the function definition is removed (that has happened before).

doc/src/base/base.md
```diff
@@ -159,6 +159,7 @@ Base.isdispatchtuple
 
 ```@docs
 Base.isimmutable
+Base.ismutable
```
fredrikekre: Maybe switch the order and put `ismutable` first?

base/deprecated.jl
ssikdar1: added in (Since Julia 1.5) in a252f5e

base/deprecated.jl
ssikdar1: changed in eed8b67

base/exports.jl
ssikdar1: export moved to `deprecated.jl` in 28f7f9f

doc/src/base/base.md
ssikdar1: order switched in aba863f

timholy:
Thanks @ssikdar1!

PR #34653, opened 2020-02-04
Few grammar changes

Mohit-Nain:
Suggesting few grammar changes in the string document


ViralBShah:
These are fine the way they are.

PR #34654, opened 2020-02-04
Implement `accumulate` and friends for Tuple

tkf:
close #26690

Demo:

```julia
julia> demo() = cumprod((1, 2, 3))
demo (generic function with 1 method)

julia> @code_typed optimize=true demo()
CodeInfo(
1 ─     return (1, 2, 6)
) => Tuple{Int64,Int64,Int64}
```

PR #34656, opened 2020-02-04
Implement `accumulate` and friends for arbitrary iterators

tkf:
This PR implements `Base.accumulate(op, itr)` simply by wrapping `Iterators.accumulate`.  I added `init` keyword argument to `Iterators.accumulate` to match the call signature.

<strike>
I'm posting this as a WIP PR as it is a patch on top #34654 which needs to go first.
</strike> (Edit: ready for review now)


base/iterators.jl
```diff
@@ -460,24 +461,31 @@ This is effectively a lazy version of [`Base.accumulate`](@ref).
...
-accumulate(f, itr) = Accumulate(f, itr)
+accumulate(f, itr; init = Base._InitialValue()) = Accumulate(f, itr, init)
 
 function iterate(itr::Accumulate)
     state = iterate(itr.itr)
     if state === nothing
         return nothing
     end
-    return (state[1], state)
+    val = Base.BottomRF(itr.f)(itr.init, state[1])
```
tkf: Since `Base.BottomRF(itr.f)` calls `reduce_first`, this patch changes the output:

Before:

```julia
julia> collect(Iterators.accumulate(+, (x for x in [true])))
1-element Array{Bool,1}:
 1

julia> collect(Iterators.accumulate(+, (x for x in [true, true, false])))
3-element Array{Integer,1}:
 true
    2
    2
```

After:

```julia
julia> collect(Iterators.accumulate(+, (x for x in [true])))
1-element Array{Int64,1}:
 1

julia> collect(Iterators.accumulate(+, (x for x in [true, true, false])))
3-element Array{Int64,1}:
 1
 2
 2
```

Ref:

https://github.com/JuliaLang/julia/blob/3720edfc06ee8687b54ee7fbb908c9f03f4fc557/base/reduce.jl#L73-L78

tkf:
Bump. Can this be reviewed and merged?

(The test failure in macOS looks like something irrelevant from `Distributed`)

test/iterators.jl
```diff
@@ -798,3 +798,9 @@ end
     @test Base.IteratorSize(Iterators.accumulate(max, rand(2,3))) === Base.IteratorSize(rand(2,3))
     @test Base.IteratorEltype(Iterators.accumulate(*, ())) isa Base.EltypeUnknown
 end
+
+@testset "Base.accumulate" begin
+    @test cumsum(x^2 for x in 1:3) == [1, 5, 14]
+    @test cumprod(x + 1 for x in 1:3) == [2, 6, 24]
+    @test accumulate(+, (x^2 for x in 1:3); init=100) == [101, 105, 114]
+end
```
rfourquet: Maybe add a test for `Iterators.accumulate` taking an `init` keyword?

rfourquet (approved):

rfourquet:
Nice addition! I was missing this functionality today. I will merge in a couple of days if no objections (please @tkf ping me if I forget).

test/iterators.jl
tkf: Some more tests are added in 78a6c440ca1c446298978e90b1090ba0edf5dc36

tkf:
ping :)

base/accumulate.jl
```diff
@@ -250,6 +262,10 @@ julia> accumulate(+, fill(1, 3, 3), dims=2)
 ```
 """
 function accumulate(op, A; dims::Union{Nothing,Integer}=nothing, kw...)
```
rfourquet: Sorry that I missed that on first review: wouldn't this function need a compat annotation, like `cumsum` and `cumprod` ?

tkf: Thanks for catching it. I added it in https://github.com/JuliaLang/julia/pull/34656/commits/7a4d7febaf219f80fbdbe4d6cb028c1ab2e2d40d.

rfourquet: Thanks, in the meantime I also thought about compat for the iterator version, and was sorry to have to ask you again another change, but you already made it :)

rfourquet:
As this changes very slightly the result of some accumulate operations (cf https://github.com/JuliaLang/julia/pull/34656/files#r374510231), I will apply the "minor change" label, but this migth be too minor to deserve it.

PR #34658, opened 2020-02-04
Overhaul of rounding for rational numbers

### LINKED ISSUE #34645: Missing RoundingModes for Rationals
`RoundUp`, `RoundDown`, and `RoundToZero` are not implemented for `Rational`s:

```julia
julia> round(2//3, RoundNearest)
1//1

julia> round(2//3, RoundToZero)
ERROR: MethodError: no method matching _round_rational(::Type{Rational}, ::Rational{Int64}, ::RoundingMode{:ToZero})
```

Edit: Since we now have `div(x, y, ::RoundingMode)`, the implementation of this seems trivial. I will make a PR later.

### LINKED ISSUE #34657: Rounding infinite Rationals
Rounding `1//0` and `-1//0` only works when a rounding mode is given:
```julia
julia> round(1//0, RoundNearest)
1//0

julia> round(1//0)
ERROR: DivideError: integer division error
```
I will make a PR for this later.

Edit: `trunc`, `floor` and `ceil` also do not support infinite `Rational`s.

sostock:
Fix #34645.
Fix #34657.

In addition, this simplifies the implementation of `round` for `Rational`s by using the recently added `div(x, y, ::RoundingMode)`.

### DIFF
--- a/base/missing.jl
+++ b/base/missing.jl
@@ -135,7 +135,7 @@ round(::Type{T}, ::Missing, ::RoundingMode=RoundNearest) where {T} =
     throw(MissingException("cannot convert a missing value to type $T: use Union{$T, Missing} instead"))
 round(::Type{T}, x::Any, r::RoundingMode=RoundNearest) where {T>:Missing} = round(nonmissingtype_checked(T), x, r)
 # to fix ambiguities
-round(::Type{T}, x::Rational, r::RoundingMode=RoundNearest) where {T>:Missing} = round(nonmissingtype_checked(T), x, r)
+round(::Type{T}, x::Rational{Tr}, r::RoundingMode=RoundNearest) where {T>:Missing,Tr} = round(nonmissingtype_checked(T), x, r)
 round(::Type{T}, x::Rational{Bool}, r::RoundingMode=RoundNearest) where {T>:Missing} = round(nonmissingtype_checked(T), x, r)
 
 # Handle ceil, floor, and trunc separately as they have no RoundingMode argument
--- a/base/rational.jl
+++ b/base/rational.jl
@@ -385,53 +385,19 @@ end
 trunc(::Type{T}, x::Rational) where {T} = convert(T,div(x.num,x.den))
 floor(::Type{T}, x::Rational) where {T} = convert(T,fld(x.num,x.den))
 ceil(::Type{T}, x::Rational) where {T} = convert(T,cld(x.num,x.den))
-round(::Type{T}, x::Rational, r::RoundingMode=RoundNearest) where {T} = _round_rational(T, x, r)
-round(x::Rational, r::RoundingMode) = round(Rational, x, r)
 
-function _round_rational(::Type{T}, x::Rational{Tr}, ::RoundingMode{:Nearest}) where {T,Tr}
-    if denominator(x) == zero(Tr) && T <: Integer
-        throw(DivideError())
-    elseif denominator(x) == zero(Tr)
+round(x::Rational, r::RoundingMode) = round(typeof(x), x, r)
+
+function round(::Type{T}, x::Rational{Tr}, r::RoundingMode=RoundNearest) where {T,Tr}
+    if iszero(denominator(x))
+        T <: Integer && throw(DivideError())
         return convert(T, copysign(one(Tr)//zero(Tr), numerator(x)))
     end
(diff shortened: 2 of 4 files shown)


sostock:
Should the addition of the missing `RoundingMode`s get a NEWS item?

Is #34657 considered a bug, i.e., should parts of this PR get backported?

base/missing.jl
```diff
@@ -135,7 +135,7 @@ round(::Type{T}, ::Missing, ::RoundingMode=RoundNearest) where {T} =
     throw(MissingException("cannot convert a missing value to type $T: use Union{$T, Missing} instead"))
 round(::Type{T}, x::Any, r::RoundingMode=RoundNearest) where {T>:Missing} = round(nonmissingtype_checked(T), x, r)
 # to fix ambiguities
-round(::Type{T}, x::Rational, r::RoundingMode=RoundNearest) where {T>:Missing} = round(nonmissingtype_checked(T), x, r)
+round(::Type{T}, x::Rational{Tr}, r::RoundingMode=RoundNearest) where {T>:Missing,Tr} = round(nonmissingtype_checked(T), x, r)
```
simonbyrne: Why add the parameter `Tr` here? It isn't used?

base/rational.jl
```diff
@@ -382,66 +382,27 @@ for (S, T) in ((Rational, Integer), (Integer, Rational), (Rational, Rational))
...
-end
+round(x::Rational, r::RoundingMode=RoundNearest) = round(typeof(x), x, r)
 
-function _round_rational(::Type{T}, x::Rational{Tr}, ::RoundingMode{:NearestTiesUp}) where {T,Tr}
-    if denominator(x) == zero(Tr) && T <: Integer
-        throw(DivideError())
-    elseif denominator(x) == zero(Tr)
+function round(::Type{T}, x::Rational{Tr}, r::RoundingMode=RoundNearest) where {T,Tr}
+    if iszero(denominator(x))
+        T <: Integer && throw(DivideError())
```
simonbyrne: Won't `div` throw a `DivideError` anyway?

base/missing.jl
sostock: Without the added parameter, the method ambiguity tests fail, see https://build.julialang.org/#/builders/22/builds/8886

I was surprised by this, too. Since there are no restrictions on the type parameter, I thought these methods would be equally specific. Does it make sense to open an issue for this?

base/rational.jl
sostock: `div` would throw a `DivideError`, but isn’t called in that branch. Instead, `convert` will throw an `InexactError`. Would it be appropriate to throw an `InexactError` instead, or would that be breaking?

One could also change the code to
```julia
function round(::Type{T}, x::Rational{Tr}, r::RoundingMode=RoundNearest) where {T,Tr}
    if  !(T <: Integer) && iszero(denominator(x))
        return convert(T, copysign(one(Tr)//zero(Tr), numerator(x)))
    end
    convert(T, div(numerator(x), denominator(x), r))
end
```
Then one would get the same behavior without explicitly throwing the `DivideError`.

simonbyrne: That does seem like a better choice, and is more explicit as to why the branch is there.

base/missing.jl
simonbyrne: Hmm, that is odd. No idea what is going on there.

base/rational.jl
sostock: Okay, I changed it to the implementation above. Note that the specialized method for `Rational{Bool}` still contains an explicit `DomainError` since that one does not call `div` at all:
https://github.com/JuliaLang/julia/blob/310b7fa60598b55e4e387ed600ddd6294353c3a1/base/rational.jl#L398-L403

sostock:
I just noticed that rounding with `digits`/`sigdigits` is type-unstable for `Rational`s. For infinite arguments, the result isn’t converted to float:
```julia
julia> round(1//2, digits=1)
0.5

julia> round(1//0, digits=1)
1//0
```
I added a commit to this PR that fixes this.

simonbyrne (approved):

sostock:
Can this be in 1.5?

StefanKarpinski:
Given that @simonbyrne has approved, I'm tempted to just merge, but let's discuss on triage.

Issue #34659, opened 2020-02-04
Wrong display for arrays of empty arrays


rfourquet:
The display of `[Int[]]` got recently broken, probably in #34387:
```julia
julia> [[1]]
1-element Array{Array{Int64,1},1}:
 [1]

julia> [Int[]]
1-element Array{Array{Int64,1},1}:
 0-element Array{Int64,1}
```

cc @fchorney

tkf:
I posted #34733 as a possible solution.

PR #34660, opened 2020-02-04
fix #34343, missing type prefix in display of some nested containers

### LINKED ISSUE #34343: missing type prefix in display of some nested containers

rfourquet:
This currently only fixes the arrays, and no tests, but wanted first feedback on the approach.
Arrays are now printed this way:
```julia
julia> [Int[]] # no change here, the context gives all the type info
 []

julia> show([Int[]]) # was `[[]]` before
[Int64[]]

julia> show([Int[1]]) # no change here, the 1 gives the type info
[[1]]
```
(Note: `[Int[]]` displays currently differently on master, cf. #34659).

Two different strategies:
1) Have an `:typeinfo_implicit` `IOContext` attribute tracking whether we "know" the current `:typeinfo` type because it was printed, or because it can be inferred from the content of the container. In the latter case, this can depend on the elements of the container being themselves non-empty containers, so if they turn out to be empty, we have then to print some type information for them (when `:typeinfo_implicit` is true). This is the 1st commit.
2) "if a container decides not to print a type prefix, it should then not set :typeinfo" (quoting [Jeff](https://github.com/JuliaLang/julia/pull/34354#issuecomment-573830284)). This is 2nd commit.

I'm not sure yet of the implications of both ideas, but think I prefer 2) for simplicity, which we can always amend if things don't go well (I'm just a little less sure that tests will even pass with 2)).

### DIFF
--- a/base/arrayshow.jl
+++ b/base/arrayshow.jl
@@ -427,8 +427,8 @@ _show_empty(io, X) = nothing # by default, we don't know this constructor
 function show(io::IO, X::AbstractArray)
     ndims(X) == 0 && return show_zero_dim(io, X)
     ndims(X) == 1 && return show_vector(io, X)
-    prefix = typeinfo_prefix(io, X)
-    io = IOContext(io, :typeinfo => eltype(X))
+    prefix, implicit = typeinfo_prefix(io, X)
+    io = IOContext(io, :typeinfo => eltype(X), :typeinfo_implicit => implicit)
     isempty(X) ?
         _show_empty(io, X) :
         _show_nonempty(io, X, prefix)
@@ -453,9 +453,10 @@ end
 # NOTE: v is not constrained to be a vector, as this function can work with iterables
 # in general (it's used e.g. by show(::IO, ::Set))
 function show_vector(io::IO, v, opn='[', cls=']')
-    print(io, typeinfo_prefix(io, v))
+    prefix, implicit = typeinfo_prefix(io, v)
+    print(io, prefix)
     # directly or indirectly, the context now knows about eltype(v)
-    io = IOContext(io, :typeinfo => eltype(v))
+    io = IOContext(io, :typeinfo => eltype(v), :typeinfo_implicit => implicit)
     limited = get(io, :limit, false)
 
     if limited && length(v) > 20
@@ -497,6 +498,8 @@ end
 # X not constrained, can be any iterable (cf. show_vector)
 function typeinfo_prefix(io::IO, X)
     typeinfo = get(io, :typeinfo, Any)::Type
+    implicit = get(io, :typeinfo_implicit, false)::Bool
+
     if !(X isa typeinfo)
         typeinfo = Any
     end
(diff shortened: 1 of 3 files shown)


JeffBezanson:
I agree the simplicity of option (2) is appealing; can you provide some examples of output that is different under the two options?

rfourquet:
> can you provide some examples of output that is different under the two options?

I don't find any, they are probably equivalent: with option 2), not setting `:typeinfo` at a certain level doesn't matter for levels below it, because all sub-objects will receive the same implicit information (which come from "leaf" objects). So it's really a matter of knowing the prefix and "was it implicit". Both options have two slightly different but equivalent (AFAICT) ways of conveying this same bit (implicitness) of information.

I added some tests, if they pass I'm happy with the current state of the PR.

timholy:
Changing this to a 2-arg output has led to some amusing errors in packages that specialized `typeinfo_prefix`. From the Revise ["Display" tests](https://github.com/timholy/Revise.jl/blob/00cf83f0b628f321fdeb8eded0123bde95b2691f/test/runtests.jl#L334-L355):

julia-1.4:
```julia
julia> str = String(take!(io))
"OrderedCollections.OrderedDict(Main.ReviseTestPrivate.##254=>ExprsSigs(<1 expressions>, <0 signatures>),Main.ReviseTestPrivate.##254.ReviseTest=>ExprsSigs(<2 expressions>, <2 signatures>),Main.ReviseTestPrivate.##254.ReviseTest.Internal=>ExprsSigs(<6 expressions>, <5 signatures>))"
```

julia-1.5:
```julia
julia> print(IOContext(stdout, :compact=>true), mexs)
O(Main.ReviseTestPrivate.##297=>ExprsSigs(<1 expressions>, <0 signatures>),Main.ReviseTestPrivate.##297.ReviseTest=>ExprsSigs(<2 expressions>, <2 signatures>),Main.ReviseTestPrivate.##297.ReviseTest.Internal=>ExprsSigs(<6 expressions>, <5 signatures>))
```

I can fix this in Revise, but would it be worth collecting the outputs to a single return value and then inspecting its type before deciding to destructure it?

rfourquet:
> Changing this to a 2-arg output has led to some amusing errors in packages that specialized typeinfo_prefix

Yeah sorry about that!

> would it be worth collecting the outputs to a single return value and then inspecting its type before deciding to destructure it?

That sounds like a good idea; do you mean it as a transition phase to not break code, or to simplify implementations not needing the `implicit` bit?  I can't appreciate whether it's worth it, but I would guess very few packages are affected by this change. How hard was it to track back the "bug" to this PR?

timholy:
Not hard, in fact two of us (@tkluck also) did it independently.

I'm not sure it's worth it either, but thought I'd mention it just in case. Since `typeinfo_prefix` doesn't even have a docstring, it's definitely not something you need to bend over backwards to support. But if it's the kind of thing that turns out to be customized by a number of packages (or, we think more packages *should* be customizing it), then it would be worth documenting and thinking a bit about a transition plan.

PR #34664, opened 2020-02-05
Add a test for `pkgdir`: submodule located in a subdirectory

DilumAluthge:
Currently, there is a test for `pkgdir(Foo.SubFoo)`, where the source file that defines `SubFoo` is located in the same directory as the source file that defines `Foo`.

This pull request adds an additional test for `pkgdir(Foo.SubFoo2)`, where the source file that defines `SubFoo2` is NOT located in the same directory as the source file that defines `Foo`.

The basic idea is that we want to make sure `pkgdir` works correctly even if you give it a submodule that has been defined in a subdirectory.

cc: @ianshmean

### DIFF
--- a/test/loading.jl
+++ b/test/loading.jl
@@ -284,7 +284,8 @@ module NotPkgModule; end
 
     @testset "pkgdir" begin
         @test pkgdir(Foo) == normpath(abspath(@__DIR__, "project/deps/Foo1"))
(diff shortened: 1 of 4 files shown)


ianshmean:
Nice, yeah. Good catch

StefanKarpinski (approved):

PR #34665, opened 2020-02-05
Added support for summing StepRangeLen{T,Float64,Float64}

chriselrod:
You can create these via, eg, `1f1:1f-1:1f2`.

Before:
```julia
julia> r32 = Float32(-10.9):Float32(3.4):Float32(200.98)
-10.9f0:3.4f0:199.90001f0

julia> sum(r32)
ERROR: MethodError: no method matching _tp_prod(::Float64, ::Int64, ::Int64)
Closest candidates are:
  _tp_prod(::Base.TwicePrecision, ::Any, ::Any...) at twiceprecision.jl:706
Stacktrace:
 [1] sum(::StepRangeLen{Float32,Float64,Float64}) at ./twiceprecision.jl:546
 [2] top-level scope at REPL[6]:1

julia> sum(r16)
ERROR: MethodError: no method matching _tp_prod(::Float64, ::Int64, ::Int64)
Closest candidates are:
  _tp_prod(::Base.TwicePrecision, ::Any, ::Any...) at twiceprecision.jl:706
Stacktrace:
 [1] sum(::StepRangeLen{Float16,Float64,Float64}) at ./twiceprecision.jl:546
 [2] top-level scope at REPL[9]:1
```

After:
```julia
julia> sum(r16)
Float16(5.95e3)

julia> sum(r32)
5953.5f0

julia> sum( -10.9:3.4:200.9 )
5953.5
```

These are less precise. I added a few tests, but I used `≈` instead of `==` for one of the `Float16` tests due to its imprecision.


base/twiceprecision.jl
```diff
@@ -535,7 +535,22 @@ function __convertSRL(::Type{StepRangeLen{T,R,S}}, r::AbstractRange{U}) where {T
...
-function sum(r::StepRangeLen)
+function sum(r::StepRangeLen{T,Float64,Float64}) where {T}
+    l = length(r)
+    # Compute the contribution of step over all indices.
+    # Indexes on opposite side of r.offset contribute with opposite sign,
+    #    r.step * (sum(1:np) - sum(1:nn))
+    np, nn = l - r.offset, r.offset - 1  # positive, negative
+    # To prevent overflow in sum(1:n), multiply its factors by the step
+    sp, sn = sumpair(np), sumpair(nn)
+    tp = r.step * sp[1] * sp[2]
```
timholy: This will indeed be safe for `r.step isa AbstractFloat` but perhaps not if it's an integer. What about something like this?

```julia
W = widen(Int)
Δn = Int(W(sp[1]) * W(sp[2]) - W(sn[1]) * W(sn[2]))
s = r.ref * Δn
```

The length of `StepRangeLen` is hard-coded as `Int`, so this seems fairly safe.

base/twiceprecision.jl
```diff
@@ -535,7 +535,22 @@ function __convertSRL(::Type{StepRangeLen{T,R,S}}, r::AbstractRange{U}) where {T
...
+    tn = r.step * sn[1] * sn[2]
+    s = tp - tn
+    # Add in contributions of ref
+    ref = r.ref * l
+    convert(T, s + ref)
+end
+function sum(r::StepRangeLen{T,Base.TwicePrecision{T},Base.TwicePrecision{T}}) where {T}
```
timholy: Probably `sum(r::StepRangeLen{T,<:TwicePrecision,<:TwicePrecision}) where {T}` would be a better declaration here.

timholy (commented):
Good catch, and thanks for the fix! As is, it will fix some problems but introduce others. As defined it will only catch a subset of `StepRangeLen` types:
```julia
using Unitful
s = 5u"m"
rng = StepRangeLen(0*s, s, 10)
```
is a type that won't fit either method. (You can mimic Unitful using `Furlongs` which is already defined in the ranges tests.) Naturally the old method would fail there too, but while we're fixing things...

chriselrod:
> Good catch, and thanks for the fix! As is, it will fix some problems but introduce others. As defined it will only catch a subset of `StepRangeLen` types:
> 
> ```julia
> using Unitful
> s = 5u"m"
> rng = StepRangeLen(0*s, s, 10)
> ```
> 
> is a type that won't fit either method. (You can mimic Unitful using `Furlongs` which is already defined in the ranges tests.) Naturally the old method would fail there too, but while we're fixing things...

```julia
julia> include(joinpath(Sys.BINDIR, "..", "..", "test", "testhelpers", "Furlongs.jl"))
Main.Furlongs

julia> using .Main.Furlongs

julia> @which sum(Furlong(1f0):Furlong(5f-1):Furlong(1f1))
sum(a::AbstractArray; dims) in Base at reducedim.jl:652

julia> typeof(Furlong(1f0):Furlong(5f-1):Furlong(1f1))
StepRange{Furlong{1,Float32},Furlong{1,Float32}}
```

Should this PR be expanded to cover `StepRange` as well as `StepRangeLen`?
`mapreduce` probably isn't how we want to be summing that.

test/ranges.jl
```diff
@@ -1326,6 +1326,15 @@ using .Main.Furlongs
           Vector(Furlong(1):Furlong(0.5):Furlong(10)) == Furlong.(1:0.5:10)
 end
 
+@testset "sum arbitrary types" begin
+    @test sum(Furlong(1):Furlong(0.5):Furlong(10)) == Furlong{1,Float64}(104.5)
```
timholy: Since StepRangeLen is an exported type you can create the range to be of whatever type you want, e.g., `StepRangeLen(Furlong(1), Furlong(0.5), 19)` if you need to make sure you're testing your new implementation.

base/twiceprecision.jl
```diff
@@ -536,6 +536,21 @@ function __convertSRL(::Type{StepRangeLen{T,R,S}}, r::AbstractRange{U}) where {T
...
 function sum(r::StepRangeLen)
+    l = length(r)
+    # Compute the contribution of step over all indices.
+    # Indexes on opposite side of r.offset contribute with opposite sign,
+    #    r.step * (sum(1:np) - sum(1:nn))
+    np, nn = l - r.offset, r.offset - 1  # positive, negative
+    # To prevent overflow in sum(1:n), multiply its factors by the step
+    sp, sn = sumpair(np), sumpair(nn)
+    W = widen(Int)
+    Δn = Int(W(sp[1]) * W(sp[2]) - W(sn[1]) * W(sn[2]))
```
timholy: You know what, I think I screwed up here. We should probably eliminate the `Int` from this, I was mentally thinking there'd have to be a cancelation here for any usable range but that's not guaranteed.

timholy (commented):
Interesting! I frankly don't understand why that `sum(r::AbstractRange{<:Real})` method is restricted to `<:Real`. It would be great to fix that too if you have time, but no worries if you can't (this is already a terrific improvement). Even if you don't switch it to a `O(1)` algorithm, I don't think it hurts to leave the Furlong test in; after all, the *result* is what we want it to be, and that just provides insurance that when the method does migrate it will continue to return the correct result.

This looks like this has accumulated an unintended deletion of a couple of checksum files. I don't actually know what those are about, but I don't think this can be merged until that's fixed.

base/twiceprecision.jl
chriselrod: Do you mean we don't need the widening, and we want this to simply be:
```julia
sp, sn = sumpair(np), sumpair(nn)
Δn = sp[1] * sp[2] - sn[1] * sn[2]
```
Or that we shouldn't convert back to an Int, leaving `typeof(Δn) == wident(Int)`?

test/ranges.jl
chriselrod: Think I should add that in addition to (following your "I don't think it hurts to leave the Furlong test in; after all, the *result* is what we want it to be" comment) or in place of the current Furlong-sum test?

chriselrod:
> This looks like this has accumulated an unintended deletion of a couple of checksum files. I don't actually know what those are about, but I don't think this can be merged until that's fixed.

I realized that happened locally (although I'm not sure how) and accidentally `git commit -am` out of habit.
Is it possible to revert that directly on github? If not, I'll either have to check out this PR at work or wait until I get back home.

base/twiceprecision.jl
timholy: The second (don't convert back to `Int`). Since we have a `convert` in the final step, we might as well wait until then. (I was thinking, "oh, any big ranges that might overflow will also have to have cancelation on either side of the offset," but that's just wrong: multiplying by the step gives an additional mechanism for avoiding overflow.)

timholy:
I don't think I can do it without doing the same thing you'd do and then submitting a PR to your branch.

base/twiceprecision.jl
chriselrod: Okay, I removed the `Int`. For what it's worth, because `::Float64 * ::Int128` yields `Float64`, both versions will be equivalent in the fairly common case where the step is a float (or a wrapper, like the Furlong).

test/ranges.jl
timholy: I like how you did it.

base/twiceprecision.jl
```diff
@@ -536,6 +536,21 @@ function __convertSRL(::Type{StepRangeLen{T,R,S}}, r::AbstractRange{U}) where {T
...
+    # To prevent overflow in sum(1:n), multiply its factors by the step
+    sp, sn = sumpair(np), sumpair(nn)
+    W = widen(Int)
+    Δn = W(sp[1]) * W(sp[2]) - W(sn[1]) * W(sn[2])
+    s = r.step * Δn
+    # Add in contributions of ref
+    ref = r.ref * l
+    convert(eltype(r), s + ref)
+end
+function sum(r::StepRangeLen{<:Any,<:TwicePrecision,<:TwicePrecision}) where {T}
```
timholy: you have a dangling `where {T}` and this is causing the test failures

timholy (commented):
Fix the test failure and this looks good to me!

base/twiceprecision.jl
chriselrod: Thanks again for the reviews!
I removed the `where {T}`, but I did run into a lot of other test failures locally. Most where BLAS related (couldn't load MKL while running the testsuite?), but also `exp2` test failures.

timholy: Seems to work on CI, and that's what matters for everyone but you :smile:. I always use the default OpenBlas, but I'm aware of why you're particularly interested in MKL.

timholy:
Thanks so much @chriselrod!

PR #30494, opened 2018-12-22
Faster and simpler sprand for SparseMatrixCSC

abraunst:
**IMPORTANT: If you need to reproduce matrices prior to this change, your best bet is to use the old version of the sprand function that you can find [here](https://github.com/JuliaLang/julia/blob/a4b32610cf484b99f69e12d0ac1e192ef5c688a9^/stdlib/SparseArrays/src/sparsematrix.jl).**

The code for sprand for `SparseMatrixCSC` picks nonempty columns first, and then fill them (with the conditional probability of being nonempty). This seems to be done for faster construction for very sparse matrices (so that empty columns are to be expected). I think however that this does not give a real advantage, because one still needs to fill the colptr array, even for empty columns. I implemented a simpler version, and seems to be twice as fast or more in most cases. Am I overlooking something? Some benchmarks:
```
for n=(10,100,1000), p=(1.0,1e-1,1e-2,1e-3,1e-4)
    @show(n,p)
    @show @benchmark sprand($n,$n,$p)
end
```

MASTER:

-------------------------------------------------------------
 |    n\p        | 1.0         |   0.1     | 0.01          |  0.001       | 0.0001
|-----------|------------|-----------|---------------|---------------|-------------
| n=10     | 2.458 μs | 2.135 μs | 830.200 ns | 704.667 ns | 688.295 ns
| n=100   | 82.119 μs | 56.133 μs | 13.024 μs | 2.480 μs | 895.000 ns
| n=1000 | 9.556 ms | 3.858 ms | 555.889 μs | 152.445 μs | 17.662 μs
----------------------------------------------------------------------------------

PR:

-----------------------------------------------------------------------------
|   n\p     | 1.0         |   0.1     | 0.01          |  0.001       | 0.0001
|-----------|------------|-----------|---------------|---------------|-------------
| n=10     |751.574 ns|835.169 ns|419.209 ns|355.906 ns|348.586 ns
| n=100   |45.407 μs|32.657 μs|4.680 μs|970.100 ns|586.613 ns
| n=1000 |4.732 ms|3.362 ms|340.277 μs|45.028 μs|7.101 μs
---------------------------------------------------------------------------------

[Raw data](https://github.com/JuliaLang/julia/files/2705628/bench3.txt)

EDIT: Some simple innest-loop optimization improves speed further:

-----------------------------------------------------------------------------
|   n\p     | 1.0         |   0.1     | 0.01          |  0.001       | 0.0001
|-----------|------------|-----------|---------------|---------------|-------------
[...]

PR #34670, opened 2020-02-05
Added doc to RNG about changing between versions

Seelengrab:
This PR is based on a few people asking about why their tests failed following a minor version change (as far as I can tell, [this one](https://github.com/JuliaLang/julia/pull/30494)). Notably on discourse following 0.6 -> 0.7 ([here](https://discourse.julialang.org/t/unit-tests-for-packages-that-use-random-number-generation-that-are-robust-to-version-changes/14477)) and because of the linked PR the [tests](https://github.com/EcoJulia/SpatialEcology.jl/blob/master/test/ComMatrix_test.jl#L31-L33) in SpatialEcology.jl also failed.

Relying on a specific stream of generated numbers using any of the interfaces of Random is imo bad form and should be avoided if possible, so a warning would be useful.

### DIFF
--- a/stdlib/Random/docs/src/index.md
+++ b/stdlib/Random/docs/src/index.md
@@ -26,6 +26,9 @@ unbounded integers, the interval must be specified (e.g. `rand(big.(1:6))`).
 Additionally, normal and exponential distributions are implemented for some `AbstractFloat` and
 `Complex` types, see [`randn`](@ref) and [`randexp`](@ref) for details.
 
+!!! warn
+    Because the precise way in which random numbers are generated is considered an implementation detail, bug fixes and speed improvements may change the stream of numbers that are generated after a version change. Relying on a specific seed or generated stream of numbers during unit testing is thus discouraged - consider testing properties of the methods in question instead.
+
 ## Random numbers module
 ```@docs
 Random.Random


mkborregaard:
I was bitten by this recently, my tests suddenly started erroring from julia-1.1 to julia-1.2

Seelengrab:
This PR is based on a few people asking about why their tests failed following a minor version change (as far as I can tell, [this one](https://github.com/JuliaLang/julia/pull/30494)). Notably on discourse following 0.6 -> 0.7 ([here](https://discourse.julialang.org/t/unit-tests-for-packages-that-use-random-number-generation-that-are-robust-to-version-changes/14477)) and because of the linked PR the [tests](https://github.com/EcoJulia/SpatialEcology.jl/blob/master/test/ComMatrix_test.jl#L31-L33) in SpatialEcology.jl also failed. 

Relying on a specific stream of generated numbers using any of the interfaces of Random is imo bad form and should be avoided if possible, so a warning would be useful.

ViralBShah:
@Seelengrab I just moved your comment to the top of the issue since it greatly helps a reader of the issue.

stdlib/Random/docs/src/index.md
```diff
@@ -26,6 +26,9 @@ unbounded integers, the interval must be specified (e.g. `rand(big.(1:6))`).
 Additionally, normal and exponential distributions are implemented for some `AbstractFloat` and
 `Complex` types, see [`randn`](@ref) and [`randexp`](@ref) for details.
 
+!!! warn
+    Because the precise way in which random numbers are generated is considered an implementation detail, bug fixes and speed improvements may change the stream of numbers that are generated after a version change. Relying on a specific seed or generated stream of numbers during unit testing is thus discouraged - consider testing properties of the methods in question instead.
```
rfourquet: Maybe specify "minor version change" ? (someone else should confirm)

Seelengrab: I guess that depends on if RNG is only allowed to change in a minor but not a bug version?

rfourquet: Ah good point; improvements changes will happen in minor versions (and be labelled as "minor change"), but a bug fix changing the streams (although rare) might happen in a patch version.

aminya (approved):

ViralBShah:
Merge?

kimikage:
Isn't the "warn" a typo of "warning"?

Seelengrab:
It was consistent with other warnings in the docs, e.g. [here](https://github.com/JuliaLang/julia/blob/96786e22ccabfdafd073122abb1fb69cea921e17/base/stream.jl#L258-L269). IIRC, the keyword for `warning` in the old documentation system was `warn`, but that seems to have changed (https://docs.julialang.org/en/v1/stdlib/Markdown/#Admonitions-1). A quick grep through the source shows some other places where the old keyword is, I'll make a new PR to fix this.

PR #23528, opened 2017-08-31
add few missing methods for set-like operations

rfourquet:
For `intersect`, `union`, `setdiff` and `symdiff`, and their mutating variants for `Set` and `IntSet`, there is now support for arbitrary many arguments consistently. What breaks is for one-argument versions: for example we had `s=Set(); union(s) !== s` but `symdiff(s) === s`. I chose here to always return a copy (this should be the only regression in terms of performance). Also, I realize that having `setdiff` accepting more or less than two arguments may be controversial; I find it natural, but can remove this bit if requested.
This should not get merged until nanosoldier runs the [corresponding benchmarks](https://github.com/JuliaCI/BaseBenchmarks.jl/pull/109). 

PS: nice to have those improvements with a net decrease of LOC! (in particular when there are 28 more lines for tests)

EDIT: also, this takes care of improving the complexity of some operations mentioned [here](https://github.com/JuliaLang/julia/pull/23139#issuecomment-320444057).
EDIT: this also adds `filter`/`filter!` for `IntSet` returning an `IntSet` (like for `Set` and `Dict`).

PR #34675, opened 2020-02-05
make setdiff not call `setdiff!` ?

oxinabox:
Calling `setdiff!` on `copy(x)` seems like its not desirable.

### DIFF
--- a/base/abstractset.jl
+++ b/base/abstractset.jl
@@ -212,7 +212,7 @@ julia> symdiff(unique([1,2,1]), unique([2, 1, 2]))
 ```
 """
 symdiff(s, sets...) = symdiff!(emptymutable(s, promote_eltype(s, sets...)), s, sets...)
-symdiff(s) = copy(s)
+symdiff(s) = union(s)
 
 """
     symdiff!(s::Union{AbstractSet,AbstractVector}, itrs...)
--- a/base/abstractset.jl
+++ b/base/abstractset.jl
@@ -212,7 +212,7 @@ julia> symdiff(unique([1,2,1]), unique([2, 1, 2]))
 ```
 """
 symdiff(s, sets...) = symdiff!(emptymutable(s, promote_eltype(s, sets...)), s, sets...)
-symdiff(s) = symdiff!(copy(s))
+symdiff(s) = copy(s)
 
 """
     symdiff!(s::Union{AbstractSet,AbstractVector}, itrs...)


rfourquet:
But the call to `symdiff!` maintains the semantics:
```julia
julia> symdiff([1, 2, 1])
[2]
```
So `symdiff(s)` is not equivalent to a `copy`.

base/abstractset.jl
```diff
@@ -212,7 +212,7 @@ julia> symdiff(unique([1,2,1]), unique([2, 1, 2]))
 ```
 """
 symdiff(s, sets...) = symdiff!(emptymutable(s, promote_eltype(s, sets...)), s, sets...)
-symdiff(s) = symdiff!(copy(s))
+symdiff(s) = copy(s)
```
oxinabox:
```suggestion
symdiff(s) = union(s)
```

rfourquet:
`symdiff(s)` is currently also not equal to `union(s)`...

oxinabox:
> `symdiff(s)` is currently also not equal to `union(s)`...

Ok, I have no idea what `setdiff` with a single argument is supposed to do.

For that matter, i have no idea what `union` is suposed to do.
Both of these are binary operations.

KristofferC:
Chesterton's fence says hello :P

oxinabox:
> Chesterton's fence says hello :P

😂 
Chesterton and [Cunningham](https://meta.wikimedia.org/wiki/Cunningham%27s_Law) would not have gotten on.

rfourquet:
> I have no idea what setdiff with a single argument is supposed to do.

The docstring of `setdiff` makes this quite clearly defined, but I realize we have conflated `setdiff` and `symdiff`!

For `symdiff`, there is also a logical reason for the current behavior, but it's far from explicit in the docstring (hint: it's a consequence of the "the multiplicity of elements matters"). Basically, as a set is the symmetric difference of all its element (taken individually as sets), the symmetric difference of multiple sets is equal to the symmetric difference of all the individual elements of all the sets. The end result is that you can put all the elements of all the sets in a multiset, and reject those which have even multiplicity, and set multiplicity 1 to the remaining ones. The behavior for non-set collections (in Julia) is just the generalisation of this same algorithms. There was not much discussions about the semantics we wanted for `symdiff` on non-set collections (which can have repeated elements), but at least it has an internal consistency and is not totally arbitrary.

StefanKarpinski:
Oh man, that is most definitely not what I would expect `symdiff` on a set with multiple copies of the same value to do.

StefanKarpinski:
By that same logic, shouldn't `setdiff([1, 2, 2, 3], [2, 3])` produce `[1, 2]` instead of `[1]`?

rfourquet:
My impression is that the current choices are made for efficiency. If I'm not mistaken, the general rule for these set operations is that they return a collection without duplicates. To have a second rule "semantically uniquifies all the arguments before processing" (i.e. doesn't consider inputs as "multisets") could make sense; this is what happens currently for `setdiff` but not to `symdiff`. And it doesn't matter for `intersect` and `union`.

For `setdiff!(s::Set, itr)`, the definition is straightforward, and so is the implementation: just remove from `s` elements from `itr`. (All other `setdiff[!]` methods probably rely on this primary one).

But for `symdiff!(s::Set, itr)`, the straightforward implementation is to add or remove `x` from `s` depending on whether `x` is already in `s` for all `x in itr`. To semantically consider only distinct elements from `itr` would require first to convert `itr` to a `Set`. 

I vaguely remember searching for precedent in other languages on this, without success. So I went for the faster implementation, which also has its own logic explained in my previous post, based on the associativity of the operation. This was also what `symdiff!(::BitSet, itr)` was doing before my change in #23528.

oxinabox:
> I realize we have conflated setdiff and symdiff

Oops, clearly I really messed up this PR. Since I thought I was changing `setdiff`.

Its been educational but anything coming Friday m it needs to go elsewhere

Issue #23680, opened 2017-09-12
Regression/bug in indmin


thofma:
On 0.6:
```julia
julia> versioninfo()
Julia Version 0.6.0
Commit 9036443 (2017-06-19 13:05 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Core(TM) i5-3320M CPU @ 2.60GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Sandybridge)
  LAPACK: libopenblas64_
  LIBM: libopenlibm
  LLVM: libLLVM-3.9.1 (ORCJIT, ivybridge)

julia> a = [4]
1-element Array{Int64,1}:
 4

julia> indmin((i for i in a))
1
```
Latest master:
```julia
julia> versioninfo()
Julia Version 0.7.0-DEV.1760
Commit a5e9844 (2017-09-12 16:50 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: Intel(R) Core(TM) i5-3320M CPU @ 2.60GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Sandybridge)
  LAPACK: libopenblas64_
  LIBM: libopenlibm
  LLVM: libLLVM-3.9.1 (ORCJIT, ivybridge)
Environment:

julia> a = [4]
1-element Array{Int64,1}:
 4

julia> indmin((i for i in a))
ERROR: MethodError: no method matching keys(::Base.Generator{Array{Int64,1},getfield(Main, Symbol("##1#2"))})
Closest candidates are:
  keys(::Cmd) at process.jl:826
  keys(::Tuple) at tuple.jl:41
  keys(::Tuple, ::Tuple...) at tuple.jl:47
  ...
Stacktrace:
 [1] pairs(::Base.Generator{Array{Int64,1},getfield(Main, Symbol("##1#2"))}) at ./associative.jl:137
 [2] findmin at ./array.jl:2141 [inlined]
 [3] indmin(::Base.Generator{Array{Int64,1},getfield(Main, Symbol("##1#2"))}) at ./array.jl:2198
```

JeffBezanson:
This was intentional. `(i for i in a)` doesn't have indices; e.g. `getindex` doesn't work on it. Could you explain the use case a bit more?

[...]

Issue #34368, opened 2020-01-14
generators should lower to `Iterators.map`, not `Base.Generator`


bramtayl:
This will allow dispatching to types that aren't always `Base.Generator`

tkf:
Thanks for opening this. I think the same holds for `Iterators.filter`, `Iterators.flatten`, and `Iterators.product`.

JeffBezanson:
Why?

bramtayl:
It would be nice to be able to define `function Iterators.map(a_function, thing::SpecialTypeOfArray)` to return something other than a `Generator`. Of course, we can currently overload the `Generator` constructor to make something that's not a `Generator`, but it seems to violate the principle of least surprise. @tkf what was your reasoning?

JeffBezanson:
Why do you need it to return something other than a `Generator`? A generator just contains the function and iterator, so no information is lost.

tkf:
@bramtayl That's my thought too.

More specifically, for example, `Iterators.map(very_special_function, ::SpecialTypeOfArray)` can return a lazy `AbstractArray` if you have a compiler-agnostic way of computing output type of the `very_special_function`.  Likewise, `Iterators.product` can easily return a lazy array, too.  `Iterators.flatten` can also return an array if the input is an array of static arrays. OTOH, `Generator` cannot be an array subtype as it has to be able wrap other things (e.g., `Iterators.filter`)

Another usecase is to overload `Iterators.map(f, ::MyTable) :: MyLazyTable`.  You can then overload `show` of `MyLazyTable` to preview computed result like Query is doing.

(But maybe the first point should be resolved using traits, as discussed in the ArrayLike PR #34196)

rapus95:
As I understand it right now, that'll easen the implementation of all types of operators which intend to propagate something. For example a caching iterator which automatically caches the outermost iteration call. Even if it's used as the innermost starter. This would work by dispatching on
```
Iterators.map(f, c::Caching{T}) = cache(f(x) for x in c.iter)
```
where `cache(iter)` creates the caching layer.

Now, `(f(x) for x in cache(1:10))` would be the same as `cache(f(x) for x in 1:10)`

Also, it'd allow custom return types which in turn could implement arbitrary traits. Currently, as I understand it, you can only implement that by either not using the generator syntax or by specializing the trait for `::Generator{myf, myiter}`

andyferris:
I support multiple dispatch for construction of generic generators based on lowering to `Iterators.map`.

[...]

PR #34678, opened 2020-02-05
keys(::Generator) pass through

### LINKED ISSUE #34674: keys(::Generator) for find* and arg*
Tl;Dr:
Add
```julia
Base.keys(gen::Base.Generator) = Base.keys(gen.iter)
```
to Base.

-----

Partially related to my comment in https://github.com/JuliaLang/julia/issues/34368#issuecomment-582365454 regarding the support of traits on generators.
Currently there is no way to use find* on generators. Reason is that there is no keys function:
```julia
julia> findmax(abs(x-1) for x in [1,5,2,9])
ERROR: MethodError: no method matching keys(::Base.Generator{Array{Int64,1},var"#97#98"})
```
But since generators are said to be the lazy variant of `map` and `map` preserves the indices it would be good to define that missing `keys` method.

That would allow for some very generic `findmax(iter)` and `argmax(iter)` as arbitrary lazy maps are allowed while maintaining good performance and low memory footprint.

Edit: adding the missing find* and arg* with custom transform functions would be appreciated aswell to allow for the convenient do-block notation.

rapus95:
Let keys pass through to allow lightweigth/lazy find* and arg* functions
Fixes https://github.com/JuliaLang/julia/issues/34674


aminya (approved):

rapus95:
What would be the intended tests here?
is that already enough?
```
@test keys(x^2 for x in 1:10) == keys(1:10)
```

KlausC:
As a test maybe more illustrative:
```
@test keys(x^2 for x in -1:0.5:1) == 1:5
```
I propose to put that into `test/functional.jl`

StefanKarpinski:
Triage finds it a little weird but approves since it makes find work and doesn't seem harmful. Would be good to add @KlausC's test and a NEWS entry, however. A use case would also be good.

mbauman (dismissed):

rapus95:
@mbauman sure, got interrupted by something else after rebasing such an old branch for the first time 😇
For the NEWS entry, into which section should this go?

test/functional.jl
```diff
@@ -134,6 +134,9 @@ let gen = ((x,y) for x in 1:10, y in 1:10 if x % 2 == 0 && y % 2 == 0),
     @test collect(gen) == collect(gen2)
 end
 
+# keys of a generator for find* and arg* (see #34678)
+@test keys(x^2 for x in -1:0.5:1) == 1:5
+
```
mbauman, line 139:
```suggestion
@test findall(!iszero, x^2 for x in -1:0.5:1) == [1, 2, 4, 5]
@test argmin(x^2 for x in -1:0.5:1) == 3

```

Let's also test your main motivation

thofma:
Kind of fixes https://github.com/JuliaLang/julia/issues/23680

Instead of `indmin` I just have to use `findmin`.

rapus95:
@thofma once this lands, you can just use `indmin` again since `keys` for `Generator` won't be missing anymore.

rapus95:
@mbauman well since I guess 1.5 is already pinned now (branched away from master), but to not to add it to the wrong news file, this'll go into 1.6, won't it? (so I can add a news entry straight to the 1.6 news file)

I'd put it into the "standard library changes" section.

tkf:
I think `keys`, `values` and `getindex` should satisfy the invariance

```julia
isequal(collect(values(A)), [A[k] for k in keys(A)])
```

So, I'm against this patch _unless_ `values` and `getindex` are defined for `Generator`.  See #36175.

rapus95:
bump for 1.6 👀

rapus95:
@fredrikekre please remove the "needs test" label. (Since there already is one). And can you add it to a 1.6 milestone? (so that it won't go unnoticed upon freezing, especially since triage already accepted the standalone and https://github.com/JuliaLang/julia/pull/37648 holds the `getindex` for generators)

oscardssmith:
Rerunning co, then good to merge in my view.

oscardssmith (approved):

oscardssmith:
@StefanKarpinski me closing and re-opening doesn't seem to have trigered CI. Can you make it re-run?

rapus95:
> @StefanKarpinski me closing and re-opening doesn't seem to have trigered CI. Can you make it re-run?

@oscardssmith To me it seems like buildbot was triggered and currently is only missing tester_macos64 and tester_win32

oscardssmith:
oh yeah. Sorry for the noise. It might be worth rebasing this onto master so we can get a fully green CI.

musm:
This looks like it still has a triage label, was this approved? Typically, the triage  label would be removed before we allow a PR to be merged.

oscardssmith:
Triage approved back in March. I think it was re-added to consider back-porting once merged, but I'm not sure.

musm:
To be sure, it would be best to hold off then until triage-reapproves and once the requested review is approved.

JeffBezanson:
Triage is ok with this again. I came around since it basically seems consistent with `axes`, which we already define here. We don't want to backport it but it can be merged on master.

rapus95:
Ok that's fine, then we should be able to discuss the long-term meaning of Generator and thus, if we also want to include the other issues/PRs (#37648, #34368) which would lead to a better rounded feature addition in 1.7

simeonschaub:
Thank you for the contribution! Sorry this took so long. In the future, also feel free to bump PRs every once in a while, because otherwise things sometimes get forgotten.

PR #34688, opened 2020-02-06
fix docstrings for & and |

tpapp:
The header of the docstring for `&` and `|` suggest a call syntax that is invalid (either needs `()`s, or be infix). This is a trivial fix for that.

Issue noted here:

https://discourse.julialang.org/t/ampersand-operator-not-consistent-with-help/34266

### DIFF
--- a/base/int.jl
+++ b/base/int.jl
@@ -254,7 +254,7 @@ false
 (~)(x::BitInteger)             = not_int(x)
 
 """
-    &(x, y)
+    x & y
 
 Bitwise and. Implements [three-valued logic](https://en.wikipedia.org/wiki/Three-valued_logic),
 returning [`missing`](@ref) if one operand is `missing` and the other is `true`.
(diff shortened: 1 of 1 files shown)


MasonProtter:
Is it worth mentioning in the doc-string that one can do `(&)(x, y)` and `(|)(x, y)`?

aminya (approved):

tpapp:
@MasonProtter: that is generic information about using infix operators as prefix syntax, so I don't think we should mention it in specific docstrings.

Incidentally, I noted that many operators have this problem in docstrings (eg `+`, `-`), I wonder if they should be fixed in general.

fredrikekre:
> Incidentally, I noted that many operators have this problem in docstrings (eg +, -), I wonder if they should be fixed in general.

But they work:
```
julia> +(1, 2)
3
```
The difference seems to be that `&` doesn't parse to a `:call` without the parenthesis. `|` actually work though:
```
julia> |(true, false)
true
```
(but the proposed change looks better anyway IMO).

GunnarFarneback:
Parsing of `&`: #34693

MasonProtter:
> That is generic information about using infix operators as prefix syntax, so I don't think we should mention it in specific docstrings.

As Fredrik pointed out, the fact that you can't do `&(a, b)` but can do `(&)(a, b)` is a bit of an oddity among the infix operators, that might make it worth mentioning in its docstring.

StefanKarpinski (approved):

kshyatt:
Test on freebsd failed in `SuiteSparse`.

PR #34689, opened 2020-02-06
Add NamedTuple pair constructor

omus:
Adds an alternative way to create a `NamedTuple`:
```julia
julia> NamedTuple(:a => 1, :b => 2.0)
(a = 1, b = 2.0)
```
It may be more sensible to call this `namedtuple` to mirror the `tuple` interface.

### DIFF
--- a/test/namedtuple.jl
+++ b/test/namedtuple.jl
@@ -276,3 +276,7 @@ let nt0 = NamedTuple(), nt1 = (a=33,), nt2 = (a=0, b=:v)
     @test Base.setindex(nt1, "value", :a) == (a="value",)
     @test Base.setindex(nt1, "value", :a) isa NamedTuple{(:a,),<:Tuple{AbstractString}}
 end
+
+# Pair constructor
+@test NamedTuple(:a => 1, :b => 2.0) === (a=1, b=2.0)
+@test_throws MethodError NamedTuple("a" => 1, "b" => 2.0)
--- a/test/namedtuple.jl
+++ b/test/namedtuple.jl
@@ -278,5 +278,5 @@ let nt0 = NamedTuple(), nt1 = (a=33,), nt2 = (a=0, b=:v)
 end
 
 # Pair constructor
-@test NamedTuple(:a => 1, :b => 2.0) === (a=1, b=2.0)
-@test_throws MethodError NamedTuple("a" => 1, "b" => 2.0)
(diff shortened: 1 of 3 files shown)


test/namedtuple.jl
```diff
@@ -276,3 +276,7 @@ let nt0 = NamedTuple(), nt1 = (a=33,), nt2 = (a=0, b=:v)
     @test Base.setindex(nt1, "value", :a) == (a="value",)
     @test Base.setindex(nt1, "value", :a) isa NamedTuple{(:a,),<:Tuple{AbstractString}}
 end
+
+# Pair constructor
+@test NamedTuple(:a => 1, :b => 2.0) === (a=1, b=2.0)
```
nickrobinson251: do we also want to test `(; :a => 1, :b => 2.0)` and/or `(:a => 1, :b => 2.0)` ?

test/namedtuple.jl
```diff
@@ -276,3 +276,7 @@ let nt0 = NamedTuple(), nt1 = (a=33,), nt2 = (a=0, b=:v)
...
+@test_throws MethodError NamedTuple("a" => 1, "b" => 2.0)
```
nickrobinson251: should add a test for `(a => 1, b => 2.0)`?

JeffBezanson:
The convention is for collection constructors to take a single argument, e.g. `Tuple([1,2])`, so I agree this would need to be a separate function. I guess just for symmetry it makes sense to have
```
tuple(x...) = (x...,)
namedtuple(x...) = (; x...)
```
but what is the use case for this?

omus:
> what is the use case for this?

Mainly I was wanting this as a more explicit form of writing `(; x...)`. As an interesting unexpected benefit there does seem to be a performance improvement:
```julia
julia> @btime (; :a => 1, :b => 2.0)
  1.425 ns (0 allocations: 0 bytes)
(a = 1, b = 2.0)

julia> @btime NamedTuple(:a => 1, :b => 2.0)
  0.032 ns (0 allocations: 0 bytes)
(a = 1, b = 2.0)
```

KristofferC:
`0.032 ns`, That isn't measuring anything sensible.

test/namedtuple.jl
omus: I'll add in the former but the later is just a regular ole' `Tuple`

test/namedtuple.jl
omus: I'm not sure what value this test adds

PR #33235, opened 2019-09-12
Allows the results of `mapfoldr` to be inferred on tuples.

tpapp:
This solution was [suggested](https://discourse.julialang.org/t/mapfoldr-and-inference-on-tuple/28657/2) by @tkf.

PR #34691, opened 2020-02-07
Eagerly return a tuple from zip of tuples

tkf:
Can we have `zip(tuples...)` return a tuple?  I think this is compatible with some methods in `Iterators` that eagerly compute results. See #33235 for similar discussion on `Iterators.reverse(x::Tuple) :: Tuple`.

### DIFF
--- a/base/iterators.jl
+++ b/base/iterators.jl
@@ -390,7 +390,12 @@ _zip_iterator_eltype(::Type{Tuple{}}) = HasEltype()
 
 reverse(z::Zip) = Zip(map(reverse, z.is))
 
-zip(tuples::NTuple{N,Any}...) where {N} = ntuple(i -> map(t -> t[i], tuples), N)
+# Not using `zip(tuples::NTuple{N,Any}...)` to avoid having unbound
+# type parameter.
+function zip(x::NTuple{N,Any}, xs::NTuple{N,Any}...) where {N}
+    tuples = (x, xs...)
+    return ntuple(i -> map(t -> t[i], tuples), N)
+end
 
 # filter
 
--- a/base/iterators.jl
+++ b/base/iterators.jl
@@ -390,6 +390,8 @@ _zip_iterator_eltype(::Type{Tuple{}}) = HasEltype()
 
 reverse(z::Zip) = Zip(map(reverse, z.is))
 
+zip(tuples::NTuple{N,Any}...) where {N} = ntuple(i -> map(t -> t[i], tuples), N)
+
 # filter
 
 struct Filter{F,I}
--- a/test/tuple.jl
+++ b/test/tuple.jl
@@ -480,3 +480,10 @@ end
 
 # tuple_type_tail on non-normalized vararg tuple
 @test Base.tuple_type_tail(Tuple{Vararg{T, 3}} where T<:Real) == Tuple{Vararg{T, 2}} where T<:Real
+
+@testset "zip" begin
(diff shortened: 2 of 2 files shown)


JeffBezanson:
This seems like a bit of a strange road to go down. My main concern is that callers are not communicating their intent; the call is to `Iterators.zip` but it in fact depends on the implementation being eager. It might be surprising for iterators to switch between eager and lazy for different types by default. If you need that level of control it might be better to have an explicit `eagerzip` or `Iterators.Eager.zip`?

tkf:
> It might be surprising for iterators to switch between eager and lazy for different types by default.

But isn't it already the case?

https://github.com/JuliaLang/julia/blob/a211abcdfacc05cb93c15774a59ce8961c16dac4/base/iterators.jl#L96-L101

> If you need that level of control it might be better to have an explicit `eagerzip` or `Iterators.Eager.zip`?

Ideally, I think it'd be nice to have:

* `Lazy.zip(itrs...)`: return `Iterators.Zip` always
* `copy(Lazy.zip(itrs...))` or `materialize(Lazy.zip(itrs...))`: eager zip.
* `DWIM.zip(itrs...)`: default to `Lazy.zip` but eagerly materialize when it can be done in O(1) space/time (e.g., ranges and tuples).

(`Lazy.` and `DWIM.` are just temporary names)

If you want `Iterators.$f` to be always lazy, I think it'd better to add some kind of `DWIM.` versions and "soft-deprecate" all the DWIM methods of `Iterators.reverse`.  Ideally, it'd be great if `DWIM.$f` is exposed as `Base.$f` in Julia 2.0.

(Alternatively, maybe expose `Iterators.Zip` as "`Lazy.zip`" and treat `Iterators.zip` as "`DWIM.zip`"? But I think DWIM `Base.$f` would be much better if possible.)

JeffBezanson:
>     * eagerly materialize when it can be done in O(1) space/time (e.g., ranges and tuples).

Tuples use O(n) space...

tkf:
Ah, right, of course. I should have written "when it consumes practically negligible space and time resources."

JeffBezanson:
Large tuples do exist, and we'd like to support them better if possible. So it's really not obvious to me that `zip` should be eager for tuples.

tkf:
> Large tuples do exist, and we'd like to support them better if possible.

Based on this reasoning, shouldn't #33235 be reverted?  It's not released yet.

---

I think there are three issues here:

1. Being consistent across all APIs in `Iterators`.  I think `Iterators.$f` would be hard to if some functions return a tuple when the input is a tuple but some others don't.  It would be better to have consistent behavior so that users can have a simple mental model.

   I think this point is **most important for the moment** since (I suppose) #33235 (special-casing `Iterators.reverse` on `Tuple`) can be reverted before releasing 1.4.

2. Having a good implementation for eager iterator transformations of small containers like `Tuple` and `NamedTuple`.

3. Having a "DWIM wrapper" of the eager and lazy iterator transformations.  This is required for Base APIs like `foldr` (which was the motivation for defining `Iterators.reverse(::Tuple)` #33235).  This is least important and I suppose it can be done as an internal functions in `Base`, at least at the beginning.

My initial aim was to get 3 (which requires 2) but I'm totally fine with it if `Base` is not having it. It's very easy to do this outside `Base`.  But I do hope we get the point 1 right.

JeffBezanson:
Yes, the point of #33235 was actually `mapfoldr`, which seems like it would be better to accomplish with a more direct specialization instead of changing `reverse`. So I agree with your summary.

PR #34692, opened 2020-02-07
try replace backslashes with forwardslashes in ccall symbols

### LINKED ISSUE #34680: `ccall()` with an absolute library path while outputting an `.o` file causes linker issues on Windows
Kristoffer and I were tracking down precompilation failures for PackageCompilerX, when we noticed [a strange error](https://github.com/JuliaComputing/PackageCompilerX.jl/pull/91#issuecomment-582503103) coming from `ld` while it was trying to link together the system image:

```
Warning: .drectve `-export:ccalllib_C:\Users\Matthis\.julia\packages\FreeType\2dE5w\deps\usr\bin\libfreetype-6.dll,data ' unrecognized
C:/Users/Matthis/.julia/artifacts/572b61b5075459e3ed62317e674398166ca98dd4/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/8.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: Cannot export FreeType: symbol not found
C:/Users/Matthis/.julia/artifacts/572b61b5075459e3ed62317e674398166ca98dd4/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/8.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: Cannot export Users: symbol not found
C:/Users/Matthis/.julia/artifacts/572b61b5075459e3ed62317e674398166ca98dd4/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/8.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: Cannot export ccalllib_C:: symbol not found
C:/Users/Matthis/.julia/artifacts/572b61b5075459e3ed62317e674398166ca98dd4/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/8.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: Cannot export packages: symbol not found
```

This is being emitted as [a global variable from `ccall()`](https://github.com/JuliaLang/julia/blob/f2d68ad6eb6b735f5beed779b9570f0d9fbbab28/src/ccall.cpp#L70-L83), but appears to have some quoting issues on Windows, as the linker doesn't like the backslashes in the symbol name.  We're not sure if the linker is truly to blame here, as we usually don't pass absolute library paths so we may have never actually tried to link a system image with this usage pattern before.

You can reproduce this with the following mini reproducer:

```julia
using Pkg.BinaryPlatforms

open("libfoo.c", "w") do io
    write(io, """
    #include <stdio.h>

    void foo() {
        printf("Hello, Foo!\\n");
    }
    """)
end

@info("Compiling C source...")
libfoo_fname = string("libfoo.", platform_dlext())
run(`gcc -o $libfoo_fname libfoo.c -shared`)

open("Foo.jl", "w") do io
    write(io, """
    module Foo
    const libfoo = joinpath(@__DIR__, "$libfoo_fname")
    ccall((:foo, libfoo), Cvoid, ())
    end
    """)
end

@info("Compiling Julia module...")
run(`$(Base.julia_cmd()) --startup-file=no --output-o foo.o Foo.jl`)

@info("Dumping `ccalllib` julia module symbols...")
run(pipeline(`nm foo.o`, `grep ccalllib`))
```

Looking at the other `ccalllib` symbols, they are all "soname" style:
```
0000000001615770 b _ccalllib_/Users/sabae/libfoo.dylib
000000000162c3d0 b _ccalllib_libcholmod
00000000016160d8 b _ccalllib_libdSFMT
0000000001615b48 b _ccalllib_libgit2
0000000001618f40 b _ccalllib_libgmp
0000000001611b50 b _ccalllib_libmpfr
000000000162ae30 b _ccalllib_libopenblas64_
0000000001623800 b _ccalllib_libopenlibm
0000000001610b60 b _ccalllib_libpcre2-8
000000000162c430 b _ccalllib_libsuitesparse_wrapper
000000000162c510 b _ccalllib_libsuitesparseconfig
```

KristofferC:
I couldn't manage to get julia to build locally on Windows so I'll just use the buildbots as a way to build for me.

Trying to fix https://github.com/JuliaLang/julia/issues/34680

### DIFF
--- a/src/ccall.cpp
+++ b/src/ccall.cpp
@@ -68,11 +68,8 @@ static bool runtime_sym_gvs(const char *f_lib, const char *f_name, MT &&M,
     }
     else {
         std::string name = "ccalllib_";
-        name += f_lib;
-        // backslashes and spaces causes problems on Windows linker
-        std::replace(name.begin(), name.end(), '\\', '/');
-        std::replace(name.begin(), name.end(), ' ', '_');
-        runtime_lib = true;
+        std::string f_lib_str = std::string(f_lib);
+        name += f_lib_str.substr(f_lib_str.find_last_of("/\\") + 1);
         auto &libgv = libMapGV[f_lib];
         if (libgv.first == NULL) {
             libptrgv = new GlobalVariable(*M, T_pint8, false,
--- a/src/ccall.cpp
+++ b/src/ccall.cpp
@@ -69,8 +69,9 @@ static bool runtime_sym_gvs(const char *f_lib, const char *f_name, MT &&M,
     else {
         std::string name = "ccalllib_";
         name += f_lib;
-        // backslashes causes problems on Windows linker
+        // backslashes and spaces causes problems on Windows linker
         std::replace(name.begin(), name.end(), '\\', '/');
+        std::replace(name.begin(), name.end(), ' ', '_');
         runtime_lib = true;
         auto &libgv = libMapGV[f_lib];
         if (libgv.first == NULL) {
--- a/src/ccall.cpp
+++ b/src/ccall.cpp
@@ -69,6 +69,8 @@ static bool runtime_sym_gvs(const char *f_lib, const char *f_name, MT &&M,
     else {
         std::string name = "ccalllib_";
         name += f_lib;
+        // backslashes causes problems on Windows linker
+        std::replace(name.begin(), name.end(), '\\', '/');
         runtime_lib = true;
         auto &libgv = libMapGV[f_lib];
         if (libgv.first == NULL) {


KristofferC:
Even with this I still get

```
C:/Users/Kristoffer Carlsson/.julia/artifacts/572b61b5075459e3ed62317e674398166ca98dd4/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/8.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: <unknown-file>:0: syntax error
Warning: .drectve `-export:ccalllib_C:/Users/Kristoffer_Carlsson/.julia/packages/CodecZlib/5t9zO/deps/usr/bin/libz-1.2.11.dll,data ' unrecognized
```

but I am not sure that is related to the symbol name at all anymore

KristofferC:
This is just trying to patch things in the wrong place (

KristofferC:
https://github.com/JuliaLang/julia/issues/34680

PR #34694, opened 2020-02-07
parser error for invalid `:` after imports

### LINKED ISSUE #34650: Parsing of import/using statements like `using A, B: x`
Right now 
```
using A, B: x
```

appears to be equivalent to

```
using A, B
nothing:x
```

This leads to the following error, for example:
```
julia> using Statistics, LinearAlgebra: 2
ERROR: MethodError: no method matching (::Colon)(::Nothing, ::Int64)
```

This feels like it should probably not parse, and it would be nice if there would be an error suggesting transforming the code into 

```
using A
using B: x
```

simeonschaub:
I wasn't quite sure, whether this should throw during parsing or lowering. I chose to have it throw during parsing, because I can't really see a situation, where the current parsing of expressions like `using A, B: x` as `(:)((using A, B;), x)` would be useful and we might eventually want to support this syntax and make this behave like users would expect it to.

Fix #34650

### DIFF
--- a/src/julia-parser.scm
+++ b/src/julia-parser.scm
@@ -1580,9 +1580,9 @@
                     (let ((ex (parse-comma-separated s (lambda (s)
                                                          (parse-import s word)))))
                       (if (eq? (peek-token s) ':)
-                          (error (string "unexpected \":\" in \"" word "\" syntax\n"
-                                         "Last import needs to be a separate \"" word
-                                         "\" statement"))
+                          (error (string "\":\" in \"" word "\" syntax can only be used "
+                                         "when importing a single module. "
+                                         "Split imports into multiple lines."))
                           ex)))))
     (if from
         `(,word (|:| ,first ,@rest))
--- a/src/julia-parser.scm
+++ b/src/julia-parser.scm
@@ -1578,11 +1578,12 @@
          (rest  (if done
                     '()
                     (let ((ex (parse-comma-separated s (lambda (s)
-                           (parse-import s word)))))
+                                                         (parse-import s word)))))
                       (if (eq? (peek-token s) ':)
-                        (error (string "unexpected \":\" in \"" word "\" syntax"
-                                 "\nLast import needs to be a separate \"" word "\" statement"))
-                        ex)))))
+                          (error (string "unexpected \":\" in \"" word "\" syntax\n"
+                                         "Last import needs to be a separate \"" word
+                                         "\" statement"))
+                          ex)))))
     (if from
         `(,word (|:| ,first ,@rest))
         (list* word first rest))))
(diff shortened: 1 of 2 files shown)


src/julia-parser.scm
```diff
@@ -1577,8 +1577,12 @@
                       (else #t)))
          (rest  (if done
                     '()
-                    (parse-comma-separated s (lambda (s)
-                                               (parse-import s word))))))
+                    (let ((ex (parse-comma-separated s (lambda (s)
+                           (parse-import s word)))))
```
JeffBezanson: Indentation needs to be fixed. (On the following lines as well.)

simeonschaub: So align `(parse-import s word)` to `(lambda`?

src/julia-parser.scm
```diff
@@ -1577,8 +1577,13 @@
                       (else #t)))
          (rest  (if done
                     '()
-                    (parse-comma-separated s (lambda (s)
-                                               (parse-import s word))))))
+                    (let ((ex (parse-comma-separated s (lambda (s)
+                                                               (parse-import s word)))))
+                      (if (eq? (peek-token s) ':)
+                          (error (string "unexpected \":\" in \"" word "\" syntax\n"
```
JeffBezanson: How about `":" may only appear at the beginning of "<word>" syntax`?

src/julia-parser.scm
```diff
@@ -1577,8 +1577,13 @@
                       (else #t)))
          (rest  (if done
                     '()
-                    (parse-comma-separated s (lambda (s)
-                                               (parse-import s word))))))
+                    (let ((ex (parse-comma-separated s (lambda (s)
+                                                               (parse-import s word)))))
```
JeffBezanson: Indentation

src/julia-parser.scm
simeonschaub: Maybe `":" in "<word>" syntax can only be used when importing a single module. Split imports into multiple lines.`? I think that way, it's a little bit clearer to new users, what exactly they should do to fix this.

src/julia-parser.scm
simeonschaub: What exactly is the rule for indentation here? I thought this is how lambdas should be aligned. Or should I make the whole lambda a separate line?

JeffBezanson: See the previous version of the code.

src/julia-parser.scm
JeffBezanson: Ok.

simeonschaub:
Just a friendly bump. Should be ready to merge now.

simeonschaub:
@JeffBezanson Have you had time to take another look at this?

PR #34699, opened 2020-02-08
Test for mapreducecols

kshyatt:
So far [not tested](https://codecov.io/gh/JuliaLang/julia/src/master/stdlib/SparseArrays/src/sparsematrix.jl#L1843). I checked and this test should call this function.

### DIFF
--- a/stdlib/SparseArrays/test/sparse.jl
+++ b/stdlib/SparseArrays/test/sparse.jl
@@ -2855,4 +2855,15 @@ end
     @test SparseArrays.getcolptr(vA) == SparseArrays.getcolptr(A[:, 1:5])
 end
 
+@testset "mapreducecols" begin
+    n = 20
+    m = 10
+    A = sprand(n, m, 0.2)
+    B = mapreduce(identity, +, A, dims=2)
+    for row in 1:n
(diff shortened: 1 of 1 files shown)


kshyatt:
Windows tester died in `Sockets`, Linux in `subarray`

stdlib/SparseArrays/test/sparse.jl
```diff
@@ -2855,4 +2855,14 @@ end
     @test SparseArrays.getcolptr(vA) == SparseArrays.getcolptr(A[:, 1:5])
 end
 
+@testset "mapreducecols" begin
+    n = 20
+    m = 10
+    A = sprand(n, m, 0.2)
+    B = mapreduce(identity, +, A, dims=2)
```
dkarrasch: I'd suggest to also test against the full matrix case. Like, add here:
```julia
@test B ≈ mapreduce(identity, +, Matrix(A), dims=2)
```
and then merge.

dkarrasch (commented):
This is going to lift coverage to 105%... at least! 😉 

Should be good to go after a little addition.

dkarrasch:
Looks like my suggestion fixed the CI failures... 🤣 😛

Issue #19168, opened 2016-10-31
make Bool <: Unsigned (i.e. UInt1)


StefanKarpinski:
Original title: "make `Bool` not a subtype of `Number`"

C.f. https://github.com/JuliaLang/julia/issues/18367. I propose simply changing the type hierarchy so that `Bool` is not a subtype of `Number` since it is so often an exceptional case. We could still keep useful arithmetic behaviors like `true + true == 2`, etc. – arithmetic need not *only* work for numbers. Once upon a time `UInt8` + `UInt8` produced an `Int`, but that's no longer the case, so now `Bool` is the "odd man out". If we make `!(Bool <: Number)` then we can at least make the behavior of `Number` completely consistent.

carlobaldassi:
Are there any problems caused by not having a completely consistent behavior of `Number`, i.e. of having an exception in their promotion rules as opposed to having that same "exception" outside of Numbers? (It's a honest question, I'm not trying to be polemic.)

jiahao:
Would we still allow an idiom like `sum(data .< 5)` to count the number of entries smaller than 5? I find this particular idiom very useful.

andyferris:
`count(data .< 5)` could count the number of entries greater than 5. 

That said, Stefan explicitly stated that arthmetic ops like + (used by sum) could stay the same. 

Overall, I think this is a great move. `Bool` is also unique in that it is the only type accepted by `if`, unlike in C, which is why I always thought it odd that `Bool` shared the same basket as integers.

StefanKarpinski:
> Are there any problems caused by not having a completely consistent behavior

Dispatch ambiguities used to be a big problem but since we don't warn about those anymore, they're less of a problem now. Otherwise, it's just a matter of making it harder to describe and anticipate what generic code is going to do.

JeffreySarnoff:
:100: (_so many months of Bool mucking about the Integer subtyping tree_)

StefanKarpinski:
@JeffBezanson want to know what problems this change would solve, and I'm drawing a blank. Probably because I've been on an issue triage call for six hours, so if people can post problems this causes, that would be helpful.

JeffreySarnoff:
omg -- any new type that subtypes Real pulls in Bool and that (in every case I have encountered) necessitates weird, nonintuitive method dispatch definitions involving Bool to keep the ambiguity away.

[...]

PR #34703, opened 2020-02-08
document strong zero property of `false`

tpapp:
Trivial docs change documenting the "strong zero" property of `false`.

This comes up from time to time, eg 

https://github.com/JuliaLang/julia/issues/19168#issuecomment-324741603
https://discourse.julialang.org/t/about-nan-0-0-and-false/34272/
https://discourse.julialang.org/t/inconsistency-of-nan/28671

so this should clear up some of the confusion. Thanks @mschauer for the reference.

### DIFF
--- a/doc/src/manual/mathematical-operations.md
+++ b/doc/src/manual/mathematical-operations.md
@@ -51,6 +51,18 @@ julia> 3*2/12
 operators. For instance, we would generally write `-x + 2` to reflect that first `x` gets negated,
 and then `2` is added to that result.)
 
+When used in multiplication, `false` acts as a *strong zero*:
+
+```jldoctest
+julia> NaN * false
+0.0
+
+julia> false * Inf
+0.0
+```
+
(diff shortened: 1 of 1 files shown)


tpapp:
Should the docstring of `*` refer to this, perhaps made a subsubsection?

mschauer:
It is also mentioned here https://github.com/JuliaLang/julia/blob/5315476d5779d04bfdb0cac4e8f3328621f8affe/base/docs/basedocs.jl#L1542

tpapp:
Then maybe the docstring of `*` could just mention it and direct the reader to `Bool`?

ViralBShah:
I'd like to merge this. Any objections? @JeffBezanson @StefanKarpinski

mschauer:
Oh, history by the way: Knuth's paper was uploaded on arXiv
```
[v1] Fri, 1 May 1992 00:00:00 UTC (25 KB)
```
I only found a picture of two years later:
image

The arxiv pirate logo still survives as favicon.

PR #34706, opened 2020-02-08
improve error message for using **

oscardssmith:
based on https://discourse.julialang.org/t/syntax-use-instead-of-and-syntax-is-not-a-unary-operator/34258/5, it seems that while well intentioned, this error message might be causing more confusion than it fixes. I took a stab at addressing it, but I think possibly the better change would be to parse `**` as an operator, and then we could define specific methods that threw better error messages. Specifically, `a**b` should have the error, but if we see `**a`, we should instead be telling people to use splatting, as that is more likely what they were trying to do.

### DIFF
--- a/src/julia-parser.scm
+++ b/src/julia-parser.scm
@@ -234,7 +234,7 @@
 
 (define (read-operator port c)
   (if (and (eqv? c #\*) (eqv? (peek-char port) #\*))
-      (error "If you were trying to raise a number to a power, you should use \"^\" instead of \"**\".\n If you used \"**\" in some other context, you did something wrong. This is invalid syntax."))
+      (error "For exponentiation, use \"^\" instead of \"**\".\n For splatting, use \"x...\" instead of \"**x\"."))
   (if (or (eof-object? (peek-char port)) (not (op-or-sufchar? (peek-char port))))
       (symbol (string c)) ; 1-char operator
       (let ((str (let loop ((str (string c))
--- a/src/julia-parser.scm
+++ b/src/julia-parser.scm
@@ -234,7 +234,7 @@
 
 (define (read-operator port c)
   (if (and (eqv? c #\*) (eqv? (peek-char port) #\*))
-      (error "For exponentiation, use \"^\" instead of \"**\".\n For splatting, use \"x...\" instead of \"**x\"."))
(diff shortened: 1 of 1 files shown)


src/julia-parser.scm
```diff
@@ -234,7 +234,7 @@
 
 (define (read-operator port c)
   (if (and (eqv? c #\*) (eqv? (peek-char port) #\*))
-      (error "use \"^\" instead of \"**\""))
```
stevengj: Maybe just `use "x^y" instead of "x**y" for exponentiation`

oscardssmith: Do you think this is the right approach for this improvement, rather than making `**` it's own operator?

oscardssmith:
Also, how do I make it so commits don't re-run ci. I know I'm being wasteful here, but I thought that putting [ci skip] in the title would fix it.

src/julia-parser.scm
stevengj: We're not going to make `**` the exponentiation operator in Julia…

oscardssmith: fair enough.

oxinabox:
> Also, how do I make it so commits don't re-run ci. I know I'm being wasteful here, but I thought that putting [ci skip] in the title would fix it.

You need to put it in the commit message, not in the PR title.

But this is not the kind of PR that should skip CI anyway, your not just updating spelling in the docs, you are changing the parser and if you had made a typo and forgotten a closing quote etc CI would catch it.

src/julia-parser.scm
stevengj: See above: maybe the shorter: `use "x^y" instead of "x**y" for exponentiation, and "x..." instead of "**x" for splatting`

DilumAluthge:
Yeah, definitely do not skip CI for this PR/these commits.

KristofferC:
> `parse-error: read: invalid escape sequence`

in CI so indeed, not really a good PR for ci skip.

oscardssmith:
You'd really think I'd be able to edit a single string without typos. Oh well. Fixed.

stevengj (approved):

oscardssmith:
Thanks for merging this!.

PR #34707, opened 2020-02-08
Test for copyto in sparsevec

kshyatt:
This line appears to be [uncovered](https://codecov.io/gh/JuliaLang/julia/src/master/stdlib/SparseArrays/src/sparsevector.jl#L471).


dkarrasch:
In the failed CI run, the sparsevector tests passed.

PR #34709, opened 2020-02-09
RFC: Add `Meta.parsefile`: provides an easy way to parse an entire Julia source file

DilumAluthge:
This pull request adds the new function `Meta.parsefile(filename; raise=true, depwarn=true)`.

`Meta.parsefile` provides a quick and easy way to parse an entire Julia source file.

`Meta.parsefile` returns a `Vector{Union{Expr, LineNumberNode}}` that contains all of the expressions in the file.

### DIFF
--- a/base/meta.jl
+++ b/base/meta.jl
@@ -224,6 +224,51 @@ function parse(str::AbstractString; raise::Bool=true, depwarn::Bool=true)
 end
 
 """
+    parsefile(filename; raise=true, depwarn=true)
+
+Parse the Julia file located at `filename`.
+Returns a vector that contains all expressions in the file.
+Throws an `IOError` if the file `filename` does not exist.
+If `raise` is `true` (default),
+syntax errors will raise an error; otherwise, `parse` will return an expression that will
+raise an error upon evaluation.  If `depwarn` is `false`, deprecation warnings will be
+suppressed.
+
+```julia
+julia> Meta.parsefile("foo.jl")
+```
+"""
+function parsefile(filename::AbstractString;
+                   raise::Bool=true,
+                   depwarn::Bool=true)::Vector{Union{Expr, LineNumberNode}}
+    if !ispath(filename)
+        throw(IOError("The file \"$(filename)\" does not exist"))
+    end
+    full_file_path = realpath(filename)::String
+    file_contents = read(full_file_path, String)::String
+    block = Meta.parse("begin $(file_contents) end";
+                       raise = raise,
+                       depwarn = depwarn)::Expr
+    block_fixed = _parsefile_fix_filename(block,
+                                          full_file_path)::Expr
+    lines = convert(Vector{Union{Expr, LineNumberNode}}, block_fixed.args)::Vector{Union{Expr, LineNumberNode}}
+    return lines
+end
+
+@inline function _parsefile_fix_filename(expr::Expr,
+                                         correct_filename::String)::Expr
+    return Expr(expr.head,
+                _parsefile_fix_filename.(expr.args, correct_filename)...)
+end
+
+@inline function _parsefile_fix_filename(lnn::LineNumberNode,
+                                         correct_filename::String)::LineNumberNode
+    return LineNumberNode(lnn.line, correct_filename)
+end
+
+@inline _parsefile_fix_filename(x, correct_filename::String) = x
+
+"""
     partially_inline!(code::Vector{Any}, slot_replacements::Vector{Any},
                       type_signature::Type{<:Tuple}, static_param_values::Vector{Any},
                       slot_offset::Int, statement_offset::Int,


base/meta.jl
```diff
@@ -223,6 +223,56 @@ function parse(str::AbstractString; raise::Bool=true, depwarn::Bool=true)
...
+Parse the Julia file located at `filename`.
+Returns a vector that contains all expressions in the file.
+Throws an `IOError` if the file `filename` does not exist.
+If `raise` is `true` (default),
+syntax errors will raise an error; otherwise, `parse` will return an expression that will
+raise an error upon evaluation.  If `depwarn` is `false`, deprecation warnings will be
+suppressed.
+
+```julia
+julia> Meta.parse("foo.jl")
```
cmcaine:
```suggestion
julia> Meta.parsefile("foo.jl")
```

base/meta.jl
```diff
@@ -223,6 +223,56 @@ function parse(str::AbstractString; raise::Bool=true, depwarn::Bool=true)
...
+    block::Expr = Meta.parse("""
+                             begin
+                                 $(file_contents)
+                             end
+                             """;
+                             raise = raise,
+                             depwarn = depwarn)::Expr
+    block_fixed::Expr = _parsefile_fix_filename(block,
+                                                full_file_path)::Expr
+    lines::Vector{Union{Expr, LineNumberNode}} = convert(Vector{Union{Expr, LineNumberNode}}, block_fixed.args)::Vector{Union{Expr, LineNumberNode}}
```
cmcaine: Just a question: this type is defined/asserted three times. Is there a nice way of not doing that?

base/meta.jl
DilumAluthge: Thanks!

base/meta.jl
DilumAluthge: Yeah I think I can remove one of those.

DilumAluthge: Take a look now.

We have to, for example, keep the `convert(TYPE, something)::TYPE`. The compiler is not able to infer the return type of `convert`.

base/meta.jl
```diff
@@ -223,6 +223,55 @@ function parse(str::AbstractString; raise::Bool=true, depwarn::Bool=true)
...
+
+@inline function _parsefile_fix_filename(expr::Expr,
+                                         correct_filename::String)::Expr
+    return Expr(expr.head,
+                _parsefile_fix_filename.(expr.args, correct_filename)...)
+end
+
+@inline function _parsefile_fix_filename(lnn::LineNumberNode,
+                                         correct_filename::String)::LineNumberNode
+    return LineNumberNode(lnn.line, correct_filename)
```
davidanthoff: This should probably be `lnn.line-1`, to adjust for the `begin` keyword that surrounds things.

DilumAluthge: Done!

DilumAluthge: @davidanthoff I removed the newline after `begin`. I think that means we don't need to subtract `1` from the line number. Am I correct?

davidanthoff: I would assume so!

DilumAluthge: If you get a chance, test it out.

cmcaine:
Surely this should just have automated tests in the PR?

DilumAluthge:
Yeah I need to write tests... but it might be a while before I have the time.

base/meta.jl
davidanthoff: I'm not setup to compile Julia itself...

timholy:
Given that this can be done as the one-liner
```julia
Base.parse_input_line(read(filename, String); filename=filename)
```
do we need a function for this?

cmcaine:
parse_input_line is undocumented and the name doesn't suggest that it could
be used for this.

FWIW, I never really know if undocumented, unexported functions are part of
the API or not, so I probably wouldn't have found or used this.

So at a minimum it would have to get some docs, and maybe we'd have to
include this example or put this example on stack overflow to get similar
discoverability.

timholy:
You do have a point. We could document it or add `Meta.parsefile` as a wrapper.

DilumAluthge:
> Given that this can be done as the one-liner
> 
> ```julia
> Base.parse_input_line(read(filename, String); filename=filename)
> ```
> 
> do we need a function for this?

Will this give LineNumberNodes with the correct filename and line numbers?

timholy:
It should, as long as you supply that keyword argument.

DilumAluthge:
> It should, as long as you supply that keyword argument.

Ah I didn't realize that. So then this functionality does exist already.

I'm going to close this PR and make a separate PR to add a docstring to `Base.parse_input_line`.

DilumAluthge:
See: https://github.com/JuliaLang/julia/pull/34715

PR #34711, opened 2020-02-09
array.c: always unshare on `push!`

### LINKED ISSUE #33143: reshape and push! inconsistent behavior
Consider the following example:

```julia
xmat = randn(5,5)
xvec = reshape(xmat, 25)
push!(xvec, 3)
xvec[1] = 0.0
xmat[1] # not zero
```

Here `push!` has decoupled `xmat` and `xvec`, though one would expect that a reshaped array always point to the same memory as the original array. I think the `push!` here should raise an error.

Contrast the example above with the following (source: https://discourse.julialang.org/t/puzzling-reshape-and-push/28326/2):

```julia
xvec = rand(4)
xmat = reshape(xvec, 2, 2)
push!(xvec, 1.0)  # ERROR: cannot resize array with shared data
```

To be consistent it seems to me that my first example above should also be an error.

anaveragehuman:
Do not merge; will fix #33143

### DIFF
--- a/src/array.c
+++ b/src/array.c
@@ -637,7 +637,6 @@ JL_DLLEXPORT void jl_arrayunset(jl_array_t *a, size_t i)
 static int NOINLINE array_resize_buffer(jl_array_t *a, size_t newlen)
 {
     jl_ptls_t ptls = jl_get_ptls_states();
-    assert(!a->flags.isshared || a->flags.how == 3);
     size_t elsz = a->elsize;
     size_t nbytes = newlen * elsz;
     size_t oldnbytes = a->maxsize * elsz;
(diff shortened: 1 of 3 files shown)


yuyichao:
If I’m reading the code correctly, I don’t think this patch works.


Also, I believe this cannot be changed as simple as this. The data is always owned by an array and you cannot reassign that.

yuyichao:
Note that the current implementation on the array side heavily depend on the fact that shared array cannot be resized. The assumptions that I used are documented in the comment that you removed,

> If inc > 0, it will always trigger the slow path and unshare the buffer

This assumption is broken by this PR.

That's really the most minor issue though. `array_resize_buffer` will happily delete the old buffer. That's harder to fix, but not impossible although you may have a hard time finding the right function to call because of the next issue.

Again, as I said above, this is not doable without major change in the way array data is managed in GC. The best way to handle it without breaking user code might be to track all array sharing although I'm not sure about that either...

JeffBezanson:
To explain a bit more, given the sequence
```
vec = ...
mat = reshape(vec, m, n)
push!(vec, y)
```
the problem is that `mat` depends on `vec` to root the underlying buffer for it when using the "how == 2" allocation strategy (`malloc`). If we use the "how == 1" strategy, then the buffer itself is a tracked object and `mat` will be able to root it. However, in that case we're unable to use `realloc` to grow the buffer. So I think one way to implement this would be to add GC support for realloc'ing a tracked buffer, then replace the how==2 strategy with how==1, and use the new realloc support when `isshared` is false.

PR #34715, opened 2020-02-10
Add `Meta.parsefile` and `Meta.parseall` functions as a wrapper around `Base.parse_input_line`

DilumAluthge:
Replacement for #34709 

This PR adds the `Meta.parsefile` function as a wrapper around the unexported `Base.parse_input_line`.

cc: @davidanthoff @cmcaine @timholy

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -57,6 +57,8 @@ New library functions
   expressions before they are evaluated ([#34595]).
 * New function `bitreverse` for reversing the order of bits in a fixed-width integer ([#34791]).
 * New function `bitrotate(x, k)` for rotating the bits in a fixed-width integer ([#33937]).
+* The new `Meta.parsefile` function parses an entire Julia source file ([#34715]).
+* The new `Meta.parseall` function parses all of the code in the input string at the top-level ([#34715]).
 
 New library features
 --------------------
--- a/base/meta.jl
+++ b/base/meta.jl
@@ -224,6 +224,47 @@ function parse(str::AbstractString; raise::Bool=true, depwarn::Bool=true)
 end
 
 """
+    parsefile(filename::AbstractString)
+
+Parse the Julia source file located at `filename` and return a `:toplevel`
+expression that contains all expressions in the file. If the source code
+contains one or more syntax errors, `parsefile` will return an expression
+that will raise an error upon evaluation.
+
+# Examples
+```julia-repl
+julia> expr = Meta.parsefile("hello.jl")
+```
+
+!!! compat "Julia 1.5"
+    This function is available as of Julia 1.5.
+"""
+function parsefile(filename::AbstractString)
+    return Base.parse_input_line(read(filename, String);
+                                 filename = String(filename))
+end
+
+"""
+    parseall(str::AbstractString)
+
+Parse the Julia source code contained in `str` and return a `:toplevel`
+expression that contains all expressions in the input string. If the source code
+contains one or more syntax errors, `parseall` will return an expression
+that will raise an error upon evaluation.
+
+# Examples
+```julia-repl
+julia> expr = Meta.parsefile("a = 3\nb = 4\nc = sqrt(a^2 + b^2)\n")
+```
+
+!!! compat "Julia 1.5"
+    This function is available as of Julia 1.5.
+"""
+function parseall(str::AbstractString)
+    return Base.parse_input_line(String(str))
+end
+
+"""
     partially_inline!(code::Vector{Any}, slot_replacements::Vector{Any},
                       type_signature::Type{<:Tuple}, static_param_values::Vector{Any},
                       slot_offset::Int, statement_offset::Int,
--- a/test/meta.jl
+++ b/test/meta.jl
@@ -210,3 +210,75 @@ let a = 1
     @test !macroexpand(@__MODULE__, :(@is_dollar_expr $a))
     @test @macroexpand @is_dollar_expr $a
 end
+
+@testset "Meta.parsefile" begin
+    @testset "successful parsing, with correct filename in the LineNumberNodes" begin
+        code = """
+            a = 3
+            b = 4
+            c = sqrt(a^2 + b^2)
+        """
+        filename, expr = mktemp() do fn, io
+            println(io, code)
+            flush(io)
+            return fn, Meta.parsefile(fn)
+        end
+        @test expr isa Expr
+        @test expr.head == :toplevel
+        @test length(expr.args) == 6
+        @test expr.args[1] == LineNumberNode(1, Symbol(filename))
+        @test expr.args[2] == :(a = 3)
+        @test expr.args[3] == LineNumberNode(2, Symbol(filename))
+        @test expr.args[4] == :(b = 4)
+        @test expr.args[5] == LineNumberNode(3, Symbol(filename))
+        @test expr.args[6] == :(c = sqrt(a ^ 2 + b ^ 2))
+    end
+    @testset "syntax error" begin
+        code = """
+            a = 1
+            b = 2.2.2
+            c = 3
+        """
+        filename, expr = mktemp() do fn, io
+            println(io, code)
+            flush(io)
+            return fn, Meta.parsefile(fn)
+        end
+        @test expr isa Expr
+        @test expr.head == :error
+        @test length(expr.args) == 1
+        @test expr == :($(Expr(:error, "invalid numeric constant \"2.2.\"")))
+    end
+end
+
+@testset "Meta.parseall" begin
+    @testset "successful parsing" begin
+        code = """
+            a = 3
+            b = 4
+            c = sqrt(a^2 + b^2)
+        """
+        expr = Meta.parseall(code)
+        @test expr isa Expr
+        @test expr.head == :toplevel
+        @test length(expr.args) == 6
+        @test expr.args[1] == LineNumberNode(1, :none)
+        @test expr.args[2] == :(a = 3)
+        @test expr.args[3] == LineNumberNode(2, :none)
+        @test expr.args[4] == :(b = 4)
+        @test expr.args[5] == LineNumberNode(3, :none)
+        @test expr.args[6] == :(c = sqrt(a ^ 2 + b ^ 2))
+    end
+    @testset "syntax errors" begin
+        code = """
+            a = 1
+            b = 2.2.2
+            c = 3
+        """
+        expr = Meta.parseall(code)
+        @test expr isa Expr
+        @test expr.head == :error
+        @test length(expr.args) == 1
+        @test expr == :($(Expr(:error, "invalid numeric constant \"2.2.\"")))
+    end
+end


base/client.jl
```diff
@@ -169,24 +169,50 @@ function _parse_input_line_core(s::String, filename::String)
...
+julia> filename = "hello.jl"
+
+julia> expr = parse_input_line(read(filename, String); filename = filename)
+```
+"""
+function parse_input_line(s::String; filename::AbstractString = "none", depwarn::Bool = true)
     # For now, assume all parser warnings are depwarns
     ex = if depwarn
-        _parse_input_line_core(s, filename)
+        _parse_input_line_core(s, String(filename))
```
c42f: I'd be inclined to leave this conversion to `_parse_input_line_core` which is the location where the concrete type is required.

base/client.jl
```diff
@@ -169,24 +169,50 @@ function _parse_input_line_core(s::String, filename::String)
...
+julia> expr = parse_input_line("c = sqrt(a^2 + b^2)")
+```
+
+```julia
+julia> filename = "hello.jl"
+
+julia> expr = parse_input_line(read(filename, String); filename = filename)
+```
+"""
+function parse_input_line(s::String; filename::AbstractString = "none", depwarn::Bool = true)
```
c42f: I suspect we can remove the `depwarn` for now; as far as I can tell, the parser doesn't depwarn anymore.

base/client.jl
```diff
@@ -169,24 +169,50 @@ function _parse_input_line_core(s::String, filename::String)
...
     return ex
 end
-parse_input_line(s::AbstractString) = parse_input_line(String(s))
+parse_input_line(s::AbstractString; kwargs...) = parse_input_line(String(s); kwargs...)
+
+"""
+    parse_input_line(io::IO; filename::AbstractString = "none", depwarn::Bool = true)
 
-function parse_input_line(io::IO)
+Read input from `io` and parse it as Julia code.
```
c42f:
```suggestion
Read lines from `io` using `readline` until the result is a complete julia expression.
```

c42f (commented):
This looks potentially useful (could you provide some background on the use cases for this outside Julia?).

I think we might come up with a clearer name if it's going to be a public API, especially because this can consume more than one line when reading from an `IO`.

JeffBezanson:
This should be in the `Meta` module along with `parse`. Then it doesn't need to be exported from Base. It could be called `parseall`, or possibly add an `all` keyword argument to `parse` instead.

The `parse_input_line(io::IO)` method seems different to me and should be a different function. It doesn't parse the entire input; it reads lines only until the parser stops returning `incomplete`. I think it could be inlined into its only caller in the stream REPL in client.jl.

DilumAluthge:
> This should be in the `Meta` module along with `parse`. Then it doesn't need to be exported from Base. It could be called `parseall`, or possibly add an `all` keyword argument to `parse` instead.

How about `Meta.parsefile`?

base/client.jl
DilumAluthge: In the new `Meta.parsefile` function, I removed the `depwarn` keyword argument.

base/client.jl
DilumAluthge: In the new `Meta.parsefile`, I kept this conversion in. The reason is that if we don't convert `filename` to a `String`, then if e.g. you pass a `SubString{String}` as the `filename`, when you try to call `parse_input_line` you get `ERROR: TypeError: in keyword argument s, expected String, got SubString{String}`.

base/client.jl
DilumAluthge: I think this is now outdated.

DilumAluthge:
> This looks potentially useful (could you provide some background on the use cases for this outside Julia?).

So far, the only use cases I've found are inside Julia. It comes up for me when I'm trying to do transformations on Julia source code at the source file level (as opposed to on the level of individual expressions).

I believe @davidanthoff also has a use case.

DilumAluthge:
> I think we might come up with a clearer name if it's going to be a public API

I've changed it to `Meta.parsefile`. How does that sound?

DilumAluthge:
FYI - I couldn't figure out where the tests for the `Meta` module go. So I put the tests at the bottom of the `test/meta.jl`. Is that the correct file?

base/meta.jl
```diff
@@ -223,6 +223,50 @@ function parse(str::AbstractString; raise::Bool=true, depwarn::Bool=true)
...
+"""
+function parsefile(filename::AbstractString; fix_line_number_nodes::Bool = true)
+    fn = String(filename)
+    expr = Base.parse_input_line(read(filename, String); filename = fn)
+    # At this point, `expr` is a `:toplevel` expression that contains
+    # everything in the source file.
+    # `expr` contains multiple `LineNumberNode`s.
+    # Unfortunately, currently some of the `LineNumberNode`s do
+    # not have a filename, i.e. the value of the `file` field is
+    # `nothing`. If `fix_line_number_nodes` is true (default), we
```
c42f: How does this happen? Naively it sounds like a bug elsewhere which should be fixed rather than worked around here.

c42f:
> I put the tests at the bottom of the `test/meta.jl`. Is that the correct file?

Yes, I believe that's correct :+1:

c42f:
> How about `Meta.parsefile`?

That's a nice name for parsing files but what if a user has the content in a string and they want to parse multiple expressions using the same top level rules that are used for files? For that case I think `parseall` is a better name, or better `parse(str; all=true)` to parse all top level expressions within the string. (Both suggested by Jeff above.)

You should also consider the error handling behavior of `_parse_input_line_core`. It's specialized for the needs of the stream REPL in client.jl but that's likely not appropriate. What should `parseall` do when it hits a parse error? `parse` has the `raise` keyword for this and it's own particular behavior. Presumably you should do the same.

base/meta.jl
JeffBezanson: I'll look into it.

JeffBezanson:
:+1: from triage.

StefanKarpinski:
Triage dixit: This seem fine, but the `fix_line_number_nodes` option seems super iffy. Shouldn't we just always fix up the line number nodes?

DilumAluthge:
> Triage dixit: This seem fine, but the `fix_line_number_nodes` option seems super iffy. Shouldn't we just always fix up the line number nodes?

Done! We now always fix up the line number nodes.

NEWS.md
```diff
@@ -56,6 +56,7 @@ New library functions
 * `include` now accepts an optional `mapexpr` first argument to transform the parsed
   expressions before they are evaluated ([#34595]).
 * New function `bitreverse` for reversing the order of bits in a fixed-width integer ([#34791]).
+* The new `Meta.parsefile` function parses an entire Julia source file.
```
StefanKarpinski:
```suggestion
* The new `Meta.parsefile` function parses an entire Julia source file ([#34715]).
```

base/meta.jl
```diff
@@ -223,6 +223,44 @@ function parse(str::AbstractString; raise::Bool=true, depwarn::Bool=true)
...
+expression that contains all expressions in the file.
+
+If `fix_line_number_nodes` is `true` (default), then all of the line number
+nodes in the result will have the correct filename. If `fix_line_number_nodes`
+is `false`, then some of the line number nodes may have `nothing` as the
+filename.
+
+julia> expr = Meta.parsefile("hello.jl")
+"""
+function parsefile(filename::AbstractString; fix_line_number_nodes::Bool = true)
```
StefanKarpinski:
```suggestion
function parsefile(filename::AbstractString)
```

test/meta.jl
```diff
@@ -210,3 +210,42 @@ let a = 1
...
 end
+
+@testset "Meta.parsefile" begin
+    @testset "fix_line_number_nodes = true" begin
+        filename, expr = mktemp() do fn, io
+            println(io, "a = 3")
+            println(io, "b = 4")
+            println(io, "c = sqrt(a^2 + b^2)")
+            flush(io)
+            return fn, Meta.parsefile(fn; fix_line_number_nodes = true)
```
StefanKarpinski:
```suggestion
            return fn, Meta.parsefile(fn)
```

base/meta.jl
DilumAluthge: Done!

test/meta.jl
```diff
@@ -210,3 +210,42 @@ let a = 1
...
+        @test expr.head == :toplevel
+        @test length(expr.args) == 6
+        @test expr.args[1] == LineNumberNode(1, Symbol(filename))
+        @test expr.args[2] == :(a = 3)
+        @test expr.args[3] == LineNumberNode(2, Symbol(filename))
+        @test expr.args[4] == :(b = 4)
+        @test expr.args[5] == LineNumberNode(3, Symbol(filename))
+        @test expr.args[6] == :(c = sqrt(a ^ 2 + b ^ 2))
+    end
+    @testset "fix_line_number_nodes = false" begin
```
StefanKarpinski: Delete this testset.

DilumAluthge: Done

base/meta.jl
```diff
@@ -223,6 +223,39 @@ function parse(str::AbstractString; raise::Bool=true, depwarn::Bool=true)
...
+    fn = String(filename)
+    expr = Base.parse_input_line(read(filename, String); filename = fn)
+    # At this point, `expr` is a `:toplevel` expression that contains
+    # everything in the source file.
+    # `expr` contains multiple `LineNumberNode`s.
+    # Unfortunately, currently some of the `LineNumberNode`s do
+    # not have a filename, i.e. the value of the `file` field is
+    # `nothing`. We go through and replace all of those `nothing`s
+    # with the correct filename. 
+    return _fix_line_number_nodes(expr, Symbol(fn))
```
timholy: I'd rather not need this. Is there an issue filed about when this happens? Seems like this should be done elsewhere and the fixing deleted from this PR.

timholy (commented):
I'd approve once the LineNumberNode fixing is deleted from this PR and an issue filed with test cases illustrating where the problem happens.

base/meta.jl
DilumAluthge: Not sure if an issue is open. I think @JeffBezanson said that he was looking into it?

timholy: If you had a reproducible example it might help him find the problem. I haven't noticed this problem myself, so I'm wondering if it only happens in a subset of circumstances. (Or perhaps I'm just blind.)

DilumAluthge: If you run the test case in this PR without fixing the line numbers, then of the three LineNumberNodes, the first one has the correct file name, and the latter two have `nothing` has the file name.

c42f: Seems to be a bug in the frontend flisp code as indicated by the following:

```
;  _
; |_ _ _ |_ _ |  . _ _
; | (-||||_(_)|__|_)|_)
;-------------------|----------------------------------------------------------

> (jl-parse-all "a = 1\nb=2\n" "blah.jl")
(toplevel (line 1 blah.jl) (= a 1) (line 2)
	  (= b 2))
```

c42f:
I believe I've fixed the issue which leads to the need for `_fix_line_number_nodes` in https://github.com/JuliaLang/julia/pull/34881.

So I think you can remove `_fix_line_number_nodes` from this PR (either wait for #34881 to merge, or immediately)

DilumAluthge:
❤️

I’ll wait for that PR to be merged, then I’ll remove the "fix line number stuff", rebase on master, etc.

StefanKarpinski:
It's merged!

DilumAluthge:
Alright, this should be ready to merge once CI is green.

base/meta.jl
```diff
@@ -223,6 +223,19 @@ function parse(str::AbstractString; raise::Bool=true, depwarn::Bool=true)
     return ex
 end
 
+"""
+    parsefile(filename)
+
+Parse the Julia source file located at `filename` and return a `:toplevel`
+expression that contains all expressions in the file.
+
+julia> expr = Meta.parsefile("hello.jl")
```
fredrikekre:
````suggestion
# Examples
```julia-repl
julia> expr = Meta.parsefile("hello.jl")
```

!!! compat "Julia 1.5"
    This function is available as of Julia 1.5.  
````

File: base/meta.jl [hunk 11]
```diff
@@ -223,6 +223,25 @@ function parse(str::AbstractString; raise::Bool=true, depwarn::Bool=true)
     return ex
 end
 
+"""
+    parsefile(filename)
+
+Parse the Julia source file located at `filename` and return a `:toplevel`
+expression that contains all expressions in the file.
+
```
c42f: What happens when there's a parse error?

I think we should consider making this (a) consistent with `Meta.parse` and (b) documented.

c42f (commented):
The implementation looks a lot cleaner now which is great.

I'd still appreciate answers to some previous questions which I don't think were addressed:
* What if someone has the content of the file in a string? `parsefile` taking a file name for opening doesn't seem like the primitive operation here. A more primitive and reusable operation is "parse all code in this string at top level". Better named `parseall` or `Meta.parse(..., all=true)`?
* What about errors? The behavior of the internal `parse_input_line` may not be what we want here. I feel like error handling should be consistent with `Meta.parse()`, or at least documented and tested.

DilumAluthge:
> What if someone has the content of the file in a string? parsefile taking a file name for opening doesn't seem like the primitive operation here. A more primitive and reusable operation is "parse all code in this string at top level". Better named parseall or Meta.parse(..., all=true)?

I think we need to have both `Meta.parsefile` and `Meta.parseall`. If you only have `Meta.parseall`, and it takes in a string, then you won't get `LineNumberNode`s with the correct filename.

base/meta.jl
DilumAluthge:
```julia-repl
shell> cat test.jl
a = 1
b = 2.2.2
c = 3

julia> parsefile("test.jl")
:($(Expr(:error, "invalid numeric constant \"2.2.\"")))
```

DilumAluthge: I have added to the docstring explaining the behavior when there is a parse error, and I have added some tests.

DilumAluthge:
> I feel like error handling should be consistent with Meta.parse(), or at least documented and tested.

Since we are adding the `Meta.parsefile` and `Meta.parseall` functions as simple wrappers around `Base.parse_input_line`, I think we are stuck with the error behavior of `Base.parse_input_line`. I have added documentation to the docstrings and have also added some tests.

If we really want to replicate error behavior of `Meta.parse`, then I think our functions `Meta.parsefile` and `Meta.parseall` would need to call `Meta.parse`. That was the approach I used originally in https://github.com/JuliaLang/julia/pull/34709, but the suggestion from @timholy was to use `Base.parse_input_line` instead.

DilumAluthge:
I have no problem switching back to using `Meta.parse`. I'm happy to do whatever the consensus is.

c42f:
Just FYI we've been discussing this exact issue (from a unifyinig-the-internals viewpoint) over at https://github.com/JuliaLang/julia/pull/35149. I feel like we'll want a single entry point `Meta.parse()` and `Meta.parsefile()` can be a utility function. (I wish we had a file path type to dispatch on and we could truly have a single generic function for this...)

Give me a little time to play with things internally — I think we should hold off on `parseall` until we have a clearer view of what the internal API will need. (In particular I'm thinking about what's a natural extension for the greedy flag; you can think of `greedy=true/false` and a potential `all=true/false` as an awkward way of communicating which grammar rule the parser will attempt to match. Not that we have a formal grammar!)

base/meta.jl
JeffBezanson: I'm not sure this is the right behavior. It doesn't make it possible to emulate `include`, which evaluates the expressions before the error. I guess we decided not to do that in the REPL, but I'm not sure why; maybe a REPL input is more of an atomic unit than a file. Anyway it seems to me that should be handled only by the REPL and not a parsing function.

c42f: Agreed. I need to sort this out while refactoring the internals so that `include` goes via the same code path that we have here. For that it seems we need to return the list of correctly parsed statements, and also include the error somehow (at the end of the list? separately?)

For this user-facing API I'd favor emulating `Meta.parse` and the `raise` keyword. In fact I'm pretty sure everything should go via `Meta.parse`; we've just got to generalize `greedy` somehow and allow a filename to be passed.

For `greedy`, I'm thinking we should allow specification of grammar rules via `rule=:atom` vs `rule=:statement` vs `rule=:toplevel`? But I'm not sure this is a good general way to enter the current parser and whether it generalizes between parsers. Also we don't have a formal grammar...

JeffBezanson: > we've just got to generalize `greedy` somehow

As you point out, we only seem to need 3 options. I don't have a strong opinion on what the interface should be though; a keyword argument with a few symbol values is not necessarily better than having 2-3 functions.

> Also we don't have a formal grammar...

If you really need to talk about specific grammar productions by name, each one is defined by a function called `parse-X`, so there you go. But, I really don't think it's too useful to make the exact set of productions part of the interface to `parse`. Even grammars written out in BNF will often have rules that are really just implementation details.

c42f:
#35149 is now merged in 1.6-DEV. It includes `Meta.parseall` and `Meta.parseatom` as internal functions — a `parseall` was required for the internal changes and having multiple functions seemed to be the API Jeff preferred.

However, they are not fully fleshed out yet with precisely defined public APIs - I thought this PR would be the logical place to finish doing that.

PR #34716, opened 2020-02-10
fix _spawn_primitive to report error correctly

tanmaykm:
Method `_spawn_primitive` in `process.jl` seems to be defining a local valiable that shadows the `error` method that it tries to use.
This renames the local variable to fix that.

Below are a few lines to simulate the condition.

Before:

```
julia> Base.rawhandle(s::String) = s

julia> Base._spawn_primitive("/tmp/xyx", `ls`, Any["abc"])
ERROR: UndefVarError: error not defined
Stacktrace:
 [1] _spawn_primitive(::String, ::Cmd, ::Array{Any,1}) at ./process.jl:79
```

With this fix:

```
julia> Base.rawhandle(s::String) = s

julia> Base._spawn_primitive("/tmp/xyx", `ls`, Any["abc"])
ERROR: invalid spawn handle abc from abc
Stacktrace:
 [2] _spawn_primitive(::String, ::Cmd, ::Array{Any,1}) at ./process.jl:79
 [3] top-level scope at REPL[4]:1
```

### DIFF
--- a/base/process.jl
+++ b/base/process.jl
@@ -84,7 +84,7 @@ const SpawnIOs = Vector{Any} # convenience name for readability
         for io in stdio]
     handle = Libc.malloc(_sizeof_uv_process)
     disassociate_julia_struct(handle) # ensure that data field is set to C_NULL
-    error = ccall(:jl_spawn, Int32,
+    err = ccall(:jl_spawn, Int32,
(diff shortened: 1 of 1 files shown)


tanmaykm:
Win32 test seem to have timed out (also it looks like all test cases in that did pass).
Maybe unrelated to this PR?

PR #34718, opened 2020-02-10
Add typetrace(t) for returning the recursive supertypes of Type `t`

IanButterworth:
A simple function that might be useful-enough to put in base?
From a discussion with @DilumAluthge and @dpsanders on slack

```julia
julia> typetrace(Int32)
"Int32 <: Signed <: Integer <: Real <: Number <: Any"

julia> typetrace(Int32, returnarray=true)
6-element Array{DataType,1}:
 Any    
 Number 
 Real   
 Integer
 Signed 
 Int32  
```

I can add tests & documentation if there's interest


ianshmean:
Perhaps `supertypetrace` is a better name?

KristofferC:
Already implemented in https://github.com/JuliaLang/julia/pull/34419/files?

ianshmean:
Ah, damn. Forgot to check master..

PR #34722, opened 2020-02-10
Add \bbsemi as a binary operator

tkf:
`;` is used as the composition operator in "diagrammatic order" ([Category (mathematics) - Wikipedia](https://en.wikipedia.org/wiki/Category_(mathematics))).  Although we can't use the normal semicolon as a binary operator in Julia, we can include `⨟` ([`\fatsemi`](https://tex.stackexchange.com/questions/125012/what-is-the-latex-code-for-an-open-semicolon)) as a binary operator.

A big thanks to the post by @schlichtanders in Discourse: https://discourse.julialang.org/t/add-composition-operator-unicode-2a1f-for-flipped/34436


stevengj:
What’s the difference between [U+2a1f](https://www.fileformat.info/info/unicode/char/2a1f/index.htm) and [U+2a3e](https://www.fileformat.info/info/unicode/char/2a3e/index.htm)

tkf:
Formal Specification— Z Notation— Syntax, Type and Semantics (http://www.open-std.org/jtc1/sc22/open/n3187.pdf) says

> schema composition ‘o9’ and the mathematical toolkit character relational composition ‘o9’ are different Z characters;

(`o9` is rendered as `o` on top of `9` in the PDF)

The motivation discussed in the aforementioned discourse thread was to define `f ⨟ g = g ∘ f` that is presumably primarily used for functions. So, IIUC, I think relational composition (2A3E) makes more sense (see Table 14). @schlichtanders @jw3126 What do you think?

As two characters are pretty much identical, I think omitting schema composition (2A1F) makes sense.  This is compatible with that we don't have, e.g., `\setminus` (`∖`).

Another consideration may be to see if there are fonts that have only one of the characters. But so far I find that it's either both of them are rendered or none of them are rendered (though maybe it's just my editor and terminal are finding fallbacks).

schlichtanders:
> What’s the difference between [U+2a1f](https://www.fileformat.info/info/unicode/char/2a1f/index.htm) and [U+2a3e](https://www.fileformat.info/info/unicode/char/2a3e/index.htm)?

From simple inspection of the unicode alternatives at this page https://unicode.org/charts/PDF/U2A00.pdf

I find 2A1F much more appealing visually (compared to its counterpart 2A3E). In addition, also its latex name is lovely self-descriptive ``\fatsemi``

tkf:
It seems there are multiple ways to enter certain symbols. So I hope it's OK to add both `\fatsemi` and `\bb;` (as how it's implemented in the current PR).

```julia
julia> length(unique(keys(REPL.REPLCompletions.latex_symbols)))
2495

julia> length(unique(values(REPL.REPLCompletions.latex_symbols)))
2475
```

Re: 2A3E vs 2A1F, I don't really mind which one is chosen.  It looks like 2A1F is rendered as a bigger symbol so maybe it's better for readability.

cormullion:
There's a bit of variety in the commonly used fonts: 



Z NOTATION SCHEMA COMPOSITION 2A1F
Z NOTATION RELATIONAL COMPOSITION 2A3E

(But a mild vote against "fatsemi" which is far from a "lovely" name... :))

JeffBezanson:
I think `\bbsemi` would be more consistent with other latex names than `\bb;`.

tkf:
I replaced `\bb;` with `\bbsemi`.  Since no one seems to be opposing U+2a1f, I also replaced u+2a3e with u+2a1f.

Can this be merged? Is there anything else to be addressed?

stdlib/REPL/src/latex_symbols.jl
```diff
@@ -524,6 +524,8 @@ const latex_symbols = Dict(
     "\\setminus" => "∖",
     "\\ast" => "∗",
     "\\circ" => "∘",
+    "\\fatsemi" => "⨟",
+    blackboard*"semi" => "⨟",
```
stevengj: My vote would be to just have the second of these.  `\fatsemi` is less consistent with our other abbreviations, and is hardly standard (it is just the name in one LaTeX add-on package as far as I can tell?)

tkf:
I removed `\fatsemi`.

Note that people wants to use this alias can simply do:

```julia
julia> using REPL

julia> REPL.REPLCompletions.latex_symbols["\\fatsemi"] = "⨟"
```

stevengj (approved):

PR #34727, opened 2020-02-11
Fix pretty printing of timings when building Julia with no stdlibs

fredrikekre:
PackageCompiler can filter out stdlibs so it can be empty, and this line fails when reducing over empty collection.

### DIFF
--- a/base/sysimg.jl
+++ b/base/sysimg.jl
@@ -49,7 +49,7 @@ let
             :Statistics,
         ]
 
-    maxlen = maximum(textwidth.(string.(stdlibs)))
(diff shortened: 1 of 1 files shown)


base/sysimg.jl
```diff
@@ -49,7 +49,7 @@ let
             :Statistics,
         ]
 
-    maxlen = maximum(textwidth.(string.(stdlibs)))
+    maxlen = 14 # number of chars in longest stdlib name
```
martinholters: `reduce(max, ...; init=0)`?

Issue #33409, opened 2019-09-28
BLAS threads should default to physical not logical core count?


ChrisRackauckas:
On an i7-8550u, OpenBLAS is defaulting to 8 threads. I was comparing to RecursiveFactorizations.jl, and saw the performance is like:

```julia
using BenchmarkTools
import LinearAlgebra, RecursiveFactorization
ccall((:openblas_get_num_threads64_, Base.libblas_name), Cint, ())
@which LinearAlgebra.BLAS.set_num_threads(4)
BenchmarkTools.DEFAULT_PARAMETERS.seconds = 0.5

luflop(m, n) = n^3÷3 - n÷3 + m*n^2
luflop(n) = luflop(n, n)

bas_mflops = Float64[]
rec_mflops = Float64[]
ns = 50:50:800
for n in ns
    A = rand(n, n)
    bt = @belapsed LinearAlgebra.lu!($(copy(A)))
    rt = @belapsed RecursiveFactorization.lu!($(copy(A)))
    push!(bas_mflops, luflop(n)/bt/1e9)
    push!(rec_mflops, luflop(n)/rt/1e9)
end

using Plots
plt = plot(ns, bas_mflops, legend=:bottomright, lab="OpenBLAS", title="LU Factorization Benchmark", marker=:auto)
plot!(plt, ns, rec_mflops, lab="RecursiveFactorization", marker=:auto)
xaxis!(plt, "size (N x N)")
yaxis!(plt, "GFLOPS")
```

1 thread

![lubenchbnumthreads1](https://user-images.githubusercontent.com/1814174/65809331-79ef0000-e16a-11e9-8e34-7ee6a5421d29.png)

4 threads

![lubenchbnumthreads4](https://user-images.githubusercontent.com/1814174/65809337-81160e00-e16a-11e9-9e74-c000b8a469ec.png)

8 threads

![lubench](https://user-images.githubusercontent.com/1814174/65809340-85422b80-e16a-11e9-99a5-78facf02fb5f.png)

Conclusion: the default that Julia chooses, 8 threads, is the worst, with 1 thread doing better. But using the number of physical cores, 4, is best.

So there were a lot of issues on Discourse and Slack #gripes where essentially "setting BLAS threads to 1 is better than the default!", but it looks like it's because the default should be the number of physical and not logical threads. I am actually very surprised it's not set that way, and so I was wondering why, and also where this default is set (I couldn't find it).

ChrisRackauckas:
[...]

PR #34728, opened 2020-02-11
Explain boundry cases for set_num_threads

oxinabox:
I wanted to know how to set number of BLAS threads to "All of them"
Looked at OpenBLAS source.
Answer is: `set_num_threads(typemax(Int))`

https://github.com/xianyi/OpenBLAS/blob/a387a23518b2794b7a9bdebdb7ca35943c50a912/driver/others/blas_server.c#L849-L917

### DIFF
--- a/stdlib/LinearAlgebra/src/blas.jl
+++ b/stdlib/LinearAlgebra/src/blas.jl
@@ -108,7 +108,7 @@ openblas_get_config() = strip(unsafe_string(ccall((@blasfunc(openblas_get_config
     set_num_threads(n)
 
 Set the number of threads the BLAS library should use.
-If n <= 0 then will make no change.
+If `n <= 0` then the number of threads will not be changed.
 If n > number of CPU threads, then caps at the number of CPU threads.
 """
 function set_num_threads(n::Integer)
--- a/stdlib/LinearAlgebra/src/blas.jl
+++ b/stdlib/LinearAlgebra/src/blas.jl
@@ -109,7 +109,7 @@ openblas_get_config() = strip(unsafe_string(ccall((@blasfunc(openblas_get_config
 
 Set the number of threads the BLAS library should use.
 If `n <= 0` then the number of threads will not be changed.
-If `n`  is greater than number of CPU threads, then the nuumber of BLAS threads wiol be set to the number of CPU threads.
+If `n`  is greater than number of CPU threads, then the number of BLAS threads will be set to the number of CPU threads.
 """
 function set_num_threads(n::Integer)
     blas = vendor()
--- a/stdlib/LinearAlgebra/src/blas.jl
+++ b/stdlib/LinearAlgebra/src/blas.jl
@@ -109,7 +109,7 @@ openblas_get_config() = strip(unsafe_string(ccall((@blasfunc(openblas_get_config
 
 Set the number of threads the BLAS library should use.
 If `n <= 0` then the number of threads will not be changed.
-If n > number of CPU threads, then caps at the number of CPU threads.
+If `n`  is greater than number of CPU threads, then the nuumber of BLAS threads wiol be set to the number of CPU threads.
 """
 function set_num_threads(n::Integer)
     blas = vendor()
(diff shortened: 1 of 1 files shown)


fchorney (approved):

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -108,6 +108,8 @@ openblas_get_config() = strip(unsafe_string(ccall((@blasfunc(openblas_get_config
     set_num_threads(n)
 
 Set the number of threads the BLAS library should use.
+If n <= 0 then will make no change.
```
KristofferC: Code quote maybe? Is this the same for MKL? And maybe a bit more descriptive like "keep the current number of threads"?

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -108,6 +108,8 @@ openblas_get_config() = strip(unsafe_string(ccall((@blasfunc(openblas_get_config
...
+If n > number of CPU threads, then caps at the number of CPU threads.
```
KristofferC: Perhaps "set the number of BLAS threads to be equal to the number of CPU threads"? Caps is a maybe a bit difficult to understand and  should be "cap" if kept to be consistent with "set" and "make".

stdlib/LinearAlgebra/src/blas.jl
oxinabox:
```suggestion
If `n <= 0` then the number of threads will not be changed.
```

I am not sure where to check MKL

stdlib/LinearAlgebra/src/blas.jl
oxinabox: Oh that is much clearer.

```suggestion
If `n`  is greater than number of CPU threads, then the nuumber of BLAS threads wiol be set to the number of CPU threads.
```

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -108,6 +108,8 @@ openblas_get_config() = strip(unsafe_string(ccall((@blasfunc(openblas_get_config
     set_num_threads(n)
 
 Set the number of threads the BLAS library should use.
+If `n <= 0` then the number of threads will not be changed.
+If `n`  is greater than number of CPU threads, then the nuumber of BLAS threads wiol be set to the number of CPU threads.
```
giordano:
```suggestion
If `n`  is greater than number of CPU threads, then the nuumber of BLAS threads will be set to the number of CPU threads.
```

mbauman: Is this hardware threads?  I think it actually includes hyper threads, which kinda sucks.  Does it also happen with MKL and other BLAS's?  Should we just do this in Julia (clamping to hardware threads) before punting to the library?

oxinabox: I assume it is logical threads, (so yes, including hyperthreading.)

My understanding is that locking down to only physical threads isn't that good except on dedicated machines that only run 1 process at a time.
And such machines just disable hyperthreading in the BIOS.

I feel like the authors of OpenBLAS would have already put the thought into it.
And if there is a better thing to do, then we should push it upstream as a PR to OpenBLAS.


I still need to go digging in the MKL docs.

oxinabox:
It is unclear as to if MKL has same behavour for upper bound.
They do not mention it here
https://software.intel.com/en-us/mkl-developer-reference-fortran-mkl-set-num-threads

but they do say: _"The number specified is a hint, and Intel® MKL may actually use a smaller number."_

tkf:
Maybe experiment it with `mkl_get_max_threads`? https://software.intel.com/en-us/mkl-developer-reference-c-mkl-get-max-threads

oxinabox:
According to @KristofferC MKL caps to number of physical threads.
https://github.com/JuliaLang/julia/issues/33409#issuecomment-583518722

stdlib/LinearAlgebra/src/blas.jl
KristofferC: https://stackoverflow.com/questions/60237626/will-mkl-set-num-threads-upper-bound-to-the-number-of-cpu-threads/60298394#60298394

KristofferC:
I just answered that stack overflow question heh.

DilumAluthge:
We have moved the LinearAlgebra stdlib to an external repo: https://github.com/JuliaLang/LinearAlgebra.jl

@oxinabox If you think that this PR is still relevant, please open a new PR on the [LinearAlgebra.jl](https://github.com/JuliaLang/LinearAlgebra.jl) repo.

PR #34729, opened 2020-02-11
Views with reshaped range indices are linear and strided

mbauman:
This came up in a DiffEq task, wherein a matrix is stored as a subset of the parameters vector — and thus to be used, it must be reshaped/subsetted before doing a matmul.  This patch gives such a view access to the blassy goodness.  The exemplar case looks something like this:

```julia
p = rand(300)
M = view(p, reshape(2:241, 60, 4))
v = rand(4)
@benchmark $M*$v
```

Before, this took nearly 200ns.  With this patch we hit BLAS and are down to 100ns.  Interestingly, this introduces a bifurcation where it's possible for view to be linear but not strided... but fortunately we've been pretty strict on keeping those two attributes distinct even though they previously were 100% overlapping.

I made sure the SubArray tests passed with JULIA_TESTFULL and while I was there I testsetified the test/subarray.jl file (as a separate and independent commit).

cc: @ChrisRackauckas.


timholy (approved):
This looks great. TBH I only looked carefully at the first commit, I assume that has all the functional changes? Thanks also for the cleanup of the subarray tests, I noticed several nice improvements there.

ChrisRackauckas:
bump?

ChrisRackauckas:
Bump again? I'd like to see this in v1.5 if possible.

ViralBShah:
@mbauman Can we get this in? Just have 4 more days!

PR #34732, opened 2020-02-12
Remove Iterators.reverse(::Tuple) specialization

tkf:
In https://github.com/JuliaLang/julia/pull/34691#issuecomment-584912856, @JeffBezanson was pointing out that it is desirable to have good support for large tuples:

> Large tuples do exist, and we'd like to support them better if possible. So it's really not obvious to me that `zip` should be eager for tuples.

_If_ we follow this principle, I think it's ideal if none of the iterator transformation APIs from `Iterators` special-case `Tuple`s.  So I suggest to revert #33235 (making `Iterators.reverse(::Tuple)` eager) and solve the original problem differently.  I think it's a safe option anyway as it makes this part of the API identical to the latest released version (1.3).

(Side note: `pairs` already supports `NamedTuple` and `Tuple` for some times so I guess it's not desirable to remove it.)

Requesting a tag: `backport 1.4`

### DIFF
--- a/base/iterators.jl
+++ b/base/iterators.jl
@@ -98,7 +98,6 @@ reverse(G::Generator) = Generator(G.f, reverse(G.iter))
 reverse(r::Reverse) = r.itr
 reverse(x::Union{Number,AbstractChar}) = x
 reverse(p::Pair) = Base.reverse(p) # copying pairs is cheap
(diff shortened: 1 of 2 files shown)


JeffBezanson:
> (Side note: `pairs` already supports `NamedTuple` and `Tuple` for some times so I guess it's not desirable to remove it.)

Yes, but it still returns a lazy iterator in those cases so it's ok.

Issue #33260, opened 2019-09-13
Can/should we take repr seriously?


tkf:
Currently, `repr` in Julia is maybe evaluatable or parseable or (often) just a string representation only consumable by humans.  However, it could be much useful if it is guaranteed to be evaluatable.  For example, you can find a lot of examples in `base/loading.jl` which is used as simple IPC mechanism:

https://github.com/JuliaLang/julia/blob/9a8b2fd72b675bb8a5bf0322943ee9451787b86a/base/loading.jl#L1148-L1151

In case of `base/loading.jl` it is OK to use `repr` because the types used there are known to have evaluatable `repr`.  However, some basic types like `PkgId` cannot be used in this way:

```julia
julia> repr(Base.PkgId(InteractiveUtils))
"InteractiveUtils [b77e0a4c-d291-57a0-90e8-8db25a27a240]"
```

Another concrete example I have in mind where `repr` can be useful is to create a plot "spec" using [DataVoyager.jl](https://github.com/queryverse/DataVoyager.jl) and dump it into a runnable Julia script.  Currently, using it needs a rather convoluted invocation like [`show(IOContext(stdout, :compact=>false), "text/plain", spec)`](https://github.com/queryverse/VegaLite.jl/pull/162) because `repr` is not a reliable entry point for evaluatable code.

We have `Serialization` module for serializing complex Julia objects.  Unfortunately, it cannot be used for communicating with different Julia versions.  For complex serializations it is probably a good idea to use something like JSON instead of Julia code.  However, I think the above examples are good enough motivation to have `repr` for simple cases.

It was suggested to add a flag like `parseable` (e.g., https://github.com/JuliaLang/julia/pull/33178#issuecomment-530294366, https://github.com/JuliaLang/julia/issues/30683#issuecomment-531223984) via the `IOContext` mechanism.  I have a concern about this direction because it is practically impossible to enforce parsability this way (a brief discussion with @JeffBezanson in https://github.com/JuliaLang/julia/pull/30757#issuecomment-456936172).  I think the overloading API should be clear that the implementer _has_ to make it evaluatable.  This can be done by introducing overloading API such as `show(::IO, ::MIME"application/julia", obj)` or `repr(::IO, obj)`.

Questions:

* Do we need an API to produce Julia code that is (very likely to be) evaluatable?  (100% guarantee is impossible when using different versions of libraries and/or Julia)
[...]

PR #34733, opened 2020-02-12
[RFC/WIP] Handle :compact context in 3-arg show for arrays

### LINKED ISSUE #34659: Wrong display for arrays of empty arrays

tkf:
Edit: now this PR fixes only `permutedims([Int[], Int[]])` case https://github.com/JuliaLang/julia/pull/34733#issuecomment-588513011

---

<strike>Before:

```
julia> [Int[]]
 0-element Array{Int64,1}
```

After:

```
julia> [Int[]]
 []
```
</strike>

<strike>I'm sending this as a WIP</strike> I was meant to send this as a WIP PR before fixing the tests as I don't know if the core devs think this is the way to go.

fix (a related bug of) #34659

### DIFF
--- a/base/arrayshow.jl
+++ b/base/arrayshow.jl
@@ -334,7 +334,7 @@ function show(io::IO, ::MIME"text/plain", X::AbstractArray)
     show_circular(io, X) && return
 
     # 1) compute new IOContext
-    if !haskey(io, :compact)
+    if !haskey(io, :compact) && length(axes(X, 2)) > 1
         io = IOContext(io, :compact => true)
     end
     if get(io, :limit, false) && eltype(X) === Method
--- a/base/arrayshow.jl
+++ b/base/arrayshow.jl
@@ -334,7 +334,7 @@ function show(io::IO, ::MIME"text/plain", X::AbstractArray)
     show_circular(io, X) && return
 
     # 1) compute new IOContext
-    if !haskey(io, :compact) && length(axes(X, 2)) > 1
+    if !haskey(io, :compact)
         io = IOContext(io, :compact => true)
     end
     if get(io, :limit, false) && eltype(X) === Method
--- a/base/arrayshow.jl
+++ b/base/arrayshow.jl
@@ -320,6 +320,13 @@ print_array(io::IO, X::AbstractArray) = show_nd(io, X, print_matrix, true)
 # typeinfo aware
 # implements: show(io::IO, ::MIME"text/plain", X::AbstractArray)
 function show(io::IO, ::MIME"text/plain", X::AbstractArray)
+    # -1) handle arrays-in-array (recursive call)
+    if get(io, :compact, false) === true
+        # Using `show` of method with the most abstract type to avoid infinite recursion.
+        invoke(show, Tuple{IO,AbstractArray}, io, X)
+        return
+    end
+
     # 0) show summary before setting :compact
     summary(io, X)
     isempty(X) && return


base/arrayshow.jl
```diff
@@ -320,14 +320,21 @@ print_array(io::IO, X::AbstractArray) = show_nd(io, X, print_matrix, true)
...
+    end
+
     # 0) show summary before setting :compact
     summary(io, X)
     isempty(X) && return
     print(io, ":")
     show_circular(io, X) && return
 
     # 1) compute new IOContext
-    if !haskey(io, :compact) && length(axes(X, 2)) > 1
```
tkf: I'm not sure if people are happy with removing `length(axes(X, 2)) > 1`.

(BTW I think this is a shortcoming of the way `:compact` is used.  It's not possible to declare compactness of the width and height separately.  This is why I've been suggesting to use `:displaysize` instead of `:compact` https://github.com/JuliaLang/julia/issues/33260#issuecomment-531991052)

base/arrayshow.jl
```diff
@@ -320,14 +320,21 @@ print_array(io::IO, X::AbstractArray) = show_nd(io, X, print_matrix, true)
 # typeinfo aware
 # implements: show(io::IO, ::MIME"text/plain", X::AbstractArray)
 function show(io::IO, ::MIME"text/plain", X::AbstractArray)
+    # -1) handle arrays-in-array (recursive call)
+    if get(io, :compact, false) === true
+        # Using `show` of method with the most abstract type to avoid infinite recursion.
```
JeffBezanson: Is this really necessary? I don't think it's common for 2-arg show methods to call 3-arg ones.

base/arrayshow.jl
JeffBezanson: This is a separate issue and will probably be controversial, so I wouldn't change this for now.

tkf:
@JeffBezanson Note that _both_ changes are required if you want to fix the bug #34659 this way. The problem is that 3-arg `show` is used for the inner array when calling 3-arg `show` for the outer array `show(io, "text/plain", [Int[]])` due to

https://github.com/JuliaLang/julia/blob/b0d1c1abb69981e80640fb1501ce88d597b9cf74/base/arrayshow.jl#L105-L111

That is to say, since 3-arg show of the inner array `show(io, "text/plain", Int[])` does not print multiple lines, we don't use the `if occursin('\n', sx)` branch.  This is why we need `if get(io, :compact, false) === true` https://github.com/JuliaLang/julia/pull/34733#discussion_r381578885.  However, for the inner array to be able to check that it is invoked via 3-arg `show`, we need to set `:compat` always even with vectors https://github.com/JuliaLang/julia/pull/34733#discussion_r378015856.

Of course, there may be another solution for fixing #34659. Please close this PR if there is a better way.

JeffBezanson:
You're right; it would only fix this case:
```
julia> permutedims([Int[], Int[]])
1×2 Array{Array{Int64,1},2}:
 0-element Array{Int64,1}  0-element Array{Int64,1}
```
but I think that's worth doing.
I don't see any way to fix the 1-column case except adding a `showelement` function and overloading it for arrays.

tkf:
OK, I put `length(axes(X, 2)) > 1` back. Let's see how many tests fail...

PR #34734, opened 2020-02-12
Recommend overloading 3-arg show; explain :compact handling

tkf:
As discussed in https://github.com/JuliaLang/julia/issues/30901#issuecomment-573450132 and implemented in #34387, it is now recommended to overload 3-arg `show` to customize the text output for humans.  This PR updates the docstring to recommend 3-arg `show` instead of 2-arg `show` and clarifies the role of `:compact` context for printing containers.

### DIFF
--- a/base/multimedia.jl
+++ b/base/multimedia.jl
@@ -94,9 +94,10 @@ your images to be displayed on any PNG-capable `AbstractDisplay` (such as IJulia
 to `import Base.show` in order to add new methods to the built-in Julia function
 `show`.
 
-The default MIME type is `MIME"text/plain"`. There is a fallback definition for `text/plain`
-output that calls `show` with 2 arguments. Therefore, this case should be handled by
-defining a 2-argument `show(io::IO, x::MyType)` method.
+To customize how an object of type `T` is shown in text-based outputs, it is highly recommended
+to overload `show(::IO, ::MIME"text/plain", ::T)` rather than [`show(::IO, ::T)`](@ref show).
+Container types should call `show(io, MIME"text/plain"(), x)` for elements `x`
+instead of `show(io, x)` with [`IOContext`](@ref) of `:compact => true`.
 
 Technically, the `MIME"mime"` macro defines a singleton type for the given `mime` string,
 which allows us to exploit Julia's dispatch mechanisms in determining how to display objects
--- a/base/multimedia.jl
+++ b/base/multimedia.jl
@@ -94,10 +94,12 @@ your images to be displayed on any PNG-capable `AbstractDisplay` (such as IJulia
 to `import Base.show` in order to add new methods to the built-in Julia function
 `show`.
 
-To customize how an object of type `T` is shown in text-based outputs, it is highly recommended
-to overload `show(::IO, ::MIME"text/plain", ::T)` rather than [`show(::IO, ::T)`](@ref show).
-Container types should call `show(io, MIME"text/plain"(), x)` for elements `x`
-instead of `show(io, x)` with [`IOContext`](@ref) of `:compact => true`.
+To customize human-readable text output for objects of type `T`, it is highly recommended
+to overload `show(::IO, ::MIME"text/plain", ::T)` rather than [`show(::IO, ::T)`](@ref show)
+which is primarily used for parseable text output.  When implementing the method
+`show(::IO, ::MIME"text/plain", ::T)` for container types, it should call
+`show(io, MIME"text/plain"(), x)` for elements `x` instead of `show(io, x)` with
+[`IOContext`](@ref) of `:compact => true` from its implementation
 
 Technically, the `MIME"mime"` macro defines a singleton type for the given `mime` string,
 which allows us to exploit Julia's dispatch mechanisms in determining how to display objects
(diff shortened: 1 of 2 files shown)


base/multimedia.jl
```diff
@@ -94,9 +94,10 @@ your images to be displayed on any PNG-capable `AbstractDisplay` (such as IJulia
 to `import Base.show` in order to add new methods to the built-in Julia function
 `show`.
 
-The default MIME type is `MIME"text/plain"`. There is a fallback definition for `text/plain`
-output that calls `show` with 2 arguments. Therefore, this case should be handled by
-defining a 2-argument `show(io::IO, x::MyType)` method.
+To customize how an object of type `T` is shown in text-based outputs, it is highly recommended
+to overload `show(::IO, ::MIME"text/plain", ::T)` rather than [`show(::IO, ::T)`](@ref show).
+Container types should call `show(io, MIME"text/plain"(), x)` for elements `x`
```
JeffBezanson: This only applies to the 3-argument `show` method for the container type.

Both here and in the other doc string I think we should use expressions such as "human readable" vs. "parseable" to clarify the difference. As it is, it almost sounds like 2-argument show should never be used.

tkf: Thanks for reviewing this. Does the new commit address your concern?

base/multimedia.jl
```diff
@@ -94,9 +94,12 @@ your images to be displayed on any PNG-capable `AbstractDisplay` (such as IJulia
 to `import Base.show` in order to add new methods to the built-in Julia function
 `show`.
 
-The default MIME type is `MIME"text/plain"`. There is a fallback definition for `text/plain`
```
nalimilan: Why remove the information about the fallback? That's quite useful to understand how the whole system works.

base/show.jl
```diff
@@ -357,7 +357,13 @@ Write an informative text representation of a value to the current output stream
 should overload `show(io::IO, x)` where the first argument is a stream. The representation used
 by `show` generally includes Julia-specific formatting and type information.
 
-[`repr`](@ref) returns the output of `show` as a string.
+[`repr`](@ref) returns the output of `show` as a string.  Therefore, `show(::IO, ::T)` is
+expected to print parseable Julia code.
+
+To customize human-readable text output for objects of type `T`, it is highly recommended
+to overload `show(::IO, ::MIME"text/plain", ::T)` rather than `show(::IO, ::T)`.  To customize
+how objects of type `T` stored in containers are shown, check `:compact` [`IOContext`](@ref)
```
nalimilan:
```suggestion
how objects of type `T` stored in containers are shown, check the `:compact` [`IOContext`](@ref) property
```

base/show.jl
```diff
@@ -357,7 +357,13 @@ Write an informative text representation of a value to the current output stream
 should overload `show(io::IO, x)` where the first argument is a stream. The representation used
 by `show` generally includes Julia-specific formatting and type information.
```
nalimilan: Mention "parseable" here, rather than just below?

base/multimedia.jl
```diff
@@ -94,9 +94,12 @@ your images to be displayed on any PNG-capable `AbstractDisplay` (such as IJulia
...
-output that calls `show` with 2 arguments. Therefore, this case should be handled by
-defining a 2-argument `show(io::IO, x::MyType)` method.
+To customize human-readable text output for objects of type `T`, it is highly recommended
+to overload `show(::IO, ::MIME"text/plain", ::T)` rather than [`show(::IO, ::T)`](@ref show)
```
nalimilan: I find it weird to say here that you shouldn't use the two-arg `show`, since we're in the docstring for the three-argument method. Anyway that's a repetition from the warning already printed in the docstring for the two-arg `show`, which will appear right before this one.

nalimilan (commented):
Thanks for doing this. Note that [the manual](https://docs.julialang.org/en/latest/manual/types/#man-custom-pretty-printing-1) should also be changed, especially since it recommends implementing a two-argument `show` to print a representation which, while parseable, doesn't produce an object of the original type...

kimikage:
I think this changes of docs would be helpful, but after all, I don't understand the best practice on v1.5.
For example:
```julia
julia> VERSION
v"1.5.0-DEV.646"

julia> struct MyType
            always_single_line::Bool
       end

julia> function Base.show(io::IO, ::MIME"text/plain", x::MyType)
          if get(io, :compact, false)
               print(io, x.always_single_line ? "single line" : "truncated", " w/ :compact")
           else
               print(io, x.always_single_line ? "SINGLE LINE BUT VERY VERBOSE" : "MULTI\nLINES", " w/o :compact")
           end
       end

julia> [MyType(false) MyType(true)]
1×2 Array{MyType,2}:
 truncated w/ :compact  single line w/ :compact

julia> [MyType(false), MyType(true)] # this uses the two-arg `show`
2-element Array{MyType,1}:
 MyType(false)
 SINGLE LINE BUT VERY VERBOSE w/o :compact
```
I think we still need to overload the two-arg `show` for now.
```julia
Base.show(io::IO, x::MyType) = show(IOContext(io, :compact=>true), MIME("text/plain"), x)

julia> [MyType(false), MyType(true)] # the customization is ok, but the result is somewhat inconsistent.
2-element Array{MyType,1}:
 truncated w/ :compact
 SINGLE LINE BUT VERY VERBOSE w/o :compact

```

PR #34736, opened 2020-02-12
Added tests for hashing secret buffers

ShriprajwalK:


test/secretbuffer.jl
```diff
@@ -80,4 +80,9 @@ using Test
         @test read(sb, String) == "\xff\xff\xff"
         shred!(sb)
     end
+    @testset "hashing secret buffers" begin
+        sb1 = SecretBuffer("hello")
+        sb2 = SecretBuffer("juliaisawesome")
+        @test hash(sb1,convert(UInt,5)) == hash(sb2,convert(UInt,5))
```
StefanKarpinski: It would be more idiomatic to write `UInt(5)` than `convert(UInt,5)` here.

StefanKarpinski:
I see from the comment in the code that it's intentional that SecretBuffers all hash the same, but it seems like a strange choice to me that's likely to cause problems. cc @mbauman, @omus: since you're the two people "blamed" for this file. Should we discuss the hashing of secret buffers a bit more before enshrining this in the tests? Or should we just test the current intended behavior and change the tests if we decide to change this?

omus:
I suspect these hash the same as a stop gap measure. We'd want to use a cryptographic hash function for secret buffers to avoid the possibility of inverting the hash.

mbauman:
Right — but do we have a secure cryptographic hash function built in to base?  I suppose we could port one.  

I didn't think this would cause serious issues since I didn't think it likely to use one of these guys in a dictionary and that hashes fall back to equality upon collision.  Am I missing something?

StefanKarpinski:
We have SHA as a stdlib and it's not going anywhere since it's heavily used in Pkg. If we don't think that SecretBuffers should be put into sets and dictionaries, wouldn't it be better to error?

test/secretbuffer.jl
vtjnash:
```suggestion
        @test hash(sb1, UInt(5)) === hash(sb2, UInt(5))
```

vtjnash (approved):

PR #34741, opened 2020-02-12
Add compact 3-arg show for Irrational

omus:
Due to the changes in #34387 any call to 3-arg show for `Irrational` (namely π) would result in the symbol and a truncated value being displayed. This PR allows for a compact option for the 3-arg show where only the symbol is shown.

Similar to: https://github.com/JuliaLang/julia/pull/34730


omus:
I'll note I just recently noticed this otherwise I would have opened this at the same time as https://github.com/JuliaLang/julia/pull/34730

StefanKarpinski (approved):

PR #34743, opened 2020-02-12
various improvements to subtyping

vtjnash:
Improves some fast paths, and fixes some intersection bugs (by visual comparison against the more-tested obvious_subtype, and by reading a comment that explains why the code it describes is not quite valid, haha)

### DIFF
--- a/src/subtype.c
+++ b/src/subtype.c
@@ -3232,6 +3232,37 @@ jl_value_t *switch_union_tuple(jl_value_t *a, jl_value_t *b)
     return ans;
 }
 
+// `a` might have a non-empty intersection with some concrete type b even if !(a<:b) and !(b<:a)
+// For example a=`Tuple{Type{<:Vector}}` and b=`Tuple{DataType}`
+int might_intersect_concrete(jl_value_t *a)
+{
+    if (jl_is_unionall(a))
+        a = jl_unwrap_unionall(a);
+    if (jl_is_typevar(a))
+        return 1; // (maybe)
+    if (jl_is_uniontype(a))
+        return might_intersect_concrete(((jl_uniontype_t*)a)->a) ||
+               might_intersect_concrete(((jl_uniontype_t*)a)->b);
+    if (jl_is_vararg_type(a))
+        return might_intersect_concrete(jl_tparam0(a));
+    if (jl_is_type_type(a))
+        return 1;
+    if (jl_is_datatype(a)) {
+        int tpl = jl_is_tuple_type(a);
+        int i, n = jl_nparams(a);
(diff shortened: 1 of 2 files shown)


src/subtype.c
```diff
@@ -304,11 +304,13 @@ static int obviously_disjoint(jl_value_t *a, jl_value_t *b, int specificity)
         return 0;
     if (specificity && a == (jl_value_t*)jl_typeofbottom_type)
         return 0;
-    if (jl_is_concrete_type(a) && jl_is_concrete_type(b) &&
-        // TODO: remove these 2 lines if and when Tuple{Union{}} === Union{}
-        (((jl_datatype_t*)a)->name != jl_tuple_typename ||
-         ((jl_datatype_t*)b)->name != jl_tuple_typename))
-        return 1;
+    // TODO: this would be a nice fast-path to have, unfortuanately,
+    //       datatype allocation fails to correctly cons them
```
JeffBezanson:
```suggestion
    //       datatype allocation fails to correctly hash-cons them
```

src/subtype.c
```diff
@@ -355,25 +357,24 @@ static int obviously_disjoint(jl_value_t *a, jl_value_t *b, int specificity)
...
         size_t i;
-        for(i=0; i < np; i++) {
-            jl_value_t *ai = jl_tparam(ad,i);
-            jl_value_t *bi = jl_tparam(bd,i);
+        for (i = 0; i < np; i++) {
+            jl_value_t *ai = jl_tparam(ad, i);
+            jl_value_t *bi = jl_tparam(bd, i);
             if (jl_is_typevar(ai) || jl_is_typevar(bi))
-                continue;
+                continue; // Union{} is always in this intersection
```
JeffBezanson: I don't think this is true, e.g. if there is a non-`Union{}` lower bound. We just skip these because vars are difficult to handle.

vtjnash:
```suggestion
                continue; // it's possible that Union{} is in this intersection
```

src/subtype.c
```diff
@@ -3226,6 +3254,35 @@ jl_value_t *switch_union_tuple(jl_value_t *a, jl_value_t *b)
     return ans;
 }
 
+int does_subtypes_concrete(jl_value_t *a)
```
JeffBezanson:
```suggestion
// `a` might have a non-empty intersection with some concrete type b even if !(a<:b) and !(b<:a)
// For example a=`Tuple{Type{<:Vector}}` and b=`Tuple{DataType}`
int might_intersect_concrete(jl_value_t *a)
```

PR #34745, opened 2020-02-12
gf: optimize jl_matching_methods for dispatch tuples

vtjnash:
When the input type is a dispatch tuple, we know that a subtype lookup
is sufficient, so we can do a slightly more direct lookup and avoid
considering `Union{}`-declared methods as possible results.

This also fixes an issue with the special `jl_matching_methods` query
being done by `jl_insert_backedges`: we would set the `done` flag when a
method "fully covered" the result, even if that method was deleted.
Rather than continue to abuse the meaning of `world == 0` (which also
has performance implications), we add a specific flag for this case.

### DIFF
--- a/src/gf.c
+++ b/src/gf.c
@@ -90,8 +90,8 @@ static int8_t jl_cachearg_offset(jl_methtable_t *mt)
 JL_DLLEXPORT jl_method_instance_t *jl_specializations_get_linfo(jl_method_t *m JL_PROPAGATES_ROOT, jl_value_t *type, jl_svec_t *sparams)
 {
     JL_LOCK(&m->writelock);
-    jl_typemap_entry_t *sf =
-        jl_typemap_assoc_by_type(m->specializations, type, NULL, /*subtype*/0, /*offs*/0, 1, /*max_world_mask*/0);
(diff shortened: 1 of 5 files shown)


src/gf.c
```diff
@@ -1296,10 +1300,10 @@ static jl_value_t *check_ambiguous_matches(jl_typemap_t *defs, jl_typemap_entry_
...
             va = NULL;
     }
-    struct ambiguous_matches_env env;
-    env.match.fptr = fptr;
-    env.match.type = (jl_value_t*)type;
-    env.match.va = va;
+    struct ambiguous_matches_env env = {.match = {.fptr = fptr, .type = (jl_value_t*)type, .va = va}};
+    //env.match.fptr = fptr;
+    //env.match.type = (jl_value_t*)type;
+    //env.match.va = va;
```
JeffBezanson: delete?

vtjnash: Ah, yes, I should do that cleanup. How do we feel about starting to use the C99 initialization syntax? (this it not valid C++ syntax, though something similar will be added in C++20)

JeffBezanson (approved):

PR #34754, opened 2020-02-13
ccall: avoid some problematic symbol mangling

vtjnash:
These fix some build issues on some platforms.


staticfloat:
This should fix https://github.com/JuliaLang/julia/issues/34680

KristofferC:
I've added a test for this in https://github.com/JuliaLang/PackageCompiler.jl/pull/341 and can confirm that it fails without this and passes with.

PR #34756, opened 2020-02-13
Changed error message for end-of-line colons

naveenarun:
I'm a Python user who's just starting to learn Julia, and I was playing around with function definitions and wrote a function with the following syntax:

```julia
function test_function(inputs):
    ...
end
```
I then got the following error message:

`ERROR: syntax: space not allowed after ":" used for quoting`

I was confused because I wasn't aware yet that colons were used for quoting, and ended up trying to search for a space character in my code (after all, I had a newline - not a space - after the colon). Wasted quite a bit of time trying to fix it.

This PR just changes the error message to the following, meant to be nudge Python users in the right direction:

`ERROR: syntax: space not allowed after ":" used for quoting. (Did you put ":" after a function/if/while/for statement?)`

### DIFF
--- a/src/julia-parser.scm
+++ b/src/julia-parser.scm
@@ -1219,10 +1219,11 @@
                     ((eqv? (peek-token s) ':)
                      (begin
                        (take-token s)
-                       (if (or (eqv? (peek-token s) #\newline)
-                               (ts:space? s)) ;; uses side effect of previous peek-token
-                           (error "space not allowed after \":\" used for quoting"))
-                       `(|.| ,ex (quote ,(parse-atom s #f)))))
+                       (cond ((eqv? (peek-token s) #\newline)
+                              (error "newline not allowed after \":\" used for quoting"))
+                             ((ts:space? s) ;; uses side effect of previous peek-token
+                              (error "whitespace not allowed after \":\" used for quoting"))
+                             (else `(|.| ,ex (quote ,(parse-atom s #f)))))))
                     ((eq? (peek-token s) '$)
                      (take-token s)
                      (let ((dollarex (parse-atom s)))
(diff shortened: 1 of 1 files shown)


yuyichao:
[As recently shown](https://github.com/JuliaLang/julia/pull/34706), error message suggesting the wrong cause can be very confusing for people that hit it for other reasons. Having guess work in the error message is fine but I don't believe having EOL `:` after begining of block is the most likely cause of this error so this new message should only be printed in more specific cases.

naveenarun:
I did a google search for exact matches of the error messages and got the following breakdown for 19 results:

- Proposal for pull request re. ternary operators: 3 results
- Use of ternary operator: 1 result (https://discourse.julialang.org/t/ternary-operator-confused-with-quoting/24536)
- Colon after function definition: 15 results

Based on this, I removed the parts of the error message talking about if/while/for statements. I think that this change, at least for the time being, would not majorly confuse people debugging ternary operator errors.

yuyichao:
While it's nice that you did the research by searching for the error message, the result, unfortunately doesn't mean anything. It's by construction biased against people whose problem is solved by the current error message. FWIW, I think mentioning while/if/for is fine, removing them makes it worse.

naveenarun:
You're right. I'll try to change this so that the message only comes up in more specific cases.

yuyichao:
Fwiw, I don't think you need to be too specific. Just end of line should be good enough and from a quick look at the code that should not be hard....

naveenarun:
Alright, I changed the error message from

`ERROR: syntax: space not allowed after ":" used for quoting`

to 

`ERROR: syntax: space or newline not allowed after ":" used for quoting`

in the two places where it's relevant.

yuyichao:
I do think this is a strict improvement. You could probably also split the conditon and give your python related error message if the first condition is true.

src/julia-parser.scm
```diff
@@ -1220,7 +1220,7 @@
                        (take-token s)
                        (if (or (eqv? (peek-token s) #\newline)
                                (ts:space? s)) ;; uses side effect of previous peek-token
-                           (error "space not allowed after \":\" used for quoting"))
+                           (error "space or newline not allowed after \":\" used for quoting"))
```
vtjnash:
```suggestion
                       (cond ((eqv? (peek-token s) #\newline)
                              (error "newline not allowed after \":\" used for quoting"))
                             ((ts:space? s) ;; uses side effect of previous peek-token
                              (error "whitespace not allowed after \":\" used for quoting"))
                             ((else
```

Keno:
Rebased and applied @vtjnash's suggestion to both instances.

Issue #34713, opened 2020-02-09
`using a` on help mode gives BoundsError


Hieronimo:
`using a` on help mode in repl gives error but `using 1` is still working and suggests for using of `using` in search (also for `import`). Is it right to give an error like this? I think, it should acts like the latter.

```julia
help?> using a
ERROR: LoadError: BoundsError: attempt to access 1-element Array{Any,1} at index [2]


help?> using 1
search: using SubString include_string unsafe_string unescape_string SubstitutionString

Couldn't find using 1
Perhaps you meant using or asin
  No documentation found.

  Binding using 1 does not exist.


help?> import a
ERROR: LoadError: BoundsError: attempt to access 1-element Array{Any,1} at index [2]


help?> import 1
search: import

Couldn't find import 1
Perhaps you meant import or issorted
  No documentation found.

  Binding import 1 does not exist.

```
_By_
**Julia Version 1.3.0**
**Commit 46ce4d7933 (2019-11-26 06:09 UTC)**

strickek:
I think help modus should be supportive,  so errors should be avoided by suggestions.

PR #34757, opened 2020-02-14
change helpdoc so using a works

ssikdar1:
For issue #34713 

https://github.com/JuliaLang/julia/blob/a211abcdfacc05cb93c15774a59ce8961c16dac4/stdlib/REPL/src/docview.jl#L34-L47

Here for `using 1`, `Meta.parse`  determines x is:
```
x = :($(Expr(:error, "invalid \"using\" statement: expected identifier")))
typeof(x) = Expr
```
as a result `expr` is 
```
expr = Symbol("using 1")
typeof(expr) = Symbol
```

while for `using a` , `Meta.parse` can parse the string without an error:
```
x = :(using a)
typeof(x) = Expr
``` 
and `expr` is 

```
expr = :(using a)
typeof(expr) = Expr
```
The difference in types leads to two different code paths. This is why one has a useful documentation, the other an exception.

This change quickly checks the input line. If it splits and the first token contains `using` just make `expr` a `Symbol` ( same type used when `using 1` is called).

With this both `using 1` and `using a` can have similar outputs:

```
help?> using a
search: using SubString include_string unsafe_string unescape_string SubstitutionString

Couldn't find using a
Perhaps you meant using or asin
  No documentation found.

  Binding using a does not exist.

help?> using 1
search: using SubString include_string unsafe_string unescape_string SubstitutionString

Couldn't find using 1
Perhaps you meant using or asin
  No documentation found.

  Binding using 1 does not exist.

julia> 
```

### DIFF
--- a/stdlib/REPL/src/docview.jl
+++ b/stdlib/REPL/src/docview.jl
@@ -37,6 +37,8 @@ function _helpmode(io::IO, line::AbstractString)
             # Docs for keywords must be treated separately since trying to parse a single
             # keyword such as `function` would throw a parse error due to the missing `end`.
             Symbol(line)
+        elseif isexpr(x, (:using, :import))
+            x.head
         else
             # Retrieving docs for macros requires us to make a distinction between the text
             # `@macroname` and `@macroname()`. These both parse the same, but are used by
--- a/stdlib/REPL/test/repl.jl
+++ b/stdlib/REPL/test/repl.jl
@@ -1027,9 +1027,11 @@ for (line, expr) in Pair[
     "while       " => :while, # keyword, trailing spaces should be stripped.
     "0"            => 0,
     "\"...\""      => "...",
-    "r\"...\""     => Expr(:macrocall, Symbol("@r_str"), LineNumberNode(1, :none), "...")
+    "r\"...\""     => Expr(:macrocall, Symbol("@r_str"), LineNumberNode(1, :none), "..."),
+    "using Foo"    => :using,
+    "import Foo"   => :import,
     ]
-    #@test REPL._helpmode(line) == Expr(:macrocall, Expr(:., Expr(:., :Base, QuoteNode(:Docs)), QuoteNode(Symbol("@repl"))), LineNumberNode(119, doc_util_path), stdout, expr)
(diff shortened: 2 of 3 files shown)


Hieronimo:
It should also covers the `import a`.

ssikdar1:
changed so  `import` and `using` work in e83571d 

```
help?> import 1
search: import

Couldn't find import 1
Perhaps you meant import or issorted
  No documentation found.

  Binding import 1 does not exist.

help?> import a
search: import

Couldn't find import a
Perhaps you meant import or issorted
  No documentation found.

  Binding import a does not exist.

help?> using a
search: using SubString include_string unsafe_string unescape_string SubstitutionString

Couldn't find using a
Perhaps you meant using or asin
  No documentation found.

  Binding using a does not exist.

help?> using a
search: using SubString include_string unsafe_string unescape_string SubstitutionString

Couldn't find using a
Perhaps you meant using or asin
  No documentation found.

  Binding using a does not exist.
```

fredrikekre:
I pushed a slightly modified version that actually shows the help for `using`/`import` instead.

ssikdar1:
@fredrikekre not sure if its crucial or not but I think `using a` and `using 1` will have different output:

```
help?> using 1
search: using SubString include_string unsafe_string unescape_string SubstitutionString

Couldn't find using 1
Perhaps you meant using or asin
  No documentation found.

  Binding using 1 does not exist.

help?> using a
search: using SubString include_string unsafe_string unescape_string SubstitutionString unsigned Unsigned occursin unsafe_pointer_to_objref

  using

  using Foo will load the module or package Foo and make its exported names available for direct use. Names can also be used via dot syntax (e.g. Foo.foo to access the name foo), whether they are
  exported or not. See the manual section about modules for details.

help?> import 1
search: import

Couldn't find import 1
Perhaps you meant import or issorted
  No documentation found.

  Binding import 1 does not exist.

help?> import a
search: import

  import

  import Foo will load the module or package Foo. Names from the imported Foo module can be accessed with dot syntax (e.g. Foo.foo to access the name foo). See the manual section about modules for
  details.

julia> 
```

PR #34762, opened 2020-02-14
Added test for bytes available with correct formatting

ShriprajwalK:


ShriprajwalK:
I've done a few changes to make the code better. Good to go from my side!

PR #34763, opened 2020-02-14
make alignment deal with float32 like float64

### LINKED ISSUE #34750: Wrong indentation when printing arrays of ComplexF32
Reproducer
```
julia> [ 1f-7 + 2.0f-11im, 2.0f-5 + 4f0im]
2-element Array{Complex{Float32},1}:
 1.0f-7 + 2.0f-11im    
       2.0f-5 + 4.0f0im
```

Compare that to 
```
julia> [ 1e-7 + 2.0e-11im, 2.0e-5 + 4e0im]
2-element Array{Complex{Float64},1}:
 1.0e-7 + 2.0e-11im
 2.0e-5 + 4.0im  
```

Tested in v1.1, v1.4 and master under macOS

birm:
closes #34750
The bug reported above is due to alignment accepting e but not f as an exponentiation, thus causing a print error for things of type float32
(I'm unsure if 2061 also needed to be changed this way, but I did anyway. I'm happy to undo this if this is not correct)

### DIFF
--- a/base/show.jl
+++ b/base/show.jl
@@ -2058,13 +2058,13 @@ alignment(io::IO, x::Number) = (length(sprint(show, x, context=io, sizehint=0)),
 alignment(io::IO, x::Integer) = (length(sprint(show, x, context=io, sizehint=0)), 0)
 "`alignment(4.23)` yields (1,3) for `4` and `.23`"
 function alignment(io::IO, x::Real)
-    m = match(r"^(.*?)((?:[\.eE].*)?)$", sprint(show, x, context=io, sizehint=0))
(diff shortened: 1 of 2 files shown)


JeffBezanson:
Welcome, and thanks! It's amazing how long this bug has been around. Could you add a test case to test/show.jl? The change to line 2061 looks reasonable; is there an example of a case it fixes?

birm:
I've added a test for the complex case. Regarding line 2061, the cases I've tried so far, the use of e or f has not impacted alignment for real numbers, so I don't yet have a test case for that.

PR #34764, opened 2020-02-14
Added test for dumptype with a Concrete and Abstract type.

kczimm:
Add a test for InteractiveUtils dumptype function.


Keno:
This test is failing on CI on master.

JeffBezanson:
What is this function supposed to do? If it's supposed to find all subtypes of a type, it doesn't work. But I'm not sure since I've never used this function once.

kczimm:
Sorry, I am feeling bad for this PR. I was looking to start contributing so I wrote a test to increase coverage but maybe the function I tried to cover is no longer needed or desired?

JeffBezanson:
Don't worry, it's not really your fault. This was a rare kind of conflict that our CI setup can't catch since two distant pieces of code interacted (they each passed separately, but not when merged together).

The function isn't exported or documented though, so maybe we should remove it.

Issue #6080, opened 2014-03-07
deprecate `&x` syntax in ccall, replace with Ref


pygy:
``` Julia
|(::String) = 4
# | (generic function with 49 methods)

&(::String) = 6
# ERROR: syntax: invalid identifier name "::"

|(123,456)
#507

&(123,456)
# ERROR: unsupported or misplaced expression &
```

`Base.(:&)` works as intended.

JeffBezanson:
Now that an alternative is available, we need to decide whether to fully deprecate `&` syntax in `ccall`.

tkelman:
There's still a performance cost to using `Ref` in `ccall` right? Guess that needs fixing before `&` in `ccall` can be deprecated without performance regressions.

StefanKarpinski:
There was a change to use Ref in the LAPACK wrappers and it was way too slow. I believe that performance issue is the blocker from completely removing the use of `&` in `ccall`.

JeffBezanson:
https://github.com/JuliaLang/julia/pull/11402

vtjnash:
need to deprecate `&` in ccall first, which is probably mostly pending https://github.com/JuliaLang/julia/pull/18632 right now (although that's not required for making this efficient – only need the easy case of https://github.com/JuliaLang/julia/pull/8134)

JeffBezanson:
#22684 and #22734 will enable this.

yuyichao:
Ah, so here's the issue..... Somehow I couldn't find this when openning the two prs....

musm:
https://github.com/JuliaLang/julia/pull/22684 has been merged 👍   so the only thing left here is to update base definitions that use `&` and add a deprecation ?

StefanKarpinski:
Yeah, I think so. There was a PR for this, but it may be easier to just do it from scratch. Do you want to tackle this, @musm? If so, that would be great.

musm:
@StefanKarpinski  I am working on this

JeffBezanson:
Deprecation is in for 0.7. For 1.0, we should change the parsing of `&x` to a `call` expression.

PR #34767, opened 2020-02-14
interpret &(a,b) as a call expresion

### LINKED ISSUE #34693: Parse `&x`as a call expression
In https://github.com/JuliaLang/julia/issues/6080#issuecomment-327029743 @JeffBezanson wrote: 
> Deprecation is in for 0.7. For 1.0, we should change the parsing of &x to a call expression.

This doesn't seem to have happened, cf https://discourse.julialang.org/t/ampersand-operator-not-consistent-with-help/34266. Is this something that can be changed for 1.x or would it have to wait for 2.0?

### LINKED ISSUE #34748: Bad syntax error for bitwise and being used non-infix
```jl
julia> 1 & 2
0

julia> &(1,2)
ERROR: syntax: invalid syntax &(Core.tuple(1, 2))
Stacktrace:
 [1] top-level scope at REPL[9]:1
```

Showing the `Core.tuple` seems potentially confusing?

birm:
closes #34693
I believe that & is included in this list by accident, but I'm not exactly sure.

### DIFF
--- a/src/julia-parser.scm
+++ b/src/julia-parser.scm
@@ -115,7 +115,7 @@
 (define syntactic-operators
   (append! (add-dots '(= += -= *= /= //= |\\=| ^= ÷= %= <<= >>= >>>= |\|=| &= ⊻=))
            '(:= --> $= && |\|\|| |.| ... ->)))
-(define syntactic-unary-operators '($ & |::|))
+(define syntactic-unary-operators '($ |::|))
 
 (define syntactic-op? (Set syntactic-operators))
 (define syntactic-unary-op? (Set syntactic-unary-operators))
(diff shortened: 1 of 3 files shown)


Keno:
> I believe that & is included in this list by accident, but I'm not exactly sure.

No, `&` used to be unary operator syntax for use in ccall. There's been talk about re-using it for automatic boxing.

birm:
Ah, it looks like this has been discussed in #34693 and earlier #6080.

yuyichao:
Yes. #34748 is only about printing and is not at all controversial. https://github.com/JuliaLang/julia/issues/34693 is about changing the parsing and is no decided on yet.

birm:
ok, I've changed the closes to reference #34693 instead of #34748

birm:
This would/could also involve reverting part of 68eaec6 if merged.

PR #34768, opened 2020-02-14
windows: Add script for automatically creating minidumps on timeout

Keno:
This adds an autodump script in `contrib/autodump.jl`. This is intended
to be used on CI as `julia autodump.jl julia runtests.jl` and will,
on timeout automatically generate a memory dump of all the running processes
for later offline analysis. The idea here is to try to debug the hangs
that have been plaguing windows CI lately.

cc @staticfloat for actually trying this out in the buildbot configuration (seems to work ok locally).

### DIFF
--- a/contrib/autodump.jl
+++ b/contrib/autodump.jl
@@ -1,0 +1,163 @@
+using Base.Filesystem
+using Dates
+
+const TIMEOUT = 2*60*60 # seconds
+
+const HANDLE = Ptr{Cvoid}
+const LPVOID = Ptr{Cvoid}
+const DWORD = UInt32
+const BOOL = Cint
+const UINT = Cuint
+const JOBOBJECTINFOCLASS = Cint
+
+function CloseHandle(handle)
+    Base.windowserror(:CloseHandle,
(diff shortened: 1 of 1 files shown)


contrib/autodump.jl
```diff
@@ -0,0 +1,163 @@
+using Base.Filesystem
+using Dates
+
+const TIMEOUT = 2*60*60 # seconds
```
staticfloat: Instead of having a `TIMEOUT` hardcoded like this, can we have this react to the buildbot's exit signal?  We can specify it to set `SIGTERM`, then respond to that here?

staticfloat (changes requested):

contrib/autodump.jl
Keno: I thought about it, but I actually think I would prefer not to use the buildbot's timeout mechanism. If this dumping doesn't complete for whatever reason, I think we should keep it untouched to allow for manual investigation of whatever weird state the kernel is in.

staticfloat:
Tested and confirmed working; I'm going to deploy this to the buildbots (and store this elsewhere) and we'll see how it goes.

PR #32117, opened 2019-05-23
RFC: More methods for Irrational

perrutquist:
The purpose of this PR is to make `Irrational` work with many more functions, including those mentioned in #31949 #26701 and #22878.

The main way that this is accomplished is by defining
```
zero(::AbstractIrrational) = false
zero(::Type{<:AbstractIrrational}) = false
one(::Type{<:AbstractIrrational}) = true
```

The documentation says that these functions should return the additive/multiplicative identities *for* the type `T` - not *of* the type T.

In order to make `false` and `true` into additive/multiplicative identities, this PR also re-defines `Irrational`-`Bool` addition, subtraction and multiplication.
This is technically breaking if someone is currently using `true * pi` or `pi + false` as a way of forcing conversion to `Float64`. (The alternatives `1 * pi` and `0 + pi` continue to work as before.)
The new +, - and * methods are not type-stable, which the old ones were. On the other hand, Julia is now much better at type inference than it was when the original methods were written.

With this, a whole range of functions start to work, as the only reason they were throwing before was a call to `one` or `zero`. To give one example, `sec(x::Number)` is defined as `one(x)/cos(x)` which now works.

This PR also extends the list of two-argument functions that convert to Float64 when given two identical `Irrational` arguments. The intent is to make all "math-like" functions in `Base` work with `Irrational`. (This should mitigate the effect of `true*pi` now returning `pi` rather than a `Float64`.)

Finally, some minor fixes: `widemul` of two `Irrational`s now returns a `BigFloat` same as `widemul` of `Float64`s, and  `sin(pi)` and `tan(pi)` now return exactly `0.0`.

PR #34773, opened 2020-02-16
zero and one for irrational numbers

mohdibntarek:
This PR takes a fraction of https://github.com/JuliaLang/julia/pull/32117 and implements `zero`, `one` and ~`oneunit` (throws)~ for irrational numbers.

### DIFF
--- a/base/irrationals.jl
+++ b/base/irrationals.jl
@@ -136,6 +136,15 @@ hash(x::Irrational, h::UInt) = 3*objectid(x) - h
 
 widen(::Type{T}) where {T<:Irrational} = T
 
+zero(::AbstractIrrational) = false
+zero(::Type{<:AbstractIrrational}) = false
+
+one(::AbstractIrrational) = true
+one(::Type{<:AbstractIrrational}) = true
+
+oneunit(T::Type{<:AbstractIrrational}) = throw(ArgumentError("The number one cannot be of type $T"))
+oneunit(x::T) where {T <: AbstractIrrational} = oneunit(T)
+
 -(x::AbstractIrrational) = -Float64(x)
 for op in Symbol[:+, :-, :*, :/, :^]
     @eval $op(x::AbstractIrrational, y::AbstractIrrational) = $op(Float64(x),Float64(y))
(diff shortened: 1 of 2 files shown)


test/numbers.jl
```diff
@@ -1035,6 +1035,15 @@ end
     @test !(1 > NaN)
 end
 
+@testset "Irrational zero, one and oneunit" begin
+	@test one(pi) === true
+	@test zero(pi) === false
+	@test one(typeof(pi)) === true
+	@test zero(typeof(pi)) === false
+	@test_throws ArgumentError oneunit(pi)
+	@test_throws ArgumentError oneunit(typeof(pi))
```
waldyrious: Heads-up: the rest of the file uses spaces for indentation, rather than tabs.

mohamed82008: Fixed. Thanks!

base/irrationals.jl
```diff
@@ -136,6 +136,15 @@ hash(x::Irrational, h::UInt) = 3*objectid(x) - h
...
 widen(::Type{T}) where {T<:Irrational} = T
 
+zero(::AbstractIrrational) = false
+zero(::Type{<:AbstractIrrational}) = false
+
+one(::AbstractIrrational) = true
+one(::Type{<:AbstractIrrational}) = true
+
+oneunit(T::Type{<:AbstractIrrational}) = throw(ArgumentError("The number one cannot be of type $T"))
+oneunit(x::T) where {T <: AbstractIrrational} = oneunit(T)
```
stevengj: Why do we need this `oneunit` method, as opposed to the default `MethodError`?

mohamed82008: We don't but the error here is clearer. Should I remove it?

stevengj: Yes, in general people using Julia are expected to understand `MethodError`… in extremely common cases, where we need a more user-friendly message, we modify `showerror`, but I don't think that is needed here.

mohamed82008: Done

stevengj (approved):

PR #34777, opened 2020-02-16
Make processor_arm.cpp compile on musl

jpsamaroo:


jpsamaroo:
Actually this still doesn't pass for me locally. Getting:
```cpp
    CC src/processor.o
In file included from /home/jpsamaroo/julia/src/processor.cpp:844:0:
/home/jpsamaroo/julia/src/processor_arm.cpp:13:72: error: missing binary operator before token "("
 #if defined(_CPU_AARCH64_) || defined(__GLIBC_PREREQ) && __GLIBC_PREREQ(2, 16)
```

yuyichao:
You neeed to split that in a second line.

jpsamaroo:
Ok now this compiles for me

src/processor_arm.cpp
```diff
@@ -10,10 +10,16 @@
 #include <fstream>
 #include <algorithm>
 
-#if defined(_CPU_AARCH64_) || __GLIBC_PREREQ(2, 16)
+// This nesting is required to allow compilation on musl
+#define DYN_GETAUXVAL
```
vtjnash: Can you change this name to `USE_DYN_GETAUXVAL`, so it's clear it's our variable and not affecting the headers? Otherwise, LGTM.

vtjnash (approved):

src/processor_arm.cpp
jpsamaroo: I'm just moving around the preprocessor statements, not adding new variables. I'm not actually sure what that variable does :smile:

vtjnash: It just defines later for us whether we've got the sys/auxv header, so that we can link against `getauxval` directly, or should try to get it dynamically.

jpsamaroo: Sorry about the delay! The change has been made :slightly_smiling_face:

jpsamaroo:
Bump, is this good to merge @vtjnash ?

PR #34778, opened 2020-02-16
Run ast_transforms with invokelatest

timholy:
The recently-introduced `ast_transforms` need to be run under `Base.invokelatest` if it is going to be useful in packages. Current master fails the test in the first commit; the second commit fixes it.

xref https://github.com/timholy/Revise.jl/pull/425.


stevengj (approved):

StefanKarpinski:
MacOS failure is unrelated, so this is good to go.

PR #34783, opened 2020-02-17
Pass arrays instead of pointers when possible in blas.jl

andreasnoack:
These weren't needed. We can just rely on the implicit convertion in `ccall` and the presence of them gave a false impression that they were needed, see https://github.com/JuliaLang/julia/pull/34320.

### DIFF
--- a/stdlib/LinearAlgebra/src/blas.jl
+++ b/stdlib/LinearAlgebra/src/blas.jl
@@ -333,7 +333,7 @@ function dot(DX::Union{DenseArray{T},AbstractVector{T}}, DY::Union{DenseArray{T}
     if n != length(DY)
         throw(DimensionMismatch("dot product arguments have lengths $(length(DX)) and $(length(DY))"))
     end
-    GC.@preserve DX DY dot(n, pointer(DX), stride(DX, 1), pointer(DY), stride(DY, 1))
+    return dot(n, DX, stride(DX, 1), DY, stride(DY, 1))
 end
 function dotc(DX::Union{DenseArray{T},AbstractVector{T}}, DY::Union{DenseArray{T},AbstractVector{T}}) where T<:BlasComplex
     require_one_based_indexing(DX, DY)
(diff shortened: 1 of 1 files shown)


andreasnoack:
Looks like the only remaining test failure is unrelated.

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -464,7 +464,7 @@ function axpy!(alpha::Number, x::Union{DenseArray{T},StridedVector{T}}, y::Union
     if length(x) != length(y)
         throw(DimensionMismatch("x has length $(length(x)), but y has length $(length(y))"))
     end
-    GC.@preserve x y axpy!(length(x), convert(T,alpha), pointer(x), stride(x, 1), pointer(y), stride(y, 1))
+    return axpy!(length(x), convert(T,alpha), x, stride(x, 1), y, stride(y, 1))
     y
```
dkarrasch: This line appears to be obsolete now.

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -529,7 +536,7 @@ function axpby!(alpha::Number, x::Union{DenseArray{T},AbstractVector{T}}, beta::
     if length(x) != length(y)
         throw(DimensionMismatch("x has length $(length(x)), but y has length $(length(y))"))
     end
-    GC.@preserve x y axpby!(length(x), convert(T,alpha), pointer(x), stride(x, 1), convert(T,beta), pointer(y), stride(y, 1))
+    return axpby!(length(x), convert(T, alpha), x, stride(x, 1), convert(T, beta), y, stride(y, 1))
     y
```
dkarrasch: This as well.

dkarrasch (approved):
Wonderful simplification, though I have no idea about the low-level stuff like GC. Just two little oversights, and then ready to merge.

PR #34784, opened 2020-02-17
Fix geqp3! documentation

### LINKED ISSUE #694: LinearAlgebra.LAPACK.geqp3!  issue
There is no concordance between the documentation of `geqp3!(A) `
```
"""
    geqp3!(A) -> (A, jpvt, tau)
Compute the pivoted `QR` factorization of `A`, `AP = QR` using BLAS level 3.
Returns `A`, modified in-place, `jpvt`, which represents the pivoting matrix `P`,
and `tau`, which stores the elementary reflectors.
"""
```
and the actual computed results: see the last line of the LAPACK wrapper function      
   
```
function geqp3!(A::AbstractMatrix{$elty}, jpvt::AbstractVector{BlasInt}, tau::AbstractVector{$elty})
.
.
.
       return A, tau, jpvt
end
```

andreasnoack:
Fixes https://github.com/JuliaLang/LinearAlgebra.jl/issues/694. The ordering of the elements in the returned tuple wasn't correct. I also merged the three docstrings into a single since it's practically a single definition with optional arguments.

### DIFF
--- a/stdlib/LinearAlgebra/src/lapack.jl
+++ b/stdlib/LinearAlgebra/src/lapack.jl
@@ -592,17 +592,28 @@ Returns `A` and `tau` modified in-place.
 geqlf!(A::AbstractMatrix, tau::AbstractVector)
 
 """
-    geqp3!(A, jpvt, tau)
+    geqp3!(A, [jpvt, tau]) -> (A, tau, jpvt)
 
 Compute the pivoted `QR` factorization of `A`, `AP = QR` using BLAS level 3.
 `P` is a pivoting matrix, represented by `jpvt`. `tau` stores the elementary
-reflectors. `jpvt` must have length length greater than or equal to `n` if `A`
-is an `(m x n)` matrix. `tau` must have length greater than or equal to the
-smallest dimension of `A`.
+reflectors. The arguments `jpvt` and `tau` are optional and allow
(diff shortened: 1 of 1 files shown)


stdlib/LinearAlgebra/src/lapack.jl
```diff
@@ -592,18 +592,29 @@ Returns `A` and `tau` modified in-place.
...
 """
-    geqp3!(A, jpvt, tau)
+    geqp3!(A, [jpvt, tau]) -> (A, tau, jpvt)
 
 Compute the pivoted `QR` factorization of `A`, `AP = QR` using BLAS level 3.
 `P` is a pivoting matrix, represented by `jpvt`. `tau` stores the elementary
-reflectors. `jpvt` must have length length greater than or equal to `n` if `A`
-is an `(m x n)` matrix. `tau` must have length greater than or equal to the
-smallest dimension of `A`.
+reflectors. The arguments `jpvt` and `tau` are optional argument that allow
```
dkarrasch: repetition of "argument", maybe change to "The arguments `jpvt` and `tau` are optional and allow ..."

dkarrasch (approved):
Single failure is timeout after LinearAlgebra/addmul, but before that, LinearAlgebra/lapack tests passed. So I'd suggest to merge.

PR #34786, opened 2020-02-17
Add multiplication/division with/by Periods

### LINKED ISSUE #34637: Multiplication of an array of numbers with `Dates.Period`
Multiplying an array of `Period`s with a number works, but multiplying an array of numbers with a `Period` does not:

```julia
julia> (Second(1):Second(1):Second(5)) * 5
5 seconds:5 seconds:25 seconds

julia> (1:1:5) * Second(5)
ERROR: MethodError: no method matching *(::StepRange{Int64,Int64}, ::Second)
```

dkarrasch:
Fixes #34637.

### DIFF
--- a/stdlib/Dates/test/periods.jl
+++ b/stdlib/Dates/test/periods.jl
@@ -175,6 +175,15 @@ end
     @test Dates.Year(2) * 3//2 == Dates.Year(3)
     @test_throws InexactError y * 0.5
     @test_throws InexactError y * 3//4
+    @test (1:1:5)*Second(5) == Second(5)*(1:1:5) == collect(1:1:5)*Second(5) == Second(5)*collect(1:1:5)
+    @test (Second(2):Second(2):Second(10))/Second(2) == 1:1:5
+    @test collect(Second(2):Second(2):Second(10))/Second(2) == 1:1:5
+    @test (Second(2):Second(2):Second(10)) / 2 == Second(1):Second(1):Second(5)
+    @test collect(Second(2):Second(2):Second(10)) / 2 == Second(1):Second(1):Second(5)
+    @test (1:1:5)*Second(5) isa AbstractRange
+    @test Second(5)*(1:1:5) isa AbstractRange
+    @test (Second(2):Second(2):Second(10))/Second(2) isa AbstractRange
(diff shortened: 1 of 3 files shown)


stdlib/Dates/test/periods.jl
```diff
@@ -175,6 +175,15 @@ end
     @test Dates.Year(2) * 3//2 == Dates.Year(3)
     @test_throws InexactError y * 0.5
     @test_throws InexactError y * 3//4
+    @test (1:1:5)*Second(5) == Second(5)*(1:1:5) == collect(1:1:5)*Second(5) == Second(5)*collect(1:1:5)
```
timholy:
```suggestion
    @test (1:1:5)*Second(5) === Second(5)*(1:1:5) === Second(5):Second(5):Second(25)
```
and get rid of the first two `isa AbstractRange` tests below.

timholy: It would also be good to test that `(1:5)*Second(5)` gives the exact same result (including it being a `StepRange`).

stdlib/Dates/test/periods.jl
```diff
@@ -175,6 +175,15 @@ end
...
+    @test (Second(2):Second(2):Second(10))/Second(2) == 1:1:5
+    @test collect(Second(2):Second(2):Second(10))/Second(2) == 1:1:5
+    @test (Second(2):Second(2):Second(10)) / 2 == Second(1):Second(1):Second(5)
+    @test collect(Second(2):Second(2):Second(10)) / 2 == Second(1):Second(1):Second(5)
+    @test (1:1:5)*Second(5) isa AbstractRange
+    @test Second(5)*(1:1:5) isa AbstractRange
+    @test (Second(2):Second(2):Second(10))/Second(2) isa AbstractRange
```
timholy: These latter two you can also probably get rid of by using an appropriate triple-equals above.

timholy (commented):
Looks good!

timholy (approved):
Once this gets through CI, I favor merging.

dkarrasch:
The only failure is a timeout (win32) after the `Dates/periods` tests passed.

mbauman:
Sorry I'm late to review here, but this looks perfect.  Thanks @dkarrasch!

PR #34791, opened 2020-02-17
add `bitreverse` function

JeffBezanson:
This can be useful for parsing binary formats, and generally helps round out the set of bit-level operations.

### DIFF
--- a/base/int.jl
+++ b/base/int.jl
@@ -860,3 +860,37 @@ for op in (:+, :-, :*, :&, :|, :xor)
         return $op(aT, bT)
     end
 end
+
+const _mask1_uint128 = (UInt128(0x5555555555555555) << 64) | UInt128(0x5555555555555555)
+const _mask2_uint128 = (UInt128(0x3333333333333333) << 64) | UInt128(0x3333333333333333)
+const _mask4_uint128 = (UInt128(0x0f0f0f0f0f0f0f0f) << 64) | UInt128(0x0f0f0f0f0f0f0f0f)
+
+"""
+    bitreverse(x)
+
+Reverse the order of bits in integer `x`. `x` must have a fixed bit width,
+e.g. be an `Int16` or `Int32`.
+
+!!! compat "Julia 1.5"
+    This function requires Julia 1.5 or later.
+
+# Examples
+```jldoctest
+julia> bitreverse(0x8080808080808080)
+0x0101010101010101
+
+julia> reverse(bitstring(0xa06e)) == bitstring(bitreverse(0xa06e))
+true
+```
+"""
+function bitreverse(x::BitInteger)
+    # TODO: consider using llvm.bitreverse intrinsic
+    z = unsigned(x)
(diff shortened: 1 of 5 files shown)


StefanKarpinski (approved):

base/int.jl
```diff
@@ -860,3 +860,67 @@ for op in (:+, :-, :*, :&, :|, :xor)
...
+
+julia> reverse(bitstring(0xa06e)) == bitstring(bitreverse(0xa06e))
+true
+```
+"""
+function bitreverse end
+
+function bitreverse(x::Union{Int8,UInt8})
+    z = x % UInt8
+    z = ((z >>> 1) & 0x55) | ((z << 1) & 0xaa)
```
Keno: why the mix of arithmetic and logic shifts?

StefanKarpinski: There is no `<<<`?

base/int.jl
```diff
@@ -860,3 +860,67 @@ for op in (:+, :-, :*, :&, :|, :xor)
...
+    z = ((z >>> 4) & 0x0f) | ((z << 4) & 0xf0)
+    return z % typeof(x)
+end
+
+function bitreverse(x::Union{Int16,UInt16})
+    z = x % UInt16
+    z = ((z >>> 1) & 0x5555) | ((z << 1) & 0xaaaa)
+    z = ((z >>> 2) & 0x3333) | ((z << 2) & 0xcccc)
+    z = ((z >>> 4) & 0x0f0f) | ((z << 4) & 0xf0f0)
+    z = ((z >>> 8) & 0x00ff) | ((z << 8) & 0xff00)
```
jakobnissen: Sorry to jump in, but why not use `bswap` for some of these operations? It's shorter, easier to understand, less reliant on magic numbers, and presumably compiles to efficient LLVM.

jakobnissen: In fact, one could use a single method for all ints, like this:
```
function bitreverse(x::Base.BitInteger)
    z = unsigned(x)
    mask1 = 0x55555555555555555555555555555555 % typeof(x)
    mask2 = 0x33333333333333333333333333333333 % typeof(x)
    mask4 = 0xffffffffffffffffffffffffffffffff % typeof(x)
    z = ((z & mask1) << 1) | ((z >> 1) & mask1)
    z = ((z & mask2) << 2) | ((z >> 2) & mask2)
    z = ((z & mask4) << 4) | ((z >> 4) & mask4)
    return bswap(z)
end
```

KristofferC: In fact, LLVM already has a `bitreverse` intrinsic: https://llvm.org/docs/LangRef.html#llvm-bitreverse-intrinsics. But it will probably compile to the same thing.

StefanKarpinski: But with the intrinsic, it's (more) not our problem to make sure it's fast on all the architectures.

JeffBezanson: I'll use this implementation, since adding intrinsics is harder and I'd rather avoid llvmcall.

PR #34793, opened 2020-02-17
llvm: address v11-dev compatibility issues

vtjnash:
LLVM is unabashedly creating more nuisance renaming changes than covered here (https://groups.google.com/forum/#!topic/llvm-dev/DZRKP9qriZU), so the last known good commit is e252293d0629066225228596a9ef9d397513eb99. But this is a start towards staying functional.

### DIFF
--- a/src/llvm-version.h
+++ b/src/llvm-version.h
@@ -15,3 +15,5 @@
 #if JL_LLVM_VERSION < 100000
 #define Align(a) (a)
 #endif
+
+#define LLVM_DISABLE_ABI_BREAKING_CHECKS_ENFORCING 0
--- a/src/codegen.cpp
+++ b/src/codegen.cpp
@@ -525,7 +525,7 @@ struct jl_returninfo_t {
     unsigned return_roots;
 };
 
-static jl_returninfo_t get_specsig_function(Module *M, const std::string &name, jl_value_t *sig, jl_value_t *jlrettype);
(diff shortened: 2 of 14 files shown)


src/llvm-version.h
```diff
@@ -15,3 +15,5 @@
 #if JL_LLVM_VERSION < 100000
 #define Align(a) (a)
 #endif
+
+#define LLVM_DISABLE_ABI_BREAKING_CHECKS_ENFORCING 0
```
Keno: Why is this necessary now?

vtjnash: I'm tired of carrying the patch to upstream

Keno: Which LLVM patch is that? I thought these checks were useful to avoid accidentally linking the wrong LLVM ABI.

vtjnash: careful--I'm disabling not disabling the checks here. https://github.com/JuliaLang/julia/blob/6aa0d924ff5431318a6c79c7c9ce3a87297170d0/deps/patches/llvm-6.0-DISABLE_ABI_CHECKS.patch

Keno: > careful--I'm disabling not disabling the checks here.

You are if the patch is applied though, right?

vchuravy: That patch is morally wrong and apparently I used to not understand C pre processor flags. I am happy with dropping that patch.

PR #34796, opened 2020-02-18
Documentation: Mention `Vararg{Any, N}` in the "Performance Tips" section of the manual

DilumAluthge:
This pull request adds two lines to the "Performance Tips" section of the manual to explain that `Vararg{Any, N}` can be used to force specialization on varargs.

I'm not sure if this should be `Vararg{<:Any, N}` or `Vararg{Any, N}`. They both seem to force specialization, so I'm not sure if there is any difference between the two.

Hat tip to @tkf for showing me that this works.

### DIFF
--- a/doc/src/manual/performance-tips.md
+++ b/doc/src/manual/performance-tips.md
@@ -534,6 +534,11 @@ but this will:
 g_vararg(x::Vararg{Int, N}) where {N} = tuple(x...)
 ```
 
+This will also specialize, and is useful when the arguments are not all of the same type:
+```julia
+h_vararg(x::Vararg{<:Any, N}) where {N} = tuple(x...)
+```
+
 Note that [`@code_typed`](@ref) and friends will always show you specialized code, even if Julia
 would not normally specialize that method call. You need to check the
 [method internals](@ref ast-lowered-method) if you want to see whether specializations are generated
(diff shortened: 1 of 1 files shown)


KristofferC:
Will it specialize on the types or just on `N`?

StefanKarpinski:
That's a lot of reviewers. I'll decline since I don't know about this.

JeffBezanson:
I'd prefer `Any` here for simplicity.

DilumAluthge:
> I'd prefer `Any` here for simplicity.

Done. I’ve changed `<:Any` to `Any`.

I’m assuming the two forms are equivalent?

doc/src/manual/performance-tips.md
```diff
@@ -534,6 +534,11 @@ but this will:
 g_vararg(x::Vararg{Int, N}) where {N} = tuple(x...)
 ```
 
+This will also specialize, and is useful when the arguments are not all of the same type:
+```julia
+h_vararg(x::Vararg{Any, N}) where {N} = tuple(x...)
+```
```
timholy: Maybe add "One only needs to introduce a single type parameter to force specialization, even if the other types are unconstrained."

DilumAluthge: Done! Take a look.

timholy (approved):

timholy:
Thanks @DilumAluthge!

PR #34797, opened 2020-02-18
Update operators.jl

githubtomtom:
why need `const` for operator definition? `>:` is the only operator that has `const`


StefanKarpinski:
LGTM but pinging @JeffBezanson in case this has some special meaning I'm not aware of.

PR #34804, opened 2020-02-18
remove error when calling fieldtype on a Module

### LINKED ISSUE #13021: Error message regression "type Module has no field val"
```
julia> module M1
       val = 42
       end
M1

julia> M1.val
42

julia> M1.val = 43
ERROR: type Module has no field val
```

The above is with a recent master build (fcb38c1). With 0.3.11 the error is `ERROR: cannot assign variables in other modules`, which seems more clear.

### LINKED ISSUE #33347: Strange error when querying `fieldtype` of `Module`
```
julia> fieldtypes(Module)
ERROR: cannot assign variables in other modules
Stacktrace:
 [1] #9 at ./reflection.jl:712 [inlined]
 [2] ntuple(::Base.var"##9#10"{DataType}, ::Int64) at ./ntuple.jl:18
 [3] fieldtypes(::Type) at ./reflection.jl:712
 [4] top-level scope at REPL[35]:1
```

KristofferC:
This was added in https://github.com/JuliaLang/julia/commit/8f08a1f8bf11628e0dff8ad1a76923e2de9e99e6 to fix https://github.com/JuliaLang/julia/issues/13021 but it doesn't seem to be needed anymore.

Fixes https://github.com/JuliaLang/julia/issues/33347.

### DIFF
--- a/src/builtins.c
+++ b/src/builtins.c
@@ -885,8 +885,6 @@ JL_CALLABLE(jl_f_fieldtype)
     }
     JL_NARGS(fieldtype, 2, 2);
     jl_datatype_t *st = (jl_datatype_t*)args[0];
-    if (st == jl_module_type)
-        jl_error("cannot assign variables in other modules");
     return get_fieldtype(args[0], args[1], 1);
 }
 
--- a/test/misc.jl
+++ b/test/misc.jl
@@ -804,3 +804,7 @@ end
 
     GC.safepoint()
 end
+
+@testset "fieldtypes Module" begin
+    @test fieldtypes(Module) isa Tuple
+end


JeffBezanson:
Since `Module` doesn't have usable fields (via `getfield` etc.) I suspect fieldtypes and fieldnames should return `()`.

vtjnash:
That fact seems a bit muddy, since it does have usable fields via `getfield`, but we call the listing of them `propertynames` (or `names`) instead of `fieldnames`:
```
julia> getfield(Main, :Main)
Main

julia> propertynames(Main)
5-element Array{Symbol,1}:
...
```

vtjnash:
Also, FWIW, we have now:
```
julia> fieldnames(Module)
(:name, :parent)

julia> fieldcount(Module)
2
```

So this PR would currently give also give the corresponding type of that pair `(Symbol, Module)` too

JeffBezanson:
I guess fieldtypes isn't the most harmful of that lot, but I think they're all pretty misleading.

PR #34807, opened 2020-02-18
partr: Fix deadlock

### LINKED ISSUE #34769: Multi-threading deadlock bug in Windows (Julia 1.3 and 1.4-rc1)
cc @ViralBShah and @Keno

I encountered a bug in Julia that results in a deadlock on Windows. The bug gets triggered by my package tests, but only when running the tests on a CI virtual machine (both Travis Windows and Appveyor). Notably, if I run the tests (via `Pkg.test()`) manually on the Appveyor VM after I RDP into it, everything works. The bug only gets triggered if the tests were initiated by the CI service. Not sure what's going on there.

Some more info and context is available in [this issue](https://github.com/Circuitscape/Omniscape.jl/issues/13) in my Julia package repo.

Here is [google drive link to a minidump](https://drive.google.com/open?id=18SqE_HtlE1UqeNao8soQ-NjWL9iA-ewv) from the Appveyor VM that hopefully has the information needed to debug this. I would happily try to debug it myself if I could, but it is beyond my capabilities.

You can see the backtrace (from an Appveyor VM) confirming that it is a deadlock [here](https://github.com/Circuitscape/Omniscape.jl/issues/13#issuecomment-586377161).

Here's the session info for the Julia version that was used on the Appveyor VM (but note that the bug is also triggered in Julia 1.3.0 and 1.3.1):
```
Julia Version 1.4.0-rc1.0
Commit b0c33b0cf5 (2020-01-23 17:23 UTC)
Platform Info:
  CPU: Intel(R) Xeon(R) CPU E5-2697 v3 @ 2.60GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-8.0.1 (ORCJIT, haswell)
Environment:
  JULIA_NUM_THREADS = 1
  JULIA_PROJECT = @.
  JULIA_VERSION = 1.4
```
and you can see the Appveyor output [here](https://ci.appveyor.com/project/vlandau/omniscape-jl/builds/30808532).

Thanks, and please let me know if I can provide any additional info.

Keno:
When there is no work to do, the first thread to be idle will attempt to
run the event loop once, waiting for any notifications (which will usually
create new work). However, there is an interesting corner case where a
notification arrives, but no work was scheduled. That doesn't usually happen,
but there are a few situations where it does:

1) Somebody used a libuv primitive outside of julia, so the callback
   doesn't schedule any julia work.
2) Another thread forcibly interrupted the waiting thread because it wants to
   take over the event loop for various reasons
3) On Windows, we occasionally get spurious wake ups of the event loop.

The existing code in partr assumed that we were in situation 2, i.e. that
there was another thread waiting to take over the event loop, so it released
the event loop and simply put the current thread to sleep in the expectation
that another thread will pick it up. However, if we instead are in one of the
other two conditions, there may not be another thread there to pick up the event
loop. Thus, with no thread owning the event loop, julia will stop responding to
events and effectively deadlock. Since both 1 and 3 are rare, and we don't actually
enter the event loop until there was no work for 4 milliseconds (which is fairly rare),
this condition rarely happens, but is occasionally observable on Windows, where it
caused #34769. To test that this fix works, we manually create situation 1 in the
test by creating an idle callback, which will prevent the event loop from blocking,
but only schedules julia work after it's been called 100 times. This reproduces
the observed failure from the issue and is fixed by this PR.

Fixes #34769

### DIFF
--- a/test/threads.jl
+++ b/test/threads.jl
@@ -31,7 +31,7 @@ function idle_callback(handle)
             # the thread to go to sleep, which only happens
             # after some default amount of time (DEFAULT_THREAD_SLEEP_THRESHOLD)
             # so spend that amount of time here.
-            ccall(:usleep, Cint, (Csize_t,), 4000)
+            Libc.systemsleep(0.004)
         elseif idle.count >= 10
             lock(idle.cond)
             try
@@ -111,14 +111,16 @@ function Base.wait(idle::UvTestIdle)
     end
 end
 
-# Spawn another process as a watchdog. If this test fails, it'll uncrecoverably
+# Spawn another process as a watchdog. If this test fails, it'll unrecoverably
 # hang in the event loop. Another process needs to kill it
 cmd = """
+    @async (Base.wait_close(stdin); exit())
     sleep(100)
+    isopen(stdin) || exit()
     println(stderr, "ERROR: Killing threads test due to watchdog expiry")
     ccall(:uv_kill, Cint, (Cint, Cint), $(getpid()), Base.SIGTERM)
 """
-proc = run(pipeline(`$(Base.julia_cmd()) -e $cmd`; stderr=stderr); wait=false)
+proc = open(pipeline(`$(Base.julia_cmd()) -e $cmd`; stderr=stderr); write=true)
 
 let idle=UvTestIdle()
     wait(idle)
@@ -134,4 +136,4 @@ end
 @test process_running(proc)
 
 # We don't need the watchdog anymore
-kill(proc, Base.SIGKILL)
+close(proc.in)
--- a/test/threads.jl
+++ b/test/threads.jl
@@ -114,7 +114,7 @@ end
 # Spawn another process as a watchdog. If this test fails, it'll unrecoverably
 # hang in the event loop. Another process needs to kill it
 cmd = """
-    @async (Base.wait_close(stdin); exit())
+    @async (Base.wait_readnb(stdin, 1); exit())
     sleep(100)
     isopen(stdin) || exit()
     println(stderr, "ERROR: Killing threads test due to watchdog expiry")
--- a/test/threads.jl
+++ b/test/threads.jl
@@ -20,3 +20,118 @@ if Sys.islinux() && Sys.CPU_THREADS > 1 && Sys.which("taskset") !== nothing
     @test endswith(run_with_affinity("1"), "2")
     @test endswith(run_with_affinity("0,1"), "3")
 end
+
+# issue #34769
+function idle_callback(handle)
+    idle = @Base.handle_as handle UvTestIdle
+    if idle.active
+        idle.count += 1
+        if idle.count == 1
+            # We want to hit the case where we're allowing
+            # the thread to go to sleep, which only happens
+            # after some default amount of time (DEFAULT_THREAD_SLEEP_THRESHOLD)
+            # so spend that amount of time here.
+            ccall(:usleep, Cint, (Csize_t,), 4000)
+        elseif idle.count >= 10
+            lock(idle.cond)
+            try
+                notify(idle.cond, true)
+            finally
+                unlock(idle.cond)
+            end
+            idle.active = false
+        end
+    end
+    nothing
+end
+
+mutable struct UvTestIdle
+    handle::Ptr{Cvoid}
+    cond::Base.ThreadSynchronizer
+    isopen::Bool
+    active::Bool
+    count::Int
+
+    function UvTestIdle()
+        this = new(Libc.malloc(Base._sizeof_uv_idle), Base.ThreadSynchronizer(), true, false, 0)
+        Base.iolock_begin()
+        Base.associate_julia_struct(this.handle, this)
+        err = ccall(:uv_idle_init, Cint, (Ptr{Cvoid}, Ptr{Cvoid}),
+            Base.eventloop(), this.handle)
+        if err != 0
+            Libc.free(this.handle)
+            this.handle = C_NULL
+            throw(_UVError("uv_idle_init", err))
+        end
+        err = ccall(:uv_idle_start, Cint, (Ptr{Cvoid}, Ptr{Cvoid}),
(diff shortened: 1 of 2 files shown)


test/threads.jl
```diff
@@ -20,3 +20,118 @@ if Sys.islinux() && Sys.CPU_THREADS > 1 && Sys.which("taskset") !== nothing
...
+        idle.active = true
+        wait(idle.cond)
+    finally
+        Base.unlock(idle.cond)
+        Base.unpreserve_handle(idle)
+        Base.iolock_end()
+    end
+end
+
+# Spawn another process as a watchdog. If this test fails, it'll uncrecoverably
```
vtjnash:
```suggestion
# Spawn another process as a watchdog. If this test fails, it'll unrecoverably
```

test/threads.jl
```diff
@@ -20,3 +20,118 @@ if Sys.islinux() && Sys.CPU_THREADS > 1 && Sys.which("taskset") !== nothing
...
+@threads for i = 1:1
+    let idle=UvTestIdle()
+        wait(idle)
+    end
+end
+
+@test process_running(proc)
+
+# We don't need the watchdog anymore
+kill(proc, Base.SIGKILL)
```
vtjnash:
```suggestion
close(proc.in)
```

test/threads.jl
```diff
@@ -20,3 +20,118 @@ if Sys.islinux() && Sys.CPU_THREADS > 1 && Sys.which("taskset") !== nothing
...
+end
+
+# Spawn another process as a watchdog. If this test fails, it'll uncrecoverably
+# hang in the event loop. Another process needs to kill it
+cmd = """
+    sleep(100)
+    println(stderr, "ERROR: Killing threads test due to watchdog expiry")
+    ccall(:uv_kill, Cint, (Cint, Cint), $(getpid()), Base.SIGTERM)
+"""
+proc = run(pipeline(`$(Base.julia_cmd()) -e $cmd`; stderr=stderr); wait=false)
```
vtjnash:
```suggestion
proc = open(pipeline(`$(Base.julia_cmd()) -e $cmd`; stderr=stderr); write=true)
```

test/threads.jl
```diff
@@ -20,3 +20,118 @@ if Sys.islinux() && Sys.CPU_THREADS > 1 && Sys.which("taskset") !== nothing
...
+        Base.unlock(idle.cond)
+        Base.unpreserve_handle(idle)
+        Base.iolock_end()
+    end
+end
+
+# Spawn another process as a watchdog. If this test fails, it'll uncrecoverably
+# hang in the event loop. Another process needs to kill it
+cmd = """
+    sleep(100)
```
vtjnash:
```suggestion
    @async (wait_close(stdin); exit())
    sleep(100)
    isopen(stdin) || exit()
```

vtjnash (approved):

test/threads.jl
```diff
@@ -20,3 +20,118 @@ if Sys.islinux() && Sys.CPU_THREADS > 1 && Sys.which("taskset") !== nothing
...
+function idle_callback(handle)
+    idle = @Base.handle_as handle UvTestIdle
+    if idle.active
+        idle.count += 1
+        if idle.count == 1
+            # We want to hit the case where we're allowing
+            # the thread to go to sleep, which only happens
+            # after some default amount of time (DEFAULT_THREAD_SLEEP_THRESHOLD)
+            # so spend that amount of time here.
+            ccall(:usleep, Cint, (Csize_t,), 4000)
```
vtjnash:
```suggestion
            Libc.systemsleep(0.004)
```

test/threads.jl
```diff
@@ -20,3 +20,120 @@ if Sys.islinux() && Sys.CPU_THREADS > 1 && Sys.which("taskset") !== nothing
...
+        Base.unlock(idle.cond)
+        Base.unpreserve_handle(idle)
+        Base.iolock_end()
+    end
+end
+
+# Spawn another process as a watchdog. If this test fails, it'll unrecoverably
+# hang in the event loop. Another process needs to kill it
+cmd = """
+    @async (Base.wait_close(stdin); exit())
```
Keno: Ah, looks like it's that bug where we don't get close notifications unless we're reading.
```suggestion
    @async (Base.wait_readnb(stdin, 1); exit())
```

JeffBezanson (approved):

ViralBShah:
Can this be backported to 1.3 and 1.4?

Keno:
Will there be another 1.3 release?

KristofferC:
Unlikely

StefanKarpinski:
Wow, that’s some serious sleuth work. Also, new record for highest test to changed code ratio?

test/threads.jl
```diff
@@ -20,3 +20,120 @@ if Sys.islinux() && Sys.CPU_THREADS > 1 && Sys.which("taskset") !== nothing
     @test endswith(run_with_affinity("1"), "2")
     @test endswith(run_with_affinity("0,1"), "3")
 end
+
+# issue #34769
+function idle_callback(handle)
+    idle = @Base.handle_as handle UvTestIdle
```
KristofferC: I cri evrytiem

ViralBShah:
I wonder is @vlandau can test this once it is merged using a nightly or an rc. Also, @vlandau, thanks for persevering through this and getting us the debug information.

vlandau:
> @vlandau can you test this once it is merged using a nightly or an rc. Also, @vlandau, thanks for persevering through this and getting us the debug information.

Yes, happy to test this out in my use case! I learned a lot in the process, so glad to have been able to help get the debug info! Thanks very much @Keno et al.


> Unlikely

@KristofferC  Does you mean this will be likely only be released in 1.4? (not that it's a big deal for me, as users of my package will likely *never* encounter this bug in 1.3)

KristofferC:
It means that it is unlikely that there will be another 1.3 release.

vlandau:
> It means that it is unlikely that there will be another 1.3 release.

yeah I got that part of it

Issue #34782, opened 2020-02-17
Floor / Ceil  methods for Number and Period Inconsistency


Nosferican:
I just noticed that the `floor` / `ceil`  methods for `Number` and `Period` have the positional arguments inverted.

```
floor(Int, 3.5)
using Dates
floor(DateTime("2000-01-31"), Month)
```

From Slack, Dates should be consistent with Base.

Related: #18574

dkarrasch:
Out of curiosity, what has changed that the arguments against this proposal from #18574 are no longer valid?

Nosferican:
I think just people keep getting confused by the order... Would it be too bad to have both methods?
```
ceil(dt::TimeType, p::Period) -> TimeType
ceil(p::Period, dt::TimeType) -> TimeType
ceil(x::Period, precision::T) where T <: Union{TimePeriod, Week, Day} -> T
ceil(precision::T, x::Period) where T <: Union{TimePeriod, Week, Day} -> T
```

rapus95:
As I already posted in https://github.com/JuliaLang/julia/pull/34810#issuecomment-593851089 I'm thinking that the original considerations were right and the current behaviour is not inconsistent. It can be seen easily if we add a function which can round to a period while converting the time type:
```julia
round(Date, now(), Month, RoundUp) == Date(2020, 4) #at least for the month of the post 😄
```
Tl;Dr: rather add those missing functions and make only the 2nd argument a required one

Moelf:
I'd like to work on this, if I understand this correctly, we'd like the following to pass as tests in the end?
```
now_ = now()
for p in (Year, Month, Day)
    for r in (RoundUp, RoundDown)
        @test round(Date, now_, p, r) == round(Date(now_), p, r)
    end
    @test floor(Date, now_, p) == round(Date, now_, p, RoundDown)
    @test ceil(Date, now_, p)  == round(Date, now_, p, RoundUp)
end
```

PR #34810, opened 2020-02-19
Update floor, ceil and round in Dates to be consistent with Base

ssikdar1:
PR for issue #34782 to have `floor`, `ceil`, and `round` in the `Dates` library to be consistent with Base where
`floor([T,] x)` converts the result to type T.

PR changes the positional arguments, update the docstrings and updates the `rounding.jl` unit tests.

## Testing
Tests pass with `$ ./julia stdlib/Dates/test/rounding.jl `

### DIFF
--- a/stdlib/Dates/docs/src/index.md
+++ b/stdlib/Dates/docs/src/index.md
@@ -549,13 +549,13 @@ julia> Dates.value(Dates.Millisecond(10))
 month or 15 minutes) with [`floor`](@ref), [`ceil`](@ref), or [`round`](@ref):
 
 ```jldoctest
-julia> floor(Date(1985, 8, 16), Dates.Month)
+julia> floor(Dates.Month, Date(1985, 8, 16))
 1985-08-01
 
-julia> ceil(DateTime(2013, 2, 13, 0, 31, 20), Dates.Minute(15))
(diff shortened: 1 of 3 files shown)


rapus95:
Imo we rather should add functions to support `round(Date, now(), Month)` as that perfectly mirrors the `round(T, val, precision)` function. I.e. when writing `round(now(), Day)` and claming that it should be `round(Day, now())` we should rather define `round(Date, now()) == convert(Date, round(now(), Day))` where in the latter statement the rounding is unneccessary though.

Tl;Dr: Proposal:
Add `round([::Type{<:TimeType},] val::TimeType [, precision::Period [, ::RoundingMode]])`

ssikdar1:
Added 
```
floor(dt::TimeType, p::Period) -> TimeType
ceil(dt::TimeType, p::Period) -> TimeType
round(dt::TimeType, p::Period, [r::RoundingMode]) -> TimeType
```
back in based on the comment in #34782. (https://github.com/JuliaLang/julia/issues/34782#issuecomment-588152602)

For functions like 
```
ceil(x::Period, precision::T) where T <: Union{TimePeriod, Week, Day} -> T
ceil(precision::T, x::Period) where T <: Union{TimePeriod, Week, Day} -> T
```
Not sure this can work since both Period and precision are of type `ConvertiblePeriod`.
( When I try to add them I get errors in the tests. )

Not sure about the proposal and convert?

PR #34816, opened 2020-02-19
Document change - Added reference link

NiteshKoushik:
Document change - Added reference link for the key words mutating and Kronecker_product

### DIFF
--- a/doc/src/manual/variables.md
+++ b/doc/src/manual/variables.md
@@ -56,7 +56,7 @@ In the Julia REPL and several other Julia editing environments, you can type man
 symbols by typing the backslash LaTeX symbol name followed by tab. For example, the variable
 name `δ` can be entered by typing `\delta`-*tab*, or even `α̂₂` by `\alpha`-*tab*-`\hat`-
 *tab*-`\_2`-*tab*. (If you find a symbol somewhere, e.g. in someone else's code,
-that you don't know how to type, the REPL help will tell you: just type `?` and
+that you don't know how to type, the [REPL](https://docs.julialang.org/en/v1/stdlib/REPL/) help will tell you: just type `?` and
 then paste the symbol.)
 
 Julia will even let you redefine built-in constants and functions if needed (although
--- a/doc/src/manual/variables.md
+++ b/doc/src/manual/variables.md
@@ -104,7 +104,7 @@ Operators like `+` are also valid identifiers, but are parsed specially. In some
 can be used just like variables; for example `(+)` refers to the addition function, and `(+) = f`
 will reassign it. Most of the Unicode infix operators (in category Sm), such as `⊕`, are parsed
 as infix operators and are available for user-defined methods (e.g. you can use `const ⊗ = kron`
-to define `⊗` as an infix Kronecker product).  Operators can also be suffixed with modifying marks,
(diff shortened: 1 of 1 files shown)


doc/src/manual/variables.md
```diff
@@ -104,7 +104,7 @@ Operators like `+` are also valid identifiers, but are parsed specially. In some
 can be used just like variables; for example `(+)` refers to the addition function, and `(+) = f`
 will reassign it. Most of the Unicode infix operators (in category Sm), such as `⊕`, are parsed
 as infix operators and are available for user-defined methods (e.g. you can use `const ⊗ = kron`
-to define `⊗` as an infix Kronecker product).  Operators can also be suffixed with modifying marks,
+to define `⊗` as an infix [Kronecker product](https://en.wikipedia.org/wiki/Kronecker_product)).  Operators can also be suffixed with modifying marks,
```
dkarrasch: You should break the line (also below) after 80 symbols.

doc/src/manual/variables.md
```diff
@@ -136,7 +136,7 @@ conventions:
     camel case instead of underscores.
   * Names of `function`s and `macro`s are in lower case, without underscores.
   * Functions that write to their arguments have names that end in `!`. These are sometimes called
-    "mutating" or "in-place" functions because they are intended to produce changes in their arguments
+    "[mutating](https://discourse.julialang.org/t/understanding-mutating-functions-correctly/30651)" or "in-place" functions because they are intended to produce changes in their arguments
```
fredrikekre: I don't think we usually add links to discourse like this. If this information is missing from the manual it can be added and then we can add a proper reference to that section here.

NiteshKoushik: Hello,
Mutating key word details is present in the manual. 
When I was going through the document, I want to know about the key word mutating - if there was a referencing link then that would help. For the same I have added this link.
Also on the key word "Kronecker product" - I feel this content needs to be added in the manual.

doc/src/manual/variables.md
NiteshKoushik: Hello,
Thanks for the suggestion. 
Please let me know which line need to be break as per 80 symbols. Is the line starting with "Operators can also be suffixed with modifying marks," should have line break.

doc/src/manual/variables.md
NiteshKoushik: Also I feel we needs to provide link to REPL key word - which has information in the same document.
Key word backslashed need to be replaced with backslash.

doc/src/manual/variables.md
```diff
@@ -53,10 +53,10 @@ julia> 안녕하세요 = "Hello"
 ```
 
 In the Julia REPL and several other Julia editing environments, you can type many Unicode math
-symbols by typing the backslashed LaTeX symbol name followed by tab. For example, the variable
+symbols by typing the backslash LaTeX symbol name followed by tab. For example, the variable
```
StefanKarpinski: This was more correct as it was.

doc/src/manual/variables.md
StefanKarpinski: All the ones that are longer than 80 chars.

vtjnash:
Thanks for the PR, but I think in general we expect people can research these terms on their own.

Issue #34771, opened 2020-02-15
`hypot` with both arguments 0 returns NaN with Unitful.jl


SebastianM-C:
https://github.com/JuliaLang/julia/pull/34316 introduced
```julia
hypot(x::T, y::T) where {T<:Number} = (z = y/x; abs(x) * sqrt(one(z) + z*z))
```
If `iszero(x)` and `iszero(y)`, the above returns NaN instead of `zero(T)`.

To trigger this one can use:
```
using Unitful
hypot(0u"m",0u"m")
```

One way to fix this would be
```julia
hypot(x::T, y::T) where {T<:Number} = !iszero(x) ? (z = y/x; abs(x) * sqrt(one(z) + z*z)) : abs(y)
```

If the above is ok, I can make a PR.

```julia
julia> versioninfo()
Julia Version 1.4.0-rc1.0
Commit b0c33b0cf5 (2020-01-23 17:23 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Core(TM) i7-7800X CPU @ 3.50GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-8.0.1 (ORCJIT, skylake)
Environment:
  JULIA_NUM_THREADS = 6
```

giordano:
I think it was https://github.com/JuliaLang/julia/pull/33224 that removed too much code

StefanKarpinski:
That attempt at avoiding overflow generically strikes me as a bit too clever, but 🤷‍♂

JeffBezanson:
Yes might as well PR that change.

JeffBezanson:
Fixed by #34817

PR #34817, opened 2020-02-19
Fix #34316

SebastianM-C:
The differences between the proposed solution in https://github.com/JuliaLang/julia/issues/34771#issue-565766128 are

- use `oneunit` instead of `one` (see https://github.com/PainterQubits/Unitful.jl/issues/291)
- add a check for `isinf` in order to have a similar behavior with the unitless case. Should this be done more generally with something like the following?
```julia
function hypot(x, y)
    #Return Inf if either or both inputs is Inf
    if isinf(x) || isinf(y)
        return promote_type(typeof(x), typeof(y))(Inf)
    end

    hypot(x, y)
end
```
- remove the `hypot` definition from `Furlongs`. I looked at why the tests were passing with the previous definition of `hypot` and I discovered that the fallback was not actually tested because of the additional method.

### DIFF
--- a/base/math.jl
+++ b/base/math.jl
@@ -630,7 +630,7 @@ hypot(x::Number, y::Number) = hypot(promote(x, y)...)
 hypot(x::Complex, y::Complex) = hypot(abs(x), abs(y))
 hypot(x::T, y::T) where {T<:Real} = hypot(float(x), float(y))
 function hypot(x::T, y::T) where {T<:Number}
-    #Return Inf if either or both imputs is Inf (Compliance with IEEE754)
+    # Return Inf if either or both inputs is Inf (Compliance with IEEE754)
     if isinf(x) || isinf(y)
         return T(Inf)
     end
@@ -639,7 +639,7 @@ function hypot(x::T, y::T) where {T<:Number}
 end
 
 function hypot(x::T, y::T) where T<:AbstractFloat
-    #Return Inf if either or both imputs is Inf (Compliance with IEEE754)
+    # Return Inf if either or both inputs is Inf (Compliance with IEEE754)
     if isinf(x) || isinf(y)
         return T(Inf)
     end
--- a/base/math.jl
+++ b/base/math.jl
@@ -630,12 +630,14 @@ hypot(x::Number, y::Number) = hypot(promote(x, y)...)
 hypot(x::Complex, y::Complex) = hypot(abs(x), abs(y))
 hypot(x::T, y::T) where {T<:Real} = hypot(float(x), float(y))
 function hypot(x::T, y::T) where {T<:Number}
-    # Return Inf if either or both inputs is Inf (Compliance with IEEE754)
-    if isinf(x) || isinf(y)
-        return T(Inf)
-    end
+    if !iszero(x)
+        z = y/x
+        z2 = z*z
 
-    !iszero(x) ? (z = y/x; abs(x) * sqrt(oneunit(z) + z*z)) : abs(y)
+        abs(x) * sqrt(oneunit(z2) + z2)
+    else
+        abs(y)
+    end
 end
 
 function hypot(x::T, y::T) where T<:AbstractFloat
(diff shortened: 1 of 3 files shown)


base/math.jl
```diff
@@ -629,7 +629,15 @@ julia> hypot(3, 4im)
 hypot(x::Number, y::Number) = hypot(promote(x, y)...)
 hypot(x::Complex, y::Complex) = hypot(abs(x), abs(y))
 hypot(x::T, y::T) where {T<:Real} = hypot(float(x), float(y))
-hypot(x::T, y::T) where {T<:Number} = (z = y/x; abs(x) * sqrt(one(z) + z*z))
+function hypot(x::T, y::T) where {T<:Number}
+    #Return Inf if either or both imputs is Inf (Compliance with IEEE754)
```
KristofferC:
```suggestion
    # Return Inf if either or both inputs is Inf (Compliance with IEEE754)
```

SebastianM-C: The is also another identical typo below.

base/math.jl
```diff
@@ -629,9 +629,17 @@ julia> hypot(3, 4im)
 hypot(x::Number, y::Number) = hypot(promote(x, y)...)
 hypot(x::Complex, y::Complex) = hypot(abs(x), abs(y))
 hypot(x::T, y::T) where {T<:Real} = hypot(float(x), float(y))
-hypot(x::T, y::T) where {T<:Number} = (z = y/x; abs(x) * sqrt(one(z) + z*z))
+function hypot(x::T, y::T) where {T<:Number}
+    # Return Inf if either or both inputs is Inf (Compliance with IEEE754)
+    if isinf(x) || isinf(y)
```
JeffBezanson: This method should not be called for IEEE754 floating point numbers; that case is caught by more specific methods. Putting float-specific behaviors here defeats the purpose of this, which is to handle very general number-like types that might not support all operations.

base/math.jl
```diff
@@ -629,9 +629,17 @@ julia> hypot(3, 4im)
...
+        return T(Inf)
+    end
+
+    !iszero(x) ? (z = y/x; abs(x) * sqrt(oneunit(z) + z*z)) : abs(y)
```
JeffBezanson: I would expect `oneunit(z)` and `z*z` to have incompatible units in general?

base/math.jl
JeffBezanson: Which is why I originally proposed `sqrt(abs2(x) + abs2(y))`, which is what you get if you pass 3 arguments to `hypot`. I'd be fine with that too.

base/math.jl
simeonschaub: Probably would be `oneunit(z)`?

base/math.jl
SebastianM-C: I see. How should I update the tests? Should I just delete the tests with `NaN` and `Inf`?

base/math.jl
SebastianM-C: Should I use `oneunit(z)*oneunit(z)` in case `^` is not defined (I suppose that's why it's `z*z` instead of `z^2`)?

SebastianM-C: I don't understand why did the tests pass with this...

JeffBezanson: Since `z` is `y/x` it is usually unitless (the units should cancel). Are there unusual cases where that doesn't hold? If not then I think the original `one(z)` is ok.

base/math.jl
JeffBezanson: Yes I think so. We'll still be following IEEE754 if you pass actual Float values, plus I think the specified behavior is really bizarre and unexpected. We already broke with it for `min` and `max` on `NaN`.

base/math.jl
SebastianM-C: The unusual case is in the tests. While `y/z` is unitless, it is represented as a `Furlong{0,Float64}` and `1+z*z` gives
```
ERROR: promotion of types Int64 and Furlong{0,Float64} failed to change any arguments
```

Should I define
```
Base.promote_rule(::Type{Furlong{0,T}}, ::Type{S}) where {T, S <: Number} = Furlong{0,promote_type(S,T)}
```
or should I use `oneunit(z)*oneunit(z)`?

JeffBezanson: I see, it uses furlongs^0 for that. Unlike Unitful.jl, but I suppose it's valid. How about
```
z = y/x; z2 = z*z; abs(x) * sqrt(oneunit(z2) + z2)
```

SebastianM-C:
Thanks for the review. I added the suggested changes in the latest commit.

PR #33758, opened 2019-11-04
Add modify! function for lookup/update/insert/delete in one go

tkf:
This implements `set!` function I proposed in https://github.com/JuliaLang/julia/pull/31367#issuecomment-504561329.

See #31367 for discussion on alternative API (e.g., token-based).

Quoting the docstring:

> ```julia
> modify!(f, dict::AbstractDict, key)
> ```
>
> Lookup and then update, insert or delete in one go without re-computing the hash.
> 
> `f` is a callable object that must accept `Union{Some{V}, Nothing}` and return `Union{T, Some{T}, Nothing}` where `T` is a type [`convert`](@ref)-able to the value type `V`.  The value `Some(d[key])` is passed to `f` if `haskey(d, key)`; otherwise `nothing` is passed.  If `f` returns `nothing`, corresponding entry in the dictionary `d`  is removed. If `f` returns non-`nothing` value `x`, `something(x)` is inserted to `d`.
> 
> `modify!` returns whatever `f` returns as-is.
> 
> **Examples**
> ```julia
> julia> dict = Dict("a" => 1);
> 
> julia> modify!(dict, "a") do val
>            Some(val === nothing ? 1 : something(val) + 1)
>        end
> Some(2)
> 
> julia> dict
> Dict{String,Int64} with 1 entry:
>   "a" => 2
> 
> julia> dict = Dict();
> 
> julia> modify!(dict, "a") do val
>            Some(val === nothing ? 1 : something(val) + 1)
>        end
> Some(1)
> 
> julia> dict
> Dict{Any,Any} with 1 entry:
>   "a" => 1
> 
> julia> modify!(_ -> nothing, dict, "a")
> 
> julia> dict
> Dict{Any,Any} with 0 entries
> ```

PR #34821, opened 2020-02-19
WIP/RFC: add 2-argument `get` returning `Union{Nothing,Some}`

JeffBezanson:
This form is a bit more general, and would eventually allow us to eliminate the loathsome `secret_table_token`.

Obviously incomplete; we should first discuss whether we want this.

### DIFF
--- a/src/iddict.c
+++ b/src/iddict.c
@@ -159,6 +159,17 @@ jl_value_t *jl_eqtable_get(jl_array_t *h, jl_value_t *key, jl_value_t *deflt)
 }
 
 JL_DLLEXPORT
+jl_value_t *jl_eqtable_get1(jl_array_t *h, jl_value_t *key, int *found)
+{
+    void **bp = jl_table_peek_bp(h, key);
+    if (found)
+        *found = (bp != NULL);
+    if (bp == NULL)
+        return jl_nothing;
+    return (jl_value_t *)*bp;
+}
+
+JL_DLLEXPORT
 jl_value_t *jl_eqtable_pop(jl_array_t *h, jl_value_t *key, jl_value_t *deflt, int *found)
 {
     void **bp = jl_table_peek_bp(h, key);
--- a/base/abstractdict.jl
+++ b/base/abstractdict.jl
@@ -500,6 +500,16 @@ function hash(a::AbstractDict, h::UInt)
     hash(hv, h)
 end
 
+function get(default::Callable, d::AbstractDict, key)
+    val = get(d, key)
+    val === nothing ? default() : something(val)
+end
+
+function get(d::AbstractDict, key, default)
+    val = get(d, key)
+    val === nothing ? default : something(val)
+end
+
 function getindex(t::AbstractDict, key)
     v = get(t, key, secret_table_token)
     if v === secret_table_token
--- a/base/compiler/compiler.jl
+++ b/base/compiler/compiler.jl
@@ -66,6 +66,7 @@ include("array.jl")
 include("abstractarray.jl")
 
 # core structures
+include("some.jl")
 include("bitarray.jl")
 include("bitset.jl")
 include("abstractdict.jl")
--- a/base/dict.jl
+++ b/base/dict.jl
@@ -488,9 +488,9 @@ julia> get(d, "c", 3)
 """
 get(collection, key, default)
 
-function get(h::Dict{K,V}, key, default) where V where K
+function get(h::Dict{K,V}, key) where V where K
     index = ht_keyindex(h, key)
-    @inbounds return (index < 0) ? default : h.vals[index]::V
+    @inbounds return (index < 0) ? nothing : Some{V}(h.vals[index]::V)
 end
 
 """
@@ -509,11 +509,6 @@ end
 ```
 """
 get(::Function, collection, key)
-
-function get(default::Callable, h::Dict{K,V}, key) where V where K
-    index = ht_keyindex(h, key)
-    @inbounds return (index < 0) ? default() : h.vals[index]::V
-end
 
 """
     haskey(collection, key) -> Bool
@@ -559,6 +554,11 @@ julia> getkey(D, 'd', 'a')
 function getkey(h::Dict{K,V}, key, default) where V where K
     index = ht_keyindex(h, key)
     @inbounds return (index<0) ? default : h.keys[index]::K
+end
+
+function getkey(h::Dict{K,V}, key) where V where K
+    index = ht_keyindex(h, key)
+    @inbounds return (index<0) ? nothing : Some{K}(h.keys[index]::K)
 end
 
 function _pop!(h::Dict, index)
@@ -769,12 +769,12 @@ function getindex(dict::ImmutableDict, key)
     end
     throw(KeyError(key))
 end
-function get(dict::ImmutableDict, key, default)
+function get(dict::ImmutableDict{K,V}, key) where {K,V}
     while isdefined(dict, :parent)
-        dict.key == key && return dict.value
+        dict.key == key && return Some{V}(dict.value)
         dict = dict.parent
     end
-    return default
+    return nothing
 end
 
 # this actually defines reverse iteration (e.g. it should not be used for merge/copy/filter type operations)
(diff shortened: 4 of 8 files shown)


rfourquet:
I have long wanted this and am happy to see this discussed after the old attempt with nullables in #18211. 

As a possible other meaning of two-args `get`: I'm currently writing a package with a couple implementation of multidicts, and while exploring possible APIs (still early stages of thinking), I currently implement, for a multidict `d`:
* `d[k]` returns an iterable of all the values associated with key `k`, possibly empty
* `d[k] = (1, 2, 3)` erase all old values of `k` and add new ones from the rhs iterable
* `get(d, k, default)` like for dicts, but returns only *one* of the possible values

This is not the place to discuss such an API, but in this case there is no function which throws when no key is found, so I though that `get(d, k)` _could_ be like `get(d, k, default)` except it would throw when `k` is no present in `d`. But for a normal dict, it would then be equivalent to `d[k]`. I think the meaning of `get` in this PR has my preference, but just wanted to share this alternative.

andyferris:
Reflecting on @rfourquet's comment I do have to say, there is a semantic that kinda works for dictionaries and a possible multi-dict. Instead of `Union{Nothing, Some{T}}` being returned from `get`, a container with zero or one elements could be returned, for example AcceleratedArrays.jl has a [`MaybeVector`](https://github.com/andyferris/AcceleratedArrays.jl/blob/master/src/MaybeVector.jl) which can be the [result](https://github.com/andyferris/AcceleratedArrays.jl/blob/master/src/UniqueHashIndex.jl#L34-L41) of `findall` (or `filter` or whatever) on a container that is known to have unique elements - in the case of `get` the uniqueness is in the `keys`.

One neat thing about this as a form "nullable" type is that many operations just fall out. "Optional chaining" can be done via broadcasting. "Nullish coalescing" is three-argument `get`. User's don't need to learn new concepts to deal with optional results (e.g. you can use `isempty` instead of `isnothing` - basically the API surface is reduced to just the container interface).

Anyway, I understand we settled on the `Union{Nothing, Some{T}}` pattern during Julia v0.7 discussions, but @rfourquet's comment reminded me of this. (I also realize this is a bit closer to the original `Nullable{T}`...)

(@rfourquet Multidict could have `getindex` require the key to be present, two argument `get` would be almost the same but give an empty vector instread of error when the key is not present... anyway just a thought)

andyferris:
Another strange idea: syntax sugar for two-argument `get` being `dict?[key]`.

tkf:
> `dict?[key]`

This is a neat syntax but I think it'd be incompatible with `f?(x)` **if** `f?(x)` is going to work on `Union{T, Missing}` instead of `Union{Some{T}, Nothing}`. https://discourse.julialang.org/t/status-of-question-mark-syntax-for-missing-values/27189/2

Having said that, I cannot help mentioning that there is a nice generalization of this in the case `dict?[key]` appears on both hand sides. For example,

```julia
dict?[key] = something(dict?[key], 0) + 1
```

can be lowered to

```julia
modify!(dict, key) do val
    something(val, 0) + 1
end
```

where `modify!` is the API I suggested in https://github.com/JuliaLang/julia/pull/33758 to do lookup/update/insert/delete in one go.

The signature of `modify!` is `modify!((Union{Some,Nothing} -> Union{Some,Nothing}), dict, key) -> Union{Some,Nothing}` and I think it's a nice counterpart of `get(dict, key) -> Union{Nothing,Some}`. In fact, I think pretty much all the getter/setter functions can be written in terms of `get` and `modify!`. This is one of the reasons why I think 2-argument `get` is a nice addition.

src/iddict.c
```diff
@@ -158,6 +158,17 @@ jl_value_t *jl_eqtable_get(jl_array_t *h, jl_value_t *key, jl_value_t *deflt)
     return (bp == NULL) ? deflt : (jl_value_t *)*bp;
 }
 
+JL_DLLEXPORT
+jl_value_t *jl_eqtable_get1(jl_array_t *h, jl_value_t *key, int *found)
+{
+    void **bp = jl_table_peek_bp(h, key);
+    if (found)
```
vtjnash:
```suggestion
```

rfourquet:
Maybe another point to consider: `pop!` and `get` are very similar in their 3-argument APIs, the difference being that `pop!` removes the element in addition to returning it. As `pop!(dict, key)` already exists and throws an error on inexistant key, that might support having the same for `get` (or eventually changing `pop!` to return `Union{Nothing,Some}` if `get` is made to do that).

Although I'm not convinced, `pop!` and `get` have different usage patterns.

andyferris:
We have that behaviour in `getindex`, which already throws if there is no key.  `get` is meant to be the “optional” variant of `getindex`. I suppose there could be a default to `pop!` - but it potentially becomes harder to reason about the mutation.

tkf:
@rfourquet For `Union{Nothing,Some}`-based (potentially) mutating API, I suggest `modify!` #33758. It's quite easy to "derive" other APIs from it:

```julia
maybepop!(dict, key) = modify!(_ -> nothing, dict, key)
```

This returns a `Union{Nothing, Some{valtype(dict)}}`.

@andyferris IIUC, @rfourquet is pointing out the similarity of the existing APIs `get!(dict, key, default)` and `pop!(dict, key, default)`.  If `get!(dict, key)` returns a `Union{Nothing,Some}`, it is reasonable to assume the same for `pop!(dict, key)` if you are new to this API.  However, `pop!(dict, key)` already exists and it throws.

I don't think we can do anything about `pop!(dict, key)` due to backward compatibility.

jw3126:
I think this would break quite some code. For instance, it will break [Setfield.jl](https://github.com/jw3126/Setfield.jl/issues/146) hard. I would guess we are not the only ones that overloaded `Base.get` in conflicting ways. I get (no pun intended) that this overload was somewhat daring, though. I am just begging for mercy :sweat_smile:

StefanKarpinski:
The life of a type pirate is fraught with peril!

jw3126:
> The life of a type pirate is fraught with peril!

I guess so. Just to clarify `Setfield.jl` overloads `get(::Any, ::Setfield.Lens)`. So this is as much piracy as `Base.*(::MyMatrix, ::AbstractMatrix)`. It hurts me, but I am ok with sacrificing the `get` overload for the greater good.

StefanKarpinski:
To be clear, I don’t have a strong opinion either way here and we should avoid causing undue breakage. Mostly just a joke.

tkf:
If we want `dict?[key]` and in general `container?[i, j, k]` to be the `Union{Some{T}, Nothing}` variant (and not `Union{T, Missing}`), I think we need something other than 2-arg `get` for this.  This is because we can't have vararg form `container?[i, j, k]` with `get` since it already has a specific 3-arg semantics.

I think @andyferris is designing "multi-dimensional" dictionary.  So, thinking about multi-dimensional key for "2-arg `get`" might be important.

jw3126:
Another possible name for this would be `probe(collection, key)`

cmcaine:
I'm probably being silly, but what's the point of this over `haskey(dict, key)`?

I'm imagining code like:

```julia
if haskey(dict, key) 
    value = dict[key]
    # do something
else
    # do something on failure
end
```

versus

```julia
x = get(dict, key)
if x !== nothing
    value = something(x)
else
    # do something on failure
end
```

There's one more hash going on in the first example, but maybe that could be optimised away?

andyferris:
> There's one more hash going on in the first example, but maybe that could be optimised away?

Unfortunately, optimizing a repeated hash lookup away seems unlikely to be feasible. There's just too much un-inlined code and Julia doesn't have the pure-functional semantics that might enable the detection of bulk repeated work. (It would also be nice to have an interface that presents the efficient path for all sorts of dictionaries - less round trips to a remote database, for example).

vtjnash:
In terms of API, perhaps this would be more consistent to call `tryget`, to go with the pattern set by `tryparse` and `trylock`, instead of `Base.maybeview` (to exhaustively list all functions that currently match that naming pattern)?

JeffBezanson:
I like `tryget` :+1:

tkf:
I'm suggesting to use `Union{Ok,Err}`-valued functions for this. See #45080

PR #34825, opened 2020-02-20
avoid leaving closure type from sysimg.jl in Main

JeffBezanson:
We currently have
```
julia> names(Main,all=true)
5-element Array{Symbol,1}:
 Symbol("#1#2")
 :Base
 :Core
 :InteractiveUtils
 :Main
```
This removes the first item to tidy things up a bit.


KristofferC:
I thought this was tested.

PR #34828, opened 2020-02-20
remove special cases for `Main` in system image saving

JeffBezanson:
`using`s and all bindings will now be preserved. I think the need for what we were doing is quite obsolete. This makes it easier to set up `Main` and make system images, since everything will come back exactly as you left it.

Also make Core a top-level module (its parent is itself) instead of setting its parent to Main.

### DIFF
--- a/src/staticdata.c
+++ b/src/staticdata.c
@@ -262,19 +262,15 @@ static void jl_serialize_module(jl_serializer_state *s, jl_module_t *m)
     for (i = 1; i < m->bindings.size; i += 2) {
         if (table[i] != HT_NOTFOUND) {
             jl_binding_t *b = (jl_binding_t*)table[i];
-            if (b->owner == m || m != jl_main_module) {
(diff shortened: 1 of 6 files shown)


vtjnash (approved):
🎉

JeffBezanson:
Ha, this also found a bug in `detect_unbound_args` on top-level modules.

src/staticdata.c
```diff
@@ -1265,6 +1246,11 @@ static void jl_reinit_item(jl_value_t *v, int how)
                 b += 1;
                 nbindings -= 1;
             }
+            if (mod->usings.items != &mod->usings._space[0]) {
+                void **newitems = (void**)malloc(mod->usings.max * sizeof(void*));
```
vtjnash:
```suggestion
                void **newitems = (void**)malloc_s(mod->usings.max * sizeof(void*));
```

PR #34832, opened 2020-02-21
Define g ⨟ f = f ∘ g

tkf:
This PR adds a simple definition `⨟(fs...) = ∘(reverse(fs)...)` to the new operator added in #34722.  This is useful for abstracting out pipelined operations like `x |> f |> g |> h` as `f ⨟ g ⨟ h`. For example:

```julia
mapping(f) = xs -> Base.Generator(f, xs)
filtering(f) = xs -> Iterators.filter(f, xs)
partitioning(n) = xs -> Iterators.partition(xs, n)
using Base.Iterators: flatten

# Iterator-to-iterator transformation to do some normalization and filtering
# with sliding window (a toy example):
ixf = partitioning(10) ⨟
    mapping(x -> (x , mean(x))) ⨟
    filtering(((_, m),) -> m > 0) ⨟
    mapping(((x, m),) -> x ./ m) ⨟
    flatten

collect(ixf(randn(100)))
```

cc @jw3126 @schlichtanders @ararslan

### DIFF
--- a/base/exports.jl
+++ b/base/exports.jl
@@ -193,6 +193,7 @@ export
     :,
     =>,
     ∘,
+    ⨟,
 
 # scalar math
     @evalpoly,
--- a/base/operators.jl
+++ b/base/operators.jl
@@ -833,6 +833,7 @@ entered in the Julia REPL (and most editors, appropriately configured) by typing
 Function composition also works in prefix form: `∘(f, g)` is the same as `f ∘ g`.
 The prefix form supports composition of multiple functions: `∘(f, g, h) = f ∘ g ∘ h`
 and splatting `∘(fs...)` for composing an iterable collection of functions.
+See also [`⨟`](@ref) for composition in the reversed order.
 
 !!! compat "Julia 1.4"
     Multiple function composition requires at least Julia 1.4.
@@ -863,6 +864,24 @@ function ∘ end
 ∘(f) = f
 ∘(f, g) = (x...)->f(g(x...))
 ∘(f, g, h...) = ∘(f ∘ g, h...)
+
+"""
+    g ⨟ f
+
+Compose functions in the "opposite order": i.e. `(g ⨟ f)(args...)` means `f(g(args...))`.
+Composition `g ⨟ f` is defined as [`f ∘ g`](@ref ∘).
+
+!!! compat "Julia 1.5"
+    Opposite composition `⨟` requires at least Julia 1.5.
+
+# Examples
+```jldoctest
+julia> 1 |> (x -> 2x) ⨟ string
+"2"
+```
+"""
+function ⨟ end
+⨟(fs...) = ∘(reverse(fs)...)
 
 """
     !f::Function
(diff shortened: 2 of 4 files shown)


ararslan:
I don't know what the symbol is

image

tkf:
In my browser:

image

ararslan:
`Char(0x2a1f)` doesn't render in my terminal either (same system: Ubuntu derivative). It'd be difficult for me to believe I'd be the only one with this issue, so it's difficult to support adding something I can only read in other people's screenshots. :stuck_out_tongue:

tkf:
I just learnt from https://github.com/JuliaLang/julia/issues/34835#issuecomment-589832898 that fileformat.info shows a list of fonts that supports a given unicode (within the fonts installed in their server).  Maybe useful as sampling points:

* U+2A1F `⨟` https://www.fileformat.info/info/unicode/char/2a1f/fontsupport.htm
* U+2218 `∘` https://www.fileformat.info/info/unicode/char/2218/fontsupport.htm

So yes, U+2218 definitely is supported by more fonts than U+2A1F.

FYI here is the list of non-ASCII names exported from `Base`:

```julia
julia> [x for x in (string(x) for x in names(Base)) if !isascii(x)]
26-element Array{String,1}:
 "÷"
 "π"
 "ℯ"
 "∈"
 "∉"
 "∋"
 "∌"
 "∘"
 "√"
 "∛"
 "∩"
 "∪"
 "≈"
 "≉"
 "≠"
 "≡"
 "≢"
 "≤"
 "≥"
 "⊆"
 "⊇"
 "⊈"
 "⊉"
 "⊊"
 "⊋"
 "⊻"
```

There are some variations in font support for these unicode characters. For example, U+212F `ℯ` is supported less than U+2218 `∘` but more than U+2A1F `⨟`:

* U+212F `ℯ` https://www.fileformat.info/info/unicode/char/212f/fontsupport.htm

laborg:
I can confirm that this is not shown on a vanilla ubuntu + firefox setup. IMHO the short list of supported fonts is good to estimator for the amount of confusion to expect from adding this.

Keno:
Somebody should file a request with whatever authority is in charge of font authoring for that combination ;).

andyferris:
I agree an operator with this semantic would be great; unfortunately, I also don't see this rendered correctly (in my browser or terminal).

jlapeyre:
Maybe it's an ubuntu quirk. It's rendered in my browsers and terminals. I'm running Arch, which doesn't give you anything out of the box.

JeffBezanson:
I knew you were up to something when you added this to the parser :)

Works fine for me on two ubuntu 18.04 systems. I hesitate to add an operator with potentially poor font coverage to Base though.

jw3126:
Works for me on ubuntu 18.04 + firefox

tkf:
Of courses, we can solve the font problem by defining something with ASCII :trollface:

schlichtanders:
font is correctly displayed for me on mac catalina 10.15.2 with all three, chrome, firefox and safari, with default font settings

adienes:
closing as an extrapolation of the recent triage discussion deciding to close several other PRs proposing "weird" new syntax, but speaking only as myself and not claiming to speak for triage

PR #34845, opened 2020-02-22
Clarify that eachrow and eachcol return vectors, add examples.

tpapp:
Some users expect matrices of single row or column from `eachrow` and `eachcol` (eg [this discussion](https://discourse.julialang.org/t/eachrow-over-array-any-2-does-not-return-rows-why/34601/)).

This trivial PR clarifies that the iteration is over vectors and adds some examples.

### DIFF
--- a/base/abstractarraymath.jl
+++ b/base/abstractarraymath.jl
@@ -411,7 +411,7 @@ end
     eachrow(A::AbstractVecOrMat)
 
 Create a generator that iterates over the first dimension of vector or matrix `A`,
-returning the row vectors as views.
+returning the rows as `AbstractVector` views.
 
 See also [`eachcol`](@ref) and [`eachslice`](@ref).
 
@@ -426,6 +426,11 @@ julia> a = [1 2; 3 4]
  1  2
  3  4
 
+julia> first(eachrow(a))
+2-element view(::Array{Int64,2}, 1, :) with eltype Int64:
+ 1
+ 2
+
 julia> collect(eachrow(a))
 2-element Array{SubArray{Int64,1,Array{Int64,2},Tuple{Int64,Base.Slice{Base.OneTo{Int64}}},true},1}:
  [1, 2]
@@ -439,7 +444,7 @@ eachrow(A::AbstractVecOrMat) = (view(A, i, :) for i in axes(A, 1))
     eachcol(A::AbstractVecOrMat)
 
 Create a generator that iterates over the second dimension of matrix `A`, returning the
-column vectors as views.
+columns as `AbstractVector` views.
 
 See also [`eachrow`](@ref) and [`eachslice`](@ref).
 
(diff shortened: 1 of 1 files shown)


andyferris (approved):

base/abstractarraymath.jl
```diff
@@ -411,12 +411,26 @@ end
     eachrow(A::AbstractVecOrMat)
 
 Create a generator that iterates over the first dimension of vector or matrix `A`,
-returning the rows as views.
+returning the row vectors as views.
```
mcabbott: I appreciate what you're trying to clarify, but I think this is confusing. Each row of the matrix gets returned, but as a vector, precisely not as a "row vector". 

Perhaps it's clearer just to say `view(r, :)` explicitly, something like
```suggestion
returning the rows as views, each a one-dimensional `view(r, :)` etc.
```

base/abstractarraymath.jl
```diff
@@ -411,12 +411,26 @@ end
...
+
+# Example
+
+```jldoctest
+julia> a = [1 2; 3 4]
+2×2 Array{Int64,2}:
+ 1  2
+ 3  4
+
+julia> collect(eachrow(a))
```
mcabbott: Likewise, I wonder if it wouldn't be clearer here to print out `first(eachrow(a))`, which will make more obvious the type of what gets returned.

base/abstractarraymath.jl
```diff
@@ -425,12 +439,26 @@ eachrow(A::AbstractVecOrMat) = (view(A, i, :) for i in axes(A, 1))
     eachcol(A::AbstractVecOrMat)
 
 Create a generator that iterates over the second dimension of matrix `A`, returning the
-columns as views.
+column vectors as views.
```
mcabbott: I would leave this alone.

base/abstractarraymath.jl
tpapp: I appreciate the further attempt at detail, but I don't think there is any room for confusion here, since "row vector" is not a type (there is no such thing in Julia), it is a vector as a row.

In any case, I included an example in both docstrings so everything should be clear to the reader and there should be absolutely no possibility of confusion. The primary focus of this PR is the examples.

mcabbott: There is no RowVector, sure, but I would say the term covers `Transpose{..., Vector}` and friends, 1 x N objects. And I thought the goal was to clarify that such objects never appear here, they play no role. 

"The rows" of a matrix are pre-existing things, but there is no pre-existing collection of "the row vectors" unless you were to make one, which this function does not.

tpapp: > but there is no pre-existing collection of "the row vectors"

The concept of [row/column vectors](https://en.wikipedia.org/wiki/Row_and_column_spaces) is generally used in linear algebra, and is the same as here.

andyferris: Hmm... I guess we use to have `RowVector` which wasn’t an `AbstractVevtor`... I can see it might get confusing on occasionally. 

Maybe “... returning the rows as `AbstractVector` views”.

base/abstractarraymath.jl
andyferris: Or show both?

tpapp: Thanks, did that. Individual views print nicer, `SubArray` may be a confusing implementation detail.

base/abstractarraymath.jl
tpapp: Thanks, that should be clear, I made that change.

tpapp:
Thanks for all the comments. I think it is as unambiguous now as it gets without risking overdocumenting.

timholy:
Thanks @tpapp and @mcabbott for a nice clarification!

PR #34850, opened 2020-02-23
Remove single-argument _chol!(A::StridedMatrix) in cholesky.jl

ShriprajwalK:


stdlib/LinearAlgebra/test/cholesky.jl
```diff
@@ -399,4 +399,17 @@ end
...
 
+@testset "covering _chol!(A::StridedMatrix)" begin
+    # Taking a hermitian, symmetric positive definite Matrix
+    b = [4 12 -16; 12 37 -43; -16 -43 98]
+    res_b = LinearAlgebra._chol!(b)
+    @test res_b[1] == [2 6 -8; 0 1 5; 0 0 3;]
+    @test res_b[2] != -1 # as b is hermitian
+    c = [2. 2+im 4.; 2-im 3 im; 4 -im 1.]
+    res_c = LinearAlgebra._chol!(c)
+    @test res_c[1] == Complex{Float64}[1.4142135623730951 + 0.0im 1.414213562373095 + 0.7071067811865475im 2.82842712474619 + 0.0im; 0.0 + 0.0im 0.7071067811865478 + 0.0im -5.656854249492377 + 4.242640687119283im; 0.0 + 0.0im 0.0 + 0.0im -56.99999999999994 + 0.0im]
```
vtjnash: Is the algorithm really this numerically stable?

andreasnoack:
```julia
julia> eigvals(c)
3-element Array{Float64,1}:
 -3.1651250644074382
  2.853079285169352
  6.312045779238089
```
The `c` matrix is not positive definite

andreasnoack (commented):
Did you check test coverage? I think this function is already covered by existing tests.

ShriprajwalK:
> Did you check test coverage? I think this function is already covered by existing tests.

I did check coveralls. This function is still in red.
https://coveralls.io/builds/28940782/source?filename=stdlib/LinearAlgebra/src/cholesky.jl
This is testing the function in line 151-156 which is in red. It is the third method of _chol!()

stdlib/LinearAlgebra/test/cholesky.jl
ShriprajwalK: The function has one behavior if the matrix A is hermitian, which a and b are testing.
I meant c to test what the function does if A is some arbitrary(not hermitian). I think I made a mistake with choosing c since it is not positive definite. Should I change it?

andreasnoack:
Just took a closer look. I think this function is dead code at this point. We used to have a `chol!` function which called it but it's gone. @fredrikekre would you be able to verify that it's no longer used anywhere?

ShriprajwalK:
Oh. So this method for the function should be deleted right?

dkarrasch:
AFAICT, there is no single-argument call of `_chol!` anywhere. All calls pass a `LowerTriangular` or `UpperTriangular` type argument as second argument.

dkarrasch:
This looks pretty good. Is there anything we need to take into account? Since this is not exported, I guess we don't need to consider external packages depending on this. Can we merge then?

ShriprajwalK:
Good to go from my side

fredrikekre (approved):

PR #34859, opened 2020-02-24
Loosen test of Cholesky factorization of diagonal matrices to approxi…

andreasnoack:
…mate equality since BLAS and Julia might use square roots that round differently.
This has occasionally caused CI failures.


fredrikekre (approved):

dkarrasch:
I'm merging this. The two CI failures happen much later than the tests.

PR #32762, opened 2019-08-01
Preserve structured matrix types more often when broadcasting

mbauman:
The primary motivation here is exponentiation (fixes JuliaLang/LinearAlgebra.jl#648). Previously `X .^ 2` would return a Matrix but `two=2; X .^ two` would return the same structured matrix type as `X`. This simply teaches LinearAlgebra a little bit more about broadcast so it can safely compute the zero-preservation property of broadcasts involving `Ref`s, which is sufficient to handle the literal pow argument list.

PR #34863, opened 2020-02-24
Always optimize Float.([32 64]) exponentiation by.(-1:3)

mbauman:
We perform these optimizations in the literal pow case, and LLVM does it when we happen to propagate constants for it, so I'd contend that they're just as worthwhile to do in the non-literal case. Calling LLVM's pow intrinsic takes ~16ns on my computer; adding these quick exits get compiled down to a simple switch statement that seems to add ~0.5ns in the cases where we don't hit the quick exits. In exchange, we get consistent numbers everywhere, they're better numbers at that (they're closer to the real answer), and these small common powers become essentially free.

(This was a hairy yak found over in #32762)

Edit: I should note I checked the other `methods(^, (Base.HWNumber, Integer))` to see if anyone else needed changing to match the `literal_pow` optimization; they all fell back the generic power by squaring which gets this right or were otherwise not applicable.

### DIFF
--- a/base/math.jl
+++ b/base/math.jl
@@ -896,8 +896,22 @@ end
     end
     z
 end
-@inline ^(x::Float64, y::Integer) = ccall("llvm.pow.f64", llvmcall, Float64, (Float64, Float64), x, Float64(y))
-@inline ^(x::Float32, y::Integer) = ccall("llvm.pow.f32", llvmcall, Float32, (Float32, Float32), x, Float32(y))
+@inline function ^(x::Float64, y::Integer)
+    y == -1 && return inv(x)
+    y == 0 && return one(x)
+    y == 1 && return x
+    y == 2 && return x*x
+    y == 3 && return x*x*x
+    ccall("llvm.pow.f64", llvmcall, Float64, (Float64, Float64), x, Float64(y))
+end
+@inline function ^(x::Float32, y::Integer)
+    y == -1 && return inv(x)
+    y == 0 && return one(x)
+    y == 1 && return x
+    y == 2 && return x*x
+    y == 3 && return x*x*x
+    ccall("llvm.pow.f32", llvmcall, Float32, (Float32, Float32), x, Float32(y))
+end
 @inline ^(x::Float16, y::Integer) = Float16(Float32(x) ^ y)
 @inline literal_pow(::typeof(^), x::Float16, ::Val{p}) where {p} = Float16(literal_pow(^,Float32(x),Val(p)))
 
--- a/test/floatfuncs.jl
+++ b/test/floatfuncs.jl
@@ -101,6 +101,67 @@ end
     @test round(Float16(1.1), sigdigits=70) === Float16(1.1)
 end
 
+@testset "literal pow matches runtime pow matches optimized pow" begin
+    two = 2
+    @test 1.0000000105367122^2 == 1.0000000105367122^two
+    @test 1.0041504f0^2 == 1.0041504f0^two
+
+    function g2(start, two, N)
+        x = start
+        n = 0
+        for _ in 1:N
+           n += (x^2 !== x^two)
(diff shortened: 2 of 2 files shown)


stevengj:
Why can't LLVM do this?

yuyichao:
I think the correct question to ask is why does the pow implementation not return the same result. Unless they do there's nothing llvm should do here as that won't be an optimization.I highly doubt any wall can helps (although aerodynamics is weird so I'm open for correction).

andyferris:
Would sufficient usage of this pattern essentially mean we won't need `literal_pow` at all? (Constant propagation will optimize the literal case further anyway...)

yuyichao:
The main reason for `literal_pow` was `x^-1` AFAIK. The optimization for other small literal power is a "bonus", if it can be called that....

andyferris:
Is it worth maybe adding -1 to the list here, then?

andyferris:
The other argument for it if I remember was around unitful calculations (constant prop clearly works for type inference in these cases now).

KristofferC:
And even when the branches are mispredicted it seems that the multiplications are fast enough to make up for it.

JeffBezanson:
Is there a concern that `x*x*x` incurs two roundoffs while `x^3.0` only has one? The literal_pow mechanism already does this though, so maybe not?

andyferris:
Well... users are free to choose between `x^3` or `x^3.0`, right?

dkarrasch:
This PR does not change how `x^3` and `x^3.0` are computed. It makes `x^3` and `three = 3; x^three`  being computed the exact same way.

mbauman:
Yeah, the wild thing is that LLVM _does_ do this in cases where it can statically see that the exponent is a constant (but Julia's parser wasn't able to).  This is what happens in cases like the test:

```Julia
    f(x, p) = x^p
    f2(x) = f(x, 2)
    x = 1.0000000105367122
     @test x^2 == f(x, 2) == f2(x) == x*x == Float64(big(x)*big(x))
```

For `x^2`, Julia's parser sees the constant; for `f2(x)` LLVM sees the constant (since it's inlined); for `x^two` and `f(x, 2)` (at global scope) neither can see it and it uses `pow`, which returns something different in this case.

JeffBezanson:
I get the same results in all those cases.

mbauman:
Hunh, that's strange. Here's what I get:

```Julia
julia> begin
       using Test
       f(x, p) = x^p
       f2(x) = f(x, 2)
       x = 1.0000000105367122
       @test x^2 == f(x, 2) == f2(x) == x*x == Float64(big(x)*big(x))
       end
Test Failed at REPL[6]:6
  Expression: x ^ 2 == f(x, 2) == f2(x) == x * x == Float64(big(x) * big(x))
   Evaluated: 1.0000000210734246 == 1.0000000210734243 == 1.0000000210734246 == 1.0000000210734246 == 1.0000000210734246
ERROR: There was an error during testing

julia> versioninfo()
Julia Version 1.5.0-DEV.391
Commit 4e609128ff (2020-03-03 07:30 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin18.7.0)
  CPU: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
```

mbauman:
Is this something we want?

mbauman:
> I get the same results in all those cases.

I was still curious about why Jeff wasn't seeing the test failure... there appears to be some OS-dependency in _which particular_ values fail.  The more thorough set of tests in this PR still fails on Linux.  And while I do agree that this is really something LLVM should do, these tests are still failing on Valentin's LLVM 10 branch.

So in short: with this branch we get more accurate results more consistently, it's _significantly_ faster for these common exponents, and results are more consistent across platforms.

PR #34864, opened 2020-02-25
x % Unsigned, x % Signed

JeffreySarnoff:
This follows from a discussion on Slack (exerpted below); the implementation is from @keno.

----
Matt Bauman:  I guess I’m in a gripey mood today.  Is there a reason why we don’t allow -1 % Unsigned?
```
julia> -1 % UInt
0xffffffffffffffff
julia> -1 % Unsigned
ERROR: InexactError: check_top_bit(Int64, -1) ...
```
Matt Bauman:  Is the “correct” way to do this `unsigned(-1)`?

Keno Fischer:  yes, but I think `% Unsigned` is reasonable.
...

Kristoffer Carlsson:  Why [is reinterpret] yuk? Isn't that the most descriptive function name to use? Who knows what semantics unsigned has?

Matt Bauman:  For some reason, reinterpret  feels like it imposes a higher cognitive burden — like I need to think about the exact bit representation.  All I care about in this operation is ensuring that negative values become absurdly large.  But it’s a moot point.

Keno Fischer:   In this case rem is just using the convert fallback. I think a definition like
```
rem(x::Signed, ::Type{Unsigned}) = x % unsigned(typeof(x))
```
 would be reasonable ... if somebody wants to submit a PR ... https://github.com/JuliaLang/julia/blob/master/base/int.jl#L508-L510

Stefan Karpinski:  @kristofferc: the reason I don’t like reinterpret is that it forces the type of the input and the result to have the same size but this is a perfectly well-defined operation even for different sizes.


base/int.jl
```diff
@@ -506,6 +506,7 @@ if nameof(@__MODULE__) === :Base
 end
 
 rem(x::T, ::Type{T}) where {T<:Integer} = x
+rem(x::Signed, ::Type{Unsigned}) = x % unsigned(typeof(x))
```
Keno: Should probably also have the opposite `x % Signed`.

JeffreySarnoff:
Covered News.md, src/int.jl, test/int.jl for both `x::Signed % Unsigned` and `x::Unsigned % Signed`.
When this builds, and there are no further notes, please mash-as-one these commits.

base/int.jl
JeffreySarnoff: done and tests are also done

test/int.jl
```diff
@@ -266,6 +266,14 @@ end
     U in [Base.BitInteger_types..., BigInt]
     @test typeof(rand(U(0):U(127)) % T) === T
 end
+@testset "x::Signed % Unsigned returns a T, T = $T" for T in [Base.BitSigned_types...]
+    @test typeof(rand(T) % Unsigned) <: Unsigned
```
Keno: I'd rather this test a few interesting cases instead (or in addition). For example `typemax`, `typemin`, `-typemax`. Also should test for `x` a BigInt, since that's an interesting one for which there is no unsigned type.

JeffreySarnoff: What do you think? Should `BigInt(x) % Unsigned` always throw an error, or should it handle values that only require 128 signed bits and throw and error for larger values?

JeffreySarnoff: I was surprised that `unsigned(Int64) == UInt64` but `signed(UInt64)` errors.  We should remedy that, defining `signed(x::T) where {T<:Base.BitUnsigned_types}` to be the size corresponding `BitSigned_types` in multinverses.jl (where the `unsigned(type{T})` are defined.

JeffreySarnoff: Why support `x::BigInt % Unsigned`? It doesn't really make sense, because `BigInt` is not a bitstype and these conversions are from one n-byte integer type (signed or unsigned) to another n-byte integer type (unsigned or signed).
The error that happens `.. no method matching Unsigned(::BigInt)` seems appropriate,

StefanKarpinski: Fixed-size integer types represent modular classes of integer values. When you do `x % T` you are saying "give me the element of that represents the modular class of `x` in `T`. That is a perfectly well-defined operation for `x::BigInt % Unsigned`. In fact, this operation is _always_ well-defined because `x` always represents a specific integer value and that value always belongs to the modular class represented by a unique element of `T`.

rfourquet: But in `x::BigInt % Unsigned`, how to choose the destination type? We won't define an `UnsignedBigInt` type. And we can't return `x` because it's expected that `(x % Unsigned) isa Unsigned`.

StefanKarpinski: Oh, I see what you mean—I interpreted that as being for some specific `T <: Unsigned` but the issue is for the literal type `Unsigned`. That is quite unclear so agree about it being an error. The alternative would be to choose `UInt` as the "default unsigned type" but that's questionable.

rfourquet: > The alternative would be to choose UInt as the "default unsigned type" but that's questionable.

I guess it's always possible to add it later if there is demand for that, but while this makes some sense, this seems a bit weird to return a smaller type than `Int128(1) % Unsigned`.

StefanKarpinski: Agree. Which leads to an issue if/when we ever add `(U)Int256` types.

JeffreySarnoff:
With Base and this PR with or without (U)Int256 bitstypes:

- x::Signed
    - `x === x % Signed`
- x::Unsigned
    - `x % Unsigned`

- x::BitInteger
    - `isa(x % Signed, Signed) & isa(x % Unsigned, Unsigned)`
    - `sizeof(x) === sizeof(x % Signed) === sizeof(x % Unsigned)`
- x::BitUnsigned
    - `isa(x, Unsigned) & isa(x, BitInteger)`
    - `x === x % Signed % Unsigned`
- x::BitSigned
    - `isa(x, Signed) & isa(x, BitInteger)`
    - `x === x % Unsigned % Signed`

JeffreySarnoff:
With Base and this PR with or without (U)Int256 bitstypes:

- x::Signed
    - `x === x % Signed`
- x::Unsigned
    - `x % Unsigned`

- x::BitInteger
    - `isa(x % Signed, Signed) & isa(x % Unsigned, Unsigned)`
    - `sizeof(x) === sizeof(x % Signed) === sizeof(x % Unsigned)`
- x::BitUnsigned
    - `isa(x, Unsigned) & isa(x, BitInteger)`
    - `x === x % Signed % Unsigned`
- x::BitSigned
    - `isa(x, Signed) & isa(x, BitInteger)`
    - `x === x % Unsigned % Signed`

Keno:
I think it's worth discussing this on triage, particularly for the BigInt case.

JeffreySarnoff:
Do those "tester" failures have anything to do with this PR?

base/int.jl
```diff
@@ -44,6 +44,22 @@ const BitUnsigned64T   = Union{Type{UInt8}, Type{UInt16}, Type{UInt32}, Type{UIn
 
 const BitIntegerType = Union{map(T->Type{T}, BitInteger_types)...}
 
+# interconvert equilength BitInteger types
+
+unsigned(::Type{Int8}) = UInt8
```
Keno: These should already be defined. Why is this necessary?

JeffBezanson:
From triage: `bigint % Signed` should be the identity, and `bigint % Unsigned` should be a method error. We should test those.

base/int.jl
JeffreySarnoff: `unsigned(::Type{IntN}) = UIntN` are defined in `Base/multinverses.jl`.
`signed(::Type{UIntN}) = IntN` are not defined.
I thought the two sets of definitions belonged in the same place, and that `Int.jl` was more appropriate than `multinverses.jl`.  I moved the defs from `multinverses.jl` to `Int.jl` and added the defs for `signed(_)` there.  I added `unsigned` to the `using Base: ` line in `multinverses.jl`.

I can revert this and I could add the `signed(_)` defs in that file, if that is more appropriate.

JeffreySarnoff:
I may have this together (the conflict is in the News file, I don't know how to deal with that).
The requested adds were added.  The suggested removes/restores have been done.
We have had `unsigned(::Type{<:BitSigned})` mapping the type to the same-sized BitUnsigned type.  The dual `signed(::Type{<:BitUnsigned})` has been missing.  I added that in `Int.jl`, in addition to the other microtasks.

JeffreySarnoff:
@Keno does this now comport with the need?

NEWS.md
```diff
@@ -66,7 +66,10 @@ New library features
 * `isapprox` (or `≈`) now has a one-argument "curried" method `isapprox(x)` which returns a function, like `isequal` (or `==`)` ([#32305]).
 * `Ref{NTuple{N,T}}` can be passed to `Ptr{T}`/`Ref{T}` `ccall` signatures ([#34199])
 * `accumulate`, `cumsum`, and `cumprod` now support `Tuple` ([#34654]).
+* `x::Signed % Unsigned` and `x::Unsigned % Signed` are supported for integer bitstypes.
+* `signed(unsigned_type)` is supported for integer bitstypes, `unsigned(signed_type)` has been supported.
 
+>>>>>>> master
```
mbauman:
```suggestion
```

JeffreySarnoff:
I don't know how to fix the NEWS.md merge conflict.  I thought I did, yet what I did did not work. Also the white space issue appears to be
``` stdlib/LinearAlgebra/test/lu.jl:346:@testset "testing Array(F::LU{T,Tridiagonal{T,V})" begin 
Error: trailing whitespace found in source file(s)
```

JeffreySarnoff:
:smile:

JeffreySarnoff:
error is unrelated
```
Error in testset Serialization:
Error During Test at none:1
  Test threw exception
  Expression: Serialization
  ProcessExitedException(3)
```
Is there anything more for me to do?

JeffreySarnoff:
lil help please -- is this right, if not what is wrong

base/int.jl
JeffreySarnoff: all that is undone

StefanKarpinski:
What's the status of this? Done? Waiting for CI to pass?

JeffreySarnoff:
This is done and it is good to go now

NEWS.md
```diff
@@ -92,6 +92,9 @@ New library features
 
 * `isapprox` (or `≈`) now has a one-argument "curried" method `isapprox(x)` which returns a function, like `isequal` (or `==`)` ([#32305]).
 * `Ref{NTuple{N,T}}` can be passed to `Ptr{T}`/`Ref{T}` `ccall` signatures ([#34199])
+* `accumulate`, `cumsum`, and `cumprod` now support `Tuple` ([#34654]).
```
rfourquet: Delete this line, which must come from rebasing (the correct version is indicated 3 lines below)

test/int.jl
```diff
@@ -267,6 +267,25 @@ end
     @test typeof(rand(U(0):U(127)) % T) === T
 end
 
+@testset "signed(<:Unsigned), unsigned(<:Signed) for bitstypes" begin
+    for (S,U) in zip(Base.BitSigned_types, Base.BitUnsigned_types)
+        @test signed(U) === S
+        @test unsigned(S) === U
+    end
+end
+@testset "x::[Un]Signed % [Un]Signed returns x" begin
```
rfourquet: Just personal taste, but I would join the three `@testsets` together, as they contain the same loop, so it's a bit of a ceremony overhead.

rfourquet:
This might need updating the docstrings (I know that `unsigned(::Type)` is currently undocumented, but this PR publicizes the feature in the NEWS.md).

JeffreySarnoff:
@rfourquet your two suggestions are implemented. If there is some "merge conflict" from NEWS.md, please reach in and adjust it -- I have tried a few times. thanks.

JeffreySarnoff:
I added the docstring for `signed(T)` where e.g `signed(::Type{UInt8}) = Int8` appears.
I do not know where the corresponding `unsigned(T)` are, so I added that docstring before the `signed(T)` section using `""" ... """ unsigned`.

JeffreySarnoff:
@StefanKarpinski there it be ✅ .. please merge before something else changes :))

PR #34868, opened 2020-02-25
Add two-argument `first` and `last` methods for any iterable

giordano:
For reference: https://discourse.julialang.org/t/extend-last-to-also-work-with-arrays/35100.

Contrary to what was asked in the discourse thread, this only applies to vectors, as I think this is the most uncontroversial option.  If applied to a generic array, the first/last n elements would be flattened, which may or may not what users expect.

### DIFF
--- a/base/abstractarray.jl
+++ b/base/abstractarray.jl
@@ -362,7 +362,7 @@ julia> first(Bool[], 1)
 0-element Array{Bool,1}
 ```
 """
-first(v::AbstractVector, n::Integer) = @inbounds v[firstindex(v):min(firstindex(v) - 1 + n, end)]
+first(v::AbstractVector, n::Integer) = @inbounds v[firstindex(v):min(firstindex(v) + n - 1, end)]
 
 """
     last(coll)
@@ -401,7 +401,7 @@ julia> last(Float64[], 1)
 0-element Array{Float64,1}
 ```
 """
-last(v::AbstractArray, n::Integer) = @inbounds v[max(firstindex(v), end + 1 - n):end]
+last(v::AbstractArray, n::Integer) = @inbounds v[max(firstindex(v), end - n + 1):end]
 
 """
     strides(A)
--- a/base/abstractarray.jl
+++ b/base/abstractarray.jl
@@ -344,6 +344,27 @@ function first(itr)
 end
 
 """
+    first(v::AbstractVector, n::Integer)
+
+Get the first n elements of the vector, without going out-of-bounds.
+
+# Examples
+```jldoctest
+julia> first(["foo", "bar", "qux"], 2)
+2-element Array{String,1}:
+ "foo"
+ "bar"
+
+julia> first(1:6, 10)
+1:6
+
+julia> first(Bool[], 1)
+0-element Array{Bool,1}
+```
+"""
+first(v::AbstractVector, n::Integer) = @inbounds v[firstindex(v):min(firstindex(v) - 1 + n, end)]
+
+"""
     last(coll)
 
 Get the last element of an ordered collection, if it can be computed in O(1) time. This is
@@ -360,6 +381,27 @@ julia> last([1; 2; 3; 4])
 ```
 """
 last(a) = a[end]
+
+"""
+    last(v::AbstractVector, n::Integer)
+
+Get the last n elements of the vector, without going out-of-bounds.
+
+# Examples
+```jldoctest
+julia> last(["foo", "bar", "qux"], 2)
+2-element Array{String,1}:
+ "bar"
+ "qux"
+
+julia> last(1:6, 10)
+1:6
+
+julia> last(Float64[], 1)
+0-element Array{Float64,1}
+```
+"""
+last(v::AbstractArray, n::Integer) = @inbounds v[max(firstindex(v), end + 1 - n):end]
 
 """
     strides(A)
--- a/base/abstractarray.jl
+++ b/base/abstractarray.jl
@@ -344,14 +344,15 @@ function first(itr)
 end
 
 """
-    first(v::AbstractVector, n::Integer)
+    first(itr, n::Integer)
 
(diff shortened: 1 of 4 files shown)


base/abstractarray.jl
```diff
@@ -343,6 +343,27 @@ function first(itr)
     x[1]
 end
 
+"""
+    first(v::AbstractVector, n::Integer)
+
+Get the first n elements of the vector, without going out-of-bounds.
```
nalimilan: Maybe something like this?
```suggestion
Get the first `n` elements of vector `v`, or fewer elements if `v` is not long enough.
```

base/abstractarray.jl
```diff
@@ -361,6 +382,27 @@ julia> last([1; 2; 3; 4])
 """
 last(a) = a[end]
 
+"""
+    last(v::AbstractVector, n::Integer)
+
+Get the last n elements of the vector, without going out-of-bounds.
```
nalimilan:
```suggestion
Get the last `n` elements of vector `v`, or fewer elements if `v` is not long enough.
```

test/abstractarray.jl
```diff
@@ -999,3 +999,15 @@ end
 @testset "vcat with mixed elements" begin
     @test vcat(Nothing[], [missing], [1.0], [Int8(1)]) isa Vector{Union{Missing, Nothing, Float64}}
 end
+
+@testset "first/last n elements of vector" begin
```
nalimilan: Could you also add tests for non-1-based indices in test/offsetarray.jl?

test/abstractarray.jl
```diff
@@ -999,3 +999,15 @@ end
...
+    v = [1, 13, 42]
+    @test first(v, -2) == []
+    @test first(v, 2) == v[1:2]
+    @test first(v, 100) == v
```
nalimilan: Just in case:
```suggestion
    @test first(v, 100) == v
    @test first(v, 100) !== v
```

test/abstractarray.jl
```diff
@@ -999,3 +999,15 @@ end
...
+    @test first(v, 1) != v[1]
+    @test last(v, -2) == []
+    @test last(v, 2) == v[end-1:end]
+    @test last(v, 100) == v
```
nalimilan:
```suggestion
    @test last(v, 100) == v
    @test last(v, 100) !== v
```

nalimilan (commented):
Thanks!

test/abstractarray.jl
giordano: Done!

nalimilan:
Thanks! Looks good, apart from one decision to make: should negative indices be allowed or not? Currently they are not allowed for `first` and `last` on strings, so maybe the same should apply for arrays.

giordano:
> Currently they are not allowed for first and last on strings, so maybe the same should apply for arrays.

Uhm, that comes automatically from the use of `nextind`,  the same check here would make the function slightly more complicated :pensive:

base/abstractarray.jl
```diff
@@ -343,6 +343,27 @@ function first(itr)
...
+ "bar"
+
+julia> first(1:6, 10)
+1:6
+
+julia> first(Bool[], 1)
+0-element Array{Bool,1}
+```
+"""
+first(v::AbstractVector, n::Integer) = @inbounds v[firstindex(v):min(firstindex(v) + n - 1, end)]
```
stevengj: You can use `begin` instead of `firstindex`

stevengj:
We might support this for arbitrary iterators via Iterators.reverse, in which case it would support multidimensional arrays too.

base/abstractarray.jl
giordano: Nice, now the two definitions are much more symmetric

nalimilan:
@stevengj Do you have an opinion about whether we should allow negative number of elements or not?

stevengj:
What would negative numbers of elements mean?

stevengj:
General iterator definitions:
```jl
first(itr, n::Integer) = collect(Iterators.take(itr, n))
last(itr, n::Integer) = reverse!(Iterators.take(Iterators.reverse(itr), n))
```

nalimilan:
> What would negative numbers of elements mean?

In the current state of the PR they are equivalent to `0`. That doesn't very useful, but that's the simplest implementation.

> `Iterators.take` throws an error for a negative argument, so I'm inclined to throw a similar error here.

Yeah, sounds like a good idea.

base/abstractarray.jl
```diff
@@ -361,6 +382,27 @@ julia> last([1; 2; 3; 4])
...
+```jldoctest
+julia> last(["foo", "bar", "qux"], 2)
+2-element Array{String,1}:
+ "bar"
+ "qux"
+
+julia> last(1:6, 10)
+1:6
+
+julia> first(Float64[], 1)
```
j-n-e: See typo

giordano: Which typo?

j-n-e: This is the examples section for `last`

Moelf: that the last line should be using `last` not `first`

StefanKarpinski:
Anything we can do to move this forward? Hate to see a PR languishing like this.

giordano:
Ok, I'll get back this weekend

giordano:
I believe I've addressed all comments above and CI seems happy

base/abstractarray.jl
```diff
@@ -343,6 +343,33 @@ function first(itr)
...
+julia> first(1:6, 10)
+1:6
+
+julia> first(Bool[], 1)
+Bool[]
+```
+"""
+first(itr, n::Integer) = collect(Iterators.take(itr, n))
+# Faster method for vectors
+function first(v::AbstractVector, n::Integer)
```
stevengj, line 368: Why is this method for `AbstractVector`, but the corresponding specialized `last` method below is for `AbstractArray`?

Issue #34682, opened 2020-02-06
`typeinf_code` documentation is wrong?


GiggleLiu:
I think its API has been changed, but I don't know how to fix the documentation

The link to the documentation

https://docs.julialang.org/en/v1/devdocs/inference/

```julia
julia> # Get the method
       atypes = Tuple{Type{Int}, UInt}  # argument types
Tuple{Type{Int64},UInt64}

julia> mths = methods(convert, atypes)  # worth checking that there is only one
# 1 method for generic function "convert":
[1] convert(::Type{T}, x::Number) where T<:Number in Base at number.jl:7

julia> m = first(mths)
convert(::Type{T}, x::Number) where T<:Number in Base at number.jl:7

julia> # Create variables needed to call `typeinf_code`
       params = Core.Compiler.Params(typemax(UInt))  # parameter is the world age,
Core.Compiler.Params(Core.Compiler.InferenceResult[], 0x000000000000688f, true, true, true, false, 100, 1000, 400, 4, 4, 8, 3, 32)

julia> #   typemax(UInt) -> most recent
       sparams = Core.svec()      # this particular method doesn't have type-parameters
svec()

julia> optimize = true            # run all inference optimizations
true

julia> cached = false             # force inference to happen (do not use cached results)
false

julia> Core.Compiler.typeinf_code(m, atypes, sparams, optimize, cached, params)
ERROR: MethodError: no method matching typeinf_code(::Method, ::Type{Tuple{Type{Int64},UInt64}}, ::Core.SimpleVector, ::Bool, ::Bool, ::Core.Compiler.Params)
Closest candidates are:
  typeinf_code(::Method, ::Any, ::Core.SimpleVector, ::Bool, ::Core.Compiler.Params) at compiler/typeinfer.jl:510
Stacktrace:
 [1] top-level scope at REPL[19]:1

julia> Core.Compiler.typeinf_code(m, atypes, sparams, optimize, params)
(CodeInfo(
    @ number.jl:7 within `convert'
1 ─     ($(Expr(:static_parameter, 1)))(x)::Union{}
└──     $(Expr(:unreachable))::Union{}
), Union{})
```

The correct inferenced type should be `Int64`.
```
julia> @code_typed convert(Int, UInt(1.0))
CodeInfo(
1 ─ %1 = Core.lshr_int(x, 63)::UInt64
│   %2 = Core.trunc_int(Core.UInt8, %1)::UInt8
[...]

PR #34869, opened 2020-02-25
Fixed the compiler.jl and inlining algorithm demos in documentation

vjhaanpaa:
Issue #34682 noted that the below demo from [Inference documentation](https://docs.julialang.org/en/v1/devdocs/inference/) gives a `MethodError`:

```julia
# Get the method
atypes = Tuple{Type{Int}, UInt}  # argument types
mths = methods(convert, atypes)  # worth checking that there is only one
m = first(mths)

# Create variables needed to call `typeinf_code`
params = Core.Compiler.Params(typemax(UInt))  # parameter is the world age,
                                                        #   typemax(UInt) -> most recent
sparams = Core.svec()      # this particular method doesn't have type-parameters
optimize = true            # run all inference optimizations
cached = false             # force inference to happen (do not use cached results)
Core.Compiler.typeinf_code(m, atypes, sparams, optimize, cached, params)
```

The specific error message is what is shown in the `REPL` when the last line is run, which complains about an additional `::Bool` in `Core.Compiler.typeinf_code()`: 

```julia
julia> Core.Compiler.typeinf_code(m, atypes, sparams, optimize, cached, params)
ERROR: MethodError: no method matching typeinf_code(::Method, ::Type{Tuple{Type{Int64},UInt64}}, ::Core.SimpleVector, ::Bool, ::Bool, ::Core.Compiler.Params)
Closest candidates are:
  typeinf_code(::Method, ::Any, ::Core.SimpleVector, ::Bool, ::Core.Compiler.Params) at compiler/typeinfer.jl:510
Stacktrace:
 [1] top-level scope at REPL[8]:1
```

In the issue, @GiggleLiu guesses that the API has been changed, and gives the below example for the latest API, in which the keyword `cached` is dropped from `Core.Compiler.typeinf_code()`.

```julia
julia> Core.Compiler.typeinf_code(m, atypes, sparams, optimize, params)
(CodeInfo(
    @ number.jl:7 within `convert'
1 ─     ($(Expr(:static_parameter, 1)))(x)::Union{}
└──     $(Expr(:unreachable))::Union{}
), Union{})
```

Dropping `cached` fixes the demo for `compiler.jl` in the documentation, and this pull request makes that change to `inference.md`. 

Additionally, the presented output of the inlining algorithm demo in the same file was corrected.


doc/src/devdocs/inference.md
```diff
@@ -109,26 +108,11 @@ cst = map(cost, ci.code)
 
 # output
 
-31-element Array{Int64,1}:
```
KristofferC: Why change this?

vjhaanpaa: If you run the demo, you get only a 5-element array. Try running this in `REPL`: 

```julia
# A demo on `fill(3.5, (2, 3))
f = fill
tt = Tuple{Float64, Tuple{Int,Int}}
# Create the objects we need to interact with the compiler
params = Core.Compiler.Params(typemax(UInt))
mi = Base.method_instances(f, tt)[1]
ci = code_typed(f, tt)[1][1]
opt = Core.Compiler.OptimizationState(mi, params)
# Calculate cost of each statement
cost(stmt::Expr) = Core.Compiler.statement_cost(stmt, -1, ci, opt.sptypes, opt.slottypes, opt.params)
cost(stmt) = 0
cst = map(cost, ci.code)
```

The current documentation lists a five-element array, and I am not entirely sure why it has been changed to a 31-element array. If I am wrong, the second demo can be just ignored, but the `doctest` failed there as well due to a reason that I don't 100% understand.

KristofferC: You might (likely) run it on a version of Julia that is different from the one where the doctests are run on (master).

doc/src/devdocs/inference.md
```diff
@@ -31,8 +31,7 @@ params = Core.Compiler.Params(typemax(UInt))  # parameter is the world age,
                                                         #   typemax(UInt) -> most recent
 sparams = Core.svec()      # this particular method doesn't have type-parameters
```
vtjnash:
```suggestion
sparams = Core.svec(Int64)      # this particular method has a type-parameter
```

doc/src/devdocs/inference.md
```diff
@@ -31,8 +31,7 @@ params = Core.Compiler.Params(typemax(UInt))  # parameter is the world age,
                                                         #   typemax(UInt) -> most recent
```
vtjnash: This is wrong now—the `typemax(UInt)` world doesn't exist.
```suggestion
params = Core.Compiler.Params(Base.get_world_counter())
```

doc/src/devdocs/inference.md
vtjnash: Running this code sample would previously have permanently corrupted the internal state of the system due to this

vtjnash:
Thanks for looking into this! There was even more problems as well, which I've noted the corrections for.

vjhaanpaa:
> Thanks for looking into this! There was even more problems as well, which I've noted the corrections for.

Thanks for your feedback. I looked into the suggestion by @KristofferC about running the second demo on an out-of-date version, and I had indeed mixed my terminal windows and `julia` instances. I'll pull together the changes and make a new commit.

vjhaanpaa:
Okay, I'm now certainly running `Version 1.5.0-DEV.339 (2020-02-25)` and there are still discrepancies with the second demo. Any suggestions?

ViralBShah:
@KristofferC Is this ok to merge?

ViralBShah:
Bump @KristofferC or should it be someone else?

PR #34878, opened 2020-02-25
Revert "Added test for dumptype with a Concrete and Abstract type."

Keno:
I don't have time at the moment to look into the actual test failure, but obviously we need to get CI back, so this reverts JuliaLang/julia#34764. If somebody else comes up with a proper fix first, obviously merge that instead.


vtjnash (approved):
FWIW, the difference is:
Prior Julia:
```
julia> parentmodule(Core)
Main
```
after secondary change included in #34828:
```
julia> parentmodule(Core)
Core
```

And that the InteractiveUtils code assumes that all modules are reachable from Main (e.g. it doesn't know about `Base.loaded_modules_array()` like it should).

PR #34879, opened 2020-02-25
Makefile wildcard globs only work at end

twhitehead:
The following doesn't work when `test` is one of the stdlibs (e.g., `REPL`)
```bash
make test-<test>
```
as the original glob does not expand. This replacement implements the same logic as is found in `choosetests.jl` (check for the existance of _$lib/src/$lib.jl_) to determine what stdlib tests are available.

Note that I have currently only tested this against the 1.3.1 tree.


Keno:
This PR came in while the buildbots were quite unstable. Closing/Re-opening to re-run the bots.

twhitehead:
Noticed I failed to remove an earlier version I had come up with that also worked (expanding `/*/` gives everything, but includes a trailing `/` on directories) but didn't match what `choosetests.jl` did. Corrected.

PR #34881, opened 2020-02-26
Add file name to all top level line number nodes

c42f:
This allows julia code to more easily consume the output of `jl-parse-all` and is consistent with the way line number nodes have file name information in all non-toplevel constructs.

See various mentions in #34715, eg https://github.com/JuliaLang/julia/pull/34715/files#discussion_r383156452.

### DIFF
--- a/src/jlfrontend.scm
+++ b/src/jlfrontend.scm
@@ -66,10 +66,7 @@
                            ;; for error, get most recent line number (#16720)
                            (lineno (if iserr (input-port-line io) lineno))
                            (next   (list* expr
-                                          ;; include filename in first line node
-                                          (if (null? exprs)
-                                              `(line ,lineno ,(symbol filename))
-                                              `(line ,lineno))
+                                          `(line ,lineno ,current-filename)
                                           exprs)))
                       (if iserr
                           (cons 'toplevel (reverse! next))


DilumAluthge:
Could be nice to have a test?

DilumAluthge:
Although, we will definitely have a test in #34715, which will indirectly test this.

c42f:
There's no test harness for the flisp code, so we'd be testing from Julia anyway. The tests for `parsefile` seem as good a place to do that as any.

DilumAluthge:
> There's no test harness for the flisp code, so we'd be testing from Julia anyway. The tests for `parsefile` seem as good a place to do that as any.

Sounds good!

c42f:
In the meantime, I get the following:

before:
```julia
julia> Base.parse_input_line("a\nb\nc", filename="X")
:($(Expr(:toplevel, :(#= X:1 =#), :a, :(#= line 2 =#), :b, :(#= line 3 =#), :c)))
```

after:
```julia
julia> Base.parse_input_line("a\nb\nc", filename="X")
:($(Expr(:toplevel, :(#= X:1 =#), :a, :(#= X:2 =#), :b, :(#= X:3 =#), :c)))
```

JeffBezanson (approved):

PR #34883, opened 2020-02-26
Update reflection.jl

githubtomtom:
make `fieldnames()` and `fieldname()` directly available for objects.


JeffBezanson:
I'm not in favor of this, since types are themselves objects. You don't know whether `fieldname(x, i)` will operate on `x` or its type, which in the past has been confusing and I've tried to eliminate where posible.

githubtomtom:
ok. thanks.

githubtomtom:
by the way, is `f(x) = (@_inline_meta; g(x) )` the correct way to inline a single-lined-function in `Base`? 
or, should I write `@inline f(x) = g(x)` ? thanks.

KristofferC:
`@inline` unless it needs to be added before that macro itself is defined

Keno:
Closing per @JeffBezanson's comment, which I agree with.

Issue #34870, opened 2020-02-25
@distributed doesnt work with OffsetArrays


oheil:
Distributed and @distributed doesn't work for arrays with arbitrary indexes.
Example:
```
using Distributed
using OffsetArrays

array=rand(10)

#works:
@distributed for index in eachindex(array)
    #do something
end

#doesn't work:
oa=OffsetArray(array,0:9)
@distributed for index in eachindex(oa)
    #do something
end
```

The reason, as far as I can analyze, is, that function 
`splitrange `
in 
`share\julia\stdlib\v1.3\Distributed\src\macros.jl`
expects indices starting with 1.

This is the error which is produced by above example:
```
julia> @sync @distributed for index in eachindex(oa)
           #do something
       end
ERROR: TaskFailedException:
BoundsError: attempt to access 10-element Base.OneTo{Int64} at index [2:11]
throw_boundserror(::Base.OneTo{Int64}, ::Tuple{UnitRange{Int64}}) at .\abstractarray.jl:537
checkbounds at .\abstractarray.jl:502 [inlined]
getindex at .\range.jl:671 [inlined]
getindex at C:\Users\USER\.julia\packages\OffsetArrays\fZSaL\src\axes.jl:148 [inlined]
(::var"#7#8")(::OffsetArrays.IdOffsetRange{Int64,Base.OneTo{Int64}}, ::Int64, ::Int64) at D:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.5\Distributed\src\macros.jl:301
#160 at D:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.5\Distributed\src\macros.jl:87 [inlined]
run_work_thunk(::Distributed.var"#160#162"{UnitRange{Int64},var"#7#8",OffsetArrays.IdOffsetRange{Int64,Base.OneTo{Int64}}}, ::Bool) at D:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.5\Distributed\src\process_messages.jl:79
run_work_thunk at D:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.5\Distributed\src\process_messages.jl:88 [inlined]
(::Distributed.var"#96#98"{Distributed.RemoteValue,Distributed.var"#160#162"{UnitRange{Int64},var"#7#8",OffsetArrays.IdOffsetRange{Int64,Base.OneTo{Int64}}}})() at .\task.jl:358
Stacktrace:
[...]

PR #34886, opened 2020-02-26
make Distributed independent of indices starting at 1

oheil:
Proposal to solve issue https://github.com/JuliaLang/julia/issues/34870

### DIFF
--- a/stdlib/Distributed/src/macros.jl
+++ b/stdlib/Distributed/src/macros.jl
@@ -262,8 +262,7 @@ end
 
 # Statically split range [firstIndex,lastIndex] into equal sized chunks for np processors
 function splitrange(firstIndex::Int, lastIndex::Int, np::Int)
-    each = div(lastIndex-firstIndex+1,np)
-    extras = rem(lastIndex-firstIndex+1,np)
+    each, extras = divrem(lastIndex-firstIndex+1, np)
     nchunks = each > 0 ? np : extras
     chunks = Vector{UnitRange{Int}}(undef, nchunks)
     lo = firstIndex
@@ -280,7 +279,7 @@ function splitrange(firstIndex::Int, lastIndex::Int, np::Int)
 end
 
 function preduce(reducer, f, R)
-    chunks = splitrange(firstindex(R),lastindex(R), nworkers())
+    chunks = splitrange(Int(firstindex(R)), Int(lastindex(R)), nworkers())
     all_w = workers()[1:length(chunks)]
 
     w_exec = Task[]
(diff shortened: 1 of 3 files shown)


stdlib/Distributed/src/macros.jl
```diff
@@ -260,9 +260,27 @@ function splitrange(N::Int, np::Int)
...
+        chunks[i] = lo:hi
+        lo = hi+1
+    end
+    return chunks
+end
+
 function preduce(reducer, f, R)
-    N = length(R)
-    chunks = splitrange(Int(N), nworkers())
+    chunks = splitrange(firstindex(R),lastindex(R), nworkers())
```
JeffBezanson:
```suggestion
    chunks = splitrange(Int(firstindex(R)), Int(lastindex(R)), nworkers())
```

Please also generally use `, ` (comma and space) consistently between arguments.

JeffBezanson:
Thanks! Seems good to me.

stdlib/Distributed/src/macros.jl
```diff
@@ -260,9 +260,27 @@ function splitrange(N::Int, np::Int)
     return chunks
 end
 
+# Statically split range [firstIndex,lastIndex] into equal sized chunks for np processors
+function splitrange(firstIndex::Int, lastIndex::Int, np::Int)
+    each = div(lastIndex-firstIndex+1,np)
+    extras = rem(lastIndex-firstIndex+1,np)
```
timholy:
```suggestion
    each, extras = divrem(lastIndex-firstIndex+1, np)
```

timholy (commented):
Looks good

oheil:
changed as suggested

timholy:
Thanks. I just realized this doesn't have a test, could you add one? There's a mock implementation of OffsetArrays, see https://github.com/JuliaLang/julia/blob/c2cd601f01849aca85c843f3eb277a0342fee0a6/test/offsetarray.jl#L3-L4

oheil:
Yes, but it will take a few days.

The original function splitrange(N,np) is still in the code. I will check if it used somewhere and if not, should I remove it? What do you think?
Or it is better to mark it debrecated? Can you give me hint on how to deprecate code properly?

timholy:
If it's not used and not exported, I suspect you can just delete it; it doesn't seem to be a likely callee by downstream packages.

If you do decide to deprecate it, try something like
```julia
@deprecate splitrange(N::Int, np::Int) splitrange(1, N, np)
```
If you want the warning to be more informative about why `1` is required, you could mimic https://github.com/JuliaLang/julia/blob/c2cd601f01849aca85c843f3eb277a0342fee0a6/base/deprecated.jl#L219-L225. (Of course this is a function rather than a macro, the only line to pay attention to is the call to `Base.depwarn`.)

oheil:
Removed old version of splitrange and added minimal tests for new splitrange

oheil:
Done. Ready for review.

Issue #34887, opened 2020-02-26
show for UnionAll types prints non-syntactic identifiers for unnamed type variables


NHDaly:
Calling `Base.show()` with `UnionAll` types with unnamed type parameters prints those type parameters with non-syntactic generated names:
```julia
julia> Vector{<:Bool}
Array{#s16,1} where #s16<:Bool
```
This of course cannot be evaluated, which can be difficult when trying to understand a very large type that's printed in an error message, for example.

My first thought was that this should be fixed following the same approach as in https://github.com/JuliaLang/julia/pull/32408: we should change these to print via the `var""` syntax, a la:
```julia
julia> Vector{<:Bool}
Array{var"#s16",1} where var"#s16"<:Bool
```

That seems like a fine approach, but then I noticed that if the variable _does_ have a name, the printing mechanism seems to be able to tweak the user-provided name to prevent conflicts with other identifiers in the type! This leads me to wonder if we can use that same smarts to just pick a standard identifier name, rather than a purposefully non-standard identifier as above?
For example, notice how the user-provided name `T` gets renamed to `T1` when there's a conflict:
```julia
julia> NTuple
Tuple{Vararg{T,N}} where T where N

julia> NTuple{N,NTuple} where N
Tuple{Vararg{Tuple{Vararg{T,N}} where T where N,N}} where N

julia> NTuple{1,NTuple{T}} where T
Tuple{Tuple{Vararg{T1,T}} where T1} where T
```

Can we similarly have julia just pick some arbitrary type names like `p1`, `p2`, `pN`... or `P1`, `P2`, ... or `_T1`, `_T2`, etc, rather than the `#s16`, `#s17` type names we're using now? And then if there's a conflict with an actual user-provided name, it can just adjust the name like it already does?
In fact, the existing mechanism already handles conflicts even for the generated names, so this should be a very easy change: just simplifying the generated names to be valid identifiers.
```julia
julia> Vector{<:Vector{<:Bool}}
Array{#s16,1} where #s16<:(Array{#s17,1} where #s17<:Bool)

julia> Vector{<:Vector{var"#s16"}} where var"#s16" <: Bool
Array{#s17,1} where #s17<:Array{#s16,1} where #s16<:Bool

julia> Vector{<:Vector{var"#s17"}} where var"#s17" <: Bool
Array{#s171,1} where #s171<:Array{#s17,1} where #s17<:Bool
```

NHDaly:
[...]

PR #32408, opened 2019-06-25
Syntax for non-standard identifiers

c42f:
Introduce syntax for non-standard identifiers. Features:

* Have a direct way to print hidden symbols (eg, from macro expansion and closure conversion)
* Directly write "weird" identifiers from other languages (for example R allows `.` as an identifier character, XRef https://github.com/JuliaInterop/RCall.jl/issues/71#issuecomment-146000972)
* Possibly distinguish between identifiers vs keywords in the future (see #26819)

Various syntax choices were explored (see below), with `var"ident"` currently chosen. Though this has the familiar visual form of a string macro, special parser support was required to make it valid in every context where a normal identifier is allowed. Therefore, this is technically a breaking change. However, the use here has exactly the same semantics as the use in RCall which is the only package I could find on github which defines `@var_str`.

`show` has been changed to use this syntax as necessary when printing expressions:
* Fixes #32354 by printing macro identifiers using `var"@ident"` rather than as a macro *call* `@ident`.
* Fixes printing of symbols generated by macro hygiene, for example `var"#89#err"` rather than the bare `#89#err`
* Simplify printing of functions and type names which are generated by the compiler.

PR #34888, opened 2020-02-26
Change TypeVars in UnionAlls to print as var"" syntax

NHDaly:
This changes unnamed type variables to print as valid syntax, via the `var""` printing macro introduced in #32408, as part of addressing #34887.

For example, this makes `Vector{<:Bool}` print as a valid, syntactic expression:
```julia
julia> Vector{<:Bool}
Array{var"#s46",1} where var"#s46"<:Bool
```

I'm not totally in love with this approach, since it makes the types much more verbose, but it's a good first step towards resolving #34887. I don't feel confident enough to delve into the type creation to change the type variable's actual name when instantiating the UnionAll to be a valid identifier, as proposed in #34887.

### DIFF
--- a/base/show.jl
+++ b/base/show.jl
@@ -1820,16 +1820,16 @@ function show(io::IO, tv::TypeVar)
     lb, ub = tv.lb, tv.ub
     if !in_env && lb !== Bottom
         if ub === Any
-            write(io, tv.name)
+            show_unquoted(io, tv.name)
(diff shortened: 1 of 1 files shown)


JeffBezanson:
Agreed on all counts --- this is a good improvement, but it would be better to use valid identifiers to begin with.

JeffBezanson (approved):

PR #34891, opened 2020-02-26
no-op for copyto!(dst, emptysrc) 

### LINKED ISSUE #34889: Bug in copyto!
The following code currently fails on master:
```
julia> copyto!([1,2], view(Int[],Int[]))
ERROR: BoundsError: attempt to access 2-element Array{Int64,1} at index [0]
Stacktrace:
 [1] throw_boundserror(::Array{Int64,1}, ::Tuple{Int64}) at ./abstractarray.jl:537
 [2] checkbounds at ./abstractarray.jl:502 [inlined]
 [3] copyto!(::Array{Int64,1}, ::SubArray{Int64,1,Array{Int64,1},Tuple{Array{Int64,1}},false}) at ./multidimensional.jl:975
 [4] top-level scope at REPL[23]:1
```

The reason is here: https://github.com/JuliaLang/julia/blob/master/base/multidimensional.jl#L975
and is most likely caused by https://github.com/JuliaLang/julia/pull/34049.

The problem is that empty source is incorrectly handled in this part of code.

CC @pablosanjose @timholy

pablosanjose:
Fixes #34889

This makes `copyto!(dst, src)` a no-op when `isempty(src)`. The shortcut does *not* check whether the eltypes of `dst` and the empty `src` can be converted one to the other. This is the current behaviour of Base, e.g. `copy!([1,2], String[])` does not throw.


pablosanjose:
(sorry about the commit mess, I'm not sure how I arrived at this, but I am not able to do a squash now, for some reason)

timholy:
Thanks for the quick fix!

PR #31214, opened 2019-02-28
More efficient indexing by CartesianIndices

mbauman:
Treat CartesianIndices as its tuple of ranges when indexing. This allows views that use CartesianIndices as indices to be strided and changes the complexity of boundschecking from O(n^d) to O(d)

Issue #24086, opened 2017-10-10
What to do about multivalue setindex!?


mbauman:
## Problem

We have two ways to assign many values to many locations in an array:

    A[2:3] = 4:5
    A[2:3] .= 4:5

In this simple case, both of these do the same thing for any array: assign `A[2] = 4` and `A[3] = 5`.  But in more complicated cases, both syntaxes are quirky in different ways:

### `setindex!`: `A[I...] = X`

`X` must have the same number of elements as the indices `I` select.  Shapes are compared, but singleton dimensions are ignored and the final dimension of `X` is allowed to linearly span multiple indices.  If `X` is not a subtype of `AbstractArray`, then the same value is simply broadcast to all indices.  Examples:

```
A = rand(4,1,3,2);
A[:,:,:,:] = 1:24 # Linear spanning
A[:,:,:,:] = reshape(1:24, 4,3,2) # Ignoring singleton dimensions in I
A[:,:,:,:] = reshape(1:24, 4,3,1,2) # Ignoring singleton dimensions in both X and I
A[:,:,:,:] = reshape(1:24, 4,6) # Ignoring singleton dimensions in I and linear spanning
A[:,:,:,:] = 1 # fills A with ones
```

Now, if `I` only selects exactly *one* index then this syntax always puts `X` into that location:

```
julia> A = Any[1,2,3];
       A[1] = 4:5;
       A
3-element Array{Any,1}:
  4:5
 2
 3
```

This is quirky when you actually want to generically assign a single value to multiple locations… and you don't care about its type!  Ref. #22747.

### `broadcast`: `A[I...] .= X`

Broadcasting dictates that the shape of the indexing expression and the shape of `X` have exactly the same shape if they have the same number of elements.  Absolutely no linear spanning is allowed.  All of the `setindex!` examples above are errors except for the scalar `1` case.  At the same time, though, `.=` adds the full set of broadcast features, allowing partial arrays to broadcast across the complete set of indices.

Now, if `I` only selects exactly *one* index, then this syntax is quirky:

```
julia> A = [[1,2,3],4:5,6]
       A[1] .= 0
       A
3-element Array{Any,1}:
 0
  4:5
 6

julia> A = [[1,2,3],4:5]
       A[1] .= 0
       A
2-element Array{AbstractArray{Int64,1},1}:
 [0, 0, 0]
 4:5
```

[...]

PR #34893, opened 2020-02-26
Fix assignment to 0-dim array indexed by CartesianIndices{0}

jmert:
The change in PR #31214 replaced `CartesianIndices` in the indexing tuple with their constituent ranges. This broke assignment to 0-dimensional arrays which have been indexed by a `CartesianIndices{0}`, e.g.
```julia
julia> A = fill(0.0);

julia> I = CartesianIndices(size(A));

julia> x = fill!(similar(A), 1.0);

# On v1.3
julia> A[I] = x
0-dimensional Array{Float64,0}:
1.0

# On master
julia> A[I] = x
ERROR: MethodError: Cannot `convert` an object of type Array{Float64,0} to an object of type Float64
...
```

This is because previously the `CartesianIndices{0}` (hereafter `CI0`) was preserved through the call to `to_indices()`:
```julia
# On v1.3
julia> to_indices(A, (I,))
(CartesianIndex(),)

julia> to_indices(A, (I,I))
(CartesianIndex(), CartesianIndex())

julia> to_indices(A, (I,I,I))
(CartesianIndex(), CartesianIndex(), CartesianIndex())
```
but now the empty tuple is returned instead:
```julia
# On master
julia> to_indices(A, (I,))
()

julia> to_indices(A, (I,I))
()

julia> to_indices(A, (I,I,I))
()
```
and that changes which internal implementation of `setindex!` is dispatched to.

This PR fixes the issue by tweaking the tuple reduction performed by `to_indices()`. The result is that
during the tuple reduction, a case where only [trailing] `CI0`s remain is shortcut and returns just a single `CI0` index.

With this PR:
```julia
julia> A[I] = x
0-dimensional Array{Float64,0}:
1.0

julia> to_indices(A, (I,))
(fill(CartesianIndex()),)

julia> to_indices(A, (I,I))
(fill(CartesianIndex()),)

julia> to_indices(A, (I,I,I))
(fill(CartesianIndex()),)
```

The one thing I couldn't figure out is how to avoid an unnecessary trailing `CI0` if a concrete index precedes it — (my attempts lead to `to_indices()` method ambiguities) — but there's always at most a single unnecessary trailing `CI0`.

```julia
julia> B = [1.0]
1-element Array{Float64,1}:
 1.0

julia> to_indices(B, (1, I))
(1, fill(CartesianIndex()))

julia> to_indices(B, (1, I, I))
(1, fill(CartesianIndex()))

julia> to_indices(B, (I, 1, I))
(1, fill(CartesianIndex()))

julia> to_indices(B, (I, I, 1, I))
(1, fill(CartesianIndex()))

julia> to_indices(B, (I, I, I, 1))
(1,)
```

### DIFF
--- a/base/multidimensional.jl
+++ b/base/multidimensional.jl
@@ -381,12 +381,12 @@ module IteratorsMD
     first(iter::CartesianIndices) = CartesianIndex(map(first, iter.indices))
     last(iter::CartesianIndices)  = CartesianIndex(map(last, iter.indices))
 
-    # Collapse trailing CartesianIndices{0} elements to a single instance. (Also
-    # maintains at least one for 0-dimensional indexed assignment.)
-    @inline to_indices(A, inds, I::Tuple{CartesianIndices{0},Vararg{CartesianIndices{0}}}) = (first(I),)
     # When used as indices themselves, CartesianIndices can simply become its tuple of ranges
     @inline to_indices(A, inds, I::Tuple{CartesianIndices, Vararg{Any}}) =
         to_indices(A, inds, (I[1].indices..., tail(I)...))
+    # but preserve CartesianIndices{0} as they consume a dimension.
+    @inline to_indices(A, inds, I::Tuple{CartesianIndices{0},Vararg{Any}}) =
+        (first(I), to_indices(A, inds, tail(I))...)
 
     @inline function in(i::CartesianIndex{N}, r::CartesianIndices{N}) where {N}
         _in(true, i.I, first(r).I, last(r).I)
--- a/base/multidimensional.jl
+++ b/base/multidimensional.jl
@@ -381,6 +381,9 @@ module IteratorsMD
     first(iter::CartesianIndices) = CartesianIndex(map(first, iter.indices))
     last(iter::CartesianIndices)  = CartesianIndex(map(last, iter.indices))
 
+    # Collapse trailing CartesianIndices{0} elements to a single instance. (Also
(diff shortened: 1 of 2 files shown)


jmert:
I found another case that has regressed, and this PR right now doesn't fix.
```
julia> I = CartesianIndices(())
0-dimensional CartesianIndices{0,Tuple{}}:
CartesianIndex()

# On v1.3
julia> setindex!(zeros(2), fill(1.0), I, 1)
2-element Array{Float64,1}:
 1.0
 0.0

# On master
julia> setindex!(zeros(2), fill(1.0), I, 1)
ERROR: MethodError: Cannot `convert` an object of type Array{Float64,0} to an object of type Float64
...
```

jmert:
It appears simply preserving all instances of `CartesianIndices(())` in the indexing tuple fixes the case in my previous comment. I've added a few new tests for that as well.

mbauman (approved):

mbauman:
Ah, very nice, thank you!  There's an ambiguity when there are no indices — was this a scalar assignment or a non scalar assignment?  https://github.com/JuliaLang/julia/issues/24086 is the root cause here, but this is a great fix.  Thanks!

jmert:
Correct me if I'm wrong, but I don't think this PR needs to be backported since #31214 was only merged after the v1.4 branch (assuming I'm reading the tag/branch "breadcrumb" in Github's view of f579d146 correctly).

mbauman:
Right you are.  Thanks again!

PR #34854, opened 2020-02-23
Missing static test

bramtayl:
Rebase #34523

Issue #24825, opened 2017-11-28
LoadError when throwing exception from macro


andreasnoack:
On 
```
Version 0.7.0-DEV.2675 (2017-11-27 20:28 UTC)
Commit f383276c5b (1 day old master)
```
I get
```julia
julia> macro myerror()
           throw(ErrorException(""))
       end
@myerror (macro with 1 method)

julia> @myerror()
ERROR: LoadError:
Stacktrace:
 [1] @myerror(::LineNumberNode, ::Module) at ./REPL[3]:2
in expression starting at REPL[4]:1
```

yuyichao:
Addict this is expected. Your error message is empty which is why it doesn't show up

andreasnoack:
Why does it throw a `LoadError`? It didn't do that on 0.6?

yuyichao:
https://github.com/JuliaLang/julia/pull/23533

andreasnoack:
Thanks. Is this
```julia
julia> @test_throws ErrorException eval(:(@myerror()))
Test Failed at REPL[26]:1
  Expression: eval($(Expr(:quote, :(#= REPL[26]:1 =# @myerror))))
    Expected: ErrorException
      Thrown: LoadError
ERROR: There was an error during testing
```
then an issue with `@test_throws`? I noticed that I can catch the `eval`ed version and not the non-evaled, i.e.
```julia
julia> try eval(:(@myerror()))
       catch
       end

julia> try @myerror()
       catch
       end
ERROR: LoadError: 
Stacktrace:
 [1] @myerror(::LineNumberNode, ::Module) at ./REPL[12]:2
in expression starting at REPL[30]:1
```

The `LoadError` is also a bit unintuitive given the documentation which states that
> An error occurred while includeing, requireing, or using a file. The error specifics should be
  available in the .error field.

and I'm probably not the only user for whom it is not obivous where to look for the `include`, `require`, or `using`.

PR #34895, opened 2020-02-26
add load=false option to atmacroexpand1

bramtayl:
This is a useful for testing whether macros parsed correctly. Ref https://github.com/JuliaLang/julia/pull/34854#discussion_r383507425

### DIFF
--- a/base/expr.jl
+++ b/base/expr.jl
@@ -151,17 +151,36 @@ With `@macroexpand` the expression expands where `@macroexpand` appears in the c
 With `macroexpand` the expression expands in the module given as the first argument.
 """
 macro macroexpand(code)
-    return :(macroexpand($__module__, $(QuoteNode(code)), recursive=true))
+    return :(macroexpand($__module__, $(QuoteNode(code))))
 end
 
-
 """
-    @macroexpand1
+    @macroexpand1 code [load = true]
 
-Non recursive version of [`@macroexpand`](@ref).
+Non recursive version of [`@macroexpand`](@ref). If `load` is set to false, will
+not load code, but instead, call the macro as a function.
+
+```jldoctest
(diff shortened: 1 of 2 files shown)


StefanKarpinski:
This would be one of the only things in the `Test` module that doesn't start with `test_` ...

StefanKarpinski:
It's also a bit weird that it has such a generic name when it is very specific: it parses a single macro call expression. Would be good to see some places in the test suite where this would clean up the code.

fredrikekre:
Could maybe just be a `raise=false` argument to `@macroexpand`? (Similar to `Meta.parse` etc.)

bramtayl:
> Could maybe just be a raise=false argument to @macroexpand? (Similar to Meta.parse etc.)

Ooh I like that better. I'll give that a shot.

bramtayl:
On further reflection, maybe `@macroexpand` throwing a load error is a bug? Compare:

```
macro test(x)
    x == "bad" && error("bad")
    :good
end
# same result
(@test "bad")
(@macroexpand @test "bad")
# different result
@test "hello world"
@macroexpand @test "hello world"
```

bramtayl:
The docs for LoadError say

An error occurred while includeing, requireing, or using a file. The error specifics should be available in the .error field.

Which all sound like eval to me, and `@macroexpand` doesn't eval...

bramtayl:
Hmm seems LoadErrors are intended based on #24825, so I've used @frederickekre s suggestion and added an option to `@macroexpand1`

bramtayl:
I don't know why the arch failed to build and I'm torn between opening a new issue, retriggering CI, and crying.

bramtayl:
Replaced by #34895

PR #34901, opened 2020-02-27
More tests for Bidiagonal

kshyatt:
Checked with `@which` that these should call the appropriate `mul!` functions.

Had to invent my sketchy type to check this path of `transpose`.

### DIFF
--- a/stdlib/LinearAlgebra/test/bidiag.jl
+++ b/stdlib/LinearAlgebra/test/bidiag.jl
@@ -498,4 +498,22 @@ end
     end
 end
 
+struct MyNotANumberType
+    n::Float64
+end
+Base.zero(n::MyNotANumberType)      = MyNotANumberType(zero(Float64))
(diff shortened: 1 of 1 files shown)


dkarrasch (commented):
The first part looks great. For the second part, I guess you want to have an eltype that is not a number. Not that I mind your type, but why not use 2x2 matrices? You could construct the "ground truth" for the transpose then as
```julia
Bidiagonal(transpose.(a), transpose.(b), :L)
```
if I'm not mistaken.

kshyatt:
Tried that, didn't work bc the `zero` method failed.

dkarrasch:
I see. No, it's not a bug, it's because the usual `Array`s don't have a size, so it would be unclear what the "zero element" of that should be. So I guess this is good then.

kshyatt:
Yay. Can you press the approve button and I'll merge after CI passes?

dkarrasch (approved):

Issue #34902, opened 2020-02-27
Threads.nthreads() not a compile-time constant.


MasonProtter:
Since the number of threads can't be changed without restarting julia, I find this behavior quite surprising:
```julia
julia> @code_llvm Threads.nthreads()
;  @ threadingconstructs.jl:19 within `nthreads'
define i64 @julia_nthreads_20335() {
top:
; ┌ @ pointer.jl:105 within `unsafe_load' @ pointer.jl:105
   %0 = load i32, i32* inttoptr (i64 140333758273848 to i32*), align 8
; └
; ┌ @ boot.jl:707 within `Int64'
; │┌ @ boot.jl:626 within `toInt64'
    %1 = sext i32 %0 to i64
; └└
  ret i64 %1
}
```
In general, I think it would be useful to do something like 
```julia
const NUM_THREADS = Int(unsafe_load(cglobal(:jl_n_threads, Cint)))
nthreads() = NUM_THREADS
```
so that branches / dispatches which depend on the number of threads can be eliminated.

KristofferC:
Code doesn't have to be recompiled just because you restart julia.

MasonProtter:
I'm not sure I understand. Are you saying this causes problems with static compilation?

yuyichao:
Making nthread a constant is going in the wrong direction so we should not have this. And yes this cause problems with compilation.

I also don't believe there are many valuable cases that benefits from a static dispatch on nthread.

MasonProtter:
> I also don't believe there are many valuable cases that benefits from a static dispatch on nthread.

using `@spawn` or `@threads` induces a lot of overhead, so often one wants to do things like 
```
if nthreads() > 1
    f(x)
else
    threaded_f(x)
end 
```
There are cases where even the presence of a `@spawn` in a branch that doesn't get hit prevents optimizations of a function body, so in a tight kernel you often want to separate those out into separate function calls. Because `nthreads()` gets recomputed each call, the above `if` block won't get elided. 

If we eventually do make machinery for changing the number of threads at runtime, it should behave like function redefinition and require you to hit the global scope for it to take effect.

yuyichao:
Assuming you `if` is not in a loop this won't matter. If it is then you should move it out.
The single thread overhead for `@thread` pretty much all come from https://github.com/JuliaLang/julia/issues/15276 apart from the cases that is fixed in a PR aready. Even if it is defined in your way it won't help anything.

[...]

PR #34903, opened 2020-02-27
Make nthreads() a compile time constant

MasonProtter:
ref https://github.com/JuliaLang/julia/issues/34902


yuyichao (changes requested):
No this is very wrong. If you really want to have this as a optimization in certain cases this should be done in codegen.

KristofferC:
This would cache the value when julia is built. Not good.

PR #34905, opened 2020-02-27
LU to Factorization tests

kshyatt:
Just making sure that the eltype conversion one works.

### DIFF
--- a/stdlib/LinearAlgebra/test/lu.jl
+++ b/stdlib/LinearAlgebra/test/lu.jl
@@ -3,7 +3,7 @@
 module TestLU
 
 using Test, LinearAlgebra, Random
-using LinearAlgebra: ldiv!, BlasInt, BlasFloat, rdiv!
+using LinearAlgebra: ldiv!, BlasReal, BlasInt, BlasFloat, rdiv!
 
 n = 10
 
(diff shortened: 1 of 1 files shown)


stdlib/LinearAlgebra/test/lu.jl
```diff
@@ -72,6 +72,11 @@ dimg  = randn(n)/2
             bft = eltya <: Real ? LinearAlgebra.LU{BigFloat} : LinearAlgebra.LU{Complex{BigFloat}}
             bflua = convert(bft, lua)
             @test bflua.L*bflua.U ≈ big.(a)[p,:] rtol=ε
+            @test Factorization{eltya}(lua) == lua
```
dkarrasch: If this is meant to be a no-op in the case of matching eltypes, maybe strengthen this to `===` comparison?

stdlib/LinearAlgebra/test/lu.jl
```diff
@@ -72,6 +72,11 @@ dimg  = randn(n)/2
...
+            # test Factorization with different eltype
+            if eltya <: BlasReal
+                @test Array(Factorization{Float16}(lua)) ≈ Array(lu(convert(Matrix{Float16}, a)))
```
dkarrasch: What exactly do you want to test here? The l.h.s. is even approx to the original `a`, so somehow you're not yet capturing that the conversion of the factorization was effective. Maybe (also) test the eltype of some factor or something like that?

kshyatt: I think this question is only answerable by my highly sleep deprived Thursday self... I'll update :)

kshyatt: OK, tried with `==` and the test failed due to what appear to be rounding issues.

PR #34908, opened 2020-02-27
REPL: make DictCompletion context module aware

aviatesk:
Makes `DictCompletion`s aware of a context-module.

Certain tools (like Juno, Debugger.jl etc) use `REPLCompletions` in a way that completions are generated in a context of a module other than `Main`, and can benefit from this change.
Not breaking at all, since `context_module` defaults to `Main`.

### DIFF
--- a/stdlib/REPL/src/REPLCompletions.jl
+++ b/stdlib/REPL/src/REPLCompletions.jl
@@ -518,7 +518,7 @@ function bslash_completions(string, pos)::Tuple{Bool, Completions}
     return (false, (Completion[], 0:-1, false))
 end
 
-function dict_identifier_key(str,tag)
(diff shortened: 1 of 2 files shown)


aviatesk:
@pfitzseb @StefanKarpinski @KristofferC any of you could give a review on this ? This change doesn't change any behaviour of usual REPL process (where we currently don't have module switching support), but should help other REPL environments where `REPLCompletions.completions` is used.

StefanKarpinski:
I've never touched `context_module` before so probably not the best reviewer. @pfitzseb, you seem to have committed code that did something with `context_module` most recently. Would you mind having a look at this and merging if it looks good to you?

StefanKarpinski:
Thanks, @KristofferC!

aviatesk:
thanks you guys !

Issue #33712, opened 2019-10-29
string `length` performance trap


JeffBezanson:
I recently saw some code that used `while i <= length(s)` where `s` is a `String`. This can happen e.g. when porting code that assumes all data is ASCII. Obviously one answer is "don't do that", but it would be nice if we could handle this better. Two ideas:

- Tell LLVM that `length(::String)` is pure and can be hoisted out of loops.
- Use the top bit of the length field to store whether a string is ASCII, speeding up `length` and `isascii` in that case.

Keno:
For the top bit, last time we discussed this, I had suggested that `length` and `isascii` could automatically set that if applicable the first time they're run. Of course we'd need to update things like egality and it'd make String even more weird of an object type, but probably worthwile. Of course telling LLVM it can LICM it is good in either case.

PR #34912, opened 2020-02-28
RFC: IteratorSize(::Type{<:AbstractString}) = SizeUnknown()

tkf:
Currently, `IteratorSize` on `AbstractString` is `HasLength` due to the generic fallback:

```
julia> @which Base.IteratorSize(String)
Base.IteratorSize(::Type) in Base at generator.jl:91
```

Since `length` on strings is not O(1), it may be better to return `SizeUnknown()` from string-related iterators.  But I don't do much string processing so I'm not sure if this is the best choice.  What do you think?  Is it a reasonable change?

### DIFF
--- a/base/strings/basic.jl
+++ b/base/strings/basic.jl
@@ -148,6 +148,7 @@ See also: [`getindex`](@ref), [`checkbounds`](@ref)
 
 ## basic generic definitions ##
 
+IteratorSize(::Type{<:AbstractString}) = SizeUnknown() # length is not O(1)
 eltype(::Type{<:AbstractString}) = Char # some string types may use another AbstractChar
 
 """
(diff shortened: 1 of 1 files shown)


vtjnash (approved):

base/strings/basic.jl
```diff
@@ -715,6 +717,7 @@ size(s::CodeUnits) = (length(s),)
 elsize(s::CodeUnits{T}) where {T} = sizeof(T)
 @propagate_inbounds getindex(s::CodeUnits, i::Int) = codeunit(s.s, i)
 IndexStyle(::Type{<:CodeUnits}) = IndexLinear()
+IteratorSize(::Type{<:CodeUnits}) = SizeUnknown()
```
JeffBezanson, line 720: CodeUnits is a vector, not a string, and its length should always be known in O(1).

JeffBezanson:
I disagree; knowing the length of an iterator is often a large performance win. Also, there may be AbstractString subtypes whose length is known in O(1) and we don't want to hurt their performance.

It's true that typically we don't want O(n) length methods, but I think strings are a corner case. Even though it's O(n) it can be computed very efficiently (and we should do better, e.g. #33712, #33988), and having a specific length just seems like a very important part of the notion of a string.

tkf:
OK, I think it makes sense.

PR #34916, opened 2020-02-28
Use typed-zero in sign() implementation

### LINKED ISSUE #34914: sign(::Float64) has more complicated Julia IR in v1.3 than v1.2
The typed/optimized IR for `sign(::Float64)` is much more complicated in v1.3 than it was in v1.2. (Despite that, though, LLVM does simplify both down to the same LLVM IR, so I don't suppose there's any real performance issue — it just made reading through a long function a bit harder during debugging of some of my own code since I initially didn't understand where the large floating point constants were coming from.)

```julia
# In v1.2
julia> @code_typed sign(2.0)
CodeInfo(
1 ─ %1  = Base.lt_float(x, 0.0)::Bool
│   %2  = Base.eq_float(x, 0.0)::Bool
│   %3  = Base.and_int(%2, true)::Bool
│   %4  = Base.and_int(%3, false)::Bool
│   %5  = Base.or_int(%1, %4)::Bool
│   %6  = Base.lt_float(0.0, x)::Bool
│   %7  = Base.eq_float(0.0, x)::Bool
│   %8  = Base.and_int(%7, false)::Bool
│   %9  = Base.or_int(%6, %8)::Bool
│   %10 = Base.ifelse(%9, 1.0, x)::Float64
│   %11 = Base.ifelse(%5, -1.0, %10)::Float64
└──       return %11
) => Float64

julia> @code_llvm debuginfo=:none sign(2.0)

define double @julia_sign_16056(double) {
top:
  %1 = fcmp uge double %0, 0.000000e+00
  %2 = fcmp ule double %0, 0.000000e+00
  %3 = select i1 %2, double %0, double 1.000000e+00
  %4 = select i1 %1, double %3, double -1.000000e+00
  ret double %4
}
```

```julia
# In v1.3
julia> @code_typed sign(2.0)                                                                                                                                                                                       
CodeInfo(                                                                                                                                                                                                          
1 ─ %1  = Base.sitofp(Float64, 0)::Float64                                                                                                                                                                         
│   %2  = Base.lt_float(x, %1)::Bool                                                                                                                                                                               
│   %3  = Base.eq_float(x, %1)::Bool                                                                                                                                                                               
│   %4  = Base.lt_float(%1, 9.223372036854776e18)::Bool                                                                                                                                                            
│   %5  = Base.and_int(%3, %4)::Bool                                                                                                                                                                               
│   %6  = Base.fptosi(Int64, %1)::Int64                                                                                                                                                                            
│   %7  = Base.slt_int(%6, 0)::Bool                                                                                                                                                                                
│   %8  = Base.and_int(%5, %7)::Bool                                                                                                                                                                               
│   %9  = Base.or_int(%2, %8)::Bool                                                                                                                                                                                
│   %10 = Base.sitofp(Float64, 0)::Float64                                                                                                                                                                         
│   %11 = Base.lt_float(%10, x)::Bool                                                                                                                                                                              
│   %12 = Base.eq_float(%10, x)::Bool                                                                                                                                                                              
│   %13 = Base.eq_float(%10, 9.223372036854776e18)::Bool                                                                                                                                                           
│   %14 = Base.fptosi(Int64, %10)::Int64                                                                                                                                                                           
│   %15 = Base.slt_int(0, %14)::Bool                                                                                                                                                                               
│   %16 = Base.or_int(%13, %15)::Bool                                                                                                                                                                              
│   %17 = Base.and_int(%12, %16)::Bool                                                                                                                                                                             
│   %18 = Base.or_int(%11, %17)::Bool                                                                                                                                                                              
│   %19 = Base.ifelse(%18, 1.0, x)::Float64
│   %20 = Base.ifelse(%9, -1.0, %19)::Float64
└──       return %20                                
) => Float64                                        

julia> @code_llvm debuginfo=:none sign(2.0)

define double @julia_sign_17326(double) {
top:                                                
  %1 = fcmp uge double %0, 0.000000e+00
  %2 = fcmp ule double %0, 0.000000e+00
  %3 = select i1 %2, double %0, double 1.000000e+00
  %4 = select i1 %1, double %3, double -1.000000e+00 
  ret double %4                                     
}                                                   
```

I'll follow up shortly with a PR which proposes a change that leads to simpler Julia IR.

jmert:
This PR fixes #34914: 
```julia
# With PR
julia> @code_typed sign(2.0)
CodeInfo(
1 ─ %1 = Base.lt_float(x, 0.0)::Bool
│   %2 = Base.lt_float(0.0, x)::Bool
│   %3 = Base.ifelse(%2, 1.0, x)::Float64
│   %4 = Base.ifelse(%1, -1.0, %3)::Float64
└──      return %4
) => Float64

julia> @code_llvm debuginfo=:none sign(2.0)

define double @julia_sign_18479(double) {
top:
  %1 = fcmp uge double %0, 0.000000e+00
  %2 = fcmp ule double %0, 0.000000e+00
  %3 = select i1 %2, double %0, double 1.000000e+00
  %4 = select i1 %1, double %3, double -1.000000e+00
  ret double %4
}
```
Previously, the type mismatch in the comparisons meant that `sign(::Furlong)` from the test helper didn't work, so I've used that as a regression test.

### DIFF
--- a/base/gmp.jl
+++ b/base/gmp.jl
@@ -10,7 +10,7 @@ import .Base: *, +, -, /, <, <<, >>, >>>, <=, ==, >, >=, ^, (~), (&), (|), xor,
              sum, trailing_zeros, trailing_ones, count_ones, tryparse_internal,
              bin, oct, dec, hex, isequal, invmod, _prevpow2, _nextpow2, ndigits0zpb,
              widen, signed, unsafe_trunc, trunc, iszero, isone, big, flipsign, signbit,
-             hastypemax
+             sign, hastypemax
 
 if Clong == Int32
     const ClongMax = Union{Int8, Int16, Int32}
@@ -668,6 +668,11 @@ flipsign!(x::BigInt, y::Integer) = (signbit(y) && (x.size = -x.size); x)
 flipsign( x::BigInt, y::Integer) = signbit(y) ? -x : x
 flipsign( x::BigInt, y::BigInt)  = signbit(y) ? -x : x
 # above method to resolving ambiguities with flipsign(::T, ::T) where T<:Signed
+function sign(x::BigInt)
+    isneg(x) && return -one(x)
+    ispos(x) && return one(x)
+    return x
+end
 
 show(io::IO, x::BigInt) = print(io, string(x))
 
--- a/base/mpfr.jl
+++ b/base/mpfr.jl
@@ -13,7 +13,7 @@ import
         nextfloat, prevfloat, promote_rule, rem, rem2pi, round, show, float,
         sum, sqrt, string, print, trunc, precision, exp10, expm1,
         log1p,
-        eps, signbit, sin, cos, sincos, tan, sec, csc, cot, acos, asin, atan,
+        eps, signbit, sign, sin, cos, sincos, tan, sec, csc, cot, acos, asin, atan,
         cosh, sinh, tanh, sech, csch, coth, acosh, asinh, atanh,
         cbrt, typemax, typemin, unsafe_trunc, floatmin, floatmax, rounding,
         setrounding, maxintfloat, widen, significand, frexp, tryparse, iszero,
@@ -786,6 +786,11 @@ cmp(x::CdoubleMax, y::BigFloat) = -cmp(y,x)
 <=(x::CdoubleMax, y::BigFloat) = !isnan(x) && !isnan(y) && cmp(y,x) >= 0
 
 signbit(x::BigFloat) = ccall((:mpfr_signbit, :libmpfr), Int32, (Ref{BigFloat},), x) != 0
+function sign(x::BigFloat)
+    c = cmp(x, 0)
+    (c == 0 || isnan(x)) && return x
+    return c < 0 ? -one(x) : one(x)
+end
 
 function precision(x::BigFloat)  # precision of an object of type BigFloat
     return ccall((:mpfr_get_prec, :libmpfr), Clong, (Ref{BigFloat},), x)
--- a/base/number.jl
+++ b/base/number.jl
@@ -114,9 +114,9 @@ signbit(x::Real) = x < 0
 
 Return zero if `x==0` and ``x/|x|`` otherwise (i.e., ±1 for real `x`).
 """
-sign(x::Number) = x == 0 ? x/abs(oneunit(x)) : x/abs(x)
-sign(x::Real) = ifelse(x < 0, oftype(one(x),-1), ifelse(x > 0, one(x), typeof(one(x))(x)))
-sign(x::Unsigned) = ifelse(x > 0, one(x), oftype(one(x),0))
+sign(x::Number) = iszero(x) ? x/abs(oneunit(x)) : x/abs(x)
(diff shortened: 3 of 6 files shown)


yuyichao:
How does this affect types where `zero` allocates?

I don't know about `BigFloat` but at least `BigInt` seems to be calling a specialized comparison function when comparing to machine precision numbers (and `BigFloat` could in principle do the same)

jmert:
It looks like you're right for both `BigInt` and `BigFloat` — they end up allocating more. I'm not sure what the right answer is. Presumably it'd be nice to have type stability for comparison of other non-standard number types, but e.g. `Unitful` defines it's own `sign` implementation to avoid comparing unitful and dimensionless number as would happen with `Base`'s current implementation.

JeffBezanson:
Could we add global constants for one and zero BigInts so we don't need to allocate a new one on each call? Harder to do for BigFloat of course due to variable precision.

JeffBezanson:
Then again, BigInt and BigFloat should just have their own implementations of `sign`.

jmert:
Following Jeff's suggestion that `BigInt` and `BigFloat` could both use their own implementations...

The performance of `sign` on my computer with the version on master:
```julia
julia> @btime sign($(big"-3"))
  176.355 ns (8 allocations: 160 bytes)
-1

julia> @btime sign($(big"0"))
  173.466 ns (8 allocations: 160 bytes)
0

julia> @btime sign($(big"3"))
  177.679 ns (8 allocations: 160 bytes)
1

julia> @btime sign($(big"-3.0"))
  218.523 ns (8 allocations: 448 bytes)
-1.0

julia> @btime sign($(big"-0.0"))
  221.188 ns (8 allocations: 448 bytes)
-0.0

julia> @btime sign($(big"0.0"))
  220.473 ns (8 allocations: 448 bytes)
0.0

julia> @btime sign($(big"3.0"))
  224.199 ns (8 allocations: 448 bytes)
1.0
```

Using this PR, which now adds specialized versions of `sign` for both `BigInt` and `BigFloat`:
```julia
julia> @btime sign($(big"-3"))
  90.159 ns (4 allocations: 80 bytes)
-1

julia> @btime sign($(big"0"))
  5.039 ns (0 allocations: 0 bytes)
0

julia> @btime sign($(big"3"))
  47.653 ns (2 allocations: 40 bytes)
1

julia> @btime sign($(big"-3.0"))
  97.525 ns (4 allocations: 224 bytes)
-1.0

julia> @btime sign($(big"-0.0"))
  11.553 ns (0 allocations: 0 bytes)
-0.0

julia> @btime sign($(big"0.0"))
  11.834 ns (0 allocations: 0 bytes)
0.0

julia> @btime sign($(big"3.0"))
  65.026 ns (2 allocations: 112 bytes)
1.0
```

jmert:
Well, the CI errors all look real — I was using `Revise` to test the changes, but upon building locally, I get the same bootstrap error. I'll have to investigate further.

jmert:
Fixed locally, so we'll see if this now passes CI again.

For the record, the issue was that I hadn't imported `Base.sign` before extending the methods, so that caused `Base.sign` to be shadowed within `Base.GMP` and `Base.MPFR`. I'm not sure why Revise was letting me get away with not importing first, so maybe that's a bug that should be reproduced and reported in Revise.

jmert:
Bump.

jmert:
Bump

PR #34491, opened 2020-01-23
Fix URLs referencing latest

fredrikekre:
Remove URLs with /latest and use correct, relative, documenter links for the manual, fixes #34488.

Issue #34488, opened 2020-01-23
Difference between latest docs and 1.5-dev docs


Hieronimo:
latest docs show 1.4-dev docs but with "v1.5" label however with v1.4-dev label in bottom-left:

https://docs.julialang.org/en/latest/NEWS/
https://docs.julialang.org/en/v1.5-dev/NEWS/
https://docs.julialang.org/en/v1.4-dev/NEWS/

fredrikekre:
Don't use the `latest` link, where did you get that?

Hieronimo:
I don't remember that, it was doing well until version 1.5-dev.
also `Edit on GitHub` link is broken on NEWS page.

Hieronimo:
Oh, I found it, just go to an invalid link and it says start at `latest` docs.

https://docs.julialang.org/en/v2/
or
https://docs.julialang.org/en/stable/

Hieronimo:
also `file issues or submit pull requests` link on 404 page is go to an archived repository.

Hieronimo:
some other links to `latest` docs:
https://github.com/JuliaLang/julia/search?q=%22docs.julialang.org%2Fen%2Flatest%22&unscoped_q=%22docs.julialang.org%2Fen%2Flatest%22

KristofferC:
Good finds, those should be fixed.

PR #34917, opened 2020-02-28
fix broken links and replace latest docs links

ghost:
Related Issues: #34488 #34491

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -317,7 +317,7 @@ runtest harness).
   - **Homepage:** <https://julialang.org>
   - **Community:** <https://julialang.org/community/>
   - **Source code:** <https://github.com/JuliaLang/julia>
-  - **Documentation:** <https://docs.julialang.org/>
+  - **Documentation:** <https://docs.julialang.org>
   - **Code coverage:** <https://coveralls.io/r/JuliaLang/julia>
 
 * Design of Julia
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1167,7 +1167,7 @@ This section lists changes that do not have deprecation warnings.
     to evaluate the wrapper. Consequently, package authors generally need to specialize
     `copy` and `copyto!` methods rather than `broadcast` and `broadcast!`. This also allows
     for more customization and control of fused broadcasts. See the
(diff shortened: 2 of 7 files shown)


CONTRIBUTING.md
```diff
@@ -317,7 +317,7 @@ runtest harness).
   - **Homepage:** <https://julialang.org>
   - **Community:** <https://julialang.org/community/>
   - **Source code:** <https://github.com/JuliaLang/julia>
-  - **Documentation:** <https://docs.julialang.org/>
+  - **Documentation:** <https://docs.julialang.org>
```
fredrikekre: Why this change (and the similar ones throughout the PR)?

doc/man/julia.1
```diff
@@ -47,7 +47,7 @@ by defining functions and overloading them for different combinations
 of argument types, which can also be user-defined.
 For a more in-depth discussion of the rationale and advantages of Julia
 over other systems, please see the online manual:
-https://docs.julialang.org/en/latest/
+https://docs.julialang.org
```
fredrikekre:
```suggestion
https://docs.julialang.org/
```

HISTORY.md
```diff
@@ -1167,7 +1167,7 @@ This section lists changes that do not have deprecation warnings.
     to evaluate the wrapper. Consequently, package authors generally need to specialize
     `copy` and `copyto!` methods rather than `broadcast` and `broadcast!`. This also allows
     for more customization and control of fused broadcasts. See the
-    [Interfaces chapter](https://docs.julialang.org/en/latest/manual/interfaces/#man-interfaces-broadcasting-1)
+    [Interfaces chapter](https://docs.julialang.org/en/v1/manual/interfaces/#man-interfaces-broadcasting-1)
```
fredrikekre: I'm not sure if it makes sense to update the HISTORY file. Did you check that the updated links still make sense for the context?

ViralBShah: I think it would be ok. v1 would certainly be better than latest in HISTORY, I think.

ViralBShah:
I am ok merging this one. On the whole, removing `latest` is a good idea.

ViralBShah:
Can we rebase this PR?

Issue #34890, opened 2020-02-26
optimize idea: bulk insert for types


vtjnash:
There's currently several O(n^2) steps inside `jl_recache_type`. The value of `n` used be very small, but it's gotten quite large. I'd be likely worthwhile to rotate this algorithm around so that we can do the work faster. In particular, instead of the second loop here, we could try using a ptrhash there. And in the first case, we could do a bulk insert of all new items, first using `qsort` to order the new items, then a do merge step.

---
![Screen Shot 2020-02-26 at 1 10 20 PM](https://user-images.githubusercontent.com/330950/75374348-1fcda480-589a-11ea-9dc9-6551439473ce.png)
---
![image](https://user-images.githubusercontent.com/330950/75374373-2f4ced80-589a-11ea-8c10-ae5dfb051baa.png)

PR #34919, opened 2020-02-28
faster type allocation lookup

vtjnash:
This previously implemented a binary search to find the type, but that required holding a lock and doing upwards of 16 steps of comparisons. This uses instead a hash set (nearly an IdDict). I observe a load factor of about 30% for Tuple, and much faster lookups.
On this microbenchmark test script, it shows a >2x improvement:
```julia
julia> T = Int; @btime Tuple{T, T}
before:  172.414 ns (0 allocations: 0 bytes)
after:    62.356 ns (0 allocations: 0 bytes)
```


JeffBezanson:
Any measurable changes in package loading?

vtjnash:
It's not a noticeable percentage of the cost there, but it's nice to set it to effectively zero (fixes much of #34890—possibly because this hits the L1/L2 caches much less hard??).
on master:
```
$ time ./usr/bin/julia -e 'using DataFrames'; time ./usr/bin/julia -e 'using Plots'; time ./usr/bin/julia -e 'using ApproxFun'
real	0m1.151s
user	0m1.289s
sys	0m0.144s

real	0m4.243s
user	0m4.248s
sys	0m0.207s

real	0m18.912s
user	0m18.582s
sys	0m0.858s
```
current measurements (with many additional changes):
```
$ time ./usr/bin/julia -e 'using DataFrames'; time ./usr/bin/julia -e 'using Plots'; time ./usr/bin/julia -e 'using ApproxFun'

real	0m1.175s
user	0m1.268s
sys	0m0.156s

real	0m3.843s
user	0m3.901s
sys	0m0.199s

real	0m16.749s
user	0m16.317s
sys	0m0.995s
```

src/jltypes.c
```diff
@@ -574,67 +581,36 @@ static int is_typekey_ordered(jl_value_t **key, size_t n)
...
-                int cmp = typekey_compare(dt, jl_svec_data(dk->parameters), jl_nparams(dk));
-                if (cmp != 0)
-                    return cmp;
-            }
-        }
+    unsigned hash = 3;
+    for (j = 0; j < n; j++) {
+        jl_value_t *kj = key[j];
+        jl_datatype_t *dk = (jl_datatype_t*)(jl_is_unionall(kj) ? jl_unwrap_unionall(kj) : kj);
+        if (!jl_is_datatype(kj))
```
JeffBezanson: Did you mean `dk` instead of `kj` here?

src/jltypes.c
```diff
@@ -748,54 +724,126 @@ static int is_cacheable(jl_datatype_t *type)
...
-        ssize_t idx = lookup_type_idx(type->name, jl_svec_data(type->parameters),
-                                      jl_svec_len(type->parameters), ord);
-        if (idx >= 0)
-            type = (jl_datatype_t*)jl_svecref(ord ? type->name->cache : type->name->linearcache, idx);
-        else
-            cache_insert_type((jl_value_t*)type, ~idx, ord);
+        assert(jl_is_datatype(type));
+        jl_value_t **key = jl_svec_data(type->parameters);
+        int n = jl_svec_len(type->parameters);
+        int ord = is_typekey_ordered(key, n);
```
JeffBezanson: Rename this `is_typekey_hashable`?

JeffBezanson (approved):

vtjnash:
I'm going to recreate this as a different PR that goes further and incorporates other work.

vtjnash:
Today's measurements:
master
```
$ time ./usr/bin/julia -e 'using DataFrames'; time ./usr/bin/julia -e 'using Plots'; time ./usr/bin/julia -e 'using ApproxFun'
real	0m1.122s
user	0m1.241s
sys	0m0.153s

real	0m3.811s
user	0m3.863s
sys	0m0.196s

real	0m20.926s
user	0m20.581s
sys	0m0.814s

time ./julia -e 'using Plots; @time display(@time plot(1:10))'
  4.062018 seconds (9.54 M allocations: 473.947 MiB, 5.78% gc time)
  9.953003 seconds (17.81 M allocations: 876.303 MiB, 3.06% gc time)

real	0m13.985s
user	0m13.827s
sys	0m0.395s

julia> T = Int; @btime Tuple{T, T}
  109.337 ns (0 allocations: 0 bytes)
```
PR
```

real	0m1.056s
user	0m1.196s
sys	0m0.152s

real	0m3.600s
user	0m3.677s
sys	0m0.206s

real	0m16.996s
user	0m16.774s
sys	0m0.772s

time ./julia -e 'using Plots; @time display(@time plot(1:10))'
  3.862645 seconds (9.55 M allocations: 474.279 MiB, 7.42% gc time)
  9.783306 seconds (17.81 M allocations: 876.683 MiB, 3.55% gc time)

real	0m13.676s
user	0m13.499s
sys	0m0.419s

julia> T = Int; @btime Tuple{T, T}
  41.030 ns (0 allocations: 0 bytes)
```

StefanKarpinski:
Those are substantial savings. It's kind of weird that with all the other things getting faster the literal time-to-first-plot doesn't get much better, but solving that isn't the _only_ concern, so 🎉

JeffBezanson:
Ick, doctests --- this seems awfully implementation-detaily to have doctests for the output; maybe we should edit that section a bit.

src/builtins.c
```diff
@@ -324,11 +332,10 @@ JL_DLLEXPORT uintptr_t jl_object_id_(jl_value_t *tv, jl_value_t *v) JL_NOTSAFEPO
...
     if (dt == jl_datatype_type) {
         jl_datatype_t *dtv = (jl_datatype_t*)v;
-        // `name->wrapper` is cacheable even though it contains TypeVars
-        // that don't have stable IDs.
-        //if (jl_egal(dtv->name->wrapper, v))
-        //    return bitmix(~dtv->name->hash, 0xaa5566aa);
-        return bitmix(~dtv->name->hash, hash_svec(dtv->parameters));
+        uintptr_t h = ~dtv->name->hash;
+        if (dtv->isconcretetype)
+            return bitmix(h, dtv->hash);
```
JeffBezanson: Why isn't `dtv->hash` sufficient here?

vtjnash:
Yeah, I need to take that doctest out (the hash is a bitmix that includes the module build_id random number). Additionally / alternatively, I've considered making small dictionaries use a linear lookup.

src/iddict.c
```diff
@@ -151,22 +148,24 @@ jl_array_t *jl_eqtable_put(jl_array_t *h, jl_value_t *key, jl_value_t *val, int
     return h;
 }
 
+// Note: lookup in the IdDict it permitted concurrently, if you avoid deletions,
```
JeffBezanson:
```suggestion
// Note: lookup in the IdDict is permitted concurrently, if you avoid deletions,
```

src/builtins.c
vtjnash: dtv->hash was only the contents of the parameters, so we need to add in the type itself now, but I could change that

JeffBezanson: Yes, it seems to make sense to me to have the name's hash already mixed in. Any downside?

JeffBezanson (approved):

PR #34920, opened 2020-02-28
Semantic Versioning block in Contributions.md

### LINKED ISSUE #31685: Package Compatibility Paragraph in Contributing.md is misleading
The paragraph about [Package Compatibility](https://github.com/JuliaLang/julia/blob/master/CONTRIBUTING.md#package-compatibility-across-releases) sounds like Julia is in contrast to Semantic Versioning.

Should this be changed in a paragraph about Semantic Versioning?

carstenbauer:
As discussed in https://github.com/JuliaLang/julia/issues/31685 some time ago, this PR updates the old "Package Compatibility Across Releases" section in contributions.md from 0.x times, where every Julia release was breaking.

It introduces a short block about Semantic Versioning and reduces the mention of Compat.jl to a minimum.

Closes #31685.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -64,11 +64,21 @@ For developers who need to wrap C libraries so that they can be called from Juli
 
 ### Semantic Versioning and Package Compatibility Across Releases
 
-When tagging a new release of your package you should follow [Semantic Versioning (SemVar)](https://semver.org/). Given a version number of the form MAJOR.MINOR.PATCH, increment the:
+When tagging a new release of your package you should follow [Semantic Versioning (SemVar)](https://semver.org/). Given a version number of the form MAJOR.MINOR.PATCH with MAJOR >= 1, increment the:
 
 * MAJOR version when you make incompatible API changes,
 * MINOR version when you add functionality in a backwards compatible manner, and
 * PATCH version when you make backwards compatible bug fixes.
+
+For packages in an early stage with version numbers 0.MINOR.PATCH, for example 0.6.2, MINOR takes over the role of MAJOR and should therefore only be increased when introducing incompatible API changes.
+
+Examples:
+
+* 0.4.2 -> 0.4.3 (non-breaking)
+* 0.4.2 -> 0.5.0 (breaking)
+* 1.2.3 -> 1.3.0 (non-breaking)
+* 1.3.0 -> 1.3.1 (non breaking)
+* 1.9.2 -> 2.0.0 (breaking)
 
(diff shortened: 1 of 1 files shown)


DilumAluthge:
If you are going to talk about SemVer, you should mention that the rules are different for pre-1.0 packages. Specifically you bump minor only if it is a breaking change, otherwise you bump patch.

crstnbr:
Good idea. I also added a few examples to make it clear.

dkarrasch:
Some of the lines are much longer than 80 characters. Would be nice if you could wrap them accordingly. This may be hard to see in your editor which wraps lines autmoatically, but it is easy to see in the change preview online.

crstnbr:
> Some of the lines are much longer than 80 characters. Would be nice if you could wrap them accordingly.

Sure, I can change that but that's the case for almost the entire file isn't it? Should I change it everywhere?

dkarrasch:
I just saw that the part that you replaced was nicely wrapped, so maybe it's enough to take part of your part now to not mess up stylistic changes from content changes.

crstnbr:
Done.

CONTRIBUTING.md
```diff
@@ -62,20 +62,38 @@ Julia has a built-in [package manager](https://julialang.github.io/Pkg.jl/v1/) b
...
-
-However, porting a package to the latest release may cause the package to break on
-earlier Julia releases. To maintain compatibility across releases, use
-[`Compat.jl`](https://github.com/JuliaLang/Compat.jl). Find the fix for your package
-from the README, and specify the minimum version of Compat that provides the fix
-in your REQUIRE file. To find the correct minimum version, refer to
+### Semantic Versioning and Package Compatibility Across Releases
+
+When tagging a new release of your package you should follow
+[Semantic Versioning (SemVar)](https://semver.org/). Given a version number
```
fredrikekre:
```suggestion
[Semantic Versioning (SemVer)](https://semver.org/). Given a version number
```

fredrikekre (commented):
Seems like maybe this section should just be removed? With the current addition it is yet another place to read about (and yet another place to maintain) how to tag packages, I think it is best to just document these kind of things in the Pkg docs maybe?

crstnbr:
Fair point. Perhaps contributions.md, because of its location, should only be about contributing to `JuliaLang/julia`. The "Contributing a Julia package" section + the ohne changed in this PR indeed seem to have quite a different character than the rest.

crstnbr:
Alright, @fredrikekre, I dropped the sections on how to develop packages as this information exists elsewhere and Contributions.md is probably not the right place for it.

StefanKarpinski:
This now appears to only delete text. Is that the intention?

fredrikekre (approved):

crstnbr:
> This now appears to only delete text. Is that the intention?

Yes. At first, I tried to update the text but then it seemed much more appropriate to just get rid of this section (the information is present elsewhere).

PR #34925, opened 2020-02-28
Improve docstrings of IEEE floating point types

### LINKED ISSUE #34833: Add info on internal structure to floating-point types
```jl
help?> Float16
search: Float16 Float64 Float32 float floatmin floatmax Cfloat BigFloat

  Float16 <: AbstractFloat

  16-bit floating point number type.
```

I suggest adding here and in the other IEEE-compatible types the information (i) that it is IEEE compatible, and (ii) the number of bits used for the exponent and mantissa.

carstenbauer:
As suggested by @dpsanders in #34833, this PR improves the docstrings of `Float16`, `Float32`. and `Float64`. Specifically, we

* mention that the types are IEEE 754 standard compatible
* describe their binary format

I went with "fraction" instead of "mantissa", mainly because this is how [wikipedia](https://en.wikipedia.org/wiki/Double-precision_floating-point_format) calls it.

Closes #34833

### DIFF
--- a/base/docs/basedocs.jl
+++ b/base/docs/basedocs.jl
@@ -1557,12 +1557,22 @@ julia> false * NaN
 """
 Bool
 
+# IEEE 754 standard
+binary = Dict(
+    16 => Dict(:sign => 1, :exp => 5, :frac => 10),
+    32 => Dict(:sign => 1, :exp => 8, :frac => 23),
+    64 => Dict(:sign => 1, :exp => 11, :frac => 52),
+)
+
 for bit in (16, 32, 64)
+    b = binary[bit]
     @eval begin
         """
             Float$($bit) <: AbstractFloat
 
-        $($bit)-bit floating point number type.
+        $($bit)-bit floating point number type (IEEE 754 standard).
+
+        Binary format: $($b[:sign]) sign, $($b[:exp]) exponent, $($b[:frac]) fraction bits.
         """
         $(Symbol("Float", bit))
     end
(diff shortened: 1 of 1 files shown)


crstnbr:
Fancy idea: It would be cool if we could [visualize](https://en.wikipedia.org/wiki/Double-precision_floating-point_format#/media/File:IEEE_754_Double_Floating_Point_Format.svg) the binary format in the REPL. :) 

But apart from that I don't know how to do it, it probably is harder to parse and doesn't provide much information.

StefanKarpinski (approved):

base/docs/basedocs.jl
```diff
@@ -1557,12 +1557,22 @@ julia> false * NaN
...
 Bool
 
+# IEEE 754 standard
+binary = Dict(
+    16 => Dict(:sign => 1, :exp => 5, :frac => 10),
+    32 => Dict(:sign => 1, :exp => 8, :frac => 23),
+    64 => Dict(:sign => 1, :exp => 11, :frac => 52),
+)
+
 for bit in (16, 32, 64)
```
JeffBezanson:
```suggestion
for (bit, sign, exp, frac) in ((16, 1, 5, 10), ...
```

StefanKarpinski: Just to expand on Jeff's comment, I think he's suggesting that it would be better to put these values inline in the for loop's RHS instead of making a Dict out of them.

crstnbr: Yeah, that's nicer.

crstnbr:
CI fail seems unrelated. Ready to go, @JeffBezanson?

PR #34929, opened 2020-02-29
Some missing tests for Cholesky

kshyatt:


stdlib/LinearAlgebra/test/cholesky.jl
```diff
@@ -62,7 +62,12 @@ end
         κ     = cond(apd, 1) #condition number
 
         unary_ops_tests(apd, capd, ε*κ*n)
-
+        if eltya != Int
+            @test Factorization{eltya}(capd) === capd
+            if eltya <: Real
+                @test Array(Factorization{complex(eltya)}(capd)) ≈ Array(factorize(complex(apd)))
```
dkarrasch: As in the other PR, shall we add an eltype test as well?

stdlib/LinearAlgebra/test/cholesky.jl
```diff
@@ -399,4 +404,25 @@ end
...
 end
 
+@testset "CholeskyPivoted and Factorization" begin
+    A = randn(8,8)
+    B = A'A
+    C = cholesky(B, Val(true), check=false)
+    @test CholeskyPivoted{eltype(C)}(C) === C
+    @test Factorization{eltype(C)}(C) === C
+    @test Array(CholeskyPivoted{complex(eltype(C))}(C)) == Array(cholesky(complex(B), Val(true), check=false))
+    @test Factorization{complex(eltype(C))}(C) == cholesky(complex(B), Val(true), check=false)
```
dkarrasch: These are tough exact equality tests. Is that safe, or vulnerable to floating point issues that we may need to relax later?

stdlib/LinearAlgebra/test/cholesky.jl
```diff
@@ -399,4 +404,25 @@ end
...
+@testset "REPL printing of CholeskyPivoted" begin
+    A = randn(8,8)
+    B = A'A
+    C = cholesky(B, Val(true), check=false)
+    cholstring = sprint((t, s) -> show(t, "text/plain", s), C)
+    rankstring = "$(C.uplo) factor with rank $(rank(C)):" 
+    factorstring = sprint((t, s) -> show(t, "text/plain", s), C.uplo == 'U' ? C.U : C.L)
+    permstring   = sprint((t, s) -> show(t, "text/plain", s), C.p)
+    @test cholstring == "$(summary(C))\n$rankstring\n$factorstring\npermutation:\n$permstring"
+end
```
dkarrasch: I'm not sure: do we test REPL printing usually? Seems like if somebody changes the printing, then (s)he is simply going to rewrite this whole test.

Well, I just saw that we do have a printing test in LU, so I guess this is fine then.

dkarrasch (commented):
LGTM. Just a few comments, the only actionable things being adding a type test after conversion and possibly relaxing the equality tests.

stdlib/LinearAlgebra/test/cholesky.jl
dkarrasch: I only now saw that these tests indeed fail for the obvious reason, so we should really relax them.

kshyatt: Oh whoops they passed locally

stdlib/LinearAlgebra/test/cholesky.jl
kshyatt: We have it for SVD as well

dkarrasch:
Watch out for trailing white space, @kshyatt.

kshyatt:
omg sorry 😰

dkarrasch (approved):

stdlib/LinearAlgebra/test/cholesky.jl
```diff
@@ -399,4 +405,26 @@ end
...
 end
 
+@testset "CholeskyPivoted and Factorization" begin
+    A = randn(8,8)
+    B = A'A
+    C = cholesky(B, Val(true), check=false)
+    @test CholeskyPivoted{eltype(C)}(C) === C
+    @test Factorization{eltype(C)}(C) === C
+    @test Array(CholeskyPivoted{complex(eltype(C))}(C)) ≈ Array(cholesky(complex(B), Val(true), check=false))
+    @test Factorization{complex(eltype(C))}(C) ≈ cholesky(complex(B), Val(true), check=false)
```
dkarrasch: The latest test failures are real and related to this line. Given the three tests above, you may simply remove this one without reducing code coverage, I'd say.

kshyatt:
I think by converting that line to an `Array` afterwards it should work...

dkarrasch:
> I think by converting that line to an `Array` afterwards it should work...

That's right, it just seems to me that you're not covering new code with this, but either way is fine.

kshyatt:
> That's right, it just seems to me that you're not covering new code with this, but either way is fine.

We are, though, bc the `Factorization{T}` line calls `CholeskyPivoted{T}` and I want to make sure that pass-through works.

kshyatt:
OK to merge?

kshyatt:
win32 failure unrelated

PR #34930, opened 2020-02-29
Add test for sincosd of missing

kshyatt:


dkarrasch:
Interesting that this works, since `sind(missing)` and `cosd(missing)` individually don't. In fact, it seems to be the only `trig.jl` function that handles `missing`. Should this be fixed by defining `deg2rad` and `rad2deg` for `missing` (in a separate PR)?

dkarrasch (approved):

PR #34934, opened 2020-02-29
A few tests for RNGs

carstenbauer:
Added a few tests for `typeof(rng)` and the generic `Random.typeof_rng(rng)`.


rfourquet:
I'm personnally not convinced it's necessary to test `Random.typeof_rng` as it's an internal, but if you do, you should probably test that `typeof_rng(default_rng()) == MersenneTwister`, as this this the whole point of `typeof_rng`.

PR #34938, opened 2020-02-29
Missing tests for TwicePrecision

kshyatt:


test/ranges.jl
```diff
@@ -168,9 +168,23 @@ end
             @test cmp_sn2(Tw(xw/yw), astuple(x/y)..., slopbits)
         end
     end
+    @testset "high precision of varying types" begin
+        x = Float32(π)
+        y = Float64(Base.MathConstants.γ)
+        @test Base.mul12(x, y)[1] ≈ Base.mul12(Float64(π), y)[1] rtol=1e-6
```
timholy: Instead of testing each component, would it be better to test the values after calling `canonicalize2` on each output? I would also recommend testing against `Float64(x)` rather than `Float64(π)` since one expects much higher level of agreement.

test/ranges.jl
```diff
@@ -168,9 +168,23 @@ end
...
+        xtp = Base.TwicePrecision{Float32}(π)
+        ytp = Base.TwicePrecision{Float64}(Base.MathConstants.γ)
+        @test Float32(xtp + ytp) ≈ Float32(Base.TwicePrecision{Float64}(π) + ytp)
+    end
 
     x1 = Base.TwicePrecision{Float64}(1)
     x0 = Base.TwicePrecision{Float64}(0)
+    @test eltype(x1) == Float64
+    @test eltype(typeof(x1)) == Float64
+    @test zero(typeof(x1)) == x0
```
timholy:
```suggestion
    @test zero(typeof(x1)) === x0
```

timholy (commented):
Thanks for doing this!

inkydragon (approved):
LGTM.

Maybe merge as is.
Further improvements could be made in a future pr.

PR #34942, opened 2020-03-01
RFC: Make `Base.manifest_names` a Vector instead of a Tuple

DilumAluthge:
This would allow you to use custom names for your manifest files by doing something like:
```julia
pushfirst!(Base.manifest_names, "CustomManifestName.toml"
```

I'm curious what the performance impact of this would be. Would it be possible for someone with the appropriate permissions to trigger a Nanosoldier run on this PR?


DilumAluthge:
Probably a bad idea.

PR #34943, opened 2020-03-01
test sprintf %p using real pointers

bicycle1885:
I discovered that `@sprintf "%p"` is not tested using real pointers while working on #34941. Since pointers are not integers, we need to test them in addition to integers. Please merge this only after #34941 or CI will fail.


fredrikekre:
> Please merge this only after #34941 or CI will fail.

Rebased on master, looks like buildbots are down so should retrigger once they are up.

Issue #34842, opened 2020-02-22
; in help mode doesn't return anything


Hieronimo:
```julia
help?> ;
search:

```

**Julia Version 1.3.0
Commit 46ce4d7933 (2019-11-26 06:09 UTC)**

mbauman:
Thanks for the report.  We do have this definition, but somehow we're losing the `;` before we get to the lookup.

https://github.com/JuliaLang/julia/blob/835e148ef7b62ac09ff3f5209745796473161d60/base/docs/basedocs.jl#L404-L409

JeffBezanson:
fixed by #34944.

PR #34944, opened 2020-03-01
help ; now has output

willywalters5:
Fixes issue #34842.
The semicolon output suppression prevented the output from being displayed.

### DIFF
--- a/stdlib/REPL/src/REPL.jl
+++ b/stdlib/REPL/src/REPL.jl
@@ -743,7 +743,7 @@ function eval_with_backend(ast, backend::REPLBackendRef)
     take!(backend.response_channel) # (val, iserr)
 end
 
-function respond(f, repl, main; pass_empty = false, supress_on_semicolon = true)
+function respond(f, repl, main; pass_empty = false, suppress_on_semicolon = true)
     return function do_respond(s, buf, ok)
         if !ok
             return transition(s, :abort)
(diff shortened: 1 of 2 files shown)


stdlib/REPL/src/REPL.jl
```diff
@@ -743,7 +743,7 @@ function eval_with_backend(ast, backend::REPLBackendRef)
     take!(backend.response_channel) # (val, iserr)
 end
 
-function respond(f, repl, main; pass_empty = false)
+function respond(f, repl, main; pass_empty = false, supress_on_semicolon = true)
```
JeffBezanson:
```suggestion
function respond(f, repl, main; pass_empty = false, suppress_on_semicolon = true)
```

stdlib/REPL/src/REPL.jl
```diff
@@ -758,7 +758,8 @@ function respond(f, repl, main; pass_empty = false)
             catch
                 response = (catch_stack(), true)
             end
-            print_response(repl, response, !ends_with_semicolon(line), Base.have_color)
+	    hide_output = supress_on_semicolon && ends_with_semicolon(line)
```
JeffBezanson: Indentation needs fixing here; use spaces instead of tab.

JeffBezanson:
Thanks!

fredrikekre (approved):

fredrikekre:
This makes it impossible to suppress help output though, e.g. this would not "work":
```
help?> sin;

julia>
```
but I guess it's kind of pointless to request suppressed help.

Issue #34739, opened 2020-02-12
Documentation on performance tips about abstract-typed fields


HenriDeh:
Following [this discussion on discourse](https://discourse.julialang.org/t/are-fields-with-structtypes-to-be-avoided-as-well/34515/4), I think the documentation should be more specific about field types.

[The performance tips documentation] states that one should avoid abstract types but really it should be non-concrete types altogether that should be avoided.

kamalojasv181:
@HenriDeh so basically we should parameterize a struct in `struct Bar{T} foo::Foo{T} end ` this way and not like this `struct Bar{T <: Foo} foo::T end` in an abstract way to improve the performance right?

HenriDeh:
No, both your definitions are correct. In your second example T will be a concrete type so it's fine. What you should not do is `struct Bar foo::Foo end` because Foo alone is not a concrete type (but not an abstract type either, hence this issue).

timholy:
Want to try a PR, @HenriDeh? You can edit the files in your browser, see https://help.github.com/en/github/managing-files-in-a-repository/editing-files-in-another-users-repository.

kamalojasv181:
@timholy if no one wants, then I can contribute to it.

HenriDeh:
I'm working on it, but your review will be welcome :)

HenriDeh:
Done, there's a test failing about white spaces. If anyone knows what I should do about it please tell me.

PR #34946, opened 2020-03-01
Extended performance tips about abstract types

HenriDeh:
Following issue [#34739](https://github.com/JuliaLang/julia/issues/34739), I added examples where fields that are not abstract are also detrimental to performance. Specifically, fields with unparameterized types should also be avoided.

### DIFF
--- a/doc/src/manual/performance-tips.md
+++ b/doc/src/manual/performance-tips.md
@@ -342,6 +342,14 @@ For reasons of length the results are not shown here, but you may wish to try th
 the type is fully-specified in the first case, the compiler doesn't need to generate any code
 to resolve the type at run-time. This results in shorter and faster code.
 
+One should also keep in mind that not-fully-parameterized types behave like abstract types. For example, even though a fully specified `Array{T,n}` is concrete, `Array` itself with no parameters given is not concrete:
+
+```jldoctest myambig3
+julia> !isconcretetype(Array), !isabstracttype(Array), isstructtype(Array), !isconcretetype(Array{Int}), isconcretetype(Array{Int,1})
+(true, true, true, true, true)
+```
+In this case, it would be better to avoid declaring `MyType` with a field `a::Array` and instead declare the field as `a::Array{T,N}` or as `a::A`, where `{T,N}` or `A` are parameters of `MyType`.
+
 ### Avoid fields with abstract containers
 
 The same best practices also work for container types:
@@ -353,6 +361,10 @@ julia> struct MySimpleContainer{A<:AbstractVector}
 
 julia> struct MyAmbiguousContainer{T}
            a::AbstractVector{T}
+       end
+
+julia> struct MyAlsoAmbiguousContainer
+           a::Array
        end
 ```
 
@@ -378,6 +390,17 @@ julia> b = MyAmbiguousContainer([1:3;]);
 
 julia> typeof(b)
 MyAmbiguousContainer{Int64}
+
+julia> d = MyAlsoAmbiguousContainer(1:3);
+
+julia> typeof(d), typeof(d.a)
+(MyAlsoAmbiguousContainer, Vector{Int64})
+
+julia> d = MyAlsoAmbiguousContainer(1:1.0:3);
+
+julia> typeof(d), typeof(d.a)
+(MyAlsoAmbiguousContainer, Vector{Float64})
+
 ```
 
 For `MySimpleContainer`, the object is fully-specified by its type and parameters, so the compiler


StefanKarpinski:
I think that adding a whole separate section for this is unnecessary, it could just be noted that parametric types without all parameters bound are not concrete, so performance issues apply to them as well and that therefore one wants to either specify concrete values for parameters or path the parameters through to the containing type.

HenriDeh:
In this case, I changed "abstract" to "non-concrete" in the title of the section. I kept the example, if you feel like it's really unnecessary and that a single sentence/paragraph should be enough then I'll try to find one.

doc/src/manual/performance-tips.md
```diff
@@ -331,7 +331,37 @@ For reasons of length the results are not shown here, but you may wish to try th
...
+       end
+ ```
+`MySecondType` and `MyThirdType` are both better choices than 
+
+```jldoctest myambig3
+julia> mutable struct MyAlsoAmbiguousType
+           mpt::MyParametricType
+       end
+```
+Even though `MyParametricType` is not an abstract type, when it is not parameterized, it is not a concrete type either:
```
johnnychen94: > `when it is not parameterized`

I read four times until I realized this belongs to `it is not a concrete type either` 😂  Perhaps you can adjust it a bit for nonnative English speakers.

timholy:
I think the abstract type is simpler than a not-fully-paramterized type. Consequently you should flip the order, keep the old material on abstract types and then add that not-full-parametrized types behave similarly. Like Stefan, I kind of think a single short example would suffice. Instead of inventing new types like `MyParametricType`, how about just distinguish `Array` and `Array{T}` from `Array{T,N}`?

johnnychen94 (commented):
Out of curiosity. If `MyParametricType` is a container like `AbstractArray`, will there be any performance difference for `MySectionType` and `MyThirdType`? IIUC `MySectionType` would generate too many concrete types compared to `MyThirdType` and thus be slower.

HenriDeh:
Okay thank you. I think this last commit is much better then.

doc/src/manual/performance-tips.md
HenriDeh: I guess this is because I'm not native either. I changed this to something clearer.

timholy (approved):

doc/src/manual/performance-tips.md
```diff
@@ -331,6 +331,14 @@ For reasons of length the results are not shown here, but you may wish to try th
 the type is fully-specified in the first case, the compiler doesn't need to generate any code
 to resolve the type at run-time. This results in shorter and faster code.
 
+One should also keep in mind that not-fully-parameterized types behave like abstract types. For example, even though `Array` is not an abstract type, it is not a concrete type either when its parameters are not bound:
```
StefanKarpinski:
```suggestion
One should also keep in mind that not-fully-parameterized types behave like abstract types. For example, even though a fully specified `Array{T,n}` is concrete, `Array` itself with no parameters given is not concrete:
```

StefanKarpinski:
The terminology is a bit fuzzy here. There's nominal abstract types, i.e. the ones you explicitly declare with `abstract type A end`. However, I would call `Array` an abstract type as well. I edited a bit to avoid saying that `Array` is not an abstract type. True, it wasn't declared with `abstract type`, but it's not concrete, so it must be abstract. Man, the word "abstract" got real weird.

crstnbr:
> However, I would call `Array` an abstract type as well.

Well, I guess you could call it *abstract* but it isn't an abstract type (`isabstract(Array) == false`), as you noted. It's a `UnionAll` since it has free parameters. I like the idea of distinguishing the two cases: an abstract type is a **specific** *abstract* type (a node in the type tree) whereas anything with a free parameter is an *abstract* "**set of types**". Both are *abstract*, they are not concrete, but only one is an abstract type.

But I do agree that this is getting weird, in particular for beginners reading the manual, and I think the current way of circumventing this point by talking about non-concreteness is a good idea..

doc/src/manual/performance-tips.md
```diff
@@ -331,6 +331,14 @@ For reasons of length the results are not shown here, but you may wish to try th
 the type is fully-specified in the first case, the compiler doesn't need to generate any code
 to resolve the type at run-time. This results in shorter and faster code.
 
+One should also keep in mind that not-fully-parameterized types behave like abstract types. For example, even though a fully specified `Array{T,n}` is concrete, `Array` itself with no parameters given is not concrete:
+
+```jldoctest myambig3
+julia> isconcretetype(Array), isabstracttype(Array), isstructtype(Array), isconcretetype(Array{Int}), isconcretetype(Array{Int,1})
+(false, false, true, false, true)
```
vtjnash, line 338:
```suggestion
julia> !isconcretetype(Array), !isabstracttype(Array), isstructtype(Array), !isconcretetype(Array{Int}), isconcretetype(Array{Int,1})
(true, true, true, true, true)
```

HenriDeh: I'm not sure about this, it seems a bit harder to decode for the reader imho.

vtjnash: My think was that it would be helpful that all of these are testing true, so the reader doesn't need to take the negation in their head

rfourquet: In this case, to emphasize the all-true-tests, perhaps:
```julia
julia> !isconcretetype(Array) && !isabstracttype(Array) && isstructtype(Array) && !isconcretetype(Array{Int}) && isconcretetype(Array{Int,1})
true
```
?

HenriDeh: It's just that I find `false` more readable than a `!` in front of a function.

HenriDeh:
By the way, I did not forget this PR, I just don't know what's missing for it to be merged.

StefanKarpinski:
Added triage to force eyeballs on this and a decision. Might just be a matter of merging it.

HenriDeh:
Since we're here. Any idea why some tests are failing? I'm not very familiar with CI.

StefanKarpinski:
I'm not sure what's going on with CI. I would try rebasing this PR onto current master and force pushing the PR.

HenriDeh:
Sorry I don't have much experience with git and collaborative development. I have my fork opened in gitkraken, what is it exactly that I should do ? I can also just make a new PR

StefanKarpinski:
I went ahead and did it for you. Hopefully CI passes and we can just merge 😁

HenriDeh:
Great thanks, there's still the triage though, unless we decide to leave it as it is because it is not that important.

StefanKarpinski:
At least with a rebase and green CI it will be mergable on the spot if it's approved.

HenriDeh:
buildbot/doctest_linux64 was failing because it was expecting `Vector{Float64}` instead of an `Array{Float64,1}`. That's weird, especially because the `Int64` example right above did not fail. All other tests were successful (see the details of commit https://github.com/JuliaLang/julia/pull/34946/commits/a9165f11829b0785f477d9cc786bda1dc2aba0d6).

Thus I changed Array to Vector and it was buildkite that failing (because of a missing key in .git...). see https://github.com/JuliaLang/julia/pull/34946/commits/c2f9c991dc78442ac8ad0f65e3672f1d954046b6

I reverted the change from Vector to Array in https://github.com/JuliaLang/julia/pull/34946/commits/a0c17c696f680066ca32c77971ac7aa970432253 to be in the exact same state as https://github.com/JuliaLang/julia/pull/34946/commits/a9165f11829b0785f477d9cc786bda1dc2aba0d6 but buildkite still fails. 

So I really don't know what's up with these tests.

JeffBezanson (approved):

JeffBezanson:
Triage approves; please rebase and adjust the doctest output to match the current printing.

HenriDeh:
I squashed all commits together. Checks passed before that so they should pass now as well. Then it's ready to merge \o/

Issue #24444, opened 2017-11-02
Deprecate `ones`?


timholy:
Now that we distinguish `one` and `oneunit`, `ones(T, sz)` seems like a misnomer. Deprecate in favor of `fill(oneunit(T), sz)`? If so, should we ditch `zeros` too?

fredrikekre:
xref  https://github.com/JuliaLang/julia/issues/11557#issuecomment-339776065 and below, and also @Sacha0's PR #24389

Sacha0:
I have work in progress to this effect that I hope to post in the next day or two :). Best!

GunnarFarneback:
To me `fill(oneunit(T), sz)` looks like a non-trivial loss in readability compared to `ones(T, sz)`.

Sacha0:
Note that you rarely need write something as verbose as `fill(oneunit(T), sz)`, as usually a literal or something similarly compact suffices in place of `oneunit(T)`. Shorter incantations may also become possible with future changes to array constructors. Additionally, once you've transitioned to `fill`, you grow fond of it I assure you :). Best!

JeffBezanson:
We could just pick whether `ones` uses `one` or `oneunit`. `ones` and `zeros` should be considered convenience functions, which is ok as long as they have clear meanings in terms of more general functions.

Sacha0:
Plan per triage: Revisit next week, specifically considering moving `ones` and `zeros` to the MATLAB compatibility layer, in favor of `fill` in base. Best!

andyferris:
`ones` and `oneunits` would distinguish the two options. 

Not sure how I feel about moving this and `zeros` to MATLAB compatibility. Just because Mathworks invented it doesn’t mean it’s not a good convenience API that we can be *proud* of. Just briefly - what were the reasons behind this thought? (Sorry, I have to say that triage seems much more efficient, but significantly less transparent when the reasoning isn’t given).

timholy:
It's not a change I'm thrilled about either, but I raised it because it's a logical inconsistency. I think it's fair to say that `ones(T, sz)` implies `fill(one(T), sz)` but really what it does under the hood is `fill(oneunit(T), sz)`.

One option would be to rename it to `oneunits`. An alternative would be to do a horrific swap: `one` -> `algebraic_one` and `oneunit`->`one`. This is breaking and not possible to do via deprecation, which is why I say "horrific."

andyferris:
[...]

PR #34948, opened 2020-03-01
add missing test for ones and zeros

johnnychen94:
I'm actually a little surprised that `eltype(ones(T)) === typeof(one(T))` not hold true for abstract types (e.g., `AbstractFloat`) given that `one(T)` is allowed, so here's one small patch that fixes this.

However, it's also a conflict that `eltype(ones(T)) != T`... so feedbacks are needed.

cc: @kimikage

---

Edit: 

Actually I'm not convinced about this change at all, with two :-1: here I think it's sufficient to just reverted the changes and simply adds some test cases for zeros and ones.

### DIFF
--- a/base/array.jl
+++ b/base/array.jl
@@ -501,15 +501,13 @@ for (fname, felt) in ((:zeros, :zero), (:ones, :one))
         $fname(dims::Tuple{Vararg{DimOrInd}}) = $fname(Float64, dims)
         $fname(::Type{T}, dims::NTuple{N, Union{Integer, OneTo}}) where {T,N} = $fname(T, map(to_dim, dims))
         function $fname(::Type{T}, dims::NTuple{N, Integer}) where {T,N}
-            init = $felt(T)
-            a = Array{typeof(init),N}(undef, dims)
-            fill!(a, init)
+            a = Array{T,N}(undef, dims)
+            fill!(a, $felt(T))
             return a
         end
         function $fname(::Type{T}, dims::Tuple{}) where {T}
-            init = $felt(T)
-            a = Array{typeof(init)}(undef)
-            fill!(a, init)
+            a = Array{T}(undef)
+            fill!(a, $felt(T))
             return a
         end
     end
--- a/base/array.jl
+++ b/base/array.jl
@@ -501,13 +501,15 @@ for (fname, felt) in ((:zeros, :zero), (:ones, :one))
         $fname(dims::Tuple{Vararg{DimOrInd}}) = $fname(Float64, dims)
         $fname(::Type{T}, dims::NTuple{N, Union{Integer, OneTo}}) where {T,N} = $fname(T, map(to_dim, dims))
         function $fname(::Type{T}, dims::NTuple{N, Integer}) where {T,N}
-            a = Array{T,N}(undef, dims)
-            fill!(a, $felt(T))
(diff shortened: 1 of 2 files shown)


kimikage:
> Create an `Array`, with element type `T`

https://docs.julialang.org/en/v1/base/arrays/#Base.ones
https://docs.julialang.org/en/v1/base/arrays/#Base.zeros

You should show the current problem and the benefits of this PR.

johnnychen94:
```julia
julia> X1 = ones(AbstractFloat, 1000, 1000); # Array{AbstractFloat, 2}

julia> X2 = ones(Float64, 1000, 1000);

julia> Y = randn(Float64, 1000, 1000);

julia> @benchmark X1 .* Y
BenchmarkTools.Trial:
  memory estimate:  22.89 MiB
  allocs estimate:  1000011
  --------------
  minimum time:     47.891 ms (0.00% GC)
  median time:      50.460 ms (0.00% GC)
  mean time:        50.898 ms (1.31% GC)
  maximum time:     58.578 ms (2.76% GC)
  --------------
  samples:          99
  evals/sample:     1

julia> @benchmark X2 .* Y
BenchmarkTools.Trial:
  memory estimate:  7.63 MiB
  allocs estimate:  4
  --------------
  minimum time:     1.157 ms (0.00% GC)
  median time:      1.488 ms (0.00% GC)
  mean time:        2.728 ms (45.41% GC)
  maximum time:     22.673 ms (92.12% GC)
  --------------
  samples:          1830
  evals/sample:     1
```

kimikage:
For example, if you need an array of `Float64`, why not specify `Float64` as eltype `T`?
`T` is not always specified directly by the user, but may be determined by promotion mechanism and so on. I think that we should not deprive us of the decision.

timholy:
When `one(T)` and `oneunit(T)` give different answers, Base will error and your version gives something that might surprise some people:
```julia
julia> struct Foo end

julia> Base.one(::Type{Foo}) = 1

julia> Base.oneunit(::Type{Foo}) = Foo()

julia> ones(Foo, 3, 3)
ERROR: MethodError: Cannot `convert` an object of type Int64 to an object of type Foo
Closest candidates are:
  convert(::Type{T}, ::T) where T at essentials.jl:168
Stacktrace:
 [1] fill!(::Array{Foo,2}, ::Int64) at ./array.jl:309
 [2] ones(::Type{Foo}, ::Tuple{Int64,Int64}) at ./array.jl:462
 [3] ones(::Type{Foo}, ::Int64, ::Int64) at ./array.jl:457
 [4] top-level scope at REPL[4]:1

julia> function jc_ones(::Type{T}, dims...) where T
           init = one(T)
           a = Array{typeof(init)}(undef, dims)
           fill!(a, init)
       end
jc_ones (generic function with 1 method)

julia> jc_ones(Foo, 3, 3)
3×3 Array{Int64,2}:
 1  1  1
 1  1  1
 1  1  1
```
Given the decision in #24444, having Base throw an error is probably useful, as it encourages people to specify what they actually want by calling `fill` directly.

Note, however, that packages like Unitful (incorrectly, IMO) specialize `ones` to call `oneunit`. It's cases like these that leave me skeptical about the decision on #24444.

johnnychen94:
Close this as the proposing change doesn't have much practical usage.

PR #34952, opened 2020-03-01
Added tests for Cholesky{T} conversion

ShriprajwalK:


dkarrasch:
It seems like this method is already covered: https://coveralls.io/builds/29064881/source?filename=stdlib/LinearAlgebra/src/cholesky.jl#L373? Or which method are you trying to test here?

ShriprajwalK:
Seems like it got covered recently. They were uncovered when I started writing the test. Sorry for this!

dkarrasch:
Don't worry. Thanks anyway for your efforts!

PR #34953, opened 2020-03-01
Add @inline for anonymous functions

### LINKED ISSUE #34939: Feature request: @inline for anonymous functions
It would be great to be able to do
```
@inline x -> 2x
```
instead, currently:
```
julia> @inline x -> 2x
ERROR: LoadError: x->begin
        #= REPL[14]:1 =#
        2x
    end is not a function expression
Stacktrace:
 [1] error(::Expr, ::String) at ./error.jl:42
 [2] findmeta(::Expr) at ./expr.jl:337
 [3] pushmeta!(::Expr, ::Symbol) at ./expr.jl:273
 [4] @inline(::LineNumberNode, ::Module, ::Any) at ./expr.jl:193
in expression starting at REPL[14]:1
```

ssikdar1:
Closes #34939

added a check for `ex.head === :->` in `findmeta` so that `@inline` will work anonymous functions.

# Testing
```

julia> @inline x -> x^2 + 2x - 1
#1 (generic function with 1 method)

julia> @inline x -> x
#3 (generic function with 1 method)
```

- [X] Add a few tests. 
Added a test in `test/compiler/inline.jl`

- [X] Add a NEWS item.


timholy (approved):

fredrikekre:
Should we also look for
```
@inline f = x -> x
```
expressions?

KristofferC:
IMO, that is not necessary, this doesn't work for example:

```
@inline g = function f(x) x end
```

test/compiler/inline.jl
```diff
@@ -51,6 +51,12 @@ function test_outer(a)
 end
 test_inlined_symbols(test_outer, Tuple{Int64})
 
+# Test case 1 for anonymous functions (issue #34939)
+function test_outer2(x)
+    @inline x -> x^2 + 2x - 1
+end
+test_inlined_symbols(test_outer2, Tuple{Int64})
```
JeffBezanson: This doesn't test that the inline declaration has been applied to the inner function; it tests that the inlining pass produced valid output for `test_outer2`.

ssikdar1: @JeffBezanson 
that makes sense. How do I test if the inline declaration has been applied to the function?
Do I have to look at the `Expr` produced when using the inline macro?

```
julia> Base.parse_input_line("x -> x^2 + 2x - 1")
:($(Expr(:toplevel, :(#= none:1 =#), :((x->begin
          #= none:1 =#
          (x ^ 2 + 2x) - 1
      end)))))

julia> Base.parse_input_line("@inline x -> x^2 + 2x - 1")
:($(Expr(:toplevel, :(#= none:1 =#), :(#= none:1 =# @inline((x->begin
              #= none:1 =#
              (x ^ 2 + 2x) - 1
          end))))))

```

Something like this?

NEWS.md
```diff
@@ -36,6 +36,8 @@ Language changes
 
 * The syntax `(;)` (which was deprecated in v1.4) now creates an empty named tuple ([#30115]).
 
+* `@inline` macro can now be applied to anonymous functions ([#34953]).
```
rfourquet: Maybe 
```suggestion
* `@inline` macro can now be applied to short-form anonymous functions ([#34953]).
```
?

ssikdar1: sure! resolved in c70bec2

test/compiler/inline.jl
JeffBezanson: This is an easy way to test whether the inlining flag is set:
```
julia> ccall(:jl_ast_flag_inlineable, Bool, (Any,), first(methods(x->x)).source)
false
```

ssikdar1: thanks! I added some tests to check `:jl_ast_flag_inlineable` for both anonymous and normal functions in b8ddf2a

JeffBezanson: This test can be removed.

test/compiler/inline.jl
```diff
@@ -275,3 +283,15 @@ let ci = code_typed(f34900, Tuple{Int, Int})[1].first
     @test length(ci.code) == 1 && isexpr(ci.code[1], :return) &&
         ci.code[1].args[1].id == 2
 end
+
+@testset "check jl_ast_flag_inlineable for inline macro" begin
+    ret = ccall(:jl_ast_flag_inlineable, Bool, (Any,), first(methods(@inline x -> x)).source)
+    @test ret == true
```
JeffBezanson: The `ret` variable is redundant here; can just do `@test ccall(...` etc. It is also more idiomatic to use `@test x` and `@test !x` instead of `@test x == true`.

test/compiler/inline.jl
ssikdar1: removed in 57ecb07

test/compiler/inline.jl
ssikdar1: fixed in 57ecb07

JeffBezanson (approved):

JeffBezanson:
Thanks for your patience in working through the review process @ssikdar1 ! LGTM now.

PR #34957, opened 2020-03-01
Add VS Code devcontainer definition

davidanthoff:
This provides integration with [VS Online](https://visualstudio.microsoft.com/services/visual-studio-online/) and [VS Code Remote Containers](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers). It should make building Julia from source quite a bit easier for some user groups (I guess in particular on Windows).

So far I've found the following ways to use this. I think there might be even more because Visual Studio (the full version) might also be integrated with this, but not entirely sure.

## VS Online

VS Online allows one to spin up a dev env in the Azure cloud and then connect to that either with a local VS Code, or via a pure web interface. One can then compile, run stuff in that cloud env. With this PR, this cloud env will be automatically configured in such a way that one can compile Julia right away.

The easiest way to try this right now is to go to VS Online at https://online.visualstudio.com, click `Create environment`, use `https://github.com/davidanthoff/julia/tree/devcontainer` as the `Git Repository` value and then create a new env. Once everything has spun up, one is dropped into a web version of VS Code (all in the browser), and then one just has to select `Tasks: Run Build Tasks...` and select `make`, and one gets a Julia build from source.

One can also use/create such an online env from a local VS Code installation, same story, all the configuration is taken care off. 

## VS Code Remote Containers

If one has VS Code, the Remote Containers extension and docker installed, then the following works: if one opens the julia folder in VS Code, it will offer as an option to open this folder in a container. If one selects that, VS Code automatically creates a docker container that has everything installed to be able to compile Julia, then connects VS Code to that locally running docker container. Again one can super easily compile Julia that way by just selecting the make build task. This should work on all platforms (certainly on Windows, where I tested it).

In both the VS Online and Remote Container scenario, the VS Code that opens will also have useful extensions pre-installed (right now the Julia extension and the C++ extension, there are probably others that would also make sense).


fredrikekre:
Why does this have to be in the Julia repo? Most things like this (e.g. editor specific configs etc) have moved elsewhere.

StefanKarpinski:
I guess because it's useful for working on Julia itself, as opposed to being just for working with it.

davidanthoff:
> Why does this have to be in the Julia repo?

It has to be in the repo here for this to work, that is the design coming from VS Code for this.

> I guess because it's useful for working on Julia itself, as opposed to being just for working with it.

Exactly. The idea is that the information on how to setup your dev environment to compile Julia (which currently is in the README and the build instructions) is included in machine readable form in the repo itself. If you then open that folder with an editor that supports this, the editor can automatically spin up an environment that has everything installed that you need to compile and work with that repo.

Maybe I should remove the stuff in the `.vscode` folder? That is kind of separate, i.e. just a VS Code task definition to call `make`. Not really needed, one can just call `make` in the VS Code REPL as well.

ViralBShah:
Would it be possible to have the VS Code stuff in a separate repo, and have it pull in julia sources upon spinning up?

davidanthoff:
> Would it be possible to have the VS Code stuff in a separate repo, and have it pull in julia sources upon spinning up?

No, the devcontainer stuff has to be in the repo that you want to edit/build.

I removed the VS Code task definition, that really doesn't add much and one can easily add that individually. So this now just has the definition of the container that should be spun up when editing this repo.

I also tested this now and it works everywhere: Windows, Mac, Linux and VS Online. You just open the folder in VS Code, wait a bit and have a fully ready environment where you can just type `make` and it will compile everything.

davidanthoff:
Oh, and we could obviously also get rid of the `Dockerfile`, if we just uploaded that docker image to docker hub, so at that point the only thing added to the repo here would be the `devcontainer.json` file.

But I almost think the way it is done now in this PR is easier, that way we don't have to maintain yet another repo with the docker image definition.

davidanthoff:
I think this could just be merged now?

davidanthoff:
https://discourse.julialang.org/t/compile-julia-from-source-with-ease-in-vs-code-even-on-windows/35696

IanButterworth:
Just noticing this root folder unchanged since merge and wondered if it's still functional/valuable?

giordano:
I've used the devcontainer in codespaces a couple of times, when I didn't have a working building setup locally. The devcontainer doesn't require much maintenance.

PR #34927, opened 2020-02-29
add readline(::AbstractCmd)

bicycle1885:
This adds the `readline(::AbstractCmd)` method in a trivial way, unless I miss something non-trivial. This closes #28975.

PR #34958, opened 2020-03-01
Revert "add readline(::AbstractCmd)"

Keno:
This broke the buildbots. Reverts JuliaLang/julia#34927.


JeffBezanson:
How did this cause win64 to fail?

Keno:
The test keeps a BusyBox command running and doesn't wait for it to exit. On win64 we delete BusyBox at the end of that testset which fails if the command is still running.

bicycle1885:
Ah, so we need to terminate the process before returning the line string.

Keno:
Yes, that would be one valid solution (which I think would be ok).

PR #34959, opened 2020-03-02
Docs on latest LLVM

PallHaraldsson:
[skip ci]


PallHaraldsson:
I didn't check this. LLVM 10 is now in RC2, so possibly 10.0.0-rc2 would be needed? I noticed a D compiler just changed to using 10 (on master) and we use 9. Possibly this should only be merged after LLVM 10 released, or ok, as is, as only docs?

ViralBShah:
Please make it refer to the current llvm version in use.

vchuravy:
I see not that much value in keeping that particular number up to date. It is just an example on how to set a different version of LLVM to build with.

PR #34965, opened 2020-03-02
Added test for factorizing LDLt

ShriprajwalK:


dkarrasch (changes requested):
LGTM. I suggested a few more tests which should help cover the uncovered conversion methods.

I realized that this test file is not run at all. So we should add it to `testgroups.jl`. This will make the PR as is fail, because the REPL printing tests fails because the `getproperty` function is broken. So we could (a) comment the REPL printing testset out, or (b) leave this PR hanging around until the `getproperty` and the printing is fixed. I'd vote for option (a), to get started with this file being tested.

ShriprajwalK:
How do I go about making the REPL comment this testset out? Where can we get more details about the bug in the `getproperty` function?
Thanks a lot!

stdlib/LinearAlgebra/test/ldlt.jl
```diff
@@ -15,4 +15,11 @@ Random.seed!(123)
     @test ldltstring == "$(summary(F))\nL factor:\n$lstring\nD factor:\n$dstring"
 end
```
dkarrasch:
```suggestion
# @testset "REPL printing of LDLT" begin
#    S = SymTridiagonal(randn(5), randn(4))
#    F = ldlt(S)
#    ldltstring = sprint((t, s) -> show(t, "text/plain", s), F)
#    lstring = sprint((t, s) -> show(t, "text/plain", s), F.L)
#    dstring = sprint((t, s) -> show(t, "text/plain", s), F.D)
#     @test ldltstring == "$(summary(F))\nL factor:\n$lstring\nD factor:\n$dstring"
# end
```

dkarrasch: This is how you comment out code.

dkarrasch:
I wasn't able to find any specialized `show` method that displays `LDLt` objects. But before that, I think we'll need to fix the `getproperty` method. Perhaps, `LDLt`s don't display nicely exactly because the `getproperty` calls are failing in the `show` method.

In any case, @ShriprajwalK, could you please add the ldlt file to the `testgroups.jl`? As I said, this will make this PR definitely fail in its current state (unless you comment out the first testset), but we need to have this being run in principle during CI.

stdlib/LinearAlgebra/test/ldlt.jl
ShriprajwalK: Oh, that's what you meant. Sorry!

stdlib/LinearAlgebra/test/ldlt.jl
```diff
@@ -13,6 +13,20 @@ Random.seed!(123)
...
     @test ldltstring == "$(summary(F))\nL factor:\n$lstring\nD factor:\n$dstring"
-end
+# @testset "REPL printing of LDLT" begin
+#    S = SymTridiagonal(randn(5), randn(4))
+#    F = ldlt(S)
+#    ldltstring = sprint((t, s) -> show(t, "text/plain", s), F)
+#    lstring = sprint((t, s) -> show(t, "text/plain", s), F.L)
+#    dstring = sprint((t, s) -> show(t, "text/plain", s), F.D)
+#     @test ldltstring == "$(summary(F))\nL factor:\n$lstring\nD factor:\n$dstring"
+# end
```
dkarrasch: Sorry, I guess my code suggestion was misleading. Please don't add these lines, but rather comment out the ones above.

stdlib/LinearAlgebra/test/testgroups
```diff
@@ -24,3 +24,4 @@ pinv
 givens
 structuredbroadcast
 addmul
+ldlt
```
dkarrasch: I think `addmul` is supposed to be on the last line, so please insert `ldlt` somewhere in the middle.

dkarrasch (changes requested):

dkarrasch:
@ShriprajwalK My sincere apologies, I messed it up completely. The `getproperty` function got fixed/changed somewhere between julia v1.3 and current master, and I was trying to run the tests on v1.3. Meanwhile, @kshyatt prepared #35018, which is ready to be merged.

dkarrasch:
Closed by 3f5c899. I'm very sorry for my misleading comments.

PR #34970, opened 2020-03-03
Warn if setting env variable is ineffective during a julia session

IanButterworth:
A suggestion to warn of ineffective changes to env variables during a julia session. 

Checks for:
- `JULIA_NUM_THREADS` which seems to be a common error
- `JULIA_PROJECT`

There may be more, I went by the list in https://docs.julialang.org/en/v1/manual/environment-variables/#Environment-Variables-1

### DIFF
--- a/base/env.jl
+++ b/base/env.jl
@@ -1,5 +1,8 @@
 # This file is a part of Julia. License is MIT: https://julialang.org/license
 
+# Environment variables that must be set before julia launches
+const sessionineffectives = ["JULIA_NUM_THREADS", "JULIA_PROJECT"]
+
 if Sys.iswindows()
     const ERROR_ENVVAR_NOT_FOUND = UInt32(203)
 
@@ -21,6 +24,7 @@ if Sys.iswindows()
     end
 
     function _setenv(svar::AbstractString, sval::AbstractString, overwrite::Bool=true)
+        in(svar, sessionineffectives) && @warn "To take effect, $svar must be set before launching a julia session"
         var = cwstring(svar)
         val = cwstring(sval)
         if overwrite || !_hasenv(var)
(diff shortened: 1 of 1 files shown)


yuyichao:
1. Unless explicitly mentioned, you should always assume environment variables are read only ones and needs to be set before it is read. These are far from the environment variables that can have this issue and most of them you won't even be able to warn against.
2. There's nothing intrinsically wrong with doing this and have a warning like this at such a level is going to be very annoy for code that deal with environment variables as black box. (Say clear the env temporarily and restore them all back)
3. This is at most a documentation problem. The page is pretty clear about it. Since that's clearly not enough, it should probably be added to the document for each individual env as well.

ianshmean:
1 & 2, fair points.
3. There are currently note boxes for each of the effected environment variables in the docs (I believe I added them back in the day when I suffered the same confusion)

To be honest, I've seen some mentions of intentions for julia to be able to dynamically use more threads than at start up, so perhaps this would soon be a moot issue

yuyichao:
> To be honest, I've seen some mentions of intentions for julia to be able to dynamically use more threads than at start up, so perhaps this would soon be a moot issue

Well, not necessarily. When that happens, this variable and/or a corresponding cmdline option, will probablby be used to set a initial thread count. And even if the thread can be dynamically changed, it won't be controlled by a ENV and there's no way setting a ENV can change the thread count on the fly. In fact, a ENV would be the worst way to do it given how non-thread safe it is...

ianshmean:
That's what I mean. There would likely be new functions for dynamically changing number of threads, after start up. I meant moot as in not relevant/not worth discussing (that term has confusingly opposite meanings..)

johnnychen94:
Does it help to only throw such warnings when `isinteractive() == true` and/or the caller isn't a Base function(via `source_path`)?

yuyichao:
> Does it help to only throw such warnings when `isinteractive() == true`

That's not useful for startup.


> the caller isn't a Base function(via source_path-- not sure if it's doable )?

That's not even well defined and is unclear what it is trying to do anyway. Base functions have nothing special.

ianshmean:
Closed because the idea didn't stick

PR #34979, opened 2020-03-03
Added test for Matrix(F::LU{T,Triadiagonal{T,V}})

ShriprajwalK:
Added test for https://codecov.io/gh/julialang/julia/src/master/stdlib/LinearAlgebra/src/lu.jl#L732

### DIFF
--- a/stdlib/LinearAlgebra/test/lu.jl
+++ b/stdlib/LinearAlgebra/test/lu.jl
@@ -92,6 +92,7 @@ dimg  = randn(n)/2
         @test lud.L*lud.U ≈ lud.P*Array(d)
         @test lud.L*lud.U ≈ Array(d)[lud.p,:]
         @test AbstractArray(lud) ≈ d
+        @test Array(lud) ≈ d
     end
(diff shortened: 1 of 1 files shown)


dkarrasch:
Actually, could you find where the method one line above

https://codecov.io/gh/julialang/julia/src/master/stdlib/LinearAlgebra/src/lu.jl#L731

is tested, and add a single line to test the `Array` conversion? That would be great. Not sure we need six new lines for testing this simple conversion.

dkarrasch:
Sorry, I wasn't being very clear. I meant to suggest to remove your entire new testset and to insert
```julia
@test Array(alu) == fa
```
after line 234 (and maybe also
```julia
@test Array(lud) ≈ d
```
after line 94) instead. That's a minimal change, and should do the job just as well.

dkarrasch:
Ok, Windows failure is weird but common, and the additions are actually not imposing any new/independent tests, they just run by different methods and thereby cover them.

ShriprajwalK:
Every PR recently made has failed for Windows for some reason.
Is there any change I should make for this PR?

PR #34980, opened 2020-03-03
dump: remove dependency on uid field [NFCI]

vtjnash:
Replaces `uid != 0` with `isconcretetype` and the `O(n^2)` iteration in jl_recache_types with a hash table. Addresses part of #34890

### DIFF
--- a/src/dump.c
+++ b/src/dump.c
@@ -51,6 +51,7 @@ static arraylist_t backref_list;
 // type-rewriting of some sort
 // (not used in MODE_IR)
 static arraylist_t flagref_list;
+static htable_t uniquing_table;
 
 // list of (size_t pos, (void *f)(jl_value_t*)) entries
 // for the serializer to mark values in need of rework by function f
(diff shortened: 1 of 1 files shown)


src/dump.c
```diff
@@ -257,16 +259,16 @@ static int type_recursively_external(jl_datatype_t *dt);
...
+    if (!jl_is_concrete_type(p0))
         return 0;
+    jl_datatype_t *p = (jl_datatype_t*)p0;
+    //while (jl_is_unionall(p)) {
+    //    if (!type_parameter_recursively_external(((jl_unionall_t*)p)->var->lb))
+    //        return 0;
+    //    if (!type_parameter_recursively_external(((jl_unionall_t*)p)->var->ub))
+    //        return 0;
+    //    p = (jl_datatype_t*)((jl_unionall_t*)p)->body;
+    //}
```
JeffBezanson: delete?

vtjnash: It seemed probably good, but wasn't really functional anymore, and we're not to the point yet where this is a performance bottleneck.

JeffBezanson (approved):

PR #34982, opened 2020-03-03
completion to make use of Docs.keywords.keys

### LINKED ISSUE #34950: REPL doc completion uses its own list of keywords
Here:
https://github.com/JuliaLang/julia/blob/ea067fb221769a43221bf422508f0f5d70fb5fde/stdlib/REPL/src/docview.jl#L579

I believe it should use the keys of `Base.Docs.keywords` instead.

suraj-shivu:
Fixes #34950

### DIFF
--- a/stdlib/REPL/src/docview.jl
+++ b/stdlib/REPL/src/docview.jl
@@ -574,7 +574,6 @@ print_correction(word) = print_correction(stdout, word)
 
 # Completion data
 
-const builtins = collect(keys(Base.Docs.keywords))
 
 moduleusings(mod) = ccall(:jl_module_usings, Any, (Any,), mod)
 
--- a/stdlib/REPL/src/docview.jl
+++ b/stdlib/REPL/src/docview.jl
@@ -582,7 +582,7 @@ filtervalid(names) = filter(x->!occursin(r"#", x), map(string, names))
 accessible(mod::Module) =
     [filter!(s -> !Base.isdeprecated(mod, s), names(mod, all = true, imported = true));
      map(names, moduleusings(mod))...;
-     builtins] |> unique |> filtervalid
+     collect(keys(Base.Docs.keywords))] |> unique |> filtervalid
 
 doc_completions(name) = fuzzysort(name, accessible(Main))
 doc_completions(name::Symbol) = doc_completions(string(name))
--- a/stdlib/REPL/src/docview.jl
+++ b/stdlib/REPL/src/docview.jl
@@ -574,12 +574,7 @@ print_correction(word) = print_correction(stdout, word)
 
 # Completion data
 
-const builtins = ["abstract type", "baremodule", "begin", "break",
-                  "catch", "ccall", "const", "continue", "do", "else",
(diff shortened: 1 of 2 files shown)


JeffBezanson:
I think this should work. Maybe add a test in REPL/test/replcompletions.jl?

suraj-shivu:
Sure, I'll add it now.

stdlib/REPL/src/docview.jl
```diff
@@ -574,12 +574,7 @@ print_correction(word) = print_correction(stdout, word)
 
 # Completion data
 
-const builtins = ["abstract type", "baremodule", "begin", "break",
-                  "catch", "ccall", "const", "continue", "do", "else",
-                  "elseif", "end", "export", "finally", "for", "function",
-                  "global", "if", "import", "let",
-                  "local", "macro", "module", "mutable struct", "primitive type",
-                  "quote", "return", "struct", "try", "using", "while"]
+const builtins = collect(keys(Base.Docs.keywords))
```
fredrikekre:
```suggestion
```
and just put `collect(keys(Base.Docs.keywords))` directly on L586

StefanKarpinski:
Testing that this is equal to the thing that it's set to feels kind of vacuous. Also, note that this list contains a lot of stuff that it probably shouldn't:
```jl
julia> sort!(collect(keys(Base.Docs.keywords)))
63-element Array{Symbol,1}:
 Symbol("")
 Symbol("#")
 Symbol("#=")
 :$
 :&&
 Symbol("'")
 Symbol("(")
 Symbol("()")
 Symbol(")")
 :...
 :(::)
 Symbol(";")
 :(=)
 Symbol("=#")
 :?
 Symbol("@var_str")
 :Julia
 Symbol("Union{}")
 Symbol("[")
 Symbol("[]")
 Symbol("]")
 :__init__
 ⋮
 :import
 :julia
 :let
 :local
 :macro
 :module
 Symbol("mutable struct")
 :new
 Symbol("primitive type")
 :quote
 :return
 :struct
 :try
 :using
 Symbol("var\"name\"")
 :where
 :while
 Symbol("{")
 Symbol("{}")
 :||
 Symbol("}")
```

stdlib/REPL/src/docview.jl
suraj-shivu: This seems much more efficient. Thank you!!

suraj-shivu:
@StefanKarpinski  May I know what it shouldn't contain or on how I should go about testing the variable as I'm pretty new to the language.

JeffBezanson:
IIUC, the goal of this is just to search over everything visible that we have doc strings for. So it seems to me everything should be included.

suraj-shivu:
Hi, how would I go about writing tests for this because some of the symbols have open issues as they do not give the same output as stored in their docStrings when put as input in REPL helpmode E.g. ";".So what do I do to access these docStrings?

JeffBezanson:
Find the function that's used to look up help completions, and test that it can find something that it couldn't before.

suraj-shivu:
Added the changes as requested.

PR #34984, opened 2020-03-03
Fix `$(PYTHON)` paths for native windows executable or cygwin

staticfloat:
We auto-detect by inspecting the output of `platform.system()`

### DIFF
--- a/Make.inc
+++ b/Make.inc
@@ -1018,7 +1018,7 @@ PYTHON_SYSTEM := $(shell $(PYTHON) -c 'from __future__ import print_function; im
 # If we're running on Cygwin, but using a native-windows Python, we need to use cygpath -w
 ifneq ($(and $(filter $(PYTHON_SYSTEM),Windows),$(findstring CYGWIN,$(BUILD_OS))),)
 define invoke_python
-$(PYTHON) "$$(shell cygpath -w "$(1)")"
+$(PYTHON) "$$(cygpath -w "$(1)")"
 endef
 else
 define invoke_python
--- a/Make.inc
+++ b/Make.inc
@@ -1018,7 +1018,7 @@ PYTHON_SYSTEM := $(shell $(PYTHON) -c 'from __future__ import print_function; im
 # If we're running on Cygwin, but using a native-windows Python, we need to use cygpath -w
 ifneq ($(and $(filter $(PYTHON_SYSTEM),Windows),$(findstring CYGWIN,$(BUILD_OS))),)
 define invoke_python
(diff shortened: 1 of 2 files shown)


Make.inc
```diff
@@ -1013,13 +1013,23 @@ endif
 # We need python for things like BB triplet recognition.  We don't really care
 # about version, generally, so just find something that works:
 PYTHON := $(shell which python 2>/dev/null || which python3 2>/dev/null || which python2 2>/dev/null || echo not found)
+PYTHON_SYSTEM := $(PYTHON) -c 'from __future__ import print_function; import platform; print(platform.system())'
+ifneq ($(filter $(call lowercase,$(PYTHON_SYSTEM)),cygwin),)
```
Keno: `$(shell $(PYTHON_SYSTEM))`?

vtjnash (approved):

Make.inc
```diff
@@ -1012,14 +1012,26 @@ endif
...
 # We need python for things like BB triplet recognition.  We don't really care
 # about version, generally, so just find something that works:
-PYTHON := $(shell which python 2>/dev/null || which python3 2>/dev/null || which python2 2>/dev/null || echo not found)
+PYTHON := "$(shell which python 2>/dev/null || which python3 2>/dev/null || which python2 2>/dev/null || echo not found)"
+PYTHON_SYSTEM := $(shell $(PYTHON) -c 'from __future__ import print_function; import platform; print(platform.system())')
+
+# If we're running on Cygwin, but using a native-windows Python, we need to use cygpath -w
+ifneq ($(and $(filter $(PYTHON_SYSTEM),Windows),$(findstring CYGWIN,$(BUILD_OS))),)
+define invoke_python
+$(PYTHON) "$(shell cygpath -w "$(1)")"
```
vtjnash:
```suggestion
$(PYTHON) "$${cygpath -w "$(1)"}"
```

staticfloat: I'm not sure what the curly braces here mean?

vtjnash: They're the same as backticks for the shell

staticfloat: I get "bad substitution" when I try that, because the `$$` escapes it all the way down to `bash`.  Here's an example:

```
$ cat Makefile 
define foo
echo $${echo $(1) | grep b}
endef

all:
        $(call foo,bar)
```

```
$ make
echo ${echo bar | grep b}
/bin/sh: ${echo bar | grep b}: bad substitution
make: *** [all] Error 1
```

I think it's sufficient to just use `$$(cygpath -w ...)`.

PR #28526, opened 2018-08-08
fix (\) SparseVector in SuiteSparse.CHOLMOD

pochoi:
(\\) for Factor and SparseVecOrMat has the following method error:
```julia
julia> using SparseArrays, LinearAlgebra
julia> sparseI = sparse(1.0I, 100, 100);
julia> sparseb = sprandn(100, 0.5); #SparseVector
julia> chI = cholesky(sparseI); #Factor
julia> chI \ sparseb ≈ sparseb #(\)(Factor, SparseVecOrMat)
ERROR: MethodError: no method matching SuiteSparse.CHOLMOD.Sparse(::SparseVector{Float64,Int64}, ::Int64)
Closest candidates are:
  SuiteSparse.CHOLMOD.Sparse(::Integer, ::Integer, ::Array{Int64,1}, ::Array{Int64,1}, ::Array{#s608,1} where #s608<:Union{Complex{Float64}, Float64}) at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v0.7/SuiteSparse/src/cholmod.jl:882
  SuiteSparse.CHOLMOD.Sparse(::Integer, ::Integer, ::Array{Int64,1}, ::Array{Int64,1}, ::Array{Tv<:Union{Complex{Float64}, Float64},1}, ::Any) where Tv<:Union{Complex{Float64}, Float64} at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v0.7/SuiteSparse/src/cholmod.jl:848
  SuiteSparse.CHOLMOD.Sparse(::SparseMatrixCSC{Tv<:Union{Complex{Float64}, Float64},Int64}, ::Integer) where Tv<:Union{Complex{Float64}, Float64} at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v0.7/SuiteSparse/src/cholmod.jl:896
  ...
Stacktrace:
 [1] \(::SuiteSparse.CHOLMOD.Factor{Float64}, ::SparseVector{Float64,Int64}) at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v0.7/SuiteSparse/src/cholmod.jl:1718
 [2] top-level scope at none:0
```

The method error is due to the call `Sparse(A::SparseMatrixCSC{Tv,SuiteSparse_long}, stype::Integer)`. 
[...]

PR #34989, opened 2020-03-04
Rebase of #28526 on master: Sparse cholesky solve with sparse RHS

ViralBShah:
Rebase #28526 on top of master.  

cc @pochoi


dkarrasch (commented):
LGTM, though I should add that I'm not familiar with the specific `spsolve` calls. Shall we have another non-trivial test with a small, densely populated sparse SPD matrix, and compare against the classic dense solution?

ViralBShah:
I think we should not need more tests because this only handles the case of the sparse rhs by making it dense and thus reducing the missing case to an existing case.

ViralBShah:
CI failures are unrelated. Merging.

pochoi:
Thank you so much!

PR #34994, opened 2020-03-04
Added tests for rdiv! in LU

ShriprajwalK:
Added test for https://codecov.io/gh/julialang/julia/src/master/stdlib/LinearAlgebra/src/lu.jl#L686 and 687.
I don't think we need all the 4 tests. I was confused about writing hardcoded values in tests or random values.
What is done in general?
Thanks!

### DIFF
--- a/stdlib/LinearAlgebra/test/lu.jl
+++ b/stdlib/LinearAlgebra/test/lu.jl
@@ -344,18 +344,12 @@ end
 end
 
 @testset "more rdiv! methods" begin
-    B = randn(Float16, 5, 5)
-    A = transpose(lu(B))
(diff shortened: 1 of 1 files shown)


dkarrasch:
Yeah, the exact tests are doomed to fail. How about something a bit more "elegant" and generic?
```julia
for elty in (Float16, Float64, ComplexF64), transform in (transpose, adjoint)
    A = randn(elty, 5, 5)
    C = copy(A)
    B = randn(elty, 5, 5)
    luB = lu(B)
    @test rdiv!(transform(A), transform(luB)) ≈ transform(C) / transform(B)
end
```

PR #34996, opened 2020-03-04
Emit structs as LLVM structures

thomasfaingnaert:
AFAICT, there is no way to call LLVM intrinsics with structure arguments/return values with `ccall`.
This PR changes the LLVM representation of Julia structs to LLVM structures (instead of LLVM arrays).

// cc: @maleadt

### DIFF
--- a/src/cgutils.cpp
+++ b/src/cgutils.cpp
@@ -733,8 +733,10 @@ static Type *_julia_struct_to_llvm(jl_codegen_params_t *ctx, jl_value_t *jt, jl_
         else if (isarray && !type_is_ghost(lasttype)) {
             if (isTuple && isvector && jl_special_vector_alignment(ntypes, jlasttype) != 0)
                 struct_decl = VectorType::get(lasttype, ntypes);
-            else
+            else if (isTuple)
                 struct_decl = ArrayType::get(lasttype, ntypes);
+            else
+                struct_decl = StructType::get(jl_LLVMContext, latypes);
         }
         else {
 #if 0 // stress-test code that tries to assume julia-index == llvm-index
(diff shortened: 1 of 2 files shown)


src/cgutils.cpp
```diff
@@ -733,8 +733,10 @@ static Type *_julia_struct_to_llvm(jl_codegen_params_t *ctx, jl_value_t *jt, jl_
         else if (isarray && !type_is_ghost(lasttype)) {
             if (isTuple && isvector && jl_special_vector_alignment(ntypes, jlasttype) != 0)
                 struct_decl = VectorType::get(lasttype, ntypes);
-            else
+            else if (isTuple)
```
vtjnash:
```suggestion
            else if (isTuple || !llvmcall)
```

src/cgutils.cpp
```diff
@@ -733,8 +733,10 @@ static Type *_julia_struct_to_llvm(jl_codegen_params_t *ctx, jl_value_t *jt, jl_
...
                 struct_decl = ArrayType::get(lasttype, ntypes);
+            else
```
KristofferC: Should inhomogeneous tuples (where the elements happen to be the same size) also be passed as `structs`, e.g, pass `(::Bool, ::Int8)` as a struct? It is now passed as a `[2 x i8]`.

thomasfaingnaert: That seems to be because https://github.com/JuliaLang/julia/blob/296223b52b9fd867f6801c1092cd5a68261e5502/src/cgutils.cpp#L717 checks the ``llvm::Type`s, and not the `jl_value_t`s of the arguments.

I'm with leaving it as is for now, since you can obtain the same behaviour now using
```julia
struct Foo
    x::Bool
    y::Int8
end
```

PR #34998, opened 2020-03-04
Distributed: add missing check on return code

vtjnash:
We were failing to check the return value, which meant we'd throw it on the subsequent line instead and self-destruct that worker.

This got lost when enabling on OSX in 0394f47f455defd5758dc7031266a0eb0adebc57


vtjnash:
Hm, I see both freebsd and macos printed some error message running Distributed test (e.g.  "Error: Fatal error on process 1"), but succeeded???

Keno:
This seems pretty obviously correct to me. What I'd like to do here is merge the two CI improvements PRs I have pending and then rebase this and see what CI says.

PR #35000, opened 2020-03-04
fix #34921, keywordargs test breaks precompile test

### LINKED ISSUE #34921: keywordargs testset breaks precompile testset
If you run the test set in a isolated build environment (i.e., no external network) then they run sequentially in the same process and several fail due to bad interactions between them.

One of these failures is that the [keywordargs.jl](../blob/release-1.3/test/keywordargs.jl#L263) declaration
```julia
struct T20804{T}
    y::T
end
(f::T20804)(;x=10) = f.y + x
```
causes the following in [precompile.jl](../blob/release-1.3/test/precompile.jl#L168) to not throw the expected exception
```julia
@test_throws ErrorException Core.kwfunc(Base.nothing)
```

You can test these two are incompatible by just putting them back-to-back in a _test.jl_ file and running them
```julia
using Test

struct T20804{T}
    y::T
end
(f::T20804)(;x=10) = f.y + x

@test_throws ErrorException Core.kwfunc(Base.nothing)
```
```bash
$ julia test.jl
```
```
Test Failed at /build/test.jl:8
  Expression: Core.kwfunc(Base.nothing)
    Expected: ErrorException
  No exception thrown
in expression starting at /build/test.jl:8
```

If you want to see it with the build system, and you don't have an isolated environment to build things in, I expect you can force tests to run sequentially by manually hacking [test/runtests.jl](../blob/release-1.3/test/runtests.jl#L62)
```
cd(@__DIR__) do
    n = 1
    if net_on
        n = min(Sys.CPU_THREADS, length(tests))
        n > 1 && addprocs_with_testenv(n)
        LinearAlgebra.BLAS.set_num_threads(1)
    end
    skipped = 0
```
to not set `n>1` (e.g., change `net_on` to `false`).

JeffBezanson:
fix #34921

### DIFF
--- a/test/precompile.jl
+++ b/test/precompile.jl
@@ -87,9 +87,9 @@ try
               (::Task)(::UInt8, ::UInt16, ::UInt32) = 2
 
               # issue 16471 (capturing references to a kwfunc)
-              Test.@test !isdefined(Base.Nothing.name.mt, :kwsorter)
-              Base.nothing(::UInt8, ::UInt16, ::UInt32; x = 52) = x
-              const nothingkw = Core.kwfunc(Base.nothing)
+              Test.@test !isdefined(typeof(sin).name.mt, :kwsorter)
+              Base.sin(::UInt8, ::UInt16, ::UInt32; x = 52) = x
+              const sinkw = Core.kwfunc(Base.sin)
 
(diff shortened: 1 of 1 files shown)


vtjnash (approved):
LGTM. I think I used `nothing` here back before the function table merger.

PR #35001, opened 2020-03-04
fix #34105, unescape triple-quoted strings after dedenting

### LINKED ISSUE #34105: Unindentation in triple-quoted strings affected by explicitly written newline (\n)
```julia
let str = "\n   yoyo"
        println("1: ")
        println("""begin line: interpolate:$str
        begin next line:""")
end

let str = "   yoyo"
        println("2: ")
        println("""begin line: interpolate:\n$str
        begin next line:""")
end
```

produces
```
1: 
begin line: interpolate:
   yoyo
begin next line:
2: 
begin line: interpolate:
   yoyo
        begin next line:
```

But I had expected both blocks of outputs under `1:` and `2:` to be the same. I did not expect `"\n"` (in the string) to be treated the same as `\n` (in the file) within a triple-quoted string.

(kind of dual error of https://github.com/JuliaLang/julia/issues/7165)

### LINKED ISSUE #34967: Handle invalid UTF-8 in code
I use this code to produce a jl file that has an invalid UTF-8 sequence in it:
```julia
ba = UInt8['a', 245, 'b']
s = String(ba)
f = open("foo.jl", "w")
println(f, "println('a')")
println(f, "# Hallo", s)
println(f, "println('c')")
close(f)
```
When I run this with `julia foo.jl`, I get this as output:
```
a
```
So it seems as if Julia crashes/exits when it encounters the invalid UTF-8.

My sense from a Slack conversation was that @StefanKarpinski thought that Julia should probably just detect these cases early on and show an error message.

JeffBezanson:
It seems clear to me that triple-quoted strings should only specially handle whitespace that directly occurs in the source text, not including escape sequences. This also makes it easier to fix #34967, since if unescaping is done first then you get a UTF-8 error when we try to strip whitespace if you write invalid UTF-8 strings using escape sequences (which is allowed of course).
But, this could be breaking since it obviously changes the values of some string literals.

fix #34105

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -36,6 +36,16 @@ Language changes
 
 * The syntax `(;)` (which was deprecated in v1.4) now creates an empty named tuple ([#30115]).
 
+* In triple-quoted string literals, whitespace stripping is now done before processing
+  escape sequences instead of after. For example, the syntax
+  ```
+  """
+    a\n b"""
+  ```
+  used to yield the string " a\nb", since the single space before `b` set the indent level.
+  Now the result is "a\n b", since the space before `b` is no longer considered to occur
+  at the start of a line. The old behavior is considered a bug ([#35001]).
+
 Multi-threading changes
 -----------------------
 
--- a/doc/src/manual/strings.md
+++ b/doc/src/manual/strings.md
@@ -642,7 +642,7 @@ julia> """
 
 Trailing whitespace is left unaltered.
 
-Triple-quoted string literals can contain `"` symbols without escaping.
+Triple-quoted string literals can contain `"` characters without escaping.
 
 Note that line breaks in literal strings, whether single- or triple-quoted, result in a newline
 (LF) character `\n` in the string, even if your editor uses a carriage return `\r` (CR) or CRLF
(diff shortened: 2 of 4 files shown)


Keno:
Yes, I'm in favor. I came across this when implementing the unindenting in one of the pure Julia parsers (can't remember which one) and thought it was an odd behavior. I concur with your assessment that this needs to happen in the opposite order.

StefanKarpinski:
While this is a behavior change, it's hard to argue that this wasn't a bug. 💯

davidanthoff:
@ZacLN probably something we need to replicate in CSTParser as well?

GunnarFarneback:
This PR changed the parsing of
```
"""\n
"""
```
from `\n` to `\n\n`.

Is this unavoidable collateral damage or something that should be fixed?

(Real code example that gets an extra newline: https://github.com/JuliaRegistries/RegistryTools.jl/blob/cbebbed17e13e7a18556a7dd62346d8610e867ea/src/types.jl#L126)

StefanKarpinski:
It's unclear to me why that had the `\n` after the opening `"""` in the first place.

GunnarFarneback:
It's unclear for me too, but one reason might be that
```
julia> """
       \na
       """
"       \na\n       "
```
before this PR. That's another thing that has changed, probably for the better.

Issue #34666, opened 2020-02-05
`@sync (@spawn @spawn f(); @spawn @spawn g())` is not thread-safe?


tkf:
`@sync @spawn @spawn f()` expands roughly to

```julia
let sync = Any[]
    task1 = Task() do
        task2 = Task() do
            f()
        end
        task2.sticky = false
        push!(sync, task2)
        schedule(task2)
        task2
    end
    task1.sticky = false
    push!(sync, task1)
    schedule(task1)
    task1
end
```

This code itself is OK. However, I don't think the following code is thread-safe:

```julia
@sync begin
    @spawn @spawn f()
    @spawn @spawn g()
end
```

This is because `push!(sync, task2)` for `@spawn f()` and `@spawn g()` can be executed in parallel.  This came up in https://github.com/JuliaLang/julia/pull/34543#discussion_r374970720

How should it be fixed? Maybe expand `@spawn ...` to something like below?

```julia
if $(Expr(:islocal, :sync))
    local sync_sub = Any[]
    task = Task() do
        local sync = sync_sub
        ...
    end
else
    task = Task() do
        ...
    end
end
task.sticky = false

if $(Expr(:islocal, :sync))
    push!(sync, task)
    @async begin
        wait(task)
        append!(sync, sync_sub)  # merge sub-tasks
    end
end

schedule(task)
```

JeffBezanson:
A simpler approach would be to lock it, and repeatedly pop a task from the vector and wait for it until the vector is empty.

I assume that sync should wait for *every* spawn inside it, but maybe a case can be made that it should only wait for the outermost level. The reason would be that then you know the full set of tasks to wait on by the time you reach the end of the sync block. Is that property useful at all?

tkf:
Actually this (rather contrived) example shows that the approach I posted above is not a correct fix:

```julia
fib(x) =
    @sync begin
        function f(x)
            x < 0 && error("Please give me an integer.")
            x in (0, 1) && return x
            a = @spawn f(x - 2)
            b = @spawn f(x - 1)
            return fetch(a) + fetch(b)
        end
        f(x)
    end
```

But I think it's conceivable to use a recursive closure with `@spawn` wrapped with `@sync` like this so that you get a good error report.  In such case, I think it makes sense to track all the nested tasks.

> I assume that sync should wait for _every_ spawn inside it, but maybe a case can be made that it should only wait for the outermost level.

[...]

PR #32477, opened 2019-07-02
remove `threadedregion` and move `jl_threading_run` to julia

JeffBezanson:
Putting this up to see if we can find any adversarial test cases.

Resolves #32970

PR #35003, opened 2020-03-04
RFC: Make Threads.@threads use the PARTR scheduler

MasonProtter:
Hello, I've occassionally encountered confusion online from people who think that `Threads.@threads` is using the new depth first scheduling routine and run into problems when trying to nest `Threads.@threads` loops within other multi-threaded code. 

I wrote up a quick macro that roughly mimics the behaviour of `Threads.@threads` but instead uses `Threads.@spawn` inside, but will only spawn at most `Threads.nthreads() + 1` tasks. 

Here's an implementation of the macro that should work on Julia 1.3+ (functionally identical to the one implemented in this PR):
```julia
macro mythreads(expr::Expr)
    @assert expr.head == :for
    @assert length(expr.args) == 2
    @assert length(expr.args[1].args) == 2
    loopvar   = expr.args[1].args[1]
    iter      = expr.args[1].args[2]
    loop_body = expr.args[2]

    rng = gensym(:rng)
    quote
        @sync for $rng in $(Iterators.partition)($iter, length($iter) ÷ Threads.nthreads())
            Threads.@spawn begin
                for $loopvar in $rng
                    $loop_body
                end
            end
        end
    end |> esc
end
```
So that
```julia
@mythreads for i in eachindex(xs)
    out[i] = sin(xs[i])
end
```
will get turned into
```julia
@sync for rng in partition(eachindex(xs), length(eachindex(xs)) ÷ Threads.nthreads())
    Threads.@spawn begin
        for i in rng
            out[i] = sin(xs[i])
        end
    end
end
```
This makes it so that a for loop only spawns a number of tasks equal to `Threads.nthreads()` (or `Threads.nthreads()+1` if the division has a remainder).

Here's a performance comparison on Julia 1.4.0-rc2 with `2` threads:
```julia
using BenchmarkTools

f1!(xs, out) = Threads.@threads for i in eachindex(xs)
    out[i] = sin(xs[i])
end

f2!(xs, out) = @mythreads for i in eachindex(xs)
    out[i] = sin(xs[i])
end

julia> let
           xs = randn(10_000)
           out1 = similar(xs)
           out2 = similar(xs)
           @btime f1!($xs, $out1)
           @btime f2!($xs, $out2)
           out1 ≈ out2
       end
  75.158 μs (13 allocations: 1.39 KiB)
  56.488 μs (15 allocations: 1.55 KiB)
true
```

_______

I'm not sure if it is intentional or not that the current implementation of `Threads.@threads` does not use PARTR and I also suspect there are things about my implementation of this macro that are not up the Base Julia snuff, hence the RFC. 

Thoughts, comments and code review much appreciated.


DilumAluthge:
Should we just deprecate `Threads.@threads` in favor of `Threads.@spawn`?

MasonProtter:
> Should we just deprecate Threads.@threads in favor of Threads.@spawn?

I'd strongly argue no. They do different things.

DilumAluthge:
Ah. Never mind then.

JeffBezanson:
#32477 also does this.

MasonProtter:
`Threads.@threads` is really handy when you have a very quick function being iterated over a very long iterator because it only spawns a number of tasks equal to `Threads.@threads`. If you naively did
```julia
@sync for i in 1:10_000_000
    out[i] = sin(x[i])
end 
```
you'd find that incredibly slow because spawning each thread task on the order of 1µs.

MasonProtter:
> #32477 also does this.

Ah I see, I wasn't able to decode that title when I saw it originally. If this is redundant feel free to close it.

MasonProtter:
Is there an argument to be made that this is desirable independent of #32477? I found the old version of the `_threadfor` function a little verbose and hard to understand. I think this nicely takes advantage of our already provided `Iterators.partition` and results in more readable / maintainable code.

base/threadingconstructs.jl
```diff
@@ -18,65 +18,27 @@ on [`threadid()`](@ref).
...
-                    f = f + rem
-                    l = l + rem
+function _threadsfor(iter_stmt, lbody)
+    loopvar   = iter_stmt.args[1]
+    iter      = iter_stmt.args[2]
+    rng = gensym(:rng)
+    out = quote
+        @sync for $rng in $(Iterators.partition)($iter, $(length)($iter) ÷ $(nthreads)())
+            Base.Threads.@spawn begin
+                for $loopvar in $rng
```
tkf: Strictly speaking, I think this is a breaking change as the following code will behave differently before and after this PR:

```julia
@threads for ...
    @spawn ...
end
```

Before this PR, `@spawn` is not synchronized at the end of `@threads` block. After this PR, it is not thread-safe: https://github.com/JuliaLang/julia/issues/34666. You can make it thread-safe by adding `@sync` before `for $loopvar in $rng` although the behavior of `@spawn` is still different (it's synchronized by `@threads`).

MasonProtter: Mhm, this could conceivably  change behaviour (though threading is experimental, so breaking changes are allowed), but before if you did
```julia
@threads for ...
    @spawn ...
end
```
then the `@spawn` would destructively interfere with the tasks spawned by `@threads`.


Are you able to cook up a valid test case that would have passed on the old version of `@threads` but would fail on this version?

MasonProtter: Ah I see an error in the test suite provides an example of exactly what you referenced.

tkf: > (though threading is experimental, so breaking changes are allowed)

Yeah, I think I agree. Though it could be surprising if `@threads` acts implicitly as `@sync`.

FWIW, I think `Threads.foreach` would be a much better API as it doesn't hide the fact that you'll have a closure. It's also easy to add keyword options, e.g., for controlling the base case size for better load balancing. Something like this:

```julia
function Threads.foreach(f, xs::AbstractArray; basesize = length(xs) ÷ Threads.nthreads())
    @sync for p in Iterators.partition(xs, basesize)
        @spawn foreach(f, p)
    end
    return
end
```

(I'm experimenting this API here: https://github.com/tkf/ThreadsX.jl/blob/1b2f2a42094986361785b617c064e5fe8cda93af/src/map.jl#L26)

`@threads` can exist as a simple wrapper around `Threads.foreach` for backward compatibility.  This trivially avoids the `@sync` scoping problem.

quinnj:
I think it's been mentioned elsewhere, but just wanted to add a voice that this would indeed be breaking; a few of us _rely_ on the current behavior in order to start worker tasks on specific threads, such as the code [here](https://github.com/tro3/ThreadPools.jl/blob/e36e82c51adfd22a7d0459912861bfaf1d43deff/src/qpool.jl#L16). I'd say as long as we have either: 1) a way to reserve a main thread where tasks couldn't be spawned, or 2) a way to still spawn a threaded task on a specific thread (e.g. `Threads.@spawn 2 expr`), then we could still get the behavior we're after.

baggepinnen:
Could the implementation in this pr simply live alongside the old one, under a different macro name? `@partr_threads` or similar

PR #35006, opened 2020-03-04
walkdir: avoid symlink loops when `follow_symlinks == false`

staticfloat:
Without this change, `walkdir()` errors out while trying to determine if a symlink is a directory or not.  When `follow_symlinks` is `false`, we should never attempt to dereference a symlink.

This has a slight behavioral change in that valid symlinks to directories are now returned in `files` instead of `dirs` when `follow_symlinks` is set to `false`, which seems like the more correct behavior here.

### DIFF
--- a/base/file.jl
+++ b/base/file.jl
@@ -853,10 +853,13 @@ function walkdir(root; topdown=true, follow_symlinks=false, onerror=throw)
     dirs = Vector{eltype(content)}()
     files = Vector{eltype(content)}()
     for name in content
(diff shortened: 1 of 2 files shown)


staticfloat:
Fantastic, it's working.  I'm rebase-squashing down to two commits, then we're good to merge.

KristofferC:
Could this have caused https://github.com/JuliaLang/Pkg.jl/issues/1716?

PR #35007, opened 2020-03-04
Added tests for +,- operations for Uniform Scaling Matrices

suraj-shivu:
This is for the functions with Uniform Scaling Matrices as the first operand.


suraj-shivu:
Sorry this was accidental

PR #35008, opened 2020-03-04
fix bug in allunique(::StepRangeLen)

### LINKED ISSUE #34997: allunique(('a':1:'c')[1:2]) errors
I didn't find a bug report for this elsewhere. I'm on 1.3.1 with Mac OS Catalina

```julia
julia> allunique(('a':1:'c')[1:2])
ERROR: MethodError: no method matching zero(::Type{Char})
Closest candidates are:
  zero(::Type{Missing}) at missing.jl:103
  zero(::Type{LibGit2.GitHash}) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.3/LibGit2/src/oid.jl:220
  zero(::Type{Pkg.Resolve.VersionWeights.VersionWeight}) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.3/Pkg/src/resolve/VersionWeights.jl:19
  ...
```

ShriprajwalK:
Fixed #34997

### DIFF
--- a/test/sets.jl
+++ b/test/sets.jl
@@ -452,6 +452,7 @@ end
     @test allunique(7:-1:1)       # negative step
     @test allunique(Date(2018, 8, 7):Day(1):Date(2018, 8, 11))  # JuliaCon 2018
     @test allunique(DateTime(2018, 8, 7):Hour(1):DateTime(2018, 8, 11))
+    @test allunique(('a':1:'c')[1:2]) == true
 end
 @testset "filter(f, ::$S)" for S = (Set, BitSet)
     s = S([1,2,3,4])
--- a/base/set.jl
+++ b/base/set.jl
@@ -377,6 +377,7 @@ allunique(::Union{AbstractSet,AbstractDict}) = true
 
 allunique(r::AbstractRange{T}) where {T} = (step(r) != zero(T)) || (length(r) <= 1)
 allunique(r::StepRange{T,S}) where {T,S} = (step(r) != zero(S)) || (length(r) <= 1)
+allunique(r::StepRangeLen{T,R,S}) where {T,R,S} = (step(r) != zero(S)) || (length(r) <= 1)
 
 filter!(f, s::Set) = unsafe_filter!(f, s)
 


test/sets.jl
```diff
@@ -452,6 +452,7 @@ end
     @test allunique(7:-1:1)       # negative step
     @test allunique(Date(2018, 8, 7):Day(1):Date(2018, 8, 11))  # JuliaCon 2018
     @test allunique(DateTime(2018, 8, 7):Hour(1):DateTime(2018, 8, 11))
+    @test allunique(('a':1:'c')[1:2]) == true
```
rfourquet: I would use `'a':'\x01':'b'` instead of (or in addition to) `('a':1:'c')[1:2]`, to be sure what the test is actually testing, and doesn't depend on possible future changes of the result type of `('a':1:'c')[1:2]`.

StefanKarpinski: Wouldn't that still error even after this change? Since the step type is `Char` which correctly does not have a zero element and also is not a valid step type. Perhaps I'm misunderstanding the suggestion.

rfourquet: Oh I got it wrong. I don't fully understand what happens here, but I just did:
```
julia> ('a':1:'c')[1:2]
'a':'\x01':'b'
```
And assumed `allunique('a':'\x01':'b')` was the problem, but `'a':'\x01':'b'` is actually not valid input:
```julia
julia> 'a':'\x01':'b'
ERROR: MethodError: no method matching zero(::Char)
```
Forget my comment above and carry on, though there might be a problem with the printing of `('a':1:'c')[1:2]`.

ShriprajwalK:
What changes do I do to make this PR better?

rfourquet:
Thanks!

test/sets.jl
rfourquet: Ok, I found the source of the confusion: until very recently, `step(::StepRangeLen)` was returning a `Char`, hence the output above `'a':'\x01':'b'`, now `step` properly doesn't convert the step to `Char` (cf. #34412), and keeps it as an `Int`, so we now have:
```julia
julia>  ('a':1:'c')[1:2]
'a':1:'b'
```
So at least now `('a':1:'c')[1:2] == 'a':1:'b'` even though the types don't match.

PR #35010, opened 2020-03-04
Added tests for +,- for UniformScaling matrices

suraj-shivu:
This is for the functions with UniformScaling Type as their first operand

### DIFF
--- a/stdlib/LinearAlgebra/test/special.jl
+++ b/stdlib/LinearAlgebra/test/special.jl
@@ -142,20 +142,18 @@ end
         push!(mats, SymTridiagonal(Vector{T}(diag), Vector{T}(offdiag)))
     end
 
-    for op in (+,*,-) # to do: fix when operation is - and the matrix has a range as the underlying representation and we get a step size of 0.
-        if(op) in (+,*)
-            for A in mats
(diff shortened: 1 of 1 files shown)


dkarrasch (changes requested):
I think we should avoid the code duplication. If I'm not mistaken, we can achieve the same code coverage by adding a line after line 154 with `A` and `B` interchanged. Moreover, I assume that the `*` in line 145 is a typo, according to the testset header. So, either we replace it by `-`, or we add `-` there, which I'd advocate for. Then we could remove the entire new testset.

suraj-shivu:
I think the * isn't a typo and the - there would cause step size 0 errors as written in the to-do, when i tried the suggested edits. So I added a couple of selection statements to go about this.

dkarrasch:
Ok, then I suggest not to touch the `A in mats, B in mats` part, because the purpose of the PR is to cover `UniformScaling` methods. Like have the old outer `for op in (+,*)` loop for the `(mats, mats)` case, and then a new `for op in (+,-)` loop for the `A in mats, B in uniformscalingmats` case.

dkarrasch (approved):
Very clean, very nice.

PR #35016, opened 2020-03-05
 Improve `fld` documentation following PR #17754 

zierenberg:
Hi,
this is my first attempt to do a pull request - so if I did something wrong I would be happy to know how to do it better.

Following the discussion in [PR #17754](https://github.com/JuliaLang/julia/pull/17754), I think that it is of value to add an example to the documentation of `fld(x,y)` that warns the user that `x/y` might not be the expected result due to floating point rounding. 

I thus added an explicit example plus description in the `#Examples` section. (see below)

As to the workflow: I forked the repository, made changes in a branch and merged with my master. However, travis checks failed because I exceeded maximum run times. I nevertheless wanted to try this PR now to see if you agree in principle, and if so, how I can properly run checks that documentations compile.

### DIFF
--- a/base/div.jl
+++ b/base/div.jl
@@ -89,7 +89,7 @@ julia> 6.0/0.1
 julia> 6.0/big(0.1)
 59.99999999999999666933092612453056361837965690217069245739573412231113406246995
 ```
-
+What is happening here is that the true value of the floating-point number written as `0.1` is slightly larger than the numerical value 1/10 while `6.0` represents the number 6 precisely. Therefore the true value of `6.0 / 0.1` is slightly less than 60. When doing division, this is rounded to precisely `60.0`, but `fld(6.0, 0.1)` always takes the floor or the true value, so the result is `59.0`.
 """
 fld(a, b) = div(a, b, RoundDown)
 
--- a/base/div.jl
+++ b/base/div.jl
@@ -80,6 +80,16 @@ See also: [`div`](@ref)
 julia> fld(7.3,5.5)
 1.0
 ```
+Because of floating-point rounding `fld(x,y)` can lead to an *alleged* violation of the fld(x,y) == x/y condition:
+```jldoctest
+julia> fld(6.0/0.1)
+59.0
+julia> 6.0/0.1
+60.0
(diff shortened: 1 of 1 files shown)


base/div.jl
```diff
@@ -80,6 +80,16 @@ See also: [`div`](@ref)
 julia> fld(7.3,5.5)
 1.0
 ```
+Because of floating-point rounding `fld(x,y)` can lead to an *alleged* violation of the fld(x,y) == x/y condition:
```
kshyatt: Might be nice to have backticks around the `fld(x,y) == x/y`

zierenberg: Good idea, fixed that together with a mistake that made doctests fail.

zierenberg:
I notice that 1/16 checks failed, namely the buildbot/tester_macos64 in "Sockets" test. I do not believe that this is due to the changes of this PR?

ViralBShah:
The failure does not matter. @StefanKarpinski Merge this?

base/div.jl
```diff
@@ -80,6 +80,16 @@ See also: [`div`](@ref)
 julia> fld(7.3,5.5)
 1.0
 ```
+Because of floating-point rounding `fld(x,y)` can lead to an *alleged* violation of the `fld(x,y) == x/y` condition:
```
StefanKarpinski, line 83:
```suggestion
Because `fld(x, y)` implements strictly correct floored rounding based on the true of floating-point numbers, unintuitive situations can arise. For example:
```

oscardssmith:
Should we merge this?

StefanKarpinski:
We should. Rerunning CI.

oscardssmith:
I think this needs a rebase for CI to pass.

ViralBShah:
Accidentally got these onto master in commit 1044621324e04d9352fe2e2cc953b4f65e823bb2. I'll revert if tests still fail.

ViralBShah:
Passes tests. One of the platforms has a test failure on a download test. So, closing this and leaving the commits on master. Sorry again for the extra commits dumped there.

PR #31553, opened 2019-03-30
Implement `for f.(args...)` syntax

tkf:
This PR implements `for f.(args...)` syntax as discussed in #19198 and planned in https://github.com/JuliaLang/julia/pull/31088#issuecomment-473030504.

close #19198
close #31088

### Implementation detail

* A new `Expr(:fordot, :(broadcasted.(expression)))` AST node is introduced to represent `for broadcasted.(expression)`.
* This is lowered to a chain of `broadcasted` calls by using the pre-existing code in `julia-syntax.scm` but skips the outer-most `materialize` call.
* `for non(dot(call))` is a syntax error at the moment.
* Lowering manually constructed `for non(dot(call))` is an error at the moment.
* There is no performance specialization in this PR.  It is done separately in #31020.

## Demo

```julia
julia> ex = :(for f.(x))
:(for f.(x))

julia> ex.head
:fordot

julia> ex.args
1-element Array{Any,1}:
 :(f.(x))

julia> Meta.lower(Main, ex)
:($(Expr(:thunk, CodeInfo(
    @ none within `top-level scope'
1 ─ %1 = Base.broadcasted(f, x)
└──      return %1
))))

julia> :(for f(x))
ERROR: syntax: invalid expression after `for`

julia> nondot = Expr(:fordot, :(f(x)))
:(for f(x))

julia> Meta.lower(Main, nondot)
:($(Expr(:error, "non-dot call after `for`")))
```

Issue #20181, opened 2017-01-22
make `any` and `all` interface consistent with `map`


bjarthur:
i'm not sure whether it's better to ask this on Discourse, but has it been discussed to have `Function{N}` where N is the number of input arguments?

my use case is that i'd like to make the interface to `any` and `all` consistent with that of `map`.  currently, the predicate versions of the former can not take a variable number of iterables like the latter can.  rather, you have to manually `zip` them up.  no big deal, but consistency is nice.

the problem is that `any` and `all` have methods which specialize on Arrays and regions.  and so a simple change of the signature to end with a var arg gets confused when there are exactly two.  a conceptually simple fix to this would be to dispatch on how many input args the predicate took.

tkelman:
or make dims a keyword argument for the array/region signatures...

ararslan:
Functions have their own types, but any function can be defined to have arbitrarily many methods with any number of arguments. So saying that, for example, `typeof(+) <: Function{2}`, isn't really meaningful; `+` is intuitively binary but you can define `+(a, b, c)`, at which point `typeof(+)` would no longer be a subtype of `Function{2}`. Perhaps I'm misunderstanding your request, but it seems to me that what you're describing would require a massive change to the type system.

pkofod:
> Functions have their own types, but any function can be defined to have arbitrarily many methods with any number of arguments.
^
I the suggestion in this issue might come from a misunderstanding of the differences between a function and a method.

JeffBezanson:
This is basically #17168 and #16191.
If changing the behavior of `any` and `all` specifically is considered useful, we can make this issue about that.

bjarthur:
great to see parameterizing `Function` has already been discussed.

using a dims keyword would make the interface inconsistent with `mean` and friends.

title of this issue changed to reflect the desire to make `any` and `all` consistent with `map`.

tkelman:
> using a dims keyword would make the interface inconsistent with `mean` and friends.

Not if we also switched to using a dims keyword for `mean` and friends at the same time.

nalimilan:
Wouldn't a keyword argument incur too large a penatly for `all` and `any`?

bjarthur:
changing `mean` and friends to use a keyword for the region would be hugely disruptive.  i'd suggest doing that only if it is decided to not parameterize `Function`.

[...]

PR #35017, opened 2020-03-05
Add function + two-argument method to reducers

baggepinnen:
## Motivation
The following style of code is very common in practice
```julia
sum(a.-b)
```
While it is easy to apply a function to all arguments without creating a temporary array and causing allocations, by means of the method `sum(f,x)`, it is slightly more cumbersome to do so when the function takes two arguments. This PR adds such a method to all reduction functions (such as `sum,maximum` etc.) accepting two arrays, implemented simply by means of zipping the two arrays and calling the `reducer(f,x)` method.
```julia
sum(-, a, b)
```
Common combinations like `sum((a.-b).^2)` can easily be realized as `sum(abs2 ∘ -, a, b)`.

`mapreduce(-, +, a, b)` does infact solve exactly this problem, but it is horribly slow compared to both alternatives above.

## Some benchmarks
In the table below, "Eager" denotes `sum(a .- b)` for various lengths of vectors. "Functional" denotes `sum(-, a, b)`. The middle colums are the timings and the "Reduction" column are the relative timings `functional/eager` meaning that values below 1 are faster for the method in this PR.
Naturally, the eager method allocates linear memory whereas the functional allocates constant `15.596 ms (2 allocations: 48 bytes)`
As for the timings, the functional approach is about 2x faster for very small arrays and very large arrays, whereas results are more even for medium sized arrays.

|   Length |    Eager [ns] | Functional [ns] | Mapreduce [ns] | Reduction |
|----------|----------|------------|-----------|-----------|
|      1 |    34.33 |      15.28 |    1023.0 |     0.445 |
|      3 |    35.83 |      18.98 |    1040.0 |    0.5296 |
|     10 |    43.78 |      28.28 |    1061.0 |    0.6458 |
|     30 |    87.04 |      53.33 |    1226.0 |    0.6128 |
|    100 |    126.9 |      142.5 |    1370.0 |     1.122 |
|    300 |    544.9 |      376.1 |    1990.0 |    0.6901 |
|   1000 |    994.4 |     1199.0 |    3232.0 |     1.206 |
|   3000 |   3260.0 |     3548.0 |    6793.0 |     1.088 |
|  10000 |  11520.0 |    11770.0 |   18980.0 |     1.022 |
|  30000 |  38410.0 |    35230.0 |   55850.0 |    0.9172 |
| 100000 | 142400.0 |   117200.0 |  193400.0 |    0.8232 |
| 300000 | 590500.0 |   380500.0 |  695400.0 |    0.6443 |
|    1.0e6 |  2.245e6 |    1.269e6 |   2.282e6 |    0.5653 |

timings

### Benchmark code
```julia
lengths = [1,3,10,30,100,300,1000,3000,10_000,30_000,100_000,300_000,1_000_000]

res1 = map(lengths) do len
    a,b, = randn(len),randn(len)
    @benchmark sum($a .- $b)
end

res2 = map(lengths) do len
    a,b, = randn(len),randn(len)
    @benchmark sum(-, $a, $b)
end

res3 = map(lengths) do len
    a,b, = randn(len),randn(len)
    @benchmark mapreduce(-, +, $a, $b)
end

f1 = plot(lengths, memory.([res1 res2 res3]), lab=["Eager" "Functional" "Mapreduce"], title="Memory", yscale=:log10, xscale=:log10)
f2 = plot(lengths, time.([res1 res2 res3]), lab=["Eager" "Functional" "Mapreduce"], title="Time", yscale=:log10, xscale=:log10)
plot(f1,f2, legend=:topleft) |> display

using PrettyTables
times = time.([res1 res2 res3])
quotient = times[:,2]./times[:,1]
T = pretty_table(round.([lengths times quotient], sigdigits=4), ["Length", "Eager", "Functional", "Mapreduce", "Reduction"], tf=markdown)
```

### DIFF
--- a/base/reducedim.jl
+++ b/base/reducedim.jl
@@ -651,6 +651,8 @@ for (fname, _fname, op) in [(:sum,     :_sum,     :add_sum), (:prod,    :_prod, 
         # User-facing methods with keyword arguments
         @inline ($fname)(a::AbstractArray; dims=:) = ($_fname)(a, dims)
         @inline ($fname)(f, a::AbstractArray; dims=:) = ($_fname)(f, a, dims)
+        @inline ($fname)(f, a::AbstractArray, b::AbstractArray; dims=:) =
+            ($_fname)(((a,b),)->f(a,b), zip(a,b), dims)
 
         # Underlying implementations using dispatch
         ($_fname)(a, ::Colon) = ($_fname)(identity, a, :)


base/reducedim.jl
```diff
@@ -651,6 +651,8 @@ for (fname, _fname, op) in [(:sum,     :_sum,     :add_sum), (:prod,    :_prod,
         # User-facing methods with keyword arguments
         @inline ($fname)(a::AbstractArray; dims=:) = ($_fname)(a, dims)
         @inline ($fname)(f, a::AbstractArray; dims=:) = ($_fname)(f, a, dims)
+        @inline ($fname)(f, a::AbstractArray, b::AbstractArray; dims=:) =
+            ($_fname)(((a,b),)->f(a,b), zip(a,b), dims)
```
johnnychen94: If so, why don't make it more general? `sum(f, args...)` as `_sum(args->f(args...), zip(args...))`

baggepinnen: Good point!

dkarrasch: > If so, why don't make it more general? `sum(f, args...)` as `_sum(args->f(args...), zip(args...))`

This could even replace the single argument method, right?

baggepinnen: It could, but I will benchmark that first to ensure there's not a performance hit for the zip.

baggepinnen:
Here's an imperfect regexp to see how common this is in practice `rg 'sum\([^)]*?-[^(]*?\)'`

baggepinnen:
Replacing the 1-arg method with a zip-only method caused quite a bit of regression on `sum(abs2, x)`
timings_1

mbauman:
Just as an alternative, you can also do `sum(Base.splat(-), zip(a, b))`. This appears to be just as fast as the definition you've written in this PR:

```Julia
julia> a,b = rand(10), rand(10);

julia> @btime sum(-, $a, $b)
  19.606 ns (2 allocations: 48 bytes)
-1.6367587009468854

julia> @btime sum(Base.splat(-), zip($a, $b))
  19.506 ns (2 allocations: 48 bytes)
-1.6367587009468854
```

tkf:
FYI there is #31020 which is more composable IMHO.

baggepinnen:
Thanks for your comments. 

- `sum(Base.splat(-), zip(a, b))` is not really any more conventient to write than `sum(((a,b),)->a-b, zip(a,b))`
- If `sum(Broadcast.broadcasted(-, a, b))` was fast, that would be awesome. It would still require syntax for making the broadcast lazy. https://github.com/JuliaLang/julia/pull/31088 seems to have stalled, so does https://github.com/JuliaLang/julia/pull/31553?

mbauman:
We've talked about reducers taking multiple arguments (and what it means) in the context of `any` and `all` in the past: https://github.com/JuliaLang/julia/issues/20181

tkf:
> * It would still require syntax for making the broadcast lazy. #31088 seems to have stalled, so does #31553?

The discussion on the syntax is in #19198. I think it's reasonable to take time to decide the syntax. Meanwhile, we can write `sum(@~ a .- b)` with `@~` from LazyArrays.jl.

baggepinnen:
Closing this in favor of lazy broadcasting

PR #35018, opened 2020-03-05
Actually test LDLt file and add tests to it

kshyatt:
Without `ldlt` in `testgroups`, the tests weren't run.

### DIFF
--- a/stdlib/LinearAlgebra/test/ldlt.jl
+++ b/stdlib/LinearAlgebra/test/ldlt.jl
@@ -6,6 +6,14 @@ using Test, LinearAlgebra, Random
 
 Random.seed!(123)
 
+@testset "Factorization conversions of LDLT" begin
+    S = SymTridiagonal(randn(5), randn(4))
+    F = ldlt(S)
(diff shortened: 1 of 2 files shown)


dkarrasch:
@kshyatt There is already some effort going on in #34965 along the exact same lines. The show-stopper there is the broken `getproperty` for `LDLt`, of which a consequence is going to be that the REPL printing tests are going to fail. I was busy reviewing loads of coverage PRs recently, so didn't have the time to dig into the `getproperty` thing. If you could take a look at that, that would be great.

dkarrasch:
I was apparently running `master` code on v1.3... and failed, obviously. Sorry for the noise.

dkarrasch (approved):
LGTM. Wasn't `addmul` put at the end of the testgroups because of long runtimes or so? Maybe we should insert `ldlt` somewhere in the middle?

kshyatt:
Oh whoops didn't see that pr. I'll see what I can do

dkarrasch:
I'm merging this, with, once again, my sincere apologies to @ShriprajwalK for messing up #34965 by my review comments.

PR #35019, opened 2020-03-05
htables: fix ordering of conditions

vtjnash:
The second condition was unreachable, which meant we might grow too slowly for the first few steps.


eulerkochy (approved):
Looks goods. The ordering of conditionals improves code readability, and infact makes more sense than the previous code.

PR #35023, opened 2020-03-05
Add broadcasting examples to `in` and `∉` docstrings.

CameronBieganek:
This PR adds examples to the `in` and `∉` docstrings that show how to use those infix operators in broadcasting expressions where the collection on the right side is intended to be treated as scalar. I've included one example with the right side enclosed in a tuple and one without the tuple to demonstrate the difference between the two, since the latter is also valid syntax that could easily be confused with the right-side-as-a-scalar behavior.

### DIFF
--- a/base/operators.jl
+++ b/base/operators.jl
@@ -1083,11 +1083,10 @@ You may also negate the `in` by doing `!(a in b)` which is logically similar to 
 
 When broadcasting with `in.(items, collection)` or `items .∈ collection`, both
 `item` and `collection` are broadcasted over, which is often not what is intended.
-For example, if both arguments are vectors, the result is a vector indicating whether
-each value in collection `items` is `in` the value at the corresponding position
-in `collection`.
-To get a vector indicating whether each value in `items` is in `collection`,
-wrap `collection` in a tuple or a `Ref` like this:
+For example, if both arguments are vectors (and the dimensions match), the result is
+a vector indicating whether each value in collection `items` is `in` the value at the
+corresponding position in `collection`. To get a vector indicating whether each value
+in `items` is in `collection`, wrap `collection` in a tuple or a `Ref` like this:
 `in.(items, Ref(collection))` or `items .∈ Ref(collection)`.
 
 # Examples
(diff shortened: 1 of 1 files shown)


base/operators.jl
```diff
@@ -1109,6 +1109,16 @@ true
...
+
+julia> [1, 2] .∈ [2, 3]
+2-element BitArray{1}:
+ 0
+ 0
+
+julia> [1, 2] .∈ ([2, 3],)
+2-element BitArray{1}:
+ 0
+ 1
```
stevengj: I think this example needs some explanation to be comprehensible.

CameronBieganek: Yeah, you're probably right. I suppose it's ok to have more than one separate block of Examples in the docstring?

CameronBieganek: Ok, I added an explanatory sentence. Not sure if it's clear enough.

base/operators.jl
```diff
@@ -1081,6 +1081,9 @@ is always a `Bool` and never `missing`.
 To determine whether an item is not in a given collection, see [`:∉`](@ref).
 You may also negate the `in` by doing `!(a in b)` which is logically similar to "not in".
 
+When broadcasting with `in` or `∈`, the collection will be broadcasted over unless the
+collection is wrapped in a tuple or a `Ref`.
```
nalimilan: Maybe something like this? Or is it too verbose?
```suggestion
When broadcasting with `in.(items, collection)` or `items .∈ collection`, both
`item` and `collection` are broadcasted over, which is often not what is intended.
For example, if both arguments are vectors, the result is a vector indicating whether
each value in collection `items` is `in` the value at the corresponding position
in `collection`.
To get a vector indicating whether each value in `items` is in `collection`,
wrap `collection` in a tuple or a `Ref` like this:
`in.(items, Ref(collection))` or `items .∈ Ref(collection)`.
```
(Same below.)

CameronBieganek: More verbose and explanatory is perfectly fine with me, but I've never contributed Base documentation before, so I'm not sure what the standard is. To make it even more clear should we insert a qualification about matching dimensions, like the following?

> if both arguments are vectors (and the dimensions match)

CameronBieganek:
Ok, I've added a more detailed explanation to the docs, as requested by @stevengj and @nalimilan . I think this PR might be ready to be merged now.

stevengj (approved):

nalimilan (approved):

PR #35024, opened 2020-03-06
fix typo in dict.jl

joaquimg:


matthias314:
Could it be that there is another typo in the revised comment
```
# @inbounds is here so the it gets propagated to isslotfiled
```
namely that it should read "... so that it gets ..."?

ViralBShah:
sigh... yes...

matthias314:
I can fix this together with another tiny change I want to make to `dict.jl` (removing an unused local variable).

Issue #35022, opened 2020-03-05
range function not eliding with constant arguments


MasonProtter:
```julia
foo(N) = range(-1, 1, length=N)
bar()  = foo(1000)
julia> @btime bar();
  118.334 ns (0 allocations: 0 bytes)
```
I've tried this on the master branch and `1.4.2-rc2`. I'd expect this function call to be completely elided, but it's not which can induce a pretty hefty performance cost sometimes in hot loops. 

```julia
julia> @code_typed foo(1000)
CodeInfo(
1 ── %1  = Base.sle_int(1, 1)::Bool
└───       goto #3 if not %1
2 ── %3  = Base.sle_int(1, 0)::Bool
└───       goto #4
3 ──       nothing::Nothing
4 ┄─ %6  = φ (#2 => %3, #3 => false)::Bool
└───       goto #6 if not %6
5 ──       invoke Base.getindex(()::Tuple, 1::Int64)::Union{}
└───       $(Expr(:unreachable))::Union{}
6 ┄─       goto #7
7 ──       goto #8
8 ──       goto #9
9 ──       goto #10
10 ─ %14 = invoke Base._linspace(Float64::Type{Float64}, -1::Int64, 1::Int64, _2::Int64, 1::Int64)::StepRangeLen{Float64,Base.TwicePrecision{Float64},Base.TwicePrecision{Float64}}
└───       goto #12
11 ─       $(Expr(:meta, :nkw, 2))
12 ┄       goto #13
13 ─       return %14
) => StepRangeLen{Float64,Base.TwicePrecision{Float64},Base.TwicePrecision{Float64}}
```

I wonder if there's anything we can do to make this infer more easily. 

@haampie on Slack tracked the problem down to `function Base._linspace(::Type{T}, start_n::Integer, stop_n::Integer, len::Integer, den::Integer) where T<:IEEEFloat` which he speculated is overwhelming the optimizer with its very complex logic.

haampie:
I don't think this will be optimized easily, many things are not inlined. Your `foo` does not inline `Base._linspace(Float64, -1, 1, 1000, 1)`, which in turn does not inline

```
invoke Base._linspace1(%3::Type{Float64}, %6::Float64, %9::Float64, _5::Int64)::StepRangeLen{Float64,Base.TwicePrecision{Float64},Base.TwicePrecision{Float64}}
[...]

PR #35029, opened 2020-03-06
Mention LinRange in the range docstring

MasonProtter:
I was recently quite surprised by https://github.com/JuliaLang/julia/issues/35022, but it was pointed out to me that perhaps I should just use `LinRange` which did indeed solve my problem.

However, I had no expectation from the docs on `range` that it can't be run at compile time if the arguments are literals, and I also didn't know that `LinRange` seems to exist to solve this problem. Hence, I think it's a good idea to mention this somehow in the docstring. 

```julia
julia> foo1() = range(-1, 1, length=10000)
foo1 (generic function with 1 method)

julia> foo2() = LinRange(-1, 1, 10000)
foo2 (generic function with 1 method)

julia> @btime foo1();
  118.294 ns (0 allocations: 0 bytes)

julia> @btime foo2();
  0.029 ns (0 allocations: 0 bytes)
```

### DIFF
--- a/base/range.jl
+++ b/base/range.jl
@@ -370,8 +370,8 @@ julia> LinRange(1.5, 5.5, 9)
  1.5,2.0,2.5,3.0,3.5,4.0,4.5,5.0,5.5
 ```
 
-Compared to [`range`](@ref), `LinRange` should have less overhead but deal with floating point errors
-less cleanly:
+Compared to [`range`](@ref), `LinRange` should have less overhead but won't try to correct
+for floating point errors:
 ```julia
 julia> collect(-0.1:0.1:0.3)
 5-element Array{Float64,1}:
@@ -383,11 +383,11 @@ julia> collect(-0.1:0.1:0.3)
 
 julia> collect(LinRange(-0.1, 0.3, 5))
 5-element Array{Float64,1}:
- -0.1                   
+ -0.1
  -1.3877787807814457e-17
-  0.09999999999999999   
-  0.19999999999999998   
-  0.3    
+  0.09999999999999999
+  0.19999999999999998
+  0.3
 ```
 """
 struct LinRange{T} <: AbstractRange{T}
--- a/base/range.jl
+++ b/base/range.jl
@@ -369,6 +369,26 @@ julia> LinRange(1.5, 5.5, 9)
 9-element LinRange{Float64}:
  1.5,2.0,2.5,3.0,3.5,4.0,4.5,5.0,5.5
 ```
+
+Compared to [`range`](@ref), `LinRange` should have less overhead but deal with floating point errors
+less cleanly:
+```julia
+julia> collect(-0.1:0.1:0.3)
+5-element Array{Float64,1}:
+ -0.1
+  0.0
+  0.1
+  0.2
+  0.3
+
+julia> collect(LinRange(-0.1, 0.3, 5))
+5-element Array{Float64,1}:
+ -0.1                   
+ -1.3877787807814457e-17
+  0.09999999999999999   
+  0.19999999999999998   
+  0.3    
+```
 """
 struct LinRange{T} <: AbstractRange{T}
     start::T
--- a/base/range.jl
+++ b/base/range.jl
@@ -374,7 +374,7 @@ julia> LinRange(1.5, 5.5, 9)
 Compared to [`range`](@ref), `LinRange` should have less overhead but won't try to correct
 for floating point errors:
 ```julia
-julia> collect(-0.1:0.1:0.3)
+julia> collect(range(-0.1, 0.3, length=5))
 5-element Array{Float64,1}:
  -0.1
   0.0
--- a/base/range.jl
+++ b/base/range.jl
@@ -59,6 +59,8 @@ automatically such that there are `length` linearly spaced elements in the range
 If `step` and `stop` are provided and `length` is not, the overall range length will be computed
 automatically such that the elements are `step` spaced.
 
+For a simpler linearly spaced range-like object with less overhead, see [`LinRange`](@ref).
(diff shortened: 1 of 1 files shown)


base/range.jl
```diff
@@ -59,6 +59,8 @@ automatically such that there are `length` linearly spaced elements in the range
 If `step` and `stop` are provided and `length` is not, the overall range length will be computed
 automatically such that the elements are `step` spaced.
 
+For a simpler linearly spaced range-like object with less overhead, see [`LinRange`](@ref).
```
mbauman:
```suggestion
Special care is taken to ensure intermediate values are computed rationally.  For a simpler linearly spaced range-like object with less overhead, see [`LinRange`](@ref).
```

Ok, maybe the "rationally" pun is a little too cute, but I think we should mention why `range` is slower in the first place.

MasonProtter: I'll wait and see if anyone has thoughts on the use of the word 'rationally' before I commit the suggestion, but I agree it's good to point out why.

timholy: I edited my old comment with a demo, you could use that: https://github.com/JuliaLang/julia/issues/35022#issuecomment-595464775.

Some people will look at that and feel very strongly that `range` is the "rational" one. Clearly you look at code, and by that definition `LinSpace` is the "rational" one.

MasonProtter: > I edited my old comment with a demo, you could use that: #35022 (comment).

I think that might be a bit too verbose to put in the docstring for `range`. Perhaps it can go in the `LinRange` docstring? Anyone have feelings on that? 

> Clearly you look at code, and by that definition LinSpace is the "rational" one. Though yes I like the pun.

Are you referring to me? I also think `range` is the 'rational' one and it's quite nice that it deals with floating point errors like that. I just didn't know there was a lower overhead alternative I should be using if I expect things to constant fold, and I suspect many others don't know about this overhead either. For someone who has never looked at the implementation of `range`, I think it's somewhat natural to think "how complicated could it be? Of course this thing should constant fold", so I thought it'd be good to put a hint in the docs.

timholy (approved):

base/range.jl
```diff
@@ -59,6 +59,8 @@ automatically such that there are `length` linearly spaced elements in the range
 If `step` and `stop` are provided and `length` is not, the overall range length will be computed
 automatically such that the elements are `step` spaced.
 
+Special care is taken to ensure intermediate values are computed rationally.  For a simpler linearly spaced range-like object with less overhead, see [`LinRange`](@ref).
```
```suggestion
Special care is taken to ensure intermediate values are computed rationally.
For a simpler linearly spaced range-like object with less overhead, see [`LinRange`](@ref).
```

StefanKarpinski (approved):

base/range.jl
```diff
@@ -59,6 +59,9 @@ automatically such that there are `length` linearly spaced elements in the range
 If `step` and `stop` are provided and `length` is not, the overall range length will be computed
 automatically such that the elements are `step` spaced.
 
+Special care is taken to ensure intermediate values are computed rationally.
+For a simpler linearly spaced range-like object with less overhead, see [`LinRange`](@ref).
```
j-n-e:
```suggestion
For a construction with less overhead, see [`LinRange`](@ref).
```
(The previous sentence implied `LinRange` is not a range but "range-like"; also, "linearly spaced" is redundant to ranges...)

base/range.jl
```diff
@@ -367,6 +370,26 @@ julia> LinRange(1.5, 5.5, 9)
 9-element LinRange{Float64}:
  1.5,2.0,2.5,3.0,3.5,4.0,4.5,5.0,5.5
 ```
+
+Compared to [`range`](@ref), `LinRange` should have less overhead but won't try to correct
+for floating point errors:
```
j-n-e:
```suggestion
Compared to [`range`](@ref), direct `LinRange` construction should have less overhead
but won't try to correct for floating point errors:
```
(The previous sentence implied `LinRange` is not one of the products of `range` calls.)

base/range.jl
```diff
@@ -367,6 +370,26 @@ julia> LinRange(1.5, 5.5, 9)
...
+```julia
+julia> collect(-0.1:0.1:0.3)
```
j-n-e:
```suggestion
julia> collect(range(-0.1, 0.3, length=5))
```
Less comparing apples to oranges.

base/range.jl
MasonProtter: I made a change with slightly different wording than you suggest. Can you let me know what you think?

base/range.jl
MasonProtter: I made a change with slightly different wording than you suggested.

j-n-e: Great

base/range.jl
j-n-e: Great, thx for checking out my thoughts

MasonProtter: 😄

j-n-e:
One more minor thought:
I wonder if the new comparison in the `LinRange` docstring might have its potential for misleading (if any) reduced by also providing `typeof(range(-0.1, 0.3, length=5))`... because it's not a `LinRange`. As I would have expected long ago before knowing more. (Also that would expose `TwicePrecision` happening in the background in the type.) Might require swapping the `LinRange` prompt ahead of the `range` prompt for sensible sequence of prompts.

timholy:
> also providing `typeof(range(-0.1, 0.3, length=5))`

I think this is an exercise better left to the reader; too much detail opens a can of worms that most readers won't/shouldn't care about. The curious will poke at it anyway and discover this soon enough.

Thanks, @MasonProtter!

j-n-e:
Yep, makes sense

PR #35034, opened 2020-03-07
rem for integers for RoundNearest

### LINKED ISSUE #34971: rem(x, y, RoundNearest) for integer x or y is not defined
`rem(x, y, RoundNearest)` for integer (`(U)Int{8|16|32|64|128}`, `Bool`) `x` or `y` is not defined.

ShriprajwalK:
Fixes #34971.

### DIFF
--- a/base/div.jl
+++ b/base/div.jl
@@ -67,7 +67,7 @@ rem(x, y, r::RoundingMode)
 rem(x, y, ::RoundingMode{:ToZero}) = rem(x,y)
 rem(x, y, ::RoundingMode{:Down}) = mod(x,y)
 rem(x, y, ::RoundingMode{:Up}) = mod(x,-y)
-rem(x, y, r::RoundingMode{:Nearest}) = x - y*div(x, y, r)
+rem(x, y, r::RoundingMode{:Nearest}) = divrem(x, y, r)[2]
 
 """
     fld(x, y)
--- a/base/div.jl
+++ b/base/div.jl
@@ -68,6 +68,7 @@ rem(x, y, ::RoundingMode{:ToZero}) = rem(x,y)
 rem(x, y, ::RoundingMode{:Down}) = mod(x,y)
 rem(x, y, ::RoundingMode{:Up}) = mod(x,-y)
 rem(x, y, r::RoundingMode{:Nearest}) = x - y*round(x/y, r)
+
 """
     fld(x, y)
 
--- a/base/div.jl
+++ b/base/div.jl
@@ -67,7 +67,7 @@ rem(x, y, r::RoundingMode)
 rem(x, y, ::RoundingMode{:ToZero}) = rem(x,y)
 rem(x, y, ::RoundingMode{:Down}) = mod(x,y)
 rem(x, y, ::RoundingMode{:Up}) = mod(x,-y)
-rem(x, y, r::RoundingMode{:Nearest}) = x - y*round(x/y, r)
+rem(x, y, r::RoundingMode{:Nearest}) = x - y*div(x, y, r)
 
 """
     fld(x, y)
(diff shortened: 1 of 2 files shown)


base/div.jl
```diff
@@ -67,6 +67,7 @@ rem(x, y, r::RoundingMode)
 rem(x, y, ::RoundingMode{:ToZero}) = rem(x,y)
 rem(x, y, ::RoundingMode{:Down}) = mod(x,y)
 rem(x, y, ::RoundingMode{:Up}) = mod(x,-y)
+rem(x, y, r::RoundingMode{:Nearest}) = x - y*round(x/y, r)
```
dkarrasch: This should probably be
```suggestion
rem(x, y, r::RoundingMode{:Nearest}) = x - y*div(x, y, r)
```
to avoid the intermediate rounding as claimed in the docstring?

ShriprajwalK: div(x, y, r) is giving me errors. Does div(x, y, r) exist?

dkarrasch: Yes, on current `master` (or `nightly`). I was running into the same issue, until I realized that there might have been additions recently.

ShriprajwalK: Oh yes, you are right. I thought I was running the latest build but I wasn't. Sorry!

dkarrasch (changes requested):
I tested a little bit and I don't think this is well-defined for unsigned integers, when you are supposed to get negative numbers. Also, I'd suggest to compare to the result when the numbers are represented as floats, so something along the following lines:
```julia
@testset "rem for $T RoundNearest" for T in (Int8, Int16, Int32, Int64, Int128)
    for (n, r) in zip(3:7, -2:2)
        @test rem(T(n), T(5), RoundNearest) == rem(float(n), 5.0, RoundNearest) == r
    end
end
```

base/div.jl
```diff
@@ -67,6 +67,7 @@ rem(x, y, r::RoundingMode)
 rem(x, y, ::RoundingMode{:ToZero}) = rem(x,y)
 rem(x, y, ::RoundingMode{:Down}) = mod(x,y)
 rem(x, y, ::RoundingMode{:Up}) = mod(x,-y)
+rem(x, y, r::RoundingMode{:Nearest}) = x - y*div(x, y, r)
```
simonbyrne: Since this `div` is defined in terms of `divrem`, it would make sense to use it for `rem` as well.
```suggestion
rem(x, y, r::RoundingMode{:Nearest}) = divrem(x, y, r)[2]
```

dkarrasch (approved):

sostock:
Defining `rem` in terms of `divrem` causes infinite recursion for `Rational`s:
```julia
julia> divrem(5//1, 2//1, RoundNearest)
ERROR: StackOverflowError:
Stacktrace:
 [1] divgcd(::Int64, ::Int64) at ./rational.jl:32
 [2] div at ./rational.jl:368 [inlined]
 [3] div at ./rational.jl:375 [inlined]
 [4] divrem(::Rational{Int64}, ::Rational{Int64}, ::RoundingMode{:Nearest}) at ./div.jl:128
 [5] rem(::Rational{Int64}, ::Rational{Int64}, ::RoundingMode{:Nearest}) at ./div.jl:70
 ... (the last 2 lines are repeated 39990 more times)
```
This is because `divrem` is calling `rem`:
https://github.com/JuliaLang/julia/blob/d7d20d3882951642ca3db2c85a83d2c50d44a244/base/div.jl#L120-L130

dkarrasch:
Thanks, @sostock, for reporting. I think I have a fix for that. The definition via `divrem` makes sense in the `Integer` case, but, as you pointed out, not in the generic case. So, having
```julia
rem(x, y, r::RoundingMode{:Nearest}) = x - y*div(x, y, r)
rem(x::Integer, y::Integer, r::RoundingMode{:Nearest}) = divrem(x, y, r)[2]
```
should solve the problem.

PR #35035, opened 2020-03-07
Added test for hypot with Complex arguments

ShriprajwalK:


simeonschaub (commented):
Might be good to have a test with a non-zero imaginary part as well? I think, we should also test that the result is actually a subtype of Real, either explicitely as `hypot(...) isa Real` or by testing equality with`===` instead of `==`, so we can make sure the types match.

test/math.jl
```diff
@@ -1049,5 +1049,6 @@ end
     using .Main.Furlongs
     @test hypot(Furlong(0), Furlong(0)) == Furlong(0.0)
     @test hypot(Furlong(3), Furlong(4)) == Furlong(5.0)
-    @test hypot(Complex(3), Complex(4)) == 5.0
+    @test hypot(Complex(3), Complex(4)) === 5.0
+    @test hypot(Complex(6 + 8im), Complex(8 + 6im)) === 10.0*sqrt(2)
```
ViralBShah: Should be either of the form `Complex(6,8)` or just `6+8im`. Both are redundant.

PR #35039, opened 2020-03-07
WIP: replace some characters in the symbol for ccall library names that can n cause the linker to fail on Windows

KristofferC:
Ref: https://github.com/JuliaLang/PackageCompiler.jl/issues/365

Just getting the build bots to run this so I can get an executable to try this with.

### DIFF
--- a/src/ccall.cpp
+++ b/src/ccall.cpp
@@ -82,6 +82,8 @@ static bool runtime_sym_gvs(const char *f_lib, const char *f_name, MT &&M,
         std::string name = "ccalllib_";
         name += llvm::sys::path::filename(f_lib);
         name += std::to_string(globalUnique++);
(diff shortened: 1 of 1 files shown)


KristofferC:
This fixes the MWE I have for this issue.

vtjnash:
The `-` is already on our blacklist (https://github.com/JuliaLang/julia/blob/master/src/aotcompile.cpp#L191), but this is possibly an llvm bug with `.drectve` emission (documentation for this linker flag suggests using quotes around it, but emitLinkerFlagsForGlobalCOFF omits them)

KristofferC:
Thanks, so can something like this PR be used to workaround the upstream issue or what would be the best way forward?

vtjnash:
Make a commit to LLVM (https://reviews.llvm.org or with `arc diff`) and add the patch here to Julia

PR #35041, opened 2020-03-08
Bump OpenBLAS yet again

Keno:
The previous patch I had was incomplete.


ViralBShah:
The number of files these things touch is a bit insane!

Keno:
yeah, each checksum is a separate file because that's easy to check in a shell script, but we should probably do something more intelligent here, like have a tab-separated file like
```
OpenBlas+8 sha512 abcd...
OpenBlas+8 sha3 abcd...
```
Shouldn't be too bad to handle with standard unix tools.

ViralBShah:
wouldn't standard unix tools break on windows?

Keno:
No, our build system already depends on them, so we assume their presence on windows (either through cygwin or msys).

PallHaraldsson:
Is there a reason to not go to the latest version (now 0.3.9)? Possibly this change was in the pipeline for a long time. I however see e.g. in latest version:

* Fixed cpu autodetection of Goldmont+, Cannon Lake and Ice Lake
and
* Improved speed of the AVX512 GEMM3M kernel on SkylakeX

PR #35046, opened 2020-03-08
fix a couple missing GC.at-preserve

rfourquet:
Some mentioned in [discourse](https://discourse.julialang.org/t/on-the-garbage-collection/35695/3). 

(tentatively applying the "GC" label, but not sure if appropriate)

### DIFF
--- a/base/hashing2.jl
+++ b/base/hashing2.jl
@@ -238,8 +238,6 @@ const memhash_seed = UInt === UInt64 ? 0x71e729fd56419c81 : 0x56419c81
 
 function hash(s::Union{String,SubString{String}}, h::UInt)
     h += memhash_seed
-    # note: use pointer(s) here (see #6058).
-    GC.@preserve s ccall(memhash, UInt, (Ptr{UInt8}, Csize_t, UInt32),
-                         pointer(s), sizeof(s), h % UInt32) + h
+    ccall(memhash, UInt, (Ptr{UInt8}, Csize_t, UInt32), s, sizeof(s), h % UInt32) + h
 end
 hash(s::AbstractString, h::UInt) = hash(String(s), h)
--- a/base/hashing2.jl
+++ b/base/hashing2.jl
@@ -18,15 +18,17 @@ end
 if GMP.Limb === UInt
     # used e.g. for Rational{BigInt}
     function hash_integer(n::BigInt, h::UInt)
-        s = n.size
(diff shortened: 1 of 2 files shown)


base/hashing2.jl
```diff
@@ -235,6 +239,7 @@ const memhash_seed = UInt === UInt64 ? 0x71e729fd56419c81 : 0x56419c81
 function hash(s::Union{String,SubString{String}}, h::UInt)
     h += memhash_seed
     # note: use pointer(s) here (see #6058).
-    ccall(memhash, UInt, (Ptr{UInt8}, Csize_t, UInt32), pointer(s), sizeof(s), h % UInt32) + h
+    GC.@preserve s ccall(memhash, UInt, (Ptr{UInt8}, Csize_t, UInt32),
```
yuyichao: I'm pretty sure just removing the `pointer` here is valid. Looking at the comment above I don't think there's a precise explaination of why one has to use `pointer`. I can't tell if it is to avoid copying by not using `bytestring` (at the time) or for avoiding the check that is now done with `Cstring`. In either case, I checked the definition of `cconvert` and `unsafe_convert` for `SubString{String}` and I believe just removing the `pointer` is fine here.

yuyichao (approved):

yuyichao:
The test failure is the usual bootstrap thing. If any of these are defined before `GC.@preserve` is available, you'll need to use the `@_gc_preserve_begin` and `@_gc_preserve_end` from `essential.jl`.

Thanks for picking this up again.

base/hashing2.jl
rfourquet: I will trust you on this one and remove it. This comment was added in https://github.com/JuliaLang/julia/commit/27478643ba8cef82a2cd6c4a5346e5d5446b425a in 2014, so pinging @stevengj if you can confirm that the reason for using `pointer` here is obsolete.

rfourquet:
> you'll need to use the @_gc_preserve_begin and @_gc_preserve_end from essential.jl.

Thanks for the hint!

base/hashing2.jl
stevengj: Yes, I think it can be removed.

(I vaguely recall that `SubString{String}` at the time didn't have `convert` function to `Ptr{UInt8}`?  Nowadays it should work fine for substrings, e.g. `Base.unsafe_convert(Ptr{UInt8}, SubString("foo",2,2))` works.)

Issue #31104, opened 2019-02-18
JVM fails to load in 1.1 (JavaCall.jl)


aviks:
In 1.1, trying to initialise the java virtual machine in Julia via JavaCall.jl causes a stack overflow error. It works correctly when the code is loaded with `-i` or  `-e` but fails when used from the REPL. 

```
ERROR: StackOverflowError:
Stacktrace:
 [1] init(::Array{String,1}) at /Users/rene/.julia/packages/JavaCall/toamy/src/jvm.jl:176
 [2] top-level scope at none:0
```

Ref: https://github.com/JuliaInterop/JavaCall.jl/issues/96

The call in question is the `JNI_CreateJavaVM` function in the JVM. More information about the JVM invocation API is at https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/invocation.html

This is a regression from 1.0.x

aviks:
@JeffBezanson suggested trying with `ALWAYS_COPY_STACKS`. However defining this in `options.h` and rebuilding julia did not change the failure.

aviks:
*Maybe* related? https://github.com/JuliaInterop/RCall.jl/pull/293

JeffBezanson:
I get a hang instead of a stack overflow:
```
julia> using JavaCall

julia> JavaCall.init(["-Xmx128M"])

```
stack trace:
```
#0  pthread_cond_timedwait@@GLIBC_2.3.2 () at ../nptl/sysdeps/unix/sysv/linux/x86_64/pthread_cond_timedwait.S:238
#1  0x00007ff55d7204e2 in os::PlatformEvent::park(long) () from /usr/lib/jvm/default-java/jre/lib/amd64/server/libjvm.so
#2  0x00007ff55d70bd93 in ObjectMonitor::EnterI(Thread*) () from /usr/lib/jvm/default-java/jre/lib/amd64/server/libjvm.so
#3  0x00007ff55d70c143 in ObjectMonitor::enter(Thread*) () from /usr/lib/jvm/default-java/jre/lib/amd64/server/libjvm.so
#4  0x00007ff55d4fbf1a in instanceKlass::initialize_impl(instanceKlassHandle, Thread*) ()
   from /usr/lib/jvm/default-java/jre/lib/amd64/server/libjvm.so
#5  0x00007ff55d4fc571 in instanceKlass::initialize(Thread*) ()
   from /usr/lib/jvm/default-java/jre/lib/amd64/server/libjvm.so
[...]

PR #35048, opened 2020-03-09
Allow the the REPL backend to run on the root Task

mkitti:
This change allows the REPL backend to run on the root `Task` rather than running using `@async` on a new `Task` as done in `REPL.start_repl_backend`. When the change is activated by setting the environmental variable `JULIA_ROOT_REPL=1`, the statement `Base.roottask === Base.current_task()` will evaluate to `true` in the REPL.

Use of the environmental variable `JULIA_ROOT_REPL` is mainly meant to permit evaluation and testing in the near term. Movement of the backend REPL loop to `REPL.repl_backend_loop` and the addition of `backend_on_current_task` keyword argument to `REPL.run_repl` also facilitates running the REPL backend on the root Task by an outside package.

JavaCall is a package for interoperability with the Java Virtual Machine that could benefit from a REPL backend running on the root Task. Currently `JULIA_COPY_STACKS=1` is required for JavaCall.jl to function properly on Mac and Linux. On the Windows operating system, `JULIA_COPY_STACKS=1` is not needed for JavaCall and may create more issues. This approach avoids the performance degradation involved in enabling `JULIA_COPY_STACKS=1`. If the REPL backend were run on the root `Task`, then `JULIA_COPY_STACKS=1` would not be required to run JavaCall on the REPL.

The status quo results in a segmentation fault on Linux run trying to run JavaCall:
```julia
$ ./julia --banner=no
julia> versioninfo()
Julia Version 1.5.0-DEV.422
Commit de4f8211b9* (2020-03-08 21:42 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: AMD FX(tm)-8350 Eight-Core Processor
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-9.0.1 (ORCJIT, bdver1)

julia> import Pkg; Pkg.status()
     Status `~/.julia/environments/v1.5/Project.toml`
   494afd89 JavaCall v0.7.3

julia> using JavaCall; JavaCall.init()
┌ Warning: JavaCall needs the environment variable `JULIA_COPY_STACKS` to be `1` or `yes`.
│ Calling the JVM may result in undefined behavior.
└ @ JavaCall ~/.julia/packages/JavaCall/0wJQZ/src/JavaCall.jl:38
Segmentation fault (core dumped)
```

With `JULIA_ROOT_REPL = 1` and the REPL backend running on the root `Task`, a segmentation fault does not occur and JavaCall functions perfectly. `JULIA_COPY_STACKS` does not need to be enabled here for JavaCall to function.
```julia
$ JULIA_ROOT_REPL=1 ./julia --banner=no
julia> versioninfo()
Julia Version 1.5.0-DEV.422
Commit de4f8211b9* (2020-03-08 21:42 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: AMD FX(tm)-8350 Eight-Core Processor
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-9.0.1 (ORCJIT, bdver1)
Environment:
  JULIA_ROOT_REPL = 1

julia> import Pkg; Pkg.status()
     Status `~/.julia/environments/v1.5/Project.toml`
   494afd89 JavaCall v0.7.3

julia> using JavaCall; JavaCall.init()
┌ Warning: JavaCall needs the environment variable `JULIA_COPY_STACKS` to be `1` or `yes`.
│ Calling the JVM may result in undefined behavior.
└ @ JavaCall ~/.julia/packages/JavaCall/0wJQZ/src/JavaCall.jl:38

julia> jlm = @jimport java.lang.Math
JavaObject{Symbol("java.lang.Math")}

julia> jcall(jlm, "sin", jdouble, (jdouble,), pi/2)
1.0

julia> jcall(jlm, "sin", jdouble, (jdouble,), pi/4)
0.7071067811865475
```

Further thoughts:
1. Perhaps running the REPL backend on the root `Task` should be the default behavior. I am not aware of a situation that actually requires the backend to run under `@async` on a non-root `Task`
2. This approach may allow Julia to drop the need for `JULIA_COPY_STACKS`, simplifying the code base.
3. Packages like JavaCall could enable functions that need to be executed from the root `Task` to be called on non-root `Tasks` through a similar `Channel` based system. Java is fully capable of asynchronous multithreaded behavior on its own.

### DIFF
--- a/base/client.jl
+++ b/base/client.jl
@@ -380,7 +380,8 @@ function run_main_repl(interactive::Bool, quiet::Bool, banner::Bool, history_fil
             # REPLDisplay
             pushdisplay(REPL.REPLDisplay(active_repl))
             _atreplinit(active_repl)
-            REPL.run_repl(active_repl, backend->(global active_repl_backend = backend))
+            backend_on_root_task = get(ENV,"JULIA_ROOT_REPL","") in ("1","yes")
+            REPL.run_repl(active_repl, backend->(global active_repl_backend = backend), backend_on_current_task=backend_on_root_task)
         end
     else
         # otherwise provide a simple fallback
--- a/base/client.jl
+++ b/base/client.jl
@@ -380,8 +380,7 @@ function run_main_repl(interactive::Bool, quiet::Bool, banner::Bool, history_fil
             # REPLDisplay
             pushdisplay(REPL.REPLDisplay(active_repl))
             _atreplinit(active_repl)
-            backend_on_root_task = get(ENV,"JULIA_ROOT_REPL","") in ("1","yes")
-            REPL.run_repl(active_repl, backend->(global active_repl_backend = backend), backend_on_current_task=backend_on_root_task)
+            REPL.run_repl(active_repl, backend->(global active_repl_backend = backend))
         end
     else
         # otherwise provide a simple fallback
--- a/stdlib/REPL/src/REPL.jl
+++ b/stdlib/REPL/src/REPL.jl
@@ -1,5 +1,16 @@
 # This file is a part of Julia. License is MIT: https://julialang.org/license
 
+"""
+Run Evaluate Print Loop (REPL)
+
+    Example minimal code
+    ```
+    import REPL
+    term = REPL.Terminals.TTYTerminal("dumb", stdin, stdout, stderr)
+    repl = REPL.LineEditREPL(term, true)
+    REPL.run_repl(repl)
+    ```
+"""
 module REPL
 
 using Base.Meta, Sockets
@@ -135,6 +146,14 @@ function eval_user_input(@nospecialize(ast), backend::REPLBackend)
     nothing
 end
 
+"""
+    start_repl_backend(repl_channel::Channel,response_channel::Channel)
+
+    Starts loop for REPL backend
+    Returns a REPLBackend with backend_task assigned
+
+    Deprecated since sync / async behavior cannot be selected
+"""
 function start_repl_backend(repl_channel::Channel, response_channel::Channel)
     # Maintain legacy behavior of asynchronous backend
     backend = REPLBackend(repl_channel, response_channel, false)
@@ -142,6 +161,15 @@ function start_repl_backend(repl_channel::Channel, response_channel::Channel)
     backend.backend_task = @async start_repl_backend(backend)
     return backend
 end
+
+"""
+    start_repl_backend(backend::REPLBackend)
+
+    Call directly to run backend loop on current Task.
+    Use @async for run backend on new Task.
+
+    Does not return backend until loop is finished.
+"""
 function start_repl_backend(backend::REPLBackend,  @nospecialize(consumer = x -> nothing))
     backend.backend_task = Base.current_task()
     consumer(backend)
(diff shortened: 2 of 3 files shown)


JeffBezanson:
I think this should be the default; I'm not sure we need the environment variable at all.

We still need `JULIA_COPY_STACKS`, but we also need a better general solution to this problem. In the meantime, I think this is a good improvement given how simple it is.

mkitti:
Would you like me to revise the PR?

I will:
1. Eliminate the environmental variable
2. Set the keyword argument `backend_on_current_task`  of `run_repl` to default to `true`
3. Revert client.jl to master

c42f:
As mentioned on slack, a nicer API to resolve the java issues would be to have a custom calling convention. Jeff mentioned this here: https://github.com/JuliaLang/julia/issues/31104#issuecomment-484256248

I wonder whether a crude implementation of this could be done right now, taking your general approach of running on the root task, but wrapping a rather different API around it than you have here.

My thought would be to extend the new `@ccall` macro to add a calling convention argument and use this for calling into the JVM inside JavaCall. For example the following ccall 

https://github.com/JuliaInterop/JavaCall.jl/blob/4989b64cb69cbd3e0409ec46a62e6124628cdb8a/src/core.jl#L222

might become

```julia
@ccall callmethod(penv::Ptr{JNIEnv},
                  obj.ptr::Ptr{Nothing},
                  jmethodId::Ptr{Nothing},
                  convertedArgs::Ptr{Nothing})::$x force_main_stack=true
```

The new `force_main_stack=true` attribute (better / more accurate name needed?) would force the call to happen on the main stack (for this, the root task will do, as you have found).

Implementation-wise, changing `@ccall` to emit the right code should be fairly accessible. I feel like the tricky part would be making sure the root task is available for running such code, and managing the communication between the root task and the task which actually wants to do the call.

The benefit of this approach is that it won't break JavaCall when people try to use it from a task which is `@async` or `@spawn`d, rather than directly from the REPL backend task.

(Having said all that, I'm not opposed to merging this change per se. It just would be nice to aim for the more general solution if you have the time to pursue it.)

vtjnash:
If we forced both front and back end to live on different tasks, the ‘force_main_stack’ option implementation becomes relatively simple to implement using existing primitives (using yieldto, with the root task running a simple ccall dispatch worker). Personally, I think this PR is closer to optimal though, until Java fixes their implementation, though I know other developers who reasonably disagree.

c42f:
> using yieldto, with the root task running a simple ccall dispatch worker

Great, I was thinking `yieldto` could come into the story, but not sure of the detail.

> Personally, I think this PR is closer to optimal though, until Java fixes their implementation, though I know other developers who reasonably disagree.

But does the current PR help JavaCall in anything other than trivial code where nobody ever calls `@async` or `@spawn`?

mkitti:
By providing access to the root Task from the standard REPL, packages like JavaCall should be able then engineer solutions like @vtjnash suggested.

As for multithreading and JavaCall, it is best to leave multiprocessing of Java functions to Java.

After this PR is merged, the JavaCall package would have a few engineering options to handle `jcall` under `@async` or `@spawn` from the standard REPL including launching the worker that @vtjnash suggested. To continue the REPL, JavaCall could launch a second REPL backend loop using `@async` to continue handling new commands from the REPL frontend.

To be clear, this PR does nothing specific for JavaCall. It simply makes the REPL commands execute on the root Task by default, which enables further engineering from within the REPL that is currently not possible. The other solutions require setting `JULIA_COPY_STACKS` before starting Julia and the REPL or starting the the REPL in a non-conventional fashion such as in [RootTaskREPL.jl](https://github.com/mkitti/RootTaskREPL.jl). JavaCall is just the motivation.

Additionally, `JULIA_COPY_STACKS` does not work on Windows leading to some confusion [ [1](https://discourse.julialang.org/t/problem-in-running-julia-on-juno/35412/2), [2](https://travis-ci.org/github/mkitti/JavaCall.jl) ].

Even if this only enabled trivial code from the REPL, it is much more user friendly for potential new Julia users who would like to interface with legacy Java code. Seeing warnings or segmentation faults and having to set OS specific environmental variables is less than ideal.

JeffBezanson (approved):

mkitti:
Should we mention something about this in NEWS?

PR #35049, opened 2020-03-09
move fld/div tests to appropriate testset

dkarrasch:
The underlying issue seems to have been long resolved, tests pass locally on my mac.

### DIFF
--- a/test/numbers.jl
+++ b/test/numbers.jl
@@ -6,11 +6,6 @@ using LinearAlgebra
 
 const ≣ = isequal # convenient for comparing NaNs
 
-# remove these tests and re-enable the same ones in the
-# testset "issue #4156" later in this file when #23866 is resolved
-@test fld(0.3,0.01) == 29.0
-@test div(0.3,0.01) == 29.0
-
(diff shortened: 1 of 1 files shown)


dkarrasch:
Hm, while the linux64 tests seem to officially pass, there is an ugly llvmcall error
```julia
llvmcall                           (6) |        started at 2020-03-09T09:28:39.612
      From worker 6:	Internal error: encountered unexpected error during compilation of undeclared_ceil:
      From worker 6:	ErrorException("Failed to parse LLVM Assembly: 
      From worker 6:	julia: llvmcall:3:18: error: use of undefined value '@llvm.ceil.f64'
      From worker 6:	%2 = call double @llvm.ceil.f64(double %0)
      From worker 6:	                 ^
      From worker 6:	")
```
which may be kind of related. It refers to `ceil`, even though the moved tests use `fld` and `div`, but still close enough to be concerned. Or has this happened before and is actually unrelated? I'm confused that I'm getting a checkmark in the tests anyway...

simonbyrne (approved):

PR #35052, opened 2020-03-09
Curried versions of endswith and startswith

### LINKED ISSUE #33193: Curried versions of endswith and startswith
In the spirit of https://github.com/JuliaLang/julia/pull/30915 I was wondering whether a PR for adding single-argument versions of `endswith`, `startswith`, and `occursin` that can be used as follows in functions like `filter` would be appreciated.

`filter(endswith(".jl"), readdir())`

I feel that I have definitely typed `x->endswith(x, ".jl")` far too often :)

carstenbauer:
See discussion in #33193.

This PR adds single-argument versions `endswith(s)` and `startswith(s)` that return functions semantically equivalent to `x->endswith(x,s)` and `x->startswith(x,s)`, respectively.

Example use case: `filter(endswith(".toml"), readdir())`

Closes #33193.


StefanKarpinski (approved):

fredrikekre (commented):
Seems like just two test cases would cover everything?

crstnbr:
> Seems like just two test cases would cover everything?

I started out with a definition of the curried versions including `::AbstractString` annotations in which case `Char` support was lost. With these tests I want to make sure that whatever the definition of these curried variants is/will be in the future they should accept every type that the parent methods can handle.

In any case, the extra tests certainly don't hurt, do they?

StefanKarpinski:
> In any case, the extra tests certainly don't hurt, do they?

They don't in this case since they're fast and cheap to do but they will get run a _lot_.

crstnbr:
> They don't in this case since they're fast and cheap to do but they will get run a _lot_.

Makes sense. I'll leave them as they are then unless someone objects.

rfourquet (approved):

crstnbr:
I've added a NEWS.md entry and compat annotations.

PR #35057, opened 2020-03-09
RFC: treat small negative λ as 0 for sqrt(::Hermitian)

stevengj:
This PR treats sufficiently small negative eigenvalues as "zero" for the purpose of taking matrix `sqrt` ~~and `log`~~, as otherwise one gets spurious complex results from semidefinite matrices due to roundoff errors.  (See also [discourse](https://discourse.julialang.org/t/matrix-square-root-of-known-positive-semi-definite-matrix-does-not-exist) discussion.)

It works by adding an `rtol` keyword parameter to `sqrt(::Hermitian)` ~~and `log(::Hermitian)`~~, specifying a relative tolerance for negative eigenvalues to ignore, which defaults to the same relative tolerance that we use for the `pinv` function.

~~This is currently just a rough draft to get reactions.~~  To do:
- [x] tests
- [x] documentation
- [x] news
- [x] remove `rtol` argument for `log`, where semidefinite matrices don't make sense

### DIFF
--- a/stdlib/LinearAlgebra/test/symmetric.jl
+++ b/stdlib/LinearAlgebra/test/symmetric.jl
@@ -669,7 +669,7 @@ end
         C = Symmetric(A*B*A'), # semidefinite up to roundoff
         Csqrt = sqrt(C)
         @test Csqrt isa Symmetric{Float64}
-        @test Csqrt*Csqrt ≈ C rtol=1e-15
+        @test Csqrt*Csqrt ≈ C rtol=1e-14
     end
     let D = Symmetric(Matrix(Diagonal([1 0; 0 -1e-14])))
         @test sqrt(D) ≈ [1 0; 0 1e-7im] rtol=1e-14
--- a/stdlib/LinearAlgebra/test/symmetric.jl
+++ b/stdlib/LinearAlgebra/test/symmetric.jl
@@ -663,4 +663,19 @@ end
     @test LinearAlgebra.hermitian_type(Int) == Int
 end
 
+@testset "sqrt(nearly semidefinite)" begin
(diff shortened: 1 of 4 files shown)


stevengj:
Oh, I guess this doesn't make sense in the case of a matrix `log`, since `log(0.0)` is `-Inf`, i.e. you shouldn't be taking the log of a semidefinite matrix anyway.

Note that this can change the *output* of `sqrt` by O(√ε), but it is *still backwards stable*.  That is, `B = sqrt(A)` is a matrix such that `‖B^2 - A‖ / ‖A‖` is O(ε).

stdlib/LinearAlgebra/test/symmetric.jl
```diff
@@ -663,4 +663,19 @@ end
     @test LinearAlgebra.hermitian_type(Int) == Int
 end
 
+@testset "sqrt(nearly semidefinite)" begin
+    let A = [0.9999999999999998 4.649058915617843e-16 -1.3149405273715513e-16 9.9959579317056e-17; -8.326672684688674e-16 1.0000000000000004 2.9280733590254494e-16 -2.9993900031619594e-16; 9.43689570931383e-16 -1.339206523454095e-15 1.0000000000000007 -8.550505126287743e-16; -6.245004513516506e-16 -2.0122792321330962e-16 1.183061278035052e-16 1.0000000000000002],
+        B = [0.09648289218436859 0.023497875751503007 0.0 0.0; 0.023497875751503007 0.045787575150300804 0.0 0.0; 0.0 0.0 0.0 0.0; 0.0 0.0 0.0 0.0],
+        C = Symmetric(A*B*A'), # semidefinite up to roundoff
+        Csqrt = sqrt(C)
+        @test Csqrt isa Symmetric{Float64}
+        @test Csqrt*Csqrt ≈ C rtol=1e-15
```
andreasnoack: The test failure on AArch64 is genuine. Since `C` has elements that are very small, I guess you'd have to use an absolute tolerance here?

andreasnoack (commented):
LGTM except for the test tolerances.

stdlib/LinearAlgebra/test/symmetric.jl
stevengj: You don't need an absolute tolerance just because elements are small — you only need an absolute tolerance if the *whole array* is going to zero.  Remember that the test is `‖M - M̃‖ ≤ rtol*‖M‖`

1e-15 is probably just too aggressive (i.e. too close to machine precision); I'll change it to 1e-15.

stevengj:
@andreasnoack, I updated the test tolerances.  Good to merge?

PR #35059, opened 2020-03-10
Windows installer: improve Desktop Icon and Start Menu group creation 

### LINKED ISSUE #34803: Windows installer creates a shortcut on the Desktop without version number
Upon installing Julia 1.3.1 Windows 64-bit from the official binaries, the final desktop shortcut that the installer created was simply called "Julia" instead of "Julia 1.3". IIRC, even in the 1.3.0 release, the shortcut used to have the minor version number, in which case this might be a regression.

### LINKED ISSUE #34867: Windows installer creates start menu group for each Julia version
Now the Windows start menu with multiple Julia versions installed looks like this
jl_startmenu
and the start menu folder like this with folders for old versions of Julia that have been uninstalled
jl_startfolder

Better arrangement could be that there was a single Julia-group under which the items would appear, like this
jl_startmenu_couldbe
and expanded
jl_startmenu_couldbe_open

This is a user visible issue when navigating the Windows start menu. Not a big issue if one has only a couple of Julia versions installed.

Default group menu name seems to be `DefaultGroupName="{#AppName} {#AppVersion}"` defined here
https://github.com/JuliaLang/julia/blob/v1.4.0-rc2/contrib/windows/build-installer.iss#L92

musm:
Fixes https://github.com/JuliaLang/julia/issues/34803 and https://github.com/JuliaLang/julia/issues/34867

- Add a page where user has the  option to create desktop icons and start menu groups
- Start menu icons are created without the folder structure (as recommended on Windows 10)
- Desktop icon now includes version number
- Installer changes the default installation back to  `C:\Users\Mus\AppData\Local\Programs\Julia - x.y` instead of `C:\Users\Mus\AppData\Local\Programs\Julia\Julia - x.y`.

I've looked into creating an auto-updater, but I'm slightly conflicted if it makes sense to do this. I suspect  people would like to have multiple versions of Julia installed for testing and having an "auto-updater" installer would defeat this.

### DIFF
--- a/contrib/windows/build-installer.iss
+++ b/contrib/windows/build-installer.iss
@@ -1,10 +1,8 @@
-; preprocessor will fill in the commented out constants below
-; #define AppVersion "x.y.z"; Julia version
-; #define AppSourceFiles "C:\ex" ; Julia build root directory
-; #define AppHomeFiles "C:\ex" ; Julia home directory
 #define AppName "Julia"
-#define AppMainExe "bin\julia.exe"
+#define AppNameLong AppName + " " + AppVersion
+#define AppMainExeName "bin\julia.exe"
 #define CurrentYear GetDateTimeString('yyyy', '', '')
+#define DirName AppName + " " + AppVersion
 
 
 [LangOptions]
@@ -16,20 +14,20 @@ TitleFontName=Segoe UI
 
 [Messages]
 SetupAppTitle={#AppName} Installer
-SetupWindowTitle=Installer - {#AppName} {#AppVersion}
+SetupWindowTitle=Installer - {#AppNameLong}
 UninstallAppTitle={#AppName} Uninstaller
-UninstallAppFullTitle=Uninstaller - {#AppName} {#AppVersion}
+UninstallAppFullTitle=Uninstaller - {#AppNameLong}
 WizardSelectDir=Select Installation Directory
 SelectDirDesc=
 SelectDirLabel3=
-SelectDirBrowseLabel=Please run installer as Administrator to install {#AppName} system wide.%n%nTo install into a different folder click Browse.
+SelectDirBrowseLabel=Restart installer as Administrator to install {#AppName} system wide.%n%n%nInstallation directory:
 WizardPreparing=Installing
 PreparingDesc=
 InstallingLabel=
 ClickFinish=
-FinishedHeadingLabel=Installation complete
+FinishedHeadingLabel=Installation Successfull
 FinishedLabelNoIcons=[name] has been successfully installed.
-FinishedLabel=
+FinishedLabel=[name] has been successfully installed.
 StatusExtractFiles=Extracting...
 StatusUninstalling=Removing...
 ConfirmUninstall=Are you sure you want to completely remove {#AppName}?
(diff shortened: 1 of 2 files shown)


oscardssmith:
How hard would optional auto updating be? My guess is that most uses who read GitHub probably have multiple versions, but that among pure users there is a sizable contingent that wants is the fastest (newest) version.

musm:
It shouldn't be too hard, but the part I'm struggling with is how to prevent the installer from auto-updating.  

In order for the auto-updater to work the default installation directory will have to be simply `Julia`  and future updates would rewrite over this directory. I'm also looking into adding a `add to path option` but that will require a bit more machinery, I think.

ViralBShah:
So long as the auto-updater can keep multiple versions around, that's fine. It is ok to replace patch releases, but minor and major releases should co-exist.

johnnychen94:
> I've looked into creating an auto-updater, but I'm slightly conflicted if it makes sense to do this.

FYI, I've already made something for this purpose: https://github.com/johnnychen94/jill.py

contrib/windows/build-installer.iss
```diff
@@ -79,17 +70,18 @@ AppName={#AppName}
...
 WizardStyle=modern
 Compression=lzma2/ultra
 SolidCompression=yes
-SetupIconFile={#AppHomeFiles}\contrib\windows\julia.ico
-DefaultDirName={autopf}\{#AppName}\{#AppName}-{#AppVersion}
-DefaultGroupName="{#AppName} {#AppVersion}"
+DefaultDirName={autopf}\{#AppNameLong}
+UsePreviousPrivileges=no
+PrivilegesRequired=lowest
+DefaultGroupName="{#AppNameLong}"
```
jaakkor2: Instead of `DefaultGroupName="{#AppNameLong}"` , could this be just `DefaultGroupName="{#AppName}" which would fix https://github.com/JuliaLang/julia/issues/34867 ?

musm: Note please see the linked issues. We shouldn't install the Julia Icon into a folder structure as a general recommendation by Windows.

The changes in this PR install the Julia x.y ver, directly into the menu.

musm:
Additional options added by this PR

image


option to install portable mode, option to create start menu entry, option to create desktop shortcut

musm:
portable mode can be created from the following command line

```
julia-installer.exe /DIR=C:\temp\julia /TASKS=portablemode /VERYSILENT
```

davidanthoff:
I think an auto-updater would be nice, but I think getting that right is difficult and needs a lot of design. In particular, I think it should be an update strategy that can handle multiple Julia versions at the same time, and ideally be integrated with the Windows Store. I've thought a fair bit about this, have some home grown set of PowerShell scripts that do this for my system. I think this might make for a good Juliacon project to come up with a strategy for this, I don't think we should rush something in at this point.

musm:
The annoying things with the portable check box is  I don't know how to make it unselect: create desktop menu and create start menu entry...

musm:
Are we happy with the final installation page?

image

musm:
Portable mode is now possible with 
```
.\julia-win64.exe  /DIR=C:\temp\julia /TYPE=portable 
```

musm:
view


Updated installer!

ViralBShah:
Is the difference between full and portable the start menu entry and desktop icon creation?

musm:
Yes. Portable mode, does not include any registry modifications, desktop or start menu groups and does not install an "uninstaller" under "programs & files". Essentially it simply extracts julia

musm:
How about ?
image

ViralBShah:
I think the portable needs a bit more explanation. Also missing the closing parentheses.

musm:
I'm going to remove that from this PR from now and just resolve the icon creation issues.

KristofferC:
For portable installations, should the `.julia` directory also go into the installation directory etc?

musm:
Updated original PR comments. Will merge after green.

ViralBShah:
I think it would be great if we could provide a zip file for people who want to install Julia without it modifying anything on their system. Should be fairly easy.

PR #35060, opened 2020-03-10
fix typo in Distributed create_worker timeout

tanmaykm:
fix typo in code that deals in timing out worker setup (introduced in #34502)


JeffBezanson:
This isn't a typo. `@async` and `@spawn` allow using `$` to interpolate the values of variables instead of capturing variable bindings.

tanmaykm:
Ah! Looks like that's something new in 1.4. I was not aware of that. Thanks!

PR #35061, opened 2020-03-10
Generalize and specialize Schur decomposition

### LINKED ISSUE #699: slow Schur factorization for Hermitian matrices
For a Hermitian matrix, Schur factorization is equivalent to the eigenvalue decomposition, but it seems to be much slower:
```jl
julia> A = rand(1000,1000); H = Symmetric(A);

julia> @btime eigen($H);
  141.423 ms (15 allocations: 23.25 MiB)

julia> @btime schur($H);
  502.031 ms (13 allocations: 23.16 MiB)
```
Seems like we just need some specialized methods for `schur`?

### LINKED ISSUE #35054: slow Schur factorization for Hermitian matrices
For a Hermitian matrix, Schur factorization is equivalent to the eigenvalue decomposition, but it seems to be much slower:
```jl
julia> A = rand(1000,1000); H = Symmetric(A);

julia> @btime eigen($H);
  141.423 ms (15 allocations: 23.25 MiB)

julia> @btime schur($H);
  502.031 ms (13 allocations: 23.16 MiB)
```
Seems like we just need some specialized methods for `schur`?

dkarrasch:
Closes JuliaLang/LinearAlgebra.jl#699.

With this PR:
```julia
julia> using LinearAlgebra, BenchmarkTools

julia> A = rand(1000,1000); H = Symmetric(A);

julia> @btime schur($H);
  206.667 ms (18 allocations: 30.88 MiB)

julia> @btime eigen($H);
  204.036 ms (15 allocations: 23.25 MiB)
```

### DIFF
--- a/stdlib/LinearAlgebra/src/schur.jl
+++ b/stdlib/LinearAlgebra/src/schur.jl
@@ -163,7 +163,7 @@ function schur(A::Union{UnitLowerTriangular{T},LowerTriangular{T}}) where {T}
     n = size(A, 1)
     J = zeros(t, n, n)
     for i in axes(J, 2)
-       J[n+1-i, i] = one(t)
+       J[n+1-i, i] = oneunit(t)
     end
     return Schur(Z, J, convert(Vector{t}, diag(A)))
 end
@@ -176,7 +176,7 @@ function schur(A::Bidiagonal{T}) where {T}
         n = size(A, 1)
         J = zeros(t, n, n)
         for i in axes(J, 2)
-            J[n+1-i, i] = one(t)
+            J[n+1-i, i] = oneunit(t)
         end
         dv = reverse!(Vector{t}(A.dv))
         ev = reverse!(Vector{t}(A.ev))
--- a/stdlib/LinearAlgebra/src/schur.jl
+++ b/stdlib/LinearAlgebra/src/schur.jl
@@ -143,7 +143,12 @@ schur(A::StridedMatrix{<:BlasFloat}) = schur!(copy(A))
 schur(A::StridedMatrix{T}) where T = schur!(copy_oftype(A, eigtype(T)))
 
 schur(A::Symmetric) = schur(copyto!(similar(parent(A)), A))
-schur(A::Hermitian) = schur(copyto!(similar(parent(A)), A))
+schur(A::Symmetric{<:Real}) = _schur(A)
+schur(A::Hermitian) = _schur(A)
+function _schur(A::RealHermSymComplexHerm)
+    F = eigen(A; sortby=nothing)
+    return Schur(typeof(F.vectors)(Diagonal(F.values)), F.vectors, F.values)
+end
 schur(A::UpperTriangular) = schur(copyto!(similar(parent(A)), A))
 schur(A::LowerTriangular) = schur(copyto!(similar(parent(A)), A))
 schur(A::Tridiagonal) = schur(Matrix(A))
--- a/stdlib/LinearAlgebra/src/schur.jl
+++ b/stdlib/LinearAlgebra/src/schur.jl
@@ -142,16 +142,36 @@ true
 schur(A::StridedMatrix{<:BlasFloat}) = schur!(copy(A))
 schur(A::StridedMatrix{T}) where T = schur!(copy_oftype(A, eigtype(T)))
 
-schur(A::Symmetric) = schur(copyto!(similar(parent(A)), A))
-schur(A::Symmetric{<:Real}) = _schur(A)
-schur(A::Hermitian) = _schur(A)
-function _schur(A::RealHermSymComplexHerm)
+schur(A::AbstractMatrix{T}) where {T} = schur!(copyto!(Matrix{eigtype(T)}(undef, size(A)...), A))
+function schur(A::RealHermSymComplexHerm)
     F = eigen(A; sortby=nothing)
     return Schur(typeof(F.vectors)(Diagonal(F.values)), F.vectors, F.values)
 end
-schur(A::UpperTriangular) = schur(copyto!(similar(parent(A)), A))
-schur(A::LowerTriangular) = schur(copyto!(similar(parent(A)), A))
-schur(A::Tridiagonal) = schur(Matrix(A))
+function schur(A::UpperTriangular{T}) where {T}
+    t = eigtype(T)
+    Z = Matrix{t}(undef, size(A)...)
+    copyto!(Z, A)
+    return Schur(Z, Matrix{t}(I, size(A)), convert(Vector{t}, diag(A)))
+end
+function schur(A::LowerTriangular{T}) where {T}
+    t = eigtype(T)
+    Z = [t(A[i, j]) for i in reverse(axes(A, 1)), j in reverse(axes(A, 2))]
+    n = size(A, 1)
+    J = [t(i + j == n + 1) for i in axes(A, 1), j in axes(A, 2)]
+    return Schur(Z, J, convert(Vector{t}, diag(A)))
+end
+function schur(A::Bidiagonal{T}) where {T}
+    t = eigtype(T)
+    if A.uplo == 'U'
+        return Schur(Matrix{t}(A), Matrix{t}(I, size(A)), Vector{t}(A.dv))
+    else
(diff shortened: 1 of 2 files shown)


stdlib/LinearAlgebra/src/schur.jl
```diff
@@ -143,7 +143,12 @@ schur(A::StridedMatrix{<:BlasFloat}) = schur!(copy(A))
 schur(A::StridedMatrix{T}) where T = schur!(copy_oftype(A, eigtype(T)))
 
 schur(A::Symmetric) = schur(copyto!(similar(parent(A)), A))
-schur(A::Hermitian) = schur(copyto!(similar(parent(A)), A))
+schur(A::Symmetric{<:Real}) = _schur(A)
+schur(A::Hermitian) = _schur(A)
+function _schur(A::RealHermSymComplexHerm)
```
stevengj: Can't this just be `function schur(A::RealHermSymComplexHerm)`?   Why do you need a separate `_schur` function here?

stevengj:
Do we have a test that exercises `schur(::Hermitian)`?

stdlib/LinearAlgebra/src/schur.jl
dkarrasch: Because otherwise I was getting an ambiguity error with `schur(A::Symmetric)`.

dkarrasch:
> Do we have a test that exercises `schur(::Hermitian)`?

Yes, we do.

https://github.com/JuliaLang/julia/blob/ff4f86714b5f90f81d421ba1fac678832eecb88f/stdlib/LinearAlgebra/test/schur.jl#L42-L45

stevengj:
I notice that schur.jl has:
```jl
schur(A::Symmetric) = schur(copyto!(similar(parent(A)), A))
schur(A::Hermitian) = schur(copyto!(similar(parent(A)), A))
schur(A::UpperTriangular) = schur(copyto!(similar(parent(A)), A))
schur(A::LowerTriangular) = schur(copyto!(similar(parent(A)), A))
schur(A::Tridiagonal) = schur(Matrix(A))
```
Should we just replace these with
```
schur(A::AbstractMatrix) = schur!(copyto!(Matrix{eigtype(A)}(undef, size(A)...), A))
```
As well as making this code more generic, this would allocate less memory and would eliminate the method ambiguity for `Symmetric` above.

(We might also eventually want specialized `schur` methods for triangular and tridiagonal matrices.)

dkarrasch:
I got two questions:

1. Do we have some high-level function to generate the "reverse identity matrix" in one go? `reverse(Matrix{eltype(A)}(I, size(A)); dims=2)` allocates twice.
2. Is there an "easy" trick what to do for tridiagonal matrices? I see what to do with bidiagonal ones, but not for tridiagonal ones.

dkarrasch:
Ok, so for the reverse identity I used a list comprehension, I guess this is not too bad. As for the tridiagonal matrices, I tried a small example, and, as expected, the Schur factor is dense, and doesn't seem to have any obvious relation with the tridiagonal matrix. Thus, the tridiagonal matrix type is best handled by the new generic `AbstractMatrix` method. Instead, I added methods for both `Bidiagonal` "types", and tests.

stevengj:
A (non-symmetric) tridiagonal matrix is a special case of an upper-Hessenberg matrix (and for the general case we also have an `UpperHessenberg` type).  Since the [first step of Schur factorization is computation of upper-Hessenberg factorization](https://www.netlib.org/lapack/lug/node50.html), this allows you to skip the first part of the computation.   In particular, you can call the xHSEQR function in LAPACK to get the Schur factorization directly from a Hessenberg matrix.

dkarrasch:
Thanks, @stevengj. Unfortunately, we don't seem to have bindings for xHSEQR in the LAPACK module yet, so this seems like some more effort. Shall we keep this PR as is, and address the tridiagonal case including the required LAPACK bindings later? As you pointed out above, we have currently
```julia
schur(A::Tridiagonal) = schur(Matrix(A))
```
so this PR introduces a potentially mild improvement by having only one matrix allocation of the appropriate `eigtype` right away. Currently, the `Tridiagonal` test exercises the generic method. 😃

stevengj (approved):

dkarrasch:
Good that I double-checked performance, and found that the list comprehensions were devastating. Now I found much nicer versions, for instance, for "flipping" a matrix. Here are some benchmarks, comparing new to the old behavior (`schur(Matrix(A))`).
```julia
let A = rand(1000, 1000)
           for T in (Symmetric, Hermitian, UnitUpperTriangular, UpperTriangular, UnitLowerTriangular, LowerTriangular)
               @btime schur($(T(A)))
               @btime schur(Matrix($(T(A))))
           end
           H = Bidiagonal(A, :U);
           @btime schur($H);
           @btime schur(Matrix($H));
           H = Bidiagonal(A, :L);
           @btime schur($H);
           @btime schur(Matrix($H));
       end

  209.768 ms (18 allocations: 30.88 MiB)
  766.071 ms (13 allocations: 23.16 MiB)

  213.434 ms (18 allocations: 30.88 MiB)
  853.173 ms (13 allocations: 23.16 MiB)

  3.138 ms (6 allocations: 15.27 MiB)
  7.059 ms (13 allocations: 23.16 MiB)

  2.801 ms (6 allocations: 15.27 MiB)
  7.517 ms (13 allocations: 23.16 MiB)

  3.123 ms (8 allocations: 15.27 MiB)
  53.193 ms (13 allocations: 23.16 MiB)

  2.757 ms (8 allocations: 15.27 MiB)
  54.063 ms (13 allocations: 23.16 MiB)

  1.240 ms (6 allocations: 15.27 MiB)
  6.936 ms (13 allocations: 23.16 MiB)

  1.108 ms (8 allocations: 15.27 MiB)
  173.280 ms (13 allocations: 23.16 MiB)
```

stdlib/LinearAlgebra/src/schur.jl
```diff
@@ -147,26 +147,37 @@ function schur(A::RealHermSymComplexHerm)
     F = eigen(A; sortby=nothing)
     return Schur(typeof(F.vectors)(Diagonal(F.values)), F.vectors, F.values)
 end
-function schur(A::UpperTriangular{T}) where {T}
+function schur(A::Union{UnitUpperTriangular{T},UpperTriangular{T}}) where {T}
```
stevengj: (Should we have an `AbstractTriangular{T,uplo}` type?)

stdlib/LinearAlgebra/src/schur.jl
```diff
@@ -147,26 +147,37 @@ function schur(A::RealHermSymComplexHerm)
...
     if A.uplo == 'U'
         return Schur(Matrix{t}(A), Matrix{t}(I, size(A)), Vector{t}(A.dv))
-    else
+    else # A.uplo == 'L'
+        # construct "reverse" identity
         n = size(A, 1)
-        J = [t(i + j == n + 1) for i in axes(A, 1), j in axes(A, 2)]
+        J = zeros(t, n, n)
+        for i in axes(J, 2)
+            J[n+1-i, i] = one(t)
```
stevengj: technically `oneunit(t)` to guarantee that it is the same type as `t`.  Though we don't actually handle dimensionful types here, I think.

stdlib/LinearAlgebra/src/schur.jl
dkarrasch: Indeed, I was searching for some type like that (or a union type), but couldn't find anything.

stdlib/LinearAlgebra/src/schur.jl
dkarrasch: Would it help to use `true` here and leave it to the implicit conversion in `setindex!`? I was unsure what to use here, but then decided to use `one(t)` just because I had that type at hand.

But I agree, if `A` had units, then those would appear in the Schur matrix, whereas the `J` just "rearranges" entries.

dkarrasch: In the latter (unit) case, I guess we would need to relax type constraints in the `Schur` struct.

dkarrasch:
Is this ready to go?

stevengj (approved):

PR #35066, opened 2020-03-10
Add guidelines for reviewing PRs on adding missing propagation

oxinabox:
This is one option to help with missing values.
By giving clear guidelines for reviewers as to what is valid and what is not.

Its just one option.
I am not myself convinced that its the best one (or even that it is all that good).
But I am putting it out there so people can think about it.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -327,8 +327,8 @@ The test added will be `@test foo(missing) === missing`.
 
 The conditions on if it is reasonable for such a overload to be added:
  - In typical use the function must not have any side-effects. (e.g. not `mkdir` or `Distributed.addprocs`).
- - The argument having missing propagation added must in typical use be a: `Number`, `AbstractString`, or `Symbol`.
- - In typical use the function must return a `Number`, `AbstractString`, or `Symbol`.
+ - The argument having missing propagation added must in typical use be a: `Number`, `AbstractString`, `Symbol`, `Dates.DatePeriod`, or `Dates.TimeType`.
+ - In typical use the function must return a `Number`, `AbstractString`, `Symbol`, `Dates.DatePeriod`, or `Dates.TimeType`.
  - The argument having pissing propagation added must not typically be being used for indexing, e.g. into a `Dict` or `Array`.
 
 If some practical reality does not align to the guidelines given in this section, rather than debate them in some PR where someone is proposing a new overload make another PR to update this section and link to it in the PR proposing the change.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -311,7 +311,31 @@ runtest harness).
    - To remove whitespace relative to the `master` branch, run
      `git rebase --whitespace=fix master`.
 
-## Resources
+# Notes for Code-Reviewers
+All people reviewing PRs to the Julia repository should be very familar with the above documentation for contributors.
+This section contains some additional notes for code-reviewers.
+Its also helpful for contibutors to read, though less important.
+
+## On reviewing PRs that add overloads to propagate missing values.
+There is often some confusion as to which overloads to functions to add missing value propagation are acceptable.
+Sometimes this leads to unproductive slippy-slope arguments: "If we add it to this function, we will have to add it to 200 more."
+Don't make slippy-slope arguments, instead refer to this section of this document.
+
+PRs to add functions must be good PRs, they must be correct and have tests like usual.
+Consider a function: `foo(x)`, where the PR is adding `foo(::Missing) = missing`.
+The test added will be `@test foo(missing) === missing`.
+
+The conditions on if it is reasonable for such a overload to be added:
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -311,7 +311,31 @@ runtest harness).
...
+Don't make slippy-slope arguments, instead refer to this section of this document.
+
+PRs to add functions must be good PRs, they must be correct and have tests like usual.
+Consider a function: `foo(x)`, where the PR is adding `foo(::Missing) = missing`.
+The test added will be `@test foo(missing) === missing`.
+
+The conditions on if it is reasonable for such a overload to be added:
+ - In typical use the function must not have any side-effects. (e.g. not `mkdir` or `Distributed.addprocs`).
+ - The argument having missing propagation added must in typical use be a: `Number`, `AbstractString`, or `Symbol`.
+ - In typical use the function must return a `Number`, `AbstractString`, or `Symbol`.
```
oxinabox: An interesting alternative set of types might be that it needs to broadcast like a scalar.

JeffBezanson:
This doesn't really make sense to me, since if we agree that these guidelines are correct then we should just go ahead and add all the applicable `missing` methods.

oxinabox:
> This doesn't really make sense to me, since if we agree that these guidelines are correct then we should just go ahead and add all the applicable missing methods

Maybe.

Though, we might get all of them right now, if we just go through some calls to `methodswith` and check everything.
But in a few releases time, some more methods will have been added, and odds are we won't have thought to check if they need missing propagation.

CONTRIBUTING.md
```diff
@@ -311,7 +311,31 @@ runtest harness).
...
+
+PRs to add functions must be good PRs, they must be correct and have tests like usual.
+Consider a function: `foo(x)`, where the PR is adding `foo(::Missing) = missing`.
+The test added will be `@test foo(missing) === missing`.
+
+The conditions on if it is reasonable for such a overload to be added:
+ - In typical use the function must not have any side-effects. (e.g. not `mkdir` or `Distributed.addprocs`).
+ - The argument having missing propagation added must in typical use be a: `Number`, `AbstractString`, `Symbol`, `Dates.DatePeriod`, or `Dates.TimeType`.
+ - In typical use the function must return a `Number`, `AbstractString`, `Symbol`, `Dates.DatePeriod`, or `Dates.TimeType`.
+ - The argument having pissing propagation added must not typically be being used for indexing, e.g. into a `Dict` or `Array`.
```
KristofferC: Since Base is not special these guidelines should also be in the docs for Missing, telling people that they should create all these extra overloads for their own functions if they match these criteria. Having this in some internal contributing doc doesn't make sense. Just document it properly then.

oxinabox: Good idea.

JeffBezanson:
>  But in a few releases time, some more methods will have been added, and odds are we won't have thought to check if they need missing propagation.

That's a big part of the problem. Adding lots of "every time you do X, you should also do Y" rules does not scale well. In Base we could maybe manage it, but over the whole ecosystem it doesn't seem realistic. That's why I think it's better to establish a pattern of handling it at the call site.

CONTRIBUTING.md
strickek: Wording in line 332 should be improved

jonas-schulze:
> That's why I think it's better to establish a pattern of handling it at the call site.

Maybe it would be good to add a Haskellesque `liftM` that does the promotion of `Missing` on the call side. `Union{T,Missing} where T` feels a lot like `Maybe T` to me.

```
liftM(f, args...) = any(ismissing, args) ? missing : f(args...)
```

oxinabox:
We have other issues and PRs about lifting operators..I recommend limiting discussion of then to "They exist and under proposal [link to relevant PR or issue] they do everything you could ever want, and thus they make the addition of any more methods that accept missings pointless. Callsite is better because ..."

And not allow this PR to get detailed into the details of lifting operators or other solutions.

StefanKarpinski:
This reads as a passive aggressive scolding to the core devs for not accepting every PR that has been made to Base to add missing methods to arbitrary functions. I agree that the current missing situation is untenable, but adding this kind of doc change is not going to help. If we could have a constructive discussion about exactly what subset of functions should have missing methods and then adding all of them, that would help. Adding `?` syntax for lifting would also help. Even adding a language mechanism for automatic lifting under the right circumstances would be fine—arguably the best solution. But in order for that to happen, we need rules on when that should happen.

oxinabox:
> This reads as a passive aggressive scolding to the core devs for not accepting every PR that has been made to Base to add missing methods to arbitrary functions. 

This was in *no* way intended that way. Sorry if it comes across that way.

You were in that discussion on slack, and said creating guidelines for what missing value PRs should have might be worth considering.
And so I transferred this off slack so it would not get lost.

PR #29240, opened 2018-09-18
implement "nearly division less" algorithm for rand(a:b)

rfourquet:
Cf. https://arxiv.org/abs/1805.10941.
Closes #20582, #29004.

This is still about 30 percent slower than the "biased" solution (e.g. #28987), but as fast as the "best case" of the current version (when the range length is a power of 2). So this is strictly better, and should become the default version in a future (major) release. How do we make this available  in the meantime? `rand(Random.fast(a:b))` is what the PR currently implements.

PR #35078, opened 2020-03-11
faster randn!(::MersenneTwister, ::Array{Float64})

rfourquet:
The main idea is that 
1. `randn()` essentially needs one random `Float64` value (a bit more in rare cases)
2. `rand!(rng::MersenneTwister, ::Array{Float})` is optimized, compared to calling `rand(rng, Float64)` in a loop

So for `randn!`, let's use this optimized version by filling the array with _uniform_ float values, and then do a second pass to transform them into normal values. The same applies to `randexp!` but with a smaller improvement.

There is a threshold on the length of the array between 10-20 below which this refined algorithm is less performant, so we switch the algorithm depending on that length. Here are the speed-ups on my machine:

| length     | 5    | 10   | 50   | 100  | 300  | 600  | 1 000 | 10 000 | 100 000 | 1 000 000 |
| ----------------- | ---- | ---  | ---  | ---  | ---  | ---  | ---   | ---    | ---     | ---       |
| randn!            | 0.88 | 1.04 | 1.57 | 1.70 | 1.81 | 1.89 | 1.92  | 1.94   | 1.92    | 1.85      |
| randexp!          | 0.87 | 1.14 | 1.27 | 1.37 | 1.37 | 1.34 | 1.42  | 1.39   | 1.37    | 1.31      |

So for small arrays there is small "speed-down".

As future work, it would be interesting to see if this also applies to some other `AbstractArray`s, and to other RNGs which have optimized array generation, in which case we could have a "trait" for enabling this improvement.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -154,6 +154,8 @@ Standard library changes
 
 #### Random
 
+* `randn!(::MersenneTwister, ::Array{Float64})` is faster, and as a result, for a given state of the RNG,
+  the corresponding generated numbers have changed ([#35078]).
 
 #### REPL
 
--- a/stdlib/LinearAlgebra/test/cholesky.jl
+++ b/stdlib/LinearAlgebra/test/cholesky.jl
@@ -39,7 +39,7 @@ end
     n1 = div(n, 2)
     n2 = 2*n1
 
-    Random.seed!(1234321)
+    Random.seed!(12343)
 
     areal = randn(n,n)/2
     aimg  = randn(n,n)/2
--- a/stdlib/LinearAlgebra/test/symmetric.jl
+++ b/stdlib/LinearAlgebra/test/symmetric.jl
@@ -4,7 +4,7 @@ module TestSymmetric
 
 using Test, LinearAlgebra, SparseArrays, Random
 
-Random.seed!(101)
+Random.seed!(1010)
 
 @testset "Pauli σ-matrices: $σ" for σ in map(Hermitian,
         Any[ [1 0; 0 1], [0 1; 1 0], [0 -im; im 0], [1 0; 0 -1] ])
--- a/stdlib/Random/src/normal.jl
+++ b/stdlib/Random/src/normal.jl
@@ -35,9 +35,11 @@ julia> randn(rng, ComplexF32, (2, 3))
   0.611224+1.56403im   0.355204-0.365563im  0.0905552+1.31012im
 ```
 """
-@inline function randn(rng::AbstractRNG=default_rng())
+@inline randn(rng::AbstractRNG=default_rng()) = _randn(rng, rand(rng, UInt52Raw()))
+
+@inline function _randn(rng::AbstractRNG, r::UInt64)
     @inbounds begin
-        r = rand(rng, UInt52())
+        r &= 0x000fffffffffffff
         rabs = Int64(r>>1) # One bit for the sign
         idx = rabs & 0xFF
         x = ifelse(r % Bool, -rabs, rabs)*wi[idx+1]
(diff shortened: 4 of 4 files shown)


KristofferC:
@nanosoldier `runtests(ALL; vs =":master")`

KristofferC:
I just want to see how much of the ecosystem tests this will break.

nanosoldier:
[Your package evaluation job](https://github.com/JuliaLang/julia/pull/35078#issuecomment-597901646) has completed - possible new issues were detected. A full report can be found [here](https://github.com/JuliaCI/NanosoldierReports/blob/9d278a7f03d4f12de0c6e5fdfca987ca5c109d82/pkgeval/by_hash/c235be7_vs_6732cb9/report.md). cc @maleadt

rfourquet:
Wow, when I had the idea for this change last week, I was surprised to not have thought about it earlier, and I just found a 5+ y.o. local branch with this exact same change :see_no_evil: (the performance improvement might have not been worthwhile at that time though?) 

Anyway, someone willing to review?

rfourquet:
Ok to merge?

StefanKarpinski:
Go for it!

KristofferC:
Would be good to open issues at the packages that started failing due to this change.

rfourquet:
> Would be good to open issues at the packages that started failing due to this change.

I'm surprised by this suggestion, and would shy away from it, I would be worried to be spamming. Is there precedent for that? I remember having fixed few repos before Julia 0.7 to adjust them to last time similar changes, but now the random streams are expected to be subject to change in minor versions and the package authors will anyway be notified either by reading the NEWS or by running their tests. I can do, but am reluctant.

StefanKarpinski:
If upgrading Julia breaks the tests, I'd imagine most package authors would want to be proactively notified about that, so I'm not sure that's terribly spammy. They can always ignore or close issues.

KristofferC:
> I'm surprised by this suggestion, and would shy away from it

Fine, but it will just end up with me filing these issues when I run PkgEval for the 1.5 release and have 63 extra regressions in the logs to go through.

rfourquet:
> but it will just end up with me filing these issues when I run PkgEval for the 1.5 release

Oh I didn't realize you were doing that, then sure I will do for these errors. But I'm afraid that in many cases this won't be handled when you run PkgEval for 1.5 and that you will have still some work :-/

KristofferC:
Sure, but at least there will be an issue there and I can quickly move on. 

People do tend to fix these issues though:

image

rfourquet:
> People do tend to fix these issues though:

Ok cool, I was just imagining many maintainners to not build Julia from sources and wait for prerelease builds before trying to fix the failures.

> there will be an issue there and I can quickly move on.

I will include "Julia 1.5" in the title, if this can help you find these issues. I might wait until after #29240 is merged in case this causes test failures in the same packages.

rfourquet:
I fixed a small bug, which makes `randn!` even slightly faster, but unfortunately it means I have to re-run PkgEval...

rfourquet:
@nanosoldier `runtests(ALL; vs =":master")`

nanosoldier:
[Your package evaluation job](https://github.com/JuliaLang/julia/pull/35078#issuecomment-618388558) has completed - possible new issues were detected. A full report can be found [here](https://github.com/JuliaCI/NanosoldierReports/blob/7de24e455342298cbef56826b5827f0d7640d2c1/pkgeval/by_hash/b89e35c_vs_098ef24/report.md). cc @maleadt

StefanKarpinski:
You definitely shouldn't have to patch things with each release. A best practices doc for RNGs would be good.

PR #35080, opened 2020-03-11
Fix MacOS Notarization

staticfloat:
We need to individually sign each Mach-O file independently, otherwise notarization bails.  Note that you must sign the overall `.app` after each individual file is signed.

I've also added more Entitlements. I've gone for a "gentle shotgun" approach, turning on everything that I think we could reasonably want.  I wait expectantly for someone to complain that they need Location access turned on for some package.  ;)

### DIFF
--- a/contrib/mac/app/Entitlements.plist
+++ b/contrib/mac/app/Entitlements.plist
@@ -4,5 +4,19 @@
 <dict>
 	<key>com.apple.security.automation.apple-events</key>
 	<true/>
+	<key>com.apple.security.cs.allow-dyld-environment-variables</key>
+	<true/>
+	<key>com.apple.security.cs.allow-jit</key>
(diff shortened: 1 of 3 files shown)


Keno:
Do we know what entitlements e.g. Python asks for?

staticfloat:
The python launcher itself asks for the following two:

```
<key>com.apple.security.cs.disable-library-validation</key>
<true/>
<key>com.apple.security.cs.disable-executable-page-protection</key>
<true/>
```

The first so that it can load unsigned libraries (which we'd need as well) and the second I assume for some kind of JIT behavior.

Keno:
All right, let's go with this for now then.

staticfloat:
This should be backported in case we ever do a 1.4.2

PR #35083, opened 2020-03-11
Build tarballs on Windows as well

musm:
Removes the conditional so that tarballs are created for windows when building with cygwin


musm:
superseded by https://github.com/JuliaLang/julia/pull/35098

PR #35086, opened 2020-03-12
use CodeGenOpt::None at optlevel<2, Default at 2, and Aggressive at >2

JeffBezanson:
This provides a little bit of speedup. I don't fully understand what these flags mean, but it seems to make sense to use `Default` at the default optimization level. Anyway I'm curious to see what the benchmarks say.

The difference in compile time between None and Default is quite large, with a much smaller difference between Default and Aggressive (which I assume is why we were using Aggressive most of the time). In fact using None for -O1 is even faster than using FastISel, so we might want to consider dropping FastISel entirely.

### DIFF
--- a/src/codegen.cpp
+++ b/src/codegen.cpp
@@ -7451,7 +7451,9 @@ extern "C" void jl_init_llvm(void)
 #ifdef DISABLE_OPT
         CodeGenOpt::None;
 #else
-        (jl_options.opt_level == 0 ? CodeGenOpt::None : CodeGenOpt::Aggressive);
+        (jl_options.opt_level < 2 ? CodeGenOpt::None :
+         jl_options.opt_level == 2 ? CodeGenOpt::Default :
+         CodeGenOpt::Aggressive);
 #endif
     jl_TargetMachine = TheTarget->createTargetMachine(
             TheTriple.getTriple(), TheCPU, FeaturesStr,


nanosoldier:
[Your benchmark job](https://github.com/JuliaLang/julia/pull/35086) has completed - possible performance regressions were detected. A full report can be found [here](https://github.com/JuliaCI/BaseBenchmarkReports/blob/14c9457d6ccc2f10dc62f08dbefa399eb019a71e/db42b58_vs_ecd58b3/report.md). cc @ararslan

vtjnash (approved):

vtjnash:
FWIW, looks like clang made a similar change recently https://reviews.llvm.org/D28409

Sacha0:
This change caused some runtime regressions in RAI's codebase between 1.4 and 1.5 at O1. For the moment we've happily resolved those regressions by patching Julia to pass `CodeGenOpt::Default` to LLVM at O1 (while leaving e.g. O1's pass set and ordering the same), and hope to ultimately resolve those regressions by migrating back to O2 (after the interesting compilation blowup at O2 that we've discussed elsewhere is addressed).

Out of curiosity, why depart from clang's selection of `CodeGenOpt::Less` for O1? :) Best!

JeffBezanson:
We should compare some latencies with `CodeGenOpt::Less`, and if it seems to be a better trade-off we can switch to it. (I assume `Less` also fixes your performance regressions?)

Sacha0:
> (I assume `Less` also fixes your performance regressions?)

Haven't tested, but will! :)

Sacha0:
A bit of testing indicates that using CodeGenOpt::Less at O1: a) mitigates some of the runtime regressions we observe with CodeGenOpt::None at O1; and b) provides most of the compilation time reduction we observe with CodeGenOpt::None at O1. That being the case, I will submit a pull request migrating O1 to CodeGenOpt::Less shortly, with the hope that it can find its way into 1.6 :). Thanks and best!

PR #35087, opened 2020-03-12
change 'e' to 'ℯ' within Doc for 'exp'

ghost:
as in https://github.com/JuliaLang/julia/blob/ecd58b3b9d0bf81cffb06a8a6fe3252d22d57f92/stdlib/LinearAlgebra/src/dense.jl#L538


fredrikekre:
It is already correct -- double backticks is math environment and rendered as LaTeX math, see https://docs.julialang.org/en/v1/base/math/#Base.exp-Tuple{Float64}.

PR #24299, opened 2017-10-23
Extensible help system for displaying MethodErrors

mbauman:
Fixes #7512.  This adds a hook for developers to add custom help messages when
`MethodError`s get displayed.  Feel free to bikeshed the naming of this method.

This is particularly helpful in cases of intentionally unimplemented methods.
I've also restructured the existing help messages in terms of this new system,
using the `info` command to display them (and a few extra messages, too):

```
julia> Int("1")
ERROR: MethodError: no method matching convert(::Type{Int64}, ::String)
INFO: Cannot `convert` an object of type String to an object of type Int64.
INFO: This may have arisen from a call to the constructor Int64(...), since type constructors fall back to convert methods.

Stacktrace:
 [1] Int64(::String) at ./sysimg.jl:114

julia> min([1,2,3])
ERROR: MethodError: no method matching min(::Array{Int64,1})
INFO: Use minimum() to find the smallest element of an array.

Closest candidates are:
  min(::AbstractArray{T1<:Real,N} where N, ::Real) where T1<:Real at deprecated.jl:55
  min(::AbstractArray{T1<:Real,N} where N, ::AbstractArray{T2<:Real,N} where N) where {T1<:Real, T2<:Real} at deprecated.jl:55
  min(::Any, ::Any) at operators.jl:404
  ...

julia> [1,2,3](1)
ERROR: MethodError: no method matching (::Array{Int64,1})(::Int64)
INFO: Use square brackets [] for indexing into arrays.

julia> [x for x in nothing]
ERROR: MethodError: no method matching start(::Void)
INFO: Objects of type Void are not iterable.

Closest candidates are:
  start(::SimpleVector) at essentials.jl:534
  start(::Base.MethodList) at reflection.jl:659
  start(::ExponentialBackOff) at error.jl:147
  ...
Stacktrace:
 [1] collect(::Base.Generator{Void,getfield(Main, Symbol("##1#2"))}) at ./array.jl:656
```

Feel free to bikeshed the name (`no_method_error_help`).

This is an [old branch/idea](https://github.com/JuliaLang/julia/issues/15781#issuecomment-209010397) that I had totally forgotten about; I was reminded by #24284, which would be very trivial in this setup.

PR #35094, opened 2020-03-12
Allow packages to provide custom hints for Exceptions

timholy:
Package authors may be able to predict likely user errors, and it can be nice to create an informative hint. For MethodErrors, we've done this repeatedly for Base; this PR makes it possible for packages to get in on the fun.

The procedure is described in the docstring for ~~Base.methoderror_hints~~`register_error_hint`.

### DIFF
--- a/base/errorshow.jl
+++ b/base/errorshow.jl
@@ -222,6 +222,9 @@ specific `MethodError`s with `push!(Base.methoderror_hints, hintmessage)`, where
         end
         return nothing    # use `nothing` to indicate that f, arg_types, kwargs didn't match
     end
+
+!!! compat "Julia 1.5"
+    Custom MethodError hints are available as of Julia 1.5.
 """
 const methoderror_hints = []
 # Note: Base should not use `methoderror_hints`, it should inline
--- a/base/errorshow.jl
+++ b/base/errorshow.jl
@@ -60,12 +60,8 @@ julia> module Hinter
        function __init__()
            register_error_hint(MethodError) do io, exc, argtypes, kwargs
                if exc.f == only_int
-                   if get(io, :color, false)
-                       print(io, "\nDid you mean to call ")
-                       printstyled(io, "`any_number`?", color=:light_magenta)
-                   else
-                       print(io, "\nDid you mean to call `any_number`?")
-                   end
+                    print(io, "\nDid you mean to call ")
+                    printstyled(io, "`any_number`?", color=:light_magenta)
                end
            end
        end
--- a/base/errorshow.jl
+++ b/base/errorshow.jl
@@ -29,6 +29,80 @@ ERROR: MyException: test exception
 """
 showerror(io::IO, ex) = show(io, ex)
 
+"""
+    register_error_hint(handler, exceptiontype)
+
+Register a "hinting" function `handler(io, exception)` that can
+suggest potential ways for users to circumvent errors.  `handler`
+should examine `exception` to see whether the conditions appropriate
+for a hint are met, and then generate output to `io`.
+Packages should call `register_error_hint` from within their
+[`__init__`](@ref) function.
+
+For specific exception types, `handler` is required to accept additional arguments:
+
+- `MethodError`: provide `handler(io, exc::MethodError, argtypes, kwargs)`,
+  which splits the combined arguments into positional and keyword arguments.
+
+When issuing a hint, the output should typically start with `\n`.
+
+If you define custom exception types, your `showerror` method can
+support hints by calling [`show_error_hints`](@ref).
+
+# Example
+
+```jldoctest
+julia> module Hinter
+
+       only_int(x::Int)      = 1
+       any_number(x::Number) = 2
+
+       function __init__()
+           register_error_hint(MethodError) do io, exc, argtypes, kwargs
+               if exc.f == only_int
+                   if get(io, :color, false)
+                       print(io, "\nDid you mean to call ")
+                       printstyled(io, "`any_number`?", color=:light_magenta)
+                   else
+                       print(io, "\nDid you mean to call `any_number`?")
+                   end
+               end
+           end
+       end
+
+       end
+Main.Hinter
+
+julia> Hinter.only_int(1.0)
+ERROR: MethodError: no method matching only_int(::Float64)
+Did you mean to call `any_number`?
+Closest candidates are:
+[...]
+```
+
+!!! compat "Julia 1.5"
+    Custom error hints are available as of Julia 1.5.
+"""
+function register_error_hint(handler, exct::Type)
+    list = get!(()->[], _hint_handlers, exct)
+    push!(list, handler)
+    return nothing
+end
+
+const _hint_handlers = IdDict{Type,Vector{Any}}()
+
+function show_error_hints(io, ex, args...)
+    hinters = get!(()->[], _hint_handlers, typeof(ex))
+    for handler in hinters
+        try
+            Base.invokelatest(handler, io, ex, args...)
+        catch err
+            tn = typeof(handler).name
+            @error "Hint-handler $handler for $(typeof(ex)) in $(tn.module) caused an error"
+        end
+    end
+end
+
 function showerror(io::IO, ex::BoundsError)
     print(io, "BoundsError")
     if isdefined(ex, :a)
@@ -45,6 +119,7 @@ function showerror(io::IO, ex::BoundsError)
             print(io, ']')
         end
     end
+    show_error_hints(io, ex)
 end
 
 function showerror(io::IO, ex::TypeError)
@@ -68,6 +143,7 @@ function showerror(io::IO, ex::TypeError)
         end
         print(io, ctx, ", expected ", ex.expected, ", got ", targs...)
     end
+    show_error_hints(io, ex)
 end
 
 function showerror(io::IO, ex, bt; backtrace=true)
@@ -106,6 +182,7 @@ function showerror(io::IO, ex::DomainError)
     if isdefined(ex, :msg)
         print(io, ":\n", ex.msg)
     end
+    show_error_hints(io, ex)
     nothing
 end
 
@@ -161,6 +238,7 @@ function showerror(io::IO, ex::InexactError)
     print(io, "InexactError: ", ex.func, '(')
     nameof(ex.T) === ex.func || print(io, ex.T, ", ")
     print(io, ex.val, ')')
+    show_error_hints(io, ex)
 end
 
 typesof(args...) = Tuple{Any[ Core.Typeof(a) for a in args ]...}
@@ -208,32 +286,6 @@ function show_convert_error(io::IO, ex::MethodError, @nospecialize(arg_types_par
     end
 end
 
-"""
-    methoderror_hints
-
-Packages can provide hints about appropriate corrective actions for
-specific `MethodError`s with `push!(Base.methoderror_hints, hintmessage)`, where
-`hintmessage` should be a function like
-
-```julia
-function hintmessage(f, arg_types, kwargs)
-    # Test to see whether the call matches the specific pattern for this message
-    if f === myfunc && length(arg_types) == 1 && arg_types[1] <: SomeType
-        return "`myfunc(::SomeType)` is not defined, did you mean to call `otherfunc`?"
-    end
-    return nothing    # use `nothing` to indicate that f, arg_types, kwargs didn't match
-end
-```
-
-Packages should perform the `push!` onto `Base.methoderror_hints` from
-their `__init__` function.
-
-!!! compat "Julia 1.5"
-    Custom MethodError hints are available as of Julia 1.5.
-"""
-const methoderror_hints = []
-# Note: Base should not use `methoderror_hints`, it should inline
-# custom hints below to avoid the slight performance hit from `invokelatest`.
 function showerror(io::IO, ex::MethodError)
     # ex.args is a tuple type if it was thrown from `invoke` and is
     # a tuple of the arguments otherwise.
@@ -337,12 +389,7 @@ function showerror(io::IO, ex::MethodError)
                       "\nYou can convert to a column vector with the vec() function.")
         end
     end
-    for predicate in methoderror_hints
-        msg = Base.invokelatest(predicate, f, arg_types_param, kwargs)
-        if msg !== nothing
-            print(io, '\n', msg)
-        end
-    end
+    show_error_hints(io, ex, arg_types_param, kwargs)
     try
         show_method_candidates(io, ex, kwargs)
     catch ex
--- a/base/errorshow.jl
+++ b/base/errorshow.jl
@@ -35,23 +35,23 @@ showerror(io::IO, ex) = show(io, ex)
 Register a "hinting" function `handler(io, exception)` that can
 suggest potential ways for users to circumvent errors.  `handler`
 should examine `exception` to see whether the conditions appropriate
-for a hint are met, and then generate output to `io`.
+for a hint are met, and if so generate output to `io`.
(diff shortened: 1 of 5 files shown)


johnnychen94:
From a package maintainer's view, adding this feature into codebase means to maintain two branches: one for julia 1.0 and one for julia >= 1.5. So it's likely that nobody is using it in other packages until the next LTS version release.

kimikage:
> he'll need to maintain two branches

If you don't intend to backport the hints to older versions of the package, I don't think you need to create branches.
```julia
if VERSION >= v"1.5"
    push!(Base.methoderror_hints, recommend_oneunit)
end
```

However, I don't really like the interface using an array. Is that unavoidable for performance?

johnnychen94:
> I don't really like the interface using an array. 

I think a shared registry is unavoidable because we want downstream packages to reuse hints defined by upstream packages. For example, let `Images.jl` reuse the hints defined by `Colors.jl`.

kimikage:
What exactly does "reuse" mean? I don't think downstream hints should depend on the upstream hints. :confused:

johnnychen94:
We'd better not redefine hints like this in `Colors`/`ColorVectorSpace`:

`For element-wise addition, use broadcasting with dot syntax: array .+ scalar`

```julia
# julia nightly build on https://github.com/JuliaLang/julia/pull/34642
julia> x = rand(4, 4);

julia> x + 2
ERROR: MethodError: no method matching +(::Array{Float64,2}, ::Int64)
For element-wise addition, use broadcasting with dot syntax: array .+ scalar
Closest candidates are:
  +(::Any, ::Any, ::Any, ::Any...) at operators.jl:529
  +(::Complex{Bool}, ::Real) at complex.jl:301
  +(::Missing, ::Number) at missing.jl:115
  ...
Stacktrace:
```

kimikage:
I don't understand the reason @timholy used the array and `for` loops instead of the multiple dispatch. I don't think we need the performance to show error (i.e. not warning) messages.

StefanKarpinski:
@kimikage, How frequently do you expect your programs to generate errors?

kimikage:
> How frequently do you expect your programs to generate errors?

It depends on what we assume. I'm not good at English, so it will take at least 5 seconds to understand an error message. So the frequency is less than 0.2 Hz. :smile:

Improving the performance of programs which won't work only makes sense in the cases of using CI systems and so on, because they won't work.

StefanKarpinski:
My point is that worrying about performance of generating error messages is unnecessary.

timholy:
I don't think it makes sense to do this by multiple dispatch. Catching the error is handled by the runtime, so the exact dispatch of which method threw the error is not going to be compiled into some specialized function. As a consequence, it's going to have to use runtime dispatch. Since all such handlers would have to be methods of the same function, and because there's probably little clear specialization by subtyping, the method table will essentially be a linear list. Consequently `jl_lookup_generic` is just going to iterate over every item on the list anyway, just like this implementation. Moreover, this implementation avoids the risk of ambiguities (and it sucks if your error-handler throws an error), and can indeed even be more optimal if necessary because it can short-circuit full subtyping.

timholy:
For reference, my test handler took 50ns to run. Imagine there are 1000 such handlers (which I don't think is out of the realm of possibility); if each takes 50ns, then the total run time is 50us. That's still way below the threshold for caring. We could have 100ns/handler and accommodate up to 10^6 such handlers before getting to 0.1s, which I'd say is the point at which I'd start caring about performance.

timholy:
@johnnychen94, just to be clear, when one handler issues a message, it doesn't block others from doing so too. One can imagine cases where that would be undesirable, but on balance I think it's the best choice if we're going to keep this simple. (Otherwise we'd need to develop a whole protocol for resolution, and that starts to smell overengineered.)

johnnychen94:
> just to be clear, when one handler issues a message, it doesn't block others from doing so too.

True, my expression wasn't very accurate. I meant to say "clear the order of displaying from nearest(downstream package) to farthest(e.g., Base)" :P

kimikage:
I simply care that `push!`ing to a global array is not a common extension interface on Julia. :smile:

timholy:
Take the `one`/`oneunit` example; rather than omitting `one(::HasNoOne)`, we *could* define it to throw a specific error, just like we could define a `x::Array + y::Number` method that throws an error. But there are circumstances where this causes problems: it might blocks a less-specific method that might, actually, be appropriate (as would happen for `getindex`), or it might cause ambiguities, or... So in general we've moved away from this kind of scheme.

This is a circumstance where we're definitely not building real code, so normal rules don't really apply. You'll note the [handlers already built into Base](https://github.com/JuliaLang/julia/blob/3935491dc5028f23587ca8df6d4e0de75cf50e53/base/errorshow.jl#L211-L319) work essentially by trying each one.

timholy:
> True, my expression wasn't very accurate. I meant to say "sort the order of displaying from nearest(downstream package) to farthest(e.g., Base)" :P

You could achieve that with `pushfirst!`. Packages have at least that level of control over ordering.

johnnychen94:
> You could achieve that with `pushfirst!`

Would this be affected by precompilation and/or package loading order?

timholy:
Loading order, yes. Not by precompilation order (you should `push!` these from your package's `__init__` function, perhaps I should mention this in the docstring).

kimikage:
This is just cosmetic, but I think it's better to quote the example with \```julia...```. :smile:

base/errorshow.jl
```diff
@@ -208,6 +208,32 @@ function show_convert_error(io::IO, ex::MethodError, @nospecialize(arg_types_par
...
+
+Packages should perform the `push!` onto `Base.methoderror_hints` from
+their `__init__` function.
+
+!!! compat "Julia 1.5"
+    Custom MethodError hints are available as of Julia 1.5.
+"""
+const methoderror_hints = []
+# Note: Base should not use `methoderror_hints`, it should inline
+# custom hints below to avoid the slight performance hit from `invokelatest`.
```
KristofferC: Is the overhead from `invokelatest` significant here?

timholy: Yes (in a relative sense, i.e., it's a large fraction of total cost), though acceptable IMO. I do think it's viable to have Base use `methoderror_hints` also, but since Base doesn't actually need to do so I thought it would be better to encourage continuing the current trajectory.

Another issue not yet discussed is where there are any circumstances where the callback needs to receive the argument values rather than just the types. The values are not always guaranteed to be available (see the comment re `invoke` at the top of `showerror`), whereas the types are.

base/errorshow.jl
```diff
@@ -208,6 +208,32 @@ function show_convert_error(io::IO, ex::MethodError, @nospecialize(arg_types_par
     end
 end
 
+"""
+    methoderror_hints
+
+Packages can provide hints about appropriate corrective actions for
+specific `MethodError`s with `push!(Base.methoderror_hints, hintmessage)`, where
+`hintmessage` should be a function like
```
tkf: Does it makes sense to expose this API as a function, like `pushdisplay` and `atreplinit`? It can be used as something like

```julia
register_methoderror_hint() do f, arg_types, kwargs
    ...
end
```

or just `register_methoderror_hint(hintmessage)`. It seems to be helpful to disentangle the API from the internal representation so that future improvements are easier.

base/errorshow.jl
```diff
@@ -208,6 +208,32 @@ function show_convert_error(io::IO, ex::MethodError, @nospecialize(arg_types_par
...
+
+```julia
+function hintmessage(f, arg_types, kwargs)
```
tkf: How about passing an `IO` object here so that the hint message can use, e.g., colors?  It can be either the `io` which `showerror` is printing to or a separate buffer so that `showerror` can do appropriate post-processing (e.g., add a newline only if needed).  Alternatively, this can be done by passing a `context` object like `sprint`.

base/errorshow.jl
timholy: I thought about that. I decided against it (but am happy to reconsider) because it would be a fair number of exports If all our exception types end up admitting hints.

timholy: Hmm, we could do it with just one: `register_error_hint(handler, MethodError)`. We'd have to document the expected arguments of `handler` for each error type separately. Thoughts?

tkf: That makes sense.  Then how about

```julia
register_error_hint(MethodError) do f, arg_types, kwargs
    ...
end
```

?

tkf: Ah, race condition.

kimikage: For example, the inability to customize `InexactError` messages is one of the problems in FixedPointNumbers.jl. However, I think it is a problem of `InexactError` and should not be generalized here. Under the principle that type piracy should be prohibited, I think only `MethodError` is the special case.

tkf: I don't think there is any type piracy going on here. I think a similar hook system can be provided for other exception types if necessary, without any type piracy.

(Type piracy occurs in the method dispatch because there is only one method you can dispatch to. Hook systems like this allow multiple matches. That's why I think type piracy is irrelevant here.)

kimikage: > I don't think there is any type piracy going on here. 

I agree with you, because the type piracy should be prohibited.
In the principle, module `Foo` have the right to control the errors which depend only on the `FooType` defined in the the module `Foo`. Thus, if you don't like the default error message, `Foo` can define a new `Exception` type as needed to show its own error message. However, `MethodError` cannot be controlled by `Foo` for now, even if it depends only on `FooType`.

kimikage:
What about adding `[methoderror]` section to Project.toml?

(Of course, this is a joke.)

base/errorshow.jl
tkf: Yes, for error _messages_, I agree that's the right way to do it. This API is about _hints_. It is for the situation multiple third parties can provide useful information to the user. Though I cannot find other exception types that are useful to have such entry-point ATM.

timholy:
Really great suggestions, @tkf! In addition, I wrapped the handler call in a `try/catch`, and I added support for hints to a few other exception types.

base/errorshow.jl
timholy: OK, in the new version I pass the exception object as well as some processed arguments, so I'm not so worried we'd want to change the API anymore.

base/errorshow.jl
```diff
@@ -29,6 +29,80 @@ ERROR: MyException: test exception
...
+
+       function __init__()
+           register_error_hint(MethodError) do io, exc, argtypes, kwargs
+               if exc.f == only_int
+                   if get(io, :color, false)
+                       print(io, "\nDid you mean to call ")
+                       printstyled(io, "`any_number`?", color=:light_magenta)
+                   else
+                       print(io, "\nDid you mean to call `any_number`?")
+                   end
```
tkf: Do you need explicit `get(io, :color, false)`?  Why not just

```suggestion
                   print(io, "\nDid you mean to call ")
                   printstyled(io, "`any_number`?", color=:light_magenta)
```

? `printstyled` is `IOContext`-aware, right?

```julia
julia> sprint() do io
           printstyled(io, "hello"; color=:red)
       end
"hello"

julia> sprint(context=stdout) do io
           printstyled(io, "hello"; color=:red)
       end
"\e[31mhello\e[39m"
```

base/errorshow.jl
```diff
@@ -29,6 +29,80 @@ ERROR: MyException: test exception
...
+for a hint are met, and then generate output to `io`.
+Packages should call `register_error_hint` from within their
+[`__init__`](@ref) function.
+
+For specific exception types, `handler` is required to accept additional arguments:
+
+- `MethodError`: provide `handler(io, exc::MethodError, argtypes, kwargs)`,
+  which splits the combined arguments into positional and keyword arguments.
+
+When issuing a hint, the output should typically start with `\n`.
```
tkf: I think less error-prone API may be to automate inserting `\n`.  I guess it can be implemented like this

```julia
buf = IOBuffer()
ioctx = IOContext(buf, :color => ..., :displaysize => ...)  # is there a good way to do this?
for handler in hinters
    seek(buf, 0)
    try
        Base.invokelatest(handler, ioctx, ex, args...)
    catch err
        ...
        continue
    end
    if position(buf) != 0
        # maybe check newlines in buf?
        println(io)
        write(io, view(buf.data, 1:buf.size))
    end
end
```

timholy: I think that's too ugly, and fragile if more than one hint fires.

tkf: I believe the `IOBuffer`-based solution is more robust as you don't have to rely on the hint authors to not forget printing `\n`.  It supports multiple hints.

timholy: Not if they're issued from the same function.

timholy:
I'll hold off a bit longer (there has been a lot of very useful feedback that has improved this considerably!), but from my standpoint this is good to merge. The only part I'm unhappy with is #35114, but that's an issue for another day.

base/errorshow.jl
```diff
@@ -29,6 +29,87 @@ ERROR: MyException: test exception
...
+julia> module Hinter
+
+       only_int(x::Int)      = 1
+       any_number(x::Number) = 2
+
+       function __init__()
+           register_error_hint(MethodError) do io, exc, argtypes, kwargs
+               if exc.f == only_int
+                    # Color is not necessary, this is just to show it's possible.
+                    print(io, "\\nDid you mean to call ")
```
mbauman: should we recommend folks use `@info`?

mbauman:
Holy cow, I missed this and it looks awesome.  This completely supersedes https://github.com/JuliaLang/julia/pull/24299 (and I like just using an array of functions much much better than leaning so heavily on multiple dispatch)... excepting there are a few more ideas about the sorts of things to be registered there.

JeffBezanson:
I'm requesting triage here since this adds new exports. I admit, I greatly dislike adding things to Base. The functionality is good, I'm mostly thinking about where it goes. It's too bad how much UI-ish stuff is in Base; in fact I think that probably extends to the whole `showerror` function as well. All of that stuff seems like it belongs in the REPL or another package like InteractiveUtils.

Global state is also always a problem. I think this could have issues with building system images; if you load packages that add hints and then save a system image, the handlers will be saved and then possibly re-added on startup. Emptying the dictionary in `Base.__init__` might be enough to fix this.

base/errorshow.jl
tkf: It's printing to a buffer `io` so I don't think we can replace it with `@info`.

(Or you are thinking to temporarily create a "buffer" logger to collect messages within a dynamic scope? It might work but I feel it's too fancy to do during error printing.)

tkf:
> Emptying the dictionary in `Base.__init__` might be enough to fix this.

I guess an alternative idea may be to register a function via `atexit` to clean it up and avoid serializing the callbacks into the sysimage in the first place.

timholy:
> I'm requesting triage here since this adds new exports.

Sounds fine. Let me know if you want changes or just go ahead and revert if necessary.

c42f:
> It's too bad how much UI-ish stuff is in Base; in fact I think that probably extends to the whole `showerror` function as well. All of that stuff seems like it belongs in the REPL or another package like InteractiveUtils.

For what it's worth I'd love if these functions were moved out of `Base`. I've been modifying `Base.showerror` recently to add some more heuristics (#35243), and it seemed quite wrong to be adding what are basically REPL-related UI hacks to `Base`.

JeffBezanson:
From triage: move the exports to the `Experimental` submodule.

timholy:
I will move it. Do you see a good solution to the global state problem? I am happy to work towards an alternative.

mbauman:
In our short discussion on triage, Jeff (and others) talked about one reason to simply move it is because we don't have an obvious alternative but it sure is helpful.

PR #35098, opened 2020-03-12
Separate stage for windows exe installer and build tarballs on Windows

musm:
Two changes:
1. Build tarballs on Windows platforms, as is done for the other supported platforms.
2. Separate out the windows  exe installer into a new stage

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -465,16 +465,15 @@ endif
 ifeq ($(OS), Darwin)
 	-cat $(JULIAHOME)/contrib/mac/startup.jl >> $(DESTDIR)$(prefix)/etc/julia/startup.jl
 endif
-
 ifeq ($(OS), WINNT)
 	cd $(BUILDROOT)/julia-$(JULIA_COMMIT)/bin && rm -f llvm* llc.exe lli.exe opt.exe LTO.dll bugpoint.exe macho-dump.exe
+endif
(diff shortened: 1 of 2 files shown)


musm:
Needs https://github.com/JuliaCI/julia-buildbot/pull/146/files to pass on windows

musm:
g2g @staticfloat?

staticfloat:
Yep, lgtm

Makefile
```diff
@@ -465,16 +465,15 @@ endif
 ifeq ($(OS), Darwin)
 	-cat $(JULIAHOME)/contrib/mac/startup.jl >> $(DESTDIR)$(prefix)/etc/julia/startup.jl
 endif
-
 ifeq ($(OS), WINNT)
 	cd $(BUILDROOT)/julia-$(JULIA_COMMIT)/bin && rm -f llvm* llc.exe lli.exe opt.exe LTO.dll bugpoint.exe macho-dump.exe
+endif
+	cd $(BUILDROOT) && $(TAR) zcvf $(JULIA_BINARYDIST_FILENAME).tar.gz julia-$(JULIA_COMMIT)
 
+exe:
```
musm: Looks like I forgot `exe: binary-dist`

musm: Do we really want to slow the build process down by adding this? FWIW copying all the files back on my computer takes 2-5 minutes

staticfloat: Yes, we do.  It's better to haev a slow build to create something we know is correct, then for there to be a possibility to have old files shipped.

PR #35101, opened 2020-03-13
Tests for powm error

kshyatt:
This little throw and the pass-through for `LowerTriangular` are not tested.

### DIFF
--- a/stdlib/LinearAlgebra/src/triangular.jl
+++ b/stdlib/LinearAlgebra/src/triangular.jl
@@ -2169,7 +2169,7 @@ end
 #     34(3), (2013) 1341–1360.
 function powm!(A0::UpperTriangular{<:BlasFloat}, p::Real)
     if abs(p) >= 1
-        ArgumentError("p must be a real number in (-1,1), got $p")
+        throw(ArgumentError("p must be a real number in (-1,1), got $p"))
     end
 
     normA0 = opnorm(A0, 1)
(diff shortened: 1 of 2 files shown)


dkarrasch:
Wow, this is an amazing example of what testing is good for. The tests currently fail because `powm[!]` is not exported. BUT, adding a `LinearAlgebra` specifier still keeps both tests failing, because
1. https://github.com/JuliaLang/julia/blob/3935491dc5028f23587ca8df6d4e0de75cf50e53/stdlib/LinearAlgebra/src/triangular.jl#L2172 requires a `throw` before the `ArgumentError`, otherwise it simply doesn't throw at all;
2. https://github.com/JuliaLang/julia/blob/3935491dc5028f23587ca8df6d4e0de75cf50e53/stdlib/LinearAlgebra/src/triangular.jl#L2231 should be
```julia
powm(A::LowerTriangular, p::Real) = copy(transpose(powm!(copy(transpose(A)), p::Real)))
```
because otherwise it throws saying there's no method matching `powm(::UpperTriangular{Complex{Float64},Array{Complex{Float64},2}}, ::Float64)`. Sure enough, we don't have such a method, only one with `!`.

@kshyatt Would you be willing to expand this PR to a "bugfix" one, or do you prefer to to do that elsewhere?

kshyatt:
I'll expand it. Shows what you get when you let CI run the tests for you.

PR #35103, opened 2020-03-13
Corrections to `timedwait` methods

omus:
Mainly I wanted to update the `timedwait` function to be able to take `Real` instead of just `Float64`. During this process I found that the `timedwait` method which takes `Period` was missing the `pollint` keyword and I made some adjustments there as well.

The `timedwait` function currently doesn't have any tests so I'll probably also do that as part of this PR.

### DIFF
--- a/base/asyncevent.jl
+++ b/base/asyncevent.jl
@@ -255,14 +255,14 @@ function Timer(cb::Function, timeout::Real; interval::Real=0.0)
 end
 
 """
-    timedwait(testcb::Function, secs::Float64; pollint::Float64=0.1)
+    timedwait(testcb::Function, secs::Real; pollint::Real=0.1)
 
 Waits until `testcb` returns `true` or for `secs` seconds, whichever is earlier.
 `testcb` is polled every `pollint` seconds.
 
 Returns :ok, :timed_out, or :error
 """
-function timedwait(testcb::Function, secs::Float64; pollint::Float64=0.1)
+function timedwait(testcb::Function, secs::Real; pollint::Real=0.1)
     pollint > 0 || throw(ArgumentError("cannot set pollint to $pollint seconds"))
     start = time_ns()
     nsecs = 1e9 * secs
--- a/base/asyncevent.jl
+++ b/base/asyncevent.jl
@@ -255,24 +255,24 @@ function Timer(cb::Function, timeout::Real; interval::Real=0.0)
 end
 
 """
-    timedwait(testcb::Function, secs::Real; pollint::Real=0.1)
+    timedwait(testcb::Function, timeout::Real; pollint::Real=0.1)
 
-Waits until `testcb` returns `true` or for `secs` seconds, whichever is earlier.
-`testcb` is polled every `pollint` seconds. The minimum duration for `secs` and `pollint` is
-1 millisecond or `0.001`.
+Waits until `testcb` returns `true` or for `timeout` seconds, whichever is earlier.
+`testcb` is polled every `pollint` seconds. The minimum duration for `timeout` and `pollint`
+is 1 millisecond or `0.001`.
 
 Returns :ok or :timed_out
 """
-function timedwait(testcb::Function, secs::Real; pollint::Real=0.1)
+function timedwait(testcb::Function, timeout::Real; pollint::Real=0.1)
     pollint >= 1e-3 || throw(ArgumentError("pollint must be ≥ 1 millisecond"))
     start = time_ns()
-    nsecs = 1e9 * secs
+    ns_timeout = 1e9 * timeout
     done = Channel(1)
     function timercb(aw)
         try
             if testcb()
                 put!(done, (:ok, nothing))
-            elseif (time_ns() - start) > nsecs
+            elseif (time_ns() - start) > ns_timeout
                 put!(done, (:timed_out, nothing))
             end
         catch e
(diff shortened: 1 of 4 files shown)


JeffBezanson:
Bump

omus:
I'll add some tests today

omus:
Alright, tests have now been added. One thing to note is that `timedwait` now rethrows exceptions raised by the callback function instead of returning `:error`. The rational behind this change is that if the callback function would error on the first call then an exception would be raised instead of returning `:error`. This approach is now consistent and should provide details as to what the exception is.

Also note I've removed the `seconds` function as the changes I made probably need additional discussion. A quick overview of the issue was that `toms` truncates units smaller than milliseconds instead of fractional number.

stdlib/Dates/src/types.jl
```diff
@@ -405,11 +405,15 @@ Base.hash(x::Time, h::UInt) =
...
-timedwait(testcb::Function, time::Period) = timedwait(testcb, toms(time) / 1000)
+Base.sleep(duration::Period) = sleep(toms(duration) / 1000)
+
+function Base.Timer(delay::Period; interval::Period=Second(0))
+    Timer(toms(delay) / 1000, interval=toms(interval) / 1000)
+end
+
+function Base.timedwait(testcb::Function, timeout::Period; pollint::Period=Millisecond(100))
+    timedwait(testcb, toms(timeout) / 1000, pollint=toms(pollint) / 1000)
+end
```
omus, line 416: I'll call out that `toms` truncates to milliseconds which means we're losing precision if a user passes in say `Nanosecond(1)` for either the `timeout` or `pollint`. However, since `timedwait` can only actually support millseconds this isn't really an issue.

This is one of those cases where two wrongs make a right ;)

base/asyncevent.jl
```diff
@@ -255,18 +255,19 @@ function Timer(cb::Function, timeout::Real; interval::Real=0.0)
...
+Returns :ok or :timed_out
 """
-function timedwait(testcb::Function, secs::Float64; pollint::Float64=0.1)
-    pollint > 0 || throw(ArgumentError("cannot set pollint to $pollint seconds"))
+function timedwait(testcb::Function, secs::Real; pollint::Real=0.1)
+    pollint >= 1e-3 || throw(ArgumentError("pollint must be ≥ 1 millisecond"))
     start = time_ns()
     nsecs = 1e9 * secs
-    done = Channel(1)
+    done = Channel{Union{Symbol,Exception}}(1)
```
JeffBezanson, line 270:
```suggestion
    done = Channel(1)
```

base/asyncevent.jl
```diff
@@ -284,7 +285,8 @@ function timedwait(testcb::Function, secs::Float64; pollint::Float64=0.1)
 
     if !testcb()
         t = Timer(timercb, pollint, interval = pollint)
-        ret = fetch(done)::Symbol
+        ret = fetch(done)
+        ret isa Exception && rethrow(ret)
```
JeffBezanson, line 289: I don't think we should rely on the type of `e`; can use a boolean flag instead.

omus: How about if we store a `Tuple{Symbol, Union{Nothing, Exception}}` in the channel? Then this code would look like:
```julia
ret, e = fetch(done)
ret === :error && rethrow(e)
```

omus:
Looks like I need to increase my timing tolerance:
```
Test Failed at /Users/julia/buildbot/worker-tabularasa/tester_macos64/build/share/julia/test/channels.jl:268
  Expression: ≈(duration, 1, atol = 0.05)
   Evaluated: 1.182448603 ≈ 1 (atol=0.05)
```

omus:
Tolerance is still to small:
```
Test Failed at /Users/julia/buildbot/worker-tabularasa/tester_macos64/build/share/julia/test/channels.jl:265
  Expression: ≈(duration, 1, atol = 0.2)
   Evaluated: 1.229485612 ≈ 1 (atol=0.2)
Error in testset channels:
Test Failed at /Users/julia/buildbot/worker-tabularasa/tester_macos64/build/share/julia/test/channels.jl:268
  Expression: ≈(duration, 1, atol = 0.2)
   Evaluated: 1.207294196 ≈ 1 (atol=0.2)
```

base/asyncevent.jl
```diff
@@ -284,7 +285,8 @@ function timedwait(testcb::Function, secs::Float64; pollint::Float64=0.1)
 
     if !testcb()
         t = Timer(timercb, pollint, interval = pollint)
-        ret = fetch(done)::Symbol
+        ret, e = fetch(done)
+        ret === :error && rethrow(e)
```
JeffBezanson, line 289: `rethrow` doesn't work outside a catch block (it's throwing `ERROR: rethrow(exc) not allowed outside a catch block`) so you have to use `throw`.

omus: Bad mistake on my part. Unfortunately this means that the exception thrown won't have the original stacktrace

omus: Only if the callback raises an exception on the first call will generate the original stack trace. Exceptions raised after will not have the original stack trace. This seems better than silencing all exceptions from the callback.

omus: Alternatively I can have `timedwait` return `:error` again and just capture exceptions from the first `testcb` call

iamed2: You could capture the stack trace and store it in a CapturedException?

omus: I've updated the code to always return a `CapturedException`

omus:
CI failures look network related

omus:
Rebased changes

Issue #34820, opened 2020-02-19
funcionloc should return line number of method sign, not line of first statement in func


davidanthoff:
It seems that `functionloc` returns the line of the first statement in a method, not the line of the method signature (i.e. the `function foo(` part).

Could this be changed, or a function be added that returns the line of the definition? Or is there a workaround right now that we could use to reliably always return the line for the `function foo` part?

We use this in the language server goto definition stuff, and getting the line of the first statement is kind of messing up the UI for us.

JeffBezanson:
Yeah, I think it would make sense to put this in the `file` and `line` fields of a `Method` object, leaving the source locations as they are.

davidanthoff:
Is there any workaround on currently shipping versions that would allow us to get the right line?

JeffBezanson:
I don't believe so; the information is just not available. But backtracking from the first line number to a `function` keyword should work pretty well (with a separate case, of course, for 1-line definitions, so there would be a bit of fiddling to detect that).

davidanthoff:
Ah, ok, I think then we'll just have to wait... The process that is extracting this information right now is super slow already, if we started to load source files to look for the location of the keyword, that would be even worse.

fredrikekre:
See also #1419, #4971. I assume there is some function in the debugger stack that gives you this information?

timholy:
See https://github.com/timholy/CodeTracking.jl

davidanthoff:
@timholy How would I use CodeTracking.jl for this problem? `whereis` also returns the location of the first line of the body, not the location of the definition.

KristofferC:
Use `CodeTracking.definition`. It does exactly the backtracking as described in https://github.com/JuliaLang/julia/issues/34820#issuecomment-588510808.

PR #35105, opened 2020-03-13
Add file and line number reporting for some lowering errors

### LINKED ISSUE #35128: Line numbers incorrectly reported for some syntax errors
In code like this:
```
if p
    # lots of code here
    #...

    a{b;c}

    # lots more code here
    #...
end
```
the unexpected semicolon is reported as being on the line with `if p`, not the line with `a{b;c}`.
This issue occurs only with errors that are detected during the eval phase (after macros are evaluated), not the parsing phase (before macros are evaluated). So in this case:
```
if p
    a b
end
```
the line with the extra token `b` is correctly identified as line 2. However, an expression like `true = false` would not be identified correctly.
The issue also occurs for errors inside a function definition, or a for loop. However, it does not appear to occur for errors directly inside module declarations, although errors in an if statement in a module declaration would be reported as being on the if statement. I haven't checked all possible constructs with multiple expressions, there may well be more problems.

c42f:
As part of looking into https://github.com/JuliaLang/julia/issues/34820 I added file and line number reporting for desugaring errors and cleaned up some problems with `jl_filename` leaking between `eval`s. I thought these could be useful in their own right (independent of finishing the fix for `functionloc`).

The approach using the dynamically scoped `*current-desugar-loc*` seemed to be the least disruptive way of making this work in lowering.

Fixes #35128

### DIFF
--- a/src/julia-syntax.scm
+++ b/src/julia-syntax.scm
@@ -1837,7 +1837,7 @@
       `(= ,lhs ,rhs)))
 
 (define (expand-forms e)
-  (if (or (atom? e) (memq (car e) '(quote inert top core globalref outerref line module toplevel ssavalue null true false meta using import export)))
+  (if (or (atom? e) (memq (car e) '(quote inert top core globalref outerref module toplevel ssavalue null true false meta using import export)))
       e
       (let ((ex (get expand-table (car e) #f)))
         (if ex
@@ -2334,6 +2334,11 @@
           (= ,r ,(expand-forms (cadr e)))
           (gc_preserve_end ,s)
           ,r)))
+
+    'line
+    (lambda (e)
+      (set! *current-desugar-loc* e)
+      e)
     ))
 
 (define (has-return? e)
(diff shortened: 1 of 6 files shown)


JeffBezanson:
Conveniently fixes #35128 :)

src/julia-syntax.scm
```diff
@@ -1837,13 +1837,19 @@
...
-      (let ((ex (get expand-table (car e) #f)))
-        (if ex
-            (ex e)
-            (cons (car e)
-                  (map expand-forms (cdr e)))))))
+  (cond ((or (atom? e)
+             (memq (car e) '(quote inert top core globalref outerref module toplevel ssavalue
+                             null true false meta using import export)))
+         e)
+        ((eq? (car e) 'line)
```
JeffBezanson: This case can be added to `expand-table` below instead.

JeffBezanson (approved):

src/julia-syntax.scm
c42f: Nice, I'll do that.

Issue #18074, opened 2016-08-17
Setting env. variables like JULIA_NUM_THREADS for remote workers


oschulz:
There currently doesn't seem to be a method to control/set/forward environment variables for remote workers, e.g. workers started with

```
addprocs([host1, ...])
```

One important variable that users may want to set is `JULIA_NUM_THREADS`. But in more complex real-world situation, one may also need to set (resp. forward) things like PATH, LD_LIBRARY_PATH (for custom software packages) and other (site- and application-dependent) variables.

Maybe addprocs could get an `env` kwarg or so?

ViralBShah:
cc @amitmurthy

xgdgsc:
Also hoping this could be added so I only have one JULIA_DEPOT_PATH JULIA_PROJECT shared on nfs between nodes.

mgkuhn:
Julia 1.6 added a new `env=["JULIA_DEPOT_PATH"=>"/depot", ...]` keyword argument to [Distributed.addprocs](https://docs.julialang.org/en/v1/stdlib/Distributed/#Distributed.addprocs) that solves this issue.

Issue #16134, opened 2016-04-30
Provide get/set functionality for jl_n_threads


josefsachsconning:
Said Sachs:
Is there any prospect for setting jl_n_threads from within Julia?  That would be more convenient than setting JULIA_NUM_THREADS before starting Julia.

Said Kelman:
Sounds like a reasonable feature request to me, to have a Threads.getnum and Threads.setnum (or equivalent names) API.

kpamnany:
High performance fork-join constructs (broadcast, reduce-barrier) typically require some setup that is dependent on the number of threads. Adding/removing threads would require teardown and recreation, and while this is certainly not impossible, I haven't run into a situation where this would be helpful. Can you describe such a situation?

josefsachsconning:
I have 2 use cases where this would be helpful.

1. I have some applications where I want a maximum of 1 thread, and other applications where I want as many threads as I have cores.  It would be much more convenient for the logic implementing this choice to be in the Julia code, rather than in a wrapper script that sets the JULIA_NUM_THREADS environment variable and then invokes the Julia script.
2. When I benchmark the effect of multithreading with varying numbers of threads, I would prefer to do it in a single Julia process rather than a different Julia process for each number of threads that I am testing.

StefanKarpinski:
Any reason this can't be done by allocating space for all potential threads up to the number of physical cores and then allowing the number of those actually used to be ramped up and down dynamically?

kpamnany:
> Any reason this can't be done by allocating space for all potential threads up to the number of physical cores and then allowing the number of those actually used to be ramped up and down dynamically?

Probably be too costly (remember KNL) this way, but like I said, it isn't that hard to do. At least in the threading infrastructure.

I'm not sure about the GC or anything in the runtime. @yuyichao, @vtjnash: can we realloc ``jl_all_heaps`` and ``jl_all_task_states``? Would it be hard to do?

yuyichao:
Runtime shouldn't be have too much trouble if the scheduler can support this (what does it mean if a worker thread request itself to be deleted?). `jl_all_heaps` should be removed (merged into `jl_tls_states_t`) and then all access to `jl_all_task_states` (should mostly be in slow path like gc and signal handling anyway) needs to be protected with a lock.

vchuravy:
@vtjnash Mentioned to me that the ability of dynamically launching threads + setting up ptls state would maybe enable calling (asynchronously) into Julia from C/C++ in a thread-safe manner.

[...]

PR #35108, opened 2020-03-14
Implement [-t|--threads] command line argument for threads

### LINKED ISSUE #26889: Feature Request: Command Line Option for Number of Threads
I like the environment variable `JULIA_NUM_THREADS`, but it should be overridable with a -t(hreads) choice.  If -p works, -t could reasonably do this, too.

fredrikekre:
Implement `[-t|--threads]` command line argument for specifying the number of Julia threads on startup, fixes #26889.

Regarding https://github.com/JuliaLang/julia/issues/26889#issuecomment-513515219: Currently this does ~not~ propagate to workers processes, ~but this commit enables that: https://github.com/JuliaLang/julia/commit/c11d9dbf0747addedd4bd8e71f9411beda696c63~

~Personally I feel like either case is fine -- for fine grained control you need to use `addprocs` anyway.~

**Edit**: I updated the PR to propagate number of threads to processes spawned with `--procs` and/or `--machine-file`. If you use those arguments in combination (e.g.  `julia -p2 -t2`) is is likely because you *want* it to propagate. And as before, if you need some other behavior you can customize all you want with `addprocs`.

### DIFF
--- a/doc/src/manual/getting-started.md
+++ b/doc/src/manual/getting-started.md
@@ -109,6 +109,7 @@ julia [switches] -- [programfile] [args...]
 |`-e`, `--eval <expr>`                  |Evaluate `<expr>`|
 |`-E`, `--print <expr>`                 |Evaluate `<expr>` and display the result|
 |`-L`, `--load <file>`                  |Load `<file>` immediately on all processors|
+|`-t`, `--threads {N\|auto`}            |Enable N threads; `auto` currently sets N to the number of local CPU threads but this might change in the future|
 |`-p`, `--procs {N\|auto`}              |Integer value N launches N additional local worker processes; `auto` launches as many workers as the number of local CPU threads (logical cores)|
 |`--machine-file <file>`                |Run processes on hosts listed in `<file>`|
 |`-i`                                   |Interactive mode; REPL runs and `isinteractive()` is true|
--- a/doc/src/manual/parallel-computing.md
+++ b/doc/src/manual/parallel-computing.md
@@ -231,27 +231,20 @@ julia> Threads.nthreads()
 1
 ```
 
-The number of threads Julia starts up with is controlled by an environment variable called `JULIA_NUM_THREADS`.
-Now, let's start up Julia with 4 threads:
+The number of threads Julia starts up with is controlled either by using the
+`-t`/`--threads` command line argument or by using the
+[`JULIA_NUM_THREADS`](@ref JULIA_NUM_THREADS) environment variable. When both are
+specified, then `-t`/`--threads` takes precedence.
 
-Bash on Linux/OSX:
+!!! compat "Julia 1.5"
+    The `-t`/`--threads` command line argument requires at least Julia 1.5.
+    In older versions you must use the environment variable instead.
+
+Lets start Julia with 4 threads:
 
 ```bash
-export JULIA_NUM_THREADS=4
+$ julia --threads 4
 ```
-
-C shell on Linux/OSX, CMD on Windows:
-
-```bash
-set JULIA_NUM_THREADS=4
-```
-
-Powershell on Windows:
-
-```powershell
-$env:JULIA_NUM_THREADS=4
-```
-
 
 Let's verify there are 4 threads at our disposal.
 
@@ -266,6 +259,29 @@ But we are currently on the master thread. To check, we use the function [`Threa
 julia> Threads.threadid()
 1
 ```
+
+!!! note
+    If you prefer to use the environment variable you can set it as follows in
+    Bash (Linux/macOS):
+    ```bash
+    export JULIA_NUM_THREADS=4
+    ```
+    C shell on Linux/macOS, CMD on Windows:
+    ```bash
+    set JULIA_NUM_THREADS=4
+    ```
+    Powershell on Windows:
+    ```powershell
+    $env:JULIA_NUM_THREADS=4
+    ```
+    Note that this must be done *before* starting Julia.
+
+!!! note
+    The number of threads specified with `-t`/`--threads` is propagated to worker processes
+    that are spawned using the `-p`/`--procs` or `--machine-file` command line options.
+    For example, `julia -p2 -t2` spawns 1 main process with 2 worker processes, and all
+    three processes have 2 threads enabled. For more fine grained control over worker
+    threads use [`addprocs`](@ref) and pass `-t`/`--threads` as `exeflags`.
 
 ## The `@threads` Macro
 
--- a/stdlib/Distributed/src/cluster.jl
+++ b/stdlib/Distributed/src/cluster.jl
@@ -1298,6 +1298,7 @@ end
 
 write_cookie(io::IO) = print(io.in, string(cluster_cookie(), "\n"))
 
+# Starts workers specified by (-n|--procs) and --machine-file command line options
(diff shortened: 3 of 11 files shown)


tknopp:
Whould it be possible to have some way of adding threads after startup. This would bei really great since it would allow packages to add threads if they are missing.

fredrikekre:
That is out of scope for this PR and should be discussed elsewhere, e.g.  #16134.

doc/src/manual/parallel-computing.md
```diff
@@ -267,6 +260,29 @@ julia> Threads.threadid()
 1
 ```
 
+!!! note
+    If you prefer to use the environment variable you can set it as follows in
+    Bash (Linux/maxOS):
```
jonas-schulze:
```suggestion
    Bash (Linux/macOS):
```

doc/src/manual/parallel-computing.md
```diff
@@ -267,6 +260,29 @@ julia> Threads.threadid()
...
+    ```bash
+    export JULIA_NUM_THREADS=4
+    ```
+    C shell on Linux/OSX, CMD on Windows:
```
jonas-schulze:
```suggestion
    C shell on Linux/macOS, CMD on Windows:
```

doc/src/manual/parallel-computing.md
```diff
@@ -267,6 +260,29 @@ julia> Threads.threadid()
...
+    ```powershell
+    $env:JULIA_NUM_THREADS=4
+    ```
+    Note that this must be done *before* starting Julia.
+
+!!! note
+    The number of threads specified with `-t`/`--threads` is propagated to worker processes
+    that are spawned using the `-p`/`--procs` or `--machine-file` command line options.
+    For example, `julia -p2 -t2` spawns 1 main process with 2 worker processes, and all
+    three processes has 2 threads enabled. For more fine grained control over worker
```
jonas-schulze:
```suggestion
    three processes have 2 threads enabled. For more fine grained control over worker
```

test/cmdlineargs.jl
```diff
@@ -186,6 +186,32 @@ let exename = `$(Base.julia_cmd()) --startup-file=no`
...
+            @test read(`$exename --threads 2 -e $code`, String) ==
+                  read(`$exename --threads=2 -e $code`, String) ==
+                  read(`$exename -t2 -e $code`, String) ==
+                  read(`$exename -t 2 -e $code`, String) == "2"
+        end end
+    end
+    @test !success(`$exename -t 0`)
+    @test !success(`$exename -t -1`)
+
+    # Combining --threads and --procs: --threads does not propagate
```
jonas-schulze:
```suggestion
    # Combining --threads and --procs: --threads does propagate
```

test/cmdlineargs.jl
```diff
@@ -186,6 +186,32 @@ let exename = `$(Base.julia_cmd()) --startup-file=no`
...
+    if cpu_threads > 1; withenv("JULIA_NUM_THREADS"=>nothing) do
+        code = "print(Threads.nthreads() + sum(remotecall_fetch(() -> Threads.nthreads(), x) for  x in workers()))"
```
jonas-schulze:
```suggestion
        code = "print(sum(remotecall_fetch(Threads.nthreads, x) for x in procs()))"
```

doc/src/manual/getting-started.md
```diff
@@ -109,6 +109,7 @@ julia [switches] -- [programfile] [args...]
 |`-e`, `--eval <expr>`                  |Evaluate `<expr>`|
 |`-E`, `--print <expr>`                 |Evaluate `<expr>` and display the result|
 |`-L`, `--load <file>`                  |Load `<file>` immediately on all processors|
+|`-t`, `--threads {N\|auto`}            |Enable N threads; `auto` sets N to the number of local CPU threads|
```
tkf, line 112: > `auto` sets N to the number of local CPU threads

Can we be a bit vague about this?  Or reserve the word `auto`?  It'd be nice to if `--threads=auto` uses the number of physical cores and also aware of cgroups.

stdlib/Distributed/src/cluster.jl
```diff
@@ -1310,14 +1311,17 @@ function process_opts(opts)
         end
     end
 
+    # Propagate --threads to workers
+    exeflags = opts.nthreads > 0 ? `--threads=$(opts.nthreads)` : ``
```
tkf, line 1315: Is it intentional that this does not look at `Threads.nthreads()`?  That is to say, `JULIA_NUM_THREADS` does not propagate to the remote workers but `--threads` does?

doc/src/manual/getting-started.md
StefanKarpinski: Or at least say that the behavior of `auto` might change in the future.

tkf: That sounds good to me too.

fredrikekre: Sure. Just FYI, I didn't "invent" `auto` for this, just reused the same implementation as we have for `--procs`.

stdlib/Distributed/src/cluster.jl
fredrikekre: My thinking was that if you use the env variable, then you probably use the env variable on the workers too. This is just a way to override it.

doc/src/manual/getting-started.md
tkf: Ah, I missed that. Can the documentation of `--proc` also include that the interpretation of `auto` may change in the future? (i.e., can relaxing `--proc=auto`'s guarantee be considered minor change?)

Though maybe the safest option is to invent a new name other than `auto` when we have the facility to detect the available number of physical cores.

stdlib/Distributed/src/cluster.jl
tkf: > you probably use the env variable on the workers too

I think it's useful and easy for "local" workers. But isn't this hard to do for "remote" workers (e.g., launched via ssh)? IIUC, there is no easy way to use environment variables in remote workers other than creating a "shim" script and setting the environment variable there.

I just thought it may be confusing that `Threads.nthreads()` does not propagate only when the workers are in different machines _and_ it is set via `JULIA_NUM_THREADS` in the main process.

I don't have a strong opinion on what is right default here, though.

fredrikekre:
Specific questions for triage:
- How should `$ julia --procs 2 --threads 3` behave? Currently 1+ 2 processes with 3 threads each.
- How should `$ JULIA_NUM_THREADS=4 julia --procs 2 --threads 3` behave? Currently 1+ 2 processes with 3 threads each. I think it makes sense that the command line flag takes precedence since it is more "local".
- What should happen when `JULIA_NUM_THREADS` are set on the remote machine?

StefanKarpinski:
How about `--threads` always takes precedence; if someone wants to start a different number of threads on each machine, they can set `JULIA_NUM_THREADS` on each machine without passing `--threads` to each process. We should also add a feature to the machine file format so that you can indicate how many processes and threads per process on each machine.

JeffBezanson:
Triage is ok with what's implemented here. We also might want to eventually phase out the environment variable.

oschulz:
Closes #18074

PR #35111, opened 2020-03-14
[WIP] enable randn(::BigFloat) and randexp(::BigFloat) (implemented in MPFR)

### LINKED ISSUE #17629: Consider adding BigFloat support for randn/randexp
https://github.com/JuliaLang/julia/pull/17627 adds MethodError domain checks that you cannot call `randn` or `randexp` for any types besides the currently supported `Float16,Float32,Float64`.

Do we also want to extend support to `BigFloat`? I don't know much about, but I've left TODO code comments in https://github.com/JuliaLang/julia/pull/17627#issuecomment-235198088 asking to consider this question.

This is a tracking issue for that discussion.

rfourquet:
Closes #17629.

### DIFF
--- a/stdlib/Random/src/normal.jl
+++ b/stdlib/Random/src/normal.jl
@@ -213,10 +213,14 @@ function Sampler(::Type{<:AbstractRNG}, d::Union{Normal{BigFloat},Exponential{Bi
     return s
 end
 
-function rand(rng::AbstractRNG, s::GMPRandState)
+function rand(rng0::AbstractRNG, s::GMPRandState)
     z = BigFloat()
-    if !ismutable(rng)
-        rng = MutableRNG(rng)
+    if ismutable(rng)
+        rng = rng0
+    elseif rng0 isa TaskLocalRNG
+        rng = MutableRNG(copy(rng0)::Xoshiro)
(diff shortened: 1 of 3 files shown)


rfourquet:
This currently doesn't support immutable RNGs. I tried the approach of wrapping an immutable RNG inside a mutable one, but this segfaults. Ideas welcome :)

stdlib/Random/src/normal.jl
```diff
@@ -116,6 +126,111 @@ end
...
+    s = GMPRandState()
+    s.alg_data = get_JuliaGMPRandFnPtr()
+    s.alg = Cint(d isa Normal{BigFloat})
+    return s
+end
+
+function rand(rng::AbstractRNG, s::GMPRandState)
+    z = BigFloat()
+    if !ismutable(rng)
+        rng = MutableRNG(rng)
```
rfourquet: replacing this line by `return rand(Mutable(rng), s)` (and fixing a method ambiguity) seems to solve the segfault problem, but I'd prefer some advice from someone who understands...

ViralBShah:
Would be nice to revive this and get it done.

PR #35112, opened 2020-03-14
LLVM11 Compatibility

vchuravy:
replaces #34793

Currently fails with asserts turned on:

```
julia: /data/vchuravy/builds/julia/usr/include/llvm/Support/Casting.h:104: static bool llvm::isa_impl_cl<To, const From*>::doit(const From*) [with To = llvm::GlobalVariable; From = llvm::GlobalValue]: Assertion `Val && "isa<> used on a null pointer"' failed.

signal (6): Aborted
gsignal at /lib/x86_64-linux-gnu/libc.so.6 (unknown line)
abort at /lib/x86_64-linux-gnu/libc.so.6 (unknown line)
unknown function (ip: 0x7f65cd891399)
__assert_fail at /lib/x86_64-linux-gnu/libc.so.6 (unknown line)
doit at /data/vchuravy/builds/julia/usr/include/llvm/Support/Casting.h:104 [inlined]
doit at /data/vchuravy/builds/julia/usr/include/llvm/Support/Casting.h:131 [inlined]
doit at /data/vchuravy/builds/julia/usr/include/llvm/Support/Casting.h:123 [inlined]
isa<llvm::GlobalVariable, llvm::GlobalValue*> at /data/vchuravy/builds/julia/usr/include/llvm/Support/Casting.h:142 [inlined]
cast<llvm::GlobalVariable, llvm::GlobalValue> at /data/vchuravy/builds/julia/usr/include/llvm/Support/Casting.h:263 [inlined]
jl_create_native at /data/vchuravy/src/julia/src/aotcompile.cpp:354
```

Compiled against 19840a307e6d98017b68b14725c53d1ad89d98f9

cc: @yhls


src/aotcompile.cpp
```diff
@@ -309,7 +309,11 @@ void *jl_create_native(jl_array_t *methods, const jl_cgparams_t cgparams)
     JL_GC_POP();
 
     // process the globals array, before jl_merge_module destroys them
+#if JL_LLVM_VERSION >= 110000
+    std::vector<StringRef> gvars;
```
vtjnash: This StringRefs get freed below (it's why we were using the implicit copy-constructor before—now you need to write the cast explicitly)

src/aotcompile.cpp
```diff
@@ -309,7 +309,11 @@ void *jl_create_native(jl_array_t *methods, const jl_cgparams_t cgparams)
...
+#else
     std::vector<std::string> gvars;
+#endif
     for (auto &global : params.globals) {
         gvars.push_back(global.second->getName());
```
vtjnash:
```suggestion
        gvars.push_back(std::string(global.second->getName()));
```

vchuravy:
Updated to do some LLVM9 patch maintenance as well. We were missing the one for D50010 that Yggdrasil was already carrying. 

Should be good to go.

PR #35116, opened 2020-03-15
Add @inline for do blocks

tkf:
This PR makes `@inline` work with `do` blocks.  This is very useful for writing high-performance code with higher-order functions such as `foreach`, `foldl`, and `reduce`.

### DIFF
--- a/base/expr.jl
+++ b/base/expr.jl
@@ -273,12 +273,11 @@ function pushmeta!(ex::Expr, sym::Symbol, args::Any...)
         inner = inner.args[end]::Expr
     end
 
-    idx, exargs = findmeta(inner)
-    if idx != 0
-        push!(exargs[idx].args, tag)
+    found, metaargs = findmeta(inner)
+    if found
+        push!(metaargs, tag)
     else
-        body::Expr = inner.args[2]
-        pushfirst!(body.args, Expr(:meta, tag))
+        pushfirst!(metaargs, Expr(:meta, tag))
     end
     ex
 end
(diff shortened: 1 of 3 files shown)


StefanKarpinski:
Alternately, should do blocks be automatically marked for probably inlining?

JeffBezanson:
> Alternately, should do blocks be automatically marked for probably inlining?

I would prefer not to do that, since many uses of do blocks don't need that level of performance, and it will add compile time cost. It's easier to mark cases where you need performance than to mark cases where you don't need performance, especially since the only symptom would be gradual degradation in compile times.

tkf:
I think a long term solution would be to have call-site inlining as the caller has a good guess for what should be inlined in this case (e.g., `foreach` requires inlining but `open` does not).

JeffBezanson:
Yes, call-site inlining would be good. Now that you mention it, the syntax in this PR conflicts with it --- `@inline f() do ... end` would mean `f` should be inlined if we had that feature.

StefanKarpinski:
IIRC, @vtjnash claimed that call site inlining would be "fairly trivial to implement" like three yrs ago.

tkf:
Maybe inline both is what you'd want in this case?  For finer control, how about

Normal call

```julia
f(a, b) do c, d
end
```

Inline only `f`:

```julia
@inline(f(a, b)) do c, d
end
```

Inline only `do` block:

```julia
f(a, b) do c, d
    @inline  # just like @_inline_meta
end
```

aviatesk:
Superseded by #41312.

PR #35119, opened 2020-03-15
clarify error message for constant redefinition

### LINKED ISSUE #28689: Segmentation fault from changing constants
I have been struggling for quite some time to debug a code by one of my students, and finally managed to trace the error down to modifying constant global variables, which are not `isbits`, and which have been referred to in functions that are compiled. Julia gives a gentle warning upon redefining constants, but you don't expect a segmentation fault resulting from it. The bug was originally in a julia 0.6 code, but the same still happens in 1.1-DEV
```julia
const A = randn(10)
function f()
   return sum(A)
end

f() # -> returns something

A = randn(10) # -> redefine, Julia prints a warning

f() # -> still returns the old value

# do some more stuff, until at some point the garbage collector is triggered
GC.gc()

f() # -> segfault, probably not directly, but eventually, just do some more times gc() and other allocating stuff

```

vtjnash:
Before it sounded like a linter message (the description was that this was a gentle notice), now it should be clearer that this warning means Julia may now be partially dysfunctional.

Closes #28689

### DIFF
--- a/src/module.c
+++ b/src/module.c
@@ -707,7 +707,7 @@ JL_DLLEXPORT void jl_checked_assignment(jl_binding_t *b, jl_value_t *rhs)
                 jl_errorf("invalid redefinition of constant %s",
                           jl_symbol_name(b->name));
             }
-            jl_printf(JL_STDERR, "WARNING: redefining constant %s\n",
+            jl_printf(JL_STDERR, "WARNING: redefinition of constant %s. This may fail, cause incorrect answers, or produce other errors.\n",
                       jl_symbol_name(b->name));
         }
     }
--- a/doc/src/manual/variables-and-scoping.md
+++ b/doc/src/manual/variables-and-scoping.md
@@ -630,7 +630,7 @@ julia> const y = 1.0
 1.0
 
 julia> y = 2.0
-WARNING: redefining constant y
+WARNING: redefinition of constant y. This may fail, cause incorrect answers, or produce other errors.
 2.0
 ```
 * if an assignment would not result in the change of variable value no message is given:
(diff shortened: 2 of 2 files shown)


src/module.c
```diff
@@ -691,7 +691,7 @@ JL_DLLEXPORT void jl_checked_assignment(jl_binding_t *b, jl_value_t *rhs)
                 jl_errorf("invalid redefinition of constant %s",
                           jl_symbol_name(b->name));
             }
-            jl_printf(JL_STDERR, "WARNING: redefining constant %s\n",
+            jl_printf(JL_STDERR, "WARNING: redefinition of constant %s may cause Julia to fail\n",
```
vtjnash:
```suggestion
            jl_printf(JL_STDERR, "WARNING: redefinition of constant %s may cause the program to fail\n",
```

StefanKarpinski:
```suggestion
            jl_printf(JL_STDERR, "WARNING: redefinition of constant %s may cause problems (including crashes), continue at your own risk\n",
```

vtjnash: Or:
```suggestion
            jl_printf(JL_STDERR, "WARNING: redefinition of constant %s may sometimes fail or cause crashes\n",
```

vtjnash: Or:
```suggestion
            jl_printf(JL_STDERR, "WARNING: redefinition of constant %s causes undefined behavior, which is not a bug\n",
```

KristofferC: This would be printed as

> redefinition of constant `foo` may sometimes fail or cause crashes

and make it sound like it is a particular statement about `foo`.

KristofferC: So something like:

> WARNING: redefinition of constant `foo`. Redefining constants may fail or cause crashes.

is better imo.

StefanKarpinski: I like that last wording 👍

vtjnash:
```suggestion
            jl_printf(JL_STDERR, "WARNING: redefinition of constant %s. This may fail, cause incorrect answers, or produce other errors.\n",
```

vtjnash:
Okay to merge?

KristofferC (approved):

PR #35124, opened 2020-03-16
Add rot to BLAS in stdlib/LinearAlgebra

amontoison:
Applies the plane rotation : 
```math
[ c  s] [x]
[-s̅  c] [y]
```


stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -11,6 +11,7 @@ using Base: require_one_based_indexing
 
 export
 # Level 1
+    rot,
```
StefanKarpinski: Should be `rot!` since it modifies `X` in place.

NEWS.md
```diff
@@ -102,7 +102,7 @@ Standard library changes
 * `normalize` now supports multidimensional arrays ([#34239])
 * `lq` factorizations can now be used to compute the minimum-norm solution to under-determined systems ([#34350]).
 * The BLAS submodule now supports the level-2 BLAS subroutine `spmv!` ([#34320]).
-
+* The BLAS submodule now supports the level-1 BLAS subroutine `rot!` ([#35124]).
```
StefanKarpinski: This should keep the following blank line.

stdlib/LinearAlgebra/test/blas.jl
```diff
@@ -78,6 +78,28 @@ Random.seed!(100)
                 @test BLAS.iamax(z) == argmax(map(x -> abs(real(x)) + abs(imag(x)), z))
             end
         end
+        @testset "rot" begin
```
StefanKarpinski:
```suggestion
        @testset "rot!" begin
```

andreasnoack:
I don't really mind adding this but notice that the functionality is already covered by `LinearAlgebra.Givens`, i.e. you can do
```
julia> LinearAlgebra.Givens(1, 3, 0.3, 0.5)*[1, 1, 1]
3-element Array{Float64,1}:
  0.8
  1.0
 -0.2

julia> lmul!(LinearAlgebra.Givens(1, 3, 0.3, 0.5), [1.0, 1, 1])
3-element Array{Float64,1}:
  0.8
  1.0
 -0.2
```
and it's probably more efficient than calling out to BLAS.

StefanKarpinski:
Maybe worth a speed comparison then?

amontoison:
@andreasnoack `x` and `y` are vectors here, can we do the same thing with LinearAlgebra.Givens? 
An example or purpose is inside an implementation of MINRES-QLP (Krylov solver for linear systems): 
```
# [ẘₖ₋₂ w̄ₖ₋₁ vₖ] [cpₖ  0   spₖ] [1   0    0 ] = [wₖ₋₂ ẘₖ₋₁ w̄ₖ] ⟷ wₖ₋₂ = cpₖ * ẘₖ₋₂ + spₖ * vₖ
#                [ 0   1    0 ] [0  cdₖ  sdₖ]                  ⟷ ẘₖ₋₁ = cdₖ * w̄ₖ₋₁ + sdₖ * (spₖ * ẘₖ₋₂ - cpₖ * vₖ)
#                [spₖ  0  -cpₖ] [0  sdₖ -cdₖ]                  ⟷ w̄ₖ   = sdₖ * w̄ₖ₋₁ - cdₖ * (spₖ * ẘₖ₋₂ - cpₖ * vₖ)

# Compute wₐᵤₓ = spₖ * ẘₖ₋₂ - cpₖ * vₖ
axpby!(n, -cpₖ, vₖ, spₖ, ẘₖ₋₂)
wₐᵤₓ = ẘₖ₋₂
# Compute ẘₖ₋₁ and w̄ₖ
ref!(n, w̄ₖ₋₁, wₐᵤₓ, cdₖ, sdₖ)
```
with
```
ref!(n, x, y, c, s) = rot!(n, y, x, s, c)
```
I update ẘₖ₋₁ and w̄ₖ without allocating extra memory. Does `LinearAlgebra.Givens` can do that without allocating?

And in general we like to used Givens reflexions (symmetric) instead of Givens rotations.

stdlib/LinearAlgebra/test/blas.jl
```diff
@@ -78,6 +78,28 @@ Random.seed!(100)
...
         end
+        @testset "rot!" begin
+            if elty <: Real
+                x = convert(Vector{elty}, randn(n))
+                y = convert(Vector{elty}, randn(n))
+                c = rand(elty)
+                s = rand(elty)
+                x2, y2 = BLAS.rot!(n,copy(x),1,copy(y),1,c,s)
+                @test x2 ≈ c*x + s*y
+                @test y2 ≈ -s*x + c*y
```
dkarrasch: I'd suggest to include testing that `x` and `y` are actually overwritten, something along the lines
```suggestion
                x2 = copy(x)
                y2 = copy(y)
                BLAS.rot!(n, x, 1, y, 1, c, s)
                @test x ≈ c*x2 + s*y2
                @test y ≈ -s*x2 + c*y2
```
and similarly below.

andreasnoack:
The current methods for `Givens` might not do exactly what you want since it's written for the situation where `x` and `y` are stored in the same array. Your example doesn't render correctly so I'm not sure exactly what you are doing. In any case, a version that works for two vectors instead of two rows in a matrix should be fairly simple to write in Julia. I still think it could easily be as fast or faster than the BLAS.

dkarrasch:
Indeed, I found and adopted this one

https://github.com/JuliaLang/julia/blob/2a5bb59facdce6c8d987a3752f2b50e2e06d7bef/stdlib/LinearAlgebra/src/givens.jl#L339-L351

to this:
```julia
function rot!(x::AbstractVector, y::AbstractVector, c, s)
    LinearAlgebra.require_one_based_indexing(x, y)
    n = length(x)
    n == length(y) || throw(DimensionMismatch("different lenghts"))
    @inbounds for i = 1:n
        xi, yi = x[i], y[i]
        x[i] =       c *xi + s*yi
        y[i] = -conj(s)*xi + c*yi
    end
    return x, y
end
```
which passes the tests. This is quick-and-dirty, one could be more permissive in terms of lengths (like pass an extra `n` and apply the rotation to the first `n` elements only, or determine `n` as the minimum of the lengths of the vectors) etc. @amontoison Want to give it a shot? It has one allocation that I don't see. Here is a little comparison with how one would obtain the same result with `Givens`:
```julia
using BenchmarkTools, Test
c, s = rand(), rand()
x0, y0 = rand(1000), rand(1000);
@btime rot!(x, y, $c, $s) setup=(x = copy(x0); y = copy(y0)); # 245 ns
@btime lmul!(LinearAlgebra.Givens(1, 2, $c, $s), A) setup=(A = [reshape(x0, 1, :); reshape(y0, 1, :)]); # 721 ns
x, y = rand(1000), rand(1000);
A = [reshape(x, 1, :); reshape(y, 1, :)];
rot!(x, y, c, s)
lmul!(LinearAlgebra.Givens(1, 2, c, s), A)
@test A ≈ [reshape(x, 1, :); reshape(y, 1, :)]
```

amontoison:
Sorry for the delay, I did some benchmarks @dkarrasch and your `rot!` performs as well as the BLAS version for me. I decided to add it in `generic.jl` with a variant `ref!` that applies the reflection instead of the rotation.

The BLAS version could be still relevant if you use `mkl` instead of `openblas` for example.

About the allocation, it's the pointers for `x` and `y` that are returned at the end :wink:

stdlib/LinearAlgebra/src/generic.jl
```diff
@@ -1416,6 +1416,49 @@ function axpby!(α, x::AbstractArray, β, y::AbstractArray)
...
+end
+
+"""
+    ref!(x, y, c, s)
+
+Overwrite `x` with `c*x + s*y` and `y` with `conj(s)*x - c*y`.
+Returns `x` and `y`.
+
+!!! compat "Julia 1.5"
+    `rot!` requires at least Julia 1.5.
```
dkarrasch: rot! -> ref!

stdlib/LinearAlgebra/src/generic.jl
```diff
@@ -1416,6 +1416,49 @@ function axpby!(α, x::AbstractArray, β, y::AbstractArray)
...
+"""
+function ref!(x::AbstractVector, y::AbstractVector, c, s)
+    n = length(x)
+    if n != length(y)
+        throw(DimensionMismatch("x has length $(length(x)), but y has length $(length(y))"))
+    end
+    @inbounds for i = 1:n
```
dkarrasch: If you start your indices with 1 here, then you will need to `require_one_based_indexing(x, y)` before, and similarly in `rot!`.

dkarrasch (changes requested):

NEWS.md
```diff
@@ -111,6 +111,7 @@ Standard library changes
 * `normalize` now supports multidimensional arrays ([#34239])
 * `lq` factorizations can now be used to compute the minimum-norm solution to under-determined systems ([#34350]).
 * The BLAS submodule now supports the level-2 BLAS subroutine `spmv!` ([#34320]).
+* The BLAS submodule now supports the level-1 BLAS subroutine `rot!` ([#35124]).
```
dkarrasch: You may want to announce the new generic functions here as well.

stdlib/LinearAlgebra/test/generic.jl
```diff
@@ -217,6 +217,25 @@ end
...
 
+@testset "rot! and ref!" begin
+    x = rand(1000)
+    y = rand(1000)
+    c = rand()
+    s = rand(ComplexF64)
+
+    x2 = copy(x)
+    y2 = copy(y)
+    rot!(n, x, y, c, s)
```
dkarrasch:
```suggestion
    rot!(x, y, c, s)
```

stdlib/LinearAlgebra/test/generic.jl
```diff
@@ -217,6 +217,25 @@ end
...
+    @test x ≈ c*x2 + s*y2
+    @test y ≈ -conj(s)*x2 + c*y2
+
+    x3 = copy(x)
+    y3 = copy(y)
+    ref!(n, x, y, c, s)
```
dkarrasch:
```suggestion
    ref!(x, y, c, s)
```

andreasnoack:
Great with generic implementations. Since they are no longer wrappers of BLAS functions, I'll suggest that the names are made more informative since the F77  variable name restrictions don't apply. So maybe `rotate!` and `reflect!`?

dkarrasch (approved):
LGTM

StefanKarpinski:
This seems good. @dkarrasch, anyone else you'd want to have look it over?

stdlib/LinearAlgebra/src/generic.jl
dpo: Late to the party but I'm just curious: why not allow more general arrays and use `eachindex` here? The documentation of `eachindex` says

> If you supply more than one AbstractArray argument, eachindex will create an iterable object that is fast for all arguments

so it seems possible to iterate over `x` and `y` even though they might have different indexing styles?!

dkarrasch: I'm not sure, but I believe that `eachindex` throws an error if you pass, say, two vectors of equal length, but one has an offset.

PR #35132, opened 2020-03-16
add contains (argument flipped occursin)

### LINKED ISSUE #35031: Bring back contains(haystack, needle)
Right now we have:
 - `startswith(haystack, needle)`
 - `endswith(haystack, needle)`
 - and `occursin(needle, haystack)`

While `occursin(needle, haystack)` is right from the perspective of reading
`f(a, b)` as `a f b`
i.e. `needle occursin haystack`

It is not consistent with `startwith` or `endswith`.

Back in julia 0.6 (IIRC) we had `contains(haystack, needle)`
which still reads right: `haystack contains needle`.

I would like to suggest we bring it back so that we can have that consistency.

----
This is part of a series of issues relating to  https://github.com/invenia/Wrangling.jl/pull/3
Will link others below for cross referencing.
https://github.com/JuliaLang/julia/issues/35032
https://github.com/JuliaLang/julia/issues/35033

oxinabox:
closes #35031 
See issue for discussion.

Includes curried form per convention in https://github.com/JuliaLang/julia/pull/35052

in 2.0 we may want to remove `occursin`

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -75,6 +75,7 @@ New library functions
 * New function `bitreverse` for reversing the order of bits in a fixed-width integer ([#34791]).
 * New function `bitrotate(x, k)` for rotating the bits in a fixed-width integer ([#33937]).
 * One argument methods `startswith(x)` and `endswith(x)` have been added, returning partially-applied versions of the functions, similar to existing methods like `isequal(x)` ([#33193]).
+* New function `contains(haystack, needle)` and its one argument partially applied form have been added, it acts like `occursin(needle, haystack)`([#35132]).
 
 New library features
 --------------------
--- a/base/exports.jl
+++ b/base/exports.jl
@@ -439,6 +439,7 @@ export
     zeros,
 
 # search, find, match and related functions
+    contains,
     eachmatch,
     endswith,
     findall,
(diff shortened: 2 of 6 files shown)


JeffBezanson:
Triage is ok with this.

oxinabox:
Is this ready for merge?

oxinabox:
bump

oxinabox:
bump

musm:
Thanks for this!

o314:
Does this really a different function  than in ?
Now, we have

```julia
using Test
@test 6 |> in(1:10)
@test "JuliaLang" |> startswith(r"Julia|Romeo")
@test_broken "aba" |> in(r"a.a") # will throw
@test "aba" |> contains(r"a.a")
```

Why do we not allow ?
```julia
using Test
using Test
@test 6 |> in(1:10)
@test "JuliaLang" |> startswith(r"Julia|Romeo")
@test"aba" |> in(r"a.a")
```

oxinabox:
In is for elements, `contains` is for substrings.

o314:
It is not in the essence of multidispatch to specialize function according to type. Then we can talk for hours of what is a type.
If it is about subsetting there is `issubset`.
There is already `in`, but also `issubset`, `occursin` the badly named `ismatch`.
Adding one at time more and more naming - close, but not exactly the same, but how much, not sure - kill slowly the power of multidispatch .

PR #35085, opened 2020-03-12
add <: to add-dots

ssikdar1:
Fixes  #34839 .
Add `<:` to `add-dots`.

# Testing
Not sure if there's somewhere to add a test?
But testing in the REPL:
```
julia> [Int, Float64, String, Bool] .<: Union{Int, String}
4-element BitArray{1}:
 1
 0
 1
 0
```

PR #35136, opened 2020-03-16
fix bug in t-function for `>:`

JeffBezanson:
Fixes issue noted in https://github.com/JuliaLang/julia/pull/35085#issuecomment-599168585


StefanKarpinski:
Should this be backported? Seems like it could have been a longish-standing bug.

JeffBezanson:
Yes it could be backported, it's just triggered very rarely.

PR #35137, opened 2020-03-16
add findall(char, str) and count(char, str)

bicycle1885:
Since `findfirst(char, string)` works, why don't we add `findall(char, string)` and `count(char, string)`?


StefanKarpinski:
Looks good to me aside from the need for compat and news.

bicycle1885:
This looks being superseded by #38675.

Issue #31458, opened 2019-03-23
Line number for first statement in short form of function definition is wrong


KristofferC:
```jl
julia> f(x) = 
           (x = 2; x = 3;
            x = 1)
f (generic function with 1 method)

julia> m = @code_lowered f(2.0);

julia> m.code
5-element Array{Any,1}:
 :(_3 = _2) 
 :(_3 = 2)  
 :(_3 = 3)  
 :(_3 = 1)  
 :(return 1)

julia> m.linetable[m.codelocs]
5-element Array{Any,1}:
 Core.LineInfoNode(Main, :f, Symbol("REPL[54]"), 1, 0)
 Core.LineInfoNode(Main, :f, Symbol("REPL[54]"), 1, 0)
 Core.LineInfoNode(Main, :f, Symbol("REPL[54]"), 2, 0)
 Core.LineInfoNode(Main, :f, Symbol("REPL[54]"), 3, 0)
 Core.LineInfoNode(Main, :f, Symbol("REPL[54]"), 3, 0)
```

Why is the statement where `x` gets assigned to 2 said to be on line `1` and the one where it is assigned to 3 on line `2`. 

Compare to the long form version of defining functions:

```
julia> function g(x) 
           (x = 2; x = 3;
            x = 1)
       end
g (generic function with 1 method)

julia> m2 = @code_lowered g(2.0);

julia> m2.code
5-element Array{Any,1}:
 :(_3 = _2) 
 :(_3 = 2)  
 :(_3 = 3)  
 :(_3 = 1)  
 :(return 1)

julia> m2.linetable[m2.codelocs]
5-element Array{Any,1}:
 Core.LineInfoNode(Main, :g, Symbol("REPL[58]"), 2, 0)
 Core.LineInfoNode(Main, :g, Symbol("REPL[58]"), 2, 0)
 Core.LineInfoNode(Main, :g, Symbol("REPL[58]"), 2, 0)
 Core.LineInfoNode(Main, :g, Symbol("REPL[58]"), 3, 0)
 Core.LineInfoNode(Main, :g, Symbol("REPL[58]"), 3, 0)
```

The code is exactly the same, the statements are on the same line, but here the line numbers are correct for the first statement (2).

c42f:
The problem is in the parser: to get the correct `functionloc`, the parser has a special case for short form functions which adds an `Expr(:block)` corresponding to the function name's location, not the location of the body:

```
julia> :(f(x) = 
           (x=1;))
[...]

PR #17470, opened 2016-07-18
Fix coverage for the first statement in the function

yuyichao:
Fixes #17442 

Not sure how to add tests but at least this fixes the coverage [on my package](https://codecov.io/github/yuyichao/LibALPM.jl/commits).

Issue #17442, opened 2016-07-15
0.5 code-coverage regression


MichaelHatherly:
`--code-coverage` appears to be skipping the first line of methods. For example with the following package:

```julia
module TestPackage

function test()
    a = 1
    b = 2
    a + b
end

end
```

and `test/runtests.jl` containing:

```julia
using TestPackage
using Base.Test

TestPackage.test()
```

On `0.5` we have

```
$ julia-0.5 --code-coverage=user test/runtests.jl
$ cat src/TestPackage.jl.11032.cov 
        - module TestPackage
        - 
        - function test()
        -     a = 1
        2     b = 2
        2     a + b
        - end
        - 
        - end
        - 
```

while on `0.4`

```
$ julia-0.4 --code-coverage=user test/runtests.jl
$ cat src/TestPackage.jl.9700.cov 
        - module TestPackage
        - 
        - function test()
        1     a = 1
        1     b = 2
        1     a + b
        - end
        - 
        - end
        - 
```

`--inline=no` doesn't change the result.

This is on `0.5.0-dev+5448` 54a6490 and `0.4.7-pre+3` eb322c0.

MichaelHatherly:
`git bisect` is giving bd06d27c001cd1c0a9d118c3614c74d33b406cb6 as the first bad commit.

PR #35138, opened 2020-03-17
Fix functionloc

### LINKED ISSUE #34820: funcionloc should return line number of method sign, not line of first statement in func

c42f:
This is a WIP of one way to fix https://github.com/JuliaLang/julia/issues/34820 — I'm not sure it's the best approach, but throwing it up there for discussion.

The approach taken here is to add a line number node to the method `signature` argument in `Expr(:method, name, signature, body)` after the signature argument types and names. I tried several options involving lowering of `Expr(:method)` and it turns out that this is quite minimally invasive way to do this. I also think it makes some sense: the method declaration is more a property of the method than it is relevant to the body `CodeInfo`.

However, there's one downside I know about, which is that functions defined in expressions (rather than as statements) can have a line number which is incorrect. In particular, the following test breaks:

```julia
let f(x) =
      g(x) = 1
    @test_broken functionloc(f(1))[2] > functionloc(f)[2]
end
```

Presumably this could be fixed by changing parsing of functions which occur inside expressions to be enclosed in an `Expr(:block)`. Perhaps that's a reasonable way to go?

### TODO
* [x] Add extra line number node to long form function bodies during parsing, and extract that in lowering.
* [x] @KristofferC mentions that https://github.com/JuliaLang/julia/issues/31458 is related — investigate whether we can fix that as part of this work.  (Decision: not directly related - fix separately)

### DIFF
--- a/src/codegen.cpp
+++ b/src/codegen.cpp
@@ -5820,10 +5820,13 @@ static std::pair<std::unique_ptr<Module>, jl_llvm_functions_t>
     auto coverageVisitStmt = [&] (size_t dbg) {
         if (dbg == 0)
             return;
+        // Compute inlining stack for current line, inner frame first
         while (dbg) {
             new_lineinfo.push_back(dbg);
             dbg = linetable.at(dbg).inlined_at;
         }
+        // Visit frames which differ from previous statement as tracked in
+        // current_lineinfo (tracked outer frame first).
         current_lineinfo.resize(new_lineinfo.size(), 0);
         for (dbg = 0; dbg < new_lineinfo.size(); dbg++) {
             unsigned newdbg = new_lineinfo[new_lineinfo.size() - dbg - 1];
@@ -5906,15 +5909,6 @@ static std::pair<std::unique_ptr<Module>, jl_llvm_functions_t>
         BB[label] = bb;
     }
 
-    if (do_coverage(mod_is_user_mod)) {
-        coverageVisitLine(ctx, ctx.file, toplineno);
-        if (linetable.size() >= 2) {
-            // avoid double-counting the entry line
-            const auto &info = linetable.at(1);
-            if (info.file == ctx.file && info.line == toplineno && info.is_user_code == mod_is_user_mod)
-                current_lineinfo.push_back(1);
-        }
-    }
     Value *sync_bytes = nullptr;
     if (do_malloc_log(true))
         sync_bytes = ctx.builder.CreateCall(prepare_call(diff_gc_total_bytes_func), {});
--- a/src/julia-syntax.scm
+++ b/src/julia-syntax.scm
@@ -231,7 +231,7 @@
   (let ((type-ex (caddr m)))
     (if (eq? (car type-ex) 'block)
         ;; extract ssavalue labels of sparams from the svec-of-sparams argument to `method`
-        (let ((sp-ssavals (cddr (last (last type-ex)))))
+        (let ((sp-ssavals (cddr (cadddr (last type-ex)))))
           (map (lambda (a)  ;; extract T from (= v (call (core TypeVar) (quote T) ...))
                  (cadr (caddr (caddr a))))
                (filter (lambda (e)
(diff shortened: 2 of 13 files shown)


JeffBezanson:
Ah, interesting. `f(x) = x` syntax has the opposite problem: it only records the line of the signature and not the first body expression. So indeed it seems a comprehensive solution requires modifying the parser. I think adding a block expression around every function is too disruptive; it would be easier for the parser to record two locations at the start of each function body, and have lowering move the first one into the signature field of method exprs as you have here.

c42f:
Ok, the doctests bring up some more serious ways that this approach can lead to unfortunate results:
* Ignoring the body leads to surprises when `eval()`ing ASTs containing function definitions which are not embedded in a block. This is hard to fix because in the current AST there's just nowhere natural to stash source location of a bare function definition unless we store it in the body statements. Wrapping in a block seems really disruptive.
* It seems that source location tracking is broken for inner constructors because `struct`s ASTs have special case traversal in `expand-forms`. This can be fixed, but leads me to think location tracking within lowering is too fragile.

In general I like this approach on the runtime side but it seems too fragile to use with confidence without parser support.

> I think adding a block expression around every function is too disruptive; it would be easier for the parser to record two locations at the start of each function body, and have lowering move the first one into the signature field of method exprs as you have here.

Absolutely agreed! Let's do it this way.

JeffBezanson (approved):

JeffBezanson:
LGTM, just has the inevitable detailed test errors.

c42f:
Thanks for taking another look, I'll sort out those test issues, then merge.

I think this is likely to incidentally break the occasional package even though it's technically-not-API-breaking. Is there any way to flag this kind of thing so users can easily understand what has changed and how to fix their code? Applying the `breaking` label seems like an overstatement.

src/julia-syntax.scm
```diff
@@ -293,6 +293,24 @@
...
+                          (car prologue-lnos)
+                          ; Fallback - take first line anywhere in body
+                          (let ((lnos (filter linenum? body)))
+                            (if (null? lnos) '(line 0 none) (car lnos))))))
+    (if (length> prologue-lnos 1)
+        ; First of two line numbers in prologue is function definition location
+        ; which should be removed from the body.
+        (let loop ((stmts body))
+          (if (eq? functionloc (cadr stmts))
+              (set-cdr! stmts (cddr stmts))
```
c42f: I felt like using `set-cdr!` here wasn't great. Is there a more natural idiomatic way to do this?

c42f: (Taking into account that we'll need to return both the line number and the body if this is pure.)

c42f:
Ok, so this should be very close. There's one remaining problem: code coverage has changed because it now includes the line of the function definition. It turns out this is because the line from the `jl_method_t` is used in codegen for coverage:

https://github.com/JuliaLang/julia/blob/f4c3f1d07474ad0377b47f1535b23c09d0487054/src/codegen.cpp#L5070

https://github.com/JuliaLang/julia/blob/f4c3f1d07474ad0377b47f1535b23c09d0487054/src/codegen.cpp#L5910

I'm not sure what the desired outcome would be here — is it useful to consider the function definition line as code which should be included in the coverage output, or not?

There's also other uses of `toplineno` related to debug info. From what I can see these relate to emitting debug info for the function and function arguments so using the true function line number should be fine for these I guess.

c42f:
To record a discussion about coverage with Jeff and Keno - the consensus was that we shouldn't visit the function definition line (unless it contains executable code), so some small adjustments to coverage codegen will be needed here.

c42f:
Some code archeology on the special case handling of `toplineno` in coverage generation:

* It was needed due to https://github.com/JuliaLang/julia/commit/bd06d27c001cd1c0a9d118c3614c74d33b406cb6 (maybe the change at line https://github.com/JuliaLang/julia/blob/bd06d27c001cd1c0a9d118c3614c74d33b406cb6/src/alloc.c#L560)
* Which led to the bug https://github.com/JuliaLang/julia/issues/17442
* Which was fixed in https://github.com/JuliaLang/julia/pull/17470

I think the code which the `toplineno` workaround related to has since been replace by the line table and having linearization as part of lowering, so should be possible to revert https://github.com/JuliaLang/julia/pull/17470 now.

src/codegen.cpp
```diff
@@ -5906,15 +5909,6 @@ static std::pair<std::unique_ptr<Module>, jl_llvm_functions_t>
...
 
-    if (do_coverage(mod_is_user_mod)) {
-        coverageVisitLine(ctx, ctx.file, toplineno);
-        if (linetable.size() >= 2) {
-            // avoid double-counting the entry line
-            const auto &info = linetable.at(1);
-            if (info.file == ctx.file && info.line == toplineno && info.is_user_code == mod_is_user_mod)
-                current_lineinfo.push_back(1);
-        }
-    }
```
c42f, line 5917: I've removed this coverage stuff for the reasons explained in the commit message: It seems like a workaround for an old bug #17442, and now we have this double-counting workaround which appears to be a workaround-workaround. So I just deleted this. I also found `coverageVisitStmt` a bit confusing so I added some comments above, which I hope correctly state what is going on.

PR #35139, opened 2020-03-17
Additional GC.@ preserve documentation

c42f:
Here's some additional documentation for `GC.@preserve`, taking into account the discussion on discourse at https://discourse.julialang.org/t/on-the-garbage-collection/35695/31

CC @yuyichao

### DIFF
--- a/base/gcutils.jl
+++ b/base/gcutils.jl
@@ -94,11 +94,50 @@ enable(on::Bool) = ccall(:jl_gc_enable, Int32, (Int32,), on) != 0
 """
     GC.@preserve x1 x2 ... xn expr
 
-Temporarily protect the given objects from being garbage collected, even if they would
-otherwise be unreferenced.
+Mark the objects `x1, x2, ...` as being *in use* during the evaluation of the
+expression `expr`. This is only required in unsafe code where `expr`
+*implicitly uses* memory or other resources owned by one of the `x`s.
 
-The last argument is the expression during which the object(s) will be preserved.
-The previous arguments are the objects to preserve.
+*Implicit use* of `x` covers any indirect use of resources logically owned by
+`x` which the compiler cannot see. Some examples:
+* Accessing memory of an object directly via a `Ptr`
+* Passing a pointer to `x` to `ccall`
+* Using resources of `x` which would be cleaned up in the finalizer.
+
+`@preserve` should generally not have any performance impact in typical use
+cases where it briefly extends object lifetime. In implementation, `@preserve`
+has effects such as protecting dynamically allocated objects from garbage
+collection.
+
+# Examples
+
+When loading from a pointer with `unsafe_load`, the underlying object is
+implicitly used, for example `x` is implicitly used by `unsafe_load(p)` in the
+following:
+
+```jldoctest
+julia> let
+           x = Ref{Int}(101)
+           p = Base.unsafe_convert(Ptr{Int}, x)
+           GC.@preserve x unsafe_load(p)
+       end
+101
+```
+
+When passing pointers to `ccall`, the pointed-to object is implicitly used and
+should be preserved. (Note however that you should normally just pass `x`
+directly to `ccall` which counts as an explicit use.)
+
+```jldoctest
+julia> let
+           x = "Hello"
+           p = pointer(x)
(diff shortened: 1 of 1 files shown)


base/gcutils.jl
```diff
@@ -94,11 +94,36 @@ enable(on::Bool) = ccall(:jl_gc_enable, Int32, (Int32,), on) != 0
...
-otherwise be unreferenced.
+Mark the objects `x1, x2, ...` as being *in use* during the evaluation of the
+expression `expr`. This is only required in unsafe code where `expr`
+*implicitly uses* memory or other resources owned by one of the `x`s.
+
+*Implicit use* of `x` covers any indirect use of resources logically owned by
+`x` which the compiler cannot see. Some examples:
+* Accessing memory of an object directly via a `Ptr`
+* Passing a pointer to `x` to `ccall` when the pointer is computed outside the
+  normal `cconvert` argument conversion pipeline.
```
yuyichao: `cconvert` -> `ccall`. It's not what function you use to do the conversion that matters here, if you call `cconvert` and `unsafe_convert` it won't help you preserving the object. It's the fact that it is done by the automatic mechaism of `ccall` that matters.

----

Also if you are calling `cconvert` and `unsafe_convert` it is the result of `cconvert` that should be preserved though that's a fairly independent issue.

c42f: Yes. I'm trying to somehow briefly say that `ccall` preserves the objects which are explicitly passed to it, but if they implicitly use some other object then those implicitly used objects must be `@preserve`d.

But without rehashing the documentation for `ccall` too much.

I've reworked this by shortening this part and adding more detail in an example.

yuyichao:
LGTM

base/gcutils.jl
rryi: I find the enhanced docstring well written and understandable. 

Maybe a drastic counterexample could raise programmer's attention, like the following. If you think it is helpful and not too long for a docstring, feel free to use or modify my suggestion:

Do not try to "optimize away" GC.@preserve because you think x is valid when a pointer to memory of x is dereferenced. Example:


```
function needs_preserve(s::String, t::String)
    do_some_allocations() # might cause garbage collection
    x = s * t * "ensure length > 0"
    p = pointer(x) # x is explicitly referenced and alive here
    c = unsafe_load(p) # x is implicitly used, not transparent to compiler
    # correct code makes use of x explicit: c = GC.preserve x unsafe_load(p)
    show(x) # x is explicitly referenced and alive here
    c
end
```

The code sequence suggests that x is valid before and after unsafe_load(p). But: without protecting pointer use by GC.@preserve, nothing prevents the compiler from reordering that code to:


```
function needs_preserve_reordered_by_compiler(s::String, t::String)
    x = s * t * "ensure length > 0"
    p = pointer(x) # x is explicitly referenced and alive here
    show(x) # after this, x is no longer referenced and available to garbage collection
    do_some_allocations() # might cause garbage collection of x
    unsafe_load(p) # possibly invalid result or even program crash by segfault
end
```

yuyichao: Reordering really isn't the issue though. There are a lot of things that actually prevent the reordering.

yuyichao: In particular, it's basically impossible to reorder anything that needs `@preserve` across anything that can trigger GC as far as the user is concerned, and you can actually rely on that if you really want to and does so correctly (I'm not sure if there's a way to do that though). However, that does **not** make the first case valid.

rryi: Ok - if there is a chance to prove my example is wrong, forget about it.

I love performance optimization, and the opportunity to avoid GC.@preserve is a temptation. However, I know about the (day-and-)nightmare of 5000 customers again and again complaining about production crashes and data loss, and the hectics to find any clue to what happened in GBytes of production logfiles. Safety first. I will follow the recommmendation to put any pointer dereferencing into a GC.@preserve block. 

However, if you come up with a bulletproof rule when to safely skip GC.@preserve, let me know ;-)

vtjnash: It's safe to skip (and can emit better code) if you avoid using pointers entirely. It's the `pointer` usages that have the performance risks, the `GC.@preserve` has no cost of its own.

c42f:
Thanks for the review!

PR #35142, opened 2020-03-17
support missing in radians-degrees conversions

aplavin:


KristofferC (changes requested):
https://github.com/JuliaLang/julia/pull/35066 probably needs to be discussed first

aplavin:
Ok, I just thought this is an accidental omission, i.e. someone just forgot to add these functions to the list. For now the choice of which math functions propagate missing and which don't seems arbitrary for me.

KristofferC:
> For now the choice of which math functions propagate missing and which don't seems arbitrary for me.

Indeed, that is the crux of it.

JeffBezanson:
A possible solution here is just to remove the type restriction from
```
rad2deg(z::Number) = (z/pi)*180
deg2rad(z::Number) = (z*pi)/180
```

Issue #35131, opened 2020-03-16
Specialization despite nospecialize annotation


timholy:
Using [`nspecializations` from here](https://gist.github.com/timholy/cb4e2afe15e9841a6a2ab4d90325cddc), I find the following behavior strange:
```julia
julia> function abstractparameters(T::Type, AT::Type)
           @nospecialize T AT
           @assert T <: AT
           Tst = supertype(T)
           while Tst <: AT
               T = Tst
               Tst = supertype(T)
           end
           return T.parameters
       end
abstractparameters (generic function with 1 method)

julia> abstractparameters(typeof(rand(3)), AbstractArray)
svec(Float64, 1)

julia> abstractparameters(typeof(rand(Float32, 3)), AbstractArray)
svec(Float32, 1)

julia> abstractparameters(typeof(rand(Int, 3)), AbstractArray)
svec(Int64, 1)

julia> m = first(methods(abstractparameters))
abstractparameters(T::Type, AT::Type) in Main at REPL[3]:2

julia> nspecializations(m)   # this is what I expect
1

julia> @generated function f(a::AbstractArray)
           T, N = abstractparameters(a, AbstractArray)    # call it from a @generated method
           :($T(1))
       end
f (generic function with 1 method)

julia> f(rand(3))
1.0

julia> f(rand(Float32, 3))
1.0f0

julia> nspecializations(m)  # this is what I don't want
2

julia> m.specializations
Core.TypeMapEntry(Core.TypeMapEntry(nothing, Tuple{typeof(abstractparameters),Type,Type}, nothing, svec(), 0x0000000000000001, 0xffffffffffffffff, MethodInstance for abstractparameters(::Type, ::Type), false, false, false), Tuple{typeof(abstractparameters),Type,Type{AbstractArray}}, nothing, svec(), 0x0000000000000001, 0xffffffffffffffff, MethodInstance for abstractparameters(::Type, ::Type{AbstractArray}), false, false, false)
```

This is a MWE of a bigger problem: in LoopVectorization, I've measured at least 33 specializations of `abstractparameters` despite the fact that both arguments are marked as not being specialiizable. In that case there are specializations on both parameters. Here's 30 of them:
```julia
julia> m.specializations.arg1[1].targ
30-element Array{Any,1}:
[...]

PR #35144, opened 2020-03-17
speed up llvmcall unique name generation

JeffBezanson:
This speeds up codegen of functions that use many duplicate llvmcalls, such as in LoopVectorization/SIMDPirates. From issue #35131

For me this takes the time to run the LoopVectorization tests from 5'40" to 5'15" (note this is a somewhat slow machine).

### DIFF
--- a/src/ccall.cpp
+++ b/src/ccall.cpp
@@ -881,7 +881,6 @@ static jl_cgval_t emit_llvmcall(jl_codectx_t &ctx, jl_value_t **args, size_t nar
     }
     if (at == NULL)
         at = try_eval(ctx, args[3], "error statically evaluating llvmcall argument tuple");
-    int i = 1;
     if (jl_is_tuple(ir)) {
         // if the IR is a tuple, we expect (declarations, ir)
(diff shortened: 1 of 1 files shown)


src/ccall.cpp
```diff
@@ -937,10 +936,11 @@ static jl_cgval_t emit_llvmcall(jl_codectx_t &ctx, jl_value_t **args, size_t nar
     Type *rettype = julia_type_to_llvm(ctx, rtt, &retboxed);
     if (isString) {
         // Make sure to find a unique name
+        static int llvmcallUnique = 1;
         std::string ir_name;
         while (true) {
             std::stringstream name;
-            name << (ctx.f->getName().str()) << "u" << i++;
+            name << (ctx.f->getName().str()) << "u" << llvmcallUnique++;
```
vtjnash:
```suggestion
            name << (ctx.f->getName().str()) << "u" << globalUnique++;
```

vtjnash (commented):
This variable needs to go away (soon), but might as well use the existing name, so that we can grep for it.

PR #35151, opened 2020-03-18
define `keys` for `Base.Generator`

KlausC:
IMO `findmin(sin(x) for x in 0:10)` should return same as `findmin([sin(x) for x in 0:10])`,
without allocating. Currently the first call throws a `MethodError`.

This PR inserts `key(::Base.Generator)` in order to achieve that.
As a consequence also `pairs(::Generator)`, `findmin(::Genearator)`, `findmax(::Generator)` are working as expected.

```
julia> @btime findmin(sin(x) for x = 1:10^6)
  21.901 ms (0 allocations: 0 bytes)
(-0.9999999999848337, 52174)

julia> @btime findmin([sin(x) for x = 1:10^6])
  25.945 ms (2 allocations: 7.63 MiB)
(-0.9999999999848337, 52174)
```

### DIFF
--- a/base/generator.jl
+++ b/base/generator.jl
@@ -51,6 +51,7 @@ length(g::Generator) = length(g.iter)
 size(g::Generator) = size(g.iter)
 axes(g::Generator) = axes(g.iter)
 ndims(g::Generator) = ndims(g.iter)
+keys(g::Generator) = OneTo(length(g))
 
 
 ## iterator traits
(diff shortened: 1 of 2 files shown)


fredrikekre:
See  #34678 (which actually has another definition of `keys(::Generator)`).

KlausC:
I retire this one in favour of  #34678 . Sorry for the inconvenience.

PR #35183, opened 2020-03-20
trying to debug Win32 loading test failures

StefanKarpinski:


StefanKarpinski:
I'm going to go ahead and merge this so that if we do see more Win32 failures here we have debug output to see what's happening.

Keno:
I don't think this is Linux-specific and it doesn't seem to be duplicate UUIDs either. I ran the test a few thousand times overnight on linux and managed to get a failure. Unfortunately, even though I printed the seed, I wasn't able to reproduce it. Will try again.

Keno:
Saw it again, though again the seed didn't help reproduce, so either I missed a source of randomness, or it's dependent on something else. I'm gonna run it again, but under rr this time.

Keno:
RR dump of the failure: https://julialang-dumps.s3.amazonaws.com/rr/julia-loading.tar.gz

PR #31223, opened 2019-03-01
RFC/WIP: creation of mapdict method to modify Dict values

ndinsmore:
This is the initial pass at the inclusion of  methods `mapdict` & `mapdict!` which allows the modification of Dict values without the overhead of using the Dict keys to access & store the values.  

The need for this method was discuss at:
https://discourse.julialang.org/t/fast-dict-value-modification-by-accessing-dict-vals
At the recommendation of @tpapp this was built out for base.

An example of the use would be as follows:
```julia
julia> D=Dict(:a=>1,:b=>2)
Dict{Symbol,Int64} with 2 entries:
  :a => 1
  :b => 2

julia> mapdict!(v->v-1,D)
Dict{Symbol,Int64} with 2 entries:
  :a => 0
  :b => 1

julia> D=mapdict!(v->Float64(v),D) #For the type mutating case you must use an =
Dict{Symbol,Float64} with 2 entries:
  :a => 0.0
  :b => 1.0
```

Currently this implementation handles most of the function cases that were brought up:
1.) Functions which return the same type as the input
2.) Functions that are mutate the value type but are typesafe
3.) Functions which are not type safe.

While the function is currently only implemented for `Dict`, it would be simple to add a naive fallback for `AbstractDict` using keys.

If there is interest in including this in Base, test would also be built out.

Benchmarking results for essentially ```mapdict!(v-> v > row ? v-1 : v,D)``` where row is set to change half the values.

image

PR #35184, opened 2020-03-20
add compat annotation for map!(f, values(dict)) (from #31223)

rfourquet:
This also didn't get a NEWS.md item, but this was for Julia 1.2... shall we add news now?


StefanKarpinski:
Maybe add it to NEWS for that branch and HISTORY on master?

fredrikekre (approved):

rfourquet:
> Maybe add it to NEWS for that branch and HISTORY on master?

Sounds good, will do.

rfourquet:
> add it to NEWS for that branch

If that's fine, I will push directly the change to that branch when this PR is merged, to avoid opening another PR (I'm not familiar how we work with release branches).

Issue #35199, opened 2020-03-21
reduced_indices is type unstable


ranocha:
Currently, `reduced_indices` is type unstable. If the branch `elseif 1 < d <= N` is used, this results in a significant delay compared to the other branches. Would it be acceptable to add a fix for some common cases such as two-dimensional `Array`s?

```julia
julia> using BenchmarkTools

julia> M = [1 2; 3 4]
2×2 Array{Int64,2}:
 1  2
 3  4

julia> for dims in 1:3
           display(@benchmark sum($M, dims=$dims))
       end
BenchmarkTools.Trial: 
  memory estimate:  96 bytes
  allocs estimate:  1
  --------------
  minimum time:     37.287 ns (0.00% GC)
  median time:      38.808 ns (0.00% GC)
  mean time:        42.092 ns (3.48% GC)
  maximum time:     1.135 μs (94.90% GC)
  --------------
  samples:          10000
  evals/sample:     992
BenchmarkTools.Trial: 
  memory estimate:  288 bytes
  allocs estimate:  7
  --------------
  minimum time:     417.427 ns (0.00% GC)
  median time:      433.241 ns (0.00% GC)
  mean time:        456.827 ns (2.17% GC)
  maximum time:     9.682 μs (93.91% GC)
  --------------
  samples:          10000
  evals/sample:     199
BenchmarkTools.Trial: 
  memory estimate:  112 bytes
  allocs estimate:  1
  --------------
  minimum time:     39.271 ns (0.00% GC)
  median time:      40.817 ns (0.00% GC)
  mean time:        44.397 ns (2.75% GC)
  maximum time:     826.044 ns (92.87% GC)
  --------------
  samples:          10000
  evals/sample:     991

julia> function Base.reduced_indices(inds::Base.Indices{2}, d::Int)
           d < 1 && throw(ArgumentError("dimension must be ≥ 1, got $d"))
           if d == 1
               return (Base.reduced_index(inds[1]), Base.tail(inds)...)::typeof(inds)
           elseif d == 2
               return tuple(inds[1], oftype(inds[d], Base.reduced_index(inds[d])))::typeof(inds)
           else
               return inds
           end
       end

julia> for dims in 1:3
           display(@benchmark sum($M, dims=$dims))
       end
BenchmarkTools.Trial: 
  memory estimate:  96 bytes
  allocs estimate:  1
  --------------
  minimum time:     35.162 ns (0.00% GC)
  median time:      36.176 ns (0.00% GC)
[...]

PR #35200, opened 2020-03-21
Make sum along dims of SymTridiagonal, Tridiagonal, Bidiagonal, Diagonal faster

ranocha:
As discussed in JuliaLang/LinearAlgebra.jl#692, the return type of a `sum` along a dimension of a `SymTridiagonal`/`Tridiagonal`/`Bidiagonal`/`Diagonal` matrix should be a dense matrix and not a sparse matrix. Additionally, I've added some specialized methods for the `sum` along a dimension of such a structured matrix.

Now, I get the following timings on my notebook for the test problem reported in JuliaLang/LinearAlgebra.jl#692.
```julia
julia> using LinearAlgebra, BenchmarkTools

julia> N = 1000;

julia> T = Tridiagonal(ones(N-1), ones(N), ones(N-1));

julia> @benchmark sum($T, dims=$1)
BenchmarkTools.Trial: 
  memory estimate:  7.94 KiB
  allocs estimate:  1
  --------------
  minimum time:     638.978 ns (0.00% GC)
  median time:      1.213 μs (0.00% GC)
  mean time:        1.404 μs (6.90% GC)
  maximum time:     18.222 μs (78.77% GC)
  --------------
  samples:          10000
  evals/sample:     92

julia> @benchmark sum($T, dims=$2)
BenchmarkTools.Trial: 
  memory estimate:  8.13 KiB
  allocs estimate:  7
  --------------
  minimum time:     952.500 ns (0.00% GC)
  median time:      1.501 μs (0.00% GC)
  mean time:        1.649 μs (6.04% GC)
  maximum time:     80.741 μs (92.46% GC)
  --------------
  samples:          10000
  evals/sample:     10

julia> @benchmark sum($T', dims=$1)
BenchmarkTools.Trial: 
  memory estimate:  7.98 KiB
  allocs estimate:  2
  --------------
  minimum time:     694.281 ns (0.00% GC)
  median time:      1.178 μs (0.00% GC)
  mean time:        1.308 μs (6.22% GC)
  maximum time:     29.076 μs (84.47% GC)
  --------------
  samples:          10000
  evals/sample:     57
```
These timings are three to four orders of magnitude lower than the ones reported in JuliaLang/LinearAlgebra.jl#692.

The difference between `dims=1` and `dims=2` is caused by JuliaLang/julia#35199:
```julia
ulia> using LinearAlgebra, BenchmarkTools

julia> N = 1000;

julia> T = Tridiagonal(ones(N-1), ones(N), ones(N-1));

julia> @benchmark Base.reduced_indices($(axes(T)), $1)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     0.018 ns (0.00% GC)
  median time:      0.021 ns (0.00% GC)
  mean time:        0.023 ns (0.00% GC)
  maximum time:     16.655 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000

julia> @benchmark Base.reduced_indices($(axes(T)), $2)
BenchmarkTools.Trial: 
  memory estimate:  192 bytes
  allocs estimate:  6
  --------------
  minimum time:     263.777 ns (0.00% GC)
  median time:      271.880 ns (0.00% GC)
  mean time:        291.368 ns (2.19% GC)
  maximum time:     6.761 μs (94.72% GC)
  --------------
  samples:          10000
  evals/sample:     346
```


dkarrasch (commented):
Which role does `init` play in general? Is it something like an initialization value? If so, then I guess basically all assignments to `res` should be `+=` assignments, shouldn't they?

ranocha:
Yes, `init` is an initialization value in general. Concerning the assignments to `res`, these could be `+=` assignments, since the initialization value is `zero(eltype(A))`, but that wouldn't change the final result (unless `x + zero(x) != x`, which seems to break the contract of `zero`).
I wrote `res[i] = ` instead of `res[i] +=` to make the purpose of `sum` more clear. Shall I change all assignments?

dkarrasch:
That's why I was asking about the purpose of `init`. Maybe I was confused by ` reduce(op, itr; [init])`, where you can set arbitrary `init`ial values. `sum` appears to have no such option, so if `init` is some sort of `zero` of the appropriate type, then the current PR is good as is.

ranocha:
Yes, that's right. For `sum`, there is no possibility to choose `init` different from `zero` of the appropriate type.

stdlib/LinearAlgebra/src/bidiag.jl
```diff
@@ -189,6 +189,10 @@ similar(B::Bidiagonal, ::Type{T}) where {T} = Bidiagonal(similar(B.dv, T), simil
 # The method below is moved to SparseArrays for now
 # similar(B::Bidiagonal, ::Type{T}, dims::Union{Dims{1},Dims{2}}) where {T} = spzeros(T, dims...)
 
+# Usually, reducedim_initarray calls similar, which yields a sparse matrix for a
+# Bidiagonal matrix. However, reducedim should yield a dense vector to increase performance.
+Base.reducedim_initarray(A::Bidiagonal, region, init) = fill(init, Base.reduced_indices(A,region))
```
tkf: It seems suboptimal to repeat similar definitions and comments. How about doing this at a single location with a single `Union`?

tkf:
> `reduce(op, itr; [init])`, where you can set arbitrary `init`ial values

Nit-picking (sorry), but this is not true. `init` must be the identity of `op`. The reason why "it works" is that `reduce` dispatches to `foldl` (where `init` can be anything).

---

But I don't think it matters here.  You are using `Base.reducedim_initarray` you just defined in `Base._sum` with explicit `init` (`zero(eltype(A))`). So, I think it is fine as long as the definition of `Base._sum` is correct.

If you want to be careful maybe you can invoke `mapreduce` with `dims` directly in the test to check other paths that calls `reducedim_initarray`.

dkarrasch:
True. I overlooked both the comment in the docstring of `reduce` and the call of `Base.reducedim_initarray` with `zero(eltype(A))`. Sorry! I'll better leave review of this PR to others. 🤣

stdlib/LinearAlgebra/src/bidiag.jl
ranocha: Okay, I'll do that. I've coded it in this way to have the appropriate methods for each type in the file for that type.

ranocha:
> If you want to be careful maybe you can invoke `mapreduce` with `dims` directly in the test to check other paths that calls `reducedim_initarray`.

Thanks for this suggestion, @tkf. After doing this, I realized that I need to change another method of `reducedim_initarray` to make it work also for `max` and `min` instead of `+`.

stdlib/LinearAlgebra/src/special.jl
```diff
@@ -3,6 +3,12 @@
 # Methods operating on different special matrix types
 
 
+# Usually, reducedim_initarray calls similar, which yields a sparse matrix for a
+# Diagonal/Bidiagonal/Tridiagonal/SymTridiagonal matrix. However, reducedim should
+# yield a dense vector to increase performance.
+Base.reducedim_initarray(A::Union{Diagonal,Bidiagonal,Tridiagonal,SymTridiagonal}, region, init, ::Type{R}) where {R} = fill(convert(R, init), Base.reduced_indices(A,region))
```
dkarrasch: Now I'm confused. None of your new `_sum` methods calls this 4-arg method...

ranocha: The method in https://github.com/JuliaLang/julia/blob/8617860b2cf3604b1ff365fa5f3f5d00f3ca47cc/base/reducedim.jl#L93 redirects the calls in my new `_sum` methods.

stdlib/LinearAlgebra/src/tridiag.jl
```diff
@@ -723,6 +723,62 @@ end
...
+    end
+    res
+end
+
+function Base._sum(A::SymTridiagonal, dims::Integer)
+    res = Base.reducedim_initarray(A, dims, zero(eltype(A)))
+    n = length(A.dv)
+    @inbounds begin
+        if dims == 1
+            res[1] = A.ev[1] + A.dv[1]
```
tkf: I guess this can cause segfault for cases like `SymTridiagonal(Int[], Int[])` and `SymTridiagonal([1], Int[])`? Apparently they are constructable...  I guess you need special handling for those cases?

Similar caution may be required for other matrix types.

ranocha: You're right, thanks for the hint. I will add additional checks in the new methods of `_sum`.

ranocha:
The failure of `buildbot/package_macos64` seems to be unrelated to the changes I've made.

stdlib/LinearAlgebra/src/tridiag.jl
ranocha: I also had to add new checks in some conversion methods to `Matrix`. Interestingly, I didn't get errors while running the tests locally in the REPL. Is there a way to disable `@inbounds` (i.e. enforce bounds checking everywhere) in the REPL?

tkf: AFAIK you need to restart `julia` with `--check-bounds=yes`.

ranocha: Okay, thanks.

tkf (approved):
I can't merge this but it LGTM :)

PR #35202, opened 2020-03-21
fix #35201, syntax error with named splatting inside function call

### LINKED ISSUE #35201: Parsing problem with nested splatting named tuple constructor in a function call
In the code below definitions `f1` and `f2` should be equivalent I think.
However, `f1` is a syntax error, while `f2` works as expected.

```
julia> h(x; k=1) = (x, k)
h (generic function with 1 method)

julia> f1(c) = h((;c...), k=true)
ERROR: syntax: "..." expression outside call
Stacktrace:
 [1] top-level scope at REPL[8]:1

julia> f2(c) = (nt=(;c...); h(nt, k=true))
f2 (generic function with 1 method)

julia> f2(Dict(:a=>1, :b=>2))
((a = 1, b = 2), true)
```

It seems that a parser does not correctly recognize that `...` in this case is used in a nested `NamedTuple` constructor call within a call to `h`.

The problem seems to be present only if the call involves keyword arguments, so e.g.
```
julia> f1(c) = h((;c...), true)
f1 (generic function with 1 method)
```
parses correctly.

Also
```
julia> f1(c) = h(;c..., k=true)
f1 (generic function with 1 method)
```
parses, but it gives a different definition of course.

JeffBezanson:
fix #35201 
I believe this was a regression in v1.2.

### DIFF
--- a/src/julia-syntax.scm
+++ b/src/julia-syntax.scm
@@ -1418,6 +1418,8 @@
                  `(,(car x) ,(arg-to-temp (cadr x))))
                 ((or (eq? (car x) 'kw) (and tup (eq? (car x) '=)))
                  `(,(car x) ,(cadr x) ,(arg-to-temp (caddr x))))
+                ((eq? (car x) 'parameters)
+                 `(parameters ,@(map arg-to-temp (cdr x))))
                 ((eq? (car x) 'tuple)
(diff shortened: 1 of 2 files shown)


bkamins:
You are fast :). Thank you!

Issue #35206, opened 2020-03-22
PermutedDimsArray, wrong results on strides method


francescoalemanno:
Hi guys this behaviour of the strides method caused a bug in some of my code that works perfect otherwise,
I strongly believe this to be a bug, since the strides provided as (1000,10,1) are pretty useless in generic array code.
```julia
A=rand(10,100,300);
B=PermutedDimsArray(A,(3,2,1));
@show strides(A) #### strides(A) = (1, 10, 1000)
@show strides(B) #### strides(B) = (1000, 10, 1) ### WRONG (supposedly)
@show strides(collect(B)) #### strides(collect(B)) = (1, 300, 30000) ## CORRECT
```

happening on
```julia
julia> versioninfo()
Julia Version 1.3.1
Commit 2d5741174c (2019-12-30 21:36 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Core(TM) i7-8565U CPU @ 1.80GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.1 (ORCJIT, skylake)

```

Quick fix if indeed bug @ https://github.com/JuliaLang/julia/pull/35208

francescoalemanno:
https://github.com/JuliaLang/julia/pull/35208   is closed,
the strides function was always supposed to care only about raw memory ordering.

PR #35208, opened 2020-03-22
Fix issue #35206: PermutedDimsArray, wrong results on strides method

francescoalemanno:
this PR is meant as a solution to
https://github.com/JuliaLang/julia/issues/35206

before PR:
```julia
A=rand(10,100,300);
B=PermutedDimsArray(A,(3,2,1));
@show strides(A) #### strides(A) = (1, 10, 1000)
@show strides(B) #### strides(B) = (1000, 10, 1) ### WRONG (supposedly)
@show strides(collect(B)) #### strides(collect(B)) = (1, 300, 30000) ## CORRECT
```

after PR:

```julia
A=rand(10,100,300);
B=PermutedDimsArray(A,(3,2,1));
@show strides(A) #### strides(A) = (1, 10, 1000)
@show strides(B) #### strides(B) = (1, 300, 30000) ## CORRECT
@show strides(collect(B)) #### strides(collect(B)) = (1, 300, 30000) ## CORRECT
```

@timholy the original method is yours i think, if i got something wrong, apologies

### DIFF
--- a/base/permuteddimsarray.jl
+++ b/base/permuteddimsarray.jl
@@ -62,13 +62,16 @@ Base.pointer(A::PermutedDimsArray, i::Integer) = throw(ArgumentError("pointer(A,
     (1,t[1].*c_strides(Base.tail(t))...)
 end
 
+@inline function c_strides(t::NTuple{1,Int})
+    1
+end
+
 @inline function c_strides(t::NTuple{0,Int})
     1
 end
 
 function Base.strides(A::PermutedDimsArray{T,N,perm}) where {T,N,perm}
-    strides_trail=c_strides(size(A))
-    ntuple(i->strides_trail[i],Val(N))
+    c_strides(size(A))
 end
 
 @inline function Base.getindex(A::PermutedDimsArray{T,N,perm,iperm}, I::Vararg{Int,N}) where {T,N,perm,iperm}
--- a/base/permuteddimsarray.jl
+++ b/base/permuteddimsarray.jl
@@ -58,9 +58,17 @@ Base.unsafe_convert(::Type{Ptr{T}}, A::PermutedDimsArray{T}) where {T} = Base.un
 # or a linear index?
 Base.pointer(A::PermutedDimsArray, i::Integer) = throw(ArgumentError("pointer(A, i) is deliberately unsupported for PermutedDimsArray"))
 
+@inline function c_strides(t::NTuple{N,Int}) where {N}
+    (1,t[1].*c_strides(Base.tail(t))...)
+end
+
+@inline function c_strides(t::NTuple{0,Int})
+    1
+end
+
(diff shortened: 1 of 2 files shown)


chriselrod:
I don't know what semantics `stride` is supposed to guarantee, but in the way I've been using it in `LoopVectorization`, `strides(B)` returns the answer that will make the library work, while `strides(collect(B))` will actually cause it to crash (unless you also replace `B` itself with `collect(B)`.

That is because the strides are supposed to be how far apart the elements are in that axis.
```julia
julia> A=rand(10,100,300);

julia> B=PermutedDimsArray(A,(3,2,1));

julia> unsafe_load(pointer(A), 1 + sum(((1,2,3) .- 1) .* strides(A)))
0.105212138004664

julia> A[1,2,3]
0.105212138004664

julia> unsafe_load(pointer(B), 1 + sum(((1,2,3) .- 1) .* strides(B)))
0.6621377953848049

julia> B[1,2,3]
0.6621377953848049
```

francescoalemanno:
```julia
julia> A=rand(10,10,10);

julia> A[7,6,3]
0.5416549391900454

julia> A[1 + sum(((7,6,3) .- 1) .* strides(A))]
0.5416549391900454

julia> B=PermutedDimsArray(A,(3,2,1));

julia> B[1 + sum(((3,6,7) .- 1) .* strides(A))]
0.5416549391900454

julia> B[1 + sum(((3,6,7) .- 1) .* strides(B))]
0.9369459472602091
```

chriselrod:
That's a good point.
On the one hand, I find it strange that the linear getindex isn't in memory order, but I can see the case for it being in `vec(B)` order.

If this gets changed from returning memory-strides to returning linear indexing strides, I would want another convenient function that returns memory-strides.

If your code relies on this, how do you handle `SubArray`s (i.e. `view`s)?

francescoalemanno:
i spent today modifying all the code that relied on `strides`  in order to use the `CartesianInd(ex/ices)` machinery without loss of performance and a great good gain in code safety.
So this issue does not affect me directly anymore, anyway there some inconsistency laying around this behaviours, but i'm not the right person to decide between:
* fix `strides`
* fix `geindex/setindex!`
* leave it as it is and add methods for "linear indexing strides"

so i guess this leaves this PR in stall, until further inputs from more experienced julia devs.

timholy:
`strides` is all about memory ordering, not linear indexing. `A[1 + sum(((7,6,3) .- 1) .* strides(A))]` is definitely not guaranteed to work. The docstring is
```
strides(A)

  Return a tuple of the memory strides in each dimension.
```
If you can think of a way to make it clearer I am certainly open to changes. See also https://docs.julialang.org/en/latest/manual/interfaces/#man-interface-strided-arrays-1.

francescoalemanno:
Thank you @timholy the doc string was indeed clear about the meaning... my bad

PR #35209, opened 2020-03-22
Replace conditional data dump by a triage function in choosetests.jl

o314:


test/choosetests.jl
```diff
@@ -82,60 +82,29 @@ function choosetests(choices = [])
...
-
-    # do ambiguous first to avoid failing if ambiguities are introduced by other tests
-    if "ambiguous" in skip_tests
-        filter!(x -> x != "ambiguous", tests)
-    elseif "ambiguous" in tests
-        filter!(x -> x != "ambiguous", tests)
-        prepend!(tests, ["ambiguous"])
-    end
-
+    function patchlib!(skip_tests,tests,name,files=[name]; stdlib::Bool=false)
```
rfourquet: Few relatively minor suggestions:
* put `tests` as first argument, as this is the variable being mutated
* maybe remove `skip_tests` ? this can use the eponyme variable from the enclosing scope
* put spaces after commas

test/choosetests.jl
```diff
@@ -82,60 +82,29 @@ function choosetests(choices = [])
...
+       flt = x -> (x != name && !(x in files))
+       testsflt = stdlib ? flt : !=(name)
```
rfourquet: I would believe you can just delete `testsflt`, and that using (the equivalent of) `!=(name)` in most code above was an oversight, as it would be unexpected from the user to specify both the `name` and the `files`. Otherwise, some file can end up appearing twice in `tests` and I don't think it can be useful. Would be good if someone can confirm (maybe @vtjnash?)

test/choosetests.jl
```diff
@@ -82,60 +82,29 @@ function choosetests(choices = [])
...
+
+   patchlib!(skip_tests,tests,"unicode", ["unicode/utf8"])
+   patchlib!(skip_tests,tests,"strings", ["strings/basic", "strings/search", "strings/util",
+                  "strings/io", "strings/types"])
+   patchlib!(skip_tests,tests,"subarray")
+   patchlib!(skip_tests,tests,"compiler", ["compiler/inference", "compiler/validation",
+       "compiler/ssair", "compiler/irpasses", "compiler/codegen",
+       "compiler/inline", "compiler/contextual"])
+   patchlib!(skip_tests,tests,"stdlib", STDLIBS; stdlib=true)
+   patchlib!(skip_tests,tests,"ambiguous")
```
rfourquet: maybe reproduce here the couple of comments seen before, e.g. `# do ambiguous first to avoid failing if ambiguities are introduced by other tests`

rfourquet (commented):
Very nice, thanks!

test/choosetests.jl
o314: Seems to be resolved now

test/choosetests.jl
o314: I have not change so much code there. May be shall we wait for a confirmation ?

rfourquet: I suggest that you update your PR according to my comment (i.e. use the same `flt` function in both `filter!` calls, and remove the `stdlib` keyword), and then I will merge. I can't think of a good reason to keep this distinction, and I will take responsability for this if anyone complains :)

rfourquet:
Superseded by #35474

PR #35213, opened 2020-03-22
fix #28581 and specialize istriu/istril for structured matrices

### LINKED ISSUE #28581: `Diagonal(::AbstractMatrix)` behavior differs from documentation in certain cases
``` 
  | | |_| | | | (_| |  |  Version 1.1.0-DEV.10 (2018-08-11)
 _/ |\__'_|_|_|\__'_|  |  sparsearrays-triangular-operations/72f777ee96 (fork: 8 commits, 0 days)
|__/                   |
```
Whereas docu states:
```
help?> Diagonal
  Diagonal(A::AbstractMatrix)
  Construct a matrix from the diagonal of A.
```
When applied to a subtype of `AbstractMatrix` we obtain the following results:
```
julia> A = UpperTriangular(sparse([1.0 0;20 0]))
2×2 UpperTriangular{Float64,SparseMatrixCSC{Float64,Int64}}:
 1.0  0.0
  ⋅   0.0

julia> Diagonal(A)
2×2 Diagonal{Float64,SparseVector{Float64,Int64}}:
 1.0   ⋅ 
  ⋅   0.0

julia> A[1,2] = 1
1

julia> Diagonal(A)
ERROR: ArgumentError: matrix cannot be represented as Diagonal
Stacktrace:
 [1] Diagonal(::UpperTriangular{Float64,SparseMatrixCSC{Float64,Int64}}) at /home/julia/julia/usr/share/julia/stdlib/v1.1/LinearAlgebra/src/special.jl:48
 [2] top-level scope at none:0
```
That means, in this case the `Diagonal` is not constructed from the diagonal of `A`, but an error is thrown, if `isdiag(A) == false`.
The same behavior applies for some other subtypes of `AbstractMatrix`, and not for others. 
IMO the check for diagonality should be removed from the constructors, and in all cases 
`Diagonal(A::AbstractMatrix) = Diagonal(diag(A))` should be true.
Alternatively the docu string should warn the user about the error check in some and which cases.

ranocha:
`Diagonal`, `Tridiagonal`, `LowerTriangular`, and `UpperTriangular` promise to have constructors taking an `AbstractMatrix` and returning a structured matrix of the corresponding parts. However, there have been additional checks if the input is structured matrix of another type. This PR fixes this inconsistent behavior. The additional checks are only made if `convert` is used.


ranocha:
I don't know how the failure of buildbot/tester_win32 could be caused by the changes I've made.

Issue #33968, opened 2019-11-28
PDF docs deployment broken


cdluminate:
The PDF document is invalid:

https://raw.githubusercontent.com/JuliaLang/docs.julialang.org/assets/julia-1.3.0.pdf

and the link is found here:

https://docs.julialang.org/en/v1.3/

aerdely:
Still broken as of today 😕

ViralBShah:
Would be nice to fix this. I am happy to do it, if there's instructions I can follow somewhere.

espectalll:
The build script should be the one at [julialang/docs.julialang.org](https://github.com/julialang/docs.julialang.org/). To run it (you might have to install dependencies like PDFLaTeX or Docker, I haven't figured that out), you have to do something like:

```sh
git clone https://github.com/JuliaLang/docs.julialang.org
git clone https://github.com/JuliaLang/julia
cd docs.julialang.org/pdf
JULIA_SOURCE=../../julia julia make.jl releases
```

I decided to modify the `make.jl` script with my own small patches, to save myself some time and effort, adding a `#!/usr/bin/env julia` header and rewriting the following function:

```julia
function collect_versions()
    versions = VersionNumber[]
    push!(versions, v"1.3.0")
    return versions
end
```

...and I got this error

```
┌ Error: LaTeXWriter: failed to compile tex with docker. Logs and partial output can be found in /tmp/jl_yokbBF.
│   exception =
│    failed process: Process(`docker exec -u zeptodoctor latex-container bash -c 'mkdir /home/zeptodoctor/build
│    cd /home/zeptodoctor/build
│    cp -r /mnt/. .
│    latexmk -f -interaction=nonstopmode -view=none -lualatex -shell-escape TheJuliaLanguage.tex
│    '`, ProcessExited(12)) [12]
│    
└ @ Documenter.Writers.LaTeXWriter ~/Projects/julia/doc/deps/packages/Documenter/cMNJA/src/Writers/LaTeXWriter.jl:173
```

espectalll:
(keep in mind: I keep my Julia repo at `~/Projects/julia`, and I couldn't find that `/tmp/jl_yokbBF` anywhere)

fredrikekre:
This is the real issue: https://github.com/JuliaDocs/Documenter.jl/issues/1200

aerdely:
The problem has been fixed!

StefanKarpinski:
[...]

PR #35214, opened 2020-03-22
Disable the pdf link generation in the manual

ViralBShah:
While #33968 is yet to be fixed upstream, we should avoid showing the pdf link in the manual.


fredrikekre:
Fixing it instead.

Issue #34759, opened 2020-02-14
Additional file extensions for deploying julia scripts/apps


rapus95:
"Taking end user deployability seriously" would've been a proper name for it aswell 😛.

I'd like to propose adding two more file extensions to the repertoire of Julia.

Current situation: Extension .JL
That's the extension for the source files of Julia. As it currently is the only file extension all julia files have it and the extension is probably bound to a custom editor preferred by the user.

-----

Case 1: Deploy julia scripts to "clicky" end users (e.g. Windows)
-----
Let them install Julia and send them the script file. Then, just open the REPL & call include or instruct them to pass the file as the first command line argument.

Proposal 1: New Extension .JLR (=julia runnable)
-----
This is just an ordinary julia source file which is bound to be run by the julia executable.
That way we don't need to mess with the binding of .jl files which most people probably bound to an editor of their choice. We'd also be able to ship the __*"run with Julia"*__ by default with all kinds of installers. As such those files would __*naturally mark script-like entry points*__ into running julia programs.

Sidenote: This file extension was used by Juliar language until late 2019 for their source files. Since that language got renamed to Juka they renamed their source files to .JUK, and thus freeing it again.

-----

Case 2: Provide a more standalone feeling similar to that you get from Java
-----
To run a Java program you need to install the JVM and from then on you can run arbitrary .JAR files which mostly are packed and precompiled code repositories together with their dependencies.

Proposal 2: New Extension .JLB (=julia binary)
-----
While technically not a binary it would be very similar to the .JAR format of Java as it'd also hold the necessary data in a packed format and be bound to be run with the runtime executable. For now it would just be an __*archive consisting of the corresponding Julia packages*__ (and maybe a runnable).
The actual binary format can change at any time since its sole use is to work out of the box on any platform which has a Julia installation (call it the JuliaVM). No internet required.
It's some sort of the __*batteries included approach which only relies on a preexisting runtime*__. 

That'd increase the spread of Julia a lot for the similar reason why that works for JAR files. Most users of Java programs aren't programmers. Nevertheless, for a a java program, __*running is dead simple*__ (once the VM installer was run). With that approach we have a very similar dead simple solution which is straight forward to install.

So while keeping to reach for the stars as in true AOT statically compiled shared shared libraries/executable binaries, this approach would be a good intermediate solution for creating conveniently shareable programs in Julia.

KristofferC:
[...]

PR #35222, opened 2020-03-22
Loading: look for vendored packages in `$project/packages`

StefanKarpinski:


DilumAluthge:
I like it.

Can you add "closes https://github.com/JuliaLang/julia/pull/35195" to the original post?

base/loading.jl
```diff
@@ -591,9 +591,15 @@ function explicit_manifest_uuid_path(project_file::String, pkg::PkgId)::Union{No
             return path
         end
         hash === nothing && return nothing
```
DilumAluthge: Is "vendored" the right word here? To me, a vendored dependency is what Pkg does with TOML.

I would say what we are doing here is "bundled" dependencies.

base/loading.jl
```diff
@@ -591,9 +591,15 @@ function explicit_manifest_uuid_path(project_file::String, pkg::PkgId)::Union{No
...
+        # Look for vendored packages in `$project/packages`
```
DilumAluthge:
```suggestion
        # Look for bundled packages in `$project/packages`
```

DilumAluthge (changes requested):

base/loading.jl
DilumAluthge: Is "vendored" the right word here? To me, a vendored dependency is what Pkg does with TOML.

I would say what we are doing here is "bundled" dependencies.

StefanKarpinski: Sure, that's a good term for it 👍

base/loading.jl
```diff
@@ -591,9 +591,15 @@ function explicit_manifest_uuid_path(project_file::String, pkg::PkgId)::Union{No
             return path
         end
         hash === nothing && return nothing
+        # Look for bundled packages in `$project/packages`
+        depot_path = DEPOT_PATH
+        project = active_project()
+        if project !== nothing
```
fredrikekre:
```suggestion
        if project !== nothing && project in load_path()
```

DilumAluthge: Could you explain this change?

fredrikekre: Seems weird to look for stuff in a path that is not on the LOAD_PATH.

DilumAluthge: Sure. I think this will be true in 99% of cases anyway. The only way that the active project wouldn’t be the first entry in the load path would be if someone manually popped it off.

fredrikekre: Yes, so my suggestion fixes that.

StefanKarpinski: The situation where this can happen is if `@` is removed from the `LOAD_PATH`. I have to think a little about whether it makes sense for that to interact with this or not...

Then this can be simplified to this:
```suggestion
        if project in load_path()
```
That's because `load_path` already filters out `nothing` values and only returns strings.

KristofferC:
How would you get the packages into `packages`? You need to set `DEPOT_PATH[1]` to the package directory, remove the other depots and instantiate, then remove the `registry` and `artifact` directories and then the leftover `packags` directory works as an implicit package depot for the specific project?

What is the use case for this?

DilumAluthge:
> How would you get the packages into `packages`?

Simple approach:
1. `cp ~/.julia/packages -Rf /path_to_my_project/packages`

Advanced approach:
1. Decide which packages you want to "bundle" in your "standalone bundle."
2. For each such package, do `cp ~/.julia/packages/packagename/packageslug -Rf /path_to_my_project/packages/packagename/packageslug `

The plan is that someone will develop an external tool to automate that process. We don’t need to add the functionality to Julia for creating the "self-contained standalone bundles." That can go in an external tool. We only need to add support for reading from the standalone bundles.

DilumAluthge:
The use case is to be able to make a single directory that contains your project, as well as all of the packages and artifacts that your project uses. You can then make a single tarball from that directory and:
1. Upload the tarball e.g. Zenodo
2. Send the tarball to e.g. journal reviewers
3. Copy the tarball onto write-once media (e.g. DVD-ROM), and then use the write-once media to transfer the tarball to e.g. an airgapped computer with no Internet access.

KristofferC:
> Simple approach:

Yeah but that will bring you all of the packages and all versions of them.


> The use case is to be able to make a single directory that contains your project, as well as all of the packages and artifacts that your project uses.

Yeah, you would do this now by adding

```
bundled_depot/packages/...
bundled_depot/artifacts/...
```

and doing `push!(DEPOT_PATH, "bundled_depot")`. Is all of this to avoid having to do that `push!`?

base/loading.jl
```diff
@@ -591,9 +591,15 @@ function explicit_manifest_uuid_path(project_file::String, pkg::PkgId)::Union{No
...
+            depot_path = [abspath(dirname(project)); depot_path]
```
tkf: Can you move it to a sub-directory? Quoting my other comments:

> Does it make sense to use `.julia` as the parent directory (i.e., `/path_to_active_project/.julia/artifacts` and `/path_to_active_project/.julia/packages`), to avoid cluttering the package directory? I think there are other things you might want to make project-local. For example, it would also be nice if the system image is automatically loaded from, e.g., `/path_to_active_project/.julia/sys.so` [JuliaLang/julia#33973 (comment)](https://github.com/JuliaLang/julia/issues/33973#issuecomment-601541817)
>
> ---  https://github.com/JuliaLang/Pkg.jl/pull/1727#issuecomment-601994269

What I'm suggesting is something like:

```suggestion
            depot_path = [abspath(dirname(project), ".julia"); depot_path]
```

`".julia"` is an example. Alternatives: https://github.com/JuliaLang/Pkg.jl/pull/1727#issuecomment-601997518

KristofferC:
Regarding https://github.com/JuliaLang/julia/pull/35222#discussion_r396136734, I agree with that a specific `.julia` directory or similar would be good since people will want to add more things `artifacts`, in the future perhaps `sysimages` etc. Best to keep all that in one directory to prevent polluting the top level package too much.

tkf:
> and doing `push!(DEPOT_PATH, "Package/.julia")`. Is all of this to avoid having to do that `push!`?

I think it is important to have a _canonical name_ for path like this.  Here is a possible source of inspiration: There is something called [`zipapp`](https://docs.python.org/3/library/zipapp.html) (see also [PEP 441](https://www.python.org/dev/peps/pep-0441/)) in Python stdlib that can be used to turn a zipped Python package into an executable application.  I think similar approach is possible if we have a canonical name for the project-local depot(-like) directory path.  We need a specific name for this since it has to be understood by tools (including `julia` binary itself).  Of course, I'm not suggesting to do this here/now.  But the approach like this PR seems to be the right direction to get there. (ref #34759)

StefanKarpinski:
Just to clarify the thinking here: what we want is the ability download a tarball of an application and run it with the guarantee that it will not need to download any resources (packages and artifacts) in order to run. We may want to go a bit further and say that the app will use _only_ resources within the tarball, not any system-installed resources. This illuminates a few things:

* We do want to allow projects to provide content-addressed resources (packages, artifacts) because that's what's required to satisfy the goal.

* We don't want to allow projects to affect the behavior of the package manager in other ways—that's both undesirable (potentially dangerous, confusing) and not a goal.

These dos and don'ts are why treating the project as a full-blown depot is not desirable. Looking for packages and artifacts in the project is not arbitrary or a slippery slope: these are the only content-addressed resources that exist in Julia/Pkg; together they are necessary and sufficient to allow a project to be self-contained in terms of dependencies.

The idea that we want to enable shipping apps in such a way that they run _only_ using the resources in the project is why it's desirable to look at a project's packages and artifacts directories first: if the the project ships with all the necessary resources, then it's guaranteed that when running the project you won't hit the filesystem outside of the project at all. While content addressing should guarantee that it doesn't matter where we load content-addressed resources from, I think it still makes sense to make projects fully self-contained in the sense that we don't even look outside of the project for resources.

The `sysimg` point is interesting: that's something that is not saved in depots, so it would be specific to projects rather than depots. @KristofferC, can you elaborate on how you imagine that working?

tkf:
Re sysimage, here is an RFC/POC: https://github.com/JuliaLang/julia/pull/35794

Keno:
Superseded by the ability to specify this in Manifest.

PR #35233, opened 2020-03-23
Bump libgit2 to 1.0

nalimilan:
Needs https://github.com/JuliaPackaging/Yggdrasil/pull/638 so that BinaryBuilder can provide the new version. We also need to update the build flags to sync with Yggdrasil if we decide to keep using LibCURL there.

This drops support for previous versions, but we could keep it if needed by adding new branches.


nalimilan:
I've updated the PR now that LibGit2 1.0.0 has been added to Yggdrasil. I've also added a commit to remove any occurrence of curl, since AFAICT the only user of it was libgit2, which now uses its own internal library by default. This could break packages that relied on libcurl being available (if any), but they should use BinaryProvider instead.

deps/Versions.make
```diff
@@ -24,9 +24,7 @@ MBEDTLS_VER = 2.16.0
 MBEDTLS_BB_REL = 1
 LIBSSH2_VER = 1.9.0
 LIBSSH2_BB_REL = 1
-CURL_VER = 7.66.0
-CURL_BB_REL = 1
-LIBGIT2_VER = 0.28.5
+LIBGIT2_VER = 1.0.0
```
nalimilan, line 27: Note to self: update to 1.0.1 in the next round of revisions (available since https://github.com/JuliaPackaging/Yggdrasil/pull/1410).
```suggestion
LIBGIT2_VER = 1.0.1
```

deps/libgit2.version
```diff
@@ -1,2 +1,2 @@
-LIBGIT2_BRANCH=v0.28.5
-LIBGIT2_SHA1=7a2b969d559b83798d93728f24d1729ffc97b717
+LIBGIT2_BRANCH=v1.0.0
+LIBGIT2_SHA1=7d3c7057f0e774aecd6fc4ef8333e69e5c4873e0
```
nalimilan, line 2:
```suggestion
LIBGIT2_BRANCH=v1.0.1
LIBGIT2_SHA1=0ced29612dacb67eefe0c562a5c1d3aab21cce96
```

simonbyrne (approved):
the changes to `stdlib/LibGit2/` look good to me. 

I can't comment on `deps/` and `Makefile` changes though

staticfloat:
We probably don't want to remove libcurl here, since we're about to add a new client of libcurl via @StefanKarpinski's new curl-based downloader.  Cool that libgit2 dropped dependence on it though!

nalimilan:
OK, I've reverted the removal of curl. Good to go?

simonbyrne (commented):
I think you need to add the necessary hashes to `julia/deps/checksums/`, and remove the old ones.

It also looks like https://github.com/JuliaLang/julia/blob/master/deps/patches/libgit2-case-sensitive.patch is no longer necessary?

nalimilan:
Ah right. Done. (It looks like the checksums for the tarball were missing for the currently used release, but nobody tests this path now...)

simonbyrne (approved):

Issue #33435, opened 2019-10-01
PermutedDimsArray constant-propagation / inference


kcajf:
```
julia> a = rand(2, 2)
2×2 Array{Float64,2}:
 0.159947  0.956223
 0.712618  0.864652

julia> @code_warntype (x -> PermutedDimsArray(x, (2, 1)))(a)
Variables
  #self#::Core.Compiler.Const(var"##28#29"(), false)
  x::Array{Float64,2}

Body::PermutedDimsArray{Float64,2,_A,_B,Array{Float64,2}} where _B where _A
1 ─ %1 = Core.tuple(2, 1)::Core.Compiler.Const((2, 1), false)
│   %2 = Main.PermutedDimsArray(x, %1)::PermutedDimsArray{Float64,2,_A,_B,Array{Float64,2}} where _B where _A
└──      return %2
```
I feel like this should be inferrable, no?

jipolanco:
The problem is that the permutation is encoded as a parameter of the `PermutedDimsArray` type.

One workaround would be to add an alternative `PermutedDimsArray` constructor taking a compile-time permutation, for instance `perm = Val((2, 1))`. In this case, the `_A` parameter in your example would be inferred by the compiler.

The `_B` parameter (the inverse permutation of `perm`) can also be evaluated at compile-time from `perm`. A possible implementation, using generated functions, is available [here](https://github.com/jipolanco/PencilFFTs.jl/blob/e8300154acb9d274b79cf249013b0984cdae2b5e/src/Pencils/permutations.jl#L35) (this actually computes the relative permutation between two permutations, which is a bit more general, but it's basically the same).

I'd be glad to submit a PR if there is interest in this kind of solution.

kcajf:
True, though in the case I gave the permutation is a compile time constant, so it seems reasonable that it would be inferred without using a Val type

adienes:
infers since 1.12

PR #35234, opened 2020-03-23
faster & type stable `invperm` and `isperm` for tuples shorter that 16 elements

francescoalemanno:
#### Regarding speed
small simple benchmarking code
```julia
using Random
using BenchmarkTools

function fast_invperm(P::NTuple{N,T}) where {N,T}
    isperm(P) || throw(ArgumentError("argument is not a permutation"))
    ntuple(Val(N)) do i
        for j in P
            @inbounds P[j]==i && return j
        end
        return first(P) # DO NOT REMOVE: necessary for type stable code_gen
    end
end

function time_impl()
    for N in 3:10:50
        T=(shuffle(1:N)...,)
        @show @benchmark invperm($T)
        @show @benchmark fast_invperm($T)
    end
end

time_impl()
```

results:

```
#= ....108 =# @benchmark(invperm($(Expr(:$, :T)))) = Trial(253.927 ns)
#= ....109 =# @benchmark(fast_invperm($(Expr(:$, :T)))) = Trial(41.746 ns)
#= ....108 =# @benchmark(invperm($(Expr(:$, :T)))) = Trial(450.132 ns)
#= ....109 =# @benchmark(fast_invperm($(Expr(:$, :T)))) = Trial(90.729 ns)
#= ....108 =# @benchmark(invperm($(Expr(:$, :T)))) = Trial(637.290 ns)
#= ....109 =# @benchmark(fast_invperm($(Expr(:$, :T)))) = Trial(194.659 ns)
#= ....108 =# @benchmark(invperm($(Expr(:$, :T)))) = Trial(1.296 μs)
#= ....109 =# @benchmark(fast_invperm($(Expr(:$, :T)))) = Trial(303.951 ns)
#= ....108 =# @benchmark(invperm($(Expr(:$, :T)))) = Trial(1.612 μs)
#= ....109 =# @benchmark(fast_invperm($(Expr(:$, :T)))) = Trial(475.883 ns)
```

#### Regarding type stability
```julia
@code_warntype invperm((1,3,2))
Variables
  #self#::Core.Compiler.Const(invperm, false)
  a::Tuple{Int64,Int64,Int64}

Body::Tuple{Vararg{Int64,N} where N}
1 ─ %1 = Core._apply_iterate(Base.iterate, Base.vect, a)::Array{Int64,1}
│   %2 = Base.invperm(%1)::Array{Int64,1}
│   %3 = Core._apply_iterate(Base.iterate, Core.tuple, %2)::Tuple{Vararg{Int64,N} where N}
└──      return %3
```

```julia
@code_warntype fast_invperm((1,3,2))

Variables
  #self#::Core.Compiler.Const(fast_invperm, false)
  P::Tuple{Int64,Int64,Int64}
  #9::var"#9#10"{Tuple{Int64,Int64,Int64}}

Body::Tuple{Int64,Int64,Int64}
1 ─       Core.NewvarNode(:(#9))
│   %2  = Main.isperm(P)::Bool
└──       goto #3 if not %2
2 ─       goto #4
3 ─ %5  = Main.ArgumentError("argument is not a permutation")::ArgumentError
└──       Main.throw(%5)
4 ┄ %7  = Main.:(var"#9#10")::Core.Compiler.Const(var"#9#10", false)
│   %8  = Core.typeof(P)::Core.Compiler.Const(Tuple{Int64,Int64,Int64}, false)
│   %9  = Core.apply_type(%7, %8)::Core.Compiler.Const(var"#9#10"{Tuple{Int64,Int64,Int64}}, false)
│         (#9 = %new(%9, P))
│   %11 = #9::var"#9#10"{Tuple{Int64,Int64,Int64}}
│   %12 = Main.Val($(Expr(:static_parameter, 1)))::Core.Compiler.Const(Val{3}(), true)
│   %13 = Main.ntuple(%11, %12)::Tuple{Int64,Int64,Int64}
└──       return %13

```

### DIFF
--- a/base/combinatorics.jl
+++ b/base/combinatorics.jl
@@ -263,12 +263,16 @@ function invperm(p::Union{Tuple{},Tuple{Int},Tuple{Int,Int}})
     p  # in dimensions 0-2, every permutation is its own inverse
 end
 
-function invperm(P::NTuple{N,Int}) where N
-    ntuple(Val(N)) do i
-        for j in eachindex(P)
-            P[j]==i && return Int(j)
+function invperm(P::NTuple{N}) where N
+    valn = Val(N)
+    ntuple(valn) do i
+        s = Base.afoldl(nothing, ntuple(identity, valn)...) do s, j
+            s !== nothing && return s
+            P[j]==i && return j
+            nothing
         end
-        throw(ArgumentError("argument is not a permutation"))
+        s === nothing && throw(ArgumentError("argument is not a permutation"))
+        s
     end
 end
 
--- a/base/combinatorics.jl
+++ b/base/combinatorics.jl
@@ -36,6 +36,15 @@ end
 
 # Basic functions for working with permutations
 
+function _isperm(A)
+    n = length(A)
+    used = falses(n)
+    for a in A
+        (0 < a <= n) && (used[a] ⊻= true) || return false
+    end
+    true
+end
+
 """
     isperm(v) -> Bool
 
@@ -50,16 +59,6 @@ julia> isperm([1; 3])
 false
 ```
 """
-
-function _isperm(A)
-    n = length(A)
-    used = falses(n)
-    for a in A
-        (0 < a <= n) && (used[a] ⊻= true) || return false
-    end
-    true
-end
-
 isperm(A) = _isperm(A)
 
 function isperm(P::NTuple{N,Int}) where N
--- a/base/combinatorics.jl
+++ b/base/combinatorics.jl
@@ -36,13 +36,20 @@ end
 
 # Basic functions for working with permutations
 
-function _isperm(A)
-    n = length(A)
-    used = falses(n)
-    for a in A
-        (0 < a <= n) && (used[a] ⊻= true) || return false
+@inline function _foldoneto(op, acc, ::Val{N}) where N
+    @assert N::Integer > 0
+    if @generated
+        quote
+            acc_0 = acc
+            Base.Cartesian.@nexprs $N i -> acc_{i} = op(acc_{i-1}, i)
+            return $(Symbol(:acc_, N))
+        end
+    else
+        for i in 1:N
+            acc = op(acc, i)
+        end
+        return acc
     end
-    true
 end
 
 """
@@ -59,28 +66,30 @@ julia> isperm([1; 3])
 false
 ```
 """
-isperm(A) = _isperm(A)
-
-function isperm(P::NTuple{N,Int}) where N
-    if N>21 # if N above 21 other algorithm is faster
-        return _isperm(P)
+function isperm(A)
+    n = length(A)
+    used = falses(n)
+    for a in A
+        (0 < a <= n) && (used[a] ⊻= true) || return false
     end
-    for i in eachindex(P)
-        flag=true
-        for j in eachindex(P)
-            if P[j]==i
-                flag=false
-                break
-            end
-        end
-        flag && return false
-    end
-    return true
+    true
 end
 
 isperm(p::Tuple{}) = true
 isperm(p::Tuple{Int}) = p[1] == 1
 isperm(p::Tuple{Int,Int}) = ((p[1] == 1) & (p[2] == 2)) | ((p[1] == 2) & (p[2] == 1))
+
+function isperm(P::NTuple{N,Integer}) where {N}
+    valn = Val(N)
+    _foldoneto(true, valn) do b,i
+        s = _foldoneto(false, valn) do s, j
+            s && return s
+            P[j]==i && return true
+            false
+        end
+        b&s
+    end
+end
 
 # swap columns i and j of a, in-place
 function swapcols!(a::AbstractMatrix, i, j)
@@ -262,10 +271,10 @@ function invperm(p::Union{Tuple{},Tuple{Int},Tuple{Int,Int}})
     p  # in dimensions 0-2, every permutation is its own inverse
 end
 
-function invperm(P::NTuple{N}) where N
+function invperm(P::NTuple{N,Integer}) where {N}
     valn = Val(N)
     ntuple(valn) do i
-        s = Base.afoldl(nothing, ntuple(identity, valn)...) do s, j
+        s = _foldoneto(nothing, valn) do s, j
             s !== nothing && return s
             P[j]==i && return j
             nothing
--- a/base/combinatorics.jl
+++ b/base/combinatorics.jl
@@ -60,12 +60,17 @@ function isperm(A)
 end
 
 function isperm(P::NTuple{N,T}) where {N,T}
-    all(ntuple(Val(N)) do i
+    for i in eachindex(P)
+        flag=true
         for j in eachindex(P)
-            P[j]==i && return true
+            if P[j]==i
+                flag=false
+                break
+            end 
         end
-        return false
-    end)
+        flag && return false
+    end
+    return true
 end
 
 isperm(p::Tuple{}) = true
--- a/base/combinatorics.jl
+++ b/base/combinatorics.jl
@@ -50,7 +50,8 @@ julia> isperm([1; 3])
 false
 ```
 """
-function isperm(A)
+
+function _isperm(A)
     n = length(A)
     used = falses(n)
     for a in A
@@ -59,7 +60,12 @@ function isperm(A)
     true
 end
 
+isperm(A) = _isperm(A)
+
 function isperm(P::NTuple{N,T}) where {N,T}
+    if N>21 # if N above 21 other algorithm is faster
+        return _isperm(P)
+    end
     for i in eachindex(P)
         flag=true
         for j in eachindex(P)
--- a/base/combinatorics.jl
+++ b/base/combinatorics.jl
@@ -244,11 +244,11 @@ function invperm(p::Union{Tuple{},Tuple{Int},Tuple{Int,Int}})
 end
 
 function invperm(P::NTuple{N,T}) where {N,T}
-    isperm(P) || throw(ArgumentError("argument is not a permutation"))
     ntuple(Val(N)) do i
         for j in P
             @inbounds P[j]==i && return j
         end
+        throw(ArgumentError("argument is not a permutation"))
         return first(P) # DO NOT REMOVE: necessary for type stable code_gen
     end
 end
--- a/base/combinatorics.jl
+++ b/base/combinatorics.jl
@@ -81,8 +81,8 @@ isperm(p::Tuple{}) = true
 isperm(p::Tuple{Int}) = p[1] == 1
 isperm(p::Tuple{Int,Int}) = ((p[1] == 1) & (p[2] == 2)) | ((p[1] == 2) & (p[2] == 1))
 
-function isperm(P::NTuple{N,Integer}) where {N}
-    valn = Val(N)
+function isperm(P::Tuple)
+    valn = Val(length(P))
     _foldoneto(true, valn) do b,i
         s = _foldoneto(false, valn) do s, j
             s || P[j]==i
@@ -273,8 +273,8 @@ function invperm(p::Union{Tuple{},Tuple{Int},Tuple{Int,Int}})
     p  # in dimensions 0-2, every permutation is its own inverse
 end
 
-function invperm(P::NTuple{N,Integer}) where {N}
-    valn = Val(N)
+function invperm(P::Tuple)
+    valn = Val(length(P))
     ntuple(valn) do i
         s = _foldoneto(nothing, valn) do s, j
             s !== nothing && return s
--- a/base/combinatorics.jl
+++ b/base/combinatorics.jl
@@ -286,7 +286,7 @@ function invperm(P::Tuple)
     end
 end
 
-invperm(P::Any16) = Tuple(invperm(collect(P)))
+invperm(P::Any16) = Tuple(invperm(collect(P)))::typeof(P)
 
 #XXX This function should be moved to Combinatorics.jl but is currently used by Base.DSP.
 """
--- a/base/combinatorics.jl
+++ b/base/combinatorics.jl
@@ -62,7 +62,7 @@ end
 
 isperm(A) = _isperm(A)
 
-function isperm(P::NTuple{N,T}) where {N,T}
+function isperm(P::NTuple{N,Int}) where N
     if N>21 # if N above 21 other algorithm is faster
         return _isperm(P)
     end
@@ -263,10 +263,10 @@ function invperm(p::Union{Tuple{},Tuple{Int},Tuple{Int,Int}})
     p  # in dimensions 0-2, every permutation is its own inverse
 end
 
-function invperm(P::NTuple{N,T}) where {N,T}
+function invperm(P::NTuple{N,Int}) where N
     ntuple(Val(N)) do i
         for j in eachindex(P)
-            P[j]==i && return T(j)
+            P[j]==i && return Int(j)
         end
         throw(ArgumentError("argument is not a permutation"))
     end
--- a/base/combinatorics.jl
+++ b/base/combinatorics.jl
@@ -66,7 +66,9 @@ julia> isperm([1; 3])
 false
 ```
 """
-function isperm(A)
+isperm(A) = _isperm(A)
+
+function _isperm(A)
     n = length(A)
     used = falses(n)
     for a in A
@@ -88,6 +90,8 @@ function isperm(P::NTuple{N,Integer}) where {N}
         b&s
     end
 end
+
+isperm(P::Any16) = _isperm(P)
 
 # swap columns i and j of a, in-place
 function swapcols!(a::AbstractMatrix, i, j)
@@ -281,6 +285,8 @@ function invperm(P::NTuple{N,Integer}) where {N}
         s
     end
 end
+
+invperm(P::Any16) = Tuple(invperm(collect(P)))
 
 #XXX This function should be moved to Combinatorics.jl but is currently used by Base.DSP.
 """
--- a/base/combinatorics.jl
+++ b/base/combinatorics.jl
@@ -59,6 +59,15 @@ function isperm(A)
     true
 end
 
+function isperm(P::NTuple{N,T}) where {N,T}
+    all(ntuple(Val(N)) do i
+        for j in eachindex(P)
+            @inbounds P[j]==i && return true
+        end
+        return false
+    end)
+end
+
 isperm(p::Tuple{}) = true
 isperm(p::Tuple{Int}) = p[1] == 1
 isperm(p::Tuple{Int,Int}) = ((p[1] == 1) & (p[2] == 2)) | ((p[1] == 2) & (p[2] == 1))
@@ -245,8 +254,8 @@ end
 
 function invperm(P::NTuple{N,T}) where {N,T}
     ntuple(Val(N)) do i
-        for j in P
-            @inbounds P[j]==i && return j
+        for j in eachindex(P)
+            @inbounds P[j]==i && return T(j)
         end
         throw(ArgumentError("argument is not a permutation"))
     end
--- a/base/combinatorics.jl
+++ b/base/combinatorics.jl
@@ -62,7 +62,7 @@ end
 function isperm(P::NTuple{N,T}) where {N,T}
     all(ntuple(Val(N)) do i
         for j in eachindex(P)
-            @inbounds P[j]==i && return true
+            P[j]==i && return true
         end
         return false
     end)
@@ -255,7 +255,7 @@ end
 function invperm(P::NTuple{N,T}) where {N,T}
     ntuple(Val(N)) do i
         for j in eachindex(P)
-            @inbounds P[j]==i && return T(j)
+            P[j]==i && return T(j)
         end
         throw(ArgumentError("argument is not a permutation"))
     end
--- a/base/combinatorics.jl
+++ b/base/combinatorics.jl
@@ -242,7 +242,16 @@ function invperm(p::Union{Tuple{},Tuple{Int},Tuple{Int,Int}})
     isperm(p) || throw(ArgumentError("argument is not a permutation"))
     p  # in dimensions 0-2, every permutation is its own inverse
 end
-invperm(a::Tuple) = (invperm([a...])...,)
+
+function invperm(P::NTuple{N,T}) where {N,T}
+    isperm(p) || throw(ArgumentError("argument is not a permutation"))
+    ntuple(Val(N)) do i
+        for j in P
+            @inbounds P[j]==i && return j
+        end
+        return zero(T) #
+    end
+end
 
 #XXX This function should be moved to Combinatorics.jl but is currently used by Base.DSP.
 """
--- a/base/combinatorics.jl
+++ b/base/combinatorics.jl
@@ -244,12 +244,12 @@ function invperm(p::Union{Tuple{},Tuple{Int},Tuple{Int,Int}})
 end
 
 function invperm(P::NTuple{N,T}) where {N,T}
-    isperm(p) || throw(ArgumentError("argument is not a permutation"))
+    isperm(P) || throw(ArgumentError("argument is not a permutation"))
     ntuple(Val(N)) do i
         for j in P
             @inbounds P[j]==i && return j
         end
-        return zero(T) #
+        return first(P) # DO NOT REMOVE: necessary for type stable code_gen
     end
 end
 
--- a/base/combinatorics.jl
+++ b/base/combinatorics.jl
@@ -286,7 +286,7 @@ function invperm(P::Tuple)
     end
 end
 
-invperm(P::Any16) = Tuple(invperm(collect(P)))::typeof(P)
+invperm(P::Any16) = Tuple(invperm(collect(P)))
 
 #XXX This function should be moved to Combinatorics.jl but is currently used by Base.DSP.
 """
--- a/base/combinatorics.jl
+++ b/base/combinatorics.jl
@@ -83,9 +83,7 @@ function isperm(P::NTuple{N,Integer}) where {N}
     valn = Val(N)
     _foldoneto(true, valn) do b,i
         s = _foldoneto(false, valn) do s, j
-            s && return s
-            P[j]==i && return true
-            false
+            s || P[j]==i
         end
         b&s
     end
--- a/base/combinatorics.jl
+++ b/base/combinatorics.jl
@@ -249,7 +249,6 @@ function invperm(P::NTuple{N,T}) where {N,T}
             @inbounds P[j]==i && return j
         end
         throw(ArgumentError("argument is not a permutation"))
-        return first(P) # DO NOT REMOVE: necessary for type stable code_gen
     end
 end
 
--- a/base/combinatorics.jl
+++ b/base/combinatorics.jl
@@ -72,7 +72,7 @@ function isperm(P::NTuple{N,T}) where {N,T}
             if P[j]==i
                 flag=false
                 break
-            end 
+            end
         end
         flag && return false
     end
--- a/test/combinatorics.jl
+++ b/test/combinatorics.jl
@@ -34,7 +34,7 @@ end
     @test invperm((1,2)) == (1,2)
     @test invperm((2,1)) == (2,1)
     @test_throws ArgumentError invperm((1,3))
-    
+
     push!(p, 1)
     @test !isperm(p)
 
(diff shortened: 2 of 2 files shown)


mcabbott:
Improvements would be nice to have. On short tuples, but not long ones, [TupleTools.jl](https://github.com/Jutho/TupleTools.jl) is quite a bit quicker:
```julia
julia> @btime fast_invperm((3,1,2,4))
  61.170 ns (2 allocations: 128 bytes)
(2, 3, 1, 4)

julia> @btime TupleTools.invperm((3,1,2,4))
  1.422 ns (0 allocations: 0 bytes)
(2, 3, 1, 4)

julia> @btime fast_invperm((3,1,2,4,5,6,7,8,10,9));
  95.121 ns (2 allocations: 128 bytes)

julia> @btime TupleTools.invperm((3,1,2,4,5,6,7,8,10,9));
  425.181 ns (25 allocations: 1.00 KiB)
```

francescoalemanno:
> Improvements would be nice to have. On short tuples, but not long ones, [TupleTools.jl](https://github.com/Jutho/TupleTools.jl) is quite a bit quicker:
> 
> ```julia
> julia> @btime fast_invperm((3,1,2,4))
>   61.170 ns (2 allocations: 128 bytes)
> (2, 3, 1, 4)
> 
> julia> @btime TupleTools.invperm((3,1,2,4))
>   1.422 ns (0 allocations: 0 bytes)
> (2, 3, 1, 4)
> 
> julia> @btime fast_invperm((3,1,2,4,5,6,7,8,10,9));
>   95.121 ns (2 allocations: 128 bytes)
> 
> julia> @btime TupleTools.invperm((3,1,2,4,5,6,7,8,10,9));
>   425.181 ns (25 allocations: 1.00 KiB)
> ```

i think the tuples must be interpolated via `$` in order for the benchmark to be meaningful.

mcabbott:
I think that what I wrote is equivalent to `$`. You do get slightly longer times with e.g. `@btime TupleTools.invperm(p)  setup=(p=(3,1,2,4));` or `@btime TupleTools.invperm($(Ref((3,1,2,4)))[]);` as these (if I understand right) circumvent constant propagation. I think we care about the speed both with and without this. But however you slice it, `TupleTools` is beaten from length=10.

francescoalemanno:
Jutho's TupleTools is much faster than both base and my code for small tuples, since it uses "lispy" tuple programs that rely on dispatch,  this is also the reason that it becomes slower for bigger tuples.
the implementations are quite a lot different. What do you suggest?

mcabbott:
We should switch, do a lispy thing for small tuples, and something else for large ones. Many functions in Base do this. At length=10, `Base.invperm` is already faster on an Array than on a Tuple, so perhaps we should just call that & convert back.

base/combinatorics.jl
```diff
@@ -242,7 +242,16 @@ function invperm(p::Union{Tuple{},Tuple{Int},Tuple{Int,Int}})
...
 end
-invperm(a::Tuple) = (invperm([a...])...,)
+
+function invperm(P::NTuple{N,T}) where {N,T}
+    isperm(P) || throw(ArgumentError("argument is not a permutation"))
+    ntuple(Val(N)) do i
+        for j in P
+            @inbounds P[j]==i && return j
+        end
+        return first(P) # DO NOT REMOVE: necessary for type stable code_gen
```
StefanKarpinski: You could make this an error instead. Can you avoid the upfront `isperm` test by putting it here instead?

francescoalemanno: doing it right now, it is much faster without `isperm` call

mcabbott:
I think it goes faster if the inner for loop is also an ntuple, and you can check invperm inside? There should be no need for `@inbounds` either. Like this, but double-check:
```julia
function fast_invperm3(P::NTuple{N,Int}) where {N}
	ntuple(N) do i
	    s = sum(ntuple(N) do j
	        P[j]==i && return j
	        0
	    end)
	    s == 0 && error("not a perm")
	    s
	end
end
```

base/combinatorics.jl
```diff
@@ -242,7 +242,16 @@ function invperm(p::Union{Tuple{},Tuple{Int},Tuple{Int,Int}})
...
 end
-invperm(a::Tuple) = (invperm([a...])...,)
+
+function invperm(P::NTuple{N,T}) where {N,T}
+    ntuple(Val(N)) do i
+        for j in P
+            @inbounds P[j]==i && return j
+        end
+        throw(ArgumentError("argument is not a permutation"))
+        return first(P) # DO NOT REMOVE: necessary for type stable code_gen
```
KristofferC: Are you sure about this? What do you get if you remove this?

StefanKarpinski: The compiler can't know that the above loop must terminate, so this makes sense. But it makes just as much sense to make this an error.

KristofferC: I don't get it. There is a `throw` just above this? Can someone spell it out to me with a `@code_warntype` or something.

francescoalemanno: Thank you! i just confirmed, that indeed replacing the return with a throw mantains the code typestable.

francescoalemanno:
```julia
using Random
using BenchmarkTools

function fast_invperm(P::NTuple{N,T}) where {N,T}
    ntuple(Val(N)) do i
        for j in P
            @inbounds P[j]==i && return j
        end
        throw(ArgumentError("argument is not a permutation"))
    end
end

function time_impl()
    for N in 3:10:50
        T=(shuffle(1:N)...,)
        @show @benchmark invperm($T)
        @show @benchmark fast_invperm($T)
    end
end

time_impl()

#= :100 =# @benchmark(invperm($(Expr(:$, :T)))) = Trial(260.141 ns)
#= :101 =# @benchmark(fast_invperm($(Expr(:$, :T)))) = Trial(7.581 ns)
#= :100 =# @benchmark(invperm($(Expr(:$, :T)))) = Trial(453.706 ns)
#= :101 =# @benchmark(fast_invperm($(Expr(:$, :T)))) = Trial(41.056 ns)
#= :100 =# @benchmark(invperm($(Expr(:$, :T)))) = Trial(652.377 ns)
#= :101 =# @benchmark(fast_invperm($(Expr(:$, :T)))) = Trial(110.756 ns)
#= :100 =# @benchmark(invperm($(Expr(:$, :T)))) = Trial(1.303 μs)
#= :101 =# @benchmark(fast_invperm($(Expr(:$, :T)))) = Trial(241.758 ns)
#= :100 =# @benchmark(invperm($(Expr(:$, :T)))) = Trial(1.609 μs)
#= :101 =# @benchmark(fast_invperm($(Expr(:$, :T)))) = Trial(415.975 ns)
```

base/combinatorics.jl
francescoalemanno: @KristofferC the compiler can never know if any of the returns in the loop are executed, so it correctly infers that the return value of the closure inside ntuple can be either a `Int` or `nothing`,  this can be rectified by adding a phantom return at the end of the closure with the correct type (but this was very inelegant), or as @StefanKarpinski said that invalid exit point can be made into an error, so the compiler know that the only possible exit point from the closure is withing the loop thus it can again infer the right return type, and keep the code typestable

mcabbott:
One reason to care a lot about very short tuples is things like #33435: PermutedDimsArray{T,4} is common, while PermutedDimsArray{T,14} may never have been spotted in the wild. Getting sub-ns `@btime` is (if I understand right) a sign that constant-propagation is allowing this to be compiled away.

francescoalemanno:
@mcabbott i agree, i will try to make a compact tuple program to solve it in small cases without overhead from the loop

francescoalemanno:
@mcabbott, yes i can see that your code for smaller tuples is able to make the compiler do all the heavy lifting, but i do not like the use of sums.

francescoalemanno:
added `isperm` and also managed to make both method even faster

```julia
function fast_isperm(P::NTuple{N,T}) where {N,T}
    all(ntuple(Val(N)) do i
        for j in eachindex(P)
            @inbounds P[j]==i && return true
        end
        return false
    end)
end

function fast_invperm(P::NTuple{N,T}) where {N,T}
    ntuple(Val(N)) do i
        for j in eachindex(P)
            @inbounds P[j]==i && return T(j)
        end
        throw(ArgumentError("argument is not a permutation"))
    end
end

using Random
using BenchmarkTools


function time_impl()
    for N in 3:10:50
        T=(shuffle(1:N)...,)
        @info "Testing size $N :"
        @btime invperm($T)
        @btime invperm($T)
        @btime fast_invperm($T)
        @btime fast_invperm($T)
    end
end

time_impl()
```

```
[ Info: Testing size 3 :
  249.991 ns (3 allocations: 256 bytes)
  254.721 ns (3 allocations: 256 bytes)
  6.849 ns (0 allocations: 0 bytes)
  6.849 ns (0 allocations: 0 bytes)
[ Info: Testing size 13 :
  458.505 ns (3 allocations: 496 bytes)
  445.848 ns (3 allocations: 496 bytes)
  24.597 ns (0 allocations: 0 bytes)
  24.591 ns (0 allocations: 0 bytes)
[ Info: Testing size 23 :
  641.355 ns (3 allocations: 736 bytes)
  652.488 ns (3 allocations: 736 bytes)
  64.287 ns (0 allocations: 0 bytes)
  64.280 ns (0 allocations: 0 bytes)
[ Info: Testing size 33 :
  1.303 μs (4 allocations: 1.25 KiB)
  1.283 μs (4 allocations: 1.25 KiB)
  222.649 ns (0 allocations: 0 bytes)
  222.863 ns (0 allocations: 0 bytes)
[ Info: Testing size 43 :
  1.611 μs (4 allocations: 1.59 KiB)
  1.614 μs (4 allocations: 1.59 KiB)
  372.524 ns (0 allocations: 0 bytes)
  371.869 ns (0 allocations: 0 bytes)
```

base/combinatorics.jl
KristofferC: The whole time I have looked at this code, there has been a `throw` above the `return` so it is based on this I have commented.

tkf:
It looks like you can make `fast_invperm3` faster for `N` ~ 7 by moving `N` to the type domain before wrapped in closure:

```julia
function fast_invperm4(P::NTuple{N}) where {N}
    valn = Val(N)
    ntuple(valn) do i
        s = sum(ntuple(valn) do j
            P[j]==i && return j
            0
        end)
        s == 0 && error("not a perm")
        s
    end
end
```

You can avoid using `sum` by explicit looping with `afoldl`

```julia
function fast_invperm5(P::NTuple{N}) where {N}
    valn = Val(N)
    ntuple(valn) do i
        s = Base.afoldl(nothing, ntuple(identity, valn)...) do s, j
            s !== nothing && return s
            P[j]==i && return j
            nothing
        end
        s === nothing && error("not a perm")
        s
    end
end
```

Interestingly, `fast_invperm3` and `fast_invperm4` are the fastest in `N < 7`, and `fast_invperm5` is the fastest for `N = 7, 8, 9, 10`.

```
N = 6
@benchmark(fast_invperm($(Expr(:$, :T)))) = Trial(16.254 ns)
@benchmark(fast_invperm3($(Expr(:$, :T)))) = Trial(0.030 ns)
@benchmark(fast_invperm4($(Expr(:$, :T)))) = Trial(0.030 ns)
@benchmark(fast_invperm5($(Expr(:$, :T)))) = Trial(5.327 ns)
N = 7
@benchmark(fast_invperm($(Expr(:$, :T)))) = Trial(18.465 ns)
@benchmark(fast_invperm3($(Expr(:$, :T)))) = Trial(30.540 ns)
@benchmark(fast_invperm4($(Expr(:$, :T)))) = Trial(19.005 ns)
@benchmark(fast_invperm5($(Expr(:$, :T)))) = Trial(7.059 ns)
N = 8
@benchmark(fast_invperm($(Expr(:$, :T)))) = Trial(22.424 ns)
@benchmark(fast_invperm3($(Expr(:$, :T)))) = Trial(34.279 ns)
@benchmark(fast_invperm4($(Expr(:$, :T)))) = Trial(28.240 ns)
@benchmark(fast_invperm5($(Expr(:$, :T)))) = Trial(8.413 ns)
N = 9
@benchmark(fast_invperm($(Expr(:$, :T)))) = Trial(25.592 ns)
@benchmark(fast_invperm3($(Expr(:$, :T)))) = Trial(36.803 ns)
@benchmark(fast_invperm4($(Expr(:$, :T)))) = Trial(32.465 ns)
@benchmark(fast_invperm5($(Expr(:$, :T)))) = Trial(10.581 ns)
N = 10
@benchmark(fast_invperm($(Expr(:$, :T)))) = Trial(30.084 ns)
@benchmark(fast_invperm3($(Expr(:$, :T)))) = Trial(47.627 ns)
@benchmark(fast_invperm4($(Expr(:$, :T)))) = Trial(39.858 ns)
@benchmark(fast_invperm5($(Expr(:$, :T)))) = Trial(12.450 ns)
```

mcabbott:
I nearly made a joke that you were going to show up with an amazing 1-line foldl answer! But `fast_invperm5` is interesting. Notice that on other tests, where the tuple is literally provided, it stays fast up to length 16, unlike the others:
```
julia> for N in 10:2:18
         p = Tuple(shuffle(1:N))
         @show N
         @eval begin
           @btime fast_invperm($p)
           @btime fast_invperm3($p)
           @btime fast_invperm4($p)
           @btime fast_invperm5($p)
         end
       end
N = 10
  32.327 ns (0 allocations: 0 bytes)
  1.697 ns (0 allocations: 0 bytes)
  2.355 ns (0 allocations: 0 bytes)
  1.696 ns (0 allocations: 0 bytes)
N = 12
  42.613 ns (0 allocations: 0 bytes)
  9.050 μs (65 allocations: 7.11 KiB)
  1.975 ns (0 allocations: 0 bytes)
  1.704 ns (0 allocations: 0 bytes)
N = 14
  54.587 ns (0 allocations: 0 bytes)
  11.509 μs (75 allocations: 9.14 KiB)
  72.662 ns (0 allocations: 0 bytes)
  1.975 ns (0 allocations: 0 bytes)
N = 16
  69.291 ns (0 allocations: 0 bytes)
  13.778 μs (85 allocations: 11.42 KiB)
  97.327 ns (0 allocations: 0 bytes)
  0.029 ns (0 allocations: 0 bytes)
N = 18
  85.220 ns (0 allocations: 0 bytes)
  16.378 μs (95 allocations: 13.95 KiB)
  2.526 ns (0 allocations: 0 bytes)
  149.934 ns (0 allocations: 0 bytes)

julia> @btime fast_invperm4((4,5,3,1,2,7,6,9,8,10,12,11,13,14));
  72.571 ns (0 allocations: 0 bytes)

julia> @btime fast_invperm5((4,5,3,1,2,7,6,9,8,10,12,11,13,14));
  1.975 ns (0 allocations: 0 bytes)
```

tkf:
Yeah, `afoldl`-based code can only be fast up to 16 because of this restriction:

https://github.com/JuliaLang/julia/blob/f449765943ba414bd57c3d1a44a73e5a0bb27534/base/operators.jl#L518-L522

(An "upside" is that we can rely on the tuple length cut-off of `afoldl`.)

BTW, another nice-to-have property is that it can handle a tuple of [StaticNumbers.jl](https://github.com/perrutquist/StaticNumbers.jl). This way, we can do `invperm` at compile-time when using static numbers. For this, we need to use `::NTuple{N,Integer}` instead of `::NTuple{N}` or `::NTuple{N,Int}`.  I think we can check this in the test by try inverting (say) `(3, true, UInt8(2))`.

If we want this property, I think we need to use one of the non-`for`-loop approaches:

```julia
julia> x = Tuple(static.(shuffle(1:6)))
(static(2), static(5), static(1), static(4), static(3), static(6))

julia> @code_warntype fast_invperm3(x)
...
│   %8 = Main.ntuple(%7, $(Expr(:static_parameter, 1)))::Core.Compiler.Const((3, 1, 5, 4, 2, 6), false)
└──      return %8

julia> @code_warntype fast_invperm(x)
...
│   %8 = Main.ntuple(%6, %7)::Tuple{StaticInteger,StaticInteger,StaticInteger,StaticInteger,StaticInteger{2},StaticInteger}
└──      return %8
```

francescoalemanno:
I'm in favor of the `afoldl` version, do you guys know why tests are failing?

tkf:
Actually I don't like the `afoldl` version so much since there are better solution for `N < 7`.  I think this is because of the temporary tuple `ntuple(identity, valn)` I need to create.  It would be nice to have a primitive for writing `for i in 1:N` for small constant `N`. Something like

```julia
function _foldoneto(op, acc, ::Val{N}) where N
    @assert N::Integer > 0
    if @generated
        quote
            acc_0 = acc
            Base.Cartesian.@nexprs $N i -> acc_{i} = op(acc_{i-1}, i)
            return $(Symbol(:acc_, N))
        end
    else
        for i in 1:N
            acc = op(acc, i)
        end
        return acc
    end
end
```

We can then use it instead of `afoldl(nothing, ntuple(identity, valn)) do ...`:

```julia
@inline function fast_invperm6(P::NTuple{N,Integer}) where {N}
    valn = Val(N)
    ntuple(valn) do i
        s = _foldoneto(nothing, valn) do s, j
            s !== nothing && return s
            P[j]==i && return j
            nothing
        end
        s === nothing && error("not a perm")
        s
    end
end
```

This works well with very small tuples (`N < 7`):

```julia
N = 2
@benchmark(fast_invperm4($(Expr(:$, :T)))) = Trial(0.028 ns)
@benchmark(fast_invperm5($(Expr(:$, :T)))) = Trial(0.029 ns)
@benchmark(fast_invperm6($(Expr(:$, :T)))) = Trial(0.028 ns)
N = 3
@benchmark(fast_invperm4($(Expr(:$, :T)))) = Trial(0.028 ns)
@benchmark(fast_invperm5($(Expr(:$, :T)))) = Trial(0.029 ns)
@benchmark(fast_invperm6($(Expr(:$, :T)))) = Trial(0.029 ns)
N = 4
@benchmark(fast_invperm4($(Expr(:$, :T)))) = Trial(0.029 ns)
@benchmark(fast_invperm5($(Expr(:$, :T)))) = Trial(3.581 ns)
@benchmark(fast_invperm6($(Expr(:$, :T)))) = Trial(0.028 ns)
N = 5
@benchmark(fast_invperm4($(Expr(:$, :T)))) = Trial(0.028 ns)
@benchmark(fast_invperm5($(Expr(:$, :T)))) = Trial(5.028 ns)
@benchmark(fast_invperm6($(Expr(:$, :T)))) = Trial(0.028 ns)
N = 6
@benchmark(fast_invperm4($(Expr(:$, :T)))) = Trial(0.028 ns)
@benchmark(fast_invperm5($(Expr(:$, :T)))) = Trial(5.447 ns)
@benchmark(fast_invperm6($(Expr(:$, :T)))) = Trial(0.028 ns)
N = 7
@benchmark(fast_invperm4($(Expr(:$, :T)))) = Trial(18.218 ns)
@benchmark(fast_invperm5($(Expr(:$, :T)))) = Trial(7.066 ns)
@benchmark(fast_invperm6($(Expr(:$, :T)))) = Trial(0.028 ns)
```

This requires a `@generated` function but maybe this is better than implementing `invperm` directly with `@generated` as `_foldoneto` would be useful for other functions, too. (Though I don't know what @Keno would say.)

This approach can be problematic for large tuples.  I think it is a good idea to implement `invperm(P::Any16)` by dispatching to the vector method: `invperm(P::Any16) = Tuple(invperm(collect(P)))`.  16 is just a conventional(?) choice as `Base.Any16` already exist.

francescoalemanno:
@tkf

> This approach can be problematic for large tuples. I think it is a good idea to implement `invperm(P::Any16)` by dispatching to the vector method: `invperm(P::Any16) = Tuple(invperm(collect(P)))`. 16 is just a conventional(?) choice as `Base.Any16` already exist.

i'm in favor of stronger optimizations for small tuples, but my method is always faster than base for larger ones, so why would you revert to base method for larger tuples?

StefanKarpinski:
> i'm trying to understand why tests are failing, is it unrelated?

That's very unlikely when all the builds are failing at the packaging step. Check out the logs here:

https://build.julialang.org/#/builders/53/builds/10921/steps/16/logs/stdio

You can click on "Details" on any of the failures to see that output. There's a doc generation failure.

tkf:
> (is it a strain for the compiler to manipulate larger tuples?)

Yes, `Any16` is a compile-time optimization, not a run-time one.  `Base` typically has some conservative cut off (e.g., `a1 + a2 + ... + a17` becomes type-unstable due to the cut off of `afoldl` I mentioned above).  I think it may be possible to convince core devs to increase the threshold (possibility to the infinity in this case?).  I just thought using `Any16` increases the chance it gets merged.

francescoalemanno:
i understand, okay :)

francescoalemanno:
@tkf, i still did not make the switching behavior after N=16, but the performance is impressive, this code constant folds at any N, very nice!

latest benchmark for https://github.com/JuliaLang/julia/pull/35234/commits/10caca556e676b9344def4e12e4029fba13746d6
```
[ Info: Testing size 3, base methods (invperm,isperm):
  261.941 ns (3 allocations: 256 bytes)
  34.989 ns (2 allocations: 128 bytes)
[ Info: Testing size 3, this PR (invperm,isperm):
  0.019 ns (0 allocations: 0 bytes)
  0.019 ns (0 allocations: 0 bytes)
[ Info: Testing size 6, base methods (invperm,isperm):
  322.317 ns (3 allocations: 320 bytes)
  40.945 ns (2 allocations: 128 bytes)
[ Info: Testing size 6, this PR (invperm,isperm):
  0.020 ns (0 allocations: 0 bytes)
  0.019 ns (0 allocations: 0 bytes)
[ Info: Testing size 9, base methods (invperm,isperm):
  364.034 ns (3 allocations: 400 bytes)
  43.915 ns (2 allocations: 128 bytes)
[ Info: Testing size 9, this PR (invperm,isperm):
  0.019 ns (0 allocations: 0 bytes)
  0.019 ns (0 allocations: 0 bytes)
[ Info: Testing size 12, base methods (invperm,isperm):
  428.258 ns (3 allocations: 464 bytes)
  50.857 ns (2 allocations: 128 bytes)
[ Info: Testing size 12, this PR (invperm,isperm):
  0.019 ns (0 allocations: 0 bytes)
  0.020 ns (0 allocations: 0 bytes)
[ Info: Testing size 15, base methods (invperm,isperm):
  492.809 ns (3 allocations: 544 bytes)
  54.673 ns (2 allocations: 128 bytes)
[ Info: Testing size 15, this PR (invperm,isperm):
  0.019 ns (0 allocations: 0 bytes)
  0.019 ns (0 allocations: 0 bytes)
[ Info: Testing size 18, base methods (invperm,isperm):
  550.370 ns (3 allocations: 608 bytes)
  60.316 ns (2 allocations: 128 bytes)
[ Info: Testing size 18, this PR (invperm,isperm):
  0.020 ns (0 allocations: 0 bytes)
  0.019 ns (0 allocations: 0 bytes)
[ Info: Testing size 21, base methods (invperm,isperm):
  621.738 ns (3 allocations: 688 bytes)
  65.784 ns (2 allocations: 128 bytes)
[ Info: Testing size 21, this PR (invperm,isperm):
  0.019 ns (0 allocations: 0 bytes)
  0.019 ns (0 allocations: 0 bytes)
[ Info: Testing size 24, base methods (invperm,isperm):
  689.846 ns (3 allocations: 752 bytes)
  70.808 ns (2 allocations: 128 bytes)
[ Info: Testing size 24, this PR (invperm,isperm):
  0.019 ns (0 allocations: 0 bytes)
  0.020 ns (0 allocations: 0 bytes)
[ Info: Testing size 27, base methods (invperm,isperm):
  752.795 ns (3 allocations: 832 bytes)
  76.069 ns (2 allocations: 128 bytes)
[ Info: Testing size 27, this PR (invperm,isperm):
  0.019 ns (0 allocations: 0 bytes)
  0.020 ns (0 allocations: 0 bytes)
[ Info: Testing size 30, base methods (invperm,isperm):
  818.729 ns (3 allocations: 928 bytes)
  81.980 ns (2 allocations: 128 bytes)
[ Info: Testing size 30, this PR (invperm,isperm):
  0.019 ns (0 allocations: 0 bytes)
  0.020 ns (0 allocations: 0 bytes)
```

base/combinatorics.jl
francescoalemanno: > The whole time I have looked at this code, there has been a `throw` above the `return` so it is based on this I have commented.

i'm very sorry @KristofferC

base/combinatorics.jl
```diff
@@ -59,29 +66,31 @@ julia> isperm([1; 3])
...
 isperm(p::Tuple{Int}) = p[1] == 1
 isperm(p::Tuple{Int,Int}) = ((p[1] == 1) & (p[2] == 2)) | ((p[1] == 2) & (p[2] == 1))
 
+function isperm(P::NTuple{N,Integer}) where {N}
+    valn = Val(N)
+    _foldoneto(true, valn) do b,i
+        s = _foldoneto(false, valn) do s, j
+            s && return s
+            P[j]==i && return true
+            false
```
tkf: I think you can simplify this to

```suggestion
            s || P[j]==i
```

tkf:
> i'm also in favor of promoting `_foldoneto` to maybe `onetofoldl` what do you think?

Do you mean to expose this as a public API from `Base`? I think that'd be great. But that's an additional API and so I think core devs need to decide it. I think it'd smoother to just put it as an internal function in this PR and then do another PR to turn this into a public API.

As for the name, I was thinking that it was a synonym of `foldl(op, OneTo(N); init=init)`. So maybe `foldl_oneto`? I agree that keeping `l` was a better choice of name.  I think `onetofoldl` is fine too.

francescoalemanno:
@tkf
> Do you mean to expose this as a public API from `Base`? I think that'd be great. But that's an additional API and so I think core devs need to decide it. I think it'd smoother to just put it as an internal function in this PR and then do another PR to turn this into a public API.

i agree :)

> As for the name, I was thinking that it was a synonym of `foldl(op, OneTo(N); init=init)`. So maybe `foldl_oneto`? I agree that keeping `l` was a better choice of name. I think `onetofoldl` is fine too.

we'll see about the name :)

now i guess that for this PR all that is left to do is introduce cutoffs via `Any16`, i will try it soon enough

francescoalemanno:
@tkf it should be good now.
@StefanKarpinski @KristofferC

tkf:
It looks like `isperm` and `invperm` on tuples are tested only for `N=0,1,2`

https://github.com/JuliaLang/julia/blob/5ecc17f0e039d1752e5e0f5da6c805d8932f5c50/test/combinatorics.jl#L25-L36

I think we need to test it for `N` between 3 and 16. Maybe also a few test for `N > 16` just to make sure we hit the `Any16` method during the test.

francescoalemanno:
@tkf 
> I think we need to test it for `N` between 3 and 16. Maybe also a few test for `N > 16` just to make sure we hit the `Any16` method during the test.

i added randomized testing for N = 3 to 20, but i just realized that the Any16 methods are not being it, and currectly i do not have a solution for this

mcabbott:
I wonder if the fallback should specify the type, like so?
```
invperm(P::Any16) = Tuple(invperm(collect(P)))::typeof(P)
```
Although I don't have any tests which obviously care.

francescoalemanno:
> I wonder if the fallback should specify the type, like so?
> 
> ```
> invperm(P::Any16) = Tuple(invperm(collect(P)))::typeof(P)
> ```
> 
> Although I don't have any tests which obviously care.

it works, so i added it :)

tkf:
> ```
> invperm(P::Any16) = Tuple(invperm(collect(P)))::typeof(P)
> ```

I think this throws for inputs like a tuple of static numbers and `(UInt8(17), ntuple(identity, 16)...)`.  Not that last case is important, but I don't think there is a good reason to not support this.

If we want better inferrability I think it's better to just increase the threshold.  I think it'd be nice to introduce `Any32` or `Any64` and use it in other places too (like `+`).

francescoalemanno:
@tkf
> If we want better inferrability I think it's better to just increase the threshold. I think it'd be nice to introduce `Any32` or `Any64` and use it in other places too (like `+`).

alright, i removed the type annotation, think we get it merged?

francescoalemanno:
@StefanKarpinski @KristofferC 
does this look good now?

tkf (approved):
LGTM (just in case it helps)

francescoalemanno:
@StefanKarpinski 
are there any further suggestions?

StefanKarpinski:
No, looks ok to me. I don't know that I'm the right person to review, however.

francescoalemanno:
> No, looks ok to me. I don't know that I'm the right person to review, however.

Thank you @StefanKarpinski , should i simply wait? or, can you ping someone else to review/merge this?

StefanKarpinski:
@KristofferC, would you be willing to take a look?

francescoalemanno:
@JeffBezanson what do you think of this?

francescoalemanno:
Bump :)

PR #27310, opened 2018-05-29
inference: re-architect entry points for more clarity

vtjnash:
The main functional changes are:
 - More nearly separates the optimization steps from the inference loop
   (still some interleaving, due to wanting to cache work results eagerly)
 - Doesn't run `widen_all_const` on the IR before returning to the user
   during reflection (e.g. code-typed). This allows the user to run
   additional passes on it outside of Julia codegen,
   or simply to better see what inference knows.
 - There's now a flag in `Compiler.Params` for whether the global
   inference cache is applicable and should be used (read / write).
   Although this flag is currently not respected, however.

PR #33802, opened 2019-11-09
Update inference devdoc

Liozou:
This PR updates the developer's documentation for the compiler reloading trick, and the use of `typeinf_code`. I hope this is indeed what was intended.

PR #35236, opened 2020-03-23
remove `cached` argument from `typeinf_code` in devdocs

KristofferC:
Was removed in https://github.com/JuliaLang/julia/pull/27310


vtjnash:
Fixed in #33802

vtjnash:
Sorry, and thanks for reminding me to merge the other one

vtjnash:
Bah, merged the wrong one though. The correct fix was actually in https://github.com/JuliaLang/julia/pull/34869, but it now conflicts.

PR #35237, opened 2020-03-23
define hasmissing

arnavs:
Defines a function `hasmissing` to test if a collection or iterator has missing values. 

This is something I've had to define myself in a few projects. I think it would be useful. 

```
"""
    hasmissing(itr)

Checks if itr has missing values.
"""
hasmissing(itr) = any(ismissing, itr)
hasmissing(x::Base.SkipMissing) = false
```

```
@testset "hasmissing" begin
    @test hasmissing([1, 2, 3, 4]) == false
    @test hasmissing([1, 2, 3, missing, 5]) == true
    @test hasmissing(skipmissing([1, 2, missing, 4])) == false
end
```

### DIFF
--- a/base/missing.jl
+++ b/base/missing.jl
@@ -264,6 +264,7 @@ end
 
 Checks if itr has missing values.
 """
+hasmissing(x::Base.SkipMissing) = false
 hasmissing(itr) = ~(collect(skipmissing(itr)) == collect(itr))
 
 # Optimized mapreduce implementation
--- a/base/missing.jl
+++ b/base/missing.jl
@@ -259,6 +259,13 @@ keys(itr::SkipMissing) =
     v
 end
 
+"""
+    hasmissing(itr)
+
+Checks if itr has missing values.
+"""
+hasmissing(itr) = ~(collect(skipmissing(itr)) == collect(itr))
+
 # Optimized mapreduce implementation
 # The generic method is faster when !(eltype(A) >: Missing) since it does not need
 # additional loops to identify the two first non-missing values of each block
--- a/base/missing.jl
+++ b/base/missing.jl
@@ -264,8 +264,8 @@ end
 
 Checks if itr has missing values.
 """
+hasmissing(itr) = any(ismissing, itr)
 hasmissing(x::Base.SkipMissing) = false
-hasmissing(itr) = ~(collect(skipmissing(itr)) == collect(itr))
 
 # Optimized mapreduce implementation
 # The generic method is faster when !(eltype(A) >: Missing) since it does not need
--- a/base/missing.jl
+++ b/base/missing.jl
@@ -266,6 +266,7 @@ Checks if itr has missing values.
 """
 hasmissing(itr) = any(ismissing, itr)
 hasmissing(x::Base.SkipMissing) = false
+hasmissing(::AbstractArray{<:Number}) = false
 
 # Optimized mapreduce implementation
 # The generic method is faster when !(eltype(A) >: Missing) since it does not need
--- a/base/missing.jl
+++ b/base/missing.jl
@@ -266,7 +266,6 @@ Checks if itr has missing values.
 """
 hasmissing(itr) = any(ismissing, itr)
 hasmissing(x::Base.SkipMissing) = false
-hasmissing(::AbstractArray{<:Number}) = false
 
 # Optimized mapreduce implementation
 # The generic method is faster when !(eltype(A) >: Missing) since it does not need
--- a/test/missing.jl
+++ b/test/missing.jl
@@ -527,3 +527,10 @@ mutable struct Obj; x; end
     @test ismissing(wref[1] == missing)
     @test ismissing(missing == wref[1])
 end
+
+@testset "hasmissing" begin
+    @test hasmissing([1, 2, 3, 4]) == false
+    @test hasmissing([1, 2, 3, missing, 5]) == true
+    @test hasmissing(skipmissing(skipmissing([1, 2, 3, 4]))) == false
+    @test hasmissing(skipmissing([1, 2, missing, 4])) == false
+end
--- a/test/missing.jl
+++ b/test/missing.jl
@@ -531,6 +531,5 @@ end
 @testset "hasmissing" begin
     @test hasmissing([1, 2, 3, 4]) == false
     @test hasmissing([1, 2, 3, missing, 5]) == true
-    @test hasmissing(skipmissing(skipmissing([1, 2, 3, 4]))) == false
     @test hasmissing(skipmissing([1, 2, missing, 4])) == false
 end
--- a/base/exports.jl
+++ b/base/exports.jl
@@ -685,6 +685,7 @@ export
     something,
     isnothing,
     nonmissingtype,
+    hasmissing
 
 # time
     sleep,


base/missing.jl
```diff
@@ -259,6 +259,14 @@ keys(itr::SkipMissing) =
     v
 end
 
+"""
+    hasmissing(itr)
+
+Checks if itr has missing values.
+"""
+hasmissing(x::Base.SkipMissing) = false
+hasmissing(itr) = ~(collect(skipmissing(itr)) == collect(itr))
```
StefanKarpinski:
```suggestion
hasmissing(itr) = collect(skipmissing(itr)) != collect(itr)
```
This would be the more idiomatic way to write this. However, there are several issues with this: it collects the iterable twice, which doesn't work for many iterables and it's really inefficient.

StefanKarpinski: Wouldn't it be better to just do `missing in itr`?

test/missing.jl
```diff
@@ -527,3 +527,10 @@ mutable struct Obj; x; end
     @test ismissing(wref[1] == missing)
     @test ismissing(missing == wref[1])
 end
+
+@testset "hasmissing" begin
+    @test hasmissing([1, 2, 3, 4]) == false
+    @test hasmissing([1, 2, 3, missing, 5]) == true
+    @test hasmissing(skipmissing(skipmissing([1, 2, 3, 4]))) == false
```
StefanKarpinski: Why two `skipmissings`?

StefanKarpinski:
Why not just write `missing in itr` in general?

base/missing.jl
arnavs: Good point about iterables not supporting double collection. But I wasn't sure how else to support both iterables and concrete collections.

test/missing.jl
arnavs: I wanted to test if it worked on nested iterators. But I'll remove that.

arnavs:
@StefanKarpinski 

```
julia> missing ∈ [1, 2, 3, 4]
missing
```

So should the generic one just be `hasmissing(itr) = ismissing(missing \in itr)`?

arnavs:
Perhaps just `collect(skipmissing(x)) == x` and not have it work for iterators?

fredrikekre:
`any(ismissing, itr) `?

arnavs:
OK. Thought that collected the `itr`, but good if it doesn't. Will switch to that.

arnavs:
It looks like the CI failures are unrelated to the PR: 

```
include at ./Base.jl:17
_jl_invoke at /buildworker/worker/package_linux64/build/src/gf.c:2152 [inlined]
jl_apply_generic at /buildworker/worker/package_linux64/build/src/gf.c:2336
jl_apply at /buildworker/worker/package_linux64/build/src/julia.h:1685 [inlined]
do_call at /buildworker/worker/package_linux64/build/src/interpreter.c:369
eval_value at /buildworker/worker/package_linux64/build/src/interpreter.c:458
eval_stmt_value at /buildworker/worker/package_linux64/build/src/interpreter.c:409 [inlined]
eval_body at /buildworker/worker/package_linux64/build/src/interpreter.c:807
jl_interpret_toplevel_thunk at /buildworker/worker/package_linux64/build/src/interpreter.c:915
top-level scope at Base.jl:48
jl_toplevel_eval_flex at /buildworker/worker/package_linux64/build/src/toplevel.c:813
jl_eval_module_expr at /buildworker/worker/package_linux64/build/src/toplevel.c:180
jl_toplevel_eval_flex at /buildworker/worker/package_linux64/build/src/toplevel.c:639
jl_parse_eval_all at /buildworker/worker/package_linux64/build/src/ast.c:910
jl_load_rewrite at /buildworker/worker/package_linux64/build/src/toplevel.c:874
include at ./boot.jl:329
_jl_invoke at /buildworker/worker/package_linux64/build/src/gf.c:2169 [inlined]
jl_apply_generic at /buildworker/worker/package_linux64/build/src/gf.c:2336
jl_apply at /buildworker/worker/package_linux64/build/src/julia.h:1685 [inlined]
do_call at /buildworker/worker/package_linux64/build/src/interpreter.c:369
eval_value at /buildworker/worker/package_linux64/build/src/interpreter.c:458
eval_stmt_value at /buildworker/worker/package_linux64/build/src/interpreter.c:409 [inlined]
eval_body at /buildworker/worker/package_linux64/build/src/interpreter.c:807
jl_interpret_toplevel_thunk at /buildworker/worker/package_linux64/build/src/interpreter.c:915
top-level scope at sysimg.jl:3
jl_toplevel_eval_flex at /buildworker/worker/package_linux64/build/src/toplevel.c:813
jl_parse_eval_all at /buildworker/worker/package_linux64/build/src/ast.c:910
jl_load_rewrite at /buildworker/worker/package_linux64/build/src/toplevel.c:874
exec_program at /buildworker/worker/package_linux64/build/ui/repl.c:35
true_main at /buildworker/worker/package_linux64/build/ui/repl.c:108
main at /buildworker/worker/package_linux64/build/ui/repl.c:217
__libc_start_main at /lib64/libc.so.6 (unknown line)
_start at /buildworker/worker/package_linux64/build/usr/bin/julia (unknown line)
```

fredrikekre:
I meant that why do we need a function for this when you can just as well use `any(ismissing, itr)`?

arnavs:
Convenience? We have `isone` when we could calll `x == one(x)`. 

And functions are helpful in loads of places, e.g. `filter`.

johnnychen94:
I believe the early-return implementation would be generally faster:

```julia
function hasmissing(itr)
    for x in itr
        ismissing(x) && return true
	end
    return false
end
```

Some types have trivial results so checking it can be merely a no-op,

```julia
hasmissing(::AbstractArray{<:Number}) = false
```

BTW, if we're going to have `hasmissing`, perhaps we can have `hasnan` as well..

arnavs:
Yeah @johnnychen94, that's a good point. I was thinking we could do something with `nonmissingtype` too, but even `AbstractArray{<:Number}` should be good enough as a special case.

arnavs:
Does stepping through an iterator collect it?

JeffBezanson:
The special case isn't necessary:
```
julia> @code_llvm any(ismissing,[1])

;  @ reducedim.jl:662 within `any'
define i8 @julia_any_109(%jl_value_t addrspace(10)* nonnull align 16 dereferenceable(40)) {
top:
  ret i8 0
}
```
Also note `any` is already short-circuiting.

I generally don't like having a large proliferation of functions like this, but maybe this one is useful enough that it's justified. The idea of adding hasnothing, hasnan, hasone, haszero, ... certainly seems like too much. In these cases I think you have to demonstrate that a specific combination comes up very often, in contexts where the extra convenience is worth the extra bulk.

StefanKarpinski:
> Edit: Sorry, was being dumb. But this always returns `missing`.

I would argue that you're not being dumb, it's the fact that `in` propagates `missing` that is dumb. Seems like a major footgun to me, but I guess that ship has sailed.

StefanKarpinski:
> Does stepping through an iterator collect it?

Not sure what this means, but stepping through an iterator does, in general, mean that you can't iterate through it again.

arnavs:
> > Does stepping through an iterator collect it?
> 
> Not sure what this means, but stepping through an iterator does, in general, mean that you can't iterate through it again.

Yeah, that's what I meant. 

Since I don't want to consume the iterators then, I removed the special case and just left it as `hasmissing(x) = any(ismissing, x)`.

StefanKarpinski:
That also consumes the iterator. There is simply no way not to consume something and check what it contains. You might be able to avoid consuming in the case where `Missing` is disjoint from the eltype of the collection, but otherwise it just can't be done.

arnavs:
I see. OK, well, then this is fine (we don't need the special short-circuiting version.)

As for the utility of the function, I'll let you guys decide. But certainly I've seen cases where I've had to check if arrays and dataframes had `missing` in them (and choose behavior based on that.)

KristofferC:
Marking triage since it is adding a new export.

nickrobinson251:
If this does not go in Base, I think this can live in the Missings.jl package?

pdeffebach:
+1 to Missings.jl, since a lot of pretty "basic" functionality lives there, like `passmissing` etc.

base/missing.jl
```diff
@@ -259,6 +259,14 @@ keys(itr::SkipMissing) =
     v
 end
 
+"""
+    hasmissing(itr)
+
+Checks if itr has missing values.
+"""
+hasmissing(itr) = any(ismissing, itr)
```
bkamins: Maybe consider this?
```suggestion
hasmissing(itr) = eltype(itr) >: Missing && any(ismissing, itr)
```
as `any(ismissing, itr)` does not seem to always optimize-out iterating `itr` even if `eltype(itr) >: Missing` is `false` (sometimes it does).

bkamins:
I would have it in Missings.jl for now.

arnavs:
Ok that works for me

base/missing.jl
arnavs: Yeah this is good.

PR #35254, opened 2020-03-24
Add sizehint! to Dict constructor

### LINKED ISSUE #36334: Use sizehint! when creating Dict from Generator
Dictionaries can be [created from generator expressions](https://en.wikibooks.org/wiki/Introducing_Julia/Dictionaries_and_sets#Creating_dictionaries), but don't seem to use `sizehint!`.  This could give better performance.

I couldn't confirm this in the source code, so I asked on [discourse](https://discourse.julialang.org/t/do-comprehensions-implicitly-use-sizehint/41452) where a benchmark showed fewer allocations when manually calling `sizehint!` than when using a generator expression.

```jl
using BenchmarkTools

println("Implicit:")
@btime begin
    dict = Dict(string(i) => sind(i) for i = 0:5:360)
end

println("Explicit:")
@btime begin
    dict = Dict{String, Int}()
    sizehint!(dict, 360÷5)
    for i in 0:5:360
        dict[string(i)] = i
    end
end
```
Results in
```
Implicit:
  8.669 μs (156 allocations: 13.42 KiB)
Explicit:
  5.737 μs (153 allocations: 9.70 KiB)
```

harryscholes:
The constructor `Dict{K,V}(kv)` does not provide a size hint, whereas the `Dict{K,V}(ps::Pair...)` constructor does. Is there any reason for this? I just tried benchmarking with and without `sizehint!` for `Dict{K,V}(kv)` and see an improvement when providing a size hint:

```julia
julia> using BenchmarkTools

julia> n = 1_000_000;

julia> kv = collect(zip(rand(n), rand(n)));

julia> @btime Dict{Float64,Float64}($kv);
  55.698 ms (54 allocations: 65.17 MiB)

julia> @eval Base begin
       function Dict{K,V}(kv) where V where K
           h = Dict{K,V}()
           sizehint!(h, length(kv))
           for (k,v) in kv
              h[k] = v
           end
           return h
       end
       end

julia> @btime Dict{Float64,Float64}($kv);
  40.509 ms (13 allocations: 51.00 MiB)
```

Closes #36334

### DIFF
--- a/base/dict.jl
+++ b/base/dict.jl
@@ -99,7 +99,10 @@ mutable struct Dict{K,V} <: AbstractDict{K,V}
 end
 function Dict{K,V}(kv) where V where K
     h = Dict{K,V}()
-    sizehint!(h, length(kv))
+    chklen = IteratorSize(kv)
+    if (chklen isa HasLength) || (chklen isa HasShape)
+        sizehint!(h, length(kv))
+    end
     for (k,v) in kv
         h[k] = v
     end
--- a/base/dict.jl
+++ b/base/dict.jl
@@ -99,6 +99,7 @@ mutable struct Dict{K,V} <: AbstractDict{K,V}
 end
 function Dict{K,V}(kv) where V where K
     h = Dict{K,V}()
+    sizehint!(h, length(kv))
     for (k,v) in kv
         h[k] = v
     end


JeffBezanson:
The problem is that not all iterators implement `length`. This needs to check `IteratorSize`.

harryscholes:
Ah yes, true. I've now added a check for whether `length` is implemented.

rfourquet (approved):

base/dict.jl
```diff
@@ -99,6 +99,10 @@ mutable struct Dict{K,V} <: AbstractDict{K,V}
 end
 function Dict{K,V}(kv) where V where K
     h = Dict{K,V}()
+    chklen = IteratorSize(kv)
+    if (chklen isa HasLength) || (chklen isa HasShape)
```
StefanKarpinski, line 103: Are both `HasLength` and `HasShape` necessary here? Wouldn't length alone be sufficient?

rfourquet: No, because `HasLength` is not a subtype of `HasShape`, but morally all `HasShape` implement `length`.

StefanKarpinski: I have to confess, I didn't follow the Cambrian explosion of collections traits, but it seems kind of bizarre that checking for `HasLength` is not sufficient to know if you can call `length` on something. But carry on 🤷‍♂️

harryscholes: Btw I stole this directly from https://github.com/JuliaLang/julia/blob/453a7ddcb6894cfcd8fc016634857b39416585cc/base/asyncmap.jl#L127-L128

garborg: I've seen `Base.haslength(iter)` used as well: https://github.com/JuliaLang/julia/blob/a9be54d1b089182a6b079224e9cab58a318b9541/base/generator.jl#L98

harryscholes: Thanks @garborg! I wasn't aware of `haslength`

PR #27401, opened 2018-06-03
WIP: Hilbert/Banach space structures

ranocha:
This is based on the discussion in JuliaLang/LinearAlgebra.jl#496 and is work in progress. It is my first pull request to julia. Thus, please be so kind to help me with improving this PR. Of course, it can also be used as a starting point for more experienced people.

The discussion in JuliaLang/LinearAlgebra.jl#496 seems to have reached a consensus about the naming of `norm`, `vecnorm`, and a new `opnorm`. Basically, `norm` should replace the old `vecnorm`, since it is used in most cases to estimate some error or magnitude. Thus, avoiding the cost of the SVD for matrices seems to be good. Moreover, most users are probably not aware of `vecnorm` and just stick to `norm` since it is okay in some cases. The old behaviour of `norm(A::AbstractMatrix, p)`, i.e. the  computation of the operator/matrix norm, is encoded in `opnorm`. Jutho has mentioned that he started implementing some similar behaviour in JuliaLang/julia#25093. However, I can not see any commits or changed files there, so this is started from scratch.

~~Moreover, the behaviour of generic `norm` functions is changed. Before, it has been along the lines of~~
```julia
norm(x, p) = sum(xi->norm(xi)^p, x)^(1/p).
```
~~Since it seems to be more natural (at least for me), I have changed it to~~
```julia
norm(x, p) = sum(xi->norm(xi, p)^p, x)^(1/p).
```
~~Thus, given the same coefficients, calling `norm` on a vector, a matrix or a vector of vectors yields the same result, which seems to be consistent in some sense.~~
```julia
julia> v = [3, -2, 6]
3-element Array{Int64,1}:
  3
 -2
  6

julia> norm(hcat(v,v), 1) == norm(vcat(v,v), 1) == norm([v,v], 1)
true

julia> norm(hcat(v,v), 2) == norm(vcat(v,v), 2) == norm([v,v], 2)
true

julia> norm(hcat(v,v), Inf) == norm(vcat(v,v), Inf) == norm([v,v], Inf)
true
```

[...]

PR #35257, opened 2020-03-25
make (*)(::Adjoint{Vector}, ::Vector) non-recursive

### LINKED ISSUE #35174: (*)(::Adjoint{Vector}, ::Vector) should use a non-recursive dot product
This is somewhat tightly related to https://github.com/JuliaLang/julia/pull/22220, but I think a bit different:
```julia
julia> σ = [[0  1
             1  0],
            [0 -im
             im 0],
            [1  0
             0  -1]]
3-element Array{Array{Complex{Int64},2},1}:
 [0 + 0im 1 + 0im; 1 + 0im 0 + 0im]
 [0 + 0im 0 - 1im; 0 + 1im 0 + 0im]
 [1 + 0im 0 + 0im; 0 + 0im -1 + 0im]

julia> [1 2 3] * σ
1-element Array{Array{Complex{Int64},2},1}:
 [3 + 0im 1 - 2im; 1 + 2im -3 + 0im]

julia> [1,2,3]'* σ
ERROR: DimensionMismatch("x and y are of different lengths!")
Stacktrace:
 [1] dot at /Users/mason/julia/usr/share/julia/stdlib/v1.4/LinearAlgebra/src/generic.jl:879 [inlined]
 [2] dot(::Array{Int64,1}, ::Array{Array{Complex{Int64},2},1}) at /Users/mason/julia/usr/share/julia/stdlib/v1.4/LinearAlgebra/src/generic.jl:894
 [3] *(::Adjoint{Int64,Array{Int64,1}}, ::Array{Array{Complex{Int64},2},1}) at /Users/mason/julia/usr/share/julia/stdlib/v1.4/LinearAlgebra/src/adjtrans.jl:245
 [4] top-level scope at REPL[46]:1
```
Here we see that `[1 2 3] * σ` did something quite different from `[1,2,3]'* σ`: it essentially did what one should expect given the definitions you see in a linear algebra class, but `[1,2,3]'* σ` tries to recursively apply `dot` down to each level of the arrays which produces the above error.

To make matters worse, if I try to 'lift' my vector `[1, 2, 3]` to the same eltype as `σ`, I get a disastrously 'wrong' answer:
```julia
julia> 𝟙 = [1 0; 0 1];

julia> [1*𝟙, 2*𝟙, 3*𝟙]' * σ
0 + 0im
```
This seems to come down to the fact that `dot` is recursive, but I'd argue that even if we want `dot` to be recursive, `*` should **not** be recursive.

Perhaps the solution is a non-recursive `dot` that `*` would lower to instead of the recursive form? 

This would be unfortunately breaking.

______

Thanks @ericphanson for making clear to me that the source of my issue is the recursive part.

dkarrasch:
Closes #35174.

These changes don't break any tests, so I believe this was more oversight than deep intention. With this PR, I get
```julia
julia> using LinearAlgebra

julia> σ = [[0  1
                    1  0],
                   [0 -im
                    im 0],
                   [1  0
                    0  -1]]
3-element Array{Array{Complex{Int64},2},1}:
 [0 + 0im 1 + 0im; 1 + 0im 0 + 0im]
 [0 + 0im 0 - 1im; 0 + 1im 0 + 0im]
 [1 + 0im 0 + 0im; 0 + 0im -1 + 0im]

julia> [1,2,3]'* σ
2×2 Array{Complex{Int64},2}:
 3+0im   1-2im
 1+2im  -3+0im
```
which used to throw before.

### DIFF
--- a/stdlib/LinearAlgebra/src/adjtrans.jl
+++ b/stdlib/LinearAlgebra/src/adjtrans.jl
@@ -242,9 +242,9 @@ Broadcast.broadcast_preserving_zero_d(f, tvs::Union{Number,TransposeAbsVec}...) 
 ## multiplication *
 
 # Adjoint/Transpose-vector * vector
-*(u::AdjointAbsVec, v::AbstractVector) = dot(u.parent, v)
+*(u::AdjointAbsVec{T}, v::AbstractVector{T}) where {T<:Number} = dot(u.parent, v)
 *(u::TransposeAbsVec{T}, v::AbstractVector{T}) where {T<:Real} = dot(u.parent, v)
-function *(u::TransposeAbsVec, v::AbstractVector)
+function *(u::AdjOrTransAbsVec, v::AbstractVector)
     require_one_based_indexing(u, v)
     @boundscheck length(u) == length(v) || throw(DimensionMismatch())
     return sum(@inbounds(u[k]*v[k]) for k in 1:length(u))
--- a/stdlib/LinearAlgebra/src/adjtrans.jl
+++ b/stdlib/LinearAlgebra/src/adjtrans.jl
@@ -244,11 +244,8 @@ Broadcast.broadcast_preserving_zero_d(f, tvs::Union{Number,TransposeAbsVec}...) 
 # Adjoint/Transpose-vector * vector
 *(u::AdjointAbsVec{T}, v::AbstractVector{T}) where {T<:Number} = dot(u.parent, v)
 *(u::TransposeAbsVec{T}, v::AbstractVector{T}) where {T<:Real} = dot(u.parent, v)
-function *(u::AdjOrTransAbsVec, v::AbstractVector)
-    require_one_based_indexing(u, v)
-    @boundscheck length(u) == length(v) || throw(DimensionMismatch())
-    return sum(@inbounds(u[k]*v[k]) for k in 1:length(u))
-end
+*(u::AdjOrTransAbsVec, v::AbstractVector) = sum(uu*vv for (uu, vv) in zip(u, v))
+
 # vector * Adjoint/Transpose-vector
 *(u::AbstractVector, v::AdjOrTransAbsVec) = broadcast(*, u, v)
 # Adjoint/Transpose-vector * Adjoint/Transpose-vector
--- a/stdlib/LinearAlgebra/test/matmul.jl
+++ b/stdlib/LinearAlgebra/test/matmul.jl
@@ -550,7 +550,7 @@ end
     @test  transpose(Xv2)*Xv2     ≈ XtX
     @test (transpose(Xv3)*Xv3)[1] ≈ XtX
     @test  Xv1'*Xv1     ≈ XcX
-    @test Xv2'*Xv2 ≈ norm(Xv2)^2
+    @test  Xv2'*Xv2     ≈ XcX
     @test (Xv3'*Xv3)[1] ≈ XcX
     @test (Xv1*transpose(Xv1))[1] ≈ XXt
     @test  Xv2*transpose(Xv2)     ≈ XXt


dkarrasch:
Currently, we don't have any tests for arrays of arrays related to the issue here.

ericphanson:
I don’t think the old behavior is really coherent; as I see it, the fact that `*(::Adjoint, ::Matrix)` dispatched to `dot` was a leaking implementation detail. And I don’t see how one could formulate a coherent system for array-of-array operations from which you would get the rule that `*(::Adjoint{Vector{T}}, Vector{Matrix{T}}) where T <: Number` gives a number.

To me, the general rule should be `*(::Matrix{Matrix{T}}, ::Matrix{Matrix{T}}) where T <: Number` should dispatch (semantically at least) to a generic matmul that uses `*` and `+` for `Matrix{T}`, and similarly for other linear algebra operations. In other words, it should just lower one level at a time by using the operations as defined for the lower level.

The recursive `dot` fits this rule if you define `dot(u,v) = sum( dot(u[i], v[i]) for i in eachindex(u,v))` and `dot` for numbers as conjugated multiplication. And the fix here, as I understand it, fits this rule too.

simeonschaub:
I would propose a small change to @ericphanson's proposal, to change that to `(*)(u::Adjoint{Vector}, v::Vector) = sum(u[k]'v[k] for k in eachindex(u,v))`, which would work for cases like this without any other changes to `dot` for matrices and numbers and would be consistent with the recursive behavior of `Adjoint` everywhere else.

dkarrasch:
Oops, I overlooked one error. Fixing it is instructive, as the failing test really sticks out of the surrounding tests. I found that it was changed in #27401, but the thread is very long. I'm pinging @ranocha to check what was the rationale there.

dkarrasch:
@simeonschaub I think that the recursive transposition behavior is coded into the `getindex` for `Adjoint`s.

simeonschaub:
Oops, yes, of course! Then I definitely agree with the implementation here

dkarrasch:
Actually, currently we have
```julia
(*)(u::TransposeAbsVec, v::AdjointAbsVec{<:Any,<:TransposeAbsVec}) =
    sum(uu*vv for (uu, vv) in zip(u, v))
(*)(u::AdjointAbsVec,   v::AdjointAbsVec{<:Any,<:TransposeAbsVec}) =
    sum(uu*vv for (uu, vv) in zip(u, v))
(*)(u::TransposeAbsVec, v::TransposeAbsVec{<:Any,<:AdjointAbsVec}) =
    sum(uu*vv for (uu, vv) in zip(u, v))
(*)(u::AdjointAbsVec,   v::TransposeAbsVec{<:Any,<:AdjointAbsVec}) =
    sum(uu*vv for (uu, vv) in zip(u, v))
```
Should we simplify the `*(u::AdjOrTransAbsVec, v::AbstractVector)` method to the same
```julia
sum(uu*vv for (uu, vv) in zip(u, v))
```
without bounds checking and `require_one_based_indexing`? Feels strange to be so strict in one case and so relaxed in so many others...

ranocha:
> Oops, I overlooked one error. Fixing it is instructive, as the failing test really sticks out of the surrounding tests. I found that it was changed in #27401, but the thread is very long. I'm pinging @ranocha to check what was the rationale there.

The intention of #27401 was to make `dot` really an inner/scalar product for various things and use the behavior of the old `vecdot` (and similarly for `norm`). Hence, `dot(Xv2, Xv2) ≈ norm(Xv2)^2` is the desired behavior. The test 
https://github.com/JuliaLang/julia/blob/cc60547e7f046272a5dcfdc013a50e70f47657e8/stdlib/LinearAlgebra/test/matmul.jl#L553 
reflects that since `*(::Adjoint, ::Matrix)` dispatched to `dot`, which is not necessarily an optimal choice, as mentioned by @ericphanson above. I didn't change `*(::Adjoint, ::Matrix)`, only `dot` in #27401. Hence, I think the rational behind this PR is fine.

marius311:
Hi guys, I ran into a few issues with this PR as I was upgrading my code to 1.5: 

1) I don't think the dimentionality checking should be removed. The 1.5 behavior now is just bizarre:

    ```julia
    # julia 1.4
    julia> [1, 2, 3]' * [1., 2.]
    ERROR: DimensionMismatch("first array has length 3 which does not match the length of the second, 2.")

    # julia 1.5rc
    julia> [1, 2, 3]' * [1., 2.]
    5.0
    ```
    and lots of `dot` implementations in `LinearAlgebra/src/generic.jl` have dimension checking. 


2) E.g. `(*)(::Adjoint{Vector{Int}}, Vector{Float})` uses slightly different summation code rather than the existing https://github.com/JuliaLang/julia/blob/release-1.4/stdlib/LinearAlgebra/src/generic.jl#L895-L897. I've benchmarked the two for simple builtin types (as you may have too) and I don't see any performance regressions, but are we sure there's no types this might cause an issue? 

3) The fact that mixed-but-still-scalar `eltype`s go down different codepaths introduces what feels like an unnecessary breaking API change because whereas previously it sufficed to override `dot(x,y)` for your custom types, now you have to do both `dot(::MyVector, ::MyVector)` and `(*)(::Adjoint{MyVector}, ::MyVector}`. A solution might be 
    ```julia
    *(u::AdjointAbsVec{<:Number}, v::AbstractVector{<:Number})
    ```
    instead of
    ```julia
    *(u::AdjointAbsVec{T}, v::AbstractVector{T}) where {T}
    ```
    in line 270, but its not totally clear.  

My impression is that while the original issue certainly deserves to be solved, maybe this PR needed to be iterated on a little more before merging? Would it be possible to fix it quickly before 1.5, or maybe revert and iterate on it a bit? For me certainly the breaking API change, while I think technically allowed by Julia's semver rules as they pertain to stdlibs, seems worrisome.

KristofferC:
cc @dkarrasch

Issue #35226, opened 2020-03-23
Doc string for Base.Generator is slightly clumsy


jlapeyre:
"The syntax `f(x) for  x in iter [if cond(x)::Bool]` is syntax for constructing an instance of this type."

is better written

"The syntax for constructing an instance of this type is `f(x) for  x in iter [if cond(x)::Bool] `."

r0cketr1kky:
I would like to take this up if possible.

StefanKarpinski:
Go for it.

PR #35259, opened 2020-03-25
Updated Doc string in Base.Generator 

r0cketr1kky:
Solves #35226


JeffBezanson:
First, I don't think `if` guards should be mentioned here, since if you have one you don't get a `Generator` but a `Filter` iterator. (I realize this PR didn't introduce that, but we might as well fix it while we're at it.)

The original wording was also chosen for a reason. The point is that a `Generator` happens to be what you get when you write `x for x in y`. It's not the only syntax for constructing a `Generator`; `Generator(f, itr)` also works like for any type.

PR #35260, opened 2020-03-25
Inline and use a non-splatting combine_axes in broadcast

ChrisRackauckas:
DifferentialEquations.jl has been carrying around a patch

https://github.com/SciML/DiffEqBase.jl/blob/master/src/diffeqfastbc.jl#L18

for awhile which stops allocations like https://discourse.julialang.org/t/strange-allocations-during-broadcasting/26605/5 due to lack of inlining and due to splatting. It would probably be good to upstream this?

@mbauman

### DIFF
--- a/base/broadcast.jl
+++ b/base/broadcast.jl
@@ -472,7 +472,7 @@ julia> Broadcast.combine_axes(1, 1, 1)
 """
 @inline combine_axes(A, B...) = broadcast_shape(axes(A), combine_axes(B...))
 @inline combine_axes(A, B) = broadcast_shape(axes(A), axes(B))
-@inline combine_axes(A) = axes(A)
+combine_axes(A) = axes(A)
 
 # shape (i.e., tuple-of-indices) inputs
 broadcast_shape(shape::Tuple) = shape
--- a/base/broadcast.jl
+++ b/base/broadcast.jl
@@ -471,7 +471,7 @@ julia> Broadcast.combine_axes(1, 1, 1)
 ```
 """
 @inline combine_axes(A, B...) = broadcast_shape(axes(A), combine_axes(B...))
-@inline combine_axes(A, B) = broadcast_shape(broadcast_axes(A), broadcast_axes(B))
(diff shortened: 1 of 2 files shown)


base/broadcast.jl
```diff
@@ -471,7 +471,8 @@ julia> Broadcast.combine_axes(1, 1, 1)
 ```
 """
 @inline combine_axes(A, B...) = broadcast_shape(axes(A), combine_axes(B...))
-combine_axes(A) = axes(A)
+@inline combine_axes(A, B) = broadcast_shape(broadcast_axes(A), broadcast_axes(B))
```
KristofferC:
```suggestion
@inline combine_axes(A, B) = broadcast_shape(axes(A), axes(B))
```

base/broadcast.jl
```diff
@@ -471,7 +471,8 @@ julia> Broadcast.combine_axes(1, 1, 1)
...
+@inline combine_axes(A) = axes(A)
```
KristofferC: I don't think the `@inline` here matters.

ChrisRackauckas:
This code was from @YingboMa so he might comment as well.

base/broadcast.jl
ChrisRackauckas: yeah, I'll remove it.

mbauman:
I've found it helpful to add `@allocated` tests to ensure these sorts of crucial things don't regress.  They can sometimes be a pain, but I think a case like this is worth it.  As simple as:

```Julia
u = rand(100)
k1 = similar(u)
k2 = similar(u)
k3 = similar(u)
k4 = similar(u)
f(a,b,c,d,e) = @. a = a + 1*(b+c+d+e)
@allocated f(u,k1,k2,k3,k4)
@test (@allocated f(u,k1,k2,k3,k4)) == 0
```

mbauman (approved):

YingboMa (commented):
Looks good to me.

PR #35261, opened 2020-03-25
updated Doc string in Base.Generator

KristofferC:
Tweaks based on JuliaLang/julia#35259


JeffBezanson:
We could just do another round of edits instead of reverting, I think.

KristofferC:
Removed the filter stuff if that is what you meant.

jonas-schulze:
I personally liked the old phrasing more. That `Generator(f, itr)` is another way to get an instance of that type is already mentioned in the first line of that docstring. Thus I don’t think the wording needs to be that implicitly vage.

Maybe smth like this would be a little more explicit about that:

`f(x) for x in itr` is syntactic sugar for creating an instance of this type. If you append a guard, you get an instance of [`Filter`](@ref) instead.

JeffBezanson:
> If you append a guard, like `f(x) for x in itr if p(x)`, you get an instance of `[Filter](@ref)` instead.

Not true; try it.

I don't see why this is the best place to document `Filter` specifically:

- When you write a guard the overall iterator is still a `Generator`, the filter is applied to the inner iterator.
- If there are multiple iterators you get a `Product` iterator.
- With multiple `for`s you get a `Flatten` iterator.

I propose we add full documentation on generator syntax to the docstring for the `for` keyword. Then this one can just mention the `Generator` part, with a link to the full thing. Similar references can be added to `Filter`, `Flatten`, and `Product` if somebody's up to it.

jonas-schulze:
> > If you append a guard, like `f(x) for x in itr if p(x)`, you get an instance of `[Filter](@ref)` instead.
> 
> Not true; try it.

I just read [your comment](https://github.com/JuliaLang/julia/pull/35259#issuecomment-604025848) on #35259 an didn't try it myself:

> if you have [a guard] you don't get a Generator but a Filter iterator.

which I thought would be an interesting addition to the documentation. Now I tried and can confirm that it's indeed a `Generator`, so all the more I think it should be mentioned in here.


> If there are multiple iterators you get a Product iterator.

How do you do that? I tried `Iterators.product`, but that leads to a `Generator`.

> I propose we add full documentation on generator syntax to the docstring for the for keyword. Then this one can just mention the Generator part, with a link to the full thing.

I agree.

PR #33680, opened 2019-10-26
Fix obsolete builtins list in devdocs/function

o314:
WAS

```julia
ns_str = """
=== typeof sizeof <: isa typeassert throw tuple getfield setfield! fieldtype
nfields isdefined arrayref arrayset arraysize applicable invoke apply_type _apply
_expr svec
"""
```

BUT

```julia
ns = ns_str |> strip |> split |> s -> map(Symbol,s)
ns2 = [n for n in names(Core) if getfield(Core,n) isa Core.Builtin] # 15 without private
ns2 = [n for n in names(Core;all=true) if getfield(Core,n) isa Core.Builtin] # 29 with private

@test setdiff(ns2, ns) == [ :_apply_latest, :_apply_pure, :_typevar, :const_arrayref, :getproperty, :ifelse, :setproperty!] # with private
```

PR #35271, opened 2020-03-27
Fix obsolete builtins list in devdocs/function (dup/rebase of #33680)

o314:
Fix obsolete builtins list in devdocs/function with dyn code in doc

Actually a dup of #33680 Rather a dup for rebasing
*i get wind of unfixable errors may be due to obsolescence of the main codebase since proposal*

### DIFF
--- a/doc/src/devdocs/functions.md
+++ b/doc/src/devdocs/functions.md
@@ -120,7 +120,8 @@ function lines(words)
     String(take!(io))
 end
 import Markdown
-[string(n) for n in names(Core;all=true) if getfield(Core,n) isa Core.Builtin] |>
+[string(n) for n in names(Core;all=true)
+    if getfield(Core,n) isa Core.Builtin && nameof(getfield(Core,n)) === n] |>
     lines |>
     s ->  "```\n$s\n```" |>
     Markdown.parse
--- a/doc/src/devdocs/functions.md
+++ b/doc/src/devdocs/functions.md
@@ -101,10 +101,29 @@ currently share a method table via special arrangement.
 
 The "builtin" functions, defined in the `Core` module, are:
 
-```
-=== typeof sizeof <: isa typeassert throw tuple getfield setfield! fieldtype
-nfields isdefined arrayref arrayset arraysize applicable invoke apply_type _apply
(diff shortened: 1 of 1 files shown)


o314:
Last errors in  #33680 were too hard to catch (time goes by, codebase become a bit more older).

A new patch #35271 has been proposed instead of.
It successfully had passed all ci tests.

Could you review it when available, please @Keno @fredrikekre ?

JeffBezanson:
This should also include the condition `nameof(getfield(Core,n)) === n`, because `Core.getproperty` is an alias of `getfield` (also `setproperty!`).

o314:
I guess there is no reflection available for those functions at the boundary between c and julia runtime.
I will provide extensional definitions for aliases (currently 1), intensional for defs (the current patch).

JeffBezanson:
The aliases should be excluded. There is a well-defined set of Builtin functions, and each has a name visible via `nameof`. One way to think of it is (1) find everything you can of type `Builtin`, (2) call `unique` on that, (3) call `nameof` on each. (Doesn't have to be implemented that way of course.)

o314:
Back from work. I have forgiven this point.

```julia
a = [n for n in names(Core;all=true) if getfield(Core,n) isa Core.Builtin]

b = [n for n in names(Core;all=true)
        if getfield(Core,n) isa Core.Builtin && nameof(getfield(Core,n)) === n ]

c = [n for n in names(Core;all=true) if getfield(Core,n) isa Core.Builtin] |>
    s -> (n -> (nameof(getfield(Core,n)), n)).(s) |>
    s -> unique(t->t[1], s) |>
    s -> (t->t[2]).(s)
    
using Test
@test length(a) > length(b)
@test b == c
```

b is simpler to me. 1 filter with 2 clauses, no need to call unique.
for information `getproperty` and `setproperty!` are concerned.
i will update with that.

vtjnash:
nearly equivalent also to `join(sort(subtypes(Core.Builtin) .|> (s -> String(s.name.mt.name))), " ")`, so I do not precisely know why this is in the devdocs, when the help system already generates this list automatically, but seems correct anyways

Issue #34264, opened 2020-01-05
Bug report: `make install` does not copy `libLLVM-8jl.so` on Linux Environment (Julia-1.5 Dev)


terasakisatoshi:
- Since `make install` does not copy `libLLVM-8jl.so` (which wll be generated by running `make` command) on linux environment, I can't run Julia that will be installed by `make install` .
- If I copy `libLLVM-8jl.so` manually to directory `path/to/prefix/lib/julia/` it will be alright on Linux environment.

# How to Reproduce

```Dockerfile
FROM ubuntu:18.04

RUN apt-get update && apt-get install -y \
    build-essential \
    libatomic1 \
    python \
    gfortran \
    perl \
    wget \
    m4 \
    cmake \
    pkg-config \
    git

ARG JL_VERSION="master"
ARG WDIR="/root"
ARG JL_BUILD_DIR=$WDIR/build
WORKDIR $JL_BUILD_DIR

RUN echo "\
CXXFLAGS=-D_GLIBCXX_USE_CXX11_ABI=0\n\
prefix=/usr/local/julia-$JL_VERSION\n\
USE_BINARYBUILDER = 0\n\
" > Make.user \
    && cat Make.user \
    && git clone --depth=1 -b $JL_VERSION https://github.com/JuliaLang/julia.git julia-$JL_VERSION\
    && cp Make.user $JL_BUILD_DIR/julia-$JL_VERSION \
    && cd julia-$JL_VERSION \
    && make -j $(nproc) \
    && make install

RUN rm -r $JL_BUILD_DIR && \
    apt-get clean && rm -rf /var/cache/apt/archives/* /var/lib/apt/lists/*
ENV PATH=/usr/local/julia-$JL_VERSION/bin:$PATH
# runtime test
RUN julia -e "using InteractiveUtils; versioninfo()" # it will fail

CMD ["julia"]
```

- Run the following command will occur an error.

```console
$ docker build -t buildmaster .
...
julia: error while loading shared libraries: libLLVM-8jl.so: cannot open shared object file: No such file or directory
```

terasakisatoshi:
- I think `JL_PRIVATE_LIBS-?` does not contain `libLLVM-8jl.so`
  - See: 
    - https://github.com/JuliaLang/julia/blob/master/Makefile#L337-L347
    - https://github.com/JuliaLang/julia/blob/master/Makefile#L182

terasakisatoshi:
Still same error happens

```
[...]

PR #35273, opened 2020-03-27
change libLLVM suffix to match build-configuration

nalimilan:
Backport of 5e0cab69966eb989f380492dda7590bca3a3ba8d.

Fixes https://github.com/JuliaLang/julia/issues/34264.


vchuravy:
This was due to a misconfiguration between Yggdrasil and base Julia. So we need to release a new LLVM v8 BB. This patch is fine if you just want to carry it locally until I get to that.

nalimilan:
So you mean it's not needed at all in release-1.4?

vchuravy:
> So you mean it's not needed at all in release-1.4?

It is needed in 1.4, but it needs a new release for BB which I am working on at https://github.com/JuliaLang/julia/pull/35300

vchuravy:
Part of  #35313

PR #35274, opened 2020-03-27
fix #35272, searchsorted(3:-1:1, 2.5, rev=true)

### LINKED ISSUE #35272: reverse searchsorted is inconsistent on 1.3
```julia
julia> searchsortedfirst(10:-1:1, 9.5; rev=true)
1

julia> searchsortedfirst(collect(10:-1:1), 9.5; rev=true)
2

julia> searchsortedlast(10:-1:1, 9.5; rev=true)
2

julia> searchsortedlast(collect(10:-1:1), 9.5; rev=true)
1
```

The range behaviour seems like the correct version, but needless to say it's extremely confusing getting different answers for these.

> searchsortedfirst(a, x; by=<transform>, lt=<comparison>, rev=false)

 > Return the index of the first value in a greater than or equal to x,
  according to the specified order. Return length(a) + 1 if x is greater than
  all values in a. a is assumed to be sorted.

It gets worse if you use all floats, now both the range and the array are wrong:

```julia
julia> searchsortedfirst(10.0:-1.0:1.0, 9.5; rev=true)
2

julia> searchsortedfirst(collect(10.0:-1.0:1.0), 9.5; rev=true)
2

julia> searchsortedlast(10.0:-1.0:1.0, 9.5; rev=true)
1

julia> searchsortedlast(collect(10.0:-1.0:1.0), 9.5; rev=true)
1
```

```julia
julia> versioninfo()
Julia Version 1.3.0
Commit 46ce4d7933 (2019-11-26 06:09 UTC)
Platform Info:
  CPU: Intel(R) Core(TM) i5-5300U CPU @ 2.30GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.1 (ORCJIT, broadwell)
Environment:
  JULIA_NUM_THREADS = 3
```

rfourquet:


rfourquet:
I like how github displays the commit title :)
funny-github

StefanKarpinski:
Probably backportable too?

rfourquet:
> Probably backportable too?

Not as-is unfortunately, as the code has changed since 1.4. I will have a look to see if I can make a similar PR against 1.4.

PR #35275, opened 2020-03-27
fix a bug in circular type detection

JeffBezanson:
Fixes one of the issues in #34126

### DIFF
--- a/src/datatype.c
+++ b/src/datatype.c
@@ -279,25 +279,25 @@ int jl_pointer_egal(jl_value_t *t)
     return 0;
 }
 
-static int references_name(jl_value_t *p, jl_typename_t *name) JL_NOTSAFEPOINT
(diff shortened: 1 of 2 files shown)


vtjnash:
Ah, right. This was introduced in https://github.com/JuliaLang/julia/pull/34223. I'm still in general skeptical (more so now that I see the mistake there that I missed on previous review) of this half-ways approach to switching this function over from the conservative approximation first implemented, to the more aggressive form (of actually checking the field types) that it could do.

CI failure is just a repr mistake with defining this test:
```
  Expression: repr(NFANode34126()) == "NFANode34126(Tuple{Nothing,NFANode34126}[])"
   Evaluated: "Main.Test78Main_core.NFANode34126(Tuple{Nothing,Main.Test78Main_core.NFANode34126}[])" == "NFANode34126(Tuple{Nothing,NFANode34126}[])"
```

PR #35277, opened 2020-03-27
Patch REPL documentation for symbols ending with '='

Liozou:
Hi there!

While randomly idling in the REPL, I was amazed to discover that:
```
help?> ==
search: == === !==

  x == y is a synonym for x = x = y

help?> <=
search: <=

  x <= y is a synonym for x = x < y

help?> .<=
search:

  x .<= y is a synonym for x = x .< y
```
This was good fun! But after an instant of dreadful uncertainty, I decided it could not be right, so here is a PR to reestablish the truth.

I believe this was unintentionally introduced in #35154. If it is intentional, let me know and I will simply re-read the entire Julia manual to make sense of things...

By the way the fix is not very nice to the eye, if there is a way get all the comparison operators in one place I would gladly use that instead.

### DIFF
--- a/stdlib/REPL/src/docview.jl
+++ b/stdlib/REPL/src/docview.jl
@@ -208,7 +208,7 @@ function lookup_doc(ex)
     end
     if isa(ex, Symbol) && Base.isoperator(ex)
         str = string(ex)
-        if endswith(str, "=")
+        if endswith(str, "=") && Base.operator_precedence(ex) == Base.prec_assignment
             op = str[1:end-1]
             return Markdown.parse("`x $op= y` is a synonym for `x = x $op y`")
         elseif startswith(str, ".")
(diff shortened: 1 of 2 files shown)


JeffBezanson:
Oops, my bad. Sorry about that everyone.

JeffBezanson (approved):

stdlib/REPL/src/docview.jl
```diff
@@ -208,7 +208,7 @@ function lookup_doc(ex)
     end
     if isa(ex, Symbol) && Base.isoperator(ex)
         str = string(ex)
-        if endswith(str, "=")
+        if endswith(str, "=") && ex ∉ (:(==), :(===), :(!=), :(!==), :(<=), :(>=), :(.==), :(.===), :(.!=), :(.!==), :(.<=), :(.>=))
```
mbauman: You could ask if `Base.operator_precedence(ex) == Base.prec_assignment` perhaps

Liozou: Thanks, that's a much more elegant alternative! I just implemented that in the new commit

Liozou:
I used @mbauman's suggestion to select the operators ending with '=' that were actually assignment operators and CI is green, let me know if you want me to add anything!

PR #30778, opened 2019-01-20
add efficient `findfirst` method for `StepRange`

marekkukan-tw:
Several functions like `sum`, `reverse`, etc. have efficient (O(1)) methods for ranges implemented in Base.
This PR adds such method for `findfirst`, where the predicate is equality, and the iterable is a `StepRange`.

before
```julia-repl
julia> @btime findfirst(==(10^5), 1:3:10^6)
  27.386 μs (3 allocations: 64 bytes)
33334
```
after
```julia-repl
julia> @btime findfirst(==(10^5), 1:3:10^6)
  15.881 ns (0 allocations: 0 bytes)
33334
```

PR #35278, opened 2020-03-27
Use inefficient `findfirst` method for `StepRange{..., Year}`

### LINKED ISSUE #35203: Errors from findfirst(..., ::StepRange{Date,Year})
This worked on Julia 1.0 and 1.1, but fails from 1.2 to 1.5:
```julia
using Dates
ys = Date(1852):Year(4):Date(1940)
y0 = ys[21+1]
findfirst(isequal(y0), ys) # MethodError: Cannot `convert` an object of type Day to an object of type Year

findall(isequal(y0), ys) # fine
```
This seems like a bug to me, as the range ought to be treated the same way as the collected array. 

The method, which I think was added in https://github.com/JuliaLang/julia/pull/30778, is here:
https://github.com/JuliaLang/julia/blob/94662324259fe2d8d6b7929c937f7da5fa6f8e6a/base/array.jl#L1794

It works fine for some dates, like `DateTime(2017):Second(1):DateTime(2020)` mentioned in the PR. But every second is the same length, while years vary.

mcabbott:
Fixes #35203, by avoiding the method added in #30778 for ranges of step Month or Year. 

The alternative would be to add a check of `Base.RangeStepStyle` to #30778. That seems tidier but it sounds like nothing else is using this trait, e.g. #30501. It seems wrong for some ranges too, like `Base.RangeStepStyle('a':'z’) == Base.RangeStepIrregular()`.

### DIFF
--- a/stdlib/Dates/src/ranges.jl
+++ b/stdlib/Dates/src/ranges.jl
@@ -65,3 +65,11 @@ Base.iterate(r::StepRange{<:TimeType}, (l, i)) = l <= i ? nothing : (r.start + r
 # Combinations of types and periods for which the range step is regular
 Base.RangeStepStyle(::Type{<:OrdinalRange{<:TimeType, <:FixedPeriod}}) =
     Base.RangeStepRegular()
+
+# Avoid special method added in https://github.com/JuliaLang/julia/pull/30778
+function findfirst(testf::Union{Fix2{typeof(isequal)},Fix2{typeof(==)}}, A::StepRange{<:TimeType, <:OtherPeriod})
+    for (i, a) in pairs(A)
+        testf(a) && return i
+    end
+    return nothing
+end
--- a/stdlib/Dates/test/ranges.jl
+++ b/stdlib/Dates/test/ranges.jl
@@ -582,4 +582,12 @@ a = Dates.Time(23, 1, 1)
       hash([Date("2018-1-03"), Date("2018-1-04"), Date("2018-1-05")]) ==
       hash(Date("2018-1-03"):Day(1):Date("2018-1-05"))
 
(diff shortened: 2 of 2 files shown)


iamed2:
I would suggest using the trait. It does need an extra definition for `Char` but that's easily added.

nalimilan:
I'm not sure the trait would really be appropriate here. The method will fail for any case in which conversion of `first(r)-x` to the step type isn't possible. AFAIK nothing requires that in the `AbstractRange` API. I'm not even sure types are required to implement `%` and `div`, which the method uses.

It seeds that a safer approach would be to:
- restrict the optimized methods to `Integer` (as @JeffBezanson suggested initially)
- add a special method for date and time types which does the right thing depending on the case (e.g. when the step is `Day` the efficient approach can be used)

mcabbott:
Since these traits appear to be basically unused, except that they correctly label date ranges with non-constant steps (Month, Year), perhaps we are free to give them this meaning? By which I mean documenting that if you set `RangeStepStyle(r) === RangeStepRegular()` for some range, then you guarantee that `div` & `rem` will work. 

The default is (and should remain) `RangeStepIrregular()`, except when `step(r)::Integer`. So using this will at present, I think, restrict the fast method to exactly the cases you suggest.

nalimilan:
@mcabbott Would you feel like making a PR to restrict the buggy specialized methods to `Integer`?

mcabbott:
My preference is for #35279, which is approved but not merged. But if you'd like to make a simpler PR that would be fine by me, too.

nalimilan:
As I said I'm not convinced #35279 is the right approach and I'd rather remove `RangeStepStyle` (which I added for reasons which no longer exist now).

PR #35279, opened 2020-03-27
Use trait `RangeStepStyle` for `findfirst(isequal, ...)`, and adjust it for `StepRange{Char,Int}`

### LINKED ISSUE #35203: Errors from findfirst(..., ::StepRange{Date,Year})

mcabbott:
Fixes #35203, and closes #35278 which is a different solution. This `findfirst` method was added in #30778. 

Edit, 2021: another PR #35278 special-cased some dates to avoid the fast method. New PR #40552 instead restricts it to only numbers. This PR uses an existing trait, which I believe correctly marks the distinction between regular and irregular dates, and marks unknown cases `RangeStepIrregular`, including many floating point ranges.

Earlier:

I'm not completely sure about things like 
```
RangeStepStyle(::Type{<:StepRangeLen{<:Any,<:Any,<:Integer}}) = RangeStepRegular()
```
and how
```
Date(2019):Month(1):Date(2021) isa StepRange
```
fits with "The step between each element is constant" here. So someone who does understand should check that this isn't wrong.

### DIFF
--- a/base/array.jl
+++ b/base/array.jl
@@ -2517,7 +2517,7 @@ function findfirst(p::Union{Fix2{typeof(isequal),T},Fix2{typeof(==),T}}, r::Abst
     return i1 + oftype(i1, p.x - first(r))
 end
 
-findfirst(p::Union{Fix2{typeof(isequal),T},Fix2{typeof(==),T}}, r::StepRange{T,S}) where {T,S}
+function findfirst(p::Union{Fix2{typeof(isequal),T},Fix2{typeof(==),T}}, r::StepRange{T,S}) where {T,S}
     RangeStepStyle(r) isa RangeStepRegular || return _findfirst(p, r)
     isempty(r) && return nothing
     minimum(r) <= p.x <= maximum(r) || return nothing
--- a/base/array.jl
+++ b/base/array.jl
@@ -2492,7 +2492,9 @@ julia> findfirst(iseven, A)
 CartesianIndex(2, 1)
 ```
 """
-function findfirst(testf::Function, A)
+findfirst(testf::Function, A) = _findfirst(testf, A)
+
+function _findfirst(testf::Function, A)
     for (i, a) in pairs(A)
         testf(a) && return i
     end
@@ -2515,10 +2517,12 @@ function findfirst(p::Union{Fix2{typeof(isequal),T},Fix2{typeof(==),T}}, r::Abst
     return i1 + oftype(i1, p.x - first(r))
 end
 
-function findfirst(p::Union{Fix2{typeof(isequal),T},Fix2{typeof(==),T}}, r::StepRange{T,S}) where {T,S}
-    if RangeStepStyle(r) !== RangeStepRegular()
-        return invoke(findfirst, Tuple{Function,Any}, p, r)
-    end
+findfirst(p::Union{Fix2{typeof(isequal),T},Fix2{typeof(==),T}}, r::StepRange{T,S}) where {T,S} =
+    _findfirst(p, RangeStepStyle(r), r)
+
+_findfirst(p, ::RangeStepStyle, r) = _findfirst(p, r)
+
+function _findfirst(p::Union{Fix2{typeof(isequal)},Fix2{typeof(==)}}, ::RangeStepRegular, r::StepRange{T,S}) where {T,S}
     isempty(r) && return nothing
     minimum(r) <= p.x <= maximum(r) || return nothing
     d = p.x - first(r)
--- a/base/array.jl
+++ b/base/array.jl
@@ -2517,7 +2517,7 @@ function findfirst(p::Union{Fix2{typeof(isequal),T},Fix2{typeof(==),T}}, r::Abst
     return i1 + oftype(i1, p.x - first(r))
 end
 
-findfirst(p::Union{Fix2{typeof(isequal),T},Fix2{typeof(==),T}}, r::StepRange{T,S}) where {T,S} =
+findfirst(p::Union{Fix2{typeof(isequal),T},Fix2{typeof(==),T}}, r::StepRange{T,S}) where {T,S}
     RangeStepStyle(r) isa RangeStepRegular || return _findfirst(p, r)
     isempty(r) && return nothing
     minimum(r) <= p.x <= maximum(r) || return nothing
--- a/base/array.jl
+++ b/base/array.jl
@@ -2463,6 +2463,9 @@ Return `nothing` if there is no such element.
 Indices or keys are of the same type as those returned by [`keys(A)`](@ref)
 and [`pairs(A)`](@ref).
 
+A faster method is used for `findfirst(isequal(x), A)` when
+`RangeStepStyle(A) === RangeStepRegular()`.
+
 # Examples
 ```jldoctest
 julia> A = [1, 4, 2, 2]
@@ -2513,6 +2516,9 @@ function findfirst(p::Union{Fix2{typeof(isequal),T},Fix2{typeof(==),T}}, r::Abst
 end
 
 function findfirst(p::Union{Fix2{typeof(isequal),T},Fix2{typeof(==),T}}, r::StepRange{T,S}) where {T,S}
+    if RangeStepStyle(r) !== RangeStepRegular()
+        return invoke(findfirst, Tuple{Function,Any}, p, r)
+    end
     isempty(r) && return nothing
     minimum(r) <= p.x <= maximum(r) || return nothing
     d = p.x - first(r)
--- a/base/array.jl
+++ b/base/array.jl
@@ -2463,8 +2463,8 @@ Return `nothing` if there is no such element.
 Indices or keys are of the same type as those returned by [`keys(A)`](@ref)
 and [`pairs(A)`](@ref).
 
-A faster method is used for `findfirst(isequal(x), A)` when
-`RangeStepStyle(A) === RangeStepRegular()`.
+An optimized method is used for `findfirst(isequal(x), A)` when `A` is a range and
+`Base.RangeStepStyle(A) === Base.RangeStepRegular()`.
 
 # Examples
 ```jldoctest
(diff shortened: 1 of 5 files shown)


base/array.jl
```diff
@@ -1798,6 +1801,9 @@ findfirst(p::Union{Fix2{typeof(isequal),T},Fix2{typeof(==),T}}, r::AbstractUnitR
     first(r) <= p.x <= last(r) ? 1+Int(p.x - first(r)) : nothing
 
 function findfirst(p::Union{Fix2{typeof(isequal),T},Fix2{typeof(==),T}}, r::StepRange{T,S}) where {T,S}
+    if RangeStepStyle(r) !== RangeStepRegular()
```
timholy: Would it be worth making this a trait-depot instead?
```julia
findfirst(p::Union{Fix2{typeof(isequal),T},Fix2{typeof(==),T}}, r::StepRange{T,S}) where {T,S} = _findfirst(p, RangeStepStyle(r), r)
_findfirst(p, ::RangeStepStyle, r) = invoke(findfirst, Tuple{Function,Any}, p, r)  # fallback
function _findfirst(p, ::RangeStepRegular, r)
    ...
end
```
That would make it possible to specialize at intermediate states.

test/ranges.jl
```diff
@@ -1587,3 +1587,12 @@ end
     @test step(r) === 2
     @test collect(r) == ['a','c','e','g']
 end
+
+@testset "Char ranges are regular" begin
+    r = 'a':'z'
+    r2 = 'α':2:'ω'
+    r3 = StepRangeLen('a',3,10)
+    @test Base.RangeStepStyle(r) === Base.RangeStepRegular()
```
timholy: Is there a test that failed before you set the step for char ranges to be regular? If so it would be good to test that as well. (It's generally better to test *outcomes* than *representations*, though I agree there's room for testing both. But representations tend to be more fragile with respect to internal changes within Julia, and what we ultimately care about is the outcome.)

timholy (approved):
Interesting issues here. "constant" is clearly in the eye of the beholder, and it depends on how things are being measured (a month is a constant step unless you are measuring time in days).

This seems correct, so despite a couple of comments below I'll give this my approval for merging. Parenthetically, for efficiency we should probably call `divrem` in that method rather than `%` followed by `÷`.

test/ranges.jl
mcabbott: On master char ranges get the fast `findfirst` method, but they won't once this checks the trait. But there's no explicit error or wrong result, it's just slower. Agree this isn't much of a test though!

base/array.jl
mcabbott: That looks neater. Should I perhaps also change the simplest findfirst method to be `findfirst_iter` or something, so that it can be called without `invoke`?

mcabbott:
Bump?

base/array.jl
mcabbott: See if you like what I did here. I made:
```julia
_findfirst(p, ::RangeStepStyle, r) = _findfirst(p, r)
```
i.e. putting the trait before the variable it refers to, and breaking up the definition of the simplest `findfirst(p, r)` so that I can call its implementation without invoke.

nalimilan (commented):
Copying here what I said at https://github.com/JuliaLang/julia/pull/35278#issuecomment-605456023:

I'm not sure the trait would really be appropriate here. The method will fail for any case in which conversion of `first(r)-x` to the step type isn't possible. AFAIK nothing requires that in the `AbstractRange` API. I'm not even sure types are required to implement `%` and `div`, which the method uses.

It seeds that a safer approach would be to:
- restrict the optimized methods to `Integer` (as @JeffBezanson suggested initially)
- add a special method for date and time types which does the right thing depending on the case (e.g. when the step is `Day` the efficient approach can be used)

mcabbott:
> will fail for any case in which conversion of first(r)-x to the step type isn't possible.

Could this simply be a requirement for `RangeStepRegular`? Being able to unambiguously divide (say) time differences into the given `step`s seems to be precisely the concept that this trait is trying to capture.

nalimilan:
The reason for the introduction of the trait was to be able to hash ranges like arrays, so the only requirement was that the difference between two subsequent elements was always the same (according to `isequal` and `hash`). The optimized method in this PR requires more than that as it uses `%` and `div`.

Do we have examples of non-`Integer` types that would use it? Otherwise redefining the trait sounds overkill. And even if we want a trait, `RangeStepRegular` may not be the best name.

mcabbott:
Dates are the non-integer types in question. The trait seems to otherwise not be used for anything at all -- was this array hash code removed at some point? 

I agree that if it didn't yet exist, then adding a trait to Base for this would indeed seem like overkill. But it does exist, and marks the border between well- and badly-behaved date ranges. The meaning is still that "difference between two subsequent elements was always the same", even if what methods ought to be implemented should be better documented. 

That said, other solutions would be fine by me too.

nalimilan:
Yes, the hashing code was changed so the trait currently isn't used at all (it should probably have been removed at that time).

Sorry for missing that some time ranges are `RangeStepRegular` (that's not visible in the PR). I'm fine with reusing the trait for that then if that allows using the optimized method for these. Though please update the trait docstring to mention which functions have to be implemented, and add tests to check that `findfirst` works for these ranges (and not only for `RangeStepIrregular` ones). The new trait methods should also be tested. Thanks!

rafaqz:
This looks good to me, thanks for getting it cleaned up @mcabbott

mcabbott:
Bump?

base/array.jl
```diff
@@ -1923,6 +1923,9 @@ Return `nothing` if there is no such element.
 Indices or keys are of the same type as those returned by [`keys(A)`](@ref)
 and [`pairs(A)`](@ref).
 
+A faster method is used for `findfirst(isequal(x), A)` when
+`RangeStepStyle(A) === RangeStepRegular()`.
```
nalimilan, line 1927:
```suggestion
An optimized method is used for `findfirst(isequal(x), A)` when `A` is a range and
`Base.RangeStepStyle(A) === Base.RangeStepRegular()`.
```

base/array.jl
```diff
@@ -1966,7 +1971,12 @@ findfirst(p::Union{Fix2{typeof(isequal),Int},Fix2{typeof(==),Int}}, r::OneTo{Int
 findfirst(p::Union{Fix2{typeof(isequal),T},Fix2{typeof(==),T}}, r::AbstractUnitRange) where {T<:Integer} =
     first(r) <= p.x <= last(r) ? 1+Int(p.x - first(r)) : nothing
 
-function findfirst(p::Union{Fix2{typeof(isequal),T},Fix2{typeof(==),T}}, r::StepRange{T,S}) where {T,S}
+findfirst(p::Union{Fix2{typeof(isequal),T},Fix2{typeof(==),T}}, r::StepRange{T,S}) where {T,S} =
+    _findfirst(p, RangeStepStyle(r), r)
+
+_findfirst(p, ::RangeStepStyle, r) = _findfirst(p, r)
+
+function _findfirst(p::Union{Fix2{typeof(isequal)},Fix2{typeof(==)}}, ::RangeStepRegular, r::StepRange{T,S}) where {T,S}
```
nalimilan, line 1979: Wouldn't this be simpler?
```suggestion
    RangeStepStyle(r) isa RangeStepRegular || return _findfirst(p, r)
```

Also maybe use a more explicit name for `_findfirst`, like `_findfirst_general`?

nalimilan (commented):
Sorry for the delay! Looks good, just a few minor comments.

base/array.jl
mcabbott: Yes this is shorter. I guess I was trying to follow how other traits are treated, but now I have no idea what examples I followed. Maybe there is no need to leave this open to later definition of more methods, `isequal, ::StepRange` is already quite narrow. 

Will see if CI agrees...

nalimilan (approved):

DilumAluthge:
This PR was approved by @nalimilan in November 2021, but since then there has been no activity on this PR for four years+.

@nalimilan @mcabbott Is this PR still relevant? Should we resurrect it?

mcabbott:
I have not thought about it since. Happy to rebase if that helps.

DilumAluthge:
@nalimilan What do you think? Is this good to be merged if CI passes? Or does it need a new review?

base/array.jl
```diff
@@ -2463,6 +2463,9 @@ Return `nothing` if there is no such element.
 Indices or keys are of the same type as those returned by [`keys(A)`](@ref)
 and [`pairs(A)`](@ref).
 
+An optimized method is used for `findfirst(isequal(x), A)` when `A` is a range and
```
adienes, line 2466: I don't think a callout in the docs like this is necessary. what does "optimized" mean? if we make all the other range types faster, does that mean the docs are violated since `RangeStepStyle(A) === RangeStepRegular()` is no longer faster?

DilumAluthge:
@mcabbott Looks like some merge conflict markers got committed (and thus maybe the merge conflicts weren't fixed correctly?).

```
Error in testset ranges:
Error During Test at /usr/home/julia/.buildkite-agent/builds/freebsd13-amdci6-3/julialang/julia-master/julia-6dd4fceb41/share/julia/test/testdefs.jl:25
  Got exception outside of a @test
  LoadError: ParseError:
  # Error @ /usr/home/julia/.buildkite-agent/builds/freebsd13-amdci6-3/julialang/julia-master/julia-6dd4fceb41/share/julia/test/ranges.jl:2584:1
  <<<<<<< HEAD
  └┘ ── not a unary operator
  Stacktrace:
```

test/ranges.jl
```diff
@@ -2581,6 +2581,7 @@ let r = Ptr{Cvoid}(20):-UInt(2):Ptr{Cvoid}(10)
     @test last(r) === Ptr{Cvoid}(10)
 end
 
+<<<<<<< HEAD
```
mcabbott: sorry...
```suggestion
```

base/array.jl
```diff
@@ -2463,6 +2463,9 @@ Return `nothing` if there is no such element.
...
+`Base.RangeStepStyle(A) === Base.RangeStepRegular()`.
+
```
mcabbott, line 2468: We can remove this, as https://github.com/JuliaLang/julia/pull/35279/changes#r2789638333 suggests

base/array.jl
mcabbott: No strong feelings, I don't remember why this is there.

This comment selects the lines https://github.com/JuliaLang/julia/pull/35279#discussion_r2790538787

adienes:
I don't think that steps being regular means that `div(x - first(r), step(r))` can be used as an index. consider

```julia
julia> r = StepRangeLen(1e16, 1, 5)
1.0e16:1.0:1.0000000000000004e16

julia> diff(r)
4-element Vector{Float64}:
 0.0
 2.0
 2.0
 0.0

julia> step(r)
1.0
```

but this PR would make it a `RangeStepRegular`

PR #35281, opened 2020-03-27
Fix Broadcasting of Bidiagonal 

### LINKED ISSUE #707: Broadcasting of Bidiagonal incorrectly returns Bidiagonal
This one seems to be similar to https://github.com/JuliaLang/julia/issues/33397
```julia
julia> Bidiagonal(ones(4), ones(3), :U) .+ Bidiagonal(ones(4), ones(3), :L)
4×4 Bidiagonal{Float64,Array{Float64,1}}:
 2.0  1.0   ⋅    ⋅
  ⋅   2.0  1.0   ⋅
  ⋅    ⋅   2.0  1.0
  ⋅    ⋅    ⋅   2.0
```

### LINKED ISSUE #35245: Broadcasting of Bidiagonal incorrectly returns Bidiagonal
This one seems to be similar to https://github.com/JuliaLang/julia/issues/33397
```julia
julia> Bidiagonal(ones(4), ones(3), :U) .+ Bidiagonal(ones(4), ones(3), :L)
4×4 Bidiagonal{Float64,Array{Float64,1}}:
 2.0  1.0   ⋅    ⋅
  ⋅   2.0  1.0   ⋅
  ⋅    ⋅   2.0  1.0
  ⋅    ⋅    ⋅   2.0
```

ssikdar1:
# Description

Tries to Fix JuliaLang/LinearAlgebra.jl#707 
Adds a check for when theres two `Bidiagnoals` with differing `uplo` types. 
In this case a `Tridiagnoal` should be returned. In the case that `uplo` is the same, a `Bidiagnoal` is still returned. ( and all other cases stay the same )

# Testing

added unit tests and ran `structuredbroadcast.jl`

```
$ ./julia  stdlib/LinearAlgebra/test/structuredbroadcast.jl
Test Summary:                                                                              | Pass  Total
broadcast[!] over combinations of scalars, structured matrices, and dense vectors/matrices |  240    240
Test Summary:                                           | Pass  Total
broadcast! where the destination is a structured matrix |   26     26
Test Summary:                                   | Pass  Total
map[!] over combinations of structured matrices |  288    288
Test Summary: | Pass  Total
Issue JuliaLang/julia#33397  |   10     10
Test Summary:            | Pass  Total
Broadcast Returned Types |    4      4
```

Also testing in the shell:

```
  julia> using LinearAlgebra

julia> Bidiagonal(ones(4), ones(3), :U) .+ Bidiagonal(ones(4), ones(3), :L)
4×4 Tridiagonal{Float64,Array{Float64,1}}:
 2.0  1.0   ⋅    ⋅ 
 1.0  2.0  1.0   ⋅ 
  ⋅   1.0  2.0  1.0
  ⋅    ⋅   1.0  2.0

julia> Bidiagonal(ones(4), ones(3), :L) .+ Bidiagonal(ones(4), ones(3), :U)
4×4 Tridiagonal{Float64,Array{Float64,1}}:
 2.0  1.0   ⋅    ⋅ 
 1.0  2.0  1.0   ⋅ 
  ⋅   1.0  2.0  1.0
  ⋅    ⋅   1.0  2.0

  julia> Bidiagonal(ones(4), ones(3), :U) .+ Bidiagonal(ones(4), ones(3), :U)
4×4 Bidiagonal{Float64,Array{Float64,1}}:
 2.0  2.0   ⋅    ⋅ 
  ⋅   2.0  2.0   ⋅ 
  ⋅    ⋅   2.0  2.0
  ⋅    ⋅    ⋅   2.0


julia> Bidiagonal(ones(4), ones(3), :L) .+ Bidiagonal(ones(4), ones(3), :L)
4×4 Bidiagonal{Float64,Array{Float64,1}}:
 2.0   ⋅    ⋅    ⋅ 
 2.0  2.0   ⋅    ⋅ 
  ⋅   2.0  2.0   ⋅ 
  ⋅    ⋅   2.0  2.0
```


stdlib/LinearAlgebra/test/structuredbroadcast.jl
```diff
@@ -146,4 +146,13 @@ end
...
 end
+@testset "Broadcast Returned Types" begin
+    # Issue 35245
+    Bu = Bidiagonal(ones(4), ones(3), :U)
+    Bl = Bidiagonal(ones(4), ones(3), :L)
+    @test typeof(Bu .+ Bl) == Tridiagonal{Float64,Array{Float64,1}}
+    @test typeof(Bl .+ Bu) == Tridiagonal{Float64,Array{Float64,1}}
+    @test typeof(Bu .+ Bu) == Bidiagonal{Float64,Array{Float64,1}}
+    @test typeof(Bl .+ Bl) == Bidiagonal{Float64,Array{Float64,1}}
+end
```
dkarrasch: Could you please add some tests that also check the numerical result, and maybe another operation? I'd suggest to use randomized (off-)diagonal vectors, to verify that the (off-)diagonals end up in the right position.

dkarrasch (changes requested):
Very nice. Just a minor layout comment and a suggestion to strengthen the tests further.

stdlib/LinearAlgebra/src/structuredbroadcast.jl
```diff
@@ -65,6 +65,9 @@ find_bidiagonal(a::Bidiagonal, rest...) = a
 find_bidiagonal(bc::Broadcast.Broadcasted, rest...) = find_bidiagonal(find_bidiagonal(bc.args...), rest...)
 find_bidiagonal(x, rest...) = find_bidiagonal(rest...)
 function structured_broadcast_alloc(bc, ::Type{<:Bidiagonal}, ::Type{ElType}, n) where {ElType}
+    if  length(bc.args) == 2 && typeof(bc.args[begin]) <: Bidiagonal && typeof(bc.args[begin+1]) <: Bidiagonal && bc.args[begin].uplo != bc.args[begin+1].uplo
+            return Tridiagonal(Array{ElType}(undef, n-1), Array{ElType}(undef, n), Array{ElType}(undef, n-1))
```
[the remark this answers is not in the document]
ssikdar1: fixed spacing issues and updated layout in cd53d56

stdlib/LinearAlgebra/test/structuredbroadcast.jl
ssikdar1: Added more tests in cd76ffb for `.*` .  With d0f3ec4 all the Tridiagonals and Bidiagonals are created with rand.

dkarrasch: To play it safe, it would be good to have different super- and sub-diagonals... Otherwise, with `+` you can't distinguish which one is which.

ssikdar1: Made the super and sub diagonal arrays distinct in afc630b

dkarrasch (approved):
I'm not a broadcasting expert, but the desired tests pass.

mbauman:
Thanks so much for taking a stab at this!  I think we want a slightly more comprehensive solution, though — this will only work for exactly adjacent pairings of Bidiagonals, but fail if there are more arguments or nesting involved.  The key will be to do something like how `find_bidiagonal` works, except I think it should be changed from short-circuiting at the first bidiagonal to combining their `uplo`s.  We could return `'U'` or `'L'` (if homogenous) or `missing` if heterogeneous.

ssikdar1:
@mbauman Apologies if this is sloppy, but is 9510c3a what you were thinking?
I've created a `find_uplo` function that has logic similar to the `find_bidiagnoal`.

stdlib/LinearAlgebra/src/structuredbroadcast.jl
```diff
@@ -64,7 +64,24 @@ find_bidiagonal() = throw(ArgumentError("could not find Bidiagonal within broadc
 find_bidiagonal(a::Bidiagonal, rest...) = a
 find_bidiagonal(bc::Broadcast.Broadcasted, rest...) = find_bidiagonal(find_bidiagonal(bc.args...), rest...)
 find_bidiagonal(x, rest...) = find_bidiagonal(rest...)
+
+find_uplo(a::Char) = a
+find_uplo(a::Missing) = a
+function find_uplo(a::Bidiagonal, rest...)
+    if length(rest) > 0
+        return a.uplo == rest[1].uplo ? a.uplo : missing
```
mbauman, line 72: Here's a missing test case:

```julia
julia> Bidiagonal(1:5, 2:5, 'U') .* 2
ERROR: type Int64 has no field uplo
```

It's a little tricky because we actually need two missing value sentinels, I think.  One that represents the fact that we've not found _any_ `Bidiagonal` yet, and another that represents the fact that we've found incompatible uplos.  Or we'll need to make two passes through the Broadcasted expression; one that short-circuits to pull out the first `Bidiagonal`'s uplo and another that merges all uplos together, starting with that one.

Other nasty test-cases include nestings like `(1 .* 2) .* (Bidiagonal(1:3, 1:2, 'U') .* (3 .* 4)) .* (5 .* Bidiagonal(1:3, 1:2, 'L'))`

ssikdar1: @mbauman 8de30ec rewrote `find_uplo` it can now handle the edge cases:
```
julia> Bidiagonal(1:5, 2:5, 'U') .* 2
5×5 Bidiagonal{Int64,Array{Int64,1}}:
 2  4  ⋅  ⋅   ⋅
 ⋅  4  6  ⋅   ⋅
 ⋅  ⋅  6  8   ⋅
 ⋅  ⋅  ⋅  8  10
 ⋅  ⋅  ⋅  ⋅  10

julia> 2 .* Bidiagonal(1:5, 2:5, 'U')
5×5 Bidiagonal{Int64,Array{Int64,1}}:
 2  4  ⋅  ⋅   ⋅
 ⋅  4  6  ⋅   ⋅
 ⋅  ⋅  6  8   ⋅
 ⋅  ⋅  ⋅  8  10
 ⋅  ⋅  ⋅  ⋅  10

julia> (1 .* 2) .* (Bidiagonal(1:3, 1:2, 'U') .* (3 .* 4)) .* (5 .* Bidiagonal(1:3, 1:2, 'L'))
3×3 Tridiagonal{Int64,Array{Int64,1}}:
 120    0     ⋅
   0  480     0
   ⋅    0  1080
```

Also added test cases to the unit tests

stdlib/LinearAlgebra/src/structuredbroadcast.jl
```diff
@@ -64,7 +64,39 @@ find_bidiagonal() = throw(ArgumentError("could not find Bidiagonal within broadc
...
+find_uplo(a::Char) = a
+find_uplo(a::Number) = Nothing
+find_uplo(a) = Nothing
+
+function find_uplo(bc::Broadcast.Broadcasted, rest...)
+
+    left = find_uplo(bc.args[1])
+    right = find_uplo(bc.args[2])
+
+    tmp = (left, right)
```
mbauman, line 78: This could be zero, one, two, or more arguments here.  Here's what I think I'd do:

```Julia
merge_uplos(::Nothing, ::Nothing) = nothing
merge_uplos(::Nothing, b) = b
merge_uplos(a, ::Nothing) = a
merge_uplos(a, b) = a == b ? a : 'T'

find_uplo(a) = nothing
find_uplo(a::Bidiagonal) = a.uplo
find_uplo(bc::Broadcasted) = mapreduce(find_uplo, merge_uplos, bc.args, init=nothing)
```

stdlib/LinearAlgebra/src/structuredbroadcast.jl
```diff
@@ -64,7 +64,39 @@ find_bidiagonal() = throw(ArgumentError("could not find Bidiagonal within broadc
...
+    myuplo = Nothing
+    for i = 1:length(tmp)
+        if typeof(tmp[i]) == Char
```
mbauman, line 81: It'd be a bit more inference-friendly to use `if tmp[I] isa Char` in places like this... but often seeing something like this in an if statement like this is a signal that you want to be using dispatch instead!

stdlib/LinearAlgebra/src/structuredbroadcast.jl
```diff
@@ -64,7 +64,39 @@ find_bidiagonal() = throw(ArgumentError("could not find Bidiagonal within broadc
...
+            if myuplo == Nothing
+                myuplo = tmp[i]
+            else
+                myuplo = tmp[i] == myuplo ? myuplo : 'T'
```
mbauman, line 85: Ah, very clever — I like this solution.

stdlib/LinearAlgebra/src/structuredbroadcast.jl
ssikdar1: This is awesome!! So clean. Changed to this in aec7bee

stdlib/LinearAlgebra/src/structuredbroadcast.jl
```diff
@@ -64,7 +64,39 @@ find_bidiagonal() = throw(ArgumentError("could not find Bidiagonal within broadc
 find_bidiagonal(a::Bidiagonal, rest...) = a
 find_bidiagonal(bc::Broadcast.Broadcasted, rest...) = find_bidiagonal(find_bidiagonal(bc.args...), rest...)
 find_bidiagonal(x, rest...) = find_bidiagonal(rest...)
+
+find_uplo(a::Bidiagonal) = a.uplo
+find_uplo(a::Char) = a
```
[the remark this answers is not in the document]
ssikdar1: Resolved in aec7bee

stdlib/LinearAlgebra/src/structuredbroadcast.jl
```diff
@@ -64,7 +64,24 @@ find_bidiagonal() = throw(ArgumentError("could not find Bidiagonal within broadc
...
+merge_uplos(a, ::Nothing) = a
+merge_uplos(::Nothing, b) = b
+merge_uplos(a, b) = a == b ? a : 'T'
+
+find_uplo(a::Bidiagonal) = a.uplo
+find_uplo(a) = nothing
+find_uplo(bc::Broadcasted) = mapreduce(find_uplo, merge_uplos, bc.args, init=nothing)
+
 function structured_broadcast_alloc(bc, ::Type{<:Bidiagonal}, ::Type{ElType}, n) where {ElType}
+    uplo = Nothing
```
mbauman, line 78: Now this can be greatly simplified, too! In fact, I don't think we need `find_bidiagonal` anymore at all — I think its definitions could even be deleted.  We can unconditionally `find_uplo(bc)` and use that answer directly in all cases.

ssikdar1: removed in 0d984ce !

mbauman (approved):

mbauman:
Thanks @dkarrasch, I forgot about this one.  Let's backport it, too.

PR #33967, opened 2019-11-27
Update for Unicode

PallHaraldsson:

PR #35282, opened 2020-03-28
bump utf8proc version, support Unicode 13

### LINKED ISSUE #33965: Unicode reference version
This might be a silly question :) but ... I noticed that the 1.3 release notes mention Unicode version 12.1.0:

> Support for Unicode 12.1.0 

[here](https://raw.githubusercontent.com/JuliaLang/julia/master/HISTORY.md)

but the only Unicode reference I can find in this repo is version 9.0.0:

```make
$(SRCCACHE)/UnicodeData.txt:
	@mkdir -p "$(SRCCACHE)"
	$(JLDOWNLOAD) "$@" http://www.unicode.org/Public/9.0.0/ucd/UnicodeData.txt
```

[from here](https://github.com/JuliaLang/julia/blob/81fb410656acabf93c9152903723ca9bb32f88c5/doc/Makefile#L27)

Are there two numbering systems? I'm happy to be educated in these arcane Unicode matters, such as what determines which Unicode symbols are in and which aren't...

stevengj:
Updates to utf8proc 2.5.0, which supports Unicode 13.0.0.

(Also updates the Unicode data files used for generating the documentation, which is not strictly necessary but may reduce confusion.  Closes #33965.  Closes #33967.)


ararslan:
I think this will also need a bump to the version used with `USE_BINARYBUILDER=1`, unless we aren't currently serving utf8proc from BinaryBuilder.

stevengj:
@ararslan, how do I do that?

ararslan:
Tbh I'm not entirely sure either. Looks like you got the Yggdrasil piece though. @staticfloat, can you provide some sagely guidance here?

stevengj:
Okay, I've dug a bit further.  utf8proc is currently not using binarybuilder at all in Julia, so this PR should be good to merge as-is.

(Since Yggdrasil is building utf8proc now, we could add binarybuilder support, but that could be done in a separate PR.)

ararslan (approved):

ararslan:
A news entry saying that the version was updated may be worthwhile but also ¯\\\_(ツ)\_/¯

StefanKarpinski (approved):
Looks good to merge once CI passes.

PR #35283, opened 2020-03-28
Support `do` block with `@code_...` macros

tkf:
This PR makes invocations like `@code_typed map(1:1) do x; x; end` work.

### DIFF
--- a/stdlib/InteractiveUtils/src/macros.jl
+++ b/stdlib/InteractiveUtils/src/macros.jl
@@ -51,11 +51,6 @@ function gen_call_with_extracted_types(__module__, fcn, ex0, kws=Expr[])
                 return Expr(:call, fcn, f,
                             Expr(:call, typesof, map(esc, ex0.args)...))
             end
-        elseif ex0.head === :do && Meta.isexpr(get(ex0.args, 1, nothing), :call)
-            f = ex0.args[1].args[1]
-            args = [[ex0.args[2]]; ex0.args[1].args[2:end]]
-            return Expr(:call, fcn, esc(f), Expr(:call, typesof, map(esc, args)...),
-                        kws...)
         else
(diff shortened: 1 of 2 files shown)


Liozou:
Nice addition! But it seems to me you are not taking into account the possible presence of keyword arguments:
```julia
julia> open("VERSION"; read=true) do f; f end
IOStream(<file VERSION>)

julia> @code_typed open("VERSION"; read=true) do f; f end
ERROR: syntax: invalid syntax ; read = true
Stacktrace:
 [1] top-level scope at REPL[15]:1
```

Liozou:
I think one option is to lower the do-block by hand near the entry point of the function, so modifying the `gen_call_with_extracted_types` this way instead of yours:
```diff
diff --git a/stdlib/InteractiveUtils/src/macros.jl b/stdlib/InteractiveUtils/src/macros.jl
index e6dd8cb7d1..91b6521bfa 100644
--- a/stdlib/InteractiveUtils/src/macros.jl
+++ b/stdlib/InteractiveUtils/src/macros.jl
@@ -8,6 +8,14 @@ separate_kwargs(args...; kwargs...) = (args, kwargs.data)
 
 function gen_call_with_extracted_types(__module__, fcn, ex0, kws=Expr[])
     if isa(ex0, Expr)
+        if ex0.head === :do && Meta.isexpr(get(ex0.args, 1, nothing), :call)
+            if length(ex0.args) != 2
+                return Expr(:call, :error, "ill-formed do call")
+            end
+            i = findlast(a->(Meta.isexpr(a, :kw) || Meta.isexpr(a, :parameters)), ex0.args[1].args)
+            insert!(ex0.args[1].args, (isnothing(i) ? 2 : i+1), ex0.args[2])
+            ex0 = ex0.args[1]
+        end
         if any(a->(Meta.isexpr(a, :kw) || Meta.isexpr(a, :parameters)), ex0.args)
             return quote
                 local arg1 = $(esc(ex0.args[1]))
```
Another option would probably be to check for keywords at the point where you originally added the do-block handling.

tkf:
@Liozou Thanks a lot for looking into this! I switched to your patch.

Liozou:
Glad I could help! :)

PR #35289, opened 2020-03-28
Add support for `begin` in at-view/at-views

### LINKED ISSUE #35115: begin for @view
`begin` was introduced in  #33946, but it would be nice if `@view` had a rewriting like [`replace_ref_end!`](https://github.com/JuliaLang/julia/blob/5808aea34db4febc2c4178810dedab57f3b19776/base/views.jl#L4) for `begin`, too.

Currently
```julia
julia> VERSION
v"1.5.0-DEV.461"

julia> v = 1:4
1:4

julia> v[(begin + 1):(end - 1)]
2:3

julia> @view v[(begin + 1):(end - 1)]
ERROR: UndefVarError: begin not defined
Stacktrace:
 [1] top-level scope at REPL[4]:1
```

sostock:
Fixes #35115.

This adds support for expressions like `@view A[begin:end]` where previously only `end` was supported.


sostock:
I added a NEWS item and compat annotations.

mbauman (approved):

tpapp:
Friendly bump — I wonder if this could get merged at some point, it provides a very nice fix for the issue.

tpapp (approved):

PR #24912, opened 2017-12-04
WIP/RFC: random unleashed

rfourquet:
#23964 decoupled the generation of random values out of "distribution specifiers" (e.g. `1:10` or `Float64`) from the generation of arrays filled with such values. This unlocked the possibility to generate other kinds of containers with the same distribution specifiers. This PR is a proposition for evolving the `rand` API in this direction, with an implementation as a proof of concept:
```
rand([rng=GLOBAL_RNG], [S], [C...])
```
It's very similar to what exists now:
1) `S` is the "distribution specifier", which controls how scalar values are generated
2) `C...` is the container specifier

The standard container specifier has this form: `C... == ContainerName, specific_parameters...`, which allows the owner of a type to plug into this API without conflict. Base has the priviledge to reserve a couple terser APIs:
- `C... == dims::Dims` or `C... == dims::Integer...` for generating Arrays
- `C... == p::AbstractFloat, m, [n]` for generating a sparse array, like `sprand(m, n, p)`

Now, how to specify `S` when we want to generate a `Dict`? we need to combine a specification for the keys, with a specification for the values. So we introduce a `Distribution` type to handle that, for example `rand(Distribution(Pair, 1:10, Float64))` generates a `Pair{Int,Float64}`, and `rand(Distribution(Pair, 1:100, Float64), Dict, 10)` will generate a `Dict` of 10 such pairs.

What about normal distributions? it's not scalable to replicate those implementations for `randn` (and `randexp`), so we introduce `Normal{T} <: Distribution{T}`.
For example, we can do `rand(Normal(), Set, 10)` to generate a length-10 `Set` of values drawn from the normal distribution, or `rand(Normal(Complex{Float64}), 0.2, 100)` to generate a sparse vector with values drawn from the "circularly symmetric complex normal distribution".

 Here is the list of other examples implemented using this API:
- `rand(Distribution(Complex, 1:10))` to generate `Complex{Int}` values where each composant is generated out of `1:10` (it's also possible to specify a different specifier for each composant)
- `rand([chars], String, [n])` superseding `randstring([chars], [n])`
- `rand(BitArray, dims...)` superseding `bitrand(dims...)`

I would favor deprecating
- `bitrand`, unless the new version is too verbose?
- `randstring` (or is it `stringrand` and `randbit` ? :stuck_out_tongue: ), which is not seriously shorter than `rand(String)`
[...]

PR #35290, opened 2020-03-28
[new feature] Random.eachrand

o314:
Propose a new `Random.eachrand` function to bring `Iterators` to `Random` in the same vein of `eachrow`, `eachcol`.

It currently allow to stream random scalar type, range type.
Stream of random array is not proposed at this stage.

### DIFF
--- a/stdlib/Random/src/Random.jl
+++ b/stdlib/Random/src/Random.jl
@@ -265,8 +265,8 @@ struct IterableRNG{X,R}
 end
 
 Base.iterate(ibl::IterableRNG{X,R}, itr=ibl.rng) where {X,R} = (rand(itr,R), itr)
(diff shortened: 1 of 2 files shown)


rfourquet:
Note that this idea was slightly discussed in #24912 (named `Rand` there), which is now part of the [RandomExtension](https://github.com/rfourquet/RandomExtensions.jl) package. I would suggest that people interested in this feature try out this package, whose initial raison-d'être was to provide a proof-of-concept for the features proposed in that PR. BTW, it was wise to not merge it at that time, as the ideas could mature in the package and 2 years later I now still have a small breaking change to the `Rand` iterator in preparation.

ViralBShah:
Should we close this in favour of trying these things out in `RandomExtension`?

o314:
That's ok for me.

ViralBShah:
Thank you!

PR #35291, opened 2020-03-28
fix struct tail initialization for inline reference allocation

JeffBezanson:
For #34126.

The fix to `init_struct_tail` is definitely needed, but I'm not sure about the change to `emit_new_struct`. It seems to me that it's currently not possible to reach `undef_derived_strct` along this path, but I might be missing something.

### DIFF
--- a/src/datatype.c
+++ b/src/datatype.c
@@ -887,19 +887,10 @@ JL_DLLEXPORT jl_value_t *jl_new_struct(jl_datatype_t *type, ...)
 
 static void init_struct_tail(jl_datatype_t *type, jl_value_t *jv, size_t na)
 {
-    size_t nf = jl_datatype_nfields(type);
-    char *data = (char*)jl_data_ptr(jv);
-    for (size_t i = na; i < nf; i++) {
-        if (jl_field_isptr(type, i)) {
-            *(jl_value_t**)(data + jl_field_offset(type, i)) = NULL;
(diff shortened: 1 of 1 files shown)


src/cgutils.cpp
```diff
@@ -2704,7 +2704,7 @@ static jl_cgval_t emit_new_struct(jl_codectx_t &ctx, jl_value_t *ty, size_t narg
             if (lt->isVectorTy() || jl_is_vecelement_type(ty)) { // maybe also check the size ?
                 init_as_value = true;
             }
-            else if (tracked.count) {
+            else if (tracked.count == 0) {
```
vtjnash:
```suggestion
            else if (tracked.count != 0) {
```

We want to keep this as a FCA if there are tracked objects. We generally don't want to `init_as_value` (though `mark_julia_type` would sink it to the stack) if we don't have to.

JeffBezanson: I'm confused about what happens below this:
```
            else if (init_as_value) {
                 ...
            }
            else {
                strct = emit_static_alloca(ctx, lt);
                if (tracked.count)
                    undef_derived_strct(ctx.builder, strct, sty, tbaa_stack);
            }
```
If `tracked.count` was nonzero, we would have set `init_as_value`, so it looks like the `undef_derived_strct` is unreachable.

PR #11242, opened 2015-05-12
NTuples made me sad (so I nixed them)

timholy:
This rabbit hole turned out to be a giant dungeon, ~~instead, and I haven't yet found the exit. I think I'm very close, but I've spent far too much time on this and I'm beginning to run out of ideas and energy.~~, but yay, I found the exit.

This is the latest attempt at `NTuple{N,T} -> Tuple{Vararg{T,N}}`, aka #10911, #10691. Julia builds and passes ~~many~~ all tests, ~~but borks on some (e.g., `subarray`). The current problem seems to be that a `VarState(Tuple{_<:Vararg{T,N}},false)` gets injected somewhere (leading ultimately to an attempted `Tuple{Vararg, ASCIIString}` and `Vararg type in non-final position` error). It's probably a 3-line fix, but I have not yet succeeded in finding the correct 3 lines (inference.jl is difficult for not-Jeff to hack on).~~

~~This still contains~~The intermediate commits contain a lot of debugging code that ~~will eventually be~~ was deleted in the later commits. In the end, the way I got this ~~far~~to work was to write new versions of `jl_intersect_type`, `jl_subtype_le`, `type_morespecific`, and `type_match_`, which (1) called the old algorithm and stored the result, then (2) translated NTuples on-the-fly and then called the new algorithm with the translated arguments, and (3) printed some output if the two disagreed. ~~This branch is partway through the deletion of all that debugging code; before completing the migration I'd love to figure out where the problem is.~~

It turns out that we have enough differences in how `NTuple`s and `Tuple{Vararg}`s behave that some disagreements are unavoidable; one of the best things about this, presumably, is that such differences will go away. I even went to the trouble to catalog the differences; in the "best" column I use "N" to indicate that I think the new algorithm yields the most sensible result, and "O" to indicate the old algorithm. Only two of those four algorithms had differences, of which representative examples are:

`jl_intersect_type`:
none (as measured by a <: b && b <: a)

`jl_subtype_le`:

| a | b | ta | invariant | old result | new result | best |
| --- | --- | --- | --- | --- | --- | --- |
| `_<:Tuple` | `Tuple` | 0 | 0 | 0 | 1 | N |
| `_<:Tuple{_<:Vararg{T<:Any, N<:Any}}` | `Tuple{Vararg{#T<:Any, N<:Any}}` | 0 | 0 | 0 | 1 | N? |
[...]

PR #35298, opened 2020-03-29
Fix offset when widening collections

timholy:
The final argument of `copyto!` is a count; subtracting 1 from the current index works only if indexing starts at 1. This also explicitly uses `LinearIndices` just as occurs in [upstream callers](https://github.com/JuliaLang/julia/blob/b637cb786a644979f6ad2752e11206d08c8ce683/base/array.jl#L686).

### DIFF
--- a/base/array.jl
+++ b/base/array.jl
@@ -696,7 +696,8 @@ end
 function setindex_widen_up_to(dest::AbstractArray{T}, el, i) where T
     @_inline_meta
     new = similar(dest, promote_typejoin(T, typeof(el)))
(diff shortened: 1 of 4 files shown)


timholy:
It should be noted that the destination array typically supports linear indexing, and so none of our current tests check this code for `IndexCartesian` destination types. Consequently, I decided to move the `TSlow` type from `test/abstractarray.jl` to `test/testhelpers/` so that it can be reused. It also seemed to be an opportune moment to update it given the merging of #11242. A trip down memory lane with @mbauman :smile:.

JeffBezanson (approved):

Issue #35217, opened 2020-03-22
Intermittent failure in loading test


Keno:
As previously discussed in #35183, we have an intermittent failure in loading.
Over the past few days I've run this test continuously on around 20 cores and
observe about a 0.5% failure rate. I tried reproducing it with just the random seed,
but that was unsuccessful. Finally, I simply serialized out the inputs to the test_find
function, which yielded a successful reproducer of the issue. The reproducer is as follows:

```
julia> roots
Dict{String,PkgId} with 3 entries:
  "B" => B [top-level]
  "A" => A [75d55ab9-8b58-4b6e-8ad9-7fc54a3c81d3]
  "C" => C [fa81875a-839c-476c-9a9a-63af890e68cb]

julia> show(IOContext(stdout, :verbose=>true, :compact=>false), graph)
Dict{PkgId,Dict{String,PkgId}}(A [75d55ab9-8b58-4b6e-8ad9-7fc54a3c81d3] => Dict("A" => A [75d55ab9-8b58-4b6e-8ad9-7fc54a3c81d3],"C" => C [fa81875a-839c-476c-9a9a-63af890e68cb]),C [fa81875a-839c-476c-9a9a-63af890e68cb] => Dict("C" => C [fa81875a-839c-476c-9a9a-63af890e68cb]))

julia> paths
Dict{PkgId,Union{Nothing, String}} with 3 entries:
  B [top-level]                            => "/tmp/jl_FyNXyg/B/src/B.jl"
  A [75d55ab9-8b58-4b6e-8ad9-7fc54a3c81d3] => "/tmp/jl_FyNXyg/A/src/A.jl"
  C [fa81875a-839c-476c-9a9a-63af890e68cb] => "/tmp/jl_FyNXyg/C/src/C.jl"

julia> push!(empty!(Base.LOAD_PATH), "/tmp/jl_FyNXyg"))

julia> test_find(roots, graph, paths)
Test Failed at REPL[16]:10
  Expression: id == get(roots, name, nothing)
   Evaluated: A [f2499fa5-f4f5-5ad4-85f6-2621ce3bb437] == A [75d55ab9-8b58-4b6e-8ad9-7fc54a3c81d3]
ERROR: There was an error during testing

julia> Base.dummy_uuid("/tmp/jl_FyNXyg/A/Project.toml")
UUID("f2499fa5-f4f5-5ad4-85f6-2621ce3bb437")
```
Both A and C have a project file, both are empty, B does not have a project file.

[...]

PR #35303, opened 2020-03-30
In open(f, ...), call flush before calling `close`

Keno:
In #35217, I noticed that if we call `close` on an IOStream, but the device
that has the underlying file is full, we silently truncate the file without
error. To remidy that, introduce an explicit flush call before closing the
file. This should either succeed and reflect the changes on disk, or fail
and throw an appropriate error.

### DIFF
--- a/base/io.jl
+++ b/base/io.jl
@@ -296,6 +296,7 @@ function open(f::Function, args...; kwargs...)
     io = open(args...; kwargs...)
     try
         f(io)
+        flush(io)
     finally
         close(io)
     end
--- a/base/io.jl
+++ b/base/io.jl
@@ -295,8 +295,9 @@ julia> rm("myfile.txt")
 function open(f::Function, args...; kwargs...)
     io = open(args...; kwargs...)
     try
-        f(io)
+        v = f(io)
         flush(io)
+        return v
     finally
         close(io)
     end


StefanKarpinski (approved):
👍

KristofferC:
[just stating the obvious, all CI failed because of errors in package loading]

StefanKarpinski:
While this is a good idea and should work, I fear that there are many kinds of IO objects where this will not work. I've seen it break things before. In general, I fear we seem to be in a state where the whole flush/close dance is very, very brittle.

vtjnash (approved):
@StefanKarpinski This change should be fine. It's just implicit-return that got broken by mistake.

vtjnash (changes requested):
```julia
julia> open("VERSION") do f; close(f); flush(f); end

julia> open(`cat`, "w") do f; close(f); flush(f); end
ERROR: IOError: stream is closed or unusable
```

^ Do we have any tests like these?

StefanKarpinski:
If it works on all built-in IO types, I'll be happy (a little surprised, but happy), but I've definitely seen external IO types that would break with this change, e.g. TranscodingStreams.

JeffBezanson:
I'm also fine with this in theory but I'm worried about breakage. Why not add a call to `flush` in `close(::IOStream)` instead?

JeffBezanson:
Ah, in fact ios_close already calls ios_flush internally, it just doesn't report the error.

Keno:
> Ah, in fact ios_close already calls ios_flush internally, it just doesn't report the error.

Right, so the question is should close throw an error instead. And when we talked about it, the sense was that `close` should not error, which is why I'm proposing this instead.

The point is at some point, we should throw an error here, because otherwise subsequent code assumes that the file operation was complete, which may not be the case.

JeffBezanson:
It seems strange to me to only throw the error if you use the `open(::Function,...)` form. It also conflicts with having a more general `with`-style construct --- is the destructor/finalizer for a file `close`, or `flush` plus `close`? What do other systems do in this situation?

Keno:
`fclose` returns an error code, but not sure if anybody ever checks it.

JeffBezanson:
`close` methods in Java and Python throw errors in this kind of situation, so I think it would be simpler and more consistent to propagate the error from the existing `flush` call inside `close`.

StefanKarpinski:
Agree. If write stuff and close a file and it didn't get successfully written to the file, I want an error.

JeffBezanson:
If `flush` fails inside close, should we continue closing the file anyway, or leave the file open? My gut reaction is that it's more predictable to close it no matter what, but that does mean you could lose buffered data.

StefanKarpinski:
Agree: close is better. If there's an error, you can't really assume the resulting file is in any kind of recoverable state anyway.

PR #35304, opened 2020-03-30
Use parent for similar(::PermutedDimsArray)

### LINKED ISSUE #658: similar(PermutedDimsArray(::CuArray)) isa Array
`similar` gives an Array not a CuArray, when acting on a PermutedDimsArray:
```julia
julia> similar(PermutedDimsArray(cu(rand(2,2)), (2,1)), 3,3)
3×3 Array{Float32,2}:
...
julia> similar(cu(rand(2,2))', 3,3) # Adjoint is fine
3×3 CuArray{Float32,2,Nothing}:
```
I'm not certain but it looks like LinearAlgebra may un-wrap Adjoint before calling `similar`, in which case the difference isn't originating in any JuliaGPU package. Where should the correction live? 

**Build log, etc**
```
julia> Pkg.build()
  Building NNlib → `~/.julia/packages/NNlib/FAI3o/deps/build.log`
false

julia> versioninfo()
Julia Version 1.3.0
Commit 46ce4d7933 (2019-11-26 06:09 UTC)
Platform Info:
  CPU: Intel(R) Core(TM) i7-8700 CPU @ 3.20GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.1 (ORCJIT, skylake)
Environment:
  JULIA_NUM_THREADS = 12

(cu) pkg> st
    Status `~/trash/cu/Project.toml`
  [3895d2a7] CUDAapi v4.0.0 #master (https://github.com/JuliaGPU/CUDAapi.jl.git)
  [c5f51814] CUDAdrv v6.2.2 #master (https://github.com/JuliaGPU/CUDAdrv.jl.git)
  [be33ccc6] CUDAnative v3.0.2 #master (https://github.com/JuliaGPU/CUDAnative.jl.git)
  [3a865a2d] CuArrays v2.0.1 [`~/.julia/dev/CuArrays`]
  [0c68f7d7] GPUArrays v3.1.0 #master (https://github.com/JuliaGPU/GPUArrays.jl.git)
  [872c559c] NNlib v0.6.6

julia> CuArrays.version()
v"10.1.243"
```

mcabbott:
This ensures that `similar` matches the underlying storage of a `PermutedDimsArray` wrapper, as happens e.g. for `Transpose`, views:
```julia
julia> using SparseArrays

julia> similar(transpose(sparse(ones(3,3))))
3×3 SparseMatrixCSC{Float64,Int64} with 0 stored entries

julia> similar(view(sparse(ones(3,3)), 1,:), 4,4)
4×4 SparseMatrixCSC{Float64,Int64} with 0 stored entries

julia> similar(PermutedDimsArray(sparse(ones(3,3)), (2,1)))
3×3 Array{Float64,2}:
 2.19322e-314  2.19306e-314  2.19306e-314
 0.0           0.0           0.0
 2.19307e-314  2.19307e-314  2.19307e-314
```
I'm not sure how to add a test for this, and don't see any in `test/subarray.jl` for views either. Can I load SparseArrays in the tests for Base?

Fixes https://github.com/JuliaGPU/CuArrays.jl/issues/658

### DIFF
--- a/base/permuteddimsarray.jl
+++ b/base/permuteddimsarray.jl
@@ -46,8 +46,10 @@ function PermutedDimsArray(data::AbstractArray{T,N}, perm) where {T,N}
 end
 
 Base.parent(A::PermutedDimsArray) = A.parent
-Base.size(A::PermutedDimsArray{T,N,perm}) where {T,N,perm}    = genperm(size(parent(A)),    perm)
+Base.size(A::PermutedDimsArray{T,N,perm}) where {T,N,perm} = genperm(size(parent(A)), perm)
 Base.axes(A::PermutedDimsArray{T,N,perm}) where {T,N,perm} = genperm(axes(parent(A)), perm)
+
+Base.similar(A::PermutedDimsArray, T::Type, dims::Base.Dims) = similar(parent(A), T, dims)
 
 Base.unsafe_convert(::Type{Ptr{T}}, A::PermutedDimsArray{T}) where {T} = Base.unsafe_convert(Ptr{T}, parent(A))
 
--- a/test/arrayops.jl
+++ b/test/arrayops.jl
@@ -4,11 +4,9 @@
 isdefined(Main, :OffsetArrays) || @eval Main include("testhelpers/OffsetArrays.jl")
 using .Main.OffsetArrays
 
-using SparseArrays
+isdefined(Main, :TSlow) || @eval Main include("testhelpers/arrayindexingtypes.jl")
 
-if !isdefined(@__MODULE__, :T24Linear)
-     include("testhelpers/arrayindexingtypes.jl")
- end
+using SparseArrays
 
 using Random, LinearAlgebra
 using Dates
--- a/test/arrayops.jl
+++ b/test/arrayops.jl
@@ -3,7 +3,12 @@
 # Array test
 isdefined(Main, :OffsetArrays) || @eval Main include("testhelpers/OffsetArrays.jl")
 using .Main.OffsetArrays
+
 using SparseArrays
+
+if !isdefined(@__MODULE__, :T24Linear)
+     include("testhelpers/arrayindexingtypes.jl")
+ end
 
 using Random, LinearAlgebra
 using Dates
@@ -683,6 +688,13 @@ end
     y = [0, 0, 0, 0]
     copyto!(y, x)
     @test y == [1, 2, 3, 4]
+
+    # similar, https://github.com/JuliaLang/julia/pull/35304
+    x = PermutedDimsArray([1 2; 3 4], (2, 1))
+    @test similar(x, 3,3) isa Array
+    z = TSlow([1 2; 3 4])
+    x_slow = PermutedDimsArray(z, (2, 1))
+    @test similar(x_slow, 3,3) isa TSlow
 end
 
 @testset "circshift" begin
--- a/test/arrayops.jl
+++ b/test/arrayops.jl
@@ -4,7 +4,7 @@
 isdefined(Main, :OffsetArrays) || @eval Main include("testhelpers/OffsetArrays.jl")
 using .Main.OffsetArrays
 
-isdefined(Main, :TSlow) || @eval Main include("testhelpers/arrayindexingtypes.jl")
+isdefined(@__MODULE__, :T24Linear) || include("testhelpers/arrayindexingtypes.jl")
 
 using SparseArrays
 


base/permuteddimsarray.jl
```diff
@@ -46,9 +46,11 @@ function PermutedDimsArray(data::AbstractArray{T,N}, perm) where {T,N}
 end
 
 Base.parent(A::PermutedDimsArray) = A.parent
-Base.size(A::PermutedDimsArray{T,N,perm}) where {T,N,perm}    = genperm(size(parent(A)),    perm)
+Base.size(A::PermutedDimsArray{T,N,perm}) where {T,N,perm} = genperm(size(parent(A)), perm)
 Base.axes(A::PermutedDimsArray{T,N,perm}) where {T,N,perm} = genperm(axes(parent(A)), perm)
 
+Base.similar(A::PermutedDimsArray, T::Type, dims::Base.Dims) = similar(parent(A), T, dims)
```
timholy: Since `Transpose` and `Adjoint` are for linear algebra which we currently support only for arrays with indexing starting with 1, it's OK for those to only support `Dims`. But `PermutedDimsArray` supports more general array types, so we have to consider the case where `similar` will receive `axes` rather than `dims` input. From https://github.com/JuliaLang/julia/blob/12acd694e096b878ebca539b03c85828191556ad/base/abstractarray.jl#L624-L645 to avoid ambiguities it looks like we might need
```julia
similar(a::PermutedDimsArray, ::Type{T}, dims::Tuple{Union{Integer, OneTo}, Vararg{Union{Integer, OneTo}}}) where T
similar(a::PermutedDimsArray, ::Type{T}, dims::Dims{N}) where {T,N}
similar(a::PermutedDimsArray, ::Type{T}, axs::Tuple{Union{Integer, AbstractUnitRange}, Vararg{Union{Integer, AbstractUnitRange}}}) where T
```

timholy (commented):
This is the right thing to do. Re testing, in #35298 I moved some useful array types into `test/testhelpers`; if that PR gets merged you could conceivably use `TSlow` for testing, since it has a custom `similar` method. It has the advantage of being guaranteed not to have any specialized methods written for it other than what's in the test file; the same cannot be said of sparse array types which tend to accumulate an ever-growing number of specializations.

base/permuteddimsarray.jl
mcabbott: OK, thanks for looking up what's needed, I will add these methods.

mcabbott:
> could conceivably use `TSlow` for testing, since it has a custom `similar` method. 

OK, that looks perfect. I see this replaces a `T24Linear` type, so perhaps I wait for #35298 to minimise churn.

base/permuteddimsarray.jl
mcabbott: Here's what happens if I do add those methods:
```julia
using SparseArrays, OffsetArrays
sp = sparse(ones(3,3))

similar(transpose(sp), 3,3)
similar(transpose(sp), 0:2,0:2)

similar(PermutedDimsArray(sp, (2,1)), 3,3)     # isa Array
similar(PermutedDimsArray(sp, (2,1)), 0:2,0:2) # OffsetArray(Array)

# initial PR:
Base.similar(A::PermutedDimsArray, T::Type, dims::Base.Dims) = similar(parent(A), T, dims)

similar(PermutedDimsArray(sp, (2,1)), 3,3)     # isa SparseMatrixCSC
similar(PermutedDimsArray(sp, (2,1)), 0:2,0:2) # OffsetArray(SparseMatrixCSC)

# add extra methods:
Base.similar(A::PermutedDimsArray, ::Type{T}, dims::Tuple{Union{Integer, Base.OneTo}, Vararg{Union{Integer, Base.OneTo}}}) where T = similar(parent(A), T, dims)
Base.similar(A::PermutedDimsArray, ::Type{T}, axs::Tuple{Union{Integer, AbstractUnitRange}, Vararg{Union{Integer, AbstractUnitRange}}}) where T = similar(parent(A), T, axs)

similar(PermutedDimsArray(sp, (2,1)), 3,3)     # isa SparseMatrixCSC
similar(PermutedDimsArray(sp, (2,1)), 0:2,0:2) # is ambiguous. Candidates:
#   similar(A::PermutedDimsArray, ::Type{T}, axs::Tuple{Union{Integer, AbstractUnitRange},Vararg{Union{Integer, AbstractUnitRange},N} where N}) where T in Main at REPL[13]:1
#   similar(A::AbstractArray, ::Type{T}, inds::Tuple{Union{Colon, Integer, Base.IdentityUnitRange, Base.OneTo, UnitRange, OffsetArrays.IdOffsetRange},Vararg{Union{Colon, Integer, Base.IdentityUnitRange, Base.OneTo, UnitRange, OffsetArrays.IdOffsetRange},N} where N}) where T in OffsetArrays at /Users/me/.julia/packages/OffsetArrays/WmVaB/src/OffsetArrays.jl:102
# Possible fix, define
#   similar(::PermutedDimsArray, ::Type{T}, ::Tuple{Union{Integer, Base.IdentityUnitRange, Base.OneTo, UnitRange, OffsetArrays.IdOffsetRange},Vararg{Union{Integer, Base.IdentityUnitRange, Base.OneTo, UnitRange, OffsetArrays.IdOffsetRange},N} where N}) where T
```

timholy: We need to solve this at the level of OffsetArrays, presumably. https://github.com/JuliaArrays/OffsetArrays.jl/issues/87. I don't have a solution yet but I haven't had time to poke at it (and neither has anyone else, apparently).

mcabbott: That doesn't look easy... but I guess until a larger change lands, this defn is OK?

test/arrayops.jl
```diff
@@ -3,8 +3,13 @@
 # Array test
 isdefined(Main, :OffsetArrays) || @eval Main include("testhelpers/OffsetArrays.jl")
 using .Main.OffsetArrays
+
 using SparseArrays
 
+if !isdefined(@__MODULE__, :T24Linear)
+     include("testhelpers/arrayindexingtypes.jl")
+ end
```
mcabbott: I copied this from #35298, not sure if it's needed at all, nor whether it should be written like `isdefined || @eval` just above. (And at very least a space needs fixing, sorry.)

mcabbott: Changed in https://github.com/JuliaLang/julia/pull/35304/commits/a524b230febe94c99ab3dcbbd3917fc13a171576, I was hoping CI would tell me if that's OK, but it's failed on some unrelated unicode thing instead.

mcabbott: I think this is needed, and have restored it to something like what I had just above. 

CI is failing, I don't completely see why, but I don't think caused by this.

timholy:
Puzzling CI failure. Can you rebase on latest master?

mcabbott:
Done, and mostly green ticks now.

timholy:
Thanks for sticking with it, @mcabbott!

PR #35309, opened 2020-03-30
fix #35305, need escaping when printing string macro calls

### LINKED ISSUE #35305: String representation of expressions with quotes in non-standard string literals
The string representation of an expression that contains quotes `"` in triple-quoted non-standard string literals (or escaped quotes in quoted non-standard string literals) is not faithfully any more. The resulting text does not represent valid Julia code and cannot be `include`d again. in Julia v1.4

```
julia> print(:(r = r"""45 " """))
r = r"45 " "
```

```
julia> print(:(r = r"45 \" "))
r = r"45 " "
```

This does not happen for "normal" strings
```
julia> print(:(r = "45 \" "))
r = "45 \" "
```

and also used to work in Julia v1.3.1. See also https://discourse.julialang.org/t/string-representation-of-expression/36729/3?u=pfarndt

JeffBezanson:
fix #35305

### DIFF
--- a/base/strings/io.jl
+++ b/base/strings/io.jl
@@ -523,6 +523,55 @@ julia> println(raw"\\\\x \\\\\\"")
 """
 macro raw_str(s); s; end
 
+"""
+    escape_raw_string(s::AbstractString)
+    escape_raw_string(io, s::AbstractString)
+
+Escape a string in the manner used for parsing raw string literals.
+For each double-quote (`"`) character in input string `s`, this
+function counts the number _n_ of preceeding backslash (`\\`) characters,
+and then increases there the number of backslashes from _n_ to 2_n_+1
+(even for _n_ = 0). It also doubles a sequence of backslashes at the end
+of the string.
+
+This escaping convention is used in raw strings and other non-standard
+string literals. (It also happens to be the escaping convention
+expected by the Microsoft C/C++ compiler runtime when it parses a
+command-line string into the argv[] array.)
+
+See also: [`escape_string`](@ref)
+"""
+function escape_raw_string(io, str::AbstractString)
+    escapes = 0
+    for c in str
+        if c == '\\'
+            escapes += 1
+        else
+            if c == '"'
+                # if one or more backslashes are followed by
+                # a double quote then escape all backslashes
+                # and the double quote
+                escapes = escapes * 2 + 1
+            end
+            while escapes > 0
+                write(io, '\\')
+                escapes -= 1
+            end
+            escapes = 0
+            write(io, c)
+        end
+    end
+    # also escape any trailing backslashes,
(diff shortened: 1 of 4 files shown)


vtjnash (approved):
I don't think this actually fixes it, because of https://github.com/JuliaLang/julia/issues/29580, but it is better.

JeffBezanson:
It looks like that is about printing `Regex` and `SubstitutionString`; this is for expressions:
```
julia> :(r"\\\\")
:(r"\\\\")
```
But I should change this to only escape backslashes and quotes I think.

vtjnash:
True, but those happen to be `raw""` strings also, like these.

Also, some part of that fix is included in https://github.com/JuliaLang/julia/pull/34111/files#diff-85598a9f3fb3fcced8c43d0472c30888R545

JeffBezanson:
Oh, nice, we should add that code and use it here.

JeffBezanson:
Added the commit from #34111. I edited it a bit to not print surrounding `"`s, since `escape_string` doesn't.

mgkuhn:
@JeffBezanson The reason I printed the surrounding `"` in [my original version](https://github.com/JuliaLang/julia/pull/34111/commits/cba9a8a561a004c098de7ec3a9fd2717296016ab) of  `escape_raw_string` was that I really couldn't think of any reason not to: the entire function really only makes sense if its output is immediately followed by a `"`, because of the way this encoding specially treats trailing backslashes at the end of the string (expecting the recipient to see that followed by a `"`). This is different from how `escape_string()` works, where not only the delimiter is configurable, but also nothing special happens to backslashes at the end of the string. I'd therefore stay with my [original version](https://github.com/JuliaLang/julia/pull/34111/commits/cba9a8a561a004c098de7ec3a9fd2717296016ab), as the analogy to `escape_string()` is not so clear cut.
(The difference in mathematical terms: `escape_string()` is an isomorphism over string concatenation, whereas `escape_string_raw()` isn't, because of the case of strings ending in a backslash.)

JeffBezanson:
That makes sense, but it's easy to add the quotes if you want them, and impossible to remove them if you don't (however unlikely that might be).

base/strings/io.jl
```diff
@@ -523,6 +523,55 @@ julia> println(raw"\\\\x \\\\\\"")
...
+                # if one or more backslashes are followed by
+                # a double quote then escape all backslashes
+                # and the double quote
+                escapes = escapes * 2 + 1
+            end
+            while escapes > 0
+                write(io, '\\')
+                escapes -= 1
+            end
+            escapes = 0
```
musm, line 560: Looks like this line is unecessary since  escape is decremented to 0 by L559

mgkuhn: Agreed.

PR #35311, opened 2020-03-30
Add parentheses when printing UnionAll constructor methods

### LINKED ISSUE #34824: invalid error message for type parameter
```julia
julia> VERSION
v"1.5.0-DEV.299"

julia> struct Foo{T<:Real,TX<:AbstractVector{T},TFX}
       Δ::T
       x::TX
       fx::TFX
       residual_norm::T
       converged::Bool
       iterations::Int
       end

julia> function Foo(Δ::T1, x::AbstractVector{T2}, fx, residual_norm::T3, converged,
       iterations) where {T1 <: Real, T2 <: Real, T3 <: Real}
       T = promote_type(T1, T2, T3)
       Tx = T.(x)
       Foo{T,typeof(Tx)}(T(Δ), Tx, fx, T(residual_norm), converged, iterations)
       end
Foo

julia> Foo(1, [1.0, 2.0], nothing, 1.0, true, 0)
ERROR: MethodError: no method matching Foo{Float64,Array{Float64,1},TFX} where TFX(::Float64, ::Array{Float64,1}, ::Nothing, ::Float64, ::Bool, ::Int64)
Stacktrace:
 [1] Foo(::Int64, ::Array{Float64,1}, ::Nothing, ::Float64, ::Bool, ::Int64) at ./REPL[3]:5
```

Note the `where TFX(::Float64, ::Array{Float64,1}, ::Nothing, ::Float64, ::Bool, ::Int64)`.

([discussion on Discourse](https://discourse.julialang.org/t/repeated-type-parameter-in-method-error-of-promoting-constructor/34892))

sostock:
Fixes #34824.

This adds parentheses around the type when printing a `MethodError` for a constructor of a `UnionAll` type if it is printed with a `where` clause.

Example:
```julia
struct Foo{T} end
const Bar = Foo{T} where T<:Integer
(Bar)(::Float64) = 1
```

Before:
```julia
julia> Bar(1)
ERROR: MethodError: no method matching Foo{T} where T<:Integer(::Int64)
Closest candidates are:
  Foo{T} where T<:Integer(::Float64) at REPL[3]:1
```

After:
```julia
julia> Bar(1)
ERROR: MethodError: no method matching (Foo{T} where T<:Integer)(::Int64)
Closest candidates are:
  (Foo{T} where T<:Integer)(::Float64) at REPL[3]:1
```

To achieve this, I added a `IOContext` property `:unionall_parens`. I thought this was the easiest way to do this, in particular this makes it easy to add this to other printing methods if necessary.


vtjnash:
Thanks for looking into this!

As the comment there says, this should duplicate the `methodshow.jl` logic (possibly factored out from there), since there's other cases it gets wrong also, and they should do the same thing. Here's a few other examples:
```julia
julia> (::Union{Nothing,Int})(::Nothing) = nothing

julia> nothing("")
ERROR: MethodError: no method matching (::Nothing)(::String)
Closest candidates are:
  Any(::Nothing) at REPL[13]:1
Stacktrace:
 [1] top-level scope at REPL[14]:1
```
```julia
julia> Type(::Nothing) = nothing

julia> Type(1)
ERROR: MethodError: no method matching Type(::Int64)
Closest candidates are:
  Type(::Nothing) at REPL[2]:1
Stacktrace:
 [1] top-level scope at REPL[3]:1

julia> methods(Type)
# 1 method for type constructor:
[1] (::Type{Type})(::Nothing) in Main at REPL[2]:1
```

sostock:
The `methodshow.jl` logic is now used for printing the “closest candidates” list (I factored the relevant logic out into a function `show_method_name`). The other three occurrences of method names when printing `MethodError`s (`no method matching $f`, `$f is ambiguous`, and `Possible fix, define $f`) still use `:unionall_parens`. Should they be changed as well?

The examples from above now print like this:
```julia
julia> struct Foo{T} end; const Bar = Foo{T} where T<:Integer; (Bar)(::Float64) = 1;

julia> Bar(1)
ERROR: MethodError: no method matching (Foo{T} where T<:Integer)(::Int64)
Closest candidates are:
  (::Type{Foo{T} where T<:Integer})(::Float64) at REPL[1]:1
```
```julia
julia> (::Union{Nothing,Int})(::Nothing) = nothing

julia> nothing("")
ERROR: MethodError: no method matching (::Nothing)(::String)
Closest candidates are:
  (::Union{Nothing, Int64})(::Nothing) at REPL[3]:1
```
```julia
julia> Type(::Nothing) = nothing;

julia> Type(1)
ERROR: MethodError: no method matching Type(::Int64)
Closest candidates are:
  (::Type{Type})(::Nothing) at REPL[5]:1
```

JeffBezanson:
I noticed one of these bugs today so I started working on this. This PR was a useful reference but there are more cases; it just kept getting more detailed and crazy the more I looked.

For example there's this:
```
julia> (Foo{T} where T<:Integer)(x) = x
ERROR: syntax: invalid function name "Foo{T} where T <: Integer" around REPL[10]:1
```
but we might want to allow that since other type expressions, e.g. `Foo{<:Integer}` are allowed in that position. Then we could print signatures like that instead of adding `(::Type{...})`.

I'll put up an alternate PR soon.

sostock:
Closed in favor of #35862

Issue #27639, opened 2018-06-18
Changing `argmax` and `argmin` behavior to match the mathematical definition


Nosferican:
The definition of `argmax` from [Wikipedia](https://en.wikipedia.org/wiki/Arg_max),
Given a collection of elements (domain), which map to an ordered set (range), what is the subset containing those elements in the domain whose image is the frontier.

Currently, `argmax` is actually `indmax` which is defined for indexed collections and returns the index of the first encountered element whose image is equal to the frontier. The generalization of `keymax` (discussed, but not implemented) extends the supported collections from indexed to `pairs`-implemented collections. This is the result of renaming `indmax` to `argmax`. A current proposal is for `argmax` to be defined for `Generator` (#27613),
```
argmax(f(x) for x ∈ domain)
```
with that change, we could fix the implementation with something like
```
import Base: Generator, argmax
function argmax(f::Function, domain::Generator)
    T = eltype(domain)
    tested = Set{T}()
    output = Set{T}()
    frontier = nothing
    for elem ∈ domain
        if !(elem ∈ tested)
            push!(tested, elem)
            value = f(elem)
            if (frontier === nothing) || (value > frontier)
                frontier = value
                output = Set(elem)
            elseif value == frontier
                push!(output, elem)
            end
        end
    end
    return output
end
```
and call it with the following syntax
```
argmax(abs2, elem for elem ∈ -3:3) == Set([-3, 3])
```

JeffBezanson:
That definition and implementation don't depend on `Generator`; any iterable would work.

I agree this is useful but I don't think `argmax` should return a Set by default; it's quite inefficient in the common case where the answer is e.g. a single integer. This could be a new function like `argmaxall` or `argmaxes`.

Nosferican:
The answer would have form of some collection of `eltype(domain)`. Maybe if the length of the container is one, it could return the value. I used `Set` in the implementation to guarantee uniqueness (no need to compute the value for an element which we already considered, but overkill since I am checking for it in the loop). Other ways could call `unique` on the domain as first step (but why do an extra pass?). There shouldn't be any efficiency loses from returning the full set rather than the first since the implementation only passes through each once (no ability to break early since it doesn't do two passes). `output` could be set to `Vector{eltype(domain)}` if that is more convenient (bit extra information of order which shouldn't  matter, but why not). Since `output` is guarantee to be non-empty, people could call `first` on it if it is a `<:AbstractVector`. Any iterable should work.

[...]

Issue #27612, opened 2018-06-16
argmax should support generators


drewrobson:
I noticed that `maximum(i for i = 1:10)` works but `argmax(i for i=1:10)` does not. This seems inconsistent.

Edit: I am proposing that `argmax(f(x) for x ∈ itr)` should work and return an `x` in `itr` that maximizes `f(x)`, which is quite literally the mathematical definition of argmax. I have proposed a simple way to implement this [here](https://github.com/JuliaLang/julia/issues/27612#issuecomment-397884789).

(originally proposed [here](https://discourse.julialang.org/t/since-maximum-itr-supports-generators-argmax-itr-should-too/11719))

stevengj:
What does argmax even mean for an unordered iterable collection like Set?

drewrobson:
That's a good point. When this function was called `indmax`, it was perhaps clearer that it should be used on collections with well defined linear indexing.

Actually, I just tried `argmax(Dict("a" => 1, "b" => 2, "c" => 3))` and it returns "c" on 0.7-alpha. Cool! On 0.6, indmax returned 1, which makes very little sense and is I think related to your point.

drewrobson:
Looks like `argmax` is now designed to work with any collection where `keys` is defined.
So `argmax` is not defined for an unordered iterable collection like Set. I suppose `argmax` could still be defined for a generator `g` if one were to take `keys(g)` to be `1:length(g)`, if that makes sense? I think this is consistent with the original spirit of `indmax`, where `keys(v)` for some vector `v` is `1:length(v)`.

drewrobson:
The following 1-liner is sufficient to implement this proposal:
`Base.keys(g::Base.Generator) = 1:length(g)`
After running that in the REPL, `argmax(i for i=1:10)` works and returns 10 as expected. The remaining question is whether this is desirable. I await further instructions!

drewrobson:
Maybe that's a bad idea, since `g[10]` doesn't actually work so `keys(g) = 1:length(g)` would be false advertising.

Alternatively, one could have a specialized definition for `argmax` for generators. The implementation would be nearly identical, except replace `keys(g)` with `1:length(g)`. This way, there's no need to define `keys(g)` in general if that is a concern.

drewrobson:
[...]

PR #35316, opened 2020-03-30
Add 2-arg versions of findmax/min, argmax/min

### LINKED ISSUE #27613: Add a 2-arg flavor, argmax(f, domain)
With the rename to argmax, and dims now a kwarg, it seems very tempting to me to define a 2-arg flavor, `argmax(f, itr)`, where `itr` is an iterable subset of the domain of `f`. This would return an `x` in `itr` that maximizes `f(x)`, which is quite literally the mathematical definition of argmax.

(originally proposed [here](https://discourse.julialang.org/t/add-a-2-arg-flavor-argmax-f-domain/11720))

Edit: Previous wording may have been confusing, so I have attempted to improve it.

cmcaine:
Fixes #27613. Related: #27639, #27612, #34674.

Thanks to @tkf, @StefanKarpinski and @drewrobson for their assistance with
this PR.

I (and Stefan) wanted this to use partial order comparisons, but that would not be backwards compatible without a bit more faffing around. If there's interest, I might introduce another PR to explore that. That would introduce two partial orderings that bend the rules for these special cases: infinitesimals, unorderables (`NaN`) and `missing`.

I think reviewers should pay particular attention to the `isgreater` function I introduce. There are a lot of corner cases when comparing or ordering values and I'm not sure I've thought of them all. I'm fairly sure that it won't introduce any regressions in `findmin` or `argmin` (and it actually makes them work with `missing` where before they did not), but it may not be quite reliable enough to export.

Despite adding the `isgreater` function, I have not updated `min`, `minimum`, or the one-arg versions of `findmin` or `argmin` to use it. I can do that, but I wanted to get some review of what I've done first and possibly get this smaller change merged.


cmcaine:
I rebased this a few times and github is showing the commits in the wrong order for me on this page. On the repo's page and in the git log they are in the right order. The right order is this:

```
commit 423dc4b3f5ddcab71f54ab8e2d83239eb007b9e2 (HEAD -> argmax-2-arg-harder, cmcaine/argmax-2-arg-harder)
Author: Colin Caine <cmcaine@gmail.com>
Date:   Mon Mar 30 18:14:26 2020 +0100

    Add 2-arg versions of findmax/min, argmax/min
    
    Fixes #27613. Related: #27639, #27612, #34674.
    
    Thanks to @tkf, @StefanKarpinski and @drewrobson for their assistance
    with this PR.

commit 7bbb84f2fa7a53545dc2cfdfd3df8feba2aac519
Author: Colin Caine <cmcaine@gmail.com>
Date:   Mon Mar 30 18:10:03 2020 +0100

    Add isgreater
    
    Defines a descending total order where unorderable values and missing are last.
    This makes defining min, argmin, etc, simpler.

commit f6a2e73d77f3bd7669126041104763ff7e12d651
Author: Colin Caine <cmcaine@gmail.com>
Date:   Mon Mar 30 21:45:10 2020 +0100

    Rename unexported isgreater in reduceddim

commit b6c06dddb199a673a9e62ae14e9eeee6939560a0 (origin/master, origin/HEAD)
Author: Steven G. Johnson <stevenj@alum.mit.edu>
Date:   Mon Mar 30 12:33:28 2020 -0400

    bump utf8proc version, support Unicode 13 (#35282)
```

base/operators.jl
```diff
@@ -157,6 +157,28 @@ isless(x::AbstractFloat, y::AbstractFloat) = (!isnan(x) & (isnan(y) | signless(x
...
+
+Values that are normally unordered, such as `NaN`,
+are ordered after regular values.
+[`missing`](@ref) values are ordered last.
+
+# Implementation
+Types should usually not implement this function. Instead, implement `isless`.
+"""
+isgreater(a, b) = _is_unorderable(a) || _is_unorderable(b) ? isless(a, b) : isless(b, a)
+_is_unorderable(x) = !isa(x == x, Bool) || x != x
```
tkf: Maybe it should be called `_irreflexive` or `_anti_reflexive`?

Also, `==` and `!=` can be expensive for containers like vectors and strings. So how about

```julia
neq = x != x
return !isa(neq, Bool) || neq
```

?

Also, is it possible to prefer `missing` over `NaN` by factoring out irreflexivity check like this?  The reason why I did it with a multi-line definition in https://github.com/JuliaLang/julia/issues/27613#issuecomment-601502228 was that I thought it was impossible. (Though you can in principle define `isgreater` on `Missing` here.)

NEWS.md
```diff
@@ -86,6 +86,8 @@ New library functions
 * New function `bitreverse` for reversing the order of bits in a fixed-width integer ([#34791]).
 * New function `bitrotate(x, k)` for rotating the bits in a fixed-width integer ([#33937]).
 * One argument methods `startswith(x)` and `endswith(x)` have been added, returning partially-applied versions of the functions, similar to existing methods like `isequal(x)` ([#33193]).
+* New function `isgreater(a, b)` defines a descending total order where unorderable values and missing are ordered smaller than any regular value.
```
tkf: This sentence sounds contradictory to me. It says `isgreater` orders unorderable values. How about phrasing it like this?:

> New function `isgreater(a, b)` defines a descending total order that is compatible with the pass-through behavior of `min`; i.e., non-`NaN` numbers are greater than `NaN` and non-`missing` values are greater than `missing`.

I think it's better to rephrase the docstring with a similar wording.

If you want to use the word "unorderable" instead, maybe "(`<`, `=`)-unorderable" is better?

cmcaine: `isless` uses the phrase "values that are normally unordered, such as `NaN`", which I had intended to imitate, but forgot :)

base/operators.jl
cmcaine: I normally just trust constant propagation to sort that kind of thing out, but I could manually do it if it's sure to make a difference.

tkf: For things like arrays, I don't think the compiler can do anything clever (imagine a worst-case scenario like `getindex` with `println` in it).

cmcaine: Adjusted roughly as suggested. I'm not persuaded by the println usecase, but I guess getindex could involve moving data from GPU to CPU or over the network or whatever.

NEWS.md
cmcaine: Updated NEWS and the docstring.

base/operators.jl
tkf: It's very easy to check that running `!=` and `==` is twice as slow as running `==` once, even for plain array:

```julia
julia> f(x) = !isa(x == x, Bool) || x != x

julia> function g(x)
           neq = x != x
           return !isa(neq, Bool) || neq
       end

julia> x = rand(1000);

julia> @btime f($x)
  2.053 μs (0 allocations: 0 bytes)
false

julia> @btime g($x)
  1.028 μs (0 allocations: 0 bytes)
false
```

cmcaine: I already made the change on the basis that it can't hurt and I hadn't benchmarked it yet :)

base/reduce.jl
```diff
@@ -659,6 +659,217 @@ julia> minimum([1,2,3])
...
+    y = iterate(p)
+    if y === nothing
+        throw(ArgumentError("collection must be non-empty"))
+    end
+    (mi, m), s = y
+    i = mi
+    while true
+        y = iterate(p, s)
+        y === nothing && break
+        m != m && break
```
tkf, line 763: This implementation of `_findmax(a, ::Colon)` does not handle `missing`. I understand this is not part of the original plan for this PR. But it may be nice to just replace the function body with `mapfoldl((k, v) -> (v, k), _rf_findmax, pairs(a))`, now that we have all the machinery.

base/reduce.jl
```diff
@@ -659,6 +659,217 @@ julia> minimum([1,2,3])
...
+
+```jldoctest
+julia> findmax(identity, 5:9)
+(9, 9)
+
+julia> findmax(-, 1:10)
+(-1, 1)
+
+julia> findmax(cos, 0:π/2:2π)
+(1.0, 0.0)
```
tkf, line 691: I think it'd be nice to do demonstrate "first match wins" behavior. Maybe something like:

```julia
julia> findmax(first, [(1, :a), (2, :b), (2, :c)])
(2, (2, :b))
```

Maybe this was the intention of `cos` example but it is not very obvious to realize this when floating point arithmetic is involved.

base/reduce.jl
cmcaine: I had planned to leave that and the more involved definition in reducedim.jl alone for this PR in the interest of getting it through quickly and with the least chance of introducing breaking changes.

I can just carefully read the code and reimplement both, though :)

tkf: Rewriting `_findmax/min(a, ::Colon)` using `_rf_findmax/min` also means that we can test `_rf_findmax/min` much more heavily via existing tests on/using `findmin/max`.

Though I guess doing this later is fine too.

base/operators.jl
```diff
@@ -157,6 +157,27 @@ isless(x::AbstractFloat, y::AbstractFloat) = (!isnan(x) & (isnan(y) | signless(x
...
+`isless` defines a fixed total order that ascends, while `isgreater` defines a
+fixed total order that descends. Both order values that are normally unordered,
+such as `NaN`, after all regular values and [`missing`](@ref) last. So for
+`isless` these values are biggest and for `isgreater` these values are
+smallest.
+
+# Implementation
+Types should not usually implement this function. Instead, implement `isless`.
+"""
+isgreater(x, y) = _is_reflexive(x) && _is_reflexive(y) ? isless(y, x) : isless(x, y)
```
stevengj, line 175: Should that be `!isless(x, y)` in the final branch?

cmcaine:
I don't think so. That's the branch for NaN and missing and I just want
those sorted the same as for isless.

base/operators.jl
cmcaine: I don't think so. That's the branch for NaN and missing and I just want those sorted the same as for isless.

ericphanson:
Maybe it's too late, but it would be awesome if this made it into 1.5. Very handy function. (Of course, it's easy enough to copy the PR code via
```julia
module FindMin
    import Base.findmin, Base.findmax
    isgreater(x, y) = _is_reflexive(x) && _is_reflexive(y) ? isless(y, x) : isless(x, y)
    _is_reflexive(x) = let eq = x == x
        isa(eq, Bool) && eq
    end
    findmax(f, domain) = mapfoldl(x -> (f(x), x), _rf_findmax, domain)
    _rf_findmax((fm, m), (fx, x)) = isless(fm, fx) ? (fx, x) : (fm, m)
    findmin(f, domain) = mapfoldl(x -> (f(x), x), _rf_findmin, domain)
    _rf_findmin((fm, m), (fx, x)) = isgreater(fm, fx) ? (fx, x) : (fm, m)
end
using .FindMin
```
to get the functionality already).

StefanKarpinski:
Yes, this would be nice to have. What remains to be done here? Is this ready to go?

cmcaine:
I paused on this because I found that the existing methods use an optimisation to avoid calling `isless` for values that are normally incomparable (so the first incomparable element will be returned, regardless of what `isless` wants).

The new definitions don't have that optimisation and are a bit slower, even when it is guaranteed that there are zero incomparables in the element type of the iterable. I wanted to try and work out why that was, but didn't get around to diving into the `isless` code to find out.

I can just not do that and then I think I only need to change the docstrings a bit to make sure we're giving a loose enough guarantee for the weird values.

base/operators.jl
tkf: This definition is the trickiest part of this PR and I think it deserves _at least_ some comments about it in the source code.

Having said that, I still have a very uneasy feeling about this definition.  In particular, I _think_ proving that `isgreater` satisfies transitivity requires that "`_is_reflexive(x) && !_is_reflexive(y)` implies `isless(x, y)`" holds.  This is the case for `NaN` and `missing` but this is not in the API.  It'd be much easier to understand and enforce that `isgreater` is a total order [*] if we just use

```julia
function isgreater2(x, y)
    xisnan = x != x
    xisnan isa Bool || return false  # x is missing
    yisnan = y != y
    yisnan isa Bool || return true  # y is missing
    xisnan && return false
    yisnan && return true
    return isless(y, x)
end
```

In other words, using `_is_reflexive` may look like it makes `isgreater` very generic but it crucially depends on the implementation details.

Anyway, I think this PR has to do at least one of the following:

* Use `isgreater2` (or something equivalent) instead of current `isgreater`.
* Add the aforementioned property of `=`-reflexivity and `isless` to the part of implementation requirement of `=` and `isless`.

(Note: Transitivity is very important as otherwise "min" is not going to be associative. If "min" is not associative, we can't have parallel argmin/findmin/minimum/extrema.)

[*] OK this is not quite enough for user-defined `missing`-like types but we don't have such facility even for `isless` yet.

tkf:
> What remains to be done here?

I think https://github.com/JuliaLang/julia/pull/35316#discussion_r407080894 is still a blocker.

tkf:
I summarized the current issues around reducers in the "`findmax` family" in #36287.  It'd be nice if we can use a more mathematically rigorous and hence optimizable definition of them.

AzamatB:
Bump. Can we get this merged? I hit the case today, where I needed it.

cmcaine:
The only remaining issues are:

1. This is a minor change in semantics.
2. The current implementations are (or were, last I checked) faster due to the thing about skipping isless I mentioned.
3. Should probably replace my isgreater() with something like tkf's which is simpler.

Need to think about if the ::Colon and reducedim methods need to be done now or not. If they're not done they may be inconsistent.

Similarly, min, and maximum will be a bit inconsistent with findmax if not updated here, but they're not consistent now, so it probably doesn't really matter. Though would be good to fix.

cmcaine:
I can rebase and stuff later and have a look, but might be good to get a steer on whether it is okay to make the existing funcs a bit slower or not.

StefanKarpinski:
While I can see the motivation for `isgreater` for helping to implement these kinds of functions, but I'm hesitant about exporting it. My worry is that people will be tempted to modify or use `isgreater` independent of `isless`, which might be problematic. The legitimate use of `isgreater` is to impelment the same kind of logic as this implements for `findmax/min` and `argmax/min`, so maybe that's ok. If that's all people use it for, ok. But if people define `isless` and then think they also need to define `isgreater` to match or that they can define `isgreater` not to match... that's what worries me.

cmcaine:
Did triage have any opinion about whether it is okay to make `findmax` and friends about ~20% slower for `Float64` in the name of consistency (implementations and benchmarking code below)? At the moment they return the first value that is not `==` to itself, if any value is like that, and that speeds up the implementation of `isless` even when the vector contains only values that are `==` to themselves. The new definitions based on `isgreater` and `isless` will always return whichever value the total ordering says is right.

The most common values that are not `==` to themselves are `NaN`, `missing` and tuples containing those values. The current methods don't support `missing`, so only users using weird types or tuples that contain these values will notice a difference in output, but all users will experience the slowdown.

Here's how the function outputs would change:

```jl
# These won't change
isless((NaN, 10), (NaN, 20)) # true
isless((NaN, 20), (NaN, 10)) # false

# Current implementation
findmax_orig([(NaN, 10), (NaN, 20)]) # (NaN, 10)
findmax_orig([(NaN, 20), (NaN, 10)]) # (NaN, 20)

# Proposed consistent implementation
findmax_new( [(NaN, 10), (NaN, 20)]) # (NaN, 20)
findmax_new( [(NaN, 20), (NaN, 10)]) # (NaN, 20)
```

<details>
<summary>Implementations and benchmarking code</summary>

```julia
findmax_orig(a) = _findmax_orig(a, :)

function _findmax_orig(a, ::Colon)
    p = pairs(a)
    y = iterate(p)
    if y === nothing
        throw(ArgumentError("collection must be non-empty"))
    end
    (mi, m), s = y
    i = mi
    while true
        y = iterate(p, s)
        y === nothing && break
        m != m && break
        (i, ai), s = y
        if ai != ai || isless(m, ai)
            m = ai
            mi = i
        end
    end
    return (m, mi)
end


# Smallest possible change (2 lines)
findmax_new(a) = _findmax_new(a, :)

function _findmax_new(a, ::Colon)
    p = pairs(a)
    y = iterate(p)
    if y === nothing
        throw(ArgumentError("collection must be non-empty"))
    end
    (mi, m), s = y
    i = mi
    while true
        y = iterate(p, s)
        y === nothing && break
        # m != m && break
        (i, ai), s = y
        if #= ai != ai || =# isless(m, ai)
            m = ai
            mi = i
        end
    end
    return (m, mi)
end


# This PR
_rf_findmax((fm, m), (fx, x)) = isless(fm, fx) ? (fx, x) : (fm, m)
findmax_2arg(f, domain) = mapfoldl(x -> (f(x), x), _rf_findmax, domain)

# How the existing functions might be implemented with this PR
findmax_fold(a) = findmax_2arg(idx -> a[idx], keys(a))
# Same speed
# findmax_fold(a) = mapfoldl(((k, v),) -> (v, k), _rf_findmax, pairs(a))


# Differences in behaviour
isless((NaN, 10), (NaN, 20)) # true
isless((NaN, 20), (NaN, 10)) # false

# Inconsistant with folding `isless`
findmax_orig([(NaN, 10), (NaN, 20)]) # (NaN, 10)
findmax_orig([(NaN, 20), (NaN, 10)]) # (NaN, 20)

# Consistant
findmax_new( [(NaN, 10), (NaN, 20)]) # (NaN, 20)
findmax_new( [(NaN, 20), (NaN, 10)]) # (NaN, 20)
findmax_fold([(NaN, 20), (NaN, 10)]) # (NaN, 20)
findmax_fold([(NaN, 10), (NaN, 20)]) # (NaN, 20)


using BenchmarkTools

input = rand(Int, 10^6);

@btime findmax_orig($input) # 920 μs
@btime findmax_new($input)  # 920 μs
@btime findmax_fold($input) # 1.1 ms

inputf = rand(Float64, 10^6);

@btime findmax_orig($inputf) # 2.3 ms
@btime findmax_new($inputf)  # 2.8 ms
@btime findmax_fold($inputf) # 2.8 ms

inputfbig = rand(Float64, 10^9);

@btime findmax_orig($inputfbig) # 2.5 s
@btime findmax_new($inputfbig)  # 3.0 s
@btime findmax_fold($inputfbig) # 2.9 s
```

</details>


I did put a note in the docs for `isgreater` saying that people should only define `isless` for their types, I don't know how much we can do to prevent people shooting themselves in the feet. Anyway, I'm happy for us to move forward without exporting `isgreater`. If someone wants it, that could be a separate PR.

base/operators.jl
```diff
@@ -157,6 +157,27 @@ isless(x::AbstractFloat, y::AbstractFloat) = (!isnan(x) & (isnan(y) | signless(x
...
+_is_reflexive(x) = let eq = x == x
+    isa(eq, Bool) && eq
+end
```
vtjnash, line 178:
```suggestion
_is_reflexive(x) = (x == x) === true
```

StefanKarpinski:
Bump. Would be great to get this rebased with Jameson's suggsted change applied (and without isgreater export).

cmcaine:
Done. Sorry for the delay :)

Jameson said on slack that making the existing findmax/min slower in the name of consistency is okay, so I've done that too and `_findmax(a, ::Colon)` is now implemented with the new 2-arg form.

The work never ends, though, because now `findmax(A; dims)` has the old behaviour and `findmax(A)` has the new behaviour. Differences will only be observed in arrays containing `NaN` at positions other than the end and in arrays containing `missing` (which will error with the old behaviour). I'll take a look at that later today and see if I can give `findmax(A; dims)` the new behaviour too.

base/reduce.jl
cmcaine: I've done `_findmax(a, ::Colon) = findmax(idx -> a[idx], keys(a))` because `pairs` materialises a whole vector of `Pair`s for an array input whereas `keys` gives a simpler iterator.

This is a bit of a tradeoff: `pairs` would be faster for `Dict`s, I think, but `Array`s are the common case, so I went with something that optimises for that.

cmcaine: I used `keys` rather than `pairs` here, but maybe I should have used `pairs`. I'll benchmark in a little bit and maybe change it.

test/operators.jl
```diff
@@ -83,6 +83,16 @@ import Base.<
...
 @test isless('a','b')
 
+@testset "isgreater" begin
+    let isgreater = Base.isgreater
+        @test !isgreater(missing, 1)
+        @test  isgreater(5, 1)
+        @test !isgreater(1, 5)
+        @test  isgreater(1, missing)
+        @test  isgreater(1, NaN)
+    end
```
tkf, line 93: `@testset` already creates a scope so we don't need `let`. Let's minimize the diff just in case we could convince @StefanKarpinski and export `isgreater` at some point :)

```suggestion
    isgreater = Base.isgreater
    @test !isgreater(missing, 1)
    @test  isgreater(5, 1)
    @test !isgreater(1, 5)
    @test  isgreater(1, missing)
    @test  isgreater(1, NaN)
```

tkf (commented):
I think this implementation of `isgreater` is incorrect? I get:

```julia
julia> isgreater(x, y) = _is_reflexive_eq(x) && _is_reflexive_eq(y) ? isless(y, x) : isless(x, y)
isgreater (generic function with 1 method)

julia> _is_reflexive_eq(x) = (x == x) === true
_is_reflexive_eq (generic function with 1 method)

julia> min1(x, y) = ifelse(isgreater(x, y), y, x)
min1 (generic function with 1 method)

julia> function isgreater2(x, y)
           xisnan = x != x
           xisnan isa Bool || return false  # x is missing
           yisnan = y != y
           yisnan isa Bool || return true  # y is missing
           xisnan && return false
           yisnan && return true
           return isless(y, x)
       end
isgreater2 (generic function with 1 method)

julia> min2(x, y) = ifelse(isgreater2(x, y), y, x)
min2 (generic function with 1 method)

julia> min((NaN, missing), (missing, NaN))
(NaN, missing)

julia> min1((NaN, missing), (missing, NaN))
(missing, NaN)

julia> min2((NaN, missing), (missing, NaN))
(NaN, missing)
```

Sorry to be persistent, but, as I mentioned in https://github.com/JuliaLang/julia/pull/35316#discussion_r419830391, I think this pattern `... ? isless(y, x) : isless(x, y)` is hard to reason about and it'd be better to have a verbose and more "honest" implementation like `isgreater2` as I've been suggesting. You can probably fix `_is_reflexive_eq` so that `min1` and `min` behave identically. So I don't think the linked comment is the source of the bug. But I think the attempt to "minimize" the code size may be a contributing factor.

test/operators.jl
cmcaine: Thanks!

cmcaine: Thanks!

cmcaine:
I'm sorry that I didn't address that earlier. I thought you were probably right, but after looking into it some more I think my initial approach is better and that `isgreater2` is wrong for tuples and other containers. Using the definitions from your comment:

```jl
julia> isgreater2((missing, NaN), (NaN, missing))                                                                                                              
false                                                                                                                                                          
                                                                                                                                                               
julia> isgreater2((NaN, missing), (missing, NaN))                                                                                                              
false
```

One of the pair has to evaluate to `true` or the function is not a total order. And then we obviously get bad stuff like this:

```julia
julia> min2((NaN, missing), (missing, NaN))                                                                                                                    
(NaN, missing)                                                                                                                                                 
                                                                                                                                                               
julia> min2((missing, NaN), (NaN, missing))                                                                                                                    
(missing, NaN)
```

Comparing with `min` is probably not the best either because `min` exhibits exactly the kind of inconsistency that prompted you to suggest `isgreater` in the first place :)

`min` uses `isless` in the general case. So to get the right result for reals, `min` uses partial order (`<`), so `NaN` gets sorted right. To get the right result for `missing`, there are special definitions of `min(::Any, ::Missing) = missing` and `min(::Missing, ::Any) = missing`.

That makes the ordering of `min` weird if you wrap things in tuples, vectors or other containers:

```jl
min(1, NaN) === NaN
min( (1,), (NaN,) ) === (1,)
min( [1], [NaN] ) == [1]
```

If `min` used `isgreater` then simply wrapping a value in a container would not change the order:

```jl
min1(x, y) = ifelse(isgreater(x, y), y, x)

min1(1, NaN) === NaN
min1( (1,), (NaN,) ) === (NaN,)
min1( [1], [NaN] ) # [NaN]
```

That's a change in behaviour, but I think it's better behaviour. We also wouldn't need special sauce for floats or missing and all the other functions would come out consistent and easy.

In general, I don't think we can satisfy all of:

1. `min` does not change it's current ordering
2. `min(x...)` and `findmin(x)` should be consistent for all `x`
3. `findmin` returns `missing` or `NaN` type in preference to any "normal" value
4. `NaN` and `missing` values are returned from both `findmin` and `findmax` in the same order

There's two ways out of it that I see:

1. `min` uses `isgreater` instead of `isless` in at least the fallback implementation.
2. `min` keeps it's current definition and we're satisfied with the functions being consistent almost all the time (not when used with tuples of weird values)

Currently, this PR implements scenario 2 and I think that's fine. A follow up PR could maybe change `min` if that is not considered too breaking.

Here's some reproducible code that explores the consequences of those two scenarios for three inputs:

<details>

```julia
x = (-1.0, NaN, 1.0)
y = ((NaN, missing), (missing, NaN))
z = ((1, 2), (missing, NaN), (NaN, missing))

# isgreater from this PR
_is_reflexive_eq(x) = (x == x) === true
isgreater(x, y) = _is_reflexive_eq(x) && _is_reflexive_eq(y) ? isless(y, x) : isless(x, y)

# min and minimum compatible with "Scenario 1"
min1(x, y) = ifelse(isgreater(x, y), y, x)
minimum1(itr) = foldl(min1, itr)

# findmin compatible with this PR
findmin_(itr) = reverse(foldl(
                    ((k1, v1), (k2, v2)) -> ifelse(isgreater(v1, v2), (k2, v2), (k1, v1)),
                    pairs(itr)
                ))

# Scenario 1
# x, y, z all consistent:

@assert minimum1(x) === NaN === findmin_(x)[1]
@assert minimum1(y) === (missing, NaN) === findmin_(y)[1]
@assert minimum1(z) === (missing, NaN) === findmin_(z)[1]

# Scenario 2
# x is fine, y and z are weird.

@assert minimum(x) === NaN === findmin(x)[1]

@assert minimum(y) === (NaN, missing)
@assert findmin_(y)[1] === (missing, NaN)

@assert minimum(z) === (1, 2)
@assert findmin_(z)[1] === (missing, NaN)

@assert minimum(x) === NaN
@assert minimum(y) === (NaN, missing)
@assert minimum(z) === (1, 2)
```

</details>

tkf:
Thanks for the analysis!

> Comparing with `min` is probably not the best either because `min` exhibits exactly the kind of inconsistency that prompted you to suggest `isgreater` in the first place :)

Ah, yes, you are right. I totally missed that `min` is not "correct" from the perspective I'm pushing towards in my last comment (i.e., proper total ordering so that `Base`-compatible parallel implementation is possible).

Hmm... This is super hard. I would even consider this a bug of `min` but it is underspecified so I'm not really sure.

@StefanKarpinski @nalimilan @andyferris What do you think? (Also please ping other people who may be interested in this.)

That is to say, do we want, for all `x`, `y`, `xs`, and `ys`,

```julia
min(x, y) === y    ⇒    min((x, xs...), (y, ys...)) === (y, ys...)
```

holds (if none of them throw)?  An example where this fails is:

```julia
julia> min(NaN, missing)
missing

julia> min((NaN, missing), (missing, NaN))
(NaN, missing)
```

Maybe define `Future.min` or `Compat.min` (and `.minimum` etc.) that does the "right" thing?

---

Separated from the above issue on `min`, I think we still have two separated issues:

### _isgreater_ for tuples and arrays

We still don't have a concrete implementation of _isgreater_ with total ordering on containers. Both `isgreater` and `isgreater2` fail `check_tuple_order` test below:

```julia
function check_tuple_order(≺)
    a ≍ b = !(a ≺ b) && !(b ≺ a)
    a ≼ b = a ≺ b || !(b ≺ a)
    p ⇒ q = !p || q
    tuples = (Tuple(p) for p in permutations([NaN, missing, 0, 1]))
    @testset for a in tuples, b in tuples
        @test (a ≍ b) === isequal(a, b)
        @test cmp(≺, a, b) == lexcmp(≺, a, b)  # lexcmp defines lexicographical ordering in terms of `cmp`

        # Total ordering w.r.t to these tuples (a bit redundant):
        @test (a ≼ b && b ≼ a) ⇒ isequal(a, b)
        @test a ≼ b || b ≼ a
        @testset for c in tuples
            @test (a ≼ b && b ≼ c) ⇒ (a ≼ c)
        end
    end
end
```

<details>
 <summary>Full code</summary>

```julia
using Combinatorics: permutations
using Test

function lexcmp(≺, as, bs)
    for (a, b) in zip(as, bs)
        y = cmp(≺, a, b)
        y == 0 || return y
    end
    return cmp(≺, length(as), length(bs))
end

function check_tuple_order(≺)
    a ≍ b = !(a ≺ b) && !(b ≺ a)
    a ≼ b = a ≺ b || !(b ≺ a)
    p ⇒ q = !p || q
    tuples = (Tuple(p) for p in permutations([NaN, missing, 0, 1]))
    @testset for a in tuples, b in tuples
        @test (a ≍ b) === isequal(a, b)
        @test cmp(≺, a, b) == lexcmp(≺, a, b)

        # Total ordering w.r.t to these tuples (a bit redundant):
        @test (a ≼ b && b ≼ a) ⇒ isequal(a, b)
        @test a ≼ b || b ≼ a
        @testset for c in tuples
            @test (a ≼ b && b ≼ c) ⇒ (a ≼ c)
        end
    end
end

isgreater(x, y) = _is_reflexive_eq(x) && _is_reflexive_eq(y) ? isless(y, x) : isless(x, y)
_is_reflexive_eq(x) = (x == x) === true

function isgreater2(x, y)
    xisnan = x != x
    xisnan isa Bool || return false  # x is missing
    yisnan = y != y
    yisnan isa Bool || return true  # y is missing
    xisnan && return false
    yisnan && return true
    return isless(y, x)
end

@testset "lexcmp" begin
    # for normal numbers, `lexcmp` on `<` and `>` should just flip things
    tuples = (Tuple(p) for p in permutations([0, 1, -1]))
    @testset for a in tuples, b in tuples
        @test lexcmp(<, a, b) == cmp(<, a, b) == -lexcmp(>, a, b) == -cmp(>, a, b)
    end
end

@testset "check_tuple_order" begin
    @testset "$f" for f in [isless, isgreater, isgreater2]
        check_tuple_order(f)
    end
end
```

</details>

The failure modes of `isgreater` and `isgreater2` are different:

```julia
julia> a = (NaN, 0.0)
       b = (NaN, 1.0);

julia> lexcmp(isgreater, a, b), cmp(isgreater, a, b)
(1, -1)

julia> lexcmp(isgreater2, a, b), cmp(isgreater2, a, b)
(1, 0)
```

I think the straightforward way to implement this is to just add `isgreater(::Tuple, ::Tuple)` and `isgreater(::AbstractArray, ::AbstractArray)`.

In the future, it'd be better for the containers to define comparison operator for `isless`/`isgreater` in "one go" (like `lexcmp` above). But I think we don't need to do this in this PR, especially since `isgreater` is not exported yet.

### `==`-reflexivity ("`_is_reflexive_eq`")

`isgreater` defined in this PR fails to derive a total-ordering for the following objects with well-defined `isless`/`isequal`

```julia
abstract type X end
struct A <: X end
struct B <: X end
struct C <: X end
Base.:(==)(::B, ::B) = false
Base.isequal(x::X, y::X) = x === y
Base.isless(::X, ::X) = false
Base.isless(::A, ::A) = false
Base.isless(::A, ::X) = true
Base.isless(::B, ::C) = true

using Test

function check_total_ordering(≺, everything = (A(), B(), C()))
    a ≍ b = !(a ≺ b) && !(b ≺ a)
    a ≼ b = a ≺ b || !(b ≺ a)
    p ⇒ q = !p || q
    @testset for a in everything, b in everything
        @test (a ≼ b && b ≼ a) ⇒ isequal(a, b)
        @test a ≼ b || b ≼ a
        @testset for c in everything
            @test (a ≼ b && b ≼ c) ⇒ (a ≼ c)
        end
    end
end

@testset "check_tuple_order" begin
    @testset "$f" for f in [isless, isgreater2, isgreater]
        check_tuple_order(f)
    end
end
```

That said, `isgreater2` fails while `isgreater` succeeds if I define

```julia
Base.:(==)(::A, ::A) = false
Base.:(==)(::C, ::C) = false
```

instead of `Base.:(==)(::B, ::B) = false`.

Also, trying other example, I think `isgreater` is well-behaved in more cases than `isgreater2` so I'll stop recommending `isgreater2`. But I think we should stop flipping `isless` based on the behavior of `==`.

`==` doesn't have well-defined algebraic properties.  I don't think we should contaminate the properties of `isgreater`/`isequal` with `==`.

I need to think about it a bit more but ATM my hunch is that we'd need to special case `AbstractFloat` and `Missing` (but I'm not sure). This is probably the easiest option to finish up the PR, though.

andyferris:
Oh wow. I always assumed that

```julia
min(x,y) = isless(x,y) ? x : y
```

and incompatible specializations are buggy. But I see that's not true for `AbstractFloat`. Things like `minimum([1,2,3,NaN])` and `maximum([1,2,3,NaN])` are *both* `NaN`.

If you aren't basing `min` and `max` on something reflexive and transitive like `isless` and `isequal` than you can't expect the result to be reflexive and transitive. That seems natural.

Similary, what is the question about `isgreater(x, y)`? Either it equals `!isequal(x, y) && !isless(x, y)` (if it doesn't throw) or else it's incorrect, right? I don't see *any* space for interpretation on that one.

(I have a definition like this in AcceleratedArrays, with which you might want to do things like seach an array already sorted via `isless` and `isequal` for the `minimum` or `maximum` element and so-on - which via the "vanilla" definition is the first and last element, and therefore a much faster question to answer. Other definitions using `<` aren't particularly useful for engineery-style work - e.g. if finding the `minimum` element of a b-tree can't be guaranteed to be O(log(N)) then we appear to be doing something rather strange.

andyferris:
Perhaps I just need favours for `lesser` (`min`), `greater` (`max`), `least` (`minimum`), `greatest` (`maximum`) for the "predictable" versions that use `isless` and `isequal`.

base/operators.jl
```diff
@@ -168,6 +168,23 @@ isless(x::AbstractFloat, y::AbstractFloat) = (!isnan(x) & (isnan(y) | signless(x
...
+`isless` defines a fixed total order that ascends, while `isgreater` defines a
+fixed total order that descends. Both order values that are normally unordered,
+such as `NaN`, after all regular values and [`missing`](@ref) last. So for
+`isless` these values are biggest and for `isgreater` these values are
+smallest.
+
+# Implementation
+Types should not usually implement this function. Instead, implement `isless`.
+"""
+isgreater(x, y) = _is_reflexive_eq(x) && _is_reflexive_eq(y) ? isless(y, x) : isless(x, y)
```
andyferris, line 186: I wonder if it would be better to call this something else. I can see it's necessary for supporting the other functions below and have them match `min`, `max`, `minimum`, `maximum` - but it might be confusing if this were ever exported as not being the opposite of `isless`?

andyferris:
Delving in slightly further, you can see that 

```julia
julia> min((NaN, missing), (missing, NaN))
(NaN, missing)
```
doesn't obey the rule that `missing` is "more missing/extremal" than `NaN` because `min` and `max` on tuples is defined via `isless` without any *special* consideration of `NaN` and `missing`.

I personally can't see any other simple way to make all of this perfectly generic or composable with nested containers, third-party types, etc. If a third party wants to define a missing-like value they can choose to overload `isless` ("it's bigger than `missing` but smaller than `NaN`") and have things flow naturally... or we'd need everyone else who overloads `minimum`, `findmin`, `argmin`, etc for their container to know about this new third-party type, or else have their container be known by this third-party type's package. I.e. you lose out on blind composition.

(Or else you could define a generic API or trait to overload to talk about degrees of extremality/missingness... `is_missinglike(NaN)` is `true`, `promote_missinglike(NaN, missing)` goes to `missing`. Which I think is the same as working through a public API for `_is_reflexive_eq(x)` and a new one to discuss a system of promotion).

tkf:
> Oh wow. I always assumed that
> 
> ```julia
> min(x,y) = isless(x,y) ? x : y
> ```
> 
> and incompatible specializations are buggy.

@andyferris So I guess you are in favor of "fixing" `min`? :)

> Delving in slightly further, you can see that
> 
> ```julia
> julia> min((NaN, missing), (missing, NaN))
> (NaN, missing)
> ```
> 
> doesn't obey the rule that `missing` is "more missing/extremal" than `NaN` because `min` and `max` on tuples is defined via `isless` without any _special_ consideration of `NaN` and `missing`.

Yeah, I think this is exactly kind of the problem we need to address in this PR. I think "`findmax` family" of functions is OK since the notion of "more missing/extremal" is compatible with the ordering defined by `isless`.

The main problem is that "`findmin` family" of functions has to treat `NaN` and `missing` specially. I believe we should do this by defining a proper ordering so that they are parallelizable.

For the first step, treating `Missing` and `AbstractFloat` separately sounds totally reasonable to me. We can design an extensible API later once we have a "correct" implementation.

cmcaine:
So the whole reason that this has been a pain to do is that users expect `NaN`, `missing` and *maybe* user-defined values to "poison" results  e.g. `minimum([1, 2, 3, NaN, 4, 5]) == NaN` despite not being orderable by `<`.

Stepping back a bit, I wonder when users actually want poisoning to happen? If they don't want NaNs in tuples to poison, then we can avoid all of the difficult stuff.

And is it actually more confusing if `min( (1,), (NaN,) ) === (NaN,)`?

If we leave containers ordered by `isless` (like `min` does) we can just special case floats and missing by moving code from `min` to `isgreater` or by re-using `isless` like this:

```jl
is_poisoning(x) = false
is_poisoning(x::AbstractFloat) = isnan(x)
is_poisoning(x::Missing) = true
isgreater(a, b) = is_poisoning(a) || is_poisoning(b) ? isless(a, b) : isless(b, a)
min1(a, b) = isgreater(a, b) ? b : a

# Examples:
min1(1, 2) === 1 === min(1, 2)
min1(1, NaN) === NaN === min(1, NaN)
min1( (1,), (NaN,) ) === (1,) === min( (1,), (NaN,) )
```

It's simple and compatible with the existing `min` and `minimum` (and they could be defined with isgreater, if we wanted). We won't provide a new descending lexigraphical order, but do users actually want that? I don't think I care.

I wrote a longer post that summarised this PR, existing issues with `isgreater` and so on, but that's all irrelevant if we can agree on this simpler behaviour.

andyferris:
> @andyferris So I guess you are in favor of "fixing" `min`? :)

I suspect there's far too much `NaN` handling code that understand the status quo out there for changes here to be considered a "minor change". Changing this in 1.x would break a lot of research or data manipulation code which we wouldn't see with PkgEval. So it's probably not feasible in the short term, and even if it were there still is:

> I wonder when users actually want poisoning to happen?

This was really never clear to me. (For myself, during my work I find that poisoning by `missing` frequently gets in the way of what I'm trying to do, but on the other hand invisible `NaN`s are the _worst_. On reflection it's not the poisoning of `missing` that upsets me but its the resulting errors of e.g. `==` or `<` or `in` no longer being predicate functions and breaking `if` statements everywhere).

> If they don't want NaNs in tuples to poison, then we can avoid all of the difficult stuff.

I think this is where my suggestion of a "simpler" family of functions `lesser`, `least`, `findleast`, etc that could use the total ordering of `isless` comes in.

The other operation you might want to do is find the non-poisoning extrema (ignoring missing, NaN, etc).

---------

To improve `min` etc I think something close to @cmcaine's version above could work nicely. Make the `is_poisining` (needs renaming) function a public API, and make the `isgreater` (also needs renaming IMO - perhaps an `ismin` or `ismax` kind of thing?) function a public API that containers like `Tuple` should use and overload. We neatly use `isless` to define the "ordering" of extremality - so I think that all the problems about blind composition that I raised are actually solved. If you want to find the `minimum` of a btree or sorted collection you can always do it quickly by just considering the first and last element (which is excellent as I'll be able to finish AcceleartedArrays, as we also know where all the non-reflexive elements go so we can also deal with `<` and `>` predicates correctly and efficiently...).

base/operators.jl
StefanKarpinski: Also perhaps it should be called `ismore` since the analogue of `isgreater` would be is `islesser`.

andyferris: I always read it like “is less than” and “is greater than” (regular comparatives) rather than “is the lesser of” and “is the greater of” (the rarer form of two-item-only comparisons like latter and former that I can’t remember the fancy grammar name of).

Nonetheless I do kind of like `ismore`.

cmcaine:
The latest series of commits:

- makes `isgreater` consistent with how `min` is at the moment
- makes the `reducedim.jl` variants use `isgreater` and `isless` in the same way as the others. I've also fixed `minimum(A; dims)` and `maximum(A; dims)` so that they don't blow up or produce bad answers if `A` contains `missing`.
- uses the `pairs` iterator in `findmax(itr)`.

I think this is ready for (final?) review.

This is the part that I am least confident of: https://github.com/JuliaLang/julia/pull/35316/commits/b92ca0cdb47faffcbfe4bbccab793be8ed72fdc5#diff-cc983f27f9c33e9b278f74989b8aaca48e461e0717bb96ea116d02754554746fL147-R157

cmcaine:
I still haven't finished adjusting all the docstrings, actually, so don't comment on those until I edit this comment 👍

base/reducedim.jl
```diff
@@ -144,10 +144,18 @@ for (f1, f2, initval) in ((:min, :max, :Inf), (:max, :min, :(-Inf)))
...
+
+            # but NaNs and missing need to be avoided as initial values
+            if (v0 != v0) === true
+                v0 = typeof(v0)($initval)
+            elseif ismissing(v0)
+                # If it's a union type, pick the initval from the other type.
+                if typeof(T) == Union
+                    v0 = (T.a == Missing ? T.b : T.a)($initval)
+                end
+            end
```
cmcaine, line 157: I think `Missing` is always `T.a` if it's in a Union, it's probably `type_morespecific` than anything but `Any`, but I didn't want to rely on that.

I was also too tired to try to work out if I could avoid defining a `v0` altogether, though I'm not hopeful about that.

cmcaine:
Test failure seems unrelated, btw.

base/reducedim.jl
Moelf: Missing is not necessary always T.a AFAIK it puts singleton to the first position and Missing is often the only singleton but this is not guaranteed (there's no intended behavior in this)

nalimilan: You could also do `nonmissingtype(T)($initval)`. Actually I'm not even sure you need to check that the `T` is a `Union` nor to extract the type, as `convert(T, $initval)` should work thanks to special `convert` methods defined for `Missing`.

BTW, this mostly replaces https://github.com/JuliaLang/julia/pull/35323 (which is less efficient but also works when the type isn't `Union{T, Missing}`).

base/reduce.jl
```diff
@@ -762,6 +762,183 @@ Inf
 """
 minimum(a; kw...) = mapreduce(identity, min, a; kw...)
 
+## findmax, findmin, argmax & argmin
+
+"""
+    findmax(f, domain) -> (f(x), x)
+    findmax(f)
+
+Returns a pair of a value in the codomain (outputs of `f`) and the corresponding
```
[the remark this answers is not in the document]
cmcaine: Thanks for the review! I would love this to be easy to understand, but this PR does also redefine these functions to be a bit more general. A big part of that is that the result is no longer a value and an index in the general case, so the old docstring is not correct.

I do explain below what happens when the `domain` argument is omitted:

```
When `domain` is omitted, `f` must have an implicit domain. In particular, if
`f` is an indexable collection, it is interpreted as a function mapping keys
(domain) to values (codomain), i.e. `findmax(itr)` returns the maximal element
of the collection `itr` and its index.
```

Do you think that is too unclear?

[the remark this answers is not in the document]
StefanKarpinski: I find the domain, codomain terminology very helpful so it would be nice to keep it even if only parenthetically.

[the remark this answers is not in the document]
cmcaine: I did consider splitting them, but I wanted to make the link between the concepts clear and encourage uses of `findmax` for other types that might have implicit domains. But I can try splitting them out and seeing what I get, I suppose.

base/reducedim.jl
cmcaine: Using `convert` seems like a good idea. Perhaps this?

```julia
            # but NaNs and missing need to be avoided as initial values
            if is_poisoning(v0)
                # convert handles Missing specially.
                v0 = convert(T, $initval)
            end
```

cmcaine: The previous code what as a bit more robust because it only special cases `NaN` and then inserts `Inf` or `-Inf`, but for unions with missing `T` could be anything. I wonder if this is better:

```jl
            if is_poisoning(v0)
                v0 = T == Any ? $initval : typemin(nonmissingtype(T))
            end
```

Or, if `nonmissingtype` isn't optimised out by the compiler:

```jl
            if ismissing(v0)
                v0 = T == Any ? $initval : typemin(nonmissingtype(T))
            elseif (v0 == v0) !== true
                # v0 is NaN
                v0 = $initval
            end
```

nalimilan: Actually I'm not sure either solution is correct. `initval` (which is `Inf` or `-Inf`) may not be convertible to `T` (e.g. for integers), and `typemin`/`typemax` may be undefined (e.g. for `BigInt` or abstract types). That's why at https://github.com/JuliaLang/julia/pull/35323 I adopted an inefficient but robust approach. Maybe we could use `$initval` when `nonmissingtype(T) <: AbstractFloat` just like the `NaN` path does, use `typemin`/`typemax` if it doesn't throw an error, and otherwise use the approach from https://github.com/JuliaLang/julia/pull/35323.

cmcaine: Good catch. What a pain. Just introducing a `try/catch` will slow things down, though maybe the typical size of `A` is large enough that it doesn't matter.

cmcaine: Maybe this code could be re-written so that it doesn't need an initial value? Or it just uses the first value or something?

cmcaine: Maybe something like this:

```jl
            if is_poisoning(v0)
                Tnm = nonmissingtype(T)
                if Tnm <: AbstractFloat
                    v0 = $initval
                else
                    try
                        v0 = $typeextreme(Tnm)
                    catch e
                        e isa MethodError || rethrow()
                        v0 = mapreduce(f, $f2, skipmissing(A))
                    end
                end
            end
```

I'd be happy to just leave `minimum(A; dims)` broken on `missing` and just get this PR merged.

nalimilan: Looks like it. But indeed you could leave that out of this PR, anyway it's been broken for a long time.

cmcaine: That's still not right because mapreduce will error if `skipmissing(A)` is empty.

I think this is right, but it's not pretty.

```jl
            T = _realtype(f, promote_union(eltype(A)))
            Tr = v0 isa T ? T : typeof(v0)

            # but NaNs and missing need to be avoided as initial values
            if is_poisoning(v0)
                Tnm = nonmissingtype(Tr)
                if Tnm <: AbstractFloat
                    v0 = $initval
                else
                    try
                        v0 = $typeextreme(Tnm)
                    catch e
                        e isa MethodError || rethrow()
                        # Find the min/max value in the array,
                        # skipping missing.
                        for v in A
                            fv = f(v)
                            if !ismissing(fv)
                                v0 = ismissing(v0) ? fv : $f2(v0, fv)
                            end
                        end
                    end
                end
            end
```

StefanKarpinski:
Wow, this was much harder than could possibly have been anticipated! Thanks for sticking it out, @cmcaine. Once @nalimilan's issues have been addressed, this seems good to go. Triage approves 🍎

StefanKarpinski:
Oh, one thing that came up on the triage call was that the definition `is_poisoning(x) = (x == x) !== true` seems to be better and if we recall correctly was used earlier in this PR (or maybe elsewhere?). Is there a reason not to use that definition? We may want to call it `isunordered` instead which I personally would be ok with exporting even.

JeffBezanson:
Yes, LGTM now!

JeffBezanson (approved):

cmcaine:
> Oh, one thing that came up on the triage call was that the definition is_poisoning(x) = (x == x) !== true seems to be better and if we recall correctly was used earlier in this PR (or maybe elsewhere?). Is there a reason not to use that definition? We may want to call it isunordered instead which I personally would be ok with exporting even.

We did previously use that :) It gives the wrong result for containers like tuples (previously I thought it was the right result, but it's not). The new behaviour matches `isless` and `min`.

The old behaviour would imply `min` should give this result `min( (2, NaN), (1, 0) ) === (2, NaN)`, which isn't consistent with how `min` or `isless` work.

Concretely:

```julia
isunordered_old(x) = (x == x) !== true

isunordered_new(x) = false
isunordered_new(x::AbstractFloat) = isnan(x)
isunordered_new(x::Missing) = true

isunordered_old( (NaN, 1) ) == true

# This result is required to match the behaviour of `min`.
isunordered_new( (NaN, 1) ) == false
```

I've renamed, added docs, and exported.

> Once @nalimilan's issues have been addressed, this seems good to go.

Great! I think I've done that and I've rebased, added a NEWS entry for `isunordered` and added compat admonitions to all the new methods and functions.

If CI passes then I think this is good to go!

cmcaine:
CI passes. I think this is ready to merge.

StefanKarpinski:
giphy

cmcaine:
At times I doubted this would ever be done ;)

I started work on this [on the 11th March 2020](https://github.com/JuliaLang/julia/issues/27613#issuecomment-597822988) and the issue was originally opened on the 16th June 2018.

```julia
julia> using Dates

julia> today() - Date(2020, 3, 11)
282 days

julia> today() - Date(2018, 6, 16)
916 days
```

Big thanks to @tkf and to the various reviewers!

mykelk:
Yay! We're using this syntax in [our new book](http://algorithmsbook.com/). Thanks for all the hard work to get this right!

PR #35321, opened 2020-03-31
Added tests for hashing unsigned integers to 32-bit hashes.

sudo-rushil:
New tests make sure that expected 32-bit hashes are generated for arbitrarily selected unsigned 32- and 64-bit integers.


JeffBezanson:
Welcome, @sudo-rushil !

Since these are internal functions used to implement `hash`, I would think they are already covered by testing `hash`. Is that not the case?

sudo-rushil:
From what I can tell, they are only used to implement hash on 32 bit systems, so technically they should be covered. However, the lack of explicit tests was causing a drop in code coverage, so I wanted to add some just in case. Thanks for taking a look!

PR #35324, opened 2020-03-31
propagate `--color` and `--startup-file` options in `julia_cmd`

JeffBezanson:
It seems like propagating these is pretty commonly wanted.


fredrikekre:
Add this to a new compat note after https://github.com/JuliaLang/julia/pull/35324/files#diff-c20eb2eead14027fab4293e10bacf73bR436 ?

PR #35325, opened 2020-03-31
Sparse addition of Symmetric and Hermitian matrices

### LINKED ISSUE #710: Adding sparse Hermitian matrices makes the result dense
Is this fixable in 1.x releases?
```julia
julia> using SparseArrays; A = sprandn(10, 10, 0.1); B = sprandn(10, 10, 0.1);

julia> typeof((A+A') + (B+B'))
SparseMatrixCSC{Float64,Int64}

julia> typeof(Hermitian(A+A') + Hermitian(B+B'))
Hermitian{Float64,Array{Float64,2}}
```
I'd argue that it's a bug that the final line doesn't give `Hermitian{Float64,SparseMatrixCSC{Float64,Int64}}`.

### LINKED ISSUE #35317: Adding sparse Hermitian matrices makes the result dense
Is this fixable in 1.x releases?
```julia
julia> using SparseArrays; A = sprandn(10, 10, 0.1); B = sprandn(10, 10, 0.1);

julia> typeof((A+A') + (B+B'))
SparseMatrixCSC{Float64,Int64}

julia> typeof(Hermitian(A+A') + Hermitian(B+B'))
Hermitian{Float64,Array{Float64,2}}
```
I'd argue that it's a bug that the final line doesn't give `Hermitian{Float64,SparseMatrixCSC{Float64,Int64}}`.

MasonProtter:
This is my very naive attempt at fixing https://github.com/JuliaLang/LinearAlgebra.jl/issues/710, but I suspect I did something subtly wrong. In particular, I worry about if I should be doing something to properly account for if the matrices are stored in the upper triangle or lower triangle and how to handle that in the `Hermitian + Hermitian` case. Thoughts?

EDIT: closes JuliaLang/LinearAlgebra.jl#710.


stdlib/SparseArrays/src/linalg.jl
```diff
@@ -7,6 +7,18 @@ using Random: rand!
 const StridedOrTriangularMatrix{T} = Union{StridedMatrix{T}, LowerTriangular{T}, UnitLowerTriangular{T}, UpperTriangular{T}, UnitUpperTriangular{T}}
 const AdjOrTransStridedOrTriangularMatrix{T} = Union{StridedOrTriangularMatrix{T},Adjoint{<:Any,<:StridedOrTriangularMatrix{T}},Transpose{<:Any,<:StridedOrTriangularMatrix{T}}}
 
++(A::Hermitian, B::SparseMatrixCSC) = parent(A) + B
++(A::SparseMatrixCSC, B::Hermitian) = A + parent(B)
+function +(A::Hermitian{<:Any, <:SparseMatrixCSC}, B::Hermitian{<:Any, <:SparseMatrixCSC})
+    Hermitian(parent(A) + parent(B))
```
dkarrasch, line 13: You were rightly concerned about the triangle issue. So the `parent`s don't need to be Hermitian themselves, plus the values that you care about maybe in different triangles for `A` and `B`. In fact, with the current implementation, they even may have the same `:L` triangle, yet the `Hermitian` constructor assumes `:U` by default. So the result could be entirely meaningless.

I think taking care of the orientation can be done by along the lines
```julia
if A.uplo == B.uplo
    return Hermitian(parent(A) + parent(B), sym_uplo(A.uplo))
elseif A.uplo == 'U'
    return Hermitian(parent(A) + permutedims(parent(B)), :U)
else
    return Hermitian(permutedims(parent(A)) + parent(B), :U)
end
```
or so...

dkarrasch (commented):
This is tricky, indeed, because the wrapped matrices don't need to be symmetric/Hermitian themselves. I left a couple of (untested) suggestions, hope they help.

stdlib/SparseArrays/src/linalg.jl
```diff
@@ -7,6 +7,18 @@ using Random: rand!
 const StridedOrTriangularMatrix{T} = Union{StridedMatrix{T}, LowerTriangular{T}, UnitLowerTriangular{T}, UpperTriangular{T}, UnitUpperTriangular{T}}
 const AdjOrTransStridedOrTriangularMatrix{T} = Union{StridedOrTriangularMatrix{T},Adjoint{<:Any,<:StridedOrTriangularMatrix{T}},Transpose{<:Any,<:StridedOrTriangularMatrix{T}}}
 
++(A::Hermitian, B::SparseMatrixCSC) = parent(A) + B
++(A::SparseMatrixCSC, B::Hermitian) = A + parent(B)
```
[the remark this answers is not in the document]
MasonProtter: Mhm, you're right. What would be the right way to approach that?

stdlib/SparseArrays/src/linalg.jl
dkarrasch: Seems like you need to `import LinearAlgebra: sym_uplo`.

stdlib/SparseArrays/src/linalg.jl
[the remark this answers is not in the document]
dkarrasch: How about
```julia
function +(A::SparseMatrixCSC, B::Hermitian)
    C = typeof(B.data)(B)
    C .+= A
    return C
end
```
? At least, that preserves potential sparsity of `B.data`.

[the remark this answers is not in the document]
dkarrasch: This
```julia
function +(A::SparseMatrixCSC, B::Hermitian{<:Any,<:AbstractSparseMatrix})
    C = sparse(B)
    return A + C
end
function +(A::SparseMatrixCSC, B::Hermitian)
    C = Matrix(B)
    C .= A .+ C
    return C
end
```
seems to solve all cases that I could come up with, including things like `B.data isa UpperTriangular` etc.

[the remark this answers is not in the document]
MasonProtter: I apologize for the late replies! Why is that preferable to your earlier one
```julia
function +(A::SparseMatrixCSC, B::Hermitian)
    C = typeof(B.data)(B)
    C .+= A
    return C
end
```
? I'm not sure I see an advantage.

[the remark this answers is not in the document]
MasonProtter: Ah, I see it now. `typeof(B.data)(B)` might not necessarily copy.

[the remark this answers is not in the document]
dkarrasch: Yeah, you were faster than I could type. Also, it could create some structured matrix type, if that's the type of `B.data`, like an `UpperTriangular`, or a `Bidiagonal` etc. So, basically I think we want either a sparse matrix to dispatch to sparse addition, or a dense matrix that we can add "into". This adds a bit of code (especially for the two orientations of each combination), but I think it's a worthy specialization.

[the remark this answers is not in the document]
MasonProtter: I ended up going with `collect` instead of `Matrix`, because I think `Matrix` might not always copy. Does that sound kosher to you?

stdlib/SparseArrays/src/linalg.jl
```diff
@@ -1,12 +1,54 @@
...
+
+function +(A::Symmetric{<:Any, <:SparseMatrixCSC}, B::Symmetric{<:Any, <:SparseMatrixCSC})
+    if A.uplo == B.uplo
+        Symmetric(parent(A) + parent(B), sym_uplo(A.uplo))
+    elseif A.uplo == 'U'
+        Symmetric(parent(A) + permutedims(parent(B)), :U)
+    else
+        Symmetric(permutedims(parent(A)) + parent(B), :U)
+    end
+end
```
MasonProtter, line 50: I'm thinking that I should replace the separate `Hermitian` and `Symmetric` parts with
```julia
for T in (:Symmetric, :Hermitian)
    @eval begin
        ...
    end
end
```
because this would cut down on code duplication. Alternatively, I could define
```julia
SymOrHerm{T, Store} = Union{Symmetric{T, Store}, Hermitian{T, Store}}
```
and then define the methods on that.

dkarrasch: I think the for-loop pattern is common and a bit more appropriate, because you'll need both the specific constructor when returning. Just in case, you can use an already existing 
```julia
const HermOrSym{T,        S} = Union{Hermitian{T,S}, Symmetric{T,S}}
```
no need to define anything new.

MasonProtter: Okay, I did the `eval` for loop, but it's failing to build locally for some reason related to the `eval`, though it works if I do the `eval` in the REPL. Pushing my changes to see if someone can spot what went wrong.

stdlib/SparseArrays/src/linalg.jl
```diff
@@ -1,12 +1,44 @@
...
+            else
+                $HermSym(permutedims(parent(A)) + parent(B), :U)
+            end
+        end
+    end
+end
+
+function (+)(A::Symmetric{<:Any, <:SparseMatrixCSC}, B::Hermitian{<:Any, <:SparseMatricCSC})
+    C = sparse(A)
+    C .= C + B
```
dkarrasch, line 37:
```suggestion
    return C + B
```

stdlib/SparseArrays/src/linalg.jl
```diff
@@ -1,12 +1,44 @@
...
+end
+(+)(A::Hermitian{<:Any, <:SparseMatrixCSC}, B::Symmetric{<:Any, <:SparseMatricCSC}) = B + A
```
dkarrasch, line 39:
```suggestion
(+)(A::Hermitian{<:Any, <:SparseMatrixCSC}, B::Symmetric{<:Any, <:SparseMatrixCSC}) = B + A
```

stdlib/SparseArrays/src/linalg.jl
```diff
@@ -1,12 +1,44 @@
...
+        (+)(A::$HermSym{<:Any,<:AbstractSparseMatrix}, B::SparseMatrixCSC) = B + A
+        (+)(A::$HermSym, B::SparseMatrixCSC) = B + A
+
+        function +(A::$HermSym{<:Any, <:SparseMatrixCSC}, B::$HermSym{<:Any, <:SparseMatrixCSC})
+            if A.uplo == B.uplo
+                $HermSym(parent(A) + parent(B), sym_uplo(A.uplo))
+            elseif A.uplo == 'U'
+                $HermSym(parent(A) + permutedims(parent(B)), :U)
+            else
+                $HermSym(permutedims(parent(A)) + parent(B), :U)
```
dkarrasch, line 29: I think it's Julia style to have a `return` in each line, because we are not exiting the function in some unique final line.

dkarrasch (changes requested):
A couple of general remarks:
- In the sparse-sparse case, I don't think there is an efficient in-place addition, so I'd return simply `A+B`.

- It's a matter of taste, but personally I prefer to have clear `return` statements, to help the eye.

- I guess one could replace all occurrences of `SparseMatrixCSC` by `AbstractSparseMatrix`, because we distinguish here between dense and sparse, and don't care about memory layout or other specifics.

stdlib/SparseArrays/src/linalg.jl
```diff
@@ -1,12 +1,44 @@
...
 
 # In matrix-vector multiplication, the correct orientation of the vector is assumed.
 const StridedOrTriangularMatrix{T} = Union{StridedMatrix{T}, LowerTriangular{T}, UnitLowerTriangular{T}, UpperTriangular{T}, UnitUpperTriangular{T}}
 const AdjOrTransStridedOrTriangularMatrix{T} = Union{StridedOrTriangularMatrix{T},Adjoint{<:Any,<:StridedOrTriangularMatrix{T}},Transpose{<:Any,<:StridedOrTriangularMatrix{T}}}
 
+for HermSym ∈ [:Hermitian, :Symmetric]
+    @eval begin
+        function (+)(A::SparseMatrixCSC, B::$HermSym{<:Any,<:AbstractSparseMatrix})
+            C  = sparse(B)
+            C .= A .+ C
```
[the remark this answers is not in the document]
MasonProtter: Doing things in place seems to be more efficient:
```julia
julia> using SparseArrays

julia> A, B, C = sprand(10000, 10000, 0.05), sprand(10000, 10000, 0.05), spzeros(10000, 10000);

julia> @btime A + B;
  63.968 ms (8 allocations: 152.66 MiB)

julia> @btime C .= A .+ B;
  61.271 ms (2 allocations: 48 bytes)
```

Regarding explicit return statements, the majority of functions in this file don't use them, and personally I dislike the `return` keyword.

stdlib/SparseArrays/src/linalg.jl
```diff
@@ -1,12 +1,44 @@
...
+            elseif A.uplo == 'U'
+                $HermSym(parent(A) + permutedims(parent(B)), :U)
+            else
+                $HermSym(permutedims(parent(A)) + parent(B), :U)
+            end
+        end
+    end
+end
+
+function (+)(A::Symmetric{<:Any, <:SparseMatrixCSC}, B::Hermitian{<:Any, <:SparseMatricCSC})
```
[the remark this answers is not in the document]
MasonProtter: Now that's a facepalm worthy typo. Sorry about that.

stdlib/SparseArrays/src/linalg.jl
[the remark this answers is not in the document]
dkarrasch: Ok, I wasn't sure if there really is an "in-place" implementation. The 48 bytes don't look reasonable to me: the vector of values of `C` initially doesn't have any entries, but needs to store all entries of `A .+ B` after the addition. And the difference in runtime (given the large matrix size and the many nonzero entries) seems negligible to me. But I don't know for sure. We'll get some advice later, probably @mbauman or @andreasnoack can help with this issue, currently it doesn't affect functionality, so we can start testing this in any case.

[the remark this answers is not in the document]
mbauman: It's because BenchmarkTools repeatedly assigns into the _same_ `C`, so you're not really measuring the time-from-empty.  Also should be using `@btime $A + $B` for more accurate answers.  You'll need to be more careful about the problem `setup` in order to get a good grasp here.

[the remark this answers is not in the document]
MasonProtter: Right, my bad (I swear I'm usually more careful about benchmarking!):
```julia
julia> @benchmark Cz .= $A .+ $B setup=(Cz = spzeros(10000, 10000))
BenchmarkTools.Trial: 
  memory estimate:  152.54 MiB
  allocs estimate:  2
  --------------
  minimum time:     65.595 ms (0.00% GC)
  median time:      94.922 ms (0.00% GC)
  mean time:        96.968 ms (2.74% GC)
  maximum time:     117.400 ms (17.02% GC)
  --------------
  samples:          50
  evals/sample:     1

julia> @benchmark $A .+ $B
BenchmarkTools.Trial: 
  memory estimate:  152.62 MiB
  allocs estimate:  8
  --------------
  minimum time:     78.753 ms (0.00% GC)
  median time:      91.994 ms (0.00% GC)
  mean time:        98.202 ms (5.81% GC)
  maximum time:     112.802 ms (15.47% GC)
  --------------
  samples:          51
  evals/sample:     1

```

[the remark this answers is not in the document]
MasonProtter: Apologies for the poor benchmarks, I think this is more representative:
```julia
julia> @benchmark B .= A .+ B setup=((A = sprandn(1000, 1000, 0.05); B = sprandn(1000, 1000, 0.05))) seconds=10
BenchmarkTools.Trial: 
  memory estimate:  1.91 MiB
  allocs estimate:  8
  --------------
  minimum time:     689.281 μs (0.00% GC)
  median time:      746.601 μs (0.00% GC)
  mean time:        765.545 μs (2.04% GC)
  maximum time:     2.161 ms (60.23% GC)
  --------------
  samples:          3951
  evals/sample:     1

julia> @benchmark A + B setup=((A = sprandn(1000, 1000, 0.05); B = sprandn(1000, 1000, 0.05))) seconds=10
BenchmarkTools.Trial: 
  memory estimate:  1.52 MiB
  allocs estimate:  7
  --------------
  minimum time:     587.142 μs (0.00% GC)
  median time:      596.713 μs (0.00% GC)
  mean time:        609.592 μs (1.45% GC)
  maximum time:     1.551 ms (58.68% GC)
  --------------
  samples:          4266
  evals/sample:     1
```
which suggests that @dkarrasch's suggestion is the way to go.

stdlib/SparseArrays/src/linalg.jl
dkarrasch: This function causes an ambiguity error with

https://github.com/JuliaLang/julia/blob/0c284839fef6c8c153edc01fddfa37a9f5ac6752/stdlib/LinearAlgebra/src/symmetric.jl#L465-L466

so I guess we should tweak the version there to get the desired behavior. In fact, I've played with it a little bit, and making the above code the generic one seems greatly beneficial:
```julia
using SparseArrays, LinearAlgebra, Test
import Base: +
import LinearAlgebra: sym_uplo
using BenchmarkTools

function myplus(A::Hermitian, B::Hermitian)
    if A.uplo == B.uplo
        Hermitian(parent(A) + parent(B), sym_uplo(A.uplo))
    elseif A.uplo == 'U'
        Hermitian(parent(A) + permutedims(parent(B)), :U)
    else
        Hermitian(permutedims(parent(A)) + parent(B), :U)
    end
end

A = sprandn(100, 100, 0.1); B = sprandn(100, 100, 0.1);

@btime $(Hermitian(A)) + $(Hermitian(B));
@btime myplus($(Hermitian(A)), $(Hermitian(B)));
@test Hermitian(A) + Hermitian(B) ≈ myplus(Hermitian(A), Hermitian(B))
Ad, Bd = Matrix.((A, B));
@btime $(Hermitian(Ad)) + $(Hermitian(Bd));
@btime myplus($(Hermitian(Ad)), $(Hermitian(Bd)));
@test Hermitian(Ad) + Hermitian(Bd) ≈ myplus(Hermitian(Ad), Hermitian(Bd))
```
I think this is because we're working with the underlying array directly, instead of calling `getindex` on a `Hermitian` many times, as happens in the above-quoted code.

stdlib/SparseArrays/src/linalg.jl
```diff
@@ -1,12 +1,35 @@
...
+
+        function +(A::$Wrapper{<:Any, <:SparseMatrixCSC}, B::$Wrapper{<:Any, <:SparseMatrixCSC})
+            if A.uplo == B.uplo
+                $Wrapper(parent(A) + parent(B), sym_uplo(A.uplo))
+            elseif A.uplo == 'U'
+                $Wrapper(parent(A) + $operation(parent(B)), :U)
+            else
+                $Wrapper($operation(parent(A)) + parent(B), :U)
+            end
+        end
```
dkarrasch, line 26: As I said [here](https://github.com/JuliaLang/julia/pull/35325#discussion_r407100805), this should go to LinearAlgebra/src/symmetric.jl around lines 465, but without the `<:SparseMatrixCSC` restriction. First, you need to do that anyway to avoid the ambiguity, and second, this approach here is even beneficial for dense matrices, compared to the code there.

stdlib/SparseArrays/src/linalg.jl
```diff
@@ -1,12 +1,35 @@
...
 using Random: rand!
 
 # In matrix-vector multiplication, the correct orientation of the vector is assumed.
 const StridedOrTriangularMatrix{T} = Union{StridedMatrix{T}, LowerTriangular{T}, UnitLowerTriangular{T}, UpperTriangular{T}, UnitUpperTriangular{T}}
 const AdjOrTransStridedOrTriangularMatrix{T} = Union{StridedOrTriangularMatrix{T},Adjoint{<:Any,<:StridedOrTriangularMatrix{T}},Transpose{<:Any,<:StridedOrTriangularMatrix{T}}}
 
+for (Wrapper, operation) ∈ [(:Hermitian, :adjoint), (:Symmetric, :transpose)]
+    @eval begin
+        (+)(A::SparseMatrixCSC, B::$Wrapper{<:Any,<:AbstractSparseMatrix}) = A + sparse(B)
+        (+)(A::$Wrapper{<:Any,<:AbstractSparseMatrix}, B::SparseMatrixCSC) = B + A
```
dkarrasch, line 13:
```suggestion
        (+)(A::$Wrapper{<:Any,<:AbstractSparseMatrix}, B::SparseMatrixCSC) = sparse(A) + B
```

stdlib/SparseArrays/src/linalg.jl
```diff
@@ -1,12 +1,35 @@
...
+
+        (+)(A::SparseMatrixCSC, B::$Wrapper) = A + collect(B)
+        (+)(A::$Wrapper, B::SparseMatrixCSC) = B + A
```
dkarrasch, line 16:
```suggestion
        (+)(A::$Wrapper, B::SparseMatrixCSC) = collect(A) + B
```

stdlib/SparseArrays/src/linalg.jl
dkarrasch: BTW, the (partially) real versions there,

https://github.com/JuliaLang/julia/blob/0c284839fef6c8c153edc01fddfa37a9f5ac6752/stdlib/LinearAlgebra/src/symmetric.jl#L467-L468

would also benefit from this more careful treatment.

stdlib/SparseArrays/src/linalg.jl
```diff
@@ -1,12 +1,35 @@
...
+                $Wrapper(parent(A) + parent(B), sym_uplo(A.uplo))
+            elseif A.uplo == 'U'
+                $Wrapper(parent(A) + $operation(parent(B)), :U)
+            else
+                $Wrapper($operation(parent(A)) + parent(B), :U)
+            end
+        end
+    end
+end
+(+)(A::Symmetric{<:Any, <:SparseMatrixCSC}, B::Hermitian{<:Any, <:SparseMatrixCSC}) = sparse(A) + B
```
dkarrasch, line 29: I'm not sure, but this boils down to
```suggestion
(+)(A::Symmetric{<:Any, <:SparseMatrixCSC}, B::Hermitian{<:Any, <:SparseMatrixCSC}) = sparse(A) + sparse(B)
```
anyway, so we might as well write it out here, given one-liner. Same below.

stdlib/SparseArrays/test/sparse.jl
```diff
@@ -2906,4 +2906,18 @@ end
     @test B ≈ mapreduce(identity, +, Matrix(A), dims=2)
 end
 
+@testset "Symmetric and Hermitian" begin
+    A = sprandn(ComplexF64, 10, 10, 0.1)
+    B = sprandn(ComplexF64, 10, 10, 0.1)
+    @test Hermitian(A + A') + B isa SparseMatrixCSC
```
dkarrasch, line 2912: I'd suggest to explicitly _not_ make the underlying array Hermitian, so better have
```suggestion
    @test Hermitian(A) + B isa SparseMatrixCSC
```
and then also have `:U` and `:L` versions. We will also need tests for the numerical values, so something like
```julia
@test Matrix(Hermitian(A, uplo)) + Matrix(B)
```

dkarrasch (changes requested):
Sorry for my annoying persistence. Besides moving some part of the code, we will also need versions for `-` of these signatures.

stdlib/SparseArrays/src/linalg.jl
MasonProtter: Hm, I'm now not really seeing the advantage of that big `if ... elseif ... else` block over the way it's done here in `LinearAlgebra/src/symmetric.jl`. I think as long as I properly handle `::Sparse + ::Hermitian` and such, the implementations of `symmetric.jl` seem like they should do the right thing.

dkarrasch: Did you try my benchmarks here https://github.com/JuliaLang/julia/pull/35325#discussion_r407100805? The advantage is a significant speedup.

dkarrasch: Oh, and the other advantage is that it actually works... and doesn't lead to ambiguity errors.

dkarrasch: One more: I thought we're discussing here the `Symmetric + Symmetric` and `Hermitian + Hermitian` case. With the most generic methods given in `LinearAlgebra`, I don't see how you avoid the method ambiguity.

MasonProtter: Right, sorry about that I see now.

MasonProtter: I really appreciate your patient and persistent help with this PR by the way!

MasonProtter:
Not exactly sure what's wrong with the CI, looks like some sort of network error? I'll just wait a bit and try to retrigger it.

stdlib/SparseArrays/src/linalg.jl
```diff
@@ -1,12 +1,32 @@
...
 
-import LinearAlgebra: checksquare
+import LinearAlgebra: checksquare, sym_uplo
 using Random: rand!
 
 # In matrix-vector multiplication, the correct orientation of the vector is assumed.
 const StridedOrTriangularMatrix{T} = Union{StridedMatrix{T}, LowerTriangular{T}, UnitLowerTriangular{T}, UpperTriangular{T}, UnitUpperTriangular{T}}
 const AdjOrTransStridedOrTriangularMatrix{T} = Union{StridedOrTriangularMatrix{T},Adjoint{<:Any,<:StridedOrTriangularMatrix{T}},Transpose{<:Any,<:StridedOrTriangularMatrix{T}}}
 
+for (Wrapper, conjugation) ∈ [(:Hermitian, :adjoint), (:Symmetric, :transpose)]
```
dkarrasch, line 10:
```suggestion
for Wrapper ∈ (:Hermitian, :Symmetric)
```
Seems like you don't need the `conjugation` here...

stdlib/LinearAlgebra/src/symmetric.jl
```diff
@@ -461,11 +461,25 @@ end
...
+end
+
 for f in (:+, :-)
-    @eval $f(A::Symmetric, B::Symmetric) = Symmetric($f(A.data, B), sym_uplo(A.uplo))
-    @eval $f(A::Hermitian, B::Hermitian) = Hermitian($f(A.data, B), sym_uplo(A.uplo))
-    @eval $f(A::Hermitian, B::Symmetric{<:Real}) = Hermitian($f(A.data, B), sym_uplo(A.uplo))
-    @eval $f(A::Symmetric{<:Real}, B::Hermitian) = Hermitian($f(A.data, B), sym_uplo(A.uplo))
+    @eval begin
+        $f(A::Hermitian, B::Symmetric{<:Real}) = $f(A, Hermitian(B, sym_uplo(B.uplo)))
+        $f(A::Symmetric{<:Real}, B::Hermitian) = $f(Hermitian(A, sym_uplo(A.uplo)), B)
```
dkarrasch, line 481:
```suggestion
        $f(A::Hermitian, B::Symmetric{<:Real}) = $f(A, Hermitian(B.data, sym_uplo(B.uplo)))
        $f(A::Symmetric{<:Real}, B::Hermitian) = $f(Hermitian(A.data, sym_uplo(A.uplo)), B)
```
I'd suggest to "re-wrap" here instead of adding another layer.

stdlib/SparseArrays/src/linalg.jl
```diff
@@ -1,12 +1,32 @@
...
+    end
+end
+for op ∈ (:+, :-)
+    @eval begin
+        $op(A::Symmetric{<:Any, <:AbstractSparseMatrix}, B::Hermitian{<:Any, <:AbstractSparseMatrix}) = $op(sparse(A), sparse(B))
+        $op(A::Hermitian{<:Any, <:AbstractSparseMatrix}, B::Symmetric{<:Any, <:AbstractSparseMatrix}) = $op(sparse(A), sparse(B))
+        $op(A::Symmetric{<:Real, <:AbstractSparseMatrix}, B::Hermitian{<:Any, <:AbstractSparseMatrix}) = $op(Hermitian(parent(A), sym_uplo(A.uplo)), B)
+        $op(A::Hermitian{<:Any, <:AbstractSparseMatrix}, B::Symmetric{<:Real, <:AbstractSparseMatrix}) = $op(A, Hermitian(parent(B), sym_uplo(B.uplo)))
+    end
+end
```
dkarrasch, line 28: I'm not exactly sure, but aren't these already handled by the generic `LinearAlgebra` methods above? I thought that those automatically generate sparse matrices in case both `parent`s are sparse, but the longer I look at these methods, the more they all look alike. 🤣

stdlib/SparseArrays/test/sparse.jl
```diff
@@ -2906,4 +2906,33 @@ end
     @test B ≈ mapreduce(identity, +, Matrix(A), dims=2)
 end
 
+@testset "Symmetric and Hermitian #35325" begin
+    A = sprandn(ComplexF64, 10, 10, 0.1)
+    B = sprandn(ComplexF64, 10, 10, 0.1)
+
+    @test Symmetric(real(A)) + Hermitian(B) isa Hermitian
+    @test Hermitian(A) + Symmetric(real(B)) isa Hermitian
+    @testset "$Wrapper $op" for op ∈ (+, -), (Wrapper, conjugation) ∈ ((Hermitian, adjoint), (Symmetric, transpose))
```
dkarrasch, line 2915: The `conjugation` seems obsolete here.

dkarrasch (changes requested):
This is shaping up nicely, very clear code. I'm not sure you need the "sparse mixed-type" methods, maybe you could try without those, since you already have tests like `isa SparseMatrixCSC`.

stdlib/LinearAlgebra/src/symmetric.jl
MasonProtter: Oops, good catch.

stdlib/SparseArrays/src/linalg.jl
MasonProtter: Yep, that's left over from before, thanks.

stdlib/SparseArrays/test/sparse.jl
MasonProtter: Correct.

stdlib/SparseArrays/src/linalg.jl
MasonProtter: Unfortunately, the methods in LinearAlgebra.jl end up causing 
```julia
::Symmetric{<:Any, <:Sparse} + ::Hermitian{<:Any, <:Sparse}
```
to be dense (not exactly sure why), so I special cased them, and then since I special cased them, that causes a method ambiguity so that then I also need to specify what to do with 
```julia
::Symmetric{<:Real, <:Sparse} + ::Hermitian{<:Real, <:Sparse}
```

dkarrasch: Okay, I think that's because the `Any-Any` case is not specifically handled by the above `LinearAlgebra` methods, only the `Symmetric{<:Real}` case. Then they fall back to something very generic. So I think you should be fine without the last two lines in this block?

MasonProtter: No, because there's an ambiguity between
```julia
 f(::Symmetric{<:Real, <:Any}) = ...
```
and
```julia
f(::Symmetric{<:Any, <:AbstractSparseMatrix}) = ... 
```
because they constraint the parameters in different ways. e.g.:
```julia
julia> using SparseArrays, LinearAlgebra; A = Symmetric(sprandn(10, 10, 0.1));

julia> f(::Symmetric{<:Real, <:Any}) = 1

julia> f(::Symmetric{<:Any, <:AbstractSparseMatrix}) = 2

julia> f(A)
ERROR: MethodError: f(::Symmetric{Float64,SparseMatrixCSC{Float64,Int64}}) is ambiguous. Candidates:
  f(::Symmetric{#s24,#s25} where #s25 where #s24<:Real) in Main at REPL[15]:1
  f(::Symmetric{#s26,#s27} where #s27<:(AbstractSparseArray{Tv,Ti,2} where Ti where Tv) where #s26) in Main at REPL[16]:1
Possible fix, define
  f(::Symmetric{#s26,#s27} where #s27<:(AbstractSparseArray{Tv,Ti,2} where Ti where Tv) where #s26<:Real)
Stacktrace:
 [1] top-level scope at REPL[17]:1
 [2] eval(::Module, ::Any) at ./boot.jl:331
 [3] eval_user_input(::Any, ::REPL.REPLBackend) at /home/mason/julia/usr/share/julia/stdlib/v1.4/REPL/src/REPL.jl:86
 [4] run_backend(::REPL.REPLBackend) at /home/mason/.julia/packages/Revise/C272c/src/Revise.jl:1075
 [5] top-level scope at none:0
```

dkarrasch: Thanks for the explanation, and sorry for the noise. 👼

stdlib/SparseArrays/test/sparse.jl
```diff
@@ -2906,4 +2906,34 @@ end
...
+        AWL = Wrapper(A, :L)
+        BWU = Wrapper(B, :U)
+        BWL = Wrapper(B, :L)
+
+        @test op(AWU, B) isa SparseMatrixCSC
+        @test op(A, BWL) isa SparseMatrixCSC
+
+        @test op(AWU, B) ≈ op(collect(AWU), B)
+        @test op(AWL, B) ≈ op(collect(AWL), B)
+        @test op(A, Wrapper(B, :U)) ≈ op(A, collect(BWU))
```
dkarrasch, line 2927:
```suggestion
        @test op(A, BWU) ≈ op(A, collect(BWU))
```
or am I missing something? It has no actual effect, but since you constructed the wrappers above anyway...

dkarrasch (approved):
Just a minor suggestion to re-use the already constructed matrix wrappers. Otherwise, this LGTM. I'm not up-to-date w.r.t. the "current" test failures on master. What I saw seemed unrelated.

stdlib/SparseArrays/test/sparse.jl
MasonProtter: Yep, my bad!

dkarrasch:
Tests are green, all is good, I'd say.

MasonProtter:
Thanks Daniel, I really appreciate your patience and persistence in helping me with this PR!

PR #35326, opened 2020-03-31
propagate errors from `flush` within `close`

JeffBezanson:
Alternative to #35303.

### DIFF
--- a/src/support/ios.c
+++ b/src/support/ios.c
@@ -648,17 +648,21 @@ int ios_flush(ios_t *s)
     return 0;
 }
 
-void ios_close(ios_t *s)
+int ios_close(ios_t *s)
 {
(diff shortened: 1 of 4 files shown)


src/support/ios.c
```diff
@@ -648,9 +648,9 @@ int ios_flush(ios_t *s)
     return 0;
 }
 
-void ios_close(ios_t *s)
+int ios_close(ios_t *s)
 {
-    ios_flush(s);
+    int err = ios_flush(s);
     if (s->fd != -1 && s->ownfd)
         close(s->fd);
```
Keno, line 655: check the return value of `close` also. Some file systems don't sync the buffers to disk until the fd is closed even with a preceding flush.

PR #35329, opened 2020-04-01
fix latest LLVM 11 compatibility issues

yhls:
- `CompositeType` class has been removed
- `FileLineInfoSpecifier::Default` has been renamed to `RawValue`

@vchuravy

### DIFF
--- a/src/cgutils.cpp
+++ b/src/cgutils.cpp
@@ -44,16 +44,18 @@ static Function *function_proto(Function *F, Module *M = nullptr)
 }
 
 #define prepare_call(Callee) prepare_call_in(jl_Module, (Callee))
-static Value *prepare_call_in(Module *M, Value *Callee)
+static FunctionCallee prepare_call_in(Module *M, Value *Callee)
 {
     if (Function *F = dyn_cast<Function>(Callee)) {
         GlobalValue *local = M->getNamedValue(Callee->getName());
         if (!local) {
             local = function_proto(F, M);
         }
-        return local;
+        Callee = local;
     }
-    return Callee;
+    FunctionType *FnTy = cast<FunctionType>(
+        Callee->getType()->getPointerElementType());
+    return {FnTy, Callee};
 }
 
 // Take an arbitrary untracked value and make it gc-tracked
@@ -1341,9 +1343,16 @@ static void emit_write_multibarrier(jl_codectx_t&, Value*, Value*);
 
 std::vector<unsigned> first_ptr(Type *T)
 {
-    if (isa<CompositeType>(T)) {
-        if (!isa<StructType>(T) && cast<SequentialType>(T)->getNumElements() == 0)
-            return {};
+    if (isa<StructType>(T) || isa<ArrayType>(T) || isa<VectorType>(T)) {
+        if (!isa<StructType>(T)) {
+            uint64_t num_elements;
+            if (auto *AT = dyn_cast<ArrayType>(T))
+                num_elements = AT->getNumElements();
+            else
+                num_elements = cast<VectorType>(T)->getNumElements();
+            if (num_elements == 0)
+                return {};
+        }
         unsigned i = 0;
         for (Type *ElTy : T->subtypes()) {
             if (isa<PointerType>(ElTy) && ElTy->getPointerAddressSpace() == AddressSpace::Tracked) {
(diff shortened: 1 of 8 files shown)


yhls:
- `SequentialType` class has been removed
- `CreateMemCpy` now takes `MaybeAlign`s instead of `unsigned`s

src/llvm-late-gc-lowering.cpp
```diff
@@ -388,22 +388,29 @@ CountTrackedPointers::CountTrackedPointers(Type *T) {
...
+            count *= cast<VectorType>(T)->getNumElements();
     }
     if (count == 0)
         all = false;
 }
 
 unsigned getCompositeNumElements(Type *T) {
-    return isa<StructType>(T) ? T->getStructNumElements() : cast<SequentialType>(T)->getNumElements();
+    if (isa<StructType>(T))
+        return T->getStructNumElements();
```
vtjnash, line 409:
```suggestion
        return T->getNumElements();
```

vtjnash (commented):
```
/Users/julia/buildbot/worker/package_macos64/build/src/llvm-late-gc-lowering.cpp:414:1: warning: control may reach end of non-void function [-Wreturn-type]
}
^
1 warning generated.
In file included from /Users/julia/buildbot/worker/package_macos64/build/src/codegen.cpp:853:
./cgutils.cpp:1349:28: warning: variable 'num_elements' is used uninitialized whenever 'if' condition is false [-Wsometimes-uninitialized]
            else if (auto *VT = dyn_cast<VectorType>(T))
                           ^~
./cgutils.cpp:1351:17: note: uninitialized use occurs here
            if (num_elements == 0)
                ^~~~~~~~~~~~
./cgutils.cpp:1349:18: note: remove the 'if' if its condition is always true
            else if (auto *VT = dyn_cast<VectorType>(T))
                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
./cgutils.cpp:1346:34: note: initialize the variable 'num_elements' to silence this warning
            uint64_t num_elements;
                                 ^
                                  = 0
1 warning generated.
```
Need to fix the generated warnings, but otherwise lgtm

src/llvm-late-gc-lowering.cpp
yhls: To use `getNumElements`, I'd have to explicitly cast `T` to `StructType` like in the following cases--do you suggest that I do this?

yhls:
I've fixed the warnings, and addressed the following change in LLVM

- deprecated overloads of `IRBuilder::CreateCall` and `CallInst::Create` have been removed

by making `prepare_call_in` in cgutils.cpp return a `FunctionCallee` instead of a `Value*`.

yhls:
- `CallSite` class has been removed

vchuravy (commented):
> To use getNumElements, I'd have to explicitly cast T to StructType like in the following cases--do you suggest that I do this?

Probably want to use a `dyn_cast` then.

src/cgutils.cpp
```diff
@@ -1341,9 +1343,16 @@ static void emit_write_multibarrier(jl_codectx_t&, Value*, Value*);
 
 std::vector<unsigned> first_ptr(Type *T)
 {
-    if (isa<CompositeType>(T)) {
```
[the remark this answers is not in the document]
yhls: I don't think there's any way to construct a `CompositeType`, so it's effectively abstract, right?

yhls:
> > To use getNumElements, I'd have to explicitly cast T to StructType like in the following cases--do you suggest that I do this?
> 
> Probably want to use a `dyn_cast` then.

I meant to say that using `getStructNumElements` allows us to avoid having to cast explicitly, as opposed to using `getNumElements` with a `dyn_cast`, so is probably preferable. But I'm pretty indifferent about one over the other here.

vchuravy:
> I meant to say that using getStructNumElements allows us to avoid having to cast explicitly, as opposed to using getNumElements with a dyn_cast, so is probably preferable. But I'm pretty indifferent about one over the other here.

It looks like LLVM is removing all the convenience getters, and it is cheaper to do the `dyn_cast` then to do an `isa` check and then do the `cast` inside `getStructNumElements` (which will do another `isa` check.)

yhls:
Ah, ok, changed to use `getNumElements`.

src/llvm-late-gc-lowering.cpp
yhls: Changed to use `getNumElements`.

PR #35332, opened 2020-04-01
Rename `arm-linux` to `armv7l-linux` on many leftover checksums

staticfloat:
In the switch to `armv7l` instead of `arm` in our triplet, we forgot to rename these checksums.  Now that I have an armv7l builder back online, let's make sure we can actually build it.  ;)


staticfloat:
We've silenced the armv7l buildbot from reporting here (because it's in such a bad state) but the build triggered by this commit is here: https://build.julialang.org/#/builders/5/builds/10281

All dependencies downloaded properly, so this is good to merge.

PR #35333, opened 2020-04-01
make Diagonal+Symmetric return Symmetric

### LINKED ISSUE #708: Diagonal + Symmetric should be Symmetric not Matrix
Right now we have
```julia
julia> Symmetric(ones(3,3)) + Diagonal(ones(3))
3×3 Array{Float64,2}:
 2.0  1.0  1.0
 1.0  2.0  1.0
 1.0  1.0  2.0
```
but the problem should probably be `Symmetric`. Likewise for `Hermitian + Diagonal{<:Real}`.

### LINKED ISSUE #35307: Diagonal + Symmetric should be Symmetric not Matrix
Right now we have
```julia
julia> Symmetric(ones(3,3)) + Diagonal(ones(3))
3×3 Array{Float64,2}:
 2.0  1.0  1.0
 1.0  2.0  1.0
 1.0  1.0  2.0
```
but the problem should probably be `Symmetric`. Likewise for `Hermitian + Diagonal{<:Real}`.

dkarrasch:
Closes JuliaLang/LinearAlgebra.jl#708.

### DIFF
--- a/stdlib/LinearAlgebra/src/diagonal.jl
+++ b/stdlib/LinearAlgebra/src/diagonal.jl
@@ -155,30 +155,32 @@ end
 (-)(A::Diagonal) = Diagonal(-A.diag)
 (+)(Da::Diagonal, Db::Diagonal) = Diagonal(Da.diag + Db.diag)
 (-)(Da::Diagonal, Db::Diagonal) = Diagonal(Da.diag - Db.diag)
+
+for f in (:+, :-)
+    @eval function $f(A::AbstractMatrix, D::Diagonal)
+        d = D.diag
+        DA = copy_oftype(A, promote_op($f, eltype(A), eltype(D)))
+        dinds = diagind(DA)
+        for (di, ai) in enumerate(dinds)
+            DA[ai] = $f(DA[ai], d[di])
+        end
+        return DA
+    end
+end
+(+)(D::Diagonal, A::AbstractMatrix) = +(A, D)
+
 for f in (:+, :-)
     @eval function $f(D::Diagonal, S::Symmetric)
-        SD = copy_oftype(S.data, promote_op($f, eltype(D), eltype(S)))
-        di = diagind(SD)
-        SD[di] .= ($f).(D.diag, SD[di])
-        return Symmetric(SD, sym_uplo(S.uplo))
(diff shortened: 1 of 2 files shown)


stdlib/LinearAlgebra/src/diagonal.jl
```diff
@@ -680,19 +706,6 @@ function getproperty(C::Cholesky{<:Any,<:Diagonal}, d::Symbol)
 end
 
 Base._sum(A::Diagonal, ::Colon) = sum(A.diag)
-function Base._sum(A::Diagonal, dims::Integer)
```
andreasnoack, line 683: Is this deletion related?

dkarrasch: No, that must have been my cat on the keyboard. I have no idea...

stdlib/LinearAlgebra/src/diagonal.jl
```diff
@@ -155,6 +155,32 @@ end
 (-)(A::Diagonal) = Diagonal(-A.diag)
 (+)(Da::Diagonal, Db::Diagonal) = Diagonal(Da.diag + Db.diag)
 (-)(Da::Diagonal, Db::Diagonal) = Diagonal(Da.diag - Db.diag)
+for f in (:+, :-)
+    @eval function $f(D::Diagonal, S::Symmetric)
+        SD = copy_oftype(S.data, promote_op($f, eltype(D), eltype(S)))
+        di = diagind(SD)
+        SD[di] .= ($f).(D.diag, SD[di])
```
andreasnoack, line 162: Is it worth avoiding the allocation associated with `SD[di]`?

dkarrasch:
I thought I'll introduce a new methods for of type `[+/-](A::AbstractMatrix, D::Diagonal)` etc. where I simply copy `A`, and add/subtract `D` to/from the diagonal. This speeds up addition and subtraction a little bit. The Symmetric/Hermitian case then simply falls back to that, and wraps the result appropriately.

dkarrasch:
Ouch, that failed. I'll remove the new generic methods, and leave `+(S.data, D)` to the ecosystem to handle.

PR #31487, opened 2019-03-26
Support `eps` on `TimeType` types

omus:
Now works similarly to how `eps` works with floating point.

PR #35334, opened 2020-04-01
Improve `eps` documentation for `TimeType`s

omus:
Also now supports calling `eps` with `TimeType` types.


omus:
Closing in favor of: https://github.com/JuliaLang/julia/pull/31487

PR #35335, opened 2020-04-01
fix detection of srets containing tracked pointers in GC lowering

JeffBezanson:
For #34126


src/llvm-late-gc-lowering.cpp
```diff
@@ -1352,7 +1352,7 @@ State LateLowerGCFrame::LocalScan(Function &F) {
                 if (CI->hasStructRetAttr()) {
                     AllocaInst *SRet = dyn_cast<AllocaInst>((CI->arg_begin()[0])->stripInBoundsOffsets());
                     if (SRet) {
```
vtjnash, line 1354: This should perhaps be an assert. There's still lingering problems with this design, but we can keep it for now.

vtjnash (approved):

Keno:
For fixes in the LLVM pass, please also add an LLVM IR level test.

Issue #34726, opened 2020-02-11
ENV is not thread safe with glibc


c42f:
For some reason I was reading our `ENV` code, and I noticed we don't have a lock around our calls to `getenv` and `setenv` on linux, or for iterating the environment via the `environ` variable.

This makes `ENV` unsafe for use with multithreaded julia, as `getenv` and `setenv` are not mutually threadsafe in glibc. (See https://github.com/bminor/glibc/blob/master/stdlib/getenv.c and https://github.com/bminor/glibc/blob/master/stdlib/setenv.c. Unsurprisingly `getenv` is safe by itself, but mysteriously `setenv` is protected by a lock which `getenv` ignores! So you can mutually `setenv` safely, but nobody can presume to `getenv` elsewhere in the same multithreaded code.)

We can easily add some more locks on our side as mitigation, but unfortunately we can't really fix `setenv` without fixing glibc; a random C library we link against may decide to call `getenv` at any point. This definitely happens in practice and I've experienced it personally: https://github.com/xianyi/OpenBLAS/issues/716#issuecomment-164339663. There's a nice discussion on the rust issue tracker including this gem: https://github.com/rust-lang/rust/pull/24741#issuecomment-99860102.

Some options:
1. Mitigation: Put our own lock around `getenv/setenv/environ` access, and just hope that no C library we link against calls these functions itself. Add a big warning to the docs. This is the current rust approach though it's fragile.
2. Avoidance: Ban using libc `setenv` entirely; create a shadow environment which is a copy of the system environment. This is the C# approach but with Julia's tradition of calling into C libraries I doubt that will work out (it creates surprises even in C#; see https://yizhang82.dev/set-environment-variable)
3. Fix glibc: Every portable language runtime has this exact problem, so people have tried fixing this in the past. In older times they were met with [surprising hostility](https://sourceware.org/bugzilla/show_bug.cgi?id=5069#c11). The seemingly-current bug is marked suspended https://sourceware.org/bugzilla/show_bug.cgi?id=15607#c2 whatever that means.

For now the only easy / possible thing to do is to clearly option (1): add some locking and a big warning. If glibc could be fixed it could morph into a long term plan. More recent bugs suggest the glibc maintainers may possibly accept a patch.

[...]

Issue #22949, opened 2017-07-25
tempnam() should be avoided in the unix version of tempname() function


sambitdash:
tempnam() is marked obsolete in the opengroup spec as well. See the link for Issue 7.

http://pubs.opengroup.org/onlinepubs/9699919799/functions/tempnam.html

tkelman:
I think we should deprecate the Julia `tempname` completely, it's got a lot of problems (xref https://github.com/JuliaLang/julia/issues/9053 as well)

nalimilan:
Right, `tempname` isn't safe due to race conditions with potential attackers. Better make this clear by requiring people to use `randstring` explicitly when they really can't use `mktemp`.

Let's close this in favor of #9053, where the discussion has already happened?

laborg:
The use of `tempnam` has been removed in https://github.com/JuliaLang/julia/pull/38879/files

PR #35342, opened 2020-04-02
Unset TMPDIR before calling tempnam in tempname

ivan-timokhin:
`tempnam` consults the `TMPDIR` environment variable before its argument (see [man page](https://linux.die.net/man/3/tempnam)), so if it is set the function will suddenly and quietly ignore the `parent` argument.

### DIFF
--- a/base/file.jl
+++ b/base/file.jl
@@ -547,7 +547,9 @@ else # !windows
 # Obtain a temporary filename.
 function tempname(parent::AbstractString=tempdir(); cleanup::Bool=true)
     isdir(parent) || throw(ArgumentError("$(repr(parent)) is not a directory"))
-    p = ccall(:tempnam, Cstring, (Cstring, Cstring), parent, temp_prefix)
+    p = withenv("TMPDIR" => nothing) do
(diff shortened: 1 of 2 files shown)


KristofferC:
Ref https://github.com/JuliaLang/julia/issues/34726

ivan-timokhin:
> Ref #34726

Ugh. Fair enough. The problem is that the current conclusion for this case (C function depending on the environment) in the linked thread is ["don't do that, use the library's configuration API"](https://github.com/JuliaLang/julia/issues/34726#issuecomment-58712445), and the whole issue here is that `tmpnam` consults `TMPDIR` *before* its explicit parameters. So I'm not sure what would a better fix look like (aside from not using `tmpnam` at all—#22949).

If this approach is deemed unacceptable, I have no objections to closing this PR; I'll open an issue instead.

ivan-timokhin:
I have thought about this some more, and convinced myself this is not going to work—as written, `tempname` would break any concurrent calls to `tempdir`, which is probably worse than the original problem. Mea culpa.

PR #35347, opened 2020-04-02
avoid passing pointers for known Type arguments

JeffBezanson:
As noticed in #34993.
Before:
```
julia> code_llvm(*, (Float16,Float16))

;  @ float.jl:398 within `*'
define i16 @"julia_*_66"(i16, i16) {
top:
  %2 = call float @j_Float32_67(%jl_value_t addrspace(10)* addrspacecast (%jl_value_t* inttoptr (i64 140254228698048 to %jl_value_t*) to %jl_value_t addrspace(10)*), i16 %0)
  %3 = call float @j_Float32_68(%jl_value_t addrspace(10)* addrspacecast (%jl_value_t* inttoptr (i64 140254228698048 to %jl_value_t*) to %jl_value_t addrspace(10)*), i16 %1)
```
After:
```
julia> code_llvm(*, (Float16,Float16))

;  @ float.jl:398 within `*'
define i16 @"julia_*_70"(i16, i16) {
top:
  %2 = call float @j_Float32_71(i16 %0)
  %3 = call float @j_Float32_72(i16 %1)
```

### DIFF
--- a/src/codegen.cpp
+++ b/src/codegen.cpp
@@ -239,6 +239,36 @@ static bool type_is_ghost(Type *ty)
     return (ty == T_void || ty->isEmptyTy());
 }
 
+// should agree with `Core.Compiler.hasuniquerep`
+static bool type_has_unique_rep(jl_value_t *t)
+{
+    if (t == (jl_value_t*)jl_typeofbottom_type)
+        return false;
+    if (t == jl_bottom_type)
+        return true;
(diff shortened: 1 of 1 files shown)


vtjnash (approved):
Should we add a comment that this can be kept in sync with `Core.Compiler.hasuniquerep`?

PR #35348, opened 2020-04-03
Reunify code for Base.include and MainInclude.include

c42f:
These are only duplicated for better stack traces, and are otherwise meant to be the same. However the code here has already started to diverge in several minor but confusing ways (#34595 etc) which is making a larger refactor related to https://github.com/JuliaLang/julia/pull/35243 more fiddly.

Factor the include bodies into a macro to improve this.

This also allows the bulk of the code to go in loading.jl which I find a little more natural.

### DIFF
--- a/base/Base.jl
+++ b/base/Base.jl
@@ -363,32 +363,9 @@ end
 for m in methods(include)
     delete_method(m)
 end
-# These functions are duplicated in client.jl/include(::String) for
-# nicer stacktraces. Modifications here have to be backported there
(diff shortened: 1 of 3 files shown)


vtjnash:
I thought this should have the same backtraces as if it were actually just a function? But we appear to already have added `mapexpr` concept there now too, so that seems like it already ruined the concept there.

c42f:
Ah that's embarrassing. I forgot macros have a special handling for backtraces :man_facepalming:

Hold on, I've got another idea.

c42f:
Replaced by #35371

PR #35349, opened 2020-04-03
RFC: Introduce Matrix^Matrix functionality

ztultrebor:
Given two matrices A and B, we can extend the scalar definition of exponentiation to state

`A^B = exp(B * log(A))`

provided A and B are square matrices of identical dimensionality.

### DIFF
--- a/stdlib/LinearAlgebra/src/dense.jl
+++ b/stdlib/LinearAlgebra/src/dense.jl
@@ -545,6 +545,22 @@ Base.:^(b::Number, A::AbstractMatrix) = exp!(log(b)*A)
 # method for ℯ to explicitly elide the log(b) multiplication
 Base.:^(::Irrational{:ℯ}, A::AbstractMatrix) = exp(A)
 
+function Base.:^(A::AbstractMatrix, B::AbstractMatrix)
+    nA1, nA2 = size(A)
+    nB1, nB2 = size(B)
+    if nA1!=nA2 || nB1!=nB2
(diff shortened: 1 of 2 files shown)


oscardssmith:
Just wondering, why did you close this? Does Julia already do it if you import linear algebra?

ztultrebor:
I may have made a logical error:  `A^B = exp(log(A) * B)` might be more natural definition.  This is not equivalent to what I stated above. I need to investigate a bit further.

StefanKarpinski:
From asking on the #linear-algebra channel on the JuliaLang Slack, it seems that picking one definition of `X^Y` for all square matrices is overly simplistic since there are both left and right exponentials, which need not be the same, and unless `X` is normal the exponential seems to be ill-defined. Starter reading (which I needed): https://en.wikipedia.org/wiki/Matrix_exponential#Matrix-matrix_exponentials. In any case, while this is a cool idea (I hadn't thought about it before), this seems like something we should not add without significantly more consideration—and I would guess that the naive definition is unlikely to be a good way to calculate this.

mschauer:
Following that trace https://web.archive.org/web/20081031231237/http://www.rockefeller.edu/labheads/cohenje/PDFs/215BarrabasCohenalApp19941.pdf, the form `exp(B*log(A))` has a relation to https://en.wikipedia.org/wiki/Von_Neumann_entropy but I doubt people will have an actual need compute `A^B` which has to be covered here.

PR #35356, opened 2020-04-03
Update `UnicodeData` checksum

staticfloat:
Upstream has apparently modified this file as of March 30th 2020, which
is annoying as it has broken all older builds of Julia attempting to
download and check this file.


ararslan (approved):

StefanKarpinski:
Wow, that's extremely uncool of them. Perhaps we should host our own copy of this file and update if upstream changes?

vtjnash (approved):
Would it help if we gave this a new name also on our side?

staticfloat:
> Would it help if we gave this a new name also on our side?

Yes, it turns out that we were storing this as an unversioned filename, so different branches conflicted.

PR #35337, opened 2020-04-01
Fix segfault on ill-formed module Expr

Liozou:
Fixes #34544

The original issue seemed to be more about module Expr whose third argument was an Expr with an empty `args`, so I added a check for this specific case.

Issue #34544, opened 2020-01-27
Eval Expr without LineNumberNode crashes Julia


mattBrzezinski:
Julia Version: 1.3.0
Operating System: macOS Catalina Version 10.15.2

Evaluating an Expr without a `LineNumberNode` crashes Julia:
```julia
eval(Expr(:call, :eval, Expr(:quote, Expr(:module, true, :bar, Expr(:block)))))
```

```
signal (11): Segmentation fault: 11
in expression starting at REPL[1]:1
fl_raise at /Users/sabae/buildbot/worker/package_macos64/build/src/flisp/flisp.c:148
type_error at /Users/sabae/buildbot/worker/package_macos64/build/src/flisp/flisp.c:190
apply_cl at /Users/sabae/buildbot/worker/package_macos64/build/src/flisp/flisp.c:0
_applyn at /Users/sabae/buildbot/worker/package_macos64/build/src/flisp/flisp.c:0
fl_applyn at /Users/sabae/buildbot/worker/package_macos64/build/src/flisp/flisp.c:722
jl_call_scm_on_ast at /Users/sabae/buildbot/worker/package_macos64/build/src/ast.c:915
jl_eval_module_expr at /Users/sabae/buildbot/worker/package_macos64/build/src/toplevel.c:170
jl_toplevel_eval_flex at /Users/sabae/buildbot/worker/package_macos64/build/src/toplevel.c:640
jl_toplevel_eval at /Users/sabae/buildbot/worker/package_macos64/build/src/toplevel.c:823 [inlined]
jl_toplevel_eval_in at /Users/sabae/buildbot/worker/package_macos64/build/src/toplevel.c:843
eval at ./boot.jl:330 [inlined]
eval at ./client.jl:425
jl_apply at /Users/sabae/buildbot/worker/package_macos64/build/src/./julia.h:1631 [inlined]
do_call at /Users/sabae/buildbot/worker/package_macos64/build/src/interpreter.c:328
eval_body at /Users/sabae/buildbot/worker/package_macos64/build/src/interpreter.c:0
jl_interpret_toplevel_thunk_callback at /Users/sabae/buildbot/worker/package_macos64/build/src/interpreter.c:888
unknown function (ip: 0xfffffffffffffffe)
unknown function (ip: 0x11f6ca70f)
unknown function (ip: 0x0)
jl_interpret_toplevel_thunk at /Users/sabae/buildbot/worker/package_macos64/build/src/interpreter.c:897
[...]

PR #35363, opened 2020-04-05
add null check for module-default-defs

ssikdar1:
# Description

Issue #34544.
In PR #35337  there was a suggestion to add a check for `body` in `module-default-defs` in `src/jlfrontend.scm`. This adds that check.

# Testing
In the julia repl I tried a couple more variations that would cause a segfault
```
julia> eval(Expr(:call, :eval, Expr(:quote, Expr(:module, true, :bar, Expr(:block)))))
Main.bar

julia> eval(Expr(:module, true, :bar, Expr(:block)))
WARNING: replacing module bar.
Main.bar

julia> eval(Expr(:module, true, :bar, Expr(:quote)))
WARNING: replacing module bar.
Main.bar

julia> eval(Expr(:call, :eval, Expr(:quote, Expr(:module, true, :bar, Expr(:block)))))
WARNING: replacing module bar.
Main.bar
```


JeffBezanson:
Great, thanks! Could you add a test case to test/syntax.jl?

ssikdar1:
5a0b565 adds a fix for #35367 as well

Output for those two lines in the REPL:
```
$ ./julia 
               _
  (_)     | (_) (_)    |
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.5.0-DEV.572 (2020-04-06)
 _/ |\__'_|_|_|\__'_|  |  linenumber/0b9ed994df* (fork: 2 commits, 4 days)
|__/                   |

julia> eval(Expr(:quote, Expr(:module, true, :bar, Expr(:quote))))
:(module bar
  end)

julia> eval(Expr(:call, :eval, Expr(:quote, Expr(:module, true, :bar, Expr(:quote)))))
Main.bar
```

src/ast.c
```diff
@@ -575,7 +575,7 @@ static jl_value_t *scm_to_julia_(fl_context_t *fl_ctx, value_t e, jl_module_t *m
             assert(jl_is_symbol(ex));
             temp = jl_module_globalref(jl_core_module, (jl_sym_t*)ex);
         }
-        else if (sym == inert_sym || (sym == quote_sym && (!iscons(car_(e))))) {
+        else if ((sym == inert_sym || (sym == quote_sym && (!iscons(car_(e))))) && iscons(e)) {
```
JeffBezanson, line 578: `iscons(e)` needs to be checked before `car_(e)`; `car` is only valid if `e` is a cons.

test/core.jl
```diff
@@ -7198,3 +7198,11 @@ struct NFANode34126
 end
 
 @test repr(NFANode34126()) == "$NFANode34126(Tuple{Nothing,$NFANode34126}[])"
+
+@testset "issue #34544/35367" begin
+    # Test these evals dont throw an error
+    eval(Expr(:call, :eval, Expr(:quote, Expr(:module, true, :bar, Expr(:block)))))
+    eval(Expr(:module, true, :bar, Expr(:block)))
+    eval(Expr(:quote, Expr(:module, true, :bar, Expr(:quote))))
+    eval(Expr(:call, :eval, Expr(:quote, Expr(:module, true, :bar, Expr(:quote)))))
```
JeffBezanson, line 7207: I don't think this should be valid; the 3rd argument to `module` has to be a `block` expression. That can be checked at the top of `jl_eval_module_expr`.

ssikdar1:
@JeffBezanson added an if check  inside `jl_eval_module_expr` in dce7897.

The following now throw an Exception:
```
julia> eval(Expr(:call, :eval, Expr(:quote, Expr(:module, true, :bar, Expr(:quote)))))
ERROR: syntax: module expression third argument must be a block

julia> eval(Expr(:module, true, :bar, Expr(:foo)))
ERROR: syntax: module expression third argument must be a block

julia> eval(Expr(:module, true, :bar, Expr(:quote)))
ERROR: syntax: module expression third argument must be a block
```

The following does not return an ErrorException (but no longer segfaults ):
```
julia> eval(Expr(:quote, Expr(:module, true, :bar, Expr(:quote))))
:(module bar
  end)
```
I think due to the quoting? Which I think is fine?

mbauman:
I just hit this bug myself — glad to see a fix already here!  My use-case was a macro:

```Julia
julia> macro foo()
           Expr(:module, true, esc(:Bar), Expr(:block))
       end
@foo (macro with 1 method)

julia> @foo
WARNING: replacing module Bar.

```

JeffBezanson:
> I think due to the quoting? Which I think is fine?

Right; the `eval` just strips off the quote, so the module expression itself is never evaluated, so the error doesn't happen.

PR #31858, opened 2019-04-27
WIP: Refactoring of README.md

ViralBShah:
The idea is to have only the default instructions that are we would like to *just work* in the top-level README, and move the detailed build and install instructions into a separate file. 

I will be further refactoring the INSTALL.md and the various other platform specific READMEs.

Issue #35122, opened 2020-03-16
Build docs should be in the manual


fredrikekre:
The [`doc/build`](https://github.com/JuliaLang/julia/tree/master/doc/build) folder should probably be included in the devdocs section of the real manual? If that's not suitable the folder should probably move elsewhere -- the `doc` folder only contains things related to the manual.

Shashankjain12:
hey @fredrikekre I would like to take this issue

ViralBShah:
@Shashankjain12 Did you make progress on this?

@musm We probably want to accomplish this as part of the better documentation of the Windows Terminal

PR #35369, opened 2020-04-06
Use relative links in doc/build/README.md

mortenpi:
This is so that you wouldn't change branches on GitHub when clicking on the links in this file.

cc @ViralBShah in case the current behaviour is intentional (ref #31858).

### DIFF
--- a/doc/build/README.md
+++ b/doc/build/README.md
@@ -7,19 +7,19 @@
 
 This directory contains various details related to building Julia:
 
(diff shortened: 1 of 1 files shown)


fredrikekre:
See also #35122.

mortenpi:
#35122 is a bit more work though (if we want to merge it into the manual), so in the meanwhile I think we could go ahead and merge this.

Issue #35125, opened 2020-03-16
Missing GCD methods on complex numbers


SyxP:
In Julia 1.3.1, there is no methods `gcd(::Complex{<:Integer}, ::Complex{<:Integer})`.

However, it is possible to define `gcd` on any Euclidean domain, of which $\mathbb{Z}[i]$ is one.

ravibitsgoa:
Hi @SyxP ,
I would like to work on this.

SyxP:
It seems like [base/intfuncs.jl](https://github.com/JuliaLang/julia/blob/master/base/intfuncs.jl) is composed of functions for Integer types. So [base/complex.jl](https://github.com/JuliaLang/julia/blob/master/base/complex.jl) might be a better home for it. Either way, we can just write the function and some tests first and then decide where it should belong in the PR.

ravibitsgoa:
For this, I wrote a function and a few tests, but I am wondering about these:
Should we ensure that the resultant gcd is normalized (rotated into the first quadrant)?
E.g. gcd(3+im, 1-im) = 1-im. It can be normalized to 1+im.
As discussed here: https://math.stackexchange.com/questions/1836682/calculating-the-gcd-of-complex-numbers
Also, for the numerical stability, during the intermediate computation, I am using Complex{<:Rational} rather than Complex{<:AbstractFloat}. Is that correct?

SyxP:
It would make sense that we ensure the resulting gcd will lie in the first quadrant (i.e. the argument lies within $[0, \pi/2)$). We currently ensure that `gcd(::Integer, ::Integer)` is non-negative.

I don't understand why one would require intermediate steps to use `Complex{<:Rational}`, you should be able to do everything without leaving `Complex{<:Integer}`

ravibitsgoa:
Just a little clarification: Since we want the argument within [0, π/2), we should return 1 for gcd(im, 2im) = im = 1 right?

I was using Complex{<:Rational} because I was trying to do z1//z2 directly (during the remainder calculation). Now I changed it to z1*z2/abs2(z2) and used only Complex{<:Integer}.

Now I am worrying about overflow cases when abs2(z) is too large to fit within a particular type. How do we handle such cases? Currently, abs2 function does not handle overflows.
(E.g. abs2(complex(Int32(10)^5, Int32(10)^5))).

SyxP:
[...]

PR #35374, opened 2020-04-06
Added GCD methods for complex numbers

ravibitsgoa:


base/complex.jl
```diff
@@ -1028,3 +1028,68 @@ function complex(A::AbstractArray{T}) where T
...
+    a - b * div(a, b, r)
+end
+
+function _first_quadrant(a::Complex)
+    ar, ai = reim(a)
+    if iszero(ar)
+        Complex(abs(ai), zero(ar))
+    elseif iszero(ai)
+        Complex(abs(ar), zero(ai))
+    elseif ar>0 && ai>0     # The complex number is already in the first quadrant
```
StefanKarpinski, line 1057:
```suggestion
    elseif ar > 0 && ai > 0     # The complex number is already in the first quadrant
```

base/complex.jl
```diff
@@ -1028,3 +1028,68 @@ function complex(A::AbstractArray{T}) where T
...
+        a
+    elseif ar<0 && ai>0     # In the second quadrant
```
StefanKarpinski, line 1059:
```suggestion
    elseif ar < 0 && ai > 0     # In the second quadrant
```

base/complex.jl
```diff
@@ -1028,3 +1028,68 @@ function complex(A::AbstractArray{T}) where T
...
+        Complex(ai, -ar)
+    elseif ar<0 && ai<0     # In the third quadrant
```
StefanKarpinski, line 1061:
```suggestion
    elseif ar < 0 && ai < 0     # In the third quadrant
```

base/complex.jl
```diff
@@ -1028,3 +1028,68 @@ function complex(A::AbstractArray{T}) where T
     end
     convert(AbstractArray{typeof(complex(zero(T)))}, A)
 end
+
+function div(a::Complex{T}, b::Complex{V}, r::RoundingMode=RoundNearest) where {T<:Integer, V<:Integer}
+    R = promote_type(T, V)
+    a, b = Complex{R}(a), Complex{R}(b)
+    b̅ = conj(b)
+    # TODO: Handle overflow when calculating a*b̅
+    t = a*b̅
```
simeonschaub, line 1037: Since `div` can't overflow for real numbers, I think, we should try to guarantee that `div` doesn't overflow for complex numbers either. I think this safety is worth the extra performance cost in most uses of this function. Implementing `Base.checked_mul(::Complex{<:Integer}, ::Complex{<:Integer})` should actually be fairly straightforward to do.

base/complex.jl
```diff
@@ -1028,3 +1028,68 @@ function complex(A::AbstractArray{T}) where T
     end
     convert(AbstractArray{typeof(complex(zero(T)))}, A)
 end
+
+function div(a::Complex{T}, b::Complex{V}, r::RoundingMode=RoundNearest) where {T<:Integer, V<:Integer}
+    R = promote_type(T, V)
+    a, b = Complex{R}(a), Complex{R}(b)
+    b̅ = conj(b)
```
simeonschaub, line 1035: `conj` can overflow if the imaginary part is `typemin(Int)`, so we should check for that.

base/complex.jl
```diff
@@ -1028,3 +1028,68 @@ function complex(A::AbstractArray{T}) where T
     end
     convert(AbstractArray{typeof(complex(zero(T)))}, A)
 end
+
+function div(a::Complex{T}, b::Complex{V}, r::RoundingMode=RoundNearest) where {T<:Integer, V<:Integer}
+    R = promote_type(T, V)
+    a, b = Complex{R}(a), Complex{R}(b)
```
simeonschaub, line 1034: I think, it would be better to use dispatch for promotion here, since that way, it's more transparent, if a user wants to have a different implementation for a custom integer type.

base/complex.jl
```diff
@@ -1028,3 +1028,68 @@ function complex(A::AbstractArray{T}) where T
...
+    b̅ = conj(b)
+    # TODO: Handle overflow when calculating a*b̅
+    t = a*b̅
+    # TODO: Create checked_abs2(::Complex{<:Integer})
+    # TODO: Handle overflow when calculating the norm of complex numbers
+    abs2_b = abs2(b)
+    abs2_b < 0 && __throw_gcd_overflow(a, b)
```
simeonschaub, line 1041: This also doesn't catch all cases of overflow, e.g.:
```julia-repl
julia> abs2(typemax(Int) + im*typemax(Int))
2
```
I think the best solution here is to add a `checked_abs2`, like you mentioned.

base/complex.jl
```diff
@@ -1028,3 +1028,68 @@ function complex(A::AbstractArray{T}) where T
...
+    Complex(div(real(t), abs2_b, r), div(imag(t), abs2_b, r))
```
simeonschaub, line 1042: It looks like you are using a 1-norm here for a notion of "nearest complex number", which is probably fine for most cases, but I think should be documented.

base/complex.jl
```diff
@@ -1028,3 +1028,68 @@ function complex(A::AbstractArray{T}) where T
...
+end
+
+function rem(a::Complex{T}, b::Complex{V}, r::RoundingMode=RoundNearest) where {T<:Integer, V<:Integer}
+    R = promote_type(T, V)
+    a, b = Complex{R}(a), Complex{R}(b)
+    a - b * div(a, b, r)
```
simeonschaub, line 1048: I think this shouldn't overflow, but would have to check with some more edge cases.

test/complex.jl
```diff
@@ -1187,3 +1187,66 @@ end
...
+
+            @test_broken gcd(complex(x, T(0)), complex(T(2), T(0))) === complex(T(1), T(0))
+            @test_broken gcd(complex(x, x), complex(T(2), T(0))) === complex(T(1), T(1))
+        end
+        let x = typemin(T)
+            @test gcd(complex(x, x), complex(T(1), T(0))) === complex(T(1), T(0))
+            @test gcd(complex(x, x), complex(T(-1), T(0))) === complex(T(1), T(0))
+            @test gcd(complex(x, x), complex(T(0), T(1))) === complex(T(1), T(0))
+            @test gcd(complex(x, x), complex(T(0), T(-1))) === complex(T(1), T(0))
+        end
```
simeonschaub, line 1238: Should test with typemax and typemin on the right as well. Especially typemin is the interesting one, because `-typemin(Int) == typemin(Int)`.

base/complex.jl
```diff
@@ -1028,3 +1028,68 @@ function complex(A::AbstractArray{T}) where T
...
+        Complex(abs(ar), zero(ai))
+    elseif ar > 0 && ai > 0     # The complex number is already in the first quadrant
+        a
+    elseif ar < 0 && ai > 0     # In the second quadrant
+        Complex(ai, -ar)
+    elseif ar < 0 && ai < 0     # In the third quadrant
+        -a
+    else                        # In the fourth quadrant
+        Complex(-ai, ar)
+    end
```
simeonschaub, line 1065: This should also check for `ar` or `ai` being `typemin(Int)`, because that breaks `abs` and `-`.

simeonschaub (changes requested):
Overall looks good, but see my comments about checking for overflow. I think the `div` and `rem` should also have their own docstring explaining how they work.

base/complex.jl
ravibitsgoa: That's true, but an issue with the current division algorithm (a/b = a*b̅/abs2(b)), is that due to abs2 overflow, we cannot ensure that a/b *never overflows*. We can indeed try to catch and handle all the possible overflows, but that will be it.

We can always find some inputs which have b::Complex{T} and abs2(b)>typemax(T), which will make div(a, b) overflow.

E.g. a = Complex{Int32}(1, 1), b= Complex{Int32}(10^5, 10^5). Here, a/b will overflow due to the current algorithm.

simeonschaub: I'm not sure, what you mean. I mean, we check in `abs2` and in `*`, that nothing overflows and then you're just calling `div` on real integers, which can't overflow, so we should catch all cases of overflow. I'm wondering though, if there aren't any better algorithms for division of complex integers, because checking in each step will prevent overflow in the result, but we will also throw errors for cases where either `a*b'` or `abs2(b)` overflows, but `a/b` should actually still be representable as the complex integer type of `a` and `b`.

ravibitsgoa: > we will also throw errors for cases where either `a*b'` or `abs2(b)` overflows

This is what I meant. It is possible that abs2(b) overflows, but a/b is still representable as a complex integer. E.g. `div(Complex{Int32}(1, 1), Complex{Int32}(10^5, 10^5))`.

base/complex.jl
```diff
@@ -1028,3 +1028,114 @@ function complex(A::AbstractArray{T}) where T
...
+0 + 2im
+
+julia> div(-5 + 5im, 2 + 2im, RoundNearestTiesAway)
+0 + 3im
+
+julia> div(-5 + 5im, 2 + 2im, RoundNearestTiesUp)
+0 + 3im
+```
+"""
+function div(a::Complex{T}, b::Complex{V}, r::RoundingMode=RoundNearest) where {T<:Integer, V<:Integer}
```
simeonschaub, line 1068: `RoundNearest` shouldn't be the default here, because `div` for real integers always rounds down. I am wondering whether all rounding modes actually make sense for complex numbers though.

base/complex.jl
simeonschaub: As soon as one of `abs(b)` or `a*b'` overflows, isn't the end result also wrong with your current algorithm, even if the result might still be representable?

base/complex.jl
ravibitsgoa: I kept RoundNearest as the default rounding mode since Euclidean GCD algorithm requires RoundNearest to work properly. I can change it to RoundToZero.

Regarding the rounding modes, we can say it is similar to rounding of complex quotient like `round(a/b, r, r)`.

base/complex.jl
simeonschaub: > Also, should checked_mul(::Complex, ::Complex) be in complex.jl or in checked.jl ?
complex.jl is probably better. Defining it in checked.jl might even cause problems during bootstrap, because I think that module gets loaded before complex.jl

ravibitsgoa: Yes. Probably we need to find a better algorithm if there exists one. The best we can do right now is catching all the overflows.

base/complex.jl
simeonschaub: You could also just accept `RoundNearest` by restricting `r` with `RoundingMode{:Nearest}`. I think it would make the most sense here though, to have separate rounding mode arguments for the real and imaginary part, to be consistent with `round`.

base/complex.jl
dpsanders: Just an idea: how about multiplying by `(1 / abs(b))^2` instead? Not sure what that will due to precision problems though. But something along those lines might work.

dpsanders: If you already have an approximate result there may be a way to improve it using a suitable Newton method.

simeonschaub: So you think this should internally use floating point arithmetic? That might not be a bad idea here.

dpsanders: Or use rationals? Or promote to a wider type? To be honest I don't quite understand what's going on in the PR and what the constraints are.

simeonschaub: @ravibitsgoa Could you explain your usecase for this? That would probably help to understand some of the constraints

ravibitsgoa: https://github.com/JuliaLang/julia/issues/35125
This PR is based on this issue. After quite a long discussion there, we decided we should avoid using floating points or rationals or wider type internally. Although those three are quite decent solutions IMO, but they will create a problem when we want to find exact integer answers of div, rem and gcd.

KlausC: I really cannot imagine those overflow-checked stuff (if you get it right) will ever be more efficient than just calculating with widened precision.

base/complex.jl
```diff
@@ -1028,3 +1028,114 @@ function complex(A::AbstractArray{T}) where T
...
+        Complex(ai, -ar)
+    elseif ar < 0 && ai < 0     # In the third quadrant
+        -a
+    else                        # In the fourth quadrant
+        Complex(-ai, ar)
+    end
+end
+
+"""
+    gcd(z1, z2)
```
KlausC, line 1115: If you implement `gcd` wouldn't you also expect `gcdx` (as for `Rational`)?

Is there a sensible extension for `Complex{<:Rational}` ?

ravibitsgoa: As it turns out, yes, gcdx can also be extended to `Complex{<:Integer}`.
https://math.stackexchange.com/questions/2911332/modular-inverse-of-gaussian-integers/2913973#2913973

Also, I think rather than `Complex{<:Rational}`, you meant `Complex{<:Integer}`, since the current methods are being added for `Complex{<:Integer}` only. In future, we might consider extending them to `Complex{<:Rational}`.

base/complex.jl
ravibitsgoa: Can you give a few possible edge cases for this, which are not covered as of now?

base/complex.jl
ravibitsgoa: Should I add gcdx method as well?

base/complex.jl
ravibitsgoa: Can someone help me find good test cases for performance testing for these methods?
Since I am new to Julia and do not have much experience with performance testing. I am unable to construct tests for which gcd would take more iterations. 
I tried running a few tests for the gcd method, which take like 0.000001 seconds to complete.

ravibitsgoa: For Euclidean GCD algorithm, the [worst case](https://en.wikipedia.org/wiki/Euclidean_algorithm#Algorithmic_efficiency) occurs when arguments are consecutive Fibonacci numbers.

Using this, I tested gcd(f1 + 0im, f2+ 0im), for f1 = 1836311903, f2= 2971215073, since these are the largest Int64 arguments for which gcd does not overflow.

Benchmarking with @time gives `0.000009 seconds (3 allocations: 96 bytes)` for the current overflow-checked implementation.

I compared it against another implementation, which calculates gcd by widening arguments into `Complex{BigInt}` first. For such a complex gcd method, @time gives around `
  0.000150 seconds (2.23 k allocations: 46.305 KiB)`. 

Hence, we can safely conclude that for arguments below up to f1, f2, calculating gcd without widening into `Complex{BigInt}` might be a better decision.

If someone finds better benchmarks, I am happy to try them as well.

KlausC:
Please fix this ambiguity error:
```
ambiguous: Test Failed at D:\buildbot\worker-tabularasa\tester_win64\build\share\julia\test\ambiguous.jl:169
  Expression: detect_ambiguities(Core, Base; imported = true, recursive = true, ambiguous_bottom = false) == []
  Evaluated: Tuple{Method,Method}[(rem(a::Complex{T}, b::Complex{T}, rr::RoundingMode) where T<:Integer in Base at complex.jl:1125, rem(x, y, ::RoundingMode{:ToZero}) in Base at div.jl:67), (rem(a::Complex{T}, b::Complex{T}, rr::RoundingMode) where T<:Integer in Base at complex.jl:1125, rem(x, y, ::RoundingMode{:Up}) in Base at div.jl:69), (rem(a::Complex{T}, b::Complex{V}, rr::RoundingMode) where {T<:Integer, V<:Integer} in Base at complex.jl:1133, rem(x, y, ::RoundingMode{:ToZero}) in Base at div.jl:67), (rem(a::Complex{T}, b::Complex{V}, rr::RoundingMode) where {T<:Integer, V<:Integer} in Base at complex.jl:1133, rem(x, y, ::RoundingMode{:Up}) in Base at div.jl:69), (rem(a::Complex{T}, b::Complex{V}, rr::RoundingMode) where {T<:Integer, V<:Integer} in Base at complex.jl:1133, rem(x, y, r::RoundingMode{:Nearest}) in Base at div.jl:70), (rem(a::Complex{T}, b::Complex{T}, rr::RoundingMode) where T<:Integer in Base at complex.jl:1125, rem(x, y, ::RoundingMode{:Down}) in Base at div.jl:68), (rem(a::Complex{T}, b::Complex{V}, rr::RoundingMode) where {T<:Integer, V<:Integer} in Base at complex.jl:1133, rem(x, y, ::RoundingMode{:Down}) in Base at div.jl:68)] == Any[]
```

base/complex.jl
```diff
@@ -1065,37 +1091,75 @@ julia> div(-5 + 5im, 2 + 2im, RoundNearestTiesUp)
 0 + 3im
 ```
 """
-function div(a::Complex{T}, b::Complex{V}, r::RoundingMode=RoundNearest) where {T<:Integer, V<:Integer}
-    R = promote_type(T, V)
-    a, b = Complex{R}(a), Complex{R}(b)
+function div(a::Complex{T}, b::Complex{T}, rr::RoundingMode=RoundNearest, ri::RoundingMode=rr) where T<:Integer
+    (T != BigInt) && (imag(b) === typemin(T)) && throw(OverflowError("Cannot compute conj($b)."))
```
KlausC, line 1095: This check for the existence of `typemin` is not reliable (there could be other subtypes of `Integer` not having it). For `T <: Unsigned` the error is thrown wrongly if `imag(b) == 0`.
You would do better with `T <: Base.BitSigned`.

ravibitsgoa:
For the ambiguity error, can I fix this TODO in div.jl line 66? (https://github.com/JuliaLang/julia/commit/0c2b7393b46a1e1eb1acb9cf7fa4982b67c308b4#diff-b1fc87557f78879f66387487c7a02689R66)

ravibitsgoa:
Hi everyone,
Can someone review this PR again?

StefanKarpinski:
@simonbyrne, @stevengj, any chance either of you could review this and merge if it looks good?

ravibitsgoa:
Hi everyone,
Can someone review and merge this PR?

base/complex.jl
```diff
@@ -254,15 +257,27 @@ end
 
 Compute the complex conjugate of a complex number `z`.
 
+When `conj` is applied to signed complex integers, overflow may occur,
+resulting in the return of a negative value of imaginary part. This overflow
+occurs only when `conj` is applied to the minimum representable value of a
+signed integer. That is, when `x == typemin(typeof(x))`,
+`imag(conj(x * im)) == x < 0`, not `-x` as might be expected.
```
stevengj, line 264: This warning seems unnecessary.   We don't have similar warnings for `-` or other basic arithmetic operators.

base/complex.jl
```diff
@@ -254,15 +257,27 @@ end
...
+
 # Examples
 ```jldoctest
 julia> conj(1 + 3im)
 1 - 3im
+
+julia> conj(1 + typemin(Int64)im)
+1 - 9223372036854775808im
```
stevengj, line 272: same … this doesn't seem the place to talk about integer overflow

ravibitsgoa: I just added it because the expected behaviour would be 1 - typemin(Int64) im.

DilumAluthge:

@ravibitsgoa Are you still interested in working on this PR?

ravibitsgoa:
Hi,
No, I'm not interested in working on this PR now.

PR #35376, opened 2020-04-06
Version check for serialized data headers

c42f:
Check that the serialization data format is compatible before attempting to read a serialized stream. New versions of Serialization are assumed to be able to read old serialized data, but attempting to read newer data with an older version of Serialization will fail with an error.

This code was written as part of an alternative solution to patching up `CodeInfo` serialization in #35371, but seems like it could be useful: it should prevent confusion if a newer julia versions somehow ends up sending data to older workers.

### DIFF
--- a/stdlib/Serialization/src/Serialization.jl
+++ b/stdlib/Serialization/src/Serialization.jl
@@ -685,6 +685,36 @@ function writeheader(s::AbstractSerializer)
     nothing
 end
 
+function readheader(s::AbstractSerializer)
+    # Tag already read
+    io = s.io
+    m1 = read(io, UInt8)
+    m2 = read(io, UInt8)
+    if m1 != UInt8('J') || m2 != UInt8('L')
+        error("Unsupported serialization format (got header magic bytes $m1 $m2)")
+    end
+    version    = read(io, UInt8)
+    flags      = read(io, UInt8)
+    reserved1  = read(io, UInt8)
+    reserved2  = read(io, UInt8)
+    reserved3  = read(io, UInt8)
+    endianflag = flags & 0x3
+    wordflag   = (flags >> 2) & 0x3
+    wordsize = wordflag == 0 ? 4 :
+               wordflag == 1 ? 8 :
+               error("Unknown word size flag in header")
+    endian_bom = endianflag == 0 ? 0x04030201 :
+                 endianflag == 1 ? 0x01020304 :
+                 error("Unknown endianness flag in header")
+    # Check protocol compatibility.
+    endian_bom == ENDIAN_BOM  || error("Serialized byte order mismatch ($(repr(endian_bom)))")
+    wordsize   == sizeof(Int) || error("Serialized word size mismatch ($wordsize)")
+    if version > ser_version
+        error("""Cannot read stream serialized with a newer version of Julia.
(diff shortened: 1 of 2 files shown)


stdlib/Serialization/src/Serialization.jl
```diff
@@ -685,6 +685,36 @@ function writeheader(s::AbstractSerializer)
...
+    wordflag   = (flags >> 2) & 0x3
+    wordsize = wordflag == 0 ? 4 :
+               wordflag == 1 ? 8 :
+               error("Unknown word size flag in header")
+    endian_bom = endianflag == 0 ? 0x04030201 :
+                 endianflag == 1 ? 0x01020304 :
+                 error("Unknown endianness flag in header")
+    # Check protocol compatibility.
+    endian_bom == ENDIAN_BOM  || error("Serialized byte order mismatch ($(repr(endian_bom)))")
+    wordsize   == sizeof(Int) || error("Serialized word size mismatch ($wordsize)")
```
JeffBezanson, line 711: I'm not sure these should be errors; in many cases data will be readable moving between a 32-bit and 64-bit machine.

c42f: Well, the endianness being mismatched will break nearly everything unless we handle it explicitly? And we don't try to handle it right now AFAICT.

But quite true regarding `sizeof(Int)` as it's explicitly encoded as `Int32` and `Int64` and should deserialize ok at the other end. Should it be a warning perhaps?

c42f:
Test failure is UnicodeData.txt

stdlib/Serialization/src/Serialization.jl
c42f: I did a couple of quick tests. It turns out that that removing the `wordsize` check does allow very simple things like homogenous arrays of integers to be deserialized between 32 and 64 bit Julia versions. So that can be helpful.

Once things get slightly complicated things start to break. For example serialized arrays of type `Vector{Tuple{Int32,Int64}}` are not portable due to platform ABI differences.

c42f:
Ok, I've done the wordsize change as discussed, and tests have passed so I'll merge this small change. Hopefully this will reduce confusion for people maintaining Julia on ad-hoc clusters.

stdlib/Serialization/src/Serialization.jl
notinaboat: Hi @c42f, @JeffBezanson, this 32/64 cross-architecture support is very useful to me. I have a cluster of 32-bit ARM sensor/control boards talking to a 64-bit intel master. I can't use `addprocs()` because of the architecture mismatch, but I've [rolled my own simple RPC](https://discourse.julialang.org/t/addprocs-to-different-achitectures/5454/2) based on `serialize(io, eval(deserialize(io)))`. It works nicely for my use case with simple data types (and seems vastly preferable to resorting to MsgPack.jl or ProtoBuf.jl)

It seems like it would be useful to have a `serialize_compat` function that guarantees to produce cross-architecture / cross-julia-version output. It could initially just throw an ArgumentError for input that isn't compatible. Support for more complex types could be added as needed.

I can see that it makes sense to keep `serialize` efficiency-focussed and keep the constraint of "must have same system image".

Do you think it would be generally useful to expose a subset of `serialize` with a compatibility guarantee?

Issue #35357, opened 2020-04-04
Customize @edit to work with VS Code Insiders


mshseek:
I would like to propose changing "code"  to "code-insiders"  in this part of [editless.jl](https://github.com/JuliaLang/julia/blob/b8e9a9ecc62ab49003bd4f1834771bdeb8cb1aa2/stdlib/InteractiveUtils/src/editless.jl#L130) so that the @edit macro would work for people using VS Code Insiders version too.

Is this worth making a PR?

ViralBShah:
I assume you are suggesting adding and not replacing. In that case it should be ok.

mshseek:
I was actually suggesting replacing. 

I expected the pattern `code-insiders` to match `code` too based on the documentation for [define_editor](https://docs.julialang.org/en/v1/stdlib/InteractiveUtils/#InteractiveUtils.define_editor) but after rereading it again I think I got it wrong.

So, instead of adding another call to `define_editor()` to support VS Code Insiders, how about we just replace `"code"` with `["code", "code-insiders"]`?

ViralBShah:
Sure.

ViralBShah:
> So, instead of adding another call to `define_editor()` to support VS Code Insiders, how about we just replace `"code"` with `["code", "code-insiders"]`?

Yes, this sounds good.

KristofferC:
https://github.com/JuliaLang/julia/pull/35382

PR #35382, opened 2020-04-06
Add support for VS Code Insiders

hsm207:
See issue #35357 for details.


hsm207:
> Have you tested locally—does this change work for you?

Yes, it works for me locally. I've tested it with `code` and `code-insiders`.

PR #16418, opened 2016-05-18
WIP: typemap split on non-leaf TypeNames

vtjnash:
There's still a few TODO items to get full use of this optimization, but I figured I'd give a sneak peek at some of the numbers I think I'm getting. Currently, this seems to give about a 50% speedup on loading precompiled files and makes the most significant dent to-date in https://github.com/JuliaLang/julia/issues/15048:

``` julia
~/julia-master$ time ./julia -e 'using DataFrames'
real    0m32.365s
user    0m31.650s
sys 0m2.069s
~/julia-master$ time ./julia -e 'using DataFrames'
real    0m0.917s
user    0m0.871s
sys 0m0.143s
```

``` julia
~/julia$ time ./julia -e 'using DataFrames'
real    0m24.753s
user    0m24.297s
sys 0m1.900s
~/julia$ time ./julia -e 'using DataFrames'
real    0m0.742s
user    0m0.696s
sys 0m0.139s
```

I think this should finally gets us to the realm where #15048 is caused by general slow startup issues (which may be helped by #5155) and not specific to reloading method tables :)

The longer technical explanation for what this patch does is that it creates a new hashmap level in the TypeMap object that splits on the uid of the names of concrete types. This gives us the following sorted levels:
- Type{LeafType} hash-table
- LeafType hash-table
- ~~linear-sorted list of the stuff that goes here (the technical definition of stuff is `jl_args_morespecific_typename`)~~
- linear-sorted list of signatures shorter than `offs` (prevents matching varargs)
- Type{Concrete TypeName} hash-table (possibly split via `jl_args_morespecific_typename` from another sig)
- Concrete TypeName hash-table (also possibly split)
- linear-sorted list of everything else
- `::Any`

PR #35384, opened 2020-04-06
typemap: attempt split on typenames also

vtjnash:
This is a recreation of https://github.com/JuliaLang/julia/pull/16418, now working due to many changes in the meantime! In measurements, it appears to successfully shave significant time off `require` calls:
```
$ time ./usr/bin/julia -e 'using DataFrames'
user 0m1.192s => 0m0.887s

$ time ./usr/bin/julia -e 'using Plots'
user 0m3.240s => 0m2.056s

$ time ./usr/bin/julia -e 'using ApproxFun
user 0m21.551s => 0m11.953s
```

### DIFF
--- a/src/typemap.c
+++ b/src/typemap.c
@@ -18,10 +18,72 @@ extern "C" {
 // compute whether the specificity of this type is equivalent to Any in the sort order
 static int jl_is_any(jl_value_t *t1)
 {
-    return (t1 == (jl_value_t*)jl_any_type ||
-            (jl_is_typevar(t1) &&
-             ((jl_tvar_t*)t1)->ub == (jl_value_t*)jl_any_type));
+    while (jl_is_typevar(t1))
+        t1 = ((jl_tvar_t*)t1)->ub;
+    return t1 == (jl_value_t*)jl_any_type;
 }
+
+static jl_value_t *jl_type_extract_name(jl_value_t *t1 JL_PROPAGATES_ROOT)
+{
+    if (jl_is_unionall(t1))
+        t1 = jl_unwrap_unionall(t1);
+    if (jl_is_vararg_type(t1)) {
+        return jl_type_extract_name(jl_tparam0(t1));
+    }
+    else if (jl_is_typevar(t1)) {
+        return jl_type_extract_name(((jl_tvar_t*)t1)->ub);
+    }
+    else if (jl_is_datatype(t1)) {
+        jl_datatype_t *dt = (jl_datatype_t*)t1;
+        if (!jl_is_kind(t1))
+            return (jl_value_t*)dt->name;
+        return NULL;
+    }
+    else if (jl_is_uniontype(t1)) {
+        jl_uniontype_t *u1 = (jl_uniontype_t*)t1;
+        jl_value_t *tn1 = jl_type_extract_name(u1->a);
+        jl_value_t *tn2 = jl_type_extract_name(u1->b);
+        if (tn1 == tn2)
+            return tn1;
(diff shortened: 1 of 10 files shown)


src/typemap.c
```diff
@@ -851,23 +1140,60 @@ static void jl_typemap_level_insert_(
     else if (l > offs) {
         t1 = jl_tparam(ttypes, offs);
     }
+    else {
+        t1 = NULL;
+    }
+    if (0 && t1 == NULL) {
```
JeffBezanson, line 1146: ?

src/typemap.c
```diff
@@ -19,9 +19,80 @@ extern "C" {
...
+        return jl_type_extract_name_imprecise(jl_tparam0(t1), abstract);
+    }
+    else if (jl_is_typevar(t1)) {
+        return jl_type_extract_name_imprecise(((jl_tvar_t*)t1)->ub, 0);
+    }
+    else if (jl_is_datatype(t1)) {
+        jl_datatype_t *dt = (jl_datatype_t*)t1;
+        if ((abstract || !dt->abstract) && !jl_is_kind(t1))
+            return 0;
+        return 1;
```
JeffBezanson, line 71: In this case we return `NULL` above, so in what sense is that imprecise? Or maybe a better question is, when does this function differ from `jl_type_extract_name() == NULL`?

src/typemap.c
```diff
@@ -280,16 +375,24 @@ static int jl_typemap_intersection_array_visitor(jl_array_t *a, jl_value_t *ty,
         JL_GC_PROMISE_ROOTED(t);
         if (t == jl_nothing || t == NULL)
             continue;
-        // `t` is a leaftype, so intersection test becomes subtype
-        if (ty == (jl_value_t*)jl_any_type || // easy case: Any always matches
-            (tparam
-             ? (jl_typeof(t) == ty || jl_isa(t, ty)) // (Type{t} <: ty), where is_leaf_type(t) => isa(t, ty)
-             : (t == ty || jl_subtype(t, ty)))) {
+        if (tparam & 2) {
```
JeffBezanson, line 378: Could use a comment on this function saying what the bits of `tparam` mean.

src/typemap.c
vtjnash: I suppose NULL is imprecise in the sense that we'll never consider asking this `jl_type_extract_name_imprecise` question if the result was NULL?

src/typemap.c
vtjnash: Whether it's the TypeName and whether it's the parameter of Type{} (but agree it should have a comment)

vtjnash:

vtjnash:
```
$ time ./julia -e '@time using Plots; @time display(@time plot(1:10))'
  4.815404 seconds (11.86 M allocations: 742.897 MiB, 4.46% gc time)
  1.499383 seconds (2.45 M allocations: 139.535 MiB, 2.96% gc time)
  4.883966 seconds (8.14 M allocations: 457.537 MiB, 2.23% gc time)

real	0m10.145s
user	0m9.966s
sys	0m0.384s
```

nanosoldier:
[Your benchmark job](https://github.com/JuliaLang/julia/pull/35384#issuecomment-655602685) has completed - possible performance regressions were detected. A full report can be found [here](https://github.com/JuliaCI/BaseBenchmarkReports/blob/8919eb56fb36ab9b1a6e403d7940607f460dfa37/2931ff9_vs_576769a/report.md). cc @ararslan

JeffBezanson (approved):

PR #35385, opened 2020-04-06
Using Val{V} instead of Val(v) in cholesky

matbesancon:
I know the type instability of Val(true) / Val(false) will be caught by the compiler here, but it doesn't cost anything and shows a preferable way to use Val (few more key strokes, type-stable)

### DIFF
--- a/stdlib/LinearAlgebra/src/cholesky.jl
+++ b/stdlib/LinearAlgebra/src/cholesky.jl
@@ -10,10 +10,10 @@
 # In the methods below, LAPACK is called when possible, i.e. StridedMatrices with Float32,
 # Float64, Complex{Float32}, and Complex{Float64} element types. For other element or
 # matrix types, the unblocked Julia implementation in _chol! is used. For cholesky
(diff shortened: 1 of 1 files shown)


fredrikekre (changes requested):
Julia 0.7/1.0 changelog:

> Uses of Val{c} in Base has been replaced with Val{c}(), which is now easily accessible via the efficient constructor Val(c). Functions are defined as f(::Val{c}) = ... and called by f(Val(c)). Notable affected functions include: ntuple, Base.literal_pow, sqrtm, lufact, lufact!, qrfact, qrfact!, cholfact, cholfact!, _broadcast!, reshape, cat and cat_t.

matbesancon:
Ok I guess we can close this then.
Still, the news file doesn't point to a corresponding PR or discussion, is is commonly admitted that the compiler will be smart enough with constant prop and not to care about type unstability anymore?

andreasnoack:
At least for this case, it's assumed that we can rely on constant propagation.

PR #33389, opened 2019-09-25
some GCLowering cleanup and add support for FCAs

vtjnash:
In preparation for some future work, change some representations and cleanup some code.

PR #32599, opened 2019-07-16
threads: further optimize scheduler based on #32551

vtjnash:
and add some debugging aids, controlled by changing the JULIA_DEBUG_SLEEPWAKE() macro in julia_threads.h

PR #35387, opened 2020-04-07
[LateGCLowering] Fix skipped Select lifting

### LINKED ISSUE #35341: Rare deadlock and segmentation fault in a parallel quicksort implementation (with an MWE)
I have a trouble with finding a cause of [a rare deadlock in my parallel quicksort](https://github.com/tkf/ThreadsX.jl/issues/104).  I tried my best, with a big help from  @chriselrod, to find out the problem but I still have no idea how to fix it.

I am reporting this here since I think there is _some_ chance that there is a bug in `julia` (scheduler? compiler?).  (In other words I'm hopelessly lost now and all I can do is "blame" the upstream.)

MWE: Install the latest ThreadsX (v0.1.1) and BenchmarkTools. Then run the following code eventually causes a deadlock when `JULIA_NUM_THREADS` is _appropriate_ (for me it's 13 for @chriselrod it's 8 but not 7):

```julia
using BenchmarkTools, ThreadsX, Random
seed = Ref(0)
while true
    @btime ThreadsX.sort($(rand(MersenneTwister(@show(seed[] += 1)), 0:0.01:1, 1_000_000)))
end
```

I've tried `git bisect` with [this script](https://github.com/tkf/ThreadsX.jl/issues/104#issuecomment-606249114) and found that #32599 changes the way the problem(s) manifest:

* After #32599 (dc46ddd857d9fcf178ee5074f635dc29c5b1ade6): running the MWE eventually hangs (it hung within 18 minuets once but it continued to run more than 30 minutes once).
* Before #32599 (f5dbc471efc3334a8c532f56fb40b06427fa52c7): running the MWE does not cause the hang.
* In both cases, the MWE can cause a segmentation fault but it is very rare; i.e., the MWE can run for more than one hour without a problem.

With f5dbc471efc3334a8c532f56fb40b06427fa52c7, I observed this segmentation fault:

```
seed[] += 1 = 247
[ Info: %CPU = 1167.0
[ Info: 45.550472802933335 minutes without a hang....

signal (11): Segmentation fault
in expression starting at none:3
setindex! at ./Base.jl:0 [inlined]
setindex! at ./subarray.jl:301 [inlined]
partition_sizes! at /root/.julia/packages/ThreadsX/OsJPr/src/quicksort.jl:167
#97 at /root/.julia/packages/ThreadsX/OsJPr/src/quicksort.jl:158 [inlined]
#3 at ./threadingconstructs.jl:126
unknown function (ip: 0x7f54cc050ee6)
_jl_invoke at /julia/src/gf.c:2144 [inlined]
jl_apply_generic at /julia/src/gf.c:2328
jl_apply at /julia/src/julia.h:1695 [inlined]
start_task at /julia/src/task.c:687
unknown function (ip: (nil))
Allocations: 6273831512 (Pool: 6273798676; Big: 32836); GC: 17817
```

However, I don't understand how [this function `partition_sizes!`](https://github.com/tkf/ThreadsX.jl/blob/2e630bff52db2e8136e3607a1d44d86811723c1e/src/quicksort.jl#L160-L172) can cause a segfault.

@chriselrod also [observed](https://github.com/tkf/ThreadsX.jl/issues/104#issuecomment-606736472) another segfault with 65b8e7ed72541d1dd5bc1f4dcabc5245e1dba718:

```
signal (11): Segmentation fault
in expression starting at REPL[3]:1
- at ./int.jl:52 [inlined]
unsafe_length at ./range.jl:517 [inlined]
unsafe_indices at ./abstractarray.jl:99 [inlined]
_indices_sub at ./subarray.jl:409 [inlined]
axes at ./subarray.jl:404 [inlined]
axes1 at ./abstractarray.jl:95 [inlined]
eachindex at ./abstractarray.jl:267 [inlined]
eachindex at ./abstractarray.jl:270 [inlined]
eachindex at ./abstractarray.jl:260 [inlined]
partition_sizes! at /home/chriselrod/.julia/packages/ThreadsX/OsJPr/src/quicksort.jl:163
unknown function (ip: 0x7f67b598804f)
unknown function (ip: (nil))
Allocations: 10113947179 (Pool: 10113915735; Big: 31444); GC: 29811
```

It is puzzling for me that how the code involved in the above stacktrace can cause a segfault:

`-` at [./int.jl:52](https://github.com/JuliaLang/julia/blob/65b8e7ed72541d1dd5bc1f4dcabc5245e1dba718/base/int.jl#L52)
`unsafe_length` at [./range.jl:517](https://github.com/JuliaLang/julia/blob/65b8e7ed72541d1dd5bc1f4dcabc5245e1dba718/base/range.jl#L517)
`unsafe_indices` at [./abstractarray.jl:99](https://github.com/JuliaLang/julia/blob/65b8e7ed72541d1dd5bc1f4dcabc5245e1dba718/base/abstractarray.jl#L99)
`_indices_sub` at [./subarray.jl:409](https://github.com/JuliaLang/julia/blob/65b8e7ed72541d1dd5bc1f4dcabc5245e1dba718/base/subarray.jl#L409)
`axes` at [./subarray.jl:404](https://github.com/JuliaLang/julia/blob/65b8e7ed72541d1dd5bc1f4dcabc5245e1dba718/base/subarray.jl#L404)
`axes1` at [./abstractarray.jl:95](https://github.com/JuliaLang/julia/blob/65b8e7ed72541d1dd5bc1f4dcabc5245e1dba718/base/abstractarray.jl#L95)
`eachindex` at [./abstractarray.jl:267](https://github.com/JuliaLang/julia/blob/65b8e7ed72541d1dd5bc1f4dcabc5245e1dba718/base/abstractarray.jl#L267)
`eachindex` at [./abstractarray.jl:270](https://github.com/JuliaLang/julia/blob/65b8e7ed72541d1dd5bc1f4dcabc5245e1dba718/base/abstractarray.jl#L270)
`eachindex` at [./abstractarray.jl:260](https://github.com/JuliaLang/julia/blob/65b8e7ed72541d1dd5bc1f4dcabc5245e1dba718/base/abstractarray.jl#L260)
`partition_sizes!` at [ThreadsX.jl/src/quicksort.jl:163](https://github.com/tkf/ThreadsX.jl/blob/v0.1.1/src/quicksort.jl#L163)

We (me and @chriselrod) tried to get a better error report by running the MWE with `--check-bounds=yes`. But [we couldn't get a hang or segfault](https://github.com/tkf/ThreadsX.jl/issues/104#issuecomment-607092045) even after running the MWE for hours.

I highly appreciate any help/advice from core devs or anyone who knows how threading works in Julia.  I'm currently trying to build ASAN-enabled `julia` with help from @vchuravy https://github.com/JuliaLang/julia/issues/35338, hoping that it gives us some information.  But I'm not sure how much I can discover by doing this as I don't know anything about LLVM or C++.

Keno:
Fixes #35341. This function was skipping the lift of the select if one of the bases was null. In this situation, the caller would assign it Number -1, which means globally rooted and is obviously wrong here. I'm not entirely sure why that code was added, but perhaps @vtjnash could comment  what the intention was and perhaps we can maintain whatever optimization that was intended to perform (even if in practice all it did was prevent rooting of `select` results).

### DIFF
--- a/src/llvm-late-gc-lowering.cpp
+++ b/src/llvm-late-gc-lowering.cpp
@@ -331,7 +331,7 @@ private:
     }
 
     void LiftPhi(State &S, PHINode *Phi);
-    bool LiftSelect(State &S, SelectInst *SI);
+    void LiftSelect(State &S, SelectInst *SI);
     Value *MaybeExtractScalar(State &S, std::pair<Value*,int> ValExpr, Instruction *InsertBefore);
     std::vector<Value*> MaybeExtractVector(State &S, Value *BaseVec, Instruction *InsertBefore);
     Value *GetPtrForNumber(State &S, unsigned Num, Instruction *InsertBefore);
@@ -600,12 +600,12 @@ Value *LateLowerGCFrame::GetPtrForNumber(State &S, unsigned Num, Instruction *In
     return MaybeExtractScalar(S, std::make_pair(Val, Idx), InsertBefore);
 }
 
-bool LateLowerGCFrame::LiftSelect(State &S, SelectInst *SI) {
+void LateLowerGCFrame::LiftSelect(State &S, SelectInst *SI) {
     if (isa<PointerType>(SI->getType()) ?
             S.AllPtrNumbering.count(SI) :
             S.AllCompositeNumbering.count(SI)) {
         // already visited here--nothing to do
-        return true;
+        return;
     }
     std::vector<int> Numbers;
     unsigned NumRoots = 1;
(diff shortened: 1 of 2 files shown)


chriselrod:
I checked out this PR, and still get frequent hangs. @tkf , can you try?

Keno:
There may be more than one issue. I'll take rr traces to debug ;).

tkf:
I've been running four instances of `ASAN_OPTIONS=detect_leaks=0:allow_user_segv_handler=1:abort_on_error=1 ASAN_SYMBOLIZER_PATH=$tooldir/llvm-symbolizer usr/bin/julia-debug -e 'using BenchmarkTools, ThreadsX, Random; for seed in 1:1000; @btime ThreadsX.sort($(rand(MersenneTwister(@show(seed)), 0:0.01:1, 1_000_000))); end'` for a while now (more than 600 iterations) and there is no failures. So I think this PR fixes the memory bug related to #35341.

@chriselrod Thanks a lot for verifying the original problem. I guess the next step is to apply this PR to pre-#32599 code and then check if the hang or segfault happens. If not, we then know some parts of #32599 introduced the problem... I'll try it later.

chriselrod:
Because you seem already set up, I'm hoping it'd not be too much trouble for you to run another example.

Here is some code that seems to reproduce the problem fairly reliably on my machine, without dependencies outside base or the standard library, other than `BenchmarkTools` (although a `foreach` would probably work too):
```julia
using Base.Threads: @spawn, @sync, nthreads
using LinearAlgebra, BenchmarkTools

function tproduct_fm(x)
    @fastmath begin
        X3 = cbrt.(x)
        X9 = cbrt.(X3).^2
    end
    lx = length(x)
    H = Matrix{eltype(x)}(undef, lx, lx)
    @sync for x_index_chunk in Iterators.partition(eachindex(x), lx ÷ 2nthreads()) 
        @spawn begin
            for i ∈ x_index_chunk
                x3 = X3[i]
                x9 = X9[i]
                js = firstindex(x):i
                       @fastmath for j in js
                           A = x3 + X3[j]
                           B = x3 * X3[j]
                           C = x9 + X9[j]
                           f = A^2 * C
                           g = -(B*A)
                           H[j, i] = f * g
                       end
            end
        end
    end
    return Symmetric(H)
end
x = rand(1000);
@benchmark tproduct_fm($x)
```

Hopefully all the issues will get fixed soon, now that the two of you seem well set up to find them.

tkf, what does it take to get everything set up to create the rr traces?

tkf:
@Keno According to https://github.com/JuliaLang/julia/issues/35341#issuecomment-609724644, this bug is due to #33389 which seems to be included in 1.4 release. Does it make sense to backport this to 1.4?

Keno:
yes

tkf:
@Keno I have a few questions:

* Do you need `rr` trace with `julia-debug`?  If we can't reproduce the bug with `julia-debug`, does it still make sense to `rr record` on plain `julia`?

* Does it make sense to use the default LLVM 9 if we are not using ASAN + rr?  Is it still better to use LLVM 10?  IIUC that was for rr+ASAN. https://github.com/JuliaLang/julia/issues/35341#issuecomment-608995165

* Does it still make sense to use `rr` master? Especially to include your patch https://github.com/mozilla/rr/pull/2488. I don't know if it was ASAN-specific.

@chriselrod If the answer to the last two questions are no, I think you can just install `rr` and run `rr record --chaos ./julia -e 'using BenchmarkTools, ThreadsX, Random; for seed in 1:100; @btime ThreadsX.sort($(rand(MersenneTwister(@show(seed)), 0:0.01:1, 1_000_000))); end'` (or run the code/script you just mentioned).

Also, if @Keno needs `julia-debug`, I think the first step for this direction is to make sure we can get the deadlock with `julia-debug`. (I'm naively assuming that `julia-debug` has some overhead that may alter how tasks are scheduled.)

Keno:
> * Do you need `rr` trace with `julia-debug`?  If we can't reproduce the bug with `julia-debug`, does it still make sense to `rr record` on plain `julia`?

Either is fine. `julia-debug` is marginally easier for me to debug.

> Does it make sense to use the default LLVM 9 if we are not using ASAN + rr? Is it still better to use LLVM 10? IIUC that was for rr+ASAN. #35341 (comment)

Whatever reproduces the bug is fine.

> Does it still make sense to use rr master? Especially to include your patch mozilla/rr#2488. I don't know if it was ASAN-specific.

Yes, please.

I'll have an updated rr jll shortly, which will simply the installation.

tkf:
> I'll have an updated rr jll shortly

Wow. I didn't know that I can install rr iva jll. This is great.

PR #35389, opened 2020-04-07
Simpler printing for Iterators.Enumerate, Iterators.Zip, and SkipMissing

### LINKED ISSUE #33695: Better printing of `zip` output
The printing of zip is quite unreadable for even simple arguments:
```julia
julia> zip((:a,:b,:c), (1,2,3))
Base.Iterators.Zip{Tuple{Tuple{Symbol,Symbol,Symbol},Tuple{Int64,Int64,Int64}}}(((:a, : :c), (1, 2, 3)))
```
Can we include better printing of `zip` outputs so that for interactive sessions, it's easier to scan?

sostock:
Fixes #33695.

This changes the printing of `Iterators.Enumerate` and `Iterators.Zip` so they are printed just as they are called:

Before:
```julia
julia> enumerate([1,2,3])
Base.Iterators.Enumerate{Array{Int64,1}}([1, 2, 3])

julia> zip([1,2,3], ('a', 'b', 'c'))
Base.Iterators.Zip{Tuple{Array{Int64,1},Tuple{Char,Char,Char}}}(([1, 2, 3], ('a', 'b', 'c')))
```
After:
```julia
julia> enumerate([1,2,3])
enumerate([1, 2, 3])

julia> zip([1,2,3], ('a', 'b', 'c'))
zip([1, 2, 3], ('a', 'b', 'c'))
```

### DIFF
--- a/base/show.jl
+++ b/base/show.jl
@@ -2223,6 +2223,15 @@ function showarg(io::IO, r::ReinterpretArray{T}, toplevel) where {T}
     print(io, ')')
 end
 
+# printing iterators from Base.Iterators
+
+function show(io::IO, e::Iterators.Enumerate)
+    print(io, "enumerate(")
+    show(io, e.itr)
+    print(io, ')')
+end
+show(io::IO, z::Iterators.Zip) = show_delim_array(io, z.is, "zip(", ',', ')', false)
+
 # pretty printing for Iterators.Pairs
 function Base.showarg(io::IO, r::Iterators.Pairs{<:Integer, <:Any, <:Any, T}, toplevel) where T<:AbstractArray
     print(io, "pairs(IndexLinear(), ::", T, ")")
--- a/base/iterators.jl
+++ b/base/iterators.jl
@@ -288,7 +288,7 @@ julia> b = ["e","d","b","c","a"]
  "a"
 
 julia> c = zip(a,b)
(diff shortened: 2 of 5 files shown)


nalimilan:
Thanks. As the CI failure shows, you'll have to update docstrings to reflect the new output.

StefanKarpinski (approved):

base/show.jl
```diff
@@ -2223,6 +2223,15 @@ function showarg(io::IO, r::ReinterpretArray{T}, toplevel) where {T}
     print(io, ')')
 end
 
+# printing iterators from Base.Iterators
+
+function show(io::IO, e::Iterators.Enumerate)
```
fredrikekre, line 2228: Should this be three-arg show?

nalimilan: IIUC 2-arg `show` is required to be parseable, but both the old and the new outputs satisfy this criterion, so probably better use the most compact form. See #34734.

sostock: If we use three-arg `show`, then `print` will use the old printing. Is that the desired behavior?

KristofferC: As Milan says, since this still parses it should be fine to use as 2 arg i think.

sostock:
I added the same printing for `skipmissing` (cf. https://github.com/JuliaLang/julia/issues/33695#issuecomment-610280399).

sostock:
I’m not sure why the CI doesn’t work for the latest commit …

nalimilan (approved):

PR #35397, opened 2020-04-07
Add space

PallHaraldsson:
[skip ci]


PallHaraldsson:
I got there from this unmerged PR: https://github.com/BenLauwens/ThinkJulia.jl/pull/24

Should we say more about the technically not keywords? While possible, I'm not sure how good an idea is is to use those as variable names...

PR #35398, opened 2020-04-07
Inference of circcopy and circshift

### LINKED ISSUE #35248: `circcopy!`does not infer for 3D arrays (does for 2D)
Hello,

Cross posting from [discourse](https://discourse.julialang.org/t/circcopy-does-not-infer-for-3d-arrays/36401) . This affects the inference of the [fft for 3D arrays](https://discourse.julialang.org/t/fft-w-does-not-infer/36393/10) for example.

Feel free to close if this is not a bug.

Cheers!

### LINKED ISSUE #35396: circshift does not infer for arrays of dimension >2
Hello,

this is related to https://github.com/JuliaLang/julia/issues/35248, they probably fail to infer for the same reason (recursive function definitions). 

```
julia> @code_warntype debuginfo=:source optimize=:true circshift!(zeros(3,3),rand(3,3),(1,1))
Variables
  #self#::Core.Compiler.Const(circshift!, false)
  dest::Array{Float64,2}
  src::Array{Float64,2}
  shiftamt::Tuple{Int64,Int64}
  inds::Tuple{Base.OneTo{Int64},Base.OneTo{Int64}}

Body::Array{Float64,2}
....

julia> @code_warntype debuginfo=:source optimize=:true circshift!(zeros(3,3,3),rand(3,3,3),(1,1,1))
Variables
  #self#::Core.Compiler.Const(circshift!, false)
  dest::Array{Float64,3}
  src::Array{Float64,3}
  shiftamt::Tuple{Int64,Int64,Int64}
  inds::Tuple{Base.OneTo{Int64},Base.OneTo{Int64},Base.OneTo{Int64}}

Body::Any
...
│           invoke Base._circshift!(_2::Array{Float64,3}, %171::Tuple{UnitRange{Int64},UnitRange{Int64}}, _3::Array{Float64,3}, %172::Tuple{UnitRange{Int64},UnitRange{Int64}}, %169::Tuple{Base.OneTo{Int64}}, %170::Tuple{Int64})
│    %174 = Core.tuple(%79, %160)::Tuple{UnitRange{Int64},UnitRange{Int64}}
│    %175 = Core.tuple(%82, %163)::Tuple{UnitRange{Int64},UnitRange{Int64}}
│    %176 = invoke Base._circshift!(_2::Array{Float64,3}, %174::Tuple{UnitRange{Int64},UnitRange{Int64}}, _3::Array{Float64,3}, %175::Tuple{UnitRange{Int64},UnitRange{Int64}}, %169::Tuple{Base.OneTo{Int64}}, %170::Tuple{Int64})::Any
└───        goto #25
25 ─        goto #26
26 ─        return %176
...
```

I will try to push an easy fix, not sure if that is the right approach though.

Cheers!

raminammour:
Hello,

Locally, this fixes https://github.com/JuliaLang/julia/issues/35396 and fixes https://github.com/JuliaLang/julia/issues/35248. I guess the recursive function definition is not inferring and the type assertion helps. It prevents the type instability to propagate further.

This is an easy/lazy fix, please feel free to close if this is not appropriate or suggest a better fix :)

Cheers!

### DIFF
--- a/test/arrayops.jl
+++ b/test/arrayops.jl
@@ -705,13 +705,13 @@ end
     src=rand(3,4,5)
     dst=similar(src)
     s=(1,2,3)
-    @inferred circshift!(dst,src,s)
+    @inferred Base.circshift!(dst,src,s)
 end
 
 @testset "circcopy" begin
     src=rand(3,4,5)
     dst=similar(src)
-    @inferred circcopy!(dst,src)
+    @inferred Base.circcopy!(dst,src)
 end
 # unique across dim
 
--- a/test/arrayops.jl
+++ b/test/arrayops.jl
@@ -702,6 +702,10 @@ end
     @test_throws ArgumentError Base.circshift!(a, a, 1)
     b = copy(a)
     @test Base.circshift!(b, a, 1) == [5,1,2,3,4]
+    src=rand(3,4,5)
+    dst=similar(src)
+    s=(1,2,3)
+    @inferred circshift!(dst,src,s)
 end
 
 # unique across dim
--- a/test/arrayops.jl
+++ b/test/arrayops.jl
@@ -708,6 +708,11 @@ end
     @inferred circshift!(dst,src,s)
 end
 
+@testset "circcopy" begin
+    src=rand(3,4,5)
+    dst=similar(src)
+    @inferred circcopy!(dst,src)
+end
 # unique across dim
 
 # With hash collisions
--- a/test/arrayops.jl
+++ b/test/arrayops.jl
@@ -701,7 +701,7 @@ end
     a = [1:5;]
     @test_throws ArgumentError Base.circshift!(a, a, 1)
     b = copy(a)
-    @test Base.circshift!(b, a, 1) == [5,1,2,3,4]
+    @test @inferred(Base.circshift!(b, a, 1) == [5,1,2,3,4])
     src=rand(3,4,5)
     dst=similar(src)
     s=(1,2,3)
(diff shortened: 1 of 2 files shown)


stevengj:
LGTM, but would be nice to know what the source of the underlying inference failure is.

@JeffBezanson, are we running into a limitation of the underlying inference code here for mutually recursive functions?

stevengj (approved):

stevengj:
Would be good to have a test, though.  Maybe just put `@inferred` into some of the existing `circshift` and `circcopy` tests?

stevengj (changes requested):
Needs regression tests.

raminammour:
I agree that this needs test so it doesn't regress again. Note that the inference succeeds for 2D arrays but not for dimensions 3 or higher. Looking at the tests quickly, all are <= 2D tests.

test/arrayops.jl
```diff
@@ -702,8 +702,17 @@ end
     @test_throws ArgumentError Base.circshift!(a, a, 1)
     b = copy(a)
     @test Base.circshift!(b, a, 1) == [5,1,2,3,4]
```
stevengj, line 704: You can just add `@inferred(...)` to this.

raminammour: Ok, I guess it can't hurt. But note that the tests I added before are needed.
Inference succeeds without this PR (the type assertion), for arrays of dimension 1 & 2. So this test would pass, even without this PR.

timholy:
This seems fine. If someone wants to put time into a more thorough fix,
- consider removing the `@inline` and `@noinline` annotations
- avoid growing tuples. Use `setindex` (without the !) to modify tuples that start out with the length of the dimensionality. `shiftamt` can still shrink as a marker of which dimension is being worked on currently.

PR #35403, opened 2020-04-08
fix #35393, use multiple TargetMachines for different optimization levels

### LINKED ISSUE #35393: LLVM ERROR: Cannot emit physreg copy instruction
I get
```
LLVM ERROR: Cannot emit physreg copy instruction
```
on 16652a6655 with Travis Linux, see https://travis-ci.org/github/fredrikekre/Literate.jl/jobs/672110054#L186

Restarting gave the same error and there are no more stacktrace info.

JeffBezanson:
`TargetMachine` does not seem to like having its optimization level repeatedly changed :man_shrugging: . Hopefully this will fix #35393; seems to work for me.

### DIFF
--- a/src/jitlayers.cpp
+++ b/src/jitlayers.cpp
@@ -17,6 +17,11 @@
 #include <llvm/Support/FormattedStream.h>
 #include <llvm/ADT/StringMap.h>
 
+#include <llvm/Support/TargetRegistry.h>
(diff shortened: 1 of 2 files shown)


vchuravy:
Might also fix the repeated AArch64 segmentation faults we have been seeing on CI.

PR #35404, opened 2020-04-08
Add busybox to download agents

jbrea:
I was puzzled that `Pkg` worked on a machine with busybox but `download` didn't.

### DIFF
--- a/base/download.jl
+++ b/base/download.jl
@@ -69,10 +69,17 @@ function download(url::AbstractString, filename::AbstractString)
             rm(filename, force=true)  # wget always creates a file
             rethrow()
         end
+    elseif Sys.which("busybox") !== nothing
+        try
+            run(`busybox wget -O $filename $url`)
+        catch
+            rm(filename, force=true)  # wget always creates a file
+            rethrow()
+        end
     elseif Sys.which("fetch") !== nothing
         run(`fetch -f $filename $url`)
     else
-        error("No download agent available; install curl, wget, or fetch.")
+        error("No download agent available; install curl, wget, busybox or fetch.")
     end
     return filename
 end


StefanKarpinski:
Presumably you've tested with busybox?

jbrea:
Pardon, I didn't get the question. I tested the edited download function on a machine that has no curl, wget or fetch but busybox; it works (don't know how to write a useful unit test, though).

musm:
`Pkg` probably copied an older version of `download` when we used to ship `busybox` alongside Julia, but we no longer ship `busybox`.  Which version on Julia are you on?

jbrea:
Sorry, I guess I wasn't clear. I am on Julia 1.4.0 (downloaded Linux binaries). On a very minimal machine I get the following:
```julia
(@v1.4) pkg> add UnicodePlots
   Updating registry at `~/.julia/registries/General`
   Updating git-repo `https://github.com/JuliaRegistries/General.git`
  Resolving package versions...
ERROR: No download engines found. We looked for: curl, wget, fetch, busybox. Install one and ensure it  is available on the path.
```
If I add busybox to the machine, I can `add UnicodePlots` without any problems.
But for example
```julia
julia> download("https://julialang.org")
ERROR: No download agent available; install curl, wget, or fetch.
```
With the changes in this PR, the `download` function works also when busybox is available.

StefanKarpinski:
> I tested the edited download function on a machine that has no curl, wget or fetch but busybox; it works (don't know how to write a useful unit test, though).

That seems good enough to me. We can't reasonably test every possible configuration of this, so the fact that you built Julia with this version and tested that it works is good enough.

musm:
@jbrea I see, makes sense. Agreed on this patch.

PR #35406, opened 2020-04-08
[RFC] haskey(d, k1, k2...): check for keys in nested dicts

StefanKarpinski:
I found myself wanting to check for the presence of a sequence of keys in a nested dictionary and this would be quite handy for that.

### DIFF
--- a/base/abstractdict.jl
+++ b/base/abstractdict.jl
@@ -15,6 +15,7 @@ end
 const secret_table_token = :__c782dbf1cf4d6a2e5e3865d7e95634f2e09b5902__
 
 haskey(d::AbstractDict, k) = in(k, keys(d))
+haskey(d::AbstractDict, k, ks...) = haskey(d, k) && haskey(d[k], ks...)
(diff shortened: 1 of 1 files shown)


laborg:
Users might think that this API checks for multiple keys in the dict at once, so basically what `all(haskey.(dict,[k1,k2,k3]))` would return if broadcasting would be defined. At least I would be a little bit surprised that this recurses into the values.

JeffBezanson:
I think this is fine in itself, but we defined multi-argument `getindex` for Dicts to look up a tuple, which conflicts somewhat. In hindsight I'm not sure that was a good idea; I don't think I've ever used that feature, and nested dicts (json!!) are much more common than tuple-key dicts.

fredrikekre:
> Users might think that this API checks for multiple keys

That's what I thought.

oscardssmith:
It seems like the best solution would be to define a `NestedDict` with a custom `getindex` method.

PR #35410, opened 2020-04-09
Rework broken-precompilation warning

c42f:
This allows me to use `Core.eval()` for replacing parts of `jl_parse_eval_all` with Julia code (part of #35243) by splitting out the warning separately. It also allows us to use the normal logging system and Base infrastructure to format the warning more nicely, including the stacktrace and pretty printing of the expression.

For bootstrap I've copied @vtjnash's bootstrap trick where `include` methods are deleted and redefined. Does this seem ok for use with `eval`? I'm unsure whether it can leave unwanted traces of the old methods in the sysimg, as I don't fully understand the world age and code caching infrastructure.

### DIFF
--- a/src/toplevel.c
+++ b/src/toplevel.c
@@ -107,6 +107,12 @@ jl_array_t *jl_get_loaded_modules(void)
     return NULL;
 }
 
+static int jl_is__toplevel__mod(jl_module_t *mod)
+{
+    return jl_base_module &&
+        (jl_value_t*)mod == jl_get_global(jl_base_module, jl_symbol("__toplevel__"));
+}
+
 // TODO: add locks around global state mutation operations
 jl_value_t *jl_eval_module_expr(jl_module_t *parent_module, jl_expr_t *ex)
 {
@@ -128,8 +134,7 @@ jl_value_t *jl_eval_module_expr(jl_module_t *parent_module, jl_expr_t *ex)
 
     // copy parent environment info into submodule
     newm->uuid = parent_module->uuid;
-    if (jl_base_module &&
-            (jl_value_t*)parent_module == jl_get_global(jl_base_module, jl_symbol("__toplevel__"))) {
+    if (jl_is__toplevel__mod(parent_module)) {
         newm->parent = newm;
         jl_register_root_module(newm);
     }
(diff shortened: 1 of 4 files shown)


vtjnash:
We strongly don't want this to go through the normal machinery—this is explicitly trying to print a fatal error message and should not be easily diverted. I'd be well okay with turning this into a call to `throw` however.

c42f:
I'd be ok with making it throw because the stacktrace is actually the most important thing for users to diagnose problems here (it's the module calling `eval` which is often at fault, not the module which is being eval'd into).

Jeff asked why this was even necessary given that `jl_current_modules` should contain the module being precompiled, but I didn't know. I'm looking into that now.

c42f:
Ok, so here's the story:

`Base.__toplevel__` is used as the parent module for package precompilation, but is not itself rooted in `jl_current_modules`. However, `create_expr_cache` in loading.jl uses `Base.include(__toplevel__, modulefile)` to bring in the code rather than `eval`.

It turns out that `Base.include` just doesn't do this check so it's currently possible to bypass the warning in released julia versions by simply using `Base.include(some_other_module, "nefarious_content.jl")`

c42f:
Ok, how about this? I've added a special case to consider `Base.__toplevel__` permanently open, and made the warning an error instead.

Changes to fix the bug in `include` will come separately as part of https://github.com/JuliaLang/julia/pull/35243

src/toplevel.c
```diff
@@ -833,11 +838,13 @@ JL_DLLEXPORT jl_value_t *jl_toplevel_eval_in(jl_module_t *m, jl_value_t *ex)
...
     if (jl_options.incremental && jl_generating_output()) {
-        if (!ptrhash_has(&jl_current_modules, (void*)m)) {
+        if (!ptrhash_has(&jl_current_modules, (void*)m) && !jl_is__toplevel__mod(m)) {
             if (m != jl_main_module) { // TODO: this was grand-fathered in
-                jl_printf(JL_STDERR, "WARNING: eval into closed module %s:\n", jl_symbol_name(m->name));
-                jl_static_show(JL_STDERR, ex);
-                jl_printf(JL_STDERR, "\n  ** incremental compilation may be fatally broken for this module **\n\n");
+                const char* name = jl_symbol_name(m->name);
+                jl_errorf("Eval into closed module `%s` breaks incremental compilation "
+                          "becuase the side effects of the `eval` will not be permanent. "
```
vchuravy, line 845:
```suggestion
                          "because the side effects of the `eval` will not be permanent. "
```

JeffBezanson:
Ah, ok, great. I like making it an error --- good to be decisive about what is and isn't allowed! Will be good to close the `include` loophole too.

c42f:
I thought I should just go ahead and close the `include` loophole so I've done that as well.

src/toplevel.c
c42f: FIxed thanks!

c42f:
Tests added. Hopefully this is good to go now provided CI passes.

c42f:
Test failures look spurious.  Win32 is a code signing problem, aarch64 presumably whatever existing problem has been causing it to segfault in several recent PRs.

KristofferC:
A rebase should fix the aarch problem.

c42f:
> A rebase should fix the aarch problem.

Done + squashed.

PR #35412, opened 2020-04-09
Add clearer documentation for asynchronous I/O to the manual.

kernelmethod:
There's heavy documentation for Tasks in the manual, but not a lot of reference to how they're used for asynchronous I/O operations. I don't think it's necessarily very clear to someone doing asynchronous I/O in Julia for the first time that

```julia
julia> f = open("foo.txt", "w");

julia> job = @async write(f, "Hello, world!");

julia> wait(job)
```

_actually_ runs asynchronously. One way to think about it is that in the following Python snippet

```python
import asyncio

async def myfun():
    with open("foo.txt", "w") as f:
        f.write("hello, world")

async def main():
    await myfun()

asyncio.run(main())
```

the `async` prefix doesn't actually cause `f.write(...)` to run asynchronously. At least as someone from a more Pythonic background, to me it's not clear in the manual that Julia's `@async` will really cause `Base.read` and `Base.write` to become asynchronous.

This addition to the documentation makes it a little clearer how `@async` works. I've added it to the bottom of the "Networking and Streams" page, which previously did not have much explicit reference to asynchronous I/O.

### DIFF
--- a/doc/src/manual/networking-and-streams.md
+++ b/doc/src/manual/networking-and-streams.md
@@ -314,6 +314,7 @@ ip"74.125.226.225"
 ```
 
 ## Asynchronous I/O
+
 All I/O operations exposed by [`Base.read`](@ref) and [`Base.write`](@ref) can be performed
 asynchronously through the use of [coroutines](@ref man-tasks). You can create a new coroutine to
 read from or write to a stream using the [`@async`](@ref) macro:
@@ -338,8 +339,8 @@ your program to block until all of the coroutines it wraps around have exited:
 julia> using Sockets
 
 julia> @sync for hostname in ("google.com", "github.com", "julialang.org")
-           conn = connect(hostname, 80)
            @async begin
+               conn = connect(hostname, 80)
                write(conn, "GET / HTTP/1.1\r\nHost:$(hostname)\r\n\r\n")
                readline(conn, keep=true)
                println("Finished connection to $(hostname)")
--- a/doc/src/manual/networking-and-streams.md
+++ b/doc/src/manual/networking-and-streams.md
@@ -315,6 +315,7 @@ ip"74.125.226.225"
 
 ## Asynchronous I/O
 
+
 All I/O operations exposed by [`Base.read`](@ref) and [`Base.write`](@ref) can be performed
 asynchronously through the use of [coroutines](@ref man-tasks). You can create a new coroutine to
 read from or write to a stream using the [`@async`](@ref) macro:
--- a/doc/src/manual/networking-and-streams.md
+++ b/doc/src/manual/networking-and-streams.md
@@ -312,3 +312,39 @@ resolution:
 julia> getaddrinfo("google.com")
 ip"74.125.226.225"
 ```
+
+## Asynchronous I/O
+All I/O operations exposed by [`Base.read`](@ref) and [`Base.write`](@ref) can be performed
+asynchronously through the use of [coroutines](@ref man-tasks). You can create a new coroutine to
+read from or write to a stream using the [`@async`](@ref) macro:
+
+```julia-repl
+julia> task = @async open("foo.txt", "w") do io
+           write(io, "Hello, World!")
+       end;
+
+julia> wait(task)
+
+julia> readlines("foo.txt")
+1-element Array{String,1}:
+ "Hello, World!"
+```
+
+It's common to run into situations where you want to perform multiple asynchronous operations
+concurrently and wait until they've all completed. You can use the [`@sync`](@ref) macro to cause
+your program to block until all of the coroutines it wraps around have exited:
+
+```julia-repl
+julia> @sync for filename in ("foo.txt", "bar.txt", "baz.txt")
+           @async open(filename, "w") do io
+               write(io, "My name is $(filename)")
+           end
+       end
+
+julia> for filename in ("foo.txt", "bar.txt", "baz.txt")
+           println(readlines(filename))
+       end
+["My name is foo.txt"]
+["My name is bar.txt"]
+["My name is baz.txt"]
+```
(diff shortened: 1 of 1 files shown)


doc/src/manual/networking-and-streams.md
```diff
@@ -312,3 +312,39 @@ resolution:
...
+```
+
+It's common to run into situations where you want to perform multiple asynchronous operations
+concurrently and wait until they've all completed. You can use the [`@sync`](@ref) macro to cause
+your program to block until all of the coroutines it wraps around have exited:
+
+```julia-repl
+julia> @sync for filename in ("foo.txt", "bar.txt", "baz.txt")
+           @async open(filename, "w") do io
+               write(io, "My name is $(filename)")
```
StefanKarpinski, line 340:
```suggestion
               println("Start writing $filename")
               write(io, "My name is $filename")
               println("Done writing $filename")
```
This seems like it would help demonstrate the asynchronous nature of the process.

kernelmethod: That seems like a good idea. I've looked at a few different options, and it seems a little difficult to find a good example using file I/O. This socket I/O example is probably fairly reproducible (with connection numbers shuffled around), but a little contrived:

```julia
julia> using Sockets

julia> @sync for i = 1:5
           conn = connect("julialang.org", 80)
           @async begin
               # Write ~32Kb of data, one byte at a time
               for j = 1:2^15
                   write(conn, rand(UInt8, 1))
               end
               println("Closed connection $(i)")
           end
       end
Closed connection 5
Closed connection 2
Closed connection 1
Closed connection 4
Closed connection 3
```

Another (possibly less reproducible) example would be

```julia
julia> using Sockets

julia> @sync for hostname in ("google.com", "github.com", "julialang.org")
           conn = connect(hostname, 80)
           @async begin
               write(conn, "GET / HTTP/1.1\r\nHost:$(hostname)\r\n\r\n")
               readline(conn, keep=true)
               println("Finished connection to $(hostname)")
           end
       end
Finished connection to google.com
Finished connection to julialang.org
Finished connection to github.com
```

StefanKarpinski: Yeah, I tried for a while to come up with good examples, but it's surprisingly hard (as you're seeing) to get the I/O subsystem to yield back to the scheduler. The simplest way by far is to call `sleep(1)` but that feels like cheating / too artificial.

StefanKarpinski: @JeffBezanson any bright ideas for something simple that will yield in this case?

kernelmethod: There hasn't been any more activity on this (I'm sure you're all pretty busy!), so I'd like to bring it back to life if possible. Asynchronous I/O is an important topic that's left almost entirely unmentioned in the manual's I/O section. I'd be happy to make any more additions / modifications needed to get this into the documentation.

It feels like the second example _ought_ to be good enough for demonstrating the `@sync` outer block / `@async` inner block. That way the Networking / Streams section will also have one code example explicitly demonstrating asynchronous file I/O, and another explicitly demonstrating asynchronous network I/O.

What else could be changed or added to this section to improve it?

StefanKarpinski:
@JeffBezanson, could you review and make suggestions?

StefanKarpinski:
Brilliant example connecting to websites!

doc/src/manual/networking-and-streams.md
```diff
@@ -312,3 +312,40 @@ resolution:
 julia> getaddrinfo("google.com")
 ip"74.125.226.225"
 ```
+
+## Asynchronous I/O
```
StefanKarpinski, line 316:
```suggestion
## Asynchronous I/O

```

doc/src/manual/networking-and-streams.md
```diff
@@ -312,3 +312,40 @@ resolution:
...
+
+It's common to run into situations where you want to perform multiple asynchronous operations
+concurrently and wait until they've all completed. You can use the [`@sync`](@ref) macro to cause
+your program to block until all of the coroutines it wraps around have exited:
+
+```julia-repl
+julia> using Sockets
+
+julia> @sync for hostname in ("google.com", "github.com", "julialang.org")
+           conn = connect(hostname, 80)
```
StefanKarpinski, line 341: Any reason not to have this line in the `@async` block too?

kernelmethod: Sorry for the delay! Nope, you're right -- the call to `connect()` would probably be better placed inside the `@async` block.

JeffBezanson:
Yes, this is a very nice example! Everything here looks fine to me. As a detail, files currently do not use libuv, so file operations never block, and on top of that most operations are buffered as well. So that might be a bit misleading. But of course the code marked `@async` still does run asynchronously anyway.

JeffBezanson:
Ah, I recently changed the `man-tasks` section to a stub, so those references should probably point to `man-asynchronous` instead.

kernelmethod:
Thanks! Any other additions / modifications I could make?

doc/src/manual/networking-and-streams.md
```diff
@@ -312,3 +312,41 @@ resolution:
 julia> getaddrinfo("google.com")
 ip"74.125.226.225"
 ```
+
+## Asynchronous I/O
+
+All I/O operations exposed by [`Base.read`](@ref) and [`Base.write`](@ref) can be performed
+asynchronously through the use of [coroutines](@ref man-tasks). You can create a new coroutine to
+read from or write to a stream using the [`@async`](@ref) macro:
```
tkf, line 320: Is this true for `IOStream` etc.?  I'm not a C programmer, but, naively reading `ios.c`, it seems to call the normal blocking `read` and `write`.  Also, how about the functions using `unused_uv_loop_arg`?

ViralBShah:
@kernelmethod Can you update the PR with the recommendations so that we can get it merged?

ViralBShah:
Bump.

Issue #35327, opened 2020-03-31
`replace` does not handle multiple patterns for String


Keno:
```
julia> replace("Keno", 'e'=>"abc", 'f'=>'g', 'a'=>'z')
ERROR: MethodError: no method matching similar(::String, ::Type{Any})
Closest candidates are:
  similar(::Array{T,1}, ::Type) where T at array.jl:357
  similar(::Array{T,2}, ::Type) where T at array.jl:358
  similar(::Array, ::Type, ::Tuple{Vararg{Int64,N}}) where N at array.jl:360
  ...
Stacktrace:
 [1] _similar_or_copy(::String, ::Type{Any}) at ./set.jl:415
 [2] replace(::String, ::Pair{Char,String}, ::Vararg{Pair,N} where N; count::Nothing) at ./set.jl:530
 [3] replace(::String, ::Pair{Char,String}, ::Pair{Char,Char}, ::Pair{Char,Char}) at ./set.jl:527
 [4] top-level scope at REPL[12]:1
```

This is surprising, because it doesn't match the API of `replace` on generic collections:
```
julia> replace([1,2,3],1=>4,2=>5,3=>6)
3-element Array{Int64,1}:
 4
 5
 6
```

Now, `replace` on string is a bit special, because it can take `String`s and Regexes also, as patterns, and perhaps we should only allow a single pattern for those cases (because otherwise the order of replacements is ambiguous), but I think it should work for `Char`, and at least give a better error message for multiple patterns of another type.

sudo-rushil:
I thought of using a recursive solution: Unfortunately, I've run into a problem.

```julia
julia> function replace(str::String, old_new::Pair...)
              for pair in old_new
                  str = replace(str, pair)
              end
              str
          end
```

However, it encountered this error.

```julia
julia> replace("Keno", 'e'=>"abc", 'f'=>'g', 'a'=>'z')
"Kzbcno"
```

I would have assumed that the desired output would be "Kabcno", but I'm not sure. It raises the question; what should be the precedence of replacements? I'll keep thinking on it and get back to you.

Keno:
For the Char replacement case, it should definitely apply the patterns only in one go.

sudo-rushil:
Got it. How's this?

```julia
julia> function replace(str::String, old_news::Pair...)
           out::String, mapping::Dict{Char,Any} = "", Dict(old_news)
           for c in str
               if c in keys(mapping)
                   out *= mapping[c]
               else
[...]

PR #30457, opened 2018-12-20
Allow multiple pair arguments in replace on strings.

MasonProtter:
This PR is intended to allow syntax like
```julia
replace("abc", 'a'=>'A', 'c'=>'C')
```
in order to be more in line with other methods of `replace` like
```julia
replace([1, 2, 1, 3], 1=>0, 2=>4)
```

PR #25396, opened 2018-01-04
replace with multiple pattern-repl pairs

KlausC:
https://github.com/JuliaLang/julia/issues/25383

PR #35414, opened 2020-04-09
[RFC] support multiple s=>r Pairs in strings replace method

francescoalemanno:
Stab @ addressing https://github.com/JuliaLang/julia/issues/35327

semantics respected:
```julia
    @test replace("foobarbaz","oo"=>"zz","ar"=>"zz","z"=>"m") == "fzzbzzbam"
    @test replace("foobarbaz","z"=>"m","oo"=>"zz","ar"=>"zz") == "fzzbzzbam"
    @test replace("foobarbaz","z"=>"m","oo"=>"zz","ar"=>"zz",count=2) == "fzzbzzbaz"
    @test replace("foobarbaz","z"=>"m",r"a.*a"=>uppercase) == "foobARBAm"
    @test replace("foobarbaz",'o'=>'z','a'=>'q','z'=>'m') == "fzzbqrbqm"
```

edit 9 apr: added fast specialization for Char=>Char mappings
edit 10 apr: generalized  fast specialization for Char=>? mappings using IOBuffer

### DIFF
--- a/base/strings/util.jl
+++ b/base/strings/util.jl
@@ -538,7 +538,7 @@ end
     c
 end
 
-function replace(str::AbstractString, subs::Pair...; count::Int=typemax(Int))
+function replace(str::ST, subs::Pair...; count::Int=typemax(Int)) where ST<:AbstractString
     @inbounds(count > 0 && for si in eachindex(subs)
         s,r=subs[si]
         R=rangify(findfirst(s,str))
--- a/base/strings/util.jl
+++ b/base/strings/util.jl
@@ -538,19 +538,35 @@ end
     c
 end
 
-function replace(str::String, subs::Pair...; count::Integer=typemax(Int))
-    @inbounds(count > 0 && for si in eachindex(subs)
+function multi_replace(str::String,count::Integer,subs::NTuple{N,Pair}) where N
+    sh=floor(Int,1.2sizeof(str))
+    @inbounds(count > 0 && for si in 1:N
         s,r=subs[si]
         R=rangify(findfirst(s,str))
         isnothing(R) && continue
+        buf=IOBuffer(sizehint=sh)
         i,j=first(R),last(R)
-        count -= 1
-        H=replace(str[begin:(i-1)], subs[(si+1):end]..., count=count)
-        length(H)>1 && (count -= 1)
-        T=replace(str[(j+1):end], subs..., count=count)
-        return H*replace(str[R],s=>r)*T
+        H,n=multi_replace(str[begin:(i-1)],count, subs[(si+1):end])
+        count-=n
+        totrepl=n
+        print(buf,H)
+        if count>0
+            print(buf,replace(str[R],s=>r))
+            count-=1
+            totrepl+=1
+        else
+            print(buf,str[R])
+        end
+        T,n=multi_replace(str[(j+1):end],count, subs)
+        totrepl+=n
+        print(buf,T)
+        return String(take!(buf)),totrepl
     end)
-    str
+    str,0
+end
+
+function replace(str::String, subs::Pair...; count::Integer=typemax(Int))
+    multi_replace(str,count,subs)[1]
 end
 
 const _ReplaceCharToTS=Pair{Char,B} where {B<:Union{AbstractChar,AbstractString,Number}}
--- a/base/strings/util.jl
+++ b/base/strings/util.jl
@@ -553,6 +553,18 @@ function replace(str::String, subs::Pair...; count::Integer=typemax(Int))
     str
 end
 
+function replace(s::String,mapping::Pair{Char,Char}...)
+    d=Dict{Char,Char}()
+    for (from,to) in mapping
+        d[from]=to
+    end
+    @inline function transform(input)
+        haskey(d,input) && return @inbounds d[input]
+        input
+    end
+    String(transform.(c for c in s))
+end
+
 # hex <-> bytes conversion
 
 """
--- a/base/strings/util.jl
+++ b/base/strings/util.jl
@@ -537,7 +537,7 @@ function tuple_replace(str::String,count::Int, subs::Pair...)
         T=tuple_replace(str[(last(R)+1):end],count,subs...)
         return (H...,replace(M,s=>r),T...)
     end
-    (str,)
+    (str,) #test
 end
 
 function replace(str::String, subs::Pair...; count::Int=typemax(Int))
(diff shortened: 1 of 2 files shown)


francescoalemanno:
@StefanKarpinski @Keno, what do you guys think?

stevengj:
If the string has length `n` and there are `N` replacements, it seems that this implementation has O(n²N) complexity in both time and space?  (Even for `N=1`, it calls itself recursively with a new string slice for each replacement, which is O(n²).)

As I also discussed in #30457 and #25396, my basic issue with this functionality is the difficulty of getting good performance with this API.  We might as well tell people to use
```
foldl(replace, patterns, init=s)
```
which at least has O(nN) complexity.

You can get better complexity by building up a regex and using a dictionary of replacments, but you still deny the caller the possibility of precomputing the regex and dictionary for the common case of calling this function for the same patterns on many strings.

stevengj:
I suppose that one option would be to define a function `stringreplacer(subs::Pair…)` that returns a "replacer" function via a regex and a dictionary of replacements, then then define `replace(s::AbstractString, subs::Pair...) = replacer(subs...)(s)`.

stevengj:
For example:
```jl
const regex_metachars = ('\\','^','$','.','|','?','*','(',')','[','{')
function backslash_metachars(s::AbstractString)
    buf = IOBuffer()
    for c in s
        if c in regex_metachars
            print(buf, '\\', c)
        else
            print(buf, c)
        end
    end
    return String(take!(buf))
end

function stringreplacer(patdict::AbstractDict{<:AbstractString,<:AbstractString})
    r = Regex(join((backslash_metachars(s) for s in keys(patdict)), '|'))
    return s::AbstractString -> replace(s, r => x -> patdict[x])
end
    
stringreplacer(pats::Pair...) = stringreplacer(Dict(pats...))
```
at which point you can do
```jl
julia> stringreplacer("oo"=>"zz","ar"=>"zz","z"=>"m")("foobarbaz")
"fzzbzzbam"
```

However, rather than having `stringreplacer` return a function, it would be even nicer to return type, so that (for example) it can be displayed nicely.

But, as has been suggested multiple times every time this has come up, it would probably make more sense to try putting together a `StringReplacer.jl` package before we consider merging this into Base.

francescoalemanno:
Thank you Steven for your feedback :) 

> If the string has length `n` and there are `N` replacements, it seems that this implementation has O(n²N) complexity in both time and space?
this general method i wrote is indeed not ideal under any perspective performance-wise

> As I also discussed in #30457 and #25396, my basic issue with this functionality is the difficulty of getting good performance with this API. 
> We might as well tell people to use
> ```
> foldl(replace, patterns, init=s)
> ```

very true, altough i do not believe the foldl approach is of any use...
with this code i wanted to match the behaviour of replace on Vectors for example, where substitutions are by definition non overlapping. for example `replace('ab','a'=>'b','b'=>'a')` should not be a no-op imho. i also wanted to allow the usage of any of the Pair{A,B} types that are allowed for Strings

> You can get better complexity by building up a regex and using a dictionary of replacments, but you still deny the caller the possibility of precomputing the regex and dictionary for the common case of calling this function for the same patterns on many strings.

that would indeed perform much better, the only problem with that approach is that it is restricted to pairs of Stringable=>Stringable, e.g a replace with `replace(s,'a'=>'b',mycustom_Char_filter => mymapping)` will definitely not be possible.

as a very first step i was aiming at a slow but interface-wise complete implementation.

> I suppose that one option would be to define a function `stringreplacer(subs::Pair…)` that returns a "replacer" function via a regex and a dictionary of replacements, then then define `replace(s::AbstractString, subs::Pair...) = replacer(subs...)(s)`.

this sounds very nice, with a "replacer" it would be possible to construct case specific optimized code for cases where the user is not using String-like pairs.

I understand now that this code is not by far Base material as there are many better options.

Do you think it is a viable option (in order to get it merged eventually) turning this PR, into a replacer + helper method, which will work only for String-like pairs (ATM)?


by the way, happy easter :)

stevengj:
To quantify my point about scaling, compare the `stringreplacer` above that uses a single precomputed regex with this PR's implementation:
```jl
replacer = stringreplacer("oo"=>"zz","ar"=>"zz","z"=>"m")
@btime $replacer("foobarbaz")
@btime $replacer("foobarbaz"^10000)

@btime replace_thisPR("foobarbaz", "oo"=>"zz","ar"=>"zz","z"=>"m")
@btime replace_thisPR($("foobarbaz"^10000), "oo"=>"zz","ar"=>"zz","z"=>"m");
```
which gives:
```
  773.821 ns (17 allocations: 720 bytes)
  4.629 ms (90009 allocations: 3.85 MiB)
  2.169 μs (80 allocations: 3.88 KiB)
  1.066 s (680050 allocations: 977.45 MiB)
```
In other words, for a short string the regex is 3x faster, and for a long string (90kB) it is 200x faster and allocates 200x less memory.

stevengj:
> Do you think it is a viable option (in order to get it merged eventually) turning this PR, into a replacer + helper method, which will work only for String-like pairs (ATM)?

I think a better approach would be to create a package.

francescoalemanno:
> However, rather than having `stringreplacer` return a function, it would be even nicer to return type, so that (for example) it can be displayed nicely.
> 
> But, as has been suggested multiple times every time this has come up, it would probably make more sense to try putting together a `StringReplacer.jl` package before we consider merging this into Base.

agreed a type would be nicer, as for making a package (unless crazy strong functionality hard to reach by Base is implemented) a very simple code similar to what you just posted, would slowly bitrot and die never tried by any user.

In any case, thank you for the detailed feedback 👍

rapus95:
sorry for coming back at this but @stevengj why do you always propose to not to add that functionality at all? You keep saying the API leads to hard to optimize situations and will only be picked when people look for performance. I'd like to question that argument since Base shouldn't be following the performance over generality mentality. I'm quite sure that people will mostly try to call that variant to replace a certain set of symbols by another set of fill ins. w/o collisions or multiple occurences. For that, `replace` is the first function to look up. So there definitively should be a generic implementation in Base. If you then see that it becomes a bottleneck, just plug in some (by then maybe available) StringReplacer.jl package. As that latter package will hold specializations for Base.replace it becomes a true drop-in replacement. IMO that's the way Base should work.
Base: Make common idioms work generically
Package: Make specific common idioms fast/specialize them

stevengj:
My concern is that people keep proposing APIs in Base that are nearly *impossible* to optimize, so you *can't* drop in a faster implementation.

That means that we are stuck forever with "use the Base API only if you don't care about performance," which is not generally what we aspire to for Base functions.

PR #35415, opened 2020-04-09
don't save implicit imports into Main

JeffBezanson:
This rolls back a small part of #34828. The problem is that the code in sysimg.jl uses various bindings from Base, and if those stay around it prevents you from importing things with the same name in a new julia session.

Caused https://github.com/rofinn/FilePathsBase.jl/pull/75#issuecomment-611310616


JeffBezanson:
Figured out a better way to do this. In general, we *do* want to save the entire state of Main, we just want to clean up the few bindings used by the bootstrap code in sysimg.jl.

Also fixes `eval` and `include` not being available during userimg.jl.

PR #35417, opened 2020-04-09
fix #34659, regression in showing empty arrays as matrix elements

### LINKED ISSUE #34659: Wrong display for arrays of empty arrays

JeffBezanson:
Fixes #34659, by just hacking it in.
This is an alternative to #34733.

Before:
```
julia> [Int[]]
 0-element Array{Int64,1}
```
After (same output as 1.4):
```
julia> [Int[]]
 []
```
If this is ok I'll add tests.

### DIFF
--- a/base/arrayshow.jl
+++ b/base/arrayshow.jl
@@ -320,6 +320,9 @@ print_array(io::IO, X::AbstractArray) = show_nd(io, X, print_matrix, true)
 # typeinfo aware
 # implements: show(io::IO, ::MIME"text/plain", X::AbstractArray)
 function show(io::IO, ::MIME"text/plain", X::AbstractArray)
+    if isempty(X) && (get(io, :compact, false) || X isa Vector)
+        return show(io, X)
+    end
     # 0) show summary before setting :compact
     summary(io, X)
     isempty(X) && return
@@ -421,7 +424,7 @@ _show_nonempty(io::IO, X::AbstractArray{T,0} where T, prefix::String) = print_ar
 
 # NOTE: it's not clear how this method could use the :typeinfo attribute
 _show_empty(io::IO, X::Array{T}) where {T} = print(io, "Array{", T, "}(undef,", join(size(X),','), ')')
-_show_empty(io, X) = nothing # by default, we don't know this constructor
+_show_empty(io, X::AbstractArray) = summary(io, X)
 
 # typeinfo aware (necessarily)
 function show(io::IO, X::AbstractArray)
@@ -443,7 +446,7 @@ function show_zero_dim(io::IO, X::AbstractArray{T, 0}) where T
         print(io, "fill(")
         show(io, X[])
     else
-        print(io, "Array{$T,0}(")
+        print(io, "Array{", T, ",0}(")
         show(io, undef)
     end
     print(io, ")")
--- a/base/abstractarray.jl
+++ b/base/abstractarray.jl
@@ -679,10 +679,10 @@ Create an empty vector similar to `v`, optionally changing the `eltype`.
 
 ```jldoctest
 julia> empty([1.0, 2.0, 3.0])
-0-element Array{Float64,1}
+Float64[]
 
 julia> empty([1.0, 2.0, 3.0], String)
-0-element Array{String,1}
+String[]
 ```
 """
 empty(a::AbstractVector{T}, ::Type{U}=T) where {T,U} = Vector{U}()
--- a/base/abstractset.jl
+++ b/base/abstractset.jl
@@ -208,7 +208,7 @@ julia> symdiff([1,2,1], [2, 1, 2])
  2
 
 julia> symdiff(unique([1,2,1]), unique([2, 1, 2]))
-0-element Array{Int64,1}
+Int64[]
 ```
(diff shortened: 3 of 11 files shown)


JeffBezanson:
Ah, so the same hack is needed for other container types:
```
julia> [Dict()]
1-element Array{Dict{Any,Any},1}:
 Dict{Any,Any} with 0 entries
```

I could add AbstractDict and AbstractSet to the hack of course. Or I think an alternative is to always use parseable syntax for empty containers, in other words
```
julia> Dict()
Dict{Any,Any}()
```

the idea being that the output `Dict{Any,Any} with 0 entries` doesn't add much value.

StefanKarpinski:
I think printing the empty versions of those with a reconstructable syntax is preferable.

JeffBezanson:
Done. I went a little bit inconsistent and decided that `3×0 Array{Float64,2}` is preferable to `Array{Float64}(undef,3,0)` though.

JeffBezanson:
Some examples:
```
julia> permutedims([zeros(3,2),zeros(2,2)])
1×2 Array{Array{Float64,2},2}:
 [0.0 0.0; 0.0 0.0; 0.0 0.0]  [0.0 0.0; 0.0 0.0]

julia> permutedims([zeros(3,0),zeros(2,0)])
1×2 Array{Array{Float64,2},2}:
 3×0 Array{Float64,2}  2×0 Array{Float64,2}

julia> permutedims([zeros(3),zeros(2)])
1×2 Array{Array{Float64,1},2}:
 [0.0, 0.0, 0.0]  [0.0, 0.0]

julia> permutedims([Int[],Int[]])
1×2 Array{Array{Int64,1},2}:
 []  []

julia> keys(Dict())
Base.KeySet for a Dict{Any,Any} with 0 entries

julia> [keys(Dict())]
1-element Array{Base.KeySet{Any,Dict{Any,Any}},1}:
 Base.KeySet for a Dict{Any,Any} with 0 entries

julia> permutedims([keys(Dict()), keys(Dict())])
1×2 Array{Base.KeySet{Any,Dict{Any,Any}},2}:
 []  []

julia> [undef,undef]
2-element Array{UndefInitializer,1}:
 UndefInitializer(): array initializer with undefined values
 UndefInitializer(): array initializer with undefined values

julia> permutedims([undef,undef])
1×2 Array{UndefInitializer,2}:
 UndefInitializer()  UndefInitializer()
```

tkf:
What'll happen to

```julia
julia> view(ones(3), 2:1)
0-element view(::Array{Float64,1}, 2:1) with eltype Float64
```

?  Is it going to be

`SubArray{Float64,1,Array{Float64,1},Tuple{UnitRange{Int64}},true}(Float64[])`

?

JeffBezanson:
That is a good question. Apparently, views don't have a reasonable `show` method:
```
julia> show(view(ones(3), 2:1))
Float64[]
```
so I'm not sure what to do.

JeffBezanson:
> Is it going to be
> 
> `SubArray{Float64,1,Array{Float64,1},Tuple{UnitRange{Int64}},true}(Float64[])`

Maybe --- the current output is:
```
julia> x=view(ones(3,3), 2:1,3:1);

julia> x
0×0 view(::Array{Float64,2}, 2:1, 3:2) with eltype Float64

julia> show(x)

```
that's right folks, the empty string. :scream: :scream:

JeffBezanson:
TIL people hate parseable output so much they would rather have NO output.

tkf:
Isn't empty string a valid Julia code?

base/arrayshow.jl
```diff
@@ -320,6 +320,7 @@ print_array(io::IO, X::AbstractArray) = show_nd(io, X, print_matrix, true)
 # typeinfo aware
 # implements: show(io::IO, ::MIME"text/plain", X::AbstractArray)
 function show(io::IO, ::MIME"text/plain", X::AbstractArray)
+    isempty(X) && ndims(X) == 1 && return show(io, X)
```
tkf, line 323: This is essentially what I suggested in #34733, but why not `get(io, :compact, false) === true && return show(io, X)`?

JeffBezanson: Hmm, there are tests for this, e.g.
```
replstr([x, x], :compact => true) == "2-element Array{Float64,1}:\n 3.14159\n 3.14159"
```
I guess the idea was to be able to keep the overall repl printing format but forcing elements to be shown compactly.

tkf: I see. I think this is another instance of the limitation for using `:compact` to mean "small width" and "single line."

rfourquet:
> 3×0 Array{Float64,2} is preferable to Array{Float64}(undef,3,0)

At least for numbers, what about `zeros(Float64, 3, 0)` ?

JeffBezanson:
That's a good idea, but I think it's a bit fiddly to need to know whether `zeros` would work for a certain type. I'll just leave it like this for now, future improvements welcome.

Issue #31389, opened 2019-03-18
Remove StridedArray restrictions on factorization objects.


dlfivefifty:
I'm going through the QR, Cholesky, etc. factorization codes as I re-write the factorizations in BandedMatrices.jl (and elsewhere) to be consistent with Base, and there are a large amount of unnecessary restrictions to `StridedArray`s. This is particularly silly as many of the factorizations have generic Julia fallbacks which do not use strided-ness at all.

I can do a PR when I get some time.

dlfivefifty:
And we also need to _add_ restrictions to `StridedArray` where LAPack is called, here is an example:

https://github.com/JuliaLang/julia/blob/8c0b550b61bcb196e1c8a9b03b38f0e6eb8a9dd7/stdlib/LinearAlgebra/src/cholesky.jl#L398

mbauman:
You may find #31149 helpful as you go through these — I use it semi-regularly when dealing with method lists that are ridiculous.  I've not merged it since nobody has reviewed or voiced their support for it.

andreasnoack:
> This is particularly silly as many of the factorizations have generic Julia fallbacks which do not use strided-ness at all.

I completely agree for some of the "outer" functions which only converts to the right element type but I have more mixed feelings for the actual implementations. While the generic implementations don't use the stridedness of `StridedArray`s explicitly, restricting to `StridedArray` ensures the elementwise access to the array is not prohibitively slow. Hitting a generic dense linear algebra routine with a sparse or distributed arrays is quite frustrating.

dlfivefifty:
We could overload for ` AbstractSparseArray ` to throw an error

andreasnoack:
> We could overload for AbstractSparseArray to throw an error

Something like might end up as the solution. GPUArrays and now also DistributedArrays have a flag to disallow scalar `getindex` to make it easier to detect when you hit `AbstractArray` implementations that fetch elements one-by-one.

Issue #17226, opened 2016-07-01
`schurfact!` fails on `StridedMatrix`


Jutho:
I noticed that the following fails:
```julia
B = randn(100,100);
A = view(B,1:50,1:50)
schurfact!(A)
```

The reason is that the `Schur` immutable assumes that `T` and `Z` are of the same type:
```julia
immutable Schur{Ty<:BlasFloat, S<:AbstractMatrix} <: Factorization{Ty}
    T::S
    Z::S
    values::Vector
    Schur(T::AbstractMatrix{Ty}, Z::AbstractMatrix{Ty}, values::Vector) = new(T, Z, values)
end
```
which is not the case, because `T` will be stored in the input matrix `A` (in this case a `SubArray`) whereas `Z` will be newly allocated in `LAPACK.gees!` using `similar` on `A`, which produces a regular `Array`. 

I could simply fix this by widening the definition of the `Schur` type (similar to how `Eigen` is currently defined), but wanted to consult the linear algebra/lapack experts (my apologies for pinging @andreasnoack , @kshyatt , @tkelman , @jiahao ) to see if that makes sense and whether any other plans or changes are in the pipeline that I could contribute to. 

For example, the type of `values` is abstract (`Vector`) so I could also change this (like it is in `Eigen`). However, the output of `schurfact` will still be type unstable because of this line in the output of `LAPACK.gees!` 
```julia
all(wi .== 0) ? wr : complex(wr, wi)
```
Would it be that bad to just always return complex values, with the advantages of having type stability? I noticed the current behaviour also appears in the various function related to `eigfact` etc. 

Another question/remark: I noticed that throughout `LAPACK`, new arrays are sometimes allocated with `Array{$elty}` and sometimes with `similar`. Given that the input array is always a `StridedMatrix`, I don't think `similar` can produce anything else than an `Array` (I checked for `SubArray` and for `SharedArray`). 

With some input, I would be happy to prepare a PR.

tkelman:
> Would it be that bad to just always return complex values, with the advantages of having type stability?

I've wanted to make this change for a while. x-ref #13362 where even the number of outputs differs between real and complex, which I think is confusing. I haven't revisited that PR in a long time though.

andreasnoack:
There are at least two somewhat separate problems here. I can see two possible solutions for the first problem of the types of `T` and `Z`. Each with some issues attached. 

[...]

PR #32147, opened 2019-05-26
AbstractVecOrMat, not StridedVecOrMat, in tridiagonal methods

stevengj:
It looks like several of the methods for tridiagonal methods were unnecessarily restricted to act on `StridedVecOrMat` rather than `AbstractVecOrMat`.

I noticed this because someone reported that `mul!` with a view was still allocating lots of memory (https://discourse.julialang.org/t/symtrididiagonal-matrices-performance-and-views-allocations)

PR #32120, opened 2019-05-23
Relax constrain on `lq` to allwo `AbstractMatrix`

oxinabox:
My own matrix type implements `lq!` and `copy`
I therefore assumed that `lq` would just work,
but it does not as it is types to only allow `StridedMatrix`, which is a (nonextensible) `Union`.

What do people think of this change?
It will still method error right now for most matrix types,
because the restriction is still in play for `lq!`

PR #35420, opened 2020-04-09
Remove overly restrictive `Strided*` constraints in LinearAlgebra

### LINKED ISSUE #713: Overly restrictive type constraints in LinearAlgebra
Many functions in `LinearAlgebra`  have the `::StridedVecOrMat` constraint. In many cases, it seems unnecessary. Just to list a few:

https://github.com/JuliaLang/julia/blob/8a5b3233a1fad479ed417345fc2b9de0a5d23c1d/stdlib/LinearAlgebra/src/lu.jl#L394-L397
https://github.com/JuliaLang/julia/blob/8a5b3233a1fad479ed417345fc2b9de0a5d23c1d/stdlib/LinearAlgebra/src/lu.jl#L339-L389

This causes problems like
```julia
julia> using LinearAlgebra
julia> A = rand(3, 3); x = rand(3)';
julia> ldiv!(lu(A), vec(x))
ERROR: MethodError: no method matching ldiv!(::LU{Float64,Array{Float64,2}}, ::Base.ReshapedArray{Float64,1,Adjoint{Float64,Array{Float64,1}},Tuple{}})
```

### LINKED ISSUE #35419: Overly restrictive type constraints in LinearAlgebra
Many functions in `LinearAlgebra`  have the `::StridedVecOrMat` constraint. In many cases, it seems unnecessary. Just to list a few:

https://github.com/JuliaLang/julia/blob/8a5b3233a1fad479ed417345fc2b9de0a5d23c1d/stdlib/LinearAlgebra/src/lu.jl#L394-L397
https://github.com/JuliaLang/julia/blob/8a5b3233a1fad479ed417345fc2b9de0a5d23c1d/stdlib/LinearAlgebra/src/lu.jl#L339-L389

This causes problems like
```julia
julia> using LinearAlgebra
julia> A = rand(3, 3); x = rand(3)';
julia> ldiv!(lu(A), vec(x))
ERROR: MethodError: no method matching ldiv!(::LU{Float64,Array{Float64,2}}, ::Base.ReshapedArray{Float64,1,Adjoint{Float64,Array{Float64,1}},Tuple{}})
```

YingboMa:
Fix JuliaLang/LinearAlgebra.jl#713

### DIFF
--- a/stdlib/LinearAlgebra/src/dense.jl
+++ b/stdlib/LinearAlgebra/src/dense.jl
@@ -13,7 +13,7 @@ const NRM2_CUTOFF = 32
 # This constant should ideally be determined by the actual CPU cache size
 const ISONE_CUTOFF = 2^21 # 2M
 
-function isone(A::StridedMatrix)
+function isone(A::AbstractMatrix)
     m, n = size(A)
     m != n && return false # only square matrices can satisfy x == one(x)
     if sizeof(A) < ISONE_CUTOFF
@@ -23,7 +23,7 @@ function isone(A::StridedMatrix)
     end
 end
 
-@inline function _isone_triacheck(A::StridedMatrix, m::Int)
+@inline function _isone_triacheck(A::AbstractMatrix, m::Int)
     @inbounds for i in 1:m, j in i:m
         if i == j
             isone(A[i,i]) || return false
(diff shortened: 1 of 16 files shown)


stdlib/LinearAlgebra/src/dense.jl
```diff
@@ -641,7 +641,7 @@ function rcswap!(i::Integer, j::Integer, X::StridedMatrix{<:Number})
 end
 
 """
-    log(A{T}::StridedMatrix{T})
+    log(A{T}::AbstractMatrix{T})
```
dkarrasch, line 644: Just came across this: the `A{T}` looks wrong here. Should be simply `A`?

mbauman:
So awesome!  Previously partially attempted in #32147 (and I think elsewhere, too).  Also looks like it fixes #31389.  Should also fix #17226, #18972, #32120, and #30097.

But the challenge is in all those ambiguities.  This is a bit of a rats nest.

Issue #34195, opened 2019-12-25
Julia 1.3 order of magnitude slowdown in function


jpjones76:
I'm noticing an order-of-magnitude slowdown in a custom function to parse "sample list" files (one ASCII value per line) to Float32 arrays. Any ideas what could be causing this?

Offending function here: https://github.com/jpjones76/SeisIO.jl/blob/master/src/Utils/Parsing/streams.jl

Affects my ASCII readers for SLIST and Lennartz SLIST files.

Before 1.3.0, `X[i] = stream_float(io, 0x00)` was 2-3x faster than `v = readline(io); X[i] = parse(Float32, v)`. Now `stream_float` is 4x slower than `readline` + `parse`. I actually need the old speeds if possible, so any alternate syntax suggestions would be welcomed.

MWE
```
using Pkg; Pkg.add("BenchmarkTools"); Pkg.add("SeisIO") 
run(`wget https://github.com/jpjones76/SeisIO-TestData/raw/master/SampleFiles/ASCII/0215162000.c00`)
using BenchmarkTools
import SeisIO:stream_float
io = open("0215162000.c00", "r")
nx = 225000
X = zeros(Float32, nx) 
@benchmark (seekstart($io); readline($io); @inbounds for i in 1:$nx; $X[i] = stream_float($io, 0x00); end)
```

Example outputs follow. I benchmarked this on a Dell laptop running Ubuntu 18.04 (bionic), kernel 5.0.0-37-generic, CPU Intel(R) Core(TM) i7-6820HQ CPU @ 2.70GHz, 16 GB RAM.

Julia 1.0.5
```
BenchmarkTools.Trial: 
  memory estimate:  272 bytes
  allocs estimate:  2
  --------------
  minimum time:     25.533 ms (0.00% GC)
  median time:      25.701 ms (0.00% GC)
  mean time:        25.867 ms (0.00% GC)
  maximum time:     30.592 ms (0.00% GC)
  --------------
  samples:          194
  evals/sample:     1
```

Julia 1.2.0:
```BenchmarkTools.Trial: 
  memory estimate:  272 bytes
  allocs estimate:  2
  --------------
  minimum time:     24.738 ms (0.00% GC)
  median time:      24.850 ms (0.00% GC)
  mean time:        24.910 ms (0.00% GC)
  maximum time:     27.172 ms (0.00% GC)
  --------------
  samples:          201
  evals/sample:     1
```

Julia 1.3.0:
```
[...]

PR #35426, opened 2020-04-10
add `lock` option to `open` to disable locking

JeffBezanson:
For #34195.

Tried reading a 240mb file a byte at a time:
julia 1.0: 1.82 sec
julia 1.4: 19 sec
this PR, lock=false: 2.6 sec

### DIFF
--- a/base/iostream.jl
+++ b/base/iostream.jl
@@ -16,8 +16,9 @@ mutable struct IOStream <: IO
     name::AbstractString
     mark::Int64
     lock::ReentrantLock
+    _dolock::Bool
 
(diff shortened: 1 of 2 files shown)


base/iostream.jl
```diff
@@ -16,8 +16,9 @@ mutable struct IOStream <: IO
     name::AbstractString
     mark::Int64
     lock::ReentrantLock
+    _dolock::Bool
```
StefanKarpinski, line 19: Why the underscore? Is this field somehow more private than others?

JeffBezanson: It's just me getting increasingly paranoid :)

StefanKarpinski: It's almost as if we need a way to actually prevent people from messing with private internals 😁

PR #35427, opened 2020-04-10
Add compat entry for @deprecated

tkf:
I think #35362 should have included the compat annotation and clarified the behavior of `@deprecated`.

https://github.com/JuliaLang/Pkg.jl/pull/1763 needs to be approved before this PR can be merged.


c42f:
Looks good to me!

PR #35433, opened 2020-04-11
Follow up to docstring changes in  #35427

oxinabox:
in #35427 I noticed that `default` was spelt as `defualt` 
when I was editting it, I also thought we could avoid saying `normal `julia` program` and be more explict.

### DIFF
--- a/base/deprecated.jl
+++ b/base/deprecated.jl
@@ -22,9 +22,9 @@ The first argument `old` is the signature of the deprecated method, the second o
 third argument is `false`.
 
 !!! compat "Julia 1.5"
-    As of Julia 1.5, functions defined by `@deprecate` do not print warning inside normal
-    `julia` program as the defualt value of `--depwarn` option is `no`.  The warnings
-    are printed from tests run by `Pkg.test()`.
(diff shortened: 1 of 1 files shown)


base/deprecated.jl
```diff
@@ -22,9 +22,9 @@ The first argument `old` is the signature of the deprecated method, the second o
 third argument is `false`.
 
 !!! compat "Julia 1.5"
-    As of Julia 1.5, functions defined by `@deprecate` do not print warning inside normal
-    `julia` program as the defualt value of `--depwarn` option is `no`.  The warnings
-    are printed from tests run by `Pkg.test()`.
+    As of Julia 1.5, functions defined by `@deprecate` do not print warning when `julia`
+    run without the `--depwarn=yes` flag set, as the default value of `--depwarn` option
```
rfourquet, line 26:
```suggestion
    runs without the `--depwarn=yes` flag set, as the default value of `--depwarn` option
```

oxinabox:
```suggestion
    is run without the `--depwarn=yes` flag set, as the default value of `--depwarn` option
```

PR #35448, opened 2020-04-12
base/range.jl: fix show(::LinRange)

### LINKED ISSUE #35437: `show(::LinRange)` output actually constructs a StepRangeLen
```
julia> show(LinRange(1,1,3))
range(1.0, stop=1.0, length=3)
julia> range(1.0, stop=1.0, length=3)
1.0:0.0:1.0

julia> typeof(ans)
```

anaveragehuman:
Fixes #35437

```julia
julia> show(LinRange(1,1,3))
LinRange(1.0, 1.0, 3)
julia> LinRange(1.0, 1.0, 3)
3-element LinRange{Float64}:
 1.0,1.0,1.0

julia> typeof(ans)
LinRange{Float64}
```

### DIFF
--- a/base/range.jl
+++ b/base/range.jl
@@ -424,12 +424,14 @@ _range(start::T, ::Nothing, stop::T, len::Integer) where {T<:Integer} =
 # for all other types we fall back to a plain old LinRange
 _linspace(::Type{T}, start::Integer, stop::Integer, len::Integer) where T = LinRange{T}(start, stop, len)
 
-function show(io::IO, r::LinRange)
-    print(io, "range(")
+function show(io::IO, r::LinRange{T}) where {T}
+    print(io, "LinRange{")
+    show(io, T)
+    print(io, "}(")
     show(io, first(r))
-    print(io, ", stop=")
+    print(io, ", ")
     show(io, last(r))
-    print(io, ", length=")
+    print(io, ", ")
     show(io, length(r))
     print(io, ')')
 end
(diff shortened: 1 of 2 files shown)


Keno:
Presumably this used to return a LinRange? What changed here to change the output type? Does `range` ever return a LinRange these days?

sostock:
The output of `show` should probably include the type parameter of `LinRange`. Otherwise, parsing the output of `show` can create a `LinRange` with different `eltype`, since the `LinRange` constructor (without type parameter) converts integers to floats:
```julia
julia> show(LinRange{Int}(1, 1, 3))
LinRange(1, 1, 3)

julia> typeof(LinRange(1, 1, 3))
LinRange{Float64}
```

timholy:
@Keno, it returns `StepRangeLen` for `IEEEFloat` subtypes or anything that gets converted to such via `float`. Everything else gets converted to `LinRange`.
```julia
julia> x = BigFloat(1.0)
1.0

julia> typeof(range(x, stop=x, length=3))
LinRange{BigFloat}

julia> x, y = rand(3), rand(3)
([0.20535507856391932, 0.5043860277330925, 0.7661816392701779], [0.027878031187783314, 0.28451886328180276, 0.26003320191090395])

julia> typeof(range(x, stop=y, length=3))
LinRange{Array{Float64,1}}
```
So the old form of printing could be used in any case where `!(T<:IEEEFloat)`.

PR #35451, opened 2020-04-13
Added tests for gcd and lcm for other integer data types.

ravibitsgoa:
Added tests for all signed and unsigned integer data types.
Improved the doctests of gcd and lcm.

Currently gcd(-5) = -5. Should we keep it that way or make it abs(-5) = 5?

### DIFF
--- a/base/intfuncs.jl
+++ b/base/intfuncs.jl
@@ -18,6 +18,21 @@ julia> gcd(6,9)
 
 julia> gcd(6,-9)
 3
+
+julia> gcd(6,0)
+6
+
+julia> gcd(0,0)
+0
+
+julia> gcd(1//3,2//3)
+1//3
+
+julia> gcd(1//3,-2//3)
+1//3
+
+julia> gcd(1//3,2)
+1//3
 ```
 """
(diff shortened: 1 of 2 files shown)


ravibitsgoa:
Can anyone review this PR which adds tests for previously untested integer data types for gcd and lcm?
Previous code tested gcd and lcm methods for only Int32 and Int64.

KristofferC:
This made different packages start to error with a method error for `checked_abs`. cc @StefanKarpinski

StefanKarpinski:
This puts us between a rock and a hard place: on is it better to break compatibility or give the wrong answer sometimes if there's overflow? One solution would be to add a `checked_abs(x::Number) = abs(x)` fallback, but I think that's too dangerous. Another would be to used
```jl
checked_abs_if_available(x::Number) = applicable(checked_abs, x) ? checked_abs(x) : abs(x)
```
for this code in gcd an lcm. That would allow it to keep working for types that haven't defined `checked_abs` but not foist that unsafeness on anyone who explicitly calls `checked_abs`. It might be even better to put a deprecation warning in the definition saying "please define checked_abs(::T)`.

PR #35453, opened 2020-04-13
Implement RefField <: Ref{T}

vtjnash:
Allows describing a GetElementPointer operation on a field of an object,
complementary to the existing ones for RefValue and RefArray.

### DIFF
--- a/base/refpointer.jl
+++ b/base/refpointer.jl
@@ -3,9 +3,10 @@
 """
     Ref{T}
 
-An object that safely references data of type `T`. This type is guaranteed to point to
-valid, Julia-allocated memory of the correct type. The underlying data is protected from
-freeing by the garbage collector as long as the `Ref` itself is referenced.
+An object that safely references data of type `T`. When constructed, it is
+therefore known to point to valid, Julia-allocated memory of the correct type.
+The underlying data is protected from freeing by the garbage collector as long
+as the `Ref` itself is referenced.
 
 In Julia, `Ref` objects are dereferenced (loaded or stored) with `[]`.
 
@@ -16,8 +17,8 @@ it can be written `Ref(a, i)` for creating a reference to the `i`-th element of 
 When passed as a `ccall` argument (either as a `Ptr` or `Ref` type), a `Ref` object will be
 converted to a native pointer to the data it references.
 
-There is no invalid (NULL) `Ref` in Julia, but a `C_NULL` instance of `Ptr` can be passed to
-a `ccall` Ref argument.
+There is no invalid (NULL) `Ref` in Julia, but a `C_NULL` instance of `Ptr` can
+be passed in C to a `ccall` Ref argument.
 
 # Use in broadcasting
 
@@ -74,7 +75,7 @@ RefArray(x::AbstractArray{T}, i::Int=1, roots::Nothing=nothing) where {T} = RefA
 convert(::Type{Ref{T}}, x::AbstractArray{T}) where {T} = RefArray(x, 1)
 
 function unsafe_convert(P::Type{Ptr{T}}, b::RefArray{T}) where T
-    if allocatedinline(T)
+    if allocatedinline(T) || T === Any
         p = pointer(b.x, b.i)
     elseif isconcretetype(T) && T.mutable
         p = pointer_from_objref(b.x[b.i])
@@ -84,10 +85,12 @@ function unsafe_convert(P::Type{Ptr{T}}, b::RefArray{T}) where T
     end
     return convert(P, p)
 end
-function unsafe_convert(P::Type{Ptr{Any}}, b::RefArray{Any})
-    return convert(P, pointer(b.x, b.i))
-end
 unsafe_convert(::Type{Ptr{Cvoid}}, b::RefArray{T}) where {T} = convert(Ptr{Cvoid}, unsafe_convert(Ptr{T}, b))
+
+getindex(b::RefArray) = b.x[b.i]
+setindex!(b::RefArray, x) = (b.x[b.i] = x; b)
+isassigned(b::RefArray) = isassigned(b.x, b.i)
+
 
 ###
 if is_primary_base_module
@@ -121,6 +124,8 @@ if is_primary_base_module
     Ref(x::AbstractArray, i::Integer) = RefArray(x, i)
 end
 
+###
+
 cconvert(::Type{Ptr{P}}, a::Array{<:Ptr}) where {P<:Ptr} = a
 cconvert(::Type{Ref{P}}, a::Array{<:Ptr}) where {P<:Ptr} = a
 cconvert(::Type{Ptr{P}}, a::Array) where {P<:Union{Ptr,Cwstring,Cstring}} = Ref{P}(a)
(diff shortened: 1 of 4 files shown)


base/refpointer.jl
```diff
@@ -136,12 +141,41 @@ unsafe_convert(::Type{Ptr{T}}, r::Ref{NTuple{N,T}}) where {N,T} =
 unsafe_convert(::Type{Ptr{T}}, r::Ptr{NTuple{N,T}}) where {N,T} =
     convert(Ptr{T}, r)
 
-###
+### Methods for a Ref object that is backed by a object field
+"""
+    RefField <: Ref{T}
+
+Create an object reflection adaptor that represents the properties of a specific object field.
```
vchuravy, line 148: Took me a while to parse this. Maybe

> Obtain a reference to a specific object field, by creating an object reflection adaptor that wraps the parent object and the properties of a specific object field.

test/ccall.jl
```diff
@@ -1581,6 +1581,67 @@ let
...
+    e::Complex{BigFloat}
+end
+mutable struct ManyFields2
+    a::Bool
+    b::Int64
+    c::Any
+    d::Integer
+    e::Complex{BigFloat}
+end
+@testset "RefField" begin
```
vchuravy, line 1599:
```suggestion
@testset "RefField struct" begin
```

test/ccall.jl
```diff
@@ -1581,6 +1581,67 @@ let
...
+    c_ptr = Base.unsafe_convert(Ptr{ManyFields}, c.x) + 16
+    d_ptr = Base.unsafe_convert(Ptr{ManyFields}, d.x) + 24
+    e_ptr = Base.unsafe_convert(Ptr{ManyFields}, e.x) + 32
+    @test Base.unsafe_convert(Ptr{Bool}, a) == a_ptr
+    @test Base.unsafe_convert(Ptr{Int64}, b) == b_ptr
+    @test Base.unsafe_convert(Ptr{Any}, c) == c_ptr
+    @test Base.unsafe_convert(Ptr{Integer}, d) == unsafe_load(Ptr{Ptr{Integer}}(d_ptr))
+    @test Base.unsafe_convert(Ptr{Complex{BigFloat}}, e) == unsafe_load(Ptr{Ptr{Integer}}(e_ptr))
+end
+@testset "RefField" begin
```
vchuravy, line 1622:
```suggestion
@testset "RefField mutable struct" begin
```

vchuravy (approved):

test/ccall.jl
```diff
@@ -1581,6 +1581,67 @@ let
...
+    @test a[] === true
+    @test b[] === 1021
+    @test c[] === "hello"
+    @test d[] === 5432
+    @test e[] == 1im
+    a_ptr = Base.unsafe_convert(Ptr{ManyFields}, a.x)
+    b_ptr = Base.unsafe_convert(Ptr{ManyFields}, b.x) + 8
+    c_ptr = Base.unsafe_convert(Ptr{ManyFields}, c.x) + 16
+    d_ptr = Base.unsafe_convert(Ptr{ManyFields}, d.x) + 24
+    e_ptr = Base.unsafe_convert(Ptr{ManyFields}, e.x) + 32
```
c42f, line 1615: Aren't these offsets system dependent?

base/refpointer.jl
c42f: Or, to focus more on the interface which people will actually use:

```julia
    RefField(object, field)

Obtain a reference to an individual field of an `object`, where the given `field` may be an index or a symbol (see also [`Base.fieldindex`](@ref)).

The returned `RefField <: Ref` is an object reflection adaptor that wraps the parent `object` and the properties of a specific object field.
```

c42f:
Looks very useful to me. One design question: is it possible to use this recursively to refer to nested fields? Reading the implementation I think not, but it would seem useful.

c42f:
> One design question: is it possible to use this recursively to refer to nested fields?

What if we made `f` an NTuple of `Int` to represent the "field path" inside a nested `struct`. Would that allow this to work fairly easily?

vchuravy:
> What if we made f an NTuple of Int to represent the "field path" inside a nested struct. Would that allow this to work fairly easily?

No I don't think so, it might be better to allow `RefField` to wrap other `Ref`s, which would allow us to chain.

tkf:
> it might be better to allow `RefField` to wrap other `Ref`s, which would allow us to chain.

I guess it'd be possible to chain to `RefArray`, too? It'd be nice to be able to mutate a field in, e.g., an array of namedtuples.

c42f:
> it might be better to allow `RefField` to wrap other `Ref`s, which would allow us to chain.

Yes I agree wrapping would be cleaner (especially considering interop with `ArrayRef`). But I wasn't sure it would work for nested immutables... there's a boundary somewhere in a chain of `Ref`s at the innermost mutable, after which wrapping seems difficult?

base/refpointer.jl
```diff
@@ -136,12 +141,41 @@ unsafe_convert(::Type{Ptr{T}}, r::Ref{NTuple{N,T}}) where {N,T} =
...
+"""
+struct RefField{f,T,S} <: Ref{T}
+    x::RefValue{S}
+    RefField{f,T,S}(x::S) where {f,T,S} = new{f::Int,T,S}(RefValue(x))
+end
```
tkf, line 153: How does this work with immutables?  For example, if I have:

```julia
x = (a=1, b=2)
ra = RefField(x, :a)
rb = RefField(x, :b)
ra[] = 10
rb[] = 20
```

After this code, I suppose `x === (a=1, b=2)` still holds?  How do I get `(a=10, b=20)`?  Is there going to be some compiler intervention?

tkf:
FYI here is my implementation of nestable Ref (for simple isbits case): https://discourse.julialang.org/t/alternative-to-mutable-named-tuple/38375/15

vtjnash:
Yeah, that code's kinda fundamentally flawed. I do need to look closer into how Ref ends up dealing with nesting though, and probably figure out how it should work before merging this.

JeffBezanson:
Supporting nesting seems pretty complex and can maybe be deferred; for example RefArray doesn't handle nested arrays either.

yuyichao:
> for example RefArray doesn't handle nested arrays either.

Arrays don't have inline storage/

JeffBezanson:
> Arrays don't have inline storage?

Some types of arrays do.

base/refpointer.jl
Gnimuc: Is this similar to C's address operator `&x.a`?

vtjnash:
yeah, I just stopped having a use case for this myself, and felt that it potentially also would ideally attempt to deal with nested pointers, for the full "mutating immutables" deal, but that is a hard problem.

StefanKarpinski:
Notes from triage: I'm personally in favor of having some version of this to round our our `RefBlah` APIs, although I'm not sure about this particular implementation. I don't think this should be mixed up in "mutating immutables" stuff and since it's no longer being considered for atomics, it's freed of that consideration, and can just be what it is: a way to pass around a reference to a field. If you try to mutate a field of an immutable through this, it should just fail.

PR #32968, opened 2019-08-20
Filter(f, tuple)

mcabbott:
Closes #30418

PR #35455, opened 2020-04-13
generalize filter[!] docstring and add compat annotation for Tuple

rfourquet:
The docstring for `filter(f, array)` was generic enough to be generally applicable (e.g. for `Set` and `Tuple`, which didn't have a corresponding docstring), so lift the type constraint from the signature.

`Filter(f, tuple)` was added in #32968.

### DIFF
--- a/base/array.jl
+++ b/base/array.jl
@@ -2363,10 +2363,13 @@ end
 ## Filter ##
 
 """
-    filter(f, a::AbstractArray)
+    filter(f, a)
 
-Return a copy of `a`, removing elements for which `f` is `false`.
+Return a copy of collection `a`, removing elements for which `f` is `false`.
 The function `f` is passed one argument.
+
+!!! compat "Julia 1.4"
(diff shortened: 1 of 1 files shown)


rfourquet:
Bump!

base/array.jl
```diff
@@ -2363,11 +2363,14 @@ end
 ## Filter ##
 
 """
-    filter(f, a::AbstractArray)
+    filter(f, a)
```
StefanKarpinski, line 2366:
```suggestion
    filter(f, itr)
```

rfourquet: > Does this support arbitrary iterables?

No, `Base.filter` is implemented only for specific types. So `itr` migth give the wrong idea here. But I would support having `filter` extended to arbitrary iterables, returning an `Vector` in this case, like `map` does.

StefanKarpinski: Ignore my suggestion then. If it changes to general iterables, we can update.

PR #35457, opened 2020-04-13
support abs2 for Dates.Period

visr:
Seems like an omission, since `abs` is defined. Currently gives a `MethodError`.

Came up in https://github.com/mcabbott/AxisKeys.jl/pull/11


StefanKarpinski (commented):
This seems to have the wrong units—a period of time squared does not have time as its unit, it has time squared as its unit.

fredrikekre:
I don't think this is correct, shouldn't it result in something with unit `Year^2`?

visr:
Hmm, I didn't consider that.

Unitful also squares the unit: https://github.com/PainterQubits/Unitful.jl/blob/7fb2dfb68c63c0a77b8c859c294d8b7f03f337d6/src/quantities.jl#L223-L224

PR #29595, opened 2018-10-11
Add Base.tail(::NamedTuple).

tpapp:
Fixes #29572.

PR #35458, opened 2020-04-13
Add Base.front(::NamedTuple)

maximerischard:
Exactly the same as  #29595 but for Base.front.


JeffBezanson:
Should this function be exported? From common lisp I'd tend to call this `butlast`. Does any other language use `front`?

rfourquet:
> Does any other language use front?

FWIW, Haskell also uses `tail`, but `init` for `Base.front` (not that I'm suggesting this name)

PR #35464, opened 2020-04-13
Add Varargs feature for the Kronecker Tensor Product

bfsmrgazph:
Add an efficient way of computing the Kronecker product of a set of matrices, vectors and/or integers.

### DIFF
--- a/stdlib/LinearAlgebra/src/dense.jl
+++ b/stdlib/LinearAlgebra/src/dense.jl
@@ -400,6 +400,52 @@ kron(a::AbstractVector, b::AbstractVector) = vec(kron(reshape(a ,length(a), 1), 
 kron(a::AbstractMatrix, b::AbstractVector) = kron(a, reshape(b, length(b), 1))
 kron(a::AbstractVector, b::AbstractMatrix) = kron(reshape(a, length(a), 1), b)
 
+_size_prod(x::Tuple{Integer,Integer},y::Tuple{Integer,Integer}) = x .* y
+_cum_prod(x) = begin
+        out = Array{Tuple}(undef,length(x))
+        out[1] = x[1]
+        for i in 2:length(x)
+                out[i] = _size_prod(out[i-1],x[i])
+        end
+        out
+end
+_getsize(a::Integer) = (1,1)
+_getsize(a::Vector) = (length(a),1)
+_getsize(a::Matrix) = size(a)
+
+function kron(A::Vararg{Union{AbstractVecOrMat,Number}})
+        size_src = _getsize.(A)
+        in_r  = _cum_prod(reverse(size_src))
+        out_r = _cum_prod(size_src)
+        dest = repeat(A[1],inner=in_r[end-1])
+        _kron!(dest,A,in_r,out_r)
+end
+
+function _kron!(dest,srcs,in_r,out_r)
+        for i in 2:length(srcs)
+                ir = i == length(srcs) ? (1,1) : in_r[end-i]
+                _kron_step!(dest,srcs[i],ir,out_r[i-1])
+        end
+        return dest
+end
+
+function _kron_step!(dest,B,in_r,out_r)
+        m=0
+        for _ in 1:out_r[2], l in 1:size(B,2)
+                for _ in 1:out_r[1], k in 1:size(B,1)
+                        f = dest[m+=1] *= B[k,l]
+                        for _ in 2:in_r[1]
+                                dest[m+=1]=f
+                        end
+                end
+
+                for _ in 2:in_r[2]
+                        copyto!(dest,m+1,dest,m+1-size(dest,1),size(dest,1))
+                        m+= size(dest,1)
+                end
+        end
+end
+
 # Matrix power
 (^)(A::AbstractMatrix, p::Integer) = p < 0 ? power_by_squaring(inv(A), -p) : power_by_squaring(A, p)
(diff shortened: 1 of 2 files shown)


dkarrasch:
I don't think you need all the code to do what you want to do. This is already working, your tests are passing in Julia v1.0 and v1.4.

dkarrasch:
Oh, or are you saying your code is more performant in any way? Could you provide some benchmark cases?

bfsmrgazph:
With the current implementation of `kron`, one have to call kron(A,kron(B,kron(C,kron(D,E)))) or reduce(kron,(A,B,C,D,E)) which is not very convenient and allocate new matrices for each call of `kron`.

This PR allows one to write `kron(set_mat...)` and becomes way more efficient for big matrices:

`julia> @time reduce(kron,(randn(100,200),randn(7,9),randn(8,9),randn(3,2)));
 14.761389 seconds (460.00 M allocations: 18.450 GiB, 5.70% gc time)

julia> @time kron(randn(100,200),randn(7,9),randn(8,9),randn(3,2));
  3.689564 seconds (100.04 k allocations: 4.059 GiB, 0.05% gc time)`

For small matrices, it is as efficient as the older kron:

`julia> @time reduce(kron,(randn(3,3),randn(3,3),1));
  0.000049 seconds (91 allocations: 4.797 KiB)

julia> @time kron(randn(3,3),randn(3,3),1);
  0.000033 seconds (80 allocations: 3.766 KiB)

This PR could actually replace the old `kron`
`
For sure there is some way of reducing the code's length, especially the part on the cumulative product of the sizes of the matrices (`_get_size` , `_size_prod`, `_cum_prod`).

The idea is to use the forward and reverse cumulative product of the matrices sizes to get the outer and inner redundancy structure of the computation of the Kronecker Tensor Product. It gives  `_kron_step!` which is the core of the solution proposed

mcabbott:
I was wondering how this compared to an approach based on just broadcasting reshaped matrices, which is nice in that it's quite generic. Here's the version I was fiddling with:
```julia
function rkron4(As::Vararg{<:AbstractArray{<:Any,N},L}) where {N,L}
    NL = times(Val(N), Val(L))
    Bs = ntuple(L) do i
        sz = ntuple(d -> rem(d-1,L)+1==i ? size(As[i], div(d-1,L)+1) : 1, NL) #N*L)
        reshape(As[i], sz)
    end
    C = .*(Bs...)
    reshape(C, ntuple(d -> prod(map(A -> size(A,d), As)), Val(N)))
end
times(::Val{N}, ::Val{L}) where {N,L} = N*L

A, B, C = rand(3,3), rand(2,2), rand(4,4);
kron(A,B,C) ≈ rkron4(C,B,A) # it's reversed, sorry
@code_warntype rkron4(A,B,C) # now ok
```
However what I think you're doing differently is slowly colouring in the output array, multiplying once and then copying that result to all the places which should (at that stage) have the same result. Which might be faster. Here's what I get, testing both with the largest array first and last. (Note that my reshape version makes things the opposite order to `kron`.)
```julia
julia> Rs = (randn(10,20),randn(7,9),randn(8,9),randn(3,2)); # smaller numbers

julia> @btime reduce(kron, $Rs); # reshape
  8.579 ms (6 allocations: 48.55 MiB)

julia> @btime pkron($Rs...); # PR
  21.242 ms (1033 allocations: 41.56 MiB)

julia> @btime rkron4($Rs...); # reshape
  35.201 ms (17 allocations: 41.53 MiB)


julia> @btime reduce(kron, reverse($Rs)); # the same, now in reversed order
  5.787 ms (5 allocations: 41.74 MiB)

julia> @btime pkron(reverse($Rs)...); 
  19.350 ms (63 allocations: 41.53 MiB)

julia> @btime rkron4(reverse($Rs)...);
  80.204 ms (17 allocations: 41.53 MiB)


julia> Rs = (randn(100,200),randn(7,9),randn(8,9),randn(3,2)); # larger, as above

julia> @time reduce(kron, Rs);                                       
  9.616191 seconds (6 allocations: 4.741 GiB, 6.48% gc time)

julia> @time pkron(Rs...);                                           
 13.703519 seconds (100.03 k allocations: 4.058 GiB, 2.85% gc time)

julia> @time rkron4(Rs...);                                          
  8.547320 seconds (17 allocations: 4.056 GiB, 6.89% gc time)


julia> @time reduce(kron, reverse(Rs)); # the same, now in reversed order
  6.779157 seconds (35 allocations: 4.056 GiB, 16.89% gc time)

julia> @time pkron(reverse(Rs)...);                                  
 12.287366 seconds (64 allocations: 4.056 GiB, 0.91% gc time)

julia> @time rkron4(reverse(Rs)...);                                 
 16.968383 seconds (18 allocations: 4.056 GiB, 0.71% gc time)
```

dkarrasch:
> With the current implementation of kron, one have to call kron(A,kron(B,kron(C,kron(D,E)))) or reduce(kron,(A,B,C,D,E)) which is not very convenient and allocate new matrices for each call of kron.

I'm not sure I follow here. You can simply call `kron(A, B, C, D)`, no need to call `reduce`. I think this is then performed as `kron(A, kron(B, kron(C, D)))`, unless one has another implementation that is more specific in terms of multiple dispatch. So this PR should be about performance only, not convenience.

DilumAluthge:

@bfsmrgazph Are you still interested in working on this PR?

adienes:
needs to move to `LinearAlgebra.jl` anyway

PR #33682, opened 2019-10-26
Add test-revise-* Makefile targets

Keno:
These targets use Revise to load any modifications to base into the
process before running tests. This is very handy when editing Base
and wanting to run the appropriate tests as a sanity check but
without waiting 3 minutes for a whole system image rebuild. Of course
it was previously possible to do this through the REPL by doing the
3-4 steps manually, but I was generally too lazy too do that, so
here's the automatic option.

Also, if we detect that you just rebuilt your sysimage before the test- command is run, you get a friendly ad for the new functionality:
Screen Shot 2019-10-26 at 12 49 33 AM

PR #35465, opened 2020-04-13
Include stdlibs in --revise

Keno:
Without this changes in the stdlibs weren't picked up.


vtjnash (approved):

timholy:
I should have noticed this when I reviewed  #33682, sorry 'bout that. Good fix.

PR #35466, opened 2020-04-13
Minor LibGit2 cleanup

Keno:
The only user facing change is adding `haskey(::GitTree, path)` to go along with getindex, but the internals are refactored.


stdlib/LibGit2/src/tree.jl
```diff
@@ -122,14 +128,13 @@ tree_entry = tree[1]
 blob = LibGit2.GitBlob(tree_entry)
 ```
 """
-function GitObject(e::GitTreeEntry) end
```
Keno, line 125: This function was declaring this signature to return nothing, which is obviously wrong. How do I declare it to just document the signature? @fredrikekre or @KristofferC do you know?

yuyichao: just `GitObject(e::GitTreeEntry)` I believe.

Keno: Thanks.

simonbyrne (approved):

PR #35474, opened 2020-04-14
Replace conditional data dump by a triage function in choosetests.jl

o314:
*Following #35209 (fork unavailable from now)*

This patch implements the comment at https://github.com/JuliaLang/julia/pull/35209#discussion_r407955751


test/choosetests.jl
```diff
@@ -82,59 +82,27 @@ function choosetests(choices = [])
...
-
-    stringtests = ["strings/basic", "strings/search", "strings/util",
-                   "strings/io", "strings/types"]
-    if "strings" in skip_tests
-        filter!(x -> (x != "strings" && !(x in stringtests)), tests)
-    elseif "strings" in tests
-        # specifically selected case
-        filter!(x -> x != "strings", tests)
-        prepend!(tests, stringtests)
+    function patchlib!(tests, name, files=[name])
```
StefanKarpinski, line 85: Calling this `patchlib!` is a little weird. That makes me think of patchelf and other such library patching utils.

o314: 1. probably not the best name one can choose. first choice was the last one o:

2. functional prg @ code leads to delta-oriented programming @ data. 
    diff, patch etc can spread and disseminate. that's great!

3. no problem at all to change this name. 
    i will try to get inspiration for that. hum, hum.

o314:
a few couple of work outdoor and i will go back to fix that.

test/choosetests.jl
rfourquet: The new name `filtertests!` is indeed more descriptive!

rfourquet:
Thanks!

PR #35475, opened 2020-04-14
Fix inference for mapreduce on skipmissing

nalimilan:


JeffBezanson:
Looks like this needs to be rebased.

PR #35483, opened 2020-04-14
Add specialized * for Rational and Integer

garborg:
Piggybacking off of #35444. Seeing that PR, I realized I have code that would be helped if multiplication got the same treatment. Trying following benchmark:

```julia
function mul_spec(x::Rational, y::Integer)
    xd,yn = Base.divgcd(x.den,y)
    Base.checked_mul(x.num,yn) // xd
end
mul_spec(x::Integer, y::Rational) = mul_spec(y, x)

function mk_rats(T, n)
    r = one(T):T(1000)
    Rational.(rand(r, n), rand(r, n))
end

function bench(f, rats)
    r = 1:length(rats)
    maximum(
        f(i, f(i2, f(f(rat, i2), i))).den
        for (rat, i, i2)
        in zip(rats, r, r .+ 1)
    )
end

n = 1_000
for T in (UInt, Int, BigInt)
    rats = mk_rats(T, n)
    @info "Type" T
    @btime bench($(*), $rats)
    @btime bench($mul_spec, $rats)
end
```
I see the following speedups:

```
┌ Info: Type
└   T = UInt64
  717.044 μs (2 allocations: 64 bytes)
  480.512 μs (2 allocations: 64 bytes)
┌ Info: Type
└   T = Int64
  753.683 μs (2 allocations: 64 bytes)
  485.841 μs (2 allocations: 64 bytes)
┌ Info: Type
└   T = BigInt
  4.653 ms (145006 allocations: 2.78 MiB)
  2.555 ms (80002 allocations: 1.53 MiB)
```

There's still a redundant `divgcd` in Rational's inner constructor.

### DIFF
--- a/base/rational.jl
+++ b/base/rational.jl
@@ -280,7 +280,7 @@ function *(x::Rational, y::Rational)
     checked_mul(xn,yn) // checked_mul(xd,yd)
 end
 function *(x::Rational, y::Integer)
-    xd,yn = divgcd(x.den,y)
+    xd, yn = divgcd(x.den, y)
     checked_mul(x.num,yn) // xd
 end
 *(x::Integer, y::Rational) = *(y, x)
--- a/base/rational.jl
+++ b/base/rational.jl
@@ -279,6 +279,11 @@ function *(x::Rational, y::Rational)
     xd,yn = divgcd(x.den,y.num)
     checked_mul(xn,yn) // checked_mul(xd,yd)
 end
+function *(x::Rational, y::Integer)
+    xd,yn = divgcd(x.den,y)
+    checked_mul(x.num,yn) // xd
+end
+*(x::Integer, y::Rational) = *(y, x)
 /(x::Rational, y::Rational) = x//y
 /(x::Rational, y::Complex{<:Union{Integer,Rational}}) = x//y
 inv(x::Rational) = Rational(x.den, x.num)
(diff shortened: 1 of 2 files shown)


base/rational.jl
```diff
@@ -279,6 +279,11 @@ function *(x::Rational, y::Rational)
     xd,yn = divgcd(x.den,y.num)
     checked_mul(xn,yn) // checked_mul(xd,yd)
 end
+function *(x::Rational, y::Integer)
+    xd,yn = divgcd(x.den,y)
```
StefanKarpinski, line 283:
```suggestion
    xd, yn = divgcd(x.den, y)
```

base/rational.jl
```diff
@@ -279,6 +279,11 @@ function *(x::Rational, y::Rational)
     xd,yn = divgcd(x.den,y.num)
     checked_mul(xn,yn) // checked_mul(xd,yd)
 end
+function *(x::Rational, y::Integer)
+    xd, yn = divgcd(x.den, y)
+    checked_mul(x.num,yn) // xd
```
StefanKarpinski, line 284:
```suggestion
    checked_mul(x.num, yn) // xd
```

StefanKarpinski (approved):

base/rational.jl
garborg: Thanks for cleaning up my misguided attempt to 'mirror surrounding syntax'!

Issue #35472, opened 2020-04-14
Bad color (magenta) in REPL


PallHaraldsson:
It's very difficult to see on black background.

A. At least for "x \times 2^n." (and "ldexp(x, n)" could also use better color), as seen with:

`help?> ldexp
`

B. I think " \times" should also substitute, as it's LaTeX, while I'm less concerned personally with that.

Keno:
Change your terminal color scheme?

Keno:
(The problem here is that the colors look different for everybody, so if we change it to something that looks fine on your terminal, it'll start looking really bad on somebody else's terminal)

JeffBezanson:
Maybe we just use too many different colors. Does math need to be purple? Maybe bold instead?

PallHaraldsson:
I mean this a bad on a black background, what I have in the terminal, and I think pretty much done (at least default in Ubuntu) and in most editors.

I'm not up to speed on "terminal color scheme" if it's about the background color or something else, and I doubt most users will neither know. I think we should have a sane default for the most common setup, and others can change.

Keno:
The problem is that unless we ship our own terminal emulator on every OS (which I'm not entirely opposed to), there's no way to pick a sane default, because whatever we pick will look different for everybody. For example, the purple is perfectly readable on my terminal, even with a black background:

![Screen Shot 2020-04-15 at 9 45 55 AM](https://user-images.githubusercontent.com/1291671/79344569-00e4a980-7efe-11ea-99af-43eb545afe97.png)
 
Blue isn't particularly readable though.

KristofferC:
Tangentially, windows updated their default colors not too long ago https://devblogs.microsoft.com/commandline/updating-the-windows-console-colors/.

PallHaraldsson:
I would suggest changing to the brighter colors, i.e. from purple ("Palette entry 5") to the lighter purple "Palette entry 13" (and e.g. 6->14). I'm just unsure where to find this in the code to make a PR (I could check).

I first found "Use colors from the system theme" checked (and I don't think I've ever changed it or to more custom) in the terminal, and toggling the switch doesn't make much difference.

Users redefining the palette, might like me first change the lighter purple to no effect, end even with finding the right color seems like a hassle, and not recommend (would screw up other terminal work? or make it better too...).

[Only changing the background to white DID make for readable, while still also an awful (while less so) scheme.]

PallHaraldsson:
[...]

PR #35488, opened 2020-04-15
Better color scheme

PallHaraldsson:
Fixing #35472


PallHaraldsson:
Not yet fixed, I changed the wrong file it seems, and if that one should also be changed then magenta is in two places.

StefanKarpinski:
We've been down this path before. The only color scheme that works well everywhere is no color scheme. The current colors work fine in a black terminal which seems to be what most people have.

Issue #35503, opened 2020-04-17
Duplicated example in document of Multi-dimensional Arrays.


newptcai:
There is a duplicated example here

https://github.com/JuliaLang/julia/blame/master/doc/src/manual/arrays.md#L139

https://github.com/JuliaLang/julia/blame/master/doc/src/manual/arrays.md#L146

OkonSamuel:
@StefanKarpinski  or @fredrikekre  since this has been resolved. I belive this can be closed

PR #35505, opened 2020-04-17
Remove duplicate example in array documentation

michiboo:
Quick PR to address Duplicated example in document of Multi-dimensional Arrays. #35503


fredrikekre (approved):
Thank you!

PR #35508, opened 2020-04-17
Add missing `image_file_specified` field in `Base.JLOptions`

staticfloat:
It's a good thing the `warn_scope` option isn't consumed by Julia-side code and is instead used by the C code that isn't sensitive to this.

### DIFF
--- a/base/options.jl
+++ b/base/options.jl
@@ -42,6 +42,7 @@ struct JLOptions
     outputji::Ptr{UInt8}
     output_code_coverage::Ptr{UInt8}
     incremental::Int8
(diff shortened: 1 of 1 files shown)


StefanKarpinski:
Would it be possible to test?

Keno:
We should assert during bootstrap that the size matches (by having a C function that returns the C sizeof, calling that and making sure the julia side has the same size).

KristofferC:
I thought we did that somewhere... I recall getting a compiler error from this somewhere..

vtjnash (approved):

PR #20531, opened 2017-02-08
Merge deleteat() into delete!() and deprecate it

nalimilan:
The generic definition of delete!() suits deleteat!() very well, and
the signatures do not conflict.

I have kept the internal `_deleteat!` function in `base/array.jl` since it mirrors `jl_array_del_at` from `src/array.c`, which itself is part of a consistent C API, with `jl_array_grow_at`, `jl_array_grow_at_end`, `jl_array_grow_at_beg`, `jl_array_del_at_beg` and `jl_array_del_at_end`.

I wasn't sure where to put the generic docstring, suggestions welcome. In particular, it would be great to find a way to have it appear before the more specific ones (which wasn't the case either before moving it out of helpdb).

PR #35513, opened 2020-04-18
add pop!(vector, idx, [default])

rfourquet:
This removes and return an element at an arbitrary position of a vector, and seems to be compatible with the docstring for `pop!(collection, key[, default])`.

Python has it, and I needed it today, which seemed good enough arguments for me to add this :)

### DIFF
--- a/base/array.jl
+++ b/base/array.jl
@@ -1129,6 +1129,21 @@ function pop!(a::Vector)
     return item
 end
 
+function pop!(a::Vector, i::Integer)
+    x = a[i]
+    _deleteat!(a, i, 1);
+    x
+end
+
+pop!(a::Vector, i::Integer, default) =
+    if 1 <= i <= length(a)
+        x = @inbounds a[i]
+        _deleteat!(a, i, 1)
+        x
+    else
+        default
+    end
+
 """
     pushfirst!(collection, items...) -> collection
 
--- a/base/array.jl
+++ b/base/array.jl
@@ -1131,11 +1131,11 @@ end
 
 function pop!(a::Vector, i::Integer)
     x = a[i]
-    _deleteat!(a, i, 1);
+    _deleteat!(a, i, 1)
     x
 end
 
-pop!(a::Vector, i::Integer, default) =
+function pop!(a::Vector, i::Integer, default)
     if 1 <= i <= length(a)
         x = @inbounds a[i]
         _deleteat!(a, i, 1)
@@ -1143,6 +1143,7 @@ pop!(a::Vector, i::Integer, default) =
     else
         default
     end
+end
 
 """
     pushfirst!(collection, items...) -> collection
--- a/NEWS.md
+++ b/NEWS.md
@@ -122,6 +122,7 @@ New library features
 * `x::Signed % Unsigned` and `x::Unsigned % Signed` are supported for integer bitstypes.
 * `signed(unsigned_type)` is supported for integer bitstypes, `unsigned(signed_type)` has been supported.
 * `accumulate`, `cumsum`, and `cumprod` now support `Tuple` ([#34654]) and arbitrary iterators ([#34656]).
+* `pop!(collection, key, [default])` now has a method for `Vector` to remove an element at an arbitrary index ([#35513]).
 * In `splice!` with no replacement, values to be removed can now be specified with an
   arbitrary iterable (instead of a `UnitRange`) ([#34524]).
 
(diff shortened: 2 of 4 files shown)


base/array.jl
```diff
@@ -1129,6 +1129,21 @@ function pop!(a::Vector)
     return item
 end
 
+function pop!(a::Vector, i::Integer)
+    x = a[i]
+    _deleteat!(a, i, 1);
+    x
```
dkarrasch, line 1135: Maybe
```suggestion
    _deleteat!(a, i, 1)
    return x
```
?

base/array.jl
```diff
@@ -1129,6 +1129,21 @@ function pop!(a::Vector)
...
+end
+
+pop!(a::Vector, i::Integer, default) =
```
KristofferC, line 1138: Boring comment but this looks like it should be a full length function?

rfourquet: Ok

base/array.jl
rfourquet: Ok, I will make a compromise and remove the useless `;` while not adding `return` ;-) 
Adding `return` in long function form is not yet officially recommended and `Base` is full of cases without `return`, so I will contribute to support this lovely style here :)

rfourquet:
I'm having after-thougths on this one, because of the `delete!` vs `deleteat!` distinction, cf. #20531. Should this be called `popat!`? 
As @nalimilan put it, "The generic definition of delete!() suits deleteat!() very well", similarly here the definition of `pop!` suits well the new method for `vector`, but the "subsequent items are shifted".

Also, I remembered that `pop!(vector, idx)` is equivalent to `splice!(vector, idx)`, and `pop!(vector, idx, default)` seems less generally useful. I still find that `pop!(vector, idx)` looks nice to have (it's more discoverable), but not at the cost of being inconsistent...

nalimilan:
It's an interesting case. `?pop!` says "Remove an item in collection and return it. If collection is an ordered container, the last item is returned.", so it implies an order. Deleting the last element in an array doesn't affect the mapping of other keys/indices to values, contrary to other elements, so that's not different from calling `pop!(dict, key)`. OTC, allowing to specify an index removes that guarantee, and would indeed justify using `popat!` instead, just like we have both `delete! and `deletea!`.

(I'm actually not completely convinced that distinction is essential, but that's our API at least until 1.x.)

rfourquet:
Note that there is a separate docstring for ` pop!(collection, key[, default])` though:

>  Delete and return the mapping for key if it exists in collection, otherwise return default, or throw an error if default is not specified.

It can easily be interpreted as working for `vector`, the unlear point being the fact that after deletion, there will be a new mapping for `key` because remaining elements are shifted.

lesshaste:
This is one of those features that arguably should not exist in Python. The performance is linear where many coders expect constant time.

rfourquet:
> The performance is linear where many coders expect constant time for a pop operation.

I find this totally subjective. There are situtation when you need an O(n) operation, and this is such a case. I'm not against finding another name for this operation, e.g. like I suggested `popat!` (for another reason). But I expect a coder to know that different data structures have different complexity trade-offs, and I don't see why a coder would expect `pop!(vector, idx)` to be O(1) but expect `deleateat!(vector, idx)` to be O(n).

lesshaste:
Am I right that it is O(1) if popping from the end?  

It is subjective as you say. I was just reporting my experience teaching python coding.

rfourquet:
> Am I right that it is O(1) if popping from the end?

Yes fortunately :)

lesshaste:
That is what causes the confusion. Lots of python students at least just can't imagine how these things are implemented.  To be fair to them, I don't know if it is O(1) time if popping the second last element without reading the source. It could be O(1) if removing the first element too but I assume not.

StefanKarpinski:
From triage: we're not that concerned about the performance issue but we're pretty concerned about the semantic distinction between what `pop!(dict, key)` and `pop!(vec, idx)` do, which is analogous to the distinction between `delete!(dict, key)` and `deleteat!(vec, idx)`—namely that `delete!(dict, key)` has no affect on the other `(key, val)` mappings in `dict` whereas `deleteat!(vec, idx)` implicitly changes all `(idx, val)` pairs following the given `idx` value in `vec` to `(idx-1, val)`. Of course, that semantic difference is precisely _why_ there's O(n) performance, but if performance was the _only_ difference we wouldn't care much about this.

JeffBezanson:
Given that this is on a release branch already, probably best to rename it to `popat!` instead of removing it completely?

PR #35518, opened 2020-04-18
Reduce allocations when deserializing svec

sjkelly:
This slightly reduces the allocations when deserializing SimpleVectors. It seems to have a decent net performance benefit when `using Plots`.


JeffBezanson:
How does Plots use the serializer?

sjkelly:
Not directly, but from what I can tell it is used to communicate with the GR backend.

JeffBezanson:
Ah, interesting. Any example speedups?

JeffBezanson (approved):

sjkelly:
I want to get a better grasp on this. I will redo the PR with timing results and some other changes soon.

Issue #35517, opened 2020-04-18
Feature request: Quarter


matthieugomez:
Could there be a `Quarter <:  DatePeriod` type in Dates?

StefanKarpinski:
Would that be different from 3 months in any way? My understanding of a quarter is that it's not a duration so much as a fixed splitting of each year. I.e. it's weird to say "a quarter later" if you mean "three months from today"; rather, you refer to the first, second, third and fourth quarters of a given year to refer to a period of time.

matthieugomez:
Public firms must report earnings every quarter — they do say "next quarter/last quarter".
Not all quarters start in January (for instance Costco's first quarter for their fiscal year starts in September). A lot of accounting datasets are at a quarterly frequency, so defining `Quarter` makes it easier to work with these datasets (for instance: what is the growth of earnings between this quarter and next quarter etc...)

It is true that a quarter is exactly three months. But then a week is exactly seven days, and we still define  `Week <: TimePeriod` as a period.

StefanKarpinski:
👍 The PR looks non-trivial enough to make sense.

matthieugomez:
Solved by https://github.com/JuliaLang/julia/pull/36364

PR #35519, opened 2020-04-18
add Quarter period

matthieugomez:
Similarly to the fact that a `Week` corresponds to seven days, a `Quarter` corresponds to three months. @quinnj 

https://github.com/JuliaLang/julia/issues/35517


matthieugomez:
I have not touched the `quarterofyear` function but maybe it should be deprecated (since it is equivalent to `quarter`).

quinnj (approved):
LGTM; no support for parsing/formatting, but it's not a very common use-case anyway.

matthieugomez:
For now, I don't export `quarter`. This is a bit weird: accessor functions are exported for all other Periods . The reason I don't do this is that there is already a function, `quarterofyear`, that does exactly the same thing. 

So, should I still export `quarter`? If yes, should I deprecate `quarterfoyear`?

quinnj:
Looks like this needs a rebase and can be merged

PR #35521, opened 2020-04-19
Sockets: Add UDP multicast group support

### LINKED ISSUE #21300: Please add a function to join an UDP multicast group
Joining a UDP multicast group is not yet supported by Julia.
Probably it requires just one operation system specific function call to set the correct options.
We really need this!
Uwe

staticfloat:
Adds two functions `join_multicast_group()` and `leave_multicast_group` for usage with UDP multicast groups.  Closes https://github.com/JuliaLang/julia/issues/21300

### DIFF
--- a/stdlib/Sockets/src/Sockets.jl
+++ b/stdlib/Sockets/src/Sockets.jl
@@ -22,6 +22,8 @@ export
     recv,
     recvfrom,
     send,
+    join_multicast_group,
+    leave_multicast_group,
     TCPSocket,
     UDPSocket,
     @ip_str,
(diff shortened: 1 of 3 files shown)


stdlib/Sockets/test/runtests.jl
```diff
@@ -412,23 +412,55 @@ end
...
         end
         [close(s) for s in [a, b, c]]
+
+        # Test ipv6 broadcast groups
+        a, b, c = [create_socket(ip"::", port) for i in 1:3]
+        try
+            # Exemplary Interface-local ipv6 multicast group, if we wanted this to actually be routed
+            # to other computers, we should use a link-local or larger address scope group
+            # bsd family do not allow broadcasting on loopbacks
+            @static if !Sys.isbsd() || !Sys.isapple()
```
tanmaykm, line 440: Not sure if the condition here is correct. Is that why freebsd64 test failed?

stdlib/Sockets/src/Sockets.jl
```diff
@@ -22,6 +22,8 @@ export
     recv,
     recvfrom,
     send,
+    join_multicast_group,
+    leave_multicast_group,
```
vtjnash, line 26: Nodejs calls these "addMembership/addSourceSpecificMembership" and "dropMembership/dropSourceSpecificMembership" (https://nodejs.org/dist/latest-v14.x/docs/api/dgram.html)

vtjnash (approved):

stdlib/Sockets/src/Sockets.jl
staticfloat: Yeah, I see a lot of different names around.  I figured that these names were as self-explanatory as possible.

stdlib/Sockets/test/runtests.jl
staticfloat: Ah, yes, I think so!

ufechner7:
Great job! :)

PR #35522, opened 2020-04-19
Add introspection macros support for dot syntax

### LINKED ISSUE #14637: `@which` does not work with fully qualified "symbol"
```
julia> @which Base.sin
ERROR: ArgumentError: argument is not a generic function
 in methods at reflection.jl:140
 in which at ./reflection.jl:292
```

expected outcome

```
julia> @which Base.sin
Base
```

### LINKED ISSUE #20620: inferred macro has trouble with broadcasted operators?
I came across this behavior in my *StaticArrays* tests:

```julia
julia> v = [1,2,3]
3-element Array{Int64,1}:
 1
 2
 3

julia> .-v
3-element Array{Int64,1}:
 -1
 -2
 -3

julia> @inferred(.-v)
ERROR: AssertionError: ($(Expr(:globalref, Base.Test, :(==))))(($(Expr(:globalref, Base.Test, :length)))(#46#inftypes), 1)

julia> b = [true,false]
2-element Array{Bool,1}:
  true
 false

julia> .!b
2-element BitArray{1}:
 false
  true

julia> @inferred(.!b)
ERROR: UndefVarError: .! not defined
```

### LINKED ISSUE #28615: Making introspection macros work for broadcasting syntax (`.`)
```julia
julia> versioninfo()
Julia Version 1.0.0
Commit 5d4eaca0c9 (2018-08-08 20:58 UTC)
Platform Info:
  CPU: Intel(R) Core(TM) i7-4700MQ CPU @ 2.40GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.0 (ORCJIT, haswell)
Environment:
  JULIA_EDITOR = code

julia> @which abs.([1,2,3])
ERROR: no unique matching method found for the specified argument types
Stacktrace:
 [2] which(::Any, ::Any) at ./reflection.jl:922
 [3] top-level scope at none:0

julia> @code_lowered abs.([1,2,3])
0-element Array{Union{Nothing, CodeInfo},1}

```
In the code above I'd expect `@which` to return `broadcast` and `@code_lowered` to return whatever  code that broadcast gets lowered to.

If the loop fusion magic behind the scenes turns out to be complicated to work with these macros, clearer error messages (or, in the case of @code_lowered, *an* error message), would be great as well.

Liozou:
Fixes #28615, fixes #14637, closes #20620. Actually the last one was already fixed on master it seems, so I only added a test.

## Broadcasting

This PR enables using the `code_` macros with dot expressions, like so:
```julia
julia> @code_warntype max.(1 .+ 3, 5 - 7)
Variables
  #self#::Core.Compiler.Const(var"##dotfunction#282#116"(), false)
  x1::Int64
  x2::Int64
  x3::Int64

Body::Int64
1 ─ %1 = Base.broadcasted(Main.:+, x1, x2)::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{0},Nothing,typeof(+),Tuple{Int64,Int64}}
│   %2 = Base.broadcasted(Main.max, %1, x3)::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{0},Nothing,typeof(max),Tuple{Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{0},Nothing,typeof(+),Tuple{Int64,Int64}},Int64}}
│   %3 = Base.materialize(%2)::Int64
└──      return %3
```

This works by wrapping the dot expression in a function and applying the introspection to the created function. In the example above, the newly created function is 
```julia
dotfunction(x1, x2, x3) = max.(x1 .+ x2, x3)
```
Note that the `5 - 7` argument has been transformed into a single argument for the new function, since it is not broadcasted upon.

Other introspection macros such as `@which` or `@edit` now error with a message pointing the user towards `Meta.@lower`:
```julia
julia> @which abs.([1,2])
ERROR: dot expressions are not lowered to a single function call, so @which cannot analyze them. You may want to use Meta.@lower to identify which function call to target.
Stacktrace:
 [2] top-level scope at none:1
```
I also added a small message mentioning `Meta.@lower` in the general error message for the introspection macros because I think it deserves to be there, but let me know what you think of it.

## `getproperty`

The dot used for broadcasting and the dot used for `getproperty` have different context so there is no confusion when handling them. When both are present, the outermost dominates (I don't know how to phrase it, but I think it just follows intuition):
```julia
julia> @code_typed optimize=false Ref.(["foo", 4])[1].x
CodeInfo(
1 ─ %1 = Base.getfield(x, f)::String
└──      return %1
) => String
```

Finally, `@which` can now handle fully qualified symbols (#14637):
```julia
julia> @which Main.Base.nothing
Core
```

### DIFF
--- a/stdlib/InteractiveUtils/src/macros.jl
+++ b/stdlib/InteractiveUtils/src/macros.jl
@@ -6,6 +6,26 @@ import Base: typesof, insert!
 
 separate_kwargs(args...; kwargs...) = (args, kwargs.data)
 
+"""
+Transform a dot expression into one where each argument has been replaced by a
+variable "xj" (with j an integer from 1 to the returned i).
+The list `args` contains the original arguments that have been replaced.
+"""
+function recursive_dotcalls!(ex, args, i=1)
+    if !(ex isa Expr) || ((ex.head !== :. || !(ex.args[2] isa Expr)) &&
+                          (ex.head !== :call || string(ex.args[1])[1] != '.'))
+        newarg = Symbol('x', i)
+        push!(args, ex)
+        return newarg, i+1
+    end
+    (start, branches) = ex.head === :. ? (1, ex.args[2].args) : (2, ex.args)
+    for j in start:length(branches)
+        branch, i = recursive_dotcalls!(branches[j], args, i)
+        branches[j] = branch
+    end
+    return ex, i
+end
+
 function gen_call_with_extracted_types(__module__, fcn, ex0, kws=Expr[])
     if isa(ex0, Expr)
         if ex0.head === :do && Meta.isexpr(get(ex0.args, 1, nothing), :call)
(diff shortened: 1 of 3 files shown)


MasonProtter:
This should probably have a NEWS.md entry, right?

Liozou:
Well it's arguably a bugfix. But if people are in favor I can PR a NEWS entry joined with #35283, it would make sense.

MasonProtter:
I see it as a feature enhancement.

PR #35526, opened 2020-04-20
Discourage definition of custom primitive types

cmcaine:
As discussed with Stefan and Matt Bauman on slack.


StefanKarpinski (approved):

kimikage:
BTW, I think this have a problem with the indentation.
https://docs.julialang.org/en/v1.6-dev/manual/types/#Primitive-Types-1

PR #35527, opened 2020-04-20
Improve threads performance when running under rr

Keno:
Partr does a user space busy loop when no events are found (until
some timeout expires), in the expectation that another thread might push some.
When running under rr, all execution is serialized, so there's no other
thread running simultaneously, and rr will happily let the thread busy loop until
its timeslice (50ms or so) is expired. The result is a dramatic
performance decrease in the threads test that is entirely unnecessary
(even during regular execution it's not clear that busy waiting
makes much sense on low-core CPUs, since we're potentially starving
another thread of the execution time it needs to actually schedule
some work). For now, just detect that we're running under rr and stop
doing any sort of busy waiting in that case.

### DIFF
--- a/src/partr.c
+++ b/src/partr.c
@@ -275,10 +275,45 @@ JL_DLLEXPORT int jl_enqueue_task(jl_task_t *task)
 }
 
 
+static int running_under_rr(void)
+{
+#ifdef _OS_LINUX_
+#define RR_CALL_BASE 1000
+#define SYS_rrcall_check_presence (RR_CALL_BASE + 8)
+    static int checked_running_under_rr = 0;
+    static int is_running_under_rr = 0;
+    if (!checked_running_under_rr) {
+        int ret = syscall(SYS_rrcall_check_presence, 0, 0, 0, 0, 0, 0);
+        if (ret == -1) {
+            // Should always be ENOSYS, but who knows what people do for
+            // unknown syscalls with their seccomp filters, so just say
+            // that we don't have rr.
+            is_running_under_rr = 0;
+        }
+        else {
+            is_running_under_rr = 1;
+        }
+        checked_running_under_rr = 1;
+    }
+    return is_running_under_rr;
+#else
+    return 0;
+#endif
+}
+
+
 //  sleep_check_after_threshold() -- if sleep_threshold ns have passed, return 1
 static int sleep_check_after_threshold(uint64_t *start_cycles)
 {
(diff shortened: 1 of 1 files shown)


src/partr.c
```diff
@@ -274,11 +274,42 @@ JL_DLLEXPORT int jl_enqueue_task(jl_task_t *task)
...
+#define SYS_rrcall_check_presence (RR_CALL_BASE + 8)
+
+static int running_under_rr(void) {
+#ifdef _OS_LINUX_
+    if (!checked_running_under_rr) {
+        int ret = syscall(SYS_rrcall_check_presence, 0, 0, 0, 0, 0, 0);
+        if (ret == -1) {
+            assert(errno == ENOSYS);
+            is_running_under_rr = 0;
+        } else {
```
vtjnash, line 289:
```suggestion
        }
        else {
```

src/partr.c
```diff
@@ -274,11 +274,42 @@ JL_DLLEXPORT int jl_enqueue_task(jl_task_t *task)
     return 0;
 }
 
+static int checked_running_under_rr = 0;
+static int is_running_under_rr = 0;
+#define RR_CALL_BASE 1000
+#define SYS_rrcall_check_presence (RR_CALL_BASE + 8)
+
+static int running_under_rr(void) {
```
vtjnash, line 282:
```suggestion
static int running_under_rr(void)
{
```

src/partr.c
```diff
@@ -274,11 +274,42 @@ JL_DLLEXPORT int jl_enqueue_task(jl_task_t *task)
     return 0;
 }
 
+static int checked_running_under_rr = 0;
+static int is_running_under_rr = 0;
+#define RR_CALL_BASE 1000
+#define SYS_rrcall_check_presence (RR_CALL_BASE + 8)
```
vtjnash, line 280:
```suggestion
#ifdef _OS_LINUX_
static int checked_running_under_rr = 0;
static int is_running_under_rr = 0;
#define RR_CALL_BASE 1000
#define SYS_rrcall_check_presence (RR_CALL_BASE + 8)
#endif
```
(though might make sense to move these inside the `running_under_rr` function)

vtjnash (approved):
FWIW, the reason this is good during normal execution is that we're bounded in the number of threads we've launched to the number of cores, so we should degrade relatively gracefully to different core counts—as long as the system scheduler (aka rr) isn't messing with our notion of a timeslice.

PR #35539, opened 2020-04-21
Test suite tweaks for rr

Keno:
threads: Pass through environment to thread_exec (for LD_PRELOAD for rr)
FileWatching: If the fds we allocated were to high, dump out what all those other fds were up to

### DIFF
--- a/base/options.jl
+++ b/base/options.jl
@@ -45,6 +45,7 @@ struct JLOptions
     image_file_specified::Int8
     warn_scope::Int8
     image_codegen::Int8
+    rr_detach::Int8
 end
 
 # This runs early in the sysimage != is not defined yet
--- a/src/jlapi.c
+++ b/src/jlapi.c
@@ -663,6 +663,17 @@ static void lock_low32(void)
 // Actual definition in `ast.c`
 void jl_lisp_prompt(void);
 
+static void rr_detach_teleport(void) {
+#ifdef _OS_LINUX_
+#define RR_CALL_BASE 1000
+#define SYS_rrcall_detach_teleport (RR_CALL_BASE + 9)
+    int err = syscall(SYS_rrcall_detach_teleport, 0, 0, 0, 0, 0, 0);
+    if (err < 0 || jl_running_under_rr(1)) {
+        jl_error("Failed to detach from rr session");
+    }
+#endif
+}
+
 JL_DLLEXPORT int repl_entrypoint(int argc, char *argv[])
 {
     // no-op on Windows, note that the caller must have already converted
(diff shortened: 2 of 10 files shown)


Keno:
I've pushed two additional test suite tweaks that are needed for the test suite to pass under rr and retitled the PR. With this merged, the test suite passes for me under rr.

base/download.jl
```diff
@@ -35,9 +35,21 @@ function find_curl()
...
 end
 
+# Clear libpath for invoking system executables that we generally don't want to be confused
+# by things like a libcurl that sits in Julia's private libdir.
+function clear_libpath(f::Function)
+    withenv(f,
+        "LD_LIBRARY_PATH" => nothing,
+        "DYLD_FALLBACK_LIBRARY_PATH" => nothing,
+        "DYLD_LIBRARY_PATH" => nothing,
+    )
```
tkf, line 45: This is not thread-safe, right? https://github.com/JuliaLang/julia/issues/34726

Why not use `setenv`?

staticfloat: You're right, that's just better.  Thanks!

base/download.jl
```diff
@@ -35,9 +35,19 @@ function find_curl()
...
 end
 
+# Clear libpath for invoking system executables that we generally don't want to be confused
+# by things like a libcurl that sits in Julia's private libdir.
+function clear_libpath(cmd)
+    setenv(cmd,
+        "LD_LIBRARY_PATH" => nothing,
+        "DYLD_FALLBACK_LIBRARY_PATH" => nothing,
+        "DYLD_LIBRARY_PATH" => nothing,
+    )
```
tkf, line 45: Don't you need something like

```suggestion
    env = copy(ENV)
    pop!(env, "LD_LIBRARY_PATH", nothing)
    pop!(env, "DYLD_FALLBACK_LIBRARY_PATH", nothing)
    pop!(env, "DYLD_LIBRARY_PATH", nothing)
    setenv(cmd, env)
```

?  I think the current implementation of `clear_libpath` can clear out some important environment variables (e.g., `PATH`). See:

```julia
julia> withenv("X" => "a") do
           cmd = `sh -c 'echo X = $X'`
           run(cmd)
           run(setenv(cmd, "Y" => "b"))
       end;
X = a
X =
```

It'd be nice to have API like `setenv(cmd, ENV, pairs...)` to do what we need here.

Keno:
Given the recent CI instability, it seems worth pushing this over the finish line. I've rebased this. Let's see what rr has to say.

Keno:
Alright, starting to work through the rr-specific test failures. The Profile test failure should be fixed by https://github.com/rr-debugger/rr/pull/2753.

I believe other than that we have issues in Distributed, Downloads and Threads. The Threads test failure just looks like a timeout caused by rr overhead, but I'm considering just not running the threads test under rr anyway, since it can hide things like memory model issues, which that test definitely needs to catch. I haven't looked into the Distributed and Downloads test failures yet.

Keno:
Alright, the rr tracing seems to finally work. Will merge this PR when regular CI Is green and then we'll turn on rr by default in a few days.

PR #32309, opened 2019-06-12
threads: expand thread-safe region for I/O

vtjnash:
This should hopefully cover most I/O operations that go through libuv (meaning that other operations like using IOStream files aren't part of this PR), to make them thread-safe.

The UDPSocket implementation of many things has been always wrong, this also fixes them.

PR #35545, opened 2020-04-21
implement LibuvStream interface for BufferStream

vchuravy:
It is noted as a non-OS stream, but has also been a subtype
of LibuvStream since forever. #32309 did not add the `readerror`
field.

It might be better to decouple `BufferStream` from the `LibuvStream`
interface, but I decided to be conservative. Are changes in the type-hierarchy
backportable?

### DIFF
--- a/base/stream.jl
+++ b/base/stream.jl
@@ -23,7 +23,6 @@ abstract type LibuvStream <: IO end
 # .  +- Pipe
 # .  +- Process (not exported)
 # .  +- ProcessChain (not exported)
-# +- BufferStream
 # +- DevNull (not exported)
 # +- Filesystem.File
 # +- LibuvStream (not exported)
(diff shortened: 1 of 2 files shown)


vchuravy:
Fixes https://github.com/JuliaParallel/MPIClusterManagers.jl/issues/9

JeffBezanson:
> Are changes in the type-hierarchy
> backportable?

No; if `A <: B` is true in 1.0 it needs to stay that way. Inserting new types is allowed though.

VasylHafych:
Hi! Looking forward to seeing this in 1.4 branch, as this PR is crucial to run my parallel Julia code with MPI.jl on a cluster.

JeffBezanson:
It looks like UDPSocket is the remaining subtype of LibuvStream that doesn't have this. Not sure if it's needed.

oschulz:
If possible at all, I'd be so glad to see this in v1.4.2 (also for MPI.jl).

StefanKarpinski:
Seems like too substantial a change for a bugfix release. Could happen in 1.5 though.

vchuravy:
> Seems like too substantial a change for a bugfix release. Could happen in 1.5 though.

What is too substantial? The struct layout change?

StefanKarpinski:
I misinterpreted this PR as changing the type hierarchy, not just changing the doc to reflect reality there. Just adding a field may be fine in a point release. @JeffBezanson's call I think.

JeffBezanson:
I think the bugfix value of this outweighs any possible problems.

PR #35554, opened 2020-04-22
Add zero for TimeType

matthieugomez:
According to docs, zero "gets the additive identity element for the type of x (x can also specify the
type itself).". Defining this method allows one to get the type of `dt::T - dt::T` in a generic way.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -167,7 +167,7 @@ Standard library changes
 
 #### Dates
 * The `eps` function now accepts `TimeType` types ([#31487]).
-* The `zero` function now accepts `TimeType` types.
+* The `zero` function now accepts `TimeType` types ([#35554]).
 
 
 #### Statistics
--- a/NEWS.md
+++ b/NEWS.md
@@ -167,6 +167,8 @@ Standard library changes
 
 #### Dates
 * The `eps` function now accepts `TimeType` types ([#31487]).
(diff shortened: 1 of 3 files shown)


quinnj (approved):
LGTM

NEWS.md
```diff
@@ -167,6 +167,8 @@ Standard library changes
 
 #### Dates
 * The `eps` function now accepts `TimeType` types ([#31487]).
+* The `zero` function now accepts `TimeType` types.
```
dkarrasch, line 170: The PR number should be added for reference.

stdlib/Dates/src/types.jl
```diff
@@ -401,6 +401,12 @@ Base.eps(::Type{Time}) = Nanosecond(1)
 Base.eps(::T) where T <: TimeType = eps(T)::Period
 
 
+Base.zero(::Type{DateTime}) = Millisecond(0)
```
nickrobinson251, line 404: why is zero of a `DateTime` a `Miilisecond`, but zero for a `Time` a `Nanosecond`? Can the answer to that be added as a comment?

matthieugomez: It returns x::T - x::T (added as a comment).

nickrobinson251: Oh that's a good reason! thanks for adding the comment 👍

PR #22007, opened 2017-05-21
jump to numbered method in the REPL

fredrikekre:
This PR makes it possible to jump to a numbered method from `methods(foo)`, as can now be done for numbered stackframes (#19680).
I think this is really convenient sometimes, instead of figuring out the correct input and then use `@edit` to get there.

Example here:
https://asciinema.org/a/1r34os4hyyugzw0e2n3il2a5a?t=12

PR #35556, opened 2020-04-22
mention methodtable Ctrl+Q trick in methodshow

simeonschaub:
Thanks to @oxinabox for pointing me to this trick on Slack. I think this deserves to be mentioned in the MethodTable output.

### DIFF
--- a/base/errorshow.jl
+++ b/base/errorshow.jl
@@ -541,13 +541,10 @@ function show_method_candidates(io::IO, ex::MethodError, @nospecialize kwargs=()
     end
 end
 
-# Contains file name and file number. Gets set when a backtrace
-# or methodlist is shown. Used by the REPL to make it possible to open
-# the location of a stackframe/method in the editor.
-global LAST_SHOWN_LINE_INFOS = Tuple{String, Int}[]
-
 function show_trace_entry(io, frame, n; prefix = "")
-    push!(LAST_SHOWN_LINE_INFOS, (string(frame.file), frame.line))
+    if haskey(io, :LAST_SHOWN_LINE_INFOS)
+        push!(io[:LAST_SHOWN_LINE_INFOS], (string(frame.file), frame.line))
+    end
     print(io, "\n", prefix)
     show(io, frame, full_path=true)
     n > 1 && print(io, " (repeats ", n, " times)")
@@ -634,7 +631,9 @@ function show_reduced_backtrace(io::IO, t::Vector, with_prefix::Bool)
 end
 
 function show_backtrace(io::IO, t::Vector)
-    resize!(LAST_SHOWN_LINE_INFOS, 0)
+    if haskey(io, :LAST_SHOWN_LINE_INFOS)
+        resize!(io[:LAST_SHOWN_LINE_INFOS], 0)
+    end
     filtered = process_backtrace(t)
     isempty(filtered) && return
 
--- a/base/methodshow.jl
+++ b/base/methodshow.jl
@@ -266,6 +266,14 @@ function show_method_table(io::IO, ms::MethodList, max::Int=-1, header::Bool=tru
             end
         end
     end
+
+    if isinteractive()
+        print(
+            io,
+            "\n\nTo edit a specific method, type the corresponding number into the " *
+            "REPL and press Ctrl+Q",
+        )
+    end
 end
 
 show(io::IO, ms::MethodList) = show_method_table(io, ms)
--- a/base/methodshow.jl
+++ b/base/methodshow.jl
@@ -240,6 +240,7 @@ function show_method_table(io::IO, ms::MethodList, max::Int=-1, header::Bool=tru
     end
     n = rest = 0
     local last
+    LAST_SHOWN_LINE_INFOS = get(io, :LAST_SHOWN_LINE_INFOS, Tuple{String,Int}[])
 
     resize!(LAST_SHOWN_LINE_INFOS, 0)
     for meth in ms
@@ -265,14 +266,6 @@ function show_method_table(io::IO, ms::MethodList, max::Int=-1, header::Bool=tru
                 print(io, " (use methods($name) to see them all)")
             end
         end
-    end
-
-    if get(io, :interactive, false)
-        print(
-            io,
-            "\n\nTo edit a specific method, type the corresponding number into the " *
-            "REPL and press Ctrl+Q",
-        )
     end
 end
 
@@ -381,6 +374,7 @@ show(io::IO, mime::MIME"text/html", mt::Core.MethodTable) = show(io, mime, Metho
 
 # pretty-printing of AbstractVector{Method}
 function show(io::IO, mime::MIME"text/plain", mt::AbstractVector{Method})
+    LAST_SHOWN_LINE_INFOS = get(io, :LAST_SHOWN_LINE_INFOS, Tuple{String,Int}[])
     resize!(LAST_SHOWN_LINE_INFOS, 0)
     first = true
     for (i, m) in enumerate(mt)
--- a/base/methodshow.jl
+++ b/base/methodshow.jl
@@ -267,7 +267,7 @@ function show_method_table(io::IO, ms::MethodList, max::Int=-1, header::Bool=tru
         end
     end
 
-    if isinteractive()
+    if get(io, :interactive, false)
         print(
             io,
             "\n\nTo edit a specific method, type the corresponding number into the " *
--- a/stdlib/REPL/src/REPL.jl
+++ b/stdlib/REPL/src/REPL.jl
@@ -212,8 +212,23 @@ function display(d::REPLDisplay, mime::MIME"text/plain", x)
         io = foldl(IOContext, d.repl.options.iocontext,
                    init=IOContext(io, :limit => true, :module => Main))
     end
+
+    if isdefined(d.repl, :last_shown_line_infos)
+        infos = d.repl.last_shown_line_infos
+        io = IOContext(io, :LAST_SHOWN_LINE_INFOS => infos)
+    end
+
     show(io, mime, x)
     println(io)
+
+    if !isempty(get(io, :last_shown_line_infos, Tuple{String,Int}[]))
+        println(
+            io,
+            "\nTo edit a specific method, type the corresponding number into the " *
+            "REPL and press Ctrl+Q",
+        )
+    end
+
     nothing
 end
 display(d::REPLDisplay, x) = display(d, MIME("text/plain"), x)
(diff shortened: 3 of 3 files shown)


simeonschaub:
Tagging @fredrikekre, since I believed you added this feature in  #22007. Could I get a review?

tkf:
This is a very nice thing to have for new users. But wouldn't this write the "tips" message into a file if I type `open(io -> show(io, methods(identity)), "file", write=true)`?  I think "UI" message like this should be implemented in the REPL module (e.g., REPL sets some custom `IOContext` and it'd be used here instead of `isinteractive()`).

simeonschaub:
Should I check for `haskey(io, :module)` instead? Otherwise, I don't think the REPL module sets any custom `IOContext`s.

tkf:
I don't think `:module` context is adequate for this. Why not add a new `IOContext` like this?

```diff
diff --git a/stdlib/REPL/src/REPL.jl b/stdlib/REPL/src/REPL.jl
index 30af00a6cd..be3e14538d 100644
--- a/stdlib/REPL/src/REPL.jl
+++ b/stdlib/REPL/src/REPL.jl
@@ -206,6 +206,7 @@ end
 function display(d::REPLDisplay, mime::MIME"text/plain", x)
     io = outstream(d.repl)
     get(io, :color, false) && write(io, answer_color(d.repl))
+    io = IOContext(io, :interactive => true)
     if isdefined(d.repl, :options) && isdefined(d.repl.options, :iocontext)
         # this can override the :limit property set initially
         io = foldl(IOContext, d.repl.options.iocontext,
```

tkf:
If we go with this direction, should we add `:interactive` to this list?

https://github.com/JuliaLang/julia/blob/65869b11d2bbf7c16c1aa81a46f331e769f7822d/base/show.jl#L266-L283

simeonschaub:
Yes, I think that would be helpful. Should this all go into this PR or would it make more sense to have a dedicated PR for this?

tkf:
This patch introduces a new convention in REPL.jl so I think it makes sense to add the documentation here.

simeonschaub:
Perhaps `REPL` or `isREPL` is a better name for this, since this trick doesn't work in Jupyter Notebooks, for example, but there `isinteractive()` still returns true?

tkf:
Ah, that's a very good point.  Yes, `:interactive` context is useless for this.

I wonder if we should just directly overload `display(d::REPLDisplay, mime::MIME"text/plain", x)` or add some overloadable hint mechanism.

A bit hacky solution may be something like

```julia
function display(d::REPLDisplay, mime::MIME"text/plain", x)
    backup = Base.LAST_SHOWN_LINE_INFOS
    empty!(Base.LAST_SHOWN_LINE_INFOS)

    ...

    if isempty(Base.LAST_SHOWN_LINE_INFOS)
        append!(Base.LAST_SHOWN_LINE_INFOS, backup)
    else
        print(io, the_hint_message)
    end
end
```

tkf:
Hmm... Actually, maybe `LAST_SHOWN_LINE_INFOS` itself should be in `IOContext`? It's not nice to communicate via a mutating global state.  I think it'd be better to do something like

```julia
infos = []
io = IOContext(io, :LAST_SHOWN_LINE_INFOS => infos)
show(io, "text/plain", x)
d.repl.last_shown_line_infos = infos
```

in `display`.

simeonschaub:
I don't really get, why we would need `Base.LAST_SHOWN_LINE_INFOS` here. Does this tell us anything about whether the output is a REPL? Is there anything wrong with just using the approach here and renaming `:interactive` to `:REPL`?

tkf:
> Does this tell us anything about whether the output is a REPL?

No, it doesn't.  My suggestion was to do it the other way around.  If we implement this in `display(d::REPLDisplay, mime::MIME"text/plain", x)`, we _know_ that the output is the REPL because of `d::REPLDisplay`.  However, now we need to know if the method table is printed or not.  If we do `empty!(Base.LAST_SHOWN_LINE_INFOS)` before calling `show` and then if it becomes non-empty after that, we know that a method table is printed.

(However, this is correct only if there is only one `Task`.  It's much better to do this without the global variable `Base.LAST_SHOWN_LINE_INFOS` now that threading (or even `@async`) becomes more common. So that's why I'm mentioning that using `IOContext` is even better.)

I think this is the right way to do it.  As you pointed out, this is very specific to REPL. So, I think REPL is the best place to implement it.

simeonschaub:
@tkf I got the same error locally, I just wanted to see whether it was just on my machine. Do you have any idea why it would fail? I don't really know where to start looking into this.

stdlib/REPL/src/REPL.jl
```diff
@@ -206,13 +206,29 @@ end
...
+
+    if isdefined(d.repl, :last_shown_line_infos)
+        infos = d.repl.last_shown_line_infos
+        io = IOContext(io, :LAST_SHOWN_LINE_INFOS => infos)
+    end
+
     show(io, mime, x)
     println(io)
+
+    if !isempty(get(io, :last_shown_line_infos, Tuple{String,Int}[]))
```
simeonschaub, line 224: Typo here, but this shouldn't cause building to fail.

stdlib/REPL/src/REPL.jl
```diff
@@ -432,9 +448,10 @@ mutable struct LineEditREPL <: AbstractREPL
     mistate::Union{MIState,Nothing}
     interface::ModalInterface
     backendref::REPLBackendRef
+    last_shown_line_infos::Vector{Tuple{String,Int}}
     LineEditREPL(t,hascolor,prompt_color,input_color,answer_color,shell_color,help_color,history_file,in_shell,in_help,envcolors) =
         new(t,hascolor,prompt_color,input_color,answer_color,shell_color,help_color,history_file,in_shell,
-            in_help,envcolors,false,nothing, Options(), nothing)
+            in_help,envcolors,false,nothing, Options(), nothing, Tuple{String,Int}[])
```
tkf, line 454: It seems that this object is partially initialized by default. The number of arguments to `new` is less than the number of fields. So, it might be adding the default argument of an incompatible type. Not sure what's the best way to fix it, though. The simplest solution seems to put `last_shown_line_infos` somewhere above `interface`.

simeonschaub: Thanks so much, that seems to have fixed it. I really did not think of that.

stdlib/REPL/src/REPL.jl
```diff
@@ -206,13 +206,30 @@ end
...
         # this can override the :limit property set initially
         io = foldl(IOContext, d.repl.options.iocontext,
                    init=IOContext(io, :limit => true, :module => Main))
     end
+
+    infos = Tuple{String,Int}[]
+    if isdefined(d.repl, :last_shown_line_infos)
+        infos = d.repl.last_shown_line_infos
+        io = IOContext(io, :LAST_SHOWN_LINE_INFOS => infos)
+    end
```
tkf, line 220:
```suggestion
    io = IOContext(io, :LAST_SHOWN_LINE_INFOS => infos)
```

stdlib/REPL/src/REPL.jl
```diff
@@ -206,13 +206,30 @@ end
...
+
     show(io, mime, x)
     println(io)
+
+    if !isempty(infos)
```
tkf, line 225:
```suggestion
    if !isempty(infos)
        d.repl.last_shown_line_infos = infos
```

tkf (commented):
Isn't it simpler to always use `io = IOContext(io, :LAST_SHOWN_LINE_INFOS => infos)`?

Also, current code seems to not update `d.repl.last_shown_line_infos` after the second time?

simeonschaub:
Yes, that makes sense. Originally, I had structured it a bit differently, but doesn't really matter either way.

stdlib/REPL/src/REPL.jl
```diff
@@ -206,13 +206,28 @@ end
 function display(d::REPLDisplay, mime::MIME"text/plain", x)
     io = outstream(d.repl)
     get(io, :color, false) && write(io, answer_color(d.repl))
+    io = IOContext(io, :interactive => true)
```
tkf, line 209:
```suggestion
```

I think we can remove this for now.

(It'd be useful to have this, but I think it's clear to do it in a separate PR that also introduces a proper docstring to `IOContext`)

tkf (approved):

simeonschaub:
Bump. Is this good to go?

tkf:
It's is rather a radical change in the "philosophy" of how `display` works on REPL. So it'd be nice if other devs can have a look at it.

ping @StefanKarpinski

StefanKarpinski:
I don't think I'm the person for this particular review. @JeffBezanson, @stevengj and @vtjnash have more of a history and familiarity with the display system.

vtjnash (approved):
Ah, this is rather clever! It's unusual, but I can't see any reason that'd be a problem.

base/errorshow.jl
```diff
@@ -634,7 +631,9 @@ function show_reduced_backtrace(io::IO, t::Vector, with_prefix::Bool)
 end
 
 function show_backtrace(io::IO, t::Vector)
-    resize!(LAST_SHOWN_LINE_INFOS, 0)
+    if haskey(io, :LAST_SHOWN_LINE_INFOS)
+        resize!(io[:LAST_SHOWN_LINE_INFOS], 0)
+    end
```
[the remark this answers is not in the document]
tkf: Is it OK to skip `resize!` here? What happens when multiple stack traces are shown (via nested errors)?  Would `IOContext` be reset appropriately?

base/methodshow.jl
```diff
@@ -373,6 +374,7 @@ show(io::IO, mime::MIME"text/html", mt::Core.MethodTable) = show(io, mime, Metho
 
 # pretty-printing of AbstractVector{Method}
 function show(io::IO, mime::MIME"text/plain", mt::AbstractVector{Method})
+    LAST_SHOWN_LINE_INFOS = get(io, :LAST_SHOWN_LINE_INFOS, Tuple{String,Int}[])
     resize!(LAST_SHOWN_LINE_INFOS, 0)
```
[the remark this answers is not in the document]
tkf: Don't you need `resize!` for something like this?

```julia
struct TwoMethodTables
    methods1
    methods2
end

function Base.show(io::IO, ::MIME"text/plain", two::TwoMethodTables)
    println(io, "TwoMethodTables:")
    show(io, MIME"text/plain"(), two.methods1)
    show(io, MIME"text/plain"(), two.methods2)
end
```

Currently, I think Ctrl+Q works for numbering in `methods2`. If we remove `resize!`, it'd refer to the numbering in `methods1` (and also in `methods2` shifted by `length(methods)`).

simeonschaub:
Bumping this again. I thought the `resize!` was still necessary too, so printing multiple stack traces still works correctly.

tkf:
I merged this as @vtjnash is OK with it at a high-level picture. I'm keeping `resize!` since both @simeonschaub and I think that's necessary.

@vtjnash Let us know if the code around `resize!` still has to be fixed.

It seems to be better to merge now so that there is more chance to play with the UI in the 1.6-DEV cycle.  It's also better to merge this before #36134.

(The test failures look like some network glitches.)

PR #35557, opened 2020-04-22
Special case empty covec-diagonal-vec product

### LINKED ISSUE #714: Adjoint times Diagonal times Vector for n = 0
```
# This works
julia> zeros(0)'*zeros(0)
0.0

# This does not, which is inconsistent
julia> zeros(0)'*Diagonal(zeros(0))*zeros(0)
Stacktrace:
 [1] _empty_reduce_error() at ./reduce.jl:212
 [2] mapreduce_empty(::Function, ::Function, ::Type) at ./reduce.jl:246
 [3] mapreduce_empty_iter(::Function, ::Function, ::Base.Iterators.Zip{Tuple{Adjoint{Float64,Array{Float64,1}},Array{Float64,1},Array{Float64,1}}}, ::Base.HasEltype) at ./reduce.jl:254
 [4] mapfoldl_impl at ./reduce.jl:57 [inlined]
 [5] #mapfoldl#186 at ./reduce.jl:72 [inlined]
 [6] mapfoldl at ./reduce.jl:72 [inlined]
 [7] #mapreduce#194 at ./reduce.jl:200 [inlined]
 [8] mapreduce at ./reduce.jl:200 [inlined]
 [9] *(::Adjoint{Float64,Array{Float64,1}}, ::Diagonal{Float64,Array{Float64,1}}, ::Array{Float64,1}) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.3/LinearAlgebra/src/diagonal.jl:640
 [10] top-level scope at none:0
```

### LINKED ISSUE #35424: Adjoint times Diagonal times Vector for n = 0
```
# This works
julia> zeros(0)'*zeros(0)
0.0

# This does not, which is inconsistent
julia> zeros(0)'*Diagonal(zeros(0))*zeros(0)
Stacktrace:
 [1] _empty_reduce_error() at ./reduce.jl:212
 [2] mapreduce_empty(::Function, ::Function, ::Type) at ./reduce.jl:246
 [3] mapreduce_empty_iter(::Function, ::Function, ::Base.Iterators.Zip{Tuple{Adjoint{Float64,Array{Float64,1}},Array{Float64,1},Array{Float64,1}}}, ::Base.HasEltype) at ./reduce.jl:254
 [4] mapfoldl_impl at ./reduce.jl:57 [inlined]
 [5] #mapfoldl#186 at ./reduce.jl:72 [inlined]
 [6] mapfoldl at ./reduce.jl:72 [inlined]
 [7] #mapreduce#194 at ./reduce.jl:200 [inlined]
 [8] mapreduce at ./reduce.jl:200 [inlined]
 [9] *(::Adjoint{Float64,Array{Float64,1}}, ::Diagonal{Float64,Array{Float64,1}}, ::Array{Float64,1}) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.3/LinearAlgebra/src/diagonal.jl:640
 [10] top-level scope at none:0
```

MasonProtter:
Closes JuliaLang/LinearAlgebra.jl#714


stdlib/LinearAlgebra/src/diagonal.jl
```diff
@@ -659,10 +659,20 @@ end
...
-*(x::Adjoint{<:Any,<:AbstractVector}, D::Diagonal, y::AbstractVector) =
-    mapreduce(t -> t[1]*t[2]*t[3], +, zip(x, D.diag, y))
-*(x::Transpose{<:Any,<:AbstractVector}, D::Diagonal, y::AbstractVector) =
-    mapreduce(t -> t[1]*t[2]*t[3], +, zip(x, D.diag, y))
+function *(x::Adjoint{<:Any,<:AbstractVector}, D::Diagonal, y::AbstractVector)
+    if all(isempty.((x, D, y)))
+        return zero(promote_type(eltype.((x, D, y))...))
+    else
+        return mapreduce(t -> t[1]*t[2]*t[3], +, zip(x, D.diag, y))
+    end
```
tkf, line 667: FWIW I think something like

```suggestion
    return mapreduce(t -> t[1]*t[2]*t[3], +, zip(x, D.diag, y);
                     init = zero(promote_type(map(eltype, (x, D, y))...)))
```

may be better because:

1. Specifying a concrete value for `init` helps inference with `reduce`-like code in general.
2. Using `map` instead of broadcasting is less pressure to the compiler.

stdlib/LinearAlgebra/src/diagonal.jl
```diff
@@ -659,10 +659,20 @@ end
...
+end
+function *(x::Transpose{<:Any,<:AbstractVector}, D::Diagonal, y::AbstractVector)
+    if all(isempty.((x, D, y)))
+        return zero(promote_type(eltype.((x, D, y))...))
+    else
+        return mapreduce(t -> t[1]*t[2]*t[3], +, zip(x, D.diag, y))
+    end
```
tkf, line 674: ditto

```suggestion
    return mapreduce(t -> t[1]*t[2]*t[3], +, zip(x, D.diag, y);
                     init = zero(promote_type(map(eltype, (x, D, y))...)))
```

stdlib/LinearAlgebra/src/diagonal.jl
MasonProtter: Sure, let's try this!

tkf:
Oops my suggestion introduced a bug:

```
Error During Test at /buildworker/worker/tester_linux64/build/share/julia/stdlib/v1.5/LinearAlgebra/test/diagonal.jl:643
  Test threw exception
  Expression: y' * D * y == (y' * D) * y == (y' * A) * y
  MethodError: no method matching zero(::Type{AbstractArray{Int64,N} where N})
  Closest candidates are:
    zero(!Matched::Type{Pkg.Resolve.FieldValue}) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Pkg/src/Resolve/fieldvalues.jl:38
    zero(!Matched::Type{LibGit2.GitHash}) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/LibGit2/src/oid.jl:220
    zero(!Matched::Type{Missing}) at missing.jl:103
    ...
  Stacktrace:
   [1] *(::LinearAlgebra.Adjoint{LinearAlgebra.Adjoint{Int64,Array{Int64,1}},Array{Array{Int64,1},1}}, ::LinearAlgebra.Diagonal{Array{Int64,2},Array{Array{Int64,2},1}}, ::Array{Array{Int64,1},1}) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/LinearAlgebra/src/diagonal.jl:663
   [2] top-level scope at /buildworker/worker/tester_linux64/build/share/julia/stdlib/v1.5/LinearAlgebra/test/diagonal.jl:643
   [3] top-level scope at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Test/src/Test.jl:1114
   [4] top-level scope at /buildworker/worker/tester_linux64/build/share/julia/stdlib/v1.5/LinearAlgebra/test/diagonal.jl:635
```

https://build.julialang.org/#/builders/22/builds/11086/steps/5/logs/stdio

Indeed the dot product should work even with non-concrete element types like `y = Any[ones(2, 2)]` as long as the input is not empty. We can't use `zero` in this case. Sorry about that.

MasonProtter:
Makes sense! Nothing to apologize for, I didn't anticipate the error either.

stdlib/LinearAlgebra/src/diagonal.jl
```diff
@@ -659,10 +659,20 @@ end
...
+end
+function *(x::Transpose{<:Any,<:AbstractVector}, D::Diagonal, y::AbstractVector)
+    if all(isempty(x), isempty(D), isempty(y))
+        return zero(promote_type(eltype(x), eltype(D), eltype(y)))
+    else
+        return mapreduce(t -> t[1]*t[2]*t[3], +, zip(x, D.diag, y))
+    end
+end
 function dot(x::AbstractVector, D::Diagonal, y::AbstractVector)
     mapreduce(t -> dot(t[1], t[2], t[3]), +, zip(x, D.diag, y))
```
tkf, line 677: I guess this function can use the same implementation too? Also, since now we have a non-trivial function body, maybe it makes sense to define a separate internal function and call it from the above three functions?

MasonProtter: Sounds good. I called the the internal function `_dot` but that's not a great name. Maybe it should be called something like `_mul` or even `_mapreduce_mul`?

tkf: I think `_dot` is good.  It's a unique name in LinearAlgebra so it's easy to rename afterwards if we need to.

stdlib/LinearAlgebra/src/diagonal.jl
```diff
@@ -659,14 +659,19 @@ end
...
-*(x::Transpose{<:Any,<:AbstractVector}, D::Diagonal, y::AbstractVector) =
-    mapreduce(t -> t[1]*t[2]*t[3], +, zip(x, D.diag, y))
-function dot(x::AbstractVector, D::Diagonal, y::AbstractVector)
-    mapreduce(t -> dot(t[1], t[2], t[3]), +, zip(x, D.diag, y))
+*(x::Adjoint{<:Any,<:AbstractVector},   D::Diagonal, y::AbstractVector) = _dot(x, D, y)
+*(x::Transpose{<:Any,<:AbstractVector}, D::Diagonal, y::AbstractVector) = _dot(x, D, y)
+dot(x::AbstractVector, D::Diagonal, y::AbstractVector) = _dot(x, D, y)
+
+function _dot(x, D::Diagonal, y)
+    if all((isempty(x), isempty(D), isempty(y)))
```
dkarrasch, line 667: This reads funny. Wouldn't
```suggestion
    if isempty(x) && isempty(D) && isempty(y)
```
be easier to read and do the same job?

stdlib/LinearAlgebra/src/diagonal.jl
```diff
@@ -659,14 +659,19 @@ end
...
 *(x::Transpose{<:Any,<:AbstractVector}, D::Diagonal) = Transpose(map((t,s) -> transpose(t)*s, D.diag, parent(x)))
-*(x::Adjoint{<:Any,<:AbstractVector}, D::Diagonal, y::AbstractVector) =
-    mapreduce(t -> t[1]*t[2]*t[3], +, zip(x, D.diag, y))
-*(x::Transpose{<:Any,<:AbstractVector}, D::Diagonal, y::AbstractVector) =
-    mapreduce(t -> t[1]*t[2]*t[3], +, zip(x, D.diag, y))
-function dot(x::AbstractVector, D::Diagonal, y::AbstractVector)
-    mapreduce(t -> dot(t[1], t[2], t[3]), +, zip(x, D.diag, y))
+*(x::Adjoint{<:Any,<:AbstractVector},   D::Diagonal, y::AbstractVector) = _dot(x, D, y)
+*(x::Transpose{<:Any,<:AbstractVector}, D::Diagonal, y::AbstractVector) = _dot(x, D, y)
+dot(x::AbstractVector, D::Diagonal, y::AbstractVector) = _dot(x, D, y)
```
dkarrasch, line 664: This changes behavior. The idea is that `*` reduces only by one level, whereas `dot` acts recursively. For that reason, the naming `_dot` is unfortunate here, because it's not a helper function for `dot`, but for `*`. Now, `dot` probably deserves the same 0-length handling, but the `mapreduce` should continue to call `dot` as before.

dkarrasch (changes requested):
The dot issue was probably just oversight, but the difference between the two `*` methods and `dot` is relevant, and must remain.

stdlib/LinearAlgebra/src/diagonal.jl
MasonProtter: Apologies, I should have read the existing `dot` more carefully.

stdlib/LinearAlgebra/src/diagonal.jl
MasonProtter: Yep

stdlib/LinearAlgebra/src/diagonal.jl
```diff
@@ -659,14 +659,19 @@ end
...
-    mapreduce(t -> t[1]*t[2]*t[3], +, zip(x, D.diag, y))
-function dot(x::AbstractVector, D::Diagonal, y::AbstractVector)
-    mapreduce(t -> dot(t[1], t[2], t[3]), +, zip(x, D.diag, y))
+*(x::Adjoint{<:Any,<:AbstractVector},   D::Diagonal, y::AbstractVector) = _mapreduce_prod(*, x, D, y)
+*(x::Transpose{<:Any,<:AbstractVector}, D::Diagonal, y::AbstractVector) = _mapreduce_prod(*, x, D, y)
+dot(x::AbstractVector, D::Diagonal, y::AbstractVector) = _mapreduce_prod(dot, x, D, y)
+
+function _mapreduce_prod(f, x, D::Diagonal, y)
+    if isempty(x) && isempty(D) && isempty(y)
+        return zero(promote_type(eltype(x), eltype(D), eltype(y)))
```
dkarrasch, line 668: I'm not 100% sure, but we may need
```suggestion
        return zero(Base.promote_op(f, eltype(x), eltype(D), eltype(y)))
```
here. Because for arrays of (say, static) arrays, `dot` is supposed to return a `Number`, not just the promoted element array type.

dkarrasch (approved):

tkf:
> the difference between the two `*` methods and `dot` is relevant

I totally missed it. Thank you!

MasonProtter:
I don't think these CI failures are related.

Issue #15985, opened 2016-04-21
access to fused multiply-subtract etc.


stevengj:
Currently, `f(x,y,z) = @fastmath muladd(x, y, -z)` generates an explicit negation operation for the `-z` in the case of floating-point arguments, as can be seen by `@code_native f(3.0, 4.0, 5.0)`.   Similarly for `muladd(-x, y, z)` etcetera.

In #9881, @yuyichao indicated that this was due to the FastSel optimization in LLVM.

As I mentioned in #6330, it would be good to have access to fused multiply-subtract and similar operations.   LLVM doesn't provide intrinsics for these, but it would be good to have a way to write code that LLVM can transform into such instructions when they are available, and that will avoid explicit negations when fused operations aren't available.

cc: @eschnett, @simonbyrne

yuyichao:
I had a hard time following the isel logic in LLVM but it seems that the logic of replacing `fmadd(x, y, -z)` with `fmsub(x, y, z)` only appears at one place and looks pretty general (eliminate the neg if it is only used by one fma-like instruction) so I don't think there's much we can do to help llvm recognizing the code pattern here.

Can we make https://github.com/JuliaLang/julia/pull/15984 close this and make this an upstream issue?

stevengj:
Sounds good @yuyichao.

yuyichao:
Might be related https://llvm.org/bugs/show_bug.cgi?id=13118

PR #35562, opened 2020-04-22
Use muladd in complex evalpoly

augustt198:
Speeds up `evalpoly` for complex inputs a bit by using `muladd` for both recurrences. The number of instructions generated by the inner loop is halved.

### DIFF
--- a/base/math.jl
+++ b/base/math.jl
@@ -153,7 +153,7 @@ function evalpoly(z::Complex, p::Tuple)
             ai = Symbol("a", i)
             push!(as, :($ai = $a))
             a = :(muladd(r, $ai, $b))
-            b = :(p[$i] - s * $ai)
+            b = :(muladd(-s, $ai, p[$i]))
         end
         ai = :a0
         push!(as, :($ai = $a))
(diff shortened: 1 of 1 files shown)


MasonProtter:
Looks like the old version had [a comment](https://github.com/JuliaLang/julia/blob/v1.3.0/base/math.jl#L137) on it regarding this [issue](https://github.com/JuliaLang/julia/issues/15985) which has since been closed.

StefanKarpinski:
I'll leave this for a bit longer in case @stevengj has any comments. Please ping in a bit if this hasn't been merged and I'll merge it.

stevengj:
Looks good to me.   LLVM seems to emit a fused multiply-subtract these days:
```jl
julia> f(a,b,c) = muladd(-a, b, c)

julia> @code_llvm f(1., 2., 3.)

;  @ REPL[1]:1 within `f'
define double @julia_f_16811(double, double, double) {
top:
; ┌ @ float.jl:410 within `muladd'
   %3 = fmul contract double %0, %1
   %4 = fsub contract double %2, %3
; └
  ret double %4
}

julia> @code_native f(1., 2., 3.)
	.section	__TEXT,__text,regular,pure_instructions
; ┌ @ REPL[1]:1 within `f'
; │┌ @ REPL[1]:1 within `muladd'
	vfnmadd213sd	%xmm2, %xmm1, %xmm0
; │└
	retq
	nopw	%cs:(%rax,%rax)
; └
```

stevengj (approved):

PR #35565, opened 2020-04-23
Bump MbedTLS version

quinnj:
cc: @staticfloat, is this all that's needed in conjuction with https://github.com/JuliaPackaging/Yggdrasil/pull/929?


deps/Versions.make
```diff
@@ -20,7 +20,7 @@ GMP_BB_REL = 4
 MPFR_VER = 4.0.2
 MPFR_BB_REL = 2
 PATCHELF_VER = 0.9
-MBEDTLS_VER = 2.16.0
+MBEDTLS_VER = 2.16.6
 MBEDTLS_BB_REL = 1
```
staticfloat, line 24: You'll need to change this `BB_REL` as well to whatever the build number it gets assigned is, then afterwards you should delete all the `MbedTLS` checksums and then run `./contrib/refresh_bb_tarballs.sh mbedtls` to get the new checksums.

ViralBShah:
@quinnj Do we need to get this into 1.5?

quinnj:
It would be nice; but I don't think I'm able to do it.

musm:
superseded by https://github.com/JuliaLang/julia/pull/37408 

What's with these two files:
https://github.com/JuliaLang/julia/tree/master/deps/checksums/mbedtls-2.16.0-apache.tgz
https://github.com/JuliaLang/julia/tree/master/deps/checksums/mbedtls-2.16.0-gpl.tgz

Is there a reason they are not handled by BB? I don't know if I need to do anything with them or if they are stragglers and should be removed.

PR #35569, opened 2020-04-23
Added a few sentences to recommend using MKL.jl instead of building with MKL

activexray:
As building Julia with MKL seems to be giving some problems for some people, I think it is best if we recommend that people first try to use the MKL.jl package if all they are building Julia for is to add MKL.

### DIFF
--- a/doc/build/build.md
+++ b/doc/build/build.md
@@ -258,6 +258,8 @@ Add the following to the `Make.user` file:
 
 It is highly recommended to start with a fresh clone of the Julia repository.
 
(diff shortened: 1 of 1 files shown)


kiranshila:
I know it's mentioned in the FAQ, but might as well mention it here too.

MasonProtter (approved):
Looks good to me. 

Just a side comment though: Your title initially made me think you wanted to recommend that *everyone* use MKL.jl, which upon closer inspection was not at all what you were suggesting.

ViralBShah:
The build failures were all some github problems, unable to clone the repo. This is a simple doc fix, so felt ok to merge.

Issue #34076, opened 2019-12-11
Base.@ccallable has not the intended effect if it is inside a package and that package has a valid ji file


KristofferC:
Using `--output-o` and including a package that has a `Base.@ccallable` function called `julia_main` when the package is precompiled from some earlier usage gives the following symbols in the object file:

```
❯ objdump -t sys.o | grep julia_main
000000000055cfa0 l     F .text  000000000000000f jfptr_julia_main_21728
000000000055ce50 l     F .text  0000000000000144 julia_julia_main_21727
```

Now, only removing `.julia/compiled` and regenerating the sysimage, gives:
```
❯ objdump -t sys.o | grep julia_main
0000000000159680 l     F .text  000000000000000f jfptr_julia_main_5387
0000000000159690 l     F .text  000000000000009f jlcapi_julia_main_5390
0000000000159730 l     F .text  0000000000000098 jlcapi_julia_main_5390_gfthunk
00000000001594f0 l     F .text  0000000000000188 julia_julia_main_5386
0000000000159690 g     F .text  000000000000009f julia_main
```

(including the unmangled `julia_main` which is the point of `@ccallable` AFAIU).

So it seems that when the package gets loaded from the .ji file the `@ccallable` annotation is lost.

vchuravy:
Yeah IIRC we don't cache that, one possible workaround would be to delay the extern marking until `init`

KristofferC:
Fixed by https://github.com/JuliaLang/julia/pull/35574.

PR #35574, opened 2020-04-23
make `ccallable` work with system images and precompilation

### LINKED ISSUE #35014: Outputting a @ccallable function to an object file broken on master
MWE from test failure in PackageCompiler.jl

```jl
# callable.jlj
Base.@ccallable function f()::Cint
    return 0
end
```

```jl
run(`$(Base.julia_cmd()) --startup=no --output-o=sys.o ccallable.jl`)
```

prints:

```
Assertion failed: (Val && "isa<> used on a null pointer"), function doit, file /Users/kristoffercarlsson/julia/usr/include/llvm/Support/Casting.h, line 105.

signal (6): Abort trap: 6
in expression starting at /Users/kristoffercarlsson/julia/ccallable.jl:1
__pthread_kill at /usr/lib/system/libsystem_kernel.dylib (unknown line)
```

JeffBezanson:
fixes #35014

The approach I took is to add a field to `Method` storing the needed return type and signature if a C entry point is requested for that method. `jl_compile_extern_c` is called to inform the JIT during loading, or after loading a .ji or system image. The signatures are also added to the data passed to `jl_create_native` to tell it to include aliases in the system image.

A possible simplification is to require the argument types to be specific enough to be specialized. Currently, we allow `@ccallable f(x::Any) = ...`, in which case we need to generate a wrapper accepting any value and doing dynamic dispatch. If we required specific types, e.g. `f(x::Cint)`, then every ccallable would be associated with a single MethodInstance, and we'd just generate the alias whenever the instance is compiled.

It's not clear whether this feature is needed in the JIT, or just system images (i.e. only when you want to link a C program with direct calls against sys.so). The JIT case takes a bunch of the code here.

A separate problem to fix in the near future is that I believe ccallable and possibly also cfunction are always using jl_apply_generic when compiled into a system image, since `gen_cfun_wrapper` doesn't use the new mechanism for looking up call targets (there is already a TODO comment about the recursive compilation done there).

### DIFF
--- a/src/aotcompile.cpp
+++ b/src/aotcompile.cpp
@@ -274,7 +274,7 @@ void *jl_create_native(jl_array_t *methods, const jl_cgparams_t cgparams)
             // each item in this list is either a MethodInstance indicating something
             // to compile, or an svec(rettype, sig) describing a C-callable alias to create.
             jl_value_t *item = jl_array_ptr_ref(methods, i);
-            if (jl_is_simplevector(item)) {
+            if (jl_is_simplevector(item) && worlds == 1) {
                 jl_compile_extern_c(shadow_output, &params, NULL, jl_svecref(item, 0), jl_svecref(item, 1));
                 continue;
             }
--- a/src/aotcompile.cpp
+++ b/src/aotcompile.cpp
@@ -271,7 +271,14 @@ void *jl_create_native(jl_array_t *methods, const jl_cgparams_t cgparams)
             continue;
         size_t i, l;
         for (i = 0, l = jl_array_len(methods); i < l; i++) {
-            mi = (jl_method_instance_t*)jl_array_ptr_ref(methods, i);
+            // each item in this list is either a MethodInstance indicating something
+            // to compile, or an svec(rettype, sig) describing a C-callable alias to create.
+            jl_value_t *item = jl_array_ptr_ref(methods, i);
+            if (jl_is_simplevector(item)) {
+                jl_compile_extern_c(shadow_output, &params, NULL, jl_svecref(item, 0), jl_svecref(item, 1));
+                continue;
+            }
+            mi = (jl_method_instance_t*)item;
             src = NULL;
             // if this method is generally visible to the current compilation world,
             // and this is either the primary world, or not applicable in the primary world
--- a/src/jitlayers.cpp
+++ b/src/jitlayers.cpp
@@ -197,47 +197,72 @@ static jl_callptr_t _jl_compile_codeinst(
     return fptr;
 }
 
-// get the address of a C-callable entry point for a function
+Function *jl_generate_ccallable(void *llvmmod, void *sysimg_handle, jl_value_t *declrt, jl_value_t *sigt, jl_codegen_params_t &params);
+
+// compile a C-callable alias
 extern "C" JL_DLLEXPORT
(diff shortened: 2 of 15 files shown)


src/aotcompile.cpp
```diff
@@ -271,7 +271,14 @@ void *jl_create_native(jl_array_t *methods, const jl_cgparams_t cgparams)
             continue;
         size_t i, l;
         for (i = 0, l = jl_array_len(methods); i < l; i++) {
-            mi = (jl_method_instance_t*)jl_array_ptr_ref(methods, i);
+            // each item in this list is either a MethodInstance indicating something
+            // to compile, or an svec(rettype, sig) describing a C-callable alias to create.
+            jl_value_t *item = jl_array_ptr_ref(methods, i);
+            if (jl_is_simplevector(item)) {
```
vtjnash, line 277:
```suggestion
            if (jl_is_simplevector(item) && worlds == 1) {
```

src/jitlayers.cpp
```diff
@@ -197,47 +197,84 @@ static jl_callptr_t _jl_compile_codeinst(
...
+    for (i = 1; i < nargs; i++) {
+        jl_value_t *ati = jl_tparam(sigt, i);
+        if (jl_is_abstract_ref_type(ati)) {
+            ati = jl_tparam0(ati);
+            if (jl_is_typevar(ati))
+                jl_error("@ccallable: argument type Ref should have an element type, not Ref{<:T}");
+        }
+        if (jl_is_pointer(ati) && jl_is_typevar(jl_tparam0(ati)))
+            jl_error("@ccallable: argument type Ptr should have an element type, Ptr{<:T}");
+    }
```
vtjnash, line 265: Ah, I see this design is just handling all of the `Ref` declared signatures incorrectly. It seems part also of why they can't appropriately be attached to a single Method.

src/jitlayers.cpp
```diff
@@ -197,47 +197,84 @@ static jl_callptr_t _jl_compile_codeinst(
...
+
+// declare a C-callable entry point; called during code loading from the toplevel
 extern "C" JL_DLLEXPORT
-void jl_extern_c(jl_function_t *f, jl_value_t *rt, jl_value_t *argt, char *name)
-{
+void jl_extern_c(jl_value_t *declrt, jl_tupletype_t *sigt)
+{
+    // validate arguments. try to do as many checks as possible here to avoid
+    // throwing errors later during codegen.
+    JL_TYPECHK(@ccallable, type, declrt);
```
vtjnash, line 231: Since this function is named `ccallable`, I don't think the `@` is really a part of the name.

src/jitlayers.cpp
JeffBezanson: Yeah, I would be fine with not supporting `Ref` at all here --- no julia values, no by-reference, only simple types you'd write directly in C code. That would make a lot of this easier.

src/jitlayers.cpp
JeffBezanson: The user-facing API is `@ccallable`. I'll add an underscore to `ccallable` to make that clear.

JeffBezanson:
Review comments hopefully addressed now. For now I'm just going to be strict about what types are allowed, and later we can see if that can be leveraged to simplify the compilation flow.

KristofferC:
Also fixes https://github.com/JuliaLang/julia/issues/34076 I guess?

Issue #18379, opened 2016-09-06
Scalar type for broadcasting


TotalVerb:
I found myself needing a `Scalar` type for broadcasting purposes, to trick `broadcast` into _not_ trying to broadcast something, even though it is a matrix. An initial (bad) implementation is [here](https://github.com/TotalVerb/Scalars.jl).

With https://github.com/JuliaLang/julia/pull/16986 close to merging, I wonder if such a feature would be desirable in `Base`.

stevengj:
@TotalVerb, can you give an example of where you needed this?

stevengj:
Wouldn't storing something in a 0-dimensional array do the trick?

pabloferz:
After #16986 you might also wrap the array with a tuple as pointed out here https://github.com/JuliaLang/julia/issues/17411#issuecomment-236414603.

TotalVerb:
That's a good point; this `Scalar` type is just a one-element tuple.

TotalVerb:
@stevengj The current implementation of `Scalar` is basically just a 0-dimensional array, but it's a little slower than necessary because of the extra allocation.

TotalVerb:
@stevengj An example of where I needed this is to support broadcasting for `Basket`, which are a collection of currencies. It is useful to support both broadcasting over the individual instruments within the collection, and to support broadcasting the collections as if they were instruments in their own right.

Another example is given in the README for Scalars.jl.

stevengj:
(An aside: I don't think you need to define `show(io, "text/plain", x)` if you just want it to call `show(io, x)`, as the former calls the latter by default.)

TotalVerb:
@stevengj It's to override the method inherited from `AbstractArray`.

stevengj:
Oh, right.

stevengj:
A simpler solution, that wouldn't require us to define any new types, is for `broadcast` to treat `Ref(x)` as a 0-dimensional array of `x`.  This seems like something we should do anyway, because that's effectively what `Ref` is, and it seems relatively easy to add once #16986 is merged.

StefanKarpinski:
Related: https://github.com/JuliaLang/julia/issues/18271#issuecomment-245276731, wherein I proposed using `Ref(x)` with the syntax `&x` as a means of not dropping a scalar dimension when slicing or reducing. Of course, that entails considering `Ref` to be a 1-dimensional container, whereas this entails considering it to be a 0-dimensional container.

[...]

PR #35591, opened 2020-04-24
Recommend to use tuples to scalarize items in broadcast expressions

MasonProtter:
There's a lot of conflicting advice out there on what one should do in order to treat a given object as a scalar in a broadcast expression. Perhaps the most common practice is to use `Ref`, but this is arguably not the intended use of `Ref` and can be less performant than `tuple` because `Ref` (currently) blocks constant propagation. For a contrived example:
```julia
julia> using StaticArrays, BenchmarkTools

julia> us = (SA[1,2,3], SA[4,5,6]); v = SA[7,8,9]
3-element SArray{Tuple{3},Int64,1,3} with indices SOneTo(3):
 7
 8
 9

julia> @btime $us .+ Ref($v)
  2.839 ns (0 allocations: 0 bytes)
([8, 10, 12], [11, 13, 15])

julia> @btime $us .+ ($v,)
  0.020 ns (0 allocations: 0 bytes)
([8, 10, 12], [11, 13, 15])
```
This difference is perhaps inconsequential currently, but in a future where Julia is better able to reason about non-mutated arrays it could become more important, so I think it's better to recommend a more future-proof style. Of course, one could say that in that future `Ref` might also play better with constant prop, but I think that may be undesirable because `Ref` is often used for explicitly blocking constant prop, i.e. in benchmarking. 

____________________

I also modified an example that wrote `ceil.((UInt8,), [1.2 3.4; 5.6 6.7])` to instead say `ceil.(UInt8, [1.2 3.4; 5.6 6.7])` because there is no reason to wrap `UInt8` in a tuple since it is already a broadcast scalar.

### DIFF
--- a/base/refpointer.jl
+++ b/base/refpointer.jl
@@ -18,6 +18,17 @@ converted to a native pointer to the data it references.
 
 There is no invalid (NULL) `Ref` in Julia, but a `C_NULL` instance of `Ptr` can be passed to
 a `ccall` Ref argument.
+
+# Use in broadcasting
+`Ref` is sometimes used in broadcasting in order to treat the referenced values as a scalar:
+
+```jldoctest
+julia> isa.(Ref([1,2,3]), [Array, Dict, Int])
+3-element BitArray{1}:
+ 1
+ 0
+ 0
+```
 """
 Ref
 
--- a/base/refpointer.jl
+++ b/base/refpointer.jl
@@ -18,17 +18,6 @@ converted to a native pointer to the data it references.
 
 There is no invalid (NULL) `Ref` in Julia, but a `C_NULL` instance of `Ptr` can be passed to
 a `ccall` Ref argument.
-
-# Use in broadcasting
-
-Broadcasting with `Ref(x)` treats `x` as a scalar:
-```jldoctest
-julia> isa.(Ref([1,2,3]), [Array, Dict, Int])
-3-element BitArray{1}:
- 1
- 0
- 0
-```
 """
 Ref
 
--- a/doc/src/manual/arrays.md
+++ b/doc/src/manual/arrays.md
@@ -945,6 +945,16 @@ julia> string.(1:3, ". ", ["First", "Second", "Third"])
  "3. Third"
 ```
 
+Sometimes, you have a non-scalar object like an array that you wish to treat as a scalar in a particular
+broadcast expression. This is best done by wrapping that item in a [`Tuple`](@ref)
+```jldoctest
+julia> ([1, 2, 3], [4, 5, 6]) .+ ([1, 2, 3],)
+([2, 4, 6], [5, 7, 9])
+
+julia> ([1, 2, 3], [4, 5, 6]) .+ tuple([1, 2, 3])
+([2, 4, 6], [5, 7, 9])
+```
+
 ## Implementation
 
 The base array type in Julia is the abstract type [`AbstractArray{T,N}`](@ref). It is parameterized by
--- a/doc/src/manual/arrays.md
+++ b/doc/src/manual/arrays.md
@@ -945,8 +945,9 @@ julia> string.(1:3, ". ", ["First", "Second", "Third"])
  "3. Third"
 ```
 
-Sometimes, you have a non-scalar object like an array that you don't wish to broadcast into in a particular
-broadcast expression. This is best done by wrapping that item in a [`Tuple`](@ref)
+Sometimes, you want a container (like an array) that would normally participate in broadcast to be "protected" 
+from broadcast's behavior of iterating over all of its elements. By placing it inside another container 
+(like a single element [`Tuple`](@ref)) broadcast will treat it as a single value.
 ```jldoctest
 julia> ([1, 2, 3], [4, 5, 6]) .+ ([1, 2, 3],)
 ([2, 4, 6], [5, 7, 9])
--- a/doc/src/manual/arrays.md
+++ b/doc/src/manual/arrays.md
@@ -933,7 +933,7 @@ julia> convert.(Float32, [1, 2])
  1.0
  2.0
 
-julia> ceil.((UInt8,), [1.2 3.4; 5.6 6.7])
+julia> ceil.(UInt8, [1.2 3.4; 5.6 6.7])
 2×2 Array{UInt8,2}:
  0x02  0x04
  0x06  0x07
--- a/doc/src/manual/arrays.md
+++ b/doc/src/manual/arrays.md
@@ -945,7 +945,7 @@ julia> string.(1:3, ". ", ["First", "Second", "Third"])
  "3. Third"
 ```
 
-Sometimes, you have a non-scalar object like an array that you wish to treat as a scalar in a particular
+Sometimes, you have a non-scalar object like an array that you don't wish to broadcast into in a particular
 broadcast expression. This is best done by wrapping that item in a [`Tuple`](@ref)
 ```jldoctest
 julia> ([1, 2, 3], [4, 5, 6]) .+ ([1, 2, 3],)
--- a/doc/src/manual/arrays.md
+++ b/doc/src/manual/arrays.md
@@ -945,8 +945,8 @@ julia> string.(1:3, ". ", ["First", "Second", "Third"])
  "3. Third"
 ```
 
-Sometimes, you want a container (like an array) that would normally participate in broadcast to be "protected" 
-from broadcast's behavior of iterating over all of its elements. By placing it inside another container 
+Sometimes, you want a container (like an array) that would normally participate in broadcast to be "protected"
+from broadcast's behavior of iterating over all of its elements. By placing it inside another container
 (like a single element [`Tuple`](@ref)) broadcast will treat it as a single value.
 ```jldoctest
 julia> ([1, 2, 3], [4, 5, 6]) .+ ([1, 2, 3],)


tkf:
Ah, it's good to know that there is a pitfall with `Ref`.

But I think the problem with recommending using tuples is that a tuple is not always the "weakest container":

```julia
julia> 1 .+ (2,)
(3,)

julia> 1 .+ Ref(2)
3
```

Perhaps we need an immutable version of `Ref` (i.e., stack-allocatable zero-dimensional container)?

mkborregaard:
It might be unrelated to the issue at hand to discuss surface syntax, but I'd think
```
us .+ &v
```
was even clearer

MasonProtter:
@tkf Ah, great point I hadn't realized that. It's fairly simple to make such a stack allocated scalarizer like `Ref` (basically the same as `StaticArrays.Scalar` but without the ambiguity here)
```julia
struct Scalar{T}
    x::T
end
Base.iterate(s::Scalar) = (s.x, nothing)
Base.iterate( ::Scalar, s) = nothing
Base.ndims(::Scalar) = 0
Base.ndims(::Type{<:Scalar}) = 0
Base.length(::Scalar) = 1
Base.eltype(::Scalar{T}) where {T} = T
Base.size(::Scalar) = ()
Base.axes(::Scalar) = ()
Base.IteratorSize(::Type{<:Scalar}) = Base.HasShape{0}()
```
```julia
julia> using StaticArrays: SA

julia> using BenchmarkTools

julia> us = (SA[1,2,3], SA[4,5,6]); v = SA[7,8,9]
3-element StaticArrays.SArray{Tuple{3},Int64,1,3} with indices SOneTo(3):
 7
 8
 9

julia> @btime $us .+ Ref($v)
  2.919 ns (0 allocations: 0 bytes)
([8, 10, 12], [11, 13, 15])

julia> @btime $us .+ Scalar($v)
  29.538 ns (1 allocation: 112 bytes)
([8, 10, 12], [11, 13, 15])
```

Perhaps there could be an appetite to add such an object to `Base` for this purpose. 

_____

@mkborregaard I think that's an interesting syntax, but such things are a a bit of a hornets nest and rather not kick it here,

MasonProtter:
@tkf  Great point, I hadn't realized that. Such a stack allocated 0-dimensional container is indeed fairly easy to make. Here's one that's similar (but not the same as `StaticArrays.Scalar` which doesn't have the desired behaviour here):
```julia
struct Scalar{T}
    x::T
end
Base.getindex(s::Scalar) = s.x
Base.getindex(s::Scalar, ::CartesianIndex{0}) = s.x
Base.iterate(s::Scalar) = (s.x, nothing)
Base.iterate( ::Scalar, s) = nothing
Base.ndims(::Scalar) = 0
Base.ndims(::Type{<:Scalar}) = 0
Base.length(::Scalar) = 1
Base.eltype(::Scalar{T}) where {T} = T
Base.eltype(::Type{Scalar{T}}) where {T} = T
Base.size(::Scalar) = ()
Base.axes(::Scalar) = ()
Base.IteratorSize(::Type{<:Scalar}) = Base.HasShape{0}()
Base.broadcastable(s::Scalar) = s
```
now
```julia
julia> using StaticArrays: SA

julia> using BenchmarkTools: @btime

julia> us = (SA[1,2,3], SA[4,5,6]); v = SA[7,8,9]
3-element StaticArrays.SArray{Tuple{3},Int64,1,3} with indices SOneTo(3):
 7
 8
 9

julia> @btime $us .+ Scalar($v)
  0.019 ns (0 allocations: 0 bytes)
([8, 10, 12], [11, 13, 15])

julia> @btime 1 .+ Scalar(1)
  0.019 ns (0 allocations: 0 bytes)
2

```

tkf:
> but I think that may be undesirable because `Ref` is often used for explicitly blocking constant prop, i.e. in benchmarking.

I think `clobber`/`escape` https://github.com/JuliaCI/BenchmarkTools.jl/pull/92 would be the best strategy for this.  Having a specific tool to do this rather than relying on an inability of the compiler sounds like a good approach to me.  I think it'd be great if constant propagation can penetrate mutable objects when they are elided.

tkf:
It's great that you get `Scalar` working! FYI I found an old (but still open) discussion here: https://github.com/JuliaLang/julia/issues/18379

mbauman:
The only reason I've not done this sooner myself is that a one-tuple doesn't really act like a scalar.  For example:

```
julia> (1,) .+ 2
(3,)

julia> Ref(1) .+ 2
3
```

`Ref`s and other 0-dimensional containers actually are treated exactly like scalars.  1-tuples are not.  In practice it's generally fine, but I don't really like calling this the "best way to treat something like a scalar."

MasonProtter:
Yeah, that's fair I hadn't realized that subtlety until @tkf pointed it out. What do you think about that `Scalar` type (maybe with a different name)? Would it be just extra cruft if we added it to base?

mbauman:
I actually initially used such a type internally in initial sketches until I realized we could use `Ref` and there was the promise of that `&` syntax.  Further, I wasn't able to see any performance hit to using a mutable container as everything inlines... but this predated constant propagation!

MasonProtter:
Do you think the fact that we have constant propagation now warrants revisiting this design?

mbauman:
Yes, I would love for a better pattern here.  It's a common stumbling block, and to me that's a far bigger motivation for this change than the performance issue.

I have a fairly strong aversion to the name `Scalar`, though, because it's _explicitly not a scalar_.  It's a zero-dimensional container.  Yes, it means that what it contains gets treated as though it were a scalar, but that's because this `Scalar` thing itself is a container!

MasonProtter:
Sounds good. Do you or anyone else have name suggestions?

tkf:
I guess `Singleton` makes sense? It's a bit long and an overloaded word, though.

ViralBShah:
Bump. What's needed to merge this?

MasonProtter:
@ViralBShah The potentially blocking concern currently is that `Tuple`s don't really behave like scalars. Unlike `Ref`, `(1,) .+ 1 == (2,) != 2`.

If it's okay with people, I could ammend the wording to instead say something like 

>if you don't want to broadcast into a given container, you can wrap it in a single argument `Tuple`

and then give the same examples, that way I don't have misleading terms like "scalar".

base/refpointer.jl
```diff
@@ -18,17 +18,6 @@ converted to a native pointer to the data it references.
 
 There is no invalid (NULL) `Ref` in Julia, but a `C_NULL` instance of `Ptr` can be passed to
 a `ccall` Ref argument.
-
-# Use in broadcasting
-
-Broadcasting with `Ref(x)` treats `x` as a scalar:
```
mbauman, line 24: I'd be inclined to keep this — it is common to see `Ref` in broadcast expressions and folks might `?Ref` to see/understand why.  I'd be okay with the status quo wording... but if you feel inspired you could make it a bit more descriptive instead of prescriptive.

MasonProtter:
Done.

doc/src/manual/arrays.md
```diff
@@ -945,6 +945,16 @@ julia> string.(1:3, ". ", ["First", "Second", "Third"])
  "3. Third"
 ```
 
+Sometimes, you have a non-scalar object like an array that you wish to treat as a scalar in a particular
```
mbauman: Here's an alternative that avoids "scalar" entirely and aims at describing the _why_ instead of just prescribing the _what_.  This could surely be improved upon, but I think I favor talking about this invertedly — that is, "protecting from iteration" instead of "treating like scalar."

> Sometimes, you want a container (like an array) that would normally participate in broadcast to be "protected" from broadcast's behavior of iterating over all of its elements.  By placing it inside _another_ container (like a one-tuple) broadcast will treat it as a single value.

base/refpointer.jl
MasonProtter: Good points!

doc/src/manual/arrays.md
MasonProtter: Yep, that makes sense.

mbauman (approved):

tkf:
It's somewhat off-topic, but I think it is somewhat inconsistent that

```
julia> (20,) .+ ones(2, 2)
2×2 Array{Float64,2}:
 21.0  21.0
 21.0  21.0
```

actually works even though

```
julia> (20,)[1, 1]
ERROR: MethodError: no method matching getindex(::Tuple{Int64}, ::Int64, ::Int64)
...
```

because we often mention that `[20] .+ ones(2, 2)` works because `[20][1, 1]` works.

I guess it's another reason to have "immutable `Ref`" somewhere.

mbauman:
> because we often mention that `[20] .+ ones(2, 2)` works because `[20][1, 1]` works.

Funny, that's not my mental model at all.  I think of tuples (and other non-array containers) as being one dimensional, even if they don't support n-dimensional access, and broadcast just extrudes over both missing and singleton dimensions.

In other words, in my mental model, the array behavior is the special case (that it allows indexing by missing dimensions)... and it's not what drives broadcasting behavior.

tkf:
Ah, OK. I thought that's what _you_ would say as (I think) I've seen you were mentioning that numbers being iterable and indexable is nice because it's compatible with broadcasting. But I agree that filling extra dimensions fits well with what we have.

MasonProtter:
I believe this PR is now good to go.

rfourquet:
On bikesheding the name: instead of `Scalar`, what about `Only`? `only` is already used in a container context with only one element, and there would be the property `only(Only(x)) === x`.

tkf:
@rfourquet The discussion is happening at #35675.

rfourquet:
Ah thanks for the pointer!

PR #35592, opened 2020-04-24
remove timing stack fields when !ENABLE_TIMINGS

JeffBezanson:


vtjnash:
`ENABLE_TIMINGS` is never defined here

JeffBezanson:
Ah, indeed. Any reason not to include options.h in julia.h?

vtjnash:
Just unnecessary, since none of the other options affect the public ABI, we don't ship this file.

JeffBezanson:
I would argue it is necessary, for removing these fields.

JeffBezanson:
This has been unconditionally removed from jl_task_t, so I'm basically happy now.

PR #35595, opened 2020-04-25
[GCChecker] teach it how to deal with more complex expressions

vtjnash:
This helps the GCChecker deal with the atomics macros (by translating them to their equivalent non-atomic operation), and causes it to discover what appears to be true missing root in subtyping. (The first two commits could be pulled out and landed on master separately.)

### DIFF
--- a/src/clangsa/GCChecker.cpp
+++ b/src/clangsa/GCChecker.cpp
@@ -399,7 +399,6 @@ PDP GCChecker::GCValueBugVisitor::ExplainNoPropagationFromExpr(
     PathDiagnosticLocation Pos, BugReporterContext &BRC, BugReport &BR) {
   const MemRegion *Region =
       N->getState()->getSVal(FromWhere, N->getLocationContext()).getAsRegion();
-  SValExplainer Ex(BRC.getASTContext());
   SymbolRef Parent = walkToRoot(
       [&](SymbolRef Sym, const ValueState *OldVState) { return !OldVState; },
       N->getState(), Region);
@@ -612,7 +611,6 @@ bool GCChecker::propagateArgumentRootedness(CheckerContext &C,
 
   bool Change = false;
   int idx = 0;
-  SValExplainer Ex(C.getASTContext());
   for (const auto P : FD->parameters()) {
     if (!isGCTrackedType(P->getType())) {
       continue;
@@ -682,7 +680,6 @@ void GCChecker::checkBeginFunction(CheckerContext &C) const {
       C.addTransition(State);
     return;
   }
-  SValExplainer Ex(C.getASTContext());
   for (const auto P : FD->parameters()) {
     if (declHasAnnotation(P, "julia_require_rooted_slot")) {
       auto Param = State->getLValue(P, LCtx);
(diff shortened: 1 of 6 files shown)


test/clangsa/MissingRoots.c
```diff
@@ -67,20 +67,6 @@ void root_value_data2() {
...
-    jl_value_t **data = jl_svec_data(val);
-    JL_GC_PUSH1(&val); // expected-note{{GC frame changed here}}
-                       // expected-note@-1{{Value was rooted here}}
-    jl_gc_safepoint();
-    look_at_value(**&data);
-    JL_GC_POP(); // expected-note{{GC frame changed here}}
-                 // expected-note@-1{{Root was released here}}
-    jl_gc_safepoint(); // expected-note{{Value may have been GCed here}}
-    **&data; // expected-warning{{Creating derivative of value that may have been GCed}}
-             // expected-note@-1{{Creating derivative of value that may have been GCed}}
```
Keno, line 80: Why remove this test?

vtjnash: It doesn't pass anymore (I think it's too much state to explore now?), but also doesn't seem too realistic.

src/clangsa/GCChecker.cpp
```diff
@@ -926,6 +921,7 @@ bool GCChecker::processArgumentRooting(const CallEvent &Call, CheckerContext &C,
     return false;
   const MemRegion *RootingRegion = nullptr;
   SymbolRef RootedSymbol = nullptr;
+  StringRef FDName = FD->getDeclName().isIdentifier() ? FD->getName() : "";
```
Keno, line 924: Where is this used?

src/clangsa/GCChecker.cpp
```diff
@@ -1106,11 +1107,19 @@ void GCChecker::checkDerivingExpr(const Expr *Result, const Expr *Parent,
     }
   }
   // This is the pointer
-  SValExplainer Ex(C.getASTContext());
-  auto ValLoc = C.getSVal(Result).getAs<Loc>();
-  if (!ValLoc) {
-    return;
+  auto ResultVal = C.getSVal(Result);
+  if (ResultVal.isUnknown()) {
+    if (!Result->getType()->isPointerType()) {
```
Keno, line 1112: Is this right? What about references? Also are you sure you want a symbol for the derived expr? Wouldn't it make more sense to have a symbol for the parent and then take a subregion for the derived expr?

test/clangsa/MissingRoots.c
Keno: Ok, that seems odd, because the state should be fairly limited here, but I'm ok dropping this test if you need to make progress.

src/clangsa/GCChecker.cpp
vtjnash: It's what you do everywhere else. And it's required for getting back something compatible with `ResultVal.getAs<Loc>()`.

vtjnash: I don't know what this code does, since it's got almost no comments, so I just hammered at it until something passed. If you have a better way, we can do that.

test/clangsa/MissingRoots.c
vtjnash: Yeah—I had added this test just for the purpose of testing this PR, so I'm okay with removing it.

KristofferC:
Doesn't feel important to backport.

Issue #35597, opened 2020-04-25
`hash(a::AbstractArray)` assumes that `first(LinearIndices(a)) == 1`


goretkin:
```julia
julia> hash(Base.IdentityUnitRange(2:4))
ERROR: BoundsError: attempt to access 3-element LinearIndices{1,Tuple{Base.IdentityUnitRange{UnitRange{Int64}}}} with indices 2:4 at index [1]
Stacktrace:
 [1] throw_boundserror(::LinearIndices{1,Tuple{Base.IdentityUnitRange{UnitRange{Int64}}}}, ::Tuple{Int64}) at ./abstractarray.jl:538
 [2] checkbounds at ./abstractarray.jl:503 [inlined]
 [3] getindex at ./indices.jl:448 [inlined]
 [4] hash(::Base.IdentityUnitRange{UnitRange{Int64}}, ::UInt64) at ./abstractarray.jl:2210
 [5] hash(::Base.IdentityUnitRange{UnitRange{Int64}}) at ./hashing.jl:18
 [6] top-level scope at REPL[3]:1
```

goretkin:
(This means you cannot hash ndims=1 `OffsetArray`s.)

Peiffap:
@goretkin Should this issue be closed? The relevant PRs (#35598, #35624) have been merged already.

goretkin:
Looks like it!

PR #35598, opened 2020-04-25
Allow hashing 1D OffsetArrays

goretkin:
Addresses issue #35597


vtjnash (approved):

mbauman:
Thanks!

test/hashing.jl
```diff
@@ -102,9 +104,9 @@ for a in vals, b in vals
 end
 
 for a in vals
-    if a isa AbstractArray
-        @test hash(a) == hash(Array(a)) == hash(Array{Any}(a))
-    end
```
timholy, line 107: Instead of disabling the test for `IdentityUnitRange`, can we instead check against `OffsetArray(2:4, 1)`? You can get that from `testhelpers/`, see the start of `offsetarray.jl`.

timholy (commented):
Looks good, but see below about a better test

test/hashing.jl
mbauman: Good call. Added in https://github.com/JuliaLang/julia/pull/35624.

PR #35599, opened 2020-04-26
SPQR: allow user-selected ordering

dpo:
This PR adds the `ordering` kwarg to `qr()` for sparse matrices (which calls SuiteSparseQR) so the user can select one of the orderings exposed by SPQR.
The default value of `ordering` is `ORDERING_DEFAULT` so that the default behavior is unchanged.

A little care must be taken when returning from `_ldiv_basic` because if `ordering == ORDERING_FIXED`, SPQR returns an empty column permutation (to indicate that no permutation occurred). ~~I tried writing the `return` statement so it's more readable but got errors in the threaded SuiteSparse tests.~~

I added tests for over- and under-determined problems.

@andreasnoack

### DIFF
--- a/stdlib/SuiteSparse/src/spqr.jl
+++ b/stdlib/SuiteSparse/src/spqr.jl
@@ -17,6 +17,9 @@ const ORDERING_METIS   = Int32(6) # metis(A'*A)
 const ORDERING_DEFAULT = Int32(7) # SuiteSparseQR default ordering
 const ORDERING_BEST    = Int32(8) # try COLAMD, AMD, and METIS; pick best
 const ORDERING_BESTAMD = Int32(9) # try COLAMD and AMD; pick best#
+const ORDERINGS = [ORDERING_FIXED, ORDERING_NATURAL, ORDERING_COLAMD, ORDERING_CHOLMOD,
+                   ORDERING_AMD, ORDERING_METIS, ORDERING_DEFAULT, ORDERING_BEST,
+                   ORDERING_BESTAMD]
 
 # Let [m n] = size of the matrix after pruning singletons.  The default
 # ordering strategy is to use COLAMD if m <= 2*n.  Otherwise, AMD(A'A) is
@@ -39,6 +42,8 @@ function _qr!(ordering::Integer, tol::Real, econ::Integer, getCTX::Integer,
         H::Union{Ref{Ptr{CHOLMOD.C_Sparse{Tv}}}        , Ptr{Cvoid}} = C_NULL,
         HPinv::Union{Ref{Ptr{CHOLMOD.SuiteSparse_long}}, Ptr{Cvoid}} = C_NULL,
         HTau::Union{Ref{Ptr{CHOLMOD.C_Dense{Tv}}}      , Ptr{Cvoid}} = C_NULL) where {Tv<:CHOLMOD.VTypes}
+
+    ordering ∈ ORDERINGS || error("unknown ordering $ordering")
 
     AA   = unsafe_load(pointer(A))
     m, n = AA.nrow, AA.ncol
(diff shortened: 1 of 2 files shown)


stdlib/SuiteSparse/src/spqr.jl
```diff
@@ -419,7 +424,8 @@ function _ldiv_basic(F::QRSparse, B::StridedVecOrMat)
                         view(X0, Base.OneTo(rnk), :))
 
     # Apply right permutation and extract solution from X
-    return getindex(X, ntuple(i -> i == 1 ? invperm(F.cpiv) : :, Val(ndims(B)))...)
+    # NB: cpiv == [] if SPQR was called with ORDERING_FIXED
+    return length(F.cpiv) == 0 ? getindex(X, ntuple(i -> i == 1 ? (1:size(F,2)) : :, Val(ndims(B)))...) : getindex(X, ntuple(i -> i == 1 ? invperm(F.cpiv) : :, Val(ndims(B)))...)
```
vtjnash, line 428: Style nit: please keep line length reasonable. So usually about 1 anonymous function per expression.

stdlib/SuiteSparse/src/spqr.jl
```diff
@@ -17,6 +17,9 @@ const ORDERING_METIS   = Int32(6) # metis(A'*A)
 const ORDERING_DEFAULT = Int32(7) # SuiteSparseQR default ordering
 const ORDERING_BEST    = Int32(8) # try COLAMD, AMD, and METIS; pick best
 const ORDERING_BESTAMD = Int32(9) # try COLAMD and AMD; pick best#
+const ORDERINGS = [ORDERING_FIXED, ORDERING_NATURAL, ORDERING_COLAMD, ORDERING_CHOLMOD,
+                   ORDERING_AMD, ORDERING_METIS, ORDERING_DEFAULT, ORDERING_BEST,
+                   ORDERING_BESTAMD]
```
vtjnash, line 22:
```suggestion
const ORDERINGS = (ORDERING_FIXED, ORDERING_NATURAL, ORDERING_COLAMD, ORDERING_CHOLMOD,
                   ORDERING_AMD, ORDERING_METIS, ORDERING_DEFAULT, ORDERING_BEST,
                   ORDERING_BESTAMD)
```

ViralBShah:
Would be nice to get this in before feature freeze.

ViralBShah:
Specifying the ordering as a value from an array is a bit difficult to use. Perhaps use `:Ordering_name`. We should add all the orderings to the documentation, and also a NEWS entry.

stdlib/SuiteSparse/src/spqr.jl
dpo: I had trouble with this when I tested. That's what I try to explain above. Adding a return statement for the special case causes threaded tests to fail. I'm not sure why. I'll try again.

dpo:
You don't refer to them as a value in an array. You refer to them by name: `ORDERING_AMD`, `ORDERING_METIS`, etc.

ViralBShah:
Ah thanks. I didn't notice that was the case only in `ErrorException`.

stdlib/SuiteSparse/src/spqr.jl
dpo: Oh well, nevermind. I cleaned it up and it seems ok.

dpo:
Sorry I forgot to update docs and news. Will do in a separate PR.

dpo:
Could you indicate where I should modify the docs? Every time I read https://docs.julialang.org/en/v1/stdlib/LinearAlgebra/#LinearAlgebra.qr I'm surprised that it doesn't seem to talk about sparse QR at all, but I can't find another place that does. The sparse QR doesn't take a `blocksize` argument, and the factorization object doesn't have the `p` and `P` attributes (it has `pcol` and `prow`).

ViralBShah:
Yes, unfortunately none of the sparse factorizations have documentation. I think the right place to put them is in the sparse matrix implementation, and the doc system should find it. At least that's my understanding.

@fredrikekre should be able to confirm that.

andreasnoack:
All the sparse factorizations have docstrings but it does look like that they aren't included in the documentation. I guess we need to generate a file similar to https://github.com/JuliaLang/julia/blob/master/stdlib/LinearAlgebra/docs/src/index.md in the `stdlib/SuiteSparse` folder.

ViralBShah:
@andreasnoack Can you open a separate issue for that?

PR #35600, opened 2020-04-26
Improve typesubtract for tuples

### LINKED ISSUE #35566: MWE: `if x isa T` does not narrow the type in `else` block
This snippet

```julia
@noinline function step(acc, x)
    xs, = acc
    y = x > 0.0 ? x : missing
    if y isa eltype(xs)
        ys = push!(xs, y)
    else
        ys = vcat(xs, [y])
    end
    return (ys,)
end

@noinline function _foldl_iter(rf, val::T, iter, state) where {T}
    while true
        ret = iterate(iter, state)
        ret === nothing && break
        x, state = ret
        y = rf(val, x)
        if y isa T
            val = y
        else
            return _foldl_iter(rf, y, iter, state)
        end
    end
    return val
end

@code_typed optimize=false _foldl_iter(step, (Missing[],), [0.0], 1)
```

prints

```julia
CodeInfo(
1 ─       $(Expr(:meta, :noinline))
│         (val@_10 = val@_3)::Tuple{Array{Missing,1}}
└──       (state@_11 = state@_5)::Int64
2 ┄       goto #9 if not true
3 ─       Core.NewvarNode(:(x))::Any
│         Core.NewvarNode(:(@_8))::Any
│         Core.NewvarNode(:(y))::Any
│         (ret = Main.iterate(iter, state@_11))::Union{Nothing, Tuple{Float64,Int64}}
│   %9  = (ret === Main.nothing)::Bool
└──       goto #5 if not %9
4 ─       goto #9
5 ─ %12 = Base.indexed_iterate(ret::Tuple{Float64,Int64}, 1)::Core.Compiler.PartialStruct(Tuple{Float64,Int64}, Any[Float64, Core.Compiler.Const(2, false)])
│         (x = Core.getfield(%12, 1))::Float64
│         (@_8 = Core.getfield(%12, 2))::Core.Compiler.Const(2, false)
│   %15 = Base.indexed_iterate(ret::Tuple{Float64,Int64}, 2, @_8::Core.Compiler.Const(2, false))::Core.Compiler.PartialStruct(Tuple{Int64,Int64}, Any[Int64, Core.Compiler.Const(3, false)])
│         (state@_11 = Core.getfield(%15, 1))::Int64
│         (y = (rf)(val@_10, x))::Tuple{Union{Array{Missing,1}, Array{Union{Missing, Float64},1}}}
│   %18 = (y isa $(Expr(:static_parameter, 1)))::Bool
└──       goto #7 if not %18
6 ─       (val@_10 = y::Tuple{Array{Missing,1}})::Tuple{Array{Missing,1}}
└──       goto #8
7 ─ %22 = Main._foldl_iter(rf, y, iter, state@_11)::Tuple{Union{Array{Missing,1}, Array{Union{Missing, Float64},1}}}
└──       return %22
8 ─       goto #2
9 ┄       return val@_10
) => Tuple{Union{Array{Missing,1}, Array{Union{Missing, Float64},1}}}
```

I expect `%22 = Main._foldl_iter(rf, y, iter, state@_11)::Tuple{Array{Union{Missing, Float64},1}}` instead of `%22 = Main._foldl_iter(rf, y, iter, state@_11)::Tuple{Union{Array{Missing,1}, Array{Union{Missing, Float64},1}}}` which has a tuple-of-union.  This seems to be due that the argument `y` in `#7` is inferred to be `Tuple{Union{Array{Missing,1}, Array{Union{Missing, Float64},1}}}`.  (Indeed, this is what `@descend` shows me.  Also, I see a corresponding `@jl_apply_generic` in the LLVM IR.  IIUC this is a sign of a dynamic dispatch.)

The type is narrowed as expected if I tweak the example to avoid using intermediate tuple:

```julia
@noinline function step2(xs, x)
    y = x > 0.0 ? x : missing
    if y isa eltype(xs)
        ys = push!(xs, y)
    else
        ys = vcat(xs, [y])
    end
    return ys
end

@code_typed optimize=false _foldl_iter(step2, Missing[], [0.0], 1)
```

prints

```julia
...
7 ─ %22 = Main._foldl_iter(rf, y::Array{Union{Missing, Float64},1}, iter, state@_11)::Array{Union{Missing, Float64},1}
...
```

It'd be great if this works in the first example too.

Checked with Julia 1.5.0-DEV.630.

tkf:
Before:

```julia
julia> Core.Compiler.typesubtract(Tuple{Union{Int,Char}}, Tuple{Char})
Tuple{Union{Char, Int64}}
```

After:

```julia
julia> Core.Compiler.typesubtract(Tuple{Union{Int,Char}}, Tuple{Char})
Tuple{Int64}
```

close #35566

Unfortunately, I couldn't construct an example where the performance is improved by this inference yet, even though I can confirm that the methods are correctly inlined as expected.

### DIFF
--- a/base/compiler/typeutils.jl
+++ b/base/compiler/typeutils.jl
@@ -70,6 +70,12 @@ function typesubtract(@nospecialize(a), @nospecialize(b))
     if isa(a, Union)
         return Union{typesubtract(a.a, b),
                      typesubtract(a.b, b)}
+    elseif a isa DataType && b isa DataType && a.name === b.name === Tuple.name &&
+            length(a.types) == length(b.types)
+        ta = switchtupleunion(a)
+        if length(ta) > 1
+            return typesubtract(Union{ta...}, b)
+        end
     end
     return a # TODO: improve this bound?
 end
--- a/base/compiler/typeutils.jl
+++ b/base/compiler/typeutils.jl
@@ -70,11 +70,14 @@ function typesubtract(@nospecialize(a), @nospecialize(b))
     if isa(a, Union)
         return Union{typesubtract(a.a, b),
                      typesubtract(a.b, b)}
-    elseif a isa DataType && b isa DataType && a.name === b.name === Tuple.name &&
-            length(a.types) == length(b.types)
-        ta = switchtupleunion(a)
-        if length(ta) > 1
-            return typesubtract(Union{ta...}, b)
+    elseif a isa DataType
+        if b isa DataType
+            if a.name === b.name === Tuple.name && length(a.types) == length(b.types)
+                ta = switchtupleunion(a)
+                if length(ta) > 1
+                    return typesubtract(Union{ta...}, b)
+                end
+            end
         end
     end
     return a # TODO: improve this bound?
--- a/test/compiler/inference.jl
+++ b/test/compiler/inference.jl
@@ -2585,4 +2585,4 @@ end
 f() = _foldl_iter(step, (Missing[],), [0.0], 1)
 end
 @test Core.Compiler.typesubtract(Tuple{Union{Int,Char}}, Tuple{Char}) == Tuple{Int}
-@test Base.return_types(Issue35566.f) == [Val(:expected)]
+@test Base.return_types(Issue35566.f) == [Val{:expected}]
--- a/test/compiler/inference.jl
+++ b/test/compiler/inference.jl
@@ -2545,3 +2545,44 @@ end
 @test only(Base.code_typed(pickvarnames, (Vector{Any},), optimize=false))[2] == Tuple{Vararg{Union{Symbol, Tuple{Vararg{Union{Symbol, Tuple}}}}}}
 
 @test map(>:, [Int], [Int]) == [true]
+
+# issue 35566
+module Issue35566
+function step(acc, x)
+    xs, = acc
+    y = x > 0.0 ? x : missing
+    if y isa eltype(xs)
+        ys = push!(xs, y)
+    else
+        ys = vcat(xs, [y])
+    end
+    return (ys,)
+end
+
+function probe(y)
+    if y isa Tuple{Vector{Missing}}
+        return Val(:missing)
+    else
+        return Val(:expected)
+    end
+end
+
+function _foldl_iter(rf, val::T, iter, state) where {T}
+    while true
+        ret = iterate(iter, state)
+        ret === nothing && break
+        x, state = ret
+        y = rf(val, x)
+        if y isa T
+            val = y
+        else
+            return probe(y)
+        end
+    end
+    return Val(:expected)
+end
+
+f() = _foldl_iter(step, (Missing[],), [0.0], 1)
+end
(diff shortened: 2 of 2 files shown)


base/compiler/typeutils.jl
```diff
@@ -70,6 +70,12 @@ function typesubtract(@nospecialize(a), @nospecialize(b))
     if isa(a, Union)
         return Union{typesubtract(a.a, b),
                      typesubtract(a.b, b)}
+    elseif a isa DataType && b isa DataType && a.name === b.name === Tuple.name &&
+            length(a.types) == length(b.types)
```
vtjnash, line 74: It can be better if you write this as an `if` statement diamond, since inference currently can only propagate the last `isa` constraint on a line
```suggestion
    elseif a isa DataType
        if b isa DataType
            if a.name === b.name === Tuple.name && length(a.types) == length(b.types)
```

base/compiler/typeutils.jl
```diff
@@ -70,6 +70,12 @@ function typesubtract(@nospecialize(a), @nospecialize(b))
...
+        ta = switchtupleunion(a)
+        if length(ta) > 1
+            return typesubtract(Union{ta...}, b)
+        end
```
tkf, line 78: Thank you!

```suggestion
                ta = switchtupleunion(a)
                if length(ta) > 1
                    return typesubtract(Union{ta...}, b)
                end
            end
        end
```

base/compiler/typeutils.jl
aviatesk: > It can be better if you write this as an if statement diamond, since inference currently can only propagate the last isa constraint on a line

@vtjnash does this limit still hold true ? I tried the following snippet 
```julia
code_typed((Any,Any); optimize = false) do a, b
    if isa(a, Int) && isa(b, Int) && (a > 0 || b > 0)
        sin(a)
    end
end |> first |> first
```
and found `a` on line 3 is inferred as `Int` as expected:
```julia
CodeInfo(
1 ─ %1  = (a isa TypeProfiler.Int)::Bool
└──       goto #8 if not %1
2 ─ %3  = (b isa TypeProfiler.Int)::Bool
└──       goto #8 if not %3
3 ─ %5  = (a::Int64 > 0)::Bool
└──       goto #5 if not %5
4 ─       (@_4 = %5)::Bool
└──       goto #6
5 ─       (@_4 = b::Int64 > 0)::Bool
6 ┄       goto #8 if not @_4
7 ─ %11 = TypeProfiler.sin(a::Int64)::Float64
└──       return %11
8 ┄       return nothing
)
```

vtjnash: No, the frontend usually rewrites this syntax now to form the diamond automatically

aviatesk: nice, could you please tell me which PR/commit I should look at to see the frontend changes ?

vtjnash: https://github.com/JuliaLang/julia/pull/36355

aviatesk: thx

Issue #35596, opened 2020-04-25
Document how to run tests on Base after `Revise.track(Base)`


goretkin:
I'm not sure, but it seems like the Base tests (via `Base.runtests`) are not running against changes that are picked up by Revise.

https://github.com/JuliaLang/julia/blob/86ee57cdd77d402ea63108c16df7117d0ce83dc7/CONTRIBUTING.md#modifying-base-more-efficiently-with-revisejl

timholy:
Demo? It should work.

goretkin:
I noticed that part of the manual, but it didn't immediately make sense to me because I didn't jump from "target" to "make target".  So thanks for bringing it up. I'll see if I can't clarify that for other readers.

I still don't really understand the subsequent wording, however.

> load any modifications to Base into the current process before running the corresponding test

There would be no current process in the context of running `make`, right? Should it say "current system image"?

timholy:
Not sure I understand your question. This should work in either of two modes:
- from the Linux prompt: `make test-revise-sometest`
- from the REPL; `Revise.track(Base); include("test/sometest.jl")`

goretkin:
If I make a change to a base file, and then do `make test-revise-sometest`, there is "no current process" to load changes into. I don't have Julia running. My guess is that it meant to say a current system image.

My original issue was that I expected `Base.runtests("sometest")` to pull in changes via Revise. And it does not. Screencast coming shortly

timholy:
Thanks for the screencast. For the Revised version to be used, you have to run the test in the current process, i.e., via `include("test/hashing.jl")` rather than `Base.runtests("hashing")`. The difference is the former runs in the current process, the latter starts a fresh Julia process that doesn't by default bring in the changes.

We could add a keyword argument to `Base.runtests` that would trigger revision using the code already in `test/choosetests.jl` and `test/runtests.jl`.

goretkin:
Thanks for explaining that `Base.runtests` runs in a new process. I guess that's like `Pkg.test`.

Regarding the second point, I should have quoted in full:

> For convenience, there are also test-revise-* targets for every test-* target that use Revise to load any modifications to Base into the current process before running the corresponding test

This also starts a fresh Julia process that *does* bring in the changes. I don't understand what "current process" could possibly mean here.

timholy:
> I don't understand what "current process" could possibly mean here.

This was probably tweaked from earlier versions, so it may not be the best choice of phrasing. Really it means "into the process that will actually run the tests."

PR #35604, opened 2020-04-26
Clarify Revise and testing workflow

goretkin:
This didn't make sense to me when I first read it, so I'm trying to make an improvement. xref: https://github.com/JuliaLang/julia/issues/35596

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -86,6 +86,7 @@ Julia's documentation source files are stored in the `doc/` directory and all do
 make docs
 ```
 
+
 from Julia's root directory. This will rebuild the Julia system image, then install or update the package dependencies required to build the documentation, and finally build the HTML documentation and place the resulting files in `doc/_build/html/`.
 
 > **Note**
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -86,6 +86,7 @@ Julia's documentation source files are stored in the `doc/` directory and all do
 make docs
 ```
 
+
```
StefanKarpinski, line 89:
```suggestion
```

StefanKarpinski:
cc @Keno since you added the revise/test stuff.

CONTRIBUTING.md
```diff
@@ -242,9 +243,9 @@ If you need to restart your Julia session, just start at step 2 above.
 built and incorporate them automatically. You only need to rebuild
 Julia if you made code-changes that Revise cannot handle.
 
-For convenience, there are also `test-revise-*` targets for every `test-*`
-target that use Revise to load any modifications to Base into the current
-process before running the corresponding test. This can be useful as a shortcut
+For convenience, there are also `test-revise-*` targets for every [`test-*`
+target](https://github.com/JuliaLang/julia/blob/master/CONTRIBUTING.md#running-specific-tests) that use Revise to load any modifications to Base into the current
+system image before running the corresponding test. This can be useful as a shortcut
```
timholy, line 248: This makes it sound like the `*.so` file on disk is modified.

goretkin: Good point. I was looking for something more specific than "process", but maybe that's the right word to begin with.

The system image is the serialization of ______ to disk. The state of the julia runtime?

ViralBShah:
I'm merging since this feels like an improvement, but if we should be rewording, please chime in and I can update it in a separate PR.

PR #35606, opened 2020-04-26
remove `Ref` allocation on task switch

JeffBezanson:
pfib(31) before:
```
  3.180960 seconds (14.34 M allocations: 1.512 GiB, 28.57% gc time)
```
after:
```
  2.835908 seconds (10.89 M allocations: 1.461 GiB, 21.35% gc time)
```

The times are pretty variable, and the GC is doing a good job here, but reducing the allocation workload here is surely worth it.

(also needs #35589 to get this result)

### DIFF
--- a/base/task.jl
+++ b/base/task.jl
@@ -615,6 +615,8 @@ function yield()
     end
 end
 
+@inline set_next_task(t::Task) = ccall(:jl_set_next_task, Cvoid, (Any,), t)
+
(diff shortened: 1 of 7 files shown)


vtjnash (approved):
It's an odd seeming API, but it is what we need to do under-the-hood, so makes sense to lift it to the higher level.

PR #35607, opened 2020-04-27
Update PCRE2 version to 10.34. Fixes #35322 #35459

### LINKED ISSUE #35322: Avoidable "PCRE compilation error" by allowing more than ASCII
Missing "PCRE_UTF8 option" for pcre_compile() or something else?

"ð" (and "þ") are some of he extra letters in the Icelandic alphabet, and I would want to make sure those work too.

From memory same kind of regex worked in Python, and as I was new to this, it took a long time to figure out what was wrong... (porting from Python something that supposedly should have worked) I had some longer regex end i was (too) tedious to count out to the right letter:

```
julia> r"""(?P<viðburðarnafn>((\w)*))"""
ERROR: LoadError: PCRE compilation error: syntax error in subpattern name (missing terminator) at offset 6
Stacktrace:
 [2] compile(::String, ::UInt32) at ./pcre.jl:123
 [3] compile(::Regex) at ./regex.jl:72
 [4] Regex(::String, ::UInt32, ::UInt32) at ./regex.jl:37
 [5] Regex(::String) at ./regex.jl:60
 [6] @r_str(::LineNumberNode, ::Module, ::Any) at ./regex.jl:109
in expression starting at REPL[51]:1
julia> versioninfo
versioninfo (generic function with 2 methods)

julia> versioninfo()
Julia Version 1.5.0-DEV.360
Commit 012b270df6 (2020-02-28 07:57 UTC)
Platform Info:
  CPU: Intel(R) Xeon(R) CPU D-1541 @ 2.10GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-9.0.1 (ORCJIT, broadwell)
Environment:
  JULIA_LLVM_ARGS = -unroll-threshold=500
  JULIA_NUM_THREADS = 4
```

http://pcre.sourceforge.net/pcre.txt

UTF-8 SUPPORT

     To build PCRE with support for UTF-8 character strings, add

       --enable-utf8

     to the configure command. Of itself, this does not make PCRE
     treat  strings as UTF-8. As well as compiling PCRE with this
     option, you also have have to set the PCRE_UTF8 option  when
     you call the pcre_compile() function.

### LINKED ISSUE #35459: Unicode named groups for regular expressions
Hello,

This issue is strongly related to my old (more than 1 year) message [here](https://discourse.julialang.org/t/how-to-use-unicode-characters-in-regex-named-groups/20178/2). 

It is to allow Unicode named groups for regular expressions, it seems it was just related to a limitation of PCRE, which was [updated](https://vcs.pcre.org/pcre2/code/trunk/src/pcre2test.c?r1=1041&r2=1066) to fix it last year.

I remember I tried near the end of last year the latest PCRE2 and it worked, and because I just tried again on the latest Julia 1.4 without success, I read again this topic and it is true they asked me to open an issue here! Done!

Sincerely.

Micket:
Fixes #35322, fixes #35459 

Depends on new release build for https://github.com/JuliaPackaging/Yggdrasil/tree/master/P/PCRE2
Semi-related https://github.com/JuliaPackaging/Yggdrasil/issues/940


vtjnash:
Good to know there's a new release with fixes. What's the status of the equivalent PR to Yggdrasil?

Micket:
@vtjnash They already bumped the version string to 10.34
https://github.com/JuliaPackaging/Yggdrasil/blob/master/P/PCRE2/build_tarballs.jl#L6
they just haven't build a tarball for release as far as I can tell. I don't know how this packaging thing works.

KristofferC:
>  they just haven't build a tarball for release as far as I can tell. I

From what I can see, the release with the tarballs has been made (I think it happens automatically when merging a PR to Yggdrasil):

https://github.com/JuliaBinaryWrappers/PCRE2_jll.jl/releases

vtjnash:
Thanks. So we just need to run `./contrib/refresh_bb_tarballs.sh pcre` on this?

KristofferC:
I think this URL needs to be tweaked?

https://github.com/JuliaLang/julia/blob/5cb3f3fbc1be7bc7a4d0b66ef69bfcf2e1417fe0/deps/pcre.mk#L58

Micket:
@KristofferC I suspect this one is also wrong then?
https://github.com/JuliaLang/julia/blob/04234fb829327ac9e056b9e7dd4788ad23e268cf/deps/libuv.mk#L67

But these ones should still use Yggdrasil?
https://github.com/JuliaLang/julia/blob/c9786e6667b436dff6501e5c6474744e44f42990/deps/dsfmt.mk#L68
https://github.com/JuliaLang/julia/blob/e6dd72f7a6afbd28e779eb7786fdebead0b4fca7/deps/unwind.mk#L107

Anyway, I'll update this PR with the correct url for pcre (when I get back home)

KristofferC:
> I suspect this one is also wrong then?

I am not sure. It might be that only some of these have transitioned to the JuliaBinaryWrappers organization.

ViralBShah:
Yes, we stopped publishing to Yggdrasil releases. The old releases may be there, but the new ones are likely to be in JuliaBinaryWrappers.

giordano:
Yes, until ~August last year the tarballs were published under Yggdrasil, but then we started using the more well organised JLL packages which are under the JuliaBinaryWrappers organisation.  In the Julia build system only when necessary a newer version of a tarball is used and all the new ones (i.e., after ~August) will be found under JuliaBinaryWrappers.

KristofferC (approved):
LGTM but would be good with a quick look from @staticfloat as well.

KristofferC:
Actually, the old checksum files should maybe be removed?

deps/pcre.mk
```diff
@@ -55,7 +55,7 @@ fastcheck-pcre: check-pcre
 check-pcre: $(BUILDDIR)/pcre2-$(PCRE_VER)/build-checked
 
 else # USE_BINARYBUILDER_PCRE
-PCRE_BB_URL_BASE := https://github.com/JuliaPackaging/Yggdrasil/releases/download/PCRE2-v$(PCRE_VER)-$(PCRE_BB_REL)
+PCRE_BB_URL_BASE := https://github.com/JuliaBinaryWrappers/PCRE2_jll.jl/releases/download/PCRE2-v$(PCRE_VER).0+$(PCRE_BB_REL)
```
KristofferC, line 58: Should this `.0` be moved to `PCRE_VER` in `Make.inc` instead?

Micket: Then it wouldn't work when downloading the sources instead:
https://github.com/JuliaLang/julia/blob/master/deps/pcre.mk#L9
(I'm not sure why PCRE2_jll insist on adding another number suffix, it really is just 10.34, https://ftp.pcre.org/pub/pcre/)

giordano: > (I'm not sure why PCRE2_jll insist on adding another number suffix, it really is just 10.34, ftp.pcre.org/pub/pcre)

Because JLL packages use Julia's `VersionNumber` representation in string interpolation:
```julia
julia> "$(v"10.34")"
"10.34.0"
```

staticfloat:
Looks good to me, but yes, please delete the old PCRE2 checksums as well.

Keno:
We've been seeing intermittent PCRE errors after this. I'm gonna revert this for 1.5 and we can re-apply and investigate for 1.6.

StefanKarpinski:
It's possible we can unrevert this after https://github.com/JuliaLang/julia/pull/35773.

BenjaminGalliot:
Does it still fail?

BenjaminGalliot:
@Keno, do people still have this OS X issue?

PR #35613, opened 2020-04-27
Fix modf(Inf)

### LINKED ISSUE #35612: modf(Inf) inconsistencies
```
julia> modf(-Inf16)
(NaN16, -Inf16)

julia> modf(-Inf32)
(-0.0f0, -Inf32)

julia> modf(-Inf64)
(-0.0, -Inf)

julia> modf(BigFloat(-Inf))
(NaN, -Inf)
```
The C documentation that I have agrees with the `Float32`/`Float64` results.

jmkuhn:
Fixes https://github.com/JuliaLang/julia/issues/35612.

### DIFF
--- a/base/math.jl
+++ b/base/math.jl
@@ -868,7 +868,7 @@ julia> modf(-3.5)
 (-0.5, -3.0)
 ```
 """
-modf(x) = rem(x,one(x)), trunc(x)
+modf(x) = isinf(x) ? (flipsign(zero(x), x), x) : (rem(x, one(x)), trunc(x))
 
 function modf(x::Float32)
     temp = Ref{Float32}()
--- a/base/mpfr.jl
+++ b/base/mpfr.jl
@@ -691,7 +691,6 @@ function min(x::BigFloat, y::BigFloat)
 end
 
 function modf(x::BigFloat)
-    isinf(x) && return (BigFloat(NaN), x)
     zint = BigFloat()
     zfloat = BigFloat()
     ccall((:mpfr_modf, :libmpfr), Int32, (Ref{BigFloat}, Ref{BigFloat}, Ref{BigFloat}, MPFRRoundingMode), zint, zfloat, x, ROUNDING_MODE[])
--- a/test/math.jl
+++ b/test/math.jl
@@ -536,6 +536,9 @@ end
         @test modf( convert(elty,1.2) )[2] ≈ convert(elty,1.0)
         @test modf( convert(elty,1.0) )[1] ≈ convert(elty,0.0)
         @test modf( convert(elty,1.0) )[2] ≈ convert(elty,1.0)
+        @test isequal(modf( convert(elty,-Inf) ), (-0.0, -Inf))
+        @test isequal(modf( convert(elty,Inf) ), (0.0, Inf))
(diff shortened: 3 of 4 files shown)


DNF2:
This has a performance implication (`modf_` is the new implementation):
```
julia> @btime y .= modf.(x) setup=(x=(10 .* rand(10^4) .- 5); y=Vector{Tuple{Float64,Float64}}(undef, length(x)));
  55.126 μs (0 allocations: 0 bytes)

julia> @btime y .= modf_.(x) setup=(x=(10 .* rand(10^4) .- 5); y=Vector{Tuple{Float64,Float64}}(undef, length(x)));
  122.888 μs (0 allocations: 0 bytes)
```

Maybe this can be fixed by dropping `rem` entirely?
```
julia> remf(x) = isinf(x) ? (flipsign(zero(x), x), x) : (a = trunc(x); (x-a, a))

julia> @btime y .= remf.(x) setup=(x=(10 .* rand(10^4) .- 5); y=Vector{Tuple{Float64,Float64}}(undef, length(x)));
  13.832 μs (0 allocations: 0 bytes)
```

(Another issue is the name, which probably shouldn't be `modf` at all.)

jmkuhn:
The performance difference that you are seeing is because `Float32` and `Float64` don't use this method at all.  They call directly into `libm`.

```
julia> modfo(x) = rem(x,one(x)), trunc(x)
modfo (generic function with 1 method)

julia> modfn(x) = isinf(x) ? (flipsign(zero(x), x), x) : (rem(x, one(x)), trunc(x))
modfn (generic function with 1 method)

julia> @btime y .= modf.(x) setup=(x=(10 .* rand(10^4) .- 5); y=Vector{Tuple{Float64,Float64}}(undef, length(x)));
  43.125 μs (0 allocations: 0 bytes)

julia> @btime y .= modfo.(x) setup=(x=(10 .* rand(10^4) .- 5); y=Vector{Tuple{Float64,Float64}}(undef, length(x)));
  109.316 μs (0 allocations: 0 bytes)

julia> @btime y .= modfn.(x) setup=(x=(10 .* rand(10^4) .- 5); y=Vector{Tuple{Float64,Float64}}(undef, length(x)));
  111.461 μs (0 allocations: 0 bytes)
```
The new version that you suggested also gives incorrect results
```
julia> modfnn(x) = isinf(x) ? (flipsign(zero(x), x), x) : (a = trunc(x); (x-a, a))
modfnn (generic function with 1 method)

julia> modf(-0.0)
(-0.0, -0.0)

julia> modfnn(-0.0)
(0.0, -0.0)
```

DNF2:
I see. So this is only for `Float16` and `BigFloat`. In that case there is no significant performance drop relative to the current implementation.


I'll just note, however, that even the libm version seems quite slow, and that all the non-`BigFloat`s improve performance if we avoid `rem`.

A faster version for `Float16` is:
```
modf(x::Float16) = Float16.(modf(Float32(x)))
```

Tweaking my original suggestion, this improves significantly on the libm performance for all the non-`BigFloat`s:
```
function truncf(x)
    isinf(x) && return (flipsign(zero(x), x), x)
    a = trunc(x)
    b = x - a
    return (flipsign(b, a*b), a)
end
```

PR #35617, opened 2020-04-27
Support setting MCPU during Make, for ARM

IanButterworth:
As is outlined in https://community.arm.com/developer/tools-software/tools/b/tools-software-ides-blog/posts/compiler-flags-across-architectures-march-mtune-and-mcpu

> In short, you cannot support both x86 and Arm with exactly the same compiler flags.  You must use -mcpu for Arm and -march for x86.

I also opened up https://github.com/JuliaCI/julia-buildbot/pull/149 to switch arm builds to mcpu, before realizing I needed to also adjust the Make file here

### DIFF
--- a/Make.inc
+++ b/Make.inc
@@ -686,7 +686,7 @@ override ARCH := $(shell $(CC) -dumpmachine | sed "s/\([^-]*\).*$$/\1/")
 else
 ifneq ($(XC_HOST),)
 XC_HOST := $(ARCH)$(shell echo $(XC_HOST) | sed "s/[^-]*\(.*\)$$/\1/")
-ifneq (,$(filter ,$(findstring arm, $(ARCH)) $(findstring aarch64, $(ARCH))))
+ifneq ($(findstring arm, $(ARCH))$(findstring aarch64, $(ARCH)),)
 MCPU := $(subst _,-,$(ARCH)) # Arm prefers MCPU over MARCH
 else
 MARCH := $(subst _,-,$(ARCH))
@@ -819,14 +819,12 @@ OPENBLAS_DYNAMIC_ARCH:=0
 OPENBLAS_TARGET_ARCH:=ARMV8
 endif
 
-do_step =
 # Set MARCH-specific flags
 ifneq ($(MARCH),)
 CC += -march=$(MARCH)
 CXX += -march=$(MARCH)
 FC += -march=$(MARCH)
 JULIA_CPU_TARGET ?= $(MARCH)
-do_step = yes
 endif
 
 # Set MCPU-specific flags
@@ -835,10 +833,9 @@ CC += -mcpu=$(MCPU)
 CXX += -mcpu=$(MCPU)
 FC += -mcpu=$(MCPU)
 JULIA_CPU_TARGET ?= $(MCPU)
-do_step = yes
 endif
 
-ifdef do_step
+ifneq ($(MARCH)$(MCPU),)
 ifeq ($(OS),Darwin)
 # on Darwin, the standalone `as` program doesn't know
 # how to handle AVX instructions, but it does know how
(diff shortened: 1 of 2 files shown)


Make.inc
```diff
@@ -815,12 +819,26 @@ OPENBLAS_DYNAMIC_ARCH:=0
...
+# Set MCPU-specific flags
+ifneq ($(MCPU),)
+CC += -mcpu=$(MCPU)
+CXX += -mcpu=$(MCPU)
+FC += -mcpu=$(MCPU)
+JULIA_CPU_TARGET ?= $(MCPU)
+do_step = yes
+endif
+
+ifdef do_step
```
ianshmean, line 841: There might be a better way to do the OR logic here

yuyichao:
I think you can just pass in `CFLAGS` etc? Different from `MARCH` it's not like this is going to be used anywhere else.

Also, this isn't very ARM specific (and FWIW the condition doesn't seem to capture aarch64 which is arguably more important) though I think it could make sense to forward native march to mtune automatically. According to that article this should work on both x86 and arm.

ianshmean:
The paradigm just seems to be different for ARM, so I thought it best to directly support MCPU?

Also, that article claims
> LLVM compilers only support “native” for the -mcpu and -mtune flags.  You cannot use -march=native with LLVM-based compilers.

Is that right? If `march=native` wasn't being passed to LLVM for user code because of that, perhaps there's reason to now add `-mcpu=native` for user code?

I'm wading into an area I'm not familiar with so apologies if these are flawed suggestions

ianshmean:
And on the condition for catching arm & aarch64, you're right.. I need to think of a better catch there

yuyichao:
> Is that right?

Not sure if that's right but this PR doesn't change that it seems.

> If march=native wasn't being passed to LLVM for user code because of that, perhaps there's reason to now add -mcpu=native for user code?

All of the flags mentioned here (`march`, `mcpu`, `mtune`), no matter what you passed in, has nothing to do with user code. `MARCH` is used as default for `JULIA_CPU_TARGET` but even that has zero effect on the user code.

ianshmean:
> Not sure if that's right but this PR doesn't change that it seems.

Sorry, I didn't mean to imply it was. That was a tangential question. I think I'm concluding that I don't think I know enough about the internals of the language to venture any further.

I just tweaked the logic to catch `*arm*` and `*aarch64*` in the `XC_HOST` section. Maybe there's a more robust method

yuyichao:
> Sorry, I didn't mean to imply it was. That was a tangential question. I think I'm concluding that I don't think I know enough about the internals of the language to venture any further.

What I meant is that if it is currently used and we didn't get an error then adding `-mtune` should be enough...

ianshmean:
As a reminder to myself, if this goes ahead we'd probably want to update the advice in the Arm build tips documentation to use `MCPU=native`

Make.inc
vtjnash:
```suggestion
ifneq ($(MARCH)$(MCPU),)
```

Make.inc
```diff
@@ -686,7 +686,11 @@ override ARCH := $(shell $(CC) -dumpmachine | sed "s/\([^-]*\).*$$/\1/")
 else
 ifneq ($(XC_HOST),)
 XC_HOST := $(ARCH)$(shell echo $(XC_HOST) | sed "s/[^-]*\(.*\)$$/\1/")
+ifneq (,$(filter ,$(findstring arm, $(ARCH)) $(findstring aarch64, $(ARCH))))
```
vtjnash, line 689: What does filter do without a filter list? Is this the same as strip?

vtjnash (approved):

Make.inc
ianshmean: Great, thanks. Incorporated

Make.inc
ianshmean: On review I think this wouldn't  have worked. I've updated to the OR format you suggested for the other case. Thanks

staticfloat (approved):

PR #35618, opened 2020-04-28
change default ios buffer size to 32k

### LINKED ISSUE #34449: Reduce IOS_BUFSIZE to 32kb
Most other languages that buffer reads from files seem to have a default of reading whatever the platform's block size is (or sometimes simply 8 kb). Julia on the other hand will always [buffer up to 128 kb](https://github.com/JuliaLang/julia/blob/faefe2ae64fb9a12824832582e2ddd3bc554d372/src/support/ios.h#L23) no matter the platform. This was introduced way back in 2011 in commit e380648 (@JeffBezanson, do you remember the justification for the change?).

There may well be a good reason to have a higher value for this, but I can't find documentation for it, and this buffer allocation does cause much higher memory usage when reading from large files as compared to other languages.

### python

```
❯ python3 -c "import io; print(io.DEFAULT_BUFFER_SIZE)"
8192
```

### Rust

8192

https://doc.rust-lang.org/1.40.0/src/std/sys_common/io.rs.html#1

### Java

8192

https://stackoverflow.com/questions/16973843/bufferedreader-default-buffer-size

JeffBezanson:
fixes #34449

Also stop over-allocating ios_t buffer by 1 byte. This should make allocations more predictable if the buffer size becomes customizable.

### DIFF
--- a/src/flisp/iostream.c
+++ b/src/flisp/iostream.c
@@ -352,10 +352,13 @@ value_t fl_ioreaduntil(fl_context_t *fl_ctx, value_t *args, uint32_t nargs)
     cv->len = n;
     if (dest.buf != data) {
         // outgrew initial space
-        cv->data = dest.buf;
+        size_t sz;
+        cv->data = ios_take_buffer(&dest, &sz);
         cv_autorelease(fl_ctx, cv);
     }
-    ((char*)cv->data)[n] = '\0';
+    else {
+        ((char*)cv->data)[n] = '\0';
+    }
     if (n == 0 && ios_eof(src))
         return fl_ctx->FL_EOF;
     return str;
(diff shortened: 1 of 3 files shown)


vtjnash (approved):

vtjnash:
From CI, looks like there's now possibly an off-by-one error somewhere, though the change looks okay by eye.

JeffBezanson:
Yeah this must have messed something up but so far I can't figure out what.

JeffBezanson:
Ok, the problem was that `vsnprintf` was returning a negative value when there wasn't enough space on windows (obviously more likely if the buffer is smaller). According to MS' docs that's the behavior of `_vsnprintf`, so we must actually be calling that. That caused us to skip some output when generating julia_flisp.boot, which caused the space character (`#\ `) to be included in the table of operator characters, causing the parse errors seen in CI. All in all, a pretty fun bug.

PR #35620, opened 2020-04-28
Enable combined BroadcastStyle dispatch in materialize!

mbauman:
In-place broadcast (the `.=` syntax) now allows for custom implementations to dispatch on the
combined `BroadcastStyle` between the destination and the right-hand side for by implementing
`materialize!(::BroadcastStyle, dest, bc)`.

cc @vchuravy

### DIFF
--- a/base/broadcast.jl
+++ b/base/broadcast.jl
@@ -820,11 +820,16 @@ Take a lazy `Broadcasted` object and compute the result
 """
 @inline materialize(bc::Broadcasted) = copy(instantiate(bc))
 materialize(x) = x
-@inline function materialize!(dest, bc::Broadcasted{Style}) where {Style}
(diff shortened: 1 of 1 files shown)


vchuravy:
Thanks Matt, to expand upon the need. Currently there are several places where we have a static union of wrapper arrays that are allowed destinations for broadcast: https://github.com/JuliaGPU/GPUArrays.jl/blob/71715e99e428636377fece74c329beebfcb58d33/src/host/broadcast.jl#L28

By adding another level of dispatch and using the broadcast style we can do away with that and select the right broadcast implementation, even if we have multiple layers of wrapper arrays.

mbauman:
I should also note that I went to go add NEWS and docs... only to remember that `materialize!` itself was never really documented as part of the official interface; `copyto!` was intended to fill that role.  Clearly that is insufficient, but I'm hesitant to yet further complicate the documentation.

PR #35622, opened 2020-04-28
gc: fix MEMDEBUG build

vtjnash:
Found this commit on my machine. I think it makes sense?

### DIFF
--- a/src/gc.c
+++ b/src/gc.c
@@ -725,6 +725,7 @@ STATIC_INLINE void gc_setmark_pool_(jl_ptls_t ptls, jl_taggedvalue_t *o,
 #ifdef MEMDEBUG
     gc_setmark_big(ptls, o, mark_mode);
 #else
+    jl_assume(page);
     if (mark_mode == GC_OLD_MARKED) {
         ptls->gc_cache.perm_scanned_bytes += page->osize;
         jl_atomic_fetch_add_relaxed(&page->nold, 1);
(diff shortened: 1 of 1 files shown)


yuyichao:
whats wrong with the current one?

vtjnash:
I expect `gc_setmark` would crash on the first GC call

yuyichao:
Because of the `jl_assume`? I think it was written this way to have the minimum conditional compilation possible.

yuyichao:
And if it was `jl_assume` the fix should be to just remove `jl_assume`. It seems that the compiler should make that assumption anyway.

vtjnash:
Yeah, that helps the compiler emit better code, but it's not valid here

yuyichao:
I think the assume isn't needed here though if you want it can just be moved into ser_pool_

vtjnash:
It helps skip several branches, and reflects the expectation on the loads on the subsequents statements.

yuyichao:
> It helps skip several branches, and reflects the expectation on the loads on the subsequents statements.

The load is exactly why I said this is not needed. The compiler should be fully aware of that.
And as I said, if explicit documentation of such assumption is needed, it can just move into `gc_setmark_pool_` without having to tripple the number of related `#if(n)def MEMDEBUG`.

vtjnash:
> The compiler should be fully aware of that

Last I checked, that'd be impossible for the compiler to know, but as I said, this was just a commit I found on my machine from awhile ago.

yuyichao:
> Last I checked, that'd be impossible for the compiler to know, but as I said, this was just a commit I found on my machine from awhile ago.

It seems that the compiler does not use this info if the unconditional load is after the check (even if the load happens in both branches).

In any case, as I said, again, I think you can just move the `assume` into `gc_setmark_pool_`. It should do exactly the same thing with a much smaller patch and easier to maintain.

Issue #33212, opened 2019-09-10
Sys.isexecutable() unreliable on Windows


staticfloat:
The current implementation of `Sys.isexecutable()` is broken on Windows; it does not properly query the relevant ACLs, and simply returns `true` if the file exists.

It is necessary for us to emulate the proper behavior here as git tree hashing requires knowing if a file is executable or not, and the native `git` tools on Windows do "the right thing", so in order to hash to the same value we will need to do the right thing as well.

The logic to do this is complex, but reducable.  An example from cygwin is given here: https://github.com/Alexpux/Cygwin/blob/9c84bfd47922aad4881f80243320422b621c95dc/winsup/cygwin/sec_acl.cc#L636-L644

musm:
@staticfloat I have a solution here, utilizing  `SHGetFileInfo`  (https://docs.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-shgetfileinfoa) By obtaining the file's `SHGFI_EXETYPE` (https://docs.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-shgetfileinfoa#return-value)

 
It can provide fine grain information on whether the file is: 
> Windows application.
> MS-DOS .exe or .com file
> Console application or .bat file

I didn't really look at the logic here, but it seems much complex than using the Windows API directly:
> The logic to do this is complex, but reducable. An example from cygwin is given here: https://github.com/Alexpux/Cygwin/blob/9c84bfd47922aad4881f80243320422b621c95dc/winsup/cygwin/sec_acl.cc#L636-L644

Should I open a PR with the stated changes?

StefanKarpinski:
@vtjnash, does this seem like a reasonable approach to you?

staticfloat:
If the docs are accurate then yes, this does seem reasonable to me.  Good find, @musm!  I think a small PoC would be really helpful to both myself and Jameson, and doesn't look too hard to put together.

musm:
I can just open a PR with a working implementation which I am close to finishing

vtjnash:
This would require having the isexecutable check calling the AccessCheck function (https://docs.microsoft.com/en-us/windows/win32/secauthz/verifying-client-access-with-acls-in-c--). We could also implement this in `uv_fs_access`, and then use that as a cross-platform solution.

staticfloat:
[...]

PR #35625, opened 2020-04-28
Update `Sys.isexecutable()` to use new `jl_fs_access()` API call

staticfloat:
This pull request bumps the `libuv` version to include https://github.com/JuliaLang/libuv/pull/6, which un-breaks `Sys.isexecutable()` on Windows.


JeffBezanson:
fixes #33212?

staticfloat:
Yes.  Finally.  :)

To test, you can use [this function](https://github.com/JuliaLang/Pkg.jl/pull/1573/files#diff-a1e7cfbd723206272a351c94f21308deR717-R774).  On current Julia, reports all filesystems on Windows to be permissions-blind, with this change, most filesystems on windows are not reported as permissions-blind. (FAT32 may still be)

staticfloat:
Alright the windows tests are failing because there's an assumption in the test suite that `chmod(file, mode)` and `stat(file) == mode` will work for the writable bit, and that not true on Windows right now because the way `stat()` works is different from how I have changed `access()` and `chmod()`.

Is it worthwhile for me to go and fix `stat()` as well, or should we just disable this test for now?  It doesn't seem critical to me right now, and it doesn't impact git tree hashing (Which is what motivates me).

base/sysinfo.jl
```diff
@@ -438,16 +438,18 @@ const WINDOWS_VISTA_VER = v"6.0"
     Sys.isexecutable(path::String)
 
 Return `true` if the given `path` has executable permissions.
+
+!!! note
+    Prior to Julia 1.4, this did not correctly interrogate filesystem
+    ACLs on Windows, therefore it would return `true` for any extant
```
StefanKarpinski, line 444:
```suggestion
    ACLs on Windows, therefore it would return `true` for any
```

staticfloat:
Turns out the error is because we aren't setting the (independent from ACLs) readonly bit anymore in this PR.  I've got a new PR to libuv here that fixes this: https://github.com/JuliaLang/libuv/pull/8

base/sysinfo.jl
```diff
@@ -438,16 +438,18 @@ const WINDOWS_VISTA_VER = v"6.0"
     Sys.isexecutable(path::String)
 
 Return `true` if the given `path` has executable permissions.
+
+!!! note
+    Prior to Julia 1.4, this did not correctly interrogate filesystem
```
StefanKarpinski, line 443:
```suggestion
    Prior to Julia 1.5, this did not correctly interrogate filesystem
```

base/sysinfo.jl
```diff
@@ -438,16 +438,18 @@ const WINDOWS_VISTA_VER = v"6.0"
...
+    ACLs on Windows, therefore it would return `true` for any
+    file.  From Julia 1.4 on, it correctly determines whether the
```
StefanKarpinski, line 445:
```suggestion
    file.  From Julia 1.5 on, it correctly determines whether the
```

staticfloat: I think we will merge this today, but your edit still stands.  :)

staticfloat: The only thing holding this up is that it's breaking the windows buildbots because the buildbots get read-only files created, and then a later process doesn't know how to delete them.  :(

staticfloat:
Thanks to Jameson's proddings, I think I've actually found a combination of flags that gives us the expected behavior of `chmod()` on all platforms.  I've added some new tests, let's see if they pass on all platforms. If they do, I'm totally happy with this PR and we can merge it.

PR #35626, opened 2020-04-28
Add Makefile caching helpers for lazy, expensive evaluations

staticfloat:
Makefiles natively offer two variable expansion modes: immediate and
deferred.  When expanding variables that require invocations of external
programs (such as `llvm-config`) immediate expansion is almost always
preferred, as it will run the external command once, exactly when the
makefile variable is defined.  Deferred mode, on the other hand, will
expand the variable every time it is used, running the external program
again and again.

When the external program is expensive, this cost can slow down the
build significantly, however when the external program requires some
setup (for instance, when it itself is downloaded through other rules in
the Makefile) it cannot always be immediately expanded.  To address
this, we build a caching layer that allows for deferred expansion, but
once it has been expanded once, the variable is replaced with the result
of running the command, and further hits to the same variable will
return the cached value.  (With the slight caveat that an empty result
will cause the external command to be run again in the future).

As an example usecase, this commit converts our relative path
calculation to use a python script and showcases how to cache this
operation.  This will be used for further JLL stdlib work where the
invocation is much more expensive.


Make.inc
```diff
@@ -277,16 +294,32 @@ private_libdir := $(libdir)/julia
...
+define hit_cache
+$(if $(_CACHE-$(1)),,$(eval _CACHE-$(1) := $($(1))))$(_CACHE-$(1))
+endef
+
 # Calculate relative paths to libdir, private_libdir, datarootdir, and sysconfdir
-build_libdir_rel := $(shell $(JULIAHOME)/contrib/relative_path.sh $(build_bindir) $(build_libdir))
-libdir_rel := $(shell $(JULIAHOME)/contrib/relative_path.sh $(bindir) $(libdir))
-build_private_libdir_rel := $(shell $(JULIAHOME)/contrib/relative_path.sh $(build_bindir) $(build_private_libdir))
-private_libdir_rel := $(shell $(JULIAHOME)/contrib/relative_path.sh $(bindir) $(private_libdir))
-datarootdir_rel := $(shell $(JULIAHOME)/contrib/relative_path.sh $(bindir) $(datarootdir))
```
KristofferC, line 285: This is lame imo. You can't grep for where this variable is defined anymore.

PR #35627, opened 2020-04-28
Teach `jl_load_dynamic_library()` about `@executable_path` on all platforms

staticfloat:
It is useful for us to be able to hard-code a dynamic library's location
relative to that of the julia executable.  We repurpose the
`@executable_path` component from MacOS, allowing all platforms to
perform executable-relative binary loading (I'm looking at you, Windows).

This is useful for the Julia binary to be able to find libraries within artifacts in https://github.com/JuliaLang/julia/pull/35193

### DIFF
--- a/src/dlload.c
+++ b/src/dlload.c
@@ -59,8 +59,6 @@ static int endswith_extension(const char *path)
 
 #define PATHBUF 4096
 
-extern char *julia_bindir;
-
 #define JL_RTLD(flags, FLAG) (flags & JL_RTLD_ ## FLAG ? RTLD_ ## FLAG : 0)
 
 #ifdef _OS_WINDOWS_
@@ -136,7 +134,7 @@ JL_DLLEXPORT int jl_dlclose(void *handle)
 
 JL_DLLEXPORT void *jl_load_dynamic_library(const char *modname, unsigned flags, int throw_err)
 {
-    char path[PATHBUF];
+    char path[PATHBUF], relocated[PATHBUF];
     int i;
 #ifdef _OS_WINDOWS_
     int err;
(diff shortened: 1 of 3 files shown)


src/dlload.c
```diff
@@ -183,13 +184,21 @@ JL_DLLEXPORT void *jl_load_dynamic_library(const char *modname, unsigned flags,
                 size_t len = strlen(dl_path);
                 if (len == 0)
                     continue;
+
+                // Is this entry supposed to be relative to the bindir?
+                if (strncmp(dl_path, "@executable_path", 16 < len ? 16 : len) == 0) {
```
vtjnash, line 189:
```suggestion
                if (len > 17 && strcmp(dl_path, "@executable_path/") == 0) {
```

src/dlload.c
```diff
@@ -183,13 +184,21 @@ JL_DLLEXPORT void *jl_load_dynamic_library(const char *modname, unsigned flags,
                 size_t len = strlen(dl_path);
                 if (len == 0)
                     continue;
+
+                // Is this entry supposed to be relative to the bindir?
+                if (len > 17 && strcmp(dl_path, "@executable_path/") == 0) {
```
StefanKarpinski, line 189: Is this valid if `dl_path` is just `@executable_path`?

src/dlload.c
staticfloat: Wait, I misread this.  This changes it to only match exactly `"@executable_path/"`; I want it to match all strings beginning with `"@executable_path/"`.

src/dlload.c
staticfloat: Ah, thanks, I had mis-read Jameson's change above.  I've reverted it now.

src/dlload.c
vchuravy: You probably still want the `len > 16`?

staticfloat: Sure, let's save a couple of cycles.  Even without it, it should still function correctly, but it will be faster than string comparing those first couple of characters. :)

vtjnash (approved):

staticfloat:
I added a `NEWS.md` entry; I'm not sure where to put the docs for this, we don't seem to have any real docs for `jl_load_dynamic_library()`.

vchuravy:
> we don't seem to have any real docs for jl_load_dynamic_library().

`Libdl.dlopen`?

JeffBezanson:
Needed in 1.5?

staticfloat:
Not needed; will merge after 1.5 branches.

PR #35637, opened 2020-04-29
Help mode documentation for ternary operator

### LINKED ISSUE #35636: docstring for the ternary operator
I wonder if it would be possible to have a docstring for the ternary operator (`?`). Currently help mode for `?` gives the [general help intro](https://github.com/JuliaLang/julia/blob/51484175dee4b0566b270fd62356b1198afdba22/base/docs/basedocs.jl#L27). Would it make sense to special-case `?` and shoehorn eg the following in there:

> `a ? b : c` 
>
> Short form for conditionals, also known as the ternary operator. Equivalent to
> ```julia
> if a
>     b
> else
>     c
> end

c42f:
Fixes #35636

Here I've stolen the single `?` as an alias for displaying the introduction text, as the intro text already has multiple other ways to show it (including simply pressing Return while in help mode).

I've also tried out something unusual which is linking to the manual documentation which describes this syntax in more detail. (Obviously we should have a better way to create such links without direct URLs, but for now I think this is the only way to do it?)

### DIFF
--- a/base/docs/basedocs.jl
+++ b/base/docs/basedocs.jl
@@ -24,7 +24,7 @@ For help on a specific function or macro, type `?` followed
 by its name, e.g. `?cos`, or `?@time`, and press enter.
 Type `;` to enter shell mode, `]` to enter package mode.
 """
-kw"help", kw"?", kw"Julia", kw"julia", kw""
+kw"help", kw"Julia", kw"julia", kw""
 
 """
     using
@@ -643,6 +643,28 @@ evaluated. The `elseif` and `else` blocks are optional, and as many `elseif` blo
 desired can be used.
 """
 kw"if", kw"elseif", kw"else"
+
+"""
+    a ? b : c
+
+Short form for conditionals; read "if `a`, evaluate `b` otherwise evaluate `c`".
+Also known as the [ternary operator](https://en.wikipedia.org/wiki/%3F:).
+
+This syntax is equivalent to `if a; b else c end`, but is often used to
+emphasize the value `b`-or-`c` which is being used as part of a larger
+expression, rather than the side effects that evaluating `b` or `c` may have.
+
+See the manual section on [control flow](@ref man-conditional-evaluation) for more details.
+
+# Examples
+```
+julia> x = 1; y = 2;
+
+julia> println(x > y ? "x is larger" : "y is larger")
+y is larger
+```
+"""
+kw"?", kw"?:"
 
 """
     for
--- a/doc/src/base/base.md
+++ b/doc/src/base/base.md
@@ -85,6 +85,7 @@ where
 ...
 ;
 =
+?:
 ```
 
 ## Standard Modules
--- a/stdlib/REPL/src/docview.jl
+++ b/stdlib/REPL/src/docview.jl
@@ -23,7 +23,8 @@ const extended_help_on = Ref{Any}(nothing)
 
 function _helpmode(io::IO, line::AbstractString)
     line = strip(line)
-    if startswith(line, '?')
+    ternary_operator_help = (line == "?" || line == "?:")
+    if startswith(line, '?') && !ternary_operator_help
         line = line[2:end]
         extended_help_on[] = line
         brief = false


base/docs/basedocs.jl
```diff
@@ -644,6 +644,31 @@ desired can be used.
 """
 kw"if", kw"elseif", kw"else"
 
+"""
+    a ? b : c
+
+Short form for conditionals; read "if `a`, evaluate `b` otherwise evaluate `c`".
+Also known as the
```
fredrikekre, line 651:
```suggestion
Also known as the [ternary operator](https://en.wikipedia.org/wiki/%3F:).
```

base/docs/basedocs.jl
```diff
@@ -644,6 +644,31 @@ desired can be used.
...
+[ternary operator](https://docs.julialang.org/en/v1/manual/control-flow/#man-conditional-evaluation-1).
```
fredrikekre, line 652:
```suggestion
```

base/docs/basedocs.jl
```diff
@@ -644,6 +644,31 @@ desired can be used.
...
+
+This syntax is equivalent to `if a; b else c end`, but is often used to
+emphasize the value `b`-or-`c` which is being used as part of a larger
+expression, rather than the side effects that evaluating `b` or `c` may have.
+
```
fredrikekre, line 657:
```suggestion

See the manual section on [control flow](@ref man-conditional-evaluation) for more details.

```

fredrikekre (approved):

base/docs/basedocs.jl
c42f: The Julia docs do have a julia-specific discussion of the ternary so I think it's better to link to that. (Only problem is that it's not in its own section, so cannot be directly linked to.)

base/docs/basedocs.jl
```diff
@@ -644,6 +644,31 @@ desired can be used.
...
+1
+
+julia> y = 2
+2
+
+julia> println(x > y ? x : y)
+2
+```
+"""
+kw"?"
```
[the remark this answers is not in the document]
c42f: Actually this won't work without special cases in the REPL help mode parser. We could special-case `?:` though, it does make sense.

base/docs/basedocs.jl
c42f: Ah I see you've linked it further down. Seems reasonable, let me check whether it works.

base/docs/basedocs.jl
c42f: By the way, multi-line suggestions now work in the github code review! (Just click and drag down over the lines you want to make into a suggestion.)

base/docs/basedocs.jl
c42f: Great idea. Unfortunately this doesn't work for the interactive help mode, which is why I didn't do it this way :-/

c42f: We'd need to add `@ref` handling to interactive help processing code to have it redirect to the manual.  I think this would be great by the way!  But I didn't want to get too distracted.

base/docs/basedocs.jl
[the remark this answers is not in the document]
c42f: Done.

base/docs/basedocs.jl
c42f: Doesn't work :-(

c42f:
Thanks @fredrikekre for the review, I think I've figured out how to incorporate those suggestions without breaking things. Any further thoughts?

base/docs/basedocs.jl
```diff
@@ -644,6 +644,32 @@ desired can be used.
...
+
+# Examples
+```
+julia> x = 1
+1
+
+julia> y = 2
+2
+
+julia> println(x > y ? "x is larger" : "y is larger")
```
fredrikekre, line 667:
```suggestion
julia> x = 1; y = 2;

julia> println(x > y ? "x is larger" : "y is larger")
```

base/docs/basedocs.jl
fredrikekre: Yes, but that URL will break soon. My suggestion is the way we handle it everywhere else:
```
$ grep -r --include \*.jl "@ref" . | wc -l 
3150
$ grep -r --include \*.jl "docs.julialang.org" . | wc -l 
8
```
The 8 occurences of that URL is in `doc/make.jl` and in the code for the banner, so there are actually no URLs in docstrings.

fredrikekre:
Would be good to include it in the manual too, maybe in the same section as `if` in https://github.com/JuliaLang/julia/blob/master/doc/src/base/base.md

base/docs/basedocs.jl
c42f: Ok let's do it that way. Having the `@ref` work in help mode will have to wait for another time.

c42f:
> Would be good to include it in the manual too

Done + rebased+squashed.

ViralBShah:
@c42f Please merge if ready to go.

PR #35641, opened 2020-04-29
make at-sync thread safe

### LINKED ISSUE #34666: `@sync (@spawn @spawn f(); @spawn @spawn g())` is not thread-safe?

JeffBezanson:
fixes #34666

### DIFF
--- a/base/experimental.jl
+++ b/base/experimental.jl
@@ -52,30 +52,34 @@ macro aliasscope(body)
 end
 
 
-function sync_end(refs)
-    local c_ex
-    defined = false
-    t = current_task()
-    cond = Threads.Condition()
-    lock(cond)
-    nremaining = length(refs)
-    for r in refs
-        schedule(Task(()->begin
-            try
-                wait(r)
-                lock(cond)
-                nremaining -= 1
-                nremaining == 0 && notify(cond)
-                unlock(cond)
-            catch e
-                lock(cond)
-                notify(cond, e; error=true)
-                unlock(cond)
-            end
-        end))
+function sync_end(c::Channel{Any})
+    if !isready(c)
+        # there must be at least one item to begin with
+        close(c)
+        return
     end
-    wait(cond)
-    unlock(cond)
+    nremaining::Int = 0
+    while true
+        event = take!(c)
+        if event === :__completion__
+            nremaining -= 1
+            if nremaining == 0
+                break
+            end
+        else
+            nremaining += 1
+            schedule(Task(()->begin
+                try
+                    wait(event)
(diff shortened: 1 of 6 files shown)


base/experimental.jl
```diff
@@ -92,9 +99,9 @@ during error handling.
 macro sync(block)
     var = esc(sync_varname)
     quote
-        let $var = Any[]
+        let $var = (Any[], Threads.SpinLock())
```
tkf, line 102: I suppose it's OK to use spinlock here because there is no contention at all in the usual use-case (i.e., no `@spawn`-in-`@spawn` pattern) so that `lock`/`unlock` have virtually no cost?

vtjnash (approved):

base/experimental.jl
```diff
@@ -52,30 +52,37 @@ macro aliasscope(body)
...
-                nremaining -= 1
-                nremaining == 0 && notify(cond)
-                unlock(cond)
-            catch e
-                lock(cond)
-                notify(cond, e; error=true)
-                unlock(cond)
-            end
-        end))
+    while true
```
vtjnash, line 57: This `while true` is conceptually invalid for this macro definition. This appears to possibly require changing `refs` into a proper Channel, rather than reimplementing it here with locks.

JeffBezanson: What is the problem exactly?

vtjnash (approved):

Issue #34267, opened 2020-01-05
Feature request: prevent spawning to thread 1


tro3:
Currently the scheduler through the `Threads.@spawn` macro will assign to random thread #'s, including the primary thread.  There may be times where there is a "traffic cop" activity on Thread 1 that can be impacted if a heavy-computational (in essence, blocking) task is put onto that thread.  It would be nice if there was a way to cordon off Thread 1 from spawning in those cases.

tro3:
Alternately, of course, the equivalent of `@spawnat` would allow the same, but I don’t know (doubt, actually) how feasible that is.

JeffBezanson:
Mostly a duplicate of #16350. We could add a thread reserved for latency-bound tasks.

tro3:
Agreed that this issue and #34240 are the same.  #16350 is highly-related but a little different, in that a single thread was being used by both a mostly-blocking task and the REPL.  (If the `yield` were not there, I suspect it would have locked completely.)  So that one just about whether the REPL gets its own thread. 

I can close this one out if you want to just track the other - your guys' call.

tknopp:
Also discussed here (for crossref): https://discourse.julialang.org/t/threading-1-3-success-story/27111/10

quinnj:
I'll just comment that I've had a desire for something like this as well; particularly in the context of running a microservice/kernel, you want to keep a "heartbeat" thread always ready to respond/beat, while being able to spawn threaded-tasks elsewhere w/o worrying about completely saturating resources to the point that the heartbeat can't respond.

What's the path forward here? I saw that @JeffBezanson suggested potentially having a `@spawn background ex` macro form that would only spawn tasks on non-main threads. Would that require some functionality on the C-side? Do @JeffBezanson / @vtjnash think it'd be worth me trying to take a stab at this, or would it be pretty hairy? I'm eager to see this, so happy to jump in and help any way I can.

tro3:
Is this as simple as...
```julia
task.jl:542

       tid = 0
        if ccall(:jl_enqueue_task, Cint, (Any,), t) != 0
            # if multiq is full, give to a random thread (TODO fix)
+           if t.background  # Perhaps nontrivial like this due to Core.Task mod, but you get the idea
+               tid = mod(time_ns() % Int, Threads.nthreads()-1) + 2
+           else
                tid = mod(time_ns() % Int, Threads.nthreads()) + 1
+           end
[...]

PR #35646, opened 2020-04-29
add `@threads static` to aid thread API evolution

JeffBezanson:
Looking into stabilizing the thread API in 1.5, I think one major likely future change is the schedule used by `@threads`. Eventually we want it to spawn un-pinned tasks (and possibly more tasks than threads), yet we know some code is already using it in a way that assumes the current schedule. I think a good approach is to allow `@threads static` now, so that code depending on the schedule can be future-proofed, and then in a later version we can change the default schedule.

### DIFF
--- a/base/threadingconstructs.jl
+++ b/base/threadingconstructs.jl
@@ -38,7 +38,7 @@ function threading_run(func)
     end
 end
 
-function _threadsfor(iter,lbody)
+function _threadsfor(iter, lbody, schedule)
     lidx = iter.args[1]         # index
     range = iter.args[2]
     quote
@@ -82,9 +82,13 @@ function _threadsfor(iter,lbody)
             end
         end
         end
-        if threadid() != 1
-            # only thread 1 can enter/exit _threadedregion
-            Base.invokelatest(threadsfor_fun, true)
+        if threadid() != 1 || ccall(:jl_in_threaded_region, Cint, ()) != 0
+            $(if schedule === :static
+              :(error("`@threads :static` can only be used from thread 1 and not nested"))
+              else
+              # only use threads when called from thread 1, outside @threads
+              :(Base.invokelatest(threadsfor_fun, true))
+              end)
         else
             threading_run(threadsfor_fun)
         end
@@ -93,31 +97,50 @@ function _threadsfor(iter,lbody)
 end
 
 """
-    Threads.@threads
+    Threads.@threads [schedule] for ... end
 
-A macro to parallelize a for-loop to run with multiple threads. This spawns [`nthreads()`](@ref)
-number of threads, splits the iteration space amongst them, and iterates in parallel.
-A barrier is placed at the end of the loop which waits for all the threads to finish
-execution, and the loop returns.
+A macro to parallelize a `for` loop to run with multiple threads. Splits the iteration
+space among multiple tasks and runs those tasks on threads according to a scheduling
+policy.
+A barrier is placed at the end of the loop which waits for all tasks to finish
+execution.
+
+The `schedule` argument can be used to request a particular scheduling policy.
+The only currently supported value is `:static`, which creates one task per thread
+and divides the iterations equally among them. Specifying `:static` is an error
+if used from inside another `@threads` loop or from a thread other than 1.
+
+The default schedule (used when no `schedule` argument is present) is subject to change.
+
+!!! compat "Julia 1.5"
+    The `schedule` argument is available as of Julia 1.5.
 """
 macro threads(args...)
     na = length(args)
-    if na != 1
+    if na == 2
+        sched, ex = args
+        if sched isa QuoteNode
+            sched = sched.value
+        elseif sched isa Symbol
+            # for now only allow quoted symbols
+            sched = nothing
+        end
+        if sched !== :static
+            throw(ArgumentError("unsupported schedule argument in @threads"))
+        end
+    elseif na == 1
+        sched = :default
+        ex = args[1]
+    else
         throw(ArgumentError("wrong number of arguments in @threads"))
     end
-    ex = args[1]
-    if !isa(ex, Expr)
-        throw(ArgumentError("need an expression argument to @threads"))
+    if !(isa(ex, Expr) && ex.head === :for)
+        throw(ArgumentError("@threads requires a `for` loop expression"))
     end
-    if ex.head === :for
-        if ex.args[1] isa Expr && ex.args[1].head === :(=)
-            return _threadsfor(ex.args[1], ex.args[2])
-        else
-            throw(ArgumentError("nested outer loops are not currently supported by @threads"))
-        end
-    else
-        throw(ArgumentError("unrecognized argument to @threads"))
+    if !(ex.args[1] isa Expr && ex.args[1].head === :(=))
+        throw(ArgumentError("nested outer loops are not currently supported by @threads"))
     end
+    return _threadsfor(ex.args[1], ex.args[2], sched)
 end
 
 """
--- a/test/threads_exec.jl
+++ b/test/threads_exec.jl
@@ -729,6 +729,17 @@ let a = zeros(nthreads())
     @test a == [1:nthreads();]
 end
 
+# static schedule
+function _atthreads_static_schedule()
+    ids = zeros(Int, nthreads())
+    Threads.@threads :static for i = 1:nthreads()
+        ids[i] = Threads.threadid()
+    end
+    return ids
+end
+@test _atthreads_static_schedule() == [1:nthreads();]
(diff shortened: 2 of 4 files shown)


tkf:
Why not sunset `@threads` and instead implement `Threads.foreach` for array-like inputs?  I think it's nice to be transparent that a threaded loop is going to create a closure.

base/threadingconstructs.jl
```diff
@@ -93,31 +93,42 @@ function _threadsfor(iter,lbody)
...
-execution, and the loop returns.
+A macro to parallelize a `for` loop to run with multiple threads. Splits the iteration
+space among multiple tasks and runs those tasks on threads according to a scheduling
+policy.
+A barrier is placed at the end of the loop which waits for all tasks to finish
+execution.
+
+The `schedule` argument can be used to request a particular scheduling policy.
+The only currently supported value is `static`, which creates one task per thread
+and divides the iterations equally among them.
```
tkf, line 106: This is not actually true, right? It only holds when `Threads.@threads` is used in the main thread. This still works (before and after #35632):

```julia
julia> Threads.@threads for i in 1:Threads.nthreads()
           @assert i == Threads.threadid()
           Threads.@threads for j in 1:Threads.nthreads()
               if i != 1
                   @assert i == Threads.threadid()  # inherits parent tid
               end
           end
       end
```

How about something like

```suggestion
The only currently supported value is `static`, which creates tasks as many as available
threads and divides the iterations equally among them.
```

?

As for "one task per thread" API, I think what people actually need is [`Threads.pin`](https://github.com/JuliaLang/julia/issues/34267#issuecomment-573507670).

JeffBezanson:
Question for triage: as tkf said, the static schedule is not exactly what's documented. The current behavior is that we spawn 1 task per thread when `@threads` is called from thread 1, otherwise all iterations run on the current thread. Should we add the behavior that *nested* calls to `@threads` on thread 1 also disable parallelism?

JeffBezanson:
From triage: disable parallelism when nested.

tkf:
How about

```julia
julia> const STOP = Threads.Atomic{Bool}(false)
       Threads.@spawn while !STOP[] end
       Threads.@threads for i in 1:Threads.nthreads()
           @show i
       end
       STOP[] = true
i = 3
i = 1
i = 4
^C
```

? It hangs currently.

JeffBezanson:
That's an example of why I think `@threads` should use a more dynamic schedule by default, but with the static schedule I don't see what else it can do. What should it do?

tkf:
Why not just

```julia
basesize = cld(length(xs), Threads.nthreads())
@sync for p in Iterators.partition(xs, basesize)
    @spawn foreach(loop_body, p)
end
```

?  It's "static" in the sense that the sub-chunk of the array to be processed by each task is determined by `nthreads()` and `size(xs)`.

IIUC, any API that would specify the assignment to thread would be impossible to compose well in task-based parallelism.  Some kind of API for this (e.g., `Threads.pin`) may be useful for power users but I don't think a high-level API like `@threads` should have this.  Maybe at least use an ugly name like `unsafe_static` or `legacy_static`?

JeffBezanson:
> IIUC, any API that would specify the assignment to thread would be impossible to compose well in task-based parallelism.

Yes, that is completely true. Some existing code assumes it though. And I believe the pinned thread assignments are part of what's meant by a "static" schedule in standard terminology. I guess I'm ok with giving it an uglier name though.

tkf:
To support existing code bases, wouldn't it be even nicer to just throw an error when `Threads.threadid() != 1` (or whenever the static assignment is impossible)?  If they assume static scheduling, I suppose it's safer to have an exception rather than silently failing?  We can "guarantee" static assignment this way (as otherwise the program either deadlocks or throws).  It's a bit harsh API but I don't think it's a huge problem if we introduce something like `Iterators.partition`-based scheduler at the same time make it default.  It's not great that we can't provide a grace period for those who are assuming static scheduling.  But I don't think it's possible to not break their code at some point anyway.

StefanKarpinski:
Why do so much work done to have a composable threading system only to refuse to allow threads to compose?!? If we wanted nested `@threads` calls to fail, we could have had that five years ago.

tkf:
I'm proposing _default_ `@threads for` to use _non_-static scheduling so that it composes well when nested in `@threads` or `@spawn`.  The extra work is for allowing existing code to still use static scheduling when they happened to assume it.  Code assuming static scheduling won't compose well anyway so I don't know what else can be done to rescue them.

StefanKarpinski:
> Code assuming static scheduling won't compose well anyway so I don't know what else can be done to rescue them.

Just run it sequentially? It's pretty hard to write code using `@threads` that wouldn't also be correct, if slower, if you just ran it sequentially on one thread. You can't always control how your code is called or even what your code calls, if it's generic. Just failing seems like the worst option.

JeffBezanson:
We also need to remove `_threadedregion` (#32477) for the full story to work anyway.

tkf:
> Just run it sequentially?

I guess it depends on how strong their assumption is and how much you want to allow it?  Falling back to sequential execution does not help a case like

```julia
const VECTORS = Vector{Vector{Float64}}[]
__init__() = Threads.resize_nthreads!(empty!(VECTORS), Float64[])

function f(outputs, inputs, callback)
    @threads for (y, x) in zip(outputs, inputs)
        tmp = VECTORS[Threads.nthreads()]
        compute_1!(tmp, x)
        callback()
        compute_2!(y, x, tmp)
    end
end
```

when `f` itself is called via `callback`.  I think failing is much better in this case.

JeffBezanson:
I guess we could give an error for nested `@threads` *if* you specify `static`? Maybe that's what you meant and I missed it. Because in general we can assume that a user of `@threads` doesn't care what schedule is used.

tkf:
Ah, yes, that's what I meant. Sorry, it wasn't clear.

test/threads_exec.jl
```diff
@@ -729,6 +729,16 @@ let a = zeros(nthreads())
     @test a == [1:nthreads();]
 end
 
+# static schedule
+function _atthreads_static_schedule()
+    ids = zeros(Int, nthreads())
+    Threads.@threads static for i = 1:nthreads()
```
jonas-schulze, line 735: This might be a stupid question: of which type is `static` in here? It’s not a symbol and looks like a keyword or a variable instead. As the tests are passing, I assume it’s not a typo.

I would like to propose that it should be a symbol instead, so that the scheduling may be controlled by a variable or similar (cf. `@spawnat :any ...` and `@spawnat 2 ...` from `Distributed`).

tkf: It's used only at macro-expansion time so it doesn't correspond to any variables. We already have something like `@simd ivdep for ... end` so I think it's OK to have this interface.

But I think using `:static` actually makes sense.  It's handy to decide scheduler at run-time like

```julia
function f(...; scheduler=:static)
    @threads scheduler for ... end
end
```

JeffBezanson: > It's handy to decide scheduler at run-time

This is actually a really fundamental decision though. Is the scheduler an object from an expression evaluated at run time, or is it known at macro expansion time, potentially allowing the macro to emit different code for different schedules?

tkf: Do you think most (all?) of the schedulers need to do something during macro expansion?  Perhaps that's the reason why you are not in favor of `Threads.foreach` approach?

I was actually thinking that a (large) subset of schedulers can be set at run-time. We can throw a run-time error when a scheduler that has to be expanded by the macro is specified at runtime.

JeffBezanson: Probably true, most schedules could be picked at run-time. I guess it's a good compromise to in general evaluate the expression, but recognize certain constant expressions like `:static` at macro expand time. I'll change it to that.

base/threadingconstructs.jl
```diff
@@ -93,31 +97,50 @@ function _threadsfor(iter,lbody)
...
+
+The `schedule` argument can be used to request a particular scheduling policy.
+The only currently supported value is `:static`, which creates one task per thread
+and divides the iterations equally among them. Specifying `:static` is an error
+if used from inside another `@threads` loop or from a thread other than 1.
+
+The default schedule (used when no `schedule` argument is present) is subject to change.
+
+!!! compat "Julia 1.5"
+    The `schedule` argument is available as of Julia 1.5.
```
vtjnash, line 116:
```suggestion
    The `schedule` argument is available as of Julia 1.5.
    Nesting `@threads` calls now throws an exception.
```

JeffBezanson: Only if you specify `:static`; I fixed that in the doc string.

PR #35648, opened 2020-04-29
turn deprecations into hard errors

StefanKarpinski:
Kind of crucial part of https://github.com/JuliaLang/julia/pull/35362. Previously deprecated features in Base are now errors. If we had released a version without doing this, then we wouldn't have been allowed to change these behaviors in following versions.

### DIFF
--- a/base/abstractdict.jl
+++ b/base/abstractdict.jl
@@ -424,22 +424,9 @@ Dict{Int64,String} with 1 entry:
 function filter(f, d::AbstractDict)
     # don't just do filter!(f, copy(d)): avoid making a whole copy of d
     df = empty(d)
-    try
-        for pair in d
-            if f(pair)
-                df[pair.first] = pair.second
-            end
-        end
-    catch e
-        if isa(e, MethodError) && e.f === f
-            depwarn("In `filter(f, dict)`, `f` is now passed a single pair instead of two arguments.", :filter)
-            for (k, v) in d
-                if f(k, v)
-                    df[k] = v
-                end
-            end
-        else
-            rethrow()
+    for pair in d
+        if f(pair)
+            df[pair.first] = pair.second
         end
     end
     return df
--- a/base/reflection.jl
+++ b/base/reflection.jl
@@ -1064,17 +1064,6 @@ function func_for_method_checked(m::Method, @nospecialize(types), sparams::Simpl
     return m
 end
 
-function func_for_method_checked(m::Method, @nospecialize(types))
-    depwarn("The two argument form of `func_for_method_checked` is deprecated. Pass sparams in addition.",
-            :func_for_method_checked)
-    if isdefined(m, :generator) && !isdispatchtuple(types)
-        error("cannot call @generated function `", m, "` ",
-              "with abstract argument types: ", types)
-    end
-    return m
-end
-
-
 """
     code_typed(f, types; optimize=true, debuginfo=:default)
 
--- a/base/shell.jl
+++ b/base/shell.jl
@@ -15,13 +15,6 @@ function rstrip_shell(s::AbstractString)
     SubString(s, 1, 0)
 end
 
-
-# needs to be factored out so depwarn only warns once
-# when removed, also need to update shell_escape for a Cmd to pass shell_special
-# and may want to use it in the test for #10120 (currently the implementation is essentially copied there)
-@noinline warn_shell_special(str,special) =
-    depwarn("Parsing command \"$str\". Special characters \"$special\" should now be quoted in commands", :warn_shell_special)
-
 function shell_parse(str::AbstractString, interpolate::Bool=true;
                      special::AbstractString="")
     s::SubString = SubString(str, firstindex(str))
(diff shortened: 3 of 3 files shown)


KristofferC:
> Kind of crucial part of #35362.

I don't really get this. How is this related to #35362. Why does this have to go in 1.5?

mbauman:
Version 1.5 will stop warning for these guys (which I think were actually deprecated in 0.7).  That'll make it seem like they're now fine behaviors — even though we had long ago agreed to deprecate them.

KristofferC:
> That'll make it seem like they're now fine behaviors. 

Yeah, that will be the case for all deprecations. So you need to run with deprecations warnings on before you update to a breaking release. So I don't see how the linked PR has any relevance. It was just a policy choice of what to show by default. 

Again, I don't see how the linked PR influences this and why it is on the milestone.

StefanKarpinski:
Julia 1.4:
```jl
julia> filter((k, v) -> isodd(v), Dict(:foo => 1, :bar => 2))
┌ Warning: In `filter(f, dict)`, `f` is now passed a single pair instead of two arguments.
│   caller = top-level scope at REPL[5]:1
└ @ Core REPL[5]:1
Dict{Symbol,Int64} with 1 entry:
  :foo => 1

julia> cmd = `echo "wtf" > test1.log`
┌ Warning: Parsing command "echo "wtf" > test1.log". Special characters "#{}()[]<>|&*?~;" should now be quoted in commands
│   caller = shell_parse(::String, ::Bool; special::String) at shell.jl:107
└ @ Base ./shell.jl:107
`echo wtf '>' test1.log`
```
Julia master since 
```jl
julia> filter((k, v) -> isodd(v), Dict(:foo => 1, :bar => 2))
Dict{Symbol,Int64} with 1 entry:
  :foo => 1

julia> cmd = `echo "wtf" > test1.log`
`echo wtf '>' test1.log`
```
Since those both work with no warning on master, if we released 1.5 as is, we would be required to keep these "features" for the rest of the 1.x series. That is NOT what we want: we want to make the filter an error and we want to make the backtick expresison construct a pipeline that writes to a file. Releasing 1.5 would make both of those changes semver violations.

This PR makes them hard errors:
```jl
julia> filter((k, v) -> isodd(v), Dict(:foo => 1, :bar => 2))
ERROR: MethodError: no method matching (::var"#1#2")(::Pair{Symbol,Int64})
Closest candidates are:
  #1(::Any, ::Any) at REPL[1]:1
Stacktrace:
 [1] filter(::var"#1#2", ::Dict{Symbol,Int64}) at ./abstractdict.jl:428
 [2] top-level scope at REPL[1]:1

julia> cmd = `echo "wtf" > test1.log`
ERROR: LoadError: parsing command `echo "wtf" > test1.log`: special characters "#{}()[]<>|&*?~;" must be quoted in commands
Stacktrace:
 [2] shell_parse(::String, ::Bool; special::String) at ./shell.jl:100
 [3] @cmd(::LineNumberNode, ::Module, ::Any) at ./cmd.jl:389
in expression starting at REPL[2]:1
```

StefanKarpinski:
We can break code that worked in a previous version but emitted a warning by default. We cannot break code that worked without any warning by default. If 1.5 went out with these constructs working with no warning, then we'd be stuck with them. That's the problem.

mbauman:
FWIW, I put it on the milestone not because I want to single-handedly force this into Julia 1.5, but rather that _we absolutely must decide_ one way or another before we feature freeze.

StefanKarpinski:
There's also some previously squeaky deprecations in base/deprecated.jl that are now silent. We should either make the squeaky again or make them errors. There are also some notes in there about deprecating things in the future, which we should maybe go ahead and turn into deprecations now that deprecations are silent by default.

tkf:
> keep these "features" for the rest of the 1.x series

I think, _technically_, this is actually what we have to do to make it completely SemVer compliment.

See [SemVer FAQ](https://semver.org/#how-should-i-handle-deprecating-functionality)

> When you deprecate part of your public API, you should do two things: (1) update your documentation to let users know about the change, (2) issue a new minor release with the deprecation in place. **Before you completely remove the functionality in a new major release** there should be at least one minor release that contains the deprecation so that users can smoothly transition to the new API.

I think the bold part (by me) implies that removing a feature should always be done by a major bump.  I think this is true even when the deprecation warnings were enabled in the previous "major" version (= 0.7).  Of course, we always have the escape hatch ("minor change")...

Perhaps the least breaking action is to create a "hard" depwarn that cannot be suppressed by `--depwarn=no`.  It may annoy more people, though.

StefanKarpinski:
These have been deprecated since 0.7.0 so we are complying with semver.

KristofferC:
> We can break code that worked in a previous version but emitted a warning by default. We cannot break code that worked without any warning by default.

This doesn't make sense to me (I don't see how the "default" printing of deprecation warning is relevant). This code has been deprecated since 0.7 and is still deprecated? Yes, you now need to run with `--depwarn=yes` to see it but that is documented.

PR #35650, opened 2020-04-29
revise parallel computing manual chapters

JeffBezanson:
This is still far from perfect, but I did a bit of a clean-up pass to remove obsolete information. I also moved everything about Tasks from the control flow chapter to the new asynchronous programming chapter, and removed redundant material.

Still need to add material on spawn, locks, and a more specific list of caveats.


doc/src/manual/multi-threading.md
```diff
@@ -1,9 +1,6 @@
-# [Multi-Threading (Experimental)](@id man-multithreading)
+# [Multi-Threading](@id man-multithreading)
```
ViralBShah, line 1: Does this mean we are taking off the Experimental label for 1.5? Should `threadcall` still have the Experimental label on it (like it used to)?

ViralBShah:
Looks pretty good. Can we merge this?

doc/src/manual/multi-threading.md
JeffBezanson: It's not SO experimental that it needs to be in the title of the chapter :)

PR #35658, opened 2020-04-30
fix #35649, copy! on vectors with shared data

### LINKED ISSUE #35649: `copy!` fails on vectors with shared data
This works fine:
```julia
a = rand(2,2)
a2 = reshape(a, 4)
copy!(a, rand(2,2))
```
but this does not:
```julia
b = rand(4)
b2 = reshape(b, 2, 2)
copy!(b, rand(4)) # ERROR: cannot resize array with shared data
```
because `copy!` for vectors currently implemented (@rfourquet) like this:
```julia
copy!(dst::AbstractVector, src::AbstractVector) = append!(empty!(dst), src)
```
Maybe we don't need the resizing if their lengths match?
```
julia> versioninfo()
Julia Version 1.4.1
Platform Info:
  CPU: AMD A10-7860K Radeon R7, 12 Compute Cores 4C+8G
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-8.0.1 (ORCJIT, bdver3)
Environment:
  JULIA_PKG_SERVER = pkg.julialang.org
```

rfourquet:
This makes the following work as expected, by not trying to `resize!` when it's not necessary:
```julia
b = rand(4)
b2 = reshape(b, 2, 2)
copy!(b, rand(4)) 
```
But as a side effect, this also fixes another "bug", but this might be considered as a "minor change": after a call `copy!(a, b)`, `a == b` is supposed to hold, but currently this fails for offset arrays:
```julia
julia> a = OffsetArray([1, 2, 3], -2); copy!(a, 1:3); a == 1:3
false
```
This is because no check on axes was made on vectors (only on arrays with more dimensions). In this PR, `eachindex(a, b)` is used, and this `eachindex` call checks for compatible axes (meaning that `copy!(a, 1:3)` above will fail, when it was working before).

### DIFF
--- a/base/abstractarray.jl
+++ b/base/abstractarray.jl
@@ -705,7 +705,15 @@ See also [`copyto!`](@ref).
     This method requires at least Julia 1.1. In Julia 1.0 this method
     is available from the `Future` standard library as `Future.copy!`.
 """
-copy!(dst::AbstractVector, src::AbstractVector) = append!(empty!(dst), src)
+function copy!(dst::AbstractVector, src::AbstractVector)
+    if length(dst) != length(src)
+        resize!(dst, length(src))
+    end
+    for i in eachindex(dst, src)
+        @inbounds dst[i] = src[i]
+    end
+    dst
+end
 
 function copy!(dst::AbstractArray, src::AbstractArray)
     axes(dst) == axes(src) || throw(ArgumentError(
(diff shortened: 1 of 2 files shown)


JeffBezanson:
> meaning that `copy!(a, 1:3)` above will fail, when it was working before

This is probably ok, as long as `copyto!` can still be used in this case.

JeffBezanson (approved):

StefanKarpinski:
Should this be backported as well?

rfourquet:
> Should this be backported as well?

I suppose it could, although this doesn't fix a bug which was producing wrong results, just something which didn't work for no good reason, so code working in 1.4 (or 1.0) won't be better off with this.

StefanKarpinski:
Ok, no backport then.

Issue #35640, opened 2020-04-29
julia 1.4.1 LibGit2 tests fails on NixOS


cstich:
`@test LibGit2.credential_helpers(cfg, mygit_cred) == expected[2:2]` introduced at this commit (https://github.com/JuliaLang/julia/commit/f1593e1db479e90033e2ce96e31cf2333ea85413) does not work properly on NixOS. When I try to compile 1.4.1 and then run the tests, I get the following error:

>Error in testset file:
Test Failed at /tmp/nix-build-julia-1.4.1.drv-0/source/test/file.jl:70
  Expression: dirname(t) == d
   Evaluated: "/tmp/nix-build-julia-1.4.1.drv-0" == "/tmp/nix-build-julia-1.4.1.drv-0/jl_aPXajU"
Error in testset LibGit2/libgit2:
Test Failed at /tmp/nix-build-julia-1.4.1.drv-0/source/usr/share/julia/stdlib/v1.4/LibGit2/test/libgit2.jl:1983
  Expression: LibGit2.credential_helpers(cfg, mygit_cred) == expected[2:2]
   Evaluated: LibGit2.GitCredentialHelper[LibGit2.GitCredentialHelper(`echo first`), LibGit2.GitCredentialHelper(`echo second`)] == LibGit2.GitCredentialHelper[LibGit2.GitCredentialHelper(`echo second`)]
ERROR: LoadError: Test run finished with errors
in expression starting at /tmp/nix-build-julia-1.4.1.drv-0/source/test/runtests.jl:67
make[1]: *** [Makefile:25: all] Error 1
make[1]: Leaving directory '/tmp/nix-build-julia-1.4.1.drv-0/source/test'
make: *** [Makefile:577: testall] Error 2
builder for '/nix/store/10xkbc3ahq406a3s2rdy226ppc69qa4y-julia-1.4.1.drv' failed with exit code 2
error: build of '/nix/store/10xkbc3ahq406a3s2rdy226ppc69qa4y-julia-1.4.1.drv' failed

Any chance anyone knows whether this is a bug in julia or actually a nix issue? Although I am using the provided `LibGit2` library to compile 1.4.1 (at least I think I am).

omus:
I suspect that the libgit2 library isn't respecting ordering for iterators. Can you try running:
```julia
using LibGit2, Test

mktempdir() do dir
    config_path = joinpath(dir, "gitconfig")
    open(config_path, "w+") do fp
        write(fp, """
            [credential]
[...]

PR #35666, opened 2020-04-30
Test ordering of `GitConfigIter`

omus:
Makes the test more specific which was helpful in trying to debug: https://github.com/JuliaLang/julia/issues/35640. Previously this was not done as `GitConfigIter` ordering was unreliable.


omus:
I'll merge this tomorrow as it's pretty uncontroversial

PR #35668, opened 2020-04-30
Fix llvm-propagate-addrspacecasts for targets with DL greater than 0

jpsamaroo:
Required for AMDGPU LLVM target usage in GPUCompiler. I assume this is necessary because AMDGPU's datalayout specifies addrspace 5 as the AS for `alloca`.


jpsamaroo:
Relies on #35645 to be effective.

@vchuravy @maleadt

Keno:
I don't think this should touch anything other than our own address spaces. Introducing extra addrspace casts is a bit dangerous, so I would prefer it it just refused to rewrite the instruction if there's an odd addrspace setup.

jpsamaroo:
Thanks for the review @Keno ! This pass was causing issues with usage of the AMDGPU target because the addrspace for the stack is non-0 (it's AS 5), and the piece of code I modified here was the source of the most immediate error (Invalid Bitcast).

Since I've managed to make things work with a custom pass pipeline (and by excluding this specific pass), I'm going to close this. I may reopen this later if it becomes necessary.

jpsamaroo:
@Keno I'm reopening this PR because I believe that it's important to properly handle non-generic-addrspaced `alloca`s (which is the case for GCN, where the stack space is in addrspace 5 in the datalayout), and this is (in my eyes) a better solution than my having to bypass part of Julia's optimization pipeline. I'm working on putting together a real-world reduced case where this PR is necessary, which will be included as part of this PR.

I think what needs to be done is to query the stack addrspace from the datalayout and use that as the addrspace for the "base" pointer that we lift to. It's my assumption that doing this will eliminate the need for inserting an addrspacecast (which I've learned is invalid according to Julia's semantics). I'm not sure how to get at the `DataLayout` when within a `FunctionPass` in C++, so I'd appreciate any help in that regard.

Keno:
Doing something targeted like selecting the alloca address space is fine. The problematic part is inserting arbitrary addrspace casts just based on what cast is missing, since in our scheme these casts have semantics, so if you insert one without checking that the pre-requisites hold, things can crash.

jpsamaroo:
Superseded by #45900

PR #35674, opened 2020-04-30
adjust deprecations for v1.5

### LINKED ISSUE #35665: now-silent deprecations that should be errors?
Now that deprecations default to not warning, base/deprecated.jl could use some updates:

https://github.com/JuliaLang/julia/blob/2b871d428a2a91124e9e2/base/deprecated.jl#L175-L197

There's some stuff in there that was squeaky that will now be silent and should perhaps be converted into a hard error. There's other deprecations that can be added now that they won't warn by default. I'm opening the issue since I'm not entirely sure what to do about all of these.

cc @mbauman, @timholy, @JeffBezanson, @simonbyrne to whomst the blame is attributed

JeffBezanson:
fixes #35665

### DIFF
--- a/base/deprecated.jl
+++ b/base/deprecated.jl
@@ -164,39 +164,17 @@ macro deprecate_moved(old, new, export_old=true)
         Expr(:call, :deprecate, __module__, Expr(:quote, old), 2))
 end
 
-# BEGIN 0.7 deprecations
-
-function promote_eltype_op end
-
-# END 0.7 deprecations
-
 # BEGIN 1.0 deprecations
 
-# @deprecate one(i::CartesianIndex) oneunit(i)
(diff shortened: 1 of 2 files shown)


base/deprecated.jl
```diff
@@ -166,36 +166,22 @@ end
...
 # END 1.0 deprecations
 
 # BEGIN 1.3 deprecations
 
-@eval Threads begin
-    Base.@deprecate_binding RecursiveSpinLock ReentrantLock
-    Base.@deprecate_binding Mutex ReentrantLock
-end
-
 # END 1.3 deprecations
```
StefanKarpinski, line 185: Seems like you can remove this entire block from `# BEGIN 1.3` to `# END 1.3`.

PR #30973, opened 2019-02-06
Workaround #28126, support SIMDing broadcast in more cases

mbauman:
This is an ugly performance hack around issue #28126 in some limited (but common) cases. The problem in short: when given many arrays of the same size, LLVM has difficulty hoisting the decision of whether a given dimension should be "extruded" out of the loop. This extra indirection in the index computation seems to foil the array bounds aliasing checks, which stymies SIMDification. The solution: check to see if _Julia_ can statically decide whether or not to extrude any dimensions in a given broadcast expression -- and if so, use a special array wrapper that flags that none of the dimensions in that array need to be extruded out in order to perform the broadcast.

No idea if tests will pass or if this actually is a good idea in general.

@nanosoldier `runbenchmarks(ALL, vs = ":master")` (edit: whoops, sorry Nanosoldier — should have waited for CI before sending you down a dead-end alley)

PR #35675, opened 2020-04-30
Remove inline annotations from broadcast kernels

mbauman:
In my cursory spot-tests, it appears as though we no longer need to force inlining the whole way through to the innermost loops of broadcast.  This is huge, and in my naive understanding I think it'll greatly improve codegen times and sizes.  It'll also allow for embedding more alternative loop designs as they no longer need to be inlined into the same function body — solving my reservations in #30973.

I've kept the preparatory `@inline`s for now, just removing them on the actual implementation.

Making this an early PR just to allow a Nanosoldier run.

### DIFF
--- a/base/broadcast.jl
+++ b/base/broadcast.jl
@@ -1309,14 +1309,14 @@ macro __dot__(x)
     esc(__dot__(x))
 end
 
-@inline function broadcasted_kwsyntax(f, args...; kwargs...)
+@inline function broadcasted_kwsyntax(f::F, args...; kwargs...) where {F}
     if isempty(kwargs) # some BroadcastStyles dispatch on `f`, so try to preserve its type
         return broadcasted(f, args...)
     else
         return broadcasted((args...) -> f(args...; kwargs...), args...)
     end
 end
-@inline function broadcasted(f, args...)
+@inline function broadcasted(f::F, args...) where {F}
     args′ = map(broadcastable, args)
     broadcasted(combine_styles(args′...), f, args′...)
 end
@@ -1324,18 +1324,18 @@ end
 # the totally generic varargs broadcasted(f, args...) method above loses Type{T}s in
 # mapping broadcastable across the args. These additional methods with explicit
 # arguments ensure we preserve Type{T}s in the first or second argument position.
-@inline function broadcasted(f, arg1, args...)
+@inline function broadcasted(f::F, arg1::T1, args...) where {F, T1}
     arg1′ = broadcastable(arg1)
     args′ = map(broadcastable, args)
     broadcasted(combine_styles(arg1′, args′...), f, arg1′, args′...)
 end
-@inline function broadcasted(f, arg1, arg2, args...)
+@inline function broadcasted(f::F, arg1::T1, arg2::T2, args...) where {F, T1, T2}
     arg1′ = broadcastable(arg1)
     arg2′ = broadcastable(arg2)
     args′ = map(broadcastable, args)
     broadcasted(combine_styles(arg1′, arg2′, args′...), f, arg1′, arg2′, args′...)
 end
-@inline broadcasted(::S, f, args...) where S<:BroadcastStyle = Broadcasted{S}(f, args)
+@inline broadcasted(::S, f::F, args...) where {S<:BroadcastStyle, F} = Broadcasted{S}(f, args)
 
 """
     BroadcastFunction{F} <: Function
--- a/base/broadcast.jl
+++ b/base/broadcast.jl
@@ -842,8 +842,8 @@ julia> string.(("one","two","three","four"), ": ", 1:4)
 broadcast(f::Tf, As...) where {Tf} = materialize(broadcasted(f, As...))
 
 # special cases defined for performance
-@inline broadcast(f, x::Number...) = f(x...)
-@inline broadcast(f, t::NTuple{N,Any}, ts::Vararg{NTuple{N,Any}}) where {N} = map(f, t, ts...)
+broadcast(f, x::Number...) = f(x...)
+broadcast(f, t::NTuple{N,Any}, ts::Vararg{NTuple{N,Any}}) where {N} = map(f, t, ts...)
 
 """
     broadcast!(f, dest, As...)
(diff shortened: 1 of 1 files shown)


mbauman:

mbauman:
On master:

```
2 $ time make -C test broadcast
    JULIA test/broadcast
Test  (Worker) | Time (s) | GC (s) | GC % | Alloc (MB) | RSS (MB)
broadcast  (1) |        started at 2020-04-30T16:43:19.276
broadcast  (1) |    47.61 |   1.28 |  2.7 |    5773.20 |   478.94

Test Summary: | Pass  Total
  Overall     |  439    439
    SUCCESS

real	0m51.426s
user	0m50.828s
sys	0m0.784s
```

On this PR:

```
$ time make -C test broadcast
    JULIA test/broadcast
Test  (Worker) | Time (s) | GC (s) | GC % | Alloc (MB) | RSS (MB)
broadcast  (1) |        started at 2020-04-30T16:43:02.881
broadcast  (1) |    37.41 |   1.06 |  2.8 |    4296.83 |   386.69

Test Summary: | Pass  Total
  Overall     |  439    439
    SUCCESS

real	0m41.086s
user	0m40.656s
sys	0m0.649s
```

base/broadcast.jl
```diff
@@ -1102,15 +1102,15 @@ struct BitMaskedBitArray{N,M}
     mask::BitArray{M}
     BitMaskedBitArray{N,M}(parent, mask) where {N,M} = new(parent, mask)
 end
-@inline function BitMaskedBitArray(parent::BitArray{N}, mask::BitArray{M}) where {N,M}
+function BitMaskedBitArray(parent::BitArray{N}, mask::BitArray{M}) where {N,M}
```
mbauman, line 1105:
```suggestion
@inline function BitMaskedBitArray(parent::BitArray{N}, mask::BitArray{M}) where {N,M}
```

This one was overzealous — we should keep the force `@inline` here for the`@boundscheck`.  It currently still inlines, but I think it's good practice to keep this here in case something else in here gets big enough to lose the auto-inline.

StefanKarpinski:
Beautiful!

nanosoldier:
[Your benchmark job](https://github.com/JuliaLang/julia/pull/35675#issuecomment-622130478) has completed - possible performance regressions were detected. A full report can be found [here](https://github.com/JuliaCI/BaseBenchmarkReports/blob/e49bed0846aca06d99cd7c22a98a5f38135cbf4a/3d3d52f_vs_2f90dde/report.md). cc @ararslan

timholy:
Very impressive gain!

StefanKarpinski:
Making the inlining cost model more accurate by making reality match the model FTW.

mbauman:
Well that turned out exceptionally well!  I wasn't expecting any run-time performance gains but was fearful of losses.  Looks like there are just a handful of extra allocations in a few cases... but it's in exchange for a 20%+ compile time speedup in the unit tests.  I'll see if anything can be done about those allocations, but I think the fact that this will allow for future runtime improvements (like #30973) will make the broadcast superusers more than pleased.

mbauman:
Ok, the regressions are limited to broadcast expressions that use literal pows (like `x.^2`) or types (like `round.(Int, x)`).  These end up using Refs inside tuples — like `(Ref(^), x, Ref(Val(2)))` and `(Ref{Type{Int}}(Int), x)`.  The problem is that these are _properly_ mutable and thus must be allocated because an inner (non-inlined) function could pull them out and store them somewhere.  In short, this exasperates the need for a true immutable scalar type for broadcast; #18379.

StefanKarpinski:
Maybe `ImmutableRef` would be a good name for it?

vtjnash:
The main confusion with that might be that it's not really a `Ref` unless it's `mutable` (as that's the Julia keyword for pinning something to the heap to make it reference-able). It might more accurately be `Scalarize`, but I think `Some` might fit here too?

mbauman:
Oh interesting.  `Some` is a new idea. It's exactly the `struct` we need and a half-decent (short!) word, but it doesn't have the `getindex`/`length`/`size` methods we need.  Before settling on `Ref` I had once called this a `struct ZeroDimArray{T} <: AbstractArray{T,0}` in the broadcast internals.  `Scalar` (or some variation thereof) sure seems to be the crowd favorite word here, but gosh [do I find it confusing](https://github.com/JuliaLang/julia/pull/35591#issuecomment-620227694).  Imagine adding this `Scalar` into the mix in [this FAQ](https://docs.julialang.org/en/v1/manual/faq/#What-are-the-differences-between-zero-dimensional-arrays-and-scalars?-1), for example.

I reviewed all the issues/discourse threads I could find to look for other alternatives, but the above is just about the universe of what I found (`Nullable` anyone? hah).  All the typical verbiage tends have mutable associations (e.g., `Box`).

mbauman:
What about using a lazy `Fill` a la [FillArrays](https://github.com/JuliaArrays/FillArrays.jl)?  Inspired by #33206, which gives:

```Julia
julia> show(reshape([1], ()))
fill(1)
```

It actually fits linguistically, too — you can imagine the `Fill()`'ed argument as filling in across the broadcasted shape.  The downside is that it's not as simple anymore — it covers far more than just a zero-dimensional array — and coordinating with FillArrays makes this more complicated yet.

mbauman:
Back to the issue at hand, I suppose this points towards another form of performance regression that this would introduce — it'd block constant propagation in the same vein that `Ref` blocked the constant propagation in the motivating example in #35591.

vtjnash:
> but it doesn't have the methods we need

Mostly, neither did `Ref` until it got co-opted for this use. And `Some` is the new name for the `Nullable` container. The `Singleton` doesn't seem quite right (it's not the only item in the set)

tkf:
Having `x[]` for `Some` would be nice also if we have more `Union{Nothing,Some}`-based API like #34821 and https://github.com/JuliaLang/julia/issues/35699#issuecomment-623809325.

```julia
x = get(dict, key)
if x !== nothing
    f(x[])  # nicer than `f(something(x))`?
else
    ...
end
```

rfourquet:
I just proposed somewhere else

> what about `Only`? `only` is already used in a container context with only one element, and there would be the property `only(Only(x)) === x`.

But `Some` and an hypothetical `Only` seem redundant. Anyway, only dropping an idea here.

mbauman:
I've rebased, but in doing so I was reminded that the blocker here was performance of `Ref`-ified scalars.  The path forward here is through https://github.com/JuliaLang/julia/pull/39184

DilumAluthge:
> The path forward here is through #39184

Looks like #39184 has been closed. Is there a different PR open that would fix #18379?

PR #35678, opened 2020-05-01
Remove redundant --threads help string

tkf:
I think this line is there by accident. Ref https://github.com/JuliaLang/julia/pull/35108#discussion_r403380123


fredrikekre:
Thanks.

PR #35679, opened 2020-05-01
Move custom error hints to Experimental

### LINKED ISSUE #35671: make register_error_hint experimental
In discussing https://github.com/JuliaLang/julia/pull/35094 on this last triage call before the 1.5 feature freeze, it was felt that this should be an experimental feature for a while. The concern is that the current form may have bad consequences for compilation time and that it may need changing if that turns out to be the case. There are two things to be done:

1. export the new exports from the `Experimental` submodule instead of directly from `Base`
2. document that it is experimental.

timholy:
Closes #35671

### DIFF
--- a/base/experimental.jl
+++ b/base/experimental.jl
@@ -116,4 +116,94 @@ macro optlevel(n::Int)
     return Expr(:meta, :optlevel, n)
 end
 
+# UI features for errors
+
+"""
+    Experimental.register_error_hint(handler, exceptiontype)
+
+Register a "hinting" function `handler(io, exception)` that can
+suggest potential ways for users to circumvent errors.  `handler`
+should examine `exception` to see whether the conditions appropriate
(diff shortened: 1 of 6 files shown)


base/experimental.jl
```diff
@@ -116,4 +116,94 @@ macro optlevel(n::Int)
...
+ERROR: MethodError: no method matching only_int(::Float64)
+Did you mean to call `any_number`?
+Closest candidates are:
+    ...
+```
+
+!!! compat "Julia 1.5"
+    Custom error hints are available as of Julia 1.5.
+!!! warning
+    This interface is experimental and subject to change or removal without notice.
```
fredrikekre, line 175: Maybe suggest that if you implement this you can (should?) put it behind a `if isdefined(Base.Experimental, :register_error_hint)` check?

JeffBezanson (approved):

PR #35686, opened 2020-05-01
upgrade threads from experimental to stable-with-caveats

JeffBezanson:
Any more warnings we should list?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -104,10 +104,13 @@ Command-line option changes
 
 Multi-threading changes
 -----------------------
+
+* Parts of the multi-threading API are now considered stable, with caveats.
+  This includes all documented identifiers from `Base.Threads` except the
+  `atomic_` operations.
 * `@threads` now allows an optional schedule argument. Use `@threads :static ...` to
   ensure that the same schedule will be used as in past versions; the default schedule
   is likely to change in the future.
-
 
 Build system changes
 --------------------
--- a/doc/src/base/multi-threading.md
+++ b/doc/src/base/multi-threading.md
@@ -1,14 +1,26 @@
 # [Multi-Threading](@id lib-multithreading)
 
-This experimental interface supports Julia's multi-threading capabilities. Types and functions
-described here might (and likely will) change in the future.
-
 ```@docs
-Base.Threads.threadid
-Base.Threads.nthreads
 Base.Threads.@threads
 Base.Threads.@spawn
+Base.Threads.threadid
+Base.Threads.nthreads
 ```
+
+## Synchronization
+
+```@docs
+Base.Threads.Condition
+Base.Threads.Event
+```
+
+See also [Synchronization](@ref lib-task-sync).
+
+## Atomic operations
+
+!!! warning
+
+    The API for atomic operations has not yet been finalized and is likely to change.
 
 ```@docs
 Base.Threads.Atomic
@@ -31,7 +43,7 @@ Base.Threads.atomic_fence
 Base.@threadcall
 ```
 
-# Low-level synchronization primitives
+## Low-level synchronization primitives
 
 These building blocks are used to create the regular synchronization objects.
 
--- a/doc/src/manual/multi-threading.md
+++ b/doc/src/manual/multi-threading.md
@@ -213,3 +213,35 @@ therefore a blocking call like other Julia APIs.
 It is very important that the called function does not call back into Julia, as it will segfault.
 
 `@threadcall` may be removed/changed in future versions of Julia.
+
+## Caveats
+
+At this time, most operations in the Julia runtime and standard libraries
+can be used in a thread-safe manner, if the user code is data-race free.
+However, in some areas work on stabilizing thread support is ongoing.
+Multi-threaded programming has many inherent difficulties, and if a program
+using threads exhibits unusual or undesirable behavior (e.g. crashes or
+mysterious results), thread interactions should typically be suspected first.
+
+There are a few specific limitations and warnings to be aware of when using
+threads in Julia:
+
+  * Base collection types require manual locking if used simultaneously by
+    multiple threads where at least one thread modifies the collection
+    (common examples include `push!` on arrays, or inserting
+    items into a `Dict`).
+  * After a task starts running on a certain thread (e.g. via `@spawn`), it
+    will always be restarted on the same thread after blocking. In the future
+    this limitation will be removed, and tasks will migrate between threads.
+  * `@threads` currently uses a static schedule, using all threads and assigning
+    equal iteration counts to each. In the future the default schedule is likely
+    to change to be dynamic.
+  * The schedule used by `@spawn` is nondeterministic and should not be relied on.
+  * Compute-bound, non-memory-allocating tasks can prevent garbage collection from
+    running in other threads that are allocating memory. In these cases it may
+    be necessary to insert a manual call to `GC.safepoint()` to allow GC to run.
+    This limitation will be removed in the future.
+  * Avoid using finalizers in conjunction with threads, particularly if they
+    operate on objects shared by multiple threads.
+  * Avoid running top-level operations, e.g. `include`, or `eval` of type,
+    method, and module definitions in parallel.
--- a/doc/src/manual/multi-threading.md
+++ b/doc/src/manual/multi-threading.md
@@ -241,7 +241,62 @@ threads in Julia:
     running in other threads that are allocating memory. In these cases it may
     be necessary to insert a manual call to `GC.safepoint()` to allow GC to run.
     This limitation will be removed in the future.
-  * Avoid using finalizers in conjunction with threads, particularly if they
-    operate on objects shared by multiple threads.
   * Avoid running top-level operations, e.g. `include`, or `eval` of type,
     method, and module definitions in parallel.
+  * Be aware that finalizers registered by a library may break if threads are enabled.
+    This may require some transitional work across the ecosystem before threading
+    can be widely adopted with confidence. See the next section for further details.
+
+## Safe use of Finalizers
+
+Because finalizers can interrupt any code, they must be very careful in how
+they interact with any global state. Unfortunately, the main reason that
+finalizers are used is to update global state (a pure function is generally
+rather pointless as a finalizer). This leads us to a bit of a conundrum.
+There are a few approaches to dealing with this problem:
+
+1. When single-threaded, code could call the internal `jl_gc_enable_finalizers`
+C function to prevent finalizers from being scheduled
+inside a critical region. Internally, this is used inside some functions (such
+as our C locks) to prevent recursion when doing certain operations (incremental
+package loading, codegen, etc.). The combination of a lock and this flag
+can be used to make finalizers safe.
+
+2. A second strategy, employed by Base in a couple places, is to explicitly
+delay a finalizer until it may be able to acquire its lock non-recursively.
+The following example demonstrates how this strategy could be applied to
+`Distributed.finalize_ref`:
+
+```
+function finalize_ref(r::AbstractRemoteRef)
+    if r.where > 0 # Check if the finalizer is already run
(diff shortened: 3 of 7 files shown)


ViralBShah:
Darn the failures look like they are from the old doc PRs. I'm looking at fixing them.

doc/src/manual/multi-threading.md
```diff
@@ -213,3 +213,33 @@ therefore a blocking call like other Julia APIs.
...
+
+There are a few specific limitations and warnings to be aware of when using
+threads in Julia:
+
+  * Base collection types require manual locking if used simultaneously by
+    multiple threads (common examples include `push!` on arrays, or inserting
+    items into a `Dict`).
+  * After a task starts running on a certain thread (e.g. via `@spawn`), it
+    will always be restarted on the same thread after blocking. In the future
+    this limitation will be removed, and tasks will migrate between threads.
```
tkf, line 234: Can you avoid thread migration by `@spwan @async f()`? If so, maybe mention that this is the forward-compatible way to rely on this property? (Not sure if we should be encouraging this pattern, though.)

doc/src/manual/multi-threading.md
```diff
@@ -213,3 +213,33 @@ therefore a blocking call like other Julia APIs.
...
+Multi-threaded programming has many inherent difficulties, and if a program
+using threads exhibits unusual or undesirable behavior (e.g. crashes or
+mysterious results), thread interactions should typically be suspected first.
+
+There are a few specific limitations and warnings to be aware of when using
+threads in Julia:
+
+  * Base collection types require manual locking if used simultaneously by
+    multiple threads (common examples include `push!` on arrays, or inserting
+    items into a `Dict`).
```
tkf, line 231: Should we be more specific than "used?"; i.e., at least one thread writes?

tkf:
I had hoped that we get structured concurrency #33248 before `@spawn` becomes stable...

JeffBezanson:
The purpose here is to define what "stable" means in this context. It obviously can't mean that we will never change or add anything ever again. `@spawn` is a primitive, and we want to say that you can use it, and we will not delete it in 1.x. Is it possible to document what it means in a way that is useful but allows the changes you want? For example, would we entirely disallow `@spawn` with no `@sync`? If so, I really think the value of that is dubious. To me it seems like we have "goto" now, and might add "while" loops in the future. What's the problem?

JeffBezanson:
Also, I believe structured concurrency applies as soon as you have Tasks. The only thing `@spawn` adds is that Tasks can run on multiple threads, and structured concurrency is not about threads (i.e. it makes sense even on 1 thread). We've had Tasks since v0.1, so if you need structured concurrency before Tasks we are definitely a bit late.

tkf:
Sorry, my comment was a bit too rant-y and not good for starting a constructive discussion.

I hate to say this, but, realistically speaking, I don't think introducing structured concurrency with/before stabilization of `@spawn` is doable or appropriate. I think there is simply too much work to do to introduce structured concurrency even in a non-strict form. Also, we can't have a strict form of structured concurrency without breaking `@async` anyway.

I'll reply to the technical parts of your comments in #33248.

doc/src/base/multi-threading.md
```diff
@@ -1,15 +1,23 @@
...
+
+```@docs
+Base.Threads.Condition
+Base.Threads.Event
+```
+
+See also [Synchronization](@ref lib-task-sync).
+
+## Atomic operations
+
```
vtjnash, line 20: I think we should mark these as subject to possible change.

doc/src/manual/multi-threading.md
```diff
@@ -213,3 +213,33 @@ therefore a blocking call like other Julia APIs.
 It is very important that the called function does not call back into Julia, as it will segfault.
 
 `@threadcall` may be removed/changed in future versions of Julia.
+
+## Caveats
+
+At this time, most operations in the Julia runtime and standard libraries
+are thread safe.
```
vtjnash, line 220:
```suggestion
are thread safe, except for `eval`.
```

doc/src/manual/multi-threading.md
vtjnash: Hm, curious. Probably not—instead seems like `@async` should inherit the sticky bit from the parent in the future.

doc/src/manual/multi-threading.md
```diff
@@ -213,3 +213,33 @@ therefore a blocking call like other Julia APIs.
...
+    this limitation will be removed, and tasks will migrate between threads.
+  * `@threads` currently uses a static schedule, using all threads and assigning
+    equal iteration counts to each. In the future the default schedule is likely
+    to change to be dynamic.
+  * The schedule used by `@spawn` is nondeterministic and should not be relied on.
+  * Compute-bound, non-memory-allocating tasks can prevent garbage collection from
+    running in other threads that are allocating memory. In these cases it may
+    be necessary to insert a manual call to `GC.safepoint()` to allow GC to run.
+    This limitation will be removed in the future.
+  * Avoid using finalizers in conjunction with threads, particularly if they
```
vtjnash, line 243: This item feels like it sticks out in the list as being unclear and unreasonable to state. I think we might need to be more explicit here about that finalizers get run simultaneously with code (kind of always have). I can work on some wording and sample code if that helps.

doc/src/manual/multi-threading.md
mbauman: If we’re going to specifically call out one thing here, I think it should be the data structures. `eval` is covered by the top-level bullet below (but could probably be specifically called out).

doc/src/manual/multi-threading.md
tkf: A function using `@async` internally may be called via `@spawn`.  If `@async` inherits the sticky bit, this internal use of `@async` becomes actually equivalent to `@spawn`, right?

I think it's possible to use `@async` in a way concurrent-safe but thread-unsafe.  If so, isn't it dangerous to inherit the sticky bit?

doc/src/base/multi-threading.md
JeffBezanson: Ok. Do you think we'd need to delete them entirely, or just add a better interface?

doc/src/manual/multi-threading.md
JeffBezanson: Right, the idea is for the subsequent list to list all the exceptions to this. I called out `include` because that will generally involve many kinds of top-level expressions and so probably isn't safe, but I don't think it's useful to say that `eval` is unsafe. Is `eval(:(1+2))` unsafe?

doc/src/manual/multi-threading.md
JeffBezanson: Yes, that would be good, thanks.

vtjnash:
```suggestion
  * Be aware that some library's finalizers may sometimes break badly if
    threads are enabled (though likely are already broken). This may require
    some transitional work across the ecosystem before threading can be widely
    adopted with confidence. Further details on this are below.
```

```md
## Safe use of Finalizers

Because finalizers can interrupt any code, they must be very careful in how
they interact with any global state. Unfortunately, the main reason that
finalizers are used is to update global state (a pure function is generally
rather pointless as a finalizer). This leads us to a bit of a tricky conundrum.
But one which has a few approaches to dealing with:

1. When singled-threaded, code could call access the internal
`jl_gc_enable_finalizers` function to prevent finalizers from being scheduled
inside a critical region. Internally, this is used inside some functions (such
as our C locks) to prevent recursion when doing certain operations (incremental
package loading, codegen, etc.). Observe that this combination (of an
inefficient spinlock and setting this flag) can be used to make finalizers safe
to work.

2. A second strategy, employed by Base in a couple places, is to explicitly try
to acquire the lock non-recursively. For example, `WeakKeyDict` and
`Distributed.finalize_ref` take this approach (though both only handle the
single-threading case correctly). This example below is updated to work in the
multi-threaded case:

        function finalize_ref(r::AbstractRemoteRef)
            if r.where > 0 # Check if the finalizer is already run
                if islocked(client_refs) || !trylock(client_refs)
                    # delay finalizer for later if we aren't free to acquire the lock
                    finalizer(finalize_ref, r)
                    return nothing
                end
                try # `lock` should always be followed by `try` in correct code
                    if r.where > 0 # Must check again here
                        # Do actual cleanup here
                        r.where = 0
                    end
                finally
                    unlock(client_refs)
                end
            end
            nothing
        end

3. A related third strategy is to use a yield-free queue. We don't currently
have a lock-free queue implemented in base, but
`Base.InvasiveLinkedListSynchronized{T}` is suitable. This can frequently be a
good strategy to use for code with event loops. For example, this strategy is
employed by `Gtk.jl` to manage lifetime ref-counting. In this approach, we
don't do any explicit work inside the `finalizer`, and merely add it to a queue
to be dealt with at a safer time. Actually, `schedule` already uses this, so
defining the finalizer as `x -> @spawn do_cleanup(x)` is one example of this
approach. Note however that this doesn't control which thread `do_cleanup` is
running on here, so `do_cleanup` would still need to acquire a lock. That
doesn't need to be true if you implement your own queue, as you can explicitly
only drain that queue from your thread. Additionally, instead of the `@spawn`
convenience macro, you could manually schedule it on a specific thread:

        t = @task do_cleanup(x)
        ccall(:jl_set_task_tid, Cvoid, (Any, Cint), t, 0) # make sticky to thread with tid=1
        schedule(t)
```

JeffBezanson: That's excellent, thanks. I'm not sure we want to mention `jl_set_task_tid`, since then it might be considered "documented" and we will have to keep it.

doc/src/base/multi-threading.md
vtjnash: Yeah, just don’t want to commit exactly to these yet

doc/src/manual/multi-threading.md
vtjnash: I think the idea generally is that if it cant be done without eval, it’s unsafe right now. While that could have been a closure or something similar.

doc/src/manual/multi-threading.md
vtjnash: I’m okay with that, though we might want to make it look less like a hack

NEWS.md
```diff
@@ -97,7 +97,9 @@ Command-line option changes
 
 Multi-threading changes
 -----------------------
-
+* Parts of the multi-threading API are now considered stable, with caveats.
+  This includes all identifiers exported from `Base.Threads` except the
```
tkf, line 101:
```suggestion
  This includes all documented identifiers from `Base.Threads` except the
```

I think it's better to not restrict to exported identifier because we may have `Threads.foreach` #34543 (or even `Threads.map`?).  As we are not going to export `Iterators.map` https://github.com/JuliaLang/julia/pull/34352#discussion_r365980331, it makes sense to not do this in `Base.Threads` as well.

doc/src/manual/multi-threading.md
JeffBezanson: Ok, that's a clearer rule. Maybe we could be even more specific and list some things like method definitions, global assignments, ... ?

doc/src/manual/multi-threading.md
```diff
@@ -213,3 +213,89 @@ therefore a blocking call like other Julia APIs.
 It is very important that the called function does not call back into Julia, as it will segfault.
 
 `@threadcall` may be removed/changed in future versions of Julia.
+
+## Caveats
+
+At this time, most operations in the Julia runtime and standard libraries
+are thread safe.
```
tkf, line 220: I find it somewhat vague to say Julia APIs "_are_ thread safe."  I think the point is that most of those APIs _themselves_ are data race-free.  But, of course, it doesn't prevent users to write data races.  Maybe it'd be clearer to phrase it like:

> At this time, most operations in the Julia runtime and standard libraries can be used in a thread-safe manner, if the user code is data race-free.

?

PR #35705, opened 2020-05-02
Add documentation to the sized identity matrix

ro-ble:
As proposed [on discourse](https://discourse.julialang.org/t/linear-algebra-tricks/38618/10?u=roble), I would like to add this to improve the documentation.

### DIFF
--- a/stdlib/LinearAlgebra/docs/src/index.md
+++ b/stdlib/LinearAlgebra/docs/src/index.md
@@ -337,6 +337,7 @@ LinearAlgebra.UnitUpperTriangular
 LinearAlgebra.UpperHessenberg
 LinearAlgebra.UniformScaling
 LinearAlgebra.I
+(LinearAlgebra.I::LinearAlgebra.UniformScaling)(::Integer)
 LinearAlgebra.Factorization
 LinearAlgebra.LU
 LinearAlgebra.lu
--- a/stdlib/LinearAlgebra/docs/src/index.md
+++ b/stdlib/LinearAlgebra/docs/src/index.md
@@ -337,7 +337,7 @@ LinearAlgebra.UnitUpperTriangular
 LinearAlgebra.UpperHessenberg
 LinearAlgebra.UniformScaling
 LinearAlgebra.I
-(::LinearAlgebra.UniformScaling)(::Integer)
+LinearAlgebra.(::LinearAlgebra.UniformScaling)(::Integer)
 LinearAlgebra.Factorization
 LinearAlgebra.LU
 LinearAlgebra.lu
--- a/stdlib/LinearAlgebra/docs/src/index.md
+++ b/stdlib/LinearAlgebra/docs/src/index.md
@@ -337,7 +337,7 @@ LinearAlgebra.UnitUpperTriangular
 LinearAlgebra.UpperHessenberg
 LinearAlgebra.UniformScaling
 LinearAlgebra.I
-(LinearAlgebra.I::LinearAlgebra.UniformScaling)(::Integer)
+(::LinearAlgebra.UniformScaling)(::Integer)
 LinearAlgebra.Factorization
 LinearAlgebra.LU
 LinearAlgebra.lu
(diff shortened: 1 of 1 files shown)


andreasnoack:
Looks like Documenter doesn't understand this. @fredrikekre do you know if there is a way to do this?

ro-ble:
The function seems to have documentation strings. However I believe that documenter struggles with the function declaration in the index.md file. From my still small Julia knowledge, I would guess, that it is something like this: https://docs.julialang.org/en/v1/manual/methods/#Function-like-objects-1
I believe functions like this have already been documented. I will search for one

fredrikekre:
I don't think it is possible to include this in `@docs`-blocks, not sure though. @mortenpi?

mortenpi:
Yeah, this is currently problematic. The docsystem is a bit buggy, but also Documenter probably can't handle it properly: https://github.com/JuliaDocs/Documenter.jl/issues/558

Although, if it would work, I think the syntax for this in at-docs blocks should be

```
(::LinearAlgebra.UniformScaling)(::Integer)
```

ro-ble:
I tried to replicate the behaviour of Documenter in REPL. Following things happened:

`@doc (::UniformScaling)(::Integer)` -> returns the correct documentation

`@doc(::LinearAlgebra.UniformScaling)(::Integer)` -> `ERROR: syntax: invalid "::" syntax`

`@doc(::(LinearAlgebra.UniformScaling))(::Integer)` -> `ERROR: syntax: invalid "::" syntax`

I am not shure how to proceed... Maybe open an issue at the Documenter repo?

mortenpi:
Since it's broken from the docsystem side, I would suggest merging the docstring for [`(I::UniformScaling)(n::Integer)`](https://github.com/JuliaLang/julia/blob/7be230bb6178ed74297a92d2e50ef8d8d801351d/stdlib/LinearAlgebra/src/uniformscaling.jl#L51-L73) into the docstring of [`struct UniformScaling`](https://github.com/JuliaLang/julia/blob/7be230bb6178ed74297a92d2e50ef8d8d801351d/stdlib/LinearAlgebra/src/uniformscaling.jl#L6-L28) and maybe also add a usage example of `I(n)` into the docstring of [`const I`](https://github.com/JuliaLang/julia/blob/7be230bb6178ed74297a92d2e50ef8d8d801351d/stdlib/LinearAlgebra/src/uniformscaling.jl#L33-L48).

ro-ble:
> I would suggest merging the docstring for [`(I::UniformScaling)(n::Integer)`](https://github.com/JuliaLang/julia/blob/7be230bb6178ed74297a92d2e50ef8d8d801351d/stdlib/LinearAlgebra/src/uniformscaling.jl#L51-L73) into the docstring of [`struct UniformScaling`](https://github.com/JuliaLang/julia/blob/7be230bb6178ed74297a92d2e50ef8d8d801351d/stdlib/LinearAlgebra/src/uniformscaling.jl#L6-L28) and maybe also add a usage example of `I(n)` into the docstring of [`const I`](https://github.com/JuliaLang/julia/blob/7be230bb6178ed74297a92d2e50ef8d8d801351d/stdlib/LinearAlgebra/src/uniformscaling.jl#L33-L48).

Yeah that would certainly be a valid solution. But before I do this, I will try to resolve this issue.  According to zekeriya.sari's [answer](https://discourse.julialang.org/t/how-to-document-this/39217/2?u=roble), I do believe that It should be possible.

stdlib/LinearAlgebra/docs/src/index.md
```diff
@@ -337,6 +337,7 @@ LinearAlgebra.UnitUpperTriangular
 LinearAlgebra.UpperHessenberg
 LinearAlgebra.UniformScaling
 LinearAlgebra.I
+LinearAlgebra.(::LinearAlgebra.UniformScaling)(::Integer)
```
mortenpi, line 340:
```suggestion
LinearAlgebra.UniformScaling(::Integer)
```

I did a bit more digging and I think this might function as a workaround.

The issue is that internally the call syntax / functor docstring gets stored with the same key (which is just the argument signature, i.e. `Tuple{Integer}` in this case) as a normal method / constructor would. So, from Documenter's perspective, the functor syntax is not a thing and only things that exist are methods on types.

However, as there is no conflicting `UniformScaling(::Integer)` constructor, we should be able to take advantage of that and just tell Documenter to include that and it will splice in the docstring for `(::LinearAlgebra.UniformScaling)(::Integer)`.

ro-ble: I am really glad you were able to find a clean solution. Thank you very much for helping!

PR #35150, opened 2020-03-18
Support `⊙` and `⊗` as elementwise- and outer-product operators

timholy:
While we have broadcasting and `a*b'`, sometimes you need to pass an operator as an argument to a function. Since we already have `⋅` as a synonym for `dot`, these seemed to make sense.

PR #35706, opened 2020-05-03
allow standalone dotted operators

### LINKED ISSUE #34156: Lower standalone dotted infix operators as anonymous functions
I could've sworn we already had an issue tracking this, but I couldn't find it.  In short, I'd love to have support for standalone `(.+)` and friends to lower to `(args...)->(+).(args...)`.  This would enable their use in higher order functions and generally make them more useful.

They currently parse as identifiers — and can be assigned to! — so this would be a breaking minor change.

simeonschaub:
This lowers dotted operators to `Base.BroadcastOp(op)`, which allows for passing them as closures to functions. One example where this can be really helpful is for array-of-array type data structures, where you want to map a broadcasted function, so you can just do `map(.+, [[1,2], [3,4]], [5, 6])`. This came up in #35150, where`hadamard` was proposed, because mapping broadcasted multiplication is currently quite cumbersome to write. Does this implementation make sense?

close #34156

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -29,8 +29,8 @@ New language features
   this to 0 or 1 can provide significant latency improvements ([#34896]).
 
 * Standalone "dotted" operators now get lowered to `Base.BroadcastOp(op)` and can be passed to
-  higher-order functions, i.e. `.op` is functionally equivalent to `(x...) -> op.(x...)`.
-  ([#34156], [#35706])
+  higher-order functions, i.e. `.op` is functionally equivalent to `(x...) -> (op).(x...)`.
+  ([#35706])
 
 Language changes
 ----------------
--- a/NEWS.md
+++ b/NEWS.md
@@ -28,6 +28,10 @@ New language features
   `Base.Experimental.@optlevel n`. For code that is not performance-critical, setting
   this to 0 or 1 can provide significant latency improvements ([#34896]).
 
+* Standalone "dotted" operators now get lowered to `Base.BroadcastOp(op)` and can be passed to
+  higher-order functions, i.e. `.op` is functionally equivalent to `(x...) -> op.(x...)`.
+  ([#34156], [#35706])
+
 Language changes
 ----------------
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -28,7 +28,7 @@ New language features
   `Base.Experimental.@optlevel n`. For code that is not performance-critical, setting
   this to 0 or 1 can provide significant latency improvements ([#34896]).
 
-* Standalone "dotted" operators now get lowered to `Base.BroadcastOp(op)` and can be passed to
+* Standalone "dotted" operators now get lowered to `Base.BroadcastFunction(op)` and can be passed to
   higher-order functions, i.e. `.op` is functionally equivalent to `(x...) -> (op).(x...)`.
   ([#35706])
 
--- a/base/broadcast.jl
+++ b/base/broadcast.jl
@@ -1290,7 +1290,7 @@ end
 
 @inline (op::BroadcastOp)(x...) = op.f.(x...)
 
-show(io::IO, op::BroadcastOp) = print(io, "Base.BroadcastOp(", op.f, ')')
+show(io::IO, op::BroadcastOp) = print(io, BroadcastOp, '(', op.f, ')')
 show(io::IO, ::MIME"text/plain", op::BroadcastOp) = show(io, op)
 
 end # module
--- a/base/broadcast.jl
+++ b/base/broadcast.jl
@@ -1265,15 +1265,15 @@ end
     BroadcastOp{F} <: Function
 
 Represents the "dotted" version of an operator, which broadcasts the operator over its
-arguments, so `BroadcastOp(op)` is functionally equivalent to `(x...) -> op.(x...)`.
+arguments, so `BroadcastOp(op)` is functionally equivalent to `(x...) -> (op).(x...)`.
 
 Can be created by just passing an operator preceded by a dot to a higher-order function.
 
 # Examples
 ```jldoctest
-julia> a = [reshape(i:i+3, 2, 2) for i in [1, 5]];
+julia> a = [[1 3; 2 4], [5 7; 6 8]]
 
-julia> b = [reshape(i:i+3, 2, 2) for i in [9, 13]];
+julia> b = [[9 11; 10 12], [13 15; 14 16]]
 
 julia> map(.*, a, b)
 2-element Array{Array{Int64,2},1}:
@@ -1289,5 +1289,8 @@ struct BroadcastOp{F} <: Function
 end
 
 @inline (op::BroadcastOp)(x...) = op.f.(x...)
+
+show(io::IO, op::BroadcastOp) = print(io, "Base.BroadcastOp(", op.f, ')')
+show(io::IO, ::MIME"text/plain", op::BroadcastOp) = show(io, op)
 
 end # module
--- a/base/broadcast.jl
+++ b/base/broadcast.jl
@@ -1261,6 +1261,29 @@ end
 end
 @inline broadcasted(::S, f, args...) where S<:BroadcastStyle = Broadcasted{S}(f, args)
 
+"""
+    BroadcastOp{F} <: Function
+
+Represents the "dotted" version of an operator, which broadcasts the operator over its
+arguments, so `BroadcastOp(op)` is functionally equivalent to `(x...) -> op.(x...)`.
+
+Can be created by just passing an operator preceded by a dot to a higher-order function.
+
+# Examples
+```jldoctest
+julia> a = [reshape(i:i+3, 2, 2) for i in [1, 5]];
+
+julia> b = [reshape(i:i+3, 2, 2) for i in [9, 13]];
+
+julia> map(.*, a, b)
+2-element Array{Array{Int64,2},1}:
+ [9 33; 20 48]
+ [65 105; 84 128]
+
+julia> Base.BroadcastOp(+)(a, b) == a .+ b
+true
+```
+"""
 struct BroadcastOp{F} <: Function
     f::F
 end
--- a/base/broadcast.jl
+++ b/base/broadcast.jl
@@ -11,7 +11,7 @@ using .Base.Cartesian
 using .Base: Indices, OneTo, tail, to_shape, isoperator, promote_typejoin,
              _msk_end, unsafe_bitgetindex, bitcache_chunks, bitcache_size, dumpbitcache, unalias
 import .Base: copy, copyto!, axes
-export broadcast, broadcast!, BroadcastStyle, broadcast_axes, broadcastable, dotview, @__dot__, broadcast_preserving_zero_d, BroadcastOp
+export broadcast, broadcast!, BroadcastStyle, broadcast_axes, broadcastable, dotview, @__dot__, broadcast_preserving_zero_d, BroadcastFunction
 
 ## Computing the result's axes: deprecated name
 const broadcast_axes = axes
@@ -1262,10 +1262,10 @@ end
 @inline broadcasted(::S, f, args...) where S<:BroadcastStyle = Broadcasted{S}(f, args)
 
 """
-    BroadcastOp{F} <: Function
+    BroadcastFunction{F} <: Function
 
 Represents the "dotted" version of an operator, which broadcasts the operator over its
-arguments, so `BroadcastOp(op)` is functionally equivalent to `(x...) -> (op).(x...)`.
+arguments, so `BroadcastFunction(op)` is functionally equivalent to `(x...) -> (op).(x...)`.
 
 Can be created by just passing an operator preceded by a dot to a higher-order function.
 
@@ -1280,17 +1280,17 @@ julia> map(.*, a, b)
  [9 33; 20 48]
  [65 105; 84 128]
 
-julia> Base.BroadcastOp(+)(a, b) == a .+ b
+julia> Base.BroadcastFunction(+)(a, b) == a .+ b
 true
 ```
 """
-struct BroadcastOp{F} <: Function
+struct BroadcastFunction{F} <: Function
     f::F
 end
 
-@inline (op::BroadcastOp)(x...) = op.f.(x...)
+@inline (op::BroadcastFunction)(x...) = op.f.(x...)
 
-Base.show(io::IO, op::BroadcastOp) = print(io, BroadcastOp, '(', op.f, ')')
-Base.show(io::IO, ::MIME"text/plain", op::BroadcastOp) = show(io, op)
+Base.show(io::IO, op::BroadcastFunction) = print(io, BroadcastFunction, '(', op.f, ')')
+Base.show(io::IO, ::MIME"text/plain", op::BroadcastFunction) = show(io, op)
 
 end # module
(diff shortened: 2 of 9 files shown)


base/broadcast.jl
```diff
@@ -1261,4 +1261,10 @@ end
 end
 @inline broadcasted(::S, f, args...) where S<:BroadcastStyle = Broadcasted{S}(f, args)
 
+struct BroadcastOp{F}
+    f::F
+end
+
+(op::BroadcastOp)(x...) = op.f.(x...)
```
stevengj, line 1268: Have you checked that the compiler specializes on F here? i.e. that performance is unchanged?

stevengj:
Does this still fuse nested dot operations with other dot calls?

base/broadcast.jl
simeonschaub: Good catch! There seems to be room for improvement:
```
julia> using BenchmarkTools

julia> a = [rand(10, 10) for _ in 1:100];

julia> b = [rand(10, 10) for _ in 1:100];

julia> @btime map((x, y) -> (Base.@_inline_meta; x .* y), $a, $b);
  14.528 μs (101 allocations: 88.38 KiB)

julia> @btime map(.*, $a, $b);
  53.660 μs (401 allocations: 97.75 KiB)
```
Shouldn't be too difficult to fix that.

simeonschaub:
No, this doesn't do any fusing with other dot calls, but I am not sure we really want that, since this would make broadcast fusing less transparent and make it not just a syntactic guarantee anymore. Since it just ends up calling `Base.materialize(Base.Broadcasted(op, x...))`, it will fuse with `Base.Broadcasted` objects though, if you explicitely construct them.

base/broadcast.jl
simeonschaub: Adding `@inline` here fixes this performance differences for me.

stevengj:
No, I mean if you do `2 .* sqrt.(x)` will it still fuse?   I haven't checked whether your lowering operation happens before fusion (bad) or after fusion (good).

base/broadcast.jl
stevengj: Maybe you just need `(op::BroadcastOp{F})(x...) where {F} = op.f.(x...)`?  Adding an explicit type parameter usually tells the compiler to specialize on function types.

simeonschaub:
Yes, those cases should not be affected at all, since broadcast fusion happens before this expansion. That's also why `(.*).(a, b)` works just like expected, for example.

base/broadcast.jl
simeonschaub: I have tried that as well, but it didn't make any difference. It seems like the compiler needs to be explicitely told to inline here.

base/broadcast.jl
```diff
@@ -1261,4 +1261,10 @@ end
 end
 @inline broadcasted(::S, f, args...) where S<:BroadcastStyle = Broadcasted{S}(f, args)
 
+struct BroadcastOp{F} <: Function
```
stevengj, line 1264: Needs a documentation string.

stevengj:
Also needs a NEWS item under "language changes".

My understanding is that the only cases that this affects were previously syntax errors, so that this is a backwards-compatible change?

stevengj:
`@nanosoldier `runbenchmarks("broadcast", vs = ":master")`

simeonschaub:
Technically, this change could be breaking, if a package assigned to `var".+"`, for example. I can't imagine that any package would have this do anything other than what's proposed in this PR though.

simeonschaub:
It should also be fairly easy to make this backwards-compatible in `Compat.jl`, by just defining `var".op" = BroadcastOp(op)` for every operator.

base/broadcast.jl
mbauman: Does this need to be a struct?  It could just be:

```julia
broadcast_op(f) = (x...) -> f.(x...)
```

That's how, for example, the somewhat-equivalent `Base.splat` function works.  The downside is that you can no longer dispatch on it, I suppose.

This also works for non-operators, which makes me want for a better name.  Maybe `broadcasting`?  The gerund actually makes sense here, and I think `Base.splatting` might even make more sense than `Base.splat`.

nanosoldier:
[Your benchmark job](https://github.com/JuliaLang/julia/pull/35706#issuecomment-623518378) has completed - possible performance regressions were detected. A full report can be found [here](https://github.com/JuliaCI/BaseBenchmarkReports/blob/7128ee31d74bfa9e154344ed94a028f2df8f4877/c1d0512_vs_f1d10e7/report.md). cc @ararslan

base/broadcast.jl
simeonschaub: The reason I decided to make this a struct is primarily for better stack traces, which is, I believe, the reason we also have `Base.Fix2`, for example. It would also open up the opportunity for higher-order functions to specialize on `BroadcastOp`, i.e. to fuse it with other broadcasting operations.
I am certainly open to suggestions for a better name! Having `broadcasting` together with `broadcast` and `broadcasted` might be a little hard to remember and perhaps throw off some non-native speakers.

stevengj: A `struct` also allows much more informative pretty-printing, e.g. if someone types `julia> .*` in the REPL.

NEWS.md
```diff
@@ -28,6 +28,10 @@ New language features
   `Base.Experimental.@optlevel n`. For code that is not performance-critical, setting
   this to 0 or 1 can provide significant latency improvements ([#34896]).
 
+* Standalone "dotted" operators now get lowered to `Base.BroadcastOp(op)` and can be passed to
+  higher-order functions, i.e. `.op` is functionally equivalent to `(x...) -> op.(x...)`.
+  ([#34156], [#35706])
```
stevengj, line 33: We typically just refer to the PR here.

stevengj: probably `(op).(x...)` too, since that is the way you would dot-call operators.

base/broadcast.jl
```diff
@@ -1261,4 +1261,33 @@ end
 end
 @inline broadcasted(::S, f, args...) where S<:BroadcastStyle = Broadcasted{S}(f, args)
 
+"""
+    BroadcastOp{F} <: Function
+
+Represents the "dotted" version of an operator, which broadcasts the operator over its
+arguments, so `BroadcastOp(op)` is functionally equivalent to `(x...) -> op.(x...)`.
```
stevengj, line 1268: `(op).(x...)`

base/broadcast.jl
```diff
@@ -1261,4 +1261,33 @@ end
...
+
+Can be created by just passing an operator preceded by a dot to a higher-order function.
+
+# Examples
+```jldoctest
+julia> a = [reshape(i:i+3, 2, 2) for i in [1, 5]];
```
stevengj, line 1274: clearer to just do
```jl
a = [ [1 3; 2 4], [5 7; 6 8] ]
b = [ [9 11; 10 12], [13 15; 14 16] ]
```

base/broadcast.jl
```diff
@@ -1261,4 +1261,33 @@ end
...
+
+julia> Base.BroadcastOp(+)(a, b) == a .+ b
+true
+```
+"""
+struct BroadcastOp{F} <: Function
+    f::F
+end
+
+@inline (op::BroadcastOp)(x...) = op.f.(x...)
```
stevengj, line 1291: I would probably add a pretty-printing method like:
```
show(io::IO, op::BroadcastOp) = print(io, "BroadcastOp(", op.f, ')')
show(io::IO, ::MIME"text/plain", op::BroadcastOp) = show(io, op)
```
(the latter is needed as otherwise it calls a method for `::Function`).

simeonschaub: Should we print this as `Base.BroadcastOp`? That way, it can just be copy-pasted

base/broadcast.jl
```diff
@@ -1261,4 +1261,36 @@ end
...
+
+show(io::IO, op::BroadcastOp) = print(io, "Base.BroadcastOp(", op.f, ')')
```
stevengj, line 1293: I think you can just do
```jl
print(io, BroadcastOp, '(', op.f, ')')
```
and it will automatically prepend the `Base.` if it `BroadcastOp` not exported.

base/broadcast.jl
tkf: > This also works for non-operators

So how about `BroadcastingFunction`?  Since it's unlikely that this is typed interactively, maybe something very descriptive makes sense?

base/broadcast.jl
simeonschaub: That prints for me as `Base.Broadcast.BroadcastOp`, which is probably fine as well

base/broadcast.jl
stevengj: Or maybe just `BroadcastFunc`.

base/broadcast.jl
stevengj: I think we want to export it from the `Broadcast` module into `Base`?

stevengj: i.e. add an `export` line.

simeonschaub: Yes, it is already exported above, `print` just doesn't realize that

base/broadcast.jl
tkf: Sounds good to me too.

base/broadcast.jl
simeonschaub: This also seems to print the type parameter, so I think I will just leave it as it was originally

simeonschaub: Yes, it is already exported above, print just doesn't realize that

simeonschaub:
One thing that's still a bit unfortunate is that just typing `.*` into the REPL still throws a `UndefVarError` and doesn't get this lowering pass. I looked into this a bit and it looks like this is because plain symbols get special-cased in `jl_toplevel_eval_flex`. Since this is going to be passed to functions directly most of the times anyways, I don't think that should really be a problem, but it would be nice to get this fixed eventually. Maybe somone who knows that part of the code base a bit better than I do has an idea for how to fix this?

stevengj:
Seems like it should be changed in the `if (jl_is_symbol(e))` branch of `interpreter.c:eval_value`, which I think is what is eventually executed by `jl_toplevel_eval_flex`.

simeonschaub:
I believe this should now work. Please don't ask me how! :laughing: It would be good if you or someone else could give this a more thorough review, since I was literally just poking at stuff until it worked.

src/ast.c
```diff
@@ -1037,6 +1037,15 @@ JL_DLLEXPORT int jl_operator_precedence(char *sym)
...
 }
 
+JL_DLLEXPORT int jl_is_dotop_named(char *sym)
+{
+    jl_ast_context_t *ctx = jl_ast_ctx_enter();
+    fl_context_t *fl_ctx = &ctx->fl;
+    int res = fl_applyn(fl_ctx, 1, symbol_value(symbol(fl_ctx, "dotop-named?")), symbol(fl_ctx, sym));
+    jl_ast_ctx_leave(ctx);
+    return res;
+}
```
stevengj, line 1047: Can't we just check if the first character is a dot?   I'm a little concerned at the interpreter overhead of calling back to scheme for every symbol.

src/toplevel.c
```diff
@@ -616,6 +616,12 @@ jl_value_t *jl_toplevel_eval_flex(jl_module_t *JL_NONNULL m, jl_value_t *e, int
             while (*n == '_') ++n;
             if (*n == 0 && n > n0)
                 jl_eval_errorf(m, "all-underscore identifier used as rvalue");
+            // .op needs special lowering
+            if (jl_is_dotop_named(n)) {
```
stevengj, line 620: Should this code be here or in `eval_value`?

src/ast.c
stevengj: I guess if we only call it for top-level symbol expressions it isn't a problem.

simeonschaub: Yes, only checking the first symbol for a dot in the c code makes sense. All the magic then happens in the scheme code afterwards anyways.

src/toplevel.c
simeonschaub: It seemed similar enough to the r-value check, so I put it here. I can certainly change that, though.

stevengj: I guess this place is good — the rvalue check is here because the symbol didn't hit the lowering code (which also checks for underscore names), and so we want the dotop check here for the same reason.

base/broadcast.jl
```diff
@@ -1261,4 +1261,36 @@ end
 end
 @inline broadcasted(::S, f, args...) where S<:BroadcastStyle = Broadcasted{S}(f, args)
 
+"""
+    BroadcastOp{F} <: Function
```
stevengj, line 1265: Change to `BroadcastFunction` as commented elsewhere.

JeffBezanson:
I'm in favor of this, but I find it pretty unsettling for a symbol to lower to a non-trivial expression. A small example is something like the `@which` macro, which assumes a symbol can just be looked up in the calling module. I think there'd be many other similar gotchas in the future.

In hindsight, we should have parsed `.+` as `(dot +)`. But, I wonder if we can still do that when `.+` is in a context where it isn't called, e.g. just an atom by itself. My thinking is that any macros that look for it must currently only be handling it as the first argument to `:call`.

simeonschaub:
> But, I wonder if we can still do that when `.+` is in a context where it isn't called, e.g. just an atom by itself. My thinking is that any macros that look for it must currently only be handling it as the first argument to `:call`.

I am afraid, I can't quite follow. Should this whole transformation occur in the parser first? What I like about doing this during lowering is that it's (almost) entirely non breaking. But I agree that parsing as `(dot op)` would have been the nicer solution.

JeffBezanson:
If we change the parsing, the main thing that can break is macros --- a macro looking for `Symbol(".+")` would not find it if it were parsed as `Expr(:dot, :+)` instead. But, that symbol is currently only useful when it's called, as in `Expr(:call, :.+, ...)`, so I'm guessing any such macros would only handle expressions of that form. Outside that context, it might be safe to change the parsing.

Or, we could potentially keep compatibility by converting `(dot +)` to `.+` in all macro arguments, and converting back on the way out.

c42f:
> Or, we could potentially keep compatibility by converting `(dot +)` to `.+` in all macro arguments, and converting back on the way out.

I do like this idea, but ASTs can come from sources other than macro arguments so I'm not sure it would work well enough for compatibility. Kind of hard to say without testing.

base/broadcast.jl
```diff
@@ -1261,4 +1261,36 @@ end
...
+
+julia> Base.BroadcastFunction(+)(a, b) == a .+ b
+true
+```
+"""
+struct BroadcastFunction{F} <: Function
+    f::F
+end
+
+@inline (op::BroadcastFunction)(x...) = op.f.(x...)
```
mbauman, line 1291: Might as well pass along kwargs here, too, yeah?

JeffBezanson:
At some point I'll try implementing my comment above --- keep parsing the same for called operators, parse as `(dot +)` outside that. Then we can at least PkgEval and see if it breaks anything.

base/broadcast.jl
simeonschaub: Makes sense! Didn't think of that.

simeonschaub:
That would be great! I tried implementing your suggestion myself, but didn't get very far with it.

JeffBezanson:
Replaced by #37583 I assume?

simeonschaub:
Oh, sure! Just forgot to close this.

Issue #33865, opened 2019-11-16
World age problem with global_logger and tasks launched before logger definition


tkf:
It seems `global_logger` does not work well with tasks in the following scenario:

1. A task using the logger is launched.
2. A new logger is defined (e.g., via import).
3. The new logger is set as the global logger.
4. The task tries to log something.

As an example, running the following code in a fresh `julia` process (i.e., the one that has not imported `TerminalLoggers` yet) throws a method error.

```julia
using Logging

ch = Channel() do ch
    for _ in 1:2
        try
            @info take!(ch)
        catch err
            bt = catch_backtrace()
            with_logger(ConsoleLogger()) do
                @error "`@info` threw" exception = (err, bt)
            end
        end
    end
end

put!(ch, "hello to ConsoleLogger")

using TerminalLoggers
# yield()

global_logger(TerminalLogger())
put!(ch, "hello to TerminalLogger")
```

Output:

```
[ Info: hello to ConsoleLogger
┌ Error: `@info` threw
│   exception =
│    MethodError: no method matching shouldlog(::TerminalLogger, ::Base.CoreLogging.LogLevel, ::Module, ::String, ::Symbol)
│    The applicable method may be too new: running in world age 26999, while current world is 27001.
│    Closest candidates are:
│      shouldlog(::TerminalLogger, ::Any, ::Any, ::Any, ::Any) at /home/takafumi/.julia/dev/TerminalLoggers/src/TerminalLogger.jl:40 (method too new to be called from this world context.)
│      shouldlog(::Base.CoreLogging.NullLogger, ::Any...) at logging.jl:84
│      shouldlog(::Base.CoreLogging.SimpleLogger, ::Any, ::Any, ::Any, ::Any) at logging.jl:529
│      ...
│    Stacktrace:
│     [1] macro expansion at ./logging.jl:316 [inlined]
│     [2] (::var"#3#5")(::Channel{Any}) at /home/takafumi/demo.jl:6
│     [3] (::Base.var"#630#631"{var"#3#5",Channel{Any}})() at ./channels.jl:129
└ @ Main ~/junk/2019/11/15-174444.jl:10
ERROR: LoadError: InvalidStateException("Channel is closed.", :closed)
Stacktrace:
 [1] try_yieldto(::typeof(Base.ensure_rescheduled), ::Base.RefValue{Task}) at ./task.jl:611
 [2] wait() at ./task.jl:668
[...]

PR #35714, opened 2020-05-03
A grab bag of latency reductions

timholy:
Yesterday I learned about `JL_DEBUG_METHOD_INVALIDATION` and have spent some time analyzing the sources of invalidation-related latencies from loading Revise. Having Revise be slow to start up is not exactly what we want for a package designed to reduce "practical" latency, so this seemed like an important case to optimize. But I expect some of these changes will have more widespread impacts.

This is organized in a series of independent commits; for details, it's probably best to look at the commit message for each one. The bottom line is that this reduces the number of lines from `julia --startup-file=no -e "using Revise" &> /tmp/invalidated.log`, when compiled with `JL_DEBUG_METHOD_INVALIDATION = 1`, by more than 200 lines.

That may sound good (and it is!), but there's an even more awesome prize waiting (EDIT: fixed below): figure out how to prevent massive invalidation due to the logging framework (CC @c42f). Revise defines its own custom logger, and this seems to invalidate basically every method throughout Base and packages that uses a `@debug`, `@info`, `@warn`, or `@error` macro. That's *really* expensive, and I'm contemplating removing all the logging from Revise just to prevent this from happening, but it would be a sad loss.

I made a guess that it was `catch_exception`, see final commit, but this was not effective. The culprit really does appear to be `shouldlog`, and I'm surprised because `ReviseLogger` ends up last on its method table so I don't understand why this invalidates so much. `ReviseLogger` ends up first on the method table for `min_enabled_level`, so maybe that's the real problem? (I honestly don't understand why it sorts that way, it would seem that having equally-specific methods insert at the end of the method table makes the most sense and that seems to be what happens most of the time.)

With these changes *and* if I disable Revise's custom logger, I go from about 1300 lines of invalidation-output to around 100. That has a very noticeable positive effect on "snappiness" when using Revise. It sure would be nice to get there!

### DIFF
--- a/base/logging.jl
+++ b/base/logging.jl
@@ -71,6 +71,15 @@ logger type.
 catch_exceptions(logger) = true
 
 
+# Prevent invalidation when packages define custom loggers
+# Using invoke in combination with @nospecialize eliminates backedges to these methods
+function _invoked_shouldlog(logger, level, _module, group, id)
+    @nospecialize
+    invoke(shouldlog, Tuple{typeof(logger), typeof(level), typeof(_module), typeof(group), typeof(id)}, logger, level, _module, group, id)
+end
+
+_invoked_min_enabled_level(@nospecialize(logger)) = invoke(min_enabled_level, Tuple{typeof(logger)}, logger)
+
 
 """
     NullLogger()
@@ -315,7 +324,7 @@ function logmsg_code(_module, file, line, level, message, exs...)
                 id = $id
                 # Second chance at an early bail-out (before computing the message),
                 # based on arbitrary logger-specific logic.
-                if shouldlog(logger, level, _module, group, id)
+                if _invoked_shouldlog(logger, level, _module, group, id)
                     file = $file
                     line = $line
                     try
@@ -374,7 +383,7 @@ struct LogState
     logger::AbstractLogger
 end
 
-LogState(logger) = LogState(LogLevel(min_enabled_level(logger)), logger)
+LogState(logger) = LogState(LogLevel(_invoked_min_enabled_level(logger)), logger)
 
 function current_logstate()
     logstate = current_task().logstate
--- a/base/logging.jl
+++ b/base/logging.jl
@@ -391,6 +391,7 @@ end
 end
 
 function with_logstate(f::Function, logstate)
+    @nospecialize
     t = current_task()
     old = t.logstate
     try
@@ -400,7 +401,6 @@ function with_logstate(f::Function, logstate)
         t.logstate = old
     end
 end
-
 
 #-------------------------------------------------------------------------------
 # Control of the current logger and early log filtering
@@ -502,7 +502,7 @@ with_logger(logger) do
 end
 ```
 """
-with_logger(f::Function, logger::AbstractLogger) = with_logstate(f, LogState(logger))
+with_logger(@nospecialize(f::Function), logger::AbstractLogger) = with_logstate(f, LogState(logger))
 
 """
     current_logger()
--- a/base/multidimensional.jl
+++ b/base/multidimensional.jl
@@ -310,6 +310,11 @@ module IteratorsMD
     convert(::Type{CartesianIndices{N,R}}, inds::CartesianIndices{N}) where {N,R} =
         CartesianIndices(convert(R, inds.indices))
 
+    # equality
+    Base.:(==)(a::CartesianIndices{N}, b::CartesianIndices{N}) where N =
+        all(map(==, a.indices, b.indices))
+    Base.:(==)(a::CartesianIndices, b::CartesianIndices) = false
+
     # AbstractArray implementation
     Base.axes(iter::CartesianIndices{N,R}) where {N,R} = map(Base.axes1, iter.indices)
     Base.IndexStyle(::Type{CartesianIndices{N,R}}) where {N,R} = IndexCartesian()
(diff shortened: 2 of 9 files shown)


timholy:
I also stole a couple of commits from the https://github.com/JuliaLang/julia/tree/kc/stale_precompile branch. (Thanks @KristofferC!)

timholy:
Build failure is just for docs (I didn't try that locally). I might wait to see if anybody has any insight about the logging issues first before I decide to tackle this further. Sure would be nice to see this in 1.5, of course, so if needed I can back that last commit out and do surgery on Revise.

timholy:
Forgot there was an important companion I needed to push: https://github.com/JuliaLang/Pkg.jl/pull/1806

timholy:
OK, so the difference in method sorting mentioned above seems to be that `min_enabled_level` gets classified `isleafsig` and the rest do not. I do not think that `min_enabled_level` is a big contributor to the problem. The bigger issue is that prior to insertion `shouldlog` has 91 direct backedges and many more indirect backedges. Curiously, `handle_message` (which appears in the [same block of code](https://github.com/JuliaLang/julia/blob/cf3d8b584a44d48b0dc66db9ab37f871f63a21c7/base/logging.jl#L318-L327)) does not have any backedges.

My guess is that this occurs because `shouldlog` is simple and gets inlined via union-splitting. I'll try rebuilding with all such methods annotated `@noinline` and see what happens.

base/generator.jl
```diff
@@ -89,13 +89,19 @@ Base.HasLength()
 """
 IteratorSize(x) = IteratorSize(typeof(x))
 IteratorSize(::Type) = HasLength()  # HasLength is the default
+# Many compiled methods iterate on tuples, but MethodInstances for
+#    IteratorSize(::Type{Tuple{...}})
+# are easily invalidated by method-table insertions because they call
+# the fallback. Calling iteratorsize instead of IteratorSize bypasses
+# this problem. Don't add methods to iteratorsize.
+iteratorsize(::Type{T}) where {T} = T<:Tuple ? HasLength() : IteratorSize(T)
```
JeffBezanson, line 97: Would it work just to add an `IteratorSize(::Type{<:Tuple})` method?

base/generator.jl
```diff
@@ -126,3 +132,9 @@ IteratorEltype(::Type) = HasEltype()  # HasEltype is the default
 IteratorEltype(::Type{Generator{I,T}}) where {I,T} = EltypeUnknown()
 
 IteratorEltype(::Type{Any}) = EltypeUnknown()
+
+# See comment for iteratorsize re invalidation
+function iteratoreltype(::Type{T}) where T
+    T<:(Tuple{Vararg{X}} where X) ? HasEltype() :
+        T<:Tuple ? EltypeUnknown() : IteratorEltype(T)
```
JeffBezanson, line 139: This definition makes sense, but currently we always give `HasEltype` for tuples (for better or worse).

base/logging.jl
```diff
@@ -502,7 +513,7 @@ with_logger(logger) do
 end
 ```
 """
-with_logger(f::Function, logger::AbstractLogger) = with_logstate(f, LogState(logger))
+with_logger(@nospecialize(f::Function), logger::AbstractLogger) = with_logstate(f, LogState(logger))
```
JeffBezanson, line 516: I would not expect this to make a difference; it doesn't affect inference and we will already avoid specializing this function on `f`.

base/multidimensional.jl
```diff
@@ -937,56 +937,6 @@ function fill!(A::AbstractArray{T}, x) where T
...
- 3.0
- 0.0
- 5.0
- 0.0
- 0.0
-```
-"""
-copyto!(dest, src)
-
-function copyto!(dest::AbstractArray{T1,N}, src::AbstractArray{T2,N}) where {T1,T2,N}
```
JeffBezanson, line 969: Is this method really not needed?

base/generator.jl
timholy: You would know much better than me, but what I was concerned about was that it would be far down on the method list and therefore easily invalidated. But maybe order isn't as important as I think it may be?

base/logging.jl
timholy: Yeah, it doesn't seem to change anything.

base/multidimensional.jl
timholy: I'm not actually sure, for now this is just a guess. Have not yet run tests or benchmarks, to me that will be the proof of the pudding.

timholy:
OK, I seem to have discovered an exit on the logging framework: the combination of `@nospecialize` and `invoke` seems to force runtime dispatch. For maximum effect I also need to modify [this line in Pkg](https://github.com/JuliaLang/Pkg.jl/blob/f84d8117b4ce1c03b2cffda4a78d38340700c646/src/PlatformEngines.jl#L129) to call `_invoked_min_enabled_level`, but I'll wait until I know whether this is a good idea before pushing that. To see the cumulative impact, see [where we started](https://gist.github.com/ianshmean/2825c2f9e118217c7855245e508c4f21#file-revise-invalidations-during-load) and [where this ended up](https://gist.github.com/timholy/c0ad9d30de098fed5e65255f6409e964): an invalidation log that went from 1384 lines down to 88.

I've not yet run timing comparisons on anything meaningful to see what this actually adds up to in practice; it's entirely possible that the improvement will be modest. (It may be complicated because it will also depend on what other packages get loaded.) It's also quite possible this will cause dreadful runtime regressions. If this passes tests, let's run our benchmarks and find out---to me it seems that the experiment is well worth doing.

timholy:
OK, this is cool. `using Revise` by itself is basically no faster. But while working on this I noticed that a huge number of `Pkg` methods get invalidated, and so that prompted another way to look for possible benefits. On 1.4, if I say `using Revise` followed by `(@v1.4) pkg> up` there's a lag of more than 4 seconds before I see any text output. On this, it's instantaneous. Woot!

c42f:
I've been vaguely wondering about how backedges interact with the logging system for a while. I have noticed that installing a new logger causes a lot more recompilation than you might expect or want, but I sure didn't realize that `shouldlog` was inlined via union splitting. That's both amazing and terrifying! Amazing because it means we could have better performance than I'd imagined, and terrifying in the amount of method invalidation it could cause.

It's also known that setting the global logger doesn't really work well with the world age - see https://github.com/JuliaLang/julia/issues/33865 . That would suggest, perhaps, we should have an `invokelatest` barrier somewhere, but I'm not exactly sure where yet.

So anyway, I'm cautiously positive about just breaking some backedges here. The logging system needs some other refactoring to fix (runtime) performance issues as well.

base/multidimensional.jl
timholy: Here's a good one: on master,
```julia
julia> a = [1 2; 3 4]
2×2 Array{Int64,2}:
 1  2
 3  4

julia> b = [3, 1, 4, 2]
4-element Array{Int64,1}:
 3
 1
 4
 2

julia> copyto!(view(a, b), a)
4-element view(::Array{Int64,1}, [3, 1, 4, 2]) with eltype Int64:
 1
 3
 1
 1
```
vs
```julia
julia> copyto!(view(a, b), copy(a))
4-element view(::Array{Int64,1}, [3, 1, 4, 2]) with eltype Int64:
 1
 3
 2
 4
```
I'm hoping to fix it in this PR, or will report it if not.

timholy:

base/generator.jl
timholy: Perhaps because of https://github.com/JuliaLang/Pkg.jl/pull/1806 I can't see any invalidation from this anymore. I've backed this out, we can re-add it if necessary.

nanosoldier:
[Your benchmark job](https://github.com/JuliaLang/julia/pull/35714#issuecomment-623430795) has completed - possible performance regressions were detected. A full report can be found [here](https://github.com/JuliaCI/BaseBenchmarkReports/blob/d796ef79a2feceb65b322d83a75c58050ab4d8e6/4ea6275_vs_f1d10e7/report.md). cc @ararslan

timholy:
Well, that's rather unremarkable (which is what I was hoping for). @JeffBezanson, let me know if there's anything objectionable here before merging.

JeffBezanson (approved):

c42f:
> Well, that's rather unremarkable

Well I doubt we have any logging performance tests ;-) But anyway I feel like this is probably a good usability tradeoff for now.

In the future we can do some proper performance benchmarking and optimization of the logging system.

PR #35715, opened 2020-05-03
Document what public API means

### LINKED ISSUE #7561: document conventions about public vs private API
In the discussion of #1974 there seems to be an implicit consensus that accessing information via the `a.b` syntax is part of the "private" API, and the "public" API consists of calls like `b(a)`.  The canonical example seems to be that if `a` is a `FloatRange` object `a.step` is not a useful value for public purposes, where `step(a)` is. 

Depending on how that issue is resolved the convention may change somewhat, but in either case there will be some preferred convention.  That convention should be documented, and so far is not as far as I could tell. Perhaps the Types section of the manual makes sense?

tkf:
I think there is confusion in Julia community for what it means to be a public API.  I think clarifying it in documentation makes sense.

Closes #7561.

### DIFF
--- a/doc/src/manual/faq.md
+++ b/doc/src/manual/faq.md
@@ -18,6 +18,28 @@ For similar reasons, automated translation to Julia would also typically generat
 
 On the other hand, language *interoperability* is extremely useful: we want to exploit existing high-quality code in other languages from Julia (and vice versa)!  The best way to enable this is not a transpiler, but rather via easy inter-language calling facilities.  We have worked hard on this, from the built-in `ccall` intrinsic (to call C and Fortran libraries) to [JuliaInterop](https://github.com/JuliaInterop) packages that connect Julia to Python, Matlab, C++, and more.
 
+## [Public API](@id man-api)
+
+### How does Julia declare public API?
+
+The only interfaces that are stable with respect to [SemVer](https://semver.org/) of `julia`
+version are the Julia `Base` and standard libraries interfaces described in
+[the documentation](https://docs.julialang.org/) and not explicitly marked as non-stable.
+Functions, types, and constants are not part of the public API if they are not included in
+the documentation, _even if they have docstrings_.
+
+### There is a useful undocumented function/type/constant. Can I use it?
+
+Please open an [issue](https://github.com/JuliaLang/julia/issues) or
+[pull request](https://github.com/JuliaLang/julia/pulls) to start a discussion for turning it
+into a public API before relying on it.
+
+### The documentation is not accurate enough. Can I rely on the existing behavior?
+
+Please open an [issue](https://github.com/JuliaLang/julia/issues) or
+[pull request](https://github.com/JuliaLang/julia/pulls) to start a discussion for turning the
+existing behavior into a public API.
+
 ## Sessions and the REPL
 
 ### How do I delete an object in memory?
--- a/doc/src/manual/faq.md
+++ b/doc/src/manual/faq.md
@@ -24,9 +24,9 @@ On the other hand, language *interoperability* is extremely useful: we want to e
 
 The only interfaces that are stable with respect to [SemVer](https://semver.org/) of `julia`
 version are the Julia `Base` and standard libraries interfaces described in
-[the documentation](https://docs.julialang.org/) and not explicitly marked as non-stable.
-Functions, types, and constants are not part of the public API if they are not included in
-the documentation, _even if they have docstrings_.
+[the documentation](https://docs.julialang.org/) and not marked as unstable (e.g.,
+experimental and internal).  Functions, types, and constants are not part of the public
+API if they are not included in the documentation, _even if they have docstrings_.
 
 ### There is a useful undocumented function/type/constant. Can I use it?
 
--- a/doc/src/manual/faq.md
+++ b/doc/src/manual/faq.md
@@ -30,9 +30,14 @@ API if they are not included in the documentation, _even if they have docstrings
 
 ### There is a useful undocumented function/type/constant. Can I use it?
 
-Please open an [issue](https://github.com/JuliaLang/julia/issues) or
+Updating Julia may break your code if you use non-public API.  If the code is
+self-contained, it may be a good idea to copy it into your project.  If you want to rely on
+a complex non-public API, especially when using it from a stable package, it is a good idea
+to open an [issue](https://github.com/JuliaLang/julia/issues) or
 [pull request](https://github.com/JuliaLang/julia/pulls) to start a discussion for turning it
-into a public API before relying on it.
+into a public API.  However, we do not discourage the attempt to create packages that expose
+stable public interfaces while relying on non-public implementation details of `julia` and
+buffering the differences across different `julia` versions.
 
 ### The documentation is not accurate enough. Can I rely on the existing behavior?
 
--- a/doc/src/manual/faq.md
+++ b/doc/src/manual/faq.md
@@ -20,7 +20,7 @@ On the other hand, language *interoperability* is extremely useful: we want to e
 
 ## [Public API](@id man-api)
 
-### How does Julia declare public API?
+### How does Julia define its public API?
 
 The only interfaces that are stable with respect to [SemVer](https://semver.org/) of `julia`
 version are the Julia `Base` and standard libraries interfaces described in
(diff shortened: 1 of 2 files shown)


stev47:
Sometimes the manual mentions "internal" methods (e.g. methods starting with underscore in the [chapter on methods](https://github.com/JuliaLang/julia/blob/master/doc/src/manual/methods.md#dispatch-on-one-argument-at-a-time)). Maybe one could clarify that apparently "non-internal" does not mean "public".

JeffBezanson:
:+1: to adding this. I'm sure @StefanKarpinski will want to weigh in.

StefanKarpinski:
This feels a little too restrictive. People can rely on internal code if they want to, but they just need to be aware that we may change it and break their code. If they're ok with that, then carry on. But if they find it useful then someone else may as well, so it would be helpful if they would open an issue to figure out a way to expose the functionality in some official way.

doc/src/manual/faq.md
```diff
@@ -18,6 +18,28 @@ For similar reasons, automated translation to Julia would also typically generat
...
+version are the Julia `Base` and standard libraries interfaces described in
+[the documentation](https://docs.julialang.org/) and not explicitly marked as non-stable.
+Functions, types, and constants are not part of the public API if they are not included in
+the documentation, _even if they have docstrings_.
+
+### There is a useful undocumented function/type/constant. Can I use it?
+
+Please open an [issue](https://github.com/JuliaLang/julia/issues) or
+[pull request](https://github.com/JuliaLang/julia/pulls) to start a discussion for turning it
+into a public API before relying on it.
```
tkf, line 35: @StefanKarpinski Do you mean this part?  What do you think about rephrasing it to something like this?

```suggestion
Updating Julia may break your code if you use non-public API.  If the code is
self-contained, it may be a good idea to copy it into your project.  If you want to rely on
a complex non-public API, especially when using it from a stable package, it is a good idea
to open an [issue](https://github.com/JuliaLang/julia/issues) or
[pull request](https://github.com/JuliaLang/julia/pulls) to start a discussion for turning it
into a public API.
```

doc/src/manual/faq.md
```diff
@@ -18,6 +18,28 @@ For similar reasons, automated translation to Julia would also typically generat
...
 
+## [Public API](@id man-api)
+
+### How does Julia declare public API?
+
+The only interfaces that are stable with respect to [SemVer](https://semver.org/) of `julia`
+version are the Julia `Base` and standard libraries interfaces described in
+[the documentation](https://docs.julialang.org/) and not explicitly marked as non-stable.
+Functions, types, and constants are not part of the public API if they are not included in
+the documentation, _even if they have docstrings_.
```
tkf, line 29: @stev47 I think using the word "non-internal" to mean "public" is fine.  I'd just say that functions that are not prefixed by `_` are not necessarily public.  But what I've written already implies that

https://github.com/JuliaLang/julia/blob/85752db3461153fffcdd522ebe7ee904e42898e5/doc/src/base/base.md#L18-L20

Is it unclear from this description?

Or maybe adding this in the FAQ helps?

```suggestion
Functions, types, and constants are not part of the public API if they are not included in
the documentation, _even if they have docstrings or not starting with `_`_.
```

stev47: Yes, maybe "... , irrespective of their docstring and irrespective of whether their names are marked as internal with a leading underscore (`_foo`) or not"

On a sidenote: will there be an official API to query whether a name is considered public or not?

tkf: I'd say "internal" is exactly the complement of "public". So "marked as internal with a leading underscore" does not make much sense; _not documenting as stable API_ is the only way to mark anything internal.

Starting with `_` is not really anything to do with public API or not.  It's just that it's used commonly for messaging "you _really_ don't want to touch this" unless documented otherwise.

But it's totally OK to use it in public API (although I guess it's not used in Base/stdlib, other than `__init__`?).  But you can find the use case elsewhere.  For example, [`OnlineStatsBase._fit!`](https://github.com/joshday/OnlineStatsBase.jl) is a public API. As you can see, these function names tend to be used as "mostly overload only" API; but they are still public.

stev47:
> I'd say "internal" is exactly the complement of "public".

If so, could you then change the corresponding wording about "internal methods"
in the manual section about methods?

> Starting with `_` is not really anything to do with public API or not.

It is assumed to be an internal marker in a lot of other programming languages
(and in some places of Julia's manual), so some clarification should be in place.

> For example,
> [`OnlineStatsBase._fit!`](https://github.com/joshday/OnlineStatsBase.jl) is a
> public API. As you can see, these function names tend to be used as "mostly
> overload only" API; but they are still public.

So looking from the other side: exported functions that are not documented are
not considered public? It will take a time for me to get used to that.

tkf:
> > I'd say "internal" is exactly the complement of "public".
>
> If so, could you then change the corresponding wording about "internal methods" in the manual section about methods?

I'm not sure what is inconsistent with the part you linked.  If you think you have a better wording in that part, the best way is to open an PR.  It's touching a different part of the documentation so it can be discussed independently.

> It is assumed to be an internal marker in a lot of other programming languages

Yes, it is _"assumed"_ to be; but not always true. In a lot of programming languages it is used in public interface when it's doing something tricky ([an example in Python](https://docs.python.org/3/library/collections.html#collections.somenamedtuple._make); [an example in C](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html#g_t_005f_005fatomic-Builtins)).

> So looking from the other side: exported functions that are not documented are not considered public?

You have to declare public API to use SemVer. That's the very first thing mentioned in [SemVer spec](https://semver.org/#spec-item-1). Writing documentation is more or less the only way to declare public API in Julia at the moment (it'd be nice to improve this, of course). `export` alone does not mean anything because it doesn't tell you what interface the exported names satisfy.

Obviously, I'm only describing an ideal case.  In practice, you may be able to rely on kindness of the library authors to avoid abrupt change in not-strictly-stable API.

stev47:
> `export` alone does not mean anything because it doesn't tell you what
> interface the exported names satisfy.

Thank you, that makes sense.

JeffBezanson:
> Writing documentation is more or less the only way to declare public API in Julia at the moment (it'd be nice to improve this, of course). `export` alone does not mean anything

I don't think this is entirely true; anything exported should be considered public, and it's considered a bug for an exported identifier not to have documentation. But even if an exported name didn't have documentation, it would at least need to remain exported and do something like what it did before to avoid breaking code.

tkf:
Yes, I agree that's the best action once undocumented `export` names are found.

doc/src/manual/faq.md
```diff
@@ -18,6 +18,28 @@ For similar reasons, automated translation to Julia would also typically generat
 
 On the other hand, language *interoperability* is extremely useful: we want to exploit existing high-quality code in other languages from Julia (and vice versa)!  The best way to enable this is not a transpiler, but rather via easy inter-language calling facilities.  We have worked hard on this, from the built-in `ccall` intrinsic (to call C and Fortran libraries) to [JuliaInterop](https://github.com/JuliaInterop) packages that connect Julia to Python, Matlab, C++, and more.
 
+## [Public API](@id man-api)
+
+### How does Julia declare public API?
```
stevengj, line 23: "How does Julie define its public API?"

mbauman:
Fun fact: merging this would close the least-recently-updated open issue (https://github.com/JuliaLang/julia/issues/7561).

doc/src/manual/faq.md
tkf: Maybe also add something like

> However, we do not discourage the attempt to create packages that expose stable public interfaces while relying on non-public implementation details of `julia` and buffering the differences across different `julia` versions.

?

c42f: Yes I think something like this would be a reasonable improvement.

Often it's unavoidable for packages to rely on implementation detail of `Base` if they want to make progress *right now* rather than in some fantasy future version of Julia. I think it's great to allow people with high risk tolerance to experiment in this way.

We just need to be extremely clear that such code will rapidly cease to work in the absence of a maintainer. (Hmm, interesting feature would be to automatically report on the private bindings a package uses as a measure of maintainability - to go along with things such as testing, code coverage and documentation. Too bad this isn't formally expressed in the code in some way.)

ViralBShah:
Bump. This would be nice to get merged and feels like we are close.

vtjnash:
I can't see any reason not to have merged this.

PR #33357, opened 2019-09-22
Add a `@ref` to the Main docstring

asinghvi17:

PR #35722, opened 2020-05-04
Revert "Add a `@ref` to the Main docstring"

Keno:
Reverts JuliaLang/julia#33357. Broke the build.


Keno:
Gonna merge this without waiting for CI. It's already broken, so it can't get any worse and maybe this way we avoid some ripple on effects from having a broken master.

PR #35723, opened 2020-05-04
Document how to create an undefined reference (Ref) and detect that using isassigned

mkitti:
We can create a `Ref` to a `undef` and figure out if what it points to is undefined as follows:
```
julia> struct Blah
           A
       end

julia> ref = Ref{Blah}()
Base.RefValue{Blah}(#undef)

julia> ref[]
Stacktrace:
 [1] getproperty at .\Base.jl:33 [inlined]
 [2] getindex(::Base.RefValue{Blah}) at .\refvalue.jl:32
 [3] top-level scope at REPL[16]:1

julia> isassigned(ref)
false

julia> ref[] = Blah(5)
Blah(5)

julia> isassigned(ref)
true
```

However, for primitive types the above does not work:

```
julia> ref = Ref{Int64}()
Base.RefValue{Int64}(140169022152048)

julia> ref[]
140169022152048

julia> isassigned(ref)
true
```

Because this is not intuitive, I propose to document this behavior.

Edited 7/15/2020: Update main comment since we document `isassigned` rather than `isdefined` for this purpose.


base/refpointer.jl
```diff
@@ -13,6 +13,11 @@ Creation of a `Ref` to a value `x` of type `T` is usually written `Ref(x)`.
 Additionally, for creating interior pointers to containers (such as Array or Ptr),
 it can be written `Ref(a, i)` for creating a reference to the `i`-th element of `a`.
 
+A `Ref` of a non-primitive type `T` referring to `undef` can be made by
```
yuyichao, line 16: Non bitstype.

base/refpointer.jl
```diff
@@ -13,6 +13,11 @@ Creation of a `Ref` to a value `x` of type `T` is usually written `Ref(x)`.
...
+`Ref{T}()`.  To check if a `Ref` refers to undef, use `isdefined(ref,1)`.
```
yuyichao, line 17: `isassigned`.

yuyichao (changes requested):

base/refpointer.jl
mkitti: Changed in 28f8806

base/refpointer.jl
mkitti: Changed in 28f8806

base/essentials.jl
```diff
@@ -623,6 +623,10 @@ getindex(v::SimpleVector, I::AbstractArray) = Core.svec(Any[ v[i] for i in I ]..
 Test whether the given array has a value associated with index `i`. Return `false`
 if the index is out of bounds, or has an undefined reference.
 
+    isassigned(ref) -> Bool
+
+Test whether the given [`Ref`](@ref) has been associated with a value. This is always true for a [`Ref`](@ref) for a bitstype. Return `false` if the reference is undefined.
+
```
mkitti, line 629: Added `isassigned(ref)` documentation with cross reference to `Ref`

base/essentials.jl
```diff
@@ -641,6 +645,18 @@ julia> v = similar(rand(3), Foo)
...
+Base.RefValue{Function}(#undef)
+
+julia> ref[] = max
+max (generic function with 12 methods)
+
+julia> isassigned(ref)
+true
+
+julia> isassigned(Ref{Int}())
+true
```
mkitti, line 659: Added examples of `isassigned` for `Ref`

base/essentials.jl
yuyichao: I think this (and the example) should add to the docstring of `isassigned(::RefValue)` instead. You can use `methods(isassigned)` (or equivalent) to find where it is.

base/refpointer.jl
```diff
@@ -13,6 +13,11 @@ Creation of a `Ref` to a value `x` of type `T` is usually written `Ref(x)`.
 Additionally, for creating interior pointers to containers (such as Array or Ptr),
 it can be written `Ref(a, i)` for creating a reference to the `i`-th element of `a`.
 
+A `Ref` of a non-bitstype `T` referring to `undef` can be made by `Ref{T}()`.
+To check if a `Ref` refers to undef, use `isassigned(ref)`.
+For example, `isassigned(Ref{T}())` is `false` if T is a struct. This does not
+apply to primitive types. `isassigned(Ref{Int64}())` results in `true`.
+
 When passed as a `ccall` argument (either as a `Ptr` or `Ref` type), a `Ref` object will be
 converted to a native pointer to the data it references.
```
yuyichao, line 23: This should probably be merged with the paragraph about `NULL` right below. The mention of `NULL` below is also wrong for exactly the case mentioned above...

mkitti: I removed the no invalid reference language in 2849f95

base/essentials.jl
mkitti: Done in 2849f95

mkitti:
I have made the changes suggested. With the phrase about no invalid `Refs` existing removed, the remaining sentence appears to be more about the use of `C_NULL` with `ccall`

mkitti:
This just came up in Slack helpdesk again: https://julialang.slack.com/archives/C6A044SQH/p1594629154297600

@yuyichao What is needed to be done on this PR?

yuyichao:
The duplicate should be removed from the docstring for `isassigned` for arrays.

base/refpointer.jl
```diff
@@ -13,11 +13,20 @@ Creation of a `Ref` to a value `x` of type `T` is usually written `Ref(x)`.
...
-converted to a native pointer to the data it references.
+`Ref{T}()` creates a reference to a value of type `T` without initialization.
+For a bitstype `T`, the value will be whatever currently resides in the memory
+allocated. For a non-bitstype `T`, the value will be `undef` and will result in
+an error, "UndefRefError: access to undefined reference" if the Ref is
+deferenced.
 
-There is no invalid (NULL) `Ref` in Julia, but a `C_NULL` instance of `Ptr` can be passed to
-a `ccall` Ref argument.
+To check if a `Ref` refers to [`undef`](@ref), use [`isassigned(ref::RefValue)`](@ref).
```
yuyichao, line 22: I highly suspect that this will link to `undef`, which is wrong. It's a missing reference, it's not refering to any object. A reference to `undef` is `Ref(undef)`.

mkitti: I just say "undefined reference" now in 94edfaa

yuyichao (approved):
There are some super long lunges but lgtm otherwise.

mkitti:
I fixed "super long lunges" in f9fbfa2

base/refpointer.jl
```diff
@@ -13,11 +13,20 @@ Creation of a `Ref` to a value `x` of type `T` is usually written `Ref(x)`.
 Additionally, for creating interior pointers to containers (such as Array or Ptr),
 it can be written `Ref(a, i)` for creating a reference to the `i`-th element of `a`.
 
-When passed as a `ccall` argument (either as a `Ptr` or `Ref` type), a `Ref` object will be
-converted to a native pointer to the data it references.
+`Ref{T}()` creates a reference to a value of type `T` without initialization.
+For a bitstype `T`, the value will be whatever currently resides in the memory
+allocated. For a non-bitstype `T`, the value will be `undef` and will result in
```
rfourquet, line 18: This seems confusing, `undef` represents a well-defined singleton value in Julia. I would just write "undefined". Also, it's not the value which will result in an error, but the act of dereferencing.

base/refpointer.jl
```diff
@@ -13,11 +13,20 @@ Creation of a `Ref` to a value `x` of type `T` is usually written `Ref(x)`.
...
+an error, "UndefRefError: access to undefined reference" if the Ref is
+deferenced.
 
-There is no invalid (NULL) `Ref` in Julia, but a `C_NULL` instance of `Ptr` can be passed to
-a `ccall` Ref argument.
+To check if a `Ref` is an undefined reference, use [`isassigned(ref::RefValue)`](@ref).
+For example, `isassigned(Ref{T}())` is `false` if T is not a bitstype struct.
```
rfourquet, line 23:
```suggestion
For example, `isassigned(Ref{T}())` is `false` if `T` is not a bitstype.
```

base/refpointer.jl
```diff
@@ -13,11 +13,20 @@ Creation of a `Ref` to a value `x` of type `T` is usually written `Ref(x)`.
...
+If T is a bitstype, `isassigned(Ref{T}())` will always be true.
```
rfourquet, line 24:
```suggestion
If `T` is a bitstype, `isassigned(Ref{T}())` will always be true.
```

base/refpointer.jl
```diff
@@ -13,11 +13,20 @@ Creation of a `Ref` to a value `x` of type `T` is usually written `Ref(x)`.
...
+
+When passed as a `ccall` argument (either as a `Ptr` or `Ref` type), a `Ref`
+object will be converted to a native pointer to the data it references.
+
+A `C_NULL` instance of `Ptr` can be passed to a `ccall` Ref argument to initialize it.
```
rfourquet, line 29:
```suggestion
A `C_NULL` instance of `Ptr` can be passed to a `ccall` `Ref` argument to initialize it.
```

base/refvalue.jl
```diff
@@ -8,6 +8,31 @@ mutable struct RefValue{T} <: Ref{T}
     RefValue{T}(x) where {T} = new(x)
 end
 RefValue(x::T) where {T} = RefValue{T}(x)
+"""
+    isassigned(ref::RefValue) -> Bool
+
+Test whether the given [`Ref`](@ref) has been associated with a value.
+This is always true for a [`Ref`](@ref) for a bitstype.
```
rfourquet, line 15: (to avoid "for" repetition... feel free to ignore)
```suggestion
This is always true for a [`Ref`](@ref) of a bitstype object.
```

base/refvalue.jl
```diff
@@ -8,6 +8,31 @@ mutable struct RefValue{T} <: Ref{T}
     RefValue{T}(x) where {T} = new(x)
 end
 RefValue(x::T) where {T} = RefValue{T}(x)
+"""
+    isassigned(ref::RefValue) -> Bool
+
+Test whether the given [`Ref`](@ref) has been associated with a value.
```
rfourquet, line 14:
```suggestion
Test whether the given [`Ref`](@ref) is associated with a value.
```

base/refpointer.jl
mkitti: I corrected this in a49c8cc

base/refpointer.jl
```diff
@@ -28,6 +37,18 @@ julia> isa.(Ref([1,2,3]), [Array, Dict, Int])
...
+julia> Ref{Function}()
+Base.RefValue{Function}(#undef)
+
+julia> try Ref{Function}()[] catch e; println(e); end
+UndefRefError()
+
+julia> isassigned(Ref{Int64}())
+true
+
+julia> Ref(C_NULL)
```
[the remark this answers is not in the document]
mkitti: This was supposed to show "normal" usage of `Ref` with a value, but it ends up being more confusing. I created an example just using `Ref(5)` and put it first.

mkitti:
Thank you for all of the reviews. Are there any other pending issues to resolve, @KristofferC or @rfourquet?

base/refpointer.jl
rfourquet: Sorry for the delay. If I'm not mistaken, my second point was not addressed. That is, the reference will not "result in an error"; trying to dereference it will result in an error.

mkitti: I revised the text in 2b9a7a7 to clarify that attempted dereferencing will cause an error. Thanks @rfourquet

rfourquet (approved):
LGTM, modulo maybe this test which might fail sometimes.

base/refpointer.jl
```diff
@@ -13,21 +13,53 @@ Creation of a `Ref` to a value `x` of type `T` is usually written `Ref(x)`.
...
+Base.RefValue{Function}(#undef)
+
+julia> try
+           Ref{Function}()[] # Dereferencing an undefined reference will result in an error
+       catch e
+           println(e)
+       end
+UndefRefError()
+
+julia> Ref{Int64}()[] == 0 # A reference to a bitstype refers to an undetermined value if not given
```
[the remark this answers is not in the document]
mkitti: Instead of doing the comparison, I just added a semicolon in 08713b2. We'll just demonstrate the lack of an error and not the undetermined output.

doc/src/base/c.md
```diff
@@ -23,6 +23,7 @@ Base.systemerror
 Base.windowserror
 Core.Ptr
 Core.Ref
+Base.isassigned(::Base.RefValue)
```
mkitti, line 26: I added this to `c.md` next to where `Ref` is introduced.. Is this the right place to insert this in the docs? Do we need additional docs on `Base.RefValue` ?

mkitti:
Hello everyone. I'm relocating over the next week or so and will be unavailable.

Please let me know if there is anything more to do on this doc PR. Thanks.

mkitti:
I'm now back online.

mkitti:
@KristofferC, is there anything left to do on this doc PR?

mkitti:
Hello. Just checking in on this documentation-only PR. @KristofferC was there anything you wanted to revise on this doc PR?

Issue #35721, opened 2020-05-04
DelimitedFiles readdlm is externally slow for Complex{Float64} 


yzcj105:
```
 using DelimitedFiles
A = randn(Complex{Float64},300,30)
B = randn(Float64,300,30)
writedlm("A.csv",A,',')
writedlm("B.csv",B,',')
 @time readdlm("A.csv",',',Complex{Float64},'\n')
 10.645386 seconds (67.52 k allocations: 3.250 GiB, 2.68% gc time)
@time readdlm("B.csv",',',Float64,'\n')
  0.005304 seconds (18.00 k allocations: 878.875 KiB)
``` 
The readdlm for Complex{Float64} is much much slower and memory requiring compared to real numbers.  
I've also compared the time consuming of pandas.read_csv in python,  pandas is very fast every for complex numbers.

yzcj105:
For Complex 300*300 matrix, the result has never come out after waiting for minutes.

yzcj105:
alternative ways to read complex matrix from csv files
A is 300*300 matrix.
```
julia> @time df = CSV.read("A.csv";header=0,delim=',')
  0.006457 seconds (22.12 k allocations: 3.413 MiB)
# return 300×300 DataFrames.DataFrame. Omitted printing of 298 columns
....
....
 A = convert(Matrix,df) .|> x->parse(Complex{Float64},x)
# return. 300×300 Array{Complex{Float64},2}:
```
Which is much much faster compared to readdlm

stevengj:
Should be fixed by #35742.

yzcj105:
Thank you!

KristofferC:
Let's not close until that merges (it will close automatically)

PR #35727, opened 2020-05-04
improve performance of complex number parsing

KristofferC:
```jl
using DelimitedFiles
A = randn(Complex{Float64},300,50)
writedlm("A.csv",A,',')
@time readdlm("A.csv",',',Complex{Float64},'\n')
```

Before:

```
  9.316321 seconds (293.64 k allocations: 3.262 GiB, 3.35% gc time)
```

After

```
  0.558150 seconds (18.06 k allocations: 1.130 MiB)
```

It is still very slow but hey.

Ref https://github.com/JuliaLang/julia/issues/35721

The super slowness came from the generic `findprev` which needs to allocate a brand new string to for the reversal:

https://github.com/JuliaLang/julia/blob/bb2aa52abd72ee64ad6a41e36959fc62560b37ac/base/strings/search.jl#L346

### DIFF
--- a/base/parse.jl
+++ b/base/parse.jl
@@ -309,7 +309,17 @@ function tryparse_internal(::Type{Complex{T}}, s::Union{String,SubString{String}
     end
 
     # find trailing im/i/j
-    iᵢ = something(findprev(in(('m','i','j')), s, e), 0)
+    iᵢ = 0
+    # findprev with a predicate function is quite slow,
+    # use three separate calls to findprev with a char instead
+    for c in ('j', 'i', 'm')
+        local i = findprev(c, s, e)
+        if i !== nothing
+            iᵢ = i
+            break
+        end
+    end
+
     if iᵢ > 0 && s[iᵢ] == 'm' # im
         iᵢ -= 1
         if s[iᵢ] != 'i'


stevengj:
Wouldn't it be better to just speed up `findprev` for strings with a predicate function?  Isn't the implementation just a loop?

stevengj:
For that matter, even the `findnext` implementatino for `AbstractString` seems to leave much to be desired (it creates a `SubString` object and a `pairs` iterator).   Seems like it should be just:

```jl
function findnext(testf::Function, s::AbstractString, i::Integer)
    z = ncodeunits(s) + 1
    1 ≤ i ≤ z || throw(BoundsError(s, i))
    @inbounds i == z || isvalid(s, i) || string_index_err(s, i)
    e = lastindex(s)
    @inbounds while i < e
        testf(s[i]) && return i
        i = nextind(s, i)
    end
    return nothing
end


function findprev(testf::Function, s::AbstractString, i::Integer)
    z = ncodeunits(s) + 1
    0 ≤ i ≤ z || throw(BoundsError(s, i))
    @inbounds i == 0 || i == z || isvalid(s, i) || string_index_err(s, i)
    @inbounds while i > 1
        testf(s[i]) && return i
        i = prevind(s, i)
    end
    return nothing
end
```

stevengj:
In fact, a lot of the code in `base/strings/search.jl` seems like it over-relies on iterator tricks rather than just looping with `nextind/prevind`.  Can't we always assume an `AbstractString` is indexable? cc @Keno.

JeffBezanson:
> (it creates a `SubString` object and a `pairs` iterator rather than just looping directly)

It would be interesting to know if this got faster since we can now stack allocate those objects.

KristofferC:
> Seems like it should be just:

Mission accomplished ;)

PR #35733, opened 2020-05-04
Eliminate most invalidations from loading FixedPointNumbers

timholy:
On master (or really, on #35714), loading FixedPointNumbers ("FPN") produces a `jl_debug_method_invalidation` log of nearly 600 lines. In conjunction with a couple of changes to FPN (below), this reduces it to 43, of which only 17 are actual `MethodInstance`s. Most of the remaining ones derive from invalidating `convert(Type{T<:Int64}, Int64) where {T<:Int64}`, which gets invalidated by
```julia
(::Type{X})(x::Number) where {X <: FixedPoint} = _convert(X, x)
```
I have not been successful in figuring out how to head that off short of `@eval`ing dedicated methods like `convert(::Type{Int64}, x::Int64) = x`.

Here's the FPN diff (EDIT: new and improved):

```diff
diff --git a/src/FixedPointNumbers.jl b/src/FixedPointNumbers.jl
index 160c133..3c3bd33 100644
--- a/src/FixedPointNumbers.jl
+++ b/src/FixedPointNumbers.jl
@@ -48,7 +48,8 @@ rawtype(::Type{X}) where {T, X <: FixedPoint{T}} = T
 *(x::Real, ::Type{X}) where {X <: FixedPoint} = _convert(X, x)
 
 # constructor-style conversions
-(::Type{X})(x::Real) where {X <: FixedPoint} = _convert(X, x)
+(::Type{X})(x::X) where {X <: FixedPoint}      = x
+(::Type{X})(x::Number) where {X <: FixedPoint} = _convert(X, x)
 
 function (::Type{<:FixedPoint})(x::AbstractChar)
     throw(ArgumentError("FixedPoint (Fixed or Normed) cannot be constructed from a Char"))
@@ -97,7 +98,6 @@ one(::Type{X}) where {X <: FixedPoint} = oneunit(X)
 inv_rawone(x) = (@generated) ? (y = 1.0 / rawone(x); :($y)) : 1.0 / rawone(x)
 
 # traits
-sizeof(::Type{X}) where {X <: FixedPoint} = sizeof(rawtype(X))
 eps(::Type{X}) where {X <: FixedPoint} = X(oneunit(rawtype(X)), 0)
 typemax(::Type{T}) where {T <: FixedPoint} = T(typemax(rawtype(T)), 0)
 typemin(::Type{T}) where {T <: FixedPoint} = T(typemin(rawtype(T)), 0)
```

### DIFF
--- a/base/essentials.jl
+++ b/base/essentials.jl
@@ -174,6 +174,8 @@ convert(::Type{Type}, x::Type) = x # the ssair optimizer is strongly dependent o
                                    # in the absence of inlining-enabled
                                    # (due to fields typed as `Type`, which is generally a bad idea)
 
+Union{}(x::Union{}) = throw(MethodError(convert, (Union{}, x)))
+
 """
     @eval [mod,] ex
 
--- a/base/essentials.jl
+++ b/base/essentials.jl
@@ -166,6 +166,7 @@ true
 """
 function convert end
 
+convert(::Type{Union{}}, x::Union{}) = throw(MethodError(convert, (Union{}, x)))
 convert(::Type{Union{}}, x) = throw(MethodError(convert, (Union{}, x)))
 convert(::Type{Any}, x) = x
 convert(::Type{T}, x::T) where {T} = x
@@ -447,6 +448,9 @@ Stacktrace:
 ```
 """
 sizeof(x) = Core.sizeof(x)
+# The next two methods prevent invalidation
+sizeof(::Type{Union{}}) = Core.sizeof(Union{})
+sizeof(::Type{T}) where T = Core.sizeof(T)
 
 # simple Array{Any} operations needed for bootstrap
 @eval setindex!(A::Array{Any}, @nospecialize(x), i::Int) = arrayset($(Expr(:boundscheck)), A, x, i)
--- a/base/number.jl
+++ b/base/number.jl
@@ -2,6 +2,10 @@
 
 ## generic operations on numbers ##
 
+# prevent invalidation
+Union{}(x::Number)                  = throw(MethodError(convert, (Union{}, x)))
+convert(::Type{Union{}}, x::Number) = throw(MethodError(convert, (Union{}, x)))
+
 # Numbers are convertible
 convert(::Type{T}, x::T)      where {T<:Number} = x
 convert(::Type{T}, x::Number) where {T<:Number} = T(x)
@@ -240,6 +244,9 @@ julia> zero(rand(2,2))
 """
 zero(x::Number) = oftype(x,0)
 zero(::Type{T}) where {T<:Number} = convert(T,0)
+# prevent invalidation
+zero(x::Union{})      = throw(MethodError(zero, (x,)))
+zero(::Type{Union{}}) = throw(MethodError(zero, (Union{},)))
 
 """
     one(x)
@@ -275,6 +282,9 @@ julia> import Dates; one(Dates.Day(1))
 """
 one(::Type{T}) where {T<:Number} = convert(T,1)
 one(x::T) where {T<:Number} = one(T)
+# prevent invalidation
+one(x::Union{})      = throw(MethodError(one, (x,)))
+one(::Type{Union{}}) = throw(MethodError(one, (Union{},)))
 # note that convert(T, 1) should throw an error if T is dimensionful,
 # so this fallback definition should be okay.
 
@@ -298,6 +308,9 @@ julia> import Dates; oneunit(Dates.Day)
 """
 oneunit(x::T) where {T} = T(one(x))
 oneunit(::Type{T}) where {T} = T(one(T))
+# prevent invalidation
+oneunit(x::Union{})      = throw(MethodError(oneunit, (x,)))
+oneunit(::Type{Union{}}) = throw(MethodError(oneunit, (Union{},)))
 
 """
     big(T::Type)
--- a/base/reduce.jl
+++ b/base/reduce.jl
@@ -306,25 +306,29 @@ with reduction `op` over an empty array with element type of `T`.
 
 If not defined, this will throw an `ArgumentError`.
 """
-reduce_empty(op, T) = _empty_reduce_error()
-reduce_empty(::typeof(+), T) = zero(T)
+reduce_empty(op, ::Type{T}) where T = _empty_reduce_error()
+reduce_empty(::typeof(+), ::Type{Union{}}) = _empty_reduce_error()   # avoid invalidation
+reduce_empty(::typeof(+), ::Type{T}) where T = zero(T)
 reduce_empty(::typeof(+), ::Type{Bool}) = zero(Int)
-reduce_empty(::typeof(*), T) = one(T)
+reduce_empty(::typeof(*), ::Type{Union{}}) = _empty_reduce_error()
+reduce_empty(::typeof(*), ::Type{T}) where T = one(T)
 reduce_empty(::typeof(*), ::Type{<:AbstractChar}) = ""
 reduce_empty(::typeof(&), ::Type{Bool}) = true
 reduce_empty(::typeof(|), ::Type{Bool}) = false
 
-reduce_empty(::typeof(add_sum), T) = reduce_empty(+, T)
+reduce_empty(::typeof(add_sum), ::Type{Union{}}) = _empty_reduce_error()
+reduce_empty(::typeof(add_sum), ::Type{T}) where T = reduce_empty(+, T)
 reduce_empty(::typeof(add_sum), ::Type{T}) where {T<:SmallSigned}  = zero(Int)
 reduce_empty(::typeof(add_sum), ::Type{T}) where {T<:SmallUnsigned} = zero(UInt)
-reduce_empty(::typeof(mul_prod), T) = reduce_empty(*, T)
+reduce_empty(::typeof(mul_prod), ::Type{Union{}}) = _empty_reduce_error()
+reduce_empty(::typeof(mul_prod), ::Type{T}) where T = reduce_empty(*, T)
 reduce_empty(::typeof(mul_prod), ::Type{T}) where {T<:SmallSigned}  = one(Int)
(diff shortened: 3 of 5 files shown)


timholy:
CC @johnnychen94, @kimikage

base/essentials.jl
```diff
@@ -166,13 +166,16 @@ true
 """
 function convert end
 
+convert(::Type{Union{}}, x::Union{}) = throw(MethodError(convert, (Union{}, x)))
```
StefanKarpinski, line 169: This is my favorite method.

timholy: Yes, this PR has been an exercise in "how on earth can inference discover an ambiguity *here*?"

timholy:
I added another couple of weird methods and got it down to 33 lines with 6 real methods. I've also been able to simplify the FPN diff; I just edited the one above to be the new version.

One benchmark (Pkg is a juicy target for invalidation, it seems):

julia 1.4:
```
julia> using Pkg

julia> @time try Pkg.add("NoPkg") catch err end
   Updating registry at `~/.julia/registries/General`
   Updating git-repo `git@github.com:JuliaRegistries/General.git`
   Updating registry at `~/.julia/registries/HolyLabRegistry`
   Updating git-repo `git@github.com:HolyLab/HolyLabRegistry.git`
  3.968906 seconds (1.31 M allocations: 70.182 MiB, 1.79% gc time)

julia> @time try Pkg.add("NoPkg") catch err end
  0.100881 seconds (365.53 k allocations: 19.821 MiB, 50.80% gc time)

julia> @time try Pkg.add("NoPkg") catch err end
  0.053736 seconds (365.53 k allocations: 19.821 MiB)

julia> using FixedPointNumbers

julia> @time try Pkg.add("NoPkg") catch err end
  2.568494 seconds (555.27 k allocations: 28.923 MiB, 0.47% gc time)
```
So a 2.5 second hit to recompile Pkg methods after loading FixedPointNumbers.

This PR:
```
julia> using Pkg

julia> @time try Pkg.add("NoPkg") catch err end
   Updating registry at `~/.julia/registries/General`
   Updating git-repo `git@github.com:JuliaRegistries/General.git`
   Updating registry at `~/.julia/registries/HolyLabRegistry`
   Updating git-repo `git@github.com:HolyLab/HolyLabRegistry.git`
  4.528111 seconds (891.47 k allocations: 52.362 MiB, 0.24% gc time)

julia> @time try Pkg.add("NoPkg") catch err end
  0.015854 seconds (98.80 k allocations: 7.125 MiB)

julia> using FixedPointNumbers

julia> @time try Pkg.add("NoPkg") catch err end
  0.067083 seconds (125.96 k allocations: 8.486 MiB, 54.02% gc time)
```
Definitely not enough for a cup of tea.

So, another package made largely "side-effect free" with respect to other functionality. Whee!

timholy:
Residual log, in case anyone has additional suggestions for improvement:
```
-- oneunit(Int32)
 first(Base.OneTo{T<:AbstractChar}) where {T<:AbstractChar}
>> FixedPointNumbers.oneunit(...) Tuple{typeof(Base.oneunit), Type{X}} where X<:(FixedPointNumbers.FixedPoint{T, f} where f where T<:Integer)
-- max(Int32, Int32)
-- *(Float64, Int64)
-- *(UInt64, Float64)
-- min(Int64, Int64)
-- min(Int32, Int32)
-- /(UInt64, Float64)
-- zero(Type{Int32})
-- <(Base.VersionNumber, Base.VersionNumber)
 promote_type(Type{Float64}, Type{S} where S<:Integer)
 promote_type(Type{Float64}, Type{S} where S<:Integer)
>> FixedPointNumbers.promote_rule(...) Tuple{typeof(Base.promote_rule), Type{T}, Type{Tf}} where Tf<:AbstractFloat where T<:(FixedPointNumbers.Normed{T, f} where f where T<:Unsigned)
oneunit(Type{T} where T<:AbstractChar)
-- one(Type{UInt64})
>> FixedPointNumbers.one(...) Tuple{typeof(Base.one), Type{X}} where X<:(FixedPointNumbers.FixedPoint{T, f} where f where T<:Integer)
-- +(Float64, Int64)
-- ==(Module, Module)
-- ==(Int64, Int64)
-- ==(Distributed.WorkerState, Distributed.WorkerState)
-- ==(Char, String)
-- ==(Base.Multimedia.TextDisplay, REPL.REPLDisplay{REPL.LineEditREPL})
-- ==(Expr, Int64)
-- ==(Int32, Int64)
-- ==(Int32, Int32)
-- ==(Symbol, Int64)
-- reinterpret(Type{Char}, UInt32)
-- <=(Int32, Int64)
-- (::Type{Char})(Int64)
convert(Type{T<:Int64}, Int64) where {T<:Int64}
 setindex!(Array{K, 1} where K<:Union{Int64, String, Symbol}, Int64, Int64)
>> FixedPointNumbers.FixedPoint(...) Tuple{Type{X}, Number} where X<:(FixedPointNumbers.FixedPoint{T, f} where f where T<:Integer)
```

JeffBezanson:
I have some code somewhere that skips doing invalidations if the intersection would be ambiguous. I think that could avoid the need for the `Union{}` methods here. In fact I can't be confident nothing will go wrong having an argument marked `::Union{}`. I didn't PR the changes before since in the cases I tried it only affected a small fraction of invalidations, but maybe it will work in this case.

timholy:
ColorTypes only needs this change:
```diff
diff --git a/src/conversions.jl b/src/conversions.jl
index b74d961..79c1014 100644
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -70,9 +70,9 @@ end
 # no-op and element-type conversions, plus conversion to and from transparency
 # Colorimetry conversions are in Colors.jl
 convert(::Type{C}, c::C) where {C<:Colorant} = c
-convert(::Type{C}, c) where {C<:Colorant} = cconvert(ccolor(C, typeof(c)), c)
+convert(::Type{C}, c::Colorant) where {C<:Colorant} = cconvert(ccolor(C, typeof(c)), c)
 cconvert(::Type{C}, c::C) where {C} = c
-cconvert(::Type{C}, c) where {C} = _convert(C, base_color_type(C), base_color_type(c), c)
+cconvert(::Type{C}, c::Colorant) where {C} = _convert(C, base_color_type(C), base_color_type(c), c)
 
 convert(::Type{C}, c::Color, alpha) where {C<:TransparentColor} = cconvert(ccolor(C, typeof(c)), c, alpha)
 cconvert(::Type{C}, c::Color, alpha) where {C<:TransparentColor} =_convert(C, base_color_type(C), base_color_type(c), c, alpha)
```
(plus some stuff to disable the error hints since that changed) and the total invalidation log for FixedPointNumbers + ColorTypes is down to 55 lines. This is from an original of 2053.

The fact that ColorTypes has largely been fixed by the same changes made for FPN is a relief; I was beginning to worry that each package would be a big lift on its own, this is the first sign I've seen of synergy.

timholy:
> skips doing invalidations if the intersection would be ambiguous

I wondered about that. I figured it would be worth seeing how much this contributed to the problem; it's certainly the majority for this package.

base/essentials.jl
```diff
@@ -447,6 +448,9 @@ Stacktrace:
 ```
 """
 sizeof(x) = Core.sizeof(x)
+# The next two methods prevent invalidation
+sizeof(::Type{Union{}}) = Core.sizeof(Union{})
+sizeof(::Type{T}) where T = Core.sizeof(T)
```
vtjnash, line 455: This'll force us to generate code for each type, and forces us to add a dispatch, which I believe we don't want for it?

base/number.jl
```diff
@@ -240,6 +244,9 @@ julia> zero(rand(2,2))
 """
 zero(x::Number) = oftype(x,0)
 zero(::Type{T}) where {T<:Number} = convert(T,0)
+# prevent invalidation
+zero(x::Union{})      = throw(MethodError(zero, (x,)))
```
vtjnash, line 248:
```suggestion
zero(x::Union{})      = unreachable
```

Though I can't promise that defining these might not cause other, unintended, side-effects.

base/reduce.jl
```diff
@@ -306,25 +306,29 @@ with reduction `op` over an empty array with element type of `T`.
 
 If not defined, this will throw an `ArgumentError`.
 """
-reduce_empty(op, T) = _empty_reduce_error()
-reduce_empty(::typeof(+), T) = zero(T)
+reduce_empty(op, ::Type{T}) where T = _empty_reduce_error()
```
vtjnash, line 309: Style nit: I think we usually write short form functions with the braces `{}` to visually distinguish them:
```suggestion
reduce_empty(op, ::Type{T}) where {T} = _empty_reduce_error()
```

base/number.jl
JeffBezanson: :100: Agreed, I would hold off on adding these --- they could cause problems, and they're the kind of thing we should be able to handle automatically (i.e. if an argument is bottom, obviously the code is unreachable).

timholy:
I'd be very happy to hold off on this, especially if there's an automatic solution on the horizon. I don't have a burning need to see these littering our method tables. Happy to treat it as a proof of principle for a more systematic approach.

KristofferC:
We should be able to keep all the `Type{Union{})` ones though? And there are some other here that could be good to get in? Can we factor out the `::Union{}` ones so the rest could be merged?

base/reduce.jl
```diff
@@ -306,25 +306,29 @@ with reduction `op` over an empty array with element type of `T`.
...
+reduce_empty(::typeof(+), ::Type{Union{}}) = _empty_reduce_error()   # avoid invalidation
```
tkf, line 310: I think we should have this. It is not because for avoiding invalidation but for "correctness."  I think it's better to do `_empty_reduce_error()` in `sum(Union{}[])` than throwing the `MethodError`.

timholy:
Superseded by lots of other PRs

PR #35737, opened 2020-05-04
fix #35708, GC seeing uninitialized struct in `jl_new_struct*`

### LINKED ISSUE #35708: Segfault in Julia runtime
Following the instructions in [Segfaults when running a script](https://docs.julialang.org/en/v1/devdocs/backtraces/#Segfaults-when-running-a-script-1):

Here is a [Gist](https://gist.github.com/bzinberg/9f065885929564d53947f07a484a845f) containing versioninfo, backtrace and valgrind output.

This crash occurs in Julia 1.4.1 and 1.4.0, but not 1.3.1.

I'm not experienced in this kind of debugging, so not sure what further information would be useful.  Happy to dialogue about it.

Thanks!

JeffBezanson:
fix #35708

### DIFF
--- a/src/datatype.c
+++ b/src/datatype.c
@@ -901,20 +901,16 @@ JL_DLLEXPORT jl_value_t *jl_new_structv(jl_datatype_t *type, jl_value_t **args, 
         jl_type_error("new", (jl_value_t*)jl_datatype_type, (jl_value_t*)type);
     if (type->ninitialized > na || na > jl_datatype_nfields(type))
         jl_error("invalid struct allocation");
-    if (type->instance != NULL) {
-        for (size_t i = 0; i < na; i++) {
-            jl_value_t *ft = jl_field_type(type, i);
-            if (!jl_isa(args[i], ft))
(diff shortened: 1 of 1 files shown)


src/datatype.c
```diff
@@ -943,14 +939,17 @@ JL_DLLEXPORT jl_value_t *jl_new_structt(jl_datatype_t *type, jl_value_t *tup)
...
     JL_GC_PUSH2(&jv, &fi);
     for (size_t i = 0; i < nargs; i++) {
         jl_value_t *ft = jl_field_type(type, i);
         fi = jl_get_nth_field(tup, i);
         if (!jl_isa(fi, ft))
             jl_type_error("new", ft, fi);
+    }
+    jv = jl_gc_alloc(ptls, jl_datatype_size(type), type);
+    for (size_t i = 0; i < nargs; i++) {
+        fi = jl_get_nth_field(tup, i);
```
vtjnash, line 952: This allocates too

PR #35738, opened 2020-05-05
Remove USING_SJLJ_EXCEPTIONS

stahta01:
The defining of USING_SJLJ_EXCEPTIONS should only be done by the compiler.

From [Common Predefined Macros](https://gcc.gnu.org/onlinedocs/cpp/Common-Predefined-Macros.html)

> __USING_SJLJ_EXCEPTIONS__ This macro is defined, with value 1, if the compiler uses the old mechanism based on setjmp and longjmp for exception handling.


yuyichao:
Seems to be an ancient hack to workaround a llvm bug. In the off change that the bug is still there it should probably be fixed with a patch instead......

PR #34101, opened 2019-12-14
Optimize copy!(::Dict, ::Dict)

tkf:
This PR implements two optimizations for `copy!(dst::Dict, src::Dict)`. This came up in another PR for implementing `setindex` for `Dict`: https://github.com/JuliaLang/julia/pull/33495#discussion_r357008424

### Case 1: `copy!(dst::D, src::D) where {D <: Dict}`

This is a straightforward case where you can copy properties as-is:

```julia
function unsafe_copy!(dst::D, src::D) where {D <: Dict}
    copy!(dst.vals, src.vals)
    copy!(dst.keys, src.keys)
    copy!(dst.slots, src.slots)

    dst.ndel = src.ndel
    dst.count = src.count
    dst.age = src.age
    dst.idxfloor = src.idxfloor
    dst.maxprobe = src.maxprobe
    return dst
end


using BenchmarkTools
d = Dict{Int,Int}(zip(1:10^5, 1:10^5))
c = Dict{Int,Int}()
@btime copy!($c, $d)
@btime unsafe_copy!($c, $d)
```

prints

```
  2.146 ms (0 allocations: 0 bytes)
  299.572 μs (0 allocations: 0 bytes)
```

in Julia 1.3.

### Case 2: `copy!(dst::Dict{Kd}, src::Dict{Ks}) where {Ks,Kd>:Ks}`

This is the case you don't need to re-compute hashes:

```julia
using Base: isslotfilled

function unsafe_copy!(dst::Dict, src::Dict)
    resize!(dst.vals, length(src.vals))
    resize!(dst.keys, length(src.keys))

    svals = src.vals
    skeys = src.keys
    dvals = dst.vals
    dkeys = dst.keys
    @inbounds for i = src.idxfloor:lastindex(svals)
        if isslotfilled(src, i)
            dvals[i] = svals[i]
            dkeys[i] = skeys[i]
        end
    end

    copy!(dst.slots, src.slots)

    dst.ndel = src.ndel
    dst.count = src.count
    dst.age = src.age
    dst.idxfloor = src.idxfloor
    dst.maxprobe = src.maxprobe
    return dst
end


using BenchmarkTools
d = Dict{Int,Int}(zip(1:10^5, 1:10^5))
c = Dict{Any,Float64}()
@btime copy!($c, $d)
@btime unsafe_copy!($c, $d)
```

This prints

```
  4.030 ms (130549 allocations: 1.99 MiB)
  1.889 ms (99489 allocations: 1.52 MiB)
```

[...]

PR #35739, opened 2020-05-05
Revert "Optimize copy!(::Dict, ::Dict)"

Keno:
Reverts JuliaLang/julia#34101. Seems to have broken CI.


tkf (approved):

staticfloat:
Confirmed that it broke CI (I bisected to this commit), and since it is getting the green flush here (especially on Windows, where it was causing the issues), let's merge.

PR #24999, opened 2017-12-09
string overhaul

StefanKarpinski:
Replacement PR for https://github.com/JuliaLang/julia/pull/24439. More description to come...

Edit see this discourse post for a description: https://discourse.julialang.org/t/changes-to-the-representation-of-char/7291/5

PR #35742, opened 2020-05-05
add better implementations for generic findprev findnext for AbstractString

### LINKED ISSUE #35721: DelimitedFiles readdlm is externally slow for Complex{Float64}

KristofferC:
Ref https://github.com/JuliaLang/julia/pull/35727#issuecomment-623499860

```jl
using DelimitedFiles
A = randn(Complex{Float64},300,50)
writedlm("A.csv",A,',')
@time readdlm("A.csv",',',Complex{Float64},'\n')
```

Before:
```
  9.316321 seconds (293.64 k allocations: 3.262 GiB, 3.35% gc time)
```

After

```
 0.013526 seconds (30.09 k allocations: 1.882 MiB)
```
Fixes #35721.

### DIFF
--- a/base/strings/search.jl
+++ b/base/strings/search.jl
@@ -130,9 +130,9 @@ function findnext(testf::Function, s::AbstractString, i::Integer)
     1 ≤ i ≤ z || throw(BoundsError(s, i))
     @inbounds i == z || isvalid(s, i) || string_index_err(s, i)
     e = lastindex(s)
-    @inbounds while i <= e
-        testf(s[i]) && return i
-        i = nextind(s, i)
+    while i <= e
+        testf(@inbounds s[i]) && return i
+        i = @inbounds nextind(s, i)
     end
     return nothing
 end
@@ -340,9 +340,9 @@ function findprev(testf::Function, s::AbstractString, i::Integer)
     0 ≤ i ≤ z || throw(BoundsError(s, i))
     i == z && return nothing
     @inbounds i == 0 || isvalid(s, i) || string_index_err(s, i)
-    @inbounds while i >= 1
-        testf(s[i]) && return i
-        i = prevind(s, i)
+    while i >= 1
+        testf(@inbounds s[i]) && return i
+        i = @inbounds prevind(s, i)
     end
     return nothing
 end
--- a/base/strings/search.jl
+++ b/base/strings/search.jl
@@ -127,15 +127,16 @@ findfirst(ch::AbstractChar, string::AbstractString) = findfirst(==(ch), string)
 function findnext(testf::Function, s::AbstractString, i::Integer)
     i = Int(i)
     z = ncodeunits(s) + 1
-    1 ≤ i ≤ z || throw(BoundsError(s, i))
+    1 ≤ i ≤ z || throw(BoundsError(s, i))
     @inbounds i == z || isvalid(s, i) || string_index_err(s, i)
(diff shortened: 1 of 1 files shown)


timholy:
An almost 1000x improvement, not too shabby.

base/strings/search.jl
```diff
@@ -127,16 +127,17 @@ findfirst(ch::AbstractChar, string::AbstractString) = findfirst(==(ch), string)
...
-    1 ≤ i ≤ z || throw(BoundsError(s, i))
+    1 ≤ i ≤ z || throw(BoundsError(s, i))
     @inbounds i == z || isvalid(s, i) || string_index_err(s, i)
-    for (j, d) in pairs(SubString(s, i))
-        if testf(d)
-            return i + j - 1
-        end
+    e = lastindex(s)
+    @inbounds while i <= e
+        testf(s[i]) && return i
```
rfourquet, line 134: This might run `testf` with bounds check off, it was [suggested](https://github.com/JuliaLang/julia/pull/34885#discussion_r384491202) to me to not do that.

stevengj:
Out of curiosity, how does the speed of `findnext` compare to the old version?

stevengj (approved):

base/strings/search.jl
```diff
@@ -334,18 +335,16 @@ findlast(ch::AbstractChar, string::AbstractString) = findlast(==(ch), string)
...
-    if i < 1
-        return i == 0 ? nothing : throw(BoundsError(s, i))
-    end
-    n = ncodeunits(s)
-    if i > n
-        return i == n+1 ? nothing : throw(BoundsError(s, i))
+    i = Int(i)
+    z = ncodeunits(s) + 1
+    0 ≤ i ≤ z || throw(BoundsError(s, i))
+    i == z && return nothing
```
stevengj, line 341: This matches the old behavior, but it mystifies me.   Was it intentional, or a bug?

stevengj: (Note that `findnext` and `findprev` for `Array` both throw a `BoundsError` if the starting index is invalid, so the string functions are inconsistent.)

KristofferC: I remember reading something about allowing indexing strings one byte after the end or something but don't recall the details. @bkamins or @StefanKarpinski might know.

bkamins: https://github.com/JuliaLang/julia/pull/24999#issuecomment-350492880

stevengj: Okay.

stevengj (approved):

stevengj:
Note: Fixes #35721.

KristofferC:
FreeBSD failure looks unrelated.

PR #35747, opened 2020-05-05
debug_method_invalidation: print the module-of-definition

timholy:
This might be helpful as we consider writing tooling to help people analyze `MethodInstance` invalidations.


timholy:
xref #35768

PR #35755, opened 2020-05-05
Make kron(a,b) preserve Adjoint-ness

mcabbott:
This makes `kron` of two `Adjoint` vectors return the same type, and solves the following important obstacle to world peace:
```julia
A = rand(3,3); B = rand(3,3); 
v = rand(3); w = rand(3);
C = rand(5,9);

# Then these are equal, kron commutes with *
C * (kron(A,B) * kron(v,w))
C * kron(A*v, B*w)

# But these are not:
C * (kron(v,w)' * kron(A,B))' # 5-element Array
C * kron(v'*A, w'*B)'         # 5×1 Array

# Because this a 1×9 matrix, not an Adjoint:
kron(v', w')
```
It also adjusts the docstring of `kron` so as not to suggest that it's closely related to an outer product in general, but only for real vectors (as discussed more than once in #35150). 

And it makes `vec(v') === v` for real vectors; this was already true for `vec(transpose(v))`, and was used in making this work before I had a better idea. (I can remove this if someone wishes to be very strict about orthogonal PRs.)

### DIFF
--- a/stdlib/LinearAlgebra/src/adjtrans.jl
+++ b/stdlib/LinearAlgebra/src/adjtrans.jl
@@ -221,6 +221,7 @@ similar(A::AdjOrTrans, ::Type{T}, dims::Dims{N}) where {T,N} = similar(A.parent,
 # sundry basic definitions
 parent(A::AdjOrTrans) = A.parent
 vec(v::TransposeAbsVec) = parent(v)
+vec(v::AdjointAbsVec{<:Real}) = parent(v)
 
 ### concatenation
 # preserve Adjoint/Transpose wrapper around vectors
--- a/stdlib/LinearAlgebra/src/dense.jl
+++ b/stdlib/LinearAlgebra/src/dense.jl
@@ -401,6 +401,9 @@ kron(a::AbstractVector, b::AbstractVector) = vec(kron(reshape(a ,length(a), 1), 
 kron(a::AbstractMatrix, b::AbstractVector) = kron(a, reshape(b, length(b), 1))
 kron(a::AbstractVector, b::AbstractMatrix) = kron(reshape(a, length(a), 1), b)
 
+kron(a::AdjointAbsVec, b::AdjointAbsVec) = adjoint(kron(parent(a), parent(b)))
+kron(a::AdjOrTransAbsVec, b::AdjOrTransAbsVec) = transpose(kron(vec(a), vec(b)))
+
 # Matrix power
 (^)(A::AbstractMatrix, p::Integer) = p < 0 ? power_by_squaring(inv(A), -p) : power_by_squaring(A, p)
 function (^)(A::AbstractMatrix{T}, p::Integer) where T<:Integer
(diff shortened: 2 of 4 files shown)


EricForgy:
As mentioned in the linked PR discussion, but specifically considering Jeff's [comment](https://github.com/JuliaLang/julia/pull/35150#issuecomment-624112184)

> It would be great to have an issue (is there one already?) listing all the dodgy stuff in LinearAlgebra that we want to deprecate or revisit. I certainly wouldn't spend much time defending kron, personally :)

I would be hesitant to make changes to LinearAlgebra, especially anything related to tensor product, until we've discussed any surrounding issues a bit more. `kron` can be represented as a 2D array of numbers, but it is not a linear map so I'm not clear on the motivation for this PR. 

Can we gather any of the issues that have been eluded to and come up with a comprehensive solution? Maybe I am worried for nothing, but I fear we might be propagating some issues that should rather be resolved at the source.

mcabbott:
This is a teeny-tiny fix of an edge case, plus a clarification of one docstring which appeared to be a source of confusion. It does not take any position on big issues, and does not modify the position implicit in what LinearAlgebra does or does not include already. 

However, if `A` and `B` are linear maps (represented as matrices) which act on vectors like `v` and `w`, then in its present incarnation, `kron(A,B)` is indeed a linear map (matrix) which acts on vectors like `kron(v,w)`. That's the universe in which my motivating example above operates.

EricForgy:
Cool. That makes sense. In that case, would it make more sense for this to go into [TensorCore.jl](https://github.com/JuliaMath/TensorCore.jl)? Your "vectors" are really higher order tensors (albeit satisfying vector axioms) and your linear maps are tensors mapping tensors to tensors, which sounds more like the realm of multilinear algebra 🤔

andreasnoack (approved):

mcabbott:
TensorCore can't define a method for `kron(::Adjoint, ...)` without committing piracy. 

Whether there are better ways to represent this, maybe there are, that's what I mean by bigger issues! Vectors like `kron(v,w)` are indeed mathematically two-index objects... and if they were represented by `Array{T,2}` instead, then my toy example would need a `*` which knows to contract two dimensions.

andreasnoack:
I think we can consider this a fix of the existing functionality. It's just a matter of preserving a little more structure than what the fallback allows for.

EricForgy:
Sure 👍 

In this PR you are fixing something that is already there.

A bigger issue for discussion elsewhere might be to move all multilinear machinery such as `kron` into `TensorCore`. I'll try to create an issue with summary and possible future design considerations separately 🙏

stevengj:
(Since `kron` works with matrices, as opposed to higher-dimensional arrays—even though there is a homomorphism between the two—I would keep it in LinearAlgebra.)

mcabbott:
Somewhat surprisingly, 
```julia
julia> @which kron
Base
```
although many of its methods are defined in LinearAlgebra.

timholy (approved):
Nice work @mcabbott, thanks!

PR #35759, opened 2020-05-05
RFC: switch coverage counters from volatile to properly atomic

vtjnash:
This makes it fully thread-safe, though also much more expensive.

Test function:
```julia
julia> f() = for i = 1:10^8; ccall(:jl_gc_safepoint, Cvoid, ()); end
```
Without coverage, this takes ~0.030090 seconds
On master, this takes ~0.214468 seconds (est. 7x increase in overhead)
On this PR, this takes ~4.553456 seconds (est. 20x additional increase in overhead)

### DIFF
--- a/src/codegen.cpp
+++ b/src/codegen.cpp
@@ -1199,16 +1199,14 @@ static uint64_t *allocLine(std::vector<logdata_block*> &vec, int line)
     return &data[line];
 }
 
-static void visitLine(jl_codectx_t &ctx, std::vector<logdata_block*> &vec, int line, Value *addend, const char* name)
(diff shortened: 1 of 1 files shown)


JeffBezanson:
Yikes, that's certainly a scary slowdown. Might warrant an application-level benchmark to see if it's viable. Otherwise I guess we might have to make this thread-local?

vtjnash:
Agreed, I’m thinking we should not do this, but thought it worth opening the PR so show the numbers.

We should also consider reviving carnivals PR too, or switch to use the infrastructure LLVM now has for doing this.

PR #35760, opened 2020-05-05
fix more PCRE error value types

JeffBezanson:
In PCRE, all of these accept and return Cint, so ideally we could stop generating the `% UInt32` for error codes in pcre_h.jl.


StefanKarpinski:
I'm looking into fixing this and making the constants of the correct type. It's tricky, of course, because C constants are macro values and don't actually have any type.

StefanKarpinski:
Closing in favor of https://github.com/JuliaLang/julia/pull/35773.

PR #35768, opened 2020-05-06
jl_debug_method_invalidation: return a list of instances

timholy:
Rather than print information about method invalidations, this allows us to store the `MethodInstance`s themselves. The reason I think this is attractive is that analyzing the reason for invalidations sometimes benefits from type intersection, and it will be easier to perform that kind of analysis if we don't have to try to reconstruct the type from its repr.

You can see a WIP analysis in SnoopCompile here: https://github.com/timholy/SnoopCompile.jl/blob/teh/invalidations/src/invalidations.jl

The biggest negative here is that this gets in the way of some potential debug information, e.g., if Julia crashes or doesn't get through bootstrap. So this isn't without some loss, CC @vtjnash. Nevertheless I think method invalidation is rarely an issue in such cases, so perhaps it's a loss we can live with.

There are some "competing" PRs, I'll list them as being closed by this one (should we choose to go this way):
Closes #35745
Closes #35747

CC @ianshmean

### DIFF
--- a/src/gf.c
+++ b/src/gf.c
@@ -1485,17 +1485,33 @@ static void update_max_args(jl_methtable_t *mt, jl_value_t *type)
         mt->max_args = na;
 }
 
-JL_DLLEXPORT int jl_debug_method_invalidation = 0;
+jl_array_t *_jl_debug_method_invalidation JL_GLOBALLY_ROOTED = NULL;
+JL_DLLEXPORT jl_value_t *jl_debug_method_invalidation(int state)
+{
+    /* After calling with `state = 1`, caller is responsible for
+       holding a reference to the returned array until this is called
+       again with `state = 0`. */
+    if (state) {
+        if (_jl_debug_method_invalidation)
+            return (jl_value_t*) _jl_debug_method_invalidation;
+        _jl_debug_method_invalidation = jl_alloc_array_1d(jl_array_any_type, 0);
+        return (jl_value_t*) _jl_debug_method_invalidation;
+    }
+    _jl_debug_method_invalidation = NULL;
+    return jl_nothing;
+}
 
 // recursively invalidate cached methods that had an edge to a replaced method
 static void invalidate_method_instance(jl_method_instance_t *replaced, size_t max_world, int depth)
 {
-    if (jl_debug_method_invalidation) {
-        int d0 = depth;
-        while (d0-- > 0)
-            jl_uv_puts(JL_STDOUT, " ", 1);
-        jl_static_show(JL_STDOUT, (jl_value_t*)replaced);
-        jl_uv_puts(JL_STDOUT, "\n", 1);
+    if (_jl_debug_method_invalidation) {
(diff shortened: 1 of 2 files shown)


KristofferC:
Should this also be updated?

https://github.com/JuliaLang/julia/blob/7ba7daf763dc670888d0e92cc342f91371a01ad8/src/dump.c#L2389-L2391

vtjnash:
Also needs to be added to the GC and marked as globally rooted. I'm not worried about the loss in early printing—this'll be useful if I start adding a lot of false positive that need to be filtered out later (and add the code to filter them).

timholy:
Thanks @vtjnash. Sorry to be dense.... Is this just call `JL_GC_PROMISE_ROOTED`? It's been a while since I looked at the GC and I think some things have changed. Are there docs?

timholy:
Also, the invalidations last "forever," right? In which case do we need to save the methods invalidated through the backedges? We could just recover those later, right?

src/gf.c
```diff
@@ -1485,17 +1485,27 @@ static void update_max_args(jl_methtable_t *mt, jl_value_t *type)
         mt->max_args = na;
 }
 
-JL_DLLEXPORT int jl_debug_method_invalidation = 0;
+jl_array_t *_jl_debug_method_invalidation = NULL;
```
vtjnash, line 1488:
```suggestion
jl_array_t *_jl_debug_method_invalidation JL_GLOBALLY_ROOTED = NULL;
```

src/gf.c
```diff
@@ -1485,17 +1485,27 @@ static void update_max_args(jl_methtable_t *mt, jl_value_t *type)
...
+JL_DLLEXPORT jl_value_t *jl_debug_method_invalidation(int state)
+{
+    if (state) {
+        if (_jl_debug_method_invalidation)
+            return (jl_value_t*) _jl_debug_method_invalidation;
+        JL_GC_PUSH1(&_jl_debug_method_invalidation);
+        _jl_debug_method_invalidation = jl_alloc_array_1d(jl_array_any_type, 0);
+        JL_GC_POP();
```
vtjnash, line 1496:
```suggestion
        _jl_debug_method_invalidation = jl_alloc_array_1d(jl_array_any_type, 0);
```
(this frame is illogical, since there's no use of the value protected inside it)

timholy:
This should pass. Still 1.5 material or are we into 1.6 now?

The last commit is worth noting, I've found it really helps make sense of why some methods are being invalidated.

timholy:
Used in https://github.com/JuliaLang/www.julialang.org/pull/794

timholy:
I've squashed this down to a single commit to make it easier to resolve conflicts with #35877. From my standpoint this is good to go and can be merged any time. With apologies to Jeff, it might make sense to merge this before #35877 so one can more readily compare before & after. (For instance, in the blog post I analyze *unique* MethodInstance invalidations.)

PR #35773, opened 2020-05-06
PCRE: use correct type for error constants (Cint), audit C calls

StefanKarpinski:


base/pcre.jl
```diff
@@ -144,20 +150,25 @@ free_jit_stack(stack) =
...
-    buffer = Vector{UInt8}(undef, 256)
-    ccall((:pcre2_get_error_message_8, PCRE_LIB), Cvoid,
-          (UInt32, Ptr{UInt8}, Csize_t), errno, buffer, sizeof(buffer))
-    GC.@preserve buffer unsafe_string(pointer(buffer))
+function err_message(errno::Integer)
+    buffer = Vector{UInt8}(undef, 1024)
+    ret = ccall((:pcre2_get_error_message_8, PCRE_LIB), Cint,
+                (Cint, Ptr{UInt8}, Csize_t), errno, buffer, length(buffer))
+    ret == ERROR_BADDATA && error("PCRE error: invalid errno ($errno)")
+    # TODO: seems like there should be a better way to get this string
```
JeffBezanson, line 158: `String(buffer)` should work.

StefanKarpinski: Won't that include the NUL byte and following data?

JeffBezanson: Oh, right, the vector would need to be resized first.

StefanKarpinski: I'm fine with this, I had a slight worry about the memory leaking though.

PR #35778, opened 2020-05-06
Define machinery so that Some can be used for broadcast

### LINKED ISSUE #18379: Scalar type for broadcasting

MasonProtter:
As mentioned in https://github.com/JuliaLang/julia/pull/35675, `Some` may be a good candidate to use basically like an immutable version of `Ref` in broadcast expressions. This makes `Some` a zero dimensional container that knows how to broadcast efficiently.

I figure it might be better to discuss the merits of this idea here rather than clogging up that PR. Another candidate brought up in that thread is to create a new singleton type `Only`. 

I think if we can make `Some` work for this without breaking anything, it would be preferable to creating yet another singleton type, but I'm wary that there could be hitherto unforeseen negative consequences to adding this behaviour to `Some`. Perhaps someone is relying on `eltype(Some(1)) == Any`

Closes https://github.com/JuliaLang/julia/issues/18379

### DIFF
--- a/base/some.jl
+++ b/base/some.jl
@@ -97,23 +97,3 @@ something() = throw(ArgumentError("No value arguments present"))
 something(x::Nothing, y...) = something(y...)
 something(x::Some, y...) = x.value
 something(x::Any, y...) = x
-
-#Methods for broadcast
-getindex(s::Some) = s.value
-getindex(s::Some, ::CartesianIndex{0}) = s.value
-
-iterate(s::Some) = (s.value, nothing)
-iterate( ::Some, s) = nothing
-
-ndims(::Some) = 0
-ndims(::Type{<:Some}) = 0
-
-length(::Some) = 1
-size(::Some) = ()
-axes(::Some) = ()
-
-IteratorSize(::Type{<:Some}) = HasShape{0}()
-broadcastable(s::Some) = s
-
-eltype(::Some{T})       where {T} = T
-eltype(::Type{Some{T}}) where {T} = T
--- a/base/some.jl
+++ b/base/some.jl
@@ -6,6 +6,8 @@
 A wrapper type used in `Union{Some{T}, Nothing}` to distinguish between the absence
 of a value ([`nothing`](@ref)) and the presence of a `nothing` value (i.e. `Some(nothing)`).
 
+`Some` is also used in broadcasting to treat it's enclosed value as a scalar.
+
 Use [`something`](@ref) to access the value wrapped by a `Some` object.
 """
 struct Some{T}
@@ -95,3 +97,23 @@ something() = throw(ArgumentError("No value arguments present"))
 something(x::Nothing, y...) = something(y...)
 something(x::Some, y...) = x.value
 something(x::Any, y...) = x
+
+#Methods for broadcast
+Base.getindex(s::Some) = s.value
+Base.getindex(s::Some, ::CartesianIndex{0}) = s.value
+
+Base.iterate(s::Some) = (s.value, nothing)
+Base.iterate( ::Some, s) = nothing
+
+Base.ndims(::Some) = 0
+Base.ndims(::Type{<:Some}) = 0
+
+Base.length(::Some) = 1
+Base.size(::Some) = ()
+Base.axes(::Some) = ()
+
+Base.IteratorSize(::Type{<:Some}) = Base.HasShape{0}()
+Base.broadcastable(s::Some) = s
+
+Base.eltype(::Some{T})       where {T} = T
+Base.eltype(::Type{Some{T}}) where {T} = T
--- a/base/some.jl
+++ b/base/some.jl
@@ -99,21 +99,21 @@ something(x::Some, y...) = x.value
 something(x::Any, y...) = x
 
 #Methods for broadcast
-Base.getindex(s::Some) = s.value
-Base.getindex(s::Some, ::CartesianIndex{0}) = s.value
+getindex(s::Some) = s.value
+getindex(s::Some, ::CartesianIndex{0}) = s.value
 
-Base.iterate(s::Some) = (s.value, nothing)
-Base.iterate( ::Some, s) = nothing
+iterate(s::Some) = (s.value, nothing)
+iterate( ::Some, s) = nothing
 
-Base.ndims(::Some) = 0
-Base.ndims(::Type{<:Some}) = 0
+ndims(::Some) = 0
+ndims(::Type{<:Some}) = 0
 
-Base.length(::Some) = 1
-Base.size(::Some) = ()
-Base.axes(::Some) = ()
+length(::Some) = 1
+size(::Some) = ()
+axes(::Some) = ()
 
-Base.IteratorSize(::Type{<:Some}) = Base.HasShape{0}()
-Base.broadcastable(s::Some) = s
+IteratorSize(::Type{<:Some}) = Base.HasShape{0}()
+broadcastable(s::Some) = s
 
-Base.eltype(::Some{T})       where {T} = T
-Base.eltype(::Type{Some{T}}) where {T} = T
+eltype(::Some{T})       where {T} = T
+eltype(::Type{Some{T}}) where {T} = T
--- a/base/some.jl
+++ b/base/some.jl
@@ -6,7 +6,7 @@
 A wrapper type used in `Union{Some{T}, Nothing}` to distinguish between the absence
 of a value ([`nothing`](@ref)) and the presence of a `nothing` value (i.e. `Some(nothing)`).
 
-`Some` is also used in broadcasting to treat it's enclosed value as a scalar.
+`Some` is also used in broadcasting to treat its enclosed value as a scalar.
 
 Use [`something`](@ref) to access the value wrapped by a `Some` object.
 """
--- a/base/some.jl
+++ b/base/some.jl
@@ -112,7 +112,7 @@ length(::Some) = 1
 size(::Some) = ()
 axes(::Some) = ()
 
-IteratorSize(::Type{<:Some}) = Base.HasShape{0}()
+IteratorSize(::Type{<:Some}) = HasShape{0}()
 broadcastable(s::Some) = s
 
 eltype(::Some{T})       where {T} = T
--- a/base/somebroadcast.jl
+++ b/base/somebroadcast.jl
@@ -1,0 +1,19 @@
+#Methods some Some can broadcast
+getindex(s::Some) = s.value
+getindex(s::Some, ::CartesianIndex{0}) = s.value
+
+iterate(s::Some) = (s.value, nothing)
+iterate( ::Some, s) = nothing
+
+ndims(::Some) = 0
+ndims(::Type{<:Some}) = 0
+
+length(::Some) = 1
+size(::Some) = ()
+axes(::Some) = ()
+
+IteratorSize(::Type{<:Some}) = HasShape{0}()
+broadcastable(s::Some) = s
+
+eltype(::Some{T})       where {T} = T
+eltype(::Type{Some{T}}) where {T} = T
--- a/base/somebroadcast.jl
+++ b/base/somebroadcast.jl
@@ -13,7 +13,7 @@ size(::Some) = ()
 axes(::Some) = ()
 
 IteratorSize(::Type{<:Some}) = HasShape{0}()
-broadcastable(s::Some) = s
+Broadcast.broadcastable(s::Some) = s
 
 eltype(::Some{T})       where {T} = T
 eltype(::Type{Some{T}}) where {T} = T
--- a/base/somebroadcast.jl
+++ b/base/somebroadcast.jl
@@ -15,5 +15,4 @@ axes(::Some) = ()
 IteratorSize(::Type{<:Some}) = HasShape{0}()
 Broadcast.broadcastable(s::Some) = s
 
-eltype(::Some{T})       where {T} = T
 eltype(::Type{Some{T}}) where {T} = T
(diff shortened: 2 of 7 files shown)


base/some.jl
```diff
@@ -95,3 +97,23 @@ something() = throw(ArgumentError("No value arguments present"))
 something(x::Nothing, y...) = something(y...)
 something(x::Some, y...) = x.value
 something(x::Any, y...) = x
+
+#Methods for broadcast
+Base.getindex(s::Some) = s.value
```
mbauman, line 102:
```suggestion
getindex(s::Some) = s.value
```

Don't need the `Base.`s here and below

MasonProtter: Whoops, that was some overzealous copy-pasting from my piracy experiment.

mbauman: I frequently use

```Julia
@eval Base begin
# experiments
end
```

MasonProtter:
Is the problem that I'm using some function that's not yet defined? I don't see what's causing the build error.

mbauman:
Yeah, that's a bootstrap error.  It looks like `some.jl` comes before both `multidimensional.jl` (so can't use `::CartesianIndex`) and `broadcast.jl` (so you'll have trouble extending `Broadcast.broadcastable`).  I think both those files have some "vacationing" definitions along these same lines (e.g., https://github.com/JuliaLang/julia/blob/a4641c8b75b93ce30b706fceed7d2af976f8e578/base/broadcast.jl#L680 and https://github.com/JuliaLang/julia/blob/master/base/multidimensional.jl#L1782)

MasonProtter:
Okay, I split out those new method definitions to a new file and put them after `broadcast`. We'll see if that helps.

base/somebroadcast.jl
```diff
@@ -0,0 +1,19 @@
...
+
+ndims(::Some) = 0
+ndims(::Type{<:Some}) = 0
+
+length(::Some) = 1
+size(::Some) = ()
+axes(::Some) = ()
+
+IteratorSize(::Type{<:Some}) = HasShape{0}()
+broadcastable(s::Some) = s
```
mbauman, line 16:
```suggestion
Broadcast.broadcastable(s::Some) = s
```

base/somebroadcast.jl
```diff
@@ -0,0 +1,19 @@
...
+ndims(::Type{<:Some}) = 0
+
+length(::Some) = 1
+size(::Some) = ()
+axes(::Some) = ()
+
+IteratorSize(::Type{<:Some}) = HasShape{0}()
+Broadcast.broadcastable(s::Some) = s
+
+eltype(::Some{T})       where {T} = T
```
JeffBezanson, line 18: This one isn't needed.

MasonProtter: One would expect it to be defined even if it’s not used by Base / stdlibs though, right?

I use `eltype(v)` where `v` is a vector a fair amount rather than `eltype(typeof(v))`

mbauman: There's the default fallback

https://github.com/JuliaLang/julia/blob/52c9e3df1301ce1ffc5da62626b8d4aadfc22089/base/array.jl#L125

MasonProtter: Oh I see, my bad!

base/some.jl
```diff
@@ -6,6 +6,8 @@
 A wrapper type used in `Union{Some{T}, Nothing}` to distinguish between the absence
 of a value ([`nothing`](@ref)) and the presence of a `nothing` value (i.e. `Some(nothing)`).
 
+`Some` is also used in broadcasting to treat it's enclosed value as a scalar.
```
tkf, line 9: Maybe more accurate to say

```suggestion
`Some` is also used in broadcasting to broadcast a single value to all axes.
```

? Though I see that a similar wording is already used in `Ref`.

tkf:
We can also replace `Ref` to `Some` in:

https://github.com/JuliaLang/julia/blob/3da2c52f4af0931e0dac7c2bbd579d340d260810/base/broadcast.jl#L678-L679

base/some.jl
MasonProtter: I feel like that suggestion is less likely to be understood. However, maybe not everyone knows the word "scalar" either.

tkf: Maybe it's fine as-is if we also explain the actual interface it implements (i.e., it is a 0-dim iterable that supports indexing).

base/some.jl
MasonProtter: Good idea, thanks!

c42f:
This is interesting and promising. I can't say I love `Some` as a word for "protect something from broadcasting", but I also don't think it's worse than `Ref` and certainly less ambiguous than `Scalar`.

Also I think it's very reasonable to have this behavior in `Some` if we can have a separate and more intuitive name for "protecting a thing from unwrapping". Broadcasting isn't the only case one might want to wrap something to protect it. With that in mind, how about defining

```julia
wrap(x) = Some(x)
```

?

c42f:
Thinking a bit more about this, I feel it's unfortunate because while `Some` has the right semantics, it has a somewhat poor name outside its original use case of wrapping `nothing`. More logical would be to have `Wrapper{T}`, and to have functions `wrap` / `unwrap`.

Then the obvious question is - what specific cases other than broadcast would benefit from such a completely generic wrapper facility? Would there be utility in user-defined types being able to overload `wrap` and `unwrap`?

MasonProtter:
Personally I like the name `Some` a lot more than I like `wrap` for this.

MasonProtter:
Not sure what's going wrong with the LinearAlgebra test here: https://build.julialang.org/#/builders/148/builds/7606 anyone got ideas?

MasonProtter:
Accidentally deleted the wrong comment, oops. It was something along the lines of 
>I prefer `Some` to `wrap` here, especially because `wrap` and `unwrap` are likely to be used out in the wild.

MasonProtter:
There's some discussion related to this [on the Zulip](https://julialang.zulipchat.com/#narrow/stream/137791-general/topic/.60Ref.60.20in.20broadcasting) right now. 

Currently, people are thinking aloud about the possibility of improving `eachslice` to work on multiple dimensions, which would naturally include the behaviour of `Ref`. 

The idea would be that `eachslice(A, dims=(2, 3))` would act like a lazy version of 
```julia
[view(A, :, i, j, :, ..., :) for i in axes(A, 2), for j in axes(A, 3)]
```
Consequently, `eachslice(A, dims=())` would naturally be a zero dimensional container enclosing `A`, i.e. `eachslice(A, dims=())` would essentially be `ImmutableRef(A)`.

This also opens up some possibilities for syntax sugar. For instance, there were suggestions along the lines of
```julia
&(A) == eachslice(A, dims=())
```
and
```julia
A[i, j, &, k] == eachslice(A, dims=(1, 2, 4))[i, j, k]
```
I'm not totally sure how I feel about this, but I think it's an interesting alternative avenue that could both solve this problem, and improve our `eachslice`.

mbauman:
I don't quite follow.  The point is to make a non-indexable object work in a broadcast expression by putting it in a 0-dimensional container that has the indexing semantics we need.  The extension of that definition of `EachSlice` there would imply using a `view(x)` as the 0-dimensional container... but views require indexability of `x` in the first place.  Views always re-index into their parent.  For that matter, I think I'd expect `eachslice` to slice into its argument, too, regardless of its implementation.

tkf:
> Would there be utility in user-defined types being able to overload `wrap` and `unwrap`?

I don't think this is what we want. You'd need something that is guaranteed to "warp" something. Letting users to overload `wrap` makes it harder for you to ensure that something is really wrapped at the use-site. I think this is why we should prefer constructor rather than factory.

> Consequently, `eachslice(A, dims=())` would naturally be a zero dimensional container enclosing `A`,

`eachslice` is kind of interesting but I don't think it's super straightforward. For `eachslice(A; dims=())` to work, `view(A)` hence `A[]` has to be already defined.  Maybe we can define `view(A) = A` so that this is well-defined.

---

What we need is the concept of 0-dimensional array but the problem is that there is no common word for this (unlike vector and matrix).  Looking at [Empty product - Wikipedia](https://en.wikipedia.org/wiki/Empty_product), the closest thing I can find is `nullary` (as `[]` operator becomes a nullary function).  Not sure if it makes sense, though.

MasonProtter:
> I don't quite follow. The point is to make a non-indexable object work in a broadcast expression by putting it in a 0-dimensional container that has the indexing semantics we need. The extension of that definition of `EachSlice` there would imply using a `view(x)` as the 0-dimensional container... but views require indexability of `x` in the first place. Views always re-index into their parent. For that matter, I think I'd expect `eachslice` to slice into its argument, too, regardless of its implementation.

`eachslice(A, dims=())` would not do a `view(A)`, it would be 'like' `ZeroDimContainer(view(A, :, ..., : ))`, except instead of returning `nothing` it'd be a zero dimensional lazy container.

base/some.jl
yurivish: This should say `its` rather than `it's` in the original formulation.

tkf:
> It would involve `view(A, :, ..., :)`

Ah, you are right. But the main point still holds, right? The input `A` has to satisfy the array API for this to work.

MasonProtter:
Yes that's true, though there's nothing stopping us from having `EachSlice(A, dims=())` never take a `view` because it's never necessary to index into it.

To be clear, I don't think there's much point of using `eachslice` for this unless we had syntax sugar for it that removed the need to actually type `eachslice` when what you really want is some 0 dimensional stack allocated container.

base/some.jl
MasonProtter: Thanks Yuri!

tkf:
> there's nothing stopping us from having `EachSlice(A, dims=())` never take a `view`

Well, it'd increase the complexity of the implementation, right?  If you don't need to special case `dims == ()`, the implementation is much simpler:

```julia
eachslice(A; dims) =
    (view(A, f(A, dims, inds)...) for inds in Iterators.product(axes.((A,), dims)...))

function f(A, dims, indx)
    fullinds = ntuple(_ -> :, ndims(A))
    for (i, d) in zip(idx, dims)
        fullinds = setindex(fullinds, i, d)
    end
    return fullinds
end
```

MasonProtter:
> Well, it'd increase the complexity of the implementation, right? 
For an `EachIndex` that's parameterized on the `dims`, it'd be simple to just dispatch on the `dims` being an empty `Tuple`.

tkf:
That's the special case and complexity I'm talking about.  I think a dedicated type is much simpler.

c42f:
Just to record a couple of more speculative options mentioned on Zulip (https://julialang.zulipchat.com/#narrow/stream/137791-general/topic/.60Ref.60.20in.20broadcasting)

* `Box` was mentioned as a potential alternative name for the immutable version of `Ref`.  (We'd probably need to rename `Core.Box` to avoid confusion, but that seems like no great loss.)
* In a Julia 2.0 future where we had immutable array literals created from `[...]` syntax, `[x]` could mean the zero dimensional array of one element (with `[x,]` being a vector of one element).

MasonProtter:
I think `Box` is a really nice name for this if we can use it without too much trouble being caused by the existing `Core.Box`.

c42f:
> I think Box is a really nice name for this

Would it make sense for `Some` to become `Box` in the future? `something(Box(nothing))` would make some kind of sense, and `Box` seems an appropriately more generic name to have in Base than `Some`. `Some` is is great name for wrapping `nothing`, but a bit of a confusing name to be reused other purposes.

tkf:
I think ultimately everything is rather arbitrary except 0-dim immutable array literal (or maybe immutable `fill(x)`).

(Though TBH I'm saying this because I want `Some(x)[] === x`)

tkf:
Hmm... Can we use `const[x]` for 0-dim immutable array and `const[x,]` for immutable vector with one element?

JeffBezanson:
I think "box" is too overloaded, e.g. for "boxed values" in languages generally.

mbauman:
So the big objection to adding functionality to `Some` is that it's supposed to be well-discriminated from a value that you potentially might have.  It also doesn't quite seem to be the best word, and while it's the immutable struct we need, it serves a very different purpose — and that purpose necessitates holding it out so it's not confusable with other values.

Triage was largely in favor of using `Fill` — a la @dlfivefifty's FillArrays.jl — for this purpose.  The 0-dimensional `Fill` is _exactly_ what we want for broadcast, and linguistically it even dovetails on broadcast's "filling" behavior across the other arguments.  The extension to support arbitrary dimensions is so trivial (and so very useful with no overhead!) that it seems to make sense to also include that in Base's definition.

The challenge is that FillArrays.jl is a living and thriving package, and it'll be hard to meaningfully implement the necessary subset that is still functional and capable without some major piracy.  It really needs to be defined _before_ `base/broadcast.jl` in order to provide the [default `broadcastable`](https://github.com/JuliaLang/julia/blob/39fc4eec54fc97efda268c9975348121d0d225f5/base/broadcast.jl#L678-L679) for many builtin types, but FillArrays also provides lots of useful LinearAlgebra functionality that'll be hard to define so early.  One potential option would be to use a minimal (unexpected) struct for internal builtin use (it's literally 5 LOC), and bring in FillArrays as a stdlib... but we still want to export `Fill` from Base, so there'd still be some piracy going on and it'd still throw FillArrays into the stdlib tarpit.

Perhaps the best option is simply to define the minimal `struct` (including the `AbstractFill` supertype), and expect that FillArrays will pirate (with a seal of approval!) to gain more comprehensive LinearAlgebra functionality.

dlfivefifty:
I think the “best option” might indeed be the best option: that’s the most settled part of FillArrays.jl, where some of the other design choices (broadcasting, ` RectDiagonal `, ` Eye `) might need some deeper thought before making the cut for StdLib

PR #35779, opened 2020-05-06
streamline a[begin] lowering via firstindex(a, n)

stevengj:
Suggested by @mbauman in https://github.com/JuliaLang/julia/pull/33946#discussion_r420433172


mbauman:
Thanks!  I do fear this is a "minor change" as it requires non-`AbstractArray`s that define `axes` to now also implement `firstindex`... but they also need to implement `lastindex` so I don't think it's that big a deal.

StefanKarpinski:
Would it make sense to include in 1.5 since we haven't branched yet?

mbauman:
The `A[begin]` lowering was added in 1.4 so it's a change either way.

mbauman:
I think this is a worthwhile change (and I'd bet will be unobjectionable to triage), just could use NEWS.md.

stevengj:
Should be good to merge?

NEWS.md
```diff
@@ -44,6 +44,9 @@ Language changes
   a function like other operators. The dotted version `.-->` is now parsed as well.
   For backwards compatibility, `-->` still parses using its own expression head
   instead of `:call`.
+* Instances of `UniformScaling` are no longer `isequal` to matrices. Previous
```
JeffBezanson, line 47: This item should be removed; rebasing artifact I imagine.

JeffBezanson (approved):

NEWS.md
```diff
@@ -44,6 +44,9 @@ Language changes
...
+  behaviour violated the rule that `isequal(x, y)` implies `hash(x) == hash(y)`.
```
mbauman, line 48:
```suggestion
```

stevengj:
linuxaarch64 is the usual unrelated `NaN == Inf` error that we've been seeing recently on CI, so this should be good to merge once the other tests are green.

nalimilan:
FWIW this technically breaks some packages like DataFrames which implemented `axes` but not `firstindex`. That's probably OK for since we can fix it soon, but it might be worth adding a fallback `firstindex` definition that calls `axes` to minimize breakage.

Cc: @bkamins

stevengj:
For that matter, a fallback `lastindex` seems worthwhile too.

StefanKarpinski:
Better to be consistent here and there's no harm since it's a fallback anyway.

bkamins:
I think it would be best to provide a fallback in Julia Base.

The problem is that the `firstindex` and `lastindex` interface https://docs.julialang.org/en/v1/manual/interfaces/ is now specified only for `firstindex(collection)` and `lastindex(collection)` (without the dimension).

If you agree I would recommend that the interface specification should be updated in the manual in a PR that implements the fallback.

stevengj:
Should be fixed by #38742.

PR #35780, opened 2020-05-06
RFC: eliminate Type{Union{}} as a source of invalidations and ambiguity

timholy:
Here's a trial balloon for reducing the number of invalidations. The logic here is that any method that ends up with a `::Type{Union{}}` is probably just junk. Yes, we have a specific error-throwing method for `convert(Union{}, x)`, but
```julia
julia> similar(rand(2,2), Union{})
2×2 Array{Union{},2}:
 #undef  #undef
 #undef  #undef
```
clearly does not have an error-throwing method defined. I would guess that almost all such signatures would be generated from intersections of `Type{<:A}` and `Type{<:B}` where `typeintersect(A, B) == Union{}`.

In terms of practical results, this achieves about half the temporal benefit of #35733, and all the remaining invalidations come from the `reduce_empty` style signatures (and thus perhaps we could keep that part of #35733).

Alternatively, one could also check whether the result was ambiguous. That seems likely to be safe (aka, more correct) but slower.

### DIFF
--- a/src/subtype.c
+++ b/src/subtype.c
@@ -3177,7 +3177,28 @@ JL_DLLEXPORT jl_value_t *jl_intersect_types(jl_value_t *x, jl_value_t *y)
 // TODO: this can probably be done more efficiently
 JL_DLLEXPORT int jl_has_empty_intersection(jl_value_t *x, jl_value_t *y)
 {
-    return intersect_types(x, y, 1) == jl_bottom_type;
+    jl_value_t *res = intersect_types(x, y, 1);
+    if (res == jl_bottom_type)
+        return 1;
+    if (jl_is_tuple_type(res)) {
+        // Check for Type{Union{}}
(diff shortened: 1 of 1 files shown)


tkf:
`Union{}` is a very important element type for empty containers for implementing `collect`/`map`-like functions using mutate-or-widen strategy.  What happens to `Union{}[]`, `empty(x, Union{})`, `Vector{Union{}}`, `Dict{Union{},Union{}}` etc. after this PR?

JeffBezanson:
Let me try to dig up and rebase the code I have that uses the ambiguity checking approach. This is an interesting experiment but it's not actually sound, so we probably can't use it.

timholy:
I kinda figured that. I just added a check of the parents to check that neither parent has `Type{Union{}}` in the same slot, if that makes it any safer.

PR #35781, opened 2020-05-07
Simple tests for IdSet

kshyatt:
Appears to be [missing](https://codecov.io/gh/julialang/julia/src/master/base/idset.jl) tests.

### DIFF
--- a/test/sets.jl
+++ b/test/sets.jl
@@ -756,3 +756,24 @@ Base.IteratorSize(::Type{<:OpenInterval}) = Base.SizeUnknown()
     @test 3 ∈ i
     @test issubset(3, i)
 end
+
+@testset "IdSet" begin
+    a = [1]
+    b = [2]
+    c = [3]
+    d = [4]
+    A = Base.IdSet{Vector{Int}}([a, b, c, d])
+    @test !isempty(A)
+    B = copy(A)
+    @test A ⊆ B
+    @test B ⊆ A
+    A = filter!(x->isodd(x[1]), A)
+    @test A ⊆ B
+    @test !(B ⊆ A)
+    @test !isempty(A)
+    a_ = pop!(A, a)
+    @test a_ === a
+    @test !isempty(A)
+    A = empty!(A)
+    @test isempty(A)
+end


test/sets.jl
```diff
@@ -709,3 +709,20 @@ Base.IteratorSize(::Type{<:OpenInterval}) = Base.SizeUnknown()
...
+    A = filter!(isodd, A)
+    @test A ⊆ B
+    @test !(B ⊆ A)
+    @test !isempty(A)
+    a = pop!(A, 1)
+    @test a == 1
+    @test !isempty(A)
+    A = empty!(A)
+    @test isempty(A)
+end
```
JeffBezanson, line 728: Would be good to test that elements are compared with `===`, e.g. using an `IdSet{Vector{Int}}`.

kshyatt: Better now?

kshyatt:
pinging again gently for a second review :)

test/sets.jl
timholy: Maybe something like
```julia
julia> s = Base.IdSet{Vector{Int}}([a, b])
Base.IdSet{Array{Int64,1}} with 2 elements:
  [1]
  [2]

julia> a ∈ s
true

julia> c = [1]
1-element Array{Int64,1}:
 1

julia> c ∈ s
false
```

Keno: Would probably be good to also add something that is `== a`, but not `=== a` to test that it's not just a regular `Dict`.

fingolfin: To make such a test, I guess one could e.g. use `Vector{Int}` keys:
```julia
julia> dict = IdDict([0] => 1, [0] => 2)
IdDict{Array{Int64,1},Int64} with 2 entries:
  [0] => 1
  [0] => 2
```

fingolfin: That said, this PR looks like a nice improvement, so even without that, it'd be nice to merge it (and improve it later?) Depending on whether @kshyatt still has any interest on "finishing" this or not, it'd be a shame to let this go to waste?

kshyatt:
I'm still interested :). When I get a free moment I'll update it.

StefanKarpinski:
Anything to be done here? Does this still need review or can we just merge it?

kshyatt:
Updated this PR. I think we can just merge it as is if it passes tests and make further improvements later.

kshyatt:
Got the ok from @StefanKarpinski to merge

PR #35791, opened 2020-05-07
broaden/fix signature of nextprod

stevengj:
I noticed that the first argument of `nextprod` was too narrow (`Vector{Int}`) whereas the second argument was too broad (`::Any`, whereas the definition requires `::Real` similar to `nextpow`).

This PR broadens the signature of the first argument to tuples and abstractvectors of `<:Integer`, while narrowing the second argument to `::Real` for consistency with `nextpow` and integers factor.   (The old method couldn't work for any non-`Real` type anyway since the factors were `Vector{Int}`, so this shouldn't be breaking.)


stevengj:
CI failures seem unrelated (`UndefVarError: @text not defined`).

base/combinatorics.jl
```diff
@@ -290,27 +290,27 @@ invperm(P::Any16) = Tuple(invperm(collect(P)))
...
 # Examples
 ```jldoctest
-julia> nextprod([2, 3], 105)
+julia> nextprod((2, 3), 105)
 108
 
 julia> 2^2 * 3^3
 108
 ```
 """
```
ararslan, line 306:
```suggestion

!!! compat "Julia 1.6"
    The method that accepts a tuple requires Julia 1.6 or later.
"""
```

Issue #32906, opened 2019-08-15
Warn when a package loads a dependencies that already got loaded and that dependency has a different version than in the current manifest.


KristofferC:
Imagine the following scenario: A user starts Julia in Juno (which loads a bunch of dependencies into Julia at startup). They then activate their project and loads a package. If that package shares any dependencies with Juno, the versions of the dependencies Juno used will be used for the package. These versions might be completely incompatible with the current project and can lead to silent errors in the code running.

Loading packages, changing project, and then loading more packages always have this risk but in the case where packages are loaded at startup, it might be hard to realize it is happening.

It might, therefore, make sense to warn in the case where a package tries to load a dependency and code loading sees that the dependency with that UUID is already loaded but at a different version than in the current project.

It could be argued that anyone that wants real reproducibility should run with a custom `LOAD_PATH` to only have the current project as the entry but I think that is quite rare in practice.

davidanthoff:
Just as a FYI, this is actually the reason why we don't load _any_ packages into the REPL process in the VS Code extension as a general policy.

Seelengrab:
This has occured in the wild now, with a custom `startup.jl` that's loading & `using JuliaFormatter` before loading `CSV` (version `0.8.5`, requiring `Parsers@v1.1.2` at the maximum), which dies during precompilation due to not being compatible with the (by then) precompiled `Parsers@v2`.

For more information, see [this](https://discourse.julialang.org/t/reverting-back-to-avoid-precompile-errors/67354/20?u=sukera) thread on discourse.

Seelengrab:
[Another thread](https://discourse.julialang.org/t/issues-with-csv-in-julia-1-6-2/67507/16), this time with IJulia. The issue seems to happen in Atom as well, which makes sense if we think about how both of them work with code loading (being intrusive in the environments and all that).

Seelengrab:
[...]

PR #35794, opened 2020-05-07
RFC: Per-project sysimage

tkf:
With this patch, you can do

```
(@v1.x) pkg> activate .

julia> Base.set_sysimage_path("PATH/TO/sys.so")

julia> exit()

$ julia --project=.  # implicitly use -J=PATH/TO/sys.so
```

To use `julia` with project-specific system image.

### Implementation/design

Given `--project=$projectdir`, `julia` binary tries to read the text file `$projectdir/.julia/sysimages/$slug.path` and use its content as the default `--sysimage` argument.  The `$slug` is computed from the path of `julia` binary.  This way, it is safe to use multiple `julia` binaries with the same project.

(Edit: it was pointed out that including `VERSION` in `slug` as well might be a better idea https://github.com/JuliaLang/Pkg.jl/issues/2008#issuecomment-687847827)

This PR does not store the system image directly at `$projectdir/.julia/sysimages/$slug.$dlext` because the system image is rather large and it is nice to be able to use the same system image in multiple projects without copying the image file itself.  Furthermore, it may make sense to distribute pre-compiled system image using the artifacts mechanism.  It then is desirable to load system image directly from artifacts datastore rather than in `$projectdir/.julia/sysimages`.  Note that symbolic link is not a good option for supporting Windows.

@staticfloat suggested to use triplet instead of `$slug` https://github.com/JuliaLang/julia/issues/33973#issuecomment-601505588 but this means to re-implement triplet detection in C. Furthermore, I don't think it is enough for supporting other variations of `julia` (e.g., debug build, different versions). Hashing the path of `julia` binary seems to be a simple robust solution.

---

If the design is good, I can add some tests.  Also, since I'm not a C programmer, I may be doing something stupid.  Let me know if there is a better way to implement this.

### DIFF
--- a/base/loading.jl
+++ b/base/loading.jl
@@ -1495,6 +1495,50 @@ function stale_cachefile(modpath::String, cachefile::String)
     end
 end
 
+function _sysimage_path_slug()
+    ptr = @ccall jl_sysimage_path_slug()::Ptr{UInt8}
+    if ptr === C_NULL
+        error("`jl_sysimage_path_slug` failed")
+    end
+    slug = unsafe_string(ptr)
+    Libc.free(ptr)
+    return slug
+end
+
+"""
+    Base.project_sysimage_path(project::AbstractString = Base.active_project())
+
+A path to the file that stores the system image path for `project` that would be
+used by default.
+"""
+function project_sysimage_path(project::AbstractString = active_project())
+    if !isfile(project)
+        error("Non-existing project file: $project")
+    end
+    slug = _sysimage_path_slug()
+    return joinpath(dirname(project), ".julia", "sysimages", slug * ".path")
+end
+# This definition of path must be matched with `jl_resolve_sysimg_location`
+# in `../src/init.c`.
+
+"""
+    Base.set_sysimage_path([project::AbstractString,] sysimg::AbstractString)
+
+Use system image `sysimg` for project `project` by default, for this
+julia binary executable.  `project` defaults to currently active project.
+"""
+function set_sysimage_path end
+set_sysimage_path(sysimg::AbstractString) = set_sysimage_path(active_project(), sysimg)
+function set_sysimage_path(project::AbstractString, sysimg::AbstractString)
+    path = project_sysimage_path(project)
+    if !abspath(sysimg)
+        sysimg = relpath(sysimg, sysimg)
+    end
+    mkpath(dirname(path))
+    write(path, sysimg)
+    return
+end
+
 """
     @__FILE__ -> AbstractString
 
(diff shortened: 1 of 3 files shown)


fredrikekre:
Shouldn't this be recorded in the project file?

tkf:
I don't think it's a good idea to add it to project or manifest file because these files have to be portable. Whether or not you have a particular system image at particular path is machine-dependent. Also, it'd be a problem if you use the same project with multiple `julia` versions.

I'd like to treat system image as an optimization. Project environment should work with and without the system image. (Of course, there is a gray zone because the package included in the image have to be consistent with what is in the manifest. I think it'd be better to handle it in a more generic mechanism like #32906.)

Roger-luo:
I'm wondering how this should interact with binaries? should the system image binary also be shipped with the package manager instead of building them locally?

tkf:
> should the system image binary also be shipped with the package manager instead of building them locally?

I'm designing the API to make it easier (via, e.g., artifacts; see the OP) but I think making it actually work is challenging, as not all packages are relocatable.

davidanthoff:
The existing implementation in VS Code for this kind of functionality compares the last modified date of the `Manifest.toml` with the sysimage, to make sure users don't end up loading a sysimage file that is out-of-date with their project definition. I think that is a useful and user friendly feature, does that somehow fit in here?

tkf:
I think it'd better to solve it with a more generic mechanism like #32906.  Checking the modified time of `Manifest.toml` is not enough anyway since you can stack environments.  It's also trivial for a frontend (e.g., VS code) to emit a warning if the sysimage and `Manifest.toml` are incompatible in some sense.

This design is about maximal flexibility and extensibility so that, e.g., you can distribute sysimage with the artifact system.

oschulz:
> Checking the modified time of Manifest.toml is not enough anyway since you can stack environments.

Would it be possible to use a hash of the environment stack, or something similar?

oschulz:
Or maybe better a hash of the part of the environment stack that forms the transitive dependencies of the packages included in the sysimage?

vchuravy:
As an aside this could allow for upgradable standard libaries, since those are baked into the system image. With a per-project dependent sysimage one could feasible update a standard library. We would need to give stdlibs independent version numbers, but that seems like the smallest issue.

brenhinkeller:
Would fix https://github.com/JuliaLang/julia/issues/36840 I think

oschulz:
> Would fix #36840 I think

Yes, also @tkf's approach here is much more refined than my original suggestion in  #36840 .

Closes #36840 .

vchuravy:
I think pkgimages has mostly superseded this.

PR #35802, opened 2020-05-08
More robust installation of BugReporting package.

fredrikekre:


stdlib/InteractiveUtils/src/InteractiveUtils.jl
```diff
@@ -363,13 +363,14 @@ function report_bug(kind)
         let Pkg = Base.require(Base.PkgId(
             Base.UUID((0x44cfe95a_1eb2_52ea,0xb672_e2afdf69b78f)), "Pkg"))
             mktempdir() do tmp
-                prev_active = Base.ACTIVE_PROJECT[]
-                env_path = joinpath(tmp, "TmpForBugReporting")
-                Pkg.generate(env_path)
-                Pkg.activate(env_path)
+                old_load_path = copy(LOAD_PATH)
+                push!(empty!(LOAD_PATH), joinpath(tmp, "Project.toml"))
```
jonas-schulze, line 367: Some project files might be called `JuliaProject.toml`. IIRC there is some machinery to detect the project file.

fredrikekre: Yea but this is in a newly created tmp folder so won't matter.

PR #35803, opened 2020-05-08
Eliminate some invalidations from loading FixedPointNumbers [tweaked #35733]

KristofferC:
https://github.com/JuliaLang/julia/pull/35733 but with the controversial `::Union{}` methods removed. Still provides a significant speedup in the test case posted in that PR.

No impact on sysimage size.

### DIFF
--- a/base/bool.jl
+++ b/base/bool.jl
@@ -3,6 +3,8 @@
 # promote Bool to any other numeric type
 promote_rule(::Type{Bool}, ::Type{T}) where {T<:Number} = T
 
+convert(::Type{Bool}, x::Bool) = x   # prevent invalidation
+
 typemin(::Type{Bool}) = false
 typemax(::Type{Bool}) = true
 
--- a/base/char.jl
+++ b/base/char.jl
@@ -49,6 +49,7 @@ Char
 (::Type{AbstractChar})(x::Number) = Char(x)
 (::Type{T})(x::AbstractChar) where {T<:Union{Number,AbstractChar}} = T(codepoint(x))
 (::Type{T})(x::T) where {T<:AbstractChar} = x
+AbstractChar(x::AbstractChar) = x
 
 """
     ncodeunits(c::Char) -> Int
@@ -179,6 +180,7 @@ convert(::Type{AbstractChar}, x::Number) = Char(x) # default to Char
 convert(::Type{T}, x::Number) where {T<:AbstractChar} = T(x)
 convert(::Type{T}, x::AbstractChar) where {T<:Number} = T(x)
 convert(::Type{T}, c::AbstractChar) where {T<:AbstractChar} = T(c)
+convert(::Type{AbstractChar}, c::AbstractChar) = c
 convert(::Type{T}, c::T) where {T<:AbstractChar} = c
 
 rem(x::AbstractChar, ::Type{T}) where {T<:Number} = rem(codepoint(x), T)
--- a/base/reduce.jl
+++ b/base/reduce.jl
@@ -306,25 +306,29 @@ with reduction `op` over an empty array with element type of `T`.
 
 If not defined, this will throw an `ArgumentError`.
 """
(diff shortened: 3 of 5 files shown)


base/bool.jl
```diff
@@ -3,6 +3,8 @@
 # promote Bool to any other numeric type
 promote_rule(::Type{Bool}, ::Type{T}) where {T<:Number} = T
 
+convert(::Type{Bool}, x::Bool) = x   # prevent invalidation
```
timholy, line 6: This one might not be necessary either, after the `Union{}` stuff is fixed. It's just a really common case that would otherwise fall back to `convert(::Type{T}, x::T)`.

base/char.jl
```diff
@@ -49,6 +49,7 @@ Char
 (::Type{AbstractChar})(x::Number) = Char(x)
 (::Type{T})(x::AbstractChar) where {T<:Union{Number,AbstractChar}} = T(codepoint(x))
 (::Type{T})(x::T) where {T<:AbstractChar} = x
+AbstractChar(x::AbstractChar) = x
```
timholy, line 52: Possibly ditto.

base/char.jl
```diff
@@ -179,6 +180,7 @@ convert(::Type{AbstractChar}, x::Number) = Char(x) # default to Char
 convert(::Type{T}, x::Number) where {T<:AbstractChar} = T(x)
 convert(::Type{T}, x::AbstractChar) where {T<:Number} = T(x)
 convert(::Type{T}, c::AbstractChar) where {T<:AbstractChar} = T(c)
+convert(::Type{AbstractChar}, c::AbstractChar) = c
```
timholy, line 183: ...possibly same for everything up to the `reduce_empty` stuff.

KristofferC:
Why is Statistics testing

```
julia> mean(())
ERROR: MethodError: reduce_empty(::typeof(Base.add_sum), ::Type{Union{}}) is ambiguous. Candidates:
  reduce_empty(::typeof(Base.add_sum), ::Type{T}) where T<:Union{Int16, Int32, Int8} in Base at reduce.jl:314
  reduce_empty(::typeof(Base.add_sum), ::Type{T}) where T<:Union{UInt16, UInt32, UInt8} in Base at reduce.jl:315
Possible fix, define
  reduce_empty(::typeof(Base.add_sum), ::Type{Union{}})
```
?

base/reduce.jl
```diff
@@ -306,25 +306,29 @@ with reduction `op` over an empty array with element type of `T`.
 
 If not defined, this will throw an `ArgumentError`.
 """
-reduce_empty(op, T) = _empty_reduce_error()
-reduce_empty(::typeof(+), T) = zero(T)
+reduce_empty(op, ::Type{T}) where T = _empty_reduce_error()
```
timholy, line 309: AFAICT this is not a signature-based invalidation, this is one that came from `jl_method_table_disable` which seems to have `Base.delete_method` as its only caller. I don't have a very good handle on what is going on here, sorry.

One interesting thing is that the inferred return type is `Union{}` (even before loading FixedPointNumbers):
```julia
julia> code_typed(Base.reduce_empty, (Base.BottomRF{typeof(max)}, Type{VersionNumber}); optimize=false)
1-element Array{Any,1}:
 CodeInfo(
1 ─ %1 = Base.getproperty(op, :rf)::Core.Compiler.Const(max, false)
│        Base.reduce_empty(%1, T)::Union{}
└──      Core.Compiler.Const(:(return %2), false)::Union{}
) => Union{}
```

timholy (commented):
I suspect that for this particular PR we will be best waiting until Jeff or Jameson weighs in...but FWIW, here are some thoughts.

KristofferC:
I thought this PR would be more trivial than it was. Now when it starts generating discussion I think it is best to just close it to keep the focus in https://github.com/JuliaLang/julia/pull/35733.

PR #35806, opened 2020-05-08
add a test that error can be inferred to bottom

KristofferC:
Test that we don't break https://github.com/JuliaLang/julia/blob/ddf7ce9a595b0c84fbed1a42e8c987a9fdcddaac/base/compiler/params.jl#L20-L21 in Base + stdlibs at least.

Ref https://github.com/JuliaLang/Pkg.jl/pull/1817

Maybe this shouldn't be a generic method if it has these constraints?

### DIFF
--- a/test/compiler/inference.jl
+++ b/test/compiler/inference.jl
@@ -2989,3 +2989,6 @@ f38888() = S38888(Base.inferencebarrier(3))
 @test f38888() isa S38888
 g38888() = S38888(Base.inferencebarrier(3), nothing)
 @test g38888() isa S38888
+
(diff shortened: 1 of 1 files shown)


StefanKarpinski:
If we had the ability to constrain the overall type signature of a generic function, then we could just declare `error` to be a function whose return type is `Union{}` which would address this without forcing it to be non-generic. Of course, that's not a feature we have and while adding the feature would be non-breaking, adding signatures to pre-existing functions would be breaking.

Keno (approved):
LGTM, but needs a rebase.

PR #35816, opened 2020-05-09
add sincospi

### LINKED ISSUE #35799: add sincospi
Looking at #35792, I think it would make sense to also have a function `sincospi`, which could have a specialized implementation over just `(sinpi(x), cospi(x))`. It does look like there is some better performance to be gained.

simeonschaub:
close #35799
ref #35792

### DIFF
--- a/test/math.jl
+++ b/test/math.jl
@@ -408,8 +408,11 @@ end
             @test sincosd(convert(T, 270))::fTsc === (  -one(fT), zero(fT) )
         end
 
-        @testset "sinpi and cospi" begin
-            for x = -3:0.3:3
+        @testset "$name" for (name, (sinpi, cospi)) in (
+            "sinpi and cospi" => (sinpi, cospi),
+            "sincospi" => (x->sincospi(x)[1], x->sincospi(x)[2])
+        )
+            @testset "pi * $x" for x = -3:0.3:3
                 @test sinpi(convert(T,x))::fT ≈ convert(fT,sin(pi*x)) atol=eps(pi*convert(fT,x))
                 @test cospi(convert(T,x))::fT ≈ convert(fT,cos(pi*x)) atol=eps(pi*convert(fT,x))
             end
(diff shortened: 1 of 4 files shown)


test/math.jl
```diff
@@ -462,20 +470,25 @@ end
 
 @testset "Irrational args to sinpi/cospi/sinc/cosc" begin
     for x in (pi, ℯ, Base.MathConstants.golden)
-        @test sinpi(x) ≈ Float64(sinpi(big(x)))
-        @test cospi(x) ≈ Float64(cospi(big(x)))
+        for (sinpi, cospi) in ((sinpi, cospi), (x->sincospi(x)[1], x->sincospi(x)[2]))
+            @test sinpi(x) ≈ Float64(sinpi(big(x)))
+            @test cospi(x) ≈ Float64(cospi(big(x)))
+            @test sinpi(complex(x, x)) ≈ Complex{Float64}(sinpi(complex(big(x), big(x))))
+            @test cospi(complex(x, x)) ≈ Complex{Float64}(cospi(complex(big(x), big(x))))
```
simeonschaub, line 477: These should really be tested a bit more

StefanKarpinski:
This is great! I'm not qualified to review though. @simonbyrne, would you be willing?

simonbyrne:
Can we factor out the range reduction so that `sinpi`/`cospi`/`sincospi` all use the same underlying function?

simeonschaub:
I thought about this as well, but `sinpi` reduces to [-1,1], whereas `cospi` does [0,1], so they work a bit differently. One could perhaps use the same reduction for `sinpi` and `sincospi`, but I think factoring that out might add more complexity than is actually saved. Or do you think the deduplication is worth that complexity?

simeonschaub:
@symonbyrne What kind of refactoring did you have in mind?

simeonschaub:
Bump. What should be done here?

simonbyrne (approved):

simeonschaub:
Does this mean this is good to go as-is?

simonbyrne:
Yes, just wanted to figure out why the tests were failing.

PR #34987, opened 2020-03-04
load keyword to at_macroexpand1

bramtayl:
Rebased version of #34895

PR #35818, opened 2020-05-09
at_macroinvoke for testing macros

bramtayl:
This is #34987 rebased. I will continue to rebase until the tests pass on all platforms. Gah.

### DIFF
--- a/base/expr.jl
+++ b/base/expr.jl
@@ -156,12 +156,32 @@ end
 
 
 """
-    @macroexpand1
+    @macroexpand1 code [load = true]
 
-Non recursive version of [`@macroexpand`](@ref).
+Non recursive version of [`@macroexpand`](@ref). If `load` is set to false, will not throw
+a `LoadError`.
+
+```jldoctest
+julia> @macroexpand1 @static "hello world"
+ERROR: LoadError: ArgumentError: invalid @static macro
+[...]
+
+julia> @macroexpand1 (@static "hello world") false
+ERROR: ArgumentError: invalid @static macro
+[...]
+```
 """
-macro macroexpand1(code)
-    return :(macroexpand($__module__, $(QuoteNode(code)), recursive=false))
+macro macroexpand1(code, load = true)
+    if load
+        return :(macroexpand($__module__, $(QuoteNode(code)), recursive=false))
+    else
+        code.head == :macrocall ||
+            ArgumentError("`@macroexpand code false` requires a macro call")
+        esc(Expr(:call, code.args[1],
+            # :quote to make sure the linenumbernode won't be scrubbed away
+            Expr(:quote, code.args[2]), __module__, code.args[3:end]...
+        ))
+    end
 end
 
 ## misc syntax ##
--- a/base/expr.jl
+++ b/base/expr.jl
@@ -179,12 +179,16 @@ ERROR: ArgumentError: invalid @static macro
 julia> @macroexpand1 @static "hello world"
 ERROR: LoadError: ArgumentError: invalid @static macro
 [...]
+
+julia> @macroinvoke @macroinvoke a + a
+ERROR: ArgumentError: `@macroinvoke` requires a macro call
 ```
 """
-macro macroinvoke(code)
+macro macroinvoke(code::Expr)
+    code.head == :macrocall ||
+            throw(ArgumentError("`@macroinvoke` requires a macro call"))
     esc(Expr(:call, code.args[1],
-        # :quote to make sure the linenumbernode won't be scrubbed away
-        Expr(:quote, code.args[2]), __module__, code.args[3:end]...
+        Meta.quot(code.args[2]), __module__, Meta.quot.(code.args[3:end])...
     ))
 end
 
--- a/base/expr.jl
+++ b/base/expr.jl
@@ -164,6 +164,30 @@ macro macroexpand1(code)
     return :(macroexpand($__module__, $(QuoteNode(code)), recursive=false))
 end
 
+"""
+    @macroinvoke code
+
+Apply a single macro, returning the transformed expression. Similar to
+[`@macroexpand1`](@ref), but requires a top-level macro and will not wrap errors in a
+`LoadError`.
+
+```jldoctest
+julia> @macroinvoke @static "hello world"
+ERROR: ArgumentError: invalid @static macro
+[...]
+
+julia> @macroexpand1 @static "hello world"
+ERROR: LoadError: ArgumentError: invalid @static macro
+[...]
+```
+"""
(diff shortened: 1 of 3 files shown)


JeffBezanson:
It seems much simpler to me just to unwrap the LoadError when needed, instead of having this option. This also partly duplicates the logic for how macros are invoked, which is asking for trouble if something about it ever changes.

JeffBezanson:
Another suggestion: you can now call macro expander functions directly with the syntax `var"@macro"(args...)`.

bramtayl:
Well, the purpose of this was to get a nice way to write a test for this line: https://codecov.io/gh/JuliaLang/julia/src/master/base/osutils.jl#L37

I had been using `var@macro` but that required getting the line number and the module too...Is there a way to unwrap a LoadError using `@test_throws`?

JeffBezanson:
You can do this:
```
@test_throws LoadError(@__FILE__,@__LINE__,ArgumentError("invalid @static macro")) @macroexpand1 (@static "")
```

bramtayl:
Oh cool. Still a little verbose though?

tkf:
I agree with @bramtayl. It'd be very useful for testing macros. I don't think it's ideal to users to know how to know the calling convention of macros (which is not a public API, I suppose) or unpack `LoadError`.

I think it'd be ideal if `load = false` behavior were the default. But that is a breaking change...

tkf:
Ref original discussion: https://github.com/JuliaLang/julia/pull/34895

JeffBezanson:
> I think it'd be ideal if `load = false` behavior were the default. But that is a breaking change...

I agree that would be better; `@macroexpand` is not really "loading" anything, and the location info is not super useful in this context.

But I really dislike the option. First, I don't think it's sound testing to have an option that makes something work differently to make it easier to test. You need to test the thing that is actually used normally. Second, it's weird that it changes the macro to require a single macro call. Maybe it should be a new macro instead, e.g. `@macrocall` or `@macroinvoke`.

bramtayl:
I'm 90% sure my original version in https://github.com/JuliaLang/julia/pull/34895 was a separate macro. I must have messed up the git history because I can't find where I called it. Anyway, a separate macro seems like it would work.

tkf:
> You need to test the thing that is actually used normally.

I agree with this. One way to pursue this direction may be to implement something like

```julia
macro exceptionof(ex)
    quote
        local err
        err = try
            $(esc(ex))
            nothing
        catch err
            Some(err)
        end
        err === nothing && error("exception not thrown")
        something(err)
    end
end

dueto(e) = x -> dueto(x, e)
dueto(x, e) = x == e
dueto(x, T::Type) = x isa T
dueto(::LoadError, ::Type{LoadError}) = true
dueto(x::LoadError, T::Type) = dueto(x.error, T)
# and so on
```

So that something like

```julia
@test (@exceptionof @macroexpand1 @mymacro(x)) |> dueto(ArgumentError)
```

would work. Or maybe even `@eval` instead of `@macroexpand1`, but it's appealing to make sure something blows up during macro expansion time.

bramtayl:
What about `@test_throws AnError a_thing`, which could expand to 

`@test_throws LoadError(@__FILE__,@__LINE__,AnError) a_thing`

tkf:
It does not work when `AnError` is a type.

bramtayl:
Ok, well I changed this to be a separate macro

base/expr.jl
```diff
@@ -164,6 +164,34 @@ macro macroexpand1(code)
...
+
+julia> @macroinvoke @macroinvoke a + a
+ERROR: ArgumentError: `@macroinvoke` requires a macro call
+```
+"""
+macro macroinvoke(code::Expr)
+    code.head == :macrocall ||
+            throw(ArgumentError("`@macroinvoke` requires a macro call"))
+    esc(Expr(:call, code.args[1],
+        Meta.quot(code.args[2]), __module__, Meta.quot.(code.args[3:end])...
```
JeffBezanson, line 191:
```suggestion
        QuoteNode(code.args[2]), __module__, QuoteNode.(code.args[3:end])...
```

JeffBezanson:
I can't shake the feeling that the real problem here is "unwrapping the LoadError in testing is too hard", and this seems like a very indirect solution to that.

tkf:
> "unwrapping the LoadError in testing is too hard"

I think it's rather "_finding_ the cause is too hard." Users shouldn't have to destructure the error manually to match it with something. It's especially true when you start considering something like `CompositeException`. That's why I suggested `dueto(err, T)` above.

base/expr.jl
bramtayl: I never can remember which one to use

bramtayl:
I'm sympathetic to the load error concerns. One option would be plan to remove the load error from `@macroexpand` in 2.0, and either unexport this or add a comment that it will deprecated in 2.0 to the docs?

JeffBezanson:
IIRC, we generally allow changing an exception type in a minor release, and I'm not sure how common it is to test errors from `@macroexpand` --- do packages tend to do that? There are other ways to do it, e.g. using `eval` or `lower`.

bramtayl:
I don't have strong options about which way we go, just hoping for a terse way to test macro-time errors

bramtayl:
Ok, I'll close this and make a new PR for unwrapping LoadError from `@macroexpand` when I get around to it

henriquebecker91:
Chiming in just to give a bit of context why this can bother someone. In my case, I create a new exception type `TimeoutError`, and I throw it from my solving method when I discover I violated a time limit parameter (these `throw_if_timeout_now` calls are scattered over all my code). In a loop inside a scientific experiment script, I have a try-catch that will basically ignore and carry on if a `TimeoutError` happens, because some instances of the problem are very hard, I just need to output a clear indication a timeout happened, and keep the execution, but if some other error happened I want to stop the machines and receive a notification about it so I can fix the problem.

I have a problem with the `LoadError` wrapping because: (1) I could have left the script executing for a long time and then be bitten by this behavior and lose time understanding where the `LoadError` come from (as this is not a `include`, `import`, `using` problem); (2) I use the [TimerOutputs.jl](https://github.com/KristofferC/TimerOutputs.jl) so many of my methods are entirely wrapped inside a `@timeit` so they are always measured when called, this makes the difference between `LoadError` or a `TimeoutError` being raised entirely arbitrary, if the time limit was violated inside a not timed method then it will be a `TimeoutError`, if it was timed then it will be a `LoadError` (wrapping a TimeoutError).

Maybe this is a kinda a extreme case, but this behavior seems like a possible source of non-deterministic able-to-stay-hidden-for-a-long-time bugs.

PR #35822, opened 2020-05-09
Markdown.plain: \ => \\

bramtayl:
I'm not 100% sure but I think that `Markdown.plain` needs to escape backslashes for round-trip-ability

### DIFF
--- a/stdlib/Markdown/test/runtests.jl
+++ b/stdlib/Markdown/test/runtests.jl
@@ -1138,3 +1138,5 @@ let m = Markdown.parse("---"), io = IOBuffer()
     show(io, "text/latex", m)
     @test String(take!(io)) == "\\rule{\\textwidth}{1pt}\n"
 end
+
+sprint(show, md"\\") == "\\\\\n"
--- a/stdlib/Markdown/src/render/plain.jl
+++ b/stdlib/Markdown/src/render/plain.jl
@@ -113,7 +113,7 @@ plaininline(io::IO, link::Link) = plaininline(io, "[", link.text, "](", link.url
 
 plaininline(io::IO, md::Image) = plaininline(io, "![", md.alt, "](", md.url, ")")
 
-plaininline(io::IO, s::AbstractString) = print(io, s)
+plaininline(io::IO, s::AbstractString) = print(io, replace(s, '\\' => "\\\\"))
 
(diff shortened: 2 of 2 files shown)


bramtayl:
Ref https://github.com/domluna/JuliaFormatter.jl/pull/231

bramtayl:
Ok, added a test

StefanKarpinski:
@MikeInnes, could you review?

stdlib/Markdown/test/runtests.jl
```diff
@@ -1138,3 +1138,5 @@ let m = Markdown.parse("---"), io = IOBuffer()
     show(io, "text/latex", m)
     @test String(take!(io)) == "\\rule{\\textwidth}{1pt}\n"
 end
+
+sprint(show, md"\\") == "\\\\\n"
```
ararslan, line 1142:
```suggestion
@test sprint(show, md"\\") == "\\\\\n"
```

bramtayl: Oops

MikeInnes:
Round-tripping is not currently a goal of the `plain` renderer; unfortunately there's a distinction here between rendering markdown _as plain text_ and rendering markdown _as markdown_ (that can be re-parsed without changes). In the first interpretation, rendering a single slash is correct (since that was expressed by the `\\`). In general it's OK for `plain` to take other liberties with markdown syntax for the sake of readability, just like all the other renderers.

```julia
julia> md"A\\*b*"
  A\b

julia> md"[ref](foo)"
  ref (foo)
```

Your options are:

(1) Re-interpret the `plain` renderer as a `markdown` renderer and always emit round-trippable syntax (as in this PR).
(2) Add a new `markdown` renderer distinct from `plain` (though it should be possible to reuse almost all of the logic).

(2) would make the subtlety clear in the code and give us a place to document it, as well as perhaps being a better method for `string`. It could also be done outside the Base, e.g. in JuliaFormatter, if we don't want to add the code here.

bramtayl:
Round-trip-ability is a goal of `show`, right? If so, it seems like it might be a good idea to add a markdown renderer to Base

bramtayl:
Ok, I'm going to close this an open a new issue for Markdown roundtripability

PR #35827, opened 2020-05-10
add an extra inbounds in nextind to prevent a gc frame

KristofferC:
Unless I miss something, this should be safe since the `@boundscheck` in `thisind`

https://github.com/JuliaLang/julia/blob/8f512f3f6d5d69403be4b8179bd80ba64229e761/base/strings/string.jl#L123

 is identical to the `@boundscheck` we just did.

https://github.com/JuliaLang/julia/blob/8f512f3f6d5d69403be4b8179bd80ba64229e761/base/strings/string.jl#L143

This avoids a GC frame and reduces the code size quite a bit (for some form of measurement, reduces the number of SSA values from 108 to 46).

### DIFF
--- a/base/strings/string.jl
+++ b/base/strings/string.jl
@@ -144,7 +144,7 @@ end
     @inbounds l = codeunit(s, i)
     (l < 0x80) | (0xf8 ≤ l) && return i+1
     if l < 0xc0
-        i′ = thisind(s, i)
+        i′ = @inbounds thisind(s, i)
         return i′ < i ? nextind(s, i′) : i+1
     end
     # first continuation byte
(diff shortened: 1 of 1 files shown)


base/strings/string.jl
```diff
@@ -144,7 +144,7 @@ end
     @inbounds l = codeunit(s, i)
     (l < 0x80) | (0xf8 ≤ l) && return i+1
     if l < 0xc0
-        i′ = thisind(s, i)
+        i′ = @inbounds thisind(s, i)
         return i′ < i ? nextind(s, i′) : i+1
```
Keno, line 148: Inbounds on the nextind also?

KristofferC: Yeah, thanks.

StefanKarpinski (approved):

StefanKarpinski:
I believe that when I wrote this I considered `@inbounds` and `@boundscheck` to only apply to whether the index was in a viable range, not that it pointed at a valid index. This is pretty much an internal mechanism, however, so we can change the meaning to whatever we want.

PR #35828, opened 2020-05-10
fix getting started link from documentation manual

blood0cean:


dkarrasch:
Thanks for fixing that, and welcome to the Julia contributors.

Issue #35757, opened 2020-05-05
Propagating context information to child-tasks and remote calls?


oschulz:
@tkf and me have been discussing ways to propagate information about available workers (or resources in general) in distributed hierarchical computations:

https://discourse.julialang.org/t/propagation-of-available-assigned-worker-ids-in-hierarchical-computations

Adding resource arguments to every function call would be impractical, and using Cassette & friends to add add context by rewriting the whole computation would be very heavy-handed, since it might tough large code stacks (and may also be not be a complete solution when remote calls are involved?).

Could we add something like a `context` field to `Task`, in addition to the `storage` field - with the difference that `context` is automatically propagated to spawned tasks and remote calls? Adding the possibility to propagate a context through a computation in this fashion could also be useful for other use cases too, I think.

vtjnash:
C.f. #34543

tkf:
Thanks for opening the issue.

I'm actually against copying `task_local_storage` API.  I think something similar to [`contextvars` in Python](https://docs.python.org/3/library/contextvars.html) (see [PEP 567](https://www.python.org/dev/peps/pep-0567/) for discussion) would be better:

```julia
const var1 = ContextVar(:var1, 42)  # with default
# possible sugar: @contextvar var1 = 42

const var2 = ContextVar{Lockable{Vector{Int}}}(:var2)  # without default, with type
# possible sugar: @contextvar var2::Lockable{Vector{Int}}
# Lockable from https://github.com/JuliaLang/julia/pull/34400

const var3 = ContextVar(:var3)  # equivalent to ContextVar{Any}(:var3)
# possible sugar: @contextvar var3


function f()
    @sync @async @show var1[] # var1[] => 42
    var1[] = 0
    @sync @async @show var1[] # var1[] => 0

    setting(var1 => 1) do
        @show var1[] # var1[] => 1
    end
    @show var1[] # var1[] => 0
end
```

In contrast to `task_local_storage(:var)`:

1. `var[]` can be inferred
2. `var` is forced to be namespaced (i.e., it has to exist in some module name space)
3. `var` can be backed up by an efficient concrete key type (e.g., UUID)
4. `var` allows small-size optimization when the value type can be inlined into the context storage

We may want to use something like HAMT (as in PEP 567) for the context storage.

---

A possible implementation of `ContextVar` may be something like

```julia
struct ContextVar{T}
    name::Symbol  # only for human readability
    key::UUID     # (for example)
[...]

Issue #32954, opened 2019-08-18
Export and document cryptographically secure random numbers


chethega:
Various applications need a source of cryptographically secure random numbers. We currently do not provide a CSPRNG.

Secure random is a very basic primitive, and imo as widely important as secure hash (which we do provide). Further, we have a very nice AbstractRNG interface. In other words, there are very few API decisions (mainly: whether to define `Random.seed!` for the CSPRNG). Secure random could live in packages, but this has lots of disadvantages: First, it is inconvenient, which will lead to people misusing the default random generator. Second, a central place for maintaining such basic functionality is preferable to many competing implementations, both security-wise and in terms of simplified audits of packages.

There was some discussion [here](https://github.com/JuliaLang/julia/issues/27614) that inspired me to grep through a lot of packages to gauge how they use randomness. I'll link all of the individual issues here, to make the point that this is important functionality, and that there is confusion about secure random in the package ecosystem. Considerations about the default RNG are separate.

A sensible solution could be to use `RandomDevice()`. In this case, we would simply update the docs.

JeffBezanson:
> Secure random could live in packages, but this has lots of disadvantages:

On the other hand, putting it in a package makes releasing bug fixes or improvements much faster.

ViralBShah:
To start with, it should certainly be a package. As it gains more widespread usage, we can figure out if it should live in stdlib. Given that we are more likely to move things out of stdlib rather than into it, having an external package seems better.

chethega:
I disagree. 

Security is still the toxic wastedump of computer science. People are rightfully reluctant of adding external dependencies. Many people are naive with respect to crypto and security. People lack the resources to properly vet packages that provide secure random. We want to support a culture of "when in doubt, then be secure". We absolutely do not want to force people to make ill-informed trade-offs whether they "can get away" with using insecure `rand`. It must be easy to do the right thing; otherwise people will invariably mess up. Bad random leads to perfectly functional test-passing vulnerable packages.

At the very least, I want the `Random` stdlib to document a best-practice for secure random. This is basic functionality and there must be an officially endorsed way of getting secure random.

If we do not want to implement a CSPRNG, then we could export a wrapper around one of the mbedtls CSPRNGs. mbedtls is already a binary dependency of julia.

[...]

PR #35833, opened 2020-05-11
RFC: Context variables

tkf:
This PR proposes *context variables* API that can be used to propagate context-dependent information across task boundaries.  It is conceptually similar to `task_local_storage` but with a main difference that it "copies" all key-value pairs to child `Task`s.  The context variables are conceptually similar to dynamically scoped variables.

## Motivations

There are several places context variables can be useful or required.

### `ENV`

It was proposed to use `task_local_storage` to fix thread-safety of `withenv` https://github.com/JuliaLang/julia/issues/34726#issuecomment-586138687 by maintaining task-local copy (or overlay) of `ENV`.  However, it would mean that the environment variables cannot cross task boundaries.  Context variable can fix this shortcoming.

### `@testset`

`@testset` uses `task_local_storage` to track current active test set.  However, following example does not work (prints `No tests`) because the information of the current test does not propagate to the child task:

```julia
@testset begin
    @sync @async @test true
end
```

### Logging

`Task` has the `logstate` field that propagates to child tasks; i.e., it works as a hard-coded context variable.  Once context variable handling is sufficiently matured, it may be possible to eliminate the special handling from `Task` and use a context variable for `Task`.  Furthermore, context variable allows users to develop logging-like interfaces.

### Custom worker pool abstraction

In #35757 and [Propagation of available/assigned worker-IDs in hierarchical computations? - Domains / Julia at Scale - JuliaLang](https://discourse.julialang.org/t/propagation-of-available-assigned-worker-ids-in-hierarchical-computations/38670), it was brought up that propagating "computation resources" (thread/process pools, etc.) across tasks and processes is required for implementing custom worker pool interfaces.

### Misc.

> Request-related data, such as security tokens and request data in web applications, language context for gettext, etc.
>
> --- [Rationale section in PEP 0567](https://www.python.org/dev/peps/pep-0567/#rationale)

It would be useful for implementing a better nestable progress information handling https://github.com/JunoLab/ProgressLogging.jl/pull/13#discussion_r371057967.

### Implicit context

* Python has [`contextvars`](https://docs.python.org/3/library/contextvars.html). See [PEP 567](https://www.python.org/dev/peps/pep-0567/) and [PEP 550](https://www.python.org/dev/peps/pep-0567/) for discussion.
* .NET has [`ExecutionContext`](https://docs.microsoft.com/en-us/dotnet/api/system.threading.executioncontext?view=netcore-3.1).
* Kotlin has [`CoroutineContext`](https://kotlinlang.org/docs/reference/coroutines/coroutine-context-and-dispatchers.html).

### Explicit context

* Go has [`context`](https://blog.golang.org/context).

## Proposed design

I propose an API with the following basic usages.

* Declare context variable: `@contextvar x`
* Get the value: `x[]`
* Set the value: `x[] = value`

For a tutorial and the full reference API of the proposed design, see https://tkf.github.io/ContextVariablesX.jl/dev/

Internally, `@contextvar` creates an instance of `ContextVar` which is defined as

```julia
struct ContextVar{T}
    key::UUID
    default::T
    ...
end
```

Then `x[]` and `x[] = value` invoke `current_task().ctxvars[x.key]` and an "immutable version" of `current_task().ctxvars[x.key] = value`, respectively (roughly speaking).

This has a couple of nice properties:

1. `x[]` can be inferred
2. `x` is forced to be namespaced (i.e., it has to exist in some module name space.)
3. `x` can be backed up by an efficient concrete key type (e.g., UUID)
4. `x` allows small-size optimization when the value type can be inlined into the context storage (in principle)

See also https://github.com/JuliaLang/julia/issues/35757 which already contains some discussion on this API.

### DIFF
--- a/base/contextvariables.jl
+++ b/base/contextvariables.jl
@@ -1,0 +1,415 @@
+# This file is a part of Julia. License is MIT: https://julialang.org/license
+
+function _uuid4 end  # to be defined by UUIDs.jl
+
+function _ContextVar end
+
+# `ContextVar` object itself does not hold any data (except the
+# default value).  It is actually just a key into the task-local
+# context storage.
+"""
+    ContextVar{T}
+
+Context variable type.  This is the type of the object `var` created by
+[`@contextvar var`](@ref @contextvar).  This acts as a reference to the
+value stored in a task-local.  The macro `@contextvar` is the only public
+API to construct this object.
+
+!!! warning
+
+    It is unspecified if this type is concrete or not. It may be
+    changed to an abstract type and/or include more type parameters in
+    the future.
+"""
+struct ContextVar{T}
+    name::Symbol
+    _module::Module
+    key::UUID
+    has_default::Bool
+    default::T
+
+    global _ContextVar
+    _ContextVar(name, _module, key, ::Type{T}, default) where {T} =
+        new{T}(name, _module, key, true, default)
+    _ContextVar(name, _module, key, ::Type{T}) where {T} = new{T}(name, _module, key, false)
+end
+
+_ContextVar(name, _module, key, ::Nothing, default) =
+    _ContextVar(name, _module, key, typeof(default), default)
+_ContextVar(name, _module, key, ::Nothing) = _ContextVar(name, _module, key, Any)
+
+eltype(::Type{ContextVar{T}}) where {T} = T
+
+function show(io::IO, var::ContextVar)
+    print(io, ContextVar)
+    if eltype(var) !== Any && !(var.has_default && typeof(var.default) === eltype(var))
+        print(io, '{', eltype(var), '}')
+    end
+    print(io, '(', repr(var.name))
+    if var.has_default
+        print(io, ", ")
+        show(io, var.default)
+    end
+    print(io, ')')
+end
+
+function show(io::IO, ::MIME"text/plain", var::ContextVar)
+    print(io, var._module, '.', var.name, " :: ContextVar")
+    if get(io, :compact, false) === false
+        print(io, " [", var.key, ']')
+        if get(var) === nothing
+            print(io, " (not assigned)")
+        else
+            print(io, " => ")
+            show(IOContext(io, :compact => true), MIME"text/plain"(), var[])
+        end
+    end
+end
+
+# The primitives that can be monkey-patched to play with new context
+# variable storage types:
+"""
+    merge_ctxvars(ctx::Union{Nothing,T}, kvs) -> ctx′:: Union{Nothing,T}
+
+!!! warning
+
+    This is not a public API.  This documentation is for making it easier to
+    experiment with different implementations of the context variable storage
+    backend, by monkey-patching it at run-time.
+
+The first argument `ctx` is either `nothing` or a dict-like object of type `T` where
+its `keytype` is `UUID` and `valtype` is `Any`.  The second argument `kvs` is an
+iterable of `Pair{UUID,<:Union{Some,Nothing}}` values.  Iterable `kvs` must have
+length.
+
+If `ctx` is `nothing` and `kvs` is non-empty, `merge_ctxvars` creates a new
+instance of `T`. If `ctx` is not `nothing`, it returns a shallow-copy `ctx′` of
+`ctx` where `k => v` is inserted to `ctx′` for each `k => Some(v)` in `kvs`
+and `k` is deleted from `ctx′` for each `k => nothing` in `kvs`.
+"""
+function merge_ctxvars(ctx, kvs)
+    # Assumption: eltype(kvs) <: Pair{UUID,<:Union{Some,Nothing}}
+    if isempty(kvs)
+        return ctx
+    else
+        # Copy-or-create-on-write:
+        vars = ctx === nothing ? Dict{UUID,Any}() : copy(ctx)
+        for (k, v) in kvs
+            if v === nothing
+                delete!(vars, k)
+            else
+                vars[k] = something(v)
+            end
+        end
+        isempty(vars) && return nothing  # should we?
+        return vars
+    end
+end
+
+get_task_ctxvars(t = current_task()) = t.ctxvars
+set_task_ctxvars(t, ctx) = t.ctxvars = ctx
+set_task_ctxvars(ctx) = set_task_ctxvars(current_task(), ctx)
+
+struct _NoValue end
+
+"""
+    get(var::ContextVar{T}) -> Union{Some{T},Nothing}
+
+Return `Some(value)` if `value` is assigned to `var`.  Return `nothing` if
+unassigned.
+"""
+function get(var::ContextVar{T}) where {T}
+    ctx = get_task_ctxvars()
+    if ctx === nothing
+        var.has_default && return Some(var.default)
+        return nothing
+    end
+    if var.has_default
+        return Some(get(ctx, var.key, var.default)::T)
+    else
+        y = get(ctx, var.key, _NoValue())
+        y isa _NoValue || return Some(ctx[var.key]::T)
+    end
+    return nothing
+end
+
+"""
+    getindex(var::ContextVar{T}) -> value::T
+
+Return the `value` assigned to `var`.  Throw a `KeyError` if unassigned.
+"""
+function getindex(var::ContextVar{T}) where {T}
+    maybe = get(var)
+    maybe === nothing && throw(KeyError(var))
+    return something(maybe)::T
+end
+
+setindex!(var::ContextVar, value) = set!(var, Some(value))
+delete!(var::ContextVar) = set!(var, nothing)
+
+"""
+    set!(var::ContextVar, Some(value))
+    set!(var::ContextVar, nothing)
+
+Set the value of context variable `var` to `value` or delete it.
+"""
+function set!(var::ContextVar{T}, value::Union{Some,Nothing}) where {T}
+    value = convert(Union{Some{T},Nothing}, value)
+    set_task_ctxvars(merge_ctxvars(get_task_ctxvars(), (var.key => value,)))
+    return var
+end
+
+"""
+    genkey(__module__::Module, varname::Symbol) -> UUID
+
+Generate a stable UUID for a context variable `__module__.\$varname`.
+"""
+function genkey(__module__::Module, varname::Symbol)
+    pkgid = PkgId(__module__)
+    if pkgid.uuid === nothing
+        throw(ArgumentError(
+            "Module `$__module__` is not a part of a package. " *
+            "`@contextvar` can only be used inside a package.",
+        ))
+    end
+    fullpath = push!(collect(fullname(__module__)), varname)
+    if any(x -> contains(string(x), "."), fullpath)
+        throw(ArgumentError(
+            "Modules and variable names must not contain a dot:\n" * join(fullpath, "\n"),
+        ))
+    end
+    return uuid5(pkgid.uuid, join(fullpath, '.'))
+end
+
+"""
+    @contextvar [local|global] var[::T] [= default]
+
+Declare a context variable named `var`.
+
+!!! warning
+
+    Context variables declared with `global` does not work with `Distributed`.
+
+# Examples
+
+Top-level context variables needs to be declared in a package:
+
+```
+module MyPackage
+@contextvar cvar1
+@contextvar cvar2 = 1
+@contextvar cvar3::Int
+end
+```
+
+Context variables can be declared in local scope by using `local` prefix:
+
+```jldoctest
+julia> function demo()
+           @contextvar local x = 1
+           function f()
+               x[] += 1
+               return x[]
+           end
+           return f()
+       end;
+
+julia> demo()
+2
+```
+
+To use `@contextvar` in a non-package namespace like REPL, prefix the variable
+with `global`:
+
+```jldoctest global_vars
+julia> @contextvar global X;
+
+julia> X[] = 1;
+
+julia> X[]
+1
+```
+"""
+macro contextvar(ex0)
+    ex = ex0
+    qualifier = :const
+    if Meta.isexpr(ex, :local)
+        length(ex.args) != 1 && throw(ArgumentError("Malformed input:\n$ex0"))
+        ex, = ex.args
+        qualifier = :local
+    elseif Meta.isexpr(ex, :global)
+        length(ex.args) != 1 && throw(ArgumentError("Malformed input:\n$ex0"))
+        ex, = ex.args
+        qualifier = :global
+    end
+    if Meta.isexpr(ex, :(=))
+        length(ex.args) != 2 && throw(ArgumentError("Unsupported syntax:\n$ex0"))
+        ex, default = ex.args
+        args = Any[esc(default)]
+    else
+        args = []
+    end
+    if Meta.isexpr(ex, :(::))
+        length(ex.args) != 2 && throw(ArgumentError("Malformed input:\n$ex0"))
+        ex, vartype = ex.args
+        pushfirst!(args, esc(vartype))
+    else
+        pushfirst!(args, nothing)
+    end
+    if !(ex isa Symbol)
+        if ex === ex0
+            throw(ArgumentError("Unsupported syntax:\n$ex0"))
+        else
+            throw(ArgumentError("""
+                Not a variable name:
+                $ex
+                Input:
+                $ex0
+                """))
+        end
+    end
+    varname = QuoteNode(ex)
+    if qualifier === :const
+        key = genkey(__module__, ex)
+    else
+        # Creating a UUID at macro expansion time because:
+        # * It would be a memory leak because context variable storage can be
+        #   filled with UUIDs created at run-time.
+        # * Creating it at run-time is doable with function-based interface like
+        #   `ContextVar(:name, default)`.
+        key = _uuid4()
+    end
+    return Expr(
+        qualifier,
+        :($(esc(ex)) = _ContextVar($varname, $__module__, $key, $(args...))),
+    )
+end
+
+"""
+    with_context(f, var1 => value1, var2 => value2, ...)
+    with_context(f, pairs)
+
+Run `f` in a context with given values set to the context variables.  Variables
+specified in this form are rolled back to the original value when `with_context`
+returns.  It act like a dynamically scoped `let`.  If `nothing` is passed as
+a value, corresponding context variable is cleared; i.e., it is unassigned or
+takes the default value.  Use `Some(value)` to set `value` if `value` can be
+`nothing`.
+
+    with_context(f, nothing)
+
+Run `f` in a new empty context.  All variables are rewind to the original values
+when `with_context` returns.
+
+Note that
+
+```julia
+var2[] = value2
+with_context(var1 => value1) do
+    @show var2[]  # shows value2
+    var3[] = value3
+end
+@show var3[]  # shows value3
+```
+
+and
+
+```julia
+var2[] = value2
+with_context(nothing) do
+    var1[] = value1
+    @show var2[]  # shows default (or throws)
+    var3[] = value3
+end
+@show var3[]  # does not show value3
+```
+
+are not equivalent.
+"""
+function with_context(f, kvs::Pair{<:ContextVar}...)
+    orig = map(((k, _),) -> k => get(k), kvs)
+    set_context(kvs)
+    try
+        return f()
+    finally
+        set_context(orig)
+    end
+end
+
+function with_context(f, ::Nothing)
+    ctx = get_task_ctxvars()
+    try
+        set_task_ctxvars(nothing)
+        return f()
+    finally
+        set_task_ctxvars(ctx)
+    end
+end
+
+# Not using `set_context!` so that `snapshot` as an input makes sense.
+"""
+    set_context(var1 => value1, var2 => value2, ...)
+    set_context(pairs)
+    set_context(snapshot::ContextSnapshot)
+
+Equivalent to `var1[] = value1`, `var2[] = value2`, and so on.  The second form
+expect an iterable of pairs of context variable and values.  This function is
+more efficient than setting individual context variables sequentially.  Like
+[`with_context`](@ref), `nothing` value means to clear the context variable
+and `Some` is always unwrapped.
+
+A "snapshot" of the context returned from [`snapshot_context`](@ref) can
+be specified as an input (the third form).
+
+# Examples
+```jldoctest
+julia> @contextvar global x
+       @contextvar global y;
+
+julia> set_context(x => 1, y => 2)
+
+julia> x[]
+1
+
+julia> y[]
+2
+
+julia> set_context([x => 10, y => 20])
+
+julia> x[]
+10
+
+julia> y[]
+20
+```
+"""
+set_context(kvs::Pair{<:ContextVar}...) = set_context(kvs)
+function set_context(kvs)
+    ctx = merge_ctxvars(get_task_ctxvars(), (k.key => v for (k, v) in kvs))
+    set_task_ctxvars(ctx)
+    return
+end
+
+struct ContextSnapshot{T}
+    vars::T
+end
+
+# TODO: Do we need to implement `Dict{ContextVar}(::ContextSnapshot)`?
+# This requires storing UUID-to-ContextVar mapping somewhere.
+
+"""
+    snapshot_context() -> snapshot::ContextSnapshot
+
+Get a snapshot of a context that can be passed to [`reset_context`](@ref) to
+rewind all changes in the context variables.
+"""
+snapshot_context() = ContextSnapshot(get_task_ctxvars())
+
+"""
+    reset_context(snapshot::ContextSnapshot)
+
+Rest the entire context of the current task to the state at which `snapshot`
+is obtained via [`snapshot_context`](@ref).
+"""
+reset_context(snapshot::ContextSnapshot) = set_task_ctxvars(snapshot.vars)
+set_context(snapshot::ContextSnapshot) = set_context(snapshot.vars)
--- a/src/jltypes.c
+++ b/src/jltypes.c
@@ -2273,10 +2273,11 @@ void jl_init_types(void) JL_GC_DISABLED
                         NULL,
                         jl_any_type,
                         jl_emptysvec,
-                        jl_perm_symsvec(11,
+                        jl_perm_symsvec(12,
                                         "next",
                                         "queue",
                                         "storage",
+                                        "ctxvars",
                                         "state",
                                         "donenotify",
                                         "result",
@@ -2285,7 +2286,8 @@ void jl_init_types(void) JL_GC_DISABLED
                                         "logstate",
                                         "code",
                                         "sticky"),
-                        jl_svec(11,
+                        jl_svec(12,
+                                jl_any_type,
                                 jl_any_type,
                                 jl_any_type,
                                 jl_any_type,
@@ -2297,7 +2299,7 @@ void jl_init_types(void) JL_GC_DISABLED
                                 jl_any_type,
                                 jl_any_type,
                                 jl_bool_type),
-                        0, 1, 9);
+                        0, 1, 10);
     jl_value_t *listt = jl_new_struct(jl_uniontype_type, jl_task_type, jl_nothing_type);
     jl_svecset(jl_task_type->types, 0, listt);
 
--- a/stdlib/UUIDs/src/UUIDs.jl
+++ b/stdlib/UUIDs/src/UUIDs.jl
@@ -94,6 +94,8 @@ function uuid4(rng::AbstractRNG=Random.default_rng())
     UUID(u)
 end
 
+Base._uuid4(rng::AbstractRNG=Random.default_rng()) = uuid4(rng)
+
 """
     uuid5(ns::UUID, name::String) -> UUID
 
--- a/base/Base.jl
+++ b/base/Base.jl
@@ -224,6 +224,8 @@ include("lock.jl")
 include("channels.jl")
 include("task.jl")
 include("weakkeydict.jl")
+include("uuid.jl")
+include("contextvariables.jl")
 
 # Logging
 include("logging.jl")
@@ -332,7 +334,6 @@ include("initdefs.jl")
 include("threadcall.jl")
 
 # code loading
-include("uuid.jl")
 include("loading.jl")
 
 # misc useful functions & macros
--- a/base/exports.jl
+++ b/base/exports.jl
@@ -38,6 +38,7 @@ export
     ComplexF64,
     ComplexF32,
     ComplexF16,
+    ContextVar,
     DenseMatrix,
     DenseVecOrMat,
     DenseVector,
@@ -525,6 +526,7 @@ export
     mergewith,
     pairs,
     reduce,
+    set!,
     setdiff!,
     setdiff,
     setindex!,
@@ -672,6 +674,11 @@ export
     timedwait,
     asyncmap,
     asyncmap!,
+    @contextvar,
+    reset_context,
+    set_context,
+    snapshot_context,
+    with_context,
 
 # channels
     take!,
--- a/src/julia.h
+++ b/src/julia.h
@@ -1728,6 +1728,7 @@ typedef struct _jl_task_t {
     jl_value_t *next; // invasive linked list for scheduler
     jl_value_t *queue; // invasive linked list for scheduler
     jl_value_t *tls;
+    jl_value_t *ctxvars;
     jl_sym_t *state;
     jl_value_t *donenotify;
     jl_value_t *result;
--- a/src/task.c
+++ b/src/task.c
@@ -570,6 +570,7 @@ JL_DLLEXPORT jl_task_t *jl_new_task(jl_function_t *start, jl_value_t *completion
     t->next = jl_nothing;
     t->queue = jl_nothing;
     t->tls = jl_nothing;
+    t->ctxvars = ptls->current_task->ctxvars;  // inherit parent context
     t->state = runnable_sym;
     t->start = start;
     t->result = jl_nothing;
@@ -1093,6 +1094,7 @@ void jl_init_root_task(void *stack_lo, void *stack_hi)
     ptls->current_task->started = 1;
     ptls->current_task->next = jl_nothing;
     ptls->current_task->queue = jl_nothing;
+    ptls->current_task->ctxvars = jl_nothing;
     ptls->current_task->state = runnable_sym;
     ptls->current_task->start = NULL;
     ptls->current_task->result = jl_nothing;
--- a/test/choosetests.jl
+++ b/test/choosetests.jl
@@ -54,7 +54,8 @@ function choosetests(choices = [])
         "checked", "bitset", "floatfuncs", "precompile",
         "boundscheck", "error", "ambiguous", "cartesian", "osutils",
         "channels", "iostream", "secretbuffer", "specificity",
-        "reinterpretarray", "syntax", "logging", "missing", "asyncmap", "atexit"
+        "reinterpretarray", "syntax", "logging", "missing", "asyncmap", "atexit",
+        "contextvariables",
     ]
 
     tests = []
--- a/test/contextvariables.jl
+++ b/test/contextvariables.jl
@@ -1,0 +1,60 @@
+module TestContextVariables
+
+using Test
+
+@contextvar global cvar1 = 42
+@contextvar global cvar2::Int
+@contextvar global cvar3
+
+@testset "typed, w/ default" begin
+    ok = Ref(0)
+    @sync @async begin
+        with_context() do
+            @test cvar1[] == 42
+            cvar1[] = 0
+            @test cvar1[] == 0
+            ok[] += 1
+            @async begin
+                @test cvar1[] == 0
+                ok[] += 1
+            end
+            with_context(cvar1 => 1) do
+                @test cvar1[] == 1
+                ok[] += 1
+            end
+            @test cvar1[] == 0
+            ok[] += 1
+        end
+    end
+    @test ok[] == 4
+end
+
+@testset "typed, w/o default" begin
+    with_context() do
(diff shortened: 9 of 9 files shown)


src/jltypes.c
```diff
@@ -2297,7 +2299,7 @@ void jl_init_types(void) JL_GC_DISABLED
                                 jl_any_type,
                                 jl_any_type,
                                 jl_bool_type),
-                        0, 1, 9);
+                        0, 1, 10);
```
tkf, line 2302: I guessed I'd need to increment this as this patch needs one more field to be initialized.  Is it the correct usage of `ninitialized` of `jl_new_datatype`?

base/contextvariables.jl
```diff
@@ -0,0 +1,415 @@
...
+end
+
+# TODO: Do we need to implement `Dict{ContextVar}(::ContextSnapshot)`?
+# This requires storing UUID-to-ContextVar mapping somewhere.
+
+"""
+    snapshot_context() -> snapshot::ContextSnapshot
+
+Get a snapshot of a context that can be passed to [`reset_context`](@ref) to
+rewind all changes in the context variables.
```
tkf, line 404: It is called [`contextvars.copy_context`](https://docs.python.org/3/library/contextvars.html#contextvars.copy_context) in Python and [`ExecutionContext.CreateCopy`](https://docs.microsoft.com/en-us/dotnet/api/system.threading.executioncontext.createcopy?view=netcore-3.1) in .NET.  But I was not sure if "copy" is the right word when the underlying data is (treated as) immutable.

src/jltypes.c
c42f: Yes, this is correct because you always initialize `ctxvars` inside `jl_init_root_task` and `jl_new_task`, so it can never be NULL when seen by Julia code.

`ninitialized` is used as a way to keep track of which fields are always initialized, so that, for example, codegen can avoid emitting NULL checks when dereferencing those fields.

stdlib/UUIDs/src/UUIDs.jl
```diff
@@ -94,6 +94,8 @@ function uuid4(rng::AbstractRNG=Random.default_rng())
     UUID(u)
 end
 
+Base._uuid4(rng::AbstractRNG=Random.default_rng()) = uuid4(rng)
```
tkf, line 97: I couldn't find a proper way to "import" stdlib to Base.  Is this an OK approach?

travigd: Why not just `rand(UInt64)` or something involving `objectid`?

travigd: It looks to me like most of the uses of `key` could just be replaced with `objectid(::ContextVar)`

tkf: Actually, the main reason I wanted to use UUID was `uuid5` because I can then use it to get _deterministic_ and namespaced key for "global const" context variables (while `uuid4` is used only for local and "global non-const" context variables):

https://github.com/JuliaLang/julia/blob/ea0e4a7f8da164ba02b2bab612ab73fae97dfbd9/base/contextvariables.jl#L181

(where `join(fullpath, '.')` is something like `"PackageName.SubModuleName.varname"`)

Making the key for "global const" context variable is important for Distributed (otherwise different processes can't agree on the key).  So no, we can't use `objectid`.

travigd:
I don't see why you've "scoped" ContextVars to modules the way you have. Why not just scope them using normal objects (i.e., why do the vars themselves need to know about what module they were declared in?)? This is the way it works in Python at least.

tkf:
As I said https://github.com/JuliaLang/julia/pull/35833#discussion_r422746979, that's a requirement for this to work with Distributed across different machines.  Consider

```julia
module MyPackage
const KEY = uuid4()
end
```

`MyPackage.KEY` changes every time I re-compile it. So, a key generated similar way (including `objectid` at compile-time and run-time) can't be used to communicate with Distributed workers on different machines.

base/contextvariables.jl
```diff
@@ -0,0 +1,415 @@
...
+            "`@contextvar` can only be used inside a package.",
+        ))
+    end
+    fullpath = push!(collect(fullname(__module__)), varname)
+    if any(x -> contains(string(x), "."), fullpath)
+        throw(ArgumentError(
+            "Modules and variable names must not contain a dot:\n" * join(fullpath, "\n"),
+        ))
+    end
+    return uuid5(pkgid.uuid, join(fullpath, '.'))
```
c42f, line 181: Ah very nice chaining these UUIDs. I was just coming here to suggest it should be done this way. But you've already done it :-)

tkf: Good to know that we independently arrived at the same solution :) It's an indication that this solution makes sense.

c42f: Well perhaps I shouldn't be so definitive as to say "should" ... but I think it makes sense :-)

stdlib/UUIDs/src/UUIDs.jl
c42f: > I couldn't find a proper way to "import" stdlib to Base. Is this an OK approach?

Yeah it's a tricky problem: you'd like a `uuid4` to use a non-terrible RNG. But we can't pull all of `Random` into Base.

Another workaround for this would be to use `root_module` to look up the `UUIDs` module without loading it. There's precedent for this in the way LibGit2 is looked up in one place:

https://github.com/JuliaLang/julia/blob/97e3fe88d7d3e4f748195a4250b34ed593419b56/base/methodshow.jl#L316

In either case, it's not great to have `Base` missing some functionality when compiled without a stdlib present.

tkf: I see. Thanks for the explanation. How about vendoring libc-based uuid4 in Base?

```julia
function _uuid4()
    u = reinterpret(UInt128, [Libc.rand() for _ in 1:(sizeof(UInt128) ÷ sizeof(Cint))])[1]
    u &= 0xffffffffffff0fff3fffffffffffffff
    u |= 0x00000000000040008000000000000000
    UUID(u)
end
```

base/contextvariables.jl
c42f: Speaking of UUIDs, where does `uuid5` come from? I guess it will need to be moved it into `base/uuid.jl` from the UUIDs stdlib?

tkf: There is a fake `uuid5` definition for loading:

https://github.com/JuliaLang/julia/blob/8f512f3f6d5d69403be4b8179bd80ba64229e761/base/loading.jl#L103-L116

stdlib/UUIDs/src/UUIDs.jl
c42f: What quality of random numbers do we need for this? I guess a large period and good seeding are desired, as the UUIDs may end up in a distributed system? Actually this would suggest that the global RNG from `Random` isn't a great choice for `UUIDs` anyway because the user may control the seeding for reasons unrelated to UUIDs.

tkf: > Actually this would suggest that the global RNG from `Random` isn't a great choice for `UUIDs` anyway because the user may control the seeding for reasons unrelated to UUIDs.

Oooh, that's a good point. Maybe we should use `Random.RandomDevice()` instead of `Random.default_rng()` in UUIDs? See also https://github.com/JuliaLang/julia/issues/32954#issuecomment-526580532

But I think `_uuid4` above may be OK as the intial implementation since:

* It is used only for `@contextvar local var` and `@contextvar global var` (the latter is available mainly for exploration in REPL and testing).  But the "main entrypoint" is `@contextvar var` which uses uuid5.
* The UUIDs are generated during macro expansion time. So, messing with it requires you to call `Libc.srand(my_seed)` at the top-level of your module. It's totally possible but somewhat unlikely.

In the long run, I think we'd need something like `Base.CoreRandom.RandomDevice`.

tkf:
### Example usecase: circular reference detection

As a short somewhat-real usecase, here is how you can use the context variable to detect circular references.

Consider this innocent-looking `show` definition:

```julia
struct WAT
    value
end

Base.show(io::IO, x::WAT) = print(io, "WAT(", repr(x.value), ")")
```

However, this causes a stack overflow with

```julia
x = Any[WAT(Any[1])]
x[1].value[1] = x
@show x  # throws StackOverflowError
```

This is because current circular reference detection relies on `IOContext`.  So, using `repr` instead of `show` "cuts" `IOContext` propagation (this is why you should be using `repr(x; context=io)` but this is very error prone) and it fools `Base.show_circular`.

We can use a context variable instead of `IOContext` to propagate the history of objects that are already `show`n:

```julia
function nocircular(f, io, @nospecialize(x))
    @contextvar local shown::Cons
    history = get(shown)
    d = 1
    for y in something(history, ())
        if x === y  # it probably should compare the "root" `io` object, too.
            print(io, "#= circular reference @-$d =#")
            return
        end
        d += 1
    end
    with_context(f, shown => Some(Cons(x, something(history, Some(nothing)))))
end

struct Cons
    car
    cdr
end

Base.iterate(list::Cons) = iterate(list, list)
Base.iterate(::Cons, list::Cons) = (list.car, list.cdr)
Base.iterate(::Cons, ::Nothing) = nothing
```

We can use this to implement our `show` method of objects that may contain circular references:

```julia
mutable struct Mutable
    value
end

function Base.show(io::IO, x::Mutable)
    nocircular(io, x) do
        print(io, "Mutable(")
        show(io, x.value)
        print(io, ")")
    end
end
```

With this approach, circular references can be detected without using `IOContext`:

```julia
x = Mutable(WAT(Mutable(1)))
x.value.value.value = x
@show x  # => Mutable(WAT(Mutable(#= circular reference @-2 =#)))
```

Of course, it works even if you switch the task inside `show`:

```julia
struct WAAAAAT
    value
end

Base.show(io::IO, x::WAAAAAT) = print(io, "WAAAAAT(", fetch(@async repr(x.value)), ")")

x = Mutable(WAAAAAT(Mutable(1)))
x.value.value.value = x
@show x  # => Mutable(WAAAAAT(Mutable(#= circular reference @-2 =#)))
```

base/contextvariables.jl
c42f: Ah I see. Not your fault at all, but ick!

It seems some of the hashing stuff should move back into Base (at least some minimal parts of the implementation of SHA1 and RandomDevice, not the whole stdlib). Good hashing and randomness is core functionality which isn't really optional in some of these cases.

tkf: Here is a PR to move `RandomDevice` to `Base`: https://github.com/JuliaLang/julia/pull/35894

c42f:
I just found a rather nice post on the use and abuse of Go's equivalent `Context.Value`:

https://medium.com/@cep21/how-to-correctly-use-context-context-in-go-1-7-8f2c0fafdf39

They argue quite persuasively against `Context.Value`, so it's a thought-provoking read. In particular, they argue against using it for request-related data as mentioned in the PEP-567 rationale, preferring a closure. I think they're right about that!

I'm not arguing against `ContextVar` btw — I believe we should eventually have something to replace `logstate` — but it does make me wonder whether we should be cautious about making it too convenient ;-)

tkf:
A bit tangential, but I find the section "Does Context.Value even belong?" interesting.  I haven't realized that the cancellation context (= nursery) and the context variable is so tightly coupled in Go.  Though given the focus on server-side application in the `context` package documentation (and how Go is used in general), maybe it's not so puzzling.

Anyway, I agree dynamically scoped variables in any form can be used to write bad programs.  It's more or less just a "less bad" form of global variables, in the end.  For example, you can put a nursery in the context variable to easily create dynamically-scoped nursery.  We already have `global` but people don't use it often. So, I'd hope it'd be used with caution.

c42f:
> A bit tangential, but I find the section "Does Context.Value even belong?" interesting.

Yes I thought this was interesting, personally I feel it's a weird mixture of abstractions :-) But perhaps it's the best which can be done with the desire to explicitly pass no more than one context, and with the limitations of Go interfaces.

A couple more half-formed thoughts about the Julia interface:
* I think the difference between `const`, `local` and `global` versions may be perplexing for users, especially because in all cases they are global to the task (IIUC the `local` contextvar doesn't behave like a local variable). Could we get away with only the `const` version? It seems a bit less convenient but might be sufficient in practice?
* I like the look of `with_context`, it's pretty clear that the variable is set in dynamic scope. Reading the variables is less nice somehow, as the syntax doesn't suggest where these variables come from, somewhat in contrast to `task_local_storage(key)`. I guess it's no different from normal global variables though.

I also think it worth considering what API we're able to offer if we want to replace `logstate` with this system and have it be just as fast or faster. In that respect I like that you've documented `@contextvar` as the only official way to create these, it gives a lot of implementation flexibility. I think the compiler could model dynamic scope enough to allow the load of a contextvar to be hoisted out of any loop. (I think it would only be necessary that the same contextvar is not set in the intervening code, and that the contextvar use is inlined into the loop body. We might need to make the UUID part of the type for that to work neatly, but that seems it could be done in future as necessary.)

tkf:
> Could we get away with only the `const` version?

Yeah, I think that's what you need most of the time. The `global` version is just too handy to use in the REPL. Otherwise, you have to create a full blown package just for playing with the context variables.

The `local` version is least important but it's needed (or it's the easiest to do it) if we want to support context variables in scripts that works with Distributed, without creating a package. This is because, since local `ContextVar` object (i.e., UUID) is captured by the closure, it'd be sent to the peer processes.

I think we can special-case `Main` to support the `const` version in the REPL. But I imagine that would be pretty hairy (involving `finalize`?).

Another aspect is that, even if we only support the `const` version, people are free to experiment with the internal functions and create the `local`/`global` version themselves.  So maybe it's a wise choice to play conservative and only provide the `const` version in the first version of the API.

> Reading the variables is less nice somehow, as the syntax doesn't suggest where these variables come from, somewhat in contrast to `task_local_storage(key)`.

Do you mean the `x[]` syntax? I just thought it's kind of cute way to do it. We can always make it uglier to make it explicit (e.g., `get_context_variable(x)`).

> I guess it's no different from normal global variables though.

Maybe we should recommend using uppercase for context variables, too?

> I think it would only be necessary that the same contextvar is not set in the intervening code, and that the contextvar use is inlined into the loop body.

Yeah, I'd hope so. Context variable lookup is just a dictionary look up in the end. So, if the compiler can assume it's effect-free, I suppose it can hoist it out?

We might need a better dictionary implantation, though. For now, I'm just using `Dict` in a copy-on-write manner. I wonder if it's better to use immutable data structures like HAMT backed up with `Base.Experimental.Const(::Vector)`.  I made this part "pluggable" so that people can easily experiment with it; i.e., you can monkey-patch a single function `merge_ctxvars` to use an alternative storage type.  It's not a public stable API, of course.

c42f:
I got a chance to chat to @JeffBezanson. Jeff had some great questions which I was ill-prepared for (and I know @tkf could have answered better) but it was productive nonetheless.  Here's a summary:

Some more minor points:
* The local/const/global thing came up again as seeming somewhat complex and confusing, and I couldn't describe the distinction off the top of my head. (I'm pretty sure I did understand this a month ago, but I found I'd largely forgotten what it was all about!)
* Jeff mentioned that UUIDs seemed kinda heavy weight as keys. (Agreed, though I'm not sure whether there are alternatives in the distributed setting?)
* Having ContextVar storage as a `Dict` didn't seem great. I mentioned that I thought this was just a simple reference implementation, with the internal `merge_ctxvars` designed to allow experimentation and gradual improvement.
* Obviously ContextVars and task local storage really do the same thing, so Jeff commented that having two of them is kind of bad, and that taking up an extra pointer in the Task isn't great. My thought/comment about this is that we just need to be sure that ContextVars are a satisfying API covering all the cases that task local storage do (and more!), and that we might tolerate the extra field for a few releases until we can delete the current task local storage API.  If we can delete the `logstate` field we'd also make up for it ;-)

I think the most important point discussed was ability to implement this efficiently in the future: exposing `set!` kills the compiler's ability to generate good code for context vars — it cannot reason locally about their value if a call to a child function can mutate the state seen by the parent. This seems like a general problem for any proposed implementation of dynamic scope, so I'll write some further thoughts about it in #35757.

tkf:
> The local/const/global thing came up again as seeming somewhat complex and confusing

I think the best way forward might be to just expose the `const` version (i.e., the default one) as the API at least for the first version.  Adding others is pretty easy because there is no change required for the backend.  For example, it can be provided as a package (provided that the package author is brave enough to rely on the implementation detail).

Anyway, here is how they work in the reference implementation:

* `const` (`@contextvar x`): Usable at top-level scope of _packages_. It cannot be used outside packages because the UUID has to be salted with the package UUID. Usable with Distributed.jl.
* `local` (`@contextvar local x`): To be used inside functions.  Usable with Distributed.jl because the closure will capture `ContextVar` object which includes the UUID.
* `global` (`@contextvar global x`): To be used in REPL or scripts.  Not usable with Distributed.jl because the UUID is not sent to remote.

It'd be possible to merge `const` and `local` if a macro can know if they are in the top-level scope or in the function-level scope.  Merging `const` and `global` may be possible since the macro can detect if the package UUID is available or not.  If the package UUID is not available, it can generate UUID4 (and maybe even broadcast the definition via Distributed.jl).

> UUIDs seemed kinda heavy weight as keys

I guess we can swap it with something better once we find out a better way to do it?  For example, maybe we can have a global "key broker" that maps an UUID to a small key `UInt32` or `UInt64` and store the small key in the `Task`-local storage.  When a `ContextVar` object is instantiated (also when it's loaded from precompilation cache), it asks the key broker for the short key and store it in its field.  It'd be tricky to make it Distributed.jl-compatible, though.

> Having ContextVar storage as a `Dict` didn't seem great. I mentioned that I thought this was just a simple reference implementation, with the internal `merge_ctxvars` designed to allow experimentation and gradual improvement.

Yeah, I just want to get API correct and make it easy to experiment with better backend storage.  HAMT is an obvious candidate.

c42f:
> HAMT is an obvious candidate.

Ah yes, I see the wikipedia page mentions persistent variants of HAMTs. This seems like the thing we want :+1:

oschulz:
> `@contextvar x` [...] `x` is forced to be namespaced 

Would that allow code to access context via `@contextvar SomeModule.x`, to avoid clashes with local names?

tkf:
How `@contextvar SomeModule.x` behaves?  Just be clear, by namespacing I mean:

```julia
module A
    module B
        @contextvar x

        get_x() = x[]

        function shadowing()
            x = 1    # shadows context variable `x`
            sum = 1  # shadows `Base.sum`
            return get_x()
        end
    end

    get_B_x() = B.x[]
end

# These work and equivalent:
A.B.x[]
A.B.get_x()
A.get_B_x()
A.B.shadowing()

# These do not work:
x[]
A.x[]
```

`@contextvar x` behaves just like ordinal global (const) variable.  You'd already make sure that local variable don't shadow functions, constants, and globals.  I don't think we need something extra here.

oschulz:
>  Just to be clear, by namespacing I mean [...]

Oh, of course - very nice!

How would you handle the case where modified context is to be passed to tasks/workers? I am in a certain context, but I want to modify part of that context in a different way for each worker (without changing by current context). E.g. to partition resources listed in the context among tasks/workers, etc. How should be express that, API-wise?

tkf:
> How would you handle the case where modified context is to be passed to tasks/workers?

It's not implemented but it's easy to add if you don't care performance.  I was thinking this would be a neat feature to have, but, initially, I also was not expecting this to be added since it's too much overheads.  I thought it'd have to use dynamic dispatch. But since @c42f and @JeffBezanson are very serious about optimizing the hell out of it, I think it might actually be possible to have this.  But I don't think it'd be available in near future since I imagine that it's a big compiler project.

---

I'm also answering your comment in https://github.com/JuliaLang/julia/issues/35757#issuecomment-657013260 here because I think it's more related to this particular API

> The "local" `@contextvar` still be propagated to tasks, though, right?

`local` in `@contextvar local` is actually nothing to do with the extent the context variable propagate.  This `local` is about how name of the context variabel is resolved.  This is always lexical and follows the usual Julia semantics.

(But your comment made me realized that it'd be very confusing if you think `local` related to the dynamic scoping of the context itself. Probably that's another reason why we should ditch `local`.)

Anywa, in the current implementation it unconditionally propagates across in-process tasks. It'd be possible to have `@shared_contextvar local`.  Some examples:

```julia
@contextvar CONST_IN_PROCESS = 1
@shared_contextvar CONST_SHARED = 1

function f()
    @contextvar local local_in_process = 1
    @shared_contextvar local_shared = 1
    local local_in_process[] += 1
    local_shared[] += 1
    CONST_IN_PROCESS[] += 1
    CONST_SHARED[] += 1
    return local_shared[]
end

function g()
    # I can't use them:
    # local local_in_process[] += 1
    # local_shared[] += 1
    CONST_IN_PROCESS[] += 1
    CONST_SHARED[] += 1
    return CONST_SHARED[]
end

function user()
    x = f()
    f()                             # => x + 1
    y = @sync @async begin
        f()                         # => x + 2
        remotecall_fetch(f, POOL)   # => x + 3
    end
    @assert y - x == 3
end
```

Note that `f` and `g` both can use `CONST_SHARED` but only `f` can use `local_shared`.  This is because `local_shared` is (lexically) local to `f`.

(Now I just realized that `user` relies on the precompilation cache to be shared across worker processes. So this is yet another reason to not have `@contextvar local`.)

tkf:
One solution for "doing" something at `@spawn` is to record the history of call tree.  Something like

```julia
struct SplitLog
    parent::Union{Nothing,SplitLog}
    id::UInt64
    left::Bool
end

@contextvar SPLIT_LOG = SplitLog(nothing, 0, true)

const _ID = Ref(UInt64(0))  # TOOD: lock it
newid() = _ID[] += 1
```

and tweak `@spwan` such that `@spwan $expr` expands to

```julia
left = SplitLog(SPLIT_LOG[], newid(), true)
right = SplitLog(SPLIT_LOG[], newid(), false)
t = Task() do
    with_context(SPLIT_LOG => left) do
        $expr
    end
end
t.sticky = false
schedule(t)
SPLIT_LOG[] = right   # should we?
```

Now you can do

```julia
@contextvar RESOURCE = ...

function with_resource(f)
    # record the "root" split id
    with_context(RESOURCE => Resource(Base.SPLIT_LOG[].id)) do
        f()
    end
end

current_resource() =
    # compute current resource from RESOURCE[] and Base.SPLIT_LOG[]
```

That is to say, `Base.SPLIT_LOG` gives you enough information for "simulating" all the `@spawn` to arrive at the current task.

However, since `SPLIT_LOG[] = right` is the thing that @c42f was arguing to remove, I'm not sure if this is a good approach.

oschulz:
> Anywa, in the current implementation it unconditionally propagates across in-process task

Yes, that's what I meant. Ok, so basically we'll have context variable that are restricted to the process, and context variable that propagate to remotes, right? I like it!

oschulz:
> One solution for "doing" something at @spawn is to record the history of call tree

So the easiest solution would be to have the task change it's context after spawning, and transfer the relevant information via closure? Like

```
function fork_work_with_procs(workers)
Threads.@spawn begin
    Distributed.available_workers[] = workers
    ... do some local stuff ...
    ... call something that distributes work and respects Distributed.available_workers[]
    ... do some more local stuff
    return something
end
```

Would that fit into your concept?

tkf:
> we'll have context variable that are restricted to the process, and context variable that propagate to remotes,

Yes, that's my idea.  This PR only deals with the in-process part, though.  I'm still waiting for an "approval" for this RFC before adding more to this.

> the task change it's context after spawning

If you don't need to change anything for the current task, it'd be a simple use-case of `with_context`:

```julia
function fork_work_with_procs(workers)
    Threads.@spawn with_context(Distributed.available_workers => workers) do
        ...
    end
end
```

oschulz:
Nice!

tkf:
I removed `setindex!`-related API from the ContextVariables.jl PoC package.  This makes it possible to ditch IRTools.jl-based proof-of-concept and use logger as "storage" backed https://github.com/tkf/ContextVariablesX.jl/pull/2.  It means that ContextVariables.jl is now kind of usable for all Julia 1.x.

So, I just renamed ContextVariables.jl ro ContextVariablesX.jl and trigger the registration https://github.com/JuliaRegistries/General/pull/19209.  This way, maybe people can use it in some real projects and understand the characteristics of the API better.

c42f:
> use logger as "storage" backed

Haha this is a really funny abuse of the logstate, I like it :-D

It's great to have something which can be experimented with right away.

tkf:
Here is a possible usage in Scratch.jl:
https://github.com/JuliaPackaging/Scratch.jl/compare/v1.0.1...tkf:contextvariablesx

ref https://github.com/JuliaPackaging/Scratch.jl/issues/15

oschulz:
Just curious, what's the status of this?

jpsamaroo:
I think this is stuck on figuring out a mechanism which is fast for both small and large sets of variables (where different dictionary structures would be necessary for good performance), but may also be optimizable by the compiler. I suspect the best way to get this implemented is:
- Have libraries try out/depend on https://github.com/tkf/ContextVariablesX.jl (for example: https://github.com/JuliaParallel/Dagger.jl/pull/328)
- Wire-up a smaller dictionary (HAMT?) in ContextVariablesX for small sets of context vars
- Benchmark this approach on all sorts of things
- If all went well, update this PR with that approach

I'm not sure what to do about making context vars optimizer-friendly; that seems like a can of worms best deferred until we're happy with the API and overhead of boxed context vars.

oschulz:
Thanks, @jpsamaroo !

tkf:
FYI, I used ContextVariablesX.jl in https://github.com/JuliaConcurrent/Julio.jl and the API was sufficient for building a structured concurrency framework and worked as expected.

I think HAMT can wait since this PR is designed in such a way that people can monkey patch the backend dictionary. I don't expect the current implementation to have a large performance overhead on task spawns. The performance penalty is incurred only when creating a new context. So, the impact on existing programs is likely negligible.

Also, another issue around uuid4 is already fixed by #35872.

I think the main blocker is the decision that it is worth 64 bits of our precious `Task`. Or we can decide to slow down logging a bit by using contextvar to implement logger. This way, we don't have to change the size of `Task`.

oschulz:
> Or we can decide to slow down logging a bit by using contextvar to implement logger.

I think it would be quite natural to handle logging via the context mechanism as well then, instead of separately. It is a "logging context", after all.

jpsamaroo:
+1 to moving `logstate` into a more general context variable system; it's somewhat silly that only logging gets its own behavior and storage space.

c42f:
> fast for both small and large sets of variables [...] but may also be optimizable by the compiler

Optimizability from the compiler point of view is mostly orthogonal to the dictionary data structure used in the runtime. The main kind of optimization which seems interesting is hoisting the dictionary lookup outside of loops (especially useful after inlining?). Such hoisting is only possible if child functions don't modify the identity of the context variable value as seen by the parent function. @tkf already adjusted this API of ContextVariablesX to make sure that's true (https://github.com/JuliaLang/julia/pull/35833#issuecomment-670982004).  This PR still includes `set!` but I assume it'd be removed in an update?

We can also optimize the Julia runtime data structures (eg, to use a HAMT); leaving that to a future PR seems fine to me.

> and overhead of boxed context vars.

I feel like the unboxing/lookup may not be a big problem, provided the compiler can move that lookup around (especially, move it out of loops).

This is where the logging system will slow down a bit if we use context variables for storing the logstate. But in various PRs people have already been making logging slower in favor of avoiding invalidations and the like. So at this point it might not slow it down too much more if we were to also have a dictionary lookup. This would be worth trying.

BTW, we previously discussed logstate and other things in #35757 (beginning around https://github.com/JuliaLang/julia/issues/35757#issuecomment-625705431)

---

To me, the biggest challenge is in context variables which need special handling during task spawning (eg, task local RNGs). Without addressing that challenge, we'll never have a something which is "like global variables but inherently threadsafe".

For now, this PR doesn't address that and that's fine. Are we happy that we can tackle that separately in the future?

tkf:
> This PR still includes `set!` but I assume it'd be removed in an update?

Ah, yes, I forgot that I didn't backport the change. Currently, the source of truth is ContextVariablesX.jl since it's much easier to tweak code in a package.

> To me, the biggest challenge is in context variables which need special handling during task spawning (eg, task local RNGs).

To make something like efficient task local RNGs plugable, we need to be able to specialize functions based on calling context. It requires compiler support and it's very out of the scope of this PR. But it'd be definitely great to have it and also I wonder if we can generalize our method invalidation system to support "multiverse" method table (calling context specialization is kind of `@eval` followed by `invokelatest`).

Also, *if* we can have context-specializations of a method, we can implement context variable using a very different strategy given that we have hooks on:

1. spawn
2. task switch
3. sync (optional)

The on-spawn hook is required for referencing parent task context. It's also where RNG splits its state. We can use task switch hook to stash the current state in `Task`'s stack. Between task switches, the state can be accessed through, say, `BUFFER[threadid()]`. It makes the context access much cheaper. In a way, we use the context-specialization to hoist the dictionary lookup to the compile time. [^1] Then, optionally, we can also use a on-sync hook to merge child's state to parent's state. It can be used for implementing Cilk's hyperobject. The hooks on spawn and sync have the duality kind of like the iterator and fold. So, it's very powerful to have both. For example, we can have multitask-compatible event collection system with task local buffers that does not require any locks or atomics. The idea is to merge task local buffers on sync, so that you get the single history of even collection of all tasks at the end of execution. This works only if child tasks are all synchronized, though (i.e., fork-join or at least structured concurrency is required).

Anyway, we can worry about an alternative strategy like this later. I think this design is compatible with the API proposed in this PR, even with `set!`. (But we should remove it to play on the safe side.)

[^1]: But this strategy moves the burden to the task switch code. So, if we have *a lot* of context variables, the strategy in this PR may be better.

c42f:
> To make something like efficient task local RNGs plugable, we need to be able to specialize functions based on calling context.

I'm not so sure it's necessary or good to have the contextvar-specific "spawn event function" inlined into the caller at the `@spawn` site. This seems like a recipe for a fairly extreme amount of code generation.

For most uses it might be sufficient to add a layer of indirection, basically calling a function pointer to deal with each contextvar which needs spawn handling. Presumably this would happen in the parent task so it's not involved in the delicate process of actual task switching, but done instead at task allocation time.

For most use cases I think this could be good enough and not lead to an explosion of code generation per context. Unfortunately RNG state update is some of the most demanding (and installed on the task by default!) so this strategy might be too slow for that.

> very out of the scope of this PR.

Definitely :-)

c42f:
> This seems like a recipe for a fairly extreme amount of code generation.

As a counterpoint to my own argument: it seems likely that most programs have a fairly restricted set of functions which actually call `@spawn`, so inlining into the spawn site based on context might not be the code generation disaster I'm worried about.

Only some prototyping and measurement will tell, I suppose.

tkf:
I don't think it has to be always "inlined" in the sense of `@inline`. The function just has to be there, like `schedule` is inside of `@spawn`. The compiler already reuses compiled `schedule` and `wait` by just not inlining them. So, I'd imagine context-specialized definition will end up being something similar. That said, for some relatively cheap function like `rand()`, I think we do want `@inline`ing.

But yes, it's important to try avoiding aggressive invalidation of the *entire* call tree just because the user switched the task local RNG type. Or maybe a more common situation is importing a package that uses context-specialized API. At the same time, I'd imagine the overhead will be at most comparable to the edit-revise loop which feels like an "OK" compromise if we can make pluggable task local RNGs fast and extensible (e.g., having multiple streams). Anyway, I totally agree that we need some prototyping to understand what it'd look like.

jpsamaroo:
Bump! This would be immensely useful as a public API to prevent having to rely on hacks like `logstate`. Is there anything blocking the specific changes in this PR?

c42f:
> Is there anything blocking the specific changes in this PR?

I think API changes from ContextVariablesX still need porting in here:

> > This PR still includes `set!` but I assume it'd be removed in an update?
> 
> Ah, yes, I forgot that I didn't backport the change. Currently, the source of truth is ContextVariablesX.jl since it's much easier to tweak code in a package.

Tokazama:
Is this something that we want to also support context in the sense that `Cassette.Context` does, or is this solely intended for concurrency?

oschulz:
> or is this solely intended for concurrency?

I don't think it's only useful for concurrency, it's about propagating argument-like "settings" that can't be well handled via function arguments. Things like logger, progress monitor, choice of computational resources, etc. This is of course very important for thread- and process-level parallelism, but I think it would benefit serial chains of function calls as well.

MasonProtter:
Just to add another potential motivation for this sort of thing, the way that `propagate_inbounds` works is basically a hacky, single use implementation of a context variable. 

It'd be pretty cool if we could have context variables be lightweight enough that things like flagging that it's okay to turn off error checks could use this mechanism.

c42f:
I don't think `propagate_inbounds` is very similar — it only propagates through one additional level of inlining. This "locality" is an important feature, especially in the reliable implementation of higher order functions. For example, here's what can go wrong with too much `@propagate_inbounds`: https://github.com/andyferris/Dictionaries.jl/pull/64

NHDaly:
I want to ask about another related usage of `Context` in Golang:
- Cancellation and timeout

https://pkg.go.dev/context#WithCancel

The Go Context object can also contain an indication of whether a request has been cancelled by the client, and/or whether a timeout/duration has been exceeded.

This requires the server code to thread the Context through all of the functions on the server, across module boundaries, and check for cancellation at very places.

Could we also cover that logic with a ContextVariable as proposed here / in https://github.com/tkf/ContextVariablesX.jl?
In particular: there a few extra pieces besides just inheriting named variables across task boundaries:
- > WithCancel returns a copy of parent with a new Done channel. The returned context's Done channel is closed when the returned cancel function is called or when the parent context's Done channel is closed, whichever happens first.
   > 
   > Canceling this context releases resources associated with it, so code should call cancel as soon as the operations running in this Context complete.

I'm thinking maybe this could be done by having a function like `with_cancel(ctx) do` that listens to the parent context's cancellation variable and propagates that state to its own?

The main place I see where this might not mesh well with the current design is:
- The current design doesn't have a local context variable to copy from, nor assign to, so i don't see how you could keep multiple separate cancellation context handles?

It seems like supporting multiple separate cancellation contexts is a bit at-odds with the feature this package seems to provide, which is avoiding having to thread the variable through all your code by instead making it a global variable.
Which I must admit is very appealing.

Maybe the approach is to have these cancellation contexts also be named context variables? Like
```julia
@contextvar request_cancelled = CancellationContext()
```
And then every new incoming server request could set its own `request_cancelled` variable?

Has anyone else thought about this already? We're working on implementing user-requested Transaction Cancellation in our async web server at RAI now, and it occurred to me that these go-style Context objects seem like the right design.

c42f:
I think you can use an implementation of dynamic scope which is inherited across `@async`/`@spawn` (such as this) to build a scope-based cancellation system.

However I don't think there's a need to bake anything like that into the context variables system itself.

Safe and convenient cancellation is a very deep topic and it's not clear to me that any language has fully solved this problem. But I still think the way that cancellation is handled in structured concurrency systems is the best option seen so far. On that note, you should really check out https://github.com/JuliaConcurrent/Julio.jl if you haven't seen it already.

vchuravy:
One comment from triage was that the `@contextvar` is slightly to magical. Would it be feasible do have something akin to `Ref` e.g. `ContextVar` so that the `v[]` is "expected"?

tkf:
I totally agree that anything that can be implemented as a macro must be implemented without a macro. However, I couldn't find any other ways to provide some desired properties without macro. (This was precisely what was discussed in https://github.com/JuliaLang/julia/pull/35833#discussion_r422996907)

Distributed computing is the main reason why the primary construction API is provided through macro. With macro-based construction, it supports the stability of the context variable key across precompilations. This key is used for identifying the value assigned to the context variable within a context (say, a task-local `Dict{UUID,Any}`). I believe it is desirable to maintain the identity of context variable `MyPackage.MyModule.MY_VARIABLE` across different processes of different machines in order to use this mechanism for distributed computing. To support this, the context variable key is generated using `uuid5(PkgId(MyPackage).uuid, "MyPackage.MyModule.MY_VARIABLE")`. This way, context variables precompiled in different machines have exactly the same UUID, thanks to the UUID-based packaging infrastructure. This makes copying entire context variables across machines as "easy" [^1] as copying the key-value pairs since the keys have the same UUIDs across machines. If we used (say) `uuid4` from a constructor invocation `ContextVariable()` the key associated with this context variable changes for each precompilation. This still works for single-process application. However, a context variable cannot be guaranteed to be identified across multiple processes anymore.

[^1]: Of course, copying everything unconditionally is a bad idea unless only immutable values -- not mutable objects -- are stored in context variables. Some additional mechanism must be built upon this PR to support automatic cross-process propagation. However, stable context variable key is still a prerequisite for deciding cross-process propagation based on context _variable_ identity (and not merely based on values and/or types).

So, it would be nice if triage can discuss:

(1) Do we want to support sending values assigned to context variables across distributed processes?

(2) If so, is it reasonable to use what is implemented in ContextVariablesX?

Even if the answer to the second question is "maybe?", I think it would be better to use the current macro-based API even we ended up using different implementation strategies because it is *constrained enough* to *allow* the implementation used by ContextVariablesX.

oschulz:
> I believe it is desirable to maintain the identity of context variable MyPackage.MyModule.MY_VARIABLE across different processes of different machines in order to use this mechanism for distributed computing.

Yes please!

oschulz:
To elaborate: I think distributed use cases would profit from context variables very much. And if the machinery doesn't forward context variables through remote calls and the like it will be very hard for the user to do so. And that will mean that code that relies on context variables would malfunction when used distributed - so lot's of special handling would need to be added manually for the distributed case, making code much less generic.

simonbyrne:
I broadly like the overall design here, but I have a minor suggestion: provide some mechanism for specifying how context variables should be passed when spawning subtasks on local and remote processes.

This was motivated by #48121: the idea is that when you spawn a task on a remote process, you could wrap the logger in a `RemoteLogger` (so that any log messages are passed back to the main process).

vchuravy:
Another implementation is `ScopedVariables` https://openjdk.org/jeps/446

vchuravy:
Replaced by #50958

PR #35836, opened 2020-05-11
add a dummy method to ! to get rid of its Union splitting

KristofferC:
Removes 4000 lines of invalidations when loading SIMD.jl...

Before

```
julia> Core.Compiler.return_type(x -> !(x[]), Tuple{Ref{Any}})
Union{Missing, Bool, Base.var"#64#65"{_A} where _A}
```

After

```
julia> Core.Compiler.return_type(x -> !(x[]), Tuple{Ref{Any}})
Any
```

Maybe not a 100% serious PR...

### DIFF
--- a/test/compiler/inference.jl
+++ b/test/compiler/inference.jl
@@ -2592,4 +2592,4 @@ end
 # Test that we do not union split ! based on the number of methods
 # because that leads to excessive invalidations if a package
 # defines an additional method for !
-Core.Compiler.return_type(x -> !(x[]), Tuple{Ref{Any}}) == Any
+@test Core.Compiler.return_type(x -> !(x[]), Tuple{Ref{Any}}) == Any
--- a/test/compiler/inference.jl
+++ b/test/compiler/inference.jl
@@ -2588,3 +2588,8 @@ f() = _foldl_iter(step, (Missing[],), [0.0], 1)
 end
 @test Core.Compiler.typesubtract(Tuple{Union{Int,Char}}, Tuple{Char}) == Tuple{Int}
 @test Base.return_types(Issue35566.f) == [Val{:expected}]
+
+# Test that we do not union split ! based on the number of methods
+# because that leads to excessive invalidations if a package
+# defines an additional method for !
+Core.Compiler.return_type(x -> !(x[]), Tuple{Ref{Any}}) == Any
--- a/base/Base.jl
+++ b/base/Base.jl
@@ -391,6 +391,12 @@ function include(mapexpr::Function, mod::Module, _path::AbstractString)
     return result
 end
 
+# Hack: add a method to Base.! to prevent union splitting
+struct _NEG_UNION_SPLITTING_BLOCKER
+end
+!(::_NEG_UNION_SPLITTING_BLOCKER) = "foo"
+
+
 end_base_include = time_ns()
 
 if is_primary_base_module


JeffBezanson:
We could actually add a hack to the compiler for this --- we've done things like that before. We could add a special case for `!`, or a general feature of per-function max_methods, i.e. you could annotate that we shouldn't bother looking at multiple matches for `!`.

test/compiler/inference.jl
```diff
@@ -2588,3 +2588,8 @@ f() = _foldl_iter(step, (Missing[],), [0.0], 1)
 end
 @test Core.Compiler.typesubtract(Tuple{Union{Int,Char}}, Tuple{Char}) == Tuple{Int}
 @test Base.return_types(Issue35566.f) == [Val{:expected}]
+
+# Test that we do not union split ! based on the number of methods
+# because that leads to excessive invalidations if a package
+# defines an additional method for !
+Core.Compiler.return_type(x -> !(x[]), Tuple{Ref{Any}}) == Any
```
ararslan, line 2595:
```suggestion
@test Core.Compiler.return_type(x -> !(x[]), Tuple{Ref{Any}}) == Any
```

timholy:
Let me float another possibility: first, while this catches 319 MethodInstance invalidations, there are "only" 88 unique methods. I'm guessing that for the majority of these, adding a `::Bool` type annotation (or possibly `::Union{Bool,Missing}`) would be an improvement on the implementation of those methods. For example, `with_output_color` (which has a single Method) leads to > 1000 invalidations, a large fraction of the total. All of these invalidations stem from https://github.com/JuliaLang/julia/blob/f12cde0e8c7f4268838d3201dd9e13b90ea44633/base/util.jl#L71
and instead writing this as `iscolor = get(io, :color, false)::Bool` would certainly improve code quality. The same can be seen for many other invalidations.

That leaves things like the fallback `!=(::Any,::Any)`, https://github.com/JuliaLang/julia/blob/f12cde0e8c7f4268838d3201dd9e13b90ea44633/base/operators.jl#L202
What if we could write it instead like this?
```julia
julia> function Base.:(!=)(x, y)
           @nospecialize(!(::Any))
           return !(x == y)
       end
```
and thereby indicate that `!(::Any)` should use runtime dispatch? That doesn't work now, but presumably it could.

JeffBezanson:
> and instead writing this as `iscolor = get(io, :color, false)::Bool` would certainly improve code quality.

:+1: PR it! :smile:

tkf:
> instead writing this as `iscolor = get(io, :color, false)::Bool` would certainly improve code quality

Would `iscolor = get(io, :color, false) === true` work too?  I always thought Boolean `IOContext` should be compared with `===` since the API is rather loose (it just says "common use"). For example, BenchmarkTools.jl used to treat `:limit` as an integer rather than a `Bool` https://github.com/JuliaCI/BenchmarkTools.jl/pull/155

timholy:
> :+1:  PR it! :smile:

It's coming. There's some fun tooling on its way to make finding this kind of thing much easier: https://github.com/JuliaDebug/Cthulhu.jl/pull/78

timholy:
> iscolor = get(io, :color, false) === true

Very similar, but with one difference: if I did `IOContext(io, :color=>:red)`, then your version would result in colorless display, whereas the `get(...)::Bool` variant throws an error. I think I prefer the error, since the former might be hard to debug ("why am I not getting red?!?")

PR #35839, opened 2020-05-11
Support `init` keyword in `maximum`/`minimum`

timholy:
We can supply `init` for general reductions, but `maximum` and `minimum` do not support this keyword. This PR allows them to do so:
```julia
julia> maximum([1,2,3])
3

julia> maximum(Int[])
Stacktrace:
 [1] _empty_reduce_error() at ./reduce.jl:299
 [2] reduce_empty(::Function, ::Type{T} where T) at ./reduce.jl:309
⋮

julia> maximum(Int[]; init=-1)
-1
```

This was motivated by writing a blog post on invalidations (https://github.com/JuliaLang/www.julialang.org/pull/794), and together with some changes to Pkg to supply the `init` keyword it eliminates all of the `reduce_empty` invalidations from loading FixedPointNumbers (which extends `reduce_empty`).

It's worth noting one unfortunate fact: `init` has different meanings for `reduce(+, src)` than for `sum!(dest, src)`. In the former case, it's the value for seeding the result; in the latter, it's a boolean indicating whether to re-initialize `dest` to the sentinel value or to keep whatever result is already there. This PR might make the situation slightly worse by supporting more cases where the conflict becomes appreciable (`maximum` vs `maximum!`). This feels like something we should fix for Julia 2.0.

Closes #35733

### DIFF
--- a/base/reducedim.jl
+++ b/base/reducedim.jl
@@ -307,21 +307,21 @@ julia> mapreduce(isodd, |, a, dims=1)
  1  1  1  1
 ```
 """
-mapreduce(f, op, A::AbstractArrayOrBroadcasted; dims=:, kw...) =
-    _mapreduce_dim(f, op, kw.data, A, dims)
+mapreduce(f, op, A::AbstractArrayOrBroadcasted; dims=:, init=_InitialValue()) =
+    _mapreduce_dim(f, op, init, A, dims)
 mapreduce(f, op, A::AbstractArrayOrBroadcasted...; kw...) =
     reduce(op, map(f, A...); kw...)
 
-_mapreduce_dim(f, op, nt::NamedTuple{(:init,)}, A::AbstractArrayOrBroadcasted, ::Colon) =
-    mapfoldl(f, op, A; nt...)
+_mapreduce_dim(f, op, nt, A::AbstractArrayOrBroadcasted, ::Colon) =
+    mapfoldl_impl(f, op, nt, A)
 
-_mapreduce_dim(f, op, ::NamedTuple{()}, A::AbstractArrayOrBroadcasted, ::Colon) =
+_mapreduce_dim(f, op, ::_InitialValue, A::AbstractArrayOrBroadcasted, ::Colon) =
     _mapreduce(f, op, IndexStyle(A), A)
 
-_mapreduce_dim(f, op, nt::NamedTuple{(:init,)}, A::AbstractArrayOrBroadcasted, dims) =
-    mapreducedim!(f, op, reducedim_initarray(A, dims, nt.init), A)
+_mapreduce_dim(f, op, nt, A::AbstractArrayOrBroadcasted, dims) =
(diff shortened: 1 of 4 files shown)


base/reducedim.jl
```diff
@@ -307,21 +307,21 @@ julia> mapreduce(isodd, |, a, dims=1)
...
 mapreduce(f, op, A::AbstractArrayOrBroadcasted...; kw...) =
     reduce(op, map(f, A...); kw...)
 
-_mapreduce_dim(f, op, nt::NamedTuple{(:init,)}, A::AbstractArrayOrBroadcasted, ::Colon) =
-    mapfoldl(f, op, A; nt...)
+_mapreduce_dim(f, op, nt, A::AbstractArrayOrBroadcasted, ::Colon) =
+    mapfoldl_impl(f, op, nt, A)
 
-_mapreduce_dim(f, op, ::NamedTuple{()}, A::AbstractArrayOrBroadcasted, ::Colon) =
+_mapreduce_dim(f, op, ::_InitialValue, A::AbstractArrayOrBroadcasted, ::Colon) =
```
tkf, line 318: :+1: That's what we did in https://github.com/JuliaArrays/StaticArrays.jl/pull/750

base/reducedim.jl
```diff
@@ -717,12 +717,12 @@ for (fname, _fname, op) in [(:sum,     :_sum,     :add_sum), (:prod,    :_prod,
...
-        @inline ($fname)(a::AbstractArray; dims=:) = ($_fname)(a, dims)
-        @inline ($fname)(f, a::AbstractArray; dims=:) = ($_fname)(f, a, dims)
+        @inline ($fname)(a::AbstractArray; dims=:, kw...) = ($_fname)(a, dims; kw...)
+        @inline ($fname)(f, a::AbstractArray; dims=:, kw...) = ($_fname)(f, a, dims; kw...)
 
         # Underlying implementations using dispatch
-        ($_fname)(a, ::Colon) = ($_fname)(identity, a, :)
-        ($_fname)(f, a, ::Colon) = mapreduce(f, $op, a)
+        ($_fname)(a, ::Colon; kw...) = ($_fname)(identity, a, :; kw...)
+        ($_fname)(f, a, ::Colon; kw...) = mapreduce(f, $op, a; kw...)
```
tkf, line 725: Can we use `init` as a positional argument for `$_fname`. I guess that would shave off some compiler work (and avoid it to get in the way of the inliner)?

tkf (approved):
The code LGTM but it looks like the doctest failed in CI.

> It's worth noting one unfortunate fact: `init` has different meanings for `reduce(+, src)` than for `sum!(dest, src)`.

Yes, it's also unfortunate that the meaning of `init` is slightly different in `foldl`/`accumulate` compared to those two families.

> This feels like something we should fix for Julia 2.0.

I think we can introduce new keyword arguments (and, technically, also even deprecate `init`) in 1.x. But I guess I should open a new issue for discussing this.

Issue #28777, opened 2018-08-20
Mean of an empty collection


bkamins:
It seems that this issue must have been discussed, but I could not find any reference to it so I open it.

`mean` behaves inconsistently between array and collection implementations in empty container case. This is especially relevant since `missing` was introduced into base. Here is a code that shows the problem:
```
julia> using Statistics

julia> x = Union{Float64,Missing}[missing]
1-element Array{Union{Missing, Float64},1}:
 missing

julia> mean(skipmissing(x))
ERROR: ArgumentError: mean of empty collection undefined: Base.SkipMissing{Array{Union{Missing, Float64},1}}(Union{Missing, Float64}[missing])

julia> mean(collect(skipmissing(x)))
NaN
```

Maybe a specialized implementation of `mean` in case when `eltype` of the collection is specified should be implemented? This is the case of `skipmissing` which knows that its `eltype` is `Float64`. For instance `sum` handles this case correctly (there is a special `mapreduce` for this case with `skipmissing`). Probably more functions in `Statistics` should be reviewed, but first I would want to understand what is the intention with `mean`.

JeffreySarnoff:
same with `cor` `cov` `std` `stdm` `var` `varm`

btw ` cov(collect(skipmissing(x))) === -0.0`

bkamins:
I can fix it, but I am not sure what do we want. Should empty collections always throw an error, or they should produce `NaN` if element type of the collection is `<:Number`? It seems that the errors were consciously hard coded>

Examples of current inconsistencies (following what was noted by @JeffreySarnoff ):
* `cov(Int[]) === -0.0`
* `var(Int[]) === NaN`
* `var(skipmissing(Int[]))` is an argument error
* `cov(skipmissing(Int[]))` is a method error

CC @nalimilan

nalimilan:
I agree that reductions over `skipmissing(::Array{T})` should return the same value as reductions over `::Array{Missings.T(T)}`. Since we can't change the return value of functions which currently don't throw, I guess we should just copy whatever behavior is currently implemented. If it's inconsistent, let's track that in a separate issue for 2.0.

Relevant: https://github.com/JuliaLang/julia/issues/5234

bkamins:
@gdkrmr: following what @nalimilan suggested currently we will fix `mean(skipmissing(y))` case only because it can be considered a bug. Other scenarios will have to wait for Julia 2.0 for unification.

[...]

PR #35843, opened 2020-05-11
Implement reduce_empty(op, Union{})

tkf:
Interestingly, the invalidation hunt #35733 actually fixed a bug in `reduce_empty`.  As #35733 may not be merged as-is, I extracted the bug fix part from #35733 and added a test to it.

It requires https://github.com/JuliaLang/Statistics.jl/pull/36


tkf:
@timholy's feedback in https://github.com/JuliaLang/www.julialang.org/pull/794#discussion_r424681687:

> I don't think that's a bug. We have lots of methods that result in an ambiguity error if you use `Union{}` as a type. Try `zero`, `one`, etc.

OK maybe saying it's a "bug" is too much.  But I _think_ it is a fault of the caller to invoke `one(Union{})`. I don't think it's a fault of caller to call `sum(Union{}[])` because `sum` should handle any iterable object and give sensible error when it doesn't work. `Union{}[]` is a reasonable input for any function on iterables because it is the "zero" of the binary "operator" `append!`/`vcat`. So it's the "best" empty vector to use as the initial state for constructing a vector that might contain numbers of unknown type. I don't think revealing the internal gut when a reasonable input is given is ideal.

tkf:
This also fixes a bug mentioned in https://github.com/JuliaLang/julia/issues/28777#issuecomment-485006981

On Julia 1.6-DEV:

```julia
julia> reduce(*, ())
""
```

After this PR:

```julia
julia> reduce(*, ())
```

nalimilan (approved):
Looks good to me. Though note that since (as you showed) `reduce(*, ()) == ""` currently, making it throw an error may technically break some code.

PR #35845, opened 2020-05-12
Inference annotations that can be read by an external tool

Keno:
This builds on top of #35831, letting inference emit a custom message
whenever it gives up on infering something. These messages are intended
to be displayed by external tools, either to debug what inference is doing
(e.g. for Cthulhu) or, if an external compiler needs to disallow certain
kinds of missing information (e.g. no dynamic dispatch on GPUs), can be
used to improve diagnostics. This is mostly a proof of concept, I think
these messages/hooks need to be a bit richer for a full solution, though
I think this can be already useful if we hook up something like Cthulhu.
As a proof of concept, I hacked up a 10 line function that reads these messagse.
It works something like the following:

```
function bar()
    sin = eval(:sin)
    sin(1)
end
foo() = bar()
```

```
julia> Compiler3.analyze_static(Tuple{typeof(foo)})
In function: bar()
ERROR: The called function was unknown
1| function bar()
2|     sin = eval(:sin)
=>     sin(1)
4| end

[1] In foo()
```

The reason this needs to sit on top of #35831 is that it needs to run
with a completely fresh cache, which #35831 provides the primitives for.
Otherwise, while you could still get the annotations out, that would
only work the first time something is inferred anywhere in the system.
With a fresh cache, everything is analyzed again, and any messages like
these that are opted in to can be collected.

cc @vchuravy for Cthulhu, @maleadt @jpsamaroo for GPU-related concerns

### DIFF
--- a/base/compiler/abstractinterpretation.jl
+++ b/base/compiler/abstractinterpretation.jl
@@ -36,7 +36,10 @@ end
 function abstract_call_gf_by_type(interp::AbstractInterpreter, @nospecialize(f), argtypes::Vector{Any}, @nospecialize(atype), sv::InferenceState,
                                   max_methods::Int = InferenceParams(interp).MAX_METHODS)
     mt = ccall(:jl_method_table_for, Any, (Any,), atype)
-    mt === nothing && return Any
+    if mt === nothing
+        add_remark!(interp, sv, "Could not identify method table for call")
+        return Any
+    end
     mt = mt::Core.MethodTable
     min_valid = UInt[typemin(UInt)]
     max_valid = UInt[typemax(UInt)]
@@ -48,7 +51,10 @@ function abstract_call_gf_by_type(interp::AbstractInterpreter, @nospecialize(f),
         for sig_n in splitsigs
             xapplicable = matching_methods(sig_n, sv.matching_methods_cache, max_methods,
                                            get_world_counter(interp), min_valid, max_valid)
-            xapplicable === false && return Any
+            if xapplicable === false
+                add_remark!(interp, sv, "For one of the union split cases, too many methods matched")
+                return Any
+            end
             append!(applicable, xapplicable)
         end
     else
@@ -57,6 +63,7 @@ function abstract_call_gf_by_type(interp::AbstractInterpreter, @nospecialize(f),
         if applicable === false
             # this means too many methods matched
             # (assume this will always be true, so we don't compute / update valid age in this case)
+            add_remark!(interp, sv, "Too many methods matched")
             return Any
         end
     end
@@ -84,6 +91,7 @@ function abstract_call_gf_by_type(interp::AbstractInterpreter, @nospecialize(f),
         sig = match[1]
         if istoplevel && !isdispatchtuple(sig)
             # only infer concrete call sites in top-level expressions
+            add_remark!(interp, sv, "Refusing to infer non-concrete call site in top-level expression")
             rettype = Any
             break
         end
@@ -135,6 +143,7 @@ function abstract_call_gf_by_type(interp::AbstractInterpreter, @nospecialize(f),
         end
     end
     if is_unused && !(rettype === Bottom)
+        add_remark!(interp, sv, "Call result type was widened because the return value is unused")
         # We're mainly only here because the optimizer might want this code,
         # but we ourselves locally don't typically care about it locally
         # (beyond checking if it always throws).
@@ -288,8 +297,11 @@ function abstract_call_method_with_const_args(interp::AbstractInterpreter, @nosp
     return result
 end
 
+const RECURSION_UNUSED_MSG = "Bounded recursion detected with unused result. Annotated return type may be wider than true result."
+
 function abstract_call_method(interp::AbstractInterpreter, method::Method, @nospecialize(sig), sparams::SimpleVector, hardlimit::Bool, sv::InferenceState)
     if method.name === :depwarn && isdefined(Main, :Base) && method.module === Main.Base
+        add_remark!(interp, sv, "Refusing to infer into `depwarn`")
         return Any, false, nothing
     end
     topmost = nothing
@@ -314,6 +326,7 @@ function abstract_call_method(interp::AbstractInterpreter, method::Method, @nosp
                 # avoid widening when detecting self-recursion
                 # TODO: merge call cycle and return right away
                 if call_result_unused(sv)
+                    add_remark!(interp, sv, RECURSION_UNUSED_MSG)
                     # since we don't use the result (typically),
                     # we have a self-cycle in the call-graph, but not in the inference graph (typically):
                     # break this edge now (before we record it) by returning early
@@ -394,6 +407,7 @@ function abstract_call_method(interp::AbstractInterpreter, method::Method, @nosp
             # continue inference, but note that we've limited parameter complexity
             # on this call (to ensure convergence), so that we don't cache this result
             if call_result_unused(sv)
+                add_remark!(interp, sv, RECURSION_UNUSED_MSG)
                 # if we don't (typically) actually care about this result,
                 # don't bother trying to examine some complex abstract signature
                 # since it's very unlikely that we'll try to inline this,
@@ -595,8 +609,8 @@ function abstract_apply(interp::AbstractInterpreter, @nospecialize(itft), @nospe
     aftw = widenconst(aft)
     if !isa(aft, Const) && (!isType(aftw) || has_free_typevars(aftw))
         if !isconcretetype(aftw) || (aftw <: Builtin)
-            # non-constant function of unknown type: bail now,
-            # since it seems unlikely that abstract_call will be able to do any better after splitting
+            add_remark!(interp, sv, "Core._apply called on a function of a non-concrete type")
+            # bail now, since it seems unlikely that abstract_call will be able to do any better after splitting
             # this also ensures we don't call abstract_call_gf_by_type below on an IntrinsicFunction or Builtin
             return Any
         end
@@ -920,6 +934,7 @@ function abstract_call(interp::AbstractInterpreter, fargs::Union{Nothing,Vector{
         # non-constant function, but the number of arguments is known
         # and the ft is not a Builtin or IntrinsicFunction
         if typeintersect(widenconst(ft), Builtin) != Union{}
+            add_remark!(interp, sv, "Could not identify method table for call")
             return Any
         end
         return abstract_call_gf_by_type(interp, nothing, argtypes, argtypes_to_type(argtypes), sv, max_methods)
--- a/base/compiler/typeinfer.jl
+++ b/base/compiler/typeinfer.jl
@@ -148,7 +148,8 @@ function cache_result!(interp::AbstractInterpreter, result::InferenceResult, min
 
     # TODO: also don't store inferred code if we've previously decided to interpret this function
     if !already_inferred
-        code_cache(interp)[result.linfo] = CodeInstance(result, min_valid, max_valid)
+        code_cache(interp)[result.linfo] = CodeInstance(result, min_valid, max_valid,
+            may_compress(interp), may_discard_trees(interp))
     end
     unlock_mi_inference(interp, result.linfo)
     nothing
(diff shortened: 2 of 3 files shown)


Keno:
Can I go ahead with this? Not having it on master is causing me lots of rebase issues when I want to make patches against master from my branch.

base/compiler/typeinfer.jl
```diff
@@ -148,7 +147,8 @@ function cache_result!(interp::AbstractInterpreter, result::InferenceResult, min
 
     # TODO: also don't store inferred code if we've previously decided to interpret this function
     if !already_inferred
-        code_cache(interp)[result.linfo] = CodeInstance(result, min_valid, max_valid)
+        code_cache(interp)[result.linfo] = CodeInstance(result, min_valid, max_valid,
+            may_compress(interp), may_discard_trees(interp))
```
vtjnash, line 151: Seems like these got accidentally added from a different commit, but otherwise seems okay

base/compiler/types.jl
```diff
@@ -189,3 +188,14 @@ lock_mi_inference(ni::NativeInterpreter, mi::MethodInstance) = (mi.inInference =
     See lock_mi_inference
 """
 unlock_mi_inference(ni::NativeInterpreter, mi::MethodInstance) = (mi.inInference = false; nothing)
+
+"""
+Emit an annotation when inference widens a result. These annotations are ignored
+by the native interpreter, but can be used by external tooling to annotate
+inference results.
+"""
+mark_dynamic!(ni::NativeInterpreter, sv, s) = nothing
```
vtjnash, line 197: This naming seems rather poor, since there's not really a sense in which the places this got called is related to "when inference widens a result". The only place that happens is `tmerge`. Perhaps this should be called simply `add_remark`?

Keno: Fair enough. The idea here was that this answers the question "why did a dynamic dispatch get introduced here". I'm fine with just generically calling it `add_remark` though. Eventually we might want to have multiple classes of remarks, not just about inference precision, so that seems like a fine direction to go.

base/compiler/typeinfer.jl
Keno: Yeah, these are useful to have the remark emitting AbstractInterpreter force results to be kept such that it can correlate the remarks with the original source, but you're right, it's conceptually separate.

Keno:
Renamed to `add_remark`

base/compiler/typeinfer.jl
vtjnash: Ah, maybe just needs a separate commit then, or at least mention in the commit message, so we don't drop this later?

I'd also perhaps have preferred we just added these to the existing configuration set for this (InferenceParams), rather than creating a duplicate strategy with dispatch.

base/compiler/abstractinterpretation.jl
```diff
@@ -327,6 +343,7 @@ function abstract_call_method(interp::AbstractInterpreter, method::Method, @nosp
                 # avoid widening when detecting self-recursion
                 # TODO: merge call cycle and return right away
                 if call_result_unused(sv)
+                    add_remark!(interp, sv, "Widening unused call result early to avoid unbounded recursion")
```
vchuravy, line 346:
```suggestion
                    add_remark!(interp, sv, "Widening unused call result early to avoid unbounded self-recursion")
```

Would be good if the annotations are unique so that one can grep for them. Not sure if my recommendation is the best here.

vchuravy (approved):

base/compiler/types.jl
```diff
@@ -189,3 +188,14 @@ lock_mi_inference(ni::NativeInterpreter, mi::MethodInstance) = (mi.inInference =
     See lock_mi_inference
 """
 unlock_mi_inference(ni::NativeInterpreter, mi::MethodInstance) = (mi.inInference = false; nothing)
+
+"""
+Emit an annotation when inference widens a result. These annotations are ignored
```
vtjnash, line 193:
```suggestion
Emit an analysis remark during inference for the current line (`sv.pc`). These annotations are ignored
```

base/compiler/abstractinterpretation.jl
```diff
@@ -37,9 +37,17 @@ function abstract_call_gf_by_type(interp::AbstractInterpreter, @nospecialize(f),
                                   max_methods::Int = InferenceParams(interp).MAX_METHODS)
     atype_params = unwrap_unionall(atype).parameters
     ft = unwrap_unionall(atype_params[1]) # TODO: ccall jl_method_table_for here
-    isa(ft, DataType) || return Any # the function being called is unknown. can't properly handle this backedge right now
+    if !isa(ft, DataType)
+        # can't properly handle this backedge right now
+        add_remark!(interp, sv, "the function being called is unknown")
```
vtjnash, line 42:
```suggestion
        add_remark!(interp, sv, "Could not identify method table for call")
```

base/compiler/abstractinterpretation.jl
```diff
@@ -608,8 +626,8 @@ function abstract_apply(interp::AbstractInterpreter, @nospecialize(itft), @nospe
     aftw = widenconst(aft)
     if !isa(aft, Const) && (!isType(aftw) || has_free_typevars(aftw))
         if !isconcretetype(aftw) || (aftw <: Builtin)
-            # non-constant function of unknown type: bail now,
-            # since it seems unlikely that abstract_call will be able to do any better after splitting
+            add_remark!(interp, sv, "non-constant function of unknown type")
```
vtjnash, line 629:
```suggestion
            add_remark!(interp, sv, "Core._apply called on a function of a non-concrete type")
```

base/compiler/abstractinterpretation.jl
```diff
@@ -407,6 +424,7 @@ function abstract_call_method(interp::AbstractInterpreter, method::Method, @nosp
             # continue inference, but note that we've limited parameter complexity
             # on this call (to ensure convergence), so that we don't cache this result
             if call_result_unused(sv)
+                add_remark!(interp, sv, "Widening unused call result early to avoid unbounded recursion")
```
vtjnash, line 427: This wording is misleading: there's never widening that happens here (except if the code-path would be discovered to stack overflow instead), since the result is known to be unused, though we also don't know what will happen next (though it's unlikely inlining would succeed regardless due to the graph cycle).
```suggestion
                add_remark!(interp, sv, "Delaying call result analysis until after optimization due to unbounded recursion")
```

base/compiler/abstractinterpretation.jl
vtjnash: As commented below, no widening is happening here (that's why we're okay with returning this early)

```suggestion
                    add_remark!(interp, sv, "Delaying call result analysis until after optimization due to bounded recursion")
```

base/compiler/abstractinterpretation.jl
```diff
@@ -148,6 +162,7 @@ function abstract_call_gf_by_type(interp::AbstractInterpreter, @nospecialize(f),
         end
     end
     if is_unused && !(rettype === Bottom)
+        add_remark!(interp, sv, "Call result was widened, because the return type is unused")
```
vtjnash, line 165:
```suggestion
        add_remark!(interp, sv, "Call result type was widened because the return value is unused")
```

vtjnash (commented):
Some alternative wording suggestions

base/compiler/abstractinterpretation.jl
```diff
@@ -49,6 +57,7 @@ function abstract_call_gf_by_type(interp::AbstractInterpreter, @nospecialize(f),
         if !isa(tname, DataType)
             # can't track the backedge to the ctor right now
             # for things like Union
+            add_remark!(interp, sv, "The called constructor is too complicated")
```
vtjnash, line 60:
```suggestion
            add_remark!(interp, sv, "Deciding to skip computing the result due to arbitrary reasons")
```

I have no clue what this `ftname === _TYPE_NAME` condition means (from 0b8f0b7dc40), but looks really arbitrarily random these days to return from here.

base/compiler/abstractinterpretation.jl
```diff
@@ -37,9 +37,17 @@ function abstract_call_gf_by_type(interp::AbstractInterpreter, @nospecialize(f),
...
+    if !isa(ft, DataType)
+        # can't properly handle this backedge right now
+        add_remark!(interp, sv, "the function being called is unknown")
+        return Any
+    end
     ftname = ft.name
-    isdefined(ftname, :mt) || return Any # not callable. should be Bottom, but can't track this backedge right now
+    if !isdefined(ftname, :mt)
+        # should be Bottom, but can't track this backedge right now
+        add_remark!(interp, sv, "The function is not callable")
```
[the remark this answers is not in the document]
Keno: I think it's worth having distinct messages for every call site, so people can grep the inference source and see what happened if necessary.

base/compiler/abstractinterpretation.jl
Keno: Are you saying we should just remove this whole code path? The suggested message seems like it could cause frustration for people seeing it.

base/compiler/abstractinterpretation.jl
Keno: Are we really delaying analysis, aren't we just skipping it entirely? When will we come back to do this analysis?

Keno:
I've applied most of the suggestions and asked two questions in others (will probably still apply those, but I'd like to understand better).

base/compiler/abstractinterpretation.jl
vtjnash: Yeah, I couldn't tell you why these specific conditions are so bad. "Too complicated" seemed like it was trying to be too PC and cause the same frustrations.

Keno: I'll put up a PR to remove it and see what breaks :)

base/compiler/abstractinterpretation.jl
vtjnash: We're skipping this here because we know it's already being analyzed and we don't care about the result.

base/compiler/types.jl
vtjnash: Yeah, but the places we're putting it (in inference) aren't introducing dynamic dispatch (that's always there), they're merely decision points where we prune our analysis of the call graph.

base/compiler/abstractinterpretation.jl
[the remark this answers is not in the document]
vtjnash: Yeah, but the only reason these are separate is that we're doing a bad job at reimplementing `jl_method_table_for` here. These reasons shouldn't actually be distinguishable.

base/compiler/abstractinterpretation.jl
Keno: Ok, but we never go back to actually annotate that result, right? In that case, I'd like to inform the user that the return type they'd be seeing in Cthulhu may not match what inference knows about this callsite. Maybe something like: "Unbounded recursion detected with unused result. Annotated return type may be wider than true result".

base/compiler/abstractinterpretation.jl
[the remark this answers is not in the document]
Keno: Ok, fair enough. What's the reason we're not calling `jl_method_table_for` here? Or is there just not good reason?

base/compiler/types.jl
Keno: Depends on what you mean by dynamic dispatch ;). I meant calls to jl_apply_generic which is often what happens when we prune the analysis here. The reason to care about that is that that'd be what's disallowed in e.g. a GPU or XLA context. Anyway, add_remark! is fine.

base/compiler/abstractinterpretation.jl
vtjnash: This is bounded recursion (in the type domain), but yeah, we're putting 'Any' here specifically because we know it can't be used by the optimizer anyways, and thus want to avoid the re-work costs of putting the real type here later.

base/compiler/abstractinterpretation.jl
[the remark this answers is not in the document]
vtjnash: Just got forgotten, I'm guessing

base/compiler/abstractinterpretation.jl
Keno: So "Unbounded recursion detected with unused result. Annotated return type may be wider than true result."? Or just leave off the remark entirely and have Cthulhu check when it seems a situation like this?

Keno:
Alright, rebased on top of #36477 and split out the may_optimize stuff into a separate commit. Also adjusted the recursion message as discussed. Will plan to merge once #36477 is in.

Issue #35638, opened 2020-04-29
Failed to compile on Fedora 32


xijiang:
Fedora 32 (just upgraded)
Pulled here a few hours ago.
Compile stopped after two errors:
1. julia/contrib/normalize_triplet.py", line 112, in <module>
    libgfortran_version = {
    KeyError: '1'
2. curl: (22) The requested URL returned error: 404 Not Found
    julia/deps/suitesparse.mk:144: 
    julia/deps/srccache/SuiteSparse.v5.4.0-6..tar.gz] Error 22

fredrikekre:
What `gfortran` version do you have installed?

xijiang:
GNU Fortran (GCC) 10.0.1 20200328 (Red Hat 10.0.1-0.11)

fredrikekre:
I guess https://github.com/JuliaLang/julia/blob/18674303bdfb5670f3bd36dd0e6ba0e6cf2bcd8b/contrib/normalize_triplet.py#L113-L118 has to be extended to include version 10.

xijiang:
I pulled this morning. It still has the same errors.

> Traceback (most recent call last):
  File "/home/xijiang/Music/julia/contrib/normalize_triplet.py", line 112, in <module>
    libgfortran_version = {
KeyError: '1'
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   453  100   453    0     0   1043      0 --:--:-- --:--:-- --:--:--  1043
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 404 Not Found
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 404 Not Found
make[1]: *** [/home/xijiang/Music/julia/deps/suitesparse.mk:144: /home/xijiang/Music/julia/deps/srccache/SuiteSparse.v5.4.0-6..tar.gz] Error 22
make: *** [Makefile:60: julia-deps] Error 2

KristofferC:
https://github.com/JuliaLang/julia/pull/35653 has not been merged yet

PR #35849, opened 2020-05-12
(1.4-only) Add GCC10 to GCC version mappings

imciner2:
This is the 1.4-specific fix for issue https://github.com/JuliaLang/julia/issues/35638. There is slightly more than just adding the new GCC version, since the old code assumed that the GCC major version would only be 1 digit, so the matching expression had to be updated to fix that.


imciner2:
The first test failure is reporting a sha512 mismatch for OpenBLAS.

KristofferC:
Changed to against release-1.4 branch

PR #35851, opened 2020-05-12
Note that views no longer allocate in NEWS

jtravs:


JeffBezanson:
If we really want to add this at this point, it needs to be added to HISTORY.md on master as well. I would also merge it with the previous item, since it is a consequence of that. "As a result, array views no longer allocate"

jtravs:
OK, I defer to you. I simply wanted this called out as for me (and I suspect many others) non-allocating array views are quite a big deal, and I had missed it.

ViralBShah:
We should definitely get this into the release notes.

PR #35854, opened 2020-05-12
add (un)signed(::Type{Bool}) method

simeonschaub:
Since `signed(true)` already works and returns an `Int`, I think this was just an omission.

### DIFF
--- a/base/int.jl
+++ b/base/int.jl
@@ -70,6 +70,7 @@ julia> signed(UInt64)
 Int64
 ```
 """
+signed(::Type{Bool}) = Int
 signed(::Type{UInt8}) = Int8
 signed(::Type{UInt16}) = Int16
 signed(::Type{UInt32}) = Int32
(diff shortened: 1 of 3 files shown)


StefanKarpinski:
Bool is already effectively an unsigned type. Not sure if we should got to 8 bits or straight to U/Int.

simeonschaub:
Wouldn't it be inconsistent if `signed(Bool)` was different from `typeof(signed(true))`? Or are you talking about changing `signed(::Bool)` as well?

StefanKarpinski:
Ah, didn't realize that was already defined. They should certainly match.

JeffBezanson (approved):

StefanKarpinski:
Test `unsigned(Bool)` as well?

StefanKarpinski (approved):

StefanKarpinski:
This looks ready to merge once CI passes.

PR #35855, opened 2020-05-12
Add some broken tests for invalidation

timholy:
Given that Jeff already has code to fix some of the invalidation issues, I'm trying to wrap up my work on this topic for the time being. I think this distills most of the triggers for the weird cases I found while writing https://github.com/JuliaLang/www.julialang.org/pull/794.

There are just ~~three~~ six `@test_broken`s here, and obviously the goal would be to turn these into `@test`s. But I know that writing tests takes time, and I had a lot of material to work with, so I figured this would be a useful head start.

### DIFF
--- a/test/worlds.jl
+++ b/test/worlds.jl
@@ -199,3 +199,115 @@ end
 notify(c26506_1)
 wait(c26506_2)
 @test result26506[1] == 3
+
+
+## Invalidation tests
+
+function instance(f, types)
+    m = which(f, types)
+    inst = nothing
+    tt = Tuple{typeof(f), types...}
+    specs = m.specializations
+    if isa(specs, Nothing)
+    elseif isa(specs, Core.SimpleVector)
+        for i = 1:length(specs)
+            if isassigned(specs, i)
+                mi = specs[i]::Core.MethodInstance
+                if mi.specTypes === tt
+                    inst = mi
+                    break
+                end
+            end
+        end
+    else
+        Base.visit(specs) do mi
+            if mi.specTypes === tt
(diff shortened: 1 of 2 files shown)


test/worlds.jl
```diff
@@ -199,3 +199,144 @@ end
...
+    data::NTuple{L,T}
+    SHermitianCompact{N,T,L}(data::NTuple{L,T}) where {N,T,L} = new{N,T,L}(data)
+end
+@inline (::Type{SSC})(a::AbstractVector) where {SSC <: SHermitianCompact} = SSC((a...,))
+x = SHermitianCompact{1,Float32,3}([1.0f0, 2.0f0, 3.0f0])
+@test_broken worlds(mi) == w
+
+## A case in which there is no intersection between signatures
+## A possibly relevant point is that the invalidated instance comes from
+##    convert(::Type{T}, x) where {T>:Nothing} = convert(nonnothingtype_checked(T), x)
```
JeffBezanson, line 336: In this case `nonnothingtype_checked(T)` returns `Union{}`, so this calls `convert(Union{}, ::String)` which is ambiguous, so hopefully it's just another case of the ambiguity issue.

vtjnash:
Broken I assume means over-aggressive and not wrong?

I'm intending to pick up work on that now, and see how that goes with the internals.

timholy:
> Broken I assume means over-aggressive and not wrong?

Right. I just meant they were `@test_broken`, but I should have been clearer. I haven't yet seem a case where we fail to invalidate, and of course that would be more serious.

PR #35856, opened 2020-05-12
support random sampling of tuple types

stev47:
This allows e.g `rand(NTuple{5,Int})` to sample a tuple of 5 `Int`s.

The implementation simply assembles a tuple by calling `rand` on the
corresponding type parameters of the tuple type.
A generated function is used to ensure type stability.

Note that this is distinct from `rand(::Tuple)` to select a random element from a tuple.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -160,6 +160,7 @@ Standard library changes
   results table ([#38042]).
 * `@testset` now supports the option `verbose` to show the test result summary
   of the children even if they all pass ([#33755]).
+* `rand` now supports sampling of tuple types ([#35856]).
 
 #### LinearAlgebra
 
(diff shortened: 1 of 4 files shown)


stev47:
Rebased and added news entry.
Can we take a look at this for 1.6? @rfourquet @StefanKarpinski

stev47:
buildbot failures seem unrelated (network errors)

rfourquet:
I will have a closer look tentatively this week, but this won't make it for 1.6. In the meantime, the functionality can be found in [RandomExtensions](https://github.com/rfourquet/RandomExtensions.jl/) :) This means I personally definitely support this feature, but given that this package is not yet very popular, I suspect people are not dying for this functionality...

rfourquet:
Sorry for forgetting about this; the functionality was added in #50251, but I rebased this in #51630 to merge the tests, which caught the fact that we didn't support `rand(Tuple{})`. Thanks!

PR #35857, opened 2020-05-12
Display argument that is duplicated in function calls

### LINKED ISSUE #34084: "syntax: function argument names not unique"-error could give the argument(s) that is(are) duplicated
Seems like a small QoL improvement.

ssikdar1:
# Description
Fixes #34084

Now error message will also give the name of a variable duplicated in the function call.

# Testing
```
julia> foo !== foo = bar
ERROR: syntax: function argument name not unique: foo around REPL[1]:1
Stacktrace:
 [1] top-level scope at REPL[1]:1
```

```
julia> function lala(;  bar=3, bar=4, boo=5)
       end
ERROR: syntax: function argument name not unique: bar around REPL[2]:1
Stacktrace:
```

### DIFF
--- a/src/julia-syntax.scm
+++ b/src/julia-syntax.scm
@@ -339,7 +339,7 @@
          (anames (map (lambda (x) (if (underscore-symbol? x) UNUSED x))
                       (llist-vars argl))))
      (if (has-dups (filter (lambda (x) (not (eq? x UNUSED))) anames))
-         (error "function argument names not unique"))
+         ((error (string "function argument names not unique " (cdr anames)))))
      (if (has-dups names)
          (error "function static parameter names not unique"))
      (if (any (lambda (x) (and (not (eq? x UNUSED)) (memq x names))) anames)
--- a/src/julia-syntax.scm
+++ b/src/julia-syntax.scm
@@ -336,8 +336,8 @@
                                    (append req opt vararg) rett)))))
    ;; no optional positional args
    (let* ((names (map car sparams))
-         (anames (map (lambda (x) (if (underscore-symbol? x) UNUSED x))(llist-vars argl)))
-         (unused_anames (filter (lambda (x) (not (eq? x UNUSED))) anames)))
+          (anames (map (lambda (x) (if (underscore-symbol? x) UNUSED x)) (llist-vars argl)))
+          (unused_anames (filter (lambda (x) (not (eq? x UNUSED))) anames)))
      (if (has-dups unused_anames)
          (error (string "function argument name not unique: \"" (car (has-dups unused_anames)) "\"")))
      (if (has-dups names)
--- a/src/julia-syntax.scm
+++ b/src/julia-syntax.scm
@@ -335,11 +335,12 @@
          (optional-positional-defs name sparams req opt dfl body
                                    (append req opt vararg) rett)))))
    ;; no optional positional args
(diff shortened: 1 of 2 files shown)


ssikdar1:
Seems many doctests / tests need to be modified, I can do that if the change to error message is aceptable?

src/julia-syntax.scm
```diff
@@ -335,11 +335,12 @@
          (optional-positional-defs name sparams req opt dfl body
                                    (append req opt vararg) rett)))))
    ;; no optional positional args
-   (let ((names (map car sparams))
+   (let* ((names (map car sparams))
          (anames (map (lambda (x) (if (underscore-symbol? x) UNUSED x))
```
JeffBezanson, line 339: Need to fix the indenting here.

src/julia-syntax.scm
```diff
@@ -335,11 +335,12 @@
...
-                      (llist-vars argl))))
-     (if (has-dups (filter (lambda (x) (not (eq? x UNUSED))) anames))
-         (error "function argument names not unique"))
+                      (llist-vars argl)))
+         (unused_anames (filter (lambda (x) (not (eq? x UNUSED))) anames)))
+     (if (has-dups unused_anames)
+         ((error (string "function argument name not unique: " (car (has-dups unused_anames))))))
```
JeffBezanson, line 343:
```suggestion
         (error (string "function argument name not unique: " (car (has-dups unused_anames)))))
```

JeffBezanson:
I think we usually quote variable names in error messages, so how about `Function argument name "x" not unique`.

src/julia-syntax.scm
ssikdar1: Accepted suggestion in 904dfe1

src/julia-syntax.scm
ssikdar1: Indentation fixed 5126389

ssikdar1:
> I think we usually quote variable names in error messages, so how about `Function argument name "x" not unique`.
kk! Added in d10dfbb and unit test fixed in 7d17ada

src/julia-syntax.scm
```diff
@@ -335,11 +335,11 @@
          (optional-positional-defs name sparams req opt dfl body
                                    (append req opt vararg) rett)))))
    ;; no optional positional args
-   (let ((names (map car sparams))
-         (anames (map (lambda (x) (if (underscore-symbol? x) UNUSED x))
-                      (llist-vars argl))))
-     (if (has-dups (filter (lambda (x) (not (eq? x UNUSED))) anames))
-         (error "function argument names not unique"))
+   (let* ((names (map car sparams))
+         (anames (map (lambda (x) (if (underscore-symbol? x) UNUSED x))(llist-vars argl)))
```
JeffBezanson, line 339:
```suggestion
         (anames (map (lambda (x) (if (underscore-symbol? x) UNUSED x)) (llist-vars argl)))
```
Plus the indentation is still wrong --- the start of each `(var value)` pair in the let* should line up. Any good code editor should have a scheme mode that does this automatically.

ssikdar1: Sorry, was using vi to edit the file. Used sublime editor to do the indentation automatically and added the space in suggestion in 02a087a

JeffBezanson:
Thanks for taking the time to fix this up. Appreciated!

PR #35862, opened 2020-05-13
fix #34824, improve various cases of showing method signatures

### LINKED ISSUE #34824: invalid error message for type parameter

JeffBezanson:
The code for this was duplicated several times (more than 3, depending how you count), with all sorts of problems. Here are some examples.

Before:
```
julia> methods(Complex)
# 7 methods for type constructor:
[1] (::Type{Complex})(x::Real) in Base at complex.jl:16
...

julia> methods(Complex{Int})
# 7 methods for type constructor:
[1] (::Type{Complex{T}})(x::AbstractIrrational) where T<:Real in Base at irrationals.jl:34
...

julia> (Complex{T} where T<:Integer)(1.2)
ERROR: MethodError: no method matching Complex{T} where T<:Integer(::Float64)
Closest candidates are:
  Complex{T} where T<:Integer(::T) where T<:Number at boot.jl:715

julia> (::Union{Nothing,Int})(::Nothing) = nothing

julia> nothing("")
ERROR: MethodError: no method matching (::Nothing)(::String)
Closest candidates are:
  Any(::Nothing) at REPL[6]:1
```

After:
```
julia> methods(Complex)
# 7 methods for type constructor:
[1] Complex(x::Real) in Base at complex.jl:16
...

julia> methods(Complex{Int})
# 7 methods for type constructor:
[1] Complex{T}(x::AbstractIrrational) where T<:Real in Base at irrationals.jl:47
...

julia> (Complex{T} where T<:Integer)(1.2)
ERROR: MethodError: no method matching (Complex{T} where T<:Integer)(::Float64)
Closest candidates are:
  (::Type{T})(::T) where T<:Number at boot.jl:715

julia> (::Union{Nothing,Int})(::Nothing) = nothing

julia> nothing("")
ERROR: MethodError: no method matching (::Nothing)(::String)
Closest candidates are:
  (::Union{Nothing, Int64})(::Nothing) at REPL[1]:1
```

Fixes #34824. Replaces #35311.

Discussion questions:
- I decided to print methods on unionalls as `(Foo{T} where T)(x)` even though that is not currently valid input syntax. With #35861 it would be, so I think it's ok.
- The "closest candidates" printing used to show the *called* type in the head of each signature, instead of the type from the definition (see third example above). That was probably intentional and I can see a case for it, but I think it's confusing because it looks like it's showing definitions that don't actually exist. I think we should show the same thing that e.g. `methods(t)` shows.

### DIFF
--- a/base/show.jl
+++ b/base/show.jl
@@ -1764,6 +1764,7 @@ function show_signature_function(io::IO, @nospecialize(ft), demangle=false, farg
             print(io, "($fargname::", ft, ")")
         end
     end
+    nothing
 end
 
 function show_tuple_as_call(io::IO, name::Symbol, sig::Type, demangle=false, kwargs=nothing)
--- a/base/show.jl
+++ b/base/show.jl
@@ -1742,6 +1742,30 @@ function demangle_function_name(name::AbstractString)
     return name
 end
 
+# show the called object in a signature, given its type `ft`
+# `io` should contain the UnionAll env of the signature
+function show_signature_function(io::IO, @nospecialize(ft), demangle=false, fargname="", html=false)
+    uw = unwrap_unionall(ft)
+    if ft <: Function && isa(uw, DataType) && isempty(uw.parameters) &&
+        isdefined(uw.name.module, uw.name.mt.name) &&
+        ft == typeof(getfield(uw.name.module, uw.name.mt.name))
+        print(io, (demangle ? demangle_function_name : identity)(uw.name.mt.name))
+    elseif isa(ft, DataType) && ft.name === Type.body.name &&
+        (f = ft.parameters[1]; !isa(f, TypeVar))
+        uwf = unwrap_unionall(f)
+        parens = isa(f, UnionAll) && !(isa(uwf, DataType) && f === uwf.name.wrapper)
+        parens && print(io, "(")
+        show(io, f)
+        parens && print(io, ")")
+    else
+        if html
+            print(io, "($fargname::<b>", ft, "</b>)")
+        else
+            print(io, "($fargname::", ft, ")")
+        end
+    end
+end
+
 function show_tuple_as_call(io::IO, name::Symbol, sig::Type, demangle=false, kwargs=nothing)
     # print a method signature tuple for a lambda definition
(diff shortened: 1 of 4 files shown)


vtjnash (approved):

base/show.jl
```diff
@@ -1742,6 +1742,30 @@ function demangle_function_name(name::AbstractString)
...
+        show(io, f)
+        parens && print(io, ")")
+    else
+        if html
+            print(io, "($fargname::<b>", ft, "</b>)")
+        else
+            print(io, "($fargname::", ft, ")")
+        end
+    end
+end
```
vtjnash, line 1767:
```suggestion
    nothing
end
```
(otherwise may be odd this function signature was ::union(nothing, false))

PR #35868, opened 2020-05-13
Migrated the docstring of unsigned method to next to the method.

ravibitsgoa:
Added a few doctests for Core.Int.


JeffBezanson (approved):

fredrikekre:
This attaches the docstring to the ::Int8 method specifically instead of the function in general (as the current code does).

JeffBezanson:
Good point. There are two cases, for passing a type or a value. How should we document that? Just combine both into a single overall docstring? Or can we document something like `signed(::Type{<:Integer})`?

The top line is also wrong; it says `signed(T::Integer)` but should say `signed(T::Type{<:Integer})` (for both functions).

ravibitsgoa:
I think the case of passing a value to the signed/unsigned methods is already correctly documented separately. I will fix the top line in the docstrings.

base/int.jl
```diff
@@ -68,8 +54,14 @@ julia> signed(UInt16)
 Int16
 julia> signed(UInt64)
 Int64
+julia> signed(Int16)
+Int16
+julia> signed(Int64)
+Int64
 ```
 """
+function signed(::Type{<:Integer}) end
```
JeffBezanson, line 63:
```suggestion
signed(::Type{<:Integer})
```

ravibitsgoa: https://docs.julialang.org/en/v1/manual/documentation/index.html#Functions-and-Methods-2
This says first syntax is preferred.

ravibitsgoa:
Hi, 
Can someone review/merge this PR?

StefanKarpinski:
It was conflicted, which I resolved, hopefully correctly? I guess we'll see if it passes CI.

base/multinverses.jl
```diff
@@ -6,6 +6,20 @@ import Base: div, divrem, rem, unsigned
...
+Convert an integer bitstype to the unsigned type of the same size.
+# Examples
+```jldoctest
+julia> unsigned(Int16)
+UInt16
+julia> unsigned(UInt64)
+UInt64
+```
+"""
+function unsigned(::Type{<:Integer}) end
```
vtjnash, line 21:
```suggestion
unsigned(::Type{<:Integer})
```

vtjnash (changes requested):
This is too early in bootstrapping to use the fancy docstring versions (which can handle this syntax form). This will need to go into `base/docs/basedocs.jl`

base/int.jl
```diff
@@ -44,22 +44,8 @@ const BitUnsigned64T   = Union{Type{UInt8}, Type{UInt16}, Type{UInt32}, Type{UIn
...
-julia> unsigned(Int16)
-UInt16
-julia> unsigned(UInt64)
-UInt64
-```
-""" unsigned
-
-"""
-    signed(T::Integer)
+    signed(T::Type{<:Integer})
```
vtjnash, line 48: (this would align the behavior of `signed(Integer)` with `signed(Signed)`, so I think it makes reasonable enough sense)
```suggestion
signed(T::Type{<:Integer}) = Signed
```

base/multinverses.jl
vtjnash, line 21:
```suggestion
unsigned(::Type{<:Integer}) = Unsigned
```

PR #35871, opened 2020-05-13
Fix ImmutableDict(:a => 1, :a => 2)[:a]

tkf:
Currently, the behavior of `Dict(pairs...)` and
`Base.ImmutableDict(pairs...)` are different:

    julia> Dict(:a => 1, :a => 2)[:a]
    2

    julia> Base.ImmutableDict(:a => 1, :a => 2)[:a]
    1

This PR fixes the latter to return 2.

cc @simeonschaub 

As I think this is a bug fix and this feature did not exist in 1.4, I add the backport label.


simeonschaub (approved):
It's a bit unfortunate that `collect(ImmutableDict(v...)) == v` doesn't hold anymore, but I agree that being consistent with the `Dict` constructor here is more important.

PR #35872, opened 2020-05-13
change uuid to use  Random.RandomDevice() instead of Random.default_rng()

### LINKED ISSUE #35860: Use RandomDevice in UUIDs instead of default_rng?
@c42f commented on https://github.com/JuliaLang/julia/pull/35833#discussion_r424052968

> Actually this would suggest that the global RNG from `Random` isn't a great choice for `UUIDs` anyway because the user may control the seeding for reasons unrelated to UUIDs.

Maybe we should use `Random.RandomDevice()` instead of `Random.default_rng()` in UUIDs?

See also https://github.com/JuliaLang/julia/issues/32954#issuecomment-526580532

ssikdar1:
# Description
Fixes #35860 

Change `Random.default_rng()`  in the UUID functions to `Random.RandomDevice()`

# Testing

Ran `$ ./julia stdlib/UUIDs/test/runtests.jl `  with no issues

### DIFF
--- a/stdlib/UUIDs/src/UUIDs.jl
+++ b/stdlib/UUIDs/src/UUIDs.jl
@@ -48,10 +48,10 @@ julia> rng = MersenneTwister(1234);
 
 julia> uuid1(rng)
 UUID("cfc395e8-590f-11e8-1f13-43a2532b2fa8")
+```
 
 !!! compat "Julia 1.6"
     The default rng has been changed to use Random.RandomDevice as of Julia 1.6
-```
 """
 function uuid1(rng::AbstractRNG=Random.RandomDevice())
     u = rand(rng, UInt128)
@@ -88,10 +88,9 @@ julia> rng = MersenneTwister(1234);
 
 julia> uuid4(rng)
 UUID("196f2941-2d58-45ba-9f13-43a2532b2fa8")
-
+```
 !!! compat "Julia 1.6"
     The default rng has been changed to use Random.RandomDevice as of Julia 1.6
-```
 """
 function uuid4(rng::AbstractRNG=Random.RandomDevice())
     u = rand(rng, UInt128)
--- a/stdlib/UUIDs/src/UUIDs.jl
+++ b/stdlib/UUIDs/src/UUIDs.jl
@@ -50,7 +50,7 @@ julia> uuid1(rng)
 UUID("cfc395e8-590f-11e8-1f13-43a2532b2fa8")
 ```
 """
-function uuid1(rng::AbstractRNG=Random.default_rng())
+function uuid1(rng::AbstractRNG=Random.RandomDevice())
     u = rand(rng, UInt128)
 
     # mask off clock sequence and node
@@ -87,7 +87,7 @@ julia> uuid4(rng)
 UUID("196f2941-2d58-45ba-9f13-43a2532b2fa8")
 ```
 """
-function uuid4(rng::AbstractRNG=Random.default_rng())
+function uuid4(rng::AbstractRNG=Random.RandomDevice())
     u = rand(rng, UInt128)
     u &= 0xffffffffffff0fff3fffffffffffffff
     u |= 0x00000000000040008000000000000000
--- a/stdlib/UUIDs/src/UUIDs.jl
+++ b/stdlib/UUIDs/src/UUIDs.jl
@@ -36,11 +36,20 @@ const namespace_oid  = UUID(0x6ba7b8129dad11d180b400c04fd430c8) # 6ba7b812-9dad-
 const namespace_x500 = UUID(0x6ba7b8149dad11d180b400c04fd430c8) # 6ba7b814-9dad-11d1-80b4-00c04fd430c8
 
 """
-    uuid1([rng::AbstractRNG=Random.RandomDevice()]) -> UUID
+    uuid1([rng::AbstractRNG]) -> UUID
 
 Generates a version 1 (time-based) universally unique identifier (UUID), as specified
 by RFC 4122. Note that the Node ID is randomly generated (does not identify the host)
 according to section 4.5 of the RFC.
+
+The default rng used by `uuid1` is not `GLOBAL_RNG` and every invocation of `uuid1()` without
+an argument should be expected to return a unique identifier. Importantly, the outputs of
+`uuid1` do not repeat even when `Random.seed!(seed)` is called. Currently (as of Julia 1.6),
+`uuid1` uses `Random.RandomDevice` as the default rng. However, this is an implementation
+detail that may change in the future.
+
+!!! compat "Julia 1.6"
+    The output of `uuid1` does not dependent on `GLOBAL_RNG` as of Julia 1.6.
 
 # Examples
 ```jldoctest; filter = r"[a-z0-9]{8}-([a-z0-9]{4}-){3}[a-z0-9]{12}"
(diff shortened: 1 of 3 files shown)


StefanKarpinski (approved):

StefanKarpinski:
Looks good to me. A simple test that could be done after this is setting the global RNG seed to the same value twice and generating a new UUID by each of these methods after and checking that they are not the same. Before this, that would have failed, so that's a good regression test.

rfourquet:
Also, this needs docs and news. And maybe a "!!! compat" annotation saying that this new behavior starts in version 1.6.

JeffBezanson:
This may be the right decision but I believe RandomDevice can be reeeally slow --- are we sure this is ok?

tkf:
(pinging @Keno who had the same concern https://github.com/JuliaLang/julia/issues/32954#issuecomment-526588327)

I think the default should be the safe one.  If you know you need tons of random UUIDs, I think you can always swap the RNG argument.  But warning this in the docstring would be nice.

rfourquet:
I'm not totally sold on the idea either. There is something nice in knowing that generally all randomness comes from the `default_rng` or explicit RNGs and reproducibility can be achieved with `seed!`. Generally, the default one is seeded only in tests, and I would tend to consider it bad practice when it's seeded in a package (not in tests). So using `RandomDevice()` here goes in the way of the user who needs reproducibility (for whatever reason, e.g. debugging). On the other hand, it's easy for applications to explicitly use `RandomDevice()` for UUIDs. Using `RandomDevice()` here is a special case and opens the door to more special cases, and special cases increase complexity.

StefanKarpinski:
I think the danger of producing the same UUIDs repeatedly when they are supposed to be universally unique is way too high. This is definitely one of those things that has caused problems because people (myself included) just don't realize when they call `seed!` that they will cause the same UUIDs to be generated repeatedly. Super dangerous. Especially because of the decision which I don't think we can roll back at this point to automatically call `seed!` at the start of each testset. The other case that's also really bad is using a default RNG for generating names of temp files, but I think that has now been fixed?

StefanKarpinski:
Combined with the fact that UUIDs are sometimes relevant for security, this seems good to me.

JeffBezanson:
How about using a dedicated default RNG for UUIDs that's initialized from RandomDevice at startup?

StefanKarpinski:
I think that would be fine. It might be good to do something that is more secure as well, e.g. initialize a seed from `RandomDevice` on the first call and then do hash-based UUID generation combining that seed value with a counter or something like that.

tkf:
When UUIDs are used for something like #35833 (kind of `gensym`), it is required that the users can't reset the RNG, for the correctness of the program.  It's easy to explicitly use `RandomDevice`, though.

StefanKarpinski:
I feel like this is pretty simple and we can just do this and see if anyone actually cares about the performance of generating UUIDs.

stdlib/UUIDs/src/UUIDs.jl
```diff
@@ -53,7 +55,7 @@ UUID("cfc395e8-590f-11e8-1f13-43a2532b2fa8")
 !!! compat "Julia 1.6"
     The default rng has been changed to use Random.RandomDevice as of Julia 1.6
 """
-function uuid1(rng::AbstractRNG=Random.RandomDevice())
+function uuid1(rng::AbstractRNG=uuid_rng)
```
rfourquet, line 58: Why did you make this change? Creating a new `RandomDevice()` is extremely cheap on non-Windows, and on Windows having a global `uuid_rng` is not thread-safe...

ssikdar1: @rfourquet I guess misunderstood @JeffBezanson 's comment above?

> How about using a dedicated default RNG for UUIDs that's initialized from RandomDevice at startup?

rfourquet: My interpretation of it was using an RNG like `default_rng()` (currently `MersenneTwister()` which is thread-safely initialized), which is therefore immune against the unqualified `Random.seed!()` but way faster than `RandomDevice()`, but "_initialized_ from `RandomDevice()`" means it's _seeded_ from `RandomDevice()` at startup.

rfourquet: But Stefan's last comment suggests to keep it how it was, because the poor performance of `RandomDevice` might considered as a non-problem until someone complains.

ssikdar1: Okay! I reverted 507b454 . I'll leave it alone for now until there's something I should explicitly add/change/remove.

tkf:
Since it is possible that the default RNG to be changed, I think it'd be better to avoid putting `Random.RandomDevice()` in the signature.  Also, it makes future changes easier if we only describe the behavior and not the implementation.  How about something like this?

>     uuid4([rng::AbstractRNG]) -> UUID
>
> Generates a version 4 (random or pseudo-random) universally unique identifier (UUID), as specified by RFC 4122.
>
> The default `rng` used by `uuid4` is not `GLOBAL_RNG` and every invocation of `uuid4()` without an argument should be expected to return a unique identifier.  Importantly, the outputs of `uuid4` do not repeat even when `Random.seed!(seed)` is called. Currently (as of Julia 1.6), `uuid4` uses `Random.RandomDevice` as the default `rng`.  However, this is an implementation detail that may change in the future.
>
> compat "Julia 1.6"
> The output of `uuid4` does not dependent on `GLOBAL_RNG` as of Julia 1.6.

c42f (approved):
This looks great to me — simple and safe, with the side benefit that it doesn't pull the complexity of all of `Random` to the default implementation of `uuid4` (which helps greatly in #35894).

Any objections to merging and seeing how it turns out?

stdlib/UUIDs/src/UUIDs.jl
```diff
@@ -36,12 +36,21 @@ const namespace_oid  = UUID(0x6ba7b8129dad11d180b400c04fd430c8) # 6ba7b812-9dad-
...
 according to section 4.5 of the RFC.
 
+The default rng used by `uuid1` is not `GLOBAL_RNG` and every invocation of `uuid1()` without
+an argument should be expected to return a unique identifier. Importantly, the outputs of
+`uuid1` do not repeat even when `Random.seed!(seed)` is called. Currently (as of Julia 1.6),
+`uuid1` uses `Random.RandomDevice` as the default rng. However, this is an implementation
+detail that may change in the future.
+
+!!! compat "Julia 1.6"
+    The output of `uuid1` does not dependent on `GLOBAL_RNG` as of Julia 1.6.
```
rfourquet, line 52:
```suggestion
    The output of `uuid1` does not depend on `GLOBAL_RNG` as of Julia 1.6.
```

stdlib/UUIDs/src/UUIDs.jl
```diff
@@ -74,11 +83,20 @@ function uuid1(rng::AbstractRNG=Random.default_rng())
...
 as specified by RFC 4122.
 
+The default rng used by `uuid4` is not `GLOBAL_RNG` and every invocation of `uuid4()` without
+an argument should be expected to return a unique identifier. Importantly, the outputs of
+`uuid4` do not repeat even when `Random.seed!(seed)` is called. Currently (as of Julia 1.6),
+`uuid4` uses `Random.RandomDevice` as the default rng. However, this is an implementation
+detail that may change in the future.
+
+!!! compat "Julia 1.6"
+    The output of `uuid4` does not dependent on `GLOBAL_RNG` as of Julia 1.6.
```
rfourquet, line 98:
```suggestion
    The output of `uuid4` does not depend on `GLOBAL_RNG` as of Julia 1.6.
```

PR #35879, opened 2020-05-13
Make view of a string return a substring

oxinabox:
the [docs say](https://docs.julialang.org/en/v1/manual/strings/)
 
> Range indexing makes a copy of the selected part of the original string. Alternatively, it is possible to create a **view** into a string using the type SubString, …

(emph, mine)

but without this PR `@view "abcd"[2:3]` just `MethodError`s

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -44,7 +44,7 @@ Standard library changes
 * The `nextprod` function now accepts tuples and other array types for its first argument ([#35791]).
 * The function `isapprox(x,y)` now accepts the `norm` keyword argument also for numeric (i.e., non-array) arguments `x` and `y` ([#35883]).
 * `view`, `@view`, and `@views` now work on `AbstractString`s, returning a `SubString` when appropriate ([#35879]).
-* All `AbstractUnitRange{<:Integer}`s now work with `SubString`, `view`,`@view` and `@views` on strings ([#35879]).
+* All `AbstractUnitRange{<:Integer}`s now work with `SubString`, `view`, `@view` and `@views` on strings ([#35879]).
 
 #### LinearAlgebra
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -43,6 +43,7 @@ Standard library changes
 ------------------------
 * The `nextprod` function now accepts tuples and other array types for its first argument ([#35791]).
 * The function `isapprox(x,y)` now accepts the `norm` keyword argument also for numeric (i.e., non-array) arguments `x` and `y` ([#35883]).
+* `view`, `@view`, and `@views` now work on `AbstractString`s, returning a `SubString` when appropriate ([#35879]).
 
 #### LinearAlgebra
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -44,6 +44,7 @@ Standard library changes
 * The `nextprod` function now accepts tuples and other array types for its first argument ([#35791]).
 * The function `isapprox(x,y)` now accepts the `norm` keyword argument also for numeric (i.e., non-array) arguments `x` and `y` ([#35883]).
 * `view`, `@view`, and `@views` now work on `AbstractString`s, returning a `SubString` when appropriate ([#35879]).
+* All `AbstractUnitRange{<:Integer}`s now work with `SubString`, `view`,`@view` and `@views` on strings ([#35879]).
 
 #### LinearAlgebra
 
--- a/base/strings/substring.jl
+++ b/base/strings/substring.jl
@@ -47,7 +47,6 @@ end
 SubString(s::AbstractString) = SubString(s, 1, lastindex(s))
 SubString{T}(s::T) where {T<:AbstractString} = SubString{T}(s, 1, lastindex(s))
 
-
 @propagate_inbounds view(s::AbstractString, r) = SubString(s, r)
 @propagate_inbounds maybeview(s::AbstractString, r::UnitRange{<:Integer}) = view(s, r)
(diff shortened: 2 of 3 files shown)


base/strings/substring.jl
```diff
@@ -47,6 +47,11 @@ end
 SubString(s::AbstractString) = SubString(s, 1, lastindex(s))
 SubString{T}(s::T) where {T<:AbstractString} = SubString{T}(s, 1, lastindex(s))
 
+
+@propagate_inbounds view(s::AbstractString, r) = SubString(s, r)
+@propagate_inbounds maybeview(s::AbstractString, r::UnitRange{<:Integer}) = view(s, r)
+@propagate_inbounds maybeview(s::AbstractString, args...) = getindex(s, args...)
```
oxinabox, line 53: `maybeview` is required for `@views` to work well on a large block that includes somethings we can take views of and others that we can't.

Its actually really nice as a setup, i was worried about this, and until i realized, I had added extra non-SubString overloads to `view`

base/strings/substring.jl
```diff
@@ -47,6 +47,11 @@ end
 SubString(s::AbstractString) = SubString(s, 1, lastindex(s))
 SubString{T}(s::T) where {T<:AbstractString} = SubString{T}(s, 1, lastindex(s))
 
+
```
oxinabox, line 50:
```suggestion
```

base/strings/substring.jl
```diff
@@ -47,6 +47,10 @@ end
 SubString(s::AbstractString) = SubString(s, 1, lastindex(s))
 SubString{T}(s::T) where {T<:AbstractString} = SubString{T}(s, 1, lastindex(s))
 
+@propagate_inbounds view(s::AbstractString, r) = SubString(s, r)
+@propagate_inbounds maybeview(s::AbstractString, r::UnitRange{<:Integer}) = view(s, r)
```
mbauman, line 51:
```suggestion
@propagate_inbounds maybeview(s::AbstractString, r::AbstractUnitRange{<:Integer}) = view(s, r)
```

Ideally this would be like this, but looks like it also needs expanding of `SubString`'s method, too:

```Julia
julia> methods(SubString)
# 6 methods for type constructor:
[1] (::Type{SubString})(s::SubString, i::Int64, j::Int64) in Base at strings/substring.jl:43
[2] (::Type{SubString})(s::AbstractString) in Base at strings/substring.jl:47
[3] (::Type{SubString})(s::T, i::Int64, j::Int64) where T<:AbstractString in Base at strings/substring.jl:38
[4] (::Type{SubString})(s::AbstractString, i::Integer) in Base at strings/substring.jl:39
[5] (::Type{SubString})(s::AbstractString, i::Integer, j::Integer) in Base at strings/substring.jl:39
[6] (::Type{SubString})(s::AbstractString, r::UnitRange{#s69} where #s69<:Integer) in Base at strings/substring.jl:40
```

May as well do that here too, yeah?

base/strings/substring.jl
```diff
@@ -47,6 +47,10 @@ end
 SubString(s::AbstractString) = SubString(s, 1, lastindex(s))
 SubString{T}(s::T) where {T<:AbstractString} = SubString{T}(s, 1, lastindex(s))
 
+@propagate_inbounds view(s::AbstractString, r) = SubString(s, r)
```
rfourquet, line 50: With this general method, `view("asd", 2)` will return `"sd"` (or am I missing something?), whereas `getindex("asd", 2)` returns `'s'`, so it seems that this method should be split into more specialized arguments.

oxinabox: I always forget that 1 argument substring is a thing. Will have to do that

test/strings/basic.jl
```diff
@@ -166,6 +166,26 @@ end
...
+
+    # We don't (at present) make non-contiguous SubStrings with views
+    @test_throws MethodError (@view x[[1,3,5]])
+    @test (@views (x[[1,3,5]])) isa String
+
+    # We don't (at present) make single character SubStrings with views
+    @test_throws MethodError (@view x[3])
+    @test (@views (x[3])) isa Char
+
+    @test (@views (x[3], x[1:2], x[[1,4]])) isa Tuple{Char, SubString, String}
```
stevengj, line 185: To be paranoid, probably test the value as well as the type:
```jl
@test (@views (x[3], x[1:2], x[[1,4]])) == ('b', "ab", "ad")
```

oxinabox:
This should be good to go now

NEWS.md
```diff
@@ -43,6 +43,8 @@ Standard library changes
 ------------------------
 * The `nextprod` function now accepts tuples and other array types for its first argument ([#35791]).
 * The function `isapprox(x,y)` now accepts the `norm` keyword argument also for numeric (i.e., non-array) arguments `x` and `y` ([#35883]).
+* `view`, `@view`, and `@views` now work on `AbstractString`s, returning a `SubString` when appropriate ([#35879]).
+* All `AbstractUnitRange{<:Integer}`s now work with `SubString`, `view`,`@view` and `@views` on strings ([#35879]).
```
StefanKarpinski, line 47: Missing space

StefanKarpinski (approved):
Aside from the missing space, this looks good to me.

NEWS.md
oxinabox:
```suggestion
* All `AbstractUnitRange{<:Integer}`s now work with `SubString`, `view`, `@view` and `@views` on strings ([#35879]).
```

PR #35881, opened 2020-05-14
Remove contrib/mac/{juliarc.jl,mac-gtk.sh}

fingolfin:
This file is referenced by the main `Makefile`; in commit
88ad3cefb42fed3f5e7ff8aefe0fd3e1d6de5356 references to juliarc.jl
where replaced by startup.jl everywhere, including the `Makefile`;
but this file here was not renamed.

That said, I don't know exactly why this might even be needed -- the fact that nobody noticed this issue for over two years suggest it might not be... So an alternative to this PR could be to remove this file, and code in `Makefile` referencing `contrib/mac/startup.jl`.

Looking at the logs, I think @staticfloat added it, so I am sure he'll now more :-)


ViralBShah:
Needs to be removed.

ViralBShah:
In fact the contrib directory is a bit of a misnomer. It mainly contains some utilities we need for the build, and scripts to build the installers. It could probably use a better name but changing it can be a bit of a chore.

staticfloat:
Yeah, I think we can just remove this file and anything referencing it.  :)

fingolfin:
I've update this PR to instead remove the file, and the reference to it in the `Makefile`, and while at it also removed `contrib/mac/mac-gtk.sh` which doesn't seem to be referenced anywhere, and seems of dubious usefulness to me.

ViralBShah:
The Mac failure doesn't look relevant.

ViralBShah:
I've started this build again. If it fails, can you rebase it on master?

PR #35653, opened 2020-04-30
normalize_triplet.py: Properly map libgfortran versions for GCC 10

staticfloat:
Closes https://github.com/JuliaLang/julia/issues/35638

PR #35882, opened 2020-05-14
Set VERSION to 1.4.2

KristofferC:


fredrikekre:
https://github.com/JuliaLang/julia/pull/35653 should maybe be included? See https://github.com/JuliaLang/julia/pull/35849 which backports that to release-1.4.

KristofferC:
Thanks missed that. Let's backport that as well indeed.

ararslan (approved):

ararslan:
Announced a testing period for 1.4.2: https://discourse.julialang.org/t/julia-1-4-2-testing-period/39540?u=ararslan

ViralBShah:
Since this hasn't happened yet, if we still do it, we should try to include https://github.com/JuliaLang/julia/pull/35979

ararslan:
Yeah, I think at this point we could just move forward with releasing 1.4.2 as-is, since there have been no reports from the testing period (as far as I've seen). That change is worthwhile for 1.5, and perhaps 1.4.3, should we do another 1.4 release.

PR #35883, opened 2020-05-14
Add `norm` kwarg to `isapprox`

### LINKED ISSUE #13132: `norm` kwarg appears in description of `isapprox`, but not in the signature
Not sure if this is a bug or feature

Attn @stevengj, who `git blame` shows was in the vicinity

sloede:
This PR (hopefully) fixes #13132.

### DIFF
--- a/base/floatfuncs.jl
+++ b/base/floatfuncs.jl
@@ -234,10 +234,8 @@ the square root of [`eps`](@ref) of the type of `x` or `y`, whichever is bigger 
 This corresponds to requiring equality of about half of the significand digits. Otherwise,
 e.g. for integer arguments or if an `atol > 0` is supplied, `rtol` defaults to zero.
 
-If `x` and `y` are scalar values (i.e., numbers), `norm` defaults to `abs`, but
-may be changed by passing a `norm::Function` keyword argument.
-`x` and `y` may also be arrays of numbers, in which case `norm` defaults to the usual
-`norm` function in LinearAlgebra.
+The `norm` keyword defaults to `abs` for numeric `(x,y)` and to `LinearAlgebra.norm` for
+arrays (where an alternative `norm` choice is sometimes useful).
 When `x` and `y` are arrays, if `norm(x-y)` is not finite (i.e. `±Inf`
 or `NaN`), the comparison falls back to checking whether all elements of `x` and `y` are
 approximately equal component-wise.
--- a/base/floatfuncs.jl
+++ b/base/floatfuncs.jl
@@ -223,7 +223,7 @@ end
 
 # isapprox: approximate equality of numbers
 """
-    isapprox(x, y; rtol::Real=atol>0 ? 0 : √eps, atol::Real=0, nans::Bool=false, norm::Function)
+    isapprox(x, y; atol::Real=0, rtol::Real=atol>0 ? 0 : √eps, nans::Bool=false[, norm::Function])
 
 Inexact equality comparison: `true` if `norm(x-y) <= max(atol, rtol*max(norm(x), norm(y)))`. The
 default `atol` is zero and the default `rtol` depends on the types of `x` and `y`. The keyword
@@ -234,10 +234,11 @@ the square root of [`eps`](@ref) of the type of `x` or `y`, whichever is bigger 
 This corresponds to requiring equality of about half of the significand digits. Otherwise,
 e.g. for integer arguments or if an `atol > 0` is supplied, `rtol` defaults to zero.
 
+If `x` and `y` are scalar values (i.e., numbers), `norm` defaults to `abs`, but
+may be changed by passing a `norm::Function` keyword argument.
 `x` and `y` may also be arrays of numbers, in which case `norm` defaults to the usual
-`norm` function in LinearAlgebra, but
-may be changed by passing a `norm::Function` keyword argument. (For numbers, `norm` is the
-same thing as `abs`.) When `x` and `y` are arrays, if `norm(x-y)` is not finite (i.e. `±Inf`
+`norm` function in LinearAlgebra.
+When `x` and `y` are arrays, if `norm(x-y)` is not finite (i.e. `±Inf`
 or `NaN`), the comparison falls back to checking whether all elements of `x` and `y` are
 approximately equal component-wise.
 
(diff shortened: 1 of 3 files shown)


base/floatfuncs.jl
```diff
@@ -234,10 +234,11 @@ the square root of [`eps`](@ref) of the type of `x` or `y`, whichever is bigger
 This corresponds to requiring equality of about half of the significand digits. Otherwise,
 e.g. for integer arguments or if an `atol > 0` is supplied, `rtol` defaults to zero.
 
+If `x` and `y` are scalar values (i.e., numbers), `norm` defaults to `abs`, but
```
stevengj, line 237: Maybe rephrase as:
```
The `norm` keyword defaults to `abs` for numeric `(x,y)` and to `LinearAlgebra.norm` for
arrays (where an alternative `norm` choice is sometimes useful). When `x` and `y` are arrays, …
```
to emphasize the fact that alternative norms are really only useful for arrays.

stevengj:
This seems is fine to me.  I guess it is useful for generic programming, e.g. if you want to pass an L1 norm for array arguments but don't want to special-case numeric arguments.

base/floatfuncs.jl
sloede: Thank you for the suggestion. I originally wanted to minimize the changes to the existing documentation, but your version is much clearer & more useful. I have updated the PR accordingly.

stevengj:
Needs a NEWS item.

sloede:
Would something like
```
* The function `isapprox(x,y)` now accepts the `norm` keyword argument also for numeric (i.e., non-array) arguments `x` and `y` ([#35883]).
```
be appropriate (added under `Standard library changes`)?

sloede:
I added the NEWS item & did a rebase. Thank you for your feedback!

base/floatfuncs.jl
```diff
@@ -223,7 +223,7 @@ end
 
 # isapprox: approximate equality of numbers
 """
-    isapprox(x, y; rtol::Real=atol>0 ? 0 : √eps, atol::Real=0, nans::Bool=false, norm::Function)
+    isapprox(x, y; atol::Real=0, rtol::Real=atol>0 ? 0 : √eps, nans::Bool=false[, norm::Function])
```
simeonschaub, line 226:
```suggestion
    isapprox(x, y; atol::Real=0, rtol::Real=atol>0 ? 0 : √eps, nans::Bool=false, norm::Function=abs)
```

base/floatfuncs.jl
```diff
@@ -273,8 +272,10 @@ julia> isapprox(1e-10, 0, atol=1e-8)
 true
 ```
 """
-function isapprox(x::Number, y::Number; atol::Real=0, rtol::Real=rtoldefault(x,y,atol), nans::Bool=false)
-    x == y || (isfinite(x) && isfinite(y) && abs(x-y) <= max(atol, rtol*max(abs(x), abs(y)))) || (nans && isnan(x) && isnan(y))
+function isapprox(x::Number, y::Number;
+                  atol::Real=0, rtol::Real=rtoldefault(x,y,atol),
+                  nans::Bool=false, norm::Function=abs)
```
simeonschaub, line 277: I don't think the restriction to `Function` is necessary here.

base/floatfuncs.jl
stevengj: `=norm` would be more accurate, since this signature is for the general `x,y` case, not specifically for numbers.

base/floatfuncs.jl
stevengj: It's `norm::Function` in the LinearAlgebra definition, so if the signature is relaxed here it should be relaxed there as well.

stevengj: I suspect that this is one of those cases where `::Function` is fine; it's hard to imagine anyone needing some other functor argument here, and declaring it as `Function` is probably clearer?

stevengj (approved):

base/floatfuncs.jl
sloede: This concerns a weird edge case. `isapprox` is originally defined in `Base` and then extended in `LinearAlgebra` (at least as far as I can tell). The docstring is in `Base`, but also covers partly the signature of the method defined in `LinearAlgebra`. Therefore I think both `norm=abs` and `norm=norm` are misleading for the casual reader. Therefore, I decided against specifying the default value in the signature and instead just mark it as optional, which should tell the inclined reader that here is probably more to it in the textual description.

The only other sane option IMHO would be to add a *second* function signature to the docstring, i.e,
```julia
    isapprox(x::Number, y::Number; atol::Real=0, rtol::Real=atol>0 ? 0 : √eps, nans::Bool=false, norm::Function=abs)
    isapprox(x::AbstractArray, y::AbstractArray; atol::Real=0, rtol::Real=atol>0 ? 0 : √eps, nans::Bool=false, norm::Function=norm)
```

However, this has the downside that it looks *horribly complicated* for what I consider a relatively straightforward (and at the same time often-used) function. Therefore, I still favor my original proposal to mark `norm::Function` as optional and to give the default values only in the description.

base/floatfuncs.jl
sloede: If we change this in `Base`, we should also change it in `LinearAlgebra` for consistency. However, this would fundamentally change the nature of this PR: Instead of *adding new features*, it would *modify existing behavior*. While the modification is arguable not going to break any existing code, I still think it's an unnecessary complication and unnecessarily complicates this relatively straightforward PR.

I also agree with @stevengj in that restriction to `::Function` this is not such a big deal in this case. However, if there is a strong consensus that `::Function` should die, I propose that it should be dealt with in a new issue.

base/floatfuncs.jl
stevengj: norm=norm is fine in the sense that `norm(::Number)` is defined as `abs`.

base/floatfuncs.jl
simeonschaub: Ok, we can just leave it for now and this can always be a follow-up.

base/floatfuncs.jl
sloede: Functionally it is the same, I agree. However, I'd assume that it is not reasonable to use functionality from `LinearAlgebra` (`norm(...)` is defined there) in `Base`, and I also do not see the benefit of having the documentation specifically explaining this mismatch of documentation and implementation. Thus I will mark this resolved for now, as I feel like there are no strong feelings either way.

However, since you are the much more experienced Julia developer, I will of course change it if you think it is necessary.

PR #35887, opened 2020-05-14
A collection of ! invalidation fixes

timholy:
On master,
```julia
struct Not end
Base.:!(::Not) = true
```
results in more than 2600 invalidated MethodInstances. On this branch, it's less than 1500. Far from fixed, but a lot better.

### DIFF
--- a/base/show.jl
+++ b/base/show.jl
@@ -1135,6 +1135,7 @@ end
 
 function show_unquoted_quote_expr(io::IO, @nospecialize(value), indent::Int, prec::Int, quote_level::Int)
     if isa(value, Symbol) && !(value in quoted_syms)
+        value = value::Symbol
         s = string(value)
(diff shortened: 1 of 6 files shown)


base/show.jl
```diff
@@ -1135,6 +1135,7 @@ end
 
 function show_unquoted_quote_expr(io::IO, @nospecialize(value), indent::Int, prec::Int, quote_level::Int)
     if isa(value, Symbol) && !(value in quoted_syms)
+        value = value::Symbol
```
timholy, line 1138: I found this change and the one below it surprising. Seems like the compiler should figure this one out.

KristofferC: I think it only handles the case where the `isa` is the last conditional so writing it like

```
if isa(value, Symbol)
    if !(value in quoted_syms)
        ...
    end
end
```

might help it.

JeffBezanson (approved):

PR #35891, opened 2020-05-14
rearrange `assemble_inline_todo!`

JeffBezanson:
This is a refactoring that should be NFC. I changed the code to loop over union-split signatures in an outer loop, and handle the cases uniformly whether they come from union splitting or just multiple matches.

There are two main reasons for this: (1) it should be faster, since now it makes the same `_methods_by_ftype` queries that inference does (inference splits unions before doing those lookups). (2) it makes a bit more sense to me and the code feels less repetitive --- e.g. now there is only one call to `analyze_method!` instead of two. I think this will make it easier to adjust the logic for when we decide to convert dispatch to branches.

A couple numbers. Before:
```
Base  ─────────── 39.296041 seconds
Stdlibs: ────  56.155352 seconds 58.8313%
TTFP: 13.783342 seconds (15.35 M allocations: 808.146 MiB, 3.25% gc time)
```
after:
```
Base  ─────────── 38.485334 seconds
Stdlibs: ────  54.897741 seconds 58.7876%
TTFP: 13.533637 seconds (15.12 M allocations: 800.206 MiB, 3.08% gc time)
```

### DIFF
--- a/base/compiler/ssair/inlining.jl
+++ b/base/compiler/ssair/inlining.jl
@@ -1009,94 +1009,99 @@ function assemble_inline_todo!(ir::IRCode, sv::OptimizationState)
             continue
         end
 
-        # Regular case: Retrieve matching methods from cache (or compute them)
-        (meth, min_valid, max_valid) = get(sv.matching_methods_cache, sig.atype) do
-            # World age does not need to be taken into account in the cache
-            # because it is forwarded from type inference through `sv.params`
-            # in the case that the cache is nonempty, so it should be unchanged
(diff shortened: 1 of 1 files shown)


JeffBezanson:
@Keno @vtjnash Ok with this? I suddenly really want to merge it, since it allows existing union-splitting cases to remain fast even if we set max_methods to 3.

vtjnash:
Sounds reasonable to me. Makes we wonder if we should make this a shared helper, but I currently suspect not.

PR #35898, opened 2020-05-15
Make text consistent with updated macro body for metaprogramming example.

tpapp:
#34181 updated `time` to `time_ns` in the macro *code*, but did not make the same change in the explanations; this trivial PR fixes that.

### DIFF
--- a/doc/src/manual/metaprogramming.md
+++ b/doc/src/manual/metaprogramming.md
@@ -780,8 +780,8 @@ macro time(ex)
 end
 ```
 
-Here, we want `t0`, `t1`, and `val` to be private temporary variables, and we want `time` to refer
-to the [`time`](@ref) function in Julia Base, not to any `time` variable the user
+Here, we want `t0`, `t1`, and `val` to be private temporary variables, and we want `time_ns` to refer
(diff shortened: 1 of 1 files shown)


doc/src/manual/metaprogramming.md
```diff
@@ -780,8 +780,8 @@ macro time(ex)
 end
 ```
 
-Here, we want `t0`, `t1`, and `val` to be private temporary variables, and we want `time` to refer
-to the [`time`](@ref) function in Julia Base, not to any `time` variable the user
+Here, we want `t0`, `t1`, and `val` to be private temporary variables, and we want `time_ns` to refer
+to the [`time_ns`](@ref) function in Julia Base, not to any `time_ns` variable the user
 might have (the same applies to `println`). Imagine the problems that could occur if the user
 expression `ex` also contained assignments to a variable called `t0`, or defined its own `time`
```
rfourquet, line 786:
```suggestion
expression `ex` also contained assignments to a variable called `t0`, or defined its own `time_ns`
```

tpapp: Thanks! I missed that one.

rfourquet (approved):

PR #35900, opened 2020-05-15
Fix codegen for sizeof for arrays with non-power-of-2 types

### LINKED ISSUE #35884: sizeof reports wrong size for array of non-power-of-2 primitive types
This grew out of a discussion on Slack about the size taken by arrays of `BitIntegers`, but basically it seems that `summarysize` will return the bytes Julia thinks the object takes in memory by just adding up the size of the array elements without regard for their actual alignment on chip.

Setup (this is actually independent of the BitIntegers package, and is true for any type created using a primitive type):
```
using BitIntegers
BitIntegers.@define_integers 40

S40 = zeros(UInt40, 10000); S64 = zeros(UInt64, 10000) 
```

Results:
```
julia> Base.summarysize( S40 )                                                                                                                                                       
50040

julia> Base.summarysize( S64 )                                                                                                                                                       
80040
```

However, on chip the size taken by each element in the array is:
```
julia> Base.elsize( S40 )                                                                                                                                                            
8

julia> Base.elsize( S64 )                                                                                                                                                            
8
```

So `summarysize` isn't actually reporting the size on chip, it is reporting the size according to the number of bytes that Julia used to create it since it just calls `Core.sizeof` on the type: https://github.com/JuliaLang/julia/blob/1b6101af566fd0e5981082d01e4ee113bf9fa0f3/base/summarysize.jl#L115.

The main question is, what should "memory" refer to in the documentation for `summarysize`? Should there be a different exposed function to actually get the on-chip memory usage?

imciner2:
This fixes the code generation optimization for the `sizeof()` function to use the array element's alignment if the element is a primitive type instead of the size of the type. This is a fix for #35884. I also found a bug in the `sizeof` overload in the `SubArrays` type, so this is fixed here as well.

I have added in tests for the Array `sizeof()` call, but there are probably more that could be added (so different types are tested) - so any suggestions are appreciated.

Fixes #35884.

### DIFF
--- a/test/abstractarray.jl
+++ b/test/abstractarray.jl
@@ -937,3 +937,39 @@ end
 @testset "vcat with mixed elements" begin
     @test vcat(Nothing[], [missing], [1.0], [Int8(1)]) isa Vector{Union{Missing, Nothing, Float64}}
 end
+
+@testset "sizeof" begin
+    let arrUInt8 = zeros(UInt8, 10)
+        @test sizeof(arrUInt8) == 10
+        @test Core.sizeof(arrUInt8) == 10
+    end
+
+    let arrUInt32 = zeros(UInt32, 10)
+        @test sizeof(arrUInt32) == 40
+        @test Core.sizeof(arrUInt32) == 40
+    end
+
+    let arrFloat64 = zeros(Float64, 10, 10)
+        @test sizeof(arrFloat64) == 800
+        @test Core.sizeof(arrFloat64) == 800
+    end
+
+    # Test union arrays (Issue #23321)
+    let arrUnion = Union{Int64, Cvoid}[rand(Bool) ? k : nothing for k = 1:10]
+        @test sizeof(arrUnion) == 80
+        @test Core.sizeof(arrUnion) == 80
+    end
+
+    # Test non-power of 2 types (Issue #35884)
+    primitive type UInt48 48 end
+    UInt48(x::UInt64) = Core.Intrinsics.trunc_int(UInt48, x)
+    UInt48(x::UInt32) = Core.Intrinsics.zext_int(UInt48, x)
+
+    a = UInt48(0x00000001);
+    b = UInt48(0x00000002);
+    c = UInt48(0x00000003);
+    let arrayOfUInt48 = [a, b, c]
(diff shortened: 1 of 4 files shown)


rfourquet:
In some circumstances `sizeof` is used to write or read the relevant number of bytes for a type, e.g. `read` at https://github.com/JuliaLang/julia/blob/43d8a285d3f0a0a432692cbdb78f1be2ea871f81/base/iobuffer.jl#L189 In that use-case, the old `sizeof` definition was "correct", so if `sizeof` is modified, it would be great to have an alternative function playing the role of the current `sizeof`.

I also don't know if the current definition of `sizeof`

> Size, in bytes, of the canonical binary representation of the given DataType T

implies that the updated behavior here is the correct one, but it would be great to update the docstring to explain this subtlelty.

imciner2:
I should have been a bit clearer, this only changes the `sizeof` results for arrays containing these non-power-of-2 types. In this case, the data is actually aligned to be on a power of  anyway. This shouldn't affect the results for just primitive types (those will still return the correct bytes).

rfourquet:
Ah thanks for the clarification!

test/abstractarray.jl
```diff
@@ -937,3 +937,39 @@ end
...
+    # Test non-power of 2 types (Issue #35884)
+    primitive type UInt48 48 end
+    UInt48(x::UInt64) = Core.Intrinsics.trunc_int(UInt48, x)
+    UInt48(x::UInt32) = Core.Intrinsics.zext_int(UInt48, x)
+
+    a = UInt48(0x00000001);
+    b = UInt48(0x00000002);
+    c = UInt48(0x00000003);
+    let arrayOfUInt48 = [a, b, c]
+        @test sizeof(arrayOfUInt48) == 24
```
JeffBezanson, line 972: I would add
```
f35900(x) = sizeof(x)
@test f35900(arrayOfUInt48) == 24
```
to be extra sure the compiled case is hit.

JeffBezanson (approved):

test/abstractarray.jl
imciner2: Thanks (although I changed it to the issue number).

PR #35903, opened 2020-05-15
More thoroughly document apropos

non-Jedi:
I've frequently found myself wishing for a regex-based apropos but
didn't realize it already existed since it's not mentioned in the
docstring.

### DIFF
--- a/stdlib/REPL/src/docview.jl
+++ b/stdlib/REPL/src/docview.jl
@@ -685,20 +685,16 @@ stripmd(x::Markdown.Footnote) = "$(stripmd(x.id)) $(stripmd(x.text))"
 stripmd(x::Markdown.Table) =
     join([join(map(stripmd, r), " ") for r in x.rows], " ")
 
-# Apropos searches through all available documentation for some string or regex
 """
-    apropos([io::IO], string)
+    apropos([io::IO=stdout], pattern::Union{AbstractString,Regex})
 
-Search through all documentation for a string, ignoring case.
+Search available docstrings for entries containing `pattern`.
+
+When `pattern` is a string, case is ignored. Results are printed to `io`.
 """
 apropos(string) = apropos(stdout, string)
 apropos(io::IO, string) = apropos(io, Regex("\\Q$string", "i"))
 
-"""
-    apropos([io::IO], regex)
-
-Search through all documentation for `regex`.
-"""
 function apropos(io::IO, needle::Regex)
     for mod in modules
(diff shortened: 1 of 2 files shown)


non-Jedi:
`apropos` also seems like it should be part of `InteractiveUtils` rather than `REPL` (that's where it's currently documented in the manual), but I suppose that would be a breaking change.

mbauman:
I also only discovered a week or two ago that using a string literal or regex literal at the help prompt does `apropos`!  It's not technically apropos' behavior, but I'm not sure where else it might live.

```Julia
help?> "aprop"
REPL.stripmd
Base.Docs.apropos

help?> r"ap..p"
Base.:∘
Base.shell_escape_posixly
Distributed.CachingPool
REPL.stripmd
Base.Docs.apropos
```

non-Jedi:
Oh neat! That should definitely be mentioned in the [help mode section of the manual](https://docs.julialang.org/en/v1/stdlib/REPL/#Help-mode-1), but I share your uncertainty over whether it belongs in this docstring.

non-Jedi:
It seems unlikely that 7598cbb broke tests (and only on macos), but when I click on "details" in the github UI, there doesn't appear to be anything there for that CI run?

non-Jedi:
I would appreciate if someone could review this pure documentation PR and merge if appropriate.

stdlib/REPL/src/docview.jl
```diff
@@ -687,12 +687,18 @@ stripmd(x::Markdown.Table) =
 
 # Apropos searches through all available documentation for some string or regex
 """
-    apropos(string)
+    apropos([io::IO], string)
 
 Search through all documentation for a string, ignoring case.
```
mbauman, line 692:
```suggestion
Search through the documentation for entries that contain the string or regular expression, ignoring case.

Print the matching entries to `io`, defaulting to `stdout`.
```

I don't really think this requires a separately documented method — especially when `contains` doesn't do so!  It's also nice to talk about `io` if we're going to mention it.  I've also changed "all documentation" to simply "the documentation."  In my mind, searching _all_ documentation is what juliahub does.

non-Jedi: Thanks for the review. I didn't quite follow your suggestion exactly, but I have consolidated into a single docstring, mentioned the `io` argument, and clarified what documentation is searched. Let me know if you think the updated version is clear.

mbauman (approved):
Looks great!

PR #30043, opened 2018-11-14
Allow setting default selections in TerminalMenus

christopher-dG:
This doesn't change any behaviour without opting into it with the `default` keyword, but allows you to set the initial cursor position (for `RadioMenu`) or checked options (`MultiSelectMenu`) when `request` is called.

Example (the menus are what appear without any input):
```julia
julia> m = RadioMenu(["a", "b", "c", "d", "e"]; default=3);

julia> request(m)
   a
   b
 > c
   d
   e
3

julia> m = MultiSelectMenu(["a", "b", "c", "d", "e"]; default=[2, 4]);

julia> request(m)
[press: d=done, a=all, n=none]
 > [ ] a
   [X] b
   [ ] c
   [X] d
   [ ] e
Set([4, 2])
```

PR #27489, opened 2018-06-08
Remove TerminalMenus

KristofferC:
Arguably this code is not up to the Base standards and got put in (without much review) because it was thought Pkg would extensively use it.
This is not really the case and @vtjnash has expressed concerns that TerminalMenus is blocking him in some work.
We are losing some existing interactive features in Pkg with this, and there was also a few more I thought about adding (see e.g. https://github.com/JuliaLang/julia/pull/26469) but to unblock Jameson this seems to be the quickest way.

PR #35915, opened 2020-05-17
Improve the API of REPL.TerminalMenus

### LINKED ISSUE #31414: julep: Allow passing an initial cursor to `request()` AbstractMenu.jl
I find myself presenting the same menus repeatedly via REPL.TerminalMenus, and it would make sense for the default option to be whatever the user last chose.

It looks like it would be super easy to hack `AbstractMenu.jl` such that the cursor can optionally be initialized with something other than 1.

I should be able to hack together a PR for this over the weekend. API-wise, I'm thinking to add something like an `initial_cursor::Int=1` keyword arg to the `request()` method that actually implements all the logic, and just forward the argument along between all the other methods. Thoughts?

### LINKED ISSUE #32349: Adding defaults to TerminalMenus
This is a feature request, and it's fixed by https://github.com/JuliaLang/julia/pull/32344

It would be good to allow for specifying default options when using TerminalMenus. For example, when using `MultiSelectMenu`: 
```julia
julia> menu = MultiSelectMenu(string.(1:10), default = 1:3);

julia> request(menu)
[press: d=done, a=all, n=none]
 > [X] 1
   [X] 2
   [X] 3
   [ ] 4
   [ ] 5
   [ ] 6
   [ ] 7
   [ ] 8
   [ ] 9
   [ ] 10
```

timholy:
This eliminates camelCase, ~fixes a typo (which unfortunately was documented)~, adds a `const` and type info to the CONFIG dict, and does some renaming/consolidation. Finally, it eliminates a dangerous escape hatch from one of the tests.

EDIT: the most recent version fixes #31414, fixes #32349, closes #30043, and closes #32344


stdlib/REPL/src/TerminalMenus/util.jl
```diff
@@ -12,54 +12,43 @@
     PAGE_UP,
     PAGE_DOWN)
 
-# Enable raw mode. Allows us to process keyboard inputs directly.
-function enableRawMode(term)
+# Set `term` mode; raw=true intercepts keyboard inputs directly
+function raw!(term, raw::Bool)
```
rfourquet, line 16: Isn't it possible to use directly `REPL.Terminals.raw!` here instead of wrapping it? It even seems they return the same `Bool`.

stdlib/REPL/test/TerminalMenus/runtests.jl
```diff
@@ -4,12 +4,9 @@ import REPL
 using REPL.TerminalMenus
 using Test
 
-TerminalMenus.config(supress_output=true)
-
-function simulateInput(expectedResult, menu::TerminalMenus.AbstractMenu, keys...)
-    # If we cannot write to the buffer, skip the test
-    !(:buffer in fieldnames(typeof(stdin))) && return true
```
rfourquet, line 11: Could you say a word about this line and why this was dangerous? I don't see in which case it would have been triggered...

rfourquet (approved):

stdlib/REPL/test/TerminalMenus/runtests.jl
timholy: It's not guaranteed if someone uses `redirect_stdin`. Looking at just the line itself, you have to dig ("what type of object is `stdin`? Where was this called from and did anyone redirect it?") just to know whether it was actually tested or not.

To me, this seems too misleading. I could change this to emit a warning, if you think that's better.

rfourquet: Ok, no I agree it's better to remove it and if an error happens in a certain testing context, it seems best to handle the error at call-site (e.g. not calling this test, but explicitly).

stdlib/REPL/src/TerminalMenus/TerminalMenus.jl
```diff
@@ -23,4 +23,9 @@ export
     MultiSelectMenu,
     request
 
+# TODO: remove Julia 2.0
+# While not exported, AbstractMenu documented these as an extension interface
+const printMenu = printmenu
+const writeLine = writeline
```
ararslan, line 29:
```suggestion
@deprecate printMenu printmenu
@deprecate writeLine writeline
```
It seems okay to me to introduce an actual deprecation for these since they aren't exported and TerminalMenus usage is fairly limit AFAIK.

timholy:
OK, the more I dug the more I found to fix. I tackled this because I was looking to create a new menu type, which gave me an opportunity to exercise the abstraction API. Perhaps not surprisingly I discovered some issues. Two issues are:
- it turns out `options(m)` was never used other than to calculate the number of items in the menu. It seems more reasonable, then, to support an interface that allows you to just pass this number rather than a whole vector of strings. (For the menu type I am implementing, it's not a static list so not having to generate a bunch of strings that will just be thrown away seems like a win.)
- there's a bad leak in the API, subtypes of AbstractMenu are expected to access TerminalMenus.CONFIG to get the characters for the cursor position and, for a multiple-selection menu, the strings that indicate the selection state. It's quite weird to have the configuration be at the level of TerminalMenus but to have the printing be at the level of the subtype and then have the subtype need to access private data of TerminalMenus.

I've tried to clean this up here, but it's still not the API we probably want for 2.0. If the subtypes are going to be responsible for printing the cursor marker and selection state strings, it seems these are options you should pass when you construct the menu rather than a configuration option for TerminalMenus as a whole. But I don't see a non-breaking way to get there. So this does the next best thing (?) and passes a copy of these settings back to `writeline`.

I think I've set this up so that all users of the legacy interface should continue to work, but that's not actually tested.

timholy:
I think this has reached a satisfying point. Changes since the last update:

- I've added an intermediate abstract type, `ConfiguredMenu`, that resolves the API "leak" I describe above. Subtypes store the configuration in the menu itself, rather than in a global variable.
- The printing of the "cursor"-selected indicator is handled automatically, just like it already was for the "up arrow" and "down arrow" characters. This ends the odd asymmetry among these indicators.
- I've made the printing algorithm more robust for dynamic menus that might change their number of options. It now passes a state variable that allows it to know how many lines it needs to back up to erase the previous menu.
- I've added a more thorough test of navigation and menu-printing (the new `dynamic_menu.jl` test file)
- I've added "legacy" tests that use the original Julia 1.0 `AbstractMenu` API, and verified that they still work.

With one important exception, this `AbstractMenu` API now preserves backward compatibility while developing a cleaner interface for the future. The exception, unfortunately, is in the exported `RadioMenu` and `MultiSelectMenu` types, which I've migrated over to the new API. That means that instead of configuring them via `TerminalMenus.config`, you configure them at construction time. If we want to preserve compatibility, we should export the `OldRadioMenu` and `OldMultiSelectMenu` that I added as test examples, and figure out some new name for the new implementations.

Thoughts?

timholy:
OK, now I think I've found a smooth path between where we are now and a nicer API. The key trick here is embedding the configuration into the menu struct, *even* if that is just a reference to the backward-compatible global variable `CONFIG`. This preserves the current mechanism for configuration while moving everybody in a saner direction.

I'm proposing this as *the* cleanup work for Julia 2.0 in REPL.TerminalMenus. Consequently it needs an actual API review. In light of #27489 I'm putting this in Stefan's pile.

tomyun:
Nice to see this moving forward! I was just wondering if we could use this chance for incorporating upstream changes made after split (merge?) circa early 2018. A notable change I can think of is single choice menu (nick-paul/TerminalMenus.jl#20) which was a sole reason prevented me from using the REPL version of this package.

timholy:
That can be viewed as a bugfix, so it would be fine to put that in any time. That said, you might as well wait for this to merge?

KristofferC (approved):
Looks like a nice improvement to the API to me.

timholy:
Added NEWS, will merge if CI passes.

yakir12:
(about the news, if this indeed allows for adding default option/s, shouldn't that be mentioned in the news?)

timholy:
Maybe. I'm not actually sure what you mean by "adding default options," can you clarify?

Note docs for TerminalMenus are at https://docs.julialang.org/en/v1.6-dev/stdlib/REPL/#TerminalMenus-1. Amazingly, they've already been updated for this PR.

yakir12:
I'm only asking because this #32349 closed by this PR. As stated in that issue, default options would allow for the prompt to have the cursor already set to some default option for the `RadioMenu`, and for some default options in `MultiSelectMenu` to be pre-checked. This is useful when you are near certain about what the user will choose and want to save the extra scrolling and clicking by allowing the user to simply press `done`.

timholy:
Oh, got it. Yes, I suppose the individual API additions are NEWS-worthy. See if #36210 helps (and feel free to edit it).

yakir12:
Perfect! Thank you very much.

GunnarFarneback:
I'm late to discover the wonders of TerminalMenus and I was delighted to find this PR when faced with the limitations of the API in Julia 1.5. However, I still see one glaring omission, namely that the `cursor` position is a local variable in the `request` loop. For reading purposes you can catch it from the `writeline` calls but there is no way to control it after the initialization.

Things that would be possible with a controllable cursor include jumping to a related option after a custom key press or seamlessly moving into a submenu related to the option at the cursor.

Ideally the cursor should have been a mandatory field in the menu struct. That's not possible to introduce now with respect to backwards compatibility of course. One possibility is to use a `cursor` field from the struct if it exists and fall back to a local variable otherwise, but I don't immediately see how to implement that in an elegant way. A simpler but less complete approach would be to allow `keypress` to return an `Int` if you want to change the cursor from that call (which is probably where it's most interesting).

@timholy, any thoughts?

timholy:
I agree that it would have been nice to allow `cursor` to be part of the menu. Indeed, the hardest part of this cleanup was doing it in a way that maintained backwards compatibility while still moving us towards a cleaner implementation.

In addition to supporting an arbitrary `Int` as a return value for `keypress`, what about supporting `Union{Int,RefValue{Int}}` for `cursor`? Then anyone who wants to observe the value from the outside can do so.

GunnarFarneback:
Ah, a `Ref` in the call to `request`. Yes, that's a clever approach. I'll do something with that.

PR #35916, opened 2020-05-17
Added tests for Iterators to cover a few edge cases.

ravibitsgoa:
Improves test coverage for Iterators.
Covers a few edge cases.


test/iterators.jl
```diff
@@ -100,6 +100,7 @@ end
 @test length(zip(cycle(1:3), 1:7, cycle(1:3))) == 7
 @test length(zip(1:3,product(1:7,cycle(1:3)))) == 3
 @test length(zip(1:3,product(1:7,cycle(1:3)),8)) == 1
+@test_throws ArgumentError length(zip()) # length of zip of empty tuple
```
jonas-schulze, line 103: Maybe, `length` should be extended to return 0 in this case.

tkf: I think `length` of empty `zip` should be infinite. This PR also tests that `Base.IteratorSize(zip()) == Base.IsInfinite()` holds.  Also, `for x in zip(); end` is an infinite loop (which, I think, is correct).

test/iterators.jl
```diff
@@ -216,6 +224,8 @@ let i = 0
         i += 1
         i <= 10 || break
     end
+    @test Base.isdone(cycle(0:3)) === Base.isdone(0:3)
```
tkf, line 227: Why not

```suggestion
    @test Base.isdone(cycle(0:3)) === Base.isdone(0:3) === missing
```

?

test/iterators.jl
```diff
@@ -841,3 +868,19 @@ end
...
+
+@testset "IteratorSize trait for zip" begin
+    @test Base.IteratorSize(zip()) == Base.IsInfinite()                     # for zip of empty tuple
+    @test Base.IteratorSize(zip((1,2,3), repeated(0))) == Base.HasLength()  # for zip of ::HasLength and ::IsInfinite
+    @test Base.IteratorSize(zip( 1:5, repeated(0) )) == Base.HasLength()    # for zip of ::HasShape and ::IsInfinite
+    @test Base.IteratorSize(zip(repeated(0), (1,2,3))) == Base.HasLength()  # for zip of ::IsInfinite and ::HasLength
+    @test Base.IteratorSize(zip(repeated(0), 1:5 )) == Base.HasLength()     # for zip of ::IsInfinite and ::HasShape
+    @test Base.IteratorSize(zip((1,2,3), 1:5) ) == Base.HasLength()         # for zip of ::HasLength and ::HasShape
+    @test Base.IteratorSize(zip(1:5, (1,2,3)) ) == Base.HasLength()         # for zip of ::HasShape and ::HasLength
+end
```
tkf, line 886: It looks like other files have newlines at the end of the file.

```suggestion
end

```

(though not sure if doing this in browser adds extra line)

test/iterators.jl
```diff
@@ -841,3 +868,19 @@ end
     @test cumprod(x + 1 for x in 1:3) == [2, 6, 24]
     @test accumulate(+, (x^2 for x in 1:3); init=100) == [101, 105, 114]
 end
+
+@testset "Iterators.tail_if_any" begin
+    @test Iterators.tail_if_any(()) == ()
+    @test Iterators.tail_if_any((1, 2)) == (2, )
+    @test Iterators.tail_if_any((1, )) == ()
```
tkf, line 875: `grep`ing the Julia files here, it looks like it's much more common to avoid the space after `,`

```suggestion
    @test Iterators.tail_if_any((1, 2)) == (2,)
    @test Iterators.tail_if_any((1,)) == ()
```

tkf (commented):
LGTM other than the minor nitpicks I mentioned.

test/iterators.jl
jonas-schulze: Why should `zip()` be infinite?

If I collect/group/zip elements from no collections, I don’t expect there to be any elements, i.e. `isempty(collect(zip())) == isempty(zip()) == true`. Just like I would expect `zip(xs...) = map(Tuple, xs...)` and that should have the length of the shortest iterable among `xs...`, which I think should represent the number of elements in a collection. If there are no collections, there no elements, hence its length should be zero.

tkf: Since `length(zip(itrs...)) == minimum(length, itrs) == mapreduce(length, min, itrs)`, the value `length(zip())` should be the identity element of `min`.  The identity element of `min` on bounded set is the _maximum_ of the set (it has to "loose" against everything else when taking `min`).  Since the output of `length` can take any non-negative integer, "infinite" is a reasonable answer.

Another way to think about is that, if you think the mapping `(a, b) -> zip(a, b)` as a monoid and define `zip(itrs...) = reduce(zip, itrs)` [*1], the iterator `zip()` should be the identity of the monoid.  An infinite stream of empty tuples is the identity element of the `zip` monoid.  The length of an infinite stream is, well, infinite.

[*1] You have to "post-process" to flatten the output tuple. But other than that, this monoid argument is accurate.

tkf: Yet another way to convince yourself is to consider a function

```julia
function zip′(itrs...)
    a, b = anysplit(itrs)
    return ((x..., y...) for (x, y) in zip(zip(a...), zip(b...)))
end
```

with a function `anysplit` that splits a tuple at _some_ location such that

```julia
ys, zs = anysplit(xs...)
@assert (ys..., zs...) === xs
```

It is desirable for the function `zip′` to behave like `zip` for any `anysplit` function.

In particular, for this to hold for `anysplit(xs...) = (), xs`, we need `zip()` to produce an infinite stream of empty tuples.

jonas-schulze: Thanks for the explanation! The monoid view was helpful. I agree that the stream of empty tuples would be the neutral element, if we consider equivalence under "generates the same elements". Though it's not intuitive to me why `zip()` would be `repeated(())`.

I know that the following is currently wrong in Julia, but as we already consider the empty tuple to be identical to "no element", I would expect this:

```
using Test

@testset "empty collections" begin
    null = Vector{Union{}}()
    @test collect(zip(())) == null
    @test collect(zip()) == null
    @test collect(()) == null
    @test collect() == null
end
```

jonas-schulze: On the other hand, having a `zip` name for the `zip`-neutral element kinda makes sense.

tkf: I think it's as counter-intuitive as the neutral element of `min` being the maximum.  Or maybe more like the intersection of empty set.  I agree it feels like a paradox but "fixing it" would break mathematical coherence behind the API.

ravibitsgoa:
Hi everyone,
Can someone review this PR and merge?

PR #35925, opened 2020-05-18
faster filesize and read(::IOStream)

JeffBezanson:
- Skip the `stat` call for reading entire files when the file is small (<32kb, the default buffer size)
- Remove a redundant extra eof check in `readbytes_all!`
- Speed up `filesize` by using lseek instead of stat. I don't know whether this will be faster on all systems. On our end, it removes an allocation, since the stat code is quite old and uses an Array instead of a Ref to get the data from C. We might be able to remove that allocation sometimes, so I can try that too.

### DIFF
--- a/base/iostream.jl
+++ b/base/iostream.jl
@@ -219,6 +219,12 @@ function position(s::IOStream)
     return pos
 end
 
+function filesize(s::IOStream)
+    sz = @_lock_ios s ccall(:ios_filesize, Int64, (Ptr{Cvoid},), s.ios)
+    systemerror("filesize", sz == -1)
+    return sz
+end
+
 _eof_nolock(s::IOStream) = ccall(:ios_eof_blocking, Cint, (Ptr{Cvoid},), s.ios) != 0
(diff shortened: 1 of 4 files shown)


base/iostream.jl
```diff
@@ -486,21 +493,33 @@ function readbytes!(s::IOStream, b::Array{UInt8}, nb=length(b); all::Bool=true)
...
-            sz -= pos
+    # First we try to fill the buffer. If that gives us the whole file,
+    # copy it out and return. Otherwise look at the file size and use it
+    # to prealloate space. Determining the size requires extra syscalls,
+    # which we want to avoid for small files.
+    @_lock_ios s begin
+        nb = ccall(:ios_fillbuf, Cssize_t, (Ptr{Cvoid},), s.ios)
+        if nb != -1
+            b = StringVector(nb)
+            readbytes_all!(s, b, nb)
```
staticfloat, line 504: We need to handle the case where this is a directory; we don't want to return a successful zero-byte read of a directory.

PR #35926, opened 2020-05-18
fix broadcasted assignment for scalar cartesian indexing

stev47:
When broadcasting left-hand side indexing is turned into a dotview.
For scalar indexing this should be equivalent to a `getindex`, but up
until this patch the `CartesianIndex` would not be treated as such.

This fixes e.g. broadcasting like `a[CartesianIndex(1)] .= 1` for a nested
array `a`.

### DIFF
--- a/base/views.jl
+++ b/base/views.jl
@@ -131,7 +131,7 @@ end
 # (while remaining equivalent to getindex for scalar indices and non-array types)
 @propagate_inbounds maybeview(A, args...) = getindex(A, args...)
 @propagate_inbounds maybeview(A::AbstractArray, args...) = view(A, args...)
-@propagate_inbounds maybeview(A::AbstractArray, args::Union{Number,CartesianIndex}...) = getindex(A, args...)
+@propagate_inbounds maybeview(A::AbstractArray, args::Union{Number,AbstractCartesianIndex}...) = getindex(A, args...)
 @propagate_inbounds maybeview(A) = getindex(A)
 @propagate_inbounds maybeview(A::AbstractArray) = getindex(A)
 
--- a/base/views.jl
+++ b/base/views.jl
@@ -131,7 +131,7 @@ end
 # (while remaining equivalent to getindex for scalar indices and non-array types)
 @propagate_inbounds maybeview(A, args...) = getindex(A, args...)
 @propagate_inbounds maybeview(A::AbstractArray, args...) = view(A, args...)
-@propagate_inbounds maybeview(A::AbstractArray, args::Number...) = getindex(A, args...)
+@propagate_inbounds maybeview(A::AbstractArray, args::Union{Number,CartesianIndex}...) = getindex(A, args...)
 @propagate_inbounds maybeview(A) = getindex(A)
 @propagate_inbounds maybeview(A::AbstractArray) = getindex(A)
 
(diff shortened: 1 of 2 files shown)


base/views.jl
```diff
@@ -132,6 +132,7 @@ end
 @propagate_inbounds maybeview(A, args...) = getindex(A, args...)
 @propagate_inbounds maybeview(A::AbstractArray, args...) = view(A, args...)
 @propagate_inbounds maybeview(A::AbstractArray, args::Number...) = getindex(A, args...)
```
mbauman, line 134:
```suggestion
@propagate_inbounds maybeview(A::AbstractArray, args::Union{Number, CartesianIndex}...) = getindex(A, args...)
```

test/broadcast.jl
```diff
@@ -313,6 +313,8 @@ end
 let a = [[4, 5], [6, 7]]
     a[1] .= 3
     @test a == [[3, 3], [6, 7]]
+    a[CartesianIndex(1)] .= 4
+    @test a == [[4, 4], [6, 7]]
 end
```
mbauman, line 318: Let's also test a mix of CartesianIndex and integer for 2+ dimensional arrays.

mbauman (changes requested):

mbauman:
Yikes, thanks for the patch.  It'd be nice if you could edit your first post to include a summary of what changes.  This does change non-erroring behaviors, but I think it does so in a way that is definitively a bug fix. On master:

```Julia
julia> a = Any[[4, 5], [6, 7]]
2-element Array{Any,1}:
 [4, 5]
 [6, 7]

julia> a[1] .= 8
2-element Array{Int64,1}:
 8
 8

julia> a
2-element Array{Any,1}:
 [8, 8]
 [6, 7]

julia> a[CartesianIndex(1)] .= 9
0-dimensional view(::Array{Any,1}, 1) with eltype Any:
9

julia> a
2-element Array{Any,1}:
 9
  [6, 7]
```

This PR makes `a[1]` and `a[CartesianIndex(1)]` behave the same — as they should.

mbauman (approved):

base/views.jl
stev47: thanks, I didn't even know mixing indices was allowed in julia

stev47:
Thanks, updated the description

stev47:
can we merge this? maybe even backport?

base/views.jl
```diff
@@ -131,7 +131,7 @@ end
 # (while remaining equivalent to getindex for scalar indices and non-array types)
 @propagate_inbounds maybeview(A, args...) = getindex(A, args...)
 @propagate_inbounds maybeview(A::AbstractArray, args...) = view(A, args...)
-@propagate_inbounds maybeview(A::AbstractArray, args::Number...) = getindex(A, args...)
+@propagate_inbounds maybeview(A::AbstractArray, args::Union{Number,CartesianIndex}...) = getindex(A, args...)
```
mbauman, line 134: Ah, we're hitting a bootstrap issue here.  This is a quick shot in the dark to fix it — and a good improvement in any case.

```suggestion
@propagate_inbounds maybeview(A::AbstractArray, args::Union{Number,AbstractCartesianIndex}...) = getindex(A, args...)
```

Issue #35922, opened 2020-05-18
Summary of non-ambiguous patterns of invalidation


timholy:
I've looked at a enough cases, and developed enough tooling, to start getting a better sense for patterns. Here are some tests that load a few packages (just to get away from looking only at Base+stdlibs) and then load a new package (or execute a new command) while capturing invalidations.

| Package or command | Invalidated sig | # children | Explanation |
|:-------------------------- |:---------------- | -----------:|:------------------|
| FixedPointNumbers | `reduce_empty(::Function, ::Type{T} where T)` | 200 | Compiler chooses not to specialize [this method](https://github.com/JuliaLang/julia/blob/cfd7f483307b406c5b07eea5bb1889677f66e8de/base/reduce.jl#L338). Many signatures don't actually intersect with the new method, but lack of specialization makes that hard to detect. A solution: add an `init` argument to `maximum` so compiler knows what to do if container is empty. |
| New `!` method | `!=(::Any, ::Any)` & similar | 383 | Union-splitting for noninferred types. Called by, e.g., `convert(::Type{T}, x::AbstractDict) where T<:AbstractDict`, `cmp(::AbstractString, ::String)` ultimately from #15276, handling `Expr` args, ... |
| StaticArrays | `unsafe_convert(::Type{Ptr{Nothing}}, b::Base.RefValue{T})` | 290 + tons in other packages | Overwhelmingly `show_default` (primarily `show(::IO, ::RawFD)` and `show_function`) |
| | `cconvert(::Type{Ptr{Nothing}}, ::Any)` | 88 | `REPL.raw!` not knowing the handle is `Ptr{Cint}` (true?) |
| | `to_indices(to_indices(A, I::Tuple)` | 80 | inference failures in concatenation |
| | `all`, `any` specializations | collectively hundreds | typically #15276 |
| OffsetArrays | `axes(::AbstractArray)` | 240 | JuliaInterpreter access to `CodeInfo.linetable` and `CodeInfo.codelocs` (both `Any`) |
| ColorTypes | `promote_rule(::Type{<:Any}, ::Type{<:Any}) = Bottom` | 35 | `+(::Int, ::Integer)` with no backedges |
| ImageCore | `convert(::Type{Array{C,N}}, ::Array{C,N} where {C<:Colorant,N}` | 80 | type-piracy, this should be defined in ColorTypes to head off any other packages implementing it |
[...]

PR #35928, opened 2020-05-18
More invalidation fixes

timholy:
These fix a fair number of the Base invalidations in #35922. If all these seem fine it could be merged, but many do arise from partial inference, and if we implement some of the suggestions in that issue then the changes to the following may not be so urgent:

- ~`convert(::Type{T}, x::AbstractDict) where T<:AbstractDict`~
- `cmp(a::AbstractString, b::AbstractString)`

I doubt either of these would cause trouble, but I'm happy to drop those if Jeff or others think they are fixable by other means. To me, the rest seem like stuff we probably want.

### DIFF
--- a/base/Enums.jl
+++ b/base/Enums.jl
@@ -159,6 +159,7 @@ macro enum(T, syms...)
         else
             throw(ArgumentError(string("invalid argument for Enum ", typename, ": ", s)))
         end
+        s = s::Symbol
         if !Base.isidentifier(s)
             throw(ArgumentError("invalid name for Enum $typename; \"$s\" is not a valid identifier"))
         end
--- a/base/Enums.jl
+++ b/base/Enums.jl
@@ -29,7 +29,7 @@ Base.print(io::IO, x::Enum) = print(io, Symbol(x))
 
 function Base.show(io::IO, x::Enum)
     sym = Symbol(x)
-    if !get(io, :compact, false)
+    if !(get(io, :compact, false)::Bool)
         from = get(io, :module, Main)
         def = typeof(x).name.module
         if from === nothing || !Base.isvisible(sym, def, from)
--- a/base/arrayshow.jl
+++ b/base/arrayshow.jl
@@ -166,7 +166,7 @@ function print_matrix(io::IO, X::AbstractVecOrMat,
                       vdots::AbstractString = "\u22ee",
                       ddots::AbstractString = "  \u22f1  ",
                       hmod::Integer = 5, vmod::Integer = 5)
-    if !get(io, :limit, false)
+    if !(get(io, :limit, false)::Bool)
         screenheight = screenwidth = typemax(Int)
     else
         sz = displaysize(io)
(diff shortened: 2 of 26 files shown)


stdlib/REPL/src/docview.jl
```diff
@@ -609,7 +609,7 @@ filtervalid(names) = filter(x->!occursin(r"#", x), map(string, names))
 accessible(mod::Module) =
     [filter!(s -> !Base.isdeprecated(mod, s), names(mod, all = true, imported = true));
      map(names, moduleusings(mod))...;
-     collect(keys(Base.Docs.keywords))] |> unique |> filtervalid
+     collect(keys(Base.Docs.keywords))]::Vector{Symbol} |> unique |> filtervalid
```
vtjnash, line 612:
```suggestion
    Symbol[filter!(s -> !Base.isdeprecated(mod, s), names(mod, all = true, imported = true));
      map(names, moduleusings(mod))...;
      collect(keys(Base.Docs.keywords))] |> unique |> filtervalid
```

base/abstractdict.jl
```diff
@@ -516,7 +516,7 @@ convert(::Type{T}, x::T) where {T<:AbstractDict} = x
 
 function convert(::Type{T}, x::AbstractDict) where T<:AbstractDict
     h = T(x)
-    if length(h) != length(x)
+    if length(h)::Int != length(x)::Int
```
JeffBezanson, line 519: I'm sure this is true in practice but it doesn't seem strictly correct to assert it. We don't know what `x` is.

base/arrayshow.jl
```diff
@@ -371,7 +371,7 @@ preceded by `prefix`, supposed to encode the type of the elements.
 """
 function _show_nonempty(io::IO, X::AbstractMatrix, prefix::String)
     @assert !isempty(X)
-    limit = get(io, :limit, false)::Bool
+    limit::Bool = get(io, :limit, false)
```
JeffBezanson, line 374: I would be surprised if this is really better --- this way it will have to do `convert(Bool, ::Any)` instead of just asserting the type.

timholy: Here I was just being consistent with the surrounding code; someone went through previously and put assertions on all the `:limit` cases, and there were all of the `limit::Bool = ` style. The difference triggered just enough OCD that I changed it.

timholy:
If folks want to keep these commits separate for easier reversion in case of trouble I can rebase, or feel free to squash.

base/arrayshow.jl
JeffBezanson: Aha, then all of those should probably be changed the other way! :)

base/Enums.jl
```diff
@@ -29,7 +29,7 @@ Base.print(io::IO, x::Enum) = print(io, Symbol(x))
 
 function Base.show(io::IO, x::Enum)
     sym = Symbol(x)
-    if !get(io, :compact, false)
+    if !(get(io, :compact, false)::Bool)
```
JeffBezanson, line 32: Fortunately `::` already binds tighter than `!`, but I can see how that might not be obvious to code readers.

timholy:
Seeing if I can trigger tests.

base/Enums.jl
timholy: i can change it, but I felt that obviousness was worth a few extra characters.

timholy:
Barring objections, I'll merge this sometime early in the work week.

PR #35929, opened 2020-05-18
Remove unnecessary restriction to `StridedVecOrMat`

dlfivefifty:
The "Strided array interface" https://docs.julialang.org/en/v1/manual/interfaces/#man-interface-strided-arrays-1 means that this is useful beyond just `StridedVecOrMat` so there is no reason to be so restrictive.


dlfivefifty:
Note ArrayLayouts.jl also has an approach for pointers to adjoints of complex arrays by defining `ConjPtr` so that, for example, `pointer(view(A',1:3,2:4)')` works. This could be moved to Base and thereby removing type piracy for ArrayLayouts.jl:
```julia
"""
    ConjPtr{T}
represents that the entry is the complex-conjugate of the pointed to entry.
"""
struct ConjPtr{T} 
    ptr::Ptr{T}
end

unsafe_convert(::Type{ConjPtr{T}}, Ac::Adjoint{<:Complex}) where T<:Complex = unsafe_convert(Ptr{T}, parent(Ac))    
function unsafe_convert(::Type{ConjPtr{T}}, V::SubArray{T,2}) where {T,N,P}
    kr, jr = parentindices(V)
    unsafe_convert(Ptr{T}, view(parent(V)', jr, kr))
end
```

stdlib/LinearAlgebra/src/adjtrans.jl
```diff
@@ -205,8 +205,8 @@ Base.strides(A::Transpose{<:Any, <:StridedVector}) = (stride(A.parent, 2), strid
 Base.strides(A::Adjoint{<:Real, <:StridedMatrix}) = reverse(strides(A.parent))
 Base.strides(A::Transpose{<:Any, <:StridedMatrix}) = reverse(strides(A.parent))
```
mbauman, line 206: Here too.

dlfivefifty: Fixed

dlfivefifty:
Can this get merged so I can start testing all the packages I manage on Julia v1.5?

KristofferC:
It seems a test was requested.

stdlib/LinearAlgebra/test/blas.jl
```diff
@@ -460,6 +460,14 @@ Base.unsafe_convert(::Type{Ptr{T}}, A::WrappedArray{T}) where T = Base.unsafe_co
 
 Base.stride(A::WrappedArray, i::Int) = stride(A.A, i)
 
+@testset "strided interface adjtrans" begin
+    x = WrappedArray([1, 2, 3, 4])
+    @test strides(x') == strides(transpose(x)) == (1,1)
+    A = WrappedArray([1 2; 3 4; 5 6])
+    @test strides(A') == strides(transpose(A)) == (3,1)
+    @test pointer(A') == pointer(transpose(A)) == pointer(A)
```
timholy, line 468: Worth a `@test_throws` for something like `strides(transpose(1:5))`?

mbauman:
Good question... but is there a case in which it matters?  You're not allowed to multiply that first value by anything other than zero, are you?

Is there a way in which `(1, 1)` is any more correct or helpful than `(4, 1)`?

dlfivefifty:
Only because `stride(::AbstractVector, 2)` is not defined at the moment. I changed the definition to work around this but if you really want `(4,1)` we can think of an alternative.

mbauman:
This is my main motivation:

```Julia
julia> strides(randn(4, 1)')
(4, 1)

julia> strides(randn(4)')
(4, 1)
```

dlfivefifty:
OK that's a strong argument. So I generalised `stride(::AbstractArray, k::Integer)` to be defined for all `k`, as in `Array`.

I also added `ConjPtr` to deal with pointers of complex adjoints. This could be removed if you aren't happy with it (though would mean ArrayLayouts.jl would still have some type piracy).

@mbauman Could you review the changes again please?

stdlib/LinearAlgebra/src/adjtrans.jl
```diff
@@ -199,15 +199,31 @@ convert(::Type{Adjoint{T,S}}, A::Adjoint) where {T,S} = Adjoint{T,S}(convert(S,
...
 
 Base.unsafe_convert(::Type{Ptr{T}}, A::Adjoint{<:Real, <:AbstractVecOrMat}) where {T} = Base.unsafe_convert(Ptr{T}, A.parent)
 Base.unsafe_convert(::Type{Ptr{T}}, A::Transpose{<:Any, <:AbstractVecOrMat}) where {T} = Base.unsafe_convert(Ptr{T}, A.parent)
 
+"""
+    ConjPtr{T}
+
+A memory address referring to the complex conjugate of data of type T. However, there is no guarantee that the memory is actually valid, or that it actually represents data of the specified type.
+"""
+struct ConjPtr{T}
```
KristofferC, line 216: What issue does this fix? Since it has a docstring it looks like a new feature which we shouldn't put into 1.5.

dlfivefifty: The issue is in the tests:
```julia
@test pointer(view(B',:,:)') == pointer(B)
```
I'll remove it though since it sounds like it will hold things up.

dlfivefifty:
I believe this is ready to merge now unless you want other changes made.

base/abstractarray.jl
```diff
@@ -392,7 +392,11 @@ julia> stride(A,3)
 12
 ```
 """
-stride(A::AbstractArray, k::Integer) = strides(A)[k]
+function stride(A::AbstractArray, k::Integer) 
+    k < 1 && return 1
+    k ≤ ndims(A) && return strides(A)[k] 
+    length(A)
```
timholy, line 398: So now `stride(1:5, 0)` and `stride(1:5, 2)` will return values but `stride(1:5, 1)` will throw an error? To me it seems better to throw a `MethodError` in all cases.

dlfivefifty: I've modified it so it always calls `strides` now, producing the errors consistently now.

timholy (approved):

KristofferC:
Whitespace needs fixing. If this is not merged soon, I'll just push for a 1.5-RC1 since the only thing this "fixes" is a package that was doing type piracy anyway and could just be worked around in the package.

base/abstractarray.jl
```diff
@@ -392,7 +392,12 @@ julia> stride(A,3)
 12
 ```
 """
-stride(A::AbstractArray, k::Integer) = strides(A)[k]
+function stride(A::AbstractArray, k::Integer)
+    st = strides(A)
+    k < 1 && return 1
```
mbauman, line 397: This feels funny to me; what is the meaning of `stride(A, 0)` or `stride(A, -1)`?  I see that this happens to be the behavior that falls out of the current `stride(::DenseArray, n)` implementation, but I'd rather keep the abstract behaviors more conservative and closer to the status quo  — especially to be considered for a back port.

base/abstractarray.jl
```diff
@@ -392,7 +392,12 @@ julia> stride(A,3)
...
+    k ≤ ndims(A) && return st[k]
+    length(A)
```
mbauman, line 399: This doesn't seem to be the right answer, but I'm not sure what the right answer is.  Is it `sum(st .* size(A))`?

I know, it has a bit of that "does it matter what number is going to be multiplied by zero" rhetorical question quality, but this seems to be a better answer.

base/abstractarray.jl
dlfivefifty: I agree, though then we shouldn't call `strides(A::AbstractVector, 2)` in the definition of `strides(A')`.

I could revert back to 
```julia
Base.strides(A::Adjoint{<:Real, <:AbstractVector}) = (1, stride(A.parent, 1))
```
despite it being inconsistent with your `strides(randn(4, 1)')` example (it's multipled by 0 anyways)

base/abstractarray.jl
mbauman: Er, no, that's not quite it either.  Is it the absolute maximum of `st .* size(A)` (but preserving its sign)?  Even that leaves a bit to be desired as an abstract behavior... for example, here's how SubArray currently works:

```Julia
julia> v = @view rand(100)[1:3:end];

julia> stride(v, 1), stride(v, 2) # This is the answer the abstract behavior I proposed would give
(3, 102)

julia> m = @view rand(100, 1)[1:3:end, :];

julia> stride(m, 1), stride(m, 2) # This seems like a "better" answer:
(3, 100)
```

base/abstractarray.jl
mbauman: There are two cases here, `n < 1` and `n > ndims(A)`.  This is only talking about the `n < 1` case — I see the motivation for the `n > ndims(A)` case.

dlfivefifty: Ah ok (I don't see the motivation for either). So perhaps just delete `n < 1` case and let it throw an error?

mbauman: Yes, that'd be my preference. :+1:

base/abstractarray.jl
mbauman: I feel like an overestimate here is somehow more defensible than an underestimate as it pushes you _out of bounds_.  Here are a few more tricky n > ndims cases:

```Julia
julia> m  = @view rand(12, 24, 60)[1:3:end, end:-4:1, 2:-1:1];

julia> strides(m)
(3, -48, -288)

julia> stride(m, 4) # This seems reasonable; increasing this would jump out-of-bounds of the view
-576

julia> m2  = @view rand(12, 24, 60, 1)[1:3:end, end:-4:1, 2:-1:1, :];

julia> strides(m2) # this would _definitely_ jump out of bounds (that's the length of the parent, and we're starting in the middle!)
(3, -48, -288, 17280)

julia> p = PermutedDimsArray(m, (3,2,1));

julia> strides(p)
(-288, -48, 3)

julia> stride(p, 4)
ERROR: BoundsError: attempt to access (-288, -48, 3)
  at index [4]

julia> vpv = @view p[1:2, 1:2, 1:2]

julia> strides(vpv)
(-288, -48, 3)

julia> stride(vpv, 4) # I think this one is wrong; SubArray is implicitly assuming column-major parents
6
```

That last SubArray behavior can and should be tackled separately, but this gives me a bit of intuition here.

dlfivefifty: Not sure I follow...

mbauman: We're adding a behavior here — I just want to make sure that it makes sense with respect to our existing behaviors and that there's a decent rationale for it.

dlfivefifty: Yes I get that, I don't follow what you are proposing. (It seems to me that it doesn't really matter what the stride is outside the range)

base/abstractarray.jl
JeffBezanson: Bump

dlfivefifty:
The test failure looks unrelated

dlfivefifty:
Does "backport-1.5" mean it's not going to be in v1.5.0? If so I'll start working around the issue in ArrayLayouts.jl for v1.5-beta

KristofferC:
> Does "backport-1.5" mean it's not going to be in v1.5.0?

No, if it gets merged before 1.5.0 is released, then it will be in.

base/abstractarray.jl
mbauman: This was by-and-large just me talking out loud.  It's not a major thing since — as you say — it shouldn't matter.  Even if we use a value that's more likely to push someone out of bounds, we don't even check pointer accesses with `--check-bounds=yes`.  It _might_ be more likely that it'd help someone identify the problem, but that's quite the shot in the dark.

Upon reflection here I still do have a slight preference for `return sum(st .* size(A))` here simply because using `length` alone is a code smell in strides computations — it just looks wrong.  Indeed, it makes it clearer that we need to call `strides` in this branch (like Tim noted above).  But as a style thing, I'll leave the final decision to you.

dlfivefifty: I'm happy with your suggestion .

mbauman:
```suggestion
    return sum(st .* size(A))
```

mbauman (approved):

stdlib/LinearAlgebra/test/blas.jl
```diff
@@ -458,7 +458,44 @@ Base.setindex!(A::WrappedArray, v, i::Int) = setindex!(A.A, v, i)
...
+    @test stride(x,1) == 1
+    @test stride(x,2) == stride(x,3) == 4
+    @test strides(x') == strides(transpose(x)) == (4,1)
+    @test pointer(x') == pointer(transpose(x)) == pointer(x)
+    @test_throws BoundsError stride(x,0)
+
+    A = WrappedArray([1 2; 3 4; 5 6])
+    @test stride(A,1) == 1
+    @test stride(A,2) == 3
+    @test stride(A,3) == stride(A,4) == 6
```
mbauman, line 474: Whoops, this needs to change too.

```suggestion
    @test stride(A,3) == stride(A,4) >= 6
```

PR #35944, opened 2020-05-19
faster repeat(::AbstractArray ...)

jw3126:
```julia
using BenchmarkTools
arr = randn(100,100)
@btime repeat(arr, inner=(2,3), outer=(1,2))
```
```
  1.775 ms (50020 allocations: 2.90 MiB) # master 
  286.722 μs (5 allocations: 1.37 MiB) # this PR
```

### DIFF
--- a/base/abstractarraymath.jl
+++ b/base/abstractarraymath.jl
@@ -323,31 +323,13 @@ to_tuple(t::Tuple) = t
 to_tuple(x::Integer) = (x,)
 to_tuple(itr) = tuple(itr...)
 
-filltuple(val, N) = ntuple(_->val, N)
-
-function pad(a::Tuple, b::Tuple, c::Tuple)
-    a1,b1 = pad(a,b)
-    b2,c2 = pad(b1, c)
-    a2,c3  = pad(a1, c2)
-    return a2, b2, c2
+function pad(a, b)
+    N = max(length(a), length(b))
+    Base.fill_to_length(a, 1, Val(N)), Base.fill_to_length(b, 1, Val(N))
 end
-function pad(a::Tuple{}, b::NTuple{N, Any}) where {N}
-    filltuple(1, Val(N)), b
-end
-function pad(a::NTuple{N, Any}, b::Tuple{}) where {N}
-    (a, filltuple(1, Val(N)))
-end
-function pad(a::NTuple{N,Any}, b::NTuple{N,Any}) where {N}
-    (a,b)
-end
-function pad(a::Tuple{}, b::Tuple{})
-    (a,b)
-end
-function pad(a::Tuple, b::Tuple)
-    a1 = first(a)
-    b1 = first(b)
-    a_tail, b_tail = pad(Base.tail(a), Base.tail(b))
-    tuple(a1, a_tail...), tuple(b1, b_tail...)
+function pad(a, b, c)
+    N = max(max(length(a), length(b)), length(c))
+    Base.fill_to_length(a, 1, Val(N)), Base.fill_to_length(b, 1, Val(N)), Base.fill_to_length(c, 1, Val(N))
 end
 
 function resolve(arr::AbstractArray{<:Any, N}, inner::NTuple{N, Any}, outer::NTuple{N,Any}) where {N}
--- a/base/abstractarraymath.jl
+++ b/base/abstractarraymath.jl
@@ -353,35 +353,22 @@ end
 
 function check(arr, inner, outer)
     if inner !== nothing
+        # TODO: Currently one based indexing is demanded for inner !== nothing,
+        # but not for outer !== nothing. Decide for something consistent.
+        Base.require_one_based_indexing(arr)
         if any(<(0), inner)
-            msg = """
-            Number of repetitions cannot be negative. Got inner = $inner
-            """
-            throw(ArgumentError(msg))
+            throw(ArgumentError("no inner repetition count may be negative; got $inner"))
         end
         if length(inner) < ndims(arr)
-            msg = """
-            Number of dimensions of repetitions and array must match. Got
-            inner = $inner
-            size(arr) = $(size(arr))
-            """
-            throw(ArgumentError(msg))
+            throw(ArgumentError("number of inner repetitions ($(length(inner))) cannot be less than number of dimensions of input array ($(ndims(arr)))"))
         end
     end
-    if outer != nothing
+    if outer !== nothing
         if any(<(0), outer)
-            msg = """
-            Number of repetitions cannot be negative. Got outer = $outer
-            """
-            throw(ArgumentError(msg))
+            throw(ArgumentError("no outer repetition count may be negative; got $outer"))
         end
-        if (length(outer) < ndims(arr)) && (inner != nothing)
-            msg = """
-            Number of dimensions of repetitions and array must match. Got
-            outer = $outer
-            size(arr) = $(size(arr))
-            """
-            throw(ArgumentError(msg))
+        if (length(outer) < ndims(arr)) && (inner !== nothing)
+            throw(ArgumentError("number of outer repetitions ($(length(outer))) cannot be less than number of dimensions of input array ($(ndims(arr)))"))
         end
     end
 end
--- a/base/abstractarraymath.jl
+++ b/base/abstractarraymath.jl
@@ -352,7 +352,7 @@ function resolve(arr, inner, outer::Nothing)
 end
 
 function check(arr, inner, outer)
-    if inner != nothing
+    if inner !== nothing
         if any(<(0), inner)
             msg = """
             Error calling repeat. all(inner . >= 0) must hold. Got
--- a/base/abstractarraymath.jl
+++ b/base/abstractarraymath.jl
@@ -355,14 +355,13 @@ function check(arr, inner, outer)
     if inner !== nothing
         if any(<(0), inner)
             msg = """
-            Error calling repeat. all(inner . >= 0) must hold. Got
-            inner = $inner
+            Number of repetitions cannot be negative. Got inner = $inner
             """
             throw(ArgumentError(msg))
         end
         if length(inner) < ndims(arr)
             msg = """
-            Error calling repeat. length(inner) >= ndims(arr) must hold. Got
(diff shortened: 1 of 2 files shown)


jw3126:
bump

jw3126:
Can this be reviewed?

base/abstractarraymath.jl
```diff
@@ -330,89 +308,171 @@ julia> repeat([1 2; 3 4], inner=(2, 1), outer=(1, 3))
...
+end
+function pad(a::Tuple{}, b::Tuple{})
+    (a,b)
+end
+function pad(a::Tuple, b::Tuple)
+    a1 = first(a)
+    b1 = first(b)
+    a_tail, b_tail = pad(Base.tail(a), Base.tail(b))
+    tuple(a1, a_tail...), tuple(b1, b_tail...)
+end
```
mbauman, line 351: You can use `Base.fill_to_length` to make this much simpler:

```suggestion
function pad(a, b)
    N = max(length(a), length(b))
    Base.fill_to_length(a, 1, Val(N)), Base.fill_to_length(b, 1, Val(N))
end
function pad(a, b, c)
    N = max(max(length(a), length(b)), length(c))
    Base.fill_to_length(a, 1, Val(N)), Base.fill_to_length(b, 1, Val(N)), Base.fill_to_length(c, 1, Val(N))
end
```

base/abstractarraymath.jl
```diff
@@ -330,89 +308,171 @@ julia> repeat([1 2; 3 4], inner=(2, 1), outer=(1, 3))
...
+    dims, outer = pad(size(arr), to_tuple(outer))
+    reshape(arr, dims), inner, outer
+end
+function resolve(arr, inner, outer::Nothing)
+    dims, inner = pad(size(arr), to_tuple(inner))
+    reshape(arr, dims), inner, outer
+end
+
+function check(arr, inner, outer)
+    if inner != nothing
```
mbauman, line 373: It's much faster and easier on the compiler to use `===` when comparing to `nothing`.

```suggestion
    if inner !== nothing
```

base/abstractarraymath.jl
```diff
@@ -330,89 +308,171 @@ julia> repeat([1 2; 3 4], inner=(2, 1), outer=(1, 3))
...
+        if any(<(0), inner)
+            msg = """
+            Error calling repeat. all(inner . >= 0) must hold. Got
+            inner = $inner
+            """
```
mbauman, line 378: This (and below) don't quite match our typical error styles.  Something more like this would fit in a bit better:

```
throw(ArgumentError("number of inner repetitions cannot be negative; got $inner"))
```

base/abstractarraymath.jl
```diff
@@ -330,89 +308,171 @@ julia> repeat([1 2; 3 4], inner=(2, 1), outer=(1, 3))
...
+
+function repeat_outer(a::AbstractMatrix, (m,n)::NTuple{2, Any})
+    o, p = size(a,1), size(a,2)
+    b = similar(a, o*m, p*n)
+    for j=1:n
+        d = (j-1)*p+1
+        R = d:d+p-1
+        for i=1:m
+            c = (i-1)*o+1
+            @inbounds b[c:c+o-1, R] = a
```
mbauman, line 422: Just for my own reference (since GitHub isn't aligning these diffs helpfully), this is nearly exactly the status quo, but with this extra `@inbounds`.  Looks good.

base/abstractarraymath.jl
```diff
@@ -330,89 +308,171 @@ julia> repeat([1 2; 3 4], inner=(2, 1), outer=(1, 3))
...
 
-    # fill the outer blocks along each dimension
-    if all(isequal(1), outer)
-        return R
+function repeat_outer(a::AbstractVector, (m,)::Tuple{Any})
+    o = length(a)
+    b = similar(a, o*m)
+    for i=1:m
+        c = (i-1)*o+1
+        @inbounds b[c:c+o-1] = a
```
mbauman, line 433: Similarly here. :+1:

base/abstractarraymath.jl
```diff
@@ -330,89 +308,171 @@ julia> repeat([1 2; 3 4], inner=(2, 1), outer=(1, 3))
...
-            dest_indices[i] = dest_indices[i] .+ inner_shape[i]
-            R[dest_indices...] = B
+    return b
+end
+
+function repeat_outer(arr::AbstractArray{<:Any,N}, dims::NTuple{N,Any}) where {N}
+    repeat_outer_generic(arr, dims)
+end
+
+function repeat_outer_generic(arr, dims)
```
mbauman, line 442: Is there a reason for the separate function here?

```suggestion
```

mbauman:
We actually have a number of BaseBenchmarks that use `repeat`:

base/abstractarraymath.jl
jw3126: Nice!

base/abstractarraymath.jl
jw3126: It was useful for checking that the 1d and 2d special implementations are needed for performance, but is no longer necessary.

base/abstractarraymath.jl
jw3126: You are referring to the error message text, not to splitting the code over multiple lines, right?

nanosoldier:
[Your benchmark job](https://github.com/JuliaLang/julia/pull/35944#issuecomment-638494246) has completed - possible performance regressions were detected. A full report can be found [here](https://github.com/JuliaCI/BaseBenchmarkReports/blob/147b9d123794ccfdd92d17253ebaeecca779ee4d/02cfcd1_vs_cf410dc/report.md). cc @ararslan

base/abstractarraymath.jl
```diff
@@ -330,89 +308,147 @@ julia> repeat([1 2; 3 4], inner=(2, 1), outer=(1, 3))
...
+    repeat_inner_outer(arr, inner, outer)
+end
+
+to_tuple(t::Tuple) = t
+to_tuple(x::Integer) = (x,)
+to_tuple(itr) = tuple(itr...)
+
+function pad(a, b)
+    N = max(length(a), length(b))
+    Base.fill_to_length(a, 1, Val(N)), Base.fill_to_length(b, 1, Val(N))
```
ararslan, line 328: No need for `Base.` qualifications inside of Base.

base/abstractarraymath.jl
```diff
@@ -330,89 +308,147 @@ julia> repeat([1 2; 3 4], inner=(2, 1), outer=(1, 3))
...
-
-    R = similar(A, shape)
-    if any(iszero, shape)
-        return R
+function check(arr, inner, outer)
+    if inner !== nothing
+        if any(<(0), inner)
+            msg = """
+            Number of repetitions cannot be negative. Got inner = $inner
+            """
```
ararslan, line 359: Note that this will add a trailing newline to the error messages, which most errors do not have.

base/abstractarraymath.jl
mbauman: This is using a somewhat clever technique of a `module _InternalImplementation` which I don't think we've done before, but it cleverly helps readability of the internals while preventing clashes with other parts of base and leaking private functions.  I like it.

Thus it does need the qualifications.

base/abstractarraymath.jl
```diff
@@ -330,89 +308,147 @@ julia> repeat([1 2; 3 4], inner=(2, 1), outer=(1, 3))
...
+end
+
+function repeat_outer(arr::AbstractArray{<:Any,N}, dims::NTuple{N,Any}) where {N}
+    insize  = size(arr)
+    outsize = map(*, insize, dims)
+    out = similar(arr, outsize)
+    for I in CartesianIndices(arr)
+        for J in CartesianIndices(dims)
+            TIJ = map(Tuple(I), Tuple(J), insize) do i, j, d
+                i + d * (j-1)
```
mbauman, line 425: Ahh, but outer repeats did previously work (but are unfortunately untested):

```Julia
# master
julia> repeat(OffsetArray(rand(3,3),-1,-1), outer=(2,1))
6×3 Matrix{Float64}:
 0.625501  0.113374  0.643544
 0.870337  0.994211  0.464454
 0.518634  0.999477  0.040367
 0.625501  0.113374  0.643544
 0.870337  0.994211  0.464454
 0.518634  0.999477  0.040367

# this PR, currently:
julia> repeat_outer(OffsetArray(rand(3,3),-1,-1), (2,1))
Segmentation fault: 11
```

base/abstractarraymath.jl
ararslan: Huh, okay.

base/abstractarraymath.jl
jw3126: A segfault is clearly bad and I need to fix it. The question is, what is the most reasonable semantics for an OffsetArray? I think returning an OffsetArray does not make sense, since it is unclear what should happen with the axes. That leaves throwing an error or silently passing to one based indexing. I am in favor of the former, but that might be considered a breaking change.

mbauman: Funny, something must have been wrong with my REPL state when I tested that (I was just doing simplistic copy/paste debugging at the REPL for a quick review).  Now that I try this again, it looks like your implementation matches the status quo — outer repeats work (returning a 1-based array) and inner repeats go out of bounds.

We could simply `Base.require_one_based_indexing` for the inner repeat case and preserve the outer's status quo, leaving it for another day... but I agree the returned axes are questionable.

base/abstractarraymath.jl
mbauman: I think I'd prefer something like:

```Julia
ArgumentError("no inner repetition count may be negative; got $inner")
ArgumentError("number of inner repetitions ($(length(inner))) cannot be less than number of dimensions of input array ($(ndims(arr)))")
ArgumentError("no outer repetition count may be negative; got $outer")
ArgumentError("number of outer repetitions ($(length(outer))) cannot be less than number of dimensions of input array ($(ndims(arr)))")
```

jw3126:
I think CI fail is unrelated to PR.

mbauman:
Thanks @jw3126!

PR #35948, opened 2020-05-20
fix AbstractArrayStyle dimension promotion rule

marius311:
The simple test I added, 

```julia
struct CustomStyle{T} <: Broadcast.AbstractArrayStyle{1} end
Broadcast.BroadcastStyle(::CustomStyle{:A}, ::CustomStyle{:B}) = CustomStyle{:A}()
@test Broadcast.result_style(CustomStyle{:A}(), CustomStyle{:B}()) ==
      Broadcast.result_style(CustomStyle{:B}(), CustomStyle{:A}()) == CustomStyle{:A}() 
```

fails on current master because the function which I deleted in this PR is doing a bit of type piracy, especially since its using `Base.typename` which means its assuming if the `typename`s of your custom style are the same, they differ only in dimensions, for which I don't think there's justification. This then breaks `BroadcastStyle` rule resolution for your custom style if this is not the case. 

In terms of Base, the offending part of the deleted function is [not covered](https://coveralls.io/builds/30897392/source?filename=base/broadcast.jl#L142) in the tests, and once you delete that, its equivalent to just deleting the whole thing.  

I checked by hand if this impacts StaticArrays.jl as a representative user of this API, and it only breaks [one test](https://github.com/JuliaArrays/StaticArrays.jl/blob/e3e06a93565cdf8e1b138d88184a568ac9cf0365/test/mapreduce.jl#L188), 

```julia
v1 = @SVector [2,4,6,8]
M = @SMatrix [1 2; 3 4; 5 6; 7 8]
broadcast(+, v1, M)
```

which now returns a `SizedArray` as opposed to previously an `SArray`. I think that can be easily fixed in StaticArrays, and it probably shouldn't have been relying on this hacky `typename` thing anyway. If you guys give the go ahead here, I can do the StaticArrays fix too (as well as any other comments you may have).


marius311:
Hmm reading over the [broadcasting](https://docs.julialang.org/en/v1/manual/interfaces/#writing-binary-broadcasting-rules-1) API again maybe you can't just delete this function, its what provides the API described in the last paragraph there that lets you just define a `Val(N)` constructor. 

Gonna mark this as draft and think about if there's a way to just do something smarter than `typename` which allows the above to test to pass while preserving the existing API. Suggestions welcome ofcourse.

marius311:
I think I've got the right solution now, which fixes the test from my OP while simultaneously not breaking any Base or StaticArrays tests (any other packages I should check?), plus I think the logic makes a decent bit of sense. 

Basically, instead of using `Base.typename(A) === Base.typename(B)` to try and guess whether to do the dimension promotion, only do it if promoting both `A` and `B` give the same result, otherwise require the user to provide a rule. 

From what I can tell inference is fine and the entire logic is optimized away at compile time, so I don't think there's runtime performance concerns. 

Ready for review from my end.

base/broadcast.jl
```diff
@@ -137,11 +137,12 @@ BroadcastStyle(a::AbstractArrayStyle, ::Style{Tuple})    = a
 BroadcastStyle(::A, ::A) where A<:ArrayStyle             = A()
 BroadcastStyle(::ArrayStyle, ::ArrayStyle)               = Unknown()
 BroadcastStyle(::A, ::A) where A<:AbstractArrayStyle     = A()
-Base.@pure function BroadcastStyle(a::A, b::B) where {A<:AbstractArrayStyle{M},B<:AbstractArrayStyle{N}} where {M,N}
-    if Base.typename(A) === Base.typename(B)
-        return A(Val(max(M, N)))
+function BroadcastStyle(::A, ::B) where {M,N,A<:AbstractArrayStyle{M},B<:AbstractArrayStyle{N}}
+    if M!=N && M!=Any && N!=Any && A(Val(max(M,N)))==B(Val(max(M,N)))
```
tkf, line 141: I think using `!==` is more compiler-friendly

```suggestion
    if M !== Any && N !== Any && M != N && A(Val(max(M,N))) == B(Val(max(M,N)))
```

(also Base code usually have spaces around the operators)

Alternatively, I guess defining

```julia
BroadcastStyle(::AbstractArrayStyle{Any}, ::AbstractArrayStyle{Any}) = Unknown()
BroadcastStyle(::AbstractArrayStyle, ::AbstractArrayStyle{Any}) = Unknown()
BroadcastStyle(::AbstractArrayStyle{Any}, ::AbstractArrayStyle) = Unknown()
```

would be the safest in terms of inferability. I think using `if` would be OK, though.

marius311: Good point, thanks. I would keep tje `if` given that the code is clearer and if inference ever breaks here that test would catch it.

tkf (approved):

mbauman:
Ah, this looks great!  It's a good solution and huge bonus points for removing the certainly-erroneous `@pure`.  It'd be good to just see what PkgEval thinks...

nanosoldier:
[Your package evaluation job](https://github.com/JuliaLang/julia/pull/35948#issuecomment-636968481) has completed - possible new issues were detected. A full report can be found [here](https://github.com/JuliaCI/NanosoldierReports/blob/a1070bb18bb2537b1e74c07d604daf0c04d5f040/pkgeval/by_hash/8e95bd0_vs_cfb9b55/report.md). cc @maleadt

marius311:
Let me know if there's anything else that needs doing (I'm not really familiar with how to interpret the output of PkgEval).

marius311:
Another friendly ping?

marius311:
Hi guys, just rebased this on master. Can this please be merged? @tkf @mbauman 

Not sure how many more times I have it in me to keep asking about this before I just give up and this gets lost.

adienes:
I think this PR is actually still basically up to date.

marius311:
Just did my bi-decadal rebase of this PR onto master, lmk if anything else is needed 😅 

(but seriously, I appreciate this is a hard one since it touches sooo much, happy to help in whatever way I can) 

I also checked StaticArrays, StructArrays, and ForwardDiff as some representative broadcast overloaders and didn't spot any new failures related to this PR (altho hard bc of other failures tbh)

adienes:
@nanosoldier `runtests()`

nanosoldier:
The package evaluation job [you requested](https://github.com/JuliaLang/julia/pull/35948#issuecomment-2876366269) has completed - possible new issues were detected.
The [**full report**](https://s3.amazonaws.com/julialang-reports/nanosoldier/pkgeval/by_hash/c0a1728_vs_2a5cfee/report.html) is available.

<details><summary>Report summary</summary>

## ❗ Packages that crashed

**1 packages crashed only on the current version.**

 - A segmentation fault happened: 1 packages

2 packages crashed on the previous version too.

## ✖ Packages that failed

**25 packages failed only on the current version.**

 - Package has test failures: 3 packages
 - Package tests unexpectedly errored: 8 packages
 - Test duration exceeded the time limit: 14 packages

1387 packages failed on the previous version too.

## ✔ Packages that passed tests

**17 packages passed tests only on the current version.**

 - Other: 17 packages

5236 packages passed tests on the previous version too.

## ~ Packages that at least loaded

**11 packages successfully loaded only on the current version.**

 - Other: 11 packages

2956 packages successfully loaded on the previous version too.

## ➖ Packages that were skipped altogether

**2 packages were skipped only on the current version.**

 - Package could not be installed: 2 packages

920 packages were skipped on the previous version too.



</details>

adienes:
pretty clean, I think the `StrideArrays.jl` issue is real, but I bet all the other failures are just downstream of that

```
MethodError: no method matching StrideArrays.CartesianStyle{Tuple{Int64}, 1}(::Val{1})
```

marius311:
I'd bet they're somehow relying on the old broken API. I can take a look.

adienes:
awesome!

I think my initial summary was wrong, and it looks like these are also some sources, so they're not all only from `StridedArrays` unfortunately.  I've created the relevant issues for:

[Tensorial.jl#250](https://github.com/KeitaNakamura/Tensorial.jl/issues/250)
[ClimaCore.jl#2329](https://github.com/CliMA/ClimaCore.jl/issues/2329)
[Modulo2.jl#8](https://github.com/matthias314/Modulo2.jl/issues/8)
[StandardPacked.jl#2](https://github.com/projekter/StandardPacked.jl/issues/2)
[RecursiveArrayTools.jl#449](https://github.com/SciML/RecursiveArrayTools.jl/issues/449)

 now these are of course not all necessarily your responsibility to fix, it would probably be polite to notify the authors before "breaking" code, so it would be reasonable give it a few days to see if any of the package authors bite on the issue

matthias314:
I'm the author of one of the packages mentioned above ([Modulo2.jl](https://github.com/matthias314/Modulo2.jl)). What about updating the Julia documentation as part of this PR? I believe the current documentation doesn't mention that style constructors have to accept argument, see the example in the [docstring](https://docs.julialang.org/en/v1/base/arrays/#Base.Broadcast.BroadcastStyle) for `BroadcastStyle`. And if they have to accept a `Val` argument, what are they supposed to do? The patch for StridedArrays.jl assumes that the `Val` argument is equal to the dimension of the style and skips it,

https://github.com/JuliaSIMD/StrideArrays.jl/blob/a2dd2e6c3a938ab79b2b1c9b495c6ee640aacc5f/src/broadcast.jl#L10-L11

In contrast, the code in `Base` allows to set and change the dimension,
https://github.com/JuliaLang/julia/blob/18d5ccd6c30eff16aca186dc65a42c39714020b0/base/broadcast.jl#L98-L99

A clarification would be most helpful.

marius311:
The Val argument stuff is described in the AbstractArrayStyle docstring, those are the only kind of styles that assume this functionally is present, with some further clarification here: https://docs.julialang.org/en/v1/manual/interfaces/#writing-binary-broadcasting-rules

I haven't looked at your package yet so not sure if that answers everything. 

In terms of the PR, I only did the trivial rule where the dimensionally is the same since that was all that was needed to fix the tests and I wasn't familiar enough with the package so just decided to be safe.

marius311:
I suppose this PR does add a requirement that the trivial rule is defined where previously there was no requirement. I wonder if all those packages can simply be fixed by that? I can look. It may make sense to just define that one generically.

matthias314:
@marius311 Thanks for the pointer to the documentation! I'm going to fix my package.

adienes:
given

```
julia> struct MyStyle{N} <: Broadcast.AbstractArrayStyle{N} end
```

I think this PR now requires a constructor on the parameterized type
```
MyStyle{N}(::Val{N}) where {N} = MyStyle{N}()
```

where some packages had been previously defining (note the lack of `{N}`)
```
MyStyle(::Val{N}) where {N} = MyStyle{N}()
```

This is at least what looks to be the case for `RecursiveArrayTools.jl`. is that an intentional change?

marius311:
Neither master nor this PR will ever call: 
```julia
MyStyle(::Val{N}) where {N} = MyStyle{N}()
```
They would only call this one: 
```julia
MyStyle{N}(::Val{N}) where {N} = MyStyle{N}()
```
If package authors had implemented the first one, I think it was just a misreading of the docs or maybe they were using it themselves explicilty elsewhere. 

This PR does require the second form to be defined in more cases than before, but at the benefit of making the logic correct.

matthias314:
> I think it was just a misreading of the docs

The docstring for `Broadcast.AbstractArrayStyle` explicitly suggests
```
(::Type{<:MyArrayStyleDim})(::Val{N}) where N = MyArrayStyleDim{N}()
```
which includes the `MyArrayStyleDim(::Val{N})` constructor. If that is not needed, then maybe one can adjust the docstring.

adienes:
so how to progress? should we try removing `DefaultArrayStyle(::Val{N}) where N = DefaultArrayStyle{N}()` (and updating that docstring) and rerun pkgeval?

DilumAluthge:

@adienes @marius311 What is the current status of this PR?

marius311:
I can refamiliarize myself with this in the next couple of days and decide on a solution to the stuff that arose in the last few comments. Still think it'd be great to get this fixed.

PR #35952, opened 2020-05-20
Split `show_ir` so line printer can be used elsewhere

timholy:
JuliaInterpreter & Debugger need to print just a subset of CodeInfo statements (those surrounding the current program counter). Currently they print the whole thing to an IOBuffer, split the lines, and keep just a subset. Unfortunately, because some statements print over multiple lines, it's easy to get misaligned.

This makes it possible to call the standard printer for a specified line.

Compared to the previous implementation, this does do a little bit of redundant work on each statement (recomputation of `max_bb_idx_size` and `maxlength_idx`), but that's <100ns of work so I don't think it's that big a deal. Alternatively I could add them as arguments.

### DIFF
--- a/base/compiler/ssair/show.jl
+++ b/base/compiler/ssair/show.jl
@@ -662,17 +662,14 @@ function show_ir(io::IO, code::IRCode, expr_type_printer=default_expr_type_print
     end
 end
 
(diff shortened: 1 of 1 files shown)


timholy:
You can see a preview of some of the kinds of display that are possible at https://juliadebug.github.io/LoweredCodeUtils.jl/previews/PR30/edges/.

Since this seems fairly noninvasive, and I've written into the docs for an important Revise update (one that will improve experience for all of Julia 1.x, not just 1.6), I plan to merge this soon.

Issue #7267, opened 2014-06-14
bundle Julia with a better terminal on Windows?


StefanKarpinski:
This seems like the nicest console emulator around on Windows:

~~http://bliker.github.io/cmder/~~
~~http://cmder.net~~
https://mintty.github.io

I think it would make sense, since Windows is not really a DIY platform, but more of a download-and-use-the-thing-that-you-can-click-on platform, that we include a really nice console emulator for people to use.

EDIT: fixed link
EDIT (@ViralBShah): mintty as the preferred choice

ncnc:
FWIW, I tried a fork of `Console2` called `ConsoleZ`. Changing the defaults I get a reasonably sane copy-paste behavior (select with left mouse button), but similar to `Console2` the unicode issues remain.

StefanKarpinski:
We still need to ship with a better terminal emulator on Windows. Watching people struggle with `cmd.exe` is awful – Julia needs to provide a better default experience out of the box.

IainNZ:
I agree, if the details can be hammered out. Although IJulia is certainly smoothing over this, at least in Julia classes I've been involved in over the last month. 

(A higher priority would be anything to speed up `Pkg` operations, which seem extra slow on Windows.)

tkelman:
As I said in December https://github.com/JuliaLang/julia/issues/7267#issuecomment-67761899, I will work on incorporating mintty. But not until we get rid of command-line git, since this will all require pretty significant rewriting of how the Windows binary builds get put together. Can we get moving on the libgit2 transition?

StefanKarpinski:
Right, thanks @tkelman – I mainly wanted to reopen the issue since it's still unresolved.

davidanthoff:
I like the plan with mintty, but I think it would be good if this was not the default if Julia is installed on a Win10 system. Win10 comes with a major overhaul over the console that probably solves all of the current issues, and I feel Julia should just adhere to the (then) good system default. Current Win10 RTM target seems to be June/summer, so certainly before Julia 0.5, and that kind of means that 0.4 will be the version that people will use on Win10.

tkelman:
[...]

PR #35960, opened 2020-05-20
Add a recommended terminal for Windows to README

musm:
I have also updated https://julialang.org/downloads/platform/ to include similar instructions, CF https://github.com/JuliaLang/www.julialang.org/pull/825

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,9 @@ The Julia REPL is quite powerful.  See the section in the manual on
 [the Julia REPL](https://docs.julialang.org/en/v1/stdlib/REPL/)
 for more details.
 
+On Windows we highly recommend running Julia using a modern terminal,
+such as installing the [Windows Terminal from the Microsoft Store](https://aka.ms/terminal).
+
 Support for editing Julia is available for many
 [widely used editors](https://github.com/JuliaEditorSupport):
 [Emacs](https://github.com/JuliaEditorSupport/julia-emacs),


StefanKarpinski (approved):

README.md
```diff
@@ -134,6 +134,9 @@ The Julia REPL is quite powerful.  See the section in the manual on
 [the Julia REPL](https://docs.julialang.org/en/v1/stdlib/REPL/)
 for more details.
 
+On Windows we highly recommend running Julia using a modern terminal,
+such as installing the [Windows Terminal from the Microsoft Store](https://aka.ms/terminal).
```
stevengj, line 138: Probably need some comment on how to use Julia from this terminal, since double-clicking `julia.exe` will open the console.

musm: We have more detailed instructions here https://julialang.org/downloads/platform/
Perhaps a little out of place to provide more details here? I think it would be better to add more details to the download page.

GregPlowman: Are you sure instructions are there? I think Steven is saying that installing Windows Terminal is not enough. How do you run Julia in Windows Terminal? And especially how to make this happen by simply clicking on `julia.exe` shortcut?

musm: Fair enough, we can add more instructions over at JuliaLang..org, but I don't think adding more than this is appropriate for the README

ViralBShah: Definitely should not be in the README. Ideally the installation instructions should be in the Julia documentation. For now, we should put them on the website - where they currently live.

Also, I feel we don't want to recommend Windows terminal, since it doesn't work on Windows 7, for example. Is there a generic way to describe what terminal is required and then list Windows terminal as one of the examples?

musm: @ViralBShah  Windows 7, is not even supported by Microsoft at this point, it was dropped at the start of the year, I don't see why we need to accommodate 100% functionality with users using an unsupported operating system from Microsoft itself. 

Windows 10 forces auto upgrades, as a result recommending the officially supported terminal sounds the most sensible thing here.

ViralBShah: Don't you need a recent Windows 10 from last year for Windows Terminal or is that fixed in the recent release? Julia currently does run on Windows 7! It can't hurt to add a single line (perhaps only on the Julia website) that a usable terminal should be provided.

If we want to bump the minimum Windows version we support to Windows 10, then let's get that decision taken.

musm: Yes you need a relatively recent version, but Windows 10 forces auto-updates on users, so it's not too important to worry about this case. 

Right, if a user is on Windows 7, they should find a better terminal and this doesn't change anything from the current scenario. This one-liner is strictly better than making no recommendation.

musm: @GregPlowman this is a fair point but that's actually an issue on the OS level that needs resolving and currently there is NO solution to this problem for any terminal on Windows. 

What we can do is create a `.bat`  called `Start Julia using the Windows Terminal` which could check if Windows Terminal exists and launch Julia from there if it exists. This would be a temporary solution, until OS level integration of the Windows Terminal is incorporated.

ViralBShah: We can merge this particular PR - but until we address the various things here, I don't think we should close #7267.

musm: @ViralBShah What exactly do we need to address to close that issue?  I want to put that issue out of its misery, since it hasn't been productive at all.

ViralBShah: I believe the following things would suffice to close it:

1. Instructions on how to use Windows Terminal, or even better, the .bat file solution that works automatically and falls back to the Console or some such thing.

2. In the installation instructions, some lines for those who don't have Windows Terminal. Many people have said in issues that their work machines are quite behind on Windows OS updates by years. So at least a clear message for those users that they can use other terminals. For those, we don't need detailed instructions or anything beyond the note.

musm: For 2, where should those instruction live?

stevengj: If using an alternative terminal with Julia is so complicated that we can't even document it clearly, that's not a good sign for the ability of new, unsophisticated users to be able to take our advice.

ViralBShah: I think the non-Terminal instructions should be literally what I do with my Windows 7 VM. Install git-bash, and start Julia with the full path.

ViralBShah:
Merging this. The other tasks can be in follow-up PRs.

README.md
musm: > If using an alternative terminal with Julia is so complicated that we can't even document it clearly, that's not a good sign for the ability of new, unsophisticated users to be able to take our advice.

It really isn't. The only issue is people on Windows 7, which is a 10 year old OS not even supported by Microsoft at this point, yet for some reason we need to accommodate them in the instructions.

ViralBShah: You don't need to do anything you don't like doing!

KristofferC: Dropping support for something doesn't mean it suddenly stops working. I think it literally means

> We will not go out of our way to support these old things.

fonsp:
From @musm:
> What we can do is create a .bat called Start Julia using the Windows Terminal which could check if Windows Terminal exists and launch Julia from there if it exists. This would be a temporary solution, until OS level integration of the Windows Terminal is incorporated.

I wrote this .bat file today, but I realized that I see no improvement in UX when using the Windows Terminal - unicode support is equally bad, the only advantage seems that I found are anti-aliased fonts.

What am I missing - why do we recommend users to install the Windows Terminal?

musm:
Unicode support is generally related to the font you are using so if you are using a font without many Unicode glyphs then that would explain your issue. What exactly is the rendering issue?
 
There's too many advantages of the Windows Terminal to list here compared to conhost. To get improved support you will need to use the nightly builds, since they require libuv updates that are not available on 1.5.  E.g. we finally get 24-bit color support in Julia (test using julia-nightly builds to try that). I'm quite surprised you say there's no improvement in UX compared to conhost, I find the opposite to be true: there's too many improvements to list compared to conhost.... multiple tabs/split panes/search/24-bit color/rendering/ the list is large.

fonsp:
Yeah I am puzzled by the rendering issue. It's not a problem with the font - it works on WSL and SSH:

image

but on powershell and cmd.exe, the glyphs don't render, and the indexing is off: (the caret is at the end of the line in this screenshot)

image

```
a😀💩
```

(I don't think that this is an issue with Julia, maybe we should discuss this elsewhere?)

> there's too many improvements to list compared to conhost.... multiple tabs/split panes/search/24-bit color/rendering/ the list is large.

You're right, thanks 👍

KristofferC:
Try with nightly, I think some of these things have gotten fixed.

musm:
Ah so you are looking for emoji support, well I'm still waiting on libuv folks to merge https://github.com/libuv/libuv/pull/2971 which we will then need to backport to our libuv branch to get emojis to  pass through.

musm:
BTW just use 

```
."C:\users\fonsv\AppData\Local\Programs\Julia 1.5.1\bin\julia.exe"
```

musm:
@fonsp  fixed on nightly now  😎

fonsp:
The emoji problem? Awesome! Thanks so much!

PR #35962, opened 2020-05-20
change Msg f type from Function to Any

### LINKED ISSUE #34333: `remotecall_fetch` and `CallMsg` don't accept type (constructor) functions
We had a problem trying to use `remotecall_fetch(OurType, worker, args...)` because `OurType` isn't a `Function`.

I'm wondering if the types at https://github.com/JuliaLang/julia/blob/master/stdlib/Distributed/src/messages.jl#L39 are too restrictive? Could this be `Callable`, or `Any`? (Or might this cause any potential problems with remote workers that I'm not aware of?)

ssikdar1:
# Description
Fix #34333 . Change `f::Function` to have a type of `Any`.

# Testing
Ran `$ ./julia stdlib/Distributed/test/runtests.jl ` without any issues.


musm:
Add test?

ssikdar1:
@musm 
Since the change was to match the `Any` type of f in these signatures 
```
remotecall(f, w::Worker, args...; kwargs...)
remotecall_wait(f, w::Worker, args...; kwargs...) 
remote_do(f, w::Worker, args...; kwargs...)
```
which call `CallMsg{:call}(f, args, kwargs)` , `CallWaitMsg(f, args, kwargs)`, `RemoteDoMsg(f, args, kwargs)` respectively , I think the preexisting tests for  `remotecall`, `remotecall_wait` and `remote_do` in `test/distributed_exec.jl` would be enough?

If there are anymore specific tests I should add, happy to add them!

JeffBezanson:
The issue is that `remotecall` does not currently work with non-Function callable objects, so that case isn't tested yet.

ssikdar1:
@JeffBezanson @musm 

Added a test of passing the example in function like objects (https://docs.julialang.org/en/v1/manual/methods/#Function-like-objects-1)
to `remotecall` and `remotecall_wait`. In that test `p` is of type `Polynomial`.

stdlib/Distributed/test/distributed_exec.jl
```diff
@@ -1682,6 +1682,30 @@ let (h, t) = Distributed.head_and_tail(Int[], 0)
...
+        end
+        return v
+    end
+
+    p = Polynomial([1,10,100])
+    args = 2
+    result = p(2)
+    @test fetch(remotecall(p, id, args)) == result
+    @test fetch(remotecall_wait(p, id, args)) == result
+end
```
JeffBezanson, line 1707: This is more elaborate than necessary, and doesn't test remote workers. `remotecall(Float64, id, 1)` would suffice.

ssikdar1: @JeffBezanson  Simplified the tests  in bcdf987, does `remotecall_fetch` test the remote workers case?

JeffBezanson: The case of `id = 1` happens to already work (calls on the local process don't need to send a message so they go through a different path), so it needs to be a different id for the test to be meaningful.

ssikdar1: @JeffBezanson changed from 1 to id_other in 2ae9479, since id_other is defined as `id_other = filter(x -> x != id_me, procs())[rand(1:(nprocs()-1))]`

stdlib/Distributed/test/distributed_exec.jl
```diff
@@ -1399,6 +1399,13 @@ let thrown = false
     @test thrown
 end
 
+# issue #34333
+let
+    @test fetch(remotecall(Float64, id_other, 1)) == Float64(1)
```
JeffBezanson, line 1404: This is fine but in the future it's better to use `===` here to check that the type also matches.

PR #35964, opened 2020-05-20
Make clamp return NaN if any of the arguments is NaN

### LINKED ISSUE #35942: clamp(x, l, u) should always propagate NaNs
```julia
julia> clamp(NaN, 0, 1)
NaN
julia> clamp(0.5, NaN, 1)
0.5
julia> clamp(0.5, 0, NaN)
0.5
```

In the last two cases, I'd expect `clamp` to propagate `NaN`s.

bicycle1885:
This fixes #35942.

I'm not perfectly sure whether this is a breaking change or not. But I regard it as a kind of bug fix because most floating-point operations involving NaNs result in a NaN.


base/math.jl
```diff
@@ -62,11 +62,7 @@ julia> clamp.([11,8,5],10,6) # an example where lo > hi
  10
 ```
 """
-clamp(x::X, lo::L, hi::H) where {X,L,H} =
-    ifelse(x > hi, convert(promote_type(X,L,H), hi),
-           ifelse(x < lo,
-                  convert(promote_type(X,L,H), lo),
-                  convert(promote_type(X,L,H), x)))
+clamp(x, lo, hi) = max(lo, min(hi, x))
```
StefanKarpinski, line 65: I appreciate the simplicity of this definition, but I'm concerned that this will significantly slower. Could you benchmark before and after?

raminammour: As Stefan noted, the use `ifelse` here is for performance (it can eliminate branches in tight loops). A benchmark with SIMD vectorization is good to have here.

cossio: But `max` and `min` use `ifelse` too, so they don't introduce new branches.

cossio: However the new definition does produce a lot more `llvm` code.

```
julia> @code_llvm debuginfo=:none clamp(x,l,u) # original definition

define double @julia_clamp_17348(double, double, double) {
top:
  %3 = fcmp uge double %2, %0
  %.inv = fcmp olt double %0, %1
  %4 = select i1 %.inv, double %1, double %0
  %5 = select i1 %3, double %4, double %2
  ret double %5
}

julia> myclamp(x, lo, hi) = max(lo, min(hi, x));

julia> @code_llvm debuginfo=:none myclamp(x,l,u)

define double @julia_myclamp_17337(double, double, double) {
top:
  %3 = fcmp olt double %0, %2
  %4 = bitcast double %0 to i64
  %5 = bitcast double %2 to i64
  %6 = icmp sgt i64 %5, -1
  %7 = icmp slt i64 %4, 0
  %8 = and i1 %7, %6
  %9 = or i1 %3, %8
  %10 = fcmp ord double %2, 0.000000e+00
  %11 = select i1 %10, double %0, double %2
  %12 = fcmp ord double %0, 0.000000e+00
  %13 = select i1 %12, double %2, double %0
  %14 = select i1 %9, double %11, double %13
  %15 = fcmp ogt double %14, %1
  %16 = bitcast double %14 to i64
  %17 = bitcast double %1 to i64
  %18 = icmp sgt i64 %16, -1
  %19 = icmp slt i64 %17, 0
  %20 = and i1 %19, %18
  %21 = or i1 %15, %20
  %22 = fcmp ord double %1, 0.000000e+00
  %23 = select i1 %22, double %14, double %1
  %24 = fcmp ord double %14, 0.000000e+00
  %25 = select i1 %24, double %1, double %14
  %26 = select i1 %21, double %23, double %25
  ret double %26
}

```

raminammour: There seems to be a performance hit, though I see vectorization instructions in both cases:

```

julia> f(x)=clamp.(x,0.2,0.9);

julia> g(x)=myclamp.(x,0.2,0.9);

julia> x=rand(10^4);

julia> @btime f($x);
  4.992 μs (2 allocations: 78.20 KiB)

julia> @btime g($x);
  6.939 μs (2 allocations: 78.20 KiB)
```
(Hand written versions of `f,g` without broadcasting, show the same results).

Hope this helps!

raminammour: Something like this does not seem to have much performance penalty:
```
function myclamp(x::X, lo::L, hi::H) where {X,L,H}
       isnan(lo) && return NaN
       isnan(hi) && return NaN
           ifelse(x > hi, convert(promote_type(X,L,H), hi),
                  ifelse(x < lo,
                         convert(promote_type(X,L,H), lo),
                         convert(promote_type(X,L,H), x)))
end

julia> @btime g($x);
  5.011 μs (2 allocations: 78.20 KiB)

julia> @btime f($x);
  4.992 μs (2 allocations: 78.20 KiB)
```
The merits of special casing `NaN` in the above implementation is another issue :)

cossio: `&&` does introduce a branch. Don't know if it's a bad thing here though.

cossio: How about:

```julia
myclamp(x::Real, l::Real, u::Real) = myclamp(promote(x,l,u)...)

function myclamp(x::Real, l::Real, u::Real)
	xf, lf, uf = float.(promote(x, l, u))
	ifelse(xf < lf, lf,
		ifelse(xf > uf, uf,
			ifelse(lf ≤ xf ≤ uf, xf,
				oftype(xf, NaN))))
end
```

cossio: `max(lo, min(hi, x))` does feel like the right answer. I don't understand why it's slower.

cossio: @raminammour's version seems best. I would write it like this:

```julia
function myclamp(x::X, lo::L, hi::H) where {X,L,H}
   T = promote_type(X,L,H)
   isnan(lo) && return convert(T, NaN)
   isnan(hi) && return convert(T, NaN)
   ifelse(x > hi, convert(T, hi),
          ifelse(x < lo,
                 convert(T, lo),
                 convert(T, x)))
end
```

so we return `NaN` of the correct type. With this I get these timing:

```
julia> @btime f($x);
  3.436 μs (2 allocations: 78.20 KiB)

julia> @btime g($x);
  3.752 μs (2 allocations: 78.20 KiB)
```

Though `clamp` is consistently slightly faster than `myclamp`, I'd say the performance hit is negligible and it seems that a cost is inevitable if add a check for `NaN`.

bicycle1885: Thanks! I'm tackling it.

bicycle1885:
After struggling to improve the performance to address [@StefanKarpinski's concern](https://github.com/JuliaLang/julia/pull/35964#discussion_r428251180), the performance is significantly improved. But it looks too good to believe: it's now almost two times faster than the original implementation in Base while it does more jobs. Here is my quick benchmark using Julia 1.4.1:

```julia
using Base: AbstractFloat

function clamp(x, lo, hi)
    ismissing(x) && return x
    let (x, lo, hi) = promote(x, lo, hi)
        if x isa AbstractFloat
            x = ifelse((x < lo) | isnan(lo), lo, x)
            x = ifelse((x > hi) | isnan(hi), hi, x)
            return x
        else
            return ifelse(x > hi, hi, ifelse(x < lo, lo, x))
        end
    end
end

function clamp!(x, lo, hi)
    @inbounds for i in eachindex(x)
        x[i] = clamp(x[i], lo, hi)
    end
    return x
end

# quick test
@assert clamp(NaN, 1, 2) === NaN
@assert clamp(0, NaN, 2) === NaN
@assert clamp(0, 1, NaN) === NaN
@assert clamp(missing, 1, 2) === missing

# Benchmarking
using BenchmarkTools
using Random

Random.seed!(12345)
N = 1024 * 16
x = randn(N)
lo, hi = -1, 1

@assert clamp!(copy(x), lo, hi) == Base.clamp!(copy(x), lo, hi)
for i in 1:3
    print("    copy:"); @btime copy(x)  # overhead
    print("   clamp:"); @btime clamp!(copy(x), lo, hi)
    print("baseline:"); @btime Base.clamp!(copy(x), lo, hi)
    i < 3 && println()
end
```

And here is the result on my machine (`copy` is overhead of `clamp` and `baseline`):
```
    copy:  9.277 μs (2 allocations: 128.08 KiB)                                                                                                                                                                                                                                              
   clamp:  11.221 μs (2 allocations: 128.08 KiB)                                                                                                                                                                                                                                             
baseline:  21.009 μs (2 allocations: 128.08 KiB)                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                             
    copy:  8.632 μs (2 allocations: 128.08 KiB)                                                                                                                                                                                                                                              
   clamp:  11.421 μs (2 allocations: 128.08 KiB)                                                                                                                                                                                                                                             
baseline:  21.449 μs (2 allocations: 128.08 KiB)                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                             
    copy:  8.546 μs (2 allocations: 128.08 KiB)                                                                                                                                                                                                                                              
   clamp:  11.060 μs (2 allocations: 128.08 KiB)                                                                                                                                                                                                                                             
baseline:  21.229 μs (2 allocations: 128.08 KiB)            
```

These are the generated code and versioninfo:
```
julia> code_native(clamp, (Float64, Float64, Float64), syntax = :intel)
        .text
; ┌ @ clamp.jl:4 within `clamp'
        vmaxsd  xmm0, xmm1, xmm0
        vcmpordsd       xmm3, xmm1, xmm1
        vblendvpd       xmm0, xmm1, xmm0, xmm3
; │ @ clamp.jl:8 within `clamp'
        vcmpordsd       xmm1, xmm2, xmm2
        vminsd  xmm0, xmm2, xmm0
        vblendvpd       xmm0, xmm2, xmm0, xmm1
; │ @ clamp.jl:9 within `clamp'
        ret
        nop
; └

julia> code_native(Base.clamp, (Float64, Float64, Float64), syntax = :intel)
        .text
; ┌ @ math.jl:65 within `clamp'
        vmaxsd  xmm1, xmm1, xmm0
        vcmpnltsd       xmm0, xmm2, xmm0
        vblendvpd       xmm0, xmm2, xmm1, xmm0
        ret
; └

julia> versioninfo()
Julia Version 1.4.1
Platform Info:
  CPU: AMD Ryzen 5 2400G with Radeon Vega Graphics
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-8.0.1 (ORCJIT, znver1)
Environment:
  JULIA_PROJECT = @.
```

cossio:
I think the `isnan` checks are elided because `lo, hi = -1,1` are ints in those benchmarks. On my laptop with `lo, hi = -1.0, 1.0` I get closer times, though still the new implementation looks faster!

```
julia> include("/home/cossio/Desktop/julia/clamp.jl")
    copy:  7.317 μs (2 allocations: 128.08 KiB)
   clamp:  10.756 μs (2 allocations: 128.08 KiB)
baseline:  14.907 μs (2 allocations: 128.08 KiB)

    copy:  9.299 μs (2 allocations: 128.08 KiB)
   clamp:  13.438 μs (2 allocations: 128.08 KiB)
baseline:  16.524 μs (2 allocations: 128.08 KiB)

    copy:  9.423 μs (2 allocations: 128.08 KiB)
   clamp:  13.232 μs (2 allocations: 128.08 KiB)
baseline:  18.988 μs (2 allocations: 128.08 KiB)
```

bicycle1885:
Oops, you're right. So the truth is `clamp` was slower for integer bounds for some reason and now the performance has been improved.

cossio:
We should tests  for when `lo > hi`. It's stated in the docs that in this case:

> If x > hi, return hi. If x < lo, return lo.

So we should test for that behavior, to make sure the implementation matches it. That is, cases like:

```
@test clamp(5, 7, 2) == 2
@test clamp(1, 7, 2) == 7
@test clamp(2, 7, 2) == 7
@test clamp(7, 7, 2) == 7
@test isnan(clamp(NaN, 7, 2))
```

raminammour:
Wouldn't this be a bit more "Julian"? (multiple dispatch vs branching on Types)
```
julia> clamp(x,lo,hi)=begin 
          ismissing(x) && return missing
          clamp(promote(x,lo,hi)...)
       end

julia> clamp(x::T,lo::T,hi::T) where T<:Real =ifelse(x > hi, hi, ifelse(x < lo, lo, x))

julia> function clamp(x::T,lo::T,hi::T) where T<:AbstractFloat
              x = ifelse((x < lo) | isnan(lo), lo, x)
                         x = ifelse((x > hi) | isnan(hi), hi, x)
                         return x
          end
```

bicycle1885:
> We should tests for when lo > hi. 

Those cases are already tested. Please check inside test/math.jl.

> Wouldn't this be a bit more "Julian"? (multiple dispatch vs branching on Types)

Really. But the current code is already enough Julian IMO.

cossio:
> Those cases are already tested. Please check inside test/math.jl.

Am I missing something?

image

All those tests have `lo < hi`.

bicycle1885:
Oh, sorry! I should have read your comment more carefully.

bicycle1885:
I'm not perfectly sure whether the behavior is intentional. For example, the behavior of `std::clamp` is undefined if `lo` is greater than `hi` (https://en.cppreference.com/w/cpp/algorithm/clamp).

cossio:
I agree the documented behavior is weird. So either we change the docs, or enforce the documented behavior.

At this point changing the docs (and the behavior) would be a breaking change (in contrast to propagating NaNs which is a bugfix).

So I'd say here we just continue to enforce the documented behavior for compatibility (that is, add the tests with `lo > hi`).

In a separate issue we can discuss whether this behavior is desired or not.

bicycle1885:
This test case `@test clamp(7, 7, 2) == 7` conflicts with the current behavior (the result is `2` on Julia 1.4.1). So, if we add the test case and change the code to match, it would be breaking. I'd like to keep the current behavior as it is because the documented behavior may not  be intentional.

cossio:
Yes, sorry was a typo. Should be:

```
@test clamp(7, 7, 2) == 2
```

which is consistent with the current documentation.

StefanKarpinski:
I don't think we need to exactly match the current behavior when `lo > hi`: this can probably be considered undefined behavior and changing it to be a minor change. The motivation for the existing behavior was mainly based on what was the fastest, not the best. Arguably, we should just clamp the input value between `min(lo, hi)` and `max(lo, hi)` so that the order of the clamp bounds doesn't actually matter.

cossio:
This version propagates `missing` on `lo` and `hi` as well as on `x`:

```julia
clamp(x, lo, hi) = clamp(promote(x, lo, hi)...)

function clamp(x::T, lo::T, hi::T) where {T}
	if ismissing(x) || ismissing(lo) || ismissing(hi)
		return missing
	else
	    x = ifelse((x < lo) | isnan(lo), lo, x)
	    x = ifelse((x > hi) | isnan(hi), hi, x)
    	return x
    end
end
```

I get these timings for this version:

```
    copy:  9.835 μs (2 allocations: 128.08 KiB)
   clamp:  16.879 μs (2 allocations: 128.08 KiB)
baseline:  16.283 μs (2 allocations: 128.08 KiB)

    copy:  9.869 μs (2 allocations: 128.08 KiB)
   clamp:  16.925 μs (2 allocations: 128.08 KiB)
baseline:  17.477 μs (2 allocations: 128.08 KiB)

    copy:  11.932 μs (2 allocations: 128.08 KiB)
   clamp:  17.350 μs (2 allocations: 128.08 KiB)
baseline:  16.848 μs (2 allocations: 128.08 KiB)
```

cossio:
It also costs us nothing to include @StefanKarpinski suggestion of clamping to the interval `(min(lo,hi), max(lo,hi))`, which I find much more natural than the currently documented behavior.

```
clamp(x, lo, hi) = clamp(promote(x, lo, hi)...)

function clamp(x::T, lo::T, hi::T) where {T}
	if ismissing(x) || ismissing(lo) || ismissing(hi)
	    return missing
	else
	    lo, hi = ifelse(lo > hi, (hi, lo), (lo, hi))
	    x = ifelse((x < lo) | isnan(lo), lo, x)
	    x = ifelse((x > hi) | isnan(hi), hi, x)
    	return x
    end
end
```

Timings

```
    copy:  9.259 μs (2 allocations: 128.08 KiB)
   clamp:  15.368 μs (2 allocations: 128.08 KiB)
baseline:  17.210 μs (2 allocations: 128.08 KiB)

    copy:  9.642 μs (2 allocations: 128.08 KiB)
   clamp:  15.493 μs (2 allocations: 128.08 KiB)
baseline:  15.957 μs (2 allocations: 128.08 KiB)

    copy:  9.615 μs (2 allocations: 128.08 KiB)
   clamp:  18.360 μs (2 allocations: 128.08 KiB)
baseline:  19.481 μs (2 allocations: 128.08 KiB)
```

JeffBezanson:
If possible, I think we should only call `isnan` for float (or at least `Real`) arguments, since that has caused problem with generic code in the past.

I also think it's ok to assume lo<=hi if we can get any speed from it. Allowing either order doesn't do anybody any favors for code readability either.

cossio:
> If possible, I think we should only call `isnan` for float (or at least `Real`) arguments, since that has caused problem with generic code in the past.

I thought `isnan(x)` would be elided for non-float `x`?

JeffBezanson:
We have `isnan(::Real)` but not `isnan(::Any)`.

cossio:
> We have isnan(::Real) but not isnan(::Any).

We do have `isnan(::Real)` and `isnan(::Complex)`. Perhaps we could merge that into `isnan(::Number)`. 

Not sure if we should go as high as `isnan(::Any)`. 

It is also fine if `clamp` errors for non-numeric inputs.

cossio:
Bump?

test/math.jl
```diff
@@ -21,6 +21,10 @@ end
     @test clamp(3.0, 1, 3) == 3.0
     @test clamp(4.0, 1, 3) == 3.0
 
+    @test clamp(NaN, 1, 2) === NaN
+    @test clamp(0, NaN, 2) === NaN
+    @test clamp(0, 1, NaN) === NaN
```
StefanKarpinski, line 26:
```suggestion
    @test clamp(NaN, 0, 1) === NaN
    @test clamp(NaN, 1, 0) === NaN
    @test clamp(0, NaN, 1) === NaN
    @test clamp(1, NaN, 0) === NaN
    @test clamp(0, 1, NaN) === NaN
    @test clamp(1, 0, NaN) === NaN
    @test clamp(NaN, NaN, 0) === NaN
    @test clamp(NaN, 0, NaN) === NaN
    @test clamp(0, NaN, NaN) === NaN
    @test clamp(NaN, NaN, NaN) === NaN
```

adienes:
clamp is confusing because if you think of it like `min(max(value, lo), hi)` it should propagate, but if you think of it like `If x > hi, return hi. If x < lo, return lo. Else, return x` which is more similar to the docstring then it won't propagate

note that in C++ , `std::clamp` also matches numpy behavior (`nan` = no bound)

in Rust, they panic when either bound is `nan` . I doubt Julia wants to copy that. there's an associated discussion here https://internals.rust-lang.org/t/clamp-function-for-primitive-types/4999

as currently documented, I don't think the docs make a prescription as to how to handle `NaN`. given that I personally think both behaviors would be viable and reasonable, I'd say Julia might as well
* follow the leader and do what numpy  and C++ do
* not make "breaking" changes to this function that's been around for a long time (maybe technically not breaking, but in practice is)

tl;dr I would close this PR and the attached issue

LilithHafner:
From triage,

We should use `min(max(lo, x), hi)` as the implementation

We should document clamp similarly to how min and max are documented (making reference to `isless`). We should explicitly state that if hi isless than lo, the result is unspecified.

This does change the behavior of `clamp(-0.0, 0.0, 0.0)`

DilumAluthge:

@bicycle1885 Are you still interested in working on this PR?

oscardssmith:
Closing as stale.

PR #35971, opened 2020-05-21
Clarify type parameters

ettersi:
This should make the list of possible type parameters easier to find and easier to parse. 

See also https://discourse.julialang.org/t/using-type-system-for-different-coordinate-systems/39722.

### DIFF
--- a/doc/src/manual/types.md
+++ b/doc/src/manual/types.md
@@ -41,11 +41,8 @@ up front are:
     type when the program is running. This is called a "run-time type" in object-oriented languages
     where the combination of static compilation with polymorphism makes this distinction significant.
   * Only values, not variables, have types -- variables are simply names bound to values.
-  * Both abstract and concrete types can be parameterized by other types. They can also be parameterized
-    by symbols, by values of any type for which [`isbits`](@ref) returns true (essentially, things
-    like numbers and bools that are stored like C types or `struct`s with no pointers to other objects),
-    and also by tuples thereof. Type parameters may be omitted when they do not need to be referenced
-    or restricted.
+  * Both abstract and concrete types can be parameterized by types, symbols, numbers and a [few other objects](@ref man-parametric-types). Type parameters may be omitted when they do not need to be referenced or 
+    restricted.
 
 Julia's type system is designed to be powerful and expressive, yet clear, intuitive and unobtrusive.
 Many Julia programmers may never feel the need to write code that explicitly uses types. Some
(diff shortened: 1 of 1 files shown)


ettersi:
Bump

doc/src/manual/types.md
```diff
@@ -41,11 +41,8 @@ up front are:
     type when the program is running. This is called a "run-time type" in object-oriented languages
     where the combination of static compilation with polymorphism makes this distinction significant.
   * Only values, not variables, have types -- variables are simply names bound to values.
-  * Both abstract and concrete types can be parameterized by other types. They can also be parameterized
-    by symbols, by values of any type for which [`isbits`](@ref) returns true (essentially, things
-    like numbers and bools that are stored like C types or `struct`s with no pointers to other objects),
-    and also by tuples thereof. Type parameters may be omitted when they do not need to be referenced
-    or restricted.
+  * Both abstract and concrete types can be parameterized by types, symbols, numbers and a [few other objects](@ref man-parametric-types). Type parameters may be omitted when they do not need to be referenced or
```
StefanKarpinski, line 44:
```suggestion
  * Both abstract and concrete types can be parameterized by types, symbols, numbers and a
    [few other objects](@ref man-parametric-types). Type parameters may be omitted when they do not need to be referenced or 
```

StefanKarpinski (commented):
On the whole this feels like a somewhat random list of things. The first addition is ok since it's clearly not meant to be definitive — it just gives some examples and then refers to another section. The second addition is much more problematic since it gives a seemingly comprehensive list of what types can be parameterized by but the list is not actually comprehensive. If we're going to have a list here, it should either clearly be a list of examples or it should be a comprehensive description of what can be used as a parameter.

doc/src/manual/types.md
```diff
@@ -546,8 +544,14 @@ because Julia is a dynamically typed language and doesn't need to make all type
 time, many traditional difficulties encountered in static parametric type systems can be relatively
 easily handled.
 
-All declared types (the `DataType` variety) can be parameterized, with the same syntax in each
-case. We will discuss them in the following order: first, parametric composite types, then parametric
+All declared types (the `DataType` variety) can be parameterized by the following objects.
```
ettersi, line 547:
```suggestion
All declared types (the `DataType` variety) can be parameterized by the following objects (as well as a few other, more niche objects). 
```

ettersi:
Ok, so two options to resolve this:

 - Make the list complete. I don't know what is missing, so I would need help figuring this out. 
 - Change the paragraph above the list as suggested. 

The first option seems more ideal to me, but the second one is the easy way out.

PR #35975, opened 2020-05-21
allow SubString{String} to participate in filter optimization for String

KristofferC:


JeffBezanson:
It looks to me like this could also be used for `AbstractString` --- both return a `String`, and the only thing they assume about the input is Char iteration and `sizeof`.

StefanKarpinski:
But sizeof is generally not required for strings and won’t be predictive of the result size unless the encoding is UTF-8. Probably better not to make that guess for other string types. Although maybe we could genetically define sizeof as `sizeof(codeunit(s))*ncodeunits(s)`. That’s the right total size even if the representation is weird.

KristofferC:
Let's keep it conservative for now then.

PR #35976, opened 2020-05-21
more precise inference of `splatnew`

JeffBezanson:
This allows constant-folding NamedTuple constructors, in turn allowing constant prop through keyword arguments.

This might make the compiler very slightly slower, but fortunately I don't see a significant effect, maybe half a second building Base. The extra type info definitely seems worth it.


vtjnash (approved):

YingboMa:
I wonder if we can have the ability to disable constant folding for a few arguments. This will help the compile time for functions with default arguments.

ChrisRackauckas:
Can this make it into v1.5 or is that unsafe?

PR #35979, opened 2020-05-21
Bump Zlib BB release to `v1.2.11+10` which enables `-O3` optimisation

giordano:
For reference, this is the corresponding PR to Yggdrasil: https://github.com/JuliaPackaging/Yggdrasil/pull/1051.  Not sure whether I need to update some build commands in case one wants to build zlib from source.

See for example https://github.com/JuliaIO/CodecZlib.jl/issues/53 about the poor performance of current build.

Given the important performance boost, it may be worth to backport this to older releases.


ViralBShah:
Any build changes in BB should also be mirrored in `deps/zlib.mk` for those building from source.

giordano:
Ok, thanks, I'll update tomorrow.  It looks like #33535 should have already switched to using CMake :smiling_imp:

deps/zlib.mk
```diff
@@ -5,7 +5,7 @@ $(eval $(call git-external,zlib,ZLIB,,,$(SRCCACHE)))
 ifneq ($(USE_BINARYBUILDER_ZLIB), 1)
 $(BUILDDIR)/$(ZLIB_SRC_DIR)/build-configured: $(SRCCACHE)/$(ZLIB_SRC_DIR)/source-extracted
 	mkdir -p $(dir $@)
-	cd $(dir $@) && $(dir $<)/configure --prefix=$(abspath $(build_prefix)) --libdir=$(abspath $(build_libdir))
+	cd $(dir $@) && $(CMAKE) -DCMAKE_INSTALL_PREFIX=$(abspath $(build_prefix)) -DCMAKE_BUILD_TYPE=Release -DUNIX=true $(dir $<)
```
giordano, line 8: I'm not entirely sure this is correct, so please triple check what I did :slightly_smiling_face:

ViralBShah: Won't be checked by CI - but I tried it out manually and looks good.

giordano: Thanks!

PR #35980, opened 2020-05-21
Fix type-instability of ∘ and Some when wrapping types

### LINKED ISSUE #34849: Specialize ∘ on type arguments?
The following does not infer:
```julia
julia> f(itr) = map(Int∘sign, itr)

julia> f((1,))
(1,)

julia> @code_warntype f((1,))
Variables
  #self#::Core.Compiler.Const(f, false)
  itr::Tuple{Int64}

Body::Tuple{Any}
1 ─ %1 = (Main.Int ∘ Main.sign)::Core.Compiler.Const(Base.var"#56#57"{DataType,typeof(sign)}(Int64, sign), false)
│   %2 = Main.map(%1, itr)::Tuple{Any}
└──      return %2
```
This occured [here](https://discourse.julialang.org/t/how-can-i-reduce-memory-allocation-and-execution-time-while-ray-tracing-using-voxelraytracers-package/34971).

tkf:
It looks fixing type instability of closure is hard (#35970). So, can we add a workaround for cases like `Float64 ∘ first`?

fix #34849

---

I also added a similar fix for `Some`

Before

```
julia> typeof(Float64 ∘ first)
Base.var"#62#63"{DataType,typeof(first)}

julia> typeof(Some(Int))
Some{DataType}
```

After

```
julia> typeof(Float64 ∘ first)
Base.ComposedFunction{Type{Float64},typeof(first)}

julia> typeof(Some(Int))
Some{Type{Int64}}
```

### DIFF
--- a/base/operators.jl
+++ b/base/operators.jl
@@ -893,8 +893,6 @@ function show(io::IO, c::ComposedFunction)
     show(io, c.g)
 end
 
-show(io::IO, ::MIME"text/plain", c::ComposedFunction) = show(io, c)
-
 """
     !f::Function
 
--- a/base/operators.jl
+++ b/base/operators.jl
@@ -873,9 +873,27 @@ julia> ∘(fs...)(3)
 ```
 """
 function ∘ end
+
+struct ComposedFunction{F,G} <: Function
+    f::F
+    g::G
+    ComposedFunction{F, G}(f, g) where {F, G} = new{F, G}(f, g)
+    ComposedFunction(f, g) = new{Core.Typeof(f),Core.Typeof(g)}(f, g)
+end
+
+(c::ComposedFunction)(x...) = c.f(c.g(x...))
+
 ∘(f) = f
-∘(f, g) = (x...)->f(g(x...))
+∘(f, g) = ComposedFunction(f, g)
 ∘(f, g, h...) = ∘(f ∘ g, h...)
+
+function show(io::IO, c::ComposedFunction)
+    show(io, c.f)
+    print(io, " ∘ ")
+    show(io, c.g)
+end
+
+show(io::IO, ::MIME"text/plain", c::ComposedFunction) = show(io, c)
 
 """
     !f::Function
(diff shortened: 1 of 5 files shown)


base/operators.jl
```diff
@@ -873,10 +873,28 @@ julia> ∘(fs...)(3)
...
+∘(f, g) = ComposedFunction(f, g)
 ∘(f, g, h...) = ∘(f ∘ g, h...)
 
+function show(io::IO, c::ComposedFunction)
+    show(io, c.f)
+    print(io, " ∘ ")
+    show(io, c.g)
+end
+
+show(io::IO, ::MIME"text/plain", c::ComposedFunction) = show(io, c)
```
JeffBezanson, line 896: There is already a fallback for this.

JeffBezanson:
Yes I think this is a reasonable approach for now. For example we already do this for Generators.

base/operators.jl
tkf: Without this, text/plain output is rather ugly:

```
julia> uppercase∘first
(::Base.ComposedFunction{typeof(uppercase),typeof(first)}) (generic function with 1 method)
```

I think this is due to `show(io::IO, ::MIME"text/plain", f::Function)`.

But defining `show` here was incorrect because MIME is not defined here. I moved it to show.jl where `show(io::IO, ::MIME"text/plain", f::Function)` is defined.

JeffBezanson: Oh, I see, I was missing that it's a subtype of Function.

nalimilan:
This is clearly and improvement for composed functions, but is it so clear for `Some`? I can't decide whether it's better to have `Some(Int)` be a `Some{Type{Int}}` or a `Some{DataType}`. For example `[Int] isa Vector{DataType}`... Do we have concrete examples that could justify one choice or the other?

tkf:
It is somewhat unfortunate that `typeof([Int]) === Vector{DataType}` because, IIUC, it is highly discouraged to dispatch on the representation of the type.  People still can use `[Int] isa Vector{<:Type}` but a lot of users may not notice if they see `Vector{DataType}` in the REPL.

> Do we have concrete examples that could justify one choice or the other?

Is type-stability not enough to justify the change?

PR #30222, opened 2018-11-30
 Inliner cost model: don't count :unreachable branches 

timholy:
This improves the heuristic for inlining by not penalizing branches that end with a `throw`. The idea is that exception paths don't reflect the typical run-time cost of a function, since generally one shouldn't be using exceptions for control-flow.

I have some local data suggesting this is helpful, but let's let nanosoldier chime in:
@nanosoldier `runbenchmarks(ALL, vs = ":master")`

Overall this seemed suspiciously easy, so I'm guessing I'm missing something. Would love to have input from someone who knows more.

Replaces #23986. Ref #29798; closing that also requires that we adjust the cost model for some `ccall`s that end up being [special-cased by the compiler](https://github.com/JuliaLang/julia/blob/3dd678f646e4403aede51389fcca7262839b5d15/src/ccall.cpp#L1586-L1793). I haven't yet invested the time to figure out exactly how those should be modeled.

PR #35982, opened 2020-05-22
skip inferring calls that lead to `throw`

JeffBezanson:
This adds a simple linear pass that finds all IR statements obviously followed by a call to `throw`, and then skips inferring generic calls in those slots. Some numbers:

   test | master | PR | PR (fixed)
-------------|-----------:|-------:|---------:
Base                | 38.4       | 35.4 | 35.8
Stdlibs               | 53.8       | 52.8 | 53.3
precompile         | 148.3      | 145.9 | 146.3
sys.so                | 149047528  | 143333400 | 145332832
TTFP                  | 13.4       | 12.7 | 12.8
plot after SIMD       | 10.9       | 9.9 | 10.1
plot after DataFrames | 13.6       | 11.5 | 11.9

### DIFF
--- a/base/compiler/optimize.jl
+++ b/base/compiler/optimize.jl
@@ -285,7 +285,7 @@ plus_saturate(x::Int, y::Int) = max(x, y, x+y)
 # known return type
 isknowntype(@nospecialize T) = (T === Union{}) || isa(T, Const) || isconcretetype(widenconst(T))
 
-function statement_cost(ex::Expr, line::Int, src::CodeInfo, sptypes::Vector{Any}, slottypes::Vector{Any}, params::OptimizationParams)
+function statement_cost(ex::Expr, line::Int, src::CodeInfo, sptypes::Vector{Any}, slottypes::Vector{Any}, params::OptimizationParams, error_path::Bool = false)
     head = ex.head
     if is_meta_expr_head(head)
         return 0
@@ -320,7 +320,7 @@ function statement_cost(ex::Expr, line::Int, src::CodeInfo, sptypes::Vector{Any}
                 return 0
             elseif (f === Main.Core.arrayref || f === Main.Core.const_arrayref) && length(ex.args) >= 3
                 atyp = argextype(ex.args[3], src, sptypes, slottypes)
-                return isknowntype(atyp) ? 4 : params.inline_nonleaf_penalty
+                return isknowntype(atyp) ? 4 : error_path ? params.inline_error_path_cost : params.inline_nonleaf_penalty
             end
             fidx = find_tfunc(f)
             if fidx === nothing
@@ -330,7 +330,11 @@ function statement_cost(ex::Expr, line::Int, src::CodeInfo, sptypes::Vector{Any}
             end
             return T_FFUNC_COST[fidx]
         end
-        return params.inline_nonleaf_penalty
+        extyp = line == -1 ? Any : src.ssavaluetypes[line]
+        if extyp === Union{}
+            return 0
+        end
+        return error_path ? params.inline_error_path_cost : params.inline_nonleaf_penalty
     elseif head === :foreigncall || head === :invoke
         # Calls whose "return type" is Union{} do not actually return:
         # they are errors. Since these are not part of the typical
@@ -347,7 +351,7 @@ function statement_cost(ex::Expr, line::Int, src::CodeInfo, sptypes::Vector{Any}
         end
         a = ex.args[2]
         if a isa Expr
-            cost = plus_saturate(cost, statement_cost(a, -1, src, sptypes, slottypes, params))
+            cost = plus_saturate(cost, statement_cost(a, -1, src, sptypes, slottypes, params, error_path))
         end
         return cost
     elseif head === :copyast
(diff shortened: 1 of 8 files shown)


vchuravy:
cc: @maleadt for possible GPU consequences.

Is this only triggered if one of the types in the signature is not concrete? E.g. if we would have add a `apply_generic` anyway?

JeffBezanson:
No, it skips them unconditionally. I can try that, but I suspect anything less will not give much benefit (since we need to remove backedges to printing code). This is somewhat experimental, because we haven't done many things like this before (flagrantly not inferring some code), but I believe it will be necessary to make a dent in the latency situation.

The best way forward for GPUs etc. might be to run inference with this change disabled. @Keno et. al. have been working on the compiler refactoring needed for that.

JeffBezanson:

nanosoldier:
[Your benchmark job](https://github.com/JuliaLang/julia/pull/35982#issuecomment-632454232) has completed - possible performance regressions were detected. A full report can be found [here](https://github.com/JuliaCI/BaseBenchmarkReports/blob/81dbcd0e7fc317f8805c660b6349a707e1c9b53d/b8b2582_vs_dd1e54d/report.md). cc @ararslan

JeffBezanson:
Ah, of course, the un-inferred callsites in many cases were blocking inlining. I decided to give those statements a flat cost. Added a column to the table above for the new numbers.

JeffBezanson:

timholy:
This looks like it would replace #30222, which I never got around to finishing. I think a fixed cost of 20 is pretty reasonable; it will still affect the inlining decision to some extent, but it will put an end to the current disincentive for crafting a careful error message that reports helpful state information back to the user. That's a win even without the very nice improvements in latency and system image size.

#30222 looks like it had a couple of other changes to the cost model; these may have long since been rendered unnecessary, but it might be worth a glance before you close it.

nanosoldier:
[Your benchmark job](https://github.com/JuliaLang/julia/pull/35982#issuecomment-632980940) has completed - possible performance regressions were detected. A full report can be found [here](https://github.com/JuliaCI/BaseBenchmarkReports/blob/9a862dbd4c3dace0890c20353ea6e35c2a485a2c/2c9ab08_vs_11c942f/report.md). cc @ararslan

JeffBezanson:
Thanks for reminding me about that PR. You were right that this is really more of a reverse-flow analysis; I could use that to add some extra precision for branches. Adding more accurate costs for certain known C functions is also a good idea.

timholy:
Feel free to steal wantonly from that PR, I got stymied and didn't know what to do.

JeffBezanson:
Ok I think this is in good shape now. The analysis should definitely be scrutinized first; it's simple but subtle.

base/compiler/optimize.jl
```diff
@@ -319,7 +319,7 @@ function statement_cost(ex::Expr, line::Int, src::CodeInfo, sptypes::Vector{Any}
                 return 0
             elseif (f === Main.Core.arrayref || f === Main.Core.const_arrayref) && length(ex.args) >= 3
                 atyp = argextype(ex.args[3], src, sptypes, slottypes)
-                return isknowntype(atyp) ? 4 : params.inline_nonleaf_penalty
+                return isknowntype(atyp) ? 4 : error_path ? 20 : params.inline_nonleaf_penalty
```
KristofferC, line 322: Maybe this `20` could be a `params.error_path_penalty`?

base/compiler/utilities.jl
```diff
@@ -212,6 +212,72 @@ function find_ssavalue_uses(e::Expr, uses::Vector{BitSet}, line::Int)
     end
 end
 
+function is_throw_call(e::Expr)
+    if e.head === :call
```
timholy, line 216: Do we also need to handle `invoke` or is that outdated? (https://docs.julialang.org/en/v1.6-dev/devdocs/ast/#Lowered-form-1)

base/compiler/utilities.jl
```diff
@@ -212,6 +212,72 @@ function find_ssavalue_uses(e::Expr, uses::Vector{BitSet}, line::Int)
...
+    stmts = BitSet()
+    n = length(code)
+    try_depth = 0
+    for i in n:-1:1
+        s = code[i]
+        if isa(s, Expr)
+            if s.head === :enter
+                try_depth -= 1
+            elseif s.head === :leave
+                try_depth += (s.args[1]::Int)
```
timholy, line 238: Nice! Does this need special treatment for `:pop_exception`?

timholy (commented):
Looks great!

base/compiler/utilities.jl
JeffBezanson: `invoke` only appears post-optimization for calls to generic functions.

base/compiler/utilities.jl
JeffBezanson: I don't believe so, since enter and leave should be sufficient to identify where handlers are.

JeffBezanson:

nanosoldier:
[Your benchmark job](https://github.com/JuliaLang/julia/pull/35982#issuecomment-643556344) has completed - possible performance regressions were detected. A full report can be found [here](https://github.com/JuliaCI/BaseBenchmarkReports/blob/dd8a1cd0e92b3d90735b0b279fc57cc435dfcbc8/f357a32_vs_1e76111/report.md). cc @ararslan

JeffBezanson:

nanosoldier:
[Your benchmark job](https://github.com/JuliaLang/julia/pull/35982#issuecomment-643684098) has completed - possible performance regressions were detected. A full report can be found [here](https://github.com/JuliaCI/BaseBenchmarkReports/blob/480c1d1eeca6123214f3e06ef3a254c1c2c45add/7ce9dff_vs_5ecb8e2/report.md). cc @ararslan

vtjnash (approved):

PR #35986, opened 2020-05-22
refactor the (number) parsing code

KristofferC:
While looking at the parsing code I noticed that quite a lot of it took three arguments, a`AbstractString`, `startpos`,  `endpos`. This is almost the definition of a `SubString` so that is completely redundant (it is possible the parsing code was written before `SubString` existed.

Therefore I eliminated these three arguments for one `AbstractString` instead and replaced the places that called this with non-default positions with `SubString`.  This actually has some performance benefit since you are not calling `lastindex` unnecessarily. 

There was also quite a lot of inconsistencies in how the parsing errors were printed between different types. Most of this was because the throws were inlined into the body itself so I factored most of this out into functions that throw.

In addition, I had quite a hard time following the dispatch logic in the parsing code. There are an awful lot of `tryparse_internal` overloads and figuring out how things are called is non-trivial. I changed it so that there is now one section for integer parsing, one for float parsing and one for complex parsing, and these sections are more or less independent of each other with a pretty obvious dispatch chain.

Also, different parts of Base (and stdlibs) reached into the internals of the parsing code (directly calling `tryparse_internal`. These now try to just implement the `parse` interface (mostly) independent of the internals of the now new module `Base.Parse`.

In the end, it seems this refactor gave some performance benefits (10%) but that wasn't really the goal.

```jl
julia> strs = string.(rand(-10^4:10^4, 10^7));
```

PR:

```jl
julia> @time parse.(Int, strs);
  0.319752 seconds (5 allocations: 76.294 MiB)
``` 

master:

```jl
julia> @time parse.(Int, strs);
  0.343541 seconds (5 allocations: 76.294 MiB)
```

### DIFF
--- a/test/parse.jl
+++ b/test/parse.jl
@@ -1,5 +1,7 @@
 # This file is a part of Julia. License is MIT: https://julialang.org/license
 
+using Test
+
 @testset "integer parsing" begin
     @test parse(Int32,"0", base = 36) === Int32(0)
     @test parse(Int32,"1", base = 36) === Int32(1)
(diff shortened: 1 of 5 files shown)


test/parse.jl
```diff
@@ -326,3 +305,19 @@ end
     tmp = @. [Int, Float64, String, Bool] <: Union{Int, String}
     @test tmp == Bool[1, 0,1, 0]
 end
+
+@testset "tryparse int on iterators" begin
+    s = Iterators.filter(!=('_'), "1_2_3_4")
+    @test Base.Parse.tryparse_internal_int(Int, s, 10, true) == 1234
```
KristofferC, line 311: This is a bit sneaky. My original intention with this PR was to be able to run `parse` on an arbitrary iterator (to support parsing `"1_2_3_4` as an integer which the TOML specification allows using `Iterators.filter`) and it happens that with the refactoring the part that does the int parsing is now general enough to allow this).

PR #35989, opened 2020-05-22
Fix minimum and maximum in the presence of missing values

nalimilan:
Another, possibly cleaner way of fixing this would be to add an `isnan(::Number) = false` fallback, and do `x isa Number && isnan(x)`. That would make the code a bit clearer and could be useful for others. Currently we would have to do `x isa Union{Real, Complex} && isnan(x)` since there are the only types for which we define `isnan` methods -- but that approach wouldn't work for custom complex types.

Similar to https://github.com/JuliaLang/julia/pull/35323 (but simpler to fix).

### DIFF
--- a/base/reduce.jl
+++ b/base/reduce.jl
@@ -573,10 +573,11 @@ function mapreduce_impl(f, op::Union{typeof(max), typeof(min)},
     simdstop  = start + chunk_len - 4
     while simdstop <= last - 3
         # short circuit in case of NaN
-        v1 == v1 || return v1
-        v2 == v2 || return v2
-        v3 == v3 || return v3
-        v4 == v4 || return v4
+        # === true is there to ignore missing
+        (v1 == v1) === true || return v1
+        (v2 == v2) === true || return v2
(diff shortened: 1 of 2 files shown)


base/reduce.jl
```diff
@@ -573,10 +573,11 @@ function mapreduce_impl(f, op::Union{typeof(max), typeof(min)},
     simdstop  = start + chunk_len - 4
     while simdstop <= last - 3
         # short circuit in case of NaN
-        v1 == v1 || return v1
-        v2 == v2 || return v2
-        v3 == v3 || return v3
-        v4 == v4 || return v4
+        # === true is there to ignore missing
```
nalimilan, line 576: "ignore" isn't the right term actually since we short-circuit. I can change this to just add "or missing" to the existing comment for NaN.

Keno: Just say something like "to handle missing"

nalimilan: Actually "short-circuit" is even more accurate I think.

tkf (approved):
LGTM

PR #35990, opened 2020-05-22
Add arrows to the punctuation list.

nathanrboyer:
My first time using Github. I don't know what I'm doing.

### DIFF
--- a/doc/src/base/punctuation.md
+++ b/doc/src/base/punctuation.md
@@ -54,5 +54,5 @@ Extended documentation for mathematical symbols & functions is [here](@ref math-
 | `=`         | single equals sign is [assignment](@ref man-variables)                                      |
 | [`==`](@ref)| double equals sign is value equality comparison                                             |
 | [`===`](@ref) | triple equals sign is programmatically identical equality comparison.                     |
-| [`=>`](@ref) | right arrow using an equals sign defines a [pair](@ref Base.pair) typically used to populate [dictionaries](@ref Dictionaries) |
-| [`->`](@ref) | right arrow using a hyphen defines an [anonymous function](@ref man-anonymous-funcitons) on a single line         |
+| [`=>`](@ref Pair) | right arrow using an equals sign defines a [`Pair`](@ref) typically used to populate [dictionaries](@ref Dictionaries) |
+| `->` | right arrow using a hyphen defines an [anonymous function](@ref man-anonymous-funcitons) on a single line         |
--- a/doc/src/base/punctuation.md
+++ b/doc/src/base/punctuation.md
@@ -12,7 +12,7 @@ Extended documentation for mathematical symbols & functions is [here](@ref math-
 | `=#`        | end a multi-line comment by immediately preceding the number sign with an equals sign       |
 | `$`         | the dollar sign is used for [string](@ref string-interpolation) and [expression](@ref man-expression-interpolation) interpolation |
 | [`%`](@ref rem) | the percent symbol is the remainder operator                                            |
-| [`^`](@ref) | the caret is the exponentiation operator.                                                   |
+| [`^`](@ref) | the caret is the exponentiation operator                                                  |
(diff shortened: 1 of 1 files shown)


doc/src/base/punctuation.md
```diff
@@ -54,3 +54,5 @@ Extended documentation for mathematical symbols & functions is [here](@ref math-
 | `=`         | single equals sign is [assignment](@ref man-variables)                                      |
 | [`==`](@ref)| double equals sign is value equality comparison                                             |
 | [`===`](@ref) | triple equals sign is programmatically identical equality comparison.                     |
+| [`=>`](@ref) | right arrow using an equals sign defines a [pair](@ref Base.pair) typically used to populate [dictionaries](@ref Dictionaries) |
+| [`->`](@ref) | right arrow using a hyphen defines an [anonymous function](@ref man-anonymous-funcitons) on a single line         |
```
mbauman, line 58:
```suggestion
| [`=>`](@ref Pair) | right arrow using an equals sign defines a [`Pair`](@ref) typically used to populate [dictionaries](@ref Dictionaries) |
| `->` | right arrow using a hyphen defines an [anonymous function](@ref man-anonymous-funcitons) on a single line         |
```

mbauman (commented):
Nice, welcome to GitHub!  This looks just about right, but needs a bit of fiddling to make the `@ref`s hit their targets as you want.  I don't think we have an `@ref` target for `->` since it's not a first-class object (and is indeed just syntax).

doc/src/base/punctuation.md
```diff
@@ -54,3 +54,5 @@ Extended documentation for mathematical symbols & functions is [here](@ref math-
 | `=`         | single equals sign is [assignment](@ref man-variables)                                      |
 | [`==`](@ref)| double equals sign is value equality comparison                                             |
 | [`===`](@ref) | triple equals sign is programmatically identical equality comparison.                     |
+| [`=>`](@ref Pair) | right arrow using an equals sign defines a [`Pair`](@ref) typically used to populate [dictionaries](@ref Dictionaries) |
+| `->` | right arrow using a hyphen defines an [anonymous function](@ref man-anonymous-funcitons) on a single line         |
```
ffevotte, line 58: I did not try to re-build the doc to check, but shouldn't this be `[anonymous function](@ref man-anonymous-functions)`, as in [Anonymous Functions](https://docs.julialang.org/en/v1/manual/functions/#man-anonymous-functions-1)

nathanrboyer: Probably. I don't know how the linking works. I tried inspecting the HTML of the documentation website to get the linking names, but I don't know HTML.  Is the title of the section all I need for the link?

mbauman, line 58: In this case I think it's as simple as a typo fix.

```suggestion
| `->` | right arrow using a hyphen defines an [anonymous function](@ref man-anonymous-functions) on a single line         |
```

ffevotte: I'm not an expert myself, but here is what I'd think you can do in order to get the id without having to know about HTML or anything:
- you want to reference something related to anonymous functions, so you first have to find the relevant page & paragraph in the documentation : this should get you [here](https://docs.julialang.org/en/v1/manual/functions/#man-anonymous-functions-1), which is where you'd like your link to point to.
- clicking on the pencil in the top right part of the page gets you to the markdown source of this part of the documentation (you now know how to do this ; it's the same process you followed to make this pull request in the first place). This should get you [here](https://github.com/JuliaLang/julia/blob/master/doc/src/manual/functions.md).
- now you can look for the relevant [paragraph](https://github.com/JuliaLang/julia/blob/master/doc/src/manual/functions.md#anonymous-functionsid-man-anonymous-functions): its id is written right next to the title: this is the reference you want to use with `(@ref)`.

Thanks and congratulations: not everybody goes through this process of helping improve the documentation!

PR #35995, opened 2020-05-22
micro-optimization for generic tuple allocation

vtjnash:
Not a huge savings, but I wanted to have this available for another reason too, and seemed simple enough to be worthwhile:
```julia
julia> x = 1; @btime (x, x);
  before: 69.130 ns (1 allocation: 32 bytes)
  after: 53.253 ns (1 allocation: 32 bytes)

julia> x = 1; @btime (x, x, x, x, x, x);
  before: 135.513 ns (1 allocation: 64 bytes)
  after: 115.842 ns (1 allocation: 64 bytes)
````

### DIFF
--- a/src/jltypes.c
+++ b/src/jltypes.c
@@ -572,7 +572,45 @@ static int typekey_eq(jl_datatype_t *tt, jl_value_t **key, size_t n)
     return 1;
 }
 
+// These `value` functions return the same values as the primary functions,
+// but operate on the typeof/Typeof each object in an array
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/jltypes.c [hunk 1]
```diff
@@ -572,7 +572,43 @@ static int typekey_eq(jl_datatype_t *tt, jl_value_t **key, size_t n)
     return 1;
 }
 
+static int typekeyvalue_eq(jl_datatype_t *tt, jl_value_t *key1, jl_value_t **key, size_t n, int leaf)
```
JeffBezanson, line 575: Would be good to have a comment that this is the same as the previous function, but with instances in the keys instead of types.

JeffBezanson (approved):

PR #35996, opened 2020-05-22
Add boolean operators to Mathematical Operators

cmcaine:
A student on exercism struggled to find the boolean not operator for Julia. And perhaps this wouldn't help, but maybe it's worth a go.

### DIFF
--- a/doc/src/manual/mathematical-operations.md
+++ b/doc/src/manual/mathematical-operations.md
@@ -70,14 +70,11 @@ The following [boolean operators](https://en.wikipedia.org/wiki/Boolean_algebra#
 | Expression | Name                                  |
 |:---------- |:--------------------------------------|
 | `!x`       | not                                   |
-| `x & y`    | and                                   |
-| `x \| y`   | or                                    |
-| `x ⊻ y`    | xor (exclusive or)                    |
 | `x && y`   | [short-circuiting and][short-circuit] |
 | `x \|\| y` | [short-circuiting or][short-circuit]  |
 | `x == y`   | equivalence                           |
 
-Note that Bool is an integer type and all the usual promotions rules and numeric operators are also defined on it (in fact, `&`, `|`, and `⊻` *are* bitwise operators).
+Note that Bool is an integer type and all the usual promotion rules and numeric operators are also defined on it.
 
 ## Bitwise Operators
 
--- a/doc/src/manual/mathematical-operations.md
+++ b/doc/src/manual/mathematical-operations.md
@@ -63,6 +63,22 @@ julia> false * Inf
 
 This is useful for preventing the propagation of `NaN` values in quantities that are known to be zero. See [Knuth (1992)](https://arxiv.org/abs/math/9205211) for motivation.
 
+## Boolean Operators
+
+The following [boolean operators](https://en.wikipedia.org/wiki/Boolean_algebra#Operations) are supported on Bool:
+
+| Expression | Name                                  |
+|:---------- |:--------------------------------------|
+| `!x`       | not                                   |
+| `x & y`    | and                                   |
+| `x \| y`   | or                                    |
+| `x ⊻ y`    | xor (exclusive or)                    |
+| `x && y`   | [short-circuiting and][short-circuit] |
+| `x \|\| y` | [short-circuiting or][short-circuit]  |
+| `x == y`   | equivalence                           |
+
+Note that Bool is an integer type and all the usual promotions rules and numeric operators are also defined on it (in fact, `&`, `|`, and `⊻` *are* bitwise operators).
+
 ## Bitwise Operators
 
 The following [bitwise operators](https://en.wikipedia.org/wiki/Bitwise_operation#Bitwise_operators)
--- a/doc/src/manual/mathematical-operations.md
+++ b/doc/src/manual/mathematical-operations.md
@@ -75,6 +75,7 @@ The following [boolean operators](https://en.wikipedia.org/wiki/Boolean_algebra#
 
 Note that Bool is an integer type and all the usual promotion rules and numeric operators are also defined on it.
 
+[short-circuit]: https://docs.julialang.org/en/v1/manual/control-flow/#Short-Circuit-Evaluation-1
 ## Bitwise Operators
 
 The following [bitwise operators](https://en.wikipedia.org/wiki/Bitwise_operation#Bitwise_operators)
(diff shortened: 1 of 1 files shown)


cmcaine:
`a <= b` is equivalent to the material implication operator, but I thought that would be best left out ;)

doc/src/manual/mathematical-operations.md
```diff
@@ -63,6 +63,22 @@ julia> false * Inf
...
+## Boolean Operators
+
+The following [boolean operators](https://en.wikipedia.org/wiki/Boolean_algebra#Operations) are supported on Bool:
+
+| Expression | Name                                  |
+|:---------- |:--------------------------------------|
+| `!x`       | not                                   |
+| `x & y`    | and                                   |
+| `x \| y`   | or                                    |
+| `x ⊻ y`    | xor (exclusive or)                    |
```
mbauman, line 75:
```suggestion
```

These are documented below and I'd rather push folks towards using `&&` and `||` for boolean logic as it has the precedence they likely want.

doc/src/manual/mathematical-operations.md
```diff
@@ -63,6 +63,22 @@ julia> false * Inf
...
+| `x && y`   | [short-circuiting and][short-circuit] |
+| `x \|\| y` | [short-circuiting or][short-circuit]  |
+| `x == y`   | equivalence                           |
+
+Note that Bool is an integer type and all the usual promotions rules and numeric operators are also defined on it (in fact, `&`, `|`, and `⊻` *are* bitwise operators).
```
cmcaine, line 80:
```suggestion
Note that Bool is an integer type and all the usual promotion rules and numeric operators are also defined on it.
```

doc/src/manual/mathematical-operations.md
cmcaine: Sure. Done.

doc/src/manual/mathematical-operations.md
```diff
@@ -63,6 +63,19 @@ julia> false * Inf
...
+## Boolean Operators
+
+The following [boolean operators](https://en.wikipedia.org/wiki/Boolean_algebra#Operations) are supported on Bool:
+
+| Expression | Name                                  |
+|:---------- |:--------------------------------------|
+| `!x`       | not                                   |
+| `x && y`   | [short-circuiting and][short-circuit] |
+| `x \|\| y` | [short-circuiting or][short-circuit]  |
+| `x == y`   | equivalence                           |
```
cmcaine, line 75:
```suggestion
```

doc/src/manual/mathematical-operations.md
```diff
@@ -63,6 +63,18 @@ julia> false * Inf
...
+The following [boolean operators](https://en.wikipedia.org/wiki/Boolean_algebra#Operations) are supported on Bool:
+
+| Expression | Name                                  |
+|:---------- |:--------------------------------------|
+| `!x`       | not                                   |
+| `x && y`   | [short-circuiting and][short-circuit] |
+| `x \|\| y` | [short-circuiting or][short-circuit]  |
+
+Note that Bool is an integer type and all the usual promotion rules and numeric operators are also defined on it.
+
```
cmcaine, line 77:
```suggestion

[short-circuit]: https://docs.julialang.org/en/v1/manual/control-flow/#Short-Circuit-Evaluation-1
```

Somehow lost this link

kimikage:
I think `||` and `&&` are not operators *in terms of Julia*. I agree with mentioning `&&` and `||` there, but they should clearly indicate that they are not really "operators".

cmcaine:
I feel like that's emphasised already by linking to the short-circuit documentation? Maybe you can come up with a better phrasing?

The overall alternatives are:

1. Status quo: don't explicitly document the Boolean operators
2. Do as I did originally and duplicate information that occurs elsewhere
3. Only show the operators people actually use in if statements and stuff

The PR's current state is at 3. I think any of these are okay (though I would rank them 2 > 3 > 1) and I'm happy to amend it as suggested if someone with merge authority wants to make a judgement call.

kimikage:
> I feel like that's emphasised already by linking to the short-circuit documentation?

The current state of this PR makes me feel like `&&` and `||` are operators in the "same" category as `!`.
That page implicitly requires general programming knowledge. Therefore I think it is important to clarify what's the same and what's different from other programming languages.
(I'm not good at elegant English writing, but what I want to say is that `&&` and `||` can be used as "so-called" binary operators, but strictly speaking, they are control flow keywords, not defined as methods.)

I also think any of those three are OK, as long as there is no misleading about Julia's syntax.

doc/src/manual/mathematical-operations.md
```diff
@@ -63,6 +63,19 @@ julia> false * Inf
...
+The following [boolean operators](https://en.wikipedia.org/wiki/Boolean_algebra#Operations) are supported on Bool:
+
+| Expression | Name                                  |
+|:---------- |:--------------------------------------|
+| `!x`       | not                                   |
+| `x && y`   | [short-circuiting and][short-circuit] |
+| `x \|\| y` | [short-circuiting or][short-circuit]  |
+
+Note that Bool is an integer type and all the usual promotion rules and numeric operators are also defined on it.
```
dkarrasch, line 76:
```suggestion
Note that `Bool` is an integer type and all the usual promotion rules and numeric operators are also defined on it.
```

doc/src/manual/mathematical-operations.md
```diff
@@ -63,6 +63,19 @@ julia> false * Inf
 
 This is useful for preventing the propagation of `NaN` values in quantities that are known to be zero. See [Knuth (1992)](https://arxiv.org/abs/math/9205211) for motivation.
 
+## Boolean Operators
+
+The following [boolean operators](https://en.wikipedia.org/wiki/Boolean_algebra#Operations) are supported on Bool:
```
dkarrasch, line 68:
```suggestion
The following [Boolean operators](https://en.wikipedia.org/wiki/Boolean_algebra#Operations) are supported on `Bool`:
```

dkarrasch (commented):
Just a few typographical comments.

kimikage:
If we choose the option 3., it might be reasonable to describe "Boolean Operators" after "Bitwise Operators". By doing so, we can simply declare that "all bitwise operators can be used as boolean operators since `Bool` is a `Integer` type".

Note that the `!x` is in the "Arithmetic Operators" section. I think the `!` should only be in the new "Boolean Operators" section.

doc/src/manual/mathematical-operations.md
cmcaine: Thanks :)

ViralBShah:
What's next here?

JeffBezanson:
This seems fine to me.

Issue #35614, opened 2020-04-27
Error by default for when using pivoted Cholesky on semidefinte matrix


olof3:
I tried to get the pivoted Cholesky factorization of a positive semidefinite matrix using
```
C = cholesky(diagm([1, 0]), Val(true))
```
However this throws `RankDeficientException(1)`, even though it seems like the matrix was factorized alright?!

It took some time for me to figure out that the following works.
```
C = cholesky(diagm([1, 0]), Val(true). check=false)
```
To me it seems non-intuitive having to turn off the check for something that should work.

According to the docs one should also be able check `issuccess(C)`, however this throws
`ERROR: MethodError: no method matching issuccess(::CholeskyPivoted{Float64,Array{Float64,2}})`

Here are the docs:

```
  cholesky(A, Val(true); tol = 0.0, check = true) -> CholeskyPivoted

  Compute the pivoted Cholesky factorization of a dense symmetric positive
  semi-definite matrix A and return a CholeskyPivoted factorization. The
  matrix A can either be a Symmetric or Hermitian StridedMatrix or a perfectly
  symmetric or Hermitian StridedMatrix. The triangular Cholesky factor can be
  obtained from the factorization F with: F.L and F.U. The following functions
  are available for CholeskyPivoted objects: size, \, inv, det, and rank. The
  argument tol determines the tolerance for determining the rank. For negative
  values, the tolerance is the machine precision.

  When check = true, an error is thrown if the decomposition fails. When check
  = false, responsibility for checking the decomposition's validity (via
  issuccess) lies with the user.
```

Another inconsistency is that the docs for `LAPACK.pstrf!` says that it works for positive-definite matrices while the [LAPACK docs](https://software.intel.com/en-us/mkl-developer-reference-c-pstrf) says that it works for positive semidefinite matrices. If it's not a typo, a remark in the docs could be in order.

andreasnoack:
I don’t remember if we did this on purpose but a challenge here is that the info code can be different from zero both when it’s PSD and when it’s indefinite.

I think the `issuccess` methods should work even though “success” is a bit ambiguous here but that’s the same for LU.

I’d think that our docs for `pstrf!` are just wrong. Would be great if you could open a PR fixing any or all of these issues.

olof3:
Hmm, yes, it is a bit confusing that LAPACK gives a non-zero info code for PSD matrices. It seems to have caused some confusion in [other places](https://github.com/pytorch/pytorch/issues/12435) as well.

[...]

PR #36002, opened 2020-05-23
add issuccess for CholeskyPivoted

palday:
Trivial change to add `issuccess` method for `CholeskyPivoted`. Let me know if (and where!) I need to add an additional test to cover this.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -47,7 +47,7 @@ Standard library changes
 * All `AbstractUnitRange{<:Integer}`s now work with `SubString`, `view`, `@view` and `@views` on strings ([#35879]).
 
 #### LinearAlgebra
-
+* New method `LinearAlgebra.issuccess(::CholeskyPivoted)` for checking whether pivoted Cholesky factorization was successful ([#36002]).
 
 #### Markdown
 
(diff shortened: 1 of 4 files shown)


ViralBShah:
The test goes in: `stdlib/LinearAlgebra/test/cholesky.jl`

ararslan:
Hey Phillip, welcome and thanks for the contribution! Some things to add here:
* At the bottom of the `issuccess` docstring, something like
  ```
  !!! compat "Julia 1.6"
      `issuccess(::CholeskyPivoted)` requires Julia 1.6 or later.
  ```
* An entry in the LinearAlgebra section of NEWS.md that mentions the addition of this method
* A test for this method in the location Viral pointed out

This looks like a very useful addition!

palday:
Okay, I think I got everything taken care of. Let me know what I missed. :)

ViralBShah:
Thanks @palday and welcome!

andreasnoack:
We discussed this and the possible issues in https://github.com/JuliaLang/julia/issues/35614

ViralBShah:
Anything we should do differently here. Sorry, I wasn't aware of #35614.

andreasnoack:
I think we concluded the definition here was reasonable even though it was argued that the factorization can also be *successful* when `info > 0` and, in fact, that it will always be successful for positive semidefinite matrices. However, it would be good to make these caveats clear in the docs.

ViralBShah:
@palday Would it be possible for you to review #35614 and update the docs accordingly?

palday:
I also gave some thought to the successful rank-deficient PSD case, but in `check=true` you would get a `RankDeficientException`, so the behavior is self-consistent if unfortunately returning false negatives for this case.

palday:
Ha, it looks like that issue is discussing exactly what I discovered. I'll take a look (might be a few days).

PR #34019, opened 2019-12-04
simplify interpreter-stacktraces code

vtjnash:
The unwinder and debuggers thought that our assembly code here was not
quite correct. Rather than attempt to fix that, let the compiler
generate it so we don't need to maintain it anymore.

This was previously also not particularly optimal for an interpreter to
need a couple extra function calls (by indirect pointer too) to setup
the call frame, so now we avoid that.

This simplifies the design by adding a new flag bit to the existing
pgcstack frames. In the future, we may end up generalizing this support
to handle stack allocation of arbitrary objects, but for now we
implement just enough support for our current needs.

It's unclear why dbghelp StackWalk glitches here a couple times (it is
reporting the stack pointer instead of the instruction pointer as the
return address), but this design is robust against that now (even though
I've manually verified that that particular glitch still happens with
this patch).

fix #33877

PR #36006, opened 2020-05-23
Profile: Fix checks of remaining backtrace buffer

### LINKED ISSUE #35870: Condition where backtrace buffer for profiling is full
On Linux, the number of stored instruction pointers will be less than the specified maximum number.
```julia
julia> versioninfo()
Julia Version 1.4.1
Platform Info:
  WORD_SIZE: 64
  LIBM: libopenlibm

julia> using Profile

julia> Profile.init(n = 100)

julia> Profile.clear()

julia> @profile rand(Float64, 100, 100);

julia> length(Profile.fetch())
99
```
In fact, the source code specifies that. ~However, I don't understand the meaning of the `-1`.~ The null sentinels are counted properly.
https://github.com/JuliaLang/julia/blob/f12cde0e8c7f4268838d3201dd9e13b90ea44633/src/signals-unix.c#L700-L706

Although the behavior is safer in terms of memory access, the warning below is not displayed correctly.
https://github.com/JuliaLang/julia/blob/3af96bcefcb0aec559d4652fd75e1bc3fcfe99da/stdlib/Profile/src/Profile.jl#L308-L315

I'm still studying how profiling works, so I am not confident in making a PR for this issue.

kimikage:
Fixes #35870

`jl_unw_stepn` requires a free space of `JL_BT_MAX_ENTRY_SIZE + 1` regardless of the actual entry size. Therefore, the tail of the backtrace buffer is eventually filled with sentinel, i.e. `0`.

It is necessary to export `JL_BT_ MAX_ENTRY_SIZE` to detect the buffer fullness in the Profile package. For this reason, and in order to commonize the code between OSs, this PR adds `jl_profile_is_buffer_full()`.

~**Edit:**
I don't feel that this change is as simple as the problem which this fixes (i.e. the display of a warning).Therefore, hoping that the backtrace handling will be radically improved in the future, I think that it is enough to add a broken test for now.~

### DIFF
--- a/src/signal-handling.c
+++ b/src/signal-handling.c
@@ -17,7 +17,8 @@ extern "C" {
 
 #include <threading.h>
 
-// Profiler control variables //
+// Profiler control variables
+// Note: these "static" variables are also used in "signals-*.c"
 static volatile jl_bt_element_t *bt_data_prof = NULL;
 static volatile size_t bt_size_max = 0;
 static volatile size_t bt_size_cur = 0;
@@ -290,6 +291,12 @@ JL_DLLEXPORT void jl_profile_clear_data(void)
 JL_DLLEXPORT int jl_profile_is_running(void)
 {
     return running;
+}
+
+JL_DLLEXPORT int jl_profile_is_buffer_full(void)
+{
+    // the latter `+ 1` is for the block terminator `0`.
+    return bt_size_cur + (JL_BT_MAX_ENTRY_SIZE + 1) + 1 > bt_size_max;
 }
 
 #ifdef __cplusplus
--- a/src/signals-mach.c
+++ b/src/signals-mach.c
@@ -1,5 +1,7 @@
 // This file is a part of Julia. License is MIT: https://julialang.org/license
 
+// Note that this file is `#include`d by "signals-unix.c"
+
 #include <mach/clock.h>
 #include <mach/clock_types.h>
 #include <mach/clock_reply.h>
@@ -491,8 +493,10 @@ void *mach_profile_listener(void *arg)
         int keymgr_locked = _keymgr_get_and_lock_processwide_ptr_2(KEYMGR_GCC3_DW2_OBJ_LIST, &unused) == 0;
         for (i = jl_n_threads; i-- > 0; ) {
             // if there is no space left, break early
-            if (bt_size_cur >= bt_size_max - 1)
+            if (jl_profile_is_buffer_full()) {
+                jl_profile_stop_timer();
                 break;
+            }
 
             unw_context_t *uc;
             jl_thread_suspend_and_get_state(i, &uc);
--- a/src/signals-unix.c
+++ b/src/signals-unix.c
@@ -1,5 +1,7 @@
 // This file is a part of Julia. License is MIT: https://julialang.org/license
 
+// Note that this file is `#include`d by "signal-handling.c"
+
 #include <signal.h>
 #include <sys/types.h>
 #include <sys/stat.h>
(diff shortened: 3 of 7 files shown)


src/signal-handling.c
```diff
@@ -288,6 +288,11 @@ JL_DLLEXPORT int jl_profile_is_running(void)
     return running;
 }
 
+JL_DLLEXPORT int jl_profile_is_buffer_full(void)
+{
+    return bt_size_cur + JL_BT_MAX_ENTRY_SIZE + 1 >= bt_size_max;
```
kimikage, line 293: I don't know the intent of the `+ 1` (changed in PR #34019). 
https://github.com/JuliaLang/julia/blob/852ff2bdab0b0cac987c0f1ef552ef11202cd2d5/src/stackwalk.c#L94
So based on the current implementation, I added extra `1` for the sentinel. (Note the difference between `>` and `>=`.)

kimikage:
I'll close this as there are other PRs which modify the `Profile`.

src/signal-handling.c
vtjnash: Yes, we now always add the unwind info from the stack walk (+1) to the current trace (n), then also may insert an entry with up to JL_BT_MAX_ENTRY_SIZE additional information.

This PR looks correct to me, modulo needing to be updated. Is there any particular reason you did not think so?

kimikage: I think this PR fix will work, but the intent or purpose of `+1` is not clear. I think one of the causes of the bug which this PR is trying to fix, is this unclear specification.

This is a matter of agreement between the `jl_unw_stepn` and its callers, and, of course, I am not saying #34019 is wrong.

We can reopen this PR, but I think it would be more beneficial to create a new PR than to open a stack walk lecture for me.

kimikage: BTW, I also don't understand the intent of timing control on Windows.
https://github.com/JuliaLang/julia/blob/29a6e66e4944c6d79e1f672ae4648a76599b2f21/src/signals-win.c#L340-L342

It was also weird before PR #37101, but I thought it was best effort, assuming the frequency was peaked in the first place.
PR #37101 seems to be trying to improve it by using `timeBeginPeriod`/`timeEndPeriod`. However, the unit of `nsecprof / GIGA` should be seconds. :confused:

vtjnash: Ah thanks for that find! I knew the logic there had been weird, but didn't notice that the units of `timeout` were wrong also. Should be even better after https://github.com/JuliaLang/julia/pull/38208 in that case. Realistically, this just meant it always used the minimum interval (1-2ms), so hardly noticeable to most people, but worth fixing.

vtjnash: Almost all looks right about that new version, if you want to make it a PR.

Note that this is using `timeout += tc.wPeriodMin` (and not `max(,1)`) since the documentation for the `Sleep` function says it returns up to that many milliseconds early. So you have to add that value to be able to set the actual desired timeout (https://docs.microsoft.com/en-us/windows/win32/sync/wait-functions#wait-functions-and-time-out-intervals)

kimikage: Oh, I'm sorry. I commented in https://github.com/JuliaLang/julia/pull/38208#issuecomment-718008334

vtjnash: Now that the units issue is fixed, do you want to rebase and land that revised version of yours for this issue? Thanks!

kimikage:
I've rebased the commit and am re-opening this PR.
However, I'm not familiar with the C and I still don't understand the current codebase.

PR #36008, opened 2020-05-23
RFC: make `getindex(::AbstractRange, i)` specific to `StepRange`

JeffBezanson:
I'm not an expert in the range code at this point, but it seemed odd to me that `getindex(::AbstractRange, ::Integer)` was defined. Based on what happens if you remove it, it seems this is only needed for `StepRange`; all other ranges (AFAIK) have their own methods. Is that correct or is this needed for some other reason?

It also looked to me like the scope of its boundscheck block needed to be bigger.

### DIFF
--- a/base/compiler/ssair/basicblock.jl
+++ b/base/compiler/ssair/basicblock.jl
@@ -1,8 +1,8 @@
 # This file is a part of Julia. License is MIT: https://julialang.org/license
 
 """
-Like UnitRange{Int}, but can handle the `last` field, being temporarily
-< first (this can happen during compacting)
+Like UnitRange{Int}, but can handle the `last` field being temporarily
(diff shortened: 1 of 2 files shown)


timholy (approved):
Provide support for `Core.Compiler.StmtRange` and this seems like an improvement. The `i > 0` is a dead giveaway that this is wrong for ranges for which indexing doesn't start at 1.

vtjnash:
Rebased, to see if this still works. Already approved, so marked "merge me" also.

DilumAluthge:
Looks like some merge conflicts need to be resolved.

LilithHafner:
The git history was messed up. I reset to master and cherrypicked the necessary commit.

PR #36013, opened 2020-05-24
Faster blockdiag for uniform input value-index types

irhum:
The current blockdiag implementation in SparseArrays uses type promotion to identify what types the values and index should be for the output SparseArray. 

This can be quite slow when the input to blockdiag is many small, sparse arrays (for the sake of example, over 2500+). If all the input sparse arrays have the same value and index types, then it's not necessary to compute them via type promotion and we can directly proceed to creating the new matrix.

This can provide significant speedups; for M generated with
```julia
M = [sprand(rand(15:50), rand(15:50), 0.02) for i in 1:2500]
``` 
and benchmarking the existing implementation with BenchmarkTools

```julia
@benchmark output = blockdiag(M...)
```

we get
```
BenchmarkTools.Trial: 
  memory estimate:  192.35 MiB
  allocs estimate:  37956
  --------------
  minimum time:     1.404 s (1.30% GC)
  median time:      1.434 s (1.31% GC)
  mean time:        1.438 s (1.24% GC)
  maximum time:     1.482 s (0.89% GC)
  --------------
  samples:          4
  evals/sample:     1
```
running the same benchmark, with same M, with the new implementation, we get a speedup of nearly 1000x

```julia
@benchmark output = blockdiag(M...)
```

```BenchmarkTools.Trial: 
  memory estimate:  3.00 MiB
  allocs estimate:  5019
  --------------
  minimum time:     1.072 ms (0.00% GC)
  median time:      1.162 ms (0.00% GC)
  mean time:        1.386 ms (11.86% GC)
  maximum time:     5.059 ms (60.41% GC)
  --------------
  samples:          3591
  evals/sample:     1
```

the current blockdiag implementation is hence split into 3 possible methods
```julia
blockdiag(X::AbstractSparseMatrixCSC...) # for the general case
blockdiag(X::AbstractSparseMatrixCSC{Tv, Ti}...) where {Tv, Ti <: Integer} # for increased speed when all input X have the same index and value types
blockdiag(::Type{Tv}, ::Type{Ti}, X::AbstractSparseMatrixCSC...) where {Tv, Ti <: Integer} # internally called by the two above methods, and available to the user when they're dealing with heterogenous inputs, and wish to specifiy the output index and value types manually for the increased speed benefits

### DIFF
--- a/stdlib/SparseArrays/src/sparsematrix.jl
+++ b/stdlib/SparseArrays/src/sparsematrix.jl
@@ -3318,7 +3318,7 @@ julia> blockdiag(sparse(2I, 3, 3), sparse(4I, 2, 2))
  ⋅  ⋅  ⋅  ⋅  4
 ```
 """
-blockdiag() = spzeros(Union{}, Int, 0, 0)
+blockdiag() = spzeros(promote_type(), Int, 0, 0)
 
 function blockdiag(X::AbstractSparseMatrixCSC{Tv, Ti}...) where {Tv, Ti <: Integer}
     _blockdiag(Tv, Ti, X...)
--- a/stdlib/SparseArrays/src/sparsematrix.jl
+++ b/stdlib/SparseArrays/src/sparsematrix.jl
@@ -3318,6 +3318,8 @@ julia> blockdiag(sparse(2I, 3, 3), sparse(4I, 2, 2))
  ⋅  ⋅  ⋅  ⋅  4
 ```
 """
+blockdiag() = spzeros(0, 0)
+
 function blockdiag(X::AbstractSparseMatrixCSC{Tv, Ti}...) where {Tv, Ti <: Integer}
     blockdiag(Tv, Ti, X...)
 end
--- a/stdlib/SparseArrays/src/sparsematrix.jl
+++ b/stdlib/SparseArrays/src/sparsematrix.jl
@@ -3318,7 +3318,9 @@ julia> blockdiag(sparse(2I, 3, 3), sparse(4I, 2, 2))
  ⋅  ⋅  ⋅  ⋅  4
 ```
 """
-function blockdiag(X::AbstractSparseMatrixCSC{Tv, Ti}...=spzeros(0,0)) where {Tv, Ti <: Integer}
+blockdiag() = spzeros(Union{}, Int, 0, 0)
+
+function blockdiag(X::AbstractSparseMatrixCSC{Tv, Ti}...) where {Tv, Ti <: Integer}
     _blockdiag(Tv, Ti, X...)
 end
 
--- a/stdlib/SparseArrays/src/sparsematrix.jl
+++ b/stdlib/SparseArrays/src/sparsematrix.jl
@@ -3321,13 +3321,13 @@ julia> blockdiag(sparse(2I, 3, 3), sparse(4I, 2, 2))
 blockdiag() = spzeros(0, 0)
 
 function blockdiag(X::AbstractSparseMatrixCSC{Tv, Ti}...) where {Tv, Ti <: Integer}
-    blockdiag(Tv, Ti, X...)
+    _blockdiag(Tv, Ti, X...)
 end
 
 function blockdiag(X::AbstractSparseMatrixCSC...)
     Tv = promote_type(map(x->eltype(nonzeros(x)), X)...)
     Ti = isempty(X) ? Int : promote_type(map(x->eltype(rowvals(x)), X)...)
-    blockdiag(Tv, Ti, X...)
+    _blockdiag(Tv, Ti, X...)
 end
 
 function _blockdiag(::Type{Tv}, ::Type{Ti}, X::AbstractSparseMatrixCSC...) where {Tv, Ti <: Integer}
--- a/stdlib/SparseArrays/src/sparsematrix.jl
+++ b/stdlib/SparseArrays/src/sparsematrix.jl
@@ -3318,15 +3318,22 @@ julia> blockdiag(sparse(2I, 3, 3), sparse(4I, 2, 2))
  ⋅  ⋅  ⋅  ⋅  4
 ```
 """
+function blockdiag(X::AbstractSparseMatrixCSC{Tv, Ti}...) where {Tv, Ti <: Integer}
+    blockdiag(Tv, Ti, X...)
+end
+
 function blockdiag(X::AbstractSparseMatrixCSC...)
+    Tv = promote_type(map(x->eltype(nonzeros(x)), X)...)
+    Ti = isempty(X) ? Int : promote_type(map(x->eltype(rowvals(x)), X)...)
+    blockdiag(Tv, Ti, X...)
+end
+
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by irhum** [comment]
added another method to deal with the empty input case, `blockdiag()`, otherwise it defaults to 
```julia
blockdiag(X::AbstractSparseMatrixCSC{Tv, Ti}...) where {Tv, Ti <: Integer}
```

instead of 
```julia
blockdiag(X::AbstractSparseMatrixCSC...)
```

resulting in a `Tv not defined` error, making an explicit definition necessary. 

The pull request, I would suggest is important given the case where all inputs have the same types for the value and index would be more common (and would benefit from increased performance) than the heterogenous case

stdlib/SparseArrays/src/sparsematrix.jl
```diff
@@ -3318,16 +3318,25 @@ julia> blockdiag(sparse(2I, 3, 3), sparse(4I, 2, 2))
...
+    blockdiag(Tv, Ti, X...)
+end
+
 function blockdiag(X::AbstractSparseMatrixCSC...)
+    Tv = promote_type(map(x->eltype(nonzeros(x)), X)...)
+    Ti = isempty(X) ? Int : promote_type(map(x->eltype(rowvals(x)), X)...)
+    blockdiag(Tv, Ti, X...)
+end
+
+function blockdiag(::Type{Tv}, ::Type{Ti}, X::AbstractSparseMatrixCSC...) where {Tv, Ti <: Integer}
```
dkarrasch, line 3333: I wonder whether we should make this function "internal", by calling it
```suggestion
function _blockdiag(::Type{Tv}, ::Type{Ti}, X::AbstractSparseMatrixCSC...) where {Tv, Ti <: Integer}
```
Otherwise, this kind of competes with the `convert` machinery, where you set the output eltype. `blockdiag` is exported, and hence this method became public.

dkarrasch (commented):
Very nice contribution, @irhum, and welcome! I have one comment that I'm curious what other people think about.

stdlib/SparseArrays/src/sparsematrix.jl
irhum: I would agree, there isn't much reason for this method to remain exposed

dkarrasch:
You'll need to rename the function at the call sites as well.

stdlib/SparseArrays/src/sparsematrix.jl
```diff
@@ -3318,16 +3318,25 @@ julia> blockdiag(sparse(2I, 3, 3), sparse(4I, 2, 2))
...
+
+function blockdiag(X::AbstractSparseMatrixCSC{Tv, Ti}...) where {Tv, Ti <: Integer}
+    _blockdiag(Tv, Ti, X...)
+end
+
 function blockdiag(X::AbstractSparseMatrixCSC...)
+    Tv = promote_type(map(x->eltype(nonzeros(x)), X)...)
+    Ti = isempty(X) ? Int : promote_type(map(x->eltype(rowvals(x)), X)...)
+    _blockdiag(Tv, Ti, X...)
+end
```
irhum, line 3331: Was wondering whether to place the no input value case as an optional argument would be cleaner? 

```suggestion
function blockdiag(X::AbstractSparseMatrixCSC{Tv, Ti}...=spzeros(0, 0)) where {Tv, Ti <: Integer}
    _blockdiag(Tv, Ti, X...)
end

function blockdiag(X::AbstractSparseMatrixCSC...)
    Tv = promote_type(map(x->eltype(nonzeros(x)), X)...)
    Ti = isempty(X) ? Int : promote_type(map(x->eltype(rowvals(x)), X)...)
    _blockdiag(Tv, Ti, X...)
end
```

irhum:
The function call sites have been changed to make the appropriate `_blockdiag` call.

Since we also need to account for the case where no inputs are given, a clean solution seems to be to pass it as an optional argument to one of the methods. Since we're explicitly handing it, `isempty(X)` in

```julia
Ti = isempty(X) ? Int : promote_type(map(x->eltype(rowvals(x)), X)...)
```

is always false, and hence can be replaced with

```
Ti = promote_type(map(x->eltype(rowvals(x)), X)...)
```

Opinions?

stdlib/SparseArrays/src/sparsematrix.jl
```diff
@@ -3318,16 +3318,23 @@ julia> blockdiag(sparse(2I, 3, 3), sparse(4I, 2, 2))
  ⋅  ⋅  ⋅  ⋅  4
 ```
 """
+function blockdiag(X::AbstractSparseMatrixCSC{Tv, Ti}...=spzeros(0,0)) where {Tv, Ti <: Integer}
```
dkarrasch, line 3321: This default value makes the function change behavior for empty arguments. Currently, we have
```julia
julia> A = blockdiag()
0×0 SparseMatrixCSC{Union{},Int64} with 0 stored entries

julia> B = spzeros(0,0)
0×0 SparseMatrixCSC{Float64,Int64} with 0 stored entries
```
Changing that behavior may be for the better, but I'm not sure. This definitely requires a broader discussion, I'd say.

dkarrasch (commented):
Looks good in principle, but there is this subtle issue with empty arguments.

stdlib/SparseArrays/src/sparsematrix.jl
irhum: I would rather not change the behavior for this pull request anyway; the goal here is to provide increased speed when dealing with CSC arrays with the same types for the index and values.

Created a new commit that fixes this (the output is now `SparseMatrixCSC{Union{},Int64}`) by explicitly spelling out the blockdiag() case, to make it clear what it does today.

stdlib/SparseArrays/src/sparsematrix.jl
```diff
@@ -3318,16 +3318,25 @@ julia> blockdiag(sparse(2I, 3, 3), sparse(4I, 2, 2))
  ⋅  ⋅  ⋅  ⋅  4
 ```
 """
+blockdiag() = spzeros(Union{}, Int, 0, 0)
```
dkarrasch, line 3321: I'd suggest to have
```suggestion
blockdiag() = spzeros(promote_type(), Int, 0, 0)
```
to avoid hardcoding some "non-sense" type and let the ecosystem decide. Pinging @tkf for this nasty "reduction over empty tuples" issue.

irhum: Thanks for pointing this out, it is more sensible this way. I believe that resolves the issue, of speeding up this function while still making sure blockdiag() output remains unchanged. Would appreciate what @tkf has to say about the reduction on empty tuple issue

tkf: I think `Union{}` makes sense (i.e., it's the identity element of `promote_type` as a monoid).  Though `promote_type()` sounds also fine.  It's just a LISPy way of saying that we want the identity of `promote_type`.

dkarrasch: Thanks, @tkf. I just had no idea what might be subject to potential future design changes and what is "set in stone". I'll go ahead and merge this as is then.

PR #36018, opened 2020-05-24
Use the TOML stdlib in code loading

### LINKED ISSUE #27414: Code loading might be better just fully parsing TOML files
Right now, code loading uses some regexes that can parse the Project files in the format that Pkg writes them. 
This has a few negative side effects:

* We do not actually support TOML file, only the subset that the specialized regexes manage to parse.
* There is no way to cache any results, instead, for everything getting loaded we start parsing from scratch.
* Giving good error messages seems hard. After we found that for some reason we cannot load a package we cannot look up things that we have already parsed, since we have thrown away all the parsed information.
* The code is kinda hard to read when parsing of TOML files is blended together with the code loading logic. There are currently ~6 functions that parse TOML files in different ways depending on what step in code loading we are doing.

I would argue that just parsing the full TOML files, cache the results and use that in code loading would lead to faster and more simple code. The drawback is that we need a full TOML parser ni Base, but since we have decided to use TOML for code loading that seems inevitable, should have called it some other file extension if we didn't intend to support TOML fully.

KristofferC:
The current way TOML files are parsed in Base during code loading has a few issues:
- The TOML "parser" in Base does pattern matching to try to find exactly the information code loading in its current state needs. This makes it a bit hard to extend since you need to write a new parsing routine for every new feature. We might want to try add some new features (e.g. sub-projects) that would require some changes in code loading which the current parser causes a bit of a hindrance to.
- The TOML parser in Base is "stateless" so every query it makes it starts to parse the file from scratch. This means that in some cases it can re-parse the file many times (https://github.com/JuliaLang/julia/issues/27414#issuecomment-555037342) unnecessarily.
- The TOML parser used by code loading and Pkg is different which means that they support different parts of the spec (AFAIU, the Base parser doesn't support inline tables for example).

Therefore, it would make sense to try:
- Make Base use a proper TOML parser that implements the spec.
- Make Pkg and Base share this TOML parser implementation.
- Eventually make the TOML parser a fully supported stdlib. TOML files are a big part of packages in Julia and it makes sense to me to have an easy option to parse these since they are needed for a lot of the ecosystem.

This PR starts this process by putting a newly written TOML parser into Base and re-implements code loading to take advantage of it. The parser is available as a package with proper API, docs, tests, etc (https://github.com/KristofferC/TOMLX.jl) but the only thing this PR adds is the parser part (with a more low-level API) since that is what is needed in Base and I wanted to make this PR have as little "noise" as possible.
It has a significant number of tests (https://github.com/KristofferC/TOMLX.jl/blob/master/test/readme.jl, https://github.com/KristofferC/TOMLX.jl/blob/master/test/toml_test.jl) and I have benchmarked it pretty thoroughly.

For this PR, I think it makes sense to mostly comment on the code loading parts (@StefanKarpinski for review). For the TOML parsing itself, perhaps it makes sense to focus that discussion to the https://github.com/KristofferC/TOMLX.jl/ repo?.

After this step is done, we update Pkg to start using the parser here for it's parsing.
Then we can try to finally make a TOML stdlib and have Pkg depend on that. This will remove the last bundled dependency for Pkg.

There is a TODO in this PR and that is that I right now use a global cache object to store the `filename => Dict` mapping in code loading which gets empited in `Base.require`. This should be moved to a function local cache that gets created in `Base.require` and threaded through all functions. I wanted to put up this PR anyway for people to get a chance to review things as soon as possible.

---------------------------------------

One possible question is why this parser is not just the one that currently lives in the `ext` folder in Pkg. The reason for this is that the `ext`- parser (forked from https://github.com/JuliaLang/TOML.jl)  was written a quite long time ago for an old spec of TOML and requires some significant updates. I was also a bit unhappy with the performance of it (it allocates scratch buffers a bit too much and makes copies of strings etc) so I felt a clean implementation might lead to a better end result.

---------------------

Fixes https://github.com/JuliaLang/julia/issues/27414#.

### DIFF
--- a/base/loading.jl
+++ b/base/loading.jl
@@ -152,68 +152,39 @@ function version_slug(uuid::UUID, sha1::SHA1, p::Int=5)
     return slug(crc, p)
 end
 
+struct TOMLCache
+    p::TOML.Parser
+    d::Dict{String, Dict{String, Any}}
+end
+TOMLCache() = TOMLCache(TOML.Parser(), Dict{String, Dict{String, Any}}())
+
+function parsed_toml(cache::TOMLCache, project_file::String)
+    get!(cache.d, project_file) do
+        TOML.reinit!(cache.p, read(project_file, String); filepath=project_file)
+        TOML.parse(cache.p)
+    end
+end
+
 ## package identification: determine unique identity of package to be loaded ##
 
-function find_package(args...)
-    pkg = identify_package(args...)
+# Used by Pkg but not used in loading itself
+function find_package(arg)
+    cache = TOMLCache()
+    pkg = identify_package(arg, cache)
     pkg === nothing && return nothing
-    return locate_package(pkg)
-end
-
-struct PkgId
-    uuid::Union{UUID,Nothing}
-    name::String
-
-    PkgId(u::UUID, name::AbstractString) = new(UInt128(u) == 0 ? nothing : u, name)
-    PkgId(::Nothing, name::AbstractString) = new(nothing, name)
-end
-PkgId(name::AbstractString) = PkgId(nothing, name)
-
-function PkgId(m::Module, name::String = String(nameof(moduleroot(m))))
-    uuid = UUID(ccall(:jl_module_uuid, NTuple{2, UInt64}, (Any,), m))
-    UInt128(uuid) == 0 ? PkgId(name) : PkgId(uuid, name)
-end
-
-==(a::PkgId, b::PkgId) = a.uuid == b.uuid && a.name == b.name
-
-function hash(pkg::PkgId, h::UInt)
-    h += 0xc9f248583a0ca36c % UInt
-    h = hash(pkg.uuid, h)
-    h = hash(pkg.name, h)
-    return h
-end
-
-show(io::IO, pkg::PkgId) =
-    print(io, pkg.name, " [", pkg.uuid === nothing ? "top-level" : pkg.uuid, "]")
-
-function binpack(pkg::PkgId)
-    io = IOBuffer()
-    write(io, UInt8(0))
-    uuid = pkg.uuid
-    write(io, uuid === nothing ? UInt128(0) : UInt128(uuid))
-    write(io, pkg.name)
-    return String(take!(io))
-end
-
-function binunpack(s::String)
-    io = IOBuffer(s)
-    @assert read(io, UInt8) === 0x00
-    uuid = read(io, UInt128)
-    name = read(io, String)
-    return PkgId(UUID(uuid), name)
+    return locate_package(pkg, cache)
 end
 
 ## package identity: given a package name and a context, try to return its identity ##
-
-identify_package(where::Module, name::String) = identify_package(PkgId(where), name)
+identify_package(where::Module, name::String, cache::TOMLCache = TOMLCache()) = identify_package(PkgId(where), name, cache)
 
 # identify_package computes the PkgId for `name` from the context of `where`
 # or return `nothing` if no mapping exists for it
-function identify_package(where::PkgId, name::String)::Union{Nothing,PkgId}
+function identify_package(where::PkgId, name::String, cache::TOMLCache=TOMLCache())::Union{Nothing,PkgId}
     where.name === name && return where
-    where.uuid === nothing && return identify_package(name) # ignore `where`
+    where.uuid === nothing && return identify_package(name, cache) # ignore `where`
     for env in load_path()
-        uuid = manifest_deps_get(env, where, name)
+        uuid = manifest_deps_get(env, where, name, cache)
         uuid === nothing && continue # not found--keep looking
         uuid.uuid === nothing || return uuid # found in explicit environment--use it
         return nothing # found in implicit environment--return "not found"
@@ -223,47 +194,33 @@ end
 
 # identify_package computes the PkgId for `name` from toplevel context
 # by looking through the Project.toml files and directories
-function identify_package(name::String)::Union{Nothing,PkgId}
+function identify_package(name::String, cache::TOMLCache=TOMLCache())::Union{Nothing,PkgId}
     for env in load_path()
-        uuid = project_deps_get(env, name)
+        uuid = project_deps_get(env, name, cache)
         uuid === nothing || return uuid # found--return it
     end
     return nothing
 end
 
-function identify_package(name::String, names::String...)
-    pkg = identify_package(name)
-    pkg === nothing && return nothing
-    return identify_package(pkg, names...)
-end
-
-# locate `tail(names)` package by following the search path graph through `names` starting from `where`
-function identify_package(where::PkgId, name::String, names::String...)
-    pkg = identify_package(where, name)
-    pkg === nothing && return nothing
-    return identify_package(pkg, names...)
-end
-
 ## package location: given a package identity, find file to load ##
-
-function locate_package(pkg::PkgId)::Union{Nothing,String}
+function locate_package(pkg::PkgId, cache::TOMLCache=TOMLCache())::Union{Nothing,String}
     if pkg.uuid === nothing
         for env in load_path()
             # look for the toplevel pkg `pkg.name` in this entry
-            found = project_deps_get(env, pkg.name)
+            found = project_deps_get(env, pkg.name, cache)
             found === nothing && continue
             if pkg == found
                 # pkg.name is present in this directory or project file,
                 # return the path the entry point for the code, if it could be found
                 # otherwise, signal failure
-                return implicit_manifest_uuid_path(env, pkg)
+                return implicit_manifest_uuid_path(env, pkg, cache)
             end
             @assert found.uuid !== nothing
-            return locate_package(found) # restart search now that we know the uuid for pkg
+            return locate_package(found, cache) # restart search now that we know the uuid for pkg
         end
     else
         for env in load_path()
-            path = manifest_uuid_path(env, pkg)
+            path = manifest_uuid_path(env, pkg, cache)
             path === nothing || return entry_path(path, pkg.name)
         end
     end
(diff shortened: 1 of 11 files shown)


timholy:
Does this make Pkg operations faster? Thinking of https://github.com/JuliaLang/julia/issues/27414#issuecomment-555037342

KristofferC:
> Does this make Pkg operations faster? 

If you mean loading packages, then I don't think it was bottlenecked on reading the TOML files, even though it did it quite inefficiently.
If you mean stuff like `Pkg.add` then TOML parsing is significant but not crucial for performance (anymore). I've already fixed the biggest performance problems with the TOML parser in Pkg (https://github.com/JuliaLang/Pkg.jl/pull/113, https://github.com/JuliaLang/Pkg.jl/pull/116, https://github.com/JuliaLang/Pkg.jl/pull/1751) so while this parser is significantly faster there are some diminishing returns in effect.

StefanKarpinski:
Any measurements on how this affects code loading speed? One thought I’d had for that was to specify what values you’d like to extract from a file and only return that data instead of constructing the whole data structure. So if you’re just looking for a single value, it could be done with no allocation at all.

KristofferC:
> One thought I’d had for that was to specify what values you’d like to extract from a file and only return that data instead of constructing the whole data structure. So if you’re just looking for a single value, it could be done with no allocation at all.

Yeah, it is possible, but the time to fully parse my pretty big v1.4 Manifest.toml file is ~300 us (including reading it from disc). I don't think this really needs to be optimized further.

StefanKarpinski:
I was thinking about the issue of parsing the same file over and over. But maybe caching is better for that.

KristofferC:
> I was thinking about the issue of parsing the same file over and over. 

I think it is good to completely avoid opening and reading through the file which a cache should help with. Also, it is simpler in the sense that you don't need a second version of the parser that does something more akin lazy parsing.

base/toml_parser.jl
```diff
@@ -0,0 +1,1099 @@
...
+
+    # We need to keep track of those tables / arrays that are defined
+    # inline since we are not allowed to add keys to those
+    inline_tables::IdSet{TOMLDict}
+    static_arrays::IdSet{TOMLArray}
+
+    # [a.b.c.d] doesn't "define" the table [a]
+    # so keys can later be added to [a], therefore
+    # we need to keep track of what tables are
+    # actualyl "defined
```
vtjnash, line 80:
```suggestion
    # actually defined
```

base/toml_parser.jl
```diff
@@ -0,0 +1,1099 @@
...
+# all the way and have this error be returned to the user
+# if the parse is called with `raise=false`. This macro
+# makes that easier
+@eval macro $(Symbol("try"))(expr)
+    :(
+        v = $(esc(expr));
+        v isa $ParserError && return v;
+        v;
+    )
+end
```
vtjnash, line 162:
```suggestion
@eval macro $(:var"try")(expr)
    return quote
        v = $(esc(expr))
        v isa ParserError && return v
        v
    end
end
```

base/toml_parser.jl
```diff
@@ -0,0 +1,1099 @@
...
+        if !(peek(l) == '\n' || peek(l) == '\r' || peek(l) == EOF_CHAR)
+            return ParserError(ErrExpectedNewLineKeyValue)
+        end
+    end
+end
+
+function recurse_dict!(l::Parser, d::Dict, dotted_keys::AbstractVector{String}, check=true)::Err{TOMLDict}
+    for i in 1:length(dotted_keys)
+        key = dotted_keys[i]
+        d = get!(() -> TOMLDict(), d, key)
```
vtjnash, line 470:
```suggestion
        d = get!(TOMLDict, d, key)
```

base/toml_parser.jl
```diff
@@ -0,0 +1,1099 @@
...
+function parse_bool(l::Parser, v::Bool)::Union{Bool, Nothing}
+    # Have eaten a 't' if `v` is true, otherwise have eaten a `f`.
+    v ? (accept(l, 'r') && accept(l, 'u') && accept(l, 'e') && return true) :
+        (accept(l, 'a') && accept(l, 'l') && accept(l, 's') && accept(l, 'e') && return false)
+    return nothing
+end
+
+isvalid_hex(c::Char) = isdigit(c) || ('a' <= c <= 'f') || ('A' <= c <= 'F')
+isvalid_oct(c::Char) = '0' <= c <= '7'
+isvalid_binary(c::Char) = '0' <= c <= '2'
```
vtjnash, line 702:
```suggestion
isvalid_binary(c::Char) = '0' <= c <= '1'
```

base/toml_parser.jl
```diff
@@ -0,0 +1,1099 @@
...
+    read_underscore |= contains_underscore
+    if accept(l, x -> x == 'e' || x == 'E')
+        accept(l, x-> x == '+' || x == '-')
+        # SPEC: (which follows the same rules as decimal integer values but may include leading zeros)
+        read_digit = accept_batch(l, isdigit)
+        ate, read_underscore = @try accept_batch_underscore(l, isdigit, !read_digit)
+        contains_underscore |= read_underscore
+    end
+    if !ok_end_value(peek(l))
+        error()
```
vtjnash, line 818: nit: message is required

base/loading.jl
```diff
@@ -490,155 +435,128 @@ end
...
-            elseif state === :deps
-                if (m = match(re_key_to_string, line)) !== nothing
-                    m.captures[1] == name && return UUID(m.captures[2])
-                end
-            end
-        end
-        return root_name == name ? root_uuid : nothing
+function explicit_project_deps_get(project_file::String, name::String, cache::TOMLCache)::Union{Nothing,UUID}
+    root_name = nothing
+    root_uuid = dummy_uuid(project_file)
```
vtjnash, line 440:
```suggestion
```

base/loading.jl
```diff
@@ -490,155 +435,128 @@ end
...
-            elseif (m = match(re_path_to_string, line)) !== nothing
-                path = String(m.captures[1])
-            elseif (m = match(re_hash_to_string, line)) !== nothing
-                hash = SHA1(m.captures[1])
+
+    d = parsed_toml(cache, manifest_file)
+    entries = get(d, pkg.name, nothing)
+    entries === nothing && return nothing # TODO: allow name to mismatch?
+    for entry in entries
+        if UUID(get(entry, "uuid", nothing)) === pkg.uuid
```
vtjnash, line 505: `ERROR: MethodError: Cannot `convert` an object of type Nothing to an object of type UInt128`

base/loading.jl
```diff
@@ -490,155 +435,128 @@ end
...
             end
-        end
-        uuid == pkg.uuid || return nothing
-        name == pkg.name || return nothing # TODO: allow a mismatch?
-        if path !== nothing
-            path = normpath(abspath(dirname(manifest_file), path))
-            return path
-        end
-        hash === nothing && return nothing
-        # Keep the 4 since it used to be the default
```
vtjnash, line 594: restore deleted functionality and comment

base/loading.jl
```diff
@@ -490,155 +435,128 @@ end
...
+    found_where || return nothing
+    found_name || return PkgId(name)
+    # Only here is deps was not a dict which mean we have a unique name for the dep
+    if !haskey(d, name) || length(d[name]) != 1
+        error("expected a single entry for $(repr(name)) in $(repr(project_file))")
+    end
+    entry = first(d[name])
+    if found_name
+        uuid = get(entry, "uuid", nothing)
+        return PkgId(UUID(uuid), name)
```
vtjnash, line 490: `ERROR: MethodError: Cannot `convert` an object of type Nothing to an object of type UInt128`

base/loading.jl
```diff
@@ -490,155 +435,128 @@ end
...
-    found_or_uuid && return PkgId(name)
-    return nothing
+    found_where || return nothing
+    found_name || return PkgId(name)
+    # Only here is deps was not a dict which mean we have a unique name for the dep
+    if !haskey(d, name) || length(d[name]) != 1
+        error("expected a single entry for $(repr(name)) in $(repr(project_file))")
+    end
+    entry = first(d[name])
+    if found_name
```
vtjnash, line 488: always true here?

base/loading.jl
```diff
@@ -490,155 +435,128 @@ end
...
+                # deps is either a list of names (deps = ["DepA", "DepB"]) or
+                # a table of entries (deps = {"DepA" = "6ea...", "DepB" = "55d..."}
+                deps = get(entry, "deps", nothing)
+                if deps isa Vector
+                    found_name = name in deps
+                    break
+                elseif deps isa Dict
+                    for (dep, uuid) in deps
+                        if dep === name
+                            return PkgId(UUID(uuid), name)
```
vtjnash, line 474: `uuid` might not be of the proper type if the file is malformed

base/loading.jl
```diff
@@ -490,155 +435,128 @@ end
...
-        seekstart(io) # rewind IO handle
-        # finally, find out the `uuid` associated with `name`
-        return something(manifest_file_name_uuid(io, name), false)
     end
-    found_or_uuid isa UUID && return PkgId(found_or_uuid, name)
-    found_or_uuid && return PkgId(name)
-    return nothing
+    found_where || return nothing
+    found_name || return PkgId(name)
+    # Only here is deps was not a dict which mean we have a unique name for the dep
```
vtjnash, line 483:
```suggestion
    # Only here if deps was not a dict which mean we have a unique name for the dep
```

base/toml_parser.jl
```diff
@@ -0,0 +1,1099 @@
...
+        minute::Int
+        second::Int
+        ms::Int
+    end
+    struct DateTime
+        date::Date
+        time::Time
+    end
+    DateTime(y, m, d, h, mi, s, ms) =
+        DateTime(Date(y,m,d), Time(h, mi, s, ms))
```
vtjnash, line 29: Would be so much better if we could either use delayed definitions here (Requires.jl?) or make these a parser error (unsupported) rather than re-implementing incompatible structs with the same names.

vtjnash:
Added the docs and tests labels since the new toml_parser.jl file will need both. Otherwise, most looks pretty straight-forward here and should be in pretty good shape to merge.

base/toml_parser.jl
```diff
@@ -0,0 +1,1099 @@
...
+    end
+    return parse_float(l, read_underscore)
+end
+
+
+function take_string_or_substring(l, contains_underscore)::Union{String, SubString}
+    subs = take_substring(l)
+    # Need to pass a AbstractString to `parse` so materialize it in case it
+    # contains underscore.
+    #                                   vvvvvvv <- this looksl like a dude flipping the bird
```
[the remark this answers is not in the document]
KristofferC: :(

base/loading.jl
```diff
@@ -151,12 +151,26 @@ function version_slug(uuid::UUID, sha1::SHA1, p::Int=5)
     return slug(crc, p)
 end
 
+struct TOMLCache
+    p::TOML.Parser
+    d::Dict{String, Dict{String, Any}}
+end
+TOMLCache() = TOMLCache(TOML.Parser(), Dict())
```
rfourquet, line 158: I guess it doesn't matter, but maybe: 
```suggestion
TOMLCache() = TOMLCache(TOML.Parser(), Dict{String,Dict{String,Any}}())
```
?

base/toml_parser.jl
KristofferC: We can't make it a parser error (we need to support arbitrary TOML files) and we want code loading to work without Dates. So this was the best I could come up with.

base/toml_parser.jl
```diff
@@ -0,0 +1,1099 @@
...
+            accept_batch(l, multiline ? stop_candidates_multiline_quoted : stop_candidates_singleline_quoted)
+        else
+            accept_batch(l, multiline ? stop_candidates_multiline : stop_candidates_singleline)
+        end
+        if !multiline && peek(l) == '\n'
+            return ParserError(ErrNewLineInString)
+        end
+        next_slash = peek(l) == '\\'
+        if !next_slash
+            # TODO: This is not true, could be """"foo""""
```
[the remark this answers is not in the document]
KristofferC: Yeah, it currently fails to parse

```
a = """""foo"""""
```

base/toml_parser.jl
```diff
@@ -0,0 +1,1099 @@
...
+    day = parse_int(l, false)
+    # Verify the real range in the constructor below
+    day in 1:31 || return ParserError(ErrParsingDateTime)
+
+    # We might have a local date now
+    read_space = false
+    if ok_end_value(peek(l))
+        if (read_space = accept(l, ' '))
+            if !isdigit(peek(l))
+                return Date(year, month, day)
```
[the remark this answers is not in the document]
KristofferC: Should me a try catch here as well.

base/toml_parser.jl
```diff
@@ -0,0 +1,1099 @@
...
+    @try accept_two(l, isdigit)
+    month = parse_int(l, false)
+    month in 1:12 || return ParserError(ErrParsingDateTime)
+    accept(l, '-') || return ParserError(ErrParsingDateTime)
+
+    # Day
+    set_marker!(l)
+    @try accept_two(l, isdigit)
+    day = parse_int(l, false)
+    # Verify the real range in the constructor below
```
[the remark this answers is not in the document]
KristofferC: it is true when you have the Dates stdlib available:

```
julia> Date(1989, 02, 29)
ERROR: ArgumentError: Day: 29 out of range (1:28)
```

timholy (commented):
In the long run, we'll probably want type annotations on a lot of the non-inferrable operations (which tend to happen a lot if you're working with `Dict{String,Any}`): invalidating package-loading code has a particularly dramatic effect on load time, since package `n` might invalidate the code needed to load package `n+1`. As a consequence, eventually this code probably has to look a bit more like the code in the compiler & JuliaInterpreter in terms of type-checks and branches.

If you don't want to have to add these by hand everywhere, one thought would be to provide an "dict-get interface" where certain keys are, via constant-propagation, able to return objects of known type. A demo of that kind of approach is #36323, which was used for `struct`s but I see little reason you couldn't do the same thing for `Dict`s (not sure if the `Symbol` vs `String` thing will make a difference for constant-prop, would be worth testing). That is, you could define `gettyped(d, key, default)` and depending on the value of `key` different types would be asserted for the return value. Then you wouldn't have to assert them at the caller.

I've made a few specific hand-annotations in various places, but stopped at a certain point.

KristofferC:
> If you don't want to have to add these by hand everywhere, one thought would be to provide an "dict-get interface" where certain keys are, via constant-propagation, able to return objects of known type. 

I can try that

timholy:
:+1: Could also be done later, though you may have the best sense of anyone of what the actual types are.

KristofferC:
Could you explain how you check these? Presumably, a `@snoopr` when loading some package that invalidates loading?

timholy:
Documented here: https://timholy.github.io/SnoopCompile.jl/stable/snoopr/. I see some errors crept in that I failed to notice, so to clarify you can minimally get `@snoopr` with SnoopCompileCore but `invalidation_trees` and everything else requires `SnoopCompile`. So the most careful approach is to load SnoopCompileCore, collect the data on invalidations, then load SnoopCompile and interpret it.

Everything else should work as stated. Let me know if you have any trouble. It takes a little while to get the hang of it, but once you're there it's pretty easy to zip around the codebase :stuck_out_tongue:.

KristofferC:
> Documented here: https://timholy.github.io/SnoopCompile.jl/stable/snoopr/. 

Yeah, I know kind of how you use it but I don't know what to load to see invalidations in loading specifically. I tried loading CSV but get nothing that points to `loading.jl` from that.

timholy:
Oh, I see. The most basic approach would be
```julia
mis = uinvalidated(invlist);
filter(mi->endswith(String(mi.def.file), "loading.jl"), mis)
```
without even building the `trees`. If you see MethodInstances in that list, try `findcaller` followed by `ascend` to reconstruct the entire call chain.

KristofferC:
It seems quite a few of them are from:

```
 >   convert(::Type{Nothing}, ::Any)
       setindex!(::IdDict{Array{Any,1},Nothing}, ::Any, ::Any)
         push!(::Base.IdSet{Array{Any,1}}, ::Array{Any,1})
           parse_array(::Base.TOML.Parser)
               ...
```

Any idea of what to do about that?

timholy:
CategoricalArrays is an untackled and huge problem. I'd probably suggest deleting that method from CategoricalArrays, it seems basically like type-piracy:
```julia
julia> @which convert(Union{Nothing,Float64}, 0.25)
convert(::Type{T}, x::T) where T>:Nothing in Base at some.jl:35
```

timholy:
You're already a pro :+1:

KristofferC:
The last commit might be interesting. There is apparently a difference between adding a type assert to a variable within the loop header vs just type asserting it in the body where inside the loop header it also inserts a `convert`. I guess it is the same as when you add a type assert to a function.

KristofferC:
> CategoricalArrays is an untackled and huge problem. 

Yeah, it even invalidates `convert(Type{Any}, x)`... what the heck is going on there.

timholy:
Thanks for pointing out the better alternative. It seems to be identical to the difference between
```julia
x::T = y
```
and
```julia
x = y::T
```

I'll switch to the style you demonstrated for future commits.

KristofferC:
I've rebased this and changed the target branch to https://github.com/JuliaLang/julia/tree/kc/load_toml for a simpler review.

KristofferC:
@timholy If you want to do some invalidation sanity checks, you could use this branch together with https://github.com/JuliaLang/Pkg.jl/pull/1984 and see how things look.

base/loading.jl
```diff
@@ -319,136 +275,82 @@ function env_project_file(env::String)::Union{Bool,String}
...
-            manifest_file = joinpath(dir, mfst)
-            isfile_casesensitive(manifest_file) && return manifest_file
-        end
-        return nothing
+function project_file_manifest_path(project_file::String, cache::TOMLCache)::Union{Nothing,String}
+    dir = abspath(dirname(project_file))
+    d = parsed_toml(cache, project_file)
+    explicit_manifest = get(d, "manifest", nothing)
+    if explicit_manifest !== nothing
+        manifest_file = normpath(joinpath(dir, explicit_manifest))
```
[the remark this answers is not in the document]
KristofferC: Doesn't inference handle that based on the type annotation above and the `!==` check?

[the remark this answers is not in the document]
KristofferC: I think so:

```
─ %12 = Base.joinpath(dir, explicit_manifest::String)::String
```

KristofferC:
This is ready to go but it breaks Revise and I don't want to break people's workflow so I'll wait a bit with merging. It should be a fairly small change to https://github.com/timholy/Revise.jl/blob/8336b10a5e7d0e4b1bb0726cf426469c8f866d2e/src/pkgs.jl#L459-L488 @timholy. I didn't really know what the function does so I had a hard time updating it myself.

KristofferC:
Revise has been updated to deal with this PR now.

PR #36020, opened 2020-05-25
Add detailed machine-file format from getting-started.md

xgdgsc:
It took me a while to find the file format when reading distributed doc. So copying that from getting-started.md might be helpful.


StefanKarpinski (approved):
Thanks for this! Can't believe we didn't have this documented ~anywhere~ here.

PR #36024, opened 2020-05-25
no longer color the exception-provided error message red

### LINKED ISSUE #36015: Remove enforcal of printing exceptions as red
The code here

https://github.com/JuliaLang/julia/blob/0413ef0e4de83b41b637ba02cc63314da45fe56b/base/errorshow.jl#L87-L95

tries its best to ensure that the `showerror` code an exception implement gets printed in red. This can mess up custom coloring that the exception itself tries to do in `showerror`. For example, here is some code that defines an exception text using colors:

```
julia> struct CustomException end

julia> function Base.showerror(io::IO, ce::CustomException)
           printstyled(io, "This is in red\n"; color=:light_red)
           println(io, "This is in white")
       end
```

Trying with the normal two arg `showerror` and things look ok. But when the error would actually be shown in REPL, the second line would have "magically" also turned red, ruining the formatting.

bild

If an exception wants to print its text in red, it can just implement that with `printstyled`.

KristofferC:
This doesn't compose well with exceptions trying to provide colors in their error messages.

Fixes https://github.com/JuliaLang/julia/issues/36015

### DIFF
--- a/base/errorshow.jl
+++ b/base/errorshow.jl
@@ -86,9 +86,7 @@ end
 
 function showerror(io::IO, ex, bt; backtrace=true)
     try
-        with_output_color(get(io, :color, false) ? error_color() : :nothing, io) do io
-            showerror(io, ex)
-        end
+        showerror(io, ex)
     finally
         backtrace && show_backtrace(io, bt)
     end


JeffBezanson (approved):

StefanKarpinski:
I agree that a wall of red isn't particularly readable. Better imo to have `ERROR:` be red and the rest of the error message be able to use color to give better messages.

PR #33681, opened 2019-10-26
Fix type rounding functions for MPFR

Keno:
Fixes #33676

PR #21970, opened 2017-05-19
Increase range of trunc/floor/ceil/round for BigFloat

omus:
Fixes #21914

PR #36027, opened 2020-05-25
Add more trunc BigFloat tests

omus:
Additional tests for #33681 which were adapted from #21970. Possibly overkill but I think the tests added in #33681 are too minimal.

### DIFF
--- a/test/mpfr.jl
+++ b/test/mpfr.jl
@@ -858,47 +858,43 @@ end
     @test trunc(UInt8, parse(BigFloat,"255.1")) == UInt8(255)
     @test_throws InexactError trunc(UInt8, parse(BigFloat,"256.1"))
 
-    # Test `InexactError` limits
-    for T in Base.BitInteger_types
-        # Use `@eval` and interpolate `$T` to make failed tests easy to read.
-        @eval begin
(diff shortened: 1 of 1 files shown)


test/mpfr.jl
```diff
@@ -857,6 +857,49 @@ end
     # Issue #33676
     @test trunc(UInt8, parse(BigFloat,"255.1")) == UInt8(255)
     @test_throws InexactError trunc(UInt8, parse(BigFloat,"256.1"))
+
+    # Test `InexactError` limits
+    for T in Base.BitInteger_types
+        # Use `@eval` and interpolate `$T` to make failed tests easy to read.
+        @eval begin
```
simonbyrne, line 864: Is the `@eval` necessary?

omus: Not strictly. I'll update to use a `@testset` for as this should provide the same context without requiring an `@eval`

simonbyrne (approved):

Keno:
I think the docs for trunc could use an update also, though. Right now they describe the behavior for `trunc(x)` and `trunc(<:Integer, ::Integer)`, but not mixed type behavior. Also, I'm not sure the `trunc(x)` documentation is quite as precise as it could be.

PR #36028, opened 2020-05-25
use a function in pkg precompile instead of writing code to stdin

JeffBezanson:
For me this reduces the time for `using Plots` (first time) from 150 seconds to 145. Not a huge improvement, but mostly I think this version of the code is easier to understand and change.

### DIFF
--- a/base/loading.jl
+++ b/base/loading.jl
@@ -1200,65 +1200,72 @@ function load_path_setup_code(load_path::Bool=true)
     return code
 end
 
+# this is called in the external process that generates precompiled package files
+function include_package_for_output(input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::typeof(_concrete_dependencies), uuid_tuple::NTuple{2,UInt64}, source::Union{Nothing,String})
+    append!(empty!(Base.DEPOT_PATH), depot_path)
+    append!(empty!(Base.DL_LOAD_PATH), dl_load_path)
+    append!(empty!(Base.LOAD_PATH), load_path)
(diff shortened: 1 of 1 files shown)


base/loading.jl
```diff
@@ -1200,66 +1200,71 @@ function load_path_setup_code(load_path::Bool=true)
...
+        Base.include(Base.__toplevel__, input)
+    catch ex
+        precompilableerror(ex) || rethrow()
+        @debug "Aborting `create_expr_cache'" exception=(ErrorException("Declaration of __precompile__(false) not allowed"), catch_backtrace())
+        exit(125) # we define status = 125 means PrecompileableError
+    end
+end
+
+precompile(include_package_for_output, (String,Vector{String},Vector{String},Vector{String},typeof(_concrete_dependencies),NTuple{2,UInt64},Nothing))
+precompile(include_package_for_output, (String,Vector{String},Vector{String},Vector{String},typeof(_concrete_dependencies),NTuple{2,UInt64},String))
```
timholy, line 1228: Since I'm guessing this must be precompiled before first use, consider putting `@assert` in front of these `precompile` directives (and perhaps the one below). If `precompile` can't find the method, [all it does is return `false`](https://github.com/JuliaLang/julia/blob/8184684657fd3eddaa116db30def36959e1e26b6/base/essentials.jl#L460). In general I tend to add them to every `precompile` statement just so I know if they've gone stale.

timholy (approved):

JeffBezanson:
I wonder if there is a risk of this exceeding the maximum argument length?

JeffBezanson:
Switched to stdin just in case.

PR #36029, opened 2020-05-26
Change getaddrinfo(host::AbstractString) to work with ipv6 localhost string "::1"

### LINKED ISSUE #36000: Sockets.connect(host::AbstractString, port::Integer) won't take IPv6 address for host
One version of connect() takes a hostname, which can be a name or a literal IPv4 address such as 127.0.0.1. But it fails on a literal IPv6 address such as ::1.
The implementation uses getaddrinfo(host) which defaults to an IPv4 address only. Better would be the first element returned from getalladdrinfo() which accepts host names and literal IPv4 and IPv6 addresses.
`julia> import Pkg; Pkg.add("Sockets")
julia> using Sockets
julia> Sockets.connect("::1", 80)`

ssikdar1:
# Description

Fixes #36000 .

The following would fail:
``` Julia
julia> using Sockets
julia> Sockets.connect("::1", 8080)
ERROR: DNSError: ::1, unknown node or service (EAI_NONAME)
Stacktrace:
 [1] getaddrinfo at /Users/ssikdar1/julia/usr/share/julia/stdlib/v1.6/Sockets/src/addrinfo.jl:136 [inlined]
 [2] getaddrinfo at /Users/ssikdar1/julia/usr/share/julia/stdlib/v1.6/Sockets/src/addrinfo.jl:139 [inlined]
 [3] connect!(::TCPSocket, ::String, ::Int64) at /Users/ssikdar1/julia/usr/share/julia/stdlib/v1.6/Sockets/src/Sockets.jl:553
 [4] connect at /Users/ssikdar1/julia/usr/share/julia/stdlib/v1.6/Sockets/src/Sockets.jl:559 [inlined]
 [5] connect(::String, ::Int64) at /Users/ssikdar1/julia/usr/share/julia/stdlib/v1.6/Sockets/src/Sockets.jl:545
 [6] top-level scope at REPL[2]:1
```

This is because Sockets.connect would call `ipaddr = getaddrinfo(host)`
Which defined here:

https://github.com/JuliaLang/julia/blob/575701248adc951038b9727e646230737e5b3c91/stdlib/Sockets/src/addrinfo.jl#L139

calls `getaddrinfo(String(host), IPv4)` with an IPv4 type. Since "::1" has no IPv4 type a DNS error is thrown here:
https://github.com/JuliaLang/julia/blob/575701248adc951038b9727e646230737e5b3c91/stdlib/Sockets/src/addrinfo.jl#L129-L137

Similarly `getaddrinfo("::1")` fails while `getaddrinfo("::1", IPv6)` works:
``` Julia
julia> getaddrinfo("::1")
ERROR: DNSError: ::1, unknown node or service (EAI_NONAME)
Stacktrace:
 [1] getaddrinfo at /Users/ssikdar1/julia/usr/share/julia/stdlib/v1.6/Sockets/src/addrinfo.jl:136 [inlined]
 [2] getaddrinfo(::String) at /Users/ssikdar1/julia/usr/share/julia/stdlib/v1.6/Sockets/src/addrinfo.jl:139
 [3] top-level scope at REPL[3]:1

julia> getaddrinfo("::1", IPv6)
ip"::1"
``` 

Change  returns the first entry from getalladdrinfo, otherwise throws a DNS error. 

# Testing

Tests added to `stdlib/Sockets/test/runtests.jl`

Manual testing of `getaddrinfo`

```Julia
julia> using Sockets

julia> getaddrinfo("::1")
ip"::1"

julia> getaddrinfo("127.0.0.1")
ip"127.0.0.1"
```

and after setting up an IPv6 server in python, using `Sockets.connect` to connect to it:
```Julia
julia> Sockets.connect("::1", 8080)
TCPSocket(RawFD(0x00000014) open, 0 bytes waiting)
```

### DIFF
--- a/stdlib/Sockets/src/addrinfo.jl
+++ b/stdlib/Sockets/src/addrinfo.jl
@@ -126,15 +126,19 @@ getalladdrinfo(host::AbstractString) = getalladdrinfo(String(host))
 Gets the first IP address of the `host` of the specified `IPAddr` type.
 Uses the operating system's underlying getaddrinfo implementation, which may do a DNS lookup.
 """
-function getaddrinfo(host::String, T::Union{Type{<:IPAddr}, Nothing})
+function getaddrinfo(host::String, T::Type{<:IPAddr})
     addrs = getalladdrinfo(host)
-    if T isa Nothing && !isempty(addrs)
-        return addrs[1]
-    end
     for addr in addrs
         if addr isa T
             return addr
         end
+    end
+    throw(DNSError(host, UV_EAI_NONAME))
+end
+function getaddrinfo(host::String, ::Nothing)
+    addrs = getalladdrinfo(host)
+    if !isempty(addrs)
+        return addrs[1]
     end
     throw(DNSError(host, UV_EAI_NONAME))
 end
--- a/stdlib/Sockets/src/addrinfo.jl
+++ b/stdlib/Sockets/src/addrinfo.jl
@@ -126,8 +126,11 @@ getalladdrinfo(host::AbstractString) = getalladdrinfo(String(host))
 Gets the first IP address of the `host` of the specified `IPAddr` type.
 Uses the operating system's underlying getaddrinfo implementation, which may do a DNS lookup.
 """
-function getaddrinfo(host::String, T::Type{<:IPAddr})
+function getaddrinfo(host::String, T::Union{Type{<:IPAddr}, Nothing})
     addrs = getalladdrinfo(host)
+    if T isa Nothing && !isempty(addrs)
+        return addrs[1]
+    end
     for addr in addrs
         if addr isa T
             return addr
@@ -136,7 +139,7 @@ function getaddrinfo(host::String, T::Type{<:IPAddr})
     throw(DNSError(host, UV_EAI_NONAME))
 end
 getaddrinfo(host::AbstractString, T::Type{<:IPAddr}) = getaddrinfo(String(host), T)
-getaddrinfo(host::AbstractString) = getaddrinfo(String(host), IPv4)
+getaddrinfo(host::AbstractString) = getaddrinfo(String(host), nothing)
 
 function uv_getnameinfocb(req::Ptr{Cvoid}, status::Cint, hostname::Cstring, service::Cstring)
     data = uv_req_data(req)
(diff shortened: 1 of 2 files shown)


ssikdar1:
regarding change c1709ad:
```
 function getaddrinfo(host::AbstractString)
     addrs = getalladdrinfo(String(host))
     if !isempty(addrs)
-        return addrs[1]
+        return addrs[end]
     end
     throw(DNSError(host, UV_EAI_NONAME))
 end
```

Turns out when getaddrinfo gets fed localhost and chooses the ipv6 address instead of the ipv4 one, the following test fails:

https://github.com/JuliaLang/julia/blob/575701248adc951038b9727e646230737e5b3c91/stdlib/Sockets/test/runtests.jl#L264

```
Test Failed at /Users/ssikdar1/julia/stdlib/Sockets/test/runtests.jl:266
  Expression: listen(randport)
    Expected: Base.IOError("listen: address already in use (EADDRINUSE)", -48)
  No exception thrown
```

Keno:
Seems reasonable to me. There are IPv6-only networks, we should not be assuming IPv4.

david-macmahon:
> There are IPv6-only networks, we should not be assuming IPv4.

I can understand the sentiment, but if anything it seems like the default should be removed altogether rather than favoring/assuming IPv6 over the previous default value of IPv4.  I mean if the API is going to change in a breaking way, might as well break it hard rather than subtly.

Seelengrab:
As I understand it, the code actually doesn't assume either way anymore - it relies on `getalladdrinfo`, which calls into the systems' version to gather all addresses for that host. The order those are returned in is system dependent.

david-macmahon:
Ah, thanks for the insight @Seelengrab.  I can confirm that the Julia 1.6 behavior mirrors the behavior of the system `getaddrinfo()` call.  Editing `/etc/gai.conf` on Ubuntu to alter the order in which the system `getaddrinfo()` call returns addresses (and address families) will also alter the order that Julia's `getalladdrinfo` function returns addresses (and address families).  Julia 1.6's `getaddrinfo(name)` just returns the first one regardless of address family.

I agree that this is preferable to the old way of assuming/favoring IPv4, but there is still a documentation inconsistency that needs to be addressed (so to speak! :)) in that IPv4 is no longer the "default".

Issue #6383, opened 2014-04-02
confusing dispatch when method parameter bounds are wider than a type parameter's declared bounds


carlobaldassi:
See a85476b2c29d9af8fab7f0a6a04809daa442a7d4 and 005197af45e28904a8cb2a346918e56ead1bc95c and comments therein.

In summary, dispatch on a parametric type whose type is restricted in the declaration (`type A{T<:Integer}` etc.) make a difference between `A`, `A{T}` and `A{T<:Integer}`. In one case, `A{T}` seems to be taken as more general than `A`; in another case, `A` seems to be taken as more general than `A{T<:Integer}` - see examples below.

Example: this shows the ambiguity

```
julia> type A{T<:Integer} <: Real
           a::T
       end

julia> type B <: FloatingPoint
       end

julia> tst{T<:FloatingPoint,S}(::Type{T}, x::A{S}) = println(1)
tst (generic function with 1 method)

julia> tst(::Type{B}, x::A) = println(2) # this should disambiguate between the one above and the one below
tst (generic function with 2 methods)

julia> tst(::Type{B}, x::Real) = println(2)
Warning: New definition 
    tst(Type{B},Real) at none:1
is ambiguous with: 
    tst(Type{T<:FloatingPoint},A{S}) at none:1.
To fix, define 
    tst(Type{B},A{S})
before the new definition.
tst (generic function with 3 methods)

julia> tst(B, A(1)) # works
2
```

Currently, the workarounds for the example above are: defining the disambiguation method with an otherwise useless parameter

```
julia> type A{T<:Integer} <: Real
           a::T
       end


julia> type B <: FloatingPoint
       end

julia> tst{T<:FloatingPoint,S}(::Type{T}, x::A{S}) = println(1)
tst (generic function with 1 method)

julia> tst{S}(::Type{B}, x::A{S}) = println(2) # useless parameter S
tst (generic function with 2 methods)

julia> tst(::Type{B}, x::Real) = println(2) # no warning though
tst (generic function with 3 methods)

julia> tst(B, A(1))
2
```

Or restricting the first method to `S<:Integer`:

```
julia> type A{T<:Integer} <: Real
           a::T
       end

julia> type B <: FloatingPoint
       end

julia> tst{T<:FloatingPoint,S<:Integer}(::Type{T}, x::A{S}) = println(1) # the <:Integer part should be implicit
tst (generic function with 1 method)

julia> tst(::Type{B}, x::A) = println(2)
[...]

PR #36030, opened 2020-05-26
Relax complex function signatures to make them ForwardDiff compatible

YingboMa:
This PR relaxes complex math function signatures so `log(::Complex{<:Dual})` can work.

Ref: https://github.com/JuliaDiff/ForwardDiff.jl/pull/455

### DIFF
--- a/base/complex.jl
+++ b/base/complex.jl
@@ -685,7 +685,7 @@ end
 function exp2(z::Complex{T}) where T
     z = float(z)
     er = exp2(real(z))
-    theta = imag(z) * log(convert(T, 2))
+    theta = imag(z) * log(convert(float(T), 2))
     s, c = sincos(theta)
     Complex(er * c, er * s)
 end
@@ -693,7 +693,7 @@ end
 function exp10(z::Complex{T}) where T
     z = float(z)
     er = exp10(real(z))
-    theta = imag(z) * log(convert(T, 10))
+    theta = imag(z) * log(convert(float(T), 10))
     s, c = sincos(theta)
     Complex(er * c, er * s)
 end
@@ -702,6 +702,7 @@ end
 function _cpow(z::Union{T,Complex{T}}, p::Union{T,Complex{T}}) where T
     z = float(z)
     p = float(p)
+    Tf = float(T)
     if isreal(p)
         pᵣ = real(p)
         if isinteger(pᵣ) && abs(pᵣ) < typemax(Int32)
@@ -709,14 +710,14 @@ function _cpow(z::Union{T,Complex{T}}, p::Union{T,Complex{T}}) where T
             # when converting p to Int, and it also turns out to be roughly
             # the crossover point for exp(p*log(z)) or similar to be faster.
             if iszero(pᵣ) # fix signs of imaginary part for z^0
-                zer = flipsign(copysign(zero(T),pᵣ), imag(z))
-                return Complex(one(T), zer)
+                zer = flipsign(copysign(zero(Tf),pᵣ), imag(z))
+                return Complex(one(Tf), zer)
             end
             ip = convert(Int, pᵣ)
             if isreal(z)
                 zᵣ = real(z)
                 if ip < 0
-                    iszero(z) && return Complex(T(NaN),T(NaN))
+                    iszero(z) && return Complex(Tf(NaN),Tf(NaN))
                     re = power_by_squaring(inv(zᵣ), -ip)
                     im = -imag(z)
                 else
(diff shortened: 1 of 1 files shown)


YingboMa:
Example:
```julia
julia> using ForwardDiff

julia> Base.nextfloat(d::ForwardDiff.Dual{T,V,N}) where {T,V,N} = ForwardDiff.Dual{T}(nextfloat(d.value), d.partials)

julia> log(ForwardDiff.Dual(1, 1) + im)
Dual{Nothing}(0.3467,0.5) + Dual{Nothing}(0.785,-0.5)*im
```

jebej:
Any potential performance impacts?

base/complex.jl
```diff
@@ -464,7 +464,7 @@ function inv(w::ComplexF64)
     return ComplexF64(p*s,q*s) # undo scaling
 end
 
-function ssqs(x::T, y::T) where T<:AbstractFloat
+function ssqs(x::T, y::T) where T
```
JeffBezanson, line 467: Maybe Real or Number?

base/complex.jl
```diff
@@ -633,7 +634,7 @@ function exp(z::Complex)
     end
 end
 
-function expm1(z::Complex{T}) where T<:Real
+function expm1(z::Complex{T}) where T
```
JeffBezanson, line 637: The parameter of `Complex` already needs to be <:Real, so this one should be ok as is?

YingboMa: I removed `<:Real` because `Complex` already only takes `Real` :p

base/complex.jl
YingboMa: This is already an internal function, so I don't think we need to constraint the type just to make it more flexible.

base/complex.jl
JeffBezanson: It's been a long-standing issue (#6383) that this method signature isn't automatically adjusted to `Complex{<:Real}` according to the declaration of `Complex`. So it's still considered good practice to put `<:Real` there.

YingboMa: I see.

YingboMa:
>     channels                        |      225     1               226
I don't think the CI failure is related.

arm61:
Even running on the nightly I am still having an error thrown with some of the examples in @YingboMa's comment (https://github.com/JuliaLang/julia/pull/36030#issuecomment-633819884). i.e. 

```
julia> using ForwardDiff
julia> sqrt(ForwardDiff.Dual(1., 1.) + 0im)
ERROR: MethodError: no method matching nextfloat(::ForwardDiff.Dual{Nothing, Float64, 1})
Closest candidates are:
  nextfloat(::Union{Float16, Float32, Float64}, ::Integer) at float.jl:595
  nextfloat(::BigFloat) at mpfr.jl:911
  nextfloat(::AbstractFloat) at float.jl:639
  ...
Stacktrace:
 [1] ssqs(x::ForwardDiff.Dual{Nothing, Float64, 1}, y::ForwardDiff.Dual{Nothing, Float64, 1})
   @ Base ./complex.jl:472
 [2] sqrt(z::Complex{ForwardDiff.Dual{Nothing, Float64, 1}})
   @ Base ./complex.jl:487
 [3] top-level scope
   @ REPL[19]:1
```
Any help/advice would be great! I am new to julia but keen to learn.

baggepinnen:
You must define the functions listed above, the following works on a 14 day old nightly.
```julia
julia> using ForwardDiff

julia> Base.float(d::ForwardDiff.Dual{T}) where T = ForwardDiff.Dual{T}(float(d.value), d.partials)

julia> Base.prevfloat(d::ForwardDiff.Dual{T,V,N}) where {T,V,N} = ForwardDiff.Dual{T}(prevfloat(float(d.value)), d.partials)

julia> Base.nextfloat(d::ForwardDiff.Dual{T,V,N}) where {T,V,N} = ForwardDiff.Dual{T}(nextfloat(float(d.value)), d.partials)

julia> function Base.ldexp(x::T, e::Integer) where T<:ForwardDiff.Dual
           if e >=0
               x * (1<<e)
           else
               x / (1<<-e)
           end
       end

julia> sqrt(ForwardDiff.Dual(1., 1.) + 0im)
Dual{Nothing}(1.0,0.5) + Dual{Nothing}(0.0,0.0)*im
```

arm61:
Amazing! It is working now for me.

kapple19:
Am I doing something wrong?

```julia
julia> using ForwardDiff

julia> Base.float(d::ForwardDiff.Dual{T}) where T = ForwardDiff.Dual{T}(float(d.value), d.partials)

julia> Base.prevfloat(d::ForwardDiff.Dual{T,V,N}) where {T,V,N} = ForwardDiff.Dual{T}(prevfloat(float(d.value)), d.partials)

julia> Base.nextfloat(d::ForwardDiff.Dual{T,V,N}) where {T,V,N} = ForwardDiff.Dual{T}(nextfloat(float(d.value)), d.partials)

julia> function Base.ldexp(x::T, e::Integer) where T<:ForwardDiff.Dual
                  if e >=0
                                 x * (1<<e)
                                            else
                                                           x / (1<<-e)
                                                                      end
                                                                             end

julia> sqrt(ForwardDiff.Dual(1., 1.) + 0im)
ERROR: StackOverflowError:
Stacktrace:
 [1] sqrt(::Complex{ForwardDiff.Dual{Nothing,Float64,1}}) at .\complex.jl:506 (repeats 79984 times)
```

arm61:
Are you using the nightly `1.6.x` build? That's what I needed.

PR #36043, opened 2020-05-27
Add a pointer to GPUs in parallel computing

ViralBShah:
cc @vchuravy @maleadt 

Should Asyncrhonous computing live here or should it be a separate chapter - not related to the parallel computing?

Would it be too much of a stretch to mention Instruction level parallelism and SIMD in this list?

### DIFF
--- a/doc/src/manual/parallel-computing.md
+++ b/doc/src/manual/parallel-computing.md
@@ -1,6 +1,6 @@
 # Parallel Computing
 
-Julia supports four main categories of features for concurrent and parallel programming:
+Julia supports these four categories of concurrent and parallel programming:
 
 1. **Asynchronous "tasks", or coroutines**:
 
--- a/doc/src/manual/parallel-computing.md
+++ b/doc/src/manual/parallel-computing.md
@@ -3,24 +3,28 @@
 Julia supports three main categories of features for concurrent and parallel programming:
 
 1. *Asynchronous "tasks", or coroutines*:
+
     Julia Tasks allow suspending and resuming computations
     for I/O, event handling, producer-consumer processes, and similar patterns.
     Tasks can synchronize through operations like [`wait`](@ref) and [`fetch`](@ref), and
     communicate via [`Channel`](@ref)s. While strictly not parallel computing by themselves,
     Julia lets you schedule `Task`s on several threads.
 
-2. *Multi-threading*
+2. *Multi-threading*:
+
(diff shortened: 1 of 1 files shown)


doc/src/manual/parallel-computing.md
```diff
@@ -2,28 +2,28 @@
 
 Julia supports three main categories of features for concurrent and parallel programming:
```
DilumAluthge, line 3:
```suggestion
Julia supports four main categories of features for concurrent and parallel programming:
```

DilumAluthge (changes requested):

vchuravy:
I would not strongly differentiate between coroutines and multi-threading. Julia uses the language of tasks and coroutines and you should use channels alongside parallel execution. In essence multi-threading is an implementation detail, and the semantic components are `Tasks`.

StefanKarpinski:
But there are two different ways to start tasks: `@async` which creates a new task that will always run on the same core and `@spawn` which creates a new task that will potentially run on a different core. That's a fairly key distinction which surely should be mentioned here.

ViralBShah:
@vchuravy Can you review?

@StefanKarpinski Mentioning `@async` vs. `@spawn` here seems a bit too early, when neither have yet been described. But it definitely makes sense to do it in the chapters that follow.

doc/src/manual/parallel-computing.md
```diff
@@ -1,20 +1,34 @@
...
+    is composable. When one multi-threaded function calls another multi-threaded function, Julia
+    will schedule all the threads globally on available resources, without oversubscribing.
+
+3. **Distributed computing**:
+
+    Distributed computing runs multiple Julia processes with separate memory spaces. These can be on the same
+    computer or multiple computers. The `Distributed` standard library provides the capability for remote execution
+    of a Julia function. With this basic building block, it is possible to build many different kinds of
+    distributed computing abstractions. Packages like [`DistributedArrays.jl`](https://github.com/JuliaParallel/DistributedArrays.jl)
+    are an example of such an abstraction. On the other hadn, packages like [`MPI.jl`](https://github.com/JuliaParallel/MPI.jl) and
```
jonas-schulze, line 27:
```suggestion
    are an example of such an abstraction. On the other hand, packages like [`MPI.jl`](https://github.com/JuliaParallel/MPI.jl) and
```

StefanKarpinski:
My point is that having those two primitive s means that concurrency and threading are different concepts in the language.

PR #36046, opened 2020-05-27
Fix a bug with break/continue/return in at-testset begin end

### LINKED ISSUE #32937: break breaks push/pop_testset invariance
```julia
julia> using Test

julia> @testset for i in 1:1
           @testset begin
               break
           end
       end
1-element Array{Any,1}:
 Test.DefaultTestSet("i = 1", Any[Test.DefaultTestSet(#= circular reference @-2 =#)], 0, false)

julia> Test.get_testset_depth()  # != 0
1

julia> @testset for i in 1:1  # same testset again
           @testset begin
               break
           end
       end
1-element Array{Any,1}:
 Test.DefaultTestSet("i = 1", Any[Test.DefaultTestSet(#= circular reference @-2 =#)], 0, false)

julia> Test.get_testset_depth()  # != 0
2
```

I think the code above should be an error.  Maybe `@testset` should be lowered into an "immediately invoked function expression"?

### LINKED ISSUE #36045: Calling return from within a testset, makes overall tests to pass even when there are failures
If from a testset you call a function that contains a testset that contains a return, then any test failures after the point the functions are called will not actually cause the outer testset to fail.

Encountered this in the wild (fix is in https://github.com/invenia/Models.jl/pull/16)
where someone thought they could return a value from a testset to use in later tests.
This ended up resulting in ony of  our packages passing CI but actually failing tons of tests, and noone noticed for months because they trusted CI and just looked at the end of the tests to see if there was a failure message.

```julia
julia> function check_foo()
           @testset "Test Foo" begin
               @test true
               return rand()
           end
       end
check_foo (generic function with 1 method)

julia> k = @testset "main" begin
           check_foo()
           @test false
       end
Test Foo: Test Failed at REPL[7]:3
  Expression: false
Stacktrace:
 [1] top-level scope at REPL[7]:3
 [2] top-level scope at /usr/local/src/julia/julia-1.5/usr/share/julia/stdlib/v1.5/Test/src/Test.jl:1114
 [3] top-level scope at REPL[7]:2
Test.DefaultTestSet("main", Any[Test.DefaultTestSet(#= circular reference @-2 =#)], 0, false)

julia> k.anynonpass
false
```

Probably related to https://github.com/JuliaLang/julia/issues/32937

tkf:
fix #32937 fix #36045

### DIFF
--- a/stdlib/Test/src/Test.jl
+++ b/stdlib/Test/src/Test.jl
@@ -1100,6 +1100,7 @@ function testset_beginend(args, tests, source)
     # action (such as reporting the results)
     ex = quote
         _check_testset($testsettype, $(QuoteNode(testsettype.args[1])))
+        local ret
         local ts = $(testsettype)($desc; $options...)
         push_testset(ts)
         # we reproduce the logic of guardseed, but this function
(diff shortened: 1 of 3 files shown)


rfourquet:
Can this be backported to 1.0 too?

PR #36047, opened 2020-05-27
support contiguous subarrays in isvalid(String, ...)

stevengj:
I noticed in JuliaWeb/HTTP.jl#538 that `isvalid(String, x)` supports `x::Vector{UInt8}` but not subarrays thereof.   This PR makes it at least support contiguous subarrays.

### DIFF
--- a/base/strings/unicode.jl
+++ b/base/strings/unicode.jl
@@ -47,6 +47,9 @@ true
 julia> isvalid(Char, 0xd799)
 true
 ```
+
+!!! compat "Julia 1.6"
+    Support for subarray values was added in Julia 1.6.
 """
 isvalid(T,value)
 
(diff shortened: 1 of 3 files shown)


stevengj:
FreeBSD failure is apparently unrelated "SuiteSparse threads test failed with nthreads == 4".

base/strings/unicode.jl
```diff
@@ -33,7 +33,8 @@ isvalid(value)
 
 Returns `true` if the given value is valid for that type. Types currently can
 be either `AbstractChar` or `String`. Values for `AbstractChar` can be of type `AbstractChar` or [`UInt32`](@ref).
-Values for `String` can be of that type, or `Vector{UInt8}` or `SubString{String}`.
+Values for `String` can be of that type, `SubString{String}`, `Vector{UInt8}`,
+or a contiguous subarray thereof.
```
KristofferC, line 37: Perhaps this needs a `compat` annotation that it requires 1.6?

stevengj: Done.

PR #36049, opened 2020-05-27
micro-optimize UUID(::AbstractString)

KristofferC:
I was profiling Pkg and saw some traces pointing to `UUID(::String)`. The current implementations allocates a brand new vector for `lowercase` (which I would guess is almost never needed) and it also does a regex call to check it the format of the string is ok. This speeds it up by a bit.

Setup:

```jl
julia> using UUIDs

julia>  a = uuid4() |> string

julia> uuid_strs = string.(UUID.(rand(UInt128, 10^6)));

julia> uuids = Vector{UUID}(undef, 10^6);
```

Master

``` 
julia> @btime UUID($a);
  319.275 ns (3 allocations: 224 bytes)

julia> @btime $uuids .= UUID.($uuid_strs)
  373.077 ms (3000000 allocations: 213.62 MiB)
```

PR
```
julia> @btime UUID($a);
  27.489 ns (0 allocations: 0 bytes)

julia> @btime $uuids .= UUID.($uuid_strs)
  111.078 ms (0 allocations: 0 bytes)
```

### DIFF
--- a/base/uuid.jl
+++ b/base/uuid.jl
@@ -37,7 +37,7 @@ let
     d = __convert_digit(_c, UInt32(16))
     d >= 16 && throw_malformed_uuid(s)
     u <<= 4
-    u |= d
+    u | d
 end
 
 global UUID
--- a/base/uuid.jl
+++ b/base/uuid.jl
@@ -30,24 +30,45 @@ end
 
 UInt128(u::UUID) = u.value
 
-let groupings = [1:8; 10:13; 15:18; 20:23; 25:36]
-    global UUID
-    function UUID(s::AbstractString)
-        s = lowercase(s)
-
-        if !occursin(r"^[0-9a-f]{8}(?:-[0-9a-f]{4}){3}-[0-9a-f]{12}$", s)
-            throw(ArgumentError("Malformed UUID string: $(repr(s))"))
-        end
-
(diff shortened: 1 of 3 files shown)


StefanKarpinski:
Wow, dude. You really got into that. Haven’t reviewed in detail, but since we do print a lot of these, 👍🏻

KristofferC:
> but since we do print a lot of these, 

I think printing them would be relevant for the `string(::UUID)` function but this is more for ingesting them.

StefanKarpinski:
Yeah, we do a lot of UUID parsing too.

base/uuid.jl
```diff
@@ -30,35 +30,56 @@ end
...
-            throw(ArgumentError("Malformed UUID string: $(repr(s))"))
-        end
+let
+@noinline throw_malformed_uuid(s) = throw(ArgumentError("Malformed UUID string: $(repr(s))"))
+@inline function uuid_kernel(s, i, u)
+    _c = UInt32(@inbounds codeunit(s, i))
+    d = __convert_digit(_c, UInt32(16))
+    d >= 16 && throw_malformed_uuid(s)
+    u <<= 4
+    u |= d
```
JeffBezanson, line 40:
```suggestion
    u | d
```

PR #36054, opened 2020-05-28
give more information in StringIndexError

KristofferC:
Before

```jl
julia> str = "abcdefghκijklmno"
"abcdefghκijklmno"

julia> str[10]
ERROR: StringIndexError("abcdefghκijklmno", 10)
Stacktrace:
```

Now

```jl
julia> str = "abcdefghκijklmno"
"abcdefghκijklmno"

julia> str[10]
ERROR: StringIndexError: invalid index [10], valid previous index [9] ('κ'), valid next index [11] ('i')
```

### DIFF
--- a/base/strings/string.jl
+++ b/base/strings/string.jl
@@ -11,6 +11,15 @@ struct StringIndexError <: Exception
 end
 @noinline string_index_err(s::AbstractString, i::Integer) =
     throw(StringIndexError(s, Int(i)))
+function Base.showerror(io::IO, exc::StringIndexError)
+    s = exc.string
+    i = thisind(s, exc.index)
+    print(io, "StringIndexError: ", "invalid index [$(exc.index)], valid previous index [$i] ('$(s[i])')")
+    inext = nextind(s, i)
+    if !(inext > ncodeunits(s))
+        print(io, ", valid next index [$(inext)] ('$(s[inext])')")
+    end
+end
 
 const ByteArray = Union{Vector{UInt8},Vector{Int8}}
 
--- a/base/strings/string.jl
+++ b/base/strings/string.jl
@@ -13,11 +13,15 @@ end
     throw(StringIndexError(s, Int(i)))
 function Base.showerror(io::IO, exc::StringIndexError)
     s = exc.string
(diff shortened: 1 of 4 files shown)


KristofferC:
Bikeshed over `valid previous index [1] ('α')`?

Perhaps `valid previous index [1]=='α'`?

base/strings/string.jl
```diff
@@ -11,6 +11,15 @@ struct StringIndexError <: Exception
 end
 @noinline string_index_err(s::AbstractString, i::Integer) =
     throw(StringIndexError(s, Int(i)))
+function Base.showerror(io::IO, exc::StringIndexError)
+    s = exc.string
+    i = thisind(s, exc.index)
```
mbauman, line 16: `StringIndexError` is exported and may be (mis)used by user code — maybe defensively check for `exc.index < firstindex(s) && return invoke(showerror, Tuple{IO, Any}, io, exc)` before doing something that may throw?

StefanKarpinski:
> Bikeshed over `valid previous index [1] ('α')`?

I like that version fine, it's certainly clearer to me than `[1]=='α'`.

StefanKarpinski:
Oh but "previous valid index" reads more naturally than "valid previous index". There's some rule about how adjectives get sorted in English and "valid previous index" violates them (in some way that I only know intuitively, I have no idea what the rules actually are, just that they exist).

StefanKarpinski:
Same for "valid next index" which would be "next valid index" instead.

mbauman:
We use pairs to connect indices to values in some cases; it could be `nearby valid indices 9=>'κ', 11=>'i'`?  Of course the irregular plural is annoying at the end of the string, but the `previous` and `next` aren't doing much heavy lifting since I can see that 9<10<11.

KristofferC:
> Oh but "previous valid index" reads more naturally than "valid previous index".

Yeah, I thought about this too but decided on this to sort of emphasise the "valid `nextind`" and "valid `prevind`" connection but I can change the order if it is too weird.

KristofferC:
Did some tweaks based on comments

PR #36056, opened 2020-05-28
Link to "noteworthy differences" early in docs

perrutquist:
It frequently happens that newcomers to Julia are tripped up by some aspect of the language, such as integer overflow, that is not present in their current favorite programming language. The list of noteworthy differences points out many of these pitfalls, so getting more people to read it can reduce user frustration.

[From discourse](https://discourse.julialang.org/t/the-speed-of-light-is-an-integer-why-should-we-care/40108/51?u=per).

### DIFF
--- a/doc/src/manual/getting-started.md
+++ b/doc/src/manual/getting-started.md
@@ -3,7 +3,7 @@
 Julia installation is straightforward, whether using precompiled binaries or compiling from source.
 Download and install Julia by following the instructions at [https://julialang.org/downloads/](https://julialang.org/downloads/).
 
-If you are coming to Julia from one of the following languages, then you should start by reading the section on noteworthy differences from [Matlab](@ref Noteworthy differences from MATLAB), [R](@ref Noteworthy differences from R), [Python](@ref Noteworthy differences from Python), [C/C++](@ref Noteworthy differences from C/C++) or [Common Lisp](@ref Noteworthy differences from Common Lisp). This will help you avoid some common pitfalls since Julia differs from those languages in many subtle ways.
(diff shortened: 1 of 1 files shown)


doc/src/manual/getting-started.md
```diff
@@ -3,6 +3,8 @@
 Julia installation is straightforward, whether using precompiled binaries or compiling from source.
 Download and install Julia by following the instructions at [https://julialang.org/downloads/](https://julialang.org/downloads/).
 
+If you are coming to Julia from one of the following languages, then you should start by reading the section on noteworthy differences from [Matlab](@ref Noteworthy differences from MATLAB), [R](@ref Noteworthy differences from R), [Python](@ref Noteworthy differences from Python), [C/C++](@ref Noteworthy differences from C/C++) or [Common Lisp](@ref Noteworthy differences from Common Lisp). This will help you avoid some common pitfalls since Julia differs from those languages in many subtle ways.
```
ViralBShah, line 6: These references don't look right.

kescobo: Easiest way to figure out the link is to just go to the page and click on the headers. This will update the url in your browser. So https://docs.julialang.org/en/v1/manual/noteworthy-differences/#Noteworthy-differences-from-MATLAB-1

should be `(@ref Noteworthy-differences-from-MATLAB-1)`. I think...

perrutquist: Thanks! Will fix.

ViralBShah: I am not sure about the 1. Probably check a few other existing references to see how to map the URLs. The doctests will tell you if you got it right. But if you ahve a source build, just do `make` in docs. It's quite easy and reasonably quick.

perrutquist: Without the -1, the docs build and the links seem to work correctly.

Issue #28778, opened 2018-08-20
Convert Enum to integer value with correct type


scls19fr:
Hello,

Following https://github.com/andrewadare/LibSerialPort.jl/pull/19#issuecomment-414151579 and discussion with @tanmaykm and @andrewadare maybe Enum docstring should be improved to show how to get BaseType.

According https://github.com/andrewadare/LibSerialPort.jl/pull/19#issuecomment-414190281 it can be done using

```julia
julia> @enum Fruit::UInt8 apple orange kiwi

julia> Base.Enums.basetype(Fruit)
UInt8
```

but we should also document how to get integer value of an Enum with correct type, I have found

```julia
julia> Base.Enums.basetype(typeof(kiwi))(kiwi)
0x02
```

but maybe there is a simpler solution which should, in such a case, be documented because in most case it's suggested to use `Int`, but it's changing type

```julia
julia> supertype(Fruit)
Enum{UInt8}

julia> typeof(Int(apple))
Int64
```

Kind regards

martinholters:
The simpler solution is `Integer(kiwi)`. Mentioning this in the docstring seems like a god idea.

scls19fr:
So maybe 
https://github.com/JuliaLang/julia/blob/f6c48ebf83983016c158f790a466b354ece7addc/base/Enums.jl#L42

should be changed for:

    julia> f(x::Fruit) = "I'm a Fruit with value: \$(Integer(x))"

PR #36057, opened 2020-05-28
Mention how to get the basetype of an enum

ViralBShah:
I feel this is sufficient to address #28778, since it is a non-standard case. The type conversion with Integer() is not specific to enums - and not necessary.


JeffBezanson:
I think it would be better to mention `Integer(x)` than to suggest using an internal function.

ViralBShah:
Yes, I didn't like having the internal method there.

Just to make sure I understand - instead of having people get the BaseType and cast to it, they should use Integer on the enum value directly. And of course you can then do `typeof` on that to get the BaseType?

PR #36067, opened 2020-05-29
Bump LibGit2 to 0.28.5

simonbyrne:
Until we can build 1.0.0 (#35233).

cc: @giordano @nalimilan


giordano:
You need also to

* update the checksum files with [`contrib/refresh_bb_tarballs.sh`](https://github.com/JuliaLang/julia/blob/master/contrib/refresh_bb_tarballs.sh),
* delete the old ones,
* update the build script at [`deps/libgit2.mk`] (https://github.com/JuliaLang/julia/blob/master/deps/libgit2.mk), adding the new patch to [`deps/patches `] (https://github.com/JuliaLang/julia/tree/master/deps/patches)

simonbyrne:
@giordano done.

giordano:
Ok, I think it's fine, but would be better if someone  more involved in Julia development could review, too :slightly_smiling_face:

nalimilan:
Looking at CI logs I see
```
/buildworker/worker/package_linux64/build/deps/tools/jldownload /tmp/srccache/LibGit2.v0.28.5-0.x86_64-linux-gnu.tar.gz https://github.com/JuliaBinaryWrappers/LibGit2_jll.jl/releases/download/LibGit2-v0.28.5+0/LibGit2.v0.28.5.x86_64-linux-gnu.tar.gz
```
and then donwload progress output. Not sure whether it's expected the first time, but it doesn't happen for other deps, so...

simonbyrne:
> Not sure whether it's expected the first time, but it doesn't happen for other deps, so...

I assume that is because the downloads are cached?

PR #36073, opened 2020-05-29
Add docstring for keys(::AbstractArray)

nalimilan:
This is helpful for users, and it matters because some Base functions rely on `keys(::AbstractArray)` returning a collection of integer ranges (e.g. `hash`).

Cc: @bkamins

### DIFF
--- a/base/abstractarray.jl
+++ b/base/abstractarray.jl
@@ -98,6 +98,16 @@ axes1(iter) = OneTo(length(iter))
 unsafe_indices(A) = axes(A)
 unsafe_indices(r::AbstractRange) = (OneTo(unsafe_length(r)),) # Ranges use checked_sub for size
 
+"""
+    keys(a::AbstractArray)
+
+Return the tuple of valid indices for array `a`. This is equivalent to
+[`LinearIndices(a)`](@ref) for vectors, and to [`CartesianIndices(axes(a))`](@ref)
+for higher-dimensional arrays.
+
+This may not return the most efficient indices type to iterate over `a`:
+use [`eachindex`](@ref) instead for maximum performance.
+"""
 keys(a::AbstractArray) = CartesianIndices(axes(a))
 keys(a::AbstractVector) = LinearIndices(a)
 
--- a/base/abstractarray.jl
+++ b/base/abstractarray.jl
@@ -101,12 +101,15 @@ unsafe_indices(r::AbstractRange) = (OneTo(unsafe_length(r)),) # Ranges use check
 """
     keys(a::AbstractArray)
 
-Return the tuple of valid indices for array `a`. This is equivalent to
-[`LinearIndices(a)`](@ref) for vectors, and to [`CartesianIndices(axes(a))`](@ref)
-for higher-dimensional arrays.
+Return an efficient array describing all valid indices for `a` arranged in the shape of `a` itself.
 
-This may not return the most efficient indices type to iterate over `a`:
-use [`eachindex`](@ref) instead for maximum performance.
+They keys of 1-dimensional arrays (vectors) are integers, whereas all other N-dimensional
+arrays use [`CartesianIndex`](@ref) to describe their locations.  Often the special array
+types [`LinearIndices`](@ref) and [`CartesianIndices`](@ref) are used to efficiently
+represent these arrays of integers and `CartesianIndex`es, respectively.
+
+Note that the `keys` of an array might not be the most efficient index type; for maximum
+performance use  [`eachindex`](@ref) instead.
 """
 keys(a::AbstractArray) = CartesianIndices(axes(a))
 keys(a::AbstractVector) = LinearIndices(a)
--- a/doc/src/base/arrays.md
+++ b/doc/src/base/arrays.md
@@ -51,6 +51,7 @@ Base.size
 Base.axes(::Any)
 Base.axes(::AbstractArray, ::Any)
 Base.length(::AbstractArray)
+Base.keys(::AbstractArray)
 Base.eachindex
 Base.IndexStyle
 Base.IndexLinear


base/abstractarray.jl
```diff
@@ -98,6 +98,16 @@ axes1(iter) = OneTo(length(iter))
...
 
+"""
+    keys(a::AbstractArray)
+
+Return the tuple of valid indices for array `a`. This is equivalent to
+[`LinearIndices(a)`](@ref) for vectors, and to [`CartesianIndices(axes(a))`](@ref)
+for higher-dimensional arrays.
+
+This may not return the most efficient indices type to iterate over `a`:
+use [`eachindex`](@ref) instead for maximum performance.
```
mbauman, line 109: I don't think we need to explicitly demand that custom arrays return either `LinearIndices` or `CartesianIndices` — the important part is simply that they're arrays of integers or `CartesianIndex`.  I do like the pointers to them, though.  I think I'd prefer to see this written more like:

```suggestion
Return an efficient array describing all valid indices for `a` arranged in the shape of `a` itself.

They keys of 1-dimensional arrays (vectors) are integers, whereas all other N-dimensional
arrays use [`CartesianIndex`es](@ref) to describe their locations.  Often the special array
types [`LinearIndices`](@ref) and [`CartesianIndices`](@ref) are used to efficiently
represent these arrays of integers and `CartesianIndex`es, respectively.

Note that the `keys` of an array might not be the most efficient index type; for maximum
performance use  [`eachindex`](@ref) instead.
```

mbauman:
> some Base functions rely on `keys(::AbstractArray)` returning a collection of integer ranges (e.g. `hash`).

Are you thinking of `axes`?

bkamins:
If I understand what @naliman wants to point that `hash(A::AbstractArray, h::UInt)` actually requires `keys` to return numeric indices.

This distinction is relevant for arrays that allow for other indexing except numbers (like NamedArrays.jl) - and the question is if such arrays can return those "other indices" though `keys`. And we came to the conclusion that it is not allowed - but maybe it is not intended?

mbauman:
Fascinating.  I'm not sure anyone has tried doing this yet — and we don't have a well-defined boundary for what needs to be overridden if `keys(::AbstractArray)` are something other than an integer or `CartesianIndex` (or if that's possible at all).

> wants to point that hash(A::AbstractArray, h::UInt) actually requires keys to return numeric indices.

That's not quite how hash uses `keys`; I think it just depends on these two rough requirements:

* `LinearIndices` contains integers
* `keys(a)` are valid not just for `a` but also for `LinearIndices(a)` and `CartesianIndices(a)`.  In fact, I think all three of these should be valid keys for each other!  For example, a rough relation that I have in my head is:

    ```julia
    a[keys(a)]==a && LinearIndices(a)[keys(a)] == LinearIndices(a) && CartesianIndices(a)[keys(a)] == CartesianIndices(a)
    ```

I'd be inclined to help make this work rather than simply documenting that it doesn't.  In fact, NamedArrays might be in good stead for this usage if it provides `NamedLinearIndices` and `NamedCartesianIndices` and sneakily overrides the `LinearIndices(::NamedArray)` constructor to pass back its named version.

mbauman:
At the same time, `keys(::AbstractVector)::AbstractVector{<:Integer}` is probably too well-entrenched to do anything about it thanks to downstream uses of `find`.

bkamins:
Well - not knowing these details we already made `DataFrameColumns` object to be an `AbstractVector` and have `keys` other than integers, but at the same time it seems we are consistent what is required:
```
julia> df = DataFrame(rand(2,3))
2×3 DataFrame
│ Row │ x1       │ x2       │ x3       │
│     │ Float64  │ Float64  │ Float64  │
├─────┼──────────┼──────────┼──────────┤
│ 1   │ 0.276473 │ 0.961011 │ 0.329211 │
│ 2   │ 0.193074 │ 0.643824 │ 0.292972 │

julia> e = eachcol(df)
2×3 DataFrameColumns
│ Row │ x1       │ x2       │ x3       │
│     │ Float64  │ Float64  │ Float64  │
├─────┼──────────┼──────────┼──────────┤
│ 1   │ 0.276473 │ 0.961011 │ 0.329211 │
│ 2   │ 0.193074 │ 0.643824 │ 0.292972 │

julia> keys(e)
3-element Array{Symbol,1}:
 :x1
 :x2
 :x3

julia> LinearIndices(keys(e))
3-element LinearIndices{1,Tuple{Base.OneTo{Int64}}}:
 1
 2
 3
```

mbauman:
Nice, that's great!  That's precisely the sort of usage that I think is acceptable (and beyond that, it's great!) and I don't want to document it away.

nalimilan:
@bkamins AFAICT `DataFrameColumns` doesn't satisfy the properties that @mbauman mentioned above. With your example, `e[keys(e)]`, `LinearIndices(e)[keys(e)]` and `CartesianIndices(e)[keys(e)]` all fail. Even `findfirst(x->false, e)` fails actually. Not sure what would be the best solution... We could probably fix `getindex` and `findfirst` (via `nextind`), but should we really implement custom `LinearIndices` and `CartesianIndices` methods?

nalimilan:
I just noticed that `keys` should probably also be documented in the Interfaces section for `AbstractArray`, right?

BTW, @mbauman, how do attempts to improve the handling of axis names in `axes` like https://github.com/JuliaArrays/AxisArrays.jl/pull/81 fit in this discussion? Should `axes` and `keys` differ in that regard?

bkamins:
> I just noticed that keys should probably also be documented in the Interfaces section for AbstractArray, right?

Actually this is what I have also checked today :).

I think that the key thing that would be great to have decided is the rule for `keys` for "array-like" collections that allow "dual" indexing (with integers and with something else).

The issue is that - at least in DataFrames.jl - we found it more convenient for `keys` to return names and `pairs` return `name => value` mappings for such types. This is exactly what `NamedTuple` type does now, but it is inconsistent with what Base now assumes for arrays.

bkamins:
So what is the decision here? Just to confirm, so that we can take appropriate steps in DataFrames.jl.

Is the contract that `keys` on `AbstractVector` must return a collection of integers? Thank you!

StefanKarpinski:
@mbauman, what's your take?

mbauman:
Good question.

> Is the contract that keys on AbstractVector must return a collection of integers?

This is definitely the "safe" thing to do, and maybe it's what we need to demand.  I was hoping that we weren't too far from making this possible, but I think it's too tricky to try to officially support at the moment.

PR #36076, opened 2020-05-29
clarify `show` doc strings

### LINKED ISSUE #36072: Display of array elements seems to use ::Mime:text/plain in juia1.5.0-beta1
I  am not sure  this is a  bug, but at  least it is  an incompatible change which  demands some explanation on how to proceed to get the same effect as before.

It  is convenient to  use a compact  display when displaying array elements which  does not reflect the type (which is no problem since the `eltype` is displayed  at the beginning of  the array display). On  the other hand when the element is displayed alone the type can be written first to show it.

Here  is an  example of  what I  mean for  a type  "integers mod  p" that I define:

```
  (_)     | (_) (_)    |
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.4.2 (2020-05-23)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> struct Mod{p} <: Number
          val::Int
          function Mod{p}(a::Integer) where {p}
                  new(mod(a,p))
          end
       end

julia> Base.show(io::IO,x::Mod{p}) where p=print(io,x.val)

julia> function Base.show(io::IO, ::MIME"text/plain", x::Mod{p}) where p
         print(io,typeof(x),": ")
         show(io,x)
       end

julia> Mod{19}(2)
Mod{19}: 2

julia> m=Mod{19}.([1 2 3;4 5 6])
2×3 Array{Mod{19},2}:
 1  2  3
 4  5  6
```

In julia1.5.0-beta1 the display of the array has changed, it seems to use `::MIME"text/plain"`

```
julia> m=Mod{19}.([1 2 3;4 5 6])
2×3 Array{Mod{19},2}:
 Mod{19}: 1  Mod{19}: 2  Mod{19}: 3
 Mod{19}: 4  Mod{19}: 5  Mod{19}: 6
```

What is the recommended way to get the same effect as before?

JeffBezanson:
fixes #36072

### DIFF
--- a/base/show.jl
+++ b/base/show.jl
@@ -265,9 +265,9 @@ the properties of that stream (note that `io` can itself be an `IOContext`).
 
 The following properties are in common use:
 
- - `:compact`: Boolean specifying that small values should be printed more compactly, e.g.
+ - `:compact`: Boolean specifying that values should be printed more compactly, e.g.
    that numbers should be printed with fewer digits. This is set when printing array
-   elements.
+   elements. `:compact` output should not contain line breaks.
  - `:limit`: Boolean specifying that containers should be truncated, e.g. showing `…` in
    place of most elements.
  - `:displaysize`: A `Tuple{Int,Int}` giving the size in rows and columns to use for text
@@ -358,13 +358,20 @@ function show_circular(io::IOContext, @nospecialize(x))
 end
 
 """
-    show(x)
+    show([io::IO = stdout], x)
 
-Write an informative text representation of a value to the current output stream. New types
-should overload `show(io::IO, x)` where the first argument is a stream. The representation used
-by `show` generally includes Julia-specific formatting and type information.
+Write a text representation of a value `x` to the output stream `io`. New types `T`
+should overload `show(io::IO, x::T)`. The representation used by `show` generally
+includes Julia-specific formatting and type information, and should be parseable
+Julia code when possible.
 
 [`repr`](@ref) returns the output of `show` as a string.
+
+To customize human-readable text output for objects of type `T`, define
+`show(io::IO, ::MIME"text/plain", ::T)` instead. Checking the `:compact`
+[`IOContext`](@ref) property of `io` in such methods is recommended,
+since some containers show their elements by calling this method with
+`:compact => true`.
 
 See also [`print`](@ref), which writes un-decorated representations.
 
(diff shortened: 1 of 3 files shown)


base/show.jl
```diff
@@ -358,14 +358,20 @@ function show_circular(io::IOContext, @nospecialize(x))
...
+should overload `show(io::IO, x::T)`. The representation used
+by `show` generally includes Julia-specific formatting and type information, and should
+be parseable Julia code when possible.
 
 [`repr`](@ref) returns the output of `show` as a string.
 
+To customize human-readable text output for objects of type `T`, define
+`show(io::IO, ::MIME"text/plain", ::T)` instead. To customize how objects
+are shown inside containers, check the `:compact` [`IOContext`](@ref)
+property of `io`.
```
tkf, line 373: It'd be nice if we could simply say

```suggestion
property of `io` in `show(io::IO, ::MIME"text/plain", ::T)`.
```

However, current situation is complicated because

* non-vector arrays calls 3-arg `show` and then fallback 2-arg `show`, with `:compact => true` set.
* vectors do the same but without setting `:compact => true`.
* dictionaries and sets use 2-arg `show`.

Maybe it's better to be honest and say that there are some rough edges in `show` system with containers and there is no consistent way to control how items with custom items are shown.  I do still think we should be using `show(IOContext(io, :compact => true), "text/plain", element)` consistently everywhere but it cannot block 1.5.

tkf (approved):
Thanks for documenting this.

base/show.jl
JeffBezanson: I'll try to reword this. We shouldn't think of it as controlling how your type is displayed inside a container --- that can never fully work, because it will always depend on what the container does. All you really need to know is "MIME type = human readable, 2-arg = parseable, compact = keep it short". For example, matrices call 3-arg show because they are themselves using a human-readable output format. But if some other container only has a parseable representation, then its elements will use 2-arg show.

tkf: > All you really need to know is "MIME type = human readable, 2-arg = parseable, compact = keep it short".

Yes, that's the ideal situation. It'd be great if we can orthogonalize target reader ("machines" vs humans) and the compactness of the output.  The reason why I suggested to add "in `show(io::IO, ::MIME"text/plain", ::T)`" was that I thought this paragraph was about human-readable format (as that's how it starts).

Issue #25904, opened 2018-02-06
Julep: `broadcast` with dictionaries


andyferris:
### Current behavior

I just learned that apparently `broadcast` doesn't work as I expected on dictionaries:

```julia
julia> d = Dict(:a => "Alice", :b => "Bob", :c => "Charlie")
Dict{Symbol,String} with 3 entries:
  :a => "Alice"
  :b => "Bob"
  :c => "Charlie"

julia> map(kv -> kv[1] => kv[2], d)
Dict{Symbol,String} with 3 entries:
  :a => "Alice"
  :b => "Bob"
  :c => "Charlie"

julia> broadcast(kv -> kv[1] => kv[2], d)
ERROR: KeyError: key 1 not found
Stacktrace:
 [1] getindex at ./dict.jl:474 [inlined]
 [2] (::##17#18)(::Dict{Symbol,String}) at ./REPL[11]:1
 [3] broadcast(::Function, ::Dict{Symbol,String}) at ./broadcast.jl:455

julia> broadcast(length, d)
3
```

AFAICT it seems `broadcast` is working on dictionaries like they are scalars or something. Someone correct me if I'm wrong - but maybe we never added any `broadcast` methods for dictionaries?

### Proposal

Make `broadcast` work with dictionaries. I can see two possibilities, of which I favor option 2:

 1. `broadcast` for 1D indexable containers generally works like `map`, so we can extend this to `AbstractDict` and have `broadcast` map the key-value pairs
 2. `broadcast` deals with matching indices and performing some operation on *values*. So `broadcast(f, dict)` could be a bit like `map(f, values(dict))` except the dictionary structure and it's indices are preserved in the output.

It seems to me that `map` deals with the iteration protocol and `broadcast` is more about matching indices and mapping those values. Using option 2 would give:

```julia
julia> length.(d)
Dict{Symbol,Int64} with 3 entries:
  :a => 5
  :b => 3
  :c => 7
```

IMO this kind of syntax could help a lot with manipulating dictionaries (one example is using `AbstractDict` for output of `groupby`-style operations, and then acting on the groups) while making no breaking changes to dictionary iteration or map. 

To me it also dovetails with the "getindices julep" via this comment https://github.com/JuliaLang/julia/issues/24019#issuecomment-363253768 which basically suggests we can get multi-valued getindex between different indexable container types including dictionaries via the lowering transformation:

```
a.[inds] --> broadcast(i -> a[i], inds)
```

[...]

PR #36085, opened 2020-05-30
Add Base.Experimental.ReservedStyle: Allow constructing Broadcasted of dictionaries and named tuples

tkf:
This PR lets people _construct_ `Broadcasted`s through the dot-call syntax with dictionaries and named tuples.  This will allow people to experiment with broadcasting with dictionaries and named tuple outside `Base` (see https://github.com/JuliaLang/julia/issues/25904#issuecomment-533309070 for more about the motivation).  Here is a toy example from the docstring:

```julia
julia> using Base.Experimental: ReservedStyle
       using Base.Broadcast: Broadcasted, broadcasted

julia> function aspairs end;

julia> function Broadcast.broadcasted(::typeof(aspairs), bc::Broadcasted{<:ReservedStyle})
           broadcasted(bc.f, collect.(pairs.(bc.args))...)
       end;

julia> aspairs.(tuple.(Dict(:a => 1), Dict(:b => 2)))
1-element Array{Tuple{Pair{Symbol,Int64},Pair{Symbol,Int64}},1}:
 (:a => 1, :b => 2)
```

Note that broadcasting with dictionaries and named tuples still throws the same exception:

```julia
julia> Dict() .+ NamedTuple()
ERROR: ArgumentError: broadcasting over dictionaries and `NamedTuple`s is reserved
Stacktrace:
 [1] copy(::Base.Broadcast.Broadcasted{Base.Broadcast.ReservedStyle,Nothing,typeof(+),Tuple{Dict{Any,Any},NamedTuple{(),Tuple{}}}}) at /home/takafumi/repos/watch/julia/base/broadcast.jl:1305
 [2] materialize(::Base.Broadcast.Broadcasted{Base.Broadcast.ReservedStyle,Nothing,typeof(+),Tuple{Dict{Any,Any},NamedTuple{(),Tuple{}}}}) at /home/takafumi/repos/watch/julia/base/broadcast.jl:836
```

However, this is done at the late materialization stage rather than in `broadcastable` (which is before constructing the `Broadcasted` object).

### DIFF
--- a/base/broadcast.jl
+++ b/base/broadcast.jl
@@ -1282,7 +1282,7 @@ struct ReservedCollection{T}
     value::T
 end
 
-get(r::ReservedCollection) = r.value
+Base.get(r::ReservedCollection) = r.value
 
 broadcastable(x::Union{AbstractDict,NamedTuple}) = ReservedCollection(x)
 BroadcastStyle(::Type{<:ReservedCollection}) = ReservedStyle()
--- a/base/broadcast.jl
+++ b/base/broadcast.jl
@@ -677,10 +677,9 @@ Base.RefValue{String}("hello")
 """
 broadcastable(x::Union{Symbol,AbstractString,Function,UndefInitializer,Nothing,RoundingMode,Missing,Val,Ptr,Regex,Pair}) = Ref(x)
 broadcastable(::Type{T}) where {T} = Ref{Type{T}}(T)
-broadcastable(x::Union{AbstractArray,Number,Ref,Tuple,Broadcasted}) = x
+broadcastable(x::Union{AbstractArray,Number,Ref,Tuple,Broadcasted,AbstractDict, NamedTuple}) = x
 # Default to collecting iterables — which will error for non-iterables
 broadcastable(x) = collect(x)
-broadcastable(::Union{AbstractDict, NamedTuple}) = throw(ArgumentError("broadcasting over dictionaries and `NamedTuple`s is reserved"))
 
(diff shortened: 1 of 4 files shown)


mbauman:
I don't have a chance for a thorough review immediately here, but my first fear is simply that some packages will happen to override the `BroadcastStyle` promotion in a way that slips these dicts into their final broadcasted expression without warning.

tkf:
Oh, that's a good point.  I haven't thought about it.

How about wrapping `AbstractDict` and `NamedTuple` in an "opaque" type

```julia
struct ReservedCollection{T}
    value::T
end

get(r::ReservedCollection) = r.value
```

that does not have `getindex` or `iterate`?

We can then define `BroadcastStyle` for `ReservedCollection`

```julia
broadcastable(x::Union{AbstractDict,NamedTuple}) = ReservedCollection(x)
BroadcastStyle(::Type{<:ReservedCollection}) = ReservedStyle()
```

This would also be useful if you have an iterable container and don't want to define broadcasting semantics immediately because you can just define

```julia
broadcastable(x::MyCollection) = ReservedCollection(x)
```

mbauman:
Ah, that's very clever and I like it much much better.  In fact, we could define an error upon call to:

```Julia
axes(::ReservedCollection) = throw(ArgumentError, "broadcasting over dictionaries and `NamedTuple`s is reserved")
```

tkf:
I thought `axes(::ReservedCollection)` was nice but this didn't cover, e.g., `Dict() .= Dict()`.

DilumAluthge:

PR #36091, opened 2020-05-31
Add passthrough of non-Markdown docs in document trimming

jmert:
The new extended docs feature added in #34226 defined how to trim [Julia's standard] Markdown docs. This breaks the docs retrieved in `PyPlot` since it just forwards through the original Python docs as plain-text (wrapped in `Base.Docs.Text`), and there's no appropriate `trimdocs()` method for that case. There is a fallback definition of the internal `_trimdocs()`, though, so it appears the restriction to specifically `trimdocs(::Markdown.MD, ::Bool)` may have been a mistake, and loosening the type on the first argument fixes the issue.

I've tested this fix on master, but it should be noted that the new v1.5 beta throws the same error when using PyPlot, so this PR should probably be backported (if accepted).

```julia
julia> using PyPlot                                                                                                                                                                                                
                                                                                                                                                                                                                   
help?> plot                                                                                                                                                                                                        
search: plot plot3D plotfile plot_date plot_trisurf plot_surface plot_wireframe triplot subplot boxplot

ERROR: MethodError: no method matching trimdocs(::Text{PyPlot.var"#1#2"{Tuple{PyPlot.LazyHelp}}}, ::Bool)
Closest candidates are:                             
  trimdocs(::Markdown.MD, ::Bool) at /home/justin/devel/ext/julia/usr/share/julia/stdlib/v1.6/REPL/src/docview.jl:111
Stacktrace:                                         
 [1] top-level scope at /home/justin/devel/ext/julia/usr/share/julia/stdlib/v1.6/REPL/src/docview.jl:366

julia> using REPL; Revise.track(REPL)

help?> plot                                         
search: plot plot3D plotfile plot_date plot_trisurf plot_surface plot_wireframe triplot subplot boxplot subplots matplotlib subplot_tool subplot2grid subplots_adjust get_plot_commands stackplot eventplot

Plot y versus x as lines and/or markers.

Call signatures::

    plot([x], y, [fmt], *, data=None, **kwargs)
    plot([x], y, [fmt], [x2], y2, [fmt2], ..., **kwargs)
[...]
```

### DIFF
--- a/stdlib/REPL/src/docview.jl
+++ b/stdlib/REPL/src/docview.jl
@@ -108,6 +108,8 @@ function Markdown.term(io::IO, msg::Message, columns)
     printstyled(io, msg.msg; msg.fmt...)
 end
 
+trimdocs(doc, brief::Bool) = doc
+
 function trimdocs(md::Markdown.MD, brief::Bool)
     brief || return md
     md, trimmed = _trimdocs(md, brief)
--- a/stdlib/REPL/test/repl.jl
+++ b/stdlib/REPL/test/repl.jl
@@ -1093,13 +1093,31 @@ Short docs
 Long docs
 """
 f() = nothing
+@doc text"""
+    f_plain()
+
+Plain text docs
+"""
+f_plain() = nothing
+@doc html"""
+<h1><code>f_html()</code></h1>
+<p>HTML docs.</p>
+"""
+f_html() = nothing
 end # module BriefExtended
+
 buf = IOBuffer()
 md = Base.eval(REPL._helpmode(buf, "$(@__MODULE__).BriefExtended.f"))
 @test length(md.content) == 2 && isa(md.content[2], REPL.Message)
 buf = IOBuffer()
 md = Base.eval(REPL._helpmode(buf, "?$(@__MODULE__).BriefExtended.f"))
 @test length(md.content) == 1 && length(md.content[1].content[1].content) == 4
+buf = IOBuffer()
+txt = Base.eval(REPL._helpmode(buf, "$(@__MODULE__).BriefExtended.f_plain"))
+@test !isempty(sprint(show, txt))
+buf = IOBuffer()
+html = Base.eval(REPL._helpmode(buf, "$(@__MODULE__).BriefExtended.f_html"))
+@test !isempty(sprint(show, html))
 
 # PR #27562
 fake_repl() do stdin_write, stdout_read, repl


jmert:
Maybe @timholy should review since his PR added the feature?

stdlib/REPL/src/docview.jl
```diff
@@ -108,7 +108,7 @@ function Markdown.term(io::IO, msg::Message, columns)
     printstyled(io, msg.msg; msg.fmt...)
 end
 
-function trimdocs(md::Markdown.MD, brief::Bool)
+function trimdocs(md, brief::Bool)
     brief || return md
     md, trimmed = _trimdocs(md, brief)
     if trimmed
```
tkf, line 114: Since `md` is not a `Markdown.MD` anymore, it's a bit misleading to name it `md`.  Also, `if trimmed` block expect `md` to be a `Markdown.MD`.  You wouldn't hit that block since `trimmed` would be always `false` for non-`Markdown.MD` object.  But it is not immediately obvious.  So, somebody might try to refactor this to the one before the patch.  I think a cleaner approach would be to add

```julia
trimdocs(doc, brief::Bool) = doc
```

instead.

jmert: I actually had something similar on my first pass, but changed it to be the absolute minimal change (in the hope that simple is easier to pass muster). I'd be happy to change more, though, if that's preferred.

tkf: Minimal does not mean maintainable.  The current patch has the problems I mentioned.  I think adding an extra method is cleaner.

jmert: Sure — that's why I pinged Tim to ask about original intent. The fallback `_trimdocs` was obviously meant to serve that purpose, so I can just change that to be a level higher in the call chain.

Commit incoming.

stdlib/REPL/src/docview.jl
```diff
@@ -137,8 +139,6 @@ function _trimdocs(md::Markdown.MD, brief::Bool)
     return Markdown.MD(content, md.meta), trimmed
 end
 
-_trimdocs(md, brief::Bool) = md, false
-
```
tkf, line 141: Don't we still need this? `Markdown.MD` can have non-`Markdown.MD` content.

jmert: Oh, can it? I didn't actually run this code — just made the "obvious" (bad assumption...) change. I'm not familiar with the Markdown data structures.

stdlib/REPL/src/docview.jl
tkf: > The fallback `_trimdocs` was obviously meant to serve that purpose

I'm not sure. It'd imagine that Tim would have tested it if so. It could just be an implementation detail for handling recursion.

stdlib/REPL/src/docview.jl
jmert: Yup, you're right. Fixed.

stdlib/REPL/src/docview.jl
jmert: You're right. Following comment was relevant. I'd assumed all Markdown-parsed content was an `MD` type, but there are more types.

tkf (approved):
LGTM. Thanks!

jmert:
Just another friendly bump. This would be nice to get onto the v1.5 branch.

tkf:
I think it's fair to say this is a bug fix and backport is worth considering. So I'm adding the backport labels.

@timholy and @fredrikekre It'd be great if one of you can have a look at it.

Issue #34737, opened 2020-02-12
Proposal: keyword argument broadcasting


jkrumbiegel:
Currently, broadcasting excludes keyword arguments:

```julia
>>> f(x; kw = 3) = x + kw

>>> f.([1, 2, 3]; kw = [4, 5, 6])

ERROR: MethodError: no method matching +(::Int64, ::Array{Int64,1})
Closest candidates are:
  +(::Any, ::Any, ::Any, ::Any...) at operators.jl:529
  +(::T, ::T) where T<:Union{Int128, Int16, Int32, Int64, Int8, UInt128, UInt16, UInt32, UInt64, UInt8} at int.jl:53
```
I don't know what the reasoning for this is, and I couldn't find previous issues about this. I often want to broadcast keyword arguments when I create different objects that offer keyword constructors. In these cases, syntactical convenience is often more important than pure performance to me. It would be a breaking change to enable keyword broadcasting without syntactical changes. However, the following syntax is still available:

```julia
>>> f.([1, 2, 3]; kw .= [4, 5, 6])

ERROR: syntax: invalid keyword argument syntax "kw .= [4, 5, 6]"
```

I propose adding the `.=` keyword syntax in order to pull selected keywords into the broadcasting expression.

c42f:
> ERROR: syntax: invalid keyword argument syntax

Note that this is actually a lowering error rather than a parser error, so you can create a macro which translates this syntax into a broadcasted closure as @mbauman noted on slack:

```julia
((x,y)->f(x; kw=y)).([1, 2, 3], [4, 5, 6])
```

That would allow you to experiment with the syntax and decide whether you like it in practice.

Personally I can see why you'd want this syntax and it has a pleasing consistency with the rest of broadcast notation.

c42f:
To capture some other insightful things @mbauman noted on slack (rather than letting them disappear into the history):

> Unfortunately the syntax isn’t available for `f(x, kw .= y)`; that’ll assign `y` into `kw` and then pass it as a second positional argument.
>
> `f(x; kw .= y)` is available though… but it’s a little fiddly to require `;`s
>
> The other question is what `f(x; kw .= y) .+ 1` does — does that fuse into the `.+`?  Do you consider `f` to be broadcast over the kwargs?

tkf:
[...]

PR #36093, opened 2020-05-31
Add Base.FixKwargs (a la Base.Fix1) for more introspectable broadcasts

tkf:
This PR adds `Base.Broadcast.FixKwargs` which can be used to match and destructure dot-calls with keyword arguments.  I think this is useful to have this as a public API.

### Toy example

```julia
julia> using LazyArrays

julia> function f end
f (generic function with 0 methods)

julia> bc = @~ f.(1, 2; a = 3, b = 4);

julia> bc.f
(::Base.Broadcast.FixKwargs{typeof(f),Base.Iterators.Pairs{Symbol,Int64,Tuple{Symbol,Symbol},NamedTuple{(:a, :b),Tuple{Int64,Int64}}}}) (generic function with 1 method)

julia> bc.f.f === f
true

julia> (; bc.f.kwargs...)
(a = 3, b = 4)

julia> bc.args
(1, 2)
```

### Example: keyword argument broadcasting

Keyword argument broadcasting https://github.com/JuliaLang/julia/issues/34737 can be implemented by users based on public API:

```julia
using Base.Broadcast: Broadcasted, FixKwargs, broadcastable

function withkwcall end
Broadcast.broadcasted(::typeof(withkwcall), x) = _fuse_kwcalls(x)

_fuse_kwcalls(x) = x
_fuse_kwcalls(bc::Broadcasted{Style}) where {Style} =
    _fuse_kwcalls(Broadcasted{Style}, bc.f, bc.args, bc.axes)

_fuse_kwcalls(BC, f, args, axes) = BC(f, map(_fuse_kwcalls, args), axes)
_fuse_kwcalls(BC, f::FixKwargs, args, axes) = BC(
    FlatKWCall{propertynames((; f.kwargs...))}(f.f),
    (map(_fuse_kwcalls, Tuple((; f.kwargs...)))..., map(_fuse_kwcalls, args)...),
    axes,
)

struct FlatKWCall{names,T} <: Function
    f::T
end
FlatKWCall{names}(f::T) where {names,T} = FlatKWCall{names,T}(f)

function (f::FlatKWCall{names})(args...) where {names}
    kw = NamedTuple{names}(args[1:length(names)])
    pos = args[length(names):end]
    return f.f(pos...; kw...)
end
```

```julia
julia> using LazyArrays

julia> f(args...; kwargs...) = (map(=>, 1:length(args), args)..., pairs(kwargs)...)

julia> withkwcall.(f.(1, [2], a = (@~ [1, 2] .+ 1), b = 3))
2-element Array{Tuple{Pair{Int64,Int64},Pair{Int64,Int64},Pair{Int64,Int64},Pair{Symbol,Int64},Pair{Symbol,Int64}},1}:
 (1 => 3, 2 => 1, 3 => 2, :a => 2, :b => 3)
 (1 => 3, 2 => 1, 3 => 2, :a => 3, :b => 3)
```

## Type stability

While I'm at it, this PR also solves the problem with closures capturing types https://github.com/JuliaLang/julia/issues/23618.  Before this PR, this test failed

https://github.com/JuliaLang/julia/blob/2290c4b64dcda05a568f8994483301c13bb3184d/test/broadcast.jl#L270-L276

and it now passes in this PR.

### DIFF
--- a/base/broadcast.jl
+++ b/base/broadcast.jl
@@ -1242,7 +1242,31 @@ macro __dot__(x)
     esc(__dot__(x))
 end
 
-@inline broadcasted_kwsyntax(f, args...; kwargs...) = broadcasted((args...)->f(args...; kwargs...), args...)
+"""
+    Base.Broadcast.FixKwargs{F,K} <: Function
+
+An instance `c` of `FixKwargs` is a function that takes positional
+arguments.  `c(args...)` is equivalent to `c.f(args...; c.kwargs...)`.
+
+The callables of type `FixKwargs` are created when function calls with
+keyword arguments are used in the dot-call syntax.  The dot-call
+sub-expression `f.(args...; kwargs...)` is lowered to a form that is
+equivalent to `broadcasted(FixKwargs(f, kwargs), args...)`.
+
+# Properties
+- `f::F`: a callable
+- `kwargs::K`: the keyword arguments passed to `f`
+"""
+struct FixKwargs{F,K} <: Function
+    f::F
+    kwargs::K
+end
+
+FixKwargs(::Type{T}, kwargs::K) where {T,K} = FixKwargs{Type{T},K}(T, kwargs)
+
+(f::FixKwargs)(args...) = f.f(args...; f.kwargs...)
+
+@inline broadcasted_kwsyntax(f, args...; kwargs...) = broadcasted(FixKwargs(f, kwargs), args...)
 @inline function broadcasted(f, args...)
     args′ = map(broadcastable, args)
     broadcasted(combine_styles(args′...), f, args′...)
--- a/base/broadcast.jl
+++ b/base/broadcast.jl
@@ -1242,31 +1242,8 @@ macro __dot__(x)
     esc(__dot__(x))
 end
 
-"""
-    Base.Broadcast.FixKwargs{F,K} <: Function
-
-An instance `c` of `FixKwargs` is a function that takes positional
-arguments.  `c(args...)` is equivalent to `c.f(args...; c.kwargs...)`.
-
-The callables of type `FixKwargs` are created when function calls with
-keyword arguments are used in the dot-call syntax.  The dot-call
-sub-expression `f.(args...; kwargs...)` is lowered to a form that is
-equivalent to `broadcasted(FixKwargs(f, kwargs), args...)`.
-
-# Properties
-- `f::F`: a callable
-- `kwargs::K`: the keyword arguments passed to `f`
-"""
-struct FixKwargs{F,K} <: Function
-    f::F
-    kwargs::K
-end
-
-FixKwargs(::Type{T}, kwargs::K) where {T,K} = FixKwargs{Type{T},K}(T, kwargs)
-
-(f::FixKwargs)(args...) = f.f(args...; f.kwargs...)
-
-@inline broadcasted_kwsyntax(f, args...; kwargs...) = broadcasted(FixKwargs(f, kwargs), args...)
+@inline broadcasted_kwsyntax(f, args...; kwargs...) =
+    broadcasted(Base.FixKwargs(f, kwargs), args...)
 @inline function broadcasted(f, args...)
     args′ = map(broadcastable, args)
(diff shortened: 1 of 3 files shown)


base/broadcast.jl
```diff
@@ -1242,7 +1242,31 @@ macro __dot__(x)
     esc(__dot__(x))
 end
 
-@inline broadcasted_kwsyntax(f, args...; kwargs...) = broadcasted((args...)->f(args...; kwargs...), args...)
+"""
+    Base.Broadcast.FixKwargs{F,K} <: Function
```
tkf, line 1246: Is it better to move this to `base/operators.jl` where `Fix1` and `Fix2` live?

mbauman: Yes, I like that idea.

base/broadcast.jl
```diff
@@ -1242,7 +1242,31 @@ macro __dot__(x)
...
+# Properties
+- `f::F`: a callable
+- `kwargs::K`: the keyword arguments passed to `f`
+"""
+struct FixKwargs{F,K} <: Function
+    f::F
+    kwargs::K
+end
+
+FixKwargs(::Type{T}, kwargs::K) where {T,K} = FixKwargs{Type{T},K}(T, kwargs)
```
mbauman, line 1265: Maybe also support a constructor with the kwargs actually being passed as kwargs?

base/broadcast.jl
```diff
@@ -1242,7 +1242,31 @@ macro __dot__(x)
     esc(__dot__(x))
 end
 
-@inline broadcasted_kwsyntax(f, args...; kwargs...) = broadcasted((args...)->f(args...; kwargs...), args...)
+"""
+    Base.Broadcast.FixKwargs{F,K} <: Function
+
+An instance `c` of `FixKwargs` is a function that takes positional
```
mbauman, line 1248: Maybe mirror the definitions from `FixN`?  I often find symmetrical docs help me understand the behaviors better.

>     Fix1(f, x)
>
>  A type representing a partially-applied version of the two-argument function
>  `f`, with the first argument fixed to the value "x". In other words, `Fix1(f, x)`
> behaves similarly to `y->f(x, y)`.

Thus:

>     FixKwargs(f; kwargs...)
>
>  A type representing a partially-applied version of the keyword argument function
>  `f`, with all keyword arguments fixed to those passed. In other words, `FixKwarg(f; kws...)`
> behaves similarly to `(args...)->f(args...; kws...)`.

base/broadcast.jl
mbauman: How does method specialization work with functions passed to the constructors here?  Do we also need a constructor that explicitly specializes on `Function`s like this?

tkf: I don't think we need a specialization for `Function` because we automatically gets it:

```julia
julia> typeof(Base.FixKwargs(reduce; init = 0))
Base.FixKwargs{typeof(reduce),Base.Iterators.Pairs{Symbol,Int64,Tuple{Symbol},NamedTuple{(:init,),Tuple{Int64}}}}
```

(see `typeof(reduce)`)

IIUC `Type` is special here because `julia`'s type system "leaks" the representation of types rather than decoupling it from subtyping.

tkf:
Thanks, I reflected the comments.

tkf:
It'd be nice if https://github.com/JuliaLang/julia/pull/36094 can be merged first so that `FixKwargs` can be listed in the documentation with `Fix1` and `Fix2`.

mbauman:
I don't think order matters much — it's just one line of code in whichever PR goes second and it totally makes sense to include that line in either PR.

I'm 👍 with the broadcasting usage, changing the title to potentially attract a reviewer that cares more about our pseudo-currying functionality.

mbauman (approved):

tkf:
It's just that @StefanKarpinski is a bit hesitant to add `Fix1`/`Fix2` to the documentation.  So I think it'd be bad to include `FixKwargs` before it's decided that it's OK to list `Fix1`/`Fix2` in the documentation.

DilumAluthge:

PR #30989, opened 2019-02-07
fix #14574, `cp` on files >2GB

JeffBezanson:
There may be a better way to do this, but this is a silly bug to have so here is a simple fix for now.

Also, any ideas on how to test this? Creating multiple GB of files seems like the kind of thing that could disagree with CI systems.

Issue #30723, opened 2019-01-15
IOError from cp with large file on Windows


davidanthoff:
This is the error:
````
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.0.3 (2018-12-18)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> cp("uniform_data_string.csv", "foo2.csv")
ERROR: IOError: sendfile: Unknown system error -756883004 (Unknown system error -756883004)
Stacktrace:
 [1] uv_error at .\libuv.jl:85 [inlined]
 [2] sendfile at .\filesystem.jl:111 [inlined]
 [3] sendfile(::String, ::String) at .\file.jl:749
 [4] #cp#12(::Bool, ::Bool, ::Function, ::String, ::String) at .\file.jl:336
 [5] cp(::String, ::String) at .\file.jl:330
 [6] top-level scope at none:0
````
I get this after I have rebooted my system, so I don't think this is related to some lingering stuff.

The file I'm trying to copy is relatively large:
````
-a----        1/15/2019  10:52 AM     3538084292 uniform_data_string.csv
````
The code actually does copy the file, as far as I can tell.

I can successfully copy files that are smaller than 2GB, so maybe there is something somewhere that makes this fail for files larger than 2GB?

The same code works on Linux and Mac.

**Comment by nalimilan** [comment]
Related: https://github.com/JuliaLang/julia/issues/14574.

**Comment by JockLawrie** [comment]
Hi there,
I am also experiencing this on Windows 10 with a 2.3GB file.
Using Julia 1.4.1.

**Comment by StefanKarpinski** [comment]
Seems like we should really do something about this. @vtjnash, I assume this is a libuv issue under the hood? Any way we can work around it and/or fix this in libuv?

**Comment by yakir12** [comment]
I have the same issue but on a mac (let me know if I should open/post on a different issue).
```
$ sw_vers
ProductName:    Mac OS X
ProductVersion: 10.15.4
BuildVersion:   19E287
$ uname -a
[...]

PR #36095, opened 2020-06-01
Test copying a large file

jaakkor2:
Test issues
* https://github.com/JuliaLang/julia/issues/14574 (fixed)
* https://github.com/JuliaLang/julia/issues/30723 (copy failure on Windows when filesize is >2 GB).

https://github.com/JuliaLang/julia/pull/30989 warns that CI might not like big files, let's see.

### DIFF
--- a/test/filesystem.jl
+++ b/test/filesystem.jl
@@ -32,4 +32,14 @@ mktempdir() do dir
   @test res == UInt8[text..., (i > 20 for i in (length(text) + 1):length(res))...]
   close(file)
 
+  # test copying a big file, issues #14574, #30723
+  filename_big = tempname(dir)
+  filename_big_copy = tempname(dir)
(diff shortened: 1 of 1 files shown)


KristofferC:
Note that testing this was explicitly mentioned in https://github.com/JuliaLang/julia/pull/30989#issue-250992382 and decided against. Just pointing it out for reference.

imciner2:
Please don't test it this way. Testing using multi GB files would not be good for low memory bandwidth systems (such as some ARM boards) since it would make the already painfully-slow unit testing even slower, and could possibly lead to out of space errors on smaller devices because there are now multiple large files being created.

jaakkor2:
Testing for https://github.com/JuliaLang/julia/issues/30723 needed, but based on comments this is not the way.

PR #36097, opened 2020-06-01
Improve sentence in the Performance Tips section

natema:
This commit improves an ambiguous sentence in the [Performance Tips](https://docs.julialang.org/en/v1/manual/performance-tips/) page.

In the section [Access arrays in memory order, along columns](https://docs.julialang.org/en/v1/manual/performance-tips/#More-dots:-Fuse-vectorized-operations-1), there is the final sentence
> (Of course, if you just do `f.(x)` then it is as fast as `fdot(x)` in this example, but in many contexts it is more convenient to just sprinkle some dots in your expressions rather than defining a separate function for each vectorized operation.)

This is a bit confusing.  What we want to say is that
- using the dot notation is faster in the given example, but in general it will not be the case,
- even though in general the dot notation is not the most performant approach, in several contexts it is more convenient.

### DIFF
--- a/doc/src/manual/performance-tips.md
+++ b/doc/src/manual/performance-tips.md
@@ -1013,11 +1013,9 @@ julia> @time f.(x);
 
 That is, `fdot(x)` is ten times faster and allocates 1/6 the
 memory of `f(x)`, because each `*` and `+` operation in `f(x)` allocates
-a new temporary array and executes in a separate loop. (Of course,
-if you just do `f.(x)` then it is as fast as `fdot(x)` in this
-example, but in many contexts it is more convenient to just sprinkle
-some dots in your expressions rather than defining a separate function
-for each vectorized operation.)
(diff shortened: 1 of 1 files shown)


mbauman:
> * using the dot notation is faster in the given example, but in general it will not be the case,
> * even though in general the dot notation is not the most performant approach, in several contexts it is more convenient.

This is definitely not how we want to frame this. In general, it should be the same speed or faster — especially as it is then able to fuse with surrounding expressions. If you have specific examples where that's not the case, please file bugs.

natema:
Thanks for the feedback. I get that my rewriting of the sentence doesn't go in the right direction. It is still unclear to me, though, how the original sentence serves the *intended* message. I guess that [I should ask in the forum for clarifications](https://discourse.julialang.org/t/unclear-sentence-in-performance-tips-doc-page/40758?u=mathematics) regarding my confusion.

PR #33553, opened 2019-10-14
give IR forms for defining types valid linear structure

JeffBezanson:
The type-defining IR forms (`struct_type` etc.) are currently very special: they violate the usual linear structure rules for IR, and evaluate their arguments in a non-standard way intertwined with some global effects. This PR tries to clean that up, removing the special cases and hopefully eventually leading to fixing #33183.

The new type-defining process here is broken into three steps at the IR level:
1. Create a new partially-initialized type object and bind it to a local variable.
2. Evaluate the supertype expression and set it.
3. Evaluate and set field types, fix up partial types.
4. Assign the type globally, doing the special check for valid redefinitions.

Uses only function calls (unfortunately requiring several new builtins though).

fixes #35416

PR #32658, opened 2019-07-23
Mutually recursive field types, the easy version

Keno:
This is an alternative to #32581, that is easier to implement, but
more restrictivive. Here, the forward declaration must contain
everything except for the field types:
```
incomplete type Foo{Bar} <: Baz; end
```
with it being an error to deviate from this specification when
completing the type
```
struct Foo <: Baz; end # Error: Missing type parameter
struct Foo{A, B} <: Baz; end # Error: Too many type parameters
struct Foo{Qux} <: Baz; end # Error: Name of type parameter doesn't match
struct Foo{Bar<:Int64} <: Baz; end # Error: Bounds of type parameter don't match
struct Foo{Bar}; end; # Error supertype dosesn't match
```
This is easier because this way we have enough information for subtyping
and type application and therefor do not need to delay this until
the entire dependency graph is complete as we did in #32581.
Of course this also means that we don't get the union feature that
was requested in #269:
```
incomplete type U; end
struct A; x::U; end
struct B; x::U; end
U = Union{A, B}; #ERROR
```
However, it could of course be emulated by wrapping the union type:
```
struct U
   data::Union{A, B}
end
```

However, given the simplicity of this change and the difficulty of #32581,
this seems like the way to go for the moment. We may want to revisit all
this if we ever want to do computed field types, which will encounter similar
challenges as #32581.

Fixes #269.

PR #36111, opened 2020-06-01
fix #36104, hack to allow accessing incomplete struct via dot syntax

### LINKED ISSUE #36104: Referring to `Foo.F` inside struct `F` defined in module `Foo` now errors with UndefVar error
```jl
module Foo
    struct F
        v::Foo.F
    end
end
```

Works in 1.4, errors in 1.5 with

```
ERROR: UndefVarError: F not defined
```

From PkgEval on 1.5 (https://github.com/JuliaCI/NanosoldierReports/blob/90885ffd4005a91d0dc4150b8c1453f2f5948eb0/pkgeval/by_hash/194ddda_vs_b5f13c3/logs/Persa/1.5.0-DEV-f11a73bd9c.log).

JeffBezanson:
This isn't perfect, but hopefully it will be good enough to catch all practical cases. In 1.6 we should officially deprecate accessing an incomplete struct via its global binding.

fix #36104

### DIFF
--- a/base/boot.jl
+++ b/base/boot.jl
@@ -741,6 +741,14 @@ Unsigned(x::Union{Float32, Float64, Bool}) = UInt(x)
 Integer(x::Integer) = x
 Integer(x::Union{Float32, Float64}) = Int(x)
 
+# During definition of struct type `B`, if an `A.B` expression refers to
+# the eventual global name of the struct, then return the partially-initialized
+# type object.
+# TODO: remove. This is a shim for backwards compatibility.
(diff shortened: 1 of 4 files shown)


vtjnash:
I think I merely commented this in person during review of https://github.com/JuliaLang/julia/pull/33553, but sometime soon in the roadmap, we're likely going to need to make it completely fine to access these incomplete structs via their global bindings at this stage of type construction (field type evaluation), per https://github.com/JuliaLang/julia/pull/32658.

I think we could just do that part now (revert to evaluating these field types in normal / global scope), and defer adding some of the error checking from that PR later?

PR #36114, opened 2020-06-02
Fix doc: pressing ctrl-c doesn't return -1 in a RadioMenu [ci skip]

laborg:
Doc clarification of what happens in a `TerminalMenu` when pressing ctrl-c:
Julia 1.0.5, 1.4.2 and 1.5-beta throw an `InterruptException` at least on Linux and Windows .


laborg:
Just seen Tim Holy #36116, which deals with this.

PR #36118, opened 2020-06-02
Example of finalizer registered within constructor

mzaffalon:


mzaffalon:
I am open to suggestions for a better example.

base/gcutils.jl
```diff
@@ -25,6 +25,22 @@ finalizer(my_mutable_struct) do x
...
+# Example
+```julia
+
+mutable struct my_mutable_struct
+    bar
+    function my_mutable_struct(bar)
+        x = new(bar)
+        f(t) = @async println("Finalizing \$x.")
+        finalizer(f, x)
+    end
```
ararslan, line 42:
```suggestion
mutable struct MyMutableStruct
    bar
    function MyMutableStruct(bar)
        x = new(bar)
        f(t) = @async println("Finalizing \$x.")
        finalizer(f, x)
    end
end
```
Also why `@async`?

mzaffalon: From the [docs](https://docs.julialang.org/en/v1.6-dev/base/base/#Base.finalizer): "`f` must not cause a task switch, which excludes most I/O operations such as println. Using the `@async` macro (to defer context switching to outside of the finalizer)..."

ararslan: Ah, I see, thanks for the reference. Code suggestion still applies though. :)

base/gcutils.jl
```diff
@@ -25,6 +25,22 @@ finalizer(my_mutable_struct) do x
...
+it.
+
+# Example
+```julia
+
+mutable struct my_mutable_struct
+    bar
+    function my_mutable_struct(bar)
+        x = new(bar)
+        f(t) = @async println("Finalizing \$x.")
```
thofma, line 40:
```suggestion
        f(t) = @async println("Finalizing \$t.")
```

base/gcutils.jl
```diff
@@ -25,6 +25,22 @@ finalizer(my_mutable_struct) do x
     ccall(:jl_safe_printf, Cvoid, (Cstring, Cstring), "Finalizing %s.", repr(x))
 end
 ```
+
+A finalizer can be registered at object constuction. In the following example note that
```
thofma, line 29:
```suggestion
A finalizer can be registered at object construction. In the following example note that
```

base/gcutils.jl
```diff
@@ -25,6 +25,22 @@ finalizer(my_mutable_struct) do x
...
+it is not necessary to returned the newly created object because the finalizer returns
```
thofma, line 30:
```suggestion
it is not necessary to return the newly created object because the finalizer returns
```

mzaffalon:
@ararslan @thofma Thanks for the suggestions. I have changed the explanation and if you approve the new version, I will squash rebase the commits.

Issue #21816, opened 2017-05-12
Redefining recursive type fails


cstjean:
Redefining non-parametric, recursive types fails:
```julia
struct Trace
    traces::Vector{Trace}
end
struct Trace
    traces::Vector{Trace}
end
> invalid redefinition of constant Trace
```

andyferris:
Seems genuinely ambiguous which `Trace` is referred to in the body of the redefinition - but I guess if we want to support this then the recursive form needs to be assumed.

StefanKarpinski:
I don't think it's actually ambiguous since only one thing can have the name `Trace` at the end of the day, and the "redefinition" can't actually change the identity of that thing.

yashbha:
What is the solution for this one?

StefanKarpinski:
You can wrap the definition in a module and reload the module, which lets you do this. Redefining a recursive definition in the same module does not yet work.

michaelkyu:
I'd like to throw in my support for this issue. Being able to redefine structs without getting this error would be useful in Jupyter notebooks, where a user is often prototyping code and changing definitions. I am running into this issue myself.

StefanKarpinski:
It would certainly be nice—I don't think there's any opposition. It's just a low priority annoyance that no one has had time or sufficient motivation to work on. If anyone wants to try, they're welcome to!

pedroborgesmelo:
Need this here.

PR #36121, opened 2020-06-02
fix #36104, assign global name during type definitions

### LINKED ISSUE #36104: Referring to `Foo.F` inside struct `F` defined in module `Foo` now errors with UndefVar error

JeffBezanson:
Alternative to #36111. Handling re-definitions is the tricky part. I realized one thing we did before was reset the global binding to undefined if an error happens during type definition, but that is invalid, so we don't do it anymore (with or without this PR). Hopefully that is the only remaining difference with v1.4.

This first checks if the new type object might be compatible with an existing one, and if so throws away the new one and uses the old type and parameters. Then we check that each step (supertype and field types) remains compatible as it runs.

fixes #36104

### DIFF
--- a/test/core.jl
+++ b/test/core.jl
@@ -7221,3 +7221,20 @@ struct AVL35416{K,V}
     avl:: Union{Nothing,Node35416{AVL35416{K,V},<:K,<:V}}
 end
 @test AVL35416(Node35416{AVL35416{Integer,AbstractString},Int,String}()) isa AVL35416{Integer,AbstractString}
+
+# issue #36104
+module M36104
+struct T36104
+    v::Vector{M36104.T36104}
+end
+struct T36104   # check that redefining it works, issue #21816
+    v::Vector{T36104}
+end
+end
+@test fieldtypes(M36104.T36104) == (Vector{M36104.T36104},)
+@test_throws ErrorException("expected") @eval(struct X36104; x::error("expected"); end)
+@test @isdefined(X36104)
+struct X36104; x::Int; end
+@test fieldtypes(X36104) == (Int,)
+primitive type P36104 8 end
+@test_throws ErrorException("invalid redefinition of constant P36104") @eval(primitive type P36104 16 end)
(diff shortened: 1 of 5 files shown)


test/core.jl
```diff
@@ -7221,3 +7221,17 @@ struct AVL35416{K,V}
...
 @test AVL35416(Node35416{AVL35416{Integer,AbstractString},Int,String}()) isa AVL35416{Integer,AbstractString}
+
+# issue #36104
+module M36104
+struct T36104
+    v::Vector{M36104.T36104}
+end
+end
+@test fieldtypes(M36104.T36104) == (Vector{M36104.T36104},)
+@test_throws ErrorException @eval(struct X36104; x::error(); end)
```
vtjnash, line 7232:
```suggestion
@test_throws ErrorException("expected") @eval(struct X36104; x::error("expected"); end)
```

vtjnash (approved):

JeffBezanson:
Well, unsurprisingly, incomplete types throw a wrench into the works. We have some in the tests just to test that case; hopefully it will not really happen outside of that. I think I will return `Any` from `supertype` if the supertype is not set, and give an error from `fieldtypes`?

vtjnash:
For soundness, the correct answer for `supertype` seems would have to be `Union{}`, not `Any`. I think #32658 was adding to the work required to handle incomplete `fieldtypes`

I don't think this PR needs to be changing our handling of `supertype`, as the current proposal (in #32658) is to not change it.

JeffBezanson:
> I don't think this PR needs to be changing our handling of `supertype`

But then it breaks callers, like `subtypes`.

JeffBezanson:
This should be ok now. As a bonus, also fixes #21816!
All it sacrifices is that you can't refer to the new type via its module within the supertype expression, e.g.
```
struct Foo <: AbstractVector{Mod.Foo}
```
but I expect that to be extremely rare.

PR #36122, opened 2020-06-02
fix #36108, printing invalid numeric juxtapositions

### LINKED ISSUE #36108: Printing of multiplication and vs. UInt16 etc.
This is a bit odd:
```julia
julia> :(2 * a)
:(2a)

julia> 2a
ERROR: UndefVarError: a not defined
Stacktrace:
 [1] top-level scope at REPL[6]:1

julia> :(0x0 * a)
:(0x00a)

julia> 0x00a == 10
true
```
From here: https://discourse.julialang.org/t/not-clear-what-0-01f0x-is-in-the-flux-docs/40553/4

JeffBezanson:
fix #36108

### DIFF
--- a/base/show.jl
+++ b/base/show.jl
@@ -1342,7 +1342,8 @@ function show_unquoted(io::IO, ex::Expr, indent::Int, prec::Int, quote_level::In
 
         # scalar multiplication (i.e. "100x")
         elseif (func === :* &&
-            length(func_args)==2 && isa(func_args[1], Real) && isa(func_args[2], Symbol))
+            length(func_args) == 2 && isa(func_args[1], Union{Int, Int64, Float32, Float64}) &&
+            isa(func_args[2], Symbol) && !in(string(func_args[2])[1], ('e', 'E', 'f')))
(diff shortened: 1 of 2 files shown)


base/show.jl
```diff
@@ -1342,7 +1342,8 @@ function show_unquoted(io::IO, ex::Expr, indent::Int, prec::Int, quote_level::In
 
         # scalar multiplication (i.e. "100x")
         elseif (func === :* &&
-            length(func_args)==2 && isa(func_args[1], Real) && isa(func_args[2], Symbol))
+            length(func_args) == 2 && isa(func_args[1], Real) && isa(func_args[2], Symbol)) &&
+            !isa(func_args[1], Unsigned) && !in(string(func_args[2])[1], ('e','E','f'))
```
vtjnash, line 1346:
```suggestion
            length(func_args) == 2 && isa(func_args[1], Union{Int, Float32, Float64}) && isa(func_args[2], Symbol)) &&
            !in(string(func_args[2])[1], ('e' ,'E' ,'f'))
```

PR #36123, opened 2020-06-02
Update build dependencies

ViralBShah:
C/C++ compiler based on LLVM recommendation: https://llvm.org/docs/GettingStarted.html#host-c-toolchain-both-compiler-and-standard-library
Powershell based on: https://julialang.org/downloads/platform/#windows

Remove version numbers for the libraries in deps since they are quite out of date, and change frequently. Instead point to Versions.make.

[ci skip]

### DIFF
--- a/doc/build/build.md
+++ b/doc/build/build.md
@@ -144,7 +144,7 @@ Notes for various architectures:
 Building Julia requires that the following software be installed:
 
 - **[GNU make]**                — building dependencies.
-- **[gcc & g++][gcc]** (>= 4.7) or **[Clang][clang]** (>= 3.1, Xcode 4.3.3 on macOS) — compiling and linking C, C++.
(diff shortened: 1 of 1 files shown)


Keno:
Broke CI due to whitespace ... and that's why we don't do CI skip in PRs that are supposed to be merged ;). Anyway, I emergency-pushed a fix.

ViralBShah:
Maybe `ci skip` should not skip whitespace ci. Anyways, until we have that, I guess one has to run the full CI.

ViralBShah:
> Broke CI due to whitespace ... and that's why we don't do CI skip in PRs that are supposed to be merged ;). Anyway, I emergency-pushed a fix.

Thanks

Keno:
> Maybe `ci skip` should not skip whitespace ci.

Yeah, good point. @staticfloat is that possible to do?

staticfloat:
Fixed in https://github.com/JuliaCI/julia-buildbot/pull/158, confirmed in https://github.com/JuliaLang/julia/pull/36439.

PR #36124, opened 2020-06-02
Revise-test: make sure all workers load Revise

timholy:
This fixes the errors
```julia
julia> Base.runtests(["REPL", "cartesian"]; revise=true)
ERROR: LoadError: On worker 2:
UndefVarError: Revise not defined
```

and then later

```julia
julia> Base.runtests(["REPL", "cartesian"]; revise=true)
ERROR: LoadError: On worker 2:
UndefVarError: STDLIBS not defined
```

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -15,6 +15,12 @@ tests = unique(tests)
 
 if use_revise
     using Revise
+    # Remote-eval the following to initialize Revise in workers
+    const revise_init_expr = quote
+        using Revise
+        const STDLIBS = $STDLIBS
+        revise_trackall()
+    end
 end
 
 const max_worker_rss = if haskey(ENV, "JULIA_TEST_MAXRSS_MB")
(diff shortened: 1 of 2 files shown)


test/runtests.jl
```diff
@@ -85,10 +81,11 @@ cd(@__DIR__) do
 
     if use_revise
         @everywhere begin
+            @eval using Revise
             Revise.track(Core.Compiler)
             Revise.track(Base)
             for (id, mod) in Base.loaded_modules
-                if id.name in STDLIBS
+                if id.name in filter!(x -> isfile(joinpath(Sys.STDLIB, x, "src", "$(x).jl")), readdir(Sys.STDLIB))
```
Keno, line 88: Do the readdir only once? As written it's done `nworkers * length(Base.loaded_modules)` times.

Keno: Actually, I think it would be better to just serialize the STDLIBS we already have here rather than re-compute it. Doesn't distributed allow something like `$STDLIBS` to refer to the local one?

timholy: Should be better now. Looks like we weren't using Revise on restarted workers, so this fixes that too.

timholy:
Trying to trigger tests...

PR #33422, opened 2019-09-30
Internally use libuv for mkdir on all platforms

musm:

PR #36126, opened 2020-06-02
Fix mkpath error handling

simonbyrne:
The error thrown by `mkdir` when the directory already exists was changed in #33422.

Unfortunately, this is hard to test, though we came across it on a cluster with a shared file system.

cc: @jakebolewski @kpamnany

### DIFF
--- a/base/file.jl
+++ b/base/file.jl
@@ -228,7 +228,7 @@ function mkpath(path::AbstractString; mode::Integer = 0o777)
     catch err
         # If there is a problem with making the directory, but the directory
         # does in fact exist, then ignore the error. Else re-throw it.
-        if !isa(err, SystemError) || !isdir(path)
+        if !isa(err, IOError) || !isdir(path)
             rethrow()
         end
     end


base/file.jl
```diff
@@ -228,7 +228,7 @@ function mkpath(path::AbstractString; mode::Integer = 0o777)
     catch err
         # If there is a problem with making the directory, but the directory
         # does in fact exist, then ignore the error. Else re-throw it.
-        if !isa(err, SystemError) || !isdir(path)
+        if !isa(err, IOError) || !isdir(path)
```
simonbyrne, line 231: Would it be better to do?
```suggestion
        if !(isa(err, IOError) && err.code == UV_EEXIST)
```

vtjnash: I assume it doesn't matter because of the `isdir` test?
```julia
julia> mkpath("/nonexistant")
ERROR: IOError: mkdir: permission denied (EACCES)
Stacktrace:
 [1] uv_error at ./libuv.jl:97 [inlined]
 [2] mkdir(::String; mode::UInt16) at ./file.jl:177
 [3] mkpath(::String; mode::UInt16) at ./file.jl:227
 [4] mkpath(::String) at ./file.jl:222
 [5] top-level scope at REPL[3]:1
```

vtjnash (approved):

base/file.jl
simonbyrne: Arguably in that case you would want the error to be propagated?

PR #36127, opened 2020-06-02
Fix a few AbstractInterpreter nits

Keno:
Addresses post-commit review by @JeffBezanson in #33955.

### DIFF
--- a/base/compiler/types.jl
+++ b/base/compiler/types.jl
@@ -5,7 +5,7 @@
 
 An abstract base class that allows multiple dispatch to determine the method of
 executing Julia code.  The native Julia LLVM pipeline is enabled by using the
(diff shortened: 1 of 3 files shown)


base/compiler/types.jl
```diff
@@ -5,7 +5,7 @@
 
 An abstract base class that allows multiple dispatch to determine the method of
 executing Julia code.  The native Julia LLVM pipeline is enabled by using the
-`TypeInference` concrete instantiatoin of this abstract class, others can be
+`NativeInterpreter` concrete instantiatoin of this abstract class, others can be
```
jonas-schulze, line 8:
```suggestion
`NativeInterpreter` concrete instantiation of this abstract class, others can be
```

PR #36128, opened 2020-06-03
Add to documentation of two-argument atan

ambyld:
Fix agrees with the listed reference and the function behavior.

### DIFF
--- a/base/math.jl
+++ b/base/math.jl
@@ -343,7 +343,7 @@ For one argument, this is the angle in radians between the positive *x*-axis and
 (1, *y*), returning a value in the interval ``[-\\pi/2, \\pi/2]``.
 
 For two arguments, this is the angle in radians between the positive *x*-axis and the
-point (*x*, *y*), returning a value in the interval ``[-\\pi, \\pi]``. This corresponds to a
+point (*x*, *y*), returning a value in the interval ``(-\\pi, \\pi]``. This corresponds to a
 standard [`atan2`](https://en.wikipedia.org/wiki/Atan2) function.
 """
 atan(x::Number)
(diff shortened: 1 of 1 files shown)


simonbyrne:
Well, not really... there are a couple of interpretations one could use here:

a. Since pi isn't exactly representable in floating point, then the result has to be in the open interval (-pi,pi).

b. Alternatively, you could view it as "what is the value it would be before rounding", and then since IEEE arithmetic supports signed 0, and [for various complicated but mathematically elegant reasons](https://people.freebsd.org/~das/kahan86branch.pdf), `atan(-0.0,x)` is defined to be `-pi` when `x < 0`, we would have [-pi,pi]. It would be good to have this noted in the docs.

c. Or, one could ask what would happen if both the result and pi were rounded, then you would have:
```
julia> -Float64(pi) == atan(-0.0,-1.0)
true
```
which again would lead to [-pi,pi].

ambyld:
Sure, pi isn't representable in floating point, but the point of this fix is to answer the question of whether atan(0,1) = pi or atan(0,1) = -pi (or whichever floating point numbers are closest to pi and -pi) without having to test it manually.

Mathematically both cannot be true, which is why the reference https://en.wikipedia.org/wiki/Atan2 chooses (-pi,pi] as the image of atan2. And indeed atan(0,1) = pi in Julia.

ambyld:
Interesting, I didn't realize `-Float64(pi) == atan(-0.0,-1.0)`. Thank you for the reference - there is much more to this than I expected.

simonbyrne:
At some point we should document all the branch cut behavior, but this is clearly an improvement.

PR #36131, opened 2020-06-03
Add suport for outer nested loops with the @threads macro

ghost:
This is my first time committing anything to open source so please be understanding of any mistakes I make, and feel free to point out what I'm doing wrong.

Recently I needed to use @threads in a nested loop, and after realizing someone else might find it useful I decided to to commit it here.

Compared to the standard @threads the function is about 3.8-3.9% slower when it is empty and does no iterations and about 1.8-2.0% per iteration, again, when empty, so whenever dealing with big loops, above the thread count, you should still just split the loop, however in my scenario the macro cut down the time by more than 60% no mater where you placed the @threads macros (4x4x4 loop, JULIA_NUM_THREADS = 10)

Since I'm not experienced I don't want to deal with the maintainability aspect, if the functions should be merged i definitely can help however, I can't decide myself.

Again, please be understanding if I'm doing something wrong and thanks in advance.

### DIFF
--- a/base/threadingconstructs.jl
+++ b/base/threadingconstructs.jl
@@ -55,12 +55,12 @@ function _threadsfor(a, body, schedule)
     initends = ((x, y) -> :($x = lastindex($y) ) ).(ends, ranges)
     initstartingindecies = ((x, y) -> :($x += $y)).(values, starts)
     #updates to variables
-    updatevalues = [:(  ( $(esc(lidx[x])) =  @inbounds $(ranges[x])[$(values[x])] ) ) for x in 1:rangelen]
+    updatevalues = [:( $(esc(lidx[x])) = @inbounds $(ranges[x])[$(values[x])] ) for x in 1:rangelen]
     initvalues = [ :($(values[1]) = (tid - 1) * len + min(tid - 1, rem))] #initial calculation of the first dimension
     for x in 2:rangelen
         push!(initvalues, :(($(values[x]), $(values[x - 1])) = divrem($(values[x - 1]), length($(ranges[x - 1]))))) #initial calculation for every dimension
     end
-    checks = :() #checks are not necesseary when there is only one variable
+    checks = :() #checks are not necessary when there is only one variable
     for x in rangelen:-1:2
         checks =
         quote
@@ -72,13 +72,12 @@ function _threadsfor(a, body, schedule)
             end
         end
     end
-
     quote
         local threadsfor_fun
         let range = [$(esc.(rang)...)]
             totallength = reduce(*, length.(range))
             function threadsfor_fun(onethread=false)
-                #Load into local variables
+                #load into local variables
                 $(initranges...)
                 tlen = totallength
                 #calculate the iteration length for the current thread
@@ -89,7 +88,7 @@ function _threadsfor(a, body, schedule)
                     tid = threadid()
                     len, rem = divrem(tlen, nthreads())
                 end
-                if len == 0 && rem < tid #no iterations abvilable for this thread
+                if len == 0 && rem < tid #no iterations available for this thread
                     return
                 end
                 #inits
@@ -102,25 +101,22 @@ function _threadsfor(a, body, schedule)
                     len += 1
                 end
                 $(updatevalues...) #set all variables to their initial values, they will be updated later in the loop
-                $(values[1]) -= 1 #reduce code duplicaiton by "ommiting" the first increment
+                $(values[begin]) -=1 #reduce code duplication by "omitting" the first increment
                 for i in 1:len
-
                     $(values[1]) += 1
                     $checks
                     $(updatevalues[1])
-
                     $(esc(body))
                 end
             end
         end
-        if threadid() != 1
-            $(
-            if schedule === :static
-                :(error("`@threads :static` can only be used from thread 1 and not nested"))
-            else
-                # only use threads when called from thread 1, outside @threads
-                :(Base.invokelatest(threadsfor_fun, true))
-            end)
+        if threadid() != 1 || ccall(:jl_in_threaded_region, Cint, ()) != 0
+            $(if schedule === :static
+              :(error("`@threads :static` can only be used from thread 1 and not nested"))
+              else
+              # only use threads when called from thread 1, outside @threads
+              :(Base.invokelatest(threadsfor_fun, true))
+              end)
         else
             threading_run(threadsfor_fun)
         end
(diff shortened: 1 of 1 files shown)


base/threadingconstructs.jl
```diff
@@ -38,6 +38,87 @@ function threading_run(func)
...
+        if threadid() != 1
+            $(
+            if schedule === :static
+                :(error("`@threads :static` can only be used from thread 1 and not nested"))
+            else
+                # only use threads when called from thread 1, outside @threads
+                :(Base.invokelatest(threadsfor_fun, true))
+            end)
+        else
+            ccall(:jl_threading_run, Cvoid, (Any,), threadsfor_fun)
```
JeffBezanson, line 114: Note this has been replaced by `threading_run` above.

JeffBezanson:
Welcome, @Giufin , this is very good work!

Note you should put backticks around `@threads` to avoid pinging that github user.

This should be written to take advantage of the fact that the number of ranges is known statically, at macro-expansion time. That way there is no need to allocate arrays and loop over indices, and it should handle ranges of different types better (though that case may be rare). Basically, use a variable for each range and index instead of an array holding all of them. That way the 1-d case would reduce to the same code we have now, and so this could be combined with the current code.

tkf:
FWIW I think a cleaner approach is to use `Iterators.product` as an intermediate representation of nested loops and then write a threaded `reduce` for it.  `@threads for` can then be derived from it.  This way, we can share the same code for `@threads` and threaded `reduce`.  (That's how https://github.com/JuliaFolds/Transducers.jl and https://github.com/tkf/ThreadsX.jl work.)

I think this approach is easier to extend, e.g., to "triangular" loops `for x in xs, y in f(x), z in g(x, y)`.

Giufin:
@JeffBezanson Here, took a little longer than I thought but it's working, and it does match the performance, though I don't have the time to check it very extensively, like I did the old one, so I hope I haven't missed any edge cases and ii wont break anything
Now that everything is done at expansion time the code got a little bit messy but I hope it's not a problem

@tkf `@threads` can't support iterators, to do that you wouldn't need to sync them(and if you are syncing you might as well just write sync the threads manually, without the use of a macro) or collect it, in which case you can just collect yourself and use the standard macro

tkf:
It is possible to support a wide range of "iterators" by defining `halve(xs) -> (left, right)` method on them.  This can be further amplified by converting iterator-transformations to transducers (`Base` already has basic support for this).  I'm happy to upstream https://github.com/JuliaFolds/SplittablesBase.jl (where `halve` is defined) for this to work in `Base`.  That's the building block package of https://github.com/JuliaFolds/Transducers.jl and https://github.com/tkf/ThreadsX.jl.

Giufin:
No matter how you approach it any other solution will be slower, either syncing, collection, or, what I assume you want, generating using modulo. I initially thought about implementing that one, You could then reuse a lot of the old macro, however it was quite a bit slower,  up to 20%, when a single iteration involved not much work and there a lot of them (you can argue that threads shouldn't be nested in that case, but now that it has roughly the same performance as splitting the loop manually, we don't need to split the loop into two variants). In the end I think we should have the macro be as low level as possible, as it's not hard to maintain, and it's performance oriented.

tkf:
I'm not suggesting locking-based or `collect`-based solution at all. It's a classic divide-and-conquer. Of course, there are situations you can do better than divide-and-conquer. However, it is pretty much usecase-dependent and the basic approach of using `Iterators.product` is still possible.

Giufin:
I know, but to implement the divide and conquer it you need to use modulo each iteration n times, or do some seriously weird stuff to generate the starting point with modulo and then continue the same way the current function does, this would involve precisely the same code as there is now, and without macros it would be slower (fastest implementation of that thing was my old function, that is about 2% slower for 60000 iterations and calculating 2 trigonometric functions each iteration)

tkf:
As I said, I'm not opposed to implementing something better than divide-and-conquer.  It's possible to implement a shallower task tree for `reduce` on `Iterators.product`.  This would give us much more composable internal than a macro-based solution.

jonas-schulze:
Does this work over function boundaries?

```julia
using Base.Threads

f(id) = @threads for _ in 1:nthreads() println("id=$id, threadid()=$(threadid())") end
g(h) = @threads for _ in 1:nthreads() h(threadid()) end

g(f)
```

base/threadingconstructs.jl
```diff
@@ -38,49 +38,77 @@ function threading_run(func)
...
+    #create variables to store information about each iterator
+    starts = [Meta.parse("start__" * string(x)) for x in lidx]
+    ends = [Meta.parse("end__" * string(x)) for x in lidx]
+    ranges = [Meta.parse("range__" * string(x)) for x in lidx]
+    values = [Meta.parse("val__" * string(x)) for x in lidx]
+    #create expressions initializing these variables
+    initranges = [:($(ranges[x]) = range[$x]) for x in 1:rangelen]
+    initstarts = ((x, y) -> :($x = firstindex($y) ) ).(starts, ranges)
+    initends = ((x, y) -> :($x = lastindex($y) ) ).(ends, ranges)
+    initstartingindecies = ((x, y) -> :($x += $y)).(values, starts)
```
jonas-schulze, line 56:
```suggestion
    initstartingindices = ((x, y) -> :($x += $y)).(values, starts)
```

base/threadingconstructs.jl
```diff
@@ -38,49 +38,77 @@ function threading_run(func)
...
-            # compute this thread's iterations
-            f = firstindex(r) + ((tid-1) * len)
-            l = f + len - 1
-            # distribute remaining iterations evenly
-            if rem > 0
+                #inits
+                $(initstarts...)
+                $(initends...)
+                $(initvalues...)
+                $(initstartingindecies...)
```
jonas-schulze, line 98:
```suggestion
                $(initstartingindices...)
```

Giufin:
No, this is specifically fot nested **outer** loops, in your example that would be converted to
```
using Base.Threads

f() = @threads for _ in 1:nthreads(), id in 1:nthreads() println("id=$id, threadid()=$(threadid())") end

f() 
```
I should probably also note that this would yield no performance improvement over splitting the expression to 2 different loops and using `@threads` on the outer one, since the number of iterations is divisible by the number of threads (they are the same), but in most cases this yields a small improvement while in specific cases (like the one that made my write this macro in the first place) the improvements are major.

jonas-schulze:
Ah, so this is for something like:

```
@threads for i in 1:10, j in 1:10
    ...
end
```

..?

Giufin:
Yes, exactly that, it's very niche but I wrote it because I had a use case and only later decided to put it here

Issue #36026, opened 2020-05-25
What's needed to include ClearStacktrace.jl in Base?


jkrumbiegel:
The title says it all. I think I figured out a good format to present stack traces, and I think all users could benefit from it. It's currently in [ClearStacktrace.jl](https://github.com/jkrumbiegel/ClearStacktrace.jl) and depends on `Crayons.jl` for the colors.

I want to know what I can do to make this PR-ready. I don't know how Julia Base handles REPL colors, which ones I'm allowed to use, how I access them without `Crayons` etc. I also don't know if there might be system-specific intricacies of stack traces that I have overlooked, testing this only a Windows and a Mac machine.

The basic idea is to have three columns, function, module and signature. Function and module together should basically always fit in a REPL horizontally, while signatures can be very long and are therefore allowed to overflow into new lines, while staying intact when resizing the REPL (compared to more complex but brittle table layouting). The code paths get a new line each, and are also allowed to overflow if they are too long. This keeps clickable links in Atom / VSCode and the like intact.

Just for reference, here is the before and after comparison from the README:
Before:
![grafik](https://user-images.githubusercontent.com/22495855/82832508-99862600-9ebb-11ea-90d3-fa0cb54eafe8.png)
After:
![grafik](https://user-images.githubusercontent.com/22495855/82832522-9f7c0700-9ebb-11ea-9076-6122ad482800.png)

cossio:
This would be really nice to have. Or at least maybe a trimmed down version which improves printing of stacktraces by default in Base.

JeffBezanson:
I'm all in favor. The main thing I don't get is the colors in the signatures. They seem...random? I guess the different colors are there to make it easier to pick out different elements, but it's a bit weird. How about color == nesting depth? I think the main weirdness is in e.g. `Tuple{Symbol,Symbol}` where the two `Symbol`s are different colors.

timholy:
There's another thing I'd *really* like to see:
```julia
julia> foo(x::T, y::T) where T = error("whoops")
foo (generic function with 1 method)

julia> function bar()
           a = rand(3, 5)
           b = view(a, :, 2:3)
           c = reinterpret(Float32, b)
           foo(c, c)
       end
bar (generic function with 1 method)

julia> bar()
ERROR: whoops
Stacktrace:
 [1] error(::String) at ./error.jl:33
[...]

PR #36134, opened 2020-06-03
include ClearStacktrace.jl in Base

jkrumbiegel:
cf. issue #36026

The base functionality works. I added handling of stackoverflow traces, too.

I have not dealt with any tests, yet. Also, somebody in the know should check if my methods of retrieving different parts of information about each frame are valid.

I am also checking for three boolean ENV variables in the code, to make printing a bit more customizable. They are:

```
JULIA_STACKTRACE_EXPAND_BASEPATHS
JULIA_STACKTRACE_CONTRACT_USERDIR
JULIA_STACKTRACE_LINEBREAKS
```

### DIFF
--- a/base/errorshow.jl
+++ b/base/errorshow.jl
@@ -609,7 +609,7 @@ function show_reduced_backtrace(io::IO, t::Vector)
     try invokelatest(update_stackframes_callback[], displayed_stackframes) catch end
 
     println(io, "\nStacktrace:")
-    ndigits = length(digits(length(t)))
+    ndigits_max = ndigits(length(t))
 
     modulecolorcycler = Iterators.Stateful(Iterators.cycle(STACKTRACE_MODULECOLORS))
 
@@ -632,7 +632,7 @@ function show_reduced_backtrace(io::IO, t::Vector)
         inlined = getfield(frame, :inlined)
 
         push!(LAST_SHOWN_LINE_INFOS, (file, line))
-        print_frame(io, frame_counter, func, inlined, modul, file, line, sigtypes, varnames, ndigits, modulecolor)
+        print_frame(io, frame_counter, func, inlined, modul, file, line, sigtypes, varnames, ndigits_max, modulecolor)
         
         if i < length(displayed_stackframes)
             println(io)
@@ -714,7 +714,7 @@ stacktrace_linebreaks()::Bool = parse(Bool, get(ENV, "JULIA_STACKTRACE_LINEBREAK
 function print_trace(io::IO, trace; print_linebreaks::Bool)
 
     n = length(trace)
-    ndigits = length(digits(n))
+    ndigits_max = ndigits(n)
 
     modulecolordict = Dict("" => :default)
     modulecolorcycler = Iterators.Stateful(Iterators.cycle(STACKTRACE_MODULECOLORS))
@@ -735,7 +735,7 @@ function print_trace(io::IO, trace; print_linebreaks::Bool)
         inlined = getfield(frame, :inlined)
 
         push!(LAST_SHOWN_LINE_INFOS, (file, line))
-        print_frame(io, i, func, inlined, modul, file, line, sigtypes, varnames, ndigits, modulecolor)
+        print_frame(io, i, func, inlined, modul, file, line, sigtypes, varnames, ndigits_max, modulecolor)
         if i < n
             println(io)
             print_linebreaks && println(io)
--- a/base/errorshow.jl
+++ b/base/errorshow.jl
@@ -549,6 +549,86 @@ end
 # replace `sf` as needed.
 const update_stackframes_callback = Ref{Function}(identity)
 
+
+function expandbasepath(str)
+
+    basefileregex = if Sys.iswindows()
+        r"^\.\\\w+\.jl$"
+    else
+        r"^\./\w+\.jl$"
+    end
+
+    if !isnothing(match(basefileregex, str))
+        sourcestring = find_source_file(str[3:end]) # cut off ./
+    else
+        str
+    end
+end
+
+function replaceuserpath(str)
+    str1 = replace(str, homedir() => "~")
+    # seems to be necessary for some paths with small letter drive c:// etc
+    replace(str1, lowercasefirst(homedir()) => "~")
+end
+
+getline(frame) = frame.line
+function getfile(frame, expandbase::Bool, contractuser::Bool)
+    file = string(frame.file)
+    if expandbase
+        file = expandbasepath(file)
+    end
+    if contractuser
+        file = replaceuserpath(file)
+    end
+
+    file
+end
+getfunc(frame) = string(frame.func)
+getmodule(frame) = try; string(frame.linfo.def.module) catch; "" end
+getsigtypes(frame) = try;  frame.linfo.specTypes.parameters[2:end] catch; "" end
+getmethod(frame) = try;  frame.linfo.def catch; nothing end
+function getvarnames(frame)
+    m = getmethod(frame)
+    if isnothing(m)
+        []
+    else
+        tv, decls, file, line = arg_decl_parts(m)
+        # this is a list of tuples (variable name, variable type)
+        # we take only the variable names
+        first.(decls[2:end])
+    end
+end
+
+const STACKTRACE_MODULECOLORS = [:light_blue, :light_yellow, :light_red,
+        :light_green,:light_magenta, :light_cyan,
+        :blue, :yellow, :red, :green, :magenta, :cyan]
+stacktrace_expand_basepaths()::Bool = parse(Bool,
+    get(ENV, "JULIA_STACKTRACE_EXPAND_BASEPATHS", "true"))
+stacktrace_contract_userdir()::Bool = parse(Bool,
+    get(ENV, "JULIA_STACKTRACE_CONTRACT_USERDIR", "true"))
+stacktrace_linebreaks()::Bool = parse(Bool, get(ENV, "JULIA_STACKTRACE_LINEBREAKS", "true"))
+
+function show_full_backtrace(io::IO, trace; print_linebreaks::Bool)
+
+    n = length(trace)
+    ndigits_max = ndigits(n)
+
+    modulecolordict = Dict("" => :default)
+    modulecolorcycler = Iterators.Stateful(Iterators.cycle(STACKTRACE_MODULECOLORS))
+
+    println(io, "\n--- STACKTRACE ---")
+    stacktrace_linebreaks() && println(io)
+
+    for (i, frame) in enumerate(trace)
+
+        print_stackframe(io, i, frame, ndigits_max, modulecolordict, modulecolorcycler)
+        if i < n
+            println(io)
+            print_linebreaks && println(io)
+        end
+    end
+end
+
 const BIG_STACKTRACE_SIZE = 50 # Arbitrary constant chosen here
 
 function show_reduced_backtrace(io::IO, t::Vector)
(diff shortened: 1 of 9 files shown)


base/errorshow.jl
```diff
@@ -610,27 +612,195 @@ function show_reduced_backtrace(io::IO, t::Vector, with_prefix::Bool)
 
     try invokelatest(update_stackframes_callback[], displayed_stackframes) catch end
 
+    println(io, "\nStacktrace:")
+    ndigits = length(digits(length(t)))
```
JeffBezanson, line 616:
```suggestion
    ndigits = Base.ndigits(length(t))
```

base/errorshow.jl
```diff
@@ -610,27 +612,195 @@ function show_reduced_backtrace(io::IO, t::Vector, with_prefix::Bool)
...
+
+    n = length(trace)
+    ndigits = length(digits(n))
+
+    modulecolordict = Dict("" => :default)
+    modulecolorcycler = Iterators.Stateful(Iterators.cycle(STACKTRACE_MODULECOLORS))
+
+    for (i, frame) in enumerate(trace)
+
+        modul = getmodule(frame)
```
JeffBezanson, line 728: Would be good to factor out this block of duplicated code.

base/errorshow.jl
```diff
@@ -644,33 +814,20 @@ function show_backtrace(io::IO, t::Vector)
...
-    if length(t) < BIG_STACKTRACE_SIZE
-        try invokelatest(update_stackframes_callback[], t) catch end
-        for entry in t
-            show_trace_entry(io, entry...)
-        end
-    else
-        show_reduced_backtrace(io, t, false)
-    end
+    # process_backtrace returns a Tuple{Frame, Int}
+    frames = first.(filtered)
```
JeffBezanson, line 825: This variable isn't used; did you mean to pass it instead of `t` in the next call?

JeffBezanson:
Thanks!

This is also going to need the code for handling keyword arguments in `show_spec_linfo` in stacktraces.jl. We could merge the new code with the `show` method for `StackFrame`, to avoid having two different ways to print the same information.

jkrumbiegel:
> This is also going to need the code for handling keyword arguments in `show_spec_linfo` in stacktraces.jl. 

I have followed your other suggestions, but I'm not quite sure what you mean by this. How should keyword arguments be handled? They do not appear in the specialization types because methods are not specialized for them, should we still print their names and the types of the passed arguments?

> We could merge the new code with the `show` method for `StackFrame`, to avoid having two different ways to print the same information.

Should I replace the show method with what I have? In what contexts is that other function used, I'm not sure what use cases I have to think about when merging.

KristofferC:
There is a bunch of code in the StackFrame module that is only releveant for the previous tacktrace printing, for example:

https://github.com/JuliaLang/julia/blob/855a08b35197e21a736d1f30e26e042c091712c7/base/stacktraces.jl#L219-L222

So either that should be removed or the code here should replace it.

KristofferC:
Maybe submodules should have the same color as the parent module?

bild

KristofferC:
> Can submodules be so different that it would make sense to distinguish them?

In many cases it is just an implementation detail of the package so I think it makes sense to keep them the same color.

JeffBezanson:
> How should keyword arguments be handled?

The current printing is:
```
julia> f(x; kwarg=1) = error();

julia> f(1,kwarg=3)
ERROR: 
Stacktrace:
 [1] error() at ./error.jl:42
 [2] f(::Int64; kwarg::Int64) at ./REPL[4]:1
 [3] top-level scope at REPL[10]:1
```
The magic is done starting at
https://github.com/JuliaLang/julia/blob/095e92d4e9ad2b90d525fec99f5474cd8d017b80/base/stacktraces.jl#L228

jkrumbiegel:
I have added keyword printing as well, but I couldn't just copy paste the code from `stacktraces.jl`. It's not really documented what many of the small operations do that seem to be there for edge cases. This is probably as far as I can take it because my knowledge of the internals is very limited.

base/errorshow.jl
```diff
@@ -610,27 +699,145 @@ function show_reduced_backtrace(io::IO, t::Vector, with_prefix::Bool)
...
+
+Print a stack frame where the module color is set manually with `modulecolor`.
+"""
+function print_stackframe(io, i, frame, digit_align_width, modulecolor)
+
+    file = getfile(frame, stacktrace_expand_basepaths(), stacktrace_contract_userdir())
+    line = getline(frame)
+
+    # add file and line info for accessing frame locations from the repl
+    push!(LAST_SHOWN_LINE_INFOS, (file, line))
```
simeonschaub, line 768: Just mentioning it here: Once #35556 is merged the global `LAST_SHOWN_LINE_INFOS` should be replaced by `get(io, :LAST_SHOWN_LINE_INFOS, Tuple{String, Int}[])`.

tk3369:
> JULIA_STACKTRACE_EXPAND_BASEPATHS
> JULIA_STACKTRACE_CONTRACT_USERDIR

Do we really want opposite directions here (`EXPAND` and `CONTRACT`)?  It would be more consistent to me if they're both `EXPAND` or both `CONTRACT`.  Just a thought.

jkrumbiegel:
Well the reason is that by default the user directory is part of the file path, so it has to be explicitly contracted. And the base files are just file names, so their file paths have to be explicitly expanded :) Those are the different behaviors compared to the current state

jkrumbiegel:
I guess if we said that the new default is to print base file paths completely, then the option could be named contract base paths to get the old behavior. It's a bit annoying that those paths tend to be super long but I personally love being able to click them in VSCode. The options are meant so everybody can decide how terse or verbose they want their traces. To me for example, even though a short stack trace can be nice without line breaks, after a certain length I don't care about the overall trace size anymore, it's all about being able to visually parse it at all

StefanKarpinski:
Seems clearer to me if they both are expand or contract but have different defaults.

jkrumbiegel:
To proceed, I need help with the following:
- what kind of tests do we need?
- is the positional / keyword argument name functionality correctly implemented?
- is the printing of anonymous and keyword shim function names correct (all the ## business)?
- are there edge cases that I didn't handle?

KristofferC:
I took the liberty of pushing a commit to this branch. Feel free to revert / make whatever changes you want on top of it. The changes are:

- Fixed a bunch of tests that this broke.
- Added a test for the new printing
-  Made it so that submodules of Main are considered distinct which seems reasonable to me: 
bild
- Changed back to `Stacktrace:` for the header. This is used everywhere in doctests so it seems best to keep this.
- Fixed some bugs in the limited stacktrace printing
- Fixed bug with the stacktrace methodloc updater
- Implemented this by doing the formatting in the stackframe printer. This meant that a lot of code could be removed since almost everything just piggy-backs on that. It also fixes some bugs in corner cases (like the function name in keyword args).

I also changed some of the defaults from some experience using this (again feel free to revert). These are:

- Changed to no line break by default. I noticed that it was very frequent that I had to scroll up to see the error message now.
- Added back the indent to the stacktrace number (similar to how it is right now)ö

jkrumbiegel:
All those changes look very reasonable to me, thanks for pushing this forward. I also think after using it for a while that the line breaks should be opt-in rather than opt-out as it can make some traces a bit long and might not add that much clarity for shorter traces.

KristofferC:
Seems tests are passing now. Would be good with another round of reviews (@JeffBezanson).

base/errorshow.jl
```diff
@@ -558,9 +549,44 @@ end
 # replace `sf` as needed.
 const update_stackframes_callback = Ref{Function}(identity)
 
+function replaceuserpath(str)
+    str = replace(str, homedir() => "~")
+    # seems to be necessary for some paths with small letter drive c:// etc
+    replace(str, lowercasefirst(homedir()) => "~")
```
JeffBezanson, line 555:
```suggestion
    str = replace(str, lowercasefirst(homedir()) => "~")
```

base/errorshow.jl
```diff
@@ -558,9 +549,44 @@ end
...
+    return str
+end
+
+const STACKTRACE_MODULECOLORS = [:light_blue, :light_yellow,
+        :light_magenta, :light_green, :light_cyan, :light_red,
+        :blue, :yellow, :magenta, :green, :cyan, :red]
+stacktrace_expand_basepaths()::Bool =
+    tryparse(Bool, get(ENV, "JULIA_STACKTRACE_EXPAND_BASEPATHS", "false")) === true
+stacktrace_contract_userdir()::Bool =
+    tryparse(Bool, get(ENV, "JULIA_STACKTRACE_CONTRACT_USERDIR", "true")) === true
```
JeffBezanson, line 565:
```suggestion
    tryparse(Bool, get(ENV, "JULIA_STACKTRACE_CONTRACT_HOMEDIR", "true")) === true
```
I think that's a more standard term for this?

base/show.jl
```diff
@@ -1787,11 +1787,17 @@ function show_signature_function(io::IO, @nospecialize(ft), demangle=false, farg
     nothing
 end
 
-function show_tuple_as_call(io::IO, name::Symbol, sig::Type, demangle=false, kwargs=nothing)
+function printstacktrace(io, s...; color, bold=false)
```
JeffBezanson, line 1790:
```suggestion
function print_within_stacktrace(io::IO, s...; color, bold=false)
```

JeffBezanson (approved):

JeffBezanson:
Thanks, looks beautiful!

KristofferC:
Let's try this then!

simeonschaub:
@KristofferC I believe this broke some doctest:
https://build.julialang.org/#/builders/39/builds/1076
https://build.julialang.org/#/builders/39/builds/1077

KristofferC:
But the doctests ran and passed?

jkrumbiegel:
Is there a place where all Julia ENV variables are listed? That was one thing I still wanted to add, so people can actually find the new options.

KristofferC:
Yes, can be added to https://github.com/JuliaLang/julia/blob/30b09dfd2b3a898237f71d4a2b0812466d017f9e/doc/src/manual/environment-variables.md.

mcabbott:
> Is is just me that thinks it is a bit strange that , and :: stands out more than types in the signatures

This also seems a bit odd to me. The intention is to make it easier to spot where the n-th type starts. An alternative would be to print the outermost type normally, and its (possibly long) type parameters in the dim (bright_black?) now used for the whole type. That way useful data marks the same division.

fredrikekre:
One idea would be to print the outermost type in the same color as the `::` but still dim the inner type parameters. Something like this:
Screenshot from 2020-09-16 15-32-49

mcabbott:
I made a fork which implemented this -- [pictures here](https://github.com/mcabbott/ClearStacktrace.jl/blob/milder/README.md#september-tweak) now updated to be on a dark background, and not to highlight the type. 

Besides the colons, IMO it's not necessary to use super-bright colours for the modules, and it would be a little nicer to have them in a rainbow-ish order.

fredrikekre:
That looks like an improvement to me. Make a PR?

KristofferC:
I think it is slightly better but it is becoming a little bit of a "color salad" where it is hard to scan for the information you want. Worth trying out though.

mcabbott:
Any thoughts on how to un-salidify it? 

Would using fewer colours help, like maybe just the first 3 above, then re-use? (Since `RecipesPipeline` is so much longer, it would still look different.)

Terminal contrast alters this a lot too, e.g. the light-mode screenshots (at my link) have much less distinction between normal and dim text.

Will try a PR when I get a minute.

StefanKarpinski:
FWIW, I think the colors are fine.

KristofferC:
> FWIW, I think the colors are fine.

The current ones or the ones in the last screenshot?

StefanKarpinski:
Last screenshot.

mcabbott:
Not a PR yet, but here's a paste-able patch:
https://gist.github.com/mcabbott/06a89aa18a0e350a2281c80320a5b01d
Mostly like the above screenshot, but it makes Base & Core red not gray.

jkrumbiegel:
I just installed 1.6 and noticed that the function names are not bold anymore, making them stick out less than I originally intended (especially if they are above and below colored Modules). This must have gotten lost in a rewrite? I've tried to follow to the point where functions are printed, but I couldn't find it. The new implementation seems to go quite deep into specialized implementations, but the gist is that the function name is printed non-bold.

See here, it's hard to see the function names.

grafik

KristofferC:
It might have gotten lost, yes. Should be easy to put back though.

mcabbott:
I tried here: https://github.com/JuliaLang/julia/pull/37773/commits/ada29c2f056cb3f5ee1b4f2cb8d14c44214ae8b4. (Without really understanding how they used to be made bold, before this.)

NHDaly:
This PR broke cycle detection in stack trace printing in some cases 😮
Please see #37587 for more details.

I've tracked the breakage down to this PR, where you can see a before/after that shows the stack trace not reporting the recursion cycle:
```julia

julia> versioninfo()  # before
Julia Version 1.6.0-DEV.353
Commit 7dba98808a* (2020-07-02 01:54 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin18.7.0)
  CPU: Intel(R) Core(TM) i7-6920HQ CPU @ 2.90GHz
  WORD_SIZE: 64
  LIBM: libopenlibm

julia> foo() = foo()

julia> foo()
ERROR: StackOverflowError:
Stacktrace:
 [1] foo() at ./REPL[2]:1 (repeats 79984 times)
```
```julia
julia> versioninfo()  # after
Julia Version 1.6.0-DEV.354
Commit 30b09dfd2b* (2020-07-02 08:02 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin18.7.0)
  CPU: Intel(R) Core(TM) i7-6920HQ CPU @ 2.90GHz
  WORD_SIZE: 64
  LIBM: libopenlibm

julia> foo() = foo()

julia> foo()
ERROR: StackOverflowError:
Stacktrace:
 [1] foo()
   @ Main ./REPL[2]:1
```

The new printing is super pretty and I love it -- can we try to track down and fix this regression so that cycle detection still works with this new framework? :) Thanks!

PR #36137, opened 2020-06-03
Remove unnecessary jl_gc_wb calls

timholy:
CC @vtjnash


yuyichao (approved):
I assume these are because the array setter already does this. These are technically wrong anyway.

PR #32185, opened 2019-05-29
Revert "Always use summary for BoundsError, even with non-AbstractArrays"

fredrikekre:
Reverts JuliaLang/julia#32058

See e.g. https://github.com/JuliaLang/julia/pull/32166/commits/2b0c2f35eee649b4690d629c369e3b2e00c64567 why this might not be a good idea, the only reason `summary` makes sense for abstract arrays is that they show the size, which is exactly what you want for a `BoundsError`, e.g.
```
julia> [1 2; 3 4][4, 5]
ERROR: BoundsError: attempt to access 2×2 Array{Int64,2} at index [4, 5]
```
but in other cases I think showing the value is better (unless the fallback `summary` shows the indexing size...).

PR #36139, opened 2020-06-03
add a summary for AbstractString

KristofferC:
A bit more useful than the current printing for e.g. BoundsError. Now it is:

```
julia> "foo"[5]
ERROR: BoundsError: attempt to access 3-codeunit String at index [5]
```

### DIFF
--- a/base/strings/basic.jl
+++ b/base/strings/basic.jl
@@ -419,11 +419,11 @@ julia> thisind("α", 3)
 3
 
 julia> thisind("α", 4)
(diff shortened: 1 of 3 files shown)


fredrikekre:
ref https://github.com/JuliaLang/julia/pull/32185

StefanKarpinski (approved):

JeffBezanson:
Maybe special case to print `empty` instead of `0-codeunit`?

PR #36143, opened 2020-06-04
fix ImmutableDict(pairs...) constructor

rfourquet:
It could only handle a couple of pairs, e.g.
`ImmutableDict(1=>1, 2=>2, 3=>3)` would throw.

The fix is implemented by adding the
`ImmutableDict(t::ImmutableDict, pairs...)` constructor,
which generalizes `ImmutableDict(t::ImmutableDict, pair)`
(with some similarity to how `push!` accepts multiple items
to be pushed).

### DIFF
--- a/base/dict.jl
+++ b/base/dict.jl
@@ -739,6 +739,8 @@ Create a new entry in the `ImmutableDict` for a `key => value` pair
 ImmutableDict
 ImmutableDict(KV::Pair{K,V}) where {K,V} = ImmutableDict{K,V}(KV[1], KV[2])
 ImmutableDict(t::ImmutableDict{K,V}, KV::Pair) where {K,V} = ImmutableDict{K,V}(t, KV[1], KV[2])
+ImmutableDict(t::ImmutableDict{K,V}, KV::Pair, rest::Pair...) where {K,V} =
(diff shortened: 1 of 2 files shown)


JeffBezanson (approved):

BenjaminGalliot:
Is there a reason for the absence of an ImmutableDict constructor from a Dict? It is pretty near the multiple pairs constructor but could be useful for automatic conversion with function return type, I think.

rfourquet:
I don't think there is any reason, besides that this `ImmutableDict` is not really for public consumption and therefore has mostly only methods needed in `Base` itself. But it turns out to be a useful type to use also in some packages.

I could have used a `ImmutableDict(d)` constructor (where `d` is an iterator) a couple of times but didn't get around to making a PR yet.

BenjaminGalliot:
That explains why it needs to be imported manually…

Honestly, I bumped into it from [here](https://stackoverflow.com/questions/38402124/immutable-dictionary) then [there](https://discourse.julialang.org/t/way-to-construct-multi-key-immutabledict/10294) because I was searching a good pattern to use a global configuration variable (almost only constants, like database parameters, regexes and paths) to avoid a lot of tramp data in functions…

PR #36145, opened 2020-06-04
Added online regex tools

cyrilbois:
Hello, I added online regex tools (regex101, and extendsclass - I am the founder) to "Regular Expressions". It can help to check regex.

### DIFF
--- a/doc/src/manual/strings.md
+++ b/doc/src/manual/strings.md
@@ -1013,6 +1013,8 @@ julia> match(regex_name,"[Jon]") === nothing
 true
 ```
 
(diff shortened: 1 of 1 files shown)


StefanKarpinski:
These are useful sites, but it feels weird for the Julia manual to be officially endorsing specific sites for this kind of thing, especially the second one which has a lot of ads and feels quite spammy.

cyrilbois:
Thank you for your reply.

ararslan:
Hi @cyrilbois and thanks for the contribution! I agree with Stefan here; I don't think the Julia manual should endorse any particular external resources. As such, I'll close this. We appreciate your willingness to contribute though!

Issue #36136, opened 2020-06-03
Change function signature of redirect_std* to allow anonymous functions & other callables


Seelengrab:
https://github.com/JuliaLang/julia/blob/2d5741174ce3e6a394010d2e470e4269ca54607f/base/stream.jl#L1147

This came up on slack. The `::Function` restriction feels like a leftover from (at least) 4 years ago, when that code was last touched. Relaxing the signature would allow `redirect_stdout(() -> call(with, arguments), devnull)` and similar to work without having to wrap the call in a proper function. 

I'm not terribly familiar with this sort of change, but I think `@nospecialize` on the `func` would be necessary to be consistent with other parts in Base where arbitrary functions can be used? Not sure, so that's why this is an issue and not a PR, even though the change itself seems trivial.

martinholters:
Note that `() -> call(with, arguments)` is a `Function`, so this may not be as important to change as you think. OTOH, I don't see much reason for the restriction `Function`, either.

Seelengrab:
> Note that () -> call(with, arguments) is a Function, so this may not be as important to change as you think. 

Hm, that's true.. So the call fails because `devnull` is not a TTY, Pipe or Socket? I would have exptected that to work, to be honest 🤔

laborg:
@Seelengrab as `devnull` is now allowed in the `redirect_std` calls, I think its time to close this issue. Or did the actual `::Function` restriction get in your way in the meantime?

Seelengrab:
Well, the original issue someone ran into with `devnull` is fixed, though the `::Function` restriction [is still there, I think](https://github.com/simeonschaub/julia/blob/d0ee29bf7e0002c353361d602fec2eeab0353e70/base/stream.jl#L1154). Not a pressing issue, still an annoyance :shrug: I'd just keep this open in case someone looks for it/wants to fix it for any callables.

PR #36146, opened 2020-06-04
allow `redirect_std*` to `devnull` (#36136)

KlausC:
solves #36136 partially. 

The methods `redirect_std*(io::Base.DevNull)` have been added. They redirect to `"/dev/null"` on non-windows systems and to `"NUL"` elsewhere.
This allows to call for example `redirect_stdout(f, devnull)` to supress all output to `stdout` produced in function call `f()`.
The first argument `f` is still restricted to `::Function`.

### DIFF
--- a/base/stream.jl
+++ b/base/stream.jl
@@ -1098,6 +1098,12 @@ for (x, writable, unix_fd, c_symbol) in
             ($f)($(writable ? :write : :read))
             return (read, write)
         end
+        function ($f)(::DevNull)
+            nulldev = @static Sys.iswindows() ? "NUL" : "/dev/null"
+            handle = open(nulldev, write=$writable)
+            $(_f)(handle)
+            return handle
+        end
     end
 end
 
@@ -1115,7 +1121,7 @@ elsewhere.
 If called with the optional `stream` argument, then returns `stream` itself.
 
 !!! note
-    `stream` must be a `TTY`, a `Pipe`, or a socket.
+    `stream` must be a `IOStream`, a `TTY`, a `Pipe`, a socket, or `devnull`.
 """
 redirect_stdout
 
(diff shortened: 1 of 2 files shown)


carpenter832:
Allow redirect

base/stream.jl
```diff
@@ -1098,6 +1098,12 @@ for (x, writable, unix_fd, c_symbol) in
             ($f)($(writable ? :write : :read))
             return (read, write)
         end
+        function ($f)(::DevNull)
+            nulldev = @static Sys.iswindows() ? "NUL" : "/dev/null"
+            handle = open(nulldev, write=$writable)
+            $(_f)(handle)
+            return handle
+        end
```
tkf, line 1106: Would `handle` be closed after `redirect_stdout(f, devnull)`? (It looks like not?)

KlausC: You are right - I am working to close the handle when it is no longer needed.

base/stream.jl
```diff
@@ -1098,6 +1099,12 @@ for (x, writable, unix_fd, c_symbol) in
             ($f)($(writable ? :write : :read))
             return (read, write)
         end
+        function ($f)(::DevNull)
+            nulldev = @static Sys.iswindows() ? "NUL" : "/dev/null"
+            handle = open(nulldev, write=$writable)
+            $(_f)(handle, isnull=true)
```
Keno, line 1105: Just `$x = devnull` right after here instead of the extra keyword argument? Seems cleaner to me.

KlausC: agree - the argument was also to close `handle`, so I added that too.

StefanKarpinski:
Let's get this merged soon if there are no more review comments!

KlausC:
bump it after a week of silence

PR #36148, opened 2020-06-04
fix recursion detection in constant prop

JeffBezanson:
#36058 was a bit too optimistic, and caused the inference stack overflow found here: https://github.com/JuliaGraphics/Gtk.jl/pull/512#issuecomment-638427192

This extra check fixes it while allowing the case in #36058 to still work.

### DIFF
--- a/base/compiler/abstractinterpretation.jl
+++ b/base/compiler/abstractinterpretation.jl
@@ -132,7 +132,7 @@ function abstract_call_gf_by_type(interp::AbstractInterpreter, @nospecialize(f),
         # if there's a possibility we could constant-propagate a better result
         # (hopefully without doing too much work), try to do that now
         # TODO: it feels like this could be better integrated into abstract_call_method / typeinf_edge
-        const_rettype = abstract_call_method_with_const_args(interp, rettype, f, argtypes, applicable[nonbot]::SimpleVector, sv)
+        const_rettype = abstract_call_method_with_const_args(interp, rettype, f, argtypes, applicable[nonbot]::SimpleVector, sv, edgecycle)
         if const_rettype ⊑ rettype
             # use the better result, if it's a refinement of rettype
             rettype = const_rettype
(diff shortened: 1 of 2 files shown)


vtjnash (commented):
Ah, right, I'd completely forgotten that the reason I had been avoiding this was due to the lack of a type-complexity limit. I'm somewhat confused though why this doesn't break the other test though.

base/compiler/abstractinterpretation.jl
```diff
@@ -259,6 +259,17 @@ function abstract_call_method_with_const_args(interp::AbstractInterpreter, @nosp
...
+        if edgecycle
+            # if there might be a cycle, check to make sure we don't end up
+            # calling ourselves here.
+            infstate = sv
+            while !(infstate === nothing)
+                if method === infstate.linfo.def
+                    return Any
+                end
+                infstate = infstate.parent
+            end
```
vtjnash, line 271:
```suggestion
            # if edgecycle is true, then there is a cycle and we're calling ourself
            return Any
```

JeffBezanson: Well, that would just revert the original PR. What we need is to find out if adding the constant information breaks the cycle. Maybe we could do a full recursion check, but only look for other frames coming from constant prop?

vtjnash (commented):
Ah, right, I'd completely forgotten that the reason I had been avoiding this was due to the lack of a recursion type-complexity limit. I'm somewhat confused though why this doesn't break the other test though, but I think it's because it's only looking through a subset of the parents (some of the parents are in the `callers_in_cycle` array, and checking both I think is equivalent to returning unconditionally here)

JeffBezanson:
> some of the parents are in the `callers_in_cycle` array, and checking both I think is equivalent to returning unconditionally here

I tried it, and that is correct as far as I can tell.

base/compiler/abstractinterpretation.jl
vtjnash: Okay, right we could probably make `edgecycle` computation more lenient to allow one identical parent to have the constant-prop flag set and still be good here.

JeffBezanson: What is the best way to check that flag? `any(result.overridden_by_const)`?

vtjnash: Sounds about right

vtjnash (approved):

Issue #36132, opened 2020-06-03
seekuntil function to mirror readuntil


non-Jedi:
I often find myself wishing for a `seekuntil` function that behaves like `readuntil` but does not allocate a string. It seems like an obviously useful function for lots of purposes. Is there interest in including such a function in `Base`?

StefanKarpinski:
Yes, that does make sense.

non-Jedi:
[On zulip it was brought up that it might be more flexible/useful to have `eachbyte`/`eachchar` iterables instead](https://julialang.zulipchat.com/#narrow/stream/225542-helpdesk/topic/findfirst.20on.20IO.20stream/near/199743019).

non-Jedi:
Turns out this does already exist which I didn't find until I had already mostly implemented this. It's `skipchars`. I'm leaving this issue open with the actionable item being to add a link in the `readuntil` docstring to `skipchars`.

StefanKarpinski:
We have done so occasionally when the name for something was really bad. We would deprecate `skipchars` for `skipuntil` since deprecations are off by default now, which would have the effect that anyone who asks for deprecation warnings would get the message to change the name.

PR #36150, opened 2020-06-04
Add readeach function for iterating streams

non-Jedi:
See the commit messages and the discussion in #36132 for justification/explanation. <del>I can break this up into 2 PRs if there's only consensus on merging half of this, but it's semi-related, so I'm submitting as a single PR first.</del>


JeffBezanson:
I think these are both basically good ideas, but I agree they should be 2 PRs.

Unfortunately with the name `skipuntil` the meaning of the predicate should be inverted, so it's not a drop-in replacement for `skipchars` (but it could still replace it eventually).

non-Jedi:
Ah right. In my head `skipchars` was using the predicate the opposite of how it actually is. I'll remove those commits while I try to figure out the best way of using the deprecation machinery for this and let this PR be about `eachof`. Will file a separate PR for `skipuntil` later.

non-Jedi:
aarch64 failure seems unrelated

StefanKarpinski:
Can I suggest the name `readeach` instead of `eachof`?

non-Jedi:
`readeach` seems much better than `eachof`. I anticipated some discussion of what the best name for this was, so I'm going to wait to make the change to see if there are other suggestions.

non-Jedi:
Would `eachread` be better than `readeach` for symmetry with `eachline`, `eachindex`, etc.?

StefanKarpinski:
I think `readeach(io, T)` sounds more natural, personally, so I'd go with that.

base/io.jl
```diff
@@ -1003,6 +1000,46 @@ eltype(::Type{<:EachLine}) = String
 
 IteratorSize(::Type{<:EachLine}) = SizeUnknown()
 
+struct EachOfIO{T, IOT <: IO}
```
tkf, line 1003: It'd make sense to rename `EachOfIO` too.

JeffBezanson: Maybe `ReadEachIterator`?

JeffBezanson:
Triage says :+1: to `readeach`.

base/io.jl
non-Jedi: Good catch. Thanks.

base/io.jl
```diff
@@ -1003,6 +1000,46 @@ eltype(::Type{<:EachLine}) = String
...
+
+# Examples
+```jldoctest
+julia> open("my_file.txt", "w") do io
+           write(io, "JuliaLang is a GitHub organization.\\n It has many members.\\n");
+       end;
+
+julia> open("my_file.txt") do io
+           for c in readeach(io, Char)
+               c == '\\n' && break
```
rfourquet, line 1025: I'm confused: the doctests pass, but `'\\n'` doesn't seem like a valid character literal...

base/io.jl
```diff
@@ -1003,6 +1000,46 @@ eltype(::Type{<:EachLine}) = String
...
+
+See also: [`skipchars`](@ref), [`eachline`](@ref), [`readuntil`](@ref)
+
+!!! compat "Julia 1.6"
+    `readeach` requires Julia 1.6 or later.
+
+# Examples
+```jldoctest
+julia> open("my_file.txt", "w") do io
+           write(io, "JuliaLang is a GitHub organization.\\n It has many members.\\n");
```
rfourquet, line 1020: For demonstration purposes, maybe using `IOBuffer("JuliaLang is a GitHub organization.\\n It has many members.\\n")` would be enough as the focus is on `readeach` ?

base/io.jl
```diff
@@ -1087,10 +1124,10 @@ julia> String(readavailable(buf))
 ```
 """
 function skipchars(predicate, io::IO; linecomment=nothing)
-    while !eof(io)
-        c = read(io, Char)
+    for c in readeach(io, Char)
         if c === linecomment
-            readline(io)
```
rfourquet, line 1093: `readline(io)` is not valid anymore in this context? (sorry if this ovious!)

rfourquet: Ok, I see from a commit message that this might be to avoid allocation which presumably happens in `readline`? you might add a comment to clarify this and prevent future "simplifications" changing this back to "readline" :)

StefanKarpinski (approved):
LGTM!

base/io.jl
non-Jedi: I believe that since this is within a string, the first backslash escapes the second.

base/io.jl
non-Jedi: I copied the format of these doctests from the surrounding functions e.g. `readlines`. I'd prefer to keep them consistent, although I agree that using `IOBuffer` rather than opening an actual file might add to their clarity?

base/io.jl
rfourquet: Oh right!

rfourquet (approved):

non-Jedi:
I've pushed two commits addressing @rfourquet's review that can be squashed once this is ready for merge.

rfourquet (approved):

tkf:
It looks like thre is a conflict in NEWS file.

PR #36153, opened 2020-06-05
Add a MSIX installer

davidanthoff:
This adds a [MSIX installer](https://docs.microsoft.com/en-us/windows/msix/overview).

The eventual goal is that we might be able to put Julia in the Windows Store with this.

At this moment it is very much an experiment, though.

The build step seems to work, but one has to sign the created msix file to be able to install it, but that should probably occur somewhere else. The command that I used locally that worked is `SignTool sign /fd SHA256 /a /f mycert.pfx /p mysupersecretpassword julia-4b9ceb55d7-win64.msix`. [This](https://docs.microsoft.com/en-us/windows/msix/package/sign-app-package-using-signtool) has instructions for other modes, for example when the certificate is installed in the cert store etc.

CC @staticfloat


davidanthoff:
Closed in favor of https://github.com/JuliaLang/julia/issues/36672 (I'll keep the branch around, though).

PR #36156, opened 2020-06-05
Support dot-calls with a do block

### LINKED ISSUE #33800: Feature request: `f.(x) do ... end`
It looks like the dot-call syntax can't be used with the do notation at the moment:

```julia
julia> :(f.(x) do; end)
ERROR: syntax: missing comma or ) in argument list
```

But I think it is handy in some situations.  For example:

```julia
first_nonempty_lines = open.(files) do io
    for line in eachline(io)
        isempty(line) || return line
    end
end
```

Does it make sense to support this?

tkf:
With this patch, you can use broadcasting with a `do` block:

```julia
julia> open.(joinpath.("base", ["reduce", "reducedim", "operators"] .* ".jl")) do io
           for line in eachline(io)
               isempty(line) || return line
           end
       end
3-element Array{String,1}:
 "# This file is a part of Julia. License is MIT: https://julialang.org/license"
 "# This file is a part of Julia. License is MIT: https://julialang.org/license"
 "# This file is a part of Julia. License is MIT: https://julialang.org/license"
```

close #33800


DilumAluthge:

Issue #30632, opened 2019-01-07
Looping over codeunits could perhaps be optimized


KristofferC:
As an example of comparing looping over a vector vs a `CodeUnits`:

```
function count_tot(x)
    s = 0
    for i in x
        s += i
    end
    return s
end

using Random
s = randstring(10^5);
cu = codeunits(s);
arr = unsafe_wrap(Array, pointer(s), sizeof(s));
```

```
julia> @btime count_tot(cu)
  72.418 μs (1 allocation: 16 bytes)
8680965

julia> @btime count_tot(arr)
  10.936 μs (1 allocation: 16 bytes)
8680965
```

KristofferC:
LLVM says:

```
julia> count_tot(cu)
remark: strings/basic.jl:741:0: loop not vectorized: could not determine number of loop iterations
remark: strings/basic.jl:741:0: loop not vectorized
8685309
```

laborg:
Doesn't appear to be a problem anymore (Julia >= 1.9.3):
```
julia> @btime count_tot(cu)
  6.480 μs (1 allocation: 16 bytes)
8688211

julia> @btime count_tot(arr)
  6.492 μs (1 allocation: 16 bytes)
8688211
```

PR #36159, opened 2020-06-05
simplify generated code for iterate on codeunits

KristofferC:
One step towards fixing https://github.com/JuliaLang/julia/issues/30632. Makes the function noexcept but still doesn't vectorize in https://github.com/JuliaLang/julia/issues/30632 like `Array` does though (not that that matters much, I'm just using it as a smoke test to see that the abstractions are seen through).

### DIFF
--- a/base/strings/basic.jl
+++ b/base/strings/basic.jl
@@ -731,7 +731,7 @@ size(s::CodeUnits) = (length(s),)
 elsize(s::CodeUnits{T}) where {T} = sizeof(T)
 @propagate_inbounds getindex(s::CodeUnits, i::Int) = codeunit(s.s, i)
 IndexStyle(::Type{<:CodeUnits}) = IndexLinear()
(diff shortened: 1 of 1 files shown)


StefanKarpinski (approved):

base/strings/basic.jl
```diff
@@ -731,7 +731,8 @@ size(s::CodeUnits) = (length(s),)
 elsize(s::CodeUnits{T}) where {T} = sizeof(T)
 @propagate_inbounds getindex(s::CodeUnits, i::Int) = codeunit(s.s, i)
 IndexStyle(::Type{<:CodeUnits}) = IndexLinear()
-iterate(s::CodeUnits, i=1) = (@_propagate_inbounds_meta; i == length(s)+1 ? nothing : (s[i], i+1))
+@inline iterate(s::CodeUnits, i=1) = (@_inline_meta; (i % UInt) - 1 < length(s) ? (@inbounds s[i], i + 1) : nothing)
```
JeffBezanson, line 734: Why both inline and _inline_meta?

KristofferC: I *really*  want it to inline?

PR #36166, opened 2020-06-05
leaf-types cache for ml-matches

vtjnash:
When we do a `methods` lookup and get back a single result, it's very easy for us to cache that information, since we essentially already have the data-structure for it (TypeMapEntry). This lets us shave off a bit of time on micro-benchmarks by putting this information into a hash table instead of a tree (which is still used to handle the general case):
```
julia> @btime methods(+, (Int, Int))
master:  3.322 μs (21 allocations: 912 bytes)
PR:      2.069 μs (18 allocations: 784 bytes)

julia> @btime Base._methods_by_ftype(Tuple{typeof(+), Int, Int}, -1, typemax(UInt64))
master:  1.496 μs (7 allocations: 464 bytes)
PR:      0.175 μs (4 allocations: 336 bytes)
```

### DIFF
--- a/src/gf.c
+++ b/src/gf.c
@@ -231,7 +231,8 @@ jl_datatype_t *jl_mk_builtin_func(jl_datatype_t *dt, const char *name, jl_fptr_a
     m->sig = (jl_value_t*)jl_anytuple_type;
     m->slot_syms = jl_an_empty_string;
 
-    JL_GC_PUSH1(&m);
+    jl_typemap_entry_t *newentry = NULL;
+    JL_GC_PUSH2(&m, &newentry);
     jl_method_instance_t *mi = jl_get_specialized(m, (jl_value_t*)jl_anytuple_type, jl_emptysvec);
     m->unspecialized = mi;
     jl_gc_wb(m, mi);
@@ -242,8 +243,10 @@ jl_datatype_t *jl_mk_builtin_func(jl_datatype_t *dt, const char *name, jl_fptr_a
     codeinst->invoke = jl_fptr_args;
 
     jl_methtable_t *mt = dt->name->mt;
-    jl_typemap_insert(&mt->cache, (jl_value_t*)mt, jl_anytuple_type,
-        NULL, jl_emptysvec, (jl_value_t*)mi, 0, &lambda_cache, 1, ~(size_t)0);
+    newentry = jl_typemap_alloc(jl_anytuple_type, NULL, jl_emptysvec,
+            (jl_value_t*)mi, 1, ~(size_t)0);
+    jl_typemap_insert(&mt->cache, (jl_value_t*)mt, newentry, 0, &lambda_cache);
+
     mt->frozen = 1;
     JL_GC_POP();
     return dt;
@@ -475,8 +478,10 @@ void jl_foreach_reachable_mtable(void (*visit)(jl_methtable_t *mt, void *env), v
 static void reset_mt_caches(jl_methtable_t *mt, void *env)
 {
     // removes all method caches
-    if (mt->defs != jl_nothing) // make sure not to reset builtin functions
+    if (mt->defs != jl_nothing) { // make sure not to reset builtin functions
+        mt->leafcache = (jl_array_t*)jl_an_empty_vec_any;
         mt->cache = jl_nothing;
+    }
     jl_typemap_visitor(mt->defs, get_method_unspec_list, env);
 }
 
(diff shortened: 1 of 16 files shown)


test/channels.jl
```diff
@@ -390,6 +390,7 @@ end
         t = Timer(0) do t
             tc[] += 1
         end
+        Libc.systemsleep(0.005)
```
Keno, line 393: Is there something better we can do here. Tests that depend on system scheduling behavior almost always turn out flakey.

src/gf.c
```diff
@@ -2803,10 +2861,54 @@ static jl_value_t *ml_matches(jl_typemap_t *defs, int offs,
...
+            jl_array_ptr_set(env.t, 0, env.matc);
+            if (*min_valid < entry->min_world)
+                *min_valid = entry->min_world;
+            if (*max_valid > entry->max_world)
+                *max_valid = entry->max_world;
+            JL_GC_POP();
+            return env.t;
+        }
+    }
+    if (((jl_datatype_t*)unw)->isdispatchtuple) {
```
Keno, line 2892: Isn't this the same condition as the previous if block?

test/channels.jl
vtjnash: Without this, the test depends on the scheduling (this test assumes that this statement takes a least a millisecond to complete). This completely fixes the test to avoid that flakiness by ensuring it always takes at least that long.

src/gf.c
vtjnash: It is in characters in this PR right now. But conceptually, they don't share a common root, so I've listed them separately. That way, if someone alters one, it won't affect the other.

vtjnash:
I'd like to plan to merge this tomorrow (after https://github.com/JuliaLang/julia/pull/36260). Please let me know if review is incomplete.

test/channels.jl
Keno: Ah, I see.

src/gf.c
Keno: Alright, you're the methodcache czar, it just looked odd as is. Maybe there should be helpers that make clear in which sense this is used, even if the implementation of these helpers is the same at the moment?

vtjnash: True, I'll keep that in mind. It won't be the last time (even this month) that I edit the code, haha.

JeffBezanson:
Looks fine to me. No reason to block merging, but I have a couple thoughts.

I wonder if we can take this farther to memoize more of `abstract_call_gf_by_type` --- i.e. avoid the "double lookup" of ml_matches followed by typeinf_edge.

src/gf.c
```diff
@@ -2346,16 +2394,23 @@ STATIC_INLINE jl_method_instance_t *jl_lookup_generic_(jl_value_t *F, jl_value_t
...
     i = 4;
-    // if no method was found in the associative cache, check the full cache
+    jl_tupletype_t *tt = NULL;
+    int64_t last_alloc;
     if (i == 4) {
+        // if no method was found in the associative cache, check the full cache
         JL_TIMING(METHOD_LOOKUP_FAST);
         mt = jl_gf_mtable(F);
         entry = jl_typemap_assoc_exact(mt->cache, F, args, nargs, jl_cachearg_offset(mt), world);
+        if (entry == NULL) {
```
JeffBezanson, line 2404: I suspect we'll lose a bit of performance from needing two lookups here sometimes. Would be nice to be able to combine the tables somehow.

src/gf.c
```diff
@@ -2803,10 +2861,54 @@ static jl_value_t *ml_matches(jl_typemap_t *defs, int offs,
...
+        jl_typemap_entry_t *entry = lookup_leafcache(leafcache, (jl_value_t*)type, world);
+        if (entry) {
+            jl_method_instance_t *mi = entry->func.linfo;
+            jl_method_t *meth = mi->def.method;
+            if (jl_egal((jl_value_t*)type, mi->specTypes)) {
+                env.match.env = mi->sparam_vals;
+                env.match.ti = mi->specTypes;
+            }
+            else {
+                // TODO: should we use jl_subtype_env instead (since we know that `type <: meth->sig` by transitivity)
```
JeffBezanson, line 2878: :+1:, plus we can skip this entirely if there are no static params.

vtjnash: Oh, wow, yeah. I'm also thinking of removing this value entirely from the results (since often it should be available from the later typeinf_edge cache lookup)

src/gf.c
vtjnash: That would be likely to hurt performance and memory usage to combine them (had that situation in a intermediate state of this PR, before I finished separating the tables). What we're doing here is expecting that one of these two tables is most likely empty (functions usually either get specialized on leaf types entirely or they don't)—and so we're just going to bypass the previous table in most cases without even attempting a lookup.

Issue #36087, opened 2020-05-30
Splat of Integer needs to be special cased


BenJWard:
Hi, this issue follows a conversation with @vchuravy and @Keno, who advised I open an issue.

Here's my version info:

```julia
Julia Version 1.5.0-beta1.0
Commit 6443f6c95a (2020-05-28 17:42 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin19.4.0)
  CPU: Intel(R) Core(TM) i7-6920HQ CPU @ 2.90GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-9.0.1 (ORCJIT, skylake)
```

So, I've been trying to implement a type using NTuples, instead of ever wider primitive types.
I successfully made a function called `shiftright` that shifts all of the 2-bit encoded elements in a NTuple one position to the right, and it is as performant as the equivalent shifting of a UInt128:

```julia
using BioSequences, BenchmarkTools
# In BioSequences.jl kmers are called Mer, not Kmer so there is no name clash.
struct Kmer{A<:NucleicAcidAlphabet{2},K,N}
    data::NTuple{N,UInt64}
end
# Shortcut
const DNAKmer{K,N} = Kmer{DNAAlphabet{2},K,N}

@inline capacity(::Type{Kmer{A,K,N}}) where {A,K,N} = div(64N, 2)

function (::Type{Kmer{A,K,N}})(seq::LongSequence{A}) where {A<:NucleicAcidAlphabet{2},K,N}
    seqlen = length(seq)
    if seqlen != K
        throw(ArgumentError("seq does not contain the correct number of nucleotides ($seqlen ≠ $K)"))
    end
    if seqlen > capacity(Kmer{A,K,N})
        throw(ArgumentError("Cannot build a mer longer than $(capacity(Kmer{A,K,N}))bp long"))
    end

    # Construct the head.
    bases_in_head = div(64 - (64N - 2K), 2)
    head = zero(UInt64)
    @inbounds for i in 1:bases_in_head
        nt = convert(eltype(typeof(seq)), seq[i])
        head = (head << 2) | UInt64(BioSequences.twobitnucs[reinterpret(UInt8, nt) + 0x01])
    end
    
    # And the rest of the sequence
    idx = Ref(bases_in_head + 1)
    
    tail = ntuple(Val{N - 1}()) do i
        Base.@_inline_meta
        body = zero(UInt64)
        @inbounds for i in 1:32
[...]

Issue #29114, opened 2018-09-10
Performance of splatting a number


KristofferC:
In a few places, the performance of splatting a number has shown up: https://github.com/JuliaLang/julia/pull/29060,  https://github.com/JuliaLang/julia/pull/28764#issuecomment-419917783 and working around this give signficiant speed-ups.
The code is well inferred but there is a call to `_apply` instead of the actual method. This dissapears if the number is wrapped in a tuple.

It would be nice to have this fixed so these type of workarounds are no longer needed.

Ref https://github.com/JuliaLang/julia/pull/27434#issuecomment-400335764

cc @mbauman, @martinholters

Keno:
We could special case numbers as we do tuples in the inliner. The more generic solution would have to emulate the iteration protocol.

mbauman:
Forgive me if this is a naive understanding of inference, but it seems like there's a middle-ground here — isn't it fairly easy to see that `iterate(::X,::Any)::Const(nothing)` for `X`s that only iterate once.

Keno:
Yes, but that's still emulating the iteration protocol in the optimizer (inference already does some of this) but the iterate isn't part of the code (since _apply does it internally). Basically you need to implement this TODO: https://github.com/JuliaLang/julia/blob/master/base/compiler/ssair/inlining.jl#L831
or special case it there.

c42f:
I just noticed we share basically the same problem over at https://github.com/JuliaArrays/StaticArrays.jl/issues/361.

What if we change lowering of splatting so that `f(x...)` turns into something like `Core._apply(f, Base.splatarg(x))`? We can have `Base.splatarg(x) = x` by default (or possibly collect into a tuple via iteration?)  For some more trivial cases like `Number` and `StaticVector` a couple of overrides would solve this issue:

```julia
Base.splatarg(x::Number) = (x,)
Base.splatarg(v::StaticVector) = Tuple(v)
```

This does mean that splatting would become configurable by user types but that seems generally consistent with the way we usually lower syntax into function calls.

yurivish:
Nice idea. Maybe it can be `Base.splattable` for consistency with `Base.broadcastable`.

c42f:
> Base.splattable ...

Interesting analogy; we could go with it, though I think the name is less evocative than `splatarg`.

[...]

PR #36169, opened 2020-06-06
Inline singleton splats

Keno:
As noted in #36087 and #29114, splatting integers currently has a
performance penalty that is unexpected. For tuples and SimpleVectors,
we have special purpose inliners that will simply inline the
tuple/SimpleVector into the call being splatted. However, for
everything else we'd have to run the iteration protocol to find
out what the values to substitute are. This change does just that,
limited to the case of length-1 (and empty) iterables. Benchmark:

```
f(x) = (x...,)
@code_typed f(1)
@benchmark f(1)
```

Before:
```
julia> @code_typed f(1)
CodeInfo(
1 ─ %1 = Core._apply_iterate(Base.iterate, Core.tuple, x)::Tuple{Int64}
└──      return %1
) => Tuple{Int64}

julia> @benchmark f(1)
BenchmarkTools.Trial:
  memory estimate:  32 bytes
  allocs estimate:  2
  --------------
  minimum time:     209.357 ns (0.00% GC)
  median time:      213.404 ns (0.00% GC)
  mean time:        218.674 ns (0.16% GC)
  maximum time:     1.922 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     540
```

After:
```
julia> @code_typed f(1)
CodeInfo(
1 ─ %1 = invoke Base.iterate(_2::Int64)::Tuple{Int64,Nothing}
│   %2 = (getfield)(%1, 1)::Int64
│   %3 = (getfield)(%1, 2)::Nothing
│        invoke Base.iterate(_2::Int64, %3::Nothing)::Nothing
│   %5 = Core.tuple(%2)::Tuple{Int64}
└──      return %5
) => Tuple{Int64}

julia> @benchmark f(1)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     3.044 ns (0.00% GC)
  median time:      3.047 ns (0.00% GC)
  mean time:        3.049 ns (0.00% GC)
  maximum time:     7.700 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000
```

Obviously this isn't 100% optimal yet, because the `iterate` calls themselves
don't get inlined, but it's a lot better. Inlining the `iterate` calls is
left for a follow up commit.

### DIFF
--- a/base/compiler/ssair/inlining.jl
+++ b/base/compiler/ssair/inlining.jl
@@ -594,50 +594,81 @@ function spec_lambda(@nospecialize(atype), sv::OptimizationState, @nospecialize(
     return mi
 end
 
+struct ExpandTuple; end
+struct InlineIterate
+    first::Tuple{MethodInstance, Any, UInt, UInt}
+    second::Union{Nothing, Tuple{MethodInstance, Any, UInt, UInt}}
+end
+
+function add_backedges!(rewrite::InlineIterate, sv::OptimizationState)
+    add_backedge!(rewrite.first[1], sv)
+    update_valid_age!(rewrite.first[3], rewrite.first[4], sv)
+    if rewrite.second !== nothing
+        add_backedge!(rewrite.second[1], sv)
+        update_valid_age!(rewrite.second[3], rewrite.second[4], sv)
+    end
+end
+
 # This assumes the caller has verified that all arguments to the _apply call are Tuples.
-function rewrite_apply_exprargs!(ir::IRCode, idx::Int, argexprs::Vector{Any}, atypes::Vector{Any}, arg_start::Int)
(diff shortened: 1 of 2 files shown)


martinholters:
I know my way around the optimizer too little to review in detail, but IIUC `analyze_type_for_apply_rewrite` duplicates a bit of work already done in `abstract_iteration` during inference. Is that just due to a lack of infrastructure to pipe the information from the latter to the former, or is there some general need to re-establish the types obtained during iteration? (Not that it matters much now, but if we want to get fancier with the inliner in the future, it would be unfortunate to duplicate more of that logic.)

Keno:
Yes, we don't currently have any infrastructure to forward sideband information from inference to the optimizer (other than the cache for method lookups that we are using here). I'd like to improve that in the future, but for now this is the best we can do.

martinholters:
Thanks for confirming. (And by all means, that does not mean I disapprove of this PR in its current form in any way.)

Keno:
Superseded by #36684

PR #36173, opened 2020-06-06
REPL prefix search: add ^x "pass through" key

rfourquet:
This fix the following bug:

1. press "Ctrl-P" to go to previous history entry, through "prefix history search"
2. "^x"

Then the Julia session is exited ("^x" is a prefix keyboard shortcut, e.g. "^x^x" does "exchange-mark-and-cursor").
This PR is a very similar to #33805. 

This is not a complete solution: if you have a 3-key combo as a keyboard shortcut, e.g. "^x^lb", then the session will again terminate. Then we could add ` "^x**" => "*",` as a "pass-through", but
1. the problem is not solved for bigger n-keys combos
2. "^x^x" doesn't work well anymore: after pressing "^x^x", the REPL first waits for a third key before processing, which is not wanted.

So let's assume 3+-key combos are not used very much.
I'm really looking forward to fixing this "pass through" problem in a general way...

### DIFF
--- a/stdlib/REPL/src/LineEdit.jl
+++ b/stdlib/REPL/src/LineEdit.jl
@@ -2218,6 +2218,7 @@ const prefix_history_keymap = merge!(
             match_input(map, s, IOBuffer(c))(s, keymap_data(ps, mode(s)))
         end,
         # match escape sequences for pass through
+        "^x*" => "*",
         "\e*" => "*",
         "\e[*" => "*",
         "\eO*"  => "*",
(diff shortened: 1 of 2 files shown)


StefanKarpinski:
Who would be good to review this?

rfourquet:
> Who would be good to review this?

Maybe @vtjnash who IIRC made the change that makes Julia exit instead of getting a runtime "REPL keymap" error?

Keno:
These kind of subtle behavior changes need a REPL test, otherwise, they're basically guaranteed to regress.

rfourquet:
I'm quite guilty as I knew tests were missing, it's just frustrating with REPL fixes to make the fix in 5 min and spend an order of magnitude more time trying to sort out how to test these things :-/ 
Anyway, that's done now, and I will mark this for backport-1.5.

rfourquet:
I will merge in a couple of days if no objection.

PR #36175, opened 2020-06-07
RFC: Define API invariance for keys, values, getindex and pairs

### LINKED ISSUE #34851: Define pairs(A) = enumerate(A)
The method for `findall` with no specialization on the collection type is
```julia
findall(testf::Function, A) = collect(first(p) for p in pairs(A) if testf(last(p)))
```

For a singly-linked list, like `DataStructures.Cons`, `findall` fails because there is no method `keys(::Cons)`. Constructing, say linear indices for `Cons` requires traversing the list to find it's length; more precisely, I don't see a way around traversal. Writing a fallback method

```julia
pairs(A) = enumerate(A)
```

would solve the problem for any iterable, including `Cons`. This assumes that
by default the keys for a collection are `1, 2, ...`.
In the case of `Cons`, it's possible that traversing twice, preallocating output after the first pass, would be more efficient. I did not try this, but I doubt it is more efficient.

See https://github.com/JuliaCollections/DataStructures.jl/issues/580 .

A somewhat related issue: I think the single argument method for `findall`, which is currently

```julia
findall(A) = collect(first(p) for p in pairs(A) if last(p))
```

would be better refactored as
```julia
findall(A) = findall(isequal(true), A)
```

tkf:
There are discussions for extending `keys` #34678 and `pairs` #34851 API.  Rather than extending the implementation purely for convenience, I suggest to document the expected relationships between the functions.  This clarifies what to expect from these APIs without looking at each implementation.  Otherwise, it is difficult to use these functions in generic code.

Concretely, for `keys`, `values`, and `getindex` I suggest:

> It is highly recommended that each `collection` that implements `keys` to satisfy the following relationship between `values` and `getindex`:
>
> ```julia
> isequal(collect(values(collection)), [collection[k] for k in keys(collection)])
> ```

For `pairs`:

> It is highly recommended that each implementation of `pairs` satisfies the following relationships:
>
> ```julia
> isequal(map(first, paris(collection)), keys(collection))
> isequal(map(last, paris(collection)), values(collection))
> ```
>
> The fallback generic implementation satisfy this invariance.

Since it is a breaking change to impose more strict API requirements for already existing and used API, I only used a very loose phrasing "It is highly recommended."  It is up to discussion if upgrading this to "must" is a `minor change`.

close #34851

### DIFF
--- a/base/abstractdict.jl
+++ b/base/abstractdict.jl
@@ -69,6 +69,16 @@ in(k, v::KeySet) = get(v.dict, k, secret_table_token) !== secret_table_token
 
 For an iterator or collection that has keys and values (e.g. arrays and dictionaries),
 return an iterator over the keys.
+
+# Implementation
+
+It is highly recommended that each `collection` that implements `keys`
+to satisfy the following relationship between [`values`](@ref) and
+[`getindex`](@ref):
+
+```julia
+isequal(collect(values(collection)), [collection[k] for k in keys(collection)])
+```
 """
 function keys end
 
@@ -130,6 +140,18 @@ values(a::AbstractDict) = ValueIterator(a)
 Return an iterator over `key => value` pairs for any
 collection that maps a set of keys to a set of values.
 This includes arrays, where the keys are the array indices.
+
+# Implementation
+
+It is highly recommended that each implementation of `pairs` satisfies
+the following relationships:
+
+```julia
+isequal(map(first, paris(collection)), keys(collection))
+isequal(map(last, paris(collection)), values(collection))
+```
+
+The fallback generic implementation satisfy this invariance.
 """
 pairs(collection) = Generator(=>, keys(collection), values(collection))
 
--- a/base/array.jl
+++ b/base/array.jl
@@ -804,6 +804,11 @@ Dict{String,Int64} with 2 entries:
 julia> getindex(A, "a")
 1
 ```
+
+# Implementation
+
+Implementation of `getindex` should satisfy the invariance explained
+in [`keys`](@ref).
 """
 function getindex end
 
--- a/base/essentials.jl
+++ b/base/essentials.jl
@@ -758,6 +758,11 @@ julia> values([2])
 1-element Array{Int64,1}:
  2
 ```
+
+# Implementation
+
+Implementation of `values` should satisfy the invariance explained in
+[`keys`](@ref).
 """
 values(itr) = itr
 


tkf:
### A case study: `Base.Generator`

With the API proposed, some care must be taken when defining `keys`, `values` and `getindex` for `Base.Generator` (#34678).  This is because it's very easy to use stateful mapping with the generator expression:

```julia
collection = let counter = 0
    (
        begin
            counter += 1
            (counter, x)
        end for x in 1:10
    )
end
```

One possible solution is to clarify the condition in which `keys` etc. for `Base.Generator` are usable:

> The generator object is "indexable" (i.e., `getindex`, `keys`, `values`, and `pairs` behave consistently) if the input collection is also indexable and the mapping is stateless in the sense that the same input produces the same output.

(This can be added in the docstring of `Iterators.map` #34352).

It might be useful to define and mention `axes` and its related methods for `Base.Generator`.

I think it makes sense to define more methods for `Base.Generator` assuming the purity of the mapping function and then explain on which conditions certain interface can be used consistently.  For example, such "narrow contract" API is required for supporting parallelism with the generator expressions via [Transducers.jl](https://github.com/JuliaFolds/Transducers.jl)/[SplittablesBase.jl](https://github.com/JuliaFolds/SplittablesBase.jl).

rapus95:
While clarification is always good(!) I think that the current understanding of Generators expects them to only depend on their input arguments (and no hidden state) since, strictly speaking, there are no guarantees/claims about the order in which it'll be iterated.
```julia
Generator(f, iter)

  Given a function f and an iterator iter, construct an iterator that yields the values of f applied to the elements
  of iter.
```

tkf:
Thanks, that's a good point. Strictly speaking, I don't think `Base.Generator`'s docstring is a public API because it's not included in the documentation (discussion about this in #35715). But at least it indicates that the intended use-case is the pure mapping. OTOH, something like `mean(exp(randn()) for _ in 1:10)` is so innocent-looking and I don't think it's reasonable to "ban" this kind of use-cases.

rapus95:
To clarify, in the example you gave most recent, I think having `keys` defined is definitively good/important for convenience, but as the function passed to the Generator is not pure, the invariance you suggested isn't expected to hold. That way we keep the Generator as a very convenient construct for stateful iterators while still increasing the overall consistency in the stateless case.

base/abstractdict.jl
```diff
@@ -130,6 +140,18 @@ values(a::AbstractDict) = ValueIterator(a)
...
 This includes arrays, where the keys are the array indices.
+
+# Implementation
+
+It is highly recommended that each implementation of `pairs` satisfies
+the following relationships:
+
+```julia
+isequal(map(first, paris(collection)), keys(collection))
+isequal(map(last, paris(collection)), values(collection))
```
KlausC, line 151: Note, that currently neither `Array` nor `Dict` satisfies these relationships, because `map` is not defined for the return types of `pairs`:
```
julia> coll = [10, 20];
julia> map(first, pairs(coll))
ERROR: map is not defined on dictionaries
```

KlausC:
> strictly speaking, there are no guarantees/claims about the order in which it'll be iterated.

IMO the claim, that the order of the generated iterator follows the order of the base iterator is so natural, that it is expected by most users. In my understanding it is implied in the description.

That would be consistent with the documentation of `map` and `Array comprehension`, which also do not explicitly mention, that the order is preserved. I would always expect these invariants for all iterables:
```
collect(f(x) for x in itr) == collect(Generator(f, itr)) == map(f, itr) == [f(x) for x in itr]
```

base/abstractdict.jl
```diff
@@ -130,6 +140,18 @@ values(a::AbstractDict) = ValueIterator(a)
...
 collection that maps a set of keys to a set of values.
 This includes arrays, where the keys are the array indices.
+
+# Implementation
+
+It is highly recommended that each implementation of `pairs` satisfies
+the following relationships:
+
+```julia
+isequal(map(first, paris(collection)), keys(collection))
```
jonas-schulze, line 150: 🇫🇷

DilumAluthge:

PR #36176, opened 2020-06-07
a few compiler improvements

JeffBezanson:
- improve fieldtype_nothrow for vararg tuple types
- remove UnionAll construction code from tuple_type_head
- add specializations to avoid some common generated functions

This shaves about 0.5 sec off the TTFP. The first two items fix some disastrous code for vararg tuple conversion:
```
julia> code_typed(convert, (Type{Tuple{Vararg{Int}}},NTuple{2,Int8}))
1-element Array{Any,1}:
 CodeInfo(
1 ─      invoke Core.TypeVar(Symbol("#s2")::Symbol, Base.Tuple::Any)::TypeVar
│        Base.fieldtype(T, 1)::Type{Int64}
│   %3 = Base.getfield(x, 1, true)::Int8
│   %4 = Core.sext_int(Core.Int64, %3)::Int64
│   %5 = (getfield)(x, 2)::Int8
│        invoke Core.TypeVar(Symbol("#s2")::Symbol, Base.Tuple::Any)::TypeVar
│        Base.fieldtype(T, 1)::Type{Int64}
│   %8 = Core.sext_int(Core.Int64, %5)::Int64
│   %9 = Core.tuple(%4, %8)::Tuple{Int64,Int64}
└──      return %9
) => Tuple{Int64,Int64}
```

### DIFF
--- a/base/compiler/tfuncs.jl
+++ b/base/compiler/tfuncs.jl
@@ -876,12 +876,12 @@ function _fieldtype_nothrow(@nospecialize(s), exact::Bool, name::Const)
     isa(fld, Int) || return false
     ftypes = datatype_fieldtypes(u)
     nf = length(ftypes)
-    (fld >= 1 && fld <= nf) || return false
     if u.name === Tuple.name && fld >= nf && isvarargtype(ftypes[nf])
-        # The length of the tuple will be determined at runtime, we can't say
-        # anything
-        return false
+        # If we don't know the exact type, the length of the tuple will be determined
+        # at runtime and we can't say anything.
+        return exact
     end
+    (fld >= 1 && fld <= nf) || return false
     return true
 end
 
--- a/base/multidimensional.jl
+++ b/base/multidimensional.jl
@@ -758,8 +758,7 @@ function _unsafe_getindex(::IndexStyle, A::AbstractArray, I::Vararg{Union{Real, 
     return dest
 end
 
-# Always index with the exactly indices provided.
-@generated function _unsafe_getindex!(dest::AbstractArray, src::AbstractArray, I::Vararg{Union{Real, AbstractArray}, N}) where N
+function _generate_unsafe_getindex!_body(N::Int)
     quote
         @_inline_meta
         D = eachindex(dest)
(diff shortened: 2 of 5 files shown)


base/compiler/tfuncs.jl
```diff
@@ -877,7 +877,7 @@ function _fieldtype_nothrow(@nospecialize(s), exact::Bool, name::Const)
     ftypes = datatype_fieldtypes(u)
     nf = length(ftypes)
     (fld >= 1 && fld <= nf) || return false
```
Keno, line 879: This seems imprecise in the `exact` case that you're fixing just below.

base/multidimensional.jl
```diff
@@ -776,6 +776,34 @@ end
     end
 end
 
+# manually written-out specializations for 1 and 2 arguments to save compile time
+@inline function _unsafe_getindex!(dest::AbstractArray, src::AbstractArray, I_1::Union{Real, AbstractArray})
```
Keno, line 780: Can we just make the body here a macro and use that three times. I'm just worried that we'll end up changing one of these and forgetting about the others.

timholy:
Worth pointing out that we might not even need those generated functions if we fix #9080.

base/compiler/tfuncs.jl
```diff
@@ -876,12 +876,12 @@ function _fieldtype_nothrow(@nospecialize(s), exact::Bool, name::Const)
...
     ftypes = datatype_fieldtypes(u)
     nf = length(ftypes)
-    (fld >= 1 && fld <= nf) || return false
     if u.name === Tuple.name && fld >= nf && isvarargtype(ftypes[nf])
-        # The length of the tuple will be determined at runtime, we can't say
-        # anything
-        return false
+        # If we don't know the exact type, the length of the tuple will be determined
+        # at runtime and we can't say anything.
+        return exact
```
Keno, line 882: This still doesn't seem right. `fieldtype(NTuple{10, T} where T, 20)` certainly does error.

base/multidimensional.jl
```diff
@@ -776,6 +775,20 @@ end
     end
 end
 
+# Always index with the exactly indices provided.
+@generated function _unsafe_getindex!(dest::AbstractArray, src::AbstractArray, I::Vararg{Union{Real, AbstractArray}, N}) where N
+    _generate_unsafe_getindex!_body(N)
```
Keno, line 780: Nice!

base/compiler/tfuncs.jl
JeffBezanson: That doesn't contain a `Vararg` but I guess it might in the future, so I can add it.

Keno: Fair enough. I guess we didn't change this yet.

PR #36183, opened 2020-06-07
Reorganize logmsg_code and line-length fixes

oxinabox:
I was considering maybe changing something in the `logmsg_code`.
but that function was uncomfortably long and put a lot of work into the processing of the arguments, which is fairly complex.
So I moved that into its own function, and tried to simplify it somewhat

Then since i was doing this kind of housekeeping i also went through and made to follow the [92 character line limit we recommend in CONTRIBUTING.md](https://github.com/JuliaLang/julia/blob/f5103cd1ae3dcc6f3b32f4c31d5136d28e631443/CONTRIBUTING.md#general-formatting-guidelines-for-julia-code-contributions)

### DIFF
--- a/base/logging.jl
+++ b/base/logging.jl
@@ -307,8 +307,10 @@ function logmsg_code(_module, file, line, level, message, exs...)
     end
 end
 
-function process_logmsg_exs(_module, _file, _line, level, message, exs...)
-    log_data = (;_module, _file, _line, kwargs=Any[])
+function process_logmsg_exs(_orig_module, _file, _line, level, message, exs...)
+    local _group, _id
+    _module = _orig_module
+    kwargs = Any[]
     for ex in exs
         if ex isa Expr && ex.head === :(=) && ex.args[1] isa Symbol
             k,v = ex.args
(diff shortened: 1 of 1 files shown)


base/logging.jl
```diff
@@ -347,6 +307,50 @@ function logmsg_code(_module, file, line, level, message, exs...)
...
+    for ex in exs
+        if ex isa Expr && ex.head === :(=) && ex.args[1] isa Symbol
+            k,v = ex.args
+            if !(k isa Symbol)
+                throw(ArgumentError("Expected symbol for key in key value pair `$ex`"))
+            end
+
+            # Recognize several special keyword arguments
+            if k ∈ (:_group, :_id, :_module, :_file, :_line)
+                log_data = Base.setindex(log_data, esc(v), k)
```
oxinabox, line 321: I think this is a reasonable enough use of `setindex`.
Like worse case senario is it copies at 4+5+6+7+8 things if everything is overridden.
and this only happens during the macro expansion anyway.

I think this is clearer than having 10 lines of `if,elseif...` because it is clear in these 2 lines that they are all special in the same way .

base/logging.jl
```diff
@@ -347,6 +307,50 @@ function logmsg_code(_module, file, line, level, message, exs...)
     end
 end
 
+function process_logmsg_exs(_module, _file, _line, level, message, exs...)
+    log_data = (;_module, _file, _line, kwargs=Any[])
```
oxinabox, line 311: my first time using names tuples that have defaulted keys, it is pretty slick.

base/logging.jl
```diff
@@ -268,75 +276,27 @@ default_group(file) = Symbol(splitext(basename(file))[1])
...
-            QuoteNode(default_group(file))
-        else
-            # memoized run-time execution
-            ref = Ref{Symbol}()
-            :(isassigned($ref) ? $ref[]
-                               : $ref[] = default_group(something($file, "")))
-        end
-    end
-
+    log_data = process_logmsg_exs(_module, file, line, level, message, exs...)
```
oxinabox, line 279: I think it is nicer to have 1 log_data object to hold all the data about the log, other than the level and message.
(tempting to put them in there too).

base/logging.jl
c42f: I'm not really sure this is an improvement in readability with having to use `setindex` like this. I mean I guess we could go with it, but overall I think it was quite straightforward before and extremely easy to see that all these five cases were the same. More verbose than unreadable ;-)

On the other hand bundling all these together into `log_data` at the bottom of process_logmsg_exs seems neat.

oxinabox: Yeah I guess I can just flip the order and do it that way

c42f:
Looks good to me, thanks.

Issue #36185, opened 2020-06-07
SIGSEGV and StackOverflow from dispatching on ::Type{T}, ::Array{Union{T,Missing},N})


oxinabox:
EDIT: MWE from below
```
 julia> typeintersect((Tuple{Type{T},Array{Union{T,Missing},N}} where {T,N}),
                     (Tuple{Type{T},Array{Union{T,Nothing},N}} where {T,N}))
ERROR: StackOverflowError
```

Original text:
---

If I do
```julia
for U in (:(Union{T, Missing}), :(Union{T, Nothing}),)
    @eval function bar(::Type{T}, xs::Array{$U,N}) where {T,N}
        T
    end
end
```

Then julia `SIGSEGV`'s (Address boundary error)

When I do:
```julia
@eval function bazz(::Type{T}, xs::Array{Union{T,Missing},N}) where {T,N}
   T
end
```
I get a `StackOverflowError`

When I do:
```
function bazz(::Type{T}, xs::Array{Union{T,Missing},N}) where {T,N}
   T
end
```
I am back to crashing out with `SIGSEGV`'s (Address boundary error)

But if i do:
```
function bazz(::Type{<:T},xs::Array{Union{T,Missing},N}) where {T,N}
   T
end
```
then it is fine.
But that is a different expression

as is
```
function bazz(::Type{T},xs::Array{<:Union{T,Missing},N}) where {T,N}
   T
end
```

This is on `Version 1.6.0-DEV.184 (2020-06-06)`

MasonProtter:
On 1.5.0-beta1, I don't get any `SIGSEV`s, but I do get `StackOverflowError`s. The `eval` appears to be unrelated to the error so far as I can tell. I get the same error with or without `eval`. 
```julia
julia> for U in (:(Union{T, Missing}), :(Union{T, Nothing}),)
           @eval function bar(::Type{T}, xs::Array{$U,N}) where {T,N}
               T
           end
       end
ERROR: StackOverflowError:
Stacktrace:
 [1] top-level scope at none:1
 [2] eval(::Module, ::Any) at ./boot.jl:331
 [3] top-level scope at REPL[4]:2

julia> @eval function bazz(::Type{T}, xs::Array{Union{T,Missing},N}) where {T,N}
          T
       end
bazz (generic function with 1 method)

julia> @eval function bar(::Type{T}, xs::Array{Union{T,Nothing},N}) where {T,N}
          T
       end
ERROR: StackOverflowError:
[...]

Issue #31152, opened 2019-02-22
Allow reinterpret(Union{Missing, T}, ::Vector{Union{Missing, U}})


cstjean:
This works:

```julia
julia> v = [1,2,3,4]
4-element Array{Int64,1}:
 1
 2
 3
 4

julia> reinterpret(Some{Int64}, v)
4-element reinterpret(Some{Int64}, ::Array{Int64,1}):
 1
 2
 3
 4
```
It would be very nice if this could be supported, too:
```julia
julia> v2 = [1,2,3,4,missing]
5-element Array{Union{Missing, Int64},1}:
 1       
 2       
 3       
 4       
  missing

julia> reinterpret(Union{Missing, Some{Int64}}, v)
ERROR: ArgumentError: cannot reinterpret `Int64` `Union{Missing, Some{Int64}}`, type `Union{Missing, Some{Int64}}` is not a bits type
Stacktrace:
 [1] (::getfield(Base, Symbol("#throwbits#200")))(::Type{Int64}, ::Type{Union{Missing, Some{Int64}}}, ::Type{Union{Missing, Some{Int64}}}) at ./reinterpretarray.jl:16
 [2] reinterpret(::Type{Union{Missing, Some{Int64}}}, ::Array{Int64,1}) at ./reinterpretarray.jl:33
 [3] top-level scope at none:0
```
I've explained the use case [on discourse](https://discourse.julialang.org/t/adding-a-unitful-unit-to-a-vector-with-missings-is-slow/21085). Briefly, when working with Unitful vectors, one often has to [take out the units](http://ajkeller34.github.io/Unitful.jl/stable/manipulations/#Unitful.ustrip) (eg. to make a linear regression), then tack them back on the result. Both operations could conceivably be done allocation-free with `reinterpret`, but are very slow on vectors that contain `missing`s.

nalimilan:
See https://github.com/JuliaLang/julia/issues/26681 about a similar issue with `convert`.

PR #36186, opened 2020-06-07
WIP: Allow convert and reinterpret for Arrays of unions with missing/nothing

### LINKED ISSUE #26681: Feature request: convert between Array{Union{T, Missing}, N} and Array{T, N} without copying
As discussed in [this thread](https://discourse.julialang.org/t/missing-data-and-namedtuple-compatibility/8136/91?) , it'd be useful to allow converting between `Array{Union{T, Missing}, N}` and `Array{T, N}` without copying the data whenever possible. From (my understanding of) the discourse discussion, it seems that  conversion from`Array{Union{T, Missing}, N}` to `Array{T, N}` should always be possible without copying whereas the other direction depends on whether the already allocated region is big enough to fit the wider type.

oxinabox:
Closes https://github.com/JuliaLang/julia/issues/26681 and #31152

Right now convert kind of breaks the compiler and i am not sure why.
Debugging that is pending, so right now it is commented out so i could run the `reinterpet` tests

Writing this is made harder by https://github.com/JuliaLang/julia/issues/36185

### DIFF
--- a/base/missing.jl
+++ b/base/missing.jl
@@ -85,10 +85,10 @@ for Cs in ((:Missing,), (:Nothing,), (:Missing, :Nothing))
         ) where {N} where {T<:Q} where Q
 
         all(x->x isa T, xs) || throw(ArgumentError("cannot convert $(eltype(xs)) to $T"))
-        if isbitstype(T) &&
+        if isbitstype(T) && T===Q
             reinterpret(T, xs)
         else
-            T[x for x in xs]
+            Q[x for x in xs]
         end
     end
     ==#
--- a/base/missing.jl
+++ b/base/missing.jl
@@ -69,6 +69,33 @@ convert(::Type{T}, x::T) where {T>:Union{Missing, Nothing}} = x
 convert(::Type{T}, x) where {T>:Missing} = convert(nonmissingtype_checked(T), x)
 convert(::Type{T}, x) where {T>:Union{Missing, Nothing}} = convert(nonmissingtype_checked(nonnothingtype_checked(T)), x)
 
+for Cs in ((:Missing,), (:Nothing,), (:Missing, :Nothing))
+    @eval function Base.reinterpret(
+        ::Type{<:T},  # Note this is effectively same as Type{T} since will error if T is abstract
+        xs::Array{Union{T, $(Cs...)},N}
+    ) where {T,N}
+        isbitstype(T) || throw(ArgumentError("cannot reinterpret `$(eltype(xs))`, type `$(T)` is not a bits type"))
+        return unsafe_wrap(Array{T,N}, Ptr{T}(pointer(xs)), length(xs))
+    end
+
+    #==
+    @eval function Base.convert(
+        ::Type{Array{Q,N}},
+        xs::Array{Union{T, $(Cs...)},N}
+        ) where {N} where {T<:Q} where Q
+
+        all(x->x isa T, xs) || throw(ArgumentError("cannot convert $(eltype(xs)) to $T"))
+        if isbitstype(T) &&
+            reinterpret(T, xs)
+        else
+            T[x for x in xs]
+        end
+    end
+    ==#
+end
+# Resolve ambiguity
+#Base.convert(::Type{Array{Any,1}}, xs::Array{Any,1}) = xs
+
 
 # Comparison operators
 ==(::Missing, ::Missing) = missing
--- a/test/missing.jl
+++ b/test/missing.jl
@@ -23,6 +23,48 @@ end
     @test_throws MethodError convert(Union{Int, Missing}, "a")
 end
 
+mutable struct NotABitsType a end
+==(n1::NotABitsType,n2::NotABitsType) = n1.a == n2.a
+
+@testset "convert/reinterpret Arrays of Unions" begin
+    unions(T) = (Union{T, Missing}, Union{T, Nothing}, Union{T, Missing, Nothing})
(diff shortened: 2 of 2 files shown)


base/regex.jl
```diff
@@ -168,6 +168,10 @@ function getindex(m::RegexMatch, name::Symbol)
 end
 getindex(m::RegexMatch, name::AbstractString) = m[Symbol(name)]
 
+iterate(m::RegexMatch, args...) = iterate(m.captures, args...)
+length(m::RegexMatch) = length(m.captures)
+eltype(m::RegexMatch) = eltype(m.captures)
```
oxinabox, line 173: what are these changes doing here?

base/missing.jl
```diff
@@ -69,6 +69,33 @@ convert(::Type{T}, x::T) where {T>:Union{Missing, Nothing}} = x
...
+        if isbitstype(T) &&
+            reinterpret(T, xs)
+        else
+            T[x for x in xs]
+        end
+    end
+    ==#
+end
+# Resolve ambiguity
+#Base.convert(::Type{Array{Any,1}}, xs::Array{Any,1}) = xs
```
oxinabox, line 97: I don't know why this is ambiguous, and the answer would probably go a long way to explaining why this breaks so much

base/missing.jl
```diff
@@ -69,6 +69,33 @@ convert(::Type{T}, x::T) where {T>:Union{Missing, Nothing}} = x
 convert(::Type{T}, x) where {T>:Missing} = convert(nonmissingtype_checked(T), x)
 convert(::Type{T}, x) where {T>:Union{Missing, Nothing}} = convert(nonmissingtype_checked(nonnothingtype_checked(T)), x)
 
+for Cs in ((:Missing,), (:Nothing,), (:Missing, :Nothing))
+    @eval function Base.reinterpret(
```
oxinabox, line 73:
```suggestion
    @eval function reinterpret(
```

base/missing.jl
```diff
@@ -69,6 +69,33 @@ convert(::Type{T}, x::T) where {T>:Union{Missing, Nothing}} = x
...
+    end
+
+    #==
+    @eval function Base.convert(
+        ::Type{Array{Q,N}},
+        xs::Array{Union{T, $(Cs...)},N}
+        ) where {N} where {T<:Q} where Q
+
+        all(x->x isa T, xs) || throw(ArgumentError("cannot convert $(eltype(xs)) to $T"))
+        if isbitstype(T) &&
```
oxinabox, line 88:
```suggestion
        if isbitstype(T) && T==Q
```

base/missing.jl
```diff
@@ -69,6 +69,33 @@ convert(::Type{T}, x::T) where {T>:Union{Missing, Nothing}} = x
...
+            reinterpret(T, xs)
+        else
+            T[x for x in xs]
```
oxinabox, line 91:
```suggestion
            Q[x for x in xs]
```

oxinabox:
So turns out there a a ton of problems with doing reinterprete this way (at least for now) like leaking memory.

So I am just going to leave this to sit here and people can take my test cases for when it is done the right way

StefanKarpinski:
In general, we want to move away from reinterpret—it has many issues. What you really want is a conversion that's very cheap (i.e. free), which has the desired semantics.

oxinabox:
> In general, we want to move away from reinterpret—it has many issues. What you really want is a conversion that's very cheap (i.e. free), which has the desired semantics.

Ok, I thought it made sense to have a `reinterpret` that is the _"I know what I am doing, trust me, take those bits in memory and interpret them as this type"_
and then a `convert` that when that is safe does this, and when it isn't doesn't

base/missing.jl
```diff
@@ -69,6 +69,33 @@ convert(::Type{T}, x::T) where {T>:Union{Missing, Nothing}} = x
 convert(::Type{T}, x) where {T>:Missing} = convert(nonmissingtype_checked(T), x)
 convert(::Type{T}, x) where {T>:Union{Missing, Nothing}} = convert(nonmissingtype_checked(nonnothingtype_checked(T)), x)
 
+for Cs in ((:Missing,), (:Nothing,), (:Missing, :Nothing))
+    @eval function Base.reinterpret(
+        ::Type{<:T},  # Note this is effectively same as Type{T} since will error if T is abstract
+        xs::Array{Union{T, $(Cs...)},N}
+    ) where {T,N}
+        isbitstype(T) || throw(ArgumentError("cannot reinterpret `$(eltype(xs))`, type `$(T)` is not a bits type"))
+        return unsafe_wrap(Array{T,N}, Ptr{T}(pointer(xs)), length(xs))
```
KristofferC, line 78: What happens when `xs` get's garbage collected?

KristofferC:
I think the only way this could be reasonably done is to "move" (think `std::move`) ownership of the original array to the new one (in a similar way how `String` does it when acting on a `Vector{UInt8}`).

base/missing.jl
oxinabox: idk, I didn't even know we did aliasing detection.

oxinabox:
> I think the only way this could be reasonably done is to "move" (think std::move) ownership of the original array to the new one (in a similar way how String does it when acting on a Vector{UInt8}).

Then it can't be called `reinterpret`, but that is Ok

KristofferC:
It would be something like `unsafe_move!`

DilumAluthge:

PR #36187, opened 2020-06-08
Allow non-Function callables to be used in count(f, itr)

tkf:
Currently, all predicate functions are wrapped with

https://github.com/JuliaLang/julia/blob/98e678fd9806856a4c0a12f711e1ee397994deeb/base/reduce.jl#L843

The signature `::Function` is overly restrictive because everything is (potentially) callable.

With this patch, following works

```julia
julia> struct NonFunctionIsZero end

julia> (::NonFunctionIsZero)(x) = iszero(x)

julia> count(NonFunctionIsZero(), [])
0
```

Since this worked in Julia 1.0 and 1.4, I'm marking this as a bugfix.

### DIFF
--- a/base/reduce.jl
+++ b/base/reduce.jl
@@ -840,7 +840,7 @@ end
 
 ## count
 
-_bool(f::Function) = x->f(x)::Bool
+_bool(f) = x->f(x)::Bool
 
 """
     count(p, itr) -> Integer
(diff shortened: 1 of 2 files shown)


Keno (approved):

tkf:
@KristofferC or someone else: I added the backport label since it's a bugfix.  Please feel free to remove it if it's inappropriate.

PR #36188, opened 2020-06-08
Support `init` keyword in `sum`/`prod`/`maximum`/`minimum`

tkf:
This PR extends/finalizes @timholy's #35839. It adds `init` kwarg also to `sum` and `prod`.

close #35839

### DIFF
--- a/base/reduce.jl
+++ b/base/reduce.jl
@@ -470,9 +470,9 @@ The return type is `Int` for signed integers of less than system word size, and
 `UInt` for unsigned integers of less than system word size.  For all other
 arguments, a common return type is found to which all arguments are promoted.
 
-The value returned for empty `itr` can be specified by `init` which must be the
-additive identity.  It is unspecified whether `init` is used for non-empty
-collections.
+The value returned for empty `itr` can be specified by `init`. It must be
+the additive identity (i.e. zero) as it is unspecified whether `init` is used
+for non-empty collections.
 
 !!! compat "Julia 1.6"
     Keyword argument `init` requires Julia 1.6 or later.
@@ -642,8 +642,9 @@ end
     maximum(f, itr; [init])
 
 Returns the largest result of calling function `f` on each element of `itr`.
-If provided, `init` must be a neutral element for `max` that will be returned
-for empty collections.
+
+If provided, `init` must be a neutral element for `max` (i.e. which is less
+than any other element) that will be returned for empty collections.
 
 !!! compat "Julia 1.6"
     Keyword argument `init` requires Julia 1.6 or later.
--- a/base/reduce.jl
+++ b/base/reduce.jl
@@ -646,6 +646,9 @@ Returns the largest result of calling function `f` on each element of `itr`.
 If provided, `init` must be a neutral element for `max` that will be returned
 for empty collections.
 
+!!! compat "Julia 1.6"
+    Keyword argument `init` requires Julia 1.6 or later.
+
 # Examples
 ```jldoctest
 julia> maximum(length, ["Julion", "Julia", "Jule"])
@@ -663,6 +666,9 @@ maximum(f, a; kw...) = mapreduce(f, max, a; kw...)
 Returns the smallest result of calling function `f` on each element of `itr`.
 If provided, `init` must be a neutral element for `min` that will be returned
 for empty collections.
+
+!!! compat "Julia 1.6"
+    Keyword argument `init` requires Julia 1.6 or later.
 
 # Examples
 ```jldoctest
--- a/base/reduce.jl
+++ b/base/reduce.jl
@@ -649,8 +649,10 @@ end
 
 Returns the largest result of calling function `f` on each element of `itr`.
 
-If provided, `init` must be a neutral element for `max` (i.e. which is less
-than or equal to any other element) that will be returned for empty collections.
+The value returned for empty `itr` can be specified by `init`. It must be the
+a neutral element for `max` (i.e. which is less than or equal to any
+other element) as it is unspecified whether `init` is used
+for non-empty collections.
 
 !!! compat "Julia 1.6"
     Keyword argument `init` requires Julia 1.6 or later.
@@ -660,8 +662,8 @@ than or equal to any other element) that will be returned for empty collections.
 julia> maximum(length, ["Julion", "Julia", "Jule"])
 6
 
-julia> maximum(length, []; init=-1)
--1
+julia> maximum(length, []; init=-Inf)
+-Inf
 ```
 """
 maximum(f, a; kw...) = mapreduce(f, max, a; kw...)
@@ -671,8 +673,10 @@ maximum(f, a; kw...) = mapreduce(f, max, a; kw...)
 
 Returns the smallest result of calling function `f` on each element of `itr`.
 
-If provided, `init` must be a neutral element for `min` (i.e. which is greater
-than or equal to any other element) that will be returned for empty collections.
+The value returned for empty `itr` can be specified by `init`. It must be the
+a neutral element for `min` (i.e. which is greater than or equal to any
+other element) as it is unspecified whether `init` is used
+for non-empty collections.
 
 !!! compat "Julia 1.6"
     Keyword argument `init` requires Julia 1.6 or later.
@@ -693,8 +697,10 @@ minimum(f, a; kw...) = mapreduce(f, min, a; kw...)
 
 Returns the largest element in a collection.
 
-If provided, `init` must be a neutral element for `max` (i.e. which is less
-than or equal to any other element) that will be returned for empty collections.
+The value returned for empty `itr` can be specified by `init`. It must be the
+a neutral element for `max` (i.e. which is less than or equal to any
+other element) as it is unspecified whether `init` is used
+for non-empty collections.
 
 !!! compat "Julia 1.6"
     Keyword argument `init` requires Julia 1.6 or later.
@@ -712,8 +718,8 @@ ERROR: ArgumentError: reducing over an empty collection is not allowed
 Stacktrace:
 [...]
 
-julia> maximum((); init=-1)
--1
+julia> maximum((); init=-Inf)
+-Inf
 ```
 """
 maximum(a; kw...) = mapreduce(identity, max, a; kw...)
@@ -723,8 +729,10 @@ maximum(a; kw...) = mapreduce(identity, max, a; kw...)
 
 Returns the smallest element in a collection.
 
-If provided, `init` must be a neutral element for `min` (i.e. which is greater
-than or equal to any other element) that will be returned for empty collections.
+The value returned for empty `itr` can be specified by `init`. It must be the
+a neutral element for `min` (i.e. which is greater than or equal to any
+other element) as it is unspecified whether `init` is used
+for non-empty collections.
 
 !!! compat "Julia 1.6"
     Keyword argument `init` requires Julia 1.6 or later.
--- a/base/reduce.jl
+++ b/base/reduce.jl
@@ -45,7 +45,7 @@ function mapfoldl_impl(f::F, op::OP, nt, itr) where {F,OP}
 end
 
 function foldl_impl(op::OP, nt, itr) where {OP}
-    v = _foldl_impl(op, get(nt, :init, _InitialValue()), itr)
+    v = _foldl_impl(op, nt, itr)
     v isa _InitialValue && return reduce_empty_iter(op, itr)
     return v
 end
@@ -157,7 +157,7 @@ Like [`mapreduce`](@ref), but with guaranteed left associativity, as in [`foldl`
 If provided, the keyword argument `init` will be used exactly once. In general, it will be
 necessary to provide `init` to work with empty collections.
 """
-mapfoldl(f, op, itr; kw...) = mapfoldl_impl(f, op, kw.data, itr)
+mapfoldl(f, op, itr; init=_InitialValue()) = mapfoldl_impl(f, op, init, itr)
 
 """
     foldl(op, itr; [init])
@@ -200,7 +200,7 @@ Like [`mapreduce`](@ref), but with guaranteed right associativity, as in [`foldr
 provided, the keyword argument `init` will be used exactly once. In general, it will be
 necessary to provide `init` to work with empty collections.
 """
-mapfoldr(f, op, itr; kw...) = mapfoldr_impl(f, op, kw.data, itr)
+mapfoldr(f, op, itr; init=_InitialValue()) = mapfoldr_impl(f, op, init, itr)
 
 
 """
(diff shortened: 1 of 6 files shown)


test/reduce.jl
```diff
@@ -172,6 +175,19 @@ for f in (sum3, sum4, sum7, sum8)
 end
 @test typeof(sum(Int8[])) == typeof(sum(Int8[1])) == typeof(sum(Int8[1 7]))
 
+@testset "`sum` of empty collections with `init`" begin
+    @testset for init in [0, 0.0]
+        @test sum([]; init = init) === init
+        @test sum((x for x in [123] if false); init = init) === init
+        @test sum(nothing, []; init = init) === init
```
nalimilan, line 182: Is passing `nothing` as the first argument documented?

base/reduce.jl
```diff
@@ -491,60 +498,82 @@ In the former case, the integers are widened to system word size and therefore
...
+additive identity.  It is unspecified whether `init` is used for non-empty
+collections.
+
+!!! compat "Julia 1.6"
+    Keyword argument `init` requires Julia 1.6 or later.
+
 # Examples
 ```jldoctest
 julia> sum(1:20)
 210
```
timholy, line 522: Would be good to add an example that illustrates using `init`.

base/reduce.jl
```diff
@@ -491,60 +498,82 @@ In the former case, the integers are widened to system word size and therefore
...
+multiplicative identity.  It is unspecified whether `init` is used for non-empty
+collections.
+
+!!! compat "Julia 1.6"
+    Keyword argument `init` requires Julia 1.6 or later.
+
 # Examples
 ```jldoctest
 julia> prod(abs2, [2; 3; 4])
 576
```
timholy, line 549: Demo here too

base/reflection.jl
```diff
@@ -1218,10 +1218,12 @@ See also [`applicable`](@ref).
 julia> hasmethod(length, Tuple{Array})
 true
 
-julia> hasmethod(sum, Tuple{Function, Array}, (:dims,))
+julia> f(; orange) = orange;
+
+julia> hasmethod(f, Tuple{}, (:orange,))
```
timholy, line 1223: Very minor point, but what value would you expect someone to supply for `orange`? How about changing this to a count of the number of oranges,
```suggestion
julia> f(; oranges=0) = oranges;

julia> hasmethod(f, Tuple{}, (:oranges,))
```
and then the `apples` and `bananas` below might be more consistent too.

timholy (commented):
Very nice, thanks so much for picking this up and getting it across the finish line!

base/reduce.jl
```diff
@@ -610,35 +639,53 @@ function mapreduce_impl(f, op::Union{typeof(max), typeof(min)},
 end
 
 """
-    maximum(f, itr)
+    maximum(f, itr; [init])
 
 Returns the largest result of calling function `f` on each element of `itr`.
+If provided, `init` must be a neutral element for `max` that will be returned
+for empty collections.
```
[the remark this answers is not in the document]
nalimilan: Ah, right, I missed that it's only used for empty collections. In that case, how about removing "neutral"? Any element will do.

tkf:
> BTW, why don't we guarantee that `init` is only used for empty collections?

> how about removing "neutral"?

I believe we should keep these caveats.  It is highly beneficial that `reduce` implementation can use `init` that is guaranteed to be the neutral element if we want to parallelize `reduce`.  Or for having some parallelized API that is `Base`-compatible.

Consider `reduce(op, xs; init = x0)` with `collect(xs) = [a, b, c, d]`.  If `xs` is a vector, we'd split `xs` in multiple threads equally:

```julia
x = reduce(op, [a, b]; init = x0)  # in thread 1
y = reduce(op, [c, d]; init = x0)  # in thread 2
ans = op(x, y)
```

However, it's not always possible to guarantee that split collections are non-empty.  For example, `xs` might be:

```julia
xs = (x for x in [bad, bad, bad, bad, a, b, c, d] if x !== bad)
```

then the collection-splitting routine might produce

```julia
left = (x for x in [bad, bad, bad, bad] if x !== bad)
right = (x for x in [a, b, c, d] if x !== bad)
```

(That's how [SplittablesBase.jl](https://github.com/JuliaFolds/SplittablesBase.jl) supports `Iterators.filter`)

Then, we'd have

```julia
x = reduce(op, []; init = x0)            # in thread 1
y = reduce(op, [a, b, c, d]; init = x0)  # in thread 2

@assert x == x0

ans = op(x, y)
```

As you can see, `x0` is used even though `xs` itself is not empty and it is important that this is the neutral element so that we have `ans == y`.

Of course, the parallel implementation of `reduce` can completely ignore the `init` keyword and internally use appropriate neutral element.  However, having a concrete `init` helps a lot for JIT'ing single-threaded basecase code.  I think it'd be better to keep the current semantics documented in `reduce`.

nalimilan:
OK, I was just curious whether we currently make use of that possibility. But for "neutral", IMHO the docstring should either mention that the behavior is unspecified (like for other docstrings), or the word should be removed.

base/reduce.jl
```diff
@@ -470,9 +470,9 @@ The return type is `Int` for signed integers of less than system word size, and
 `UInt` for unsigned integers of less than system word size.  For all other
 arguments, a common return type is found to which all arguments are promoted.
 
-The value returned for empty `itr` can be specified by `init` which must be the
-additive identity.  It is unspecified whether `init` is used for non-empty
-collections.
+The value returned for empty `itr` can be specified by `init`. It must be
+the additive identity (i.e. zero) as it is unspecified whether `init` is used
```
nalimilan, line 474: Note that this change should also be applied to other docstrings for `sum` and `prod`.

test/reduce.jl
tkf: I added a comment about it:

https://github.com/JuliaLang/julia/blob/6918213ef2081430e7b6f586b1013aee6641523e/test/reduce.jl#L189-L191

I can instead do

```julia
@testset "`sum` of empty collections with `init`" begin
    function noncallable end
    @testset for init in [0, 0.0]
        # ...
        @test sum(noncallable, []; init = init) === init
        # ...
    end
end

```

if that's better.

tkf:
> I was just curious whether we currently make use of that possibility

It's possible but I'm not sure. I just think it's a reasonable API to keep it as-is.

base/reduce.jl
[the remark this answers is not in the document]
tkf: @timholy I think `init=-1` with `maximum(length, ...)` is perfect but it's dangerous with `minimum(length, ...)`. I think it's better to _require_ `init` to be the neutral element to make it easy to parallelize. See: https://github.com/JuliaLang/julia/pull/36188#issuecomment-640841688

test/reduce.jl
tkf: Actually, now I think `function noncallable end` is better. (But we can rollback if you want.)

base/reduce.jl
tkf: I just pushed the commits to apply the same change to other docstrings.

base/reduce.jl
```diff
@@ -610,62 +645,108 @@ function mapreduce_impl(f, op::Union{typeof(max), typeof(min)},
...
+
+!!! compat "Julia 1.6"
+    Keyword argument `init` requires Julia 1.6 or later.
+
 # Examples
 ```jldoctest
 julia> minimum(length, ["Julion", "Julia", "Jule"])
 4
+
+julia> minimum(length, []; init=-1)
```
timholy, line 685: Given your comment about parallelization, do you want to change this to `init=typemax(Int)` or something?

base/reduce.jl
```diff
@@ -491,60 +498,82 @@ In the former case, the integers are widened to system word size and therefore
 the result is 128. In the latter case, no such widening happens and integer
 overflow results in -128.
 """
-sum(f, a) = mapreduce(f, add_sum, a)
+sum(f, a; kw...) = mapreduce(f, add_sum, a; kw...)
```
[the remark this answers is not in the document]
tkf: My thought was that some non-array collection might want to implement `dims`.

timholy (approved):

base/reduce.jl
timholy: Sorry, there's a second one I missed on first reading...

tkf: Thanks, I missed it. I used `init=Inf` for simplicity. What do you think?

timholy: Awesome. I would have approved this even without, but with I think it's perfect. Thanks again for doing this!!

base/reduce.jl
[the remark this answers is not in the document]
tkf: Another issue is that passing `init` breaks something like `sum(inv, 1)` because `mapreduce(f, op, a::Number)` does not take any keyword arguments

https://github.com/JuliaLang/julia/blob/7301dc61bdeb5d66e94e15bdfcd4c54f7c90f068/base/reduce.jl#L419-L420

It also breaks collection that dispatches to external `mapreduce` implementation that does not use `Base._InitialValue`.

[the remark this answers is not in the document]
nalimilan: OK, that's fine then.

base/reduce.jl
```diff
@@ -610,62 +645,108 @@ function mapreduce_impl(f, op::Union{typeof(max), typeof(min)},
 end
 
 """
-    maximum(f, itr)
+    maximum(f, itr; [init])
 
 Returns the largest result of calling function `f` on each element of `itr`.
 
+If provided, `init` must be a neutral element for `max` (i.e. which is less
+than or equal to any other element) that will be returned for empty collections.
```
nalimilan, line 653: How about using the same formulation as for other functions?
```suggestion
The value returned for empty `itr` can be specified by `init`. It must be the
a neutral element for `max` (i.e. which is less than or equal to any
other element) as it is unspecified whether `init` is used
for non-empty collections.
```

base/reduce.jl
```diff
@@ -610,62 +645,108 @@ function mapreduce_impl(f, op::Union{typeof(max), typeof(min)},
...
+maximum(f, a; kw...) = mapreduce(f, max, a; kw...)
 
 """
-    minimum(f, itr)
+    minimum(f, itr; [init])
 
 Returns the smallest result of calling function `f` on each element of `itr`.
 
+If provided, `init` must be a neutral element for `min` (i.e. which is greater
+than or equal to any other element) that will be returned for empty collections.
```
nalimilan, line 675: How about using the same formulation as for other functions?
```suggestion
The value returned for empty `itr` can be specified by `init`. It must be the
a neutral element for `min` (i.e. which is greater than or equal to any
other element) as it is unspecified whether `init` is used
for non-empty collections.
```

base/reduce.jl
```diff
@@ -610,62 +645,108 @@ function mapreduce_impl(f, op::Union{typeof(max), typeof(min)},
...
+!!! compat "Julia 1.6"
+    Keyword argument `init` requires Julia 1.6 or later.
+
 # Examples
 ```jldoctest
 julia> maximum(length, ["Julion", "Julia", "Jule"])
 6
+
+julia> maximum(length, []; init=-1)
+-1
```
nalimilan, line 664:
```suggestion
julia> maximum(length, []; init=-Inf)
-Inf
```

base/reduce.jl
```diff
@@ -610,62 +645,108 @@ function mapreduce_impl(f, op::Union{typeof(max), typeof(min)},
...
+maximum(a; kw...) = mapreduce(identity, max, a; kw...)
 
 """
-    minimum(itr)
+    minimum(itr; [init])
 
 Returns the smallest element in a collection.
 
+If provided, `init` must be a neutral element for `min` (i.e. which is greater
+than or equal to any other element) that will be returned for empty collections.
```
nalimilan, line 727: How about using the same formulation as for other functions?
```suggestion
The value returned for empty `itr` can be specified by `init`. It must be the
a neutral element for `min` (i.e. which is greater than or equal to any
other element) as it is unspecified whether `init` is used
for non-empty collections.
```

base/reduce.jl
```diff
@@ -610,62 +645,108 @@ function mapreduce_impl(f, op::Union{typeof(max), typeof(min)},
...
+minimum(f, a; kw...) = mapreduce(f, min, a; kw...)
 
 """
-    maximum(itr)
+    maximum(itr; [init])
 
 Returns the largest element in a collection.
 
+If provided, `init` must be a neutral element for `max` (i.e. which is less
+than or equal to any other element) that will be returned for empty collections.
```
nalimilan, line 697: How about using the same formulation as for other functions?
```suggestion
The value returned for empty `itr` can be specified by `init`. It must be the
a neutral element for `max` (i.e. which is less than or equal to any
other element) as it is unspecified whether `init` is used
for non-empty collections.
```

base/reduce.jl
```diff
@@ -610,62 +645,108 @@ function mapreduce_impl(f, op::Union{typeof(max), typeof(min)},
...
 julia> maximum([1,2,3])
 3
+
+julia> maximum(())
+ERROR: ArgumentError: reducing over an empty collection is not allowed
+Stacktrace:
+[...]
+
+julia> maximum((); init=-1)
+-1
```
nalimilan, line 716:
```suggestion
julia> maximum((); init=-Inf)
-Inf
```

base/reduce.jl
tkf: I actually think `init=-1` here is a good illustrative example.  Since you know that `length(thing)` cannot be less than `0`, `init=-1` is a good neutral element.  So, I suggest to keep it this way.

I also changed the `minimum` demo to `minimum(length, []; init=typemax(Int64))` based on @timholy's suggestion.  Maybe this version clarifies the usage more.  I also added `minimum(tanh, Real[]; init=1.0)` and `maximum(tanh, Real[]; init=-1.0)` just in case it's useful.

base/reduce.jl
```diff
@@ -610,62 +645,122 @@ function mapreduce_impl(f, op::Union{typeof(max), typeof(min)},
 end
 
 """
-    maximum(f, itr)
+    maximum(f, itr; [init])
 
 Returns the largest result of calling function `f` on each element of `itr`.
 
+The value returned for empty `itr` can be specified by `init`. It must be the
```
nalimilan, line 652:
```suggestion
The value returned for empty `itr` can be specified by `init`. It must be
```

base/reduce.jl
```diff
@@ -610,62 +645,122 @@ function mapreduce_impl(f, op::Union{typeof(max), typeof(min)},
...
-maximum(f, a) = mapreduce(f, max, a)
+maximum(f, a; kw...) = mapreduce(f, max, a; kw...)
 
 """
-    minimum(f, itr)
+    minimum(f, itr; [init])
 
 Returns the smallest result of calling function `f` on each element of `itr`.
 
+The value returned for empty `itr` can be specified by `init`. It must be the
```
nalimilan, line 679:
```suggestion
The value returned for empty `itr` can be specified by `init`. It must be
```

base/reduce.jl
```diff
@@ -610,62 +645,122 @@ function mapreduce_impl(f, op::Union{typeof(max), typeof(min)},
...
-minimum(f, a) = mapreduce(f, min, a)
+minimum(f, a; kw...) = mapreduce(f, min, a; kw...)
 
 """
-    maximum(itr)
+    maximum(itr; [init])
 
 Returns the largest element in a collection.
 
+The value returned for empty `itr` can be specified by `init`. It must be the
```
nalimilan, line 706:
```suggestion
The value returned for empty `itr` can be specified by `init`. It must be
```

base/reduce.jl
```diff
@@ -610,62 +645,122 @@ function mapreduce_impl(f, op::Union{typeof(max), typeof(min)},
...
-maximum(a) = mapreduce(identity, max, a)
+maximum(a; kw...) = mapreduce(identity, max, a; kw...)
 
 """
-    minimum(itr)
+    minimum(itr; [init])
 
 Returns the smallest element in a collection.
 
+The value returned for empty `itr` can be specified by `init`. It must be the
```
nalimilan, line 738:
```suggestion
The value returned for empty `itr` can be specified by `init`. It must be
```

base/reduce.jl
nalimilan: Fine with me. (Though I must admit I don't remember offhand how `tanh` behaves...)

nalimilan (approved):
Thanks!

tkf:
@nalimilan @timholy Thanks a lot for your detailed review!

cossio:
Will this make it into 1.5?

PR #36189, opened 2020-06-08
Implement stdout option for --trace-compile

mkitti:
The command line option `--trace-compile` supposedly supports `stdout` as a valid option according to `--help-hidden`. Currently, using `--trace-compile=stdout` would result in precompile statements being written to an ordinary file called "stdout".

This commit implements `--trace-compile=stdout` such that the precompile statements are written to standard output stream rather than a file.
```
$ ./julia --help-hidden
julia [switches] -- [programfile] [args...]
...
 --trace-compile={stdout,stderr}
                           Print precompile statements for methods compiled during execution.
```

### DIFF
--- a/src/jloptions.c
+++ b/src/jloptions.c
@@ -161,8 +161,8 @@ static const char opts_hidden[]  =
     " --output-bc name          Generate LLVM bitcode (.bc)\n"
     " --output-asm name         Generate an assembly file (.s)\n"
     " --output-incremental=no   Generate an incremental output file (rather than complete)\n"
-    " --trace-compile={stdout,stderr,name}\n"
-    "                           Print precompile statements for methods compiled during execution or save to a file (.jl)\n\n"
+    " --trace-compile={stderr,name}\n"
+    "                           Print precompile statements for methods compiled during execution or save to a path\n\n"
 ;
 
 JL_DLLEXPORT void jl_parse_opts(int *argcp, char ***argvp)
--- a/src/jloptions.c
+++ b/src/jloptions.c
@@ -161,8 +161,8 @@ static const char opts_hidden[]  =
     " --output-bc name          Generate LLVM bitcode (.bc)\n"
     " --output-asm name         Generate an assembly file (.s)\n"
     " --output-incremental=no   Generate an incremental output file (rather than complete)\n"
-    " --trace-compile={stdout,stderr}\n"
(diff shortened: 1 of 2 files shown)


src/jloptions.c
```diff
@@ -161,8 +161,8 @@ static const char opts_hidden[]  =
     " --output-bc name          Generate LLVM bitcode (.bc)\n"
     " --output-asm name         Generate an assembly file (.s)\n"
     " --output-incremental=no   Generate an incremental output file (rather than complete)\n"
-    " --trace-compile={stdout,stderr}\n"
-    "                           Print precompile statements for methods compiled during execution.\n\n"
+    " --trace-compile={stdout,stderr,name}\n"
```
StefanKarpinski, line 164:
```suggestion
    " --trace-compile={stdout,stderr,path}\n"
```

src/jloptions.c
```diff
@@ -161,8 +161,8 @@ static const char opts_hidden[]  =
...
+    "                           Print precompile statements for methods compiled during execution or save to a file (.jl)\n\n"
```
StefanKarpinski, line 165:
```suggestion
    "                           Print precompile statements for methods compiled during execution or save to a path\n\n"
```
Otherwise it makes it seem like this flag will write to the file `$name.jl` which is not what it should do.

KristofferC:
Imo the erroneous `stdout` documentation should just be removed.

src/jloptions.c
mkitti: I changed the word to just "path" as suggested.

mkitti:
I removed the `stdout` code and the reference to `stdout` in the --help-hidden message.

src/jloptions.c
```diff
@@ -161,8 +161,8 @@ static const char opts_hidden[]  =
     " --output-bc name          Generate LLVM bitcode (.bc)\n"
     " --output-asm name         Generate an assembly file (.s)\n"
     " --output-incremental=no   Generate an incremental output file (rather than complete)\n"
-    " --trace-compile={stdout,stderr}\n"
-    "                           Print precompile statements for methods compiled during execution.\n\n"
+    " --trace-compile={stderr,name}\n"
+    "                           Print precompile statements for methods compiled during execution or save to a path\n\n"
```
mkitti, line 165: This is now just a two line documentation change for the help text.

PR #36192, opened 2020-06-08
update `lock` example

scheidan:
The old example gave the wrong impression that `lock` takes a data structure and not a lock as argument.

I'm not an expert on this topic at all!

### DIFF
--- a/doc/src/manual/multi-threading.md
+++ b/doc/src/manual/multi-threading.md
@@ -73,20 +73,20 @@ can be observed from multiple threads. For example, in most cases you should
 use the following code pattern:
 
 ```julia-repl
-julia> lock(l) do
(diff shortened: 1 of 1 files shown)


doc/src/manual/multi-threading.md
```diff
@@ -73,19 +73,20 @@ can be observed from multiple threads. For example, in most cases you should
 use the following code pattern:
 
 ```julia-repl
-julia> lock(a) do
+julia> lock(l) do
```
ViralBShah, line 76: Maybe use `lk` instead of `l`? It looked like 1 to me and I was quite puzzled.

scheidan: good point! I've updated that

PR #36200, opened 2020-06-08
in inference, disallow 3-element Unions if one elt is abstract

JeffBezanson:
Simpler alternative to #35904; just the type-widening part. With this we can still split `!` as follows:
```
julia> g(x) = !x

julia> code_typed(g,(Any,))
1-element Array{Any,1}:
 CodeInfo(
1 ─ %1  = (isa)(x, Missing)::Bool
└──       goto #3 if not %1
2 ─       goto #6
3 ─ %4  = (isa)(x, Bool)::Bool
└──       goto #5 if not %4
4 ─ %6  = π (x, Bool)
│   %7  = Base.not_int(%6)::Bool
└──       goto #6
5 ─ %9  = !x::Any
└──       goto #6
6 ┄ %11 = φ (#2 => $(QuoteNode(missing)), #4 => %7, #5 => %9)::Any
└──       return %11
) => Any
```
Those branches are potentially helpful, and the `!x::Any` avoids adding the method table backedge, so we still get most of the latency improvement.

### DIFF
--- a/base/compiler/typelimits.jl
+++ b/base/compiler/typelimits.jl
@@ -268,8 +268,11 @@ function type_more_complex(@nospecialize(t), @nospecialize(c), sources::SimpleVe
     return true
 end
 
+union_count_abstract(x::Union) = union_count_abstract(x.a) + union_count_abstract(x.b)
+union_count_abstract(@nospecialize(x)) = !isdispatchelem(x)
+
 function issimpleenoughtype(@nospecialize t)
-    return unionlen(t) <= MAX_TYPEUNION_LENGTH && unioncomplexity(t) <= MAX_TYPEUNION_COMPLEXITY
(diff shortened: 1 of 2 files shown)


Keno (approved):

fingolfin:
This fixes a type interference error we are experiencing in Julia 1.4 and 1.5 (see https://github.com/oscar-system/Oscar.jl/issues/104). Any chance this could be backported to Julia 1.5?

JeffBezanson:
This isn't a bug fix --- it's probably just hiding the bug. But I'll see if the real cause can be fixed.

fingolfin:
OK, that makes sense (sorry, I didn't really look at the content of this PR, else I could have guessed).

Should I then open an issue for the inference problem we are seeing, and how to reproduce it?

JeffBezanson:
Yes, please.

PR #36202, opened 2020-06-08
Remove unnecessary use of ans keyword from manual

clarkfitzg:
The `ans` keyword adds to the cognitive load for new users reading the
manual, potentially distracting their focus from the examples. This
commit replaces non essential uses of `ans` in the manual with clear
variable assignments.

Thanks @fredrikekre for the [suggestion and encouragement on Slack](https://julialang.slack.com/archives/C69DZ83C6/p1591540869043200).
For context, I'm a new user coming from R and Python.
While reading the string docs, I went down a short rabbit hole with `ans`:

> Where did this `ans` come from?
> There must have been some hidden assignment that happened somewhere, maybe in a hidden block in the docs that they       forgot to show.
> (Looks up docs), ah, `ans` is a keyword.
> What are the semantics?
> Will it write over my variable `ans`?
> Yes, that's too bad, some people like to use ans as a variable name
> ... (and so on) 

I did read the [Getting Started](https://docs.julialang.org/en/v1/manual/getting-started/), but a couple days later when I returned to the docs I forgot about `ans`. `ans` strikes me as a convenience, not a core feature of the language. Why make something potentially confusing the first thing that a new user sees in the manual?

For the integers and floating point numbers, continually reassigning the variable `x` with objects of different types reinforces that Julia is dynamically typed, which is a core feature of the language.

It seems that the doctests are failing locally on my Mac with the `typemax(Int)` example in the FAQ. I'm not sure it's related to my changes?

### DIFF
--- a/doc/src/manual/faq.md
+++ b/doc/src/manual/faq.md
@@ -410,16 +410,16 @@ is bounded and wraps around at either end so that adding, subtracting and multip
 can overflow or underflow, leading to some results that can be unsettling at first:
 
 ```jldoctest
-julia> typemax(Int)
+julia> m = typemax(Int)
 9223372036854775807
 
-julia> ans+1
+julia> m+1
 -9223372036854775808
 
-julia> -ans
+julia> -m
 -9223372036854775808
 
-julia> 2*ans
+julia> 2*m
 0
 ```
 
--- a/doc/src/manual/faq.md
+++ b/doc/src/manual/faq.md
@@ -410,16 +410,16 @@ is bounded and wraps around at either end so that adding, subtracting and multip
 can overflow or underflow, leading to some results that can be unsettling at first:
 
 ```jldoctest
-julia> m = typemax(Int)
+julia> x = typemax(Int)
 9223372036854775807
 
-julia> m+1
+julia> y = x+1
 -9223372036854775808
 
-julia> -m
+julia> z = -y
 -9223372036854775808
 
-julia> 2*m
+julia> 2*z
 0
 ```
 
--- a/doc/src/manual/getting-started.md
+++ b/doc/src/manual/getting-started.md
@@ -21,7 +21,9 @@ To exit the interactive session, type `CTRL-D` (press the Control/`^` key togeth
 `exit()`. When run in interactive mode, `julia` displays a banner and prompts the user for input.
 Once the user has entered a complete expression, such as `1 + 2`, and hits enter, the interactive
 session evaluates the expression and shows its value. If an expression is entered into an interactive
-session with a trailing semicolon, its value is not shown.
+session with a trailing semicolon, its value is not shown. The variable `ans` is bound to the
+value of the last evaluated expression whether it is shown or not. The `ans` variable is only
+bound in interactive sessions, not when Julia code is run in other ways.
 
 To evaluate expressions written in a source file `file.jl`, write `include("file.jl")`.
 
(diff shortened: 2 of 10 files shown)


doc/src/manual/faq.md
```diff
@@ -410,16 +410,16 @@ is bounded and wraps around at either end so that adding, subtracting and multip
...
-julia> typemax(Int)
+julia> m = typemax(Int)
 9223372036854775807
 
-julia> ans+1
+julia> m+1
 -9223372036854775808
 
-julia> -ans
+julia> -m
```
KristofferC, line 419: This looks wrong.

clarkfitzg: ah yes, not the same `ans`.

doc/src/manual/getting-started.md
```diff
@@ -14,16 +14,14 @@ io = IOBuffer()
...
 import Markdown
-Markdown.parse("```\n\$ julia\n\n$(banner)\njulia> 1 + 2\n3\n\njulia> ans\n3\n```")
+Markdown.parse("```\n\$ julia\n\n$(banner)\njulia> 1 + 2\n3\n```")
 ```
 
 To exit the interactive session, type `CTRL-D` (press the Control/`^` key together with the `d` key), or type
 `exit()`. When run in interactive mode, `julia` displays a banner and prompts the user for input.
 Once the user has entered a complete expression, such as `1 + 2`, and hits enter, the interactive
 session evaluates the expression and shows its value. If an expression is entered into an interactive
-session with a trailing semicolon, its value is not shown. The variable `ans` is bound to the
```
KristofferC, line 24: I think it is ok to remove the use of `ans` in the docstrings but this seems like a quite innocent piece of documentation that can actually be pretty useful to know while getting started.  I would argue for putting this part back.

clarkfitzg: I agree that it's a good and useful piece of documentation. It's the prominent position in the "Getting Started" section that seems a little out of place, and potentially distracting.

How about moving this explanation near the top of the "Variables" section, which is the next section after "Getting Started"? This seems like a natural place. A new user working their way through the docs will still see it just as early, so they can use `ans` as they play around on the REPL.

doc/src/manual/getting-started.md
```diff
@@ -14,7 +14,7 @@ io = IOBuffer()
 Base.banner(io)
 banner = String(take!(io))
 import Markdown
-Markdown.parse("```\n\$ julia\n\n$(banner)\njulia> 1 + 2\n3\n\njulia> ans\n3\n```")
```
Keno, line 17: I think it's useful to keep this in the getting started docs, since people may not be familiar with this.

Keno: (If you want to make it more explicit, you can do something like)
```
julia> ans # `ans` retrieves the previous REPL answer
3
```

clarkfitzg:
Here are the uses of `ans` in the docs after this commit.

```
~/dev/julia/doc/src $ grep -r "[^a-z]ans[^a-z]" *
manual/getting-started.md:session with a trailing semicolon, its value is not shown. The variable `ans` is bound to the
manual/getting-started.md:value of the last evaluated expression whether it is shown or not. The `ans` variable is only
manual/noteworthy-differences.md:  * In both Julia and MATLAB, the variable `ans` is set to the value of the last expression issued
manual/noteworthy-differences.md:    in an interactive session. In Julia, unlike MATLAB, `ans` is not set when Julia code is run in
stdlib/REPL.md:the REPL also binds the result to the variable `ans`. A trailing semicolon on the line can be
~/dev/julia/doc/src $ 
```

Keno (approved):

clarkfitzg:
Should I rebase and squash these commits?

Keno:
No need, they can be squashed on merge.

PR #30485, opened 2018-12-21
Make return type of broadcast inferrable with heterogeneous arrays

nalimilan:
Inference is not able to detect the element type automatically, but we can do it manually since we know `promote_typejoin` is used for widening.

Fixes #28382. The same changes should be applied to `map`.

PR #36208, opened 2020-06-09
set default max_methods to 3

JeffBezanson:
Until we settle on a generally-better algorithm (which may not exist), this is a good way to get some free performance. I'm curious to find out what PkgEval says.

Here are some numbers. Baseline from master (note my environment also includes https://github.com/JuliaPlots/PlotUtils.jl/pull/96):
```
Base  ─────────── 38.462095 seconds
Stdlibs: ────  54.332004 seconds 58.551%
-rw-r--r-- 1 jeff jeff  12831441 Jun  8 19:59 corecompiler.ji
-rwxr-xr-x 1 jeff jeff 151354656 Jun  8 20:07 sys.so
julia> @time using Plots
144.572065 seconds (9.39 M allocations: 534.732 MiB, 0.14% gc time)

 11.177675 seconds (11.11 M allocations: 574.409 MiB, 1.57% gc time)

julia> @time using SIMD
[ Info: Precompiling SIMD [fdea26ae-647d-5447-a871-4b548cad5224]
  2.551328 seconds (2.15 M allocations: 110.859 MiB, 0.87% gc time)

  9.076342 seconds (20.95 M allocations: 1.047 GiB, 6.05% gc time)
```

This PR:
```
Base  ─────────── 36.959470 seconds
Stdlibs: ────  51.727160 seconds 58.3257%
-rw-r--r-- 1 jeff jeff  12750431 Jun  8 21:34 corecompiler.ji
-rwxr-xr-x 1 jeff jeff 145160120 Jun  8 21:41 sys.so
julia> @time using Plots
125.626309 seconds (7.76 M allocations: 445.493 MiB, 0.10% gc time)

 10.892562 seconds (10.28 M allocations: 530.079 MiB, 2.52% gc time)

julia> @time using SIMD
[ Info: Precompiling SIMD [fdea26ae-647d-5447-a871-4b548cad5224]
  1.415139 seconds (285.21 k allocations: 18.519 MiB)

  0.409602 seconds (982.29 k allocations: 49.991 MiB, 2.96% gc time)
```

The last item is clearly volatile; there are probably just a few critical backedges that this plus #36200 manage to cut.

### DIFF
--- a/base/compiler/typelimits.jl
+++ b/base/compiler/typelimits.jl
@@ -391,7 +391,20 @@ function tmerge(@nospecialize(typea), @nospecialize(typeb))
                         widen = tuplemerge(unwrap_unionall(ti)::DataType, unwrap_unionall(tj)::DataType)
                         widen = rewrap_unionall(rewrap_unionall(widen, ti), tj)
                     else
-                        widen = typenames[i].wrapper
+                        wr = typenames[i].wrapper
+                        uw = unwrap_unionall(wr)::DataType
+                        ui = unwrap_unionall(ti)::DataType
+                        uj = unwrap_unionall(tj)::DataType
+                        merged = wr
+                        for k = 1:length(uw.parameters)
+                            ui_k = ui.parameters[k]
+                            if ui_k === uj.parameters[k] && !has_free_typevars(ui_k)
+                                merged = merged{ui_k}
+                            else
+                                merged = merged{uw.parameters[k]}
+                            end
+                        end
+                        widen = rewrap_unionall(merged, wr)
                     end
                     types[i] = Union{}
                     typenames[i] = Any.name
--- a/base/array.jl
+++ b/base/array.jl
@@ -154,7 +154,7 @@ end
 size(a::Array, d::Integer) = arraysize(a, convert(Int, d))
 size(a::Vector) = (arraysize(a,1),)
 size(a::Matrix) = (arraysize(a,1), arraysize(a,2))
-size(a::Array{<:Any,N}) where {N} = (@_inline_meta; ntuple(M -> size(a, M), Val(N)))
+size(a::Array{<:Any,N}) where {N} = (@_inline_meta; ntuple(M -> size(a, M), Val(N))::Dims)
 
 asize_from(a::Array, n) = n > ndims(a) ? () : (arraysize(a,n), asize_from(a, n+1)...)
 
--- a/base/compiler/types.jl
+++ b/base/compiler/types.jl
@@ -58,7 +58,7 @@ struct OptimizationParams
             inline_cost_threshold::Int = 100,
             inline_nonleaf_penalty::Int = 1000,
             inline_tupleret_bonus::Int = 400,
-            max_methods::Int = 4,
+            max_methods::Int = 3,
             tuple_splat::Int = 32,
             union_splitting::Int = 4,
         )
@@ -107,7 +107,7 @@ struct InferenceParams
     function InferenceParams(;
             ipo_constant_propagation::Bool = true,
             aggressive_constant_propagation::Bool = false,
-            max_methods::Int = 4,
+            max_methods::Int = 3,
             union_splitting::Int = 4,
             apply_union_enum::Int = 8,
             tupletype_depth::Int = 3,
--- a/stdlib/Random/src/Random.jl
+++ b/stdlib/Random/src/Random.jl
@@ -253,7 +253,7 @@ rand(rng::AbstractRNG, ::UniformT{T}) where {T} = rand(rng, T)
 rand(rng::AbstractRNG, X)                                           = rand(rng, Sampler(rng, X, Val(1)))
 # this is needed to disambiguate
 rand(rng::AbstractRNG, X::Dims)                                     = rand(rng, Sampler(rng, X, Val(1)))
-rand(rng::AbstractRNG=default_rng(), ::Type{X}=Float64) where {X} = rand(rng, Sampler(rng, X, Val(1)))
+rand(rng::AbstractRNG=default_rng(), ::Type{X}=Float64) where {X} = rand(rng, Sampler(rng, X, Val(1)))::X
 
 rand(X)                   = rand(default_rng(), X)
 rand(::Type{X}) where {X} = rand(default_rng(), X)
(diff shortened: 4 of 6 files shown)


JeffBezanson:

nanosoldier:
[Your package evaluation job](https://github.com/JuliaLang/julia/pull/36208#issuecomment-640983676) has completed - possible new issues were detected. A full report can be found [here](https://github.com/JuliaCI/NanosoldierReports/blob/a217ad650cfee1af3fd68fd8c492777536eae1d8/pkgeval/by_hash/3890d1d_vs_d765e59/report.md). cc @maleadt

JeffBezanson:
Added some mitigations. Here's the analysis:

- POMDPModelTools: makes some calls to `rand(::AbstractRNG, ...)`, can be fixed with one type declaration in Random.
- FieldMetadata: adds one method per annotated field, and the test case has 4 fields. Inference depends on the number of fields so this is not a meaningful regression.
- PencilFFTs: not clear, similar to #35800 (in the caller the type is `Any` but when you drill down the type is `Int`).
- TypedTables: fixed by making `tmerge` sharper: when merging concrete types from the same family but with different parameters, keep parameters they have in common (e.g. merging various types of Vector gives `Vector` instead of `Array`).

JeffBezanson:
@rfourquet Is the type declaration in Random valid?

rfourquet:
> @rfourquet Is the type declaration in Random valid?

Yes, fortunately `rand(X::Type)` is required to return an `X` object by the current documented API.

JeffBezanson:

nanosoldier:
[Your benchmark job](https://github.com/JuliaLang/julia/pull/36208#issuecomment-644901310) has completed - possible performance regressions were detected. A full report can be found [here](https://github.com/JuliaCI/BaseBenchmarkReports/blob/36b296674fc1835dce70fe17bd20c972ff5a874d/444a973_vs_9ed10de/report.md). cc @ararslan

PallHaraldsson:
This improves, I see e.g. 0.43 but also down to 41.24 times slower (are some of these false?);
["union", "array", "(\"map\", *, Float32, (false, false))"]

JeffBezanson:
Unsurprisingly, the big regressions are in union-splitting cases that involve 4 cases (`Union{T,Nothing} x Union{T,Nothing}`). Inference will actually look at more than max_methods methods if max_union_split is greater, but inlining won't. #35891 should fix that, so we should merge that first and retry.

JeffBezanson:

nanosoldier:
[Your benchmark job](https://github.com/JuliaLang/julia/pull/36208#issuecomment-645596943) has completed - possible performance regressions were detected. A full report can be found [here](https://github.com/JuliaCI/BaseBenchmarkReports/blob/0d8398eb91f2fe15b7845efc97b83de87b8fca08/00f1a7a_vs_8ca6e8d/report.md). cc @ararslan

JeffBezanson:
I think this is ready to merge after CI. @Keno @vtjnash ok?

base/compiler/typelimits.jl
```diff
@@ -391,7 +391,19 @@ function tmerge(@nospecialize(typea), @nospecialize(typeb))
                         widen = tuplemerge(unwrap_unionall(ti)::DataType, unwrap_unionall(tj)::DataType)
                         widen = rewrap_unionall(rewrap_unionall(widen, ti), tj)
                     else
-                        widen = typenames[i].wrapper
+                        wr = typenames[i].wrapper
+                        uw = unwrap_unionall(wr)
+                        ui = unwrap_unionall(ti)
+                        uj = unwrap_unionall(tj)
```
vtjnash, line 397:
```suggestion
                        uw = unwrap_unionall(wr)::DataType
                        ui = unwrap_unionall(ti)::DataType
                        uj = unwrap_unionall(tj)::DataType
```

base/compiler/typelimits.jl
```diff
@@ -391,7 +391,19 @@ function tmerge(@nospecialize(typea), @nospecialize(typeb))
...
+                        merged = wr
+                        for k = 1:length(uw.parameters)
+                            if ui.parameters[k] === uj.parameters[k]
+                                merged = merged{ui.parameters[k]}
+                            else
+                                merged = merged{uw.parameters[k]}
+                            end
+                        end
+                        widen = rewrap_unionall(merged, wr)
```
vtjnash, line 406: I wonder if this is entirely sound, due mainly to losing `where` bounds, but also wondering if it'd mismatch TypeVars coming from different places. What if we had something like:
```julia
struct MyType{A, B, C, D} end
T = TypeVar(:T)
S = TypeVar(:S)
A = UnionAll(T, MyType{T, S, T, Pair{T, S}})
B = UnionAll(T, MyType{S, T, T, Pair{T, S}})
```
The merge result would appear to be:
```julia
MyType{A, B, T, Pair{T, S}} where B where A
```

Which, I think the only issue it has is that it's missing `where T`, but just checking

vtjnash (commented):
Yeah, those results seem good. I think it should be split into 3ish commits though.

base/compiler/typelimits.jl
JeffBezanson: You're right. I'll exclude anything with free typevars.

jipolanco:
For what it's worth, the PencilFFTs inference issue is fixed by replacing a broadcast over a tuple using the dot syntax, by an explicit broadcast using `map` or `broadcast` (see the commit message above).

Maybe this can help the mitigation effort. I can also try to provide a simpler example that reproduces the issue.

nalimilan:
Bisection revealed that this PR is reponsible for the fact that the return type of `Base.Broadcast.copyto_nonleaf!([1], Base.Broadcast.Broadcasted(+, ([1,2], [1,1])), [1, 2], 1, 1)` is no longer inferred as `Array`, which was the case in Julia 1.5. Unfortunately this means that even if after https://github.com/JuliaLang/julia/pull/30485#issuecomment-673396551 the return *element* type of `broadcast` is inferred for small `Union` inputs, the *container* type isn't inferred, which defeats the goal of inferring a concrete type.

Do you think something can be done about this? FWIW, I've noticed that in `copyto_nonleaf!` the compiler doesn't seem to realize that `typeof(val)` cannot be of type `T` in the second branch of the `if` (since it's handled in the first branch). This adds a type instability which isn't real and which may be the reason why inference bails out.

PR #36211, opened 2020-06-09
Normalize (simplify) `UnionAll`s when used as type parameter

### LINKED ISSUE #35130: datatype cache normalization bugs
These types should print the same (since the type parameters are type-equal), if the datatype cache (and new hash computation) is working correctly:

```julia
julia> Val{Tuple{Any}}
Val{Tuple{Any}}

julia> Val{Tuple{T} where T}
Val{Tuple{T} where T}
```
```
julia> Val{Tuple{Float64,Int64,Int64}}
Val{Tuple{Float64,Int64,Int64}}

julia> Val{Tuple{Float64,Vararg{T,2}} where T<:Int64}
Val{Tuple{Float64,Vararg{T,2}} where T<:Int64}

julia> Val{Tuple{Float64,T,T} where T<:Int64}
Val{Tuple{Float64,Vararg{T,2}} where T<:Int64}
```

This is actually an improvement from before the new hash computation code (the last one was previously also creating a unique type). Found these by some quick fuzzing, so I'm pleased to see there weren't more!

martinholters:
A stab at #35130 although I'm not totally convinced this is a good idea, see https://github.com/JuliaLang/julia/issues/35130#issuecomment-640391209.

This normalizes `UnionAll`s (and `Union`s thereof) when they are used as type parameters. Normalization here means that the typevar is replaced with its upper bound if it does not appear in invariant position and is not subject to the diagonal rule (or in the trivial case where lower and upper bound are identical). Parameters to `Type` are excluded from this normalization to prevent the change in dispatch also mentioned in https://github.com/JuliaLang/julia/issues/35130#issuecomment-640391209. This is sufficient to cover the cases from #35130:

```julia
julia> Val{Tuple{T} where T}
Val{Tuple{Any}}

julia> Val{Tuple{Float64,Vararg{T,2}} where T<:Int64}
Val{Tuple{Float64,Int64,Int64}}

julia> Val{Tuple{Float64,T,T} where T<:Int64}
Val{Tuple{Float64,Int64,Int64}}

julia> Val{Union{Nothing, T} where T}
Val{Any}
```

This mostly also makes the tests from #34272 pass (except for where the `Type` exclusion is relevant), but I'm not convinced there aren't other cases where type-equal concrete types are not normalized to identical types, so #34272 might still be necessary in all its glory.

Closes #35130.

### DIFF
--- a/src/jltypes.c
+++ b/src/jltypes.c
@@ -1153,6 +1153,92 @@ static jl_value_t *extract_wrapper(jl_value_t *t JL_PROPAGATES_ROOT) JL_GLOBALLY
     return NULL;
 }
 
+int _may_substitute_ub(jl_value_t *v, jl_tvar_t *var, int inside_inv, int *cov_count) JL_NOTSAFEPOINT
+{
+    if (v == (jl_value_t*)var) {
+        if (inside_inv) {
+            return 0;
+        }
+        else {
+            (*cov_count)++;
+            return *cov_count <= 1 || jl_is_concrete_type(var->ub);
+        }
+    }
+    else if (jl_is_uniontype(v)) {
+        return _may_substitute_ub(((jl_uniontype_t*)v)->a, var, inside_inv, cov_count) &&
+            _may_substitute_ub(((jl_uniontype_t*)v)->b, var, inside_inv, cov_count);
+    }
+    else if (jl_is_unionall(v)) {
+        jl_unionall_t *ua = (jl_unionall_t*)v;
+        if (ua->var == var)
+            return 1;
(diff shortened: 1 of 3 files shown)


JeffBezanson:
I updated and rebased this. I think we should probably merge it, since it fixes some crashes. Seems to be a strict improvement.

KristofferC:
@JeffBezanson, this needs a manual backport to https://github.com/JuliaLang/julia/pull/40209

Sacha0:
Thanks for shepherding this along Jeff, and for having an eye on the backport Kristoffer! :)

PR #36222, opened 2020-06-10
Fix undef in cat for eltypes that are not <:Number

### LINKED ISSUE #36041: Problem with cat
It  is a problem I posted on discourse,  but I did not get an authoritative answer. I hope to get a better one posting as an issue.

I  have a problem  with `cat`. I  am using `cat`  all the time to construct block-diagonal matrices:

```
julia> m=[1 2;3 4]
2×2 Array{Int64,2}:
 1  2
 3  4

julia> cat(m,m;dims=(1,2))
4×4 Array{Int64,2}:
 1  2  0  0
 3  4  0  0
 0  0  1  2
 0  0  3  4
```

I am using polynomials

```
julia> using Polynomials

julia> q=Polynomial([0,1])
Polynomial(x)

julia> m=[q q^0;0 q^0]
2×2 Array{Polynomial{Int64},2}:
 Polynomial(x)  Polynomial(1)
 Polynomial(0)  Polynomial(1)

julia> cat(m,m;dims=(1,2))
4×4 Array{Polynomial{Int64},2}:
    Polynomial(x)     Polynomial(1)  #undef            #undef
    Polynomial(0)     Polynomial(1)  #undef            #undef
 #undef            #undef               Polynomial(x)     Polynomial(1)
 #undef            #undef               Polynomial(0)     Polynomial(1)
```

I wonder why these `#undef` while `Polynomials` have a well-defined `zero` function:

```
julia> zero(q)
Polynomial(0)
```

I tracked down the problem to lines 1442–1444 of abstractarray.jl (julia1.4.2):

```
  if T <: Number && count(!iszero, catdims) > 1
        fill!(A, zero(T))
    end
```

Is  not the test T<:Number  wrong? Should not the  right test be that T has the method zero ("duck-typing")?

lassepe:
Fixes #36041 as discussed therein.

### DIFF
--- a/base/abstractarray.jl
+++ b/base/abstractarray.jl
@@ -1509,7 +1509,7 @@ _cat(dims, X...) = cat_t(promote_eltypeof(X...), X...; dims=dims)
     catdims = dims2cat(dims)
     shape = cat_shape(catdims, (), map(cat_size, X)...)
     A = cat_similar(X[1], T, shape)
-    if T <: Number && count(!iszero, catdims) > 1
+    if count(!iszero, catdims) > 1
         fill!(A, zero(T))
     end
     return __cat(A, shape, catdims, X...)
--- a/test/abstractarray.jl
+++ b/test/abstractarray.jl
@@ -610,6 +610,10 @@ function test_cat(::Type{TestAbstractArray})
     # 29172
     @test_throws ArgumentError cat([1], [2], dims=0)
     @test_throws ArgumentError cat([1], [2], dims=[5, -3])
+
+    # 36041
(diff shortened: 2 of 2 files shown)


JeffBezanson:
Thanks! Could you add a test case?

lassepe:
Done. Let me know if you would suggest another way to test this.

lassepe:
The failed test is in [`test/download.jl:50`](https://github.com/JuliaLang/julia/blob/4441245e096431c4210010ee4163cd009ade5ce5/test/download.jl#L50) which seems unrelated.

PR #36223, opened 2020-06-10
julia_versionnumber

### LINKED ISSUE #36221: a URL that stores the current nightly `VERSION`
I want to write a function that connects to the internet on a certain day and gets the latest nightly version. For that, I need a URL that stores this version in it.

Something like (which can run from any julia):
```julia
JuliaVersionNumber(v::VersionNumber) = v
function JuliaVersionNumber(v::String)
	if v == "nightly" || v == "latest"
		# return the VERSION from that URL
	end
end
```
```julia
julia> JuliaVersionNumber("nightly")
v"1.6.0-DEV.192"
```

Is there a way to add this to the nightly build-script?

One way is to program a cron job using `julia-setup` GitHub action, and run it using `nightly` option and then run `VERSION` and then upload that to a URL. But depending on the time of the day that the cron job runs, the nightly `VERSION` may get outdated after some hours. 

This is why I think it should be part of the same script that uploads the nightly binaries.

aminya:
Fixes #36221 

This is a function that connects to GitHub and gets the latest nightly version.

Note: I just picked a name that I thought is suitable. Let me know if you have a better suggestion.

### DIFF
--- a/base/version.jl
+++ b/base/version.jl
@@ -300,3 +300,39 @@ function banner(io::IO = stdout)
         """)
     end
 end
+
+"""
+    julia_versionnumber(v::String)
+    julia_versionnumber(v::VersionNumber)
+
+returns the Julia version number by following the same specifications as VersionNumber.
+
+`julia_versionnumber` will fetch the latest nightly version number if `"nightly"` or `"latest"` is given as the input.
+
+# Examples
+```julia
+julia> julia_versionnumber("nightly")
+v"1.6.0-DEV"
+```
+
+```jldoctest
+julia> julia_versionnumber("1.2.3")
+v"1.2.3"
+
+julia> julia_versionnumber(v"1.2.3")
+v"1.2.3"
+```
+"""
+julia_versionnumber(v::VersionNumber) = v
+function julia_versionnumber(v::String)
+    if in(v, ["nightly", "latest"])
+        version_path = download(
+            "https://raw.githubusercontent.com/JuliaLang/julia/master/VERSION",
+            joinpath(tempdir(), "VERSION.txt"),
+        )
+        version_str = replace(Base.read(version_path, String), "\n" => "")
+        return VersionNumber(version_str)
+    else
+        return VersionNumber(v)
+    end
+end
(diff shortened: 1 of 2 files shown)


Keno (changes requested):
This doesn't seem like it should be in Base. It also commits us to a location of the version file that's fairly random.

aminya:
> Why is this needed in base? This looks like it would be better suited in user package that requires this functionality directly.

It is already there, but [per comment](https://github.com/timholy/SnoopCompile.jl/pull/71#pullrequestreview-407412651) of @timholy, we should move them to other packages that have similar functionality. 

>  We should consider putting the bot utils in a separate package if they aren't specific to SnoopCompile.

I could not find any Version.jl to commit this to. The only version.jl lives in the Julia Base itself.


> It also commits us to a location of the version file that's fairly random.

Exactly that's why I think this belongs to Base. 

There is no doubt that there should be a way to fetch the current nightly version, and if this goes to some other package, the maintainer of that package should monitor Julia repository constantly.

yuyichao:
> Exactly that's why I think this belongs to Base.

It is actually quite conterintuitiviely the opposite. While I don't expect the VERSION file to move (it hasn't fore YEARS...), suppose if it did in 1.7, then the 1.6 version of julia won't be able to access it anymore. OTOH, if it was in a package, the package can just release a new version that works for all julia versions.

Basically the property you are querying is not a property of the julia version you run on, but a property of the latest setup, which shouldn't be tied to a specific julia version.

aminya:
> > Exactly that's why I think this belongs to Base.
> 
> It is actually quite conterintuitiviely the opposite. While I don't expect the VERSION file to move (it hasn't fore YEARS...), suppose if it did in 1.7, then the 1.6 version of julia won't be able to access it anymore. OTOH, if it was in a package, the package can just release a new version that works for all julia versions.
> 
> Basically the property you are querying is not a property of the julia version you run on, but a property of the latest setup, which shouldn't be tied to a specific julia version.

Yes, this is true. However, this has been a common thing. Other than the backports, Julia rarely fixes the compatibility issues of the older versions. That is the job of Compat package.

KristofferC:
> I might just move this to a package of itself and call it JuliaVersionNumber

The user case seems specific enough that just pasting it where you need it right now might be the best.

PR #36227, opened 2020-06-11
allow passing multiple arguments to append! & prepend!

rfourquet:
E.g. `append!([1], [2], [3]) == [1, 2, 3]`.
The equivalent operations for sets (`union!`) and dictionaries (`merge!`)
already support multiple arguments, as well as `push!`.

For `prepend!`, order is maintained in the same way as in `pushfirst!`:
`prepend!([3], [1], [2]) == [1, 2, 3] == pushfirst!([3], 1, 2)`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -92,6 +92,8 @@ New library features
 * New constructor `NamedTuple(iterator)` that constructs a named tuple from a key-value pair iterator.
 * A new `reinterpret(reshape, T, a::AbstractArray{S})` reinterprets `a` to have eltype `T` while potentially
   inserting or consuming the first dimension depending on the ratio of `sizeof(T)` and `sizeof(S)`.
+* New `append!(vector, collections...)` and `prepend!(vector, collections...)` methods accept multiple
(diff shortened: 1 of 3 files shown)


tkf (approved):
LGTM

Keno:
What's the rational for the `prepend!` behavior? Usually when the function modify their first argument, we have varargs behave as lfold.

tkf:
`append!(xs, a, b, c)` is equivalent to `push!(xs, a..., b..., c...)`. So, it makes sense if `prepend!(xs, a, b, c)` is equivalent to `pushfirst!(xs, a..., b..., c...)`, as discussed in the OP.

It might make sense if `pushfirst!` had left-fold behavior but that's too late now.

Keno:
Alright I buy that, but @StefanKarpinski or @JeffBezanson want to give a quick second opinion?

StefanKarpinski:
Agree with what @tkf said, including the rationale and the potential regret.

rfourquet:
News just added. As I think it's an uncontroversial change, I will merge soon.

PR #36229, opened 2020-06-11
add keepat! (the opposite of deleteat!)

rfourquet:
It feels like I'm missing something, but I asked on slack and this functionality doesn't seem to exist yet in `Base`.
`keepat!(vector, indices)` deletes items at all indices of `vector` except those which are specified in `indices`.
In other words, after `keepat!(a, inds)`, `a` is equal to the array which would have been obtained by `a[inds]`.

I'm not sure about the name, it could be maybe `getindex!`, or the feature could be incorporated in `deleteat!` with a keyword.

### DIFF
--- a/base/array.jl
+++ b/base/array.jl
@@ -1368,6 +1368,41 @@ function insert!(a::Array{T,1}, i::Integer, item) where T
 end
 
 """
+    keepat!(a::Vector, inds)
+
+Remove the items at all the indices which are not given by `inds`, and return the modified `a`.
+Items which are kept are shifted to fill the resulting gaps.
+
+`inds` can be either an iterator or a collection of sorted and unique integer indices.
+See also [`deleteat!`](@ref).
+
+# Examples
+```jldoctest
+julia> keepat!([6, 5, 4, 3, 2, 1], 1:2:5)
+3-element Array{Int64,1}:
+ 6
+ 4
+ 2
+```
+"""
+function keepat!(a::Vector, inds)
+    n = length(a)
+    l = 0
+    i = 1
+    for k in inds
+        1 <= k <= n || throw(BoundsError(a, k))
+        l < k || throw(ArgumentError("indices must be unique and sorted"))
+        if i != k
+            @inbounds a[i] = a[k]
+        end
+        l = k
+        i += 1
+    end
+    _deleteend!(a, n-i+1)
+    a
+end
+
+"""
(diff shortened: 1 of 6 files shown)


StefanKarpinski:
Maybe better expressed with https://github.com/mbauman/InvertedIndices.jl and `deleteat!`?

rfourquet:
> Maybe better expressed with https://github.com/mbauman/InvertedIndices.jl and deleteat!?

Maybe! Someone pointed out this package to me and I forgot to mention it here. But it's quite less discoverable than `deleteat!`, while the operation of "keeping only these indices" doesn't seem to be any less fundamental than "delete all these indices". Although it might be less often needed given that this feature isn't in `Base` yet and doesn't seem to have been requested before. But given how often the non-mutating equivalent `vector[inds]` is used, I'm quite surprised by this fact.

base/array.jl
```diff
@@ -1288,6 +1288,41 @@ function insert!(a::Array{T,1}, i::Integer, item) where T
...
+        l < k || throw(ArgumentError("indices must be unique and sorted"))
+        if i != k
+            @inbounds a[i] = a[k]
+        end
+        l = k
+        i += 1
+    end
+    _deleteend!(a, n-i+1)
+    a
+end
```
vtjnash, line 1324: Should we make this generic and move it to abstractarrays.jl?

```suggestion
function keepat!(a::AbstractVector, inds)
    isempty(inds) && return a
    local prev
    i = firstindex(a)
    for k in inds
        @isdefined(prev) && (prev <= k || throw(ArgumentError("indices must be sorted")))
        if i != k
            a[i] = a[k]
        end
        prev = k
        i = nextind(a, i)
    end
    deleteat!(a, i:lastindex(a))
    return a
end
```

vtjnash (approved):
Sounds reasonable to me. I think the implementation can be generalized, but otherwise, lgtm conceptually. I'd merge if you finish it :)

base/array.jl
rfourquet: I added an `@inbouds` in the generic version to mitigate a bit the slow-down, removed the `isempty` check which was incorrect. I also re-added the constraint to have unique indices, otherwise we have to add another check, unless we would want `keepat!([1, 2, 3], [1, 2, 2]) == [1, 2, 2])` ? For now I prefer the non-ambiguous behavior, which might also catch bugs from the user.

PR #36231, opened 2020-06-11
deleteat! : check bounds for the first passed index

rfourquet:
All other indices are bound-checked.
Currently, the behavior looks like:
```julia
julia> deleteat!([1:1000;], [-100])

signal (11): Segmentation fault
[...]

julia>  deleteat!(BigInt[0], [0])
free(): invalid next size (normal)

signal (6): Aborted
[...]

julia> deleteat!(BigInt[0], [-100])
[...]

julia> deleteat!([0], [0])
Int64[]

julia> deleteat!([0], [2])
1-element Array{Int64,1}:
 0

julia> deleteat!([0], [3])
ERROR: InexactError: check_top_bit(UInt64, -1)
[...]
```
With this commit, all these expressions throw a `BoundsError`.

### DIFF
--- a/base/array.jl
+++ b/base/array.jl
@@ -1353,7 +1353,7 @@ function _deleteat!(a::Vector, inds, dltd=Nowhere())
     y = iterate(inds)
     y === nothing && return a
     (p, s) = y
-    1 <= p <= n || throw(BoundsError(a, p))
+    checkbounds(a, p)
     push!(dltd, @inbounds a[p])
     q = p+1
     while true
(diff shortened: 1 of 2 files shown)


base/array.jl
```diff
@@ -1352,9 +1352,9 @@ function _deleteat!(a::Vector, inds, dltd=Nowhere())
     n = length(a)
     y = iterate(inds)
     y === nothing && return a
-    n == 0 && throw(BoundsError(a, inds))
     (p, s) = y
-    p <= n && push!(dltd, @inbounds a[p])
+    1 <= p <= n || throw(BoundsError(a, p))
```
simeonschaub, line 1356:
```suggestion
    checkbounds(a, p)
```

JeffBezanson:
Wow, good catch.

PR #24836, opened 2017-11-29
WIP/RFC: add push function: non-mutable counterpart of push!

rfourquet:
This is for
1. a bit of convenience for `IOContext`. I always found it a bit strange to describe `IOContext` as an immutable dictionary, but to not have a way to push elements to if without calling its constructor, so now we can do `push(io, :compact => true)` instead of `IOContext(io, :compact => true)`;
2. and to claim `push` in base (i.e. set/reserve the meaning, for use by package implementing immutable data structures, see e.g. https://github.com/JuliaCollections/FunctionalCollections.jl/); few more functions could be added, like `pop` and `append`.

PR #36239, opened 2020-06-11
add push(::IO, pair...) and push(::ImmutableDict, pair...)

rfourquet:
This is part of #24836 with what seems to be the least controversial: this doesn't add a generic `push` implementation for mutable collections, and doesn't even implement specific implementation for specific containers, but just defines push on `ImmutableDict`, and on `IO`.

Both types are considered as immutable dictionaries, they support `get` and `getindex`, and it seems natural that they also support `push`ing.
Triage label was already applied to the other PR and removed, I'm not sure if it needs to be applied here.

### DIFF
--- a/base/exports.jl
+++ b/base/exports.jl
@@ -475,6 +475,7 @@ export
     pop!,
     popat!,
     prepend!,
+    push,
     push!,
     resize!,
     popfirst!,
--- a/base/show.jl
+++ b/base/show.jl
@@ -217,7 +217,7 @@ print(io::IO, s::Symbol) = (write(io,s); nothing)
 `IOContext` provides a mechanism for passing output configuration settings among [`show`](@ref) methods.
 
 In short, it is an immutable dictionary that is a subclass of `IO`. It supports standard
-dictionary operations such as [`getindex`](@ref), and can also be used as an I/O stream.
+dictionary operations such as [`getindex`](@ref) and [`push`](@ref), and can also be used as an I/O stream.
 """
 struct IOContext{IO_t <: IO} <: AbstractPipe
     io::IO_t
@@ -323,6 +323,15 @@ short
 ```
 """
 IOContext(io::IO, KV::Pair, KVs::Pair...) = IOContext(IOContext(io, KV), KVs...)
+
+"""
+    push(io::IO, KV::Pair...)
+
+Create an `IOContext` that wraps a given stream, adding the specified `key=>value` pairs to
+the properties of that stream (note that `io` can itself be an `IOContext`).
+This is equivalent to [`IOContext(io, KV...)`](@ref).
+"""
+push(io::IO, KV::Pair...) = IOContext(io, KVs...)
 
 show(io::IO, ctx::IOContext) = (print(io, "IOContext("); show(io, ctx.io); print(io, ")"))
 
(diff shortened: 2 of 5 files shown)


base/exports.jl
```diff
@@ -475,6 +475,7 @@ export
     pop!,
     popat!,
     prepend!,
+    push,
```
tkf, line 478: As I explained in https://github.com/JuliaLang/julia/pull/24836#issuecomment-643067749, it's better to avoid exporting this.

rfourquet: I don't understand well this other comment in the context of this present PR, where the meaning of `push` is made minimal: `push(dict::ImmutableDict, pair...)` cannot mean anything else I think than what it does in this PR. It doesnt seem breaking to add widening so that another dictionary type is returned (or very slightly so, in case we widen the type instead of convert the new pair to the `eltype` of `dict` but `ImmutableDict`). In any case, I'm fine not exported it but I don't understand what this changes, i.e. I don't see how "It is a big loss if we lose this API" depends on whether the symbol is exported or not...  is it about making it public API (that has to be stable) or not?

base/show.jl
```diff
@@ -324,6 +324,15 @@ short
 """
 IOContext(io::IO, KV::Pair, KVs::Pair...) = IOContext(IOContext(io, KV), KVs...)
 
+"""
+    push(io::IO, KV::Pair...)
+
+Create an `IOContext` that wraps a given stream, adding the specified `key=>value` pairs to
+the properties of that stream (note that `io` can itself be an `IOContext`).
+This is equivalent to [`IOContext(io, KV...)`](@ref).
```
fredrikekre, line 332: Whats the advantage of `push` then? Personally I would prefer reading code that had `IOContext(io, kv...)`. `push(io, ...)` feels like it should write something to the `io`, not returning an io with modified context.

base/exports.jl
tkf: The loss I mentioned was about `push(::AbstractVector, _)` and `push` on mutable containers in general. The major concern in #24836 for implementing `push(::AbstractVector, _)` was that some new users might try `push` instead of `push!`.  Not exporting `push` allows us to add `push(::AbstractVector, _)` without worrying about confusing new users.  (Implementing `push(::AbstractVector, _)` doesn't have to be done in this PR, of course.)

base/show.jl
tkf: Alternative API might be `Base.setindex(io, V, K)` and/or `merge(io, (K1 = V1, K2 = V2))`.

base/exports.jl
rfourquet: Ok I understand now. I'm not conviced (we have other non-mutating functions which are exported), but the `export` can be the subject of another PR then.

base/show.jl
rfourquet: I agree that `push` is less convincing for `IOContext` than for `ImmutableDict`. Still, `IO` types already have the "dictionary" semantics and `getindex`. But `IO` is special enough compared to "normal" collections that I see this can be more controversial to add `push` for it. Seems like I will have to shrink this shrinked PR even more ;-)

base/exports.jl
tkf: Yeah, I fully agree that it's not convincing that defining (e.g.) `merge`/`merge!` is OK and `push`/`push!` is not. But it's better to avoid the situation that we can't define `push` on `AbstractVector` just because it's exported.

PR #36247, opened 2020-06-11
When using mismatched deps version, add a removal suggestion

DilumAluthge:
cc: @JeffBezanson @staticfloat

### DIFF
--- a/deps/tools/uninstallers.mk
+++ b/deps/tools/uninstallers.mk
@@ -30,5 +30,5 @@ $(addprefix version-check-,$(DEP_LIBS_STAGED)) : version-check-% : install-%
 		echo "  want $(UNINSTALL_$*)" ; \
 		echo "  To resolve this warning, you could try either of the following suggestions: " ; \
 		echo "  1. Run the following command: make -C deps uninstall" ; \
-		echo "  2. Remove the JULIA_HOME/usr directory, where you replace JULIA_HOME with the path of the directory in which you are building Julia " ; \
+		echo "  2. Remove the following directory: $(JULIAHOME)/usr" ; \
 	fi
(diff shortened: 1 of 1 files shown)


deps/tools/uninstallers.mk
```diff
@@ -28,4 +28,5 @@ $(addprefix version-check-,$(DEP_LIBS_STAGED)) : version-check-% : install-%
 			[ "$(UNINSTALL_$*)" != "`cat $(build_prefix)/manifest/$*`" ]) ; then \
 		echo "WARNING: using mismatched version for $$(cat $(build_prefix)/manifest/$*):" ; \
 		echo "  want $(UNINSTALL_$*)" ; \
+		echo "  Try deleting the usr directory." ; \
```
staticfloat, line 31:
```suggestion
		echo "  Try deleting the $(JULIAHOME)/usr directory." ; \
```

Let's not leave any ambiguity as to which `usr` directory we're advocating the deletion of.  ;)

staticfloat (approved):

deps/tools/uninstallers.mk
vtjnash: Tries incorrectly to test on my system...
```
$ echo $(JULIAHOME)/usr 
/usr
```

Yeah, let's leave no doubt 😂

We could also suggest some of the less nuclear options (e.g. `make -C deps uninstall`)?

DilumAluthge: "Try deleting the usr subdirectory in the build directory"?

DilumAluthge: @staticfloat @vtjnash I have made the path of the usr directory more explicit. What do you think?

I also added @vtjnash's suggestion for a less drastic option

staticfloat: I’m not sure Jameson was serious about not using the actual $(JULIAHOME) directory, I think he was making a joke about copy-pasting the value into his terminal from the makefile source. If you use the actual variable value, it will print out the actual folder that should be deleted, which seems like the best option overall.

DilumAluthge: Ah hahaha whoops my bad. Okay let me fix that. Should I leave in his suggestion about `make -C deps uninstall`?

DilumAluthge: I've fixed the message to use the actual value of the `JULIAHOME` variable.

staticfloat (approved):

PR #36249, opened 2020-06-11
Rewording related to disability

goretkin:
I believe these small changes to be worthwhile. I believe there are some appropriate figurative uses of "blind" (e.g. https://www.realsocialskills.org/blog/blind-spot-is-not-a-disability-analogy), but I don't think the uses in this PR are like that.

### DIFF
--- a/base/abstractarray.jl
+++ b/base/abstractarray.jl
@@ -2260,7 +2260,7 @@ function hash(A::AbstractArray, h::UInt)
     # hashes will often subsequently be compared by equality -- and equality between arrays
     # works elementwise forwards and is short-circuiting. This means that a collision
     # between arrays that differ by elements at the beginning is cheaper than one where the
-    # difference is towards the end. Furthermore, blindly choosing log(N) entries from a
+    # difference is towards the end. Furthermore, naively choosing log(N) entries from a
     # sparse array will likely only choose the same element repeatedly (zero in this case).
 
     # To achieve this, we work backwards, starting by hashing the last element of the
--- a/base/abstractarray.jl
+++ b/base/abstractarray.jl
@@ -2260,7 +2260,7 @@ function hash(A::AbstractArray, h::UInt)
     # hashes will often subsequently be compared by equality -- and equality between arrays
     # works elementwise forwards and is short-circuiting. This means that a collision
     # between arrays that differ by elements at the beginning is cheaper than one where the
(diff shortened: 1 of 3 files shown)


base/abstractarray.jl
```diff
@@ -2260,7 +2260,7 @@ function hash(A::AbstractArray, h::UInt)
     # hashes will often subsequently be compared by equality -- and equality between arrays
     # works elementwise forwards and is short-circuiting. This means that a collision
     # between arrays that differ by elements at the beginning is cheaper than one where the
-    # difference is towards the end. Furthermore, blindly choosing log(N) entries from a
+    # difference is towards the end. Furthermore, naively choosing log(N) entries from a
```
Keno, line 2263: `arbitrarily`?

doc/src/manual/methods.md
```diff
@@ -906,7 +906,7 @@ f(x::Int, y::Int) = 3
 ```
 
 This is often the right strategy; however, there are circumstances
-where following this advice blindly can be counterproductive. In
+where following this advice carelessly can be counterproductive. In
```
Keno, line 909: `mindlessly`? It's more about acting without thinking about whether it is the correct solution than acting without care, though of course the meanings are close.

base/abstractarray.jl
goretkin: Though as I'm re-reading I'm confused by what the mitigation for not getting a bunch of zeros from a sparse array is

Keno (approved):

base/abstractarray.jl
Keno: The problem with sparse arrays is that zeros tend to be clustered in blocks and along dimensions, so you don't want to do things like pick the first `log(N)` elements, or all the elements in some dimension or along some diagonal or something with regular structure like that. So this does a fibonacci selection. Peudo-random with some low-entropy seed would probably also work.

goretkin: Right, but aren't all those mitigations important in the dense case as well? For two `size`-`(10, 10)` arrays, you could have the majority of elements differ but you keep sampling every 11th element, and you just pick out some structured subset (like the diagonal).

Just asking in case there's an additional opportunity to clarify the comment.

Keno: Sparse here is used a bit loosely. If the array is a `DenseArray` that just happens to have a sparsity pattern, the same concern applies. This function is applicable to all `AbstractArray`s.

PR #18218, opened 2016-08-24
Make matrix multiplication work for more types

blegat:
Matrix multiplication does not work for arrays of elements types `T` and `S` such that the type `TS = promote_op(*, T, S)` does not satisfy `TS == promote_op(+, TS, TS)`.

For example, in [this package](https://github.com/blegat/SumOfSquares.jl), I have the type `Term` which is one term of a polynomial and the type `Polynomial` which is the sum of several terms.
Suppose I am trying to compute `A * x` where `A` is a matrix of `Int64` and `x` is a vector of `Term`.
Then `TS` will be `Term` so it will allocate a vector of `Term` to store the result. However when doing the sum of `Term` it will get `Polynomial` so what should have been allocated is a vector of `Polynomial`.

The fix is simple, allocate an array of type `promote_op(+, TS, TS)`.

PR #36256, opened 2020-06-12
Fix Matrix on diagonal and triangular matrices for types without zero

blegat:
Similar to https://github.com/JuliaLang/julia/pull/35743

### DIFF
--- a/stdlib/LinearAlgebra/src/bidiag.jl
+++ b/stdlib/LinearAlgebra/src/bidiag.jl
@@ -163,7 +163,7 @@ function Matrix{T}(A::Bidiagonal) where T
     B[n,n] = A.dv[n]
     return B
 end
-Matrix(A::Bidiagonal{T}) where {T} = Matrix{T}(A)
+Matrix(A::Bidiagonal{T}) where {T} = Matrix{promote_with_zero(T)}(A)
 Array(A::Bidiagonal) = Matrix(A)
 promote_rule(::Type{Matrix{T}}, ::Type{<:Bidiagonal{S}}) where {T,S} =
     @isdefined(T) && @isdefined(S) ? Matrix{promote_type(T,S)} : Matrix
--- a/stdlib/LinearAlgebra/src/dense.jl
+++ b/stdlib/LinearAlgebra/src/dense.jl
@@ -299,11 +299,15 @@ function diagm_size(size::Tuple{Int,Int}, kv::Pair{<:Integer,<:AbstractVector}..
     (m ≥ mmax && n ≥ nmax) || throw(DimensionMismatch("invalid size=$size"))
     return m, n
 end
+# For some type `T`, `zero(T)` is not a `T` and `zeros(T, ...)` fails.
+# `zero_type(T::Type) = typeof(zero(T))` would not work for types such
+# as `Array` for which `zero(::T)` is defined but not `zero(::Type{T})`
+# so we use `promote_op` instead.
+zero_type(T::Type) = promote_op(zero, T)
+promote_with_zero(T::Type) = promote_type(T, zero_type(T))
 function diagm_container(size, kv::Pair{<:Integer,<:AbstractVector}...)
     T = promote_type(map(x -> eltype(x.second), kv)...)
-    # For some type `T`, `zero(T)` is not a `T` and `zeros(T, ...)` fails.
-    U = promote_type(T, typeof(zero(T)))
-    return zeros(U, diagm_size(size, kv...)...)
+    return zeros(promote_with_zero(T), diagm_size(size, kv...)...)
 end
 diagm_container(size, kv::Pair{<:Integer,<:BitVector}...) =
     falses(diagm_size(size, kv...)...)
--- a/stdlib/LinearAlgebra/src/triangular.jl
+++ b/stdlib/LinearAlgebra/src/triangular.jl
@@ -31,7 +31,7 @@ for t in (:LowerTriangular, :UnitLowerTriangular, :UpperTriangular,
             Anew = convert(AbstractMatrix{T}, A.data)
             $t(Anew)
         end
-        Matrix(A::$t{T}) where {T} = Matrix{T}(A)
+        Matrix(A::$t{T}) where {T} = Matrix{promote_with_zero(T)}(A)
 
         size(A::$t, d) = size(A.data, d)
         size(A::$t) = size(A.data)
--- a/stdlib/LinearAlgebra/src/tridiag.jl
+++ b/stdlib/LinearAlgebra/src/tridiag.jl
@@ -138,7 +138,7 @@ function Matrix{T}(M::SymTridiagonal) where T
     end
     return Mf
 end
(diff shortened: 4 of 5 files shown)


blegat:
The failing tests are due to the types `T` such as `Matrix{Float64}` for which `zero(::T)` is defined but `zero(::Type{T})` is not defined.
For these types, `Matrix{T}(::LowerTriangular{T})` works as it only requires `zero(::T)` to be defined but the current PR does not work as it calls `typeof(zero(T))`.
I see two solutions to this:
1. Define a function `zero_type(T::Type) = typeof(zero(T))` that should be implemented for `Array` next to
https://github.com/JuliaLang/julia/blob/8d91b1123f4be3c03e587710fa8b0777762c5e07/base/abstractarray.jl#L970
and for `AbstractSparseArray` next to
https://github.com/JuliaLang/julia/blob/8d91b1123f4be3c03e587710fa8b0777762c5e07/stdlib/SparseArrays/src/SparseArrays.jl#L55
The issue with this is that it can be considered a breaking change.
2. Use `Base.promote_op(zero, T)` instead of `typeof(zero(T))`.

andreasnoack:
I'm a little reluctant about these changes. My gut feeling is that the `zero(T) isa T` is generally assumed in the linear algebra code here and elsewhere. I'm not saying this is change is necessarily wrong but I'm not sure if we understand the implications.

blegat:
I understand that many parts in `LinearAlgebra` were written with `Number`s in mind and in fact, most of it continues to work with other types such as JuMP functions and polynomial functions.
There are a few places where too much is assumed and this triggers error when, e.g. users of JuMP (resp. MultivariatePolynomials) are trying to do linear algebra with JuMP functions (resp. polynomials).
We define a type `AbstractMutable` in MutableArithmetics. Subtypes of `AbstractMutable` are mutable types. Some methods in stdlib (e.g. `mul!`, `dot`) would be more efficient if they exploit their mutability and we redefine these methods for `AbstractMutable`.
It happens that the mutable types are also often types with non-`Number` behaviors (e.g. `zero(T) != T`) so we also try to redefine methods for `AbstractMutable` where these assumptions are made.
This is however a bit hacky and was no the role of `AbstractMutable` (it is meant to rewrite methods for performance by exploiting mutability).
We made a list here of the different issues with assumptions made in `LinearAlgebra` in:
https://github.com/jump-dev/MutableArithmetics.jl/issues/47
You can see that many places where these kind of assumptions were made in LinearAlgebra were already fixed, this is currently the last one remaining.
In fact, this PR is similar in spirit with https://github.com/JuliaLang/julia/pull/18218/ where we also needed `promote_op` to handle more general cases.

tkf:
> Some methods in stdlib (e.g. `mul!`, `dot`) would be more efficient if they exploit their mutability and we redefine these methods for `AbstractMutable`.

I don't think relying on computations on the type domain is a good practice in Julia. It doesn't compose well because it's rather unavoidable to _rely_ on `promote_op` (or very complicated trait system) if you go with such design principle.  Using `return_type` (via `promote_op`) like this means that your API is undefined.  (Yes, it applies to a lot of Base/stdlib APIs, too.)

I do agree exploiting in-place operations on nested mutable data structure would be highly beneficial.  But I think the logic has to occur in the value domain rather than in type domain.  The only route I can imagine for this to happen is to embrace mutate-or-widen strategy everywhere recursively in LinearAlgebra/Base functions.  I'm exploring this in [BangBang.jl](https://github.com/JuliaFolds/BangBang.jl).  It is primary for "containers" but I added a few compute-oriented APIs like `materialize!!`, `mul!!`, etc. and I find it powerful.  Of course, using mutate-or-widen strategy doesn't mean that we can't use `return_type`.  We just have to make sure that it is used solely for optimization.

I understand that this is not a short-term solution but I think it's something worth keep in mind before using `promote_op` everywhere and designing APIs relying on that.

blegat:
Thanks for the link to `BangBang`, it looks interesting. The difference with MutableArithmetics (MA) is that the latter focus more on arithmetics/algebra and operation that originally do not mutate and return the result such as `*`, `+`, `-`, ...

> it's rather unavoidable to _rely_ on `promote_op` (or very complicated trait system) if you go with such design principle.

In order to write algorithms that work for any type and still exploit mutability when possible, you must be able to apply an operation on some objects and store the result in some other object by mutating it **if possible** and, in any case, return the result (for instance, with `Int`, it would not mutate it and with `BigInt` it will but as your code is written for any type, you don't know which case it is). This is provided by `MA.operate!` and `MA.operate_to!`.
It turns out that this is quite boilerplate and error-prone to implement as it's doing two things (checking whether we should mutate and mutate if we can).
Moreover, sometimes, you have a code that is not type-independent and you know that you want to mutate the object and if it's not possible then it's a bug and you want an error, not silently inefficient code. This is provided by `MA.mutable_operate!` and `MA.mutable_operate_to!`.
The implementation of `MA.operate!` can just redirect to `MA.mutable_operate!` but as mentioned it is quite error-prone to implement and in fact there is one way to implement it that relies on a fact that seems to work for all types that implements MA:
Either `output` is not mutable (e.g. `Int`, `Float64`) and `MA.operate_to!(op, output, ...)` should just redirect to `op(...)` or it is and then, if `op(...)` has the same type as `output` then we should redirect to `MA.mutable_operate_to!(op, output, ...)`, otherwise, we should just redirect to `op(...)`.
To do this, you need to be able to get the return type so types should implement `MA.promote_operation` (as it does not rely on `promote_op`).
While it seems like a lot of work to require the types to implement `MA.promote_operation`, it is still less work than implementing `MA.operate!` and `MA.operate_to!` and you get a reliable `MA.promote_operation` that user can use for free (and it's ok to use it, not like `Base.promote_op` which should be avoided). So it's a case when by implementing less, you get more features!
I'm not saying that in Julia, all types should implement `MA.promote_operation` for all functions in addition to implementing the functions.
I'm just saying that for certain functions (typically the one encountered in arithmetics). The approach seems very fruitful.

Another example where we could get more features with less effort is matrix multiplication.
Currently in LinearAlgebra and SparseArrays (`spmatmul!` is not doing exactly that but it could be rewritten like this), there are many implementations of `*(A, B)` that basically do
1) Guess the type of the resulting array
2) Create a new array to store the result
4) Redirect to `LinearAlgebra.mul!`

So in fact, we could have only one method for `*` and have many methods for giving the type of the resulting array and many methods for creating an new array from the type of an array (which already exist, e.g. `similar`).
This is the approach we used in MA and it seems to work quite well: https://github.com/jump-dev/MutableArithmetics.jl/blob/7057356d3f154b1a76ca385e03a650d834c80665/src/linear_algebra.jl#L215-L226.
By doing this, we get more features, as now there is a function that can be used to get the return type of an array multiplication, and it requires less effort as implementing this function is less work than implementing `*` since it is one of the task that is currently done in `*` anyway.

inkydragon:
We have moved the LinearAlgebra stdlib to an external repo: [JuliaLang/LinearAlgebra.jl
](https://github.com/JuliaLang/LinearAlgebra.jl)

If you think that this PR is still relevant, please open a new PR on the LinearAlgebra.jl repo.

PR #36257, opened 2020-06-12
 vscode julia path setting

xgdgsc:
This would enable vscode julia extension to be able to start julia REPL out of the box

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,3 +1,5 @@
 FROM julia:latest
 
 RUN apt-get update && apt-get install -y build-essential libatomic1 python gfortran perl wget m4 cmake pkg-config git
+RUN mkdir -p /root/.vscode-server/data/Machine/
+COPY settings.json /root/.vscode-server/data/Machine/
--- a/.devcontainer/settings.json
+++ b/.devcontainer/settings.json
@@ -1,0 +1,4 @@
+{
+    "julia.executablePath": "/usr/local/julia/bin/julia",
+    "julia.NumThreads": 4
+}


StefanKarpinski:
cc @davidanthoff

davidanthoff:
I don't understand why we would need any of this, and some of it seems like a really bad idea (modifying things inside `/root/.vscode-server`).

The docker image we use has `julia` on the path, the VS Code extension picks that up and should just work, so I don't think there is any need to manually configure a path there. I just tried it, and it all seems to work smoothly.

Setting a fixed number of threads (i.e. four) also seems like a bad idea, that is a user setting, not something that should be committed to the repo here.

xgdgsc:
But why I didin' t have julia extension able to find julia out of the box until I set this?

davidanthoff:
I tried it this this morning by using the VS Code command `Remote-Containers: Open Repository in Container...`, entering `julialang/julia` as the repo to clone and then just waiting. Once all the prep was done, everything worked and the Julia extension found the Julia binary just fine.

If it doesn't for you, it is either a bug in the extension, or something else goes haywire in your setup, because the extension should just work if there is a `julia` on the path, and the docker image in this devcontainer _has_ julia on the path.

Can you describe in more detail what you are actually doing to use this feature here?

xgdgsc:
I just `git clone https://github.com/JuliaLang/julia.git` and open that folder with vscode. Then click the right bottom notification that tells you a devcontainer is found and then it builds that image. And after finish, if I run `julia: start REPL` it shows:

```
Command 'Julia: Start REPL' resulted in an error (command 'language-julia.startREPL' not found)
```

If I add those in settings.json and rebuild it works. If I remove them and rebuild it shows the error again.

davidanthoff:
Ah, that is very helpful!

I think what is happening here is that you have a configuration value set on your local system that points to a julia binary on your local system, and then that config value is _also_ used in the container, which of course makes no sense.

I think this is really a bug in the Julia VS Code extension. We should mark the configuration setting as `machine-overridable` (I think), so that your user setting from your main machine isn't used in the container.

I think a workaround for now would be to use the description [here](https://code.visualstudio.com/remote-tutorials/containers/how-it-works#_devcontainerjson) to make sure the container gets a machine specific settings override. That is essentially the same strategy that you are using in this PR here, but instead of copying the file it uses the supported default settings story of devcontainers. I think I would just change the devcontainer.json to read:
```
{
    "extensions": [
      "julialang.language-julia",
      "ms-vscode.cpptools"
    ],
    "settings": {
      "julia.executablePath": "julia"
    },
    "dockerFile": "Dockerfile"
}
```
That way we don't encode a full path there, and I think that should be enough.

But I still think that once we have shipped a new version of the extension where the configuration scope issue is fixed, we should probably remove that setting here again.

davidanthoff:
Fix on the extension side is at https://github.com/julia-vscode/julia-vscode/pull/1340.

xgdgsc:
Thanks! I guess we can skip the workaround and wait for the release.

Issue #36263, opened 2020-06-12
`show_default`, but without type parameters


goretkin:
Definitions like https://github.com/JuliaLang/julia/blob/1e761112038518bafe6ce37b646ae38f38d28253/base/range.jl#L742 could be no more than something like

```julia
show(io::IO, r::OneTo) = show_default(io, r; type_parameters=false)
```

There are many places where that's a default behavior that is wanted. For example, when all the type parameters correspond to fields that when `show`'d reveal their type. For example

```julia
struct Foo{T}
  x::T
end

show(Foo(Base.OneTo(3)))
```

produces

`Foo{Base.OneTo{Int64}}(Base.OneTo(3))`,
which is unnecessary.

Also, `show_default` has great logic to use `IOContext` to see which names need to be qualified or not:

```julia
julia> Base.show_default(stdout, Base.OneTo(3))
Base.OneTo{Int64}(3)
julia> Base.show_default(stdout, Foo(Base.OneTo(3)))
Foo{Base.OneTo{Int64}}(Base.OneTo(3))
julia> using Base: OneTo

julia> Base.show_default(stdout, Base.OneTo(3))
OneTo{Int64}(3)
julia> Base.show_default(stdout, Foo(Base.OneTo(3)))
Foo{OneTo{Int64}}(Base.OneTo(3))
```

Notice the inconsistency in `Foo{OneTo{Int64}}(Base.OneTo(3))`. 

x-ref: https://github.com/JuliaLang/julia/issues/33260

goretkin:
Seems suspiciously relevant
https://github.com/JuliaLang/julia/blob/3222aba1d32ab9af41c82a1dd0bce0a7f72d8cd5/base/show.jl#L477-L483 but doesn't do the job fully.

One idea is a trait-based API, where a type could register "when you `show` me, my type is obvious", but that would require more careful design since it would likely be public.

goretkin:
Related: https://github.com/MrVPlusOne/SimpleTypePrint

timholy:
If a change like this is made, it shouldn't be done without an awareness that at least until recently `show_default` was a major invalidation risk. One would want to run the tests in #37193, which had to be reverted (see #38461), to make sure this kind of change doesn't cause unintended problems.

timholy:
Also,
```julia
julia> r = Base.OneTo(3)
Base.OneTo(3)

julia> io = IOBuffer();

julia> @btime show($io, $r)
[...]

PR #36264, opened 2020-06-13
Generalize `Base.show_default` 

goretkin:
Summary of behavior:

```julia
julia> struct Foo{T}
       x::T
       end

julia> Foo(Base.OneTo(3))
Foo{Base.OneTo{Int64}}(Base.OneTo(3))

julia> using Base: OneTo

julia> Foo(Base.OneTo(3))
Foo{OneTo{Int64}}(OneTo(3))

julia> Base.show(io::IO, f::Foo) = Base.show_default(io, f; f_type=t -> t.name)

julia> Foo(Base.OneTo(3))
Foo(OneTo(3))
```

See #36263 

This needs to look better than requiring the anonymous function. The reason I didn't just use a flag is because I thought there might be a good way to selectively print some type parameters and not other. For example, for a type like

```julia
struct Bar{S, T}
    x::S
end
```
one might desire to print it as `Bar{_, 2}(Base.OneTo(3))` instead of
`Bar(Base.OneTo(3))` or `Bar{Base.OneTo{Int64}, 2}(Base.OneTo(3))`

This also needs tests. Question: how should I test the following behavior:

```julia
julia> Foo(Base.OneTo(3))
Foo{Base.OneTo{Int64}}(Base.OneTo(3))

julia> using Base: OneTo

julia> Foo(Base.OneTo(3))
Foo{OneTo{Int64}}(OneTo(3))
```

### DIFF
--- a/base/indices.jl
+++ b/base/indices.jl
@@ -358,7 +358,14 @@ unsafe_length(S::Slice) = unsafe_length(S.indices)
 getindex(S::Slice, i::Int) = (@_inline_meta; @boundscheck checkbounds(S, i); i)
 getindex(S::Slice, i::AbstractUnitRange{<:Integer}) = (@_inline_meta; @boundscheck checkbounds(S, i); i)
 getindex(S::Slice, i::StepRange{<:Integer}) = (@_inline_meta; @boundscheck checkbounds(S, i); i)
-show(io::IO, r::Slice) = show_default(io, r; f_type = t -> t.name)
+
+function show(io::IO, r::Slice)
+    show(io, typeof(r).name)
+    print(io, "(")
+    show_fields(io, r)
+    print(io, ")")
+end
+
 iterate(S::Slice, s...) = iterate(S.indices, s...)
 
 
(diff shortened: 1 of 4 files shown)


JeffBezanson:
This definitely needs a less flexible API for the common case. For more customization, I think a good way to do it instead of passing a callback is to rename `_show_default_body` to something nicer like `show_fields`, and then you can implement a `show` method as
```
    # custom type printing code here
    print(io, "(")
    Base.show_fields(io, x)
    print(io, ")")
```

goretkin:
Thanks for the suggestion. I wasn't sure if there should be a function

```
function show_no_type_parameters(io::IO, r)
    show(io, typeof(r).name)
    print(io, "(")
    show_fields(io, r)
    print(io, ")")
end
```

or if it should be repeated where necessary.

goretkin:
Wondering about 

> This also needs tests. Question: how should I test the following behavior:
> 
> ```julia
> julia> Foo(Base.OneTo(3))
> Foo{Base.OneTo{Int64}}(Base.OneTo(3))
> 
> julia> using Base: OneTo
> 
> julia> Foo(Base.OneTo(3))
> Foo{OneTo{Int64}}(OneTo(3))
> ```

and any other suggestions to move this along.

timholy:
Worth cross-posting https://github.com/JuliaLang/julia/issues/36263#issuecomment-728805556 here. Benchmarks needed before this can be merged.

DilumAluthge:

@goretkin If you're still interested in working on this PR, let me know, and I can re-open it.

Issue #31442, opened 2019-03-22
Extrema is slower than maximum + minimum


mbauman:
On AVX-512 systems, calling `extrema(x)` is slower than calling `maximum` and then `minimum`.  Looks like we need to incorporate the same techniques from https://github.com/JuliaLang/julia/pull/30320 over to extrema in order to get it to vectorize in a similar manner.

sagartewari01:
@mbauman Can you provide the test case you used? I ran the following code:

```
arrs = []

for i in 1:100000
    insert!(arrs, 1, rand(1000))
end

println("running extrema")
@time map(arrs) do arr
    extrema(arr)
end

println("running minimum, maximum")
@time map(arrs) do arr
    (minimum(arr), maximum(arr))
end
```

and got this output:
```
running extrema
  0.520760 seconds (301.98 k allocations: 14.811 MiB)
running minimum, maximum
  0.741328 seconds (279.30 k allocations: 13.814 MiB)
```

julia version:
```
julia version 1.1.0
```

mbauman:
Yes, that test case reproduces the behavior on master for me.  It won't on 1.1.0 as the PR that improved maximum and minimum was after 1.1.0.  Of course, note that you'll also need to have a relatively recent computer to see the difference.

nlw0:
I got a slower `extrema` in my Core i5 machine.

```
running extrema
  314.316 ms (100003 allocations: 4.58 MiB)
running minimum, maximum
  224.977 ms (100003 allocations: 4.58 MiB)
```

nlw0:
I would like to try working on this, and it would be great to hear what kind of patch is expected. I was thinking maybe `mapreduce_impl` could be turned into a generated function that can handle multiple functions simultaneously. Does that sound good?

One question I have is about the unrolling. Ideally this should be adapted for each architecture and data type. How much of this should be up to LLVM or to Julia? Should we prioritize generating code that the LLVM can vectorize? The alternative I imagine would be controlling parameters like this unrolling depending on the architecture.

nlw0:
Thinking a bit more about this, maybe the best of worlds would be if this function could simply be implemented with a general fold-left, and it should result in a good vectorized machine code relying on the LLVM automatic vectorization. Should we concentrate our efforts on making sure that a generic fold works better in more cases, or should we first make sure that a few specific functions have high-performance implementations?

mbauman:
[...]

PR #36265, opened 2020-06-13
Define `extrema` using `mapreduce`; support `init`

tkf:
Like #36188, but now for `extrema`.

This patch also switches the implementation of `extrema` to use `mapreduce`.  It improves the performance even for simple arrays:

```julia
julia> @btime extrema($(randn(1000)));
  5.502 μs (0 allocations: 0 bytes)      # before (1.6.0-DEV.220)
  2.988 μs (0 allocations: 0 bytes)      # after (7b0e72efc8f7fbf2c0bd5d30a8b6e2e22288547f)
```

The improvement is much more drastic when mixing it with iterator transforms because it hits the transducer path.

```julia
julia> @btime extrema($(Iterators.flatten(((0, 1, 2), randn(1000)))));
  195.393 μs (5489 allocations: 133.03 KiB)  # before (1.6.0-DEV.220)
  3.469 μs (0 allocations: 0 bytes)          # after (7b0e72efc8f7fbf2c0bd5d30a8b6e2e22288547f)
```

Of course, for arrays, we might want to use the same unrolling trick as in `minimum`/`maximum`.  But that's tracked by #31442 and can be done later.  I think this PR would be a good generic fallback that is useful as a baseline for #31442.

One thing not so great about this PR is that I had to copy the old `iterate`-only `extrema` implementation to compiler.jl as apparently it is required for bootstrap.  I think it might be nice to separate out a very minimal portion of `reduce.jl` so that it can be `include`d from the compiler.  But that's a bit too large refactoring to do in this PR.


base/multidimensional.jl
```diff
@@ -1566,9 +1566,13 @@ _unique_dims(A::AbstractArray, dims::Colon) = invoke(unique, Tuple{Any}, A)
...
 """
-    extrema(A::AbstractArray; dims) -> Array{Tuple}
+    extrema([f,] A::AbstractArray; dims, [init]) -> Array{Tuple}
 
-Compute the minimum and maximum elements of an array over the given dimensions.
+Compute the minimum and maximum of `f` applied to each element (if given) in the given
+dimensions of `A`.
+
+!!! compat "Julia 1.2"
+    2-argument `extrema` method requires Julia 1.2 or later.
```
nalimilan, line 1575:
```suggestion
    The `extrema(f, A)` method requires Julia 1.2 or later.
```
(To avoid confusion about whether keyword arguments are counted or not.)

base/operators.jl
```diff
@@ -443,58 +443,6 @@ julia> minmax('c','b')
...
-        (x, s) = y
-        fx = f(x)
-        vmax = max(fx, vmax)
-        vmin = min(fx, vmin)
-    end
-    return (vmin, vmax)
-end
-
-extrema(x::Real) = (x, x)
-extrema(f, x::Real) = (y = f(x); (y, y))
```
nalimilan, line 496: Aren't these still useful?

base/multidimensional.jl
```diff
@@ -1566,9 +1566,13 @@ _unique_dims(A::AbstractArray, dims::Colon) = invoke(unique, Tuple{Any}, A)
 end
 
 """
-    extrema(A::AbstractArray; dims) -> Array{Tuple}
+    extrema([f,] A::AbstractArray; dims, [init]) -> Array{Tuple}
 
-Compute the minimum and maximum elements of an array over the given dimensions.
+Compute the minimum and maximum of `f` applied to each element (if given) in the given
+dimensions of `A`.
```
nalimilan, line 1572: This sentence is weird when `f` is not given. Maybe this?
```suggestion
 Compute the minimum and maximum elements of `A` over dimensions `dims`.
 If `f` is provided, return the minimum and maximum elements after applying `f` to them.
```

base/reduce.jl
```diff
@@ -762,6 +762,75 @@ Inf
...
+
+"""
+    extrema(f, itr; [init]) -> Tuple
+
+Compute both the minimum and maximum of `f` applied to each element in `itr` and return
+them as a 2-tuple. Only one pass is made over `itr`.
+
+The value returned for empty `itr` can be specified by `init`. It must be a 2-tuple whose
+first/second element is a neutral element for `min`/`max` (i.e. which is greater/less than
+or equal to any other element). This is required because it is unspecified whether `init`
```
nalimilan, line 802:
```suggestion
The value returned for empty `itr` can be specified by `init`. It must be a 2-tuple whose
first and second elements are neutral elements for `min` and `max` respectively
(i.e. which are greater/less than or equal to any other element).
```

base/reduce.jl
```diff
@@ -762,6 +762,75 @@ Inf
...
 minimum(a; kw...) = mapreduce(identity, min, a; kw...)
 
+"""
+    extrema(itr; [init]) -> Tuple
+
+Compute both the minimum and maximum element in a single pass, and return them as a 2-tuple.
+
+The value returned for empty `itr` can be specified by `init`. It must be a 2-tuple whose
+first/second element is a neutral element for `min`/`max` (i.e. which is greater/less than
+or equal to any other element). This is required because it is unspecified whether `init`
```
nalimilan, line 772:
```suggestion
The value returned for empty `itr` can be specified by `init`. It must be a 2-tuple whose
first and second elements are neutral elements for `min` and `max` respectively
(i.e. which are greater/less than or equal to any other element).
```

base/compiler/compiler.jl
```diff
@@ -48,6 +48,24 @@ include("operators.jl")
 include("pointer.jl")
 include("refvalue.jl")
 
+# required for bootstrap
+extrema(itr) = extrema(identity, itr)
+function extrema(f, itr)
```
nalimilan, line 53: Maybe define this as `_extrema_itr` and do `extrema(f, itr) = _extrema_itr(f, itr)`, so that you can test that `_extrema_itr` gives the expected results? Otherwise AFAICT tests won't cover it.

nalimilan (commented):
Just a few superficial comments.

base/operators.jl
tkf: `mapreduce` has a specialization for `Number`:

https://github.com/JuliaLang/julia/blob/5142abf4507525a7a02e7d984ce39802ee4fe134/base/reduce.jl#L419

So, I think the compiler will generate the equivalent machine code.

base/compiler/compiler.jl
tkf: This function is in `Core.Compiler` and not in `Base`. So, this won't be tested in the test suite. We can test `Core.Compiler.extrema`, though.

tkf: I added a few simple tests for `Core.Compiler.extrema` 4a27df97501dc9ee02581480b7076a1ed69e785c

base/reduce.jl
```diff
@@ -762,6 +762,73 @@ Inf
...
+"""
+    extrema(itr; [init]) -> Tuple
+
+Compute both the minimum and maximum element in a single pass, and return them as a 2-tuple.
+
+The value returned for empty `itr` can be specified by `init`. It must be a 2-tuple whose
+first and second elements are neutral elements for `min` and `max` respectively
+(i.e. which are greater/less than or equal to any other element). It is used for non-empty
+collections. Note: it implies that, for empty `itr`, the first element is typically
+_greater_ than the last element. This is a "paradoxical" but yet expected result.
```
timholy, line 774: Not sure this is ideal, but what about something like this?

```suggestion
The value returned for empty `itr` can be specified by `init`. It must be a 2-tuple whose
first and second elements are neutral elements for `min` and `max` respectively
(i.e. which are greater/less than or equal to any other element). As a consequence, when `itr` is empty the
returned `(mn, mx)` tuple will satisfy `mn > mx`. When `init` is specified it is used even for non-empty `itr`.
```

timholy (approved):
Looks great! Needing a special version in compiler isn't so bad, IMO.

base/reduce.jl
tkf: Hmm... We'd need something like `reduce_empty(::typeof(max), ::Type{T}) where {T} = typemin(T)` for "when `itr` is empty the returned `(mn, mx)` tuple will satisfy `mn > mx`" to make sense, right?  (I commented in https://github.com/JuliaLang/julia/issues/36266#issuecomment-644424299 but I'm in favor of _not_ defining this `reduce_empty`.)

But I think naming the returned value is a good idea! So I stole it and updated the docstrings: 360b3463cef1f03681ffdbffbb6cb005f9ac09f4

timholy: I don't think it has anything to do with `reduce_empty`? By supplying `init` you bypass it, right?

Maybe it's easier if I explain the problem. Having "It is used for non-empty collections" come between your specification of the neutral elements and the consequences for the returned value feels like an interruption of thought. I recognize these issues are interleaved (the reason you need certain sentinel values is because they are used for non-empty collections), but this version of the docstring seems unnecessarily confusing.

base/operators.jl
nalimilan: Cool. Maybe worth checking just in case?

simeonschaub:
Bump. Looks like this just needs to be rebased.

Moelf:
bump, hitting `extrema` slower than `(maximum, minimum)` again recently

vtjnash (approved):

base/multidimensional.jl
```diff
@@ -1715,23 +1719,21 @@ julia> extrema(A, dims = (1,2))
...
+_extrema_dims(f::F, A::AbstractArray, ::Colon, ::_InitialValue) where {F} =
+    mapreduce(_DupY(f), _extrema_rf, A)
+# Note: not passing `init = _InitialValue()` since user-defined
+# `reduce`/`foldl` cannot be aware of `Base._InitialValue` that is an
+# internal implementation detail.
+
+_extrema_dims(f::F, A::AbstractArray, dims, init) where {F} =
+    mapreduce(_DupY(f), _extrema_rf, A; dims = dims, init = init)
+function _extrema_dims(f::F, A::AbstractArray, dims, ::_InitialValue) where {F}
+    sz = [size(A)...]
```
simeonschaub, line 1736: Was this change intentional? Seems like this is causing the test failures.

tkf: I'm guessing it shows up in the diff because this PR is behind #39321.

tkf: Ah never mind, I think we need to "merge" this line manually. Thanks for catching this!

```suggestion
    sz = size(A)
```

base/reduce.jl
tkf: Thanks, after re-reading your version again, I think it makes sense now. It's added to the patch 3ceaaae94bb5a26c36a72c5e824f90c1821ca8a4

DilumAluthge:
@tkf The CI failures seem related to the contents of the PR.

tkf:
This passed the tests for SparseArrays locally but I noticed that it's not hitting the right method. Unfortunately, the build process ignores missing name on import and happy define `HigherOrderFns._extrema_itr`

```
Profile  ──────────  0.286842 seconds
WARNING: could not import Base._extrema_itr into HigherOrderFns
SparseArrays  ─────  2.713728 seconds
```

We need some fix in SparseArrays to make it reasonably correct and practically OK. (100% correctness is impossible with the current facility in Base because SparseArrays seem to want to assume `max` and `min` are commutative. We need to revive #36424 for this.)

JeffBezanson (approved):

tkf:
#43604 took care of this

PR #36267, opened 2020-06-13
Fix bug on introspection macros on broadcasting syntax

### LINKED ISSUE #36261: `@code_lowered` etc errors when used on broadcasting function call
With dot-function calls, `@code_lowered` & friends error when the argument is a variable name. e.g. The dot syntax works when some arguments of a multi-argument function is just a number and not an array, such as for the example [here](https://docs.julialang.org/en/v1/manual/functions/#man-vectorized-1),
```
f(x,y) = 3x + 4y
A = [1.0, 2.0, 3.0]
f.(pi, A)
```
where the dot is broadcasting for the 2nd argument of `f`. For this example:
`@code_lowered f.(pi, A)` errors with
```
type Symbol has no field head

Stacktrace:
 [1] getproperty(::Symbol, ::Symbol) at ./Base.jl:33
...
```
(same with `@code_typed` and `@code_llvm`, `@code_native`).

`@code_lowered f.(A,A)` also fails
```
ERROR: LoadError: type Symbol has no field head
Stacktrace:
 [1] getproperty(::Symbol, ::Symbol) at ./Base.jl:33
...
```

however, `@code_lowered f.(rand(3), rand(3))` does not error.

There was https://github.com/JuliaLang/julia/pull/35522 but this issue also occurs on master
```
julia> versioninfo()
Julia Version 1.6.0-DEV.173
Commit 8c65f8d987* (2020-06-05 00:08 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin19.5.0)
  WORD_SIZE: 64
  LIBM: libopenlibm
```

Liozou:
Fix #36261

This was a simple omission from #35522. I added the appropriate tests.

### DIFF
--- a/stdlib/InteractiveUtils/src/macros.jl
+++ b/stdlib/InteractiveUtils/src/macros.jl
@@ -15,7 +15,7 @@ function recursive_dotcalls!(ex, args, i=1)
     if !(ex isa Expr) || ((ex.head !== :. || !(ex.args[2] isa Expr)) &&
                           (ex.head !== :call || string(ex.args[1])[1] != '.'))
         newarg = Symbol('x', i)
-        if ex isa Expr && ex.head === :...
+        if Meta.isexpr(ex, :...)
             push!(args, only(ex.args))
             return Expr(:..., newarg), i+1
         else
--- a/stdlib/InteractiveUtils/src/macros.jl
+++ b/stdlib/InteractiveUtils/src/macros.jl
@@ -15,7 +15,7 @@ function recursive_dotcalls!(ex, args, i=1)
     if !(ex isa Expr) || ((ex.head !== :. || !(ex.args[2] isa Expr)) &&
                           (ex.head !== :call || string(ex.args[1])[1] != '.'))
         newarg = Symbol('x', i)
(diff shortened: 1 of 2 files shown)


stdlib/InteractiveUtils/src/macros.jl
```diff
@@ -15,7 +15,7 @@ function recursive_dotcalls!(ex, args, i=1)
     if !(ex isa Expr) || ((ex.head !== :. || !(ex.args[2] isa Expr)) &&
                           (ex.head !== :call || string(ex.args[1])[1] != '.'))
         newarg = Symbol('x', i)
-        if ex.head === :...
+        if ex isa Expr && ex.head === :...
```
simeonschaub, line 18:
```suggestion
        Meta.isexpr(ex, :...)
```

Liozou: I didn't know about this syntax, thanks!

Liozou: There is a missing `if`, let me fix this

Liozou:
CI failures look unrelated.

PR #36273, opened 2020-06-14
fix #36272: Error with optional argument in anonymous function defined in macro

### LINKED ISSUE #36272: Error with optional argument in anonymous function defined in macro
In Julia 1.5.0 beta 1, the following fails:
```julia
macro m()
    :((a, b=1) -> a*b)
end
@m()(1)
```
with
```
ERROR: UndefVarError: b not defined
Stacktrace:
 [1] (::var"#1#2")(::Int64, ::Int64) at ./REPL[1]:3 (repeats 2 times)
 [2] top-level scope at REPL[1]:3
```

The reason I want to do this is here: https://github.com/JuliaPOMDP/POMDPs.jl/blob/609fe87094d4b1b01993deb0514565450266547f/src/generative.jl#L189. A workaround is to name the function.

simeonschaub:
I hope this is the right location for the test. Is this implementation correct?

Fix #36272

### DIFF
--- a/src/macroexpand.scm
+++ b/src/macroexpand.scm
@@ -210,7 +210,7 @@
         ((atom? v) '())
         (else
          (case (car v)
-           ((... kw |::|) (try-arg-name (cadr v)))
+           ((... kw |::| =) (try-arg-name (cadr v)))
            ((escape) (list v))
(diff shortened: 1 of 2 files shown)


JeffBezanson:
LGTM, thanks!

Issue #36251, opened 2020-06-12
varinfo()'s summary display of tuples of arrays is too long 


briochemc:
Maybe it's my setup, but I tried with v1.4.2 and v1.5.0-beta1.0 with the same output. MWE where I just define a tuple of vectors and then call `varinfo()`:

```julia
julia> tup = (rand(100),) ;

julia> varinfo()
  name                    size summary
[...]

PR #36276, opened 2020-06-14
Fix summary for tuples

briochemc:
Removed `sumary(io: IO, t::tuple)` line as suggested in https://github.com/JuliaLang/julia/issues/36251#issuecomment-643320532.


JeffBezanson:
Thanks. I think there are some tests of this to fix as well.

StefanKarpinski:
The good news is you don't need to add tests, just fix the existing ones. Might be worth a note in NEWS.

Keno (changes requested):
As per previous comments, tests need to be fixed (changing the review status drops it from my queue until comments are addressed).

briochemc:
>  tests need to be fixed

Yes, I'll try to do that early next week, sorry got very little time right now!

briochemc:
I just looked for `summary` in `test/show.jl` and changed those in the last commit (hopefully correctly)... But is that all that was needed?

briochemc:
I *think* I fixed the tests correctly. Please let me know if I need to do anything else!

JeffBezanson:
Looks good, just needs a rebase now.

briochemc:
Again, I *think* I just did the correct thing by merging master back into the PR branch. Do I need to do anything else?

KristofferC:
Looks like some test need fixing up.

test/errorshow.jl
```diff
@@ -291,7 +291,7 @@ let undefvar
     @test occursin("DomainError with [0.0 -1.0 …", err_str)
 
     err_str = @except_str (1, 2, 3)[4] BoundsError
-    @test err_str == "BoundsError: attempt to access (1, 2, 3) at index [4]"
+    @test err_str == "BoundsError: attempt to access Tuple{$Int,$Int,$Int} at index [4]"
```
JeffBezanson, line 294: Please rebase this (which will hopefully make the test failure go away), and add spaces after these commas (and in the other file as well). Hopefully then we can finally merge this. Sorry this has taken so long!

briochemc: @JeffBezanson Are you sure about the spaces after commas? If we are reverting the output of `summary` for tuples to use `typeof`, then the output has no commas, right? E.g., with Julia v1.5.1:
```julia
julia> typeof((1,2,3))
 Tuple{Int64,Int64,Int64} 
```
Or am I missing something?

fredrikekre: https://github.com/JuliaLang/julia/pull/37085

briochemc: OK I was not aware of that 😅 will change it!

briochemc: OK Done! 🙂

briochemc: @JeffBezanson is it OK this time?

briochemc:
(Gentle bump!)

PR #36279, opened 2020-06-14
Promote on Rational binary operations

### LINKED ISSUE #36277: Return type of *(::Rational, ::Integer) changed in 1.5
#35483 introduced a new method for `*(::Rational, ::Integer)` which changed the return type of that operation. There was no specialized method for this before, so promotion was used.

Julia 1.4.2:
```julia
julia> Rational{Int64}(1,1) * UInt8(1) # returns Rational{Int64}
1//1
```

Julia 1.5.0-beta1:
```julia
julia> Rational{Int64}(1,1) * UInt8(1) # returns Rational{UInt64}
0x0000000000000001//0x0000000000000001
```

Should this return type be preserved between minor versions?

Liozou:
Fix #36277

I think this is a bugfix (see issue) that needs backporting to the 1.5 branch.

### DIFF
--- a/test/rational.jl
+++ b/test/rational.jl
@@ -575,3 +575,14 @@ end
     @test -Int32(1) // typemax(Int32) - Int32(1) == typemin(Int32) // typemax(Int32)
     @test 1 // (typemax(Int128) + BigInt(1)) - 2 == (1 + BigInt(2)*typemin(Int128)) // (BigInt(1) + typemax(Int128))
 end
+
+@testset "Promotions on binary operations with Rationals (#36277)" begin
+    @test (-40//3) // 0x5 == 0x5 // (-15//8) == -8//3
+    @test (-4//7) // (0x1//0x3) == (0x4//0x7) // (-1//3) == -12//7
+    @test -3//2 + 0x1//0x1 == -3//2 + 0x1 == 0x1//0x1 + (-3//2) == 0x1 + (-3//2) == -1//2
+    @test 0x3//0x5 - 2//3 == 3//5 - 0x2//0x3 == -1//15
+    @test rem(-12//5, 0x2//0x1) == rem(-12//5, 0x2) == -2//5
+    @test mod(0x3//0x1, -4//7) == mod(0x3, -4//7) == -3//7
+    @test -1//5 * 0x3//0x2 == 0x3//0x2 * -1//5 == -3//10
+    @test -2//3 * 0x1 == 0x1 * -2//3 == -2//3
+end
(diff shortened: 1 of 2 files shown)


sostock:
Note that this restores the pre-1.5 behavior of `*(::Rational, ::Integer)` and `*(::Integer, ::Rational)`, but it *changes* the behavior of `*(x::Rational, y::Rational)`, which didn’t use promotion before.

test/rational.jl
```diff
@@ -575,3 +575,14 @@ end
...
+
+@testset "Promotions on binary operations with Rationals (#36277)" begin
+    @test (-40//3) // 0x5 == 0x5 // (-15//8) == -8//3
+    @test (-4//7) // (0x1//0x3) == (0x4//0x7) // (-1//3) == -12//7
+    @test -3//2 + 0x1//0x1 == -3//2 + 0x1 == 0x1//0x1 + (-3//2) == 0x1 + (-3//2) == -1//2
+    @test 0x3//0x5 - 2//3 == 3//5 - 0x2//0x3 == -1//15
+    @test rem(-12//5, 0x2//0x1) == rem(-12//5, 0x2) == -2//5
+    @test mod(0x3//0x1, -4//7) == mod(0x3, -4//7) == -3//7
+    @test -1//5 * 0x3//0x2 == 0x3//0x2 * -1//5 == -3//10
+    @test -2//3 * 0x1 == 0x1 * -2//3 == -2//3
```
simeonschaub, line 587: This should probably either use `===` for checking equality or check the return type explicitly. Otherwise the return type is still allowed to differ as long as equality by `==` holds.

sostock: I also prepared a fix for this and used the following tests:
```julia
inttypes = [Int8, Int16, Int32, Int64, Int128, BigInt, UInt8, UInt16, UInt32, UInt64, UInt128]
for T in inttypes, S in inttypes
    @test typeof(one(Rational{T}) * one(S)) == typeof(one(S) * one(Rational{T})) == promote_type(Rational{T}, S)
end
```
Note that this doesn’t check `*(::Rational, ::Rational)`, since I did not include that in my fix (I only restored pre-1.5 behavior).

Liozou:
> Note that this restores the pre-1.5 behavior of `*(::Rational, ::Integer)` and `*(::Integer, ::Rational)`, but it _changes_ the behavior of `*(x::Rational, y::Rational)`, which didn’t use promotion before.

That's a fair point. But I think that's a good thing since it makes the return types more consistent. when using binary operations with either Rationals or Integers.

test/rational.jl
Liozou: @simeonschaub You are right, thanks for pointing this out. I added your tests @sostock (and others, including the case of `*(::Rational, ::Rational)`) in the last commit, thanks!

JeffBezanson (approved):

PR #20317, opened 2017-01-29
Improve inferability of unique

p-zubieta:
Fixes #20105 and supersedes #20106

PR #36280, opened 2020-06-14
Improve inference for `unique` with abstract eltypes

timholy:
#20317 improved inference of `unique`, but problematic cases still arise for containers with known but abstract eltypes. Here, we short-circuit the `typejoin` when the return type is determined by the element type of the input container.

For `unique(f, itr)`, this commit also allows the caller to supply `seen::Set` to circumvent the inference challenges.

This resolves a number of invalidations in `Pkg` and `REPL`.

~TODO: update NEWS entry with issue number, once assigned.~

### DIFF
--- a/base/set.jl
+++ b/base/set.jl
@@ -120,19 +120,25 @@ julia> unique(Real[1, 1.0, 2])
 ```
 """
 function unique(itr)
-    T = @default_eltype(itr)
-    out = Vector{T}()
-    seen = Set{T}()
-    y = iterate(itr)
-    y === nothing && return out
-    x, i = y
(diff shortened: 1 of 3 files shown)


base/set.jl
```diff
@@ -120,19 +120,25 @@ julia> unique(Real[1, 1.0, 2])
...
+        for x in itr
+            if !in(x, seen)
+                push!(seen, x)
+                push!(out, x)
+            end
+        end
+        return out
+    end
     y = iterate(itr)
     y === nothing && return out
```
JeffBezanson, line 136: Is `out` undefined here?

Issue #36005, opened 2020-05-23
Second "using" is slow (slower than third, can even be slower than the first)


PallHaraldsson:
Even on my recent:

```
$ ~/julia-1.6.0-DEV-8f512f3f6d/bin/julia -O3

julia> @time using Plots
 10.669931 seconds (11.02 M allocations: 655.843 MiB, 3.90% gc time)

julia> @time using Plots  # for some other packages if I recall I've seen up to 1+ sec.
  0.611072 seconds (935.63 k allocations: 45.940 MiB, 2.12% gc time)

julia> @time using Plots
  0.000317 seconds (230 allocations: 14.219 KiB)
```

Second and all subsequent `using` should all be as fast (not only about Plots.jl). You could ask yourself, why do them, but could well happen in real-world code, and anyway this seems like a low-hanging fruit that should be fixed.

Note, this doesn't happen with all packages, so first I thought it would be a problem with those packages, but I'm not sure. At least none of those I've informed knew of this or have any idea why it happens.

I'm not complaining--here--about the first slow `using` (a whole separate issue), nor the the even worse (i.e. first one above is not that):

```
julia> @time using Plots
[ Info: Precompiling Plots [91a5bcdd-55d7-5caf-9e0b-520d859cae80]
122.433167 seconds (11.63 M allocations: 684.720 MiB, 0.18% gc time)
```

https://github.com/JuliaPlots/Plots.jl/issues/2590

KristofferC:
This is because loading eg Plots invalidates some of the code loading code itself.

PallHaraldsson:
Ok, good to know, even better would be to know what hat means and how to fix. Is this documented?

timholy:
Great question. There's lots of attention to this issue right now, and after a few issues are fixed in Julia itself there will be a blog post: https://github.com/JuliaLang/www.julialang.org/pull/794.

PallHaraldsson:
I'm not sure it's invalidations as Kristoffer states. Why does it happen (I guess your blog explains, if that's the fact), or more importantly why doesn't it happen for 3rd and all subsequent?

I took a look at the very long (that's not bad, I want to read it eventually) blog post as you can see from my review, but I admit I did so far only scan it (and neither read most comments to it).

[...]

PR #36282, opened 2020-06-14
Fix more invalidations from overloading `==`

timholy:
Other than a PR to Pkg (which needs #36280), this is pretty much everything that I think can be done on #36005 without changes to the compiler itself. The combination of #36255, #36280, #36281, the upcoming Pkg PR, some PRs to JuliaInterpreter/LoweredCodeUtils/Revise, and this PR is to reduce the total count of (non-unique) invalidations from more than 3k to 529. Of those, 281 are from `==(::Symbol, ::Any)`, which I do not see a reasonable way to address without changes to the compiler itself (some details were mentioned in #36255).

### DIFF
--- a/base/cartesian.jl
+++ b/base/cartesian.jl
@@ -356,7 +356,7 @@ function exprresolve_arith(ex::Expr)
     if ex.head === :call
         callee = ex.args[1]
         if isa(callee, Symbol)
-            if haskey(exprresolve_arith_dict, callee) && all([isa(ex.args[i], Number) for i = 2:length(ex.args)])
+            if haskey(exprresolve_arith_dict, callee) && all(Bool[isa(ex.args[i], Number) for i = 2:length(ex.args)])
                 return true, exprresolve_arith_dict[callee](ex.args[2:end]...)
             end
         end
--- a/base/cartesian.jl
+++ b/base/cartesian.jl
@@ -353,8 +353,13 @@ const exprresolve_cond_dict = Dict{Symbol,Function}(:(==) => ==,
     :(<) => <, :(>) => >, :(<=) => <=, :(>=) => >=)
 
 function exprresolve_arith(ex::Expr)
-    if ex.head === :call && haskey(exprresolve_arith_dict, ex.args[1]) && all([isa(ex.args[i], Number) for i = 2:length(ex.args)])
-        return true, exprresolve_arith_dict[ex.args[1]](ex.args[2:end]...)
+    if ex.head === :call
+        callee = ex.args[1]
+        if isa(callee, Symbol)
+            if haskey(exprresolve_arith_dict, callee) && all([isa(ex.args[i], Number) for i = 2:length(ex.args)])
+                return true, exprresolve_arith_dict[callee](ex.args[2:end]...)
+            end
+        end
     end
     false, 0
 end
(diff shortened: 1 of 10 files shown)


timholy:
BTW, we are making progress but there's still a lot of work ahead. Numbers in the left two columns are where we were in May (around the time of the 1.5 feature freeze) when I was actively working on the blog post, numbers in the right two columns are where we are now on this PR and all the ones referenced above:

| Package | more specific (May) | ambiguous (May) | more specific (now) | ambiguous (now) |
|:------- | ------------------:| --------:| -----:| -----:|
| Example | 0 | 0  | 0  | 0  |
| Revise | 7 |  0 | 0 | 0 |
| FixedPointNumbers | 170 | 387 | 34 | 470 |
| SIMD | 3903 | 187 | 5  | 874 |
| StaticArrays | 989 | 3133 | 593 | 2060 |
| Optim | 1643 | 2921 | 760 | 2103 |
| Images | 1749 | 3671 | 610 | 2515 |
| Flux | 1991 | 3460 | not loadable (CUDAnative) |  |
| Plots | 1542 | 4302 | 1268 | 3028 |
| DataFrames | 4919 | 783 | 3520 | 1833 |
| JuMP | 2145 | 4670 | not loadable (MutableArithmetics) |  |
| Makie | 6233 | 5526 | 710 | 4081 |
| DifferentialEquations | 5152 | 6218 | 3787 | 3432 |

Obviously, the single biggest advance will come when the ambiguous cases go away, but #36200 and all these PRs to improve inferrability are starting to have a pretty big cumulative effect. @SimonDanisch and @eschnett should be particularly pleased.

I'm not quite sure why the ambiguity numbers have changed so much, but either it's changes in the packages (I wasn't careful to insist on the same versions) or the changes in inference have changed the dependency trees in a way that affects the ambiguity cases, too. (That actually makes sense.)

base/cartesian.jl
```diff
@@ -353,17 +353,27 @@ const exprresolve_cond_dict = Dict{Symbol,Function}(:(==) => ==,
     :(<) => <, :(>) => >, :(<=) => <=, :(>=) => >=)
 
 function exprresolve_arith(ex::Expr)
-    if ex.head === :call && haskey(exprresolve_arith_dict, ex.args[1]) && all([isa(ex.args[i], Number) for i = 2:length(ex.args)])
-        return true, exprresolve_arith_dict[ex.args[1]](ex.args[2:end]...)
+    if ex.head === :call
+        callee = ex.args[1]
+        if isa(callee, Symbol)
+            if haskey(exprresolve_arith_dict, callee) && all([isa(ex.args[i], Number) for i = 2:length(ex.args)])
```
JeffBezanson, line 359:
```suggestion
            if haskey(exprresolve_arith_dict, callee) && all(Bool[isa(ex.args[i], Number) for i = 2:length(ex.args)])
```

stdlib/REPL/src/REPLCompletions.jl
```diff
@@ -626,7 +628,7 @@ function completions(string, pos, context_module=Main)::Completions
         if isa(ex, Expr)
             if ex.head==:call
```
JeffBezanson, line 629:
```suggestion
            if ex.head === :call
```

stdlib/REPL/src/REPLCompletions.jl
```diff
@@ -626,7 +628,7 @@ function completions(string, pos, context_module=Main)::Completions
...
                 return complete_methods(ex, context_module), first(frange):method_name_end, false
-            elseif ex.head==:. && ex.args[2] isa Expr && ex.args[2].head==:tuple
+            elseif ex.head==:. && ex.args[2] isa Expr && (ex.args[2]::Expr).head==:tuple
```
JeffBezanson, line 631:
```suggestion
            elseif ex.head === :. && ex.args[2] isa Expr && (ex.args[2]::Expr).head === :tuple
```

timholy:
Wow, that `downloads` test fails a lot...

PR #36292, opened 2020-06-15
Add keyword argument constant propogation to News

ChrisRackauckas:
I think https://github.com/JuliaLang/julia/pull/35976 is a pretty big deal since it makes all of SciML suddenly infer a lot better, and probably another big library that's splatting keyword arguments. So I added a note to the compiler/runtime improvements.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -17,6 +17,7 @@ Compiler/Runtime improvements
 * All platforms can now use `@executable_path` within `jl_load_dynamic_library()`.
   This allows executable-relative paths to be embedded within executables on all
   platforms, not just MacOS, which the syntax is borrowed from. ([#35627])
+* Constant propogation now occurs through keyword arguments ([#35976])
 
 Command-line option changes
 ---------------------------


mateuszbaran:
Isn't this wording regarding that change a bit too optimistic? When I read that I was happy to see many places in StaticArrays that require `Val` to be type-stable (like dims keyword arguments or full in SVD) to become much nicer but I just checked and it's not the case. Even after forcing inlining I get things like
```julia
julia> using StaticArrays

julia> using LinearAlgebra

julia> A = @SMatrix rand(3,4)
3×4 SArray{Tuple{3,4},Float64,2,12} with indices SOneTo(3)×SOneTo(4):
 0.532864  0.414811  0.430558  0.482411
 0.758695  0.324281  0.946536  0.643921
 0.877494  0.386115  0.168173  0.733471

julia> @code_warntype svd(A, full=true)
Variables
  #unused#::Core.Compiler.Const(LinearAlgebra.var"#svd##kw"(), false)
  @_2::NamedTuple{(:full,),Tuple{Bool}}
  @_3::Core.Compiler.Const(LinearAlgebra.svd, false)
  A::SArray{Tuple{3,4},Float64,2,12}
  full::Bool
  @_6::Bool

Body::Union{StaticArrays.SVD{Float64,SArray{Tuple{3,3},Float64,2,9},SArray{Tuple{3},Float64,1,3},SArray{Tuple{3,4},Float64,2,12}}, StaticArrays.SVD{Float64,SArray{Tuple{3,3},Float64,2,9},SArray{Tuple{3},Float64,1,3},SArray{Tuple{4,4},Float64,2,16}}}
1 ─ %1  = Base.haskey(@_2, :full)::Core.Compiler.Const(true, false)
│         %1
│         (@_6 = Base.getindex(@_2, :full))
└──       goto #3
2 ─       Core.Compiler.Const(:(@_6 = StaticArrays.Val(false)), false)
3 ┄       (full = @_6)
│   %7  = (:full,)::Core.Compiler.Const((:full,), false)
│   %8  = Core.apply_type(Core.NamedTuple, %7)::Core.Compiler.Const(NamedTuple{(:full,),T} where T<:Tuple, false)
│   %9  = Base.structdiff(@_2, %8)::Core.Compiler.Const(NamedTuple(), false)
│   %10 = Base.pairs(%9)::Core.Compiler.Const(Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}(), false)
│   %11 = Base.isempty(%10)::Core.Compiler.Const(true, false)
│         %11
└──       goto #5
4 ─       Core.Compiler.Const(:(Base.kwerr(@_2, @_3, A)), false)
5 ┄ %15 = StaticArrays.:(var"#svd#627")(full, @_3, A)::Union{StaticArrays.SVD{Float64,SArray{Tuple{3,3},Float64,2,9},SArray{Tuple{3},Float64,1,3},SArray{Tuple{3,4},Float64,2,12}}, StaticArrays.SVD{Float64,SArray{Tuple{3,3},Float64,2,9},SArray{Tuple{3},Float64,1,3},SArray{Tuple{4,4},Float64,2,16}}}
└──       return %15

```
Or should I open an issue about this problem with an MWE?

ChrisRackauckas:
Yes, please open an issue about this with an MWE. I would've thought this would be fixed as well.

mateuszbaran:
Hmm... that particular example with `svd` results in proper constant propagation when the call is in a function. It isn't propagated when I add a second, non-constant, keyword argument. I don't know if it's supposed to work, my mental model of constant propagation is lacking even for non-keyword arguments.

PR #36301, opened 2020-06-16
FileWatching: cleanup timers on return

vtjnash:
Following discussion and my own advice at #36217, this avoids a minor temporary resource leak when using the `timeout` parameter. Can't really be tested as there's no visible side effects to this.

### DIFF
--- a/stdlib/FileWatching/src/FileWatching.jl
+++ b/stdlib/FileWatching/src/FileWatching.jl
@@ -649,10 +649,13 @@ giving the result of the polling.
 function poll_fd(s::Union{RawFD, Sys.iswindows() ? WindowsRawSocket : Union{}}, timeout_s::Real=-1; readable=false, writable=false)
     wt = Condition()
     fdw = _FDWatcher(s, readable, writable)
+    local timer
     try
         if timeout_s >= 0
             result::FDEvent = FDEvent()
-            @async (sleep(timeout_s); notify(wt))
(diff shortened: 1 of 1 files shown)


stdlib/FileWatching/src/FileWatching.jl
```diff
@@ -649,10 +649,13 @@ giving the result of the polling.
 function poll_fd(s::Union{RawFD, Sys.iswindows() ? WindowsRawSocket : Union{}}, timeout_s::Real=-1; readable=false, writable=false)
     wt = Condition()
     fdw = _FDWatcher(s, readable, writable)
+    local timer
```
tkf, line 652: Just curious, but why not `timer = nothing` and later `timer === nothing || close(timer)`?  Is it some kind of optimization or just a coding style?

tkf (commented):
Just looking at the diff, but yeah, using `Timer` is a nice solution for this :+1:

stdlib/FileWatching/src/FileWatching.jl
vtjnash: True, those would be essentially same. So just equivalent variations based on how I started writing it.

staticfloat (approved):

vtjnash:
There's perhaps a chance this broke something (https://build.julialang.org/#/builders/79/builds/863/steps/5/logs/stdio), though I don't see how, but let's keep an eye on that.

PR #36303, opened 2020-06-16
show: differentiate TypeName and wrapper when printing

vtjnash:
Currently we print a TypeName identically to the wrapper object Type it is pointed from. This prints them slightly differently, to help distinguish which one you have.

### DIFF
--- a/base/show.jl
+++ b/base/show.jl
@@ -499,7 +499,7 @@ function show(io::IO, @nospecialize(x::Type))
         show_datatype(io, x)
         return
     elseif x isa Union
-        if x.a isa DataType && Core.Compiler.typename(x.a) === Core.Compiler.typename(DenseArray)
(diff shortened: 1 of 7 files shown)


StefanKarpinski:
Would be good to have a test somewhere for the `typename(...)` printing output, if for no other reason than to demonstrate what this changed.

base/show.jl
```diff
@@ -647,7 +648,9 @@ macro show(exs...)
 end
 
 function show(io::IO, tn::Core.TypeName)
+    print(io, "typename(")
     show_type_name(io, tn)
+    print(io, ")")
```
JeffBezanson, line 653: Maybe something like `TypeName Array`, so it doesn't look like syntax that might actually work?

vtjnash: We actually do have a `typename` function that works this way

JeffBezanson (approved):

Issue #36266, opened 2020-06-13
(Soft-)deprecate `init` for `sum!` etc?


tkf:
As @timholy pointed out in #35839, the meaning of `init` in in-place multi-dimensional reducers like `sum!` and `count!` are incompatible with `reduce` etc.

Now that `sum(...; init)` #36188 is merged and would be shipped with Julia 1.6, it's a bit concerning that `init` in `sum!(r, A; init = false)` and `sum(A; init = false)` means something completely different.  I think it might make sense to deprecate or "soft-deprecate" `init` for `sum!` etc.  By "soft-deprecate" I mean to add a new keyword argument that has `init` as the fallback without `depwarn`.  This new keyword argument would be documented as a preferred way to control the initialization of the destination array.  Since `sum!` etc. do not actually document `init`, going directly to deprecation might be OK.  But, since this keyword argument exists for a long time, I'm inclined to treat this as a fully-fledged public API.  (This is a bit tricky case because it's unclear if performance-breaking but semantics-preserving change should be treated as non-breaking.)

For the actual name of the new keyword argument, how about `sum!(r, A; fill = true)` instead of `sum!(r, A; init = true)`?

tkf:
Looking at it again, now I'm not so sure about `fill` as the new keyword argument. Brainstorming some other names:

```
fillinit
initialize
initfirst
needinit
```

Alternatively, it also makes sense to rename `init` of `reduce` etc. to `identity`, `id`, `neutral`, or something.  We need something similar to `init` for `foldl`/`foldr`/`accumulate`, though.

(cc'ing other people who might be interested in this: @nalimilan @mbauman @JeffBezanson)

mcabbott:
How about `sum!(r, A; keep = false)` or `sum!(r, A; discard = true)`?

tkf:
I think `discard` is a good candidate! I guess `keep` is kind of a same idea but `keep`ing some value happens only if the corresponding region of the input array is zero (or `NaN`/`missing` is in the destination). I think `discard` is better because `discard`ing old values happen always.

timholy:
I also like `discard`, great suggestion @mcabbott and thanks @tkf for spearheading this and so many other advances!

StefanKarpinski:
Doesn't the keyword argument to `sum!` control whether the array gets filled first or not? This keyword isn't documented anywhere that I can see and even figuring out what it does requires perusing some fairly confusing code in `reducedim.jl`

tkf:
[...]

PR #36305, opened 2020-06-16
Remove `init` from `count!` docstring

tkf:
To prepare for renaming keyword argument `init` for in-place multi-dimensional reducers #36266, it'd be better to avoid increasing the user of `init` in Julia 1.5 phase.  So, maybe it makes sense to "downgrade" the status of `init` for `count!` to the same level as the one for `sum!` etc.


nalimilan (approved):

timholy (approved):

KristofferC:
CI errors is download failures.

Issue #35495, opened 2020-04-15
Some improvements for the PDF of the manual


ViralBShah:
- [x] Remove release notes from the PDF (we start with 25 pages of release notes - mostly blank) (https://github.com/JuliaLang/julia/pull/36312)
- [x] The spacing for chapter numbers in the TOC is not sufficient (https://github.com/JuliaDocs/Documenter.jl/pull/1785)

Nice to have: 
- [ ] Have a mechanism for a custom cover page (https://github.com/JuliaDocs/Documenter.jl/pull/1788)

Edit: I updated the title of this issue to cover various issues in the PDF of the manual.

Basically have to go 50 pages to get to the "Getting started" part of the manual, which is chapter 14.

ViralBShah:
I suspect this happens because in the web version, the next page from the landing page is the release notes. However, in the online docs, one probably uses the toc on the left for navigation.

mortenpi:
The reason is that the [`NEWS.md` page, while "hidden", is actually the same level as the Home/Manual/Base/etc. chapters](https://github.com/JuliaLang/julia/blob/cc6e121386758dff6ba7911770e48dfd59520199/doc/make.jl#L53), and so it becomes a separate `\chapter`. It might be possible to make it a lower-level page in the PDF or something along those lines.

ViralBShah:
I guess showing release notes to someone that early on in even in the web version doesn't seem like a good idea. Perhaps we can include it as a link rather than as a separate page? The level of information in the NEWS file as we currently ship it is significantly more technical than the rest of the manual.

ViralBShah:
Actually, is it possible to remove the NEWS altogether from the PDF? It is good the way it is in the web version, but doesn't add much value in the pdf.

ViralBShah:
The TOC has way too much detail. Would be nice to present fewer levels of information in the Table of Contents.

ViralBShah:
Spacing issues in TOC:  

![Screen Shot 2020-04-27 at 6 14 57 PM](https://user-images.githubusercontent.com/744411/80426085-043f4400-88b3-11ea-9b7a-b9cbf5add40b.png)

ViralBShah:
How does one go about fixing the latex generated that finally builds this PDF?

ViralBShah:
Self-answer: https://github.com/JuliaDocs/Documenter.jl/blob/master/src/Writers/LaTeXWriter.jl

inkydragon:
[...]

PR #36312, opened 2020-06-16
Remove NEWS.md from the PDF of the manual

ViralBShah:
This removes the NEWS.md link at the bottom of the first page of the
manual. Given that we link to it prominently on top of the page, I
feel this is ok.

This will remove Section II of the PDF manual, where the release notes
take 15 pages (because ofthe way markdown gets converted to PDF).

Partial fix for #35495


KristofferC:
Won't this break the link to the release notes in the html version=

ViralBShah:
My understanding is that it won't generate the link to the release notes at the bottom of the page, but we still have the link to release notes on the top.

ViralBShah:
Verified. With this commit, the link at the bottom goes to "Getting Started" instead of "NEWS".

ViralBShah:
I'll merge this tomorrow unless there are objections.

ViralBShah:
How often does the 1.6-Dev documentation on docs.julialang.org get generated/updated? The pdf there has a date of May 8.

KristofferC:
This removed the NEWS section from the manual when you click release notes:

image

Pretty much what I meant with https://github.com/JuliaLang/julia/pull/36312#issuecomment-644874314.

ViralBShah:
Please explain what you found misleading in the PR title. What more does it do than what it said?

I didn't realize the NEWS showed up in the menu - that's not how I understood your comment.  That UI is quite strange - but anyways, I can add that back.

KristofferC:
It seems like it changes some headings in the PDF now as well.

ViralBShah:
If something is wrong, let me know and I'll fix it in the PR I make to reinstate the release notes in the html version.

KristofferC:
The pdf on the latest docs (if it has updated) looks like:

image

I don't know if it looked like that before?

ViralBShah:
Where are you looking for the PDF on the latest docs?

KristofferC:
Did it look fine on your local version?

ViralBShah:
I think it did. But I am building again to see what happens.

ViralBShah:
Before, there were 20 pages of Release notes, because of the way Markdown gets converted into latex. This removes the release notes from the PDF version, and merges the Introduction into the manual.

ViralBShah:
I think you have an old PDF. This is what it looks locally.
Screen Shot 2020-06-17 at 4 42 34 PM

KristofferC:
Yeah, I just got the one from the website which I guess hasn't updated.

ViralBShah:
Right - it has the date of May 8, 2020. Something seems to have broken somewhere a while ago.

ViralBShah:
The reason you see all those sections empty is because the NEWS file was probably empty after we branched.

ViralBShah:
I wonder if `article` would be better than the `memoir` latex documentclass which is what Documenter is using.

ViralBShah:
Follow-up PR https://github.com/JuliaLang/julia/pull/36333

PR #36314, opened 2020-06-16
Fix: lastindex for Numbers

### LINKED ISSUE #36311: 1[end, end] errors: no multiarg lastindex for Numbers
while silly, we should be constistent
```julia
julia> 1[1,1]
1

julia> 1[end]
1

julia> [1][end, end]
1

julia> 1[end, end]
ERROR: MethodError: no method matching lastindex(::Int64, ::Int64)
```

mgautam98:
Fixes: #36311 
Added a method to `lastindex` to handel 1[end,end] for consistency.
Also added similar function to `firstindex`

### DIFF
--- a/base/number.jl
+++ b/base/number.jl
@@ -68,7 +68,9 @@ ndims(x::Number) = 0
 ndims(::Type{<:Number}) = 0
 length(x::Number) = 1
 firstindex(x::Number) = 1
+firstindex(x::Number, ::Int) = 1
 lastindex(x::Number) = 1
+lastindex(x::Number, ::Int) = 1
 IteratorSize(::Type{<:Number}) = HasShape{0}()
 keys(::Number) = OneTo(1)
 
--- a/base/number.jl
+++ b/base/number.jl
@@ -68,9 +68,9 @@ ndims(x::Number) = 0
 ndims(::Type{<:Number}) = 0
 length(x::Number) = 1
 firstindex(x::Number) = 1
-firstindex(x::Number, ::Int) = 1
+firstindex(x::Number, d::Int) = d < 1 ? throw(BoundsError()) : 1
 lastindex(x::Number) = 1
-lastindex(x::Number, ::Int) = 1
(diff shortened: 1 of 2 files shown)


base/number.jl
```diff
@@ -68,7 +68,9 @@ ndims(x::Number) = 0
 ndims(::Type{<:Number}) = 0
 length(x::Number) = 1
 firstindex(x::Number) = 1
+firstindex(x::Number, ::Int) = 1
```
simeonschaub, line 71: Since `size` and `axes` do this check as well, this should probably be:
```suggestion
firstindex(x::Number, d::Int) = d < 1 ? throw(BoundsError()) : 1
```

base/number.jl
```diff
@@ -68,7 +68,9 @@ ndims(x::Number) = 0
...
 lastindex(x::Number) = 1
+lastindex(x::Number, ::Int) = 1
```
simeonschaub, line 73: Here as well.

mgautam98: Okay

mgautam98: I made the changes. When I was testing it was throwing `BoundsError` so I didn't thought of it

simeonschaub:
This still needs tests, probably somewhere here: https://github.com/JuliaLang/julia/blob/6cdfcf9b48c0c4d4cd4022539ec7e5083df9e8e3/test/numbers.jl#L2403

mgautam98:
@simeonschaub Okay I will include test

simeonschaub (approved):
If you want, you could also include @oxinabox's `1[end, end]
` example in the tests, but it's not that important.

mbauman (approved):
Thanks!

oxinabox (approved):

mgautam98:
@mbauman will it me merged?

mbauman:
Yes, closing and re-opening is a little trick to easily re-run CI.  A website we hit to test HTTP connectivity went down, which meant that all the tests here failed.  I'm quite certain that this will pass tests, but it's better to confidently merge with 18 ✅s from CI.

Issue #265, opened 2011-11-26
automatic recompilation of dependent functions


vtjnash:
if I define a function d2 before a function d1 which calls d2 then change d2, d1 uses the old definition for d2.
I assume this is because it is all precompiled, but maybe there should be a note warning of this? Or would it be possible to replace the old definition with a longjmp to the new one?
(Mostly important for the REPL, since I don't always do a full load)

``` julia
julia> function d2()
       a
       end

julia> function d()
         d2()
       end

julia> d()
in d: a not defined

julia> function d2()
       b=2
       end

julia> d()
in d: a not defined

julia> d2
Methods for generic function d2
d2() at prompt:2

julia> function d()
         d2()
       end

julia> d()
2
```

yuyichao:
I was a little bit surprized to know this problem, not because I think it is easy to handle this, but because the following works as expected.

```julia
julia> f(a, b) = a + b
f (generic function with 1 method)

julia> g(args...) = f(args...)
g (generic function with 1 method)

julia> g(1, 2)
3

julia> f(a::Int, b::Int) = a - b
f (generic function with 2 methods)

julia> g(1, 2)
-1
```

yuyichao:
Actually, the function returns the right result but the `@code_typed` returns wrong result.....

```julia
julia> f(a, b) = a + b
f (generic function with 1 method)

julia> g(args...) = f(args...)
g (generic function with 1 method)

julia> g(1, 2)
3

julia> @code_typed g(1, 2)
1-element Array{Any,1}:
 :($(Expr(:lambda, Any[:(args::(top(apply_type))(Vararg,Any)::Any::Any)], Any[Any[],Any[Any[:args,(Int64,Int64),0]],Any[]], :(begin  # none, line 1:
        return (top(box))(Int64,(top(add_int))((top(tupleref))(args::(Int64,Int64),1)::Int64,(top(tupleref))(args::(Int64,Int64),2)::Int64))
    end::Int64))))

julia> f(a::Int, b::Int) = a - b
f (generic function with 2 methods)

julia> g(1, 2)
-1

julia> @code_typed g(1, 2)
1-element Array{Any,1}:
 :($(Expr(:lambda, Any[:(args::(top(apply_type))(Vararg,Any)::Any::Any)], Any[Any[],Any[Any[:args,(Int64,Int64),0]],Any[]], :(begin  # none, line 1:
[...]

PR #36317, opened 2020-06-17
Add a few hooks that make external AbstractInterpreters easier

Keno:
Details, in the individual commit messages. I don't think these are necessarily the correct abstractions longterm, but they are quite useful for prototyping in the short term. With these hooks (and prior work on refactoring the caches), it's possible to have a completely external Julia compiler, complete with its own implementation of apply_generic that will use the correct interpreter for its analysis, while still running on the Julia execution engine.

### DIFF
--- a/base/compiler/inferencestate.jl
+++ b/base/compiler/inferencestate.jl
@@ -46,6 +46,10 @@ mutable struct InferenceState
     # `max_valid`, to be used in inlining
     matching_methods_cache::IdDict{Any, Tuple{Any, UInt, UInt}}
 
+    # The interpreter that created this inference state. Not looked at by
+    # NativeInterpreter. But other interpreters may use this to detect cycles
+    interp::AbstractInterpreter
+
     # src is assumed to be a newly-allocated CodeInfo, that can be modified in-place to contain intermediate results
     function InferenceState(result::InferenceResult, src::CodeInfo,
                             cached::Bool, interp::AbstractInterpreter)
@@ -107,7 +111,8 @@ mutable struct InferenceState
             Vector{InferenceState}(), # callers_in_cycle
             #=parent=#nothing,
             cached, false, false, false,
-            IdDict{Any, Tuple{Any, UInt, UInt}}())
+            IdDict{Any, Tuple{Any, UInt, UInt}}(),
+            interp)
         result.result = frame
         cached && push!(get_inference_cache(interp), result)
         return frame
--- a/src/julia.h
+++ b/src/julia.h
@@ -2100,6 +2100,10 @@ typedef struct {
 
     // Cache access. Default: jl_rettype_inferred.
     jl_codeinstance_lookup_t lookup;
+
+    // If not `nothing`, rewrite all generic calls to call
+    // generic_context(f, args...) instead of f(args...).
+    jl_value_t *generic_context;
 } jl_cgparams_t;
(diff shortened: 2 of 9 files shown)


vtjnash:
I'd been intentionally avoiding this (meaning `_invoke_codeinst`, and created the `:invoke` expr head instead) as it seems likely to be semantically wrong with-respect-to #265, and therefore the existence of it may violate the assumptions required for doing incremental compilation, and so it could possibly cause other issues (such as committing us to using a specific internal cache structure that we don't want). 

Defining this to take a `Method` should be a better match for Julia's regular semantics (e.g. internally it would be the same as the `invoke` function), though a MethodInstance probably also could be made to work feasibly.

base/compiler/inferencestate.jl
```diff
@@ -46,6 +46,10 @@ mutable struct InferenceState
     # `max_valid`, to be used in inlining
     matching_methods_cache::IdDict{Any, Tuple{Any, UInt, UInt}}
 
+    # The interpreter that created this inference state. Not looked at by
+    # NativeInterpreter. But other interpreters may use this to detect cycles
+    interp::AbstractInterpreter
+
```
vtjnash, line 52: This seems like a slight design smell that this other state is getting mixed into into our state here. Why isn't this part of the `InferenceParams` configuration values?

Keno: Well, it's not really for configuration, it's so that the interpreter can detect that this InferenceState was not created by it and handle it appropriately.

Keno:
I don't want it to take a Method, because I really just want a low-level way to say, "no, invoke this CodeInstance specifically" for debugging and experimenting purposes. I agree it's a bad high-level API, but it's useful for testing. If you think a builtin is too high-level, we could maybe just have it be something you can ccall?

src/julia.h
```diff
@@ -2090,6 +2090,10 @@ typedef struct {
 
     // Cache access. Default: jl_rettype_inferred.
     jl_codeinstance_lookup_t lookup;
+
+    // If non-null, rewrite all generic calls to call
+    // generic_context(f, args...) instead of f(args...).
+    jl_value_t *generic_context;
```
JeffBezanson, line 2096: Needs a default value in `jl_default_cgparams`?

Keno: Yes. Also I realize the comment is outdated, since I settled on `nothing` rather than null to indicate absence.

vtjnash:
These appear to be unrelated features in each commit—can you make them separate PRs?

Keno:
That's the point of having them each be a separate commit, so they can be reviewed commit by commit. It didn't seem worth making 4 separate PRs

Keno:
This whole PR is 65 lines. Are 4 15 line PRs really better, taking 4x the CI time? If the issue is that you like some of the features, but not others?

vtjnash:
Yeah, it seems like we generally shouldn't be complicating review for the sake of a CI bot. I don't mind if a PR starts this way (I do that frequently enough myself, especially when one piece of work reveals another), but if you want part of it faster, then it's good to separate that out. Otherwise, it's not the reviewer's fault if the whole thing gets blocked on one piece. And I know I might get yelled at for interleaving thoughts and trying to discuss parts of this while they're entangled with the other parts, so I'd appreciate not needing to try to interleave replies in one PR thread (since GitHub is not particularly great UI for having unrelated discussions in one PR).

Keno:
Split out into #36398 through #36401.

PR #36318, opened 2020-06-17
use ReturnNode, GotoIfNot, and Argument more consistently in IR

JeffBezanson:
These are more efficient representations, and I would like to make the optimizer IR and codegen/inference IR more similar, reducing the amount of conversion needed. Seems to make the compiler a hair faster too.

- Always uses ReturnNode and GotoIfNot within IR (AST/macros unaffected)
- Inference still uses Slots but codegen can understand Argument, removing a conversion step

I also plan to replace `:unreachable` Exprs with `ReturnNode()`. Do we still like the ReturnNode-with-undefined-field approach, or should we add a singleton type for this?

### DIFF
--- a/base/compiler/optimize.jl
+++ b/base/compiler/optimize.jl
@@ -339,12 +339,6 @@ function statement_cost(ex::Expr, line::Int, src::CodeInfo, sptypes::Vector{Any}
         # prevent inlining.
         extyp = line == -1 ? Any : src.ssavaluetypes[line]
         return extyp === Union{} ? 0 : 20
-    elseif head === :return
-        a = ex.args[1]
-        if a isa Expr
-            return statement_cost(a, -1, src, sptypes, slottypes, params)
-        end
-        return 0
     elseif head === :(=)
         if ex.args[1] isa GlobalRef
             cost = 20
@@ -364,12 +358,6 @@ function statement_cost(ex::Expr, line::Int, src::CodeInfo, sptypes::Vector{Any}
         # since these aren't usually performance-sensitive functions,
         # and llvm is more likely to miscompile them when these functions get large
         return typemax(Int)
-    elseif head === :gotoifnot
-        target = ex.args[2]::Int
-        # loops are generally always expensive
-        # but assume that forward jumps are already counted for from
-        # summing the cost of the not-taken branch
-        return target < line ? 40 : 0
     end
     return 0
 end
@@ -386,6 +374,8 @@ function inline_worthy(body::Array{Any,1}, src::CodeInfo, sptypes::Vector{Any}, 
             # but assume that forward jumps are already counted for from
             # summing the cost of the not-taken branch
             thiscost = stmt.label < line ? 40 : 0
+        elseif stmt isa GotoIfNot
+            thiscost = stmt.dest < line ? 40 : 0
         else
             continue
         end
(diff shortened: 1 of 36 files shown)


timholy:
Would be awesome to add `ReturnNode` and `Argument` to https://docs.julialang.org/en/v1.6-dev/devdocs/ast/, and to delete mention of `:gotoifnot`.

base/compiler/optimize.jl
```diff
@@ -386,6 +374,14 @@ function inline_worthy(body::Array{Any,1}, src::CodeInfo, sptypes::Vector{Any},
             # but assume that forward jumps are already counted for from
             # summing the cost of the not-taken branch
             thiscost = stmt.label < line ? 40 : 0
+        elseif stmt isa GotoIfNot
+            thiscost = stmt.dest < line ? 40 : 0
+        elseif stmt isa ReturnNode
+            if stmt.val isa Expr
+                thiscost = statement_cost(stmt.val, -1, src, sptypes, slottypes, params)::Int
+            else
```
vtjnash, line 382: Shouldn't this be unreachable now (malformed IR)?

JeffBezanson: I guess so. There can still be some `Expr`s there, but I guess they all have 0 cost.

src/codegen.cpp
```diff
@@ -5016,7 +5034,8 @@ static std::pair<std::unique_ptr<Module>, jl_llvm_functions_t>
     jl_codectx_t ctx(jl_LLVMContext, params);
     JL_GC_PUSH2(&ctx.code, &ctx.roots);
     ctx.code = src->code;
-
+    //if (jl_is_method(lam->def.method) && !strcmp(jl_symbol_name(lam->def.method->name),"promote_rule"))
+    //    jl_(src->code);
```
vtjnash, line 5038:
```suggestion

```

vtjnash (approved):

src/codegen.cpp
JeffBezanson: oops

Keno (approved):

Keno:
CI has some complaints though (beyond the issues currently plaguing us on master).

Keno:
While we're here, I think we may also want to give `:static_parameter` it's own node type as well as disallowing it in anything other than statement position, since it can throw. I'd also kind of like to split out `GlobalRef` to constant and non-constant bindings and disallow the latter in anything other than statement position also. I think if we did that, then there could only ever be one exception side-effect per statement, which i think would be desirable.

Keno:
That can be a separate PR though.

Keno:
Actually, now that I think about it, we may want to get rid of `:static_parameter` entirely and just have static parameters assigned to slots. I think the reason they aren't currently is because static parameters used to be special inside ccall, but I think we got rid of that behavior?

JeffBezanson:
I'm not sure we want to further overload slots with static parameters. Arguments, static parameters, and local variables get their values from different places, effectively --- I mean, I believe you were the one who added `Argument`, and I can see why :)

JeffBezanson:
Also what do you think about removing `:unreachable`?

Keno:
> I'm not sure we want to further overload slots with static parameters.

I was imagining more of an implicit (or maybe even explicit):
```
Slot(:T) = StaticParameter(:T)
```
at the top of the function (except being an assignment that propagates undefined-ness). I don't really like that you can't re-assign static parameters in the function body. Just seems weird.

Keno:
> Also what do you think about removing `:unreachable`?

`ReturnNode` with an undef field for that value? That's how IRCode represents it.

Keno (approved):

PR #36323, opened 2020-06-17
Improve inferrability of stream IO

timholy:
A lot of IO subtypes wrap other IO objects, and the field type declaration is typically abstract (e.g., `::IO` or `::LibuvStream`). This leads to widespread inference problems, because it's not possible for inference to know what `io.io.status` will return, even if it knows that `io.io::LibuvStream`.

To fix this, this PR defines a formal interface for three abstract types (`LibuvStream`, `LibuvServer`, and `AbstractPipe`) by specializing `getproperty` for them and adding name-specific type assertions. Since these types are unexported, it should not be considered a breaking change to specify that certain fields must have certain types.

This change makes most functions in `base/stream.jl` pretty well-inferred, which is a substantial change from where we are now.

### DIFF
--- a/base/io.jl
+++ b/base/io.jl
@@ -333,8 +333,27 @@ function open(f::Function, args...; kwargs...)
     end
 end
 
-# Generic wrappers around other IO objects
+"""
+    AbstractPipe
+
+`AbstractPipe` is the abstract supertype for IO pipes that provide for communication between processes.
+
+If `pipe isa AbstractPipe`, it must obey the following interface:
+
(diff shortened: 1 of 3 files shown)


timholy:
If there are no objections, I might merge this soon. Among my invalidation PRs, this isn't one of the most consequential in terms of latency, but I think it has a certain elegance: it provides an example of how one might non-obtrusively define interfaces for abstract types.

JeffBezanson (approved):

Issue #36321, opened 2020-06-17
Unexpected 1d/2d array method error message with unsupported keyword argument


tkf:
I couldn't find any issue tracking this:

```julia
julia> identity([1 1]; unsupported = :keyword)
ERROR: MethodError: no method matching identity(::Array{Int64,2}; unsupported=:keyword)

You might have used a 2d row vector where a 1d column vector was required.
Note the difference between 1d column vector [1,2,3] and 2d row vector [1 2 3].
You can convert to a column vector with the vec() function.
Closest candidates are:
  identity(::Any) at operators.jl:511 got unsupported keyword argument "unsupported"
```

This seems to happen only with keyword arguments:

```julia
julia> identity([1 1], nothing)
ERROR: MethodError: no method matching identity(::Array{Int64,2}, ::Nothing)
Closest candidates are:
  identity(::Any) at operators.jl:511
```

Moelf:
this fix may be problematic if some it was a 1d/2d method error && keywords are presented?

tkf:
Wow, that's a super quick fix :) Yeah, agree about that the user can remove the unsupported keyword argument.

PR #36325, opened 2020-06-17
fix 1d/2d array method error message with unsupported keyword arg

Moelf:
based on #36321

### DIFF
--- a/base/errorshow.jl
+++ b/base/errorshow.jl
@@ -316,7 +316,7 @@ function showerror(io::IO, ex::MethodError)
             hasrows |= isrow
             push!(vec_args, isrow ? vec(arg) : arg)
         end
-        if hasrows && applicable(f, vec_args...)
+        if hasrows && applicable(f, vec_args...) && isempty(kwargs)
             print(io, "\n\nYou might have used a 2d row vector where a 1d column vector was required.",
                       "\nNote the difference between 1d column vector [1,2,3] and 2d row vector [1 2 3].",
                       "\nYou can convert to a column vector with the vec() function.")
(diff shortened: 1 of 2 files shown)


base/errorshow.jl
```diff
@@ -315,7 +315,7 @@ function showerror(io::IO, ex::MethodError)
             hasrows |= isrow
             push!(vec_args, isrow ? vec(arg) : arg)
         end
-        if hasrows && applicable(f, vec_args...)
+        if hasrows && applicable(f, vec_args...) && isempty(kwargs)
```
mbauman, line 318: It'd be awesome if we could do:

```julia
        if hasrows && applicable(f, vec_args...; kwargs...)
```

but looks like applicable doesn't understand kwargs.  This seems like a good enough stop-gap.

Moelf:
bump

StefanKarpinski:
Who would be good to review this?

mbauman:
This is a decent band-aid, IMO.  It'd be easier to do something better with the just-filed #36611 or by transforming the usage here to call `hasmethod` instead, but I think the current workaround is good enough (even if not ideal)

It'd be nice to add a few simple tests, though.

Moelf:
tests added ;)

Moelf:
not sure why doctest failed; anything else needed?

Moelf:
bump for 1.6 train

mbauman (approved):

test/errorshow.jl
```diff
@@ -750,3 +750,12 @@ end
     @test contains(bt_str, "@ $m.B.C")
     @test contains(bt_str, "@ $m.B.D")
 end
+# 1d/2d error shouldn't appear in unsupported keywords arg #36325
+let err = nothing
+    try
+        identity([1,1]; bad_kwards = :julia)
```
JeffBezanson, line 756:
```suggestion
        identity([1 1]; bad_kwards = :julia)
```

PR #36330, opened 2020-06-17
macOS lazily saves the AVX512 context

### LINKED ISSUE #36329: LLVM-9.0.1 doesn't recognize AVX-512 in macOS
See https://github.com/SciML/DifferentialEquations.jl/issues/623#issuecomment-644739630

@maleadt mentioned that https://github.com/llvm/llvm-project/commit/82921bf2baed96b700f90b090d5dc2530223d9c0 might fix it. I wonder if we can patch Julia's LLVM to make it generate more efficient code for new Apple hardware.

YingboMa:
Fix #36329

### DIFF
--- a/src/processor_x86.cpp
+++ b/src/processor_x86.cpp
@@ -545,7 +545,14 @@ static NOINLINE std::pair<uint32_t,FeatureList<feature_sz>> _get_host_cpu(void)
     unset_bits(features, 32 + 27);
     if (!hasavx)
         features_disable_avx(features);
+#ifdef _OS_DARWIN_
+    // See https://github.com/llvm/llvm-project/commit/82921bf2baed96b700f90b090d5dc2530223d9c0
+    // and https://github.com/apple/darwin-xnu/blob/a449c6a3b8014d9406c2ddbdc81795da24aa7443/osfmk/i386/fpu.c#L174
+    // Darwin lazily saves the AVX512 context on first use
+    bool hasavx512save = hasavx;
+#else
     bool hasavx512save = hasavx && test_all_bits(xcr0, 0xe0);
+#endif
     if (!hasavx512save)
         features_disable_avx512(features);
     // Ignore feature bits that we are not interested in.


src/processor_x86.cpp
```diff
@@ -545,7 +545,13 @@ static NOINLINE std::pair<uint32_t,FeatureList<feature_sz>> _get_host_cpu(void)
     unset_bits(features, 32 + 27);
     if (!hasavx)
         features_disable_avx(features);
+#ifdef OS_DARWIN
```
yuyichao, line 548: `_OS_DARWIN_` i believe.

src/processor_x86.cpp
```diff
@@ -545,7 +545,13 @@ static NOINLINE std::pair<uint32_t,FeatureList<feature_sz>> _get_host_cpu(void)
...
+    // See https://github.com/llvm/llvm-project/commit/82921bf2baed96b700f90b090d5dc2530223d9c0
+    // Darwin lazily saves the AVX512 context on first use
```
yuyichao, line 550: also https://github.com/apple/darwin-xnu/blob/a449c6a3b8014d9406c2ddbdc81795da24aa7443/osfmk/i386/fpu.c#L174 or whatever appropriate direct document from apple.

yuyichao (changes requested):

yuyichao (approved):
LGTM. Do note that since it seems that loop vectorizer might be involved there'll still be other issues that you'll need to hack around until it is moved to the LLVM level.

vchuravy:
We should also back-port the llvm patch.

Keno:
This seems like a terrible idea on their part, but I guess it's what we have to do.

yuyichao:
IIUC it sounds like they are reimplementing `xsave`. It's pretty unfortunate that they deviate from the intel document...

I assume it also means that running a darwin vm that doesn't support avx512 on hardware that does is basically unsupported then... (which is probably something that apple couldn't care less about...)

Keno:
I doubt this is the last we've heard along these lines, but let's try this for now, since it matches LLVM.

YingboMa:
Is it possible to back port this change?

PR #36332, opened 2020-06-17
Introduce keyword argument `reset` for `sum!` etc.

### LINKED ISSUE #36266: (Soft-)deprecate `init` for `sum!` etc?

tkf:
See #36266 for the discussion. The name of the argument is still discussed in #36266 (`reset` is a placeholder ATM).

close #36266

### DIFF
--- a/base/reducedim.jl
+++ b/base/reducedim.jl
@@ -537,7 +537,7 @@ julia> prod!([0 0], A)
 1×2 Array{Int64,2}:
  3  8
 
-julia> prod!([0 0], A; reset = true)
+julia> prod!([0 0], A; reset = false)
 1×2 Array{Int64,2}:
  0  0
 ```
@@ -663,7 +663,7 @@ julia> minimum!([0 1], A; reset = false)  # initial values "win"
 1×2 Array{Int64,2}:
  0  1
 
-julia> minimum!([4, 4], A; reset = false)
+julia> minimum!([4 4], A; reset = false)
 1×2 Array{Int64,2}:
  1  2
 ```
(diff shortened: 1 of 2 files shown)


base/reducedim.jl
```diff
@@ -850,15 +976,44 @@ end
...
-                  init::Bool=true)
-    findminmax!(isgreater, init && !isempty(A) ? fill!(rval, first(A)) : rval, fill!(rind,zero(eltype(keys(A)))), A)
+                  kw...)
+    if haskey(kw, :init)  # backward compatibility
+        _kw_reset(kw) && !isempty(A) && fill!(rval, first(A))
+        fill!(rind, zero(eltype(keys(A))))
+    elseif _kw_reset(kw)
+        !isempty(A) && fill!(rval, first(A))
+        fill!(rind, zero(eltype(keys(A))))
+    end
```
tkf, line 1015: The behavior `init=false` was not consistent with other multi-dimensional reducers.  So I suggest to fix it.

It was added in https://github.com/JuliaLang/julia/pull/6894 by @timholy but it looks like it was an internal function back then. So, probably the consistency with other reducers and the usability for end-users were not the biggest concern.

Also, I'm not sure if we should keep the backward compatibility branch. But it's rather cheap so why not?

base/reducedim.jl
```diff
@@ -732,15 +805,39 @@ all(a::AbstractArray; dims=:)              = _all(a, dims)
...
+    _kw_reset((init = bool,)) -> bool
+    _kw_reset((reset = bool,)) -> bool
+
+Extract a keyword argument `reset` from the keyword arguments.  Throw an error
+if unsupported.  For backward compatibility, allow `init` as an alias.
+"""
+_kw_reset(kw::Iterators.Pairs) = _kw_reset(kw.data)
+_kw_reset(::NamedTuple{(), Tuple{}}) = true
+_kw_reset(kw::NamedTuple{(:reset,), Tuple{Bool}}) = kw.reset
+_kw_reset(kw::NamedTuple{(:init,), Tuple{Bool}}) = kw.init
```
tkf, line 818: Should we `depwarn` when `init` is passed?

DilumAluthge:

PR #36333, opened 2020-06-17
Add Release Notes back to the html version of the documentation

ViralBShah:
Adding them back to the PDF version as well, but now they are at the end rather than in the beginning.


ViralBShah:
Note that the very first link in the document in the Introduction section, which links to the Release notes, is a broken link. It was broken before and is still broken. Not sure how to fix it. The easiest way is perhaps to make it a web URL.

ViralBShah:
cc @KristofferC

PR #36336, opened 2020-06-17
Switch `httpbin` tests over to JuliaLang-hosted `httpbin` mock server

staticfloat:
This should fix the recent testing issues we've had with `httpbin`.

### DIFF
--- a/test/download.jl
+++ b/test/download.jl
@@ -1,6 +1,6 @@
 # This file is a part of Julia. License is MIT: https://julialang.org/license
 
-let urls = ["http://httpbin.org/ip", "https://httpbin.org/ip"]
+let urls = ["http://httpbin.julialang.org/ip", "https://httpbin.julialang.org/ip"]
     for url in urls
         @test Base.download_url(url) == url
     end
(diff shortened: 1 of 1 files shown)


test/download.jl
```diff
@@ -1,6 +1,6 @@
 # This file is a part of Julia. License is MIT: https://julialang.org/license
 
-let urls = ["http://httpbin.org/ip", "https://httpbin.org/ip"]
+let urls = ["http://httpbin.julialang.org/ip", "https://httpbin.julialang.org/ip"]
```
Keno, line 3: Looks like we only get `https` here.

Keno:
```suggestion
let urls = ["http://julialang.org", "https://httpbin.julialang.org/ip"]
```

Keno: Also needs to be adjusted below though.

staticfloat: I'm not sure what you're saying here; this chunk is only testing that `Base.download_url()` is affected by `Base.DOWNLOAD_HOOKS`; nothing is actually downloaded.

Keno: ah, my bad.

staticfloat: I've added a comment to make this more explicit.

Keno (approved):
Should add the API definition in contrib or something, in case we get hit by two badly timed buses and somebody else needs to re-deploy this, but LGTM otherwise.

Issue #36335, opened 2020-06-17
Performance of intersect(::Set, ::Set) depends on order of arguments


CameronBieganek:
```julia
julia> using BenchmarkTools, Random

julia> s = Set(rand(1:10_000, 1000));

julia> t = Set(1000:9000);

julia> @btime intersect($s, $t);
  187.572 μs (19 allocations: 49.59 KiB)

julia> @btime intersect($t, $s);
  25.782 μs (19 allocations: 49.59 KiB)
```

According to the [Python wiki](https://wiki.python.org/moin/TimeComplexity#set), the python intersection method has time complexity
`O(min(len(s), len(t))`. This could be achieved in Julia with something like the following, where the sizes of the sets are tested:

```julia
julia> function myintersect(s::AbstractSet, t::AbstractSet)
           if length(s) > length(t)
               intersect(s, t)
           else
               intersect(t, s)
           end
       end
myintersect (generic function with 1 method)

julia> @btime myintersect($s, $t);
  25.655 μs (19 allocations: 49.59 KiB)

julia> @btime myintersect($t, $s);
  25.998 μs (19 allocations: 49.59 KiB)
```

laborg:
The implementation of https://github.com/JuliaLang/julia/pull/36339 has been merged:
```julia
# on Julia master
julia> using BenchmarkTools, Random

julia> s = Set(rand(1:10_000, 1000));

julia> t = Set(1000:9000);

julia> @btime intersect($s, $t);
  23.570 μs (17 allocations: 49.14 KiB)

julia> @btime intersect($t, $s);
  23.543 μs (17 allocations: 49.14 KiB)
```

PR #36339, opened 2020-06-18
change intersect order to reduce time

Moelf:
based on #36335

Before:
```
julia> @btime intersect($t, $s);
  42.603 μs (17 allocations: 49.56 KiB)

julia> @btime intersect($s, $t);
  317.603 μs (17 allocations: 49.56 KiB)
```

After:
```
julia> @btime intersect($t, $s);
  53.400 μs (17 allocations: 49.56 KiB)

julia> @btime intersect($s, $t);
  53.687 μs (17 allocations: 49.56 KiB)
```

### DIFF
--- a/base/abstractset.jl
+++ b/base/abstractset.jl
@@ -169,9 +169,9 @@ end
 intersect(s) = union(s)
 function intersect(s::AbstractSet, itr)
     if haslength(itr) && hasfastin(itr) && length(s) < length(itr)
-        return mapfilter(_in(itr), push!, s, emptymutable(s))
+        return mapfilter(∈(itr), push!, s, emptymutable(s))
     else
-        return mapfilter(_in(s), push!, itr, emptymutable(s))
+        return mapfilter(∈(s), push!, itr, emptymutable(s))
     end
 end
 
--- a/base/abstractset.jl
+++ b/base/abstractset.jl
@@ -149,7 +149,8 @@ function intersect(s::AbstractSet, itr, itrs...)
     # determine if swap order is useful and viable
     if length(s)>50 && haslength(itr) && all(map(haslength, itrs))
         itrs_lengths = map(length, itrs)
-        min_length = minimum(itrs_lengths)
+        min_idx = argmin(itrs_lengths)
+        min_length = itrs_lengths[min_idx]
         # do nothing if s is longer than the rest or itr is already the shortest
         if length(itr) > min_length >= length(s)
             min_idx = argmin(itrs_lengths)
--- a/base/abstractset.jl
+++ b/base/abstractset.jl
@@ -153,7 +153,7 @@ function intersect(s::AbstractSet, itr, itrs...)
         min_length = itrs_lengths[min_idx]
         # do nothing if s is longer than the rest or itr is already the shortest
         if length(itr) > min_length >= length(s)
-            new_itrs = (itrs[1:min_idx-1]..., itr, itrs[min_idx+1:end]...)
+            new_itrs = setindex(itrs, itr, min_idx)
             return intersect(s, itrs[min_idx], new_itrs...)
         end
     end
--- a/base/abstractset.jl
+++ b/base/abstractset.jl
@@ -11,36 +11,30 @@ copy!(dst::AbstractSet, src::AbstractSet) = union!(empty!(dst), src)
     union(s, itrs...)
     ∪(s, itrs...)
 
-Construct an object containing all distinct elements from all of the arguments.
+Construct the union of sets. Maintain order with arrays.
 
-The first argument controls what kind of container is returned.
-If this is an array, it maintains the order in which elements first appear.
-
-Unicode `∪` can be typed by writing `\\cup` then pressing tab in the Julia REPL, and in many editors.
-This is an infix operator, allowing `s ∪ itr`.
-
-See also [`unique`](@ref), [`intersect`](@ref), [`isdisjoint`](@ref), [`vcat`](@ref), [`Iterators.flatten`](@ref).
+See also: [`intersect`](@ref), [`isdisjoint`](@ref), [`vcat`](@ref), [`Iterators.flatten`](@ref).
 
 # Examples
 ```jldoctest
-julia> union([1, 2], [3])
-3-element Vector{Int64}:
+julia> union([1, 2], [3, 4])
+4-element Vector{Int64}:
  1
  2
  3
+ 4
 
-julia> union([4 2 3 4 4], 1:3, 3.0)
-4-element Vector{Float64}:
- 4.0
- 2.0
- 3.0
- 1.0
+julia> union([1, 2], [2, 4])
+3-element Vector{Int64}:
+ 1
+ 2
+ 4
 
-julia> (0, 0.0) ∪ (-0.0, NaN)
-3-element Vector{Real}:
-   0
-  -0.0
- NaN
+julia> union([4, 2], 1:2)
+3-element Vector{Int64}:
+ 4
+ 2
+ 1
 
 julia> union(Set([1, 2]), 2:3)
 Set{Int64} with 3 elements:
@@ -59,14 +53,14 @@ const ∪ = union
 """
     union!(s::Union{AbstractSet,AbstractVector}, itrs...)
 
-Construct the [`union`](@ref) of passed in sets and overwrite `s` with the result.
+Construct the union of passed in sets and overwrite `s` with the result.
 Maintain order with arrays.
 
 # Examples
 ```jldoctest
-julia> a = Set([3, 4, 5]);
+julia> a = Set([1, 3, 4, 5]);
 
-julia> union!(a, 1:2:7);
+julia> union!(a, 1:2:8);
 
 julia> a
 Set{Int64} with 5 elements:
@@ -108,15 +102,10 @@ end
     intersect(s, itrs...)
     ∩(s, itrs...)
 
-Construct the set containing those elements which appear in all of the arguments.
+Construct the intersection of sets.
+Maintain order with arrays.
 
-The first argument controls what kind of container is returned.
-If this is an array, it maintains the order in which elements first appear.
-
-Unicode `∩` can be typed by writing `\\cap` then pressing tab in the Julia REPL, and in many editors.
-This is an infix operator, allowing `s ∩ itr`.
-
-See also [`setdiff`](@ref), [`isdisjoint`](@ref), [`issubset`](@ref Base.issubset), [`issetequal`](@ref).
+See also: [`setdiff`](@ref), [`isdisjoint`](@ref), [`issubset`](@ref Base.issubset), [`issetequal`](@ref).
 
 !!! compat "Julia 1.8"
     As of Julia 1.8 intersect returns a result with the eltype of the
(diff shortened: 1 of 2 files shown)


Moelf:
a little awkward considering a `intersect(s::AbstractSet, t::AbstractSet)` would be more a specific method thus causing issues.

base/abstractset.jl
```diff
@@ -119,7 +119,7 @@ Set{Int64} with 1 element:
 """
 intersect(s::AbstractSet, itr, itrs...) = intersect!(intersect(s, itr), itrs...)
 intersect(s) = union(s)
-intersect(s::AbstractSet, itr) = mapfilter(_in(s), push!, itr, emptymutable(s))
+intersect(s::AbstractSet, itr) = length(s) > length(itr) ? mapfilter(_in(s), push!, itr, emptymutable(s)) : mapfilter(_in(itr), push!, s, emptymutable(itr))
```
tkf, line 122: Thanks for the patch!  Unfortunately, this has a few problems:

* `itr` may not define `length`. We need to support generic iterables.  You can use `haslength(itr)` to check if you can call `length(itr)`.
* `emptymutable(itr)` may not return a set. We should always use `emptymutable(s)`.

tkf: Also, we need to check `hasfastin(itr)` if we can use `_in(itr)`.

Finally, please limit the length of a line to 92 characters (as mentioned in CONTRIBUTING.md)

Moelf: does it look better now? (force pushed

base/abstractset.jl
```diff
@@ -119,7 +119,15 @@ Set{Int64} with 1 element:
...
 intersect(s) = union(s)
-intersect(s::AbstractSet, itr) = mapfilter(_in(s), push!, itr, emptymutable(s))
+function intersect(s::AbstractSet, itr) 
+    if haslength(itr) && hasfastin(itr)
+        if length(s) > length(itr) 
+            return mapfilter(_in(s), push!, itr, emptymutable(s)) 
+        else
+            return mapfilter(_in(itr), push!, s, emptymutable(s))
+        end
+    end
```
tkf, line 129: You need something like

```suggestion
    if hasfastin(itr) && haslength(itr) && length(s) < length(itr) 
        return mapfilter(_in(itr), push!, s, emptymutable(s))
    else
        return mapfilter(_in(s), push!, itr, emptymutable(s)) 
    end
```

tkf: Also please make sure both branches are tested.

CameronBieganek:
By the way, it looks like the `_in` function defined on line 46 of abstractset.jl is just the curried version of `in`, which now exists. So it seems like `_in(itr)` could be replaced with `in(itr)`, unless I'm missing something...

Also, it appears that `intersect!` suffers from the same problem as `intersect`:

```julia
julia> @benchmark intersect!(s, t) setup=(s = Set(rand(1:10_000, 1000))) setup=(t = Set(1000:9000)) evals=1
BenchmarkTools.Trial: 
  memory estimate:  32 bytes
  allocs estimate:  2
  --------------
  minimum time:     11.558 μs (0.00% GC)
  median time:      12.407 μs (0.00% GC)
  mean time:        13.226 μs (0.00% GC)
  maximum time:     80.413 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1

julia> @benchmark intersect!(t, s) setup=(s = Set(rand(1:10_000, 1000))) setup=(t = Set(1000:9000)) evals=1
BenchmarkTools.Trial: 
  memory estimate:  32 bytes
  allocs estimate:  2
  --------------
  minimum time:     219.543 μs (0.00% GC)
  median time:      222.558 μs (0.00% GC)
  mean time:        230.665 μs (1.53% GC)
  maximum time:     3.513 ms (93.35% GC)
  --------------
  samples:          9632
  evals/sample:     1
```

So it would be nice if we could also fix that in this PR. :)

tkf:
You can't flip the order of `intersect!` because the first argument has to be mutated.  Not very sure if we can do something with `intersect!(s, t)`.  It might be possible to get some speedup from `intersect!(a, b, c)` by flipping `b` and `c` though.

base/abstractset.jl
Moelf: sorry for the delay, it should be fine now also added tests for it

Moelf:
im not sure how to deal with `intersect(s::AbstractSet, itr, itrs...)`, the problem is that `itrs` is a tuple and to swap place of `itr` and the shortest element in `itrs` seems expensive?

Moelf:
ok I tried to add short cut for `intersect(a,b,c...)` as well plz review

```
julia> s1 = Set(rand(1:100_000, 10000));

julia> s2 = Set(rand(1:100_000, 30000));

julia> s3 = Set(rand(1:100_000, 50000));

julia> t = Set(10000:90000);

julia> @btime intersect($t, $s1, $s2, $s3);
  1.199 ms (29 allocations: 193.98 KiB)

julia> @btime intersect($t, $s3, $s2, $s1);
  4.596 ms (34 allocations: 770.28 KiB)
```

after:
```
julia> @btime intersect($t, $s1, $s2, $s3);
  1.112 ms (22 allocations: 193.86 KiB)

julia> @btime intersect($t, $s3, $s2, $s1);
  1.112 ms (31 allocations: 194.02 KiB)
```

laborg:
This might make `intersect` slower for small sets:
1.4.2:
```
julia> a = [Set([7, 9, 8]),
           Set([7, 4, 2, 8, 6]),
           Set([9, 10, 2, 5, 8, 6, 1]),
           Set([7, 4, 9, 2, 8, 1]),
           Set([7, 9, 4, 10, 2, 8, 5, 6]),
           Set([7, 4, 10, 2, 8, 6, 1]),
           Set([9, 4, 3, 5, 8, 6, 1]),
           Set([4, 2, 5, 7, 9, 10, 8, 1]),
           Set([4, 2, 3, 7, 9, 10, 8, 6, 1])];

julia> b = reverse(a);

julia> @btime intersect($a...)
  750.129 ns (22 allocations: 816 bytes)
Set{Int64} with 1 element:
  8

julia> @btime intersect($b...)
  865.911 ns (22 allocations: 816 bytes)
Set{Int64} with 1 element:
  8
```
This PR:
```
julia> @btime intersect($a...)
  1.216 μs (27 allocations: 944 bytes)
Set{Int64} with 1 element:
  8

julia> @btime intersect($b...)

  2.220 μs (33 allocations: 1.25 KiB)
Set{Int64} with 1 element:
  8
```

Moelf:
yea this is inevitable, what can be done is add a ` length(s) > 100 &&`; the thing is as soon as the the leading `Set` has a small size, the process would be very fast, so only really need to swap `itr, itrs` if `s` is long .

Moelf:
after rebase the tests are all passed now

CameronBieganek:
Just curious, what's the status of this PR?

StefanKarpinski:
Appears to be in good shape but needs review.

oscardssmith:
@StefanKarpinski why does this need the triage label?

StefanKarpinski:
Seemed like a good way to get some collective code review on it.

base/abstractset.jl
```diff
@@ -117,9 +117,28 @@ Set{Int64} with 1 element:
...
-intersect(s::AbstractSet, itr, itrs...) = intersect!(intersect(s, itr), itrs...)
+function intersect(s::AbstractSet, itr, itrs...)
+    # determine if swap order is useful and viable
+    if length(s)>50 && haslength(itr) && all(haslength.(itrs))
+        itrs_lengths = length.(itrs)
+        min_length = minimum(itrs_lengths)
+        # do nothing if s is longer than the rest or itr is already the shortest
+        if length(itr) > min_length >= length(s)
+            min_idx = argmin(itrs_lengths)
+            new_itrs = (itrs[1:min_idx-1]..., itr, itrs[min_idx+1:end]...)
```
vtjnash, line 128:
```suggestion
            new_itrs = setindex(itrs, itr, min_idx)
```

base/abstractset.jl
```diff
@@ -117,9 +117,28 @@ Set{Int64} with 1 element:
   2
 ```
 """
-intersect(s::AbstractSet, itr, itrs...) = intersect!(intersect(s, itr), itrs...)
+function intersect(s::AbstractSet, itr, itrs...)
+    # determine if swap order is useful and viable
+    if length(s)>50 && haslength(itr) && all(haslength.(itrs))
```
vtjnash, line 122:
```suggestion
    if length(s)>50 && haslength(itr) && all(map(haslength, itrs))
```

base/abstractset.jl
```diff
@@ -117,9 +117,28 @@ Set{Int64} with 1 element:
...
+        itrs_lengths = length.(itrs)
```
vtjnash, line 123:
```suggestion
        itrs_lengths = map(length, itrs)
```

base/abstractset.jl
```diff
@@ -117,9 +117,28 @@ Set{Int64} with 1 element:
...
+        min_length = minimum(itrs_lengths)
```
vtjnash, line 124:
```suggestion
        min_idx = argmin(itrs_lengths)
        min_length = itrs_lengths[min_idx]
```

base/abstractset.jl
```diff
@@ -117,9 +117,28 @@ Set{Int64} with 1 element:
...
+        # do nothing if s is longer than the rest or itr is already the shortest
+        if length(itr) > min_length >= length(s)
+            min_idx = argmin(itrs_lengths)
```
vtjnash, line 127:
```suggestion
```

vtjnash:
I can't see a reason that triage is necessary either. I've made a few small comments. If you could test that and rebase, we can merge soon.

base/abstractset.jl
Moelf: `setindex!`？

oscardssmith:
Is this good to merge?

base/abstractset.jl
```diff
@@ -117,9 +117,28 @@ Set{Int64} with 1 element:
   2
 ```
 """
-intersect(s::AbstractSet, itr, itrs...) = intersect!(intersect(s, itr), itrs...)
+function intersect(s::AbstractSet, itr, itrs...)
+    # determine if swap order is useful and viable
+    if length(s)>50 && haslength(itr) && all(map(haslength, itrs))
```
JeffBezanson, line 122:
```suggestion
    if length(s)>50 && haslength(itr) && all(haslength, itrs)
```

base/abstractset.jl
```diff
@@ -117,9 +117,28 @@ Set{Int64} with 1 element:
...
+        itrs_lengths = map(length, itrs)
+        min_idx = argmin(itrs_lengths)
+        min_length = itrs_lengths[min_idx]
+        # do nothing if s is longer than the rest or itr is already the shortest
+        if length(itr) > min_length >= length(s)
```
JeffBezanson, line 127: Why is this looking at `length(s)`? Why not always intersect `s` with the shortest itr?

JeffBezanson:
Needs a small change & rebase, then we can merge.

Moelf:
oops, my hacky edit has resulted in weird diff but it should be fine with respect to master branch and both test cases (two-element `interset` and multi-element `intersect(s, itr, itrs...)` work now):
```

julia> @btime intersect($t, $s);
  14.217 μs (17 allocations: 49.14 KiB)

julia> @btime intersect($s, $t);
  14.828 μs (17 allocations: 49.14 KiB)

julia> @btime intersect($t, $s1, $s2, $s3);
  534.275 μs (22 allocations: 193.34 KiB)

julia> @btime intersect($t, $s3, $s2, $s1);
  520.268 μs (22 allocations: 193.34 KiB)
```

PR #36341, opened 2020-06-18
fix #36340: Ctrl+Q stopped working for stack traces etc.

### LINKED ISSUE #36340: Ctrl+Q stopped working for stack traces etc.
After #35556, Ctrl+Q works only if the method table is shown via `display(d::REPLDisplay, mime::MIME"text/plain", x)`.  So, for example, Ctrl+Q still works with `methods(Base._in)` but not with `?Base._in`.  Ctrl+Q also does not work with stacktraces anymore.

I still think #35556 is a good approach because it makes `show` on method table thread-safe.  So, I'd suggest to keep #35556 and somehow find a way to use the display system for the cases where Ctrl+Q does not work.  But I'm also OK with reverting #35556 if Ctrl+Q on, e.g., stacktraces is essential for people and the fix is hard.

cc @simeonschaub

simeonschaub:
Works locally for me, but this should really have some tests. I would really appreciate any hints on how to test this kind of behavior in the REPL, as I didn't have much success previously.
cc @tkf 
closes #36340

### DIFF
--- a/stdlib/REPL/src/REPL.jl
+++ b/stdlib/REPL/src/REPL.jl
@@ -234,7 +234,20 @@ display(d::REPLDisplay, x) = display(d, MIME("text/plain"), x)
 function print_response(repl::AbstractREPL, @nospecialize(response), show_value::Bool, have_color::Bool)
     repl.waserror = response[2]
     io = IOContext(outstream(repl), :module => Main)
+
+    infos = Tuple{String,Int}[]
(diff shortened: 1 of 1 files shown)


simeonschaub:
Ah, wait. This still doesn't work in help mode, for example.

simeonschaub:
Closing in favor of #36344.

PR #36309, opened 2020-06-16
Update to GMP 6.2.0

fingolfin:
This fixes a few bugs and brings performance improvements; for the full
release notes, see <https://gmplib.org/gmp6.2>.

Two patches were removed as they are no longer necessary.

I guess merging this should be coordinated with https://github.com/JuliaPackaging/Yggdrasil/pull/1159

PR #36343, opened 2020-06-18
Update to GMP 6.2.0

fingolfin:
This fixes a few bugs and brings performance improvements; for the full release notes, see <https://gmplib.org/gmp6.2>.

Removed `deps/patches/gmp-config-ldflags.patch` as it is no longer needed.

Adjusted `deps/patches/gmp-exception.patch` (in principle it could go, but then on Windows division by zero of GMP integers would not lead to a `DivideError` but instead a generic `ErrorException`)

This is <s>the same as</s> an improved version of #36309 which was reverted in #36328 as it was failing the pre-merge CI tests on Win32 and Win64. Moreover, there were other issues (see https://github.com/JuliaPackaging/Yggdrasil/pull/1159 and https://github.com/JuliaRegistries/General/pull/16517), so naturally this PR here shouldn't be merged hastily. But I wanted to submit it so that I can see what the failures on Win32 and Win64 are so that hopefully they can be debugged and resolved.

UPDATE: I added back a variant of the GMP exception pass to ensure behavior for division by zero of GMP bigints stayed the same on Windows, too. See also https://github.com/JuliaPackaging/Yggdrasil/pull/1188


fingolfin:
The Win32 precheck now works, but the main test fails -- I am pretty sure this is because https://github.com/JuliaPackaging/Yggdrasil/pull/1188 isn't merged, so of course we are using the "old" GMP_jll 6.2 which had the old exceptions patch.

PallHaraldsson:
@StefanKarpinski Should this get into 1.5 RC2 as it has bugfixes and "major speedup" (and supposedly tested aslready, as a separate project)? [And maybe your "active project" PR today as a small change?]

StefanKarpinski:
No, upgrading dependencies is quite risky and not something to do in a release candidate. It could go into 1.5.1 if it's unproblematic on master for a while.

fingolfin:
This is indeed quite delicate. For now, GMP_jll 6.2.0 is set to require Julia >= 1.6 in the general registry; that would have to be adapted if it was put into Julia 1.5.1.

All in all, though, this might be a breaking change for other packages using GMP, so personally I would recommend to not put this into 1.5 but rather leave it for 1.6 (in the hope that 1.6 in turn won't be *that* far off ;-).

Issue #34949, opened 2020-03-01
help display code uses stdout directly instead of REPL output stream


JeffBezanson:
From the test in #34944:

```
julia> REPLTests.fake_repl() do in, out, repl
           repltask = @async begin
               REPL.run_repl(repl)
           end
           write(in, "?;\n")
           write(in, '\x04')
           wait(repltask)
       end
search:

```

Issue #32377, opened 2019-06-20
Base.runtests("REPL") fails, but not when run together with some other testset


sostock:
`Base.runtests("REPL")` (or `Base.runtests(["REPL"])`) fails, but running another testset together with it does not:

```julia
julia> Base.runtests(["REPL"])
[...]
Test Summary: | Pass  Fail  Broken  Total
  Overall     | 1007     1       5   1013
    REPL      | 1007     1       5   1013
    FAILURE

The global RNG seed was 0x975ed040f99437173b72ce7bc2fec875.

Error in testset REPL:
Test Failed at /home/sebastian/Development/julia/usr/share/julia/stdlib/v1.3/REPL/test/replcompletions.jl:107
  Expression: count(isequal("REPL"), c) == 1
   Evaluated: 2 == 1
[...]

julia> Base.runtests(["REPL", "cartesian"])
[...]
Test Summary: | Pass  Broken  Total
  Overall     | 1029       5   1034
    SUCCESS
```

This hapens on Julia 1.0.4, 1.1.1, 1.2.0-rc1, and master.

tgflynn:
This seems to be related to the way runtests uses worker threads.

If you do:

```
export JULIA_CPU_THREADS=1
./julia
Base.runtests(["REPL", "cartesian"])
```
The tests still fail, like in the single test case.

tgflynn:
Also of note:

```
include("stdlib/REPL/test/runtests.jl")
``` succeeds

but

```
using REPL
include("stdlib/REPL/test/runtests.jl")
```
fails.

twhitehead:
I believe the issue is that the test system doesn't truly isolate tests. Specifically, the Main module space is shared by the test running system and the tests themselves. This has bad effects in the REPL module which has several implicit and explicit references to Main.

What is screwing up this particular test is that test/runtests.jl imports REPL to get REPL.Terminal. This adds the module REPL to the Main namespace and causes the using REP completion test in stdlib/REPL/test/replcompletions.jl to return the module REPL brought into scope by the test runner in addition to the expected package REPL.

I tried to see if this issue could be worked around by wrapping replcompletions.jl in a module Main statement to hide the global one. While it did resolve this issue, it causes a few of the other later tests to stop working for reasons I haven't fully figured out. My wild guess would be that it might have to do with my Main not having the magic global Main export everything automatically property.

[...]

PR #36344, opened 2020-06-18
Fix Ctrl+Q for stack traces

### LINKED ISSUE #36340: Ctrl+Q stopped working for stack traces etc.

simeonschaub:
This should be a more general fix than #36341. It currently silently gets stuck while generating precompile statements on my machine, but not sure if that's reproducible.

close #36340

### DIFF
--- a/base/methodshow.jl
+++ b/base/methodshow.jl
@@ -240,9 +240,9 @@ function show_method_table(io::IO, ms::MethodList, max::Int=-1, header::Bool=tru
     end
     n = rest = 0
     local last
-    LAST_SHOWN_LINE_INFOS = get(io, :LAST_SHOWN_LINE_INFOS, Tuple{String,Int}[])
+    last_shown_line_infos = get(io, :last_shown_line_infos, Tuple{String,Int}[])
 
-    resize!(LAST_SHOWN_LINE_INFOS, 0)
+    resize!(last_shown_line_infos, 0)
     for meth in ms
         if max==-1 || n<max
             n += 1
@@ -250,7 +250,7 @@ function show_method_table(io::IO, ms::MethodList, max::Int=-1, header::Bool=tru
             print(io, "[$n] ")
             show(io, meth)
             file, line = updated_methodloc(meth)
-            push!(LAST_SHOWN_LINE_INFOS, (string(file), line))
+            push!(last_shown_line_infos, (string(file), line))
         else
             rest += 1
             last = meth
@@ -374,8 +374,8 @@ show(io::IO, mime::MIME"text/html", mt::Core.MethodTable) = show(io, mime, Metho
 
 # pretty-printing of AbstractVector{Method}
 function show(io::IO, mime::MIME"text/plain", mt::AbstractVector{Method})
-    LAST_SHOWN_LINE_INFOS = get(io, :LAST_SHOWN_LINE_INFOS, Tuple{String,Int}[])
-    resize!(LAST_SHOWN_LINE_INFOS, 0)
+    last_shown_line_infos = get(io, :last_shown_line_infos, Tuple{String,Int}[])
+    resize!(last_shown_line_infos, 0)
     first = true
     for (i, m) in enumerate(mt)
         first || println(io)
@@ -383,7 +383,7 @@ function show(io::IO, mime::MIME"text/plain", mt::AbstractVector{Method})
         print(io, "[$(i)] ")
         show(io, m)
         file, line = updated_methodloc(m)
-        push!(LAST_SHOWN_LINE_INFOS, (string(file), line))
+        push!(last_shown_line_infos, (string(file), line))
     end
 end
 
--- a/base/methodshow.jl
+++ b/base/methodshow.jl
@@ -240,9 +240,10 @@ function show_method_table(io::IO, ms::MethodList, max::Int=-1, header::Bool=tru
     end
     n = rest = 0
     local last
-    last_shown_line_infos = get(io, :last_shown_line_infos, Tuple{String,Int}[])
 
-    resize!(last_shown_line_infos, 0)
+    last_shown_line_infos = get(io, :last_shown_line_infos, nothing)
+    last_shown_line_infos === nothing || empty!(last_shown_line_infos)
+
     for meth in ms
         if max==-1 || n<max
             n += 1
@@ -250,7 +251,9 @@ function show_method_table(io::IO, ms::MethodList, max::Int=-1, header::Bool=tru
             print(io, "[$n] ")
             show(io, meth)
             file, line = updated_methodloc(meth)
-            push!(last_shown_line_infos, (string(file), line))
+            if last_shown_line_infos !== nothing
+                push!(last_shown_line_infos, (string(file), line))
+            end
         else
             rest += 1
             last = meth
@@ -374,8 +377,8 @@ show(io::IO, mime::MIME"text/html", mt::Core.MethodTable) = show(io, mime, Metho
 
 # pretty-printing of AbstractVector{Method}
 function show(io::IO, mime::MIME"text/plain", mt::AbstractVector{Method})
-    last_shown_line_infos = get(io, :last_shown_line_infos, Tuple{String,Int}[])
-    resize!(last_shown_line_infos, 0)
+    last_shown_line_infos = get(io, :last_shown_line_infos, nothing)
+    last_shown_line_infos === nothing || empty!(last_shown_line_infos)
     first = true
     for (i, m) in enumerate(mt)
         first || println(io)
@@ -383,7 +386,9 @@ function show(io::IO, mime::MIME"text/plain", mt::AbstractVector{Method})
         print(io, "[$(i)] ")
         show(io, m)
         file, line = updated_methodloc(m)
-        push!(last_shown_line_infos, (string(file), line))
+        if last_shown_line_infos !== nothing
+            push!(last_shown_line_infos, (string(file), line))
+        end
     end
 end
 
--- a/stdlib/REPL/src/REPL.jl
+++ b/stdlib/REPL/src/REPL.jl
@@ -212,21 +212,8 @@ function display(d::REPLDisplay, mime::MIME"text/plain", x)
                    init=IOContext(io, :limit => true, :module => Main))
     end
 
-    infos = Tuple{String,Int}[]
-    io = IOContext(io, :LAST_SHOWN_LINE_INFOS => infos)
-
     show(io, mime, x)
     println(io)
-
-    if !isempty(infos)
-        d.repl.last_shown_line_infos = infos
-        println(
-            io,
-            "\nTo edit a specific method, type the corresponding number into the " *
-            "REPL and press Ctrl+Q",
-        )
-    end
-
     nothing
 end
 display(d::REPLDisplay, x) = display(d, MIME("text/plain"), x)
@@ -234,7 +221,16 @@ display(d::REPLDisplay, x) = display(d, MIME("text/plain"), x)
 function print_response(repl::AbstractREPL, @nospecialize(response), show_value::Bool, have_color::Bool)
     repl.waserror = response[2]
     io = IOContext(outstream(repl), :module => Main)
+
     print_response(io, response, show_value, have_color, specialdisplay(repl))
+
+    if repl isa LineEditREPL && !isempty(repl.last_shown_line_infos)
+        println(
+            io,
+            "\nTo edit a specific method, type the corresponding number into the " *
+            "REPL and press Ctrl+Q",
+        )
+    end
     nothing
 end
 function print_response(errio::IO, @nospecialize(response), show_value::Bool, have_color::Bool, specialdisplay=nothing)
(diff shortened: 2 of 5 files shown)


tkf:
Maybe try `debug_output = stdout`?

https://github.com/JuliaLang/julia/blob/bf3ed5788eb9993db71798372128d47855d86715/contrib/generate_precompile.jl#L69

stdlib/REPL/src/REPL.jl
```diff
@@ -451,7 +456,7 @@ mutable struct LineEditREPL <: AbstractREPL
         new(t,hascolor,prompt_color,input_color,answer_color,shell_color,help_color,history_file,in_shell,
             in_help,envcolors,false,nothing, Options(), nothing, Tuple{String,Int}[])
 end
-outstream(r::LineEditREPL) = r.t
+outstream(r::LineEditREPL) = IOContext(r.t, :LAST_SHOWN_LINE_INFOS => r.last_shown_line_infos)
```
tkf, line 459: Don't you need to `empty!` it? Or maybe just use `Tuple{String,Int}[]` as before?

simeonschaub: No, `r.last_shown_line_infos` is still needed later to actually jump to the correct definition. The functions that then write to `LAST_SHOWN_LINE_INFOS` should actually `empty!` this array already.

stdlib/REPL/src/REPL.jl
```diff
@@ -212,29 +212,34 @@ function display(d::REPLDisplay, mime::MIME"text/plain", x)
...
 display(d::REPLDisplay, x) = display(d, MIME("text/plain"), x)
 
 function print_response(repl::AbstractREPL, @nospecialize(response), show_value::Bool, have_color::Bool)
     repl.waserror = response[2]
     io = IOContext(outstream(repl), :module => Main)
+
     print_response(io, response, show_value, have_color, specialdisplay(repl))
+
+    if repl isa LineEditREPL && !isempty(repl.last_shown_line_infos)
+        repl.last_shown_line_infos = infos
```
simeonschaub, line 236: :facepalm: That line should have been deleted

stdlib/REPL/src/REPL.jl
simeonschaub: Apparently, this doesn't always get cleared though...

simeonschaub: I think you are right, this should still get emptied, because otherwise `last_shown_line_infos` is not empty when other things get printed to the REPL and the hint still gets shown.

simeonschaub:
@tkf What do you think should be done about the helpmode case? The relevant code is in `docview.jl`, but the problem is that `io` is not passed through as an argument and some of those functions are documented, so I don't know if we can just easily change them.

tkf:
I guess we need https://github.com/JuliaLang/julia/issues/34949 to fix help mode? This PR as-is is still a big usability improvement/fix so I think we should just merge this.

Do you think it's possible to write a few tests for this?

simeonschaub:
It's probably possible to write a test here, but I'm not really familiar with how `REPLTests.fake_repl` works, so I would appreciate any pointers from people who have written those existing REPL tests. Last time I looked into it, I didn't get very far.

tkf:
I think it'd be easier to unit-test this than doing an integration test. It's a bit cumbersome to setup a REPL instance, though:

```julia
julia> term = REPL.TTYTerminal("dumb",IOBuffer(),IOBuffer(),IOBuffer());

julia> repl = REPL.LineEditREPL(term, false);

julia> repl.specialdisplay = REPL.REPLDisplay(repl);

julia> REPL.print_response(repl, (:hello, false), true, false)

julia> String(take!(term.out_stream))
":hello\n"
```

simeonschaub:
> I guess we need #34949 to fix help mode? This PR as-is is still a big usability improvement/fix so I think we should just merge this.

 #36289 doesn't really buy us a lot here, because `methods` actually gets called in e.g. `Docs.@doc Base._in`, which in turn calls `REPL.lookup_doc`, so fixing this would mean having to change how parts of the Docs module work.

tkf:
Ah, I see. So `Docs.@doc` macro creates a string before it is printed again. Yeah, it sounds like a non-trivial problem.

stdlib/REPL/src/REPL.jl
```diff
@@ -451,7 +447,11 @@ mutable struct LineEditREPL <: AbstractREPL
         new(t,hascolor,prompt_color,input_color,answer_color,shell_color,help_color,history_file,in_shell,
             in_help,envcolors,false,nothing, Options(), nothing, Tuple{String,Int}[])
 end
-outstream(r::LineEditREPL) = r.t
+function outstream(r::LineEditREPL)
+    linfos = r.last_shown_line_infos
+    empty!(linfos)
+    IOContext(r.t, :LAST_SHOWN_LINE_INFOS => linfos)
```
tkf, line 453:
```suggestion
    IOContext(r.t, :LAST_SHOWN_LINE_INFOS => Tuple{String,Int}[])
```

stdlib/REPL/src/REPL.jl
```diff
@@ -212,29 +212,25 @@ function display(d::REPLDisplay, mime::MIME"text/plain", x)
...
 end
 display(d::REPLDisplay, x) = display(d, MIME("text/plain"), x)
 
 function print_response(repl::AbstractREPL, @nospecialize(response), show_value::Bool, have_color::Bool)
     repl.waserror = response[2]
     io = IOContext(outstream(repl), :module => Main)
+
     print_response(io, response, show_value, have_color, specialdisplay(repl))
+
+    if repl isa LineEditREPL && !isempty(repl.last_shown_line_infos)
```
tkf, line 227:
```suggestion
    if repl isa LineEditREPL && !isempty(infos)
        repl.last_shown_line_infos = infos
```

stdlib/REPL/src/REPL.jl
```diff
@@ -212,29 +212,25 @@ function display(d::REPLDisplay, mime::MIME"text/plain", x)
...
-        )
-    end
-
     nothing
 end
 display(d::REPLDisplay, x) = display(d, MIME("text/plain"), x)
 
 function print_response(repl::AbstractREPL, @nospecialize(response), show_value::Bool, have_color::Bool)
     repl.waserror = response[2]
     io = IOContext(outstream(repl), :module => Main)
```
tkf, line 223:
```suggestion
    infos = Tuple{String,Int}[]
    io = IOContext(outstream(repl), :module => Main, :LAST_SHOWN_LINE_INFOS => infos)
```

stdlib/REPL/src/REPL.jl
```diff
@@ -451,7 +447,11 @@ mutable struct LineEditREPL <: AbstractREPL
         new(t,hascolor,prompt_color,input_color,answer_color,shell_color,help_color,history_file,in_shell,
             in_help,envcolors,false,nothing, Options(), nothing, Tuple{String,Int}[])
 end
-outstream(r::LineEditREPL) = r.t
+function outstream(r::LineEditREPL)
+    linfos = r.last_shown_line_infos
+    empty!(linfos)
+    IOContext(r.t, :LAST_SHOWN_LINE_INFOS => linfos)
+end
```
tkf, line 454:
```suggestion
outstream(r::LineEditREPL) = r.t
```

tkf (commented):
I think it's better to set `IOContext` outside `outstream`.  This function sounds like an accessor function and having a side-effect is not great.  For example, it'd mean some innocent use like

https://github.com/JuliaLang/julia/blob/5142abf4507525a7a02e7d984ce39802ee4fe134/stdlib/REPL/src/REPL.jl#L982

would clear out `last_shown_line_infos`.  I think it'd be cleaner if we can keep `outstream` as a simple accessor.  We've already discussed this in https://github.com/JuliaLang/julia/pull/36344#discussion_r442324513 but do you think the change I suggested here work?  It's just moving what we were doing in `display` to `print_response`.

stdlib/REPL/src/REPL.jl
simeonschaub: Interestingly enough, that doesn't work for `show(::MethodTable)` anymore. Will have to investigate.

KristofferC:
Bump.

stdlib/REPL/test/replcompletions.jl
```diff
@@ -110,7 +110,7 @@ end
 
 let s = "using REP"
     c, r = test_complete(s)
-    @test count(isequal("REPL"), c) == 1
+    @test count(isequal("REPL"), c) == 2
```
simeonschaub, line 113: I have no clue what's going on here...

simeonschaub (commented):
Sorry, my laptop crashed, that's why it took me some time to restore the changes I made. I refactored most of the logic behind the hint message into `with_methodtable_hint` to avoid duplicating this code. I also realized that `LAST_SHOWN_LINE_INFOS` should probably be lowercase now, since it's no longer a global variable.

stdlib/REPL/test/replcompletions.jl
simeonschaub: Weird, this now breaks the CI test,  but was needed locally on my machine for tests to pass 🤔.

KristofferC: > Weird, this now breaks the CI test, but was needed locally on my machine for tests to pass

Yeah, it's a bug in the REPL tests.

simeonschaub: Does CI run the tests with different options or why isn't this bug present there?

KristofferC: https://github.com/JuliaLang/julia/issues/32377 has some more info.

base/methodshow.jl
```diff
@@ -240,17 +240,17 @@ function show_method_table(io::IO, ms::MethodList, max::Int=-1, header::Bool=tru
     end
     n = rest = 0
     local last
-    LAST_SHOWN_LINE_INFOS = get(io, :LAST_SHOWN_LINE_INFOS, Tuple{String,Int}[])
+    last_shown_line_infos = get(io, :last_shown_line_infos, Tuple{String,Int}[])
 
-    resize!(LAST_SHOWN_LINE_INFOS, 0)
+    resize!(last_shown_line_infos, 0)
```
KristofferC, line 245: What's the point in modifying `last_shown_line_infos` when it isn't provided by the `io`? The result will never be used in that case, right?

simeonschaub: Yes, you are correct, this was just the easiest way to implement this at the time. Do you think it's worth refactoring this logic?

stdlib/REPL/test/replcompletions.jl
simeonschaub: Ah, thanks for the pointer!

base/methodshow.jl
KristofferC: I think it would be better to do 

```
last_linfo = get(io, :last_shown_line_infos, nothing)
...
last_linfo === nothing || push!(last_shown_line_infos, (string(file), line))
```

Is the `resize!` needed now? There is always a length zero coming from the `io`, no?

staticfloat:
Just a note that this is somehow freezing up the buildbots :(

simeonschaub:
Are they actually frozen? To me, it looked like they finished, but somehow didn't report back correctly to GitHub. Might have been because GitHub was down for a while?

tkf:
Yeah it looks like the problem is on GitHub side. For example, `tester_linux64` is finished but `package_linux64` still says it's pending.

base/methodshow.jl
tkf: I think we still need `resize!` since you can print multiple stack traces in one `io`. Previous discussion:

https://github.com/JuliaLang/julia/pull/35556#discussion_r435535024
https://github.com/JuliaLang/julia/pull/35556#issuecomment-643663448

staticfloat:
> Are they actually frozen? To me, it looked like they finished, but somehow didn't report back correctly to GitHub.

Yes they're freezing, I go in manually and end them.  Here's one that was been running for 21 hours before I killed it: https://build.julialang.org/#/builders/77/builds/1247

simeonschaub:
Do you think that's related to this PR? That's kind of the annoying thing about changing something in REPL – You change some small thing and suddenly some other thing gets silently broken/stuck. :smile:

staticfloat:
Yes, it's been consistent with this PR, unfortunately.  Here's another one: https://build.julialang.org/#/builders/77/builds/1281

That being said, the latest version of this seems to have made it through the `package_*` builders for `win64`, so maybe you fixed it?  I'm only seeing it on Windows, perhaps there's something particularly weird about that platform's handling of this?

simeonschaub:
Weird... I don't see how any of these changes could have fixed this.

simeonschaub:
I've rebased onto the latest master, let's hope the Windows build still goes through. :grin:

simeonschaub:
@staticfloat All tests now passed, so it seems like this should be ok now. @tkf Would you be willing to review?

stdlib/REPL/src/REPL.jl
```diff
@@ -495,6 +484,22 @@ function complete_line(c::LatexCompletions, s)
     return unique!(map(completion_text, ret)), partial[range], should_complete
 end
 
+with_methodtable_hint(f, repl) = f(outstream(repl))
+function with_methodtable_hint(f, repl::LineEditREPL)
```
tkf, line 488: :+1: I think this is a clean approach!

tkf (approved):
LGTM!

tkf:
@KristofferC Do you want to review it again?

If no objections, let's merge this in a few days.

simeonschaub:
Bump.

tkf:
Thanks for the fix! (and the ping)

PR #36346, opened 2020-06-18
Consider terminal emacs in EDITOR_CALLBACKS before graphical

non-Jedi:
Since the regex `r"\bemacs"` matches the cmd `emacs -nw`, the regex
`r"\bemacs\b.*\s(-nw|--no-window-system)\b"` should be considered in the
`EDITOR_CALLBACKS` list before it.

Without this change, when `JULIA_EDITOR="emacs -nw"`, julia tries to
open the editor with `run(...; wait=false)` which causes the error:

    emacs: standard input is not a tty

or for emacsclient:

    emacsclient: could not get terminal name
    emacsclient: error executing alternate editor ""

cc @rfourquet

### DIFF
--- a/stdlib/InteractiveUtils/src/editless.jl
+++ b/stdlib/InteractiveUtils/src/editless.jl
@@ -58,6 +58,7 @@ Note that many editors are already defined. All of the following commands should
 already work:
 
 - emacs
+- emacsclient
 - vim
 - nvim
 - nano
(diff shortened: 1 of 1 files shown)


stdlib/InteractiveUtils/src/editless.jl
```diff
@@ -113,15 +114,16 @@ function define_default_editors()
     define_editor(r".*") do cmd, path, line
         `$cmd $path`
     end
+    define_editor([r"\bemacs", "gedit", r"\bgvim"]) do cmd, path, line
```
rfourquet, line 117: Just out of curiosity, changing `r"\bemacs"` to `r"\bemacs\b"` would also work as a fix, right?

rfourquet (approved):

stdlib/InteractiveUtils/src/editless.jl
yuyichao: That still matches `emacs -nw`?

non-Jedi: If we changed this to `r"\bemacs\b"`, then graphical emacsclient would no longer be matched, so it would end up being caught by the fallback `r".*"` regex which doesn't jump to line.

rfourquet: Ah right, thank you both.

rfourquet:
For people seeing this issue and wanting a fix for Julia 1.4 and 1.5, you can put the following in startup.jl:
```
using InteractiveUtils
InteractiveUtils.define_editor([
    "vim", "vi", "nvim", "mvim", "nano",
    r"\bemacs\b.*\s(-nw|--no-window-system)\b",
    r"\bemacsclient\b.\s*-(-?nw|t|-?tty)\b"], wait=true) do cmd, path, line
        `$cmd +$line $path`
    end
```
(this is copy-pasted and could be simplified but I didn't bother).

PR #36350, opened 2020-06-18
Refactor `abstract_eval` to separate out statements and values

Keno:
In preparation for adding the ability for statements to return
additional info, while values will only ever have a type.


vtjnash:
> while values will only ever have a type

This seems a bit arbitrary, since might values eventually also eventually participate in computing other attribute information (e.g. uses/escapes, pure/effects)? But I guess this PR is essentially working towards demonstrating what we meant by having linear IR? (And even "type" already includes some non-`Type` information in the form of `Const`)

Keno:
Perhaps. For the moment I mostly want to add additional information to `:call`s. I agree that we may want to do something with `GlobalRef` in the future also, but I think we can cross that bridge when we get there. My sense is that if something is complex enough to need extra statement information, it probably should be its own statement anyway, so it can be moved around and processed independently. Otherwise we're in the weird "not quite linear" itermediate state.

JeffBezanson (approved):

PR #36352, opened 2020-06-18
add doctests for frexp, expm1, exponent

mgautam98:
I added doctests for some of the maths functions
1. frexp
2. exponent
3. expm1


base/math.jl
```diff
@@ -360,6 +360,14 @@ asinh(x::Number)
     expm1(x)
 
 Accurately compute ``e^x-1``.
+# Examples
+```jldoctest
+julia> expm1(2)
```
simeonschaub, line 365: Would comparing `expm1(1e-16)` with `exp(1e-16) - 1` be a better example here? Otherwise I could see new users being confused why this exists at all.

mgautam98: Yes, thats a nice idea I will do that

base/math.jl
```diff
@@ -360,6 +360,14 @@ asinh(x::Number)
     expm1(x)
 
 Accurately compute ``e^x-1``.
+# Examples
+```jldoctest
+julia> exp(1e-16) - 1 ≈ expm1(1e-16)
```
simeonschaub, line 365: Using `isapprox` here might be a bit confusing, because from a first look, it's not really clear that this is just because you didn't specify an absolute tolerance. I would probably just evaluate these two expressions in spearate lines, so it's clear what each of them returns.

mgautam98: Also should we also add into the docstring that `function avoids the loss of precision involved in the direct evaluation of exp(x)-1 for small x` ?

simeonschaub: That sounds good to me!

simeonschaub (approved):

mgautam98:
@simeonschaub I don't know why the tests are failing

simeonschaub:
The failure seems unrelated to this PR, so I wouldn't worry about it. That sometimes happens with the IO and threading tests.

simeonschaub:
@dkarrasch Could this be merged, once the merge conflicts are resolved?

simeonschaub:
@mgautam98 Can you rebase this to resolve the conflicts?

ViralBShah:
Let's merge if tests pass.

dkarrasch:
Test failures seem unrelated: time outs, internet connection, Distributed.

PR #36355, opened 2020-06-19
fix #36230, more efficient lowering of `if` with a chain of `&&`

### LINKED ISSUE #36230: Missing TypedSlot in boolean `if` chains
Inference fails to take full advantage of the knowledge that `a` and `b` are both `Symbol`s:
```julia
julia> function symcmp(vec)
           a, b = vec[1], vec[2]
           if isa(a, Symbol) && isa(b, Symbol)
               return a == b
           end
           return false
       end
symcmp (generic function with 1 method)

julia> code_typed(symcmp, (Vector{Any},))
1-element Array{Any,1}:
 CodeInfo(
1 ── %1  = Base.arrayref(true, vec, 1)::Any
│    %2  = Base.arrayref(true, vec, 2)::Any
│    %3  = (%1 isa Main.Symbol)::Bool
└───       goto #3 if not %3
2 ── %5  = (%2 isa Main.Symbol)::Bool
└───       goto #4
3 ──       nothing::Nothing
4 ┄─ %8  = φ (#2 => %5, #3 => false)::Bool
└───       goto #11 if not %8
5 ── %10 = π (%2, Symbol)
│    %11 = Main.:(==)::Core.Compiler.Const(==, false)
│    %12 = (isa)(%1, WeakRef)::Bool
└───       goto #7 if not %12
6 ── %14 = π (%1, WeakRef)
│    %15 = invoke %11(%14::WeakRef, %10::Symbol)::Union{Missing, Bool}
└───       goto #10
7 ── %17 = (isa)(%1, Missing)::Bool
└───       goto #9 if not %17
8 ──       goto #10
9 ── %20 = (%1 == %10)::Union{Missing, Bool}
└───       goto #10
10 ┄ %22 = φ (#6 => %15, #8 => $(QuoteNode(missing)), #9 => %20)::Union{Missing, Bool}
└───       return %22
11 ─       return false
) => Union{Missing, Bool}
```

Interestingly, the pure `&&` chain does not have the same problem:

```julia
julia> function symcmp(vec)
           a, b = vec[1], vec[2]
           isa(a, Symbol) && isa(b, Symbol) && return a == b
           return false
       end
symcmp (generic function with 1 method)

julia> code_typed(symcmp, (Vector{Any},))
1-element Array{Any,1}:
 CodeInfo(
1 ─ %1 = Base.arrayref(true, vec, 1)::Any
│   %2 = Base.arrayref(true, vec, 2)::Any
│   %3 = (%1 isa Main.Symbol)::Bool
└──      goto #5 if not %3
2 ─ %5 = (%2 isa Main.Symbol)::Bool
└──      goto #4 if not %5
3 ─ %7 = π (%1, Symbol)
│   %8 = π (%2, Symbol)
│   %9 = (%7 === %8)::Bool
└──      return %9
4 ─      nothing::Nothing
5 ┄      return false
) => Bool
```

It may not be exactly the same issue (it also involves constant-propagation), but this is also not inferred successfully:

```julia
julia> code_typed(Meta.isexpr, (Any, Symbol, Int))
1-element Array{Any,1}:
 CodeInfo(
1 ─ %1  = (ex isa Base.Meta.Expr)::Bool
└──       goto #3 if not %1
2 ─ %3  = π (ex, Expr)
│   %4  = Base.getfield(%3, :head)::Symbol
│   %5  = (%4 === heads)::Bool
└──       goto #4
3 ─       goto #4
4 ┄ %8  = φ (#2 => %5, #3 => false)::Bool
└──       goto #6 if not %8
5 ─ %10 = Base.getproperty(ex, :args)::Any
│   %11 = Base.Meta.length(%10)::Any
│   %12 = (%11 == n)::Any
└──       return %12
6 ─       return false
) => Any
```
And finally, in a dream world, inference would know that a successful `isexpr` call implies that `ex::Expr`. This is assumed all over the place, e.g., https://github.com/JuliaLang/julia/blob/38f2c594c49cec46d38be6dd614a5592582a3016/base/show.jl#L1038-L1039 to pick just one of many, but currently `item.args` is inferred as `Any`. This affects even the 2-argument form, e.g.,
```julia
julia> function useisexpr(ex)
           if isexpr(ex, :(=))
               lhs, rhs = ex.args[1], ex.args[2]
               return lhs
           end
           return nothing
       end

julia> code_typed(useisexpr, (Any,))
1-element Array{Any,1}:
 CodeInfo(
1 ─ %1  = (ex isa Base.Meta.Expr)::Bool
└──       goto #3 if not %1
2 ─ %3  = π (ex, Expr)
│   %4  = Base.getfield(%3, :head)::Symbol
│   %5  = (%4 === :(=))::Bool
└──       goto #4
3 ─       goto #4
4 ┄ %8  = φ (#2 => %5, #3 => false)::Bool
└──       goto #6 if not %8
5 ─ %10 = Base.getproperty(ex, :args)::Any
│   %11 = Base.getindex(%10, 1)::Any
│   %12 = Base.getproperty(ex, :args)::Any
│         Base.getindex(%12, 2)::Any
└──       return %11
6 ─       return Main.nothing
) => Any
```
I recognize this might be more than one issue and will be happy to split it up if desired.

I've tagged this with latency because it seems to be a really common source of invalidations for packages like JuliaInterpreter and similar that have to deal with a lot of poorly-typed data. I constantly think I'm being careful, but often it's not working. (Noticed while investigating #36005, though @KristofferC pointed this out to me earlier.)

JeffBezanson:
This lowers `if a && b && ...` to just a series of conditional branches instead of introducing a temporary slot, allowing inference's branch condition propagation to work effectively. Does not fix the `if isexpr(...) && ...` case though; that is really pretty far out of reach for now.

fix #36230

### DIFF
--- a/base/expr.jl
+++ b/base/expr.jl
@@ -347,9 +347,9 @@ function findmeta_block(exargs, argsmatch=args->true)
     for i = 1:length(exargs)
         a = exargs[i]
         if isa(a, Expr)
-            if (a::Expr).head === :meta && argsmatch((a::Expr).args)
+            if a.head === :meta && argsmatch(a.args)
                 return i, exargs
-            elseif (a::Expr).head === :block
+            elseif a.head === :block
                 idx, exa = findmeta_block(a.args, argsmatch)
                 if idx != 0
                     return idx, exa
--- a/base/meta.jl
+++ b/base/meta.jl
@@ -62,8 +62,9 @@ true
 ```
 """
 isexpr(@nospecialize(ex), head::Symbol) = isa(ex, Expr) && ex.head === head
-isexpr(@nospecialize(ex), heads::Union{Set,Vector,Tuple}) = isa(ex, Expr) && in(ex.head, heads)
-isexpr(@nospecialize(ex), heads, n::Int) = isexpr(ex, heads) && length((ex::Expr).args) == n
+isexpr(@nospecialize(ex), heads) = isa(ex, Expr) && in(ex.head, heads)
+isexpr(@nospecialize(ex), head::Symbol, n::Int) = isa(ex, Expr) && ex.head === head && length(ex.args) == n
+isexpr(@nospecialize(ex), heads, n::Int) = isa(ex, Expr) && in(ex.head, heads) && length(ex.args) == n
 
 """
     Meta.show_sexpr([io::IO,], ex)
--- a/src/julia-syntax.scm
+++ b/src/julia-syntax.scm
@@ -1895,6 +1895,13 @@
                 (else
                  (error (string "invalid " syntax-str " \"" (deparse el) "\""))))))))
 
+(define (expand-if e)
+  (if (and (pair? (cadr e)) (eq? (car (cadr e)) '&&))
+      (let ((clauses (cdr (flatten-ex '&& (cadr e)))))
+        `(if (&& ,@(map expand-forms clauses))
+             ,@(map expand-forms (cddr e))))
+      (cons (car e) (map expand-forms (cdr e)))))
+
 ;; move an assignment into the last statement of a block to keep more statements at top level
 (define (sink-assignment lhs rhs)
   (if (and (pair? rhs) (eq? (car rhs) 'block))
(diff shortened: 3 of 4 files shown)


vtjnash (approved):
Oh, this is great! I didn't realize lowering could just map all of the else clauses to the same point and solve this.

timholy:
Yay, this is wonderful!

Advertising campaign queued and ready: "Julia 1.6: now with fewer `::`!"

PR #36356, opened 2020-06-19
Use haslength to check for HasShape and HasLength traits

harryscholes:
@garborg drew my attention to `haslength` in #35254. Here, I tidy up two `HasShape` and `HasLength` trait checks with `haslength`


rfourquet (approved):

StefanKarpinski:
Looks good. I restarted the macOS run although there's basically zero chance that this breaks macOS. Feel free to merge when it passes.

PR #36360, opened 2020-06-19
add BLAS.get_num_threads

jw3126:
From slack, also came up [on discourse](https://discourse.julialang.org/t/how-to-get-number-of-current-blas-threads/32090)

### DIFF
--- a/stdlib/LinearAlgebra/src/blas.jl
+++ b/stdlib/LinearAlgebra/src/blas.jl
@@ -144,7 +144,7 @@ function get_num_threads()
         s = get(ENV, key, "")
         nt = Base.tryparse(Cint, s)
         if nt === nothing
-            @warn "Failed to read environment variable $key"
+            @warn "Failed to read environment variable $key" maxlog=1
         else
             return nt
         end
--- a/stdlib/LinearAlgebra/src/blas.jl
+++ b/stdlib/LinearAlgebra/src/blas.jl
@@ -107,9 +107,14 @@ end
 openblas_get_config() = strip(unsafe_string(ccall((@blasfunc(openblas_get_config), libblas), Ptr{UInt8}, () )))
 
 """
-    set_num_threads(n)
+    set_num_threads(n::Integer)
+    set_num_threads(::Nothing)
 
-Set the number of threads the BLAS library should use.
+Set the number of threads the BLAS library should use equal to `n::Integer`.
+If `nothing` is passed to this function, julia tries to figure out the optimial number of threads.
+The exact heuristic is an implementation detail.
+
+On exotic variants of `BLAS` this function can fail, which is indicated by returning `nothing`.
 """
 function set_num_threads(n::Integer)
     blas = vendor()
@@ -128,10 +133,17 @@ function set_num_threads(n::Integer)
     return nothing
 end
 
+function set_num_threads(::Nothing)
+    n = max(1, Sys.CPU_THREADS ÷ 2)
+    set_num_threads(n)
+end
+
 """
     get_num_threads()
 
 Get the number of threads the BLAS library is using.
+
+On exotic variants of `BLAS` this function can fail, which is indicated by returning `nothing`.
 """
 function get_num_threads()
     blas = LinearAlgebra.BLAS.vendor()
@@ -149,8 +161,9 @@ function get_num_threads()
             return nt
         end
     end
-    @warn "Could not get number of BLAS threads. Returning Sys.CPU_THREADS instead." maxlog=1
-    return Sys.CPU_THREADS
+    ret = nothing
+    @warn "Could not get number of BLAS threads. Returning $ret instead." maxlog=1
+    return ret
 end
 
 const _testmat = [1.0 0.0; 0.0 -1.0]
--- a/stdlib/LinearAlgebra/src/blas.jl
+++ b/stdlib/LinearAlgebra/src/blas.jl
@@ -106,6 +106,17 @@ end
 
 openblas_get_config() = strip(unsafe_string(ccall((@blasfunc(openblas_get_config), libblas), Ptr{UInt8}, () )))
 
+function guess_vendor()
+    # like determine_vendor, but guesses blas in some cases
+    # where determine_vendor returns :unknown
+    ret = vendor()
+    if Sys.isapple() && (ret == :unknown )
+        ret = :osxblas
+    end
+    ret
+end
+
+
 """
     set_num_threads(n::Integer)
     set_num_threads(::Nothing)
@@ -116,17 +127,17 @@ The exact heuristic is an implementation detail.
 
 On exotic variants of `BLAS` this function can fail.
 """
-function set_num_threads(n::Integer)::Nothing
-    blas = vendor()
-    if blas === :openblas || blas == :openblas64
+function set_num_threads(n::Integer, _blas=guess_vendor())::Nothing
+    if _blas === :openblas || _blas == :openblas64
         return ccall((@blasfunc(openblas_set_num_threads), libblas), Cvoid, (Cint,), n)
-    elseif blas === :mkl
+    elseif _blas === :mkl
         # MKL may let us set the number of threads in several ways
         return ccall((:MKL_Set_Num_Threads, libblas), Cvoid, (Cint,), n)
-    elseif Sys.isapple()
+    elseif _blas === :osxblas
         # OSX BLAS looks at an environment variable
         ENV["VECLIB_MAXIMUM_THREADS"] = n
     else
+        @assert _blas === :unknown
         @warn "Failed to set number of BLAS threads." maxlog=1
     end
     return nothing
@@ -134,13 +145,13 @@ end
 
 _tryparse_env_int(key) = tryparse(Int, get(ENV, key, ""))
 
-function set_num_threads(::Nothing)
+function set_num_threads(::Nothing, _blas=guess_vendor())
     n = something(
         _tryparse_env_int("OPENBLAS_NUM_THREADS"),
         _tryparse_env_int("OMP_NUM_THREADS"),
         max(1, Sys.CPU_THREADS ÷ 2),
     )
-    set_num_threads(n)
+    set_num_threads(n, _blas)
 end
 
 """
@@ -150,13 +161,12 @@ Get the number of threads the BLAS library is using.
 
 On exotic variants of `BLAS` this function can fail, which is indicated by returning `nothing`.
 """
-function get_num_threads()::Union{Int, Nothing}
-    blas = vendor()
-    if blas === :openblas || blas === :openblas64
+function get_num_threads(_blas=guess_vendor())::Union{Int, Nothing}
+    if _blas === :openblas || _blas === :openblas64
         return Int(ccall((@blasfunc(openblas_get_num_threads), libblas), Cint, ()))
-    elseif blas == :mkl
+    elseif _blas === :mkl
         return Int(ccall((:mkl_get_max_threads, libblas), Cint, ()))
-    elseif Sys.isapple()
+    elseif _blas === :osxblas
         key = "VECLIB_MAXIMUM_THREADS"
         nt = _tryparse_env_int(key)
         if nt === nothing
@@ -164,6 +174,8 @@ function get_num_threads()::Union{Int, Nothing}
         else
             return nt
         end
+    else
+        @assert _blas === :unknown
     end
     @warn "Could not get number of BLAS threads. Returning `nothing` instead." maxlog=1
     return nothing
--- a/stdlib/LinearAlgebra/src/blas.jl
+++ b/stdlib/LinearAlgebra/src/blas.jl
@@ -122,10 +122,26 @@ end
     set_num_threads(::Nothing)
 
 Set the number of threads the BLAS library should use equal to `n::Integer`.
-If `nothing` is passed to this function, julia tries to figure out the optimial number of threads.
-The exact heuristic is an implementation detail.
 
-On exotic variants of `BLAS` this function can fail.
+Also accepts `nothing`, in which case julia tries to guess the default number of threads.
+Passing `nothing` is discouraged and mainly exists for the following reason:
+
+On exotic variants of BLAS, `nothing` may be returned by `get_num_threads()`.
+Thus on exotic variants of BLAS, the following pattern may fail to set the number of threads:
+
+```julia
+old = get_num_threads()
+set_num_threads(1)
+@threads for i in 1:10
+    # single-threaded BLAS calls
+end
+set_num_threads(old)
+```
+Because `set_num_threads` accepts `nothing`, this code can still run
+on exotic variants of BLAS without error. Warnings will be raised instead.
+
+!!! compat "Julia 1.6"
+    `set_num_threads(::Nothing)` requires at least Julia 1.6.
 """
 set_num_threads(n)::Nothing = _set_num_threads(n)
 
@@ -162,6 +178,9 @@ end
 Get the number of threads the BLAS library is using.
 
 On exotic variants of `BLAS` this function can fail, which is indicated by returning `nothing`.
+
+!!! compat "Julia 1.6"
+    `get_num_threads` requires at least Julia 1.6.
 """
 get_num_threads(;_blas=guess_vendor())::Union{Int, Nothing} = _get_num_threads()
 
--- a/stdlib/LinearAlgebra/src/blas.jl
+++ b/stdlib/LinearAlgebra/src/blas.jl
@@ -127,7 +127,9 @@ The exact heuristic is an implementation detail.
 
 On exotic variants of `BLAS` this function can fail.
 """
-function set_num_threads(n::Integer; _blas=guess_vendor())::Nothing
+set_num_threads(n)::Nothing = _set_num_threads(n)
+
+function _set_num_threads(n::Integer; _blas = guess_vendor())
     if _blas === :openblas || _blas == :openblas64
         return ccall((@blasfunc(openblas_set_num_threads), libblas), Cvoid, (Cint,), n)
     elseif _blas === :mkl
--- a/stdlib/LinearAlgebra/src/blas.jl
+++ b/stdlib/LinearAlgebra/src/blas.jl
@@ -127,7 +127,7 @@ The exact heuristic is an implementation detail.
 
 On exotic variants of `BLAS` this function can fail.
 """
-function set_num_threads(n::Integer, _blas=guess_vendor())::Nothing
+function set_num_threads(n::Integer; _blas=guess_vendor())::Nothing
     if _blas === :openblas || _blas == :openblas64
         return ccall((@blasfunc(openblas_set_num_threads), libblas), Cvoid, (Cint,), n)
     elseif _blas === :mkl
@@ -145,13 +145,13 @@ end
 
 _tryparse_env_int(key) = tryparse(Int, get(ENV, key, ""))
 
-function set_num_threads(::Nothing, _blas=guess_vendor())
+function set_num_threads(::Nothing; _blas=guess_vendor())
     n = something(
         _tryparse_env_int("OPENBLAS_NUM_THREADS"),
         _tryparse_env_int("OMP_NUM_THREADS"),
         max(1, Sys.CPU_THREADS ÷ 2),
     )
-    set_num_threads(n, _blas)
+    set_num_threads(n; _blas)
 end
 
 """
@@ -161,7 +161,7 @@ Get the number of threads the BLAS library is using.
 
 On exotic variants of `BLAS` this function can fail, which is indicated by returning `nothing`.
 """
-function get_num_threads(_blas=guess_vendor())::Union{Int, Nothing}
+function get_num_threads(;_blas=guess_vendor())::Union{Int, Nothing}
     if _blas === :openblas || _blas === :openblas64
         return Int(ccall((@blasfunc(openblas_get_num_threads), libblas), Cint, ()))
     elseif _blas === :mkl
--- a/stdlib/LinearAlgebra/src/blas.jl
+++ b/stdlib/LinearAlgebra/src/blas.jl
@@ -122,7 +122,7 @@ function set_num_threads(n::Integer)
         # OSX BLAS looks at an environment variable
         ENV["VECLIB_MAXIMUM_THREADS"] = n
     else
-        @warn "Failed to set number of BLAS threads."
+        @warn "Failed to set number of BLAS threads." maxlog=1
     end
 
     return nothing
--- a/stdlib/LinearAlgebra/src/blas.jl
+++ b/stdlib/LinearAlgebra/src/blas.jl
@@ -163,7 +163,9 @@ Get the number of threads the BLAS library is using.
 
 On exotic variants of `BLAS` this function can fail, which is indicated by returning `nothing`.
 """
-function get_num_threads(;_blas=guess_vendor())::Union{Int, Nothing}
+get_num_threads(;_blas=guess_vendor())::Union{Int, Nothing} = _get_num_threads()
+
+function _get_num_threads(; _blas = guess_vendor())::Union{Int, Nothing}
     if _blas === :openblas || _blas === :openblas64
         return Int(ccall((@blasfunc(openblas_get_num_threads), libblas), Cint, ()))
     elseif _blas === :mkl
--- a/stdlib/LinearAlgebra/src/blas.jl
+++ b/stdlib/LinearAlgebra/src/blas.jl
@@ -133,8 +133,14 @@ function set_num_threads(n::Integer)
     return nothing
 end
 
+_tryparse_env_cint(key) = tryparse(Cint, get(ENV, key, ""))
+
 function set_num_threads(::Nothing)
-    n = max(1, Sys.CPU_THREADS ÷ 2)
+    n = something(
+        _tryparse_env_cint("OPENBLAS_NUM_THREADS"),
+        _tryparse_env_cint("OMP_NUM_THREADS"),
+        max(1, Sys.CPU_THREADS ÷ 2),
+    )
     set_num_threads(n)
 end
 
--- a/stdlib/LinearAlgebra/src/blas.jl
+++ b/stdlib/LinearAlgebra/src/blas.jl
@@ -153,7 +153,7 @@ function _set_num_threads(::Nothing; _blas = guess_vendor())
         _tryparse_env_int("OMP_NUM_THREADS"),
         max(1, Sys.CPU_THREADS ÷ 2),
     )
-    set_num_threads(n; _blas)
+    _set_num_threads(n; _blas)
 end
 
 """
--- a/stdlib/LinearAlgebra/src/blas.jl
+++ b/stdlib/LinearAlgebra/src/blas.jl
@@ -136,8 +136,8 @@ _tryparse_env_int(key) = tryparse(Int, get(ENV, key, ""))
 
 function set_num_threads(::Nothing)
     n = something(
-        _tryparse_env_cint("OPENBLAS_NUM_THREADS"),
-        _tryparse_env_cint("OMP_NUM_THREADS"),
+        _tryparse_env_int("OPENBLAS_NUM_THREADS"),
+        _tryparse_env_int("OMP_NUM_THREADS"),
         max(1, Sys.CPU_THREADS ÷ 2),
     )
     set_num_threads(n)
--- a/stdlib/LinearAlgebra/src/blas.jl
+++ b/stdlib/LinearAlgebra/src/blas.jl
@@ -132,7 +132,7 @@ function set_num_threads(n::Integer)::Nothing
     return nothing
 end
 
-_tryparse_env_cint(key) = tryparse(Cint, get(ENV, key, ""))
+_tryparse_env_int(key) = tryparse(Int, get(ENV, key, ""))
 
 function set_num_threads(::Nothing)
     n = something(
--- a/stdlib/LinearAlgebra/src/blas.jl
+++ b/stdlib/LinearAlgebra/src/blas.jl
@@ -147,7 +147,7 @@ end
 
 _tryparse_env_int(key) = tryparse(Int, get(ENV, key, ""))
 
-function set_num_threads(::Nothing; _blas=guess_vendor())
+function _set_num_threads(::Nothing; _blas = guess_vendor())
     n = something(
         _tryparse_env_int("OPENBLAS_NUM_THREADS"),
         _tryparse_env_int("OMP_NUM_THREADS"),
--- a/stdlib/LinearAlgebra/src/blas.jl
+++ b/stdlib/LinearAlgebra/src/blas.jl
@@ -114,12 +114,12 @@ Set the number of threads the BLAS library should use equal to `n::Integer`.
 If `nothing` is passed to this function, julia tries to figure out the optimial number of threads.
 The exact heuristic is an implementation detail.
 
-On exotic variants of `BLAS` this function can fail, which is indicated by returning `nothing`.
+On exotic variants of `BLAS` this function can fail.
 """
-function set_num_threads(n::Integer)
+function set_num_threads(n::Integer)::Nothing
     blas = vendor()
     if blas === :openblas || blas == :openblas64
-        return ccall((@blasfunc(openblas_set_num_threads), libblas), Cvoid, (Int32,), n)
+        return ccall((@blasfunc(openblas_set_num_threads), libblas), Cvoid, (Cint,), n)
     elseif blas === :mkl
         # MKL may let us set the number of threads in several ways
         return ccall((:MKL_Set_Num_Threads, libblas), Cvoid, (Cint,), n)
@@ -129,7 +129,6 @@ function set_num_threads(n::Integer)
     else
         @warn "Failed to set number of BLAS threads." maxlog=1
     end
-
     return nothing
 end
 
@@ -151,16 +150,16 @@ Get the number of threads the BLAS library is using.
 
 On exotic variants of `BLAS` this function can fail, which is indicated by returning `nothing`.
 """
-function get_num_threads()
+function get_num_threads()::Union{Int, Nothing}
     blas = LinearAlgebra.BLAS.vendor()
     if blas === :openblas || blas === :openblas64
-        return ccall((@blasfunc(openblas_get_num_threads), libblas), Cint, ())
+        return Int(ccall((@blasfunc(openblas_get_num_threads), libblas), Cint, ()))
     elseif blas == :mkl
-        return ccall((:mkl_get_max_threads, libblas), Cint, ())
+        return Int(ccall((:mkl_get_max_threads, libblas), Cint, ()))
     elseif Sys.isapple()
         key = "VECLIB_MAXIMUM_THREADS"
         s = get(ENV, key, "")
-        nt = Base.tryparse(Cint, s)
+        nt = Base.tryparse(Int, s)
         if nt === nothing
             @warn "Failed to read environment variable $key" maxlog=1
         else
--- a/stdlib/LinearAlgebra/src/blas.jl
+++ b/stdlib/LinearAlgebra/src/blas.jl
@@ -110,7 +110,7 @@ function guess_vendor()
     # like determine_vendor, but guesses blas in some cases
     # where determine_vendor returns :unknown
     ret = vendor()
-    if Sys.isapple() && (ret == :unknown )
+    if Sys.isapple() && (ret == :unknown)
         ret = :osxblas
     end
     ret
--- a/stdlib/LinearAlgebra/src/blas.jl
+++ b/stdlib/LinearAlgebra/src/blas.jl
@@ -113,10 +113,8 @@ Set the number of threads the BLAS library should use.
 """
 function set_num_threads(n::Integer)
     blas = vendor()
-    if blas === :openblas
-        return ccall((:openblas_set_num_threads, libblas), Cvoid, (Int32,), n)
-    elseif blas === :openblas64
-        return ccall((:openblas_set_num_threads64_, libblas), Cvoid, (Int32,), n)
+    if blas === :openblas || blas == :openblas64
+        return ccall((@blasfunc(openblas_set_num_threads), libblas), Cvoid, (Int32,), n)
     elseif blas === :mkl
         # MKL may let us set the number of threads in several ways
         return ccall((:MKL_Set_Num_Threads, libblas), Cvoid, (Cint,), n)
@@ -128,6 +126,24 @@ function set_num_threads(n::Integer)
     end
 
     return nothing
+end
+
+"""
+    get_num_threads()
+
+Get the number of threads the BLAS library is using.
+"""
+function get_num_threads()
+    blas = LinearAlgebra.BLAS.vendor()
+    if blas === :openblas || blas === :openblas64
+        return ccall((@blasfunc(openblas_get_num_threads), libblas), Cint, ())
+    elseif blas == :mkl
+        return ccall((:MKL_Get_Max_Num_Threads, libblas), Cint, ())
+    elseif Sys.isapple()
+        return Base.parse(Cint, ENV["VECLIB_MAXIMUM_THREADS"])
+    else
+        error("Unknown BLAS") # better error? return nothing
+    end
 end
 
 const _testmat = [1.0 0.0; 0.0 -1.0]
--- a/stdlib/LinearAlgebra/src/blas.jl
+++ b/stdlib/LinearAlgebra/src/blas.jl
@@ -162,7 +162,7 @@ function get_num_threads()
         end
     end
     ret = nothing
-    @warn "Could not get number of BLAS threads. Returning $ret instead." maxlog=1
+    @warn "Could not get number of BLAS threads. Returning `$ret` instead." maxlog=1
     return ret
 end
 
--- a/stdlib/LinearAlgebra/src/blas.jl
+++ b/stdlib/LinearAlgebra/src/blas.jl
@@ -118,11 +118,11 @@ function set_num_threads(n::Integer)
     elseif blas === :mkl
         # MKL may let us set the number of threads in several ways
         return ccall((:MKL_Set_Num_Threads, libblas), Cvoid, (Cint,), n)
-    end
-
-    # OSX BLAS looks at an environment variable
-    @static if Sys.isapple()
+    elseif Sys.isapple()
+        # OSX BLAS looks at an environment variable
         ENV["VECLIB_MAXIMUM_THREADS"] = n
+    else
+        @warn "Failed to set number of BLAS threads."
     end
 
     return nothing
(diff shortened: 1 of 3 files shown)


stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -113,10 +113,8 @@ Set the number of threads the BLAS library should use.
 """
 function set_num_threads(n::Integer)
```
jw3126, line 114: I just noticed that on exotic BLAS this fails silently. Should  we throw an error instead?

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -130,6 +128,24 @@ function set_num_threads(n::Integer)
...
+function get_num_threads()
+    blas = LinearAlgebra.BLAS.vendor()
+    if blas === :openblas || blas === :openblas64
+        return ccall((@blasfunc(openblas_get_num_threads), libblas), Cint, ())
+    elseif blas == :mkl
+        return ccall((:MKL_Get_Max_Num_Threads, libblas), Cint, ())
+    elseif Sys.isapple()
+        return Base.parse(Cint, ENV["VECLIB_MAXIMUM_THREADS"])
+    else
+        error("Unknown BLAS") # better error? return nothing
```
jw3126, line 145: Suggesions for better error?

stdlib/LinearAlgebra/test/blas.jl
```diff
@@ -553,4 +553,15 @@ Base.stride(A::WrappedArray, i::Int) = stride(A.A, i)
     end
 end
 
+@testset "get_set_num_threads" begin
+    default = BLAS.get_num_threads()
+    @test default isa Integer
+    @test default > 0
+    new = rand(1:10)
```
OkonSamuel, line 560: consider `new=rand(1:Sys.CPU_THREADS)`

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -130,6 +128,24 @@ function set_num_threads(n::Integer)
...
 end
 
+"""
+    get_num_threads()
+
+Get the number of threads the BLAS library is using.
+"""
+function get_num_threads()
+    blas = LinearAlgebra.BLAS.vendor()
+    if blas === :openblas || blas === :openblas64
```
OkonSamuel, line 138: replace 
```julia
if blas === :openblas || blas === :openblas64
        return ccall((@blasfunc(openblas_get_num_threads), libblas), Cint, ())
    elseif blas == :mkl
        return ccall((:mkl_get_max_threads, libblas), Cint, ())
    elseif Sys.isapple()
        return Base.parse(Cint, ENV["VECLIB_MAXIMUM_THREADS"])
    else
        error("Unknown BLAS") # better error? return n
```
with
```julia
if blas === :openblas || blas === :openblas64
        return ccall((@blasfunc(openblas_get_num_threads), libblas), Cint, ())
    elseif blas == :mkl
        return ccall((:mkl_get_max_threads, libblas), Cint, ())
   end 
   @static if Sys.isapple()
        return Base.parse(Cint, ENV["VECLIB_MAXIMUM_THREADS"])
    end

    error("Unknown BLAS") # better error? return n
```

OkonSamuel: I don't know if the error would be needed since we covered all cases in `set_num_threads`

jw3126: Ok, can you explain why?

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -130,6 +128,24 @@ function set_num_threads(n::Integer)
...
+        return ccall((@blasfunc(openblas_get_num_threads), libblas), Cint, ())
+    elseif blas == :mkl
+        return ccall((:mkl_get_max_threads, libblas), Cint, ())
+    elseif Sys.isapple()
+        return Base.parse(Cint, ENV["VECLIB_MAXIMUM_THREADS"])
```
mcabbott, line 144: This doesn't run for me, on a mac.

stdlib/LinearAlgebra/test/blas.jl
```diff
@@ -553,4 +553,15 @@ Base.stride(A::WrappedArray, i::Int) = stride(A.A, i)
     end
 end
 
+@testset "get_set_num_threads" begin
+    default = BLAS.get_num_threads()
+    @test default isa Integer
+    @test default > 0
+    new=rand(1:Sys.CPU_THREADS)
```
mcabbott, line 560: `Sys.CPU_THREADS == 4` but with 2 physical cores I can't set more than 2 threads.

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -130,6 +128,24 @@ function set_num_threads(n::Integer)
...
+    else
+        error("Unknown BLAS") # better error? return nothing
```
mcabbott: I don't know whether other options are possible, but if they are, `set_num_threads` won't do anything. I think this should just return some integer, not an error. Could be `Sys.CPU_THREADS`.

stdlib/LinearAlgebra/src/blas.jl
OkonSamuel: `@static if Sys.isapple()` will evaluate the block at parse time when `Sys.isapple()` is true. The code block won't exist when `Sys.isapple()` is false

jw3126: Yeah, but why is this important that the code block disappears at macro expansion time?

jw3126: Is this a performance optimization?

mcabbott (commented):
This would be nice to have. I think the obvious use is 
```
old = get_num_threads()
@threads for i in 1:100
    # use *
end
set_num_threads(old)
```
If it's possible to have Julia without one of the libraries handled by `set_num_threads`, then I think this should do nothing, and not give an error.

stdlib/LinearAlgebra/src/blas.jl
jw3126: I like this part from the Zen of python
```
Errors should never pass silently.
Unless explicitly silenced.
In the face of ambiguity, refuse the temptation to guess.
```
If we want to break it, that needs justification. I think the user should just do
```julia
try
    BLAS.get_num_threads()
...
```
BTW for `BLAS.set_num_threads` it is broken, so maybe there are good reasons, that I am not aware of.

stdlib/LinearAlgebra/src/blas.jl
OkonSamuel: It just so just incase the block call code that's only valid for a given OS.

stdlib/LinearAlgebra/src/blas.jl
OkonSamuel: try this
```julia
@static if Sys.isapple()
        ntreads = Base.tryparse(Cint, get(ENV, "VECLIB_MAXIMUM_THREADS", "$(Sys.CPU_THREADS)"))
        nthreads isa Nothing && return Sys.CPU_THREADS
        return nthreads
    end
```

jw3126: what do you mean?

stdlib/LinearAlgebra/src/blas.jl
jw3126: > This would be nice to have. I think the obvious use is
> 
> ```
> old = get_num_threads()
> @threads for i in 1:100
>     # use *
> end
> set_num_threads(old)
> ```
> 
> If it's possible to have Julia without one of the libraries handled by `set_num_threads`, then I think this should do nothing, and not give an error.

How about 
```julia
BLAS.with_num_threads(1, ignore_failure=true) do
    ...
end
```

stdlib/LinearAlgebra/src/blas.jl
OkonSamuel: i was referring to @mcabbott to check if the updated code above runs on his mac

jw3126: Oh sorry. It is still magical to me, that this block needs to be elided on mac. Can somebody explain what exactly fails on mac?

stdlib/LinearAlgebra/src/blas.jl
mcabbott: What's the use case where an error would be better? 

In my example above there's no ambiguity, in the sense that setting the number of threads controls speed but not correctness. And it would be pretty annoying to need two try/catch blocks. 

I presume that making `set_num_threads` give an error would also count as a breaking change (if this can every happen).

stdlib/LinearAlgebra/test/blas.jl
jw3126: So I guess, `new = 1` is the only safe thing here?

mcabbott: Perhaps `div(Sys.CPU_THREADS, 2)` too? But `1` would test it.

stdlib/LinearAlgebra/src/blas.jl
mcabbott: Why is this this block needed at all?

As far as I know, `@static if` is only useful to shield things like macros which aren't defined.

stdlib/LinearAlgebra/src/blas.jl
jw3126: This error helps to debug why performance is bad. Otherwise, you think BLAS threading is turned off, so you conclude the problem is with your `Threads.@threads`. Depending on the setting bad performance is a bug as severe as incorrectness.

stdlib/LinearAlgebra/src/blas.jl
jw3126: So what does it shield in the case at hand?

OkonSamuel: That option is what's used in existing `set_num_threads` to set number of threads for MacOS

OkonSamuel: > So what does it shield in the case at hand?

@jw3126. I guess you could skip the `@static` and write it this way instead.(since all constructs are valid on all OS platforms)
```julia
if Sys.isapple()
        ntreads = Base.tryparse(Cint, get(ENV, "VECLIB_MAXIMUM_THREADS", "$(Sys.CPU_THREADS)"))
        nthreads isa Nothing && return Sys.CPU_THREADS
        return nthreads
end
```

stdlib/LinearAlgebra/src/blas.jl
jw3126: How about raising a warning instead of an error?

stdlib/LinearAlgebra/src/blas.jl
mcabbott: Still seems a bit baroque to parse "Sys.CPU_THREADS" as a string. Shouldn't it be more like 
```
elseif Sys.isapple()
    val = get(ENV, "VECLIB_MAXIMUM_THREADS", nothing)
    isnothing(val) || return parse(Cint, val)
end

@warn "unable to read number of BLAS threads" maxlog=1
return Sys.CPU_THREADS
```

stdlib/LinearAlgebra/src/blas.jl
mcabbott: There are lots of slow fallback methods which don't give errors. It could give a warning, I guess. 

There's some precedent for returning `nothing`, which would be fine for save-and-reset as long as there's a method like this:
```
set_num_threads(::Nothing) = set_num_threads(Sys.CPU_THREADS)
```

tkf: I think error or `return nothing` is better than returning `Sys.CPU_THREADS`. It's better to provide a way to let the user know that `get_num_threads` failed in some sense.  I'm slightly inclined to `nothing` since it's easier to compose.  A negative side is that, _if_ we can figure out better error information, there is no way to propagate this programmatically.  But I think the chance we get to this stage is low.

Also, I don't think defining `set_num_threads(::Nothing)` is the best option.  For example, the user may have access to the number of physical CPUs (e.g., calling `lscpu`).  Also, it's trivial to write `set_num_threads(something(old, Sys.CPU_THREADS))` or `old === nothing || bset_num_threads(old)`.

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -113,23 +113,46 @@ Set the number of threads the BLAS library should use.
...
         return ccall((:MKL_Set_Num_Threads, libblas), Cvoid, (Cint,), n)
-    end
-
-    # OSX BLAS looks at an environment variable
-    @static if Sys.isapple()
+    elseif Sys.isapple()
+        # OSX BLAS looks at an environment variable
         ENV["VECLIB_MAXIMUM_THREADS"] = n
+    else
+        @warn "Failed to set number of BLAS threads."
```
tkf, line 125:
```suggestion
        @warn "Failed to set number of BLAS threads." maxlog=1
```

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -113,23 +113,46 @@ Set the number of threads the BLAS library should use.
...
+    if blas === :openblas || blas === :openblas64
+        return ccall((@blasfunc(openblas_get_num_threads), libblas), Cint, ())
+    elseif blas == :mkl
+        return ccall((:mkl_get_max_threads, libblas), Cint, ())
+    elseif Sys.isapple()
+        key = "VECLIB_MAXIMUM_THREADS"
+        s = get(ENV, key, "")
+        nt = Base.tryparse(Cint, s)
+        if nt === nothing
+            @warn "Failed to read environment variable $key"
```
tkf, line 147:
```suggestion
            @warn "Failed to read environment variable $key" maxlog=1
```

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -113,23 +113,46 @@ Set the number of threads the BLAS library should use.
...
+        else
+            return nt
+        end
+    end
+    @warn "Could not get number of BLAS threads. Returning Sys.CPU_THREADS instead."
```
tkf, line 152:
```suggestion
    @warn "Could not get number of BLAS threads. Returning Sys.CPU_THREADS instead." maxlog=1
```

stdlib/LinearAlgebra/src/blas.jl
mcabbott: It would be a little weird to accept `nothing`, but `set_num_threads(something(old, Sys.CPU_THREADS))` also feels like easy-to-forget boilerplate for the off-chance that someone has a non-standard library (which your tests won't have). For the use I outlined, of my package re-setting it after the loop. If you know more, then it needn't get in the way.

tkf: I don't think `set_num_threads(::Nothing) = set_num_threads(Sys.CPU_THREADS)` is weird. It'd be a good API if `Sys.CPU_THREADS` was the best choice. But we know that using the number of physical cores and respecting cgroup would be a better choice here. Having said that, I'm less opposed to defining `set_num_threads(::Nothing)` than returning `Sys.CPU_THREADS` if we clearly document that `set_num_threads(::Nothing)` will always be a best-effort and the exact number that would be set is undefined.

mcabbott: There's no easy way to get the number of cores, right? 

Openblas lets me set 32 threads, while MKL restricts to 2 (on a 2/4 processor). But for both of these, you can save the default first. Not sure whether the ENV variable will always be set for Apple Accelerate. For even more exotic libraries, `set_num_threads` won't do anything anyway (right now).

OkonSamuel: > There's no easy way to get the number of cores, right?

Yes it's not easy. I think https://github.com/JuliaParallel/Hwloc.jl does give this. But it would be nice to have this option in Base.

tkf: I think it's straightforward in Linux. You just have to parse some text files in `/proc` and `/sys`.  Using CPUID might be the best fallback option for all OSes.  Hwloc.jl is probably not going to be in Base or stdlib.  See #33409

I think it has to be written in C (at least for now) if we want to use it also for `Threads.nthreads()`.  But nobody wants to touch it (so far).

mcabbott: OK. I guess if that gets solved, this can follow later. For now I guess `Threads.nthreads()` and `div(Sys.CPU_THREADS,2)` are other possible defaults.

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -107,29 +107,65 @@ end
...
+        s = get(ENV, key, "")
+        nt = Base.tryparse(Cint, s)
+        if nt === nothing
+            @warn "Failed to read environment variable $key" maxlog=1
+        else
+            return nt
+        end
+    end
+    ret = nothing
+    @warn "Could not get number of BLAS threads. Returning $ret instead." maxlog=1
```
tkf, line 165:
```suggestion
    @warn "Could not get number of BLAS threads. Returning `$ret` instead." maxlog=1
```

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -107,29 +107,65 @@ end
...
+        s = get(ENV, key, "")
+        nt = Base.tryparse(Cint, s)
+        if nt === nothing
+            @warn "Failed to read environment variable $key" maxlog=1
+        else
+            return nt
+        end
+    end
+    ret = nothing
+    @warn "Could not get number of BLAS threads. Returning `$ret` instead." maxlog=1
```
OkonSamuel, line 165:
```suggestion
    @warn "Could not get number of BLAS threads. Returning `nothing` instead." maxlog=1
```

OkonSamuel: since `ret` is always nothing in this case we can drop the interpolation.

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -107,29 +107,65 @@ end
...
+        @warn "Failed to set number of BLAS threads." maxlog=1
     end
 
     return nothing
 end
 
+function set_num_threads(::Nothing)
+    n = max(1, Sys.CPU_THREADS ÷ 2)
+    set_num_threads(n)
+end
```
tkf, line 139: I'm a bit uneasy about `Sys.CPU_THREADS ÷ 2` (e.g., the user may have hyperthreading disabled). But I'm not sure how much this code path would be used...

Also, it makes sense to use whatever the logic used in `Base.__init__`

https://github.com/JuliaLang/julia/blob/35c1f87176fecab5a8e8077fb3a62275363a5aa5/base/Base.jl#L380-L388

So, maybe it's better to read the environment variable (e.g., if the user or  `Base.__init__` sets this environment variable, it's going to be used in `set_num_threads`).

```suggestion
_tryparse_env_cint(key) = tryparse(Cint, get(ENV, key, ""))

function set_num_threads(::Nothing)
    n = something(
        _tryparse_env_cint("OPENBLAS_NUM_THREADS"),
        _tryparse_env_cint("OMP_NUM_THREADS"),
        max(1, Sys.CPU_THREADS ÷ 2),
    )
    set_num_threads(n)
end
```

(or maybe without `÷ 2`)

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -107,29 +107,65 @@ end
...
+        nt = Base.tryparse(Cint, s)
+        if nt === nothing
+            @warn "Failed to read environment variable $key" maxlog=1
+        else
+            return nt
+        end
+    end
+    ret = nothing
+    @warn "Could not get number of BLAS threads. Returning `$ret` instead." maxlog=1
+    return ret
```
tkf, line 166: I think it's nice that the current code ties the implementation and the warning message. But I fear other core devs might not like "unnecessary" dynamism (as @OkonSamuel pointed out).  I think it might be safe to just do:

```suggestion
    @warn "Could not get number of BLAS threads. Returning `nothing` instead." maxlog=1
    return nothing
```

stdlib/LinearAlgebra/src/blas.jl
tkf: If we go with it, we should just do `return nothing`. See: https://github.com/JuliaLang/julia/pull/36360#discussion_r443108316

stdlib/LinearAlgebra/src/blas.jl
OkonSamuel: I think the fallback to ` max(1, Sys.CPU_THREADS ÷ 2)` is good as a last resort until julia gets functionality for determining number of physical cores (Who knows this might be sooner than later)

stdlib/LinearAlgebra/src/blas.jl
mcabbott: If we go with returning `nothing`, should there be any warnings at all? 

I thought these two were complementary approaches, either return something clearly marking the failure to read how many threads, or else return an integer & warn that it may not be correct.

jw3126: I think a warning won't do any hurt.

tkf: If this is used as

```julia
old = get_num_threads()
# ...
set_num_threads(old)
```

then the end-user (which may not be who writes the function) cannot notice something fishy is going on unless there is a warning.

Also, I'm guessing this code pass is rarely used. So, I guess it's OK to make things verbose until somebody complains.  It's also possible for a user to suppress the warning.

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -107,26 +107,66 @@ end
...
+    elseif Sys.isapple()
+        # OSX BLAS looks at an environment variable
         ENV["VECLIB_MAXIMUM_THREADS"] = n
+    else
+        @warn "Failed to set number of BLAS threads." maxlog=1
     end
+    return nothing
+end
+
+_tryparse_env_cint(key) = tryparse(Cint, get(ENV, key, ""))
```
tkf, line 135:
```suggestion
_tryparse_env_int(key) = tryparse(Int, get(ENV, key, ""))
```

tkf, line 135:
```suggestion
_tryparse_env_int(key) = tryparse(Int, get(ENV, key, ""))
```

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -107,26 +107,66 @@ end
...
+
+function set_num_threads(::Nothing)
+    n = something(
+        _tryparse_env_cint("OPENBLAS_NUM_THREADS"),
+        _tryparse_env_cint("OMP_NUM_THREADS"),
```
tkf, line 140:
```suggestion
        _tryparse_env_int("OPENBLAS_NUM_THREADS"),
        _tryparse_env_int("OMP_NUM_THREADS"),
```

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -107,26 +107,66 @@ end
...
+function get_num_threads()::Union{Int, Nothing}
+    blas = LinearAlgebra.BLAS.vendor()
+    if blas === :openblas || blas === :openblas64
+        return Int(ccall((@blasfunc(openblas_get_num_threads), libblas), Cint, ()))
+    elseif blas == :mkl
+        return Int(ccall((:mkl_get_max_threads, libblas), Cint, ()))
+    elseif Sys.isapple()
+        key = "VECLIB_MAXIMUM_THREADS"
+        s = get(ENV, key, "")
+        nt = Base.tryparse(Int, s)
```
tkf, line 162:
```suggestion
        key = "VECLIB_MAXIMUM_THREADS"
        nt = _tryparse_env_int(key)
```

tkf (commented):
We can refactor this a bit more.

stdlib/LinearAlgebra/src/blas.jl
mcabbott: But this is a performance optimisation, those tend to give up silently. (Although I wish more of them had `@debug` statements.) That said I agree it shouldn't be triggered often.

If there is to be a warning, just one is enough, not two.

tkf: I think the question is if it is OK to consider changing number of threads as _pure_ optimization.  It changes the result slightly:

```julia
julia> A = randn(300, 300); B = randn(300, 300);

julia> LinearAlgebra.BLAS.set_num_threads(1)

julia> C1 = @btime $A * $B;
  1.112 ms (2 allocations: 703.20 KiB)

julia> LinearAlgebra.BLAS.set_num_threads(2)

julia> C2 = @btime $A * $B;
  650.866 μs (2 allocations: 703.20 KiB)

julia> extrema(C1 .- C2)
(-3.552713678800501e-14, 4.263256414560601e-14)
```

I think it is worth notifying the user that something unexpected is going on.

mcabbott: Perhaps one warning when `set` fails, and nothing on `get`, is the right way to go?

tkf: So I guess you mean to merge the two warnings for macOS? But, since this code path would be rarely exercised, I think it's better to go with simplicity than a nicely formatted warning.

tkf (approved):
LGTM

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -107,26 +107,65 @@ end
...
+
+"""
+    get_num_threads()
 
+Get the number of threads the BLAS library is using.
+
+On exotic variants of `BLAS` this function can fail, which is indicated by returning `nothing`.
+"""
+function get_num_threads()::Union{Int, Nothing}
+    blas = LinearAlgebra.BLAS.vendor()
```
dkarrasch, line 154: Is this different from just `vendor()`, i.e., without specifier?

dkarrasch (commented):
LGTM, just one minor comment. In our CI systems, do we cover all the `vendor` and/or OS cases that we handle here? Like the `isapple` case that is not covered by the `:openblas*` and `:mkl` branches?

stdlib/LinearAlgebra/src/blas.jl
OkonSamuel: It's the same so should just have been `vendor()`

OkonSamuel:
```suggestion
    blas = vendor()
```

OkonSamuel:
> do we cover all the `vendor` and/or OS cases that we handle here? Like the `isapple` case that is not covered by the `:openblas*` and `:mkl` branches?

I think sticking to the recognized cases in `vendor()` ( `:openblas*` and `:mkl`) on CI is fine since they are very common. The `isapple`(apples accelerate BLAS) case is considered `:unknown` by `vendor()` but `set_num_threads` and `get_num_threads` takes that case into account for users who specifically may prefer using BLAS provided by apple's accelerate framework.

OkonSamuel:
LGTM except for the slight suggestion above

OkonSamuel (approved):

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -107,26 +107,65 @@ end
...
 
+"""
+    get_num_threads()
+
+Get the number of threads the BLAS library is using.
+
+On exotic variants of `BLAS` this function can fail, which is indicated by returning `nothing`.
+"""
+function get_num_threads()::Union{Int, Nothing}
+    blas = vendor()
```
tkf, line 154:
```suggestion
get_num_threads()::Union{Int, Nothing} = _get_num_threads()

function _get_num_threads(; _vendor = vendor, _isapple = Sys.isapple)
    blas = _vendor()
```

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -107,26 +107,65 @@ end
...
+    if blas === :openblas || blas === :openblas64
+        return Int(ccall((@blasfunc(openblas_get_num_threads), libblas), Cint, ()))
+    elseif blas == :mkl
+        return Int(ccall((:mkl_get_max_threads, libblas), Cint, ()))
+    elseif Sys.isapple()
```
tkf, line 159:
```suggestion
    elseif _isapple()
```

tkf (commented):
> In our CI systems, do we cover all the `vendor` and/or OS cases that we handle here?

One way to test the "exotic" branches to make the implementation patchable by writing something like this and calling `_get_num_threads(blas = :unknown, _isapple = true)` in the test.

We can't cover the mkl route because it needs the C API to exist. But we can cover other paths.

stdlib/LinearAlgebra/src/blas.jl
jw3126: I also think we should not overengineer this rare case. About silent fails of performance optimization, I think it makes a difference if an API level function fails at its one job or if some "optimization detail" burried deep in a long algorithm fails.

stdlib/LinearAlgebra/src/blas.jl
jw3126: Mhh we could do this, but is it a desirable tradeoff? It makes the code less clear and does it really give a useful test? Seems a bit like testing for the sake of testing.

tkf: Yeah, I get it that looks stupid to have "testing for the sake of testing" (and it's not like I'm 100% certain this is the Right Thing™ to do).  But it's plausible that

```julia
key = "VECLIB_MAXIMUM_THREADS"
nt = _tryparse_env_int(key)
```

is "refactored" into

```julia
nt = _tryparse_env_int("VECLIB_MAXIMUM_THREADS")
```

unless everyone is careful enough to notice `$key` is used in `@warn`.  I find it unsatisfactory that the only way to avoid a bug in stdlib is "be careful."

Mocking/monkey-patching during test is a thing in other languages so I think it would have been a nobrainer to test this if Julia had a good monkey-patching support.  Since making `get_num_threads` "patchable" is not difficult, I don't see too much downside in this direction.

jw3126: I added a something in this direction, can you take a look if you like it?

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -106,27 +106,78 @@ end
...
+On exotic variants of `BLAS` this function can fail.
 """
-function set_num_threads(n::Integer)
-    blas = vendor()
-    if blas === :openblas
-        return ccall((:openblas_set_num_threads, libblas), Cvoid, (Int32,), n)
-    elseif blas === :openblas64
-        return ccall((:openblas_set_num_threads64_, libblas), Cvoid, (Int32,), n)
-    elseif blas === :mkl
+function set_num_threads(n::Integer, _blas=guess_vendor())::Nothing
```
OkonSamuel, line 130: is there any reason for exposing the `_blas` argument here. 
Why not
```julia
function set_num_threads(n::Integer)::Nothing
         _blas=guess_vendor()
   ....
```

jw3126: Yeah, its for testing. We can simulate `:osxblas` this way, which is otherwise not covered by CI. See the above discussion with @tkf.

OkonSamuel: in that case why not make `_blas` a kwarg instead

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -106,27 +106,78 @@ end
...
+On exotic variants of `BLAS` this function can fail.
 """
-function set_num_threads(n::Integer)
-    blas = vendor()
-    if blas === :openblas
-        return ccall((:openblas_set_num_threads, libblas), Cvoid, (Int32,), n)
-    elseif blas === :openblas64
-        return ccall((:openblas_set_num_threads64_, libblas), Cvoid, (Int32,), n)
-    elseif blas === :mkl
+function set_num_threads(n::Integer; _blas=guess_vendor())::Nothing
```
tkf, line 130:
```suggestion
set_num_threads(n)::Nothing = _set_num_threads(n)

function _set_num_threads(n::Integer; _blas=guess_vendor())
```

tkf, line 130:
```suggestion
set_num_threads(n)::Nothing = _set_num_threads(n)

function _set_num_threads(n::Integer; _blas = guess_vendor())
```

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -106,27 +106,78 @@ end
...
+    else
+        @assert _blas === :unknown
+        @warn "Failed to set number of BLAS threads." maxlog=1
     end
+    return nothing
+end
 
+_tryparse_env_int(key) = tryparse(Int, get(ENV, key, ""))
+
+function set_num_threads(::Nothing; _blas=guess_vendor())
```
tkf, line 148:
```suggestion
function _set_num_threads(::Nothing; _blas = guess_vendor())
```

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -106,27 +106,78 @@ end
...
+end
+
+"""
+    get_num_threads()
+
+Get the number of threads the BLAS library is using.
+
+On exotic variants of `BLAS` this function can fail, which is indicated by returning `nothing`.
+"""
+function get_num_threads(;_blas=guess_vendor())::Union{Int, Nothing}
```
tkf, line 164:
```suggestion
get_num_threads(;_blas=guess_vendor())::Union{Int, Nothing} = _get_num_threads()

function _get_num_threads(; _blas = guess_vendor())::Union{Int, Nothing}
```

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -106,27 +106,78 @@ end
 
 openblas_get_config() = strip(unsafe_string(ccall((@blasfunc(openblas_get_config), libblas), Ptr{UInt8}, () )))
 
+function guess_vendor()
+    # like determine_vendor, but guesses blas in some cases
+    # where determine_vendor returns :unknown
+    ret = vendor()
+    if Sys.isapple() && (ret == :unknown )
```
tkf, line 113:
```suggestion
    if Sys.isapple() && (ret == :unknown)
```

stdlib/LinearAlgebra/test/blas.jl
```diff
@@ -553,4 +553,22 @@ Base.stride(A::WrappedArray, i::Int) = stride(A.A, i)
...
+    @test BLAS.get_num_threads() === 1
+    BLAS.set_num_threads(default)
+    @test BLAS.get_num_threads() === default
+
+    if BLAS.guess_vendor() !== :osxblas
+        # test osxblas which is not covered by CI
+        BLAS.set_num_threads(1, _blas=:osxblas)
+        @test BLAS.get_num_threads(_blas=:osxblas) === 1
+        BLAS.set_num_threads(2, _blas=:osxblas)
+        @test BLAS.get_num_threads(_blas=:osxblas) === 2
```
tkf, line 570: Using `@test_logs` so that we don't get suspicious warnings in the test logs.

```suggestion
        withenv() do
            @test_logs (:warn,) match_mode=:any BLAS._set_num_threads(1, _blas=:osxblas)
            @test @test_logs(
                (:warn,),
                (:warn,),
                match_mode=:any,
                BLAS._get_num_threads(_blas=:osxblas),
            ) === 1
            @test_logs (:warn,) match_mode=:any BLAS._set_num_threads(2, _blas=:osxblas)
            @test @test_logs(
                (:warn,),
                (:warn,),
                match_mode=:any,
                BLAS._get_num_threads(_blas=:osxblas),
            ) === 2
        end
```

tkf (commented):
Let's avoid exposing the internal arguments at the interface.

(Underscoring the keyword argument is OK. But we can do better, especially for a stdlib.)

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -106,27 +106,82 @@ end
...
+        ret = :osxblas
+    end
+    ret
+end
+
+
 """
-    set_num_threads(n)
+    set_num_threads(n::Integer)
+    set_num_threads(::Nothing)
```
fredrikekre, line 122: This seems like an unusual API, I would prefer `set_num_threads()` instead of `set_num_threads(nothing)`.

stdlib/LinearAlgebra/test/blas.jl
```diff
@@ -553,4 +553,34 @@ Base.stride(A::WrappedArray, i::Int) = stride(A.A, i)
...
 
+@testset "get_set_num_threads" begin
+    default = BLAS.get_num_threads()
+    @test default isa Int
+    @test default > 0
+    BLAS.set_num_threads(1)
+    @test BLAS.get_num_threads() === 1
+    BLAS.set_num_threads(default)
+    @test BLAS.get_num_threads() === default
+
```
tkf, line 564:
```suggestion

    @test_logs (:warn,) match_mode=:any BLAS._set_num_threads(1, _blas=:unknown)
```

stdlib/LinearAlgebra/test/blas.jl
```diff
@@ -553,4 +553,34 @@ Base.stride(A::WrappedArray, i::Int) = stride(A.A, i)
...
+                match_mode=:any,
+                BLAS._get_num_threads(_blas=:osxblas),
+            ) === 1
+            @test_logs (:warn,) match_mode=:any BLAS._set_num_threads(2, _blas=:osxblas)
+            @test @test_logs(
+                (:warn,),
+                (:warn,),
+                match_mode=:any,
+                BLAS._get_num_threads(_blas=:osxblas),
+            ) === 2
```
tkf, line 581:
```suggestion
        withenv("VECLIB_MAXIMUM_THREADS" => nothing) do
            @test @test_logs(
                (:warn,),
                (:warn,),
                match_mode=:any,
                BLAS._get_num_threads(_blas=:osxblas),
            ) === nothing
            @test_logs BLAS._set_num_threads(1, _blas=:osxblas)
            @test @test_logs(BLAS._get_num_threads(_blas=:osxblas)) === 1
            @test_logs BLAS._set_num_threads(2, _blas=:osxblas)
            @test @test_logs(BLAS._get_num_threads(_blas=:osxblas)) === 2
```

stdlib/LinearAlgebra/src/blas.jl
jw3126: This is to allow the pattern
```julia
default = BLAS.get_num_threads() # returns nothing on exotic platforms
BLAS.set_num_threads(1)
# do stuff
BLAS.set_num_threads(default)
```

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -106,27 +106,82 @@ end
...
 
+_tryparse_env_int(key) = tryparse(Int, get(ENV, key, ""))
+
+function _set_num_threads(::Nothing; _blas = guess_vendor())
+    n = something(
+        _tryparse_env_int("OPENBLAS_NUM_THREADS"),
+        _tryparse_env_int("OMP_NUM_THREADS"),
+        max(1, Sys.CPU_THREADS ÷ 2),
+    )
+    set_num_threads(n; _blas)
```
OkonSamuel, line 156:
```suggestion
    _set_num_threads(n; _blas)
```

stdlib/LinearAlgebra/src/blas.jl
fredrikekre: That will still work if you make the signature `set_num_threads(n=nothing)`

jw3126: To me allowing `nothing` is a hack to allow the above pattern on strange platforms. It is not a thing I would encourage or that I think needs more convenient syntax.

jw3126: I imagine if you have code like `set_num_threads()` it is more likely you forgot to pass the number of threads, than that you really want to invoce the `nothing` method.

tkf: I don't think we should define `set_num_threads() = set_num_threads(nothing)`. It would send a wrong message that `set_num_threads(nothing)` is somehow a reasonable default.  But it's not.  It is the last resort that exists only for supporting the rollback use case.

But this is not clear from the current docstring. I think it's better to clarify this.

mcabbott: Perhaps clearest to show the pattern which motivates this:
```
Set the number of threads the BLAS library should use.

Also accepts `nothing`, in which case it tries to set set the default number of threads.
On exotic variants of BLAS, `nothing` may be returned by ` get_num_threads()`.
Accepting it means that the following pattern will not give an error:

old = get_num_threads()
set_num_threads(1)
@threads for i in 1:10
    # single-threaded BLAS calls
end
set_num_threads(old)
```

tkf:
We need a news item for `BLAS.get_num_threads`. I think it's good to go.

stdlib/LinearAlgebra/src/blas.jl
tkf: @fredrikekre Does the argument here makes sense? It'd be nice if you can have a look at the docstring.

mcabbott:
Has anyone tried this with Apple's BLAS? 

I tried & failed to install this today, `make USE_SYSTEM_BLAS=1 USE_SYSTEM_LAPACK=0` led to various errors. Paging @dlfivefifty perhaps -- you once said on Discourse that you used this?

dlfivefifty:
I stopped using Apple BLAS around Julia v0.7: I think MKL is faster

jw3126:
Can somebody rerun the failing CI?

StefanKarpinski:
I tried restarting but it's not doing my bidding even though I'm logged in. Hitting the rebuild button just gives me a bunch of info about the PR and doesn't do anything. @staticfloat, any idea what's up with that?

staticfloat:
If you click `rebuild` and it shows you a screen like this:

image

That means that the build is queued, but hasn't started yet.  You can see at the top instead of `builds` it says `buildrequests`.  Eventually, the build will run and the status here on GH will change.  If you click on the `macos64` build here, you can see that it finished about 5 hours ago.

jw3126:
Can this be merged?

mcabbott:
This is late but notice (1) tests will fail if `get_num_threads()` returns `nothing`, as it claims to sometimes do, and (2) there seems to be no test which runs the `set_num_threads(::Nothing)` method. 
```
default = BLAS.get_num_threads()
@test default isa Int
```

jw3126:
> This is late but notice (1) tests will fail if `get_num_threads()` returns `nothing`, 

Arguably thats a desirable property. It means that the BLAS library is not properly supported.

> (2) there seems to be no test which runs the `set_num_threads(::Nothing)` method.

The exact number of threads set this way is an implementation detail and should not be tested. But it makes sense, that we just call the method and test that julia does not e.g. crash.

mcabbott:
It's documented to fail gracefully on weird BLAS libraries, so I would say they ought to pass tests.

jw3126:
My view is that `get_num_threads` returning nothing is a bug even if it is a documented bug. But if somebody really wants to add such a test I can also live with that.

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -106,27 +106,101 @@ end
...
+    get_num_threads()
 
+Get the number of threads the BLAS library is using.
+
+On exotic variants of `BLAS` this function can fail, which is indicated by returning `nothing`.
+
+!!! compat "Julia 1.6"
+    `get_num_threads` requires at least Julia 1.6.
+"""
+get_num_threads(;_blas=guess_vendor())::Union{Int, Nothing} = _get_num_threads()
```
giordano, line 185: Any reason why `_blas` isn't passed to `_get_num_threads`?

jw3126: Thanks for reporting, I will fix it!

PR #36364, opened 2020-06-19
Add Quarter period

matthieugomez:
This is the same as https://github.com/JuliaLang/julia/pull/35519, except that it is rebased (the only difference is to resolve the NEWS file conflict).  Sorry to open a different pull request but I cannot modify the original pull request anymore — not sure what I did wrong.


PallHaraldsson:
I was looking at the code, and noticed:

https://github.com/JuliaLang/julia/blob/e36b7b48692a82431cf2610cede6fc48ead63a71/stdlib/Dates/src/query.jl#L655

I think this is still wrong, for a leap year:
```
julia> dayofquarter(Date(2020, 4, 1))
2
```

matthieugomez:
Good point! I think you could open an issue about it --- this is not something I have introduced with my changes.

matthieugomez:
Thanks!

PaulSoderlind:
Did you forget adding the following to adjusters.jl, or was it intentional?:
```
Base.trunc(dt::Date, p::Type{Quarter}) = firstdayofquarter(dt)
Base.trunc(dt::DateTime, p::Type{Quarter}) = DateTime(trunc(Date(dt), Quarter))
```

matthieugomez:
Indeed, I forgot to add these methods. Could you open a pull request to add them?

PaulSoderlind:
done. https://github.com/JuliaLang/julia/pull/38277

Issue #36322, opened 2020-06-17
Passing through NamedTuple throws away known type info from @kwdef constructor


timholy:
Some new `convert` method definitions cause substantial invalidation due to usage of `convert` in constructors. A good example is [`Pkg.Operations.load_direct_deps`](https://github.com/JuliaLang/Pkg.jl/blob/3e3f9d7a3baf3a96c860d2b1bc895d5fd2d40d35/src/Operations.jl#L64-L83):

```julia
julia> using Pkg, Cthulhu

julia> descend(Pkg.Operations.load_direct_deps, (Pkg.Types.Context,))
```

From here, you can navigate to the "body method" where you get to the core issue:
```julia
│    %47 = uuid::Base.UUID
│    %48 = name::String
│    %49 = Base.getproperty(entry::Pkg.Types.PackageEntry, :path)::Union{Nothing, String}
│    %50 = Base.getproperty(entry::Pkg.Types.PackageEntry, :repo)::Pkg.Types.GitRepo
│    %51 = Base.getproperty(entry::Pkg.Types.PackageEntry, :pinned)::Bool
│    %52 = Base.getproperty(entry::Pkg.Types.PackageEntry, :tree_hash)::Union{Nothing, Base.SHA1}
│    %53 = Base.getproperty(entry::Pkg.Types.PackageEntry, :version)::Union{Nothing, VersionNumber}
│    %54 = Pkg.Operations.isfixed(entry::Pkg.Types.PackageEntry)::Bool
│    %55 = Pkg.Operations.load_version(%53, %54, preserve)::Union{Nothing, VersionNumber, Pkg.Types.VersionSpec}
│    %56 = Core.tuple(%47, %48, %49, %50, %51, %52, %55)::Tuple{Base.UUID,String,Union{Nothing, String},Pkg.Types.GitRepo,Bool,Union{Nothing, Base.SHA1},Union{Nothing, VersionNumber, Pkg.Types.VersionSpec}}
│    %57 = (%46)(%56)::NamedTuple{(:uuid, :name, :path, :repo, :pinned, :tree_hash, :version),_A} where _A<:Tuple
│    %58 = Core.kwfunc(Pkg.Operations.PackageSpec)::Core.Compiler.Const(Core.var"#Type##kw"(), false)
└───       (@_13 = (%58)(%57, Pkg.Operations.PackageSpec))
```
[...]

PR #33472, opened 2019-10-04
improve `instanceof_tfunc` to take declared parameter bounds into account

JeffBezanson:
This fixes the following inference performance issue identified by @andreasnoack :
```
julia> using DataFrames

julia> @time cumsum(rand(2, 2), dims=1);
 59.742860 seconds (72.02 M allocations: 4.530 GiB, 2.89% gc time)
```
(except on master it's even worse)

What happened here was that a type got inferred as `CartesianIndices{A,B} where {A,B}` --- there are no limits on A and B even though the type has some declared bounds. That in turn caused us to infer into methods that actually could not be applicable, because we thought `iterate(::CartesianIndices)` might not cover every case. That led to a lot of recursion that made it very hard to infer a lot of code.

It's a bit lucky that this ends up being enough to fix it. Does not help #33336.

PR #36366, opened 2020-06-19
fix #32699, more accurate apply_type_tfunc for NamedTuple

### LINKED ISSUE #32699: better upper bounds in apply_type_tfunc for NamedTuple
Here's an example where the resulting `NamedTuple{names, _A) where _A` type is not ideal since it completely discards valid information up to that point.
```julia
8 ── %21 = Base.getfield(%14, :id)::Array{Int64,1}
│    %22 = Base.arrayref(true, %21, %15)::Int64
│    %23 = Base.getfield(%16, :salary)::Array{Union{Missing, Float64},1}
│    %24 = Base.arrayref(true, %23, %17)::Union{Missing, Float64}
│    %25 = (isa)(%24, Missing)::Bool
└───       goto #10 if not %25
9 ──       goto #13
10 ─ %28 = (isa)(%24, Float64)::Bool
└───       goto #12 if not %28
11 ─ %30 = π (%24, Float64)
│    %31 = Base.sitofp(Float64, 2)::Float64
│    %32 = Base.mul_float(%30, %31)::Float64
└───       goto #13
12 ─       Core.throw(ErrorException("fatal error in type inference (type bound)"))::Union{}
└───       $(Expr(:unreachable))::Union{}
13 ┄ %36 = φ (#9 => $(QuoteNode(missing)), #11 => %32)::Union{Missing, Float64}
│    %37 = Core.tuple(%22, %36)::Tuple{Int64,Union{Missing, Float64}}
│    %38 = (NamedTuple{(:id, :y),T} where T<:Tuple)(%37)::NamedTuple{(:id, :y),_A} where _A
```
@JeffBezanson thought it should be valid/fine to include an upper bound here, so that we end up with `NamedTuple{(:id, :y), _A} where _A <: Tuple{Int64, Union{Missing, Float64}`

JeffBezanson:
This can make inference a bit slower, so I tried limiting it to NamedTuple types with known names, to mitigate the effects a bit.

@quinnj @timholy See how this works for you.

fixes #32699

### DIFF
--- a/base/compiler/tfuncs.jl
+++ b/base/compiler/tfuncs.jl
@@ -587,7 +587,7 @@ function fieldcount_noerror(@nospecialize t)
     if !(t isa DataType)
         return nothing
     end
-    if t.name === NamedTuple.body.body.name
+    if t.name === _NAMEDTUPLE_NAME
         names, types = t.parameters
         if names isa Tuple
             return length(names)
@@ -1079,11 +1079,14 @@ function apply_type_tfunc(@nospecialize(headtypetype), @nospecialize args...)
     if !istuple && !isa(headtype, UnionAll)
         return Union{}
     end
+    uw = unwrap_unionall(headtype)
+    isnamedtuple = isa(uw, DataType) && uw.name === _NAMEDTUPLE_NAME
     uncertain = false
     canconst = true
     tparams = Any[]
     outervars = Any[]
     varnamectr = 1
+    ua = headtype
     for i = 1:largs
         ai = widenconditional(args[i])
         if isType(ai)
@@ -1105,8 +1108,12 @@ function apply_type_tfunc(@nospecialize(headtypetype), @nospecialize args...)
             #    ai = rename_unionall(ai)
             #    unw = unwrap_unionall(ai)
             #end
+            ai_w = widenconst(ai)
+            ub = ai_w isa Type && ai_w <: Type ? instanceof_tfunc(ai)[1] : Any
             if istuple
-                if i == largs
+                # in the last parameter of a Tuple type, if the upper bound is Any
+                # then this could be a Vararg type.
+                if i == largs && ub === Any
                     push!(tparams, Vararg)
                 # XXX
                 #elseif isT
@@ -1122,12 +1129,28 @@ function apply_type_tfunc(@nospecialize(headtypetype), @nospecialize args...)
             #        ai = ai.body
             #    end
             else
+                # Is this the second parameter to a NamedTuple?
+                if isnamedtuple && isa(ua, UnionAll) && uw.parameters[2] === ua.var
+                    # If the names are known, keep the upper bound, but otherwise widen to Tuple.
+                    # This is a widening heuristic to avoid keeping type information
+                    # that's unlikely to be useful.
+                    if !(uw.parameters[1] isa Tuple || (i == 2 && tparams[1] isa Tuple))
+                        ub = Any
+                    end
+                else
+                    ub = Any
+                end
                 tvname = varnamectr <= length(_tvarnames) ? _tvarnames[varnamectr] : :_Z
                 varnamectr += 1
-                v = TypeVar(tvname)
+                v = TypeVar(tvname, ub)
                 push!(tparams, v)
                 push!(outervars, v)
             end
+        end
+        if isa(ua, UnionAll)
+            ua = ua.body
+        else
+            ua = nothing
         end
     end
     local appl
--- a/test/compiler/inference.jl
+++ b/test/compiler/inference.jl
@@ -2678,3 +2678,10 @@ partial_return_2(x) = Val{partial_return_1(x)[2]}
 # Precision of abstract_iteration
 f_splat(x) = (x...,)
 @test Base.return_types(f_splat, (Pair{Int,Int},)) == Any[Tuple{Int, Int}]
+
+# issue #32699
+f32699(a) = (id = a[1],).id
+@test Base.return_types(f32699, (Vector{Union{Int,Missing}},)) == Any[Union{Int,Missing}]
+g32699(a) = Tuple{a}
+@test Base.return_types(g32699, (Type{<:Integer},))[1] == Type{<:Tuple{Any}}
+@test Base.return_types(g32699, (Type,))[1] == Type{<:Tuple}


quinnj:
image

timholy:
The inference looks great! In #36322 I failed to record which package's invalidations I was looking at. I think it was FilePathsBase, and I haven't seen `load_direct_deps` pop up, so I think it's likely this fixes it. Thanks!

quinnj:
Yep, the inference is looking super spiffy, though it seems the last part of my work-flow is failing to inline (`Tables.eachcolumns`):
```julia
│    %33 = Core.tuple(%20, %32)::Tuple{Int64,Union{Missing, Float64}}
│    %34 = (NamedTuple{(:x, :y),T} where T<:Tuple)(%33)::NamedTuple{(:x, :y),_A} where _A<:Tuple{Int64,Union{Missing, Float64}}
└───       goto #14
14 ─       goto #15
15 ┄ %37 = φ (#7 => true, #14 => false)::Bool
│    %38 = φ (#14 => %34)::NamedTuple{(:x, :y),_A} where _A<:Tuple{Int64,Union{Missing, Float64}}
│    %39 = φ (#14 => %16)::Int64
└───       goto #17 if not %37
16 ─       goto #20
17 ─ %42 = Base.add_int(%3, 1)::Int64
│          Tables.eachcolumns(Tables.add_or_widen!, sch, %38, columns, %42, updated, $(QuoteNode(Base.HasLength())))::Any
```
Is that failure to inline related? Like, the inline-cost is calculated too expensive because we have `NamedTuple{(:x, :y),_A} where _A<:Tuple{Int64,Union{Missing, Float64}}`?

(I've double checked that my definitions of `eachcolumns` and related calls all have `@inline` appropriately).

base/compiler/tfuncs.jl
```diff
@@ -1122,13 +1128,34 @@ function apply_type_tfunc(@nospecialize(headtypetype), @nospecialize args...)
...
+                # Is this the second parameter to a NamedTuple?
+                if isnamedtuple && isa(ua, UnionAll) && uw.parameters[2] === ua.var
+                    # If the names are known, keep the upper bound (which we know is
+                    # at most Tuple), but otherwise widen to Tuple.
+                    # This is a widening heuristic to avoid keeping type information
+                    # that's unlikely to be useful.
+                    if uw.parameters[1] isa Tuple || (i == 2 && tparams[1] isa Tuple)
+                        if ub >: Tuple
+                            ub = Tuple
+                        end
```
vtjnash, line 1140: `ub = typeintersect(ub, Tuple)`?

base/compiler/tfuncs.jl
```diff
@@ -1122,13 +1128,34 @@ function apply_type_tfunc(@nospecialize(headtypetype), @nospecialize args...)
             #        ai = ai.body
             #    end
             else
+                # Is this the second parameter to a NamedTuple?
+                if isnamedtuple && isa(ua, UnionAll) && uw.parameters[2] === ua.var
+                    # If the names are known, keep the upper bound (which we know is
+                    # at most Tuple), but otherwise widen to Tuple.
+                    # This is a widening heuristic to avoid keeping type information
+                    # that's unlikely to be useful.
+                    if uw.parameters[1] isa Tuple || (i == 2 && tparams[1] isa Tuple)
```
vtjnash, line 1137: This feels like it assumes NamedTuple can only be used in very specific ways. But I don't see a particular reason to believe that `i == 2` implies anything about it being the bound for the second parameter. As a trivial counter example, we can just flip the order of them:

```
julia> NamedTuple{T, S} where {S, T}
NamedTuple{T,S} where T where S

julia> ans{Tuple{Int}, (:a,)}
NamedTuple{(:a,),Tuple{Int64}}
```

base/compiler/tfuncs.jl
```diff
@@ -1105,8 +1108,11 @@ function apply_type_tfunc(@nospecialize(headtypetype), @nospecialize args...)
             #    ai = rename_unionall(ai)
             #    unw = unwrap_unionall(ai)
             #end
+            ub = ai <: Type ? instanceof_tfunc(ai)[1] : Any
```
vtjnash, line 1111:
```suggestion
            ub = widenconst(ai) <: Type ? instanceof_tfunc(ai)[1] : Any
```

base/compiler/tfuncs.jl
JeffBezanson: That's what `uw.parameters[2] === ua.var` above is trying to check for --- that the var we're substituting for corresponds to the second parameter.

base/compiler/tfuncs.jl
```diff
@@ -1122,13 +1129,32 @@ function apply_type_tfunc(@nospecialize(headtypetype), @nospecialize args...)
...
+                    # This is a widening heuristic to avoid keeping type information
+                    # that's unlikely to be useful.
+                    if uw.parameters[1] isa Tuple || (i == 2 && tparams[1] isa Tuple)
+                        ub = typeintersect(ub, Tuple)
+                    else
+                        ub = Tuple
+                    end
+                else
+                    ub = Any
+                end
```
vtjnash, line 1145:
```suggestion
                    if !(uw.parameters[1] isa Tuple || (i == 2 && tparams[1] isa Tuple))
                        ub = Any
                    end
                else
                    ub = Any
                end
```

vtjnash (commented):
Very late night thought: didn't we attempt part of this 6 months ago, before we realized this bound would be wrong, and did https://github.com/JuliaLang/julia/pull/33472 instead?

Counter-example:
```julia
julia> NamedTuple{<:Any, <:Any}
NamedTuple{var"#s1",var"#s2"} where var"#s2" where var"#s1"
```

base/compiler/tfuncs.jl
vtjnash: Ah. I guess it's just the `i == 2 && tparams[1] isa Tuple` that seemed strange, but I suppose that variable can be assumed that it must appear as the names for the NamedTuple in this case. And it's hopefully not important to deal with the case where the variables are swapped.

base/compiler/tfuncs.jl
JeffBezanson: Ah, yes. The `Tuple` bound should be valid only if `ai_w <: Type` above holds, i.e. if whatever we're putting in there is not a typevar.

vtjnash: I think not even then?
```
julia> NamedTuple{<:Any, Val{T}} where T
NamedTuple{var"#s1",Val{T}} where var"#s1" where T
```

JeffBezanson: Ouch.

vtjnash (approved):

quinnj:
Exciting! @JeffBezanson , do you have any thoughts on my inlining question from above? Just wondered if you were aware of something obvious. I'm planning on looking into it a little deeper as well.

JeffBezanson:
Since we infer `eachcolumns` as `Any`, I imagine there might be dynamic dispatches in it, which make the inlining cost very high.

Issue #36345, opened 2020-06-18
accept path of startup file in --startup-file command argument


simeonschaub:
AFAIK, there isn't currently a way to explicitly specify a custom startup file for Julia to use. It would be quite useful to allow specifying a file path to `--startup-file`. The startup file gets executed before REPL startup, sothis would make it easier to quickly try out a different startup file, without making any changes to the default config.

PR #36368, opened 2020-06-20
RFC: Allow specifying startup.jl via environment variable

mortenpi:
Adds the `JULIA_STARTUP_FILE` environment variable which can be used to specify or override the default `~/config/startup.jl` file(s).

The implementation actually allows for multiple startup files to be specified, similar to `JULIA_LOAD_PATH` (i.e. you separate them with `:` or `;`). Also, depending on whether an empty entry is in that list, you can choose whether you want to load the global system & user startup files. `--startup-file=no` will still override this, as you'd expect.

**Motivation.** When I work on a project, I usually want everything on live in a fully self-contained repository. I'll have a shell script that I can source which sets up e.g. `JULIA_PROJECT`, `JULIA_LOAD_PATH` etc., and then I can run `julia foo.jl` on scripts to run my calculations etc. But you can't easily have a project-specific startup file.

With this, I would just specify this additional environment variable in that shell script I source. Furthermore, I can use this to disable the system/user startup files, so the project environment becomes more reproducible.

Technically, I could also use a project-specific depot with `JULIA_DEPOT_PATH`, as Julia looks for the startup file in the topmost depot. But that means initializing a whole depot for each project, which feels a bit heavy. Also, it will also hide the startup script under two layers of directories -- it feels cleaner to have it at the top level of the project repo.

This is a POC/RFC implementation. Will clean it up and add docs etc., if there is support for this.

### DIFF
--- a/base/client.jl
+++ b/base/client.jl
@@ -303,7 +303,42 @@ function exec_options(opts)
     nothing
 end
 
+split_env_path(paths) = split(paths, Sys.iswindows() ? ';' : ':')
+
 function load_julia_startup()
+    # We'll check the JULIA_STARTUP_FILE environment variable for user-specified startup
+    # files.
+    JULIA_STARTUP_FILE = get(ENV, "JULIA_STARTUP_FILE", nothing)
+    if JULIA_STARTUP_FILE !== nothing
+        # We need to store the startup files we've already included, to make sure that
+        # we don't include it twice.
+        startup_files_loaded = String[]
+        for path in split(JULIA_STARTUP_FILE, Sys.iswindows() ? ';' : ':')
+            path = expanduser(path)
+            path in startup_files_loaded && continue
+            push!(startup_files_loaded, path)
+            # If an empty path is passed, e.g. by tucking a `:` at the beginning/end, or
+            # `::` in the middle, we'll load the Julia default startup files.
+            if isempty(path)
+                load_julia_startup_defaults()
+                continue
+            end
+            # If the startup file exists (relative to pwd), then include it. Otherwise,
+            # _print_ an error. Julia will not exit/error if a file is missing.
+            fullpath = abspath(path)
+            if isfile(fullpath)
+                include(Main, fullpath)
+            else
+                @error "User-specified startup file not found: $path"
+            end
+        end
+    else
+        load_julia_startup_defaults()
+    end
+    return nothing
+end
+
+function load_julia_startup_defaults()
     # If the user built us with a specific Base.SYSCONFDIR, check that location first for a startup.jl file
     #   If it is not found, then continue on to the relative path based on Sys.BINDIR
     BINDIR = Sys.BINDIR::String


StefanKarpinski:
I’m generally anti-startup files, but if we’re going to support them, this seems ok (but more below). Might be better to call it `JULIA_STARTUP_FILES` since it’s a list of files.

Can this need not be met by specifying one file that loads others? For example, that one file could look for an environment variable named, say, `JULIA_STARTUP_FILES`, split it on the end path separator and then load each of those. You see what I’m getting at: the single startup file mechanism is already maximally flexible so I wonder whether it really needs additional bells and whistles.

mortenpi:
> the single startup file mechanism is already maximally flexible so I wonder whether it really needs additional bells and whistles.

Just to make sure: you mean having a single file that could still be specified using `JULIA_STARTUP_FILE` and not the status quo? Because you currently can't have project-specific startup files without modifying the global ones, which this is trying to solve.

If you just mean the path splitting etc., then sure, I would be equally happy if you can just specify a single file. As you say, if your environment needs something more complex, you can just do it in the startup file. In fact, as I was implementing this, I was on the fence as well whether it should support multiple files or not.

However, if you do go with a single file, I suppose it should then always override the user and global startup files?

StefanKarpinski:
The point I was making was that you can have a global startup file that looks at `JULIA_STARTUP_FILE` and loads that. So the purpose of this would be to override a startup file that’s already present in cases where you don’t control the startup file? It seems like you could also turn off the startup file and just load a file at the start of whatever code you evaluate, no?

mortenpi:
> So the purpose of this would be to override a startup file that’s already present in cases where you don’t control the startup file?

Not exactly. Rather, it's where you don't _want_ to modify the startup file. Relying on a snippet in `~/.julia/config/startup.jl` means that I have to keep that in sync between computers (e.g. I shuffle my Julia environments between my desktop, laptop and the cluster) and projects. Also, if I want to share it with a collaborator, I would have to have them modify their global startup file. 

> It seems like you could also turn off the startup file and just load a file at the start of whatever code you evaluate, no?

For scripts that would indeed work. But you would then still have to do that every time you start an interactive session (REPL or IDE). I actually had forgotten that this was the original motivation for this idea.

So, to clarify my use case: I want to modify `LOAD_PATH` and set up a few environment variables for a project/environment in a way that works seamlessly on both on the command line _and_ in IDEs, like Juno (which don't know how to source shell scripts). And to do all that without touching the global environment (because of the reasons above).

At one point I was experimenting with [adding various environment stuff to `Project.toml` files](https://github.com/mortenpi/ProjectX.jl). But for that to work I still need to do `using ProjectX` somewhere, and I prefer if that doesn't happen in the global startup file.

However, I am no longer sure how useful this is for my particular use case. It's not really enough to get the Juno part working seamlessly. I would minimally also need #36345.

All that said, as startup files are already a thing, it does feel like it should be possible to control them with environment variables (like you can for `LOAD_PATH`, `DEPOT_PATH` etc). For some prior art, e.g. [Python has `PYTHONSTARTUP`](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONSTARTUP) (only seems to apply to interactive mode though). But it might be best to wait until someone comes up with a more convincing use case.

davidanthoff:
I believe this would also be quite useful for the VS Code extension: right now, if users start a shell in VS Code in the integrated terminal and then start Julia inside that shell, none of the integrations we provide work because none of the extension specific code is loaded into that Julia instance. With this we could "inject" ourselves into Julia instances that are started from a shell that is started in VS Code and provide the full integration for such sessions.

davidanthoff:
Bump :) I didn't follow the 1.6 cycle very closely, so this might be too late already, but is there any chance that this could still make it into 1.6? Or did we have a feature freeze already?

If this made it into the next LTS release it would mean we could really rely on it in the VS Code extension, which would allow us to ship a much nicer integration between the extension and Julia instances that are started within a VS Code shell, which is something we've gotten a fair number of requests for.

For our scenario it would actually be key that we can control the startup script _without_ modifying the `startup.jl` file in the depot and _without_ passing another command line argument. What we would want to achieve is that if someone types in `julia` in a shell in VS Code, that another script from the extension runs.

simeonschaub:
Yes, I agree this would be nice to have, I have also wanted something like this before.

> is there any chance that this could still make it into 1.6?

The branching happened a while ago and the beta is already released, so it's probably too late for that, unfortunately.

davidanthoff:
And bump again, this would still be super useful for the VS Code extension :)

PR #36369, opened 2020-06-20
Allow single option with REPL.TerminalMenus

tomyun:
As briefly [mentioned](https://github.com/JuliaLang/julia/pull/35915#issuecomment-638316817) in the recent API change of REPL.TerminalMenus (#35915), single choice menu (nick-paul/TerminalMenus.jl#20) is a feature implemented in the original repo then somehow missed a chance to be part of the stdlib.

The first half of this PR is mostly identical to the above patch with small changes. Basically it relaxes minimum value of pagesize to 1 instead of 2. A notable thing added to this PR was to fix the condition for page offset update in `move_down!()` to ensure proper scrolling when there is only one line in the page (pagesize=1). The recently changed code for this function has a slightly different condition than to what was originally implemented ([old](https://github.com/nick-paul/TerminalMenus.jl/blob/809727726d6b8a27865fe6e9da953bc40891239a/src/AbstractMenu.jl#L110) vs [new](https://github.com/JuliaLang/julia/blob/35c1f87176fecab5a8e8077fb3a62275363a5aa5/stdlib/REPL/src/TerminalMenus/AbstractMenu.jl#L259)).

The remaining part is not directly related to the single choice menu, but to improve handling of pagesize=1 which is now inevitably open to multiple options if users insist on using it. A main issue is that the current scrolling arrow can be only placed for one direction, but now we may end up with a single line with multiple options that can be scrolled to either direction.

For this, the logic in `printcursor()` to print out empty space when no arrow presents is now taken over by `print_arrow()`. `updown_arrow()` provides an additional symbol for the new arrow (`I` in ASCII, `↕` in Unicode by default). Then `printcursor()` is always called for each line regardless of existence of arrows.

### DIFF
--- a/stdlib/REPL/src/TerminalMenus/AbstractMenu.jl
+++ b/stdlib/REPL/src/TerminalMenus/AbstractMenu.jl
@@ -256,7 +256,8 @@ end
 function move_down!(m::AbstractMenu, cursor::Int, lastoption::Int=numoptions(m))
     if cursor < lastoption
         cursor += 1 # move selection down
-        if m.pagesize + m.pageoffset <= cursor
+        pagepos = m.pagesize + m.pageoffset
+        if pagepos <= cursor && pagepos < lastoption
             m.pageoffset += 1 # scroll page down
         end
     elseif scroll_wrap(m)
--- a/stdlib/REPL/src/TerminalMenus/AbstractMenu.jl
+++ b/stdlib/REPL/src/TerminalMenus/AbstractMenu.jl
@@ -369,7 +369,7 @@ down_arrow(c::Config) = c.down_arrow
 down_arrow(::AbstractMenu) = CONFIG[:down_arrow]
 
 updown_arrow(m::ConfiguredMenu) = updown_arrow(m.config)
-updown_arrow(m::AbstractConfig) = updown_arrow(c.config)
+updown_arrow(c::AbstractConfig) = updown_arrow(c.config)
 updown_arrow(c::Config) = c.updown_arrow
 updown_arrow(::AbstractMenu) = CONFIG[:updown_arrow]
 
--- a/stdlib/REPL/src/TerminalMenus/AbstractMenu.jl
+++ b/stdlib/REPL/src/TerminalMenus/AbstractMenu.jl
@@ -315,9 +315,14 @@ function printmenu(out, m::AbstractMenu, cursoridx::Int; oldstate=nothing, init:
         # clearline
         print(buf, "\x1b[2K")
 
-        if i == firstline && m.pageoffset > 0
+        upscrollable = i == firstline && m.pageoffset > 0
+        downscrollable = i == lastline && i != lastoption
+
+        if upscrollable && downscrollable
+            print_arrow(buf, m, updown_arrow(m))
+        elseif upscrollable
             print_arrow(buf, m, up_arrow(m))
-        elseif i == lastline && i != lastoption
+        elseif downscrollable
             print_arrow(buf, m, down_arrow(m))
         else
             print_arrow(buf, m, ' ')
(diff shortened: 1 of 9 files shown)


stdlib/REPL/src/TerminalMenus/AbstractMenu.jl
```diff
@@ -362,10 +368,15 @@ down_arrow(c::AbstractConfig) = down_arrow(c.config)
 down_arrow(c::Config) = c.down_arrow
 down_arrow(::AbstractMenu) = CONFIG[:down_arrow]
 
-print_arrow(buf, ::ConfiguredMenu, c::Char) = print(buf, c, "  ")
+updown_arrow(m::ConfiguredMenu) = updown_arrow(m.config)
+updown_arrow(m::AbstractConfig) = updown_arrow(c.config)
```
timholy, line 372:
```suggestion
updown_arrow(c::AbstractConfig) = updown_arrow(c.config)
```
Maybe a sign that there is a missing test?

stdlib/REPL/src/TerminalMenus/AbstractMenu.jl
```diff
@@ -362,10 +368,15 @@ down_arrow(c::AbstractConfig) = down_arrow(c.config)
...
+updown_arrow(c::Config) = c.updown_arrow
+updown_arrow(::AbstractMenu) = CONFIG[:updown_arrow]
+
+print_arrow(buf, ::ConfiguredMenu, c::Char) = print(buf, c)
```
timholy, line 376:
```suggestion
```
Do we even need `print_arrow` anymore?

stdlib/REPL/test/TerminalMenus/multiselect_menu.jl
```diff
@@ -57,3 +53,5 @@ end
 # Test SDTIN
 multi_menu = MultiSelectMenu(string.(1:10), charset=:ascii)
 @test simulate_input(Set([1,2]), multi_menu, :enter, :down, :enter, 'd')
+multi_menu = MultiSelectMenu(["single option"])
```
timholy, line 56: You have a CI failure due to tests being run with `--depwarn=error`.

```suggestion
multi_menu = MultiSelectMenu(["single option"], charset=:ascii)
```

and similarly for `RadioMenu` below.

timholy (commented):
Nice!

stdlib/REPL/src/TerminalMenus/AbstractMenu.jl
tomyun: I believe now it should get covered by a test with `charset=:ascii` added to trigger ConfiguredMenu instead of legacy Dict-based menu.

stdlib/REPL/src/TerminalMenus/AbstractMenu.jl
tomyun: Good point! They're now replaced by plain `print`.

stdlib/REPL/test/TerminalMenus/multiselect_menu.jl
tomyun: Added `charset=:ascii` to other non-legacy tests.

I feel like it's a bit confusing behavior though, especially when we have an exception like `pagesize` kwarg doesn't trigger new interface.

tomyun:
To cope with some `dynamic_menu` UI errors, `move_down!()` is updated to behave identical to the [original implementation](https://github.com/nick-paul/TerminalMenus.jl/blob/809727726d6b8a27865fe6e9da953bc40891239a/src/AbstractMenu.jl#L110).

stdlib/REPL/test/TerminalMenus/multiselect_menu.jl
timholy: I know. I don't think there's another way to get to a sane interface without breaking backwards compatibility. I was extremely pleased to find something so unintrusive, even if it's far from perfect. We can drop the `CONFIG` interface in Julia 2.0, but that will likely be years away.

timholy:
Thanks, @tomyun!

PR #35693, opened 2020-05-01
Fix links in the manual

ViralBShah:
Fix a couple of broken links in the manual. I am unable to track these four:

```
┌ Warning: invalid local link: unresolved path in base/arrays.md
│   link.text =
│    1-element Array{Any,1}:
│     Markdown.Code("", "^")
│   link.url = "math.html#Base.:^-Tuple{Number,Number}"
└ @ Documenter.Writers.HTMLWriter ~/julia/doc/deps/packages/Documenter/QQWIJ/src/Writers/HTMLWriter.jl:1676
┌ Warning: invalid local link: unresolved path in base/arrays.md
│   link.text =
│    1-element Array{Any,1}:
│     Markdown.Code("", "c^r")
│   link.url = "math.html#Base.:^-Tuple{Number,Number}"
└ @ Documenter.Writers.HTMLWriter ~/julia/doc/deps/packages/Documenter/QQWIJ/src/Writers/HTMLWriter.jl:1676
┌ Warning: invalid local link: unresolved path in base/base.md
│   link.text =
│    1-element Array{Any,1}:
│     Markdown.Code("", "one")
│   link.url = "../base/numbers.html#Base.one"
└ @ Documenter.Writers.HTMLWriter ~/julia/doc/deps/packages/Documenter/QQWIJ/src/Writers/HTMLWriter.jl:1676
┌ Warning: invalid local link: unresolved path in base/base.md
│   link.text =
│    1-element Array{Any,1}:
│     Markdown.Code("", "Float64")
│   link.url = "../base/numbers.html#Core.Float64"
└ @ Documenter.Writers.HTMLWriter ~/julia/doc/deps/packages/Documenter/QQWIJ/src/Writers/HTMLWriter.jl:1676
```

PR #36376, opened 2020-06-21
Fix broken links in docstring of `repeat`

kimikage:
This changes the link targets of `^` parseable and more specific.

This fixes the problem in the first half of https://github.com/JuliaLang/julia/pull/35693#issue-412313318. These four seem like a problem with Documenter.jl (cf. https://github.com/JuliaDocs/Documenter.jl/issues/505) (or Julia). 

**Edit:**
The `^` without signatures should refer to
`base/math/#Base.:^-Tuple{Number,Number}`
but in "base/foo.md", the relative URL is resolved into:
`../math/#Base.:^-Tuple{Number,Number}` in "base/punctuation.md",
`math/#Base.:^-Tuple{Number,Number}` in "base/arrays.md".

In any case, the `math/#Base.:^` is not a suitable reference for `^` in the docstring of `repeat`. This is the reason for fixing only the first half.

### DIFF
--- a/base/strings/basic.jl
+++ b/base/strings/basic.jl
@@ -686,7 +686,7 @@ reverseind(s::AbstractString, i::Integer) = thisind(s, ncodeunits(s)-i+1)
 
 Repeat a string `r` times. This can be written as `s^r`.
 
-See also: [`^`](@ref)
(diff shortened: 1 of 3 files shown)


kimikage:
I'm sorry, but the problem doesn't seem to be totally fixed by this. Probably, we should remove the duplicated docs for `repeat` in "base/arrays".

PR #36380, opened 2020-06-22
Generic `isfinite` method

omus:
Implements a generic `isfinite` method which should work for any type supporting subtraction and `iszero`. My motivation behind the change is that I was writing some generic code which needed `isfinite` to work with `Char` and `Period` instances.

### DIFF
--- a/base/float.jl
+++ b/base/float.jl
@@ -537,20 +537,6 @@ isnan(x::AbstractFloat) = (x != x)::Bool
 isnan(x::Float16) = reinterpret(UInt16,x)&0x7fff > 0x7c00
 isnan(x::Real) = false
 
-"""
-    isfinite(f) -> Bool
-
-Test whether a number is finite.
-
-# Examples
-```jldoctest
-julia> isfinite(5)
-true
-
-julia> isfinite(NaN32)
-false
-```
-"""
 isfinite(x::AbstractFloat) = x - x == 0
 isfinite(x::Float16) = reinterpret(UInt16,x)&0x7c00 != 0x7c00
 isfinite(x::Real) = decompose(x)[3] != 0
--- a/stdlib/Dates/src/periods.jl
+++ b/stdlib/Dates/src/periods.jl
@@ -54,6 +54,7 @@ Base.zero(::Union{Type{P},P}) where {P<:Period} = P(0)
 Base.one(::Union{Type{P},P}) where {P<:Period} = 1  # see #16116
 Base.typemin(::Type{P}) where {P<:Period} = P(typemin(Int64))
 Base.typemax(::Type{P}) where {P<:Period} = P(typemax(Int64))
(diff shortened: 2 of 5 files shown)


omus:
Looks like I should remove this method: `isfinite(x::AbstractFloat) = x - x == 0`

base/float.jl
```diff
@@ -537,20 +537,6 @@ isnan(x::AbstractFloat) = x != x
...
-# Examples
-```jldoctest
-julia> isfinite(5)
-true
-
-julia> isfinite(NaN32)
-false
-```
-"""
 isfinite(x::AbstractFloat) = x - x == 0
```
omus, line 540: I should probably remove this method

omus:
Rebased and resolved conflict

base/float.jl
omus: Method is required otherwise I get:
```
...
docs/utils.jl
Base  ───────────error during bootstrap:
LoadError("sysimg.jl", 19, StackOverflowError())
jl_apply_generic at /Users/omus/Development/Julia/latest/src/gf.c:2368
isfinite at ./float.jl:541
isinf at ./float.jl:549
decompose at ./hashing2.jl:186
unknown function (ip: 0x119e3d79f)
jl_apply_generic at /Users/omus/Development/Julia/latest/src/gf.c:0
isfinite at ./float.jl:541
...
isinf at ./float.jl:549
decompose at ./hashing2.jl:186

*** This error might be fixed by running `make clean`. If the error persists, try `make cleanall`. ***
make[1]: *** [/Users/omus/Development/Julia/latest/usr/lib/julia/sys.ji] Error 1
make: *** [julia-sysimg-ji] Error 2
```

omus:
From triage. Seems better to narrow the scope to `isfinite(::Number)` and also define `isfinite(::Period)`. As Julia has moved away from treating characters as numbers we should leave `isfinite(::Char)` as undefined.

stdlib/Dates/src/periods.jl
```diff
@@ -54,6 +54,7 @@ Base.zero(::Union{Type{P},P}) where {P<:Period} = P(0)
 Base.one(::Union{Type{P},P}) where {P<:Period} = 1  # see #16116
 Base.typemin(::Type{P}) where {P<:Period} = P(typemin(Int64))
 Base.typemax(::Type{P}) where {P<:Period} = P(typemax(Int64))
+Base.isfinite(::Union{Type{P}, P}) where {P<:Period} = true
```
omus, line 57: Following the lead of how `isfinite` is implemented for `TimeType` as that also supports taking the type

omus:
The usefulness of `isfinite(::Number)` is definitely debatable as currently all numeric types in base implement a more specific version. I was thinking about removing it from this PR but it should still be useful for custom `Number` types as well as providing a template for extending this to other types in the future.

PR #36382, opened 2020-06-22
Use NFD-normalized identifiers to match LaTeX hints

### LINKED ISSUE #36378: NFC normalisation of identifiers prevents some latex hints
For example `?ẋ` doesn't print the expected message `"ẋ" can be typed by x\dot<tab>`, because the combining mark is normalized to u1e8b.

I think a possible fix would be to use `Unicode.normalize(s, :NFD)` here:
https://github.com/JuliaLang/julia/blob/f6d34c3c32c535422491bfe980b1ebdb6bc123db/stdlib/REPL/src/docview.jl#L336

sijow:
This fixes #36378.

### DIFF
--- a/stdlib/REPL/src/docview.jl
+++ b/stdlib/REPL/src/docview.jl
@@ -13,6 +13,8 @@ using Base: with_output_color
 
 using InteractiveUtils: subtypes
 
+using Unicode: normalize
+
 ## Help mode ##
 
 # This is split into helpmode and _helpmode to easier unittest _helpmode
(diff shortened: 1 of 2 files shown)


stevengj (approved):

JeffBezanson:
Causes
```
┌ Warning: Package REPL does not have Unicode in its dependencies:
```

knuesel:
@JeffBezanson sorry about that, I fixed it in my branch. Would you like a new PR for the fix?

stevengj:
@knuesel, please do.

Issue #30908, opened 2019-01-30
Julep: lowering of keyword-function methods


timholy:
## TL;DR version

Proposal: for the "body method" produced when lowering kwarg functions, don't gensym the method name. For example, when `foo` has one or more keyword arguments, use `#foo#` instead of `#foo#37`. 

If you're not yet convinced, then the following tome might help.

## Background

The current situation makes it hard to compute valid signatures for all methods defined in a file. This turns out to be problematic for several purposes, most notably for debuggers (both Rebugger and ASTInterpreter2):

1. IDE debuggers need to be able to navigate to the first line of a method so the user can watch as it steps through the source code line-by-line.
1. when using a debugger, there's a temptation to edit the source code to, e.g., fix bugs
1. editing the code can confuse the debugger, because it changes the line numbering of the file. The risk is that when you step into `foo`, the editor (following the debugger's request to present file "filename.jl" at line 555) might land you halfway through the middle of an unrelated method because you deleted a few comments higher up in the file.
1. a solution is available in Revise, which tries to fix this by maintaining a dict of "corrected location" info indexed by the method's signature. The signature allows you to uniquely identify a method, so you can go from methods (obtained from stacktraces or by `which` when you want to `step in`) to the correct current position in the source file.

Hence, it's important to be able to parse a source file and generate an `IdDict{Signature,LocationInfo}` containing all signatures.

## The problem

Parsing signatures is generally easy using the lowered representation of a source file. An annoying exception occurs for functions with keyword arguments. Consider the following:

```julia
julia> module Lowering end
Main.Lowering

julia> ex = :(foo(x::Int8, y=0; z::String="hello") = x+y)
:(foo(x::Int8, y=0; z::String="hello") = begin
          #= REPL[2]:1 =#
          x + y
      end)

julia> Core.eval(Lowering, ex)
foo (generic function with 2 methods)

julia> names(Lowering; all=true)
10-element Array{Symbol,1}:
 Symbol("##foo#1") 
 Symbol("#eval")   
 Symbol("#foo")    
 Symbol("#foo#1")  
 Symbol("#include")
 Symbol("#kw##foo")
 :Lowering         
 :eval             
 :foo              
 :include          

julia> Meta.lower(Lowering, ex)
:($(Expr(:thunk, CodeInfo(
1 ─       $(Expr(:method, Symbol("#foo#2")))
│   %2  = (Core.typeof)(#foo#2)
[...]

PR #36385, opened 2020-06-22
Add `bodyfunction(m::Method)`

timholy:
This finds the function that implements the "body" of keyword-method `m`, as generated by lowering. You can think of this as a complement to `Core.kwfunc`.

This addresses the "discovery" part of #30908. The renaming part is done by LoweredCodeUtils. I'll let @JeffBezanson decide whether to close that issue.

This is basically copy/pasted from SnoopCompile.

### DIFF
--- a/base/reflection.jl
+++ b/base/reflection.jl
@@ -1281,6 +1281,51 @@ function hasmethod(@nospecialize(f), @nospecialize(t), kwnames::Tuple{Vararg{Sym
 end
 
 """
+    fbody = bodyfunction(basemethod::Method)
+
+Find the keyword "body function" (the function that contains the body of the method
+as written, called after all missing keyword-arguments have been assigned default values).
+`basemethod` is the method you obtain via [`which`](@ref) or [`methods`](@ref).
+"""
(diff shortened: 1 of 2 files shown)


timholy:
There haven't been any complaints here so I'll merge tomorrow.

base/reflection.jl
```diff
@@ -1253,6 +1253,51 @@ function hasmethod(@nospecialize(f), @nospecialize(t), kwnames::Tuple{Vararg{Sym
...
 end
 
+"""
+    fbody = bodyfunction(mnokw::Method)
+
+Find the keyword "body function" (the function that contains the body of the method
+as written, called after all missing keyword-arguments have been assigned default values).
+`mnokw` is the method that gets called when you invoke it without supplying any keywords.
+"""
+function bodyfunction(mnokw::Method)
```
KristofferC, line 1263: `meth_no_kw` maybe.. Right now, it feels like I'm reading some foreign language :P

timholy: I went with `basemethod` and referenced `which` and `methods`, seem reasonable?

PR #36389, opened 2020-06-23
add a version of `code_typed` that accepts a function type

JeffBezanson:
...instead of an instance. I need this all the time; is there another convenient way to do it that I missed?


Keno:
No, this is good to have. I think there's like 15 different versions of this in external packages.

vtjnash (approved):

timholy:
We probably also want a variant of `which` that takes a tuple-type signature.

yuyichao:
Does it still have to be a separate argument in this case? (Shouldn’t it just/also accept a triple type or tulle of types)

Keno:
Maybe we should also have a version that just takes the whole signature tuple.

JeffBezanson:
Ok, will add.

PR #36390, opened 2020-06-23
1.5: Set VERSION to 1.5.0-beta2

KristofferC:


KristofferC:
I think we can make a proper RC instead (probably today)

PR #36397, opened 2020-06-23
Check axes in Array(::AbstractArray) (fixes #36220)

### LINKED ISSUE #36220: `copyto!` breaking change in v1.5
```julia
using OffsetArrays

oa = fill(1, (3:5, 3:5))
copyto!(Array{Int,2}(undef, size(oa)), oa)

```

On v1.4: 
```
ERROR: BoundsError: attempt to access 3×3 Array{Int64,2} at index [3:5, 3:5]
Stacktrace:
 [1] throw_boundserror(::Array{Int64,2}, ::Tuple{OffsetArrays.IdOffsetRange{Int64,Base.OneTo{Int64}},OffsetArrays.IdOffsetRange{Int64,Base.OneTo{Int64}}}) at ./abstractarray.jl:537
 [2] checkbounds at ./abstractarray.jl:502 [inlined]
 [3] copyto!(::Array{Int64,2}, ::OffsetArray{Int64,2,Array{Int64,2}}) at ./multidimensional.jl:959
 [4] top-level scope at none:0
```
via a dispatch to https://github.com/JuliaLang/julia/blob/44fa15b1502a45eac76c9017af94332d4557b251/base/multidimensional.jl#L958-L965

whereas on v1.5 it does not throw.

https://github.com/JuliaLang/julia/commit/f345c5f3f81291989e80822b86710de712018069 may be the cause.

timholy:
Now, `Array(A::AbstractArray)` checks to make sure that `A` has axes starting at 1. It's not entirely clear whether this check should be done by `Array(A)` or `convert(Array, A)`, but the more conservative option is to avoid changing behavior, and this is how it works on 1.4 (indirectly via the `copyto!` bug).

EDIT: Fixes #36220

### DIFF
--- a/base/abstractarray.jl
+++ b/base/abstractarray.jl
@@ -940,6 +940,12 @@ function copyto!(B::AbstractVecOrMat{R}, ir_dest::AbstractRange{Int}, jr_dest::A
     return B
 end
 
+function copyto_axcheck!(dest, src)
+    @noinline checkaxs(axd, axs) = axd == axs || throw(DimensionMismatch("axes must agree, got $axd and $axs"))
+
+    checkaxs(axes(dest), axes(src))
+    copyto!(dest, src)
+end
 
 """
     copymutable(a)
--- a/base/array.jl
+++ b/base/array.jl
@@ -561,8 +561,8 @@ promote_rule(a::Type{Array{T,n}}, b::Type{Array{S,n}}) where {T,n,S} = el_same(p
 
 if nameof(@__MODULE__) === :Base  # avoid method overwrite
 # constructors should make copies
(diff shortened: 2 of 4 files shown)


base/abstractarray.jl
```diff
@@ -940,6 +940,11 @@ function copyto!(B::AbstractVecOrMat{R}, ir_dest::AbstractRange{Int}, jr_dest::A
     return B
 end
 
+function copyto_axcheck!(dest, src)
```
mbauman, line 943: This likely needs to be `@inline`d to make use of the `@boundscheck` block, but I'm not sure I'd worry much about that in the first place since you can always just call `copyto!` instead of `@inbounds copyto_axcheck!`.

PR #36399, opened 2020-06-23
Add "_invoke_codeinst" builtin

Keno:
This builtin is a low-level mechanism to execute a specific codeinst.
Higher-level invokers need to make sure that any world-age limits
are observed.

### DIFF
--- a/src/builtin_proto.h
+++ b/src/builtin_proto.h
@@ -24,6 +24,7 @@ DECLARE_BUILTIN(typeof);     DECLARE_BUILTIN(sizeof);
 DECLARE_BUILTIN(issubtype);  DECLARE_BUILTIN(isa);
 DECLARE_BUILTIN(_apply);     DECLARE_BUILTIN(_apply_pure);
 DECLARE_BUILTIN(_apply_latest); DECLARE_BUILTIN(_apply_iterate);
+DECLARE_BUILTIN(_invoke_codeinst);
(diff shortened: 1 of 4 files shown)


vtjnash:
I feel like conceptually this wants to provide https://github.com/yuyichao/FunctionWrappers.jl, but implementation-wise seems to promise a rather awkward set of premises. For example, this sounds like it'd be impossible to infer through or optimize in any way in codegen, due to your aforementioned constraint that the higher-level needs to ensure this call is well-formed (seeming to imply the lower-level can't do so). This also seems to accidentally bypass some of the work that we usually expect to do at a callsite, as illustrated by `_jl_invoke` for example.

Seems like you could just write this in Julia, as demonstrated below. Like this PR, it is also a possibly an unstable hack (as we don't promise the internal calling convention is stable), but perhaps it's just stable enough (and doesn't add any internal support expectations)?
```julia
ccall(codeinst.invoke, Any, (Any, Ptr{Any}, Csize_t, Any), func, args, length(args), codeinst)
```

JeffBezanson:
Ah nice, if this is just for debugging then I definitely like the `ccall` implementation.

Keno:
I guess `args` here would be a SimpleVector to match the expected ABI of invoke? I agree I like that better. If we ever need it to be fast, we could even teach codegen to put that SimpleVector on the stack.

PR #36401, opened 2020-06-23
Add hook in inference recursion resolution for external AbstractInter…

Keno:
…preter

This extends hookability to the same-frame comparison in inference's
recursion cycle detection. The case I ran into that made this necessary
is a recursive, nested AD transform. In this case, inference must detect
if two frames have different orders of derivatives (e.g. the primitive
for `-`, again calls `-`; the external interpreter makes sure that
inference results for these end up in different caches).

### DIFF
--- a/base/compiler/inferencestate.jl
+++ b/base/compiler/inferencestate.jl
@@ -46,6 +46,10 @@ mutable struct InferenceState
     # `max_valid`, to be used in inlining
     matching_methods_cache::IdDict{Any, Tuple{Any, UInt, UInt}}
 
+    # The interpreter that created this inference state. Not looked at by
+    # NativeInterpreter. But other interpreters may use this to detect cycles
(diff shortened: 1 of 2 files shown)


vtjnash (approved):
Seems a little awkward, since we assume elsewhere that these must merge, to ensure that inference converges. But that's already slightly broken (I think all of this logic probably should be over in `abstract_call_method` instead of here), and this is experimental, so I guess that's okay with you?

Keno:
> Seems a little awkward, since we assume elsewhere that these must merge, to ensure that inference converges

I believe convergence is still guaranteed as long as call cycles that are at the same AD gradation merge properly.

Yes, I'm fine with moving this wherever. I'll also imagine that once we figure out the high level abstractions for AbstractInterpreter, there'll be a more explicit way to tag inference frames.

PR #36404, opened 2020-06-23
fix and de-dup cached calls to `methods_by_ftype` in compiler

JeffBezanson:
This does a couple things:

1. Fixes this issue: https://github.com/JuliaLang/julia/commit/8ca6e8d25fbb25cc8d72055f5dcfd61b4a758220#r40082739 but I'm not sure how to trigger it (or if that's even possible).
2. I believe we were pessimizing world ranges in some cases (pretty rarely, it seems), due to the cache updating function not being pure (it updates the outer `{min,max}_valid` arrays during inference).
3. Refactors the common code into one function.
4. The closure here was capturing a `@nospecialize` argument. I boxed it so that the closure can have a concrete type inferred. That could be done generally in lowering but I'll leave that for later.

### DIFF
--- a/base/compiler/ssair/inlining.jl
+++ b/base/compiler/ssair/inlining.jl
@@ -1041,18 +1041,13 @@ function assemble_inline_todo!(ir::IRCode, sv::OptimizationState)
         local fully_covered = true
         for atype in splits
             # Regular case: Retrieve matching methods from cache (or compute them)
-            (meth, min_valid, max_valid) = get(sv.matching_methods_cache, atype) do
-                # World age does not need to be taken into account in the cache
-                # because it is forwarded from type inference through `sv.params`
-                # in the case that the cache is nonempty, so it should be unchanged
-                # The max number of methods should be the same as in inference most
-                # of the time, and should not affect correctness otherwise.
-                min_val = UInt[typemin(UInt)]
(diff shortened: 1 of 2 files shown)


base/compiler/ssair/inlining.jl
```diff
@@ -1100,11 +1095,8 @@ function assemble_inline_todo!(ir::IRCode, sv::OptimizationState)
         if signature_fully_covered && length(cases) == 0 && only_method isa Method
             if length(splits) > 1
                 # get match information for a single overall match instead of union splits
-                meth = get(sv.matching_methods_cache, sig.atype) do
-                    ms = _methods_by_ftype(sig.atype, sv.params.MAX_METHODS,
-                        sv.world, UInt[typemin(UInt)], UInt[typemin(UInt)])
-                    return ms
-                end
+                (meth, min_valid, max_valid) =
```
vtjnash, line 1098: Seems like we should be doing something with these values (`update_valid_age!(min_valid, max_valid, sv)`) if analyze_method is successful

vtjnash (approved):
> capturing a `@nospecialize` argument

That seems like an interesting idea. Even in general, we could put `@nospecialize` on any (captured) local to select this behavior (and not necessarily need a `Box`, but just an `::Any` typed field).

PR #20385, opened 2017-02-02
Support pointer(A::PermDimsArray) if parent(A) supports it

timholy:
Useful for wrapping certain C routines, where the storage order differs from Julia's.

PR #36405, opened 2020-06-23
Fix pointer to no longer assume contiguity

### LINKED ISSUE #36402: 1.5: Surprising new pointer behaviour with adjoints and views
This seems wrong to me:
```julia
julia> A = randn(5,3);

julia> V = view(A',:,2:4)
3×3 view(::Adjoint{Float64,Array{Float64,2}}, :, 2:4) with eltype Float64:
 -0.6594    -0.124149   0.492452
  1.20338    0.224742  -0.0431509
 -0.425888   0.725104   0.976229

julia> Base.unsafe_load(pointer(V))
0.49245221105087544
```

I expected it to return the pointer to `V[1,1]`, not `V[1,end]`. 

Related to https://github.com/JuliaLang/julia/pull/35929 

@mbauman

mbauman:
Fixes #36402.  Turns out this affected both `pointer(::SubArray)` as well as `pointer(::AbstractArray, i)`.

### DIFF
--- a/base/abstractarray.jl
+++ b/base/abstractarray.jl
@@ -1001,7 +1001,14 @@ end
 pointer(x::AbstractArray{T}) where {T} = unsafe_convert(Ptr{T}, x)
 function pointer(x::AbstractArray{T}, i::Integer) where T
     @_inline_meta
-    unsafe_convert(Ptr{T}, x) + (i - first(LinearIndices(x)))*elsize(x)
+    unsafe_convert(Ptr{T}, x) + _memory_offset(x, i)
+end
+
+# The distance from pointer(x) to the element at x[I...] in bytes
+_memory_offset(x::DenseArray, I...) = (_to_linear_index(x, I...) - first(LinearIndices(x)))*elsize(x)
+function _memory_offset(x::AbstractArray, I...)
+    J = _to_subscript_indices(x, I...)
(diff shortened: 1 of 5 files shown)


mbauman:
As I went through this, I realized that there's a question mark in how `pointer(A, i)` is documented: does the `i` mean an offset of elements or does it mean an index?

```
help?> pointer
search: pointer pointer_from_objref unsafe_pointer_to_objref codepoint

  pointer(array [, index])

  Get the native address of an array or string, optionally at a given location index.
```

To me, `location index` reads like a `getindex`y index (and not a memory offset), but this was intentionally disabled by PermutedDimsArray because it wasn't clear.  This would definitively make it an index-location.

https://github.com/JuliaLang/julia/blob/52c55d7934f71c5b2d9f6e6fa98cb48817def57c/base/permuteddimsarray.jl#L56-L61

I'm really just trying to fix #36402 in this PR (and intentionally structured this to allow for further improvements to pointer, and we could re-enable PermutedDimsArray later, too), but I just wanted to call this ambiguity out.  Cc @timholy and #20385.

vtjnash:
It's intended to be the index of the ith element (`getindex`y)

KristofferC:
Good to go?

mbauman:
Yes, this is good to go on my side once tests pass.

PR #36408, opened 2020-06-24
Avoid potential buffer overflow/missing zero termination of string

martinholters:
Ref https://github.com/JuliaLang/julia/commit/00abaf8f5f0fac23e93e140172c10b11383a042f#r39659418. Replace a `strncpy(dest, src, strlen(src))` with `strncpy(dest, src,  DEST_SIZE); dest[DEST_SIZE-1]='\0'`. The old code could overflow the destination if the source was too long (not sure that could actually happen), defeating the purpose of `strncpy`. Further, it failed to put the terminating zero in the destination, relying on it being properly initialized.

I have no idea about the context here, but this code path isn't exercised at all when I do `make test`. Is that somehow system specific? Does CI cover this?

### DIFF
--- a/stdlib/Libdl/test/runtests.jl
+++ b/stdlib/Libdl/test/runtests.jl
@@ -224,11 +224,6 @@ end
 
 # test DL_LOAD_PATH handling and @executable_path expansion
 mktempdir() do dir
-    # Skip these tests if the temporary directory is not on the same filesystem
-    # as the BINDIR, as in that case, a relative path will never work.
-    if Base.Filesystem.splitdrive(dir)[1] != Base.Filesystem.splitdrive(Sys.BINDIR)[1]
-        return
-    end
     # Create a `libdcalltest` in a directory that is not on our load path
     src_path = joinpath(private_libdir, "libccalltest.$(Libdl.dlext)")
     dst_path = joinpath(dir, "libdcalltest.$(Libdl.dlext)")
@@ -237,8 +232,8 @@ mktempdir() do dir
     # Add an absurdly long entry to the load path to verify it doesn't lead to a buffer overflow
     push!(Base.DL_LOAD_PATH, joinpath(dir, join(rand('a':'z', 10000))))
 
-    # Add this temporary directory to our load path, using `@executable_path` to do so.
-    push!(Base.DL_LOAD_PATH, joinpath("@executable_path", relpath(dir, Sys.BINDIR)))
+    # Add the temporary directors to load path by absolute path
+    push!(Base.DL_LOAD_PATH, dir)
 
     # Test that we can now open that file
     Libdl.dlopen("libdcalltest") do dl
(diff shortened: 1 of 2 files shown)


staticfloat:
This code path is only exercised when using `dlopen("libname")` after having run `push!(Base.DL_LOAD_PATH, "@executable_path/path/to/libraries")`.  I've added a test case to this PR.

staticfloat (approved):

stdlib/Libdl/test/runtests.jl
```diff
@@ -222,4 +222,27 @@ let dl = C_NULL
...
 
+# test @executable_path expansion
+mktempdir() do dir
+    # Create a `libdcalltest` in a directory that is not on our load path
+    src_path = joinpath(private_libdir, "libccalltest.$(Libdl.dlext)")
+    dst_path = joinpath(dir, "libdcalltest.$(Libdl.dlext)")
+    cp(src_path, dst_path)
+
+    # Add this temporary directory to our load path, using `@executable_path` to do so.
+    push!(Base.DL_LOAD_PATH, joinpath("@executable_path", relpath(dir, Sys.BINDIR)))
```
martinholters, line 233: Thanks for the test case, but it's actually the branch without `@executable_path` that is relevant here, i.e.
```suggestion
    push!(Base.DL_LOAD_PATH, dir)
```
But it might be useful to include this test as is and another one without `@executable_path`?

martinholters: I've added another test after this one where the last entry in `Base.DL_LOAD_PATH` is replaced with `dir`. Also, I first add an overly long entry to `Base.DL_LOAD_PATH` which makes these tests segfault without this PR. (I couldn't trigger an error due to the missing zero termination; probably we're lucky and the buffer is usually zeroed out when we get to use it.)

@staticfloat does this look good to you?

martinholters: Ah, I had missed this test (before and after my change) fails on Windows. Any idea?

staticfloat (approved):
It works locally for me, and when running on the buildbots.  Strange.

stdlib/Libdl/test/runtests.jl
staticfloat: Aha, figured it out.  It's because on windows our temporary directory and our executable are on different filesystems.  Darn.

stdlib/Libdl/test/runtests.jl
```diff
@@ -222,4 +222,45 @@ let dl = C_NULL
...
 
+# test DL_LOAD_PATH handling and @executable_path expansion
+mktempdir() do dir
+    # Create a `libdcalltest` in a directory that is not on our load path
+    src_path = joinpath(private_libdir, "libccalltest.$(Libdl.dlext)")
+    dst_path = joinpath(dir, "libdcalltest.$(Libdl.dlext)")
+    cp(src_path, dst_path)
+
+    # Add an absurdly long entry to the load path to verify it doesn't lead to a buffer overflow
+    push!(Base.DL_LOAD_PATH, joinpath(dir, join(rand(0:25, 10000) .+ 'a')))
```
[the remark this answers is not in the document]
martinholters: :facepalm: Of course. I feel ashamed...

stdlib/Libdl/test/runtests.jl
```diff
@@ -222,4 +222,45 @@ let dl = C_NULL
...
+        @test_throws ErrorException Libdl.dlsym(dl, :foo)
+
+        fptr = Libdl.dlsym_e(dl, :set_verbose)
+        @test fptr != C_NULL
+        fptr = Libdl.dlsym_e(dl, :foo)
+        @test fptr == C_NULL
+    end
+
+    # Now use the absolute path
+    Base.DL_LOAD_PATH[end] = dir
```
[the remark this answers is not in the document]
martinholters: Good point. But we should then also re-add the looooooooooong path from above.

stdlib/Libdl/test/runtests.jl
martinholters: Ah, great, that might have taken me quite a bit of time to track down...

martinholters:
I've moved the test using the absolute path first so that it can run unconditionally (it can, right?). If this passes tests, I think it should be ready to go.

martinholters:
Thanks for helping out with the tests here, Elliot. Funny how they turned out to be much more involved than the fix itself...

JeffBezanson:
@staticfloat This seems to involve code that doesn't exist on 1.5 --- I don't see what can be backported?

staticfloat:
My bad; I thought it was in 1.5 but I guess I was wrong. :)

PR #36409, opened 2020-06-24
Fix spelling of readdir.

GunnarFarneback:
Trivial typo.


StefanKarpinski (approved):
That once gets me pretty often.

PR #36415, opened 2020-06-24
Clarify that path to julia can be set

goretkin:
Here is one example that I believe points to the wording being misleading: https://discourse.julialang.org/t/multiple-computer-example/10518/8


ViralBShah:
@amitmurthy is not really around...

PR #30174, opened 2018-11-28
Make compilecache atomic

tkf:
Previously, the cache path is directly passed to the --output-ji option of the subprocess for generating compile cache.  However, when several processes are compiling the same package, one process may modify the file while another process is calculating the checksum, yielding a broken cache.  This patch let each process create a cache in a private temporary path and then atomically rename it to the final cache path.

Before this PR, I can consistently invoke `Warning: The call to compilecache failed to create a usable precompiled cache file for Example` with:

```sh
touch ~/.julia/dev/Example/src/Example.jl
for _ in $(seq 30)
do
    julia --startup-file=no -e "using Example" &
done
wait
```

With this PR it does not show any warnings (so far).

ref: https://discourse.julialang.org/t/precompilation-error-using-hpc/17094

PR #36416, opened 2020-06-24
Make compilecache atomic

kpamnany:
Replaces #30174 and is identical except for additionally removing the temporary file code in `src/dump.c`.


vtjnash (approved):

KristofferC:
> and is identical

Would be good to cherry pick that commit then so tha the original author get credit.

kpamnany:
> It would be good to cherry-pick that commit then so the original author gets credit.

It's a set of commits, not one, and in a different repo. I'll add a 'Co-authored-by' line to the commit message instead.

KristofferC:
You can also commit as another author with `--author`

Issue #32888, opened 2019-08-14
`rem2pi` called with `NaN` returns constant number instead of `NaN`


JKrehl:
Calling `rem2pi` with a `NaN` (also `NaN16`, `NaN32`, `NaN128`) returns, depending on the rounding mode, the constants `1.3470949413735402` or `-4.936090365806046` which is obviously incorrect.

This also means that `mod2pi` is incorrect.

ravibitsgoa:
I think we can solve this bug by checking for NaN argument explicitly in rem2pi.
I would like to work on this.

ravibitsgoa:
Currently, `rem2pi` methods are defined in [base/math.jl](https://github.com/JuliaLang/julia/blob/master/base/math.jl), but the tests for them are in [test/numbers.jl](https://github.com/JuliaLang/julia/blob/master/test/numbers.jl). I think they should be moved to test/math.jl too.

ravibitsgoa:
I think `rem2pi(NaN, r)` should return `NaN` and `rem2pi(Inf, r)` should throw a `DomainError`. Is it correct behaviour?

JKrehl:
For performance reasons alone i would prefer `rem2pi` returning `NaN` (or `NaN32`..., according to input type) whenever an invalid value is passed. Errors are interrupts and cost far more (when not optimized away by the compiler) than a poison value. All float types offer a mathematically sound poison value (i.e. `NaN`) for this situation so that should be used in my opinion.
I cannot judge how that relates to possibly existing on the use of poison values and errors in numerical computations.

Thank you for working on this. :)

ravibitsgoa:
I agree about the performance impact of Errors, but most of the trigonometric functions in base/math.jl throw Errors on `Inf` input (E.g. `sin`, `cos`, `tan` etc.). Although `sin(NaN)=NaN`.

Hence, I thought it's a good idea to continue that behaviour.

StefanKarpinski:
These functions are generally expensive enough that the difference between returning nan and throwing an error doesn’t matter.

KristofferC:
> These functions are generally expensive enough that the difference between returning nan and throwing an error doesn’t matter.

With the exception that functions that throw now has a side effect and can't be optimized away if the result is unused / multiple calls with the same argument can't be folded together etc. It's a pretty big semantic difference.

JeffreySarnoff:
```
julia> mod2pi.((-Inf, Inf, NaN))
[...]

PR #36420, opened 2020-06-24
Fix rem2pi for NaN

mgautam98:
Fix for #32888 

This will also fix the `mod2pi`. 

There was a discussion that Domain Error should be thrown or `NaN `should be returned. To be consistent with `rem` I returned `NaN`

### DIFF
--- a/base/math.jl
+++ b/base/math.jl
@@ -1239,6 +1239,8 @@ julia> rem2pi(7pi/4, RoundDown)
 """
 function rem2pi end
 function rem2pi(x::Float64, ::RoundingMode{:Nearest})
+    isnan(x) && return NaN
+
     abs(x) < pi && return x
 
     n,y = rem_pio2_kernel(x)
@@ -1262,6 +1264,8 @@ function rem2pi(x::Float64, ::RoundingMode{:Nearest})
     end
 end
 function rem2pi(x::Float64, ::RoundingMode{:ToZero})
+    isnan(x) && return NaN
+
     ax = abs(x)
     ax <= 2*Float64(pi,RoundDown) && return x
 
@@ -1287,6 +1291,8 @@ function rem2pi(x::Float64, ::RoundingMode{:ToZero})
     copysign(z,x)
 end
 function rem2pi(x::Float64, ::RoundingMode{:Down})
+    isnan(x) && return NaN
+
     if x < pi4o2_h
         if x >= 0
             return x
@@ -1316,6 +1322,8 @@ function rem2pi(x::Float64, ::RoundingMode{:Down})
     end
 end
 function rem2pi(x::Float64, ::RoundingMode{:Up})
+    isnan(x) && return NaN
+
     if x > -pi4o2_h
         if x <= 0
             return x
--- a/test/numbers.jl
+++ b/test/numbers.jl
@@ -2675,6 +2675,13 @@ end
     @test rem2pi(T(-8), RoundUp)      ≈ -8+2pi
 end
 
+@testset "PR #36420 $T" for T in (Float16, Float32, Float64)
+    @test rem2pi(T(NaN), RoundToZero)  === T(NaN)
+    @test rem2pi(T(NaN), RoundNearest) === T(NaN)
+    @test rem2pi(T(NaN), RoundDown)    === T(NaN)
+    @test rem2pi(T(NaN), RoundUp)      === T(NaN)
+end
+
 import Base.^
 struct PR20530; end
 struct PR20889; x; end


mgautam98:
@simeonschaub can you please review it?

base/math.jl
```diff
@@ -975,6 +975,8 @@ julia> rem2pi(7pi/4, RoundDown)
 """
 function rem2pi end
 function rem2pi(x::Float64, ::RoundingMode{:Nearest})
+    isnan(x) && return NaN
+
     abs(x) < pi && return x
```
kimikage, line 980: I guess this can be optimized:
```suggestion
    abs(x) >= pi || return x
```

mgautam98: Yes. but wouldn't that will make it hard to understand and debug later?

kimikage: That's right. However, for such low-level functions, I prefer the speed. Is it not enough to just add comments into the source?

StefanKarpinski: Worth checking if LLVM figures that out and eliminates the `isnan` check.

simeonschaub: It seems like LLVM with `optimize=true` can't eliminate the `fcmp ord double %0, 0.000000e+00` instruction and extra branch from `isnan` here. So I think it makes sense to use `>=` here to replace the `isnan` check and just add an explanatory comment about the `NaN` case.

oscardssmith:
what's the performance impact of this pr?

giordano:
About 5% according to
```
julia> using BenchmarkTools

julia> @benchmark x .= mod2pi.(x) setup=(x=randn(10_000))
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  64.060 μs … 177.371 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     66.712 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   68.426 μs ±   6.567 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

   ▅█▇▅▇▇▄▄▄▂        ▁▁▁▁▂▁                 ▁                  ▂
  ▇██████████▆▅▅▁▅▆▇█████████▇▄▅▄▄▄▅▃▅▅██▇███▇▄▄▃▃▁▄▃▄▄▄▅▅▇▇██ █
  64.1 μs       Histogram: log(frequency) by time        99 μs <

 Memory estimate: 0 bytes, allocs estimate: 0.

julia> @eval Base.Math function rem2pi(x::Float64, ::RoundingMode{:Down})
           isnan(x) && return NaN

           if x < pi4o2_h
               if x >= 0
                   return x
               elseif x > -pi4o2_h
                   return add22condh(x,0.0,pi4o2_h,pi4o2_l)
               end
           end

           n,y = rem_pio2_kernel(x)

           if iseven(n)
               if n & 2 == 2 # n % 4 == 2: add pi
                   return add22condh(y.hi,y.lo,pi2o2_h,pi2o2_l)
               else          # n % 4 == 0: add 0 or 2pi
                   if y.hi > 0
                       return y.hi+y.lo
                   else      # negative: add 2pi
                       return add22condh(y.hi,y.lo,pi4o2_h,pi4o2_l)
                   end
               end
           else
               if n & 2 == 2 # n % 4 == 3: add 3pi/2
                   return add22condh(y.hi,y.lo,pi3o2_h,pi3o2_l)
               else          # n % 4 == 1: add pi/2
                   return add22condh(y.hi,y.lo,pi1o2_h,pi1o2_l)
               end
           end
       end
rem2pi (generic function with 9 methods)

julia> @benchmark x .= mod2pi.(x) setup=(x=randn(10_000))
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  67.314 μs … 171.758 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     69.521 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   72.705 μs ±   8.752 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▂██▅▇▅▄▅▂    ▁ ▂▁▁▁▁                                         ▂
  ██████████▇▇████████▇▇▇▇█▇▇▇████▇▆▆▅▇▅▆▆▆▆▇▇▇▆▅▆▅▆▆▆▆▄▅▄▅▄▅▇ █
  67.3 μs       Histogram: log(frequency) by time       113 μs <

 Memory estimate: 0 bytes, allocs estimate: 0.
```
Unless you have smarter ways to deal with non-finite values, this seems a decent trade-off.

@JeffreySarnoff just reported on Slack that also `Inf` is wrong
```
julia> mod2pi.((-Inf, Inf, NaN))
(1.1963318081441687, 5.086853499035418, 1.3470949413735402)
```

oscardssmith:
removing back-port labels since https://github.com/JuliaLang/julia/pull/46163 is a more complete fix.

PR #36421, opened 2020-06-24
line 55 Base.Math.exponent

igotfr:
Base.Math.exponent

```
exponent(x::BigFloat)
exponent(x::T) where T<:Union{Float16, Float32, Float64}
```

returns the x for 2^y which is closest to x rounded toward zero

### DIFF
--- a/doc/src/base/numbers.md
+++ b/doc/src/base/numbers.md
@@ -51,7 +51,12 @@ Base.signed
 Base.unsigned
 Base.float(::Any)
 Base.Math.significand
+
 Base.Math.exponent
+exponent(x::BigFloat)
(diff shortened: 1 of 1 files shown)


StefanKarpinski:
What's up with this PR?

cindRoberta:
> What's up with this PR?

for me is more intuitive than "Get the exponent of a normalized floating-point number."

my description is from the book: Beginning Julia Programming: For Engineers and Scientists page 109

I could not understand by the documentation, but I clearly understood it by the book. Will help those who are starting, I guarantee

StefanKarpinski:
The PR seems to add text to a block of signatures that are used to generate documentation from docstrings, which breaks the doc generation.

cindRoberta:
#36438

PR #36424, opened 2020-06-25
RFC: Document that mapreduce(f, op, _) may exploit commutativity of op and purity of f

### LINKED ISSUE #36081: maximum(::Broadcasted) can produce buggy outputs with stateful functions
Combining a stateful function, lazy broadcasted object, and reduction can produce a buggy result.

Here is a demo:

```julia
function badzero_before_second_pass(i)
    global FIRST_PASS
    if i == 1
        if FIRST_PASS
            FIRST_PASS = false
            return -0.0  # "badzero"
        else
            return 0.0
        end
    end
    return -1
end

using Base.Broadcast: broadcasted, instantiate

FIRST_PASS = true
@show maximum(instantiate(broadcasted(badzero_before_second_pass, 1:17)))

FIRST_PASS = true
@show maximum(badzero_before_second_pass.(1:17))
```

prints

```
maximum(instantiate(broadcasted(badzero_before_second_pass, 1:17))) = 0.0
maximum(badzero_before_second_pass.(1:17)) = -0.0
```

This is because `maximum` go over an array twice:

https://github.com/JuliaLang/julia/blob/acd0e83b8ba152094f4548a71e4bbd56f97726c2/base/reduce.jl#L594-L609

This is not new in the sense as it could have happened with something like https://github.com/JuliaArrays/MappedArrays.jl.

(I'm not sure what should be done with it but I just thought it's worth recording in the issue. Personally, I think stateful function should be "banned" (i.e., declared to be an undefined behavior) in broadcasting but I've seen, e.g., @mbauman mentioning `rand.()` is a nice idiom so this is likely not everyone's favorite definition.  In general, I think we need to formalize what to be expected from broadcasting.  For example, sparse style broadcasting already expects some kind of pureness in the functions.)

tkf:
Currently `reduce` etc. document that it may exploit associativity of the operator.  However, we already exploit commutativity of certain operations like `max`

https://github.com/JuliaLang/julia/blob/98a845f99541e6ece591a3c0528cdf400a39ff88/base/reduce.jl#L627-L630

Also, using `@simd` arguably requires this caveat.  So, I propose to document the current behavior.

If we do this, I think it makes sense to also require that `f` and `getindex(A, _)` for `mapreduce(f, op, A)` are reasonably pure and they may be called out of order.  Again, this is already assumed in `mapreduce(f, max, array)` since it runs over the array (up to) twice.

https://github.com/JuliaLang/julia/blob/98a845f99541e6ece591a3c0528cdf400a39ff88/base/reduce.jl#L642-L650

In particular, it means that it's possible to implement a more efficient `maximum(f, ::PermutedDimsArray)` etc.

close #36081

### DIFF
--- a/base/reduce.jl
+++ b/base/reduce.jl
@@ -270,6 +270,10 @@ In general, it will be necessary to provide `init` to work with empty collection
 intermediate collection needs to be created. See documentation for [`reduce`](@ref) and
 [`map`](@ref).
 
+Known commutativity of the operation `op` such as `+` and `max` may be exploited in the
+implementation.  It means that the elements of `itr` may be accessed out of order and
+the order in which `f` is called is implementation-defined.
+
 !!! compat "Julia 1.2"
     `mapreduce` with multiple iterators requires Julia 1.2 or later.
 
@@ -439,6 +443,9 @@ The associativity of the reduction is implementation dependent. This means that 
 use non-associative operations like `-` because it is undefined whether `reduce(-,[1,2,3])`
 should be evaluated as `(1-2)-3` or `1-(2-3)`. Use [`foldl`](@ref) or
 [`foldr`](@ref) instead for guaranteed left or right associativity.
+
+Known commutativity of the operation `op` such as `+` and `max` may be exploited in the
+implementation.  It means that the elements of `itr` may be accessed out of order.
 
 Some operations accumulate error. Parallelism will be easier if the reduction can be
 executed in groups. Future versions of Julia might change the algorithm. Note that the
--- a/base/reduce.jl
+++ b/base/reduce.jl
@@ -271,7 +271,7 @@ intermediate collection needs to be created. See documentation for [`reduce`](@r
 [`map`](@ref).
 
 Known commutativity of the operation `op` such as `+` and `max` may be exploited in the
-implementation.  It means that the elements of `itr` may be accessed out of order and
+implementation. This means that the elements of `itr` may be accessed out of order and
 the order in which `f` is called is implementation-defined.
 
 !!! compat "Julia 1.2"
@@ -445,7 +445,7 @@ should be evaluated as `(1-2)-3` or `1-(2-3)`. Use [`foldl`](@ref) or
 [`foldr`](@ref) instead for guaranteed left or right associativity.
 
 Known commutativity of the operation `op` such as `+` and `max` may be exploited in the
-implementation.  It means that the elements of `itr` may be accessed out of order.
+implementation. This means that the elements of `itr` may be accessed out of order.
 
 Some operations accumulate error. Parallelism will be easier if the reduction can be
 executed in groups. Future versions of Julia might change the algorithm. Note that the


base/reduce.jl
```diff
@@ -270,6 +270,10 @@ In general, it will be necessary to provide `init` to work with empty collection
 intermediate collection needs to be created. See documentation for [`reduce`](@ref) and
 [`map`](@ref).
 
+Known commutativity of the operation `op` such as `+` and `max` may be exploited in the
+implementation.  It means that the elements of `itr` may be accessed out of order and
```
StefanKarpinski, line 274:
```suggestion
implementation. This means that the elements of `itr` may be accessed out of order and
```

base/reduce.jl
```diff
@@ -440,6 +444,9 @@ use non-associative operations like `-` because it is undefined whether `reduce(
 should be evaluated as `(1-2)-3` or `1-(2-3)`. Use [`foldl`](@ref) or
 [`foldr`](@ref) instead for guaranteed left or right associativity.
 
+Known commutativity of the operation `op` such as `+` and `max` may be exploited in the
+implementation.  It means that the elements of `itr` may be accessed out of order.
```
StefanKarpinski, line 448:
```suggestion
implementation. This means that the elements of `itr` may be accessed out of order.
```

StefanKarpinski (approved):

base/reduce.jl
```diff
@@ -270,6 +270,10 @@ In general, it will be necessary to provide `init` to work with empty collection
 intermediate collection needs to be created. See documentation for [`reduce`](@ref) and
 [`map`](@ref).
 
+Known commutativity of the operation `op` such as `+` and `max` may be exploited in the
```
mbauman, line 273: What does "known commutativity" mean?  Are we strict about the fact that `max` _should_ be commutative, or are we limiting ourselves to only some `max(::T, ::T)`s?

We already say that the associativity of `reduce` isn't specified (and that if you want to demand a particular associativity to use fold[lr])... I know that's not the same as commutativity, but it feels very similar since it means the accumuland might appear on either side.

tkf: What I meant by "known commutativity" was that `max` implementer must make it commutative.

I guess we can also restrict this to known functions and input types (so methods).  But then we can't use this for something like `maximum(f, xs)` because robustly dispatching on return type is not really possible.

Are there use cases for non-commutative `max`, `min`, `+`, `&`, `|` etc.?  If I need something similar to `max` but not commutative, I think I'm OK with creating a new function `maxish`.  I think it's better to demand more to the implementer so that the higher order functions can optimize more.

mbauman: The bigger point here is simply: are there reduction algorithms that care about commutativity but not associativity?  Associativity seems like the bigger deal to me, but perhaps that's just a limitation of my own imagination.

My gut reaction would lean towards making `(map)fold[lr]` the implementations that guarantee both associativity and commutativity, and free `mapreduce` to do whatever it wants on both properties.

tkf: > Associativity seems like the bigger deal to me

That's my understanding, too.

From a sequential computation tree

```
        +
       / \
      +   4
     / \
    +   3
   / \
  +   2
 / \
0   1
```

you can't lower the height of the tree (= data dependency) by using only commutativity:

```
  +
 / \
4    +
    / \
   +   3
  / \
 2   +
    / \
   0   1
```

while it's possible with associativity

```
     +
   /   \
  +     +
 / \   / \
1   2 3   4
```

I can't see how modifying the tree using commutativity helps the computation.  I _think_ commutativity is useful only if the `op` already is associative.  (Though I'd love to know if there is something you can do only with commutativity.)

> free `mapreduce` to do whatever it wants on both properties

I think my proposal is a bit more subtle. For the associativity, it's the *caller* that asserts it. For the commutativity, it's the implementer of `op`.  I guess this "asymmetry" of who asserts what is not really the most beautiful API.  Some alternatives are:

(1) Add a wrapper type

```julia
struct Commutative{F} <: Function
    op::F
end

(f::Commutative)(a, b) = f.op(a, b)
```

so that `mapreduce` can dispatch on `op::Commutative{<:Union{typeof(max), typeof(min)}}`.  This way, it'd always be the caller who asserts certain properties.  For example, you'd write `maximum(f, xs) = mapreduce(f, Commutative(max), xs)` and document that `max(::eltype(xs), ::eltype(xs))` must be commutative. (This is something I meant to explore in https://github.com/JuliaFolds/Transducers.jl/issues/143.)

(2) Add `reduce_commutative` (with a better name) that requires `op` to be associative and commutative.  If we go this way, maybe we can use consisten naming schemes like `folda` (= `reduce`) for associative `op` and `foldac` for associative and commutative `op`.

(3) Add `reduce_dwim` (with a better name) that also auto-detect known associativity.  This way, it's always the implementer who asserts the associativity.

Ultimately, I think using `reduce` for (3) would be ideal for users. They can just call `reduce` and it dispatches to an optimal implementation.  But it is not possible to do this until Julia 2.0.  I think what I suggest in this PR can be used as an intermediate step for implementing (3) in Julia 2.0.

base/reduce.jl
```diff
@@ -270,6 +270,10 @@ In general, it will be necessary to provide `init` to work with empty collection
...
+implementation. This means that the elements of `itr` may be accessed out of order and
+the order in which `f` is called is implementation-defined.
```
tkf, line 275: @mbauman What do you think about the purity/out-of-order part?

vtjnash:
@mbauman should we close this, since the docs already mention associativity, and the additional mention of commutativity doesn't sound that useful? perhaps we should actually just suggest that all implementations be commutative, and someday worry about changing some like `max` to do that?

jtrakk:
Commutativity can be useful for table join operations, because join(smalltable, bigtable) is often written with a full sequential scan on the left and a hash lookup on the right. Being able to swap the operands can make it much faster. It doesn't have to be written this way of course, but I think it's a good example of where commutativity is helpful.

Issue #36354, opened 2020-06-19
[Feature Request] Sys.iswsl()


mgautam98:
Background:
I am doing a project where I need to open Host system browser. The WSL is causing some problems if I detect it as Linux so I need to detect if Host is WSL to fix it. 

While I can do someting like this: https://github.com/microsoft/WSL/issues/4071 It would be convenient if we could have a Sys.iswsl() in Julia.

I would like to work on it myself but I couldn't understand Sys module .

yuyichao:
Note that this would be different from **all** other `Sys.is*` since this would have to be a runtime detection. Whether having it or not, giving it such similar a name would be misleading.

yuyichao:
(It's more similar to detection of linux distribution)

mgautam98:
@yuyichao why this would be a runtime detection? How does exactly OS is detected in Julia please help me understand.

yuyichao:
> why this would be a runtime detection

Because you are not running a different binary from the linux one.

> How does exactly OS is detected in Julia please help me understand.

This has nothing to do with why it needs to be done at runtime. It has to be done at runtime simply because this is a runtime property. It is not known at compile time and it can't be stored in the code as a constant.

There's nothing special about OS detection in julia. In fact there is none. All what `Sys.is*` do is returning a constant.

mgautam98:
We can check the version/release of Linux during compile time which will tell us if its wsl ?  see [this](https://github.com/microsoft/WSL/issues/4071#issuecomment-496737093)

yuyichao:
No that's the wrong thing to do, that cannot tell you if you are running at runtime.

Again, it isn't about how to do the detection, it's about when to do the detection. The detection must be done **after** you download the binary from https://julialang.org/downloads/ when you start julia. It cannot be done when generating the binary you download.

yuyichao:
In another word, you don't care about whether your julia is compiled on WSL (it's not). You care about if it is running on WSL. And that's different from every other `Sys.is*`. It's not a property of the OS/target (i.e. ABI) it's a property of the distribution.

mgautam98:
Okay thanks for clearing confusion. I compile my Julia on wsl so I forgot that people download binaries 😄 
Now I know why it can only be done at runtime 👍

Keno:
[...]

PR #36425, opened 2020-06-25
Sys.detectwsl()

mgautam98:
Tries to solve #36354 

Since it will be a runtime function I did it with Julia itself.

I looked into libuv docs and there is no wrapper function to get proc-version since Its too system specific and libuv abstracts it.

Looking at https://github.com/microsoft/WSL/issues/4071 and [javascript/wsl](https://github.com/sindresorhus/is-wsl/blob/master/index.js) seems that we can only find it from version information

If its not the correct way then please give me some pointers in the correct direction.

### DIFF
--- a/base/sysinfo.jl
+++ b/base/sysinfo.jl
@@ -31,7 +31,8 @@ export BINDIR,
        iswindows,
        isjsvm,
        isexecutable,
-       which
+       which,
+       detectwsl
 
 import ..Base: show
 
@@ -411,7 +412,18 @@ including e.g. a WebAssembly JavaScript embedding in a web browser.
 """
 isjsvm(os::Symbol) = (os === :Emscripten)
 
-for f in (:isunix, :islinux, :isbsd, :isapple, :iswindows, :isfreebsd, :isopenbsd, :isnetbsd, :isdragonfly, :isjsvm)
+"""
+    Sys.detectwsl([os])
+
+Runtime Predicate for testing if Julia is running inside WSL.
+"""
+function detectwsl(os::Symbol)
+    islinux(os) &&
+    isfile("/proc/sys/kernel/osrelease") &&
+    contains(read("/proc/sys/kernel/osrelease", String), r"Microsoft|WSL"i)
+end
+
+for f in (:detectwsl, :isunix, :islinux, :isbsd, :isapple, :iswindows, :isfreebsd, :isopenbsd, :isnetbsd, :isdragonfly, :isjsvm)
     @eval $f() = $(getfield(@__MODULE__, f)(KERNEL))
 end
 
--- a/test/osutils.jl
+++ b/test/osutils.jl
@@ -28,6 +28,11 @@
     else
         @test Sys.windows_version() >= v"1.0.0-"
     end
+
+    @test !Sys.detectwsl(:Windows)
+    if !Sys.islinux()
+        @test !Sys.detectwsl()
+    end
 end
 
 @testset "@static" begin


DilumAluthge:
Should there be a fallback in case an error occurs while trying to open and/or read from `/proc/version`?

mgautam98:
> Should there be a fallback in case an error occurs while trying to open and/or read from `/proc/version`?
Yes I should wrap it around try-catch

StefanKarpinski:
How about this:
```
detectwsl(os::Symbol) =
    islinux(os) &&
    isfile("/proc/version") &&
    contains(lowercase(readline("/proc/version")), "microsoft")
```

StefanKarpinski:
I guess the try-catch might still be good but the main thing is just checking that the file exists.

StefanKarpinski:
And actually, if the file exists but we can't read it, what value should be returned? An error might be best. That's a pretty weird situation and if you can't open and read the file, how do you know if you're on WSL or not?

DilumAluthge:
Yeah. Is there ever a non-pathological situation in which `/proc/version` exists but is not world-readable?

StefanKarpinski:
I did just try changing the permissions on a Linux system and you can if you have root access change them, so it's possible that someone could make this file not readable to some users. But in that case what's the right answer here? One should generally assume not WSL as the norm, but if you can't read that file we don't know.

DilumAluthge:
Does `Sys.detectwsl()` have to definitely return a `Bool`? Can it instead return `Union{Bool, Missing}`? Then, if e.g. we can't read `/proc/version`, we return `missing`?

DilumAluthge:
(Could replace `Missing` with `Nothing`, if you want.)

StefanKarpinski:
I feel like that's just deferring the error and making it harder to debug. The inevitable usage of this is to test it with a conditional so that's converting a clear "we cannot tell if this system is running WSL or not" error into an obscure " TypeError: non-boolean (Missing) used in boolean context" error.

DilumAluthge:
Hmm. Maybe then throw an error if we can't read the file? At least it will then be clear what the source of the error is.

StefanKarpinski:
Yes, that's why my version just checks if the file exists and then reads it. If it can't be read it will produce a fairly clear error and TBH I anticipate this never happening because it would be a pretty odd situation (either the kernel is badly broken and can't produce the contents of `/proc/version` or someone has changed the permissions on that file to make it unreadable, which, why?).

DilumAluthge:
Makes sense to me!

base/sysinfo.jl
```diff
@@ -411,7 +411,22 @@ including e.g. a WebAssembly JavaScript embedding in a web browser.
 """
 isjsvm(os::Symbol) = (os === :Emscripten)
 
-for f in (:isunix, :islinux, :isbsd, :isapple, :iswindows, :isfreebsd, :isopenbsd, :isnetbsd, :isdragonfly, :isjsvm)
+"""
+    Sys.detectwsl([os])
+
+Predicate for testing if Julia is running inside WSL.
```
StefanKarpinski, line 417:
```suggestion
Runtime predicate for testing if Julia is running inside WSL.
```

imciner2:
So, the problem is the string `microsoft` can appear in non-WSL Linux kernels from Microsoft: https://github.com/microsoft/WSL/issues/423#issuecomment-611086412, since apparently some of their cloud images will also contain that string. Instead, the way that seems to be recommended moving forward is to check for the string `WSL` inside `/proc/sys/kernel/osrelease`. This seems to be the current "official" way of detecting it for WSL2, since that change comes direct from the WSL repo and devs.

mgautam98:
@imciner2 systemd says its official way of detecting WSL https://github.com/systemd/systemd/pull/15325/files#diff-0e08ed2dbc97131dbf8deb9eb7af4163R465-R471 but I am inside WSL2 
```
$ cat /proc/sys/kernel/osrelease
4.19.104-microsoft-standard

$ cat /proc/version
Linux version 4.19.104-microsoft-standard (oe-user@oe-host) (gcc version 8.2.0 (GCC)) #1 SMP Wed Feb 19 06:37:35 UTC 2020
```
`/proc/sys/kernel/osrelease` is returning a substring of `/proc/version`

imciner2:
> but for me (Inside WSL2) not having`WSL` string

I guess that change hasn't trickled down yet. It looks like it was fairly recently done in the upstream WSL code, so it might not have made it into updaes yet. I would suggest adding a comment in the code referencing the ambiguity we get with the `microsoft` string (that it could be both a cloud server or WSL), and reference that WSL issue. That way if someone reports incorrect detection of something, this is documented and doesn't have to be researched everytime.

base/sysinfo.jl
```diff
@@ -411,7 +412,18 @@ including e.g. a WebAssembly JavaScript embedding in a web browser.
...
-for f in (:isunix, :islinux, :isbsd, :isapple, :iswindows, :isfreebsd, :isopenbsd, :isnetbsd, :isdragonfly, :isjsvm)
+"""
+    Sys.detectwsl([os])
+
+Runtime Predicate for testing if Julia is running inside WSL.
+"""
+function detectwsl(os::Symbol)
+    islinux(os) &&
+    isfile("/proc/sys/kernel/osrelease") &&
+    contains(read("/proc/sys/kernel/osrelease", String), r"Microsoft|WSL"i)
```
StefanKarpinski, line 423:
```suggestion
    contains(read("/proc/sys/kernel/osrelease", String), r"\b(Microsoft|WSL)\b"i)
```
Presumably we don't want to match if say the hypothetical "BungalowsLinux" distro had its name in here.

test/osutils.jl
```diff
@@ -28,6 +28,9 @@
     else
         @test Sys.windows_version() >= v"1.0.0-"
     end
+
+    @test !Sys.detectwsl(:Windows)
+    @test !Sys.detectwsl(:Linux)
```
StefanKarpinski, line 33: Won't this test fail if you run it under WSL?

mgautam98: I don't know how we are going to test it. Yes these tests will fail inside WSL how are we going to test it?

mgautam98: I don't know how we are going to test it. Yes these tests will fail inside WSL how are we going to test it?

mgautam98: I don't know how we are going to test it. Yes these tests will fail inside WSL how are we going to test it?

StefanKarpinski: I would test that it's not true when the argument is not `:Linux` since that's always correct and then add this test for the zero-argument version:
```jl
if !Sys.islinux()
    @test !Sys.detectwsl()
end
```

mgautam98: Should the Sys.detectwsl(:Windows) test be removed?

fonsp:
We could call it `Sys.iswsl()` instead of `Sys.detectwsl()`? Then it fits in the list of OS checks, and "detect WSL" sounds like an action, not a question - the name does not suggest that it returns a `Bool`.

imciner2:
> We could call it `Sys.iswsl()` instead of `Sys.detectwsl()`? Then it fits in the list of OS checks, and "detect WSL" sounds like an action, not a question - the name does not suggest that it returns a `Bool`.

See the discussion that occurred in the linked issue about the naming (https://github.com/JuliaLang/julia/issues/36354). The `detect` prefix was chosen because this is a runtime detection of WSL, not a compile-time constant like all the other `sys.is*` functions.

fonsp:
Ah sorry, I didn't see it - I can see why `Sys.is*` should be reserved. 

But I still feel like `detectwsl` is semantically off - it sounds like it has return type "`Union{WSL, Missing}`", not `Bool`. To me, it is comparable to `findnext`, `getfield` and `getkey`, which all return the found index/object, not whether or not it was found.

Some final suggestions:

- `Sys.runtimeiswsl()` (and `Sys.runtimeiscontainer()`, etc)

or we return a struct:

```julia
struct RuntimeProperties
	WSL::Bool
	container::Bool
	etc
end

Sys.runtimeproperties()::RuntimeProperties = ...
```

Just wanted to give some alternatives, but feel free to ignore :)

StefanKarpinski:
Another possible name would be `Sys.checkwsl()` if you think that would be clearer?

fonsp:
Yeah that somehow sounds a lot better... I like it!

fonsp:
Should it take `os::Symbol` as an argument? Or is it there for convenience?

StefanKarpinski:
It's a little weird to pass the symbol in this case since you can't really simulate being on another system with this.

kdheepak:
Was this function removed? I'm not able to access it from my Julia instance in the REPL and I'm not able to see it in the `base/sysinfo.jl` file.

I'm not able to find it using github in this repo as well:

image

base/sysinfo.jl
```diff
@@ -411,7 +412,18 @@ including e.g. a WebAssembly JavaScript embedding in a web browser.
 """
 isjsvm(os::Symbol) = (os === :Emscripten)
 
-for f in (:isunix, :islinux, :isbsd, :isapple, :iswindows, :isfreebsd, :isopenbsd, :isnetbsd, :isdragonfly, :isjsvm)
+"""
+    Sys.detectwsl([os])
+
+Runtime Predicate for testing if Julia is running inside WSL.
```
fingolfin, line 418: How about this? (The 1.11 would have to be changed if this were to be backported to 1.10)
```suggestion
Runtime predicate for testing if Julia is running inside
Windows Subsystem for Linux (WSL).

!!! note
    Unlike `Sys.iswindows`, `Sys.islinux` etc., this is a runtime tests, and thus
    cannot meaningfully be used in `@static if ` constructs.

!!! compat "Julia 1.11"
    This function requires at least Julia 1.11.
```

fingolfin:
This has sat here for a long time unfortunately. So I don't know if @mgautam98 still has any interest in this? If so I'd be happy to help get it over the line. If not perhaps someone would like to "adopt" it. Otherwise we'll have to close it :-(

PR #36430, opened 2020-06-25
LinearAlgebra.givens with unitful arguments

KlausC:
Solves JuliaLang/LinearAlgebra.jl#738

### DIFF
--- a/stdlib/LinearAlgebra/src/givens.jl
+++ b/stdlib/LinearAlgebra/src/givens.jl
@@ -253,6 +253,10 @@ end
 function givensAlgorithm(f::T, g::T) where T
     fs = f / oneunit(T)
     gs = g / oneunit(T)
+    typeof(fs) === T && typeof(gs) === T &&
+    !isa(fs, Union{AbstractFloat,Complex{<:AbstractFloat}}) &&
+    throw(MethodError(givensAlgorithm, (fs, gs)))
+
     c, s, r = givensAlgorithm(fs, gs)
     c, s, r * oneunit(T)
 end
--- a/stdlib/LinearAlgebra/src/givens.jl
+++ b/stdlib/LinearAlgebra/src/givens.jl
@@ -258,7 +258,7 @@ function givensAlgorithm(f::T, g::T) where T
     throw(MethodError(givensAlgorithm, (fs, gs)))
 
     c, s, r = givensAlgorithm(fs, gs)
-    c, s, r * oneunit(T)
+    return c, s, r * oneunit(T)
 end
 
 givensAlgorithm(f, g) = givensAlgorithm(promote(float(f), float(g))...)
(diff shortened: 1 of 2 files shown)


StefanKarpinski:
Thanks! Would be great to add the example from the issue as a test.

KlausC:
The example is from external package `Unitful`. Is there a type `T` in `Base` with `one(T) != oneunit(T)`?

StefanKarpinski:
Right. Sometimes we'll mock up very minimal types for this kind of testing. Not sure if we have one already...

KlausC:
> Right. Sometimes we'll mock up very minimal types for this kind of testing. Not sure if we have one already...

I did not find one. Will make a minimal type similar to `Unitful.Quantity` in module `TestGivens`.

StefanKarpinski:
This seems pretty simple, but it would be good if someone who actually knows what a Givens rotation is took a look. @andreasnoack, could you review?

stdlib/LinearAlgebra/src/givens.jl
```diff
@@ -248,6 +249,18 @@ function givensAlgorithm(f::Complex{T}, g::Complex{T}) where T<:AbstractFloat
...
+# enable for unitful quantities
+function givensAlgorithm(f::T, g::T) where T
+    fs = f / oneunit(T)
+    gs = g / oneunit(T)
+    typeof(fs) === T && typeof(gs) === T &&
+    !isa(fs, Union{AbstractFloat,Complex{<:AbstractFloat}}) &&
+    throw(MethodError(givensAlgorithm, (fs, gs)))
+
+    c, s, r = givensAlgorithm(fs, gs)
+    c, s, r * oneunit(T)
```
dkarrasch, line 261: Nitpicking, but the above methods do have a `return` statement.
```suggestion
    return c, s, r * oneunit(T)
```

dkarrasch (commented):
LGTM, except for one minor style comment.

andreasnoack (approved):

dkarrasch (approved):

dkarrasch (changes requested):
There is some nasty method ambiguity error, so this cannot be merged as is, unfortunately. It's a strange error, because it seems to relate to multiplication methods defined in the tests, not in the code. Would it be possible to reformulate the MockUnitful tests in terms of Furlongs?

KlausC:
> Would it be possible to reformulate the MockUnitful tests in terms of Furlongs?

Unfortunately not, because `Furlong` does not have the property that `f / oneunit(f) isa Union{AbstractFloat,Complex{<:AbstractFloat}}` for `f::Furlong`, which is needed to call `givensAlgorithm`.

The fix assumes, that `f / oneunit(f)` strips off the unit from the type of `f`; there seems no other way inside `Base` to accomplish that :-1: . (I could imagine something like `stripunit` as a companion of `oneunit` to do that).

The detected ambiguities are an insane interaction of `Furlong` and `MockUnitful`. I change the definition of `MockUnitful` to get rid of it. If they are detected or not seems to be a timing problem, when test-ambiguities is still working while test-LinearAlgebra is executed.

dkarrasch (approved):

stevengj:
> Unfortunately not, because `Furlong` does not have the property that `f / oneunit(f)` isa `Union{AbstractFloat,Complex{<:AbstractFloat}}` for `f::Furlong`, which is needed to call `givensAlgorithm`.

Arguably this was a bug in the `Furlong` test type, which should have been fixed rather than defining yet another unitful test type.   It's fixed in #39852.

PR #36438, opened 2020-06-25
Base.Math.exponent

igotfr:
Base.Math.exponent

```
exponent(x::BigFloat)
exponent(x::T) where T<:Union{Float16, Float32, Float64}
```

returns the x for 2^y which is closest to x rounded toward zero

### DIFF
--- a/base/math.jl
+++ b/base/math.jl
@@ -778,7 +778,7 @@ ldexp(x::Float16, q::Integer) = Float16(ldexp(Float32(x), q))
 """
     exponent(x::AbstractFloat) -> Int
 
-Get the exponent of a normalized floating-point number
+Get the exponent of a normalized floating-point number.
 Returns the `y` for `2^y` which is closest to `x` rounded toward zero
 """
 function exponent(x::T) where T<:IEEEFloat
--- a/base/math.jl
+++ b/base/math.jl
@@ -776,10 +776,10 @@ end
 ldexp(x::Float16, q::Integer) = Float16(ldexp(Float32(x), q))
 
 """
-    exponent(x::BigFloat)
-    exponent(x::T) where T<:Union{Float16, Float32, Float64}
+    exponent(x::AbstractFloat) -> Int
 
-Returns the x for 2^y which is closest to x rounded toward zero
+Get the exponent of a normalized floating-point number
+Returns the `y` for `2^y` which is closest to `x` rounded toward zero
 """
 function exponent(x::T) where T<:IEEEFloat
     @noinline throw1(x) = throw(DomainError(x, "Cannot be NaN or Inf."))
--- a/base/math.jl
+++ b/base/math.jl
@@ -779,7 +779,7 @@ ldexp(x::Float16, q::Integer) = Float16(ldexp(Float32(x), q))
     exponent(x::AbstractFloat) -> Int
 
 Get the exponent of a normalized floating-point number.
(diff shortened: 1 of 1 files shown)


base/math.jl
```diff
@@ -776,9 +776,10 @@ end
 ldexp(x::Float16, q::Integer) = Float16(ldexp(Float32(x), q))
 
 """
-    exponent(x) -> Int
+    exponent(x::BigFloat)
+    exponent(x::T) where T<:Union{Float16, Float32, Float64}
 
-Get the exponent of a normalized floating-point number.
+Returns the x for 2^y which is closest to x rounded toward zero
```
dkarrasch, line 782:
```suggestion
Returns the `y` for `2^y` which is closest to `x` rounded toward zero.
```
? Also, I don't think removing the previous explanation is a good idea. It is just another way of saying it, and may be closer to some users depending on their background.

dkarrasch (dismissed):

base/math.jl
```diff
@@ -776,9 +776,10 @@ end
 ldexp(x::Float16, q::Integer) = Float16(ldexp(Float32(x), q))
 
 """
-    exponent(x) -> Int
+    exponent(x::BigFloat)
+    exponent(x::T) where T<:Union{Float16, Float32, Float64}
```
StefanKarpinski, line 780: This would be clearer in the original form just putting an `::AbstractFloat` annotation on `x`, i.e.
```jl
exponent(x::AbstractFloat) -> Int
```

base/math.jl
cindRoberta: I did the changes

base/math.jl
cindRoberta: I did the changes

base/math.jl
```diff
@@ -776,9 +776,10 @@ end
 ldexp(x::Float16, q::Integer) = Float16(ldexp(Float32(x), q))
 
 """
-    exponent(x) -> Int
+    exponent(x::AbstractFloat) -> Int
 
-Get the exponent of a normalized floating-point number.
+Get the exponent of a normalized floating-point number
```
StefanKarpinski, line 781:
```suggestion
Get the exponent of a normalized floating-point number.
```

base/math.jl
```diff
@@ -776,9 +776,10 @@ end
 ldexp(x::Float16, q::Integer) = Float16(ldexp(Float32(x), q))
 
 """
-    exponent(x) -> Int
+    exponent(x::AbstractFloat) -> Int
 
 Get the exponent of a normalized floating-point number.
+Returns the `y` for `2^y` which is closest to `x` rounded toward zero
```
StefanKarpinski, line 782:
```suggestion
Returns the integer `y` for which `2^y` is closest to `x` rounded toward zero.
```

base/math.jl
```diff
@@ -776,9 +776,10 @@ end
 ldexp(x::Float16, q::Integer) = Float16(ldexp(Float32(x), q))
 
 """
-    exponent(x) -> Int
+    exponent(x::AbstractFloat) -> Int
 
 Get the exponent of a normalized floating-point number.
+Returns the integer `y` for which `2^y` is closest to `x` rounded toward zero.
```
StefanKarpinski, line 782:
```suggestion
Returns the largest integer `y` such that `2^y ≤ abs(x)`.
```

PR #36444, opened 2020-06-26
use isnothing instead of `=== nothing`

mzaffalon:


KristofferC (changes requested):
This is not a good idea. Using `!==` allows inference to do more checking, see https://docs.julialang.org/en/v1/manual/performance-tips/#Checking-for-equality-with-a-singleton-1.

mzaffalon:
For my understanding: isn't this what [`isnothing(::Nothing) = true
`](https://github.com/JuliaLang/julia/blob/master/base/some.jl#L67) does?

KristofferC:
> For my understanding: isn't this what isnothing(::Nothing) = true does?

No, it needs to be inline `===` (or `!==`).

mzaffalon:
When would one use `isnothing`?

KristofferC:
Can use it in something like `filter(!isnothing, x)`

Issue #36393, opened 2020-06-23
Julep: attempting to measure (& test?) "absolute" risk of invalidation


timholy:
Short version: as protection against invalidation (and possibly "code quality" in general), perhaps we should consider some kind of nanosoldier test on the count of backedges of poorly-inferred `MethodInstance`s. Also an update on progress (& regressions) from 1.4 to master.

There are two things that bother me about my recent work on improving inferrability of Base, Pkg, REPL, etc.:

- it's all package-specific: I care about PkgX, so I search for and stomp on invalidations that it triggers, but that may or may not help users of PkgY. Worse, is there any risk it could make things worse for PkgY?
- it's easy to imagine regressions. The fixes I've been making only rarely have a consequence that can be caught by `@inferred`, as many methods return a known type even if their internals are not well-inferred.

For these reasons, I've begun to wonder whether we can & should add some tests that look for "problematic" `MethodInstance`s and tally their number of backedges. The goal would be to develop an absolute measure of our vulnerability to invalidation: at-risk `MethodInstances` that have the most backedges can cause the most invalidation. If we can reliably measure our vulnerability, then we could also introduce some "hold the line" tests, which prevent our vulnerability from increasing above a certain level. I recognize that there are serious downsides in adding such tests---someone wants to add cool feature X but it's blocked because it's not easy to do that inferrably---but there are also quality-concern risks with *not* having such tests. So, I thought I'd begin by at least trying to come up with some kind of overall measure of our vulnerability.

At the bottom of this I present a couple of scripts that seem useful. What they try to do is pull out `MethodInstances` that seem possibly problematic, in the sense that they fit patterns that I've seen while studying invalidations. Overwhelmingly, they involve abstract types, but many methods with abstract types seem reasonably OK, so I've tried to accommodate that in some specific ways. I should say right from the outset that I'm not at all confident that I've made the selection well, but at least it does pull out a lot of `MethodInstance`s that have featured heavily in my invalidation work (as well as many that haven't). For more detail you should really just open up the code below.

The output of the analysis is a scatter plot, comparing where we were on Julia 1.4.2 vs where we are now on master. Each dot represents one possibly-problematic `MethodInstance` signature. Magenta is for exported names, green for non-exported names:

![method_scatter_log](https://user-images.githubusercontent.com/1525481/85400641-5b157280-b51e-11ea-8d3f-1e6a22c020c3.png)

[...]

PR #34274, opened 2020-01-06
Define push! and popfirst! for AbstractChannel

tkf:
This PR defines

https://github.com/JuliaLang/julia/blob/a143db189501ccffc08dee58c9ab8a35d80e8294/base/channels.jl#L10-L11

which are previously only defined for `Channel`.

This PR also fixes the inconsistency of `push!(c::Channel, v)` as reported in #34202; i.e., `push!(c::Channel, v)` now returns the first argument `c` like any other containers.

close #34202

PR #36449, opened 2020-06-26
More improvements to inference in IO

timholy:
Also continues the work in #36280.

### DIFF
--- a/base/io.jl
+++ b/base/io.jl
@@ -357,35 +357,35 @@ end
 function pipe_reader end
 function pipe_writer end
 
-write(io::AbstractPipe, byte::UInt8) = write(pipe_writer(io), byte)
-unsafe_write(io::AbstractPipe, p::Ptr{UInt8}, nb::UInt) = unsafe_write(pipe_writer(io), p, nb)
-buffer_writes(io::AbstractPipe, args...) = buffer_writes(pipe_writer(io), args...)
-flush(io::AbstractPipe) = flush(pipe_writer(io))
+write(io::AbstractPipe, byte::UInt8) = write(pipe_writer(io)::IO, byte)
+unsafe_write(io::AbstractPipe, p::Ptr{UInt8}, nb::UInt) = unsafe_write(pipe_writer(io)::IO, p, nb)
+buffer_writes(io::AbstractPipe, args...) = buffer_writes(pipe_writer(io)::IO, args...)
+flush(io::AbstractPipe) = flush(pipe_writer(io)::IO)
 
-read(io::AbstractPipe, byte::Type{UInt8}) = read(pipe_reader(io), byte)
-unsafe_read(io::AbstractPipe, p::Ptr{UInt8}, nb::UInt) = unsafe_read(pipe_reader(io), p, nb)
-read(io::AbstractPipe) = read(pipe_reader(io))
-readuntil(io::AbstractPipe, arg::UInt8; kw...) = readuntil(pipe_reader(io), arg; kw...)
-readuntil(io::AbstractPipe, arg::AbstractChar; kw...) = readuntil(pipe_reader(io), arg; kw...)
-readuntil(io::AbstractPipe, arg::AbstractString; kw...) = readuntil(pipe_reader(io), arg; kw...)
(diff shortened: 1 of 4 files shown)


base/set.jl
```diff
@@ -289,7 +289,7 @@ function _unique!(f, A::AbstractVector, seen::Set, current::Integer, i::Integer)
         end
         i += 1
     end
-    return resize!(A, current - firstindex(A) + 1)
+    return resize!(A, current - firstindex(A) + 1)::typeof(A)
```
tkf, line 292: I've been wondering if we should separate "call API" and "overload API."  It makes much easier to imposing an invariance like this one.  For example, we can have

```julia
function resize!(a, n)
    __resize!__(a, n)
    return a::typeof(a)
end
```

It asserts the return type and also makes sure that the object `a` is returned.  This is useful for something like `push!` where some implementations forget to return the correct object #34274.

base/io.jl
```diff
@@ -357,35 +357,35 @@ end
...
-read(io::AbstractPipe, byte::Type{UInt8}) = read(pipe_reader(io), byte)
-unsafe_read(io::AbstractPipe, p::Ptr{UInt8}, nb::UInt) = unsafe_read(pipe_reader(io), p, nb)
-read(io::AbstractPipe) = read(pipe_reader(io))
-readuntil(io::AbstractPipe, arg::UInt8; kw...) = readuntil(pipe_reader(io), arg; kw...)
-readuntil(io::AbstractPipe, arg::AbstractChar; kw...) = readuntil(pipe_reader(io), arg; kw...)
-readuntil(io::AbstractPipe, arg::AbstractString; kw...) = readuntil(pipe_reader(io), arg; kw...)
-readuntil(io::AbstractPipe, arg::AbstractVector; kw...) = readuntil(pipe_reader(io), arg; kw...)
-readuntil_vector!(io::AbstractPipe, target::AbstractVector, keep::Bool, out) = readuntil_vector!(pipe_reader(io), target, keep, out)
-readbytes!(io::AbstractPipe, target::AbstractVector{UInt8}, n=length(target)) = readbytes!(pipe_reader(io), target, n)
+write(io::AbstractPipe, byte::UInt8) = write(pipe_writer(io)::IO, byte)
```
tkf, line 360: Does it make sense to define `_pipe_writer(io) = pipe_writer(io)::IO`?

timholy: Not a bad idea, but annotating each call site seems reasonable too. The `_pipe_writer` solution would fall apart if people start defining new `_pipe_writer` methods :frowning_face:. Sometimes it might be nice to declare a function to be non-specializable!

timholy: I decided to adopt that strategy in #36453, since there were so many calls to `buffer`. Here, however, I could imagine there being external extensions of `pipe_writer`.

base/set.jl
```diff
@@ -352,24 +352,15 @@ julia> unique!(B)
...
-# elements with isless/< defined.
-function unique!(A)
-    if isempty(A)
-        return A
-    else
-        return _unique!(A)
+function unique!(itr)
+    if isa(itr, AbstractVector)
+        T = eltype(itr)
+        if hasmethod(isless, (T, T))
```
JeffBezanson, line 358: Unfortunately this is still expensive.

base/io.jl
```diff
@@ -902,7 +902,7 @@ end
 function readuntil(io::IO, target::AbstractVector{T}; keep::Bool=false) where T
     out = (T === UInt8 ? StringVector(0) : Vector{T}())
     readuntil_vector!(io, target, keep, out)
-    return out
+    return out::Vector{T}
```
JeffBezanson, line 905: It seems to me that if we know `T` we can already infer this, and if we don't know `T` there is no difference?

timholy: That sounds right. This might have been something I added hoping it would help, but then found the real answer later. I am checking now whether omitting this makes any difference, though I am strongly inclined to agree with you.

I need to get in the habit of specifying which packages I happened to be studying for invalidations when I make individual commites. Otherwise I struggle to re-create the test case that led to this. We should really think about #36393 a bit.

timholy: Seems fine without (checked Images, Plots, FilePathsBase, and of course indirectly all the packages they depend on).

timholy:
I'll merge soon barring any further concerns.

PR #30830, opened 2019-01-24
some better type info in Base to improve latency a bit

JeffBezanson:

Issue #29166, opened 2018-09-13
adding a method to Base.Int can result in (seemingly) unrelated recompilation


rfourquet:
This is a MWE:
```julia
struct X <: Integer end
Base.Int(::X) = 0
```
After entering this in the REPL, it becomes annoyingly slow (even for simple things as entering `1+2`), while Julia apparently recompiles a bunch of functions. This problem apparently disappears for other Base builtin integers, or when using e.g. `Real` instead of `Integer` as the supertype. 
Also, this doesn't happen on 0.7, so I will label as a regression.

JeffBezanson:
I'll note that this is not really a bug --- AFAICT we were not invalidating any methods incorrectly --- but it did point to some places we could improve the typing of Base code. This new definition invalidates anything that calls `Int(x::Integer)`. In general it is not easy to guess how many inferred methods might contain that, or any other function/type pair.

PR #36459, opened 2020-06-27
Fix invalidations from novel Integer conversions

timholy:
Defining

```julia
    struct MyInt <: Integer
        x::Int
    end
    (::Type{T})(x::MyInt) where T<:Integer = T(x.x)
```
triggers about 830 unique method invalidations. This fixes the majority of them, and most of the rest are nothing to worry about (they are in Core.Compiler, or run in the `normalize_keymap` phase of REPL initialization). One package that adds this kind of method is GeometryTypes, which is used by Plots. This PR, together with https://github.com/JuliaWeb/HTTP.jl/pull/549, fix the last "big" sources of invalidation from `using Plots`. Altogether there are "only" 440 invalidations left, many of which are probably unconcerning; when I started, my estimate (which is probably ballpark but not entirely comparable due to some changes in how I'm counting) was nearly 10-fold higher (see https://github.com/JuliaLang/www.julialang.org/pull/794/files).

EDIT: fun fact, the opposite case (#29166) was handled in #30830.

This PR by itself is not a huge effect, but for the record:

Just the predecessors of this PR (which is better than current `master`), including my customizations of #35877:
```julia
julia> @time using Plots; @time p = plot(rand(10)); @time display(p)
  5.339177 seconds (9.02 M allocations: 561.481 MiB, 4.25% gc time)
  1.910392 seconds (2.81 M allocations: 156.618 MiB, 1.55% gc time)
  4.631717 seconds (6.53 M allocations: 364.880 MiB, 1.48% gc time)
```

Adding this PR on top of the previous:
```julia
julia> @time using Plots; @time p = plot(rand(10)); @time display(p)
  4.853647 seconds (9.23 M allocations: 571.112 MiB, 3.34% gc time)
  1.898238 seconds (2.81 M allocations: 156.582 MiB, 3.03% gc time)
  4.382853 seconds (6.46 M allocations: 361.079 MiB, 1.48% gc time)
```

Compared to the timings I posted [here for master](https://github.com/JuliaLang/julia/pull/35877#issuecomment-649572533), this is almost 2s faster, so the cumulative effect is mounting (but also coming to its end, I don't expect much further progress just from fixing invalidations). There's still about 0.8s of precompilation, so getting more to precompile would add a little more benefit.

However, this specific PR should not be merged immediately. For most of my invalidation PRs, I think the improvements in inference can also be justified as improving code-quality. While that applies to some of this PR, a lot of it is just type-annotation, and one might really wish that most of those annotations aren't necessary. Consequently I've marked this RFC because it's worth asking whether there is a more systematic way to achieve the same benefits without all the annotation.

### DIFF
--- a/base/int.jl
+++ b/base/int.jl
@@ -369,7 +369,7 @@ julia> count_ones(7)
 3
 ```
 """
-count_ones(x::BitInteger) = ctpop_int(x) % Int
+count_ones(x::BitInteger) = (ctpop_int(x) % Int)::Int
 
 """
     leading_zeros(x::Integer) -> Integer
@@ -382,7 +382,7 @@ julia> leading_zeros(Int32(1))
 31
 ```
 """
-leading_zeros(x::BitInteger) = ctlz_int(x) % Int
+leading_zeros(x::BitInteger) = (ctlz_int(x) % Int)::Int
 
 """
     trailing_zeros(x::Integer) -> Integer
@@ -395,7 +395,7 @@ julia> trailing_zeros(2)
 1
 ```
 """
-trailing_zeros(x::BitInteger) = cttz_int(x) % Int
+trailing_zeros(x::BitInteger) = (cttz_int(x) % Int)::Int
 
 """
     count_zeros(x::Integer) -> Integer
--- a/base/regex.jl
+++ b/base/regex.jl
@@ -422,11 +422,11 @@ struct SubstitutionString{T<:AbstractString} <: AbstractString
     string::T
 end
 
-ncodeunits(s::SubstitutionString) = ncodeunits(s.string)
-codeunit(s::SubstitutionString) = codeunit(s.string)
-codeunit(s::SubstitutionString, i::Integer) = codeunit(s.string, i)
-isvalid(s::SubstitutionString, i::Integer) = isvalid(s.string, i)
-iterate(s::SubstitutionString, i::Integer...) = iterate(s.string, i...)
+ncodeunits(s::SubstitutionString) = ncodeunits(s.string)::Int
+codeunit(s::SubstitutionString) = codeunit(s.string)::Type{<:Union{UInt8, UInt16, UInt32}}
+codeunit(s::SubstitutionString, i::Integer) = codeunit(s.string, i)::Union{UInt8, UInt16, UInt32}
+isvalid(s::SubstitutionString, i::Integer) = isvalid(s.string, i)::Bool
+iterate(s::SubstitutionString, i::Integer...) = iterate(s.string, i...)::Union{Nothing,Tuple{AbstractChar,Int}}
 
 function show(io::IO, s::SubstitutionString)
     print(io, "s")
--- a/base/arrayshow.jl
+++ b/base/arrayshow.jl
@@ -166,18 +166,19 @@ function print_matrix(io::IO, X::AbstractVecOrMat,
                       vdots::AbstractString = "\u22ee",
                       ddots::AbstractString = "  \u22f1  ",
                       hmod::Integer = 5, vmod::Integer = 5)
+    hmod, vmod = Int(hmod)::Int, Int(vmod)::Int
     if !(get(io, :limit, false)::Bool)
         screenheight = screenwidth = typemax(Int)
     else
-        sz = displaysize(io)
+        sz = displaysize(io)::Tuple{Int,Int}
         screenheight, screenwidth = sz[1] - 4, sz[2]
     end
-    screenwidth -= length(pre) + length(post)
-    presp = repeat(" ", length(pre))  # indent each row to match pre string
+    screenwidth -= length(pre)::Int + length(post)::Int
+    presp = repeat(" ", length(pre)::Int)  # indent each row to match pre string
     postsp = ""
     @assert textwidth(hdots) == textwidth(ddots)
-    sepsize = length(sep)
-    rowsA, colsA = UnitRange(axes(X,1)), UnitRange(axes(X,2))
+    sepsize = length(sep)::Int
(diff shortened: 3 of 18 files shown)


timholy:
One thought (essentially an alternative to #35901 and the suggestion to use runtime dispatch for any non-specialized argument) would be to be able to annotate methods like this:
```julia
@ifabstract Int ncodeunits(s::AbstractString) = #= implementation=#
```
The idea here is that specific methods can override the return type, but that the supplied type is used if inference is purely abstract, i.e., using just the constraints in the method definition itself.

base/int.jl
```diff
@@ -382,7 +382,7 @@ julia> leading_zeros(Int32(1))
 31
 ```
 """
-leading_zeros(x::BitInteger) = ctlz_int(x) % Int
+leading_zeros(x::BitInteger) = (ctlz_int(x) % Int)::Int
```
kimikage, line 385: Is this effective?
```julia
julia> code_warntype(leading_zeros, (Base.BitInteger,))
Variables
  #self#::Core.Compiler.Const(leading_zeros, false)
  x::Union{Int128, Int16, Int32, Int64, Int8, UInt128, UInt16, UInt32, UInt64, UInt8}

Body::Int64
1 ─ %1 = Base.ctlz_int(x)::Union{Int128, Int16, Int32, Int64, Int8, UInt128, UInt16, UInt32, UInt64, UInt8}
│   %2 = Base.Int(%1)::Int64
└──      return %2
```
Rather, I think that the `pad`s in `bin()` and so on need assertions.

base/intfuncs.jl
```diff
@@ -605,8 +605,8 @@ function bin(x::Unsigned, pad::Integer, neg::Bool)
     i = neg + max(pad,sizeof(x)<<3-leading_zeros(x))
     a = StringVector(i)
     while i > neg
-        @inbounds a[i] = 48+(x&0x1)
-        x >>= 1
+        @inbounds a[i] = (0x30+(x&0x1)::Unsigned)::Unsigned
+        x = (x >> 1)::Unsigned
```
kimikage, line 609:
```suggestion
        @inbounds a[i] = 0x30 + ((x % UInt8)::UInt8 & 0x1)
        x >>= 0x1
```
I think this may clarify the intent. As mentioned above, `(pad % Int)::Int` may be safer.

kimikage (commented):
In general, I don't like "whack-a-mole" type assertions.:confused:

This is off-topic, but there seems to be room for optimization in integer-->string conversions(e.g. `bin`). So I think it's better to tackle the conversion functions on another PR.
**Edit:**
I submitted PR #36470, which only targets `bin()` and so on.

base/int.jl
kimikage: Ah, I see...
```julia
julia> code_warntype(leading_zeros, (Base.BitInteger,)) # 1.6.0-DEV.312
Variables
  #self#::Core.Compiler.Const(leading_zeros, false)
  x::Union{Int128, Int16, Int32, Int64, Int8, UInt128, UInt16, UInt32, UInt64, UInt8}

Body::Any
1 ─ %1 = Base.ctlz_int(x)::Any
│   %2 = (%1 % Base.Int)::Any
└──      return %2
```

timholy:
> In general, I don't like "whack-a-mole" type assertions.confused

I'm not a fan either, which is why this PR feels ugly compared to most of the former ones. The issue is that these are code paths that end up getting used by callers with poor inference (sometimes due to `@nospecialize`, meaning that the poor inference is deliberate). These annotation "snip" call chains at strategic points that would otherwise lead to the invalidation of hundreds of methods, which is why it may be justified despite the ugliness. Nevertheless, if we can find a more systematic way (which is essentially what #36463 is about) I'd prefer it.

KristofferC:
I think it is perfectly fine to live with a bit of ugliness to get rid of the most egregious invalidations  (based on empirical data). When the low hanging fruits are gone, then we are in a better position to evaluate if a more systematic method is needed or if it is better to focus on another angle for attacking invalidations etc.

I'm a heavy proponent of not letting perfect be the enemy of good.

base/regex.jl
```diff
@@ -422,11 +422,11 @@ struct SubstitutionString{T<:AbstractString} <: AbstractString
...
 end
 
-ncodeunits(s::SubstitutionString) = ncodeunits(s.string)
-codeunit(s::SubstitutionString) = codeunit(s.string)
-codeunit(s::SubstitutionString, i::Integer) = codeunit(s.string, i)
-isvalid(s::SubstitutionString, i::Integer) = isvalid(s.string, i)
-iterate(s::SubstitutionString, i::Integer...) = iterate(s.string, i...)
+ncodeunits(s::SubstitutionString) = ncodeunits(s.string)::Int
+codeunit(s::SubstitutionString) = codeunit(s.string)::Type{<:Union{UInt8, UInt16, UInt32}}
+codeunit(s::SubstitutionString, i::Integer) = codeunit(s.string, i)::Union{UInt8, UInt16, UInt32}
```
JeffBezanson, line 427: I'm pretty sure this is true for all string types in practice, but it's not technically required.

timholy: Docs say this:
```julia
codeunit(s::AbstractString) -> Type{<:Union{UInt8, UInt16, UInt32}}
```
I've been taking the docs as license to enforce stuff. Should I not?

StefanKarpinski: I guess someone could have a different code unit, but it would be very weird. Not sure all of the logic in base would work correctly with some other code unit.

timholy:
I've stripped out the part that is now being handled by #36470. If there are no objections, I'll merge this fairly soon.

Issue #36463, opened 2020-06-28
Julep: `tfunc`s for all


timholy:
TL;DR: one of the harder compiler problems is handling inference on abstract types. This proposes to cheat by adding lookup tables for a few functions. This also analyzes how many poorly-inferred calls we actually have.

There's been a lot of thinking lately about how to handle abstract inference, e.g., #34742, "typocalypse" #36208, and many of the invalidation PRs are nibbling at individual pieces of this problem. Some of the discussion in #35904 centered on whether it would be possible for inference be more assertive about output types, but the cost of running inference on everything is large and recent progress has focused on doing less inference, not more.

I have a growing sense that the majority of inference problems affect a relatively small fraction of functions. One potential solution (which I'm sure must have been considered before) is to do something akin to the `tfunc` mechanism† for generic functions, automatically adding type-assertions as needed for particular functions (EDIT: previous was clarified in response to @martinholters). I'm not exactly sure how it would work (and therein lies the rub), but something like "if you're calling a method in `ModuleA` with abstract types as wide as the method signature itself, then its permissible to impose custom tfuncs on its callees." This is essentially a strong form of piracy-prevention.

To see how this might work, let's consider the example of `eof`. We document that it returns a `Bool`. In a sense, if you're calling one of "our" `eof` implementations, and your object is a subtype of `IO`, then this would seem to be a restriction you should have to live with (meaning, we'd enforce it by automatically adding a type-assert). So if you have a method `myreader(io::IO, args...)` and you call `eof(io)`, then it seems fair to require that this `eof` call return a `Bool`. Conversely, if you duck-type the method, `myreader(io, args...)`, and if inference ends up producing a purely-abstract version without knowing the type of `io`, then the abstract call is on `eof(::Any)`, and in that case we would not make any guarantees about the return type. (Even though Base "owns" `Any`, we make an exception and pretend it doesn't.) If this is still too restrictive, it could conceivably be loosened by exempting known *concrete* types from the pseudo-`tfunc` constraints. I suspect this isn't a good idea, though, as program behavior would change based on the success or failure of inference, and moreover it just seems weird to allow specific concrete subtypes to escape constraints that we otherwise apply to the abstract type itself.

[...]

PR #36470, opened 2020-06-29
Optimize integer-->string conversions

kimikage:
~This avoids invalidations caused by invalidating `StringVector(::Integer)`. (cf. PR #36459)~ (**Edit:** The problem has been solved by PR #36459.)
This also makes `bin()`, `dec()` and `hex()` slightly faster, but does not change the Printf.

```julia
julia> versioninfo() # on WSL
Julia Version 1.6.0-DEV.312
Commit 6185d24ac3 (2020-06-28 22:42 UTC)
Platform Info:
  WORD_SIZE: 64
  LIBM: libopenlibm

julia> @btime string.(x, base=2) setup=(x=rand(Int64,1000));
  57.300 μs (2001 allocations: 151.67 KiB) # before
  38.400 μs (2001 allocations: 151.69 KiB) # after

julia> @btime string.(x, base=8) setup=(x=rand(Int64,1000));
  34.700 μs (2001 allocations: 117.30 KiB) # before
  34.800 μs (2001 allocations: 117.30 KiB) # after

julia> @btime string.(x, base=10) setup=(x=rand(Int64,1000));
  45.900 μs (2001 allocations: 117.27 KiB) # before
  35.400 μs (2001 allocations: 117.28 KiB) # after

julia> @btime string.(x, base=10) setup=(x=rand(Int32,1000));
  35.300 μs (2001 allocations: 101.69 KiB) # before
  31.300 μs (2001 allocations: 101.69 KiB) # after

julia> @btime string.(x, base=16) setup=(x=rand(Int64,1000));
  38.300 μs (2001 allocations: 115.84 KiB) # before
  33.600 μs (2001 allocations: 115.80 KiB) # after
```

### DIFF
--- a/base/intfuncs.jl
+++ b/base/intfuncs.jl
@@ -616,11 +616,26 @@ ndigits(x::Integer; base::Integer=10, pad::Integer=1) = max(pad, ndigits0z(x, ba
 ## integer to string functions ##
 
 function bin(x::Unsigned, pad::Integer, neg::Bool)
-    i = neg + max(pad,sizeof(x)<<3-leading_zeros(x))
-    a = StringVector(i)
+    m = 8sizeof(x) - leading_zeros(x)::Int
+    n = neg + max((pad % Int)::Int, m)
+    a = StringVector(n)
+    # for i in 0x0:UInt(n-1) # automatic vectorization produces redundant codes
+    #     @inbounds a[n - i] = 0x30 + (((x >> i) % UInt8)::UInt8 & 0x1)
+    # end
+    i = n
+    @inbounds while i >= 4
+        b = UInt32((x % UInt8)::UInt8)
+        d = 0x30303030 + ((b * 0x08040201) >> 0x3) & 0x01010101
+        a[i-3] = (d >> 0x00) % UInt8
+        a[i-2] = (d >> 0x08) % UInt8
+        a[i-1] = (d >> 0x10) % UInt8
+        a[i]   = (d >> 0x18) % UInt8
+        x >>= 0x4
+        i -= 4
+    end
     while i > neg
-        @inbounds a[i] = 48+(x&0x1)
-        x >>= 1
+        @inbounds a[i] = 0x30 + ((x % UInt8)::UInt8 & 0x1)
+        x >>= 0x1
         i -= 1
     end
     if neg; @inbounds a[1]=0x2d; end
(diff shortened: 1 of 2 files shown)


JeffBezanson:
Awesome!

base/intfuncs.jl
```diff
@@ -602,74 +602,111 @@ ndigits(x::Integer; base::Integer=10, pad::Integer=1) = max(pad, ndigits0z(x, ba
 ## integer to string functions ##
 
 function bin(x::Unsigned, pad::Integer, neg::Bool)
-    i = neg + max(pad,sizeof(x)<<3-leading_zeros(x))
-    a = StringVector(i)
+    m = 8sizeof(x) - leading_zeros(x)::Int
+    n = neg + max((pad % Int)::Int, m)
```
StefanKarpinski, line 606: Is the type assertion really necessary in `(pad % Int)::Int`? That seems like something the compiler should know.

timholy: > Is the type assertion really necessary in (pad % Int)::Int

In some circumstances, yes. All applicable methods infer to `Int` but there are 14 of them:

```julia
julia> sig = (Integer, Type{Int})
(Integer, Type{Int64})

julia> cts = code_typed(%, sig);

julia> collect(zip(methods(%, sig), last.(cts)))
14-element Array{Tuple{Method,DataType},1}:
 (rem(x::BigInt, ::Type{T}) where T<:Union{Int16, Int32, Int64, Int8, UInt16, UInt32, UInt64, UInt8} in Base.GMP at gmp.jl:332, Int64)
 (rem(x::BigInt, ::Type{T}) where T<:Union{Int128, Int16, Int32, Int64, Int8, UInt128, UInt16, UInt32, UInt64, UInt8} in Base.GMP at gmp.jl:335, Int64)
 (rem(x::UInt32, ::Type{Int64}) in Base at int.jl:479, Int64)
 (rem(x::Int8, ::Type{Int64}) in Base at int.jl:477, Int64)
 (rem(x::Bool, ::Type{Int64}) in Base at int.jl:474, Int64)
 (rem(x::UInt16, ::Type{Int64}) in Base at int.jl:479, Int64)
 (rem(x::UInt128, ::Type{Int64}) in Base at int.jl:472, Int64)
 (rem(x::UInt64, ::Type{Int64}) in Base at int.jl:482, Int64)
 (rem(x::UInt8, ::Type{Int64}) in Base at int.jl:479, Int64)
 (rem(x::Int128, ::Type{Int64}) in Base at int.jl:472, Int64)
 (rem(x::Int32, ::Type{Int64}) in Base at int.jl:477, Int64)
 (rem(x::Int16, ::Type{Int64}) in Base at int.jl:477, Int64)
 (rem(x::T, ::Type{T}) where T<:Integer in Base at int.jl:540, Int64)
 (rem(x::Integer, T::Type{var"#s71"} where var"#s71"<:Integer) in Base at int.jl:543, Int64)
```
and `max_methods` is 3. So the abstractly-inferred call is judged to have a return type of `Any`, even if any concrete type is inferred correctly:
```julia
julia> f(x) = x % Int

julia> code_typed(f, (Int,))
1-element Array{Any,1}:
 CodeInfo(
1 ─     return x
) => Int64

1-element Array{Any,1}:
 CodeInfo(
1 ─ %1 = (x % Main.Int)::Any
└──      return %1
) => Any
```

Because we are discovering more and more good reasons *not* to run inference to exhaustion (and developing more infrastructure to limit inference, like `@nospecialize`), this is becoming more of an issue. This is basically what #36463 is about.

StefanKarpinski: Gotcha. Carry on.

timholy (approved):
This is a pretty performance-sensitive operation, so unrolling those loops is a great idea.

base/intfuncs.jl
rfourquet, line 606: Using `pad % Int` is invalid in some corner cases (e.g. for `pad == typemax(Int)+1`). It's not necessarily a problem for `bin` as it's an internal function, but some form of validation should probably remain somewhere, e.g. in client functions like `string` (which could take care of converting to `Int`).

kimikage: Is that a security issue?

rfourquet: I don't know, but this makes the behavior of `string` sometimes incorrect, in an  and observable way.

kimikage: > an observable way

Can you give a specific example? Being secure is a requirement of the correctness, but otherwise I don't think the correct behavior is obvious.

rfourquet: Currently, `string(0, pad=typemin(Int128)+3)` returns `""`, which seems correct as per the docstring. With this change (if I'm not mistaken), it will return `"000"`, which seems wrong.

kimikage: I will try to fix it. However, this PR starts with avoiding invalidation in the first place.

rfourquet: > However, this PR starts with avoiding invalidation in the first place.

I don't understand how this is related. I'm not familiar with invalidations hunting, but hopefully we can both avoid invalidations and be correct. In this case, what about `pad = clamp(pad, Int)::Int` ?

kimikage: It's "usually" not a bad idea. However, I don't really want to limit it to the range of `Int` (since it is too "wide" for practical use cases). I want to just avoid the invalidation. 
The `pad`s are always `Int`, in the use cases in `Base` i.e. in the default sysimage. Therefore, `pad % Int` is just insurance. In that respect, I think `clamp` is not better than `rem`.

Perhaps a reasonable fix would be to get rid of the `% Int` (i.e. reverting). Do you have any thoughts?

rfourquet: > I don't really want to limit it to the range of Int (since it is too "wide" for practical use cases)

I don't really understand this sentence. In my view, everything which is `<0` can be set to `0`, and everything which is bigger than `typemax(Int)` probably makes no sense has it's not practical to have a string of length `> typemax(Int)`, so such number can be replaced by `typemax(Int)`. I just described `clamp`. What doesn't seem OK is to use `pad % Int` as this has a weird behavior of for these extreme values of `pad`.

> The pads are always Int, in the use cases in Base i.e. in the default sysimage. 

Maybe, but I'm concerned by usage outside of base where your current change leads to an incorrect result as I showed with an example. But unfortunately I have no other suggestion, as I didn't follow how invalidation work, and I don't know which is better, `clamp` or the `max` that was there or writing the equivalent of `clamp` by hand (which would be sad) etc.

kimikage: >> I don't really want to limit it to the range of Int (since it is too "wide" for practical use cases)
>
> I don't really understand this sentence.

I meant a user friendly error message or docstring.

kimikage:
@rfourquet, I appreciate your comment

I added the clamping into the `string()`, and changed the types of the `pad` of internal functions (e.g. `bin()`) from `Integer` to `Int`. Although we can leave them as `Integer` for "compatibility", I think there are no practical advantages.

I don't know the reason, but there is a little slowdown (~1 ns/number) from the original commit, but the unrolling is still effective.

timholy:
Bumping this. I don't have any objections to merging this.

kimikage:
Since PR #36459 was merged, the `::Int` for `leading_zeros(x)` doesn't seem to be needed anymore.

musm:
Looks like this PR has been approved, is it ok to merge?

base/intfuncs.jl
```diff
@@ -601,75 +601,112 @@ ndigits(x::Integer; base::Integer=10, pad::Integer=1) = max(pad, ndigits0z(x, ba
 
 ## integer to string functions ##
 
-function bin(x::Unsigned, pad::Integer, neg::Bool)
-    i = neg + max(pad,sizeof(x)<<3-leading_zeros(x))
-    a = StringVector(i)
+function bin(x::Unsigned, pad::Int, neg::Bool)
+    m = 8sizeof(x) - leading_zeros(x)
```
musm, line 605:
```suggestion
    m = sizeof(x) - leading_zeros(x)
```

base/intfuncs.jl
```diff
@@ -601,75 +601,112 @@ ndigits(x::Integer; base::Integer=10, pad::Integer=1) = max(pad, ndigits0z(x, ba
...
     end
     if neg; @inbounds a[1]=0x2d; end
     String(a)
 end
 
-function oct(x::Unsigned, pad::Integer, neg::Bool)
-    i = neg + max(pad,div((sizeof(x)<<3)-leading_zeros(x)+2,3))
-    a = StringVector(i)
+function oct(x::Unsigned, pad::Int, neg::Bool)
+    m = div(8sizeof(x) - leading_zeros(x) + 2, 3)
```
musm, line 632:
```suggestion
    m = div(sizeof(x) - leading_zeros(x) + 2, 3)
```

musm: Any reason the tests did not catch this?

base/intfuncs.jl
```diff
@@ -601,75 +601,112 @@ ndigits(x::Integer; base::Integer=10, pad::Integer=1) = max(pad, ndigits0z(x, ba
...
-function hex(x::Unsigned, pad::Integer, neg::Bool)
-    i = neg + max(pad,(sizeof(x)<<1)-(leading_zeros(x)>>2))
-    a = StringVector(i)
-    while i > neg
-        d = x & 0xf
-        @inbounds a[i] = 48+d+39*(d>9)
-        x >>= 4
-        i -= 1
+function hex(x::Unsigned, pad::Int, neg::Bool)
+    m = 2sizeof(x) - (leading_zeros(x) >> 2)
```
musm, line 668:
```suggestion
    m = 2*sizeof(x) - (leading_zeros(x) >> 2)
```

base/intfuncs.jl
musm: for clarity, I think this is an improvement

StefanKarpinski:
I've kicked off the failed CI runs, which looked unrelated. This seems good once CI passes.

kimikage:
Thanks!

kimikage:
I'm sorry, but I may have broken something.
cc: @rfourquet
https://github.com/JuliaCI/NanosoldierReports/blob/c1103cd2398e1d0442f115264893823991742a66/pkgeval/by_date/2020-10/01/logs/BitIntegers/1.6.0-DEV-2fe2b43606.log

rfourquet:
> Perhaps the cause is the type of bit shifting.

@kimikage Thank you for the report, I had no idea this `x << 0x4` was broken in `BitIntegers`.

Issue #8005, opened 2014-08-14
export keyword in front of functions / types


tknopp:
This came up on the mailing list and having worked on Gtk.jl (see https://github.com/JuliaLang/Gtk.jl/blob/master/src/long_exports.jl) I think it would be a good thing to allow writing

```
export type MyType
    x::Int
end

export myfunc(m::MyType) = m.x
```

Don't know if it causes issues that `export` operates on methods rather than functions.

davidmanzanares:
+1 to this idea. This would make it closer to ES6 modules (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export) and it would reduce the number of lines avoiding noise in my opinion.

Regarding the utility of having a single place to look upon the exposed functions, I think this is useful when the exported functions are the API of a library, but I think that for application code the proposal is more convenient.

tk3369:
I love this idea too. 

Is it possible to export only the specific functions that are marked export, taking into account the complete signature?  That would solve the include/exclude-all issue above.

StefanKarpinski:
All that's required to make this happen is someone hacking the parser to make it happen.

dalum:
If anyone is interested, I made a small package that provides an `@export` macro here: https://github.com/dalum/InlineExports.jl

c42f:
For what it's worth I'm not really in favor of this; it seems a little frivolous as surface syntax when a macro achieves the same thing with one extra character to be typed.

If people do really want the functionality, why not just add `@export` to the standard library rather than the parser? That should be strictly more conservative than adding it as actual language syntax.

(I also happen to like the up-front export list which has become the convention in package source and it can be a nice opportunity to group the package's public API in a meaningful way for people who will be working on the code.)

It's worth also considering Jeff's points here: https://github.com/JuliaLang/julia/pull/36474#issuecomment-651380656

One final point — I feel like exports are kind of a big deal and should deserve some careful thinking by package authors. So it might be ok to not have the utmost convenience in their syntax :-)

tknopp:
I don't buy that argument. If `export` is a special keyword, it is an additional thing to remember whether an `@` in-front of it and sometimes not. I favor consistency here.

StefanKarpinski:
True, but one doesn't _need_ the ability to export functions by doing `export function` instead of separately. If that was clearly a desirable feature, then sure, we should do this instead of `InlineExports` but I think the deeper question is whether it's desirable or not in the first place.

[...]

PR #36474, opened 2020-06-29
[RFC] Inline export statements

dalum:
This PR implements the language feature described in #8005.

It works by extending `export` to parse into generic:
```julia
Expr(:export, Expr(...), ...)
```
expressions, which are then traversed to extract exportable symbols, rather than force all args of export expressions to be symbols.

I'm not super familiar with C, and this is the first time I've touched the C code of Julia, so I am not sure if there is already something similar to the `get_exportable_sym` function.  In either case, this seems to work for exporting most types of expressions that someone might want to export.

The only problem with the current implementation, that I have identified, is that:
```
module C

"""docstring ..."""
export f(x) = x

end
```
results in a syntax error, `invalid assignment location "f(x)"`.  This seems to originate from the macro expansion that happens by the documenter, and can be reproduced by
```
julia> eval(Expr(:escape, :(export f(x) = x)))
ERROR: syntax: invalid assignment location "f(x)"
```
I have been trying to debug the error in `julia-syntax.scm`, but I think I could use some help or useful pointers on how to get it right.


JeffBezanson:
I initially thought this was a good idea, until others pointed out that (1) it's nice to have all the exports listed together, (2) this might create the impression that individual methods can be exported, so now I'm not so sure.

This looks correct, but I think it should be implemented in lowering, by converting `export function f end` to `export f; function f end`. Better to avoid doing symbolic manipulation in C. I would wait for the triage discussion before bothering with that though.

dalum:
Thanks for the comment! :blush: I think it is probably best not to discuss this in this PR, but I would just like to add my opinion on the two points you raised:

(1) I also think that for most packages, the convention of placing all exports together is nice. But this convention is not currently enforced by syntax, and I do not think it should be.  When I have been writing research code, I prefer to use inline exports. I sometimes change the name of a function, or have a lot of small "scratch" functions that run a specific setup with a few parameters, and in this case, not having to maintain exported names in two places is very convenient. This type of code is not a case where you use `export` to expose a public API of a well-documented package with multiple maintainers, but where you use `export` to have them available in the REPL when `using` your research package.  It bothered me enough that I have made a small package with an `@export` macro in https://github.com/dalum/InlineExports.jl that I use quite often for those types of projects.  I made this PR because it was mentioned in #8005, that someone just had to hack the parser, and I thought it would be a fun challenge and a way for me to get more familiar with the parser. :slightly_smiling_face: And because I would like to see this as an actual language feature.

By the way, for the issue that some people have raised, related to having a central place where you can see all exported names from a module, the function `Base.names` provides that.

(2) I'm not sure this would be an issue, if the behaviour is documented.

JeffBezanson:
Triage is ambivalent about this. It creates another opportunity for spurious variations in coding style (some already argue that e.g. the two function syntaxes are confusing). There is also complexity around combining it with other function annotations, e.g. `@inline export function ...` or `export @inline function`. Some are in favor though. So not conclusive so far I'm afraid.

non-Jedi:
Is there a reason to add extra syntax for this instead of just a `@export` macro as in https://github.com/dalum/InlineExports.jl?

LilithHafner:
Triage agrees with @c42f https://github.com/JuliaLang/julia/issues/8005#issuecomment-660077430

PR #36477, opened 2020-06-30
Remove outdated restriction on Type{} calls in abstract interpretation

Keno:
`@vtjnash` said in #35845 that he didn't know what this code was for
and didn't think it was necessary. Since he's the one that added it
and the comment says it's about backedges, I'm inclined to believe
him :). In any case, removing it passes base tests, so let's try it.


Keno:
I also pushed the other change here (using `jl_method_table_for`), since it's in the same place and I'd like to PkgEval them together. If one of them turns out problematic, we can split the PR again.

Keno:
@nanosoldier `runtests(ALL, vs = "master")`

Keno:

JeffBezanson (approved):

Keno:
Looks like PkgEval is having some issues. I'll give it until tomorrow morning to get back. Otherwise, I'll just merge and we can see if anything breaks during the regular PkgEval runs.

PR #36478, opened 2020-06-30
Clarify BLAS.asum docstring for complex arrays

sethaxen:
The `BLAS.asum` docstring claims it computes the sum of the absolute values of the elements of the array. However, for complex arrays, it actually does something different. This PR clarifies the docstring.

### DIFF
--- a/stdlib/LinearAlgebra/src/blas.jl
+++ b/stdlib/LinearAlgebra/src/blas.jl
@@ -423,7 +423,11 @@ nrm2(x::Union{AbstractVector,DenseArray}) = nrm2(length(x), x, stride1(x))
 """
     asum(n, X, incx)
 
-Sum of the absolute values of the first `n` elements of array `X` with stride `incx`.
+Sum of the magnitudes of the first `n` elements of array `X` with stride `incx`.
+
+For a real array, the magnitude is the absolute value. For a complex array, the
+magnitude is the sum of the absolute value of the real part and the absolute value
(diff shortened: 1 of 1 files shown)


andreasnoack (approved):

stdlib/LinearAlgebra/src/blas.jl
```diff
@@ -423,7 +423,11 @@ nrm2(x::Union{AbstractVector,DenseArray}) = nrm2(length(x), x, stride1(x))
 """
     asum(n, X, incx)
 
-Sum of the absolute values of the first `n` elements of array `X` with stride `incx`.
+Sum of the magnitudes of the first `n` elements of array `X` with stride `incx`.
+
+For a real array, the magnitude is the absolute value. For a complex array, the
+magnitude is the sum of the absolute value of the real part and the absolute value
+of the imaginary part.
```
oxinabox, line 430: We should make it clear that this is not the normal definition of magnitude for a complex number.
Maybe we should say the `L1 norm` instead?

andreasnoack: I think the proposed formulation is pretty clear. Maybe we could mention explicitly that no square roots are taken but I don't really think it's required. I assume you mean `ℓ1`-norm of each element but I think it would just be unnecessarily confusing to use functional analysis language for applying `abs2` to the elements of a complex array.

PR #36480, opened 2020-06-30
Fix exception stack lowering in finally handlers

### LINKED ISSUE #34579: Exception stack not cleaned up after return inside of a catch handler with a finally clause
I ran into issues with the `exception` test suite failing in a particular build environment when ran in the complete test suite but not by itself.

Through a process of elimination I eventually traced the issue down to [test12806 in core.jl](../blob/d58a805413820014071c4bd33a1c698a57613e3b/test/core.jl#L934) leaving things on the exception stack.

The following `error.jl` code, which is just core `test12806` followed by the very first exception test, seems to reliably demonstrate the issue for both 1.3.0 and 1.3.1 across several linux distributions (nix, gento, arch, debian).

```julia
using Test
using Base: catch_stack

function test12806()
    let catchb = false, catchc = false, catchr = false, a = []
        for i in 1:3
            try
                throw("try err")
            catch e
                i == 1 && break
                i == 2 && continue
                i == 3 && return (catchb, catchc, catchr, a)
            finally
                i == 1 && (catchb = true; continue)
                i == 2 && (catchc = true; )
                i == 3 && (catchr = true; push!(a, 1))
            end
        end
    end
end
@test test12806() == (true, true, false, [1])
@test length(catch_stack()) == 0
```

Here is the output I get from running it

```bash
$ julia error.jl
Test Failed at /home/tyson/julia/error.jl:22
  Expression: length(catch_stack()) == 0
   Evaluated: 1 == 0
in expression starting at /home/tyson/julia/error.jl:22
caused by [exception 1]
"try err"
Stacktrace:
 [1] test12806() at /home/tyson/julia/error.jl:8
 [2] top-level scope at /home/tyson/julia/error.jl:21
 [3] include_relative(::Module, ::String) at /nix/store/0xhszcg5kaw5kl4z2lyhnwsb2b9k8z27-julia-1.3.0/lib/julia/sys.so:?
 [4] include(::Module, ::String) at /nix/store/0xhszcg5kaw5kl4z2lyhnwsb2b9k8z27-julia-1.3.0/lib/julia/sys.so:?
 [5] exec_options(::Base.JLOptions) at /nix/store/0xhszcg5kaw5kl4z2lyhnwsb2b9k8z27-julia-1.3.0/lib/julia/sys.so:?
 [6] _start() at /nix/store/0xhszcg5kaw5kl4z2lyhnwsb2b9k8z27-julia-1.3.0/lib/julia/sys.so:?
```

c42f:
When using `return` or `break` nested inside `finally` handlers,
exception stack lowering failed to pop exceptions from the stack
correctly:

* For `return`, the exception stack was not popped at all. If done
  inside a loop this could eventually cause the runtime to run out of
  memory.
* For `break`, the exception stack was popped too early, causing subtle
  inconsistency in intermediate finally handlers.

Fix these issues by storing the current exception token stack with the
current finally handler information and using it to pop the stack before
jumping into the finally block.

Fixes #34579

CC @KristofferC

### DIFF
--- a/src/julia-syntax.scm
+++ b/src/julia-syntax.scm
@@ -3711,10 +3711,10 @@ f(x) = yt(x)
         (label-counter 0)     ;; counter for generating label addresses
         (label-map (table))   ;; maps label names to generated addresses
         (label-nesting (table)) ;; exception handler and catch block nesting of each label
-        (finally-handler #f)  ;; `(var label map level)` where `map` is a list of `(tag . action)`.
-                              ;; To exit the current finally block, set `var` to integer `tag`,
-                              ;; jump to `label`, and put `(tag . action)` in the map, where `action`
-                              ;; is `(return x)`, `(break x)`, or a call to rethrow.
+        (finally-handler #f)  ;; Current finally block info: `(var label map level tokens)`
+                              ;; `map` is a list of `(tag . action)` which will
+                              ;; be emitted at the exit of the block. Code
+                              ;; should enter the finally block via `enter-finally-block`.
         (handler-goto-fixups '())  ;; `goto`s that might need `leave` exprs added
         (handler-level 0)     ;; exception handler nesting depth
         (catch-token-stack '())) ;; tokens identifying handler enter for current catch blocks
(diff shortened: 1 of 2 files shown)


src/julia-syntax.scm
```diff
@@ -3732,14 +3732,27 @@ f(x) = yt(x)
...
+        ;; which will run at finally block exit.
         (set-car! (cddr finally-handler) (cons (cons tag action) tags))
         (emit `(= ,(car finally-handler) ,tag))
         (if need-goto
-            (begin (emit `(leave ,(+ 1 (- handler-level (cadddr finally-handler)))))
-                   (emit `(goto ,(cadr finally-handler)))))
+            (let ((label (cadr finally-handler))
+                  (dest-handler-level (cadddr finally-handler))
+                  (dest-tokens        (caddddr finally-handler)))
+              ; Leave current exception handling scope and jump to finally block
```
JeffBezanson, line 3751:
```suggestion
              ;; Leave current exception handling scope and jump to finally block
```

JeffBezanson (approved):

PR #36488, opened 2020-06-30
Tests for in-place kron

kshyatt:
Just some [missed methods](https://codecov.io/gh/JuliaLang/julia/src/master/stdlib/LinearAlgebra/src/dense.jl#L415).

### DIFF
--- a/stdlib/LinearAlgebra/test/dense.jl
+++ b/stdlib/LinearAlgebra/test/dense.jl
@@ -399,12 +399,12 @@ end
     a = [1.0, 0.0]
     b = [0.0, 1.0]
     @test kron!([1.0, 0.0], b, 0.5) == [0.0; 0.5]
-    b = [0.0, 1.0]
+    @test kron!([1.0, 0.0], 0.5, b) == [0.0; 0.5]
     c = Vector{Float64}(undef, 4)
-    c = kron!(c, a, b)
+    kron!(c, a, b)
     @test c == [0.0; 1.0; 0.0; 0.0]
     c = Matrix{Float64}(undef, 2, 2)
-    c = kron!(c, a, b')
+    kron!(c, a, b')
     @test c == [0.0 1.0; 0.0 0.0]
 end
 
--- a/stdlib/LinearAlgebra/test/dense.jl
+++ b/stdlib/LinearAlgebra/test/dense.jl
@@ -395,6 +395,19 @@ end
     @test kron(b',2) == [8 10 12]
 end
 
+@testset "kron!" begin
+    a = [1.0, 0.0]
+    b = [0.0, 1.0]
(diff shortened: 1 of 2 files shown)


stdlib/LinearAlgebra/test/dense.jl
```diff
@@ -395,6 +395,19 @@ end
     @test kron(b',2) == [8 10 12]
 end
 
+@testset "kron!" begin
+    a = [1.0, 0.0]
+    b = [0.0, 1.0]
+    @test kron!([1.0, 0.0], b, 0.5) == [0.0; 0.5]
+    b = [0.0, 1.0]
```
dkarrasch, line 402:
```suggestion
```
This line is redundant.

stdlib/LinearAlgebra/test/dense.jl
```diff
@@ -395,6 +395,19 @@ end
...
+    c = Vector{Float64}(undef, 4)
+    c = kron!(c, a, b)
```
dkarrasch, line 404: Since `kron!` is in-place, this call looks a bit strange, but we could make it a return value test:
```
    d = kron!(c, a, b)
    @test d === c
```
or simply replace it with
```suggestion
    kron!(c, a, b)
```
? I guess we don't test the return value usually, so the shorter call may be more appropriate.

stdlib/LinearAlgebra/test/dense.jl
```diff
@@ -395,6 +395,19 @@ end
...
+    @test c == [0.0; 1.0; 0.0; 0.0]
+    c = Matrix{Float64}(undef, 2, 2)
+    c = kron!(c, a, b')
```
dkarrasch, line 407:
```suggestion
    kron!(c, a, b')
```

stdlib/LinearAlgebra/test/dense.jl
```diff
@@ -395,6 +395,19 @@ end
     @test kron(b',2) == [8 10 12]
 end
 
+@testset "kron!" begin
+    a = [1.0, 0.0]
+    b = [0.0, 1.0]
+    @test kron!([1.0, 0.0], b, 0.5) == [0.0; 0.5]
```
dkarrasch, line 401: I'm confused we don't have a
```julia
kron!(c::AbstractVecOrMat, a:: Number, b::AbstractVecOrMat) = mul!(c, a, b)
```
method. It wouldn't get called by `kron(a::Number, b::AbstractVecOrMat)`, but the method you're testing here doesn't get called by `kron` either. Maybe this would be an opportunity to add that simple method and add a test with flipped arguments here?

dkarrasch (changes requested):
Thanks, @kshyatt. I've got some small suggestions for consideration.

Issue #36440, opened 2020-06-26
API for setting default env for `run()` invocations


staticfloat:
JLL packages would benefit greatly from an API to set default environment variables for future `run()` invocations; like a default `setenv()` that other `setenv()` invocations eventually get merged into.  We don't want to modify `ENV` itself, as that is thread-unsafe.

Proposed API: Provide a `with_child_env(pairs...)` API just like `setenv` and `withenv`.  This will set a dictionary that is merged into by `run()` that causes child processes to be initialized with that environment.

StefanKarpinski:
What had been proposed in the previous github issue discussing that changing `ENV` is inherently thread-unsafe was that we initialize `ENV` as a normal `Dict{String,String}` that is a copy of the real environment and when spawning child tasks we populate the child tasks environment from `ENV`. Would that suffice to cover what you want? The only case that didn't cover was communicating with in-process libraries that are configured via the environment, which is hopefully rare and which we could potentially have some sort of very slow stop-the-world API for doing.

staticfloat:
In the world of binary dependencies, we have to `setenv()` (right now, done by `ENV[key] = value`) all the time, as that's the main way that we are able to communicate to those libraries.  Two examples off the top of my head:

* https://github.com/JuliaGraphics/Cairo.jl/blob/a591b52ccc809d3179c6cd1dca2eadb947e5986e/src/Cairo.jl#L22-L23
* https://github.com/JuliaGraphics/Gtk.jl/blob/4b929f248443ebefc780fc1b6d698b8268bea2f7/src/Gtk.jl#L111-L113

From what I have read online, as long as we're (a) using glibc, and (b) loading external libraries (that can start their own threads) it will always be unsafe to touch the environment at all, because if at any point anyone does a `getenv()` while someone else does a `setenv()`, bad things can happen.  That being said, if we assume that most programs are generally well-behaved, putting a lock around it to prevent Julia from stepping on its own toes should probably work for most things.

Your proposed changes would work for this usecase; we wouldn't even need to change anything; using `withenv()` as we're doing now would continue to "just work", although the Julia packages we would need to alter our `ENV[key] = value` mappings to use whatever `setenv()` API we end up creating.

StefanKarpinski:
Unclear if that's a "yes" or a "no" on if my proposed approach would work...

staticfloat:
Yes, it will work.  That's what I was trying to communicate in my last paragraph.

tkf:
[...]

PR #36494, opened 2020-06-30
ENV as a normal Dict: populate at boot, copy when spawning children

StefanKarpinski:
Instead of modifying the process environment in-place, which is inherently thread-unsafe because of how most `libc` libraries are implemented, this changes the `ENV` global to be populated from the process environment at Julia's startup time and when starting a child process it populates the child's environment with a copy of `ENV` so the process's environment is never modified.

### DIFF
--- a/base/initdefs.jl
+++ b/base/initdefs.jl
@@ -25,8 +25,7 @@ Stop the program with an exit code. The default exit code is zero, indicating th
 program completed successfully. In an interactive session, `exit()` can be called with
 the keyboard shortcut `^D`.
 """
-exit(n) = ccall(:jl_exit, Cvoid, (Int32,), n)
-exit() = exit(0)
+exit(n::Integer = 0) = ccall(:jl_exit, Cvoid, (Int32,), n)
 
 const roottask = current_task()
 
--- a/base/initdefs.jl
+++ b/base/initdefs.jl
@@ -19,6 +19,67 @@ An array of the command line arguments passed to Julia, as strings.
 const ARGS = String[]
 
 """
+    ENV
+
+A dictionary that is populated from the initial process environment and whose contents
+are used to populate the environments of spawned subprocesses.
+
+!!! note
+    As of Julia 1.6, this is a plain `Dict` rather than a special type. When you modify
+    it, the process environment is not modified directly, rather when a child process
+    is started, the value of the `ENV` dict is used to initialize the environment of
+    the child process. This means that `ENV` cannot be used to communicate via the
+    current proccess's environment with libraries.
+"""
+const ENV = Dict{String,String}()
+
+if !Sys.iswindows()
+    function init_env()
+        empty!(ENV)
+        idx = 0
+        while true
+            entry = ccall(:jl_environ, Any, (Int32,), idx)
+            entry === nothing && break
+            entry = entry::String
+            m = match(r"^(.*?)=(.*)$"s, entry)
+            m === nothing && error("malformed environment entry: $entry")
+            ENV[m.captures[1]] = m.captures[2]
+            idx += 1
+        end
+    end
+else
+    function winuppercase(s::AbstractString)
+        isempty(s) && return s
+        LOCALE_INVARIANT = 0x0000007f
+        LCMAP_UPPERCASE  = 0x00000200
+        ws = transcode(UInt16, String(s))
+        result = ccall(:LCMapStringW, stdcall, Cint,
+            (UInt32, UInt32, Ptr{UInt16}, Cint, Ptr{UInt16}, Cint),
+            LOCALE_INVARIANT, LCMAP_UPPERCASE, ws, length(ws), ws, length(ws))
+        systemerror(:LCMapStringW, iszero(result))
+        return transcode(String, ws)
+    end
+
+    function init_env()
+        env = ccall(:GetEnvironmentStringsW, stdcall, Ptr{UInt16}, ())
+        try pos = env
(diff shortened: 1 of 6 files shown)


Keno:
I'm ok with it in general, but I'm not quite sure we can do it in 1.x, since there's certainly packages relying on being able to modify `ENV` and having that picked up by C libraries that they load.

StefanKarpinski:
There was a lot of concern in a recent discussion thread about the fact that modifying `ENV` is currently not thread-safe and this was the only viable solution anyone could come up with. Is that concern about thread safety not actual?

Keno:
It's a problem, but somewhat tempered for the moment by threading being not very common and even when people use threading having these kinds of modifications happen mostly at load time. I'm just not sure we can do it now. At the very least, we'd need to survey existing packages and have a list of things that need to get fixed.

base/initdefs.jl
```diff
@@ -18,15 +18,75 @@ An array of the command line arguments passed to Julia, as strings.
...
+are used to populate the environments of spawned subprocesses.
+
+!!! note
+    As of Julia 1.6, this is a plain `Dict` rather than a special type. When you modify
+    it, the process environment is not modified directly, rather when a child process
+    is started, the value of the `ENV` dict is used to initialize the environment of
+    the child process. This means that `ENV` cannot be used to communicate via the
+    current proccess's environment with libraries.
+"""
+const ENV = Dict{String,String}()
```
tkf, line 34: Don't you need a lock?

tkf:
This is a breaking change because, as discussed in #34726, modifying the process environment is the only way to communicate with some C libraries.

I think we need to make the thread-safe version opt-in with something like `using Future: ENV, withenv`.  If the new API is not going to mutate/reflect the process environment, I think it might be better to use some kind of context variables system (e.g., #35833) to get rid of the lock completely.

base/initdefs.jl
StefanKarpinski: If it were possible to lock on an arbitrary object, then I'd say it's the callers responsibility to do the locking, but that's not possible so I guess we either need an official `ENV_LOCK` object or need to make this a threadsafe dict type, which I guess would therefore need to live in Base and might as well be exported.

DilumAluthge: A thread-safe Dict would be useful for many use-cases in addition to this one.

StefanKarpinski:
I don't really have a dog in this fight: I neither particularly care about the libc thread safety issues, nor do I particularly care about C libraries that need to be communicated to via environment variables. I made this PR because I was looking at the relevant code and the upshot of the discussion in https://github.com/JuliaLang/julia/issues/34726 seemed to be that this was the best approach. So if people don't want this, feel free to close this PR.

base/initdefs.jl
DilumAluthge: Also, regarding being able to lock on arbitrary objects, this PR doesn't implement that directly but implements a struct that bundles objects and their locks together: https://github.com/JuliaLang/julia/pull/34400

DilumAluthge:
We could do a PkgEval run and fix the packages that break?

tkf:
I think there are three concurrency concerns:

* _Trying_ to make C-level environment variables data-race free (DRF), at least when used via non-`unsafe_` Julia API.  Of course, this can only be a best-effort basis API since C libraries might mutate environment variables.
* Making `ENV[key] = value` and `ENV[key]` DRF.
* Making `withenv(f, ...)` or an equivalent API "safe"; i.e., concurrent executions rollback the set of environment variables to the one before. (This is what @staticfloat requested in #36440)

(The first two points would be identical if we keep `ENV` as the wrapper of the C-level environment variable.)

Note that using a lock is an OK solution for the first two points but rather terrible one for the last point.  This is because `withenv(f, ...)` has to hold a lock while executing `f`.  It'd make things DRF but you can't run, e.g., multiple processes concurrently using different executables installed via JLLs.  If the lock is only at `getindex`/`setindex!`-level, we can't grantee that the environment variables are consistent during and after executing `withenv` (although it's DRF).

On top of these complications, there is a concern of backward compatibility.  I think the best approach is to just add a lock in the current `EnvDict` (hope-for-the-best "DRF") _and_ then implement Plan 1 of https://github.com/JuliaLang/julia/issues/34726#issuecomment-586138687 in `Future` (which requires something like #35833).

> We could do a PkgEval run and fix the packages that break?

I think we are pretty sure that this is breaking.  IIUC, Cairo.jl and Gtk.jl examples in https://github.com/JuliaLang/julia/issues/36440#issuecomment-650308864 using environment variables to configure external libraries.  Even [`Base` uses it](https://github.com/JuliaLang/julia/blob/35c1f87176fecab5a8e8077fb3a62275363a5aa5/base/Base.jl#L382-L388).

Also (it's not particularity criticizing your comment, @DilumAluthge), I think we need to remember that there are tons of private code that is invisible to PkgEval.  I think PkgEval should primary be used to catch the breaking changes we _missed_ in the review but not for breaking the backward compatibility promise on purpose (although I know there are exceptions).

base/initdefs.jl
tkf: > If it were possible to lock on an arbitrary object, then I'd say it's the callers responsibility to do the locking,

All the callers must agree on one single lock.  So, not providing it as the part of API pretty much makes it impossible to use a lock with it.

DilumAluthge:
That's a good point. Fixing the PkgEval failures would not be sufficient in this case.

PR #36501, opened 2020-07-01
Fixed @test return value on Error (#31495)

### LINKED ISSUE #31495: return types of @test don't match the documentation
The docs says

> Returns a `Pass` `Result` if it does, a `Fail` `Result` if it is
`false`, and an `Error` `Result` if it could not be evaluated.

but (note `Tuple`s)

```julia
julia> VERSION
v"1.2.0-DEV.557"

julia> using Test

julia> @testset "example" begin
           @show typeof(@test 1 == 1)
           @show typeof(@test 1 == 2)
           @show typeof(@test (1 - "a"))
       end
typeof(#= REPL[15]:2 =# @test(1 == 1)) = Test.Pass
example: Test Failed at REPL[15]:3
  Expression: 1 == 2
   Evaluated: 1 == 2
Stacktrace:
 [1] top-level scope at show.jl:573
 [2] top-level scope at REPL[15]:3
 [3] top-level scope at /home/tamas/src/julia-git/usr/share/julia/stdlib/v1.2/Test/src/Test.jl:1113
 [4] top-level scope at REPL[15]:2
typeof(#= REPL[15]:3 =# @test(1 == 2)) = Tuple{Test.Fail,Array{Union{Ptr{Nothing}, Base.InterpreterIP},1}}
example: Error During Test at REPL[15]:4
  Test threw exception
  Expression: 1 - "a"
  MethodError: no method matching -(::Int64, ::String)
  Closest candidates are:
    -(::Union{Int128, Int16, Int32, Int64, Int8, UInt128, UInt16, UInt32, UInt64, UInt8}) at int.jl:51
    -(::T<:Union{Int128, Int16, Int32, Int64, Int8, UInt128, UInt16, UInt32, UInt64, UInt8}, !Matched::T<:Union{Int128, Int16, Int32, Int64, Int8, UInt128, UInt16, UInt32, UInt64, UInt8}) where T<:Union{Int128, Int16, Int32, Int64, Int8, UInt128, UInt16, UInt32, UInt64, UInt8} at int.jl:52
    -(::Union{Int16, Int32, Int64, Int8}, !Matched::BigInt) at gmp.jl:470
    ...
  Stacktrace:
   [1] top-level scope at show.jl:573
   [2] top-level scope at REPL[15]:4
   [3] top-level scope at /home/tamas/src/julia-git/usr/share/julia/stdlib/v1.2/Test/src/Test.jl:1113
   [4] top-level scope at REPL[15]:2
  
typeof(#= REPL[15]:4 =# @test(1 - "a")) = Tuple{Test.Error,Bool}
Test Summary: | Pass  Fail  Error  Total
example       |    1     1      1      3
ERROR: Some tests did not pass: 1 passed, 1 failed, 1 errored, 0 broken.
```

cf [discussion](https://discourse.julialang.org/t/print-debug-info-for-failed-test/22311/) where I wanted to condition on these types to print some extra information.

tisztamo:
This is directly mergeable and closes the bug #31495 by copying the two-liner proposed there by @tpapp , thus I do not feel comfortable prefixing RFC.

But this is also my first contribution as a [boy scout](https://www.oreilly.com/library/view/97-things-every/9780596809515/ch08.html) and it would be great to get some feedback, especially because adding the missing tests was nontrivial and is not without issues:

- I was only able to create the missing tests for `record(::DefaultTestSet, ::T <: Result)` methods, but not for `@test` itself.
- Awkward output is printed, as seen here:

Screenshot from 2020-07-01 16-01-03

I think this is acceptable when a test system tests itself, but I would be happy to fix it. I just don't know how to start.

### DIFF
--- a/stdlib/Test/src/Test.jl
+++ b/stdlib/Test/src/Test.jl
@@ -731,9 +731,9 @@ end
 
 # Records nothing, and throws an error immediately whenever a Fail or
 # Error occurs. Takes no action in the event of a Pass or Broken result
-record(ts::FallbackTestSet, res::Union{Pass,Broken}) = res
-function record(ts::FallbackTestSet, res::Union{Fail,Error})
-    println(res)
+record(ts::FallbackTestSet, t::Union{Pass,Broken}) = t
+function record(ts::FallbackTestSet, t::Union{Fail,Error})
+    println(t)
     throw(FallbackTestSetException("There was an error during testing"))
 end
 # We don't need to do anything as we don't record anything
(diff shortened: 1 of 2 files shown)


tpapp:
Thanks for fixing this and providing a test, but I am not sure why you renamed the function arguments in unrelated code, it will make them inconsistent with the rest of the file.

tisztamo:
> Thanks for fixing this and providing a test, but I am not sure why you renamed the function arguments in unrelated code, it will make them inconsistent with the rest of the file.

I changed the names in methods of `record()` that I have found in this file, because they seemed to be inconsistent with the documentation:

https://github.com/JuliaLang/julia/blob/f9ee4dbd1f0011b5fd739e098c3cfc1562fb4323/stdlib/Test/src/Test.jl#L669-L677 

II failed to read other parts of the file, but now I see that `t` is used in quite a few places. As it always refers to a subtype of `Result` I still think that `res` would be a better name in every case, recovering file-level consistency. But I understand that this is subjective.

What do you think is the way to go? Should I rename them back to t, leave it as it is, change the documentation or change all occurrences of `t` to `res`?

tpapp:
Generally I would suggest going for minimal changes in a PR like this.

tisztamo:
Reverted the renaming. Can I do anything else to improve this PR?

tpapp (approved):

tpapp:
I think it is fine, you just have to wait for someone with commit rights to the repo to review it. Please give it some time, and ping here if it does not happen within a few days.

tisztamo:
bump

StefanKarpinski:
Thanks for the fix and the bump!

Issue #35215, opened 2020-03-22
Julia 1.4 fails on startup (AMD Phenom on Linux)


alea54:
uname -a 
Linux odie 4.15.0-88-generic #88-Ubuntu SMP Tue Feb 11 20:11:34 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

[cpuinfo.txt](https://github.com/JuliaLang/julia/files/4364650/cpuinfo.txt)
[errjulia-1.4.txt](https://github.com/JuliaLang/julia/files/4364651/errjulia-1.4.txt)

jdadavid:
Similar error here, on AMD Turion :

uname -a
Linux qube.jdad.org 5.5.9-desktop-1.mga7 #1 SMP Thu Mar 12 08:02:44 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

[err-julia-1.4.0-turion.txt](https://github.com/JuliaLang/julia/files/4364973/err-julia-1.4.0-turion.txt)
[cpuinfo-turion.txt](https://github.com/JuliaLang/julia/files/4364974/cpuinfo-turion.txt)

Attaching also versioninfo from (working) julia-1.3.1
[julia-1.3-versioninfo-turion.txt](https://github.com/JuliaLang/julia/files/4364975/julia-1.3-versioninfo-turion.txt)

ViralBShah:
Did you happen to try RC1 or RC2?

benz0li:
Fails on startup (QEMU Virtual CPU):

```bash
$ julia
LLVM ERROR: 64-bit code requested on a subtarget that doesn't support it!
```

```bash
$ uname -a
Linux aa614eb95626 4.15.0-91-generic #92-Ubuntu SMP Fri Feb 28 11:09:48 UTC 2020 x86_64 GNU/Linux
```

[cpuinfo.txt](https://github.com/JuliaLang/julia/files/4365416/cpuinfo.txt)

---

Related to https://github.com/llvm/llvm-project/commit/b7b353be6044051cc4fa8888dd3d86a2a1b56d8c?

alea54:
@ViralBShah : Just tried. Both RC1 and RC2 start normally.

fredrikekre:
Has to be the new llvm binaries then I guess? https://github.com/JuliaLang/julia/compare/v1.4.0-rc2...v1.4.0

fabmazz:
[...]

PR #36502, opened 2020-07-01
Update X86 feature and CPU detection

yuyichao:
First three commits assumes LLVM >= 5.0
Last two commits assumes LLVM >= 7.0 (really 8.0 but nothing was needed from 8.0)

At least the first commit should be backported to all versions that supports LLVM 8.0 ref https://github.com/JuliaLang/julia/issues/35215#issuecomment-631496751

### DIFF
--- a/src/processor_x86.cpp
+++ b/src/processor_x86.cpp
@@ -195,7 +195,7 @@ static constexpr CPUSpec<CPU, feature_sz> cpus[] = {
     {"generic", CPU::generic, CPU::generic, 0, Feature::generic},
     {"bonnell", CPU::intel_atom_bonnell, CPU::generic, 0, Feature::bonnell},
     {"silvermont", CPU::intel_atom_silvermont, CPU::generic, 0, Feature::silvermont},
-    {"goldmont", CPU::intel_atom_goldmont, CPU::generic, 50000, Feature::goldmont},
+    {"goldmont", CPU::intel_atom_goldmont, CPU::generic, 0, Feature::goldmont},
     {"core2", CPU::intel_core2, CPU::generic, 0, Feature::core2},
     {"yonah", CPU::intel_yonah, CPU::generic, 0, Feature::yonah},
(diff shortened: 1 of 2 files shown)


src/processor_x86.cpp
```diff
@@ -840,6 +935,15 @@ get_llvm_target_noext(const TargetData<feature_sz> &data)
     features.push_back("+sse2");
     features.push_back("+mmx");
     features.push_back("+fxsr");
+#ifdef _CPU_X86_64_
+    // This is required to make LLVM happy if LLVM's feature based CPU arch guess
+    // returns a value that may not have 64bit support.
+    // This can happen with vitalization.
```
vchuravy, line 941:
```suggestion
    // This can happen with virtualization.
```

vchuravy (approved):

src/processor_x86.cpp
yuyichao: Ooops....

KristofferC:
Based on 

> First three commits assumes LLVM >= 5.0
> Last two commits assumes LLVM >= 7.0 (really 8.0 but nothing was needed from 8.0)

I assume all 5 commits should be backported to 1.5 (so I did that in https://github.com/JuliaLang/julia/pull/36577). You could take a look if you want to see that it looks reasonable, there was only one fairly trivial (hopefully) merge conflict.

yuyichao:
All 5 should be safe, just that the latter 3 has a higher chance of mistake/introducing errors....

https://github.com/JuliaLang/julia/pull/36502/commits/6fd5b82b2fb704346871579be15dc4ab2459de79 seems to be missing? That's actually the only bugfix that should be backported even if others don't. That's probablly also the origin of the conflict.

KristofferC:
> 6fd5b82 seems to be missing?

Hm, yes. So I did mess something up. Will fix it.

KristofferC:
Added

KristofferC:
I thought about it a bit and maybe it is better to just backport the bugfix commit since we are so late in the release cycle. I changed the backport branch to that.

PR #36503, opened 2020-07-01
Small fix in nonmissingtype documentation

martincornejo:
This corrects the markdown string of the `nonmissingtype` documentation to include the message `This function is exported as of Julia 1.3.` inside the admonition block.


martincornejo:
See https://docs.julialang.org/en/v1/base/base/#Base.nonmissingtype

PR #25796, opened 2018-01-29
Fix some overly optimistic fastpaths in subtyping

martinholters:
Fixes #25430, see discussion therein.

WIP because it leads to this regression (hence the disabled core test):
```julia
julia> struct T22624{A,B,C}; v::Vector{T22624{Int64,A}}; end
ERROR: StackOverflowError:
```
Hints welcome about where I should look to fix this!

PR #36504, opened 2020-07-01
RFC: update jltypes reference diagram for the current v1.6 type system

vtjnash:
Attempted updated of the old type-system categorization diagram to show some of the intent with the current system: https://github.com/JuliaLang/julia/blob/0b5777f1cd0ed1cd193eee08192412d07b2f1ecc/doc/src/assets/jltypes.svg

(somewhat inspired by https://github.com/JuliaLang/julia/pull/25796)


JeffBezanson:
Found a couple issues:
- The `isa` arrows from UnionAll and Union to DataType are pointing the wrong direction
- I would not connect `Union` and `Union{}`. Instead put `TypeofBottom` next to Union.
- `isa` arrow from `Union{}` to `TypeofBottom` goes the wrong way
- For the criteria for when an immutable is boxed, I would say - might have undefined fields, - is self-referential, or - has too many pointers.

vtjnash:
While I think this was a helpful exercise, I think we don't need this anymore

PR #32446, opened 2019-06-28
Fix REPL test result uniqueness

Moelf:
Fixes #32377

Makes the behavior match whats in production code? Since `REPL` can come from Package and module.
https://github.com/JuliaLang/julia/blob/67cdc550a0c5d615c4071b7fae12caaf85d01e30/stdlib/REPL/src/REPL.jl#L363

PR #36506, opened 2020-07-01
fix #32377: `make test-REPL` fails

### LINKED ISSUE #32377: Base.runtests("REPL") fails, but not when run together with some other testset

simeonschaub:
This fixes the issue locally for me.
closes #32377

### DIFF
--- a/stdlib/REPL/test/replcompletions.jl
+++ b/stdlib/REPL/test/replcompletions.jl
@@ -101,6 +101,9 @@ test_scomplete(s) =  map_completion_text(shell_completions(s,lastindex(s)))
 test_bslashcomplete(s) =  map_completion_text(bslash_completions(s,lastindex(s))[2])
 test_complete_context(s) =  map_completion_text(completions(s,lastindex(s),Main.CompletionFoo))
 
+module M32377 end
+test_complete_32377(s) = map_completion_text(completions(s,lastindex(s), M32377))
+
 let s = ""
     c, r = test_complete(s)
     @test "CompletionFoo" in c
@@ -109,10 +112,10 @@ let s = ""
 end
 
 let s = "using REP"
-    c, r = test_complete(s)
+    c, r = test_complete_32377(s)
     @test count(isequal("REPL"), c) == 1
     # issue #30234
-    @test !Base.isbindingresolved(Main, :tanh)
+    @test !Base.isbindingresolved(M32377, :tanh)
 end
 
 let s = "Comp"


stdlib/REPL/test/replcompletions.jl
```diff
@@ -101,6 +101,9 @@ test_scomplete(s) =  map_completion_text(shell_completions(s,lastindex(s)))
 test_bslashcomplete(s) =  map_completion_text(bslash_completions(s,lastindex(s))[2])
 test_complete_context(s) =  map_completion_text(completions(s,lastindex(s),Main.CompletionFoo))
 
+module M32377 end
+test_complete_32377(s) = map_completion_text(completions(s,lastindex(s), M32377))
+
```
timholy, line 106: I'm confused by the apparent lack of a module-end statement.

simeonschaub:
Oh, I just saw #32446. It seems like this should be a less controversial change, since I think it still preserves the original intent of the test. Could someone review?

simeonschaub:
@vtjnash You had some thoughts on #32446, do you think this is the right way to go?

stdlib/REPL/test/replcompletions.jl
```diff
@@ -101,6 +101,9 @@ test_scomplete(s) =  map_completion_text(shell_completions(s,lastindex(s)))
 test_bslashcomplete(s) =  map_completion_text(bslash_completions(s,lastindex(s))[2])
 test_complete_context(s) =  map_completion_text(completions(s,lastindex(s),Main.CompletionFoo))
 
+module M32377 end
```
vtjnash, line 104: I'd be interested to know if it still works with:
```suggestion
module M32377
    using REPL
end
```
and if not, confirmation that it doesn't represent a real issue.

But yes, this looks like a great solution!

vtjnash (approved):

PR #36507, opened 2020-07-01
Annotate (most) Julia reference accesses as atomic unordered

vtjnash:
Per discussion in https://github.com/JuliaLang/julia/pull/35535 (though I made a new PR as I wanted to preserve the current state of that branch for comparison), this works towards guaranteeing that references stored in memory are valid and cannot corrupt the GC. It's free on essentially all processors, and just requests the compiler to respect our expectations of that.

### DIFF
--- a/src/codegen.cpp
+++ b/src/codegen.cpp
@@ -2223,7 +2223,10 @@ static jl_cgval_t emit_globalref(jl_codectx_t &ctx, jl_module_t *mod, jl_sym_t *
         if (bnd->constp) {
             return mark_julia_const(bnd->value);
         }
-        return mark_julia_type(ctx, tbaa_decorate(tbaa_binding, ctx.builder.CreateLoad(bp)), true, (jl_value_t*)jl_any_type);
+        LoadInst *v = ctx.builder.CreateAlignedLoad(T_prjlvalue, bp, sizeof(void*));
+        v->setOrdering(AtomicOrdering::Unordered);
+        tbaa_decorate(tbaa_binding, v);
+        return mark_julia_type(ctx, v, true, (jl_value_t*)jl_any_type);
     }
     // todo: use type info to avoid undef check
     return emit_checked_var(ctx, bp, name, false, tbaa_binding);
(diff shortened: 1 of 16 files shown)


Keno:
What's the thing that `relaxed` guarantees? Basically just that there are no torn writes, so it doesn't which value you see, but it guarantees that any value you see was written at some point?

vtjnash:
Correct: this mostly just ensures it doesn't turn into a memcpy. There's also a monotonicity promise with relaxed (once you observe a new value, it can't un-happen) that's absent from llvm::Unordered (and present in Java, but absent from C). That can also make hoisting difficult (it's a stronger conditions on aliasing)—but our UndefVar check depends on that being prohibited already.

Keno:
Alright, makes sense to me. Presumably x86 memcpy could still be `rep stosq`, but I guess LLVM will do that if we use the right intrinsic (or maybe the loop idiom recognition pass will do the right thing already).

src/datatype.c
```diff
@@ -1057,11 +1057,12 @@ JL_DLLEXPORT jl_value_t *jl_get_nth_field_checked(jl_value_t *v, size_t i)
 
 void set_nth_field(jl_datatype_t *st, void *v, size_t i, jl_value_t *rhs) JL_NOTSAFEPOINT
 {
+    if (rhs == NULL) // TODO: this should be invalid, but it happens frequently in ircode.c
```
Keno, line 1060: What's the purpose of this behavior change in the context of this change (don't really have an independent opinion on the change, but seems odd here).

vtjnash: We're going to potentially attempt to read the type tag (for a write barrier or to figure out how to copy the data). I suppose we should actually store the NULL, but the use case (ircode) doesn't care. It's mostly here because I had added an assert here in April, and it started failing when I tried to rebase this.

src/codegen.cpp
```diff
@@ -2209,7 +2209,10 @@ static jl_cgval_t emit_globalref(jl_codectx_t &ctx, jl_module_t *mod, jl_sym_t *
         if (bnd->constp) {
             return mark_julia_const(bnd->value);
         }
-        return mark_julia_type(ctx, tbaa_decorate(tbaa_binding, ctx.builder.CreateLoad(bp)), true, (jl_value_t*)jl_any_type);
+        LoadInst *v = ctx.builder.CreateAlignedLoad(T_prjlvalue, bp, sizeof(void*));
+        v->setOrdering(AtomicOrdering::Unordered);
+        tbaa_decorate(tbaa_binding, v);
```
Keno, line 2214: This is technically fine since it mutates in place, but I think traditionally we've used the return value of this call, in case it ever needs to rewrite the instruction or allocate  anew one.

vtjnash:
I'm not sure why it'd ever pick `rep stosq` over `mov %eax, [%eax]`, but this just kindly asks it not to.

Keno:
`rep stosq` is faster and I think still legal here (since stosq work operates on 64 bit chunks at a time).

src/codegen.cpp
vtjnash: I think we just did that for the convenience of chaining calls, since TBAA shouldn't need to replace the instruction.

src/datatype.c
Keno: In the ircode usage is the previous field always NULL? If so maybe assert that here?

vtjnash:
Ah, I leave it entirely up to LLVM to decide then. There's usually no chunks though, just single loads (followed by UndefVar checks).

src/codegen.cpp
Keno: I agree it shouldn't need to, as currently designed, but it's just the kind of thing that's bound to cause a use after free if we ever do need to change this. Maybe not worth defending against until we decide to rewrite the whole C++ part in Rust ;).

src/datatype.c
vtjnash: Yeah, it's a new object (GC hasn't run yet since allocation), otherwise we'd sometimes segfault on the write barrier a couple lines later.

Keno: Alright, I'd say let's add that assert then, since that way there's at least no silent behavior change if anybody else thought they'd be clever by using this to unset a field.

vtjnash: Might as well. We should probably also look into avoiding it in that file also, but it's not causing problems yet.

PR #36508, opened 2020-07-02
RFC: Add inference->optimize analysis forwarding mechanism

Keno:
This change attempts to be a solution to the generalized problem
encountered in #36169. In short, we do a whole bunch of analysis
during inference to figure out the final type of an expression,
but sometimes, we may need intermediate results that were
computed along the way. So far, we don't really have a great
place to put those results, so we end up having to re-compute
them during the optimization phase. That's what #36169 did,
but is clearly not a scalable solution.

I encountered the exact same issue while working on a new AD
compiler plugin, that needs to do a whole bunch of work during
inference to determine what to do (e.g. call a primitive, recurse,
or increase the derivative level), and optimizations need to have
access to this information.

This PR adds an additional `info` field to CodeInfo and IRCode
that can be used to forward this kind of information. As a proof
of concept, it forwards method match info from inference to
inlining (we do already cache these, so there's little performance
gain from this per se - it's more to exercise the infrastructure).

The plan is to do an alternative fix to #36169 on top of this
as the next step, but I figured I'd open it up for discussion first.

### DIFF
--- a/base/compiler/abstractinterpretation.jl
+++ b/base/compiler/abstractinterpretation.jl
@@ -31,13 +31,13 @@ function matching_methods(@nospecialize(atype), cache::IdDict{Any, Tuple{Any, UI
     ms, minvalid, maxvalid, ambig = matching_methods(atype, cache, max_methods, world)
     min_valid[1] = max(min_valid[1], minvalid)
     max_valid[1] = min(max_valid[1], maxvalid)
-    return ms
+    return ms, ambig
 end
 
 function abstract_call_gf_by_type(interp::AbstractInterpreter, @nospecialize(f), argtypes::Vector{Any}, @nospecialize(atype), sv::InferenceState,
                                   max_methods::Int = InferenceParams(interp).MAX_METHODS)
     if sv.currpc in sv.throw_blocks
-        return Any
+        return CallMeta(Any, false)
     end
     min_valid = UInt[typemin(UInt)]
     max_valid = UInt[typemax(UInt)]
@@ -48,19 +48,21 @@ function abstract_call_gf_by_type(interp::AbstractInterpreter, @nospecialize(f),
     if splitunions
         splitsigs = switchtupleunion(atype)
         applicable = Any[]
+        infos = MethodMatchInfo[]
         for sig_n in splitsigs
             mt = ccall(:jl_method_table_for, Any, (Any,), sig_n)
             if mt === nothing
                 add_remark!(interp, sv, "Could not identify method table for call")
-                return Any
+                return CallMeta(Any, false)
             end
             mt = mt::Core.MethodTable
-            xapplicable = matching_methods(sig_n, sv.matching_methods_cache, max_methods,
-                                           get_world_counter(interp), min_valid, max_valid)
+            xapplicable, ambig = matching_methods(sig_n, sv.matching_methods_cache, max_methods,
+                                                  get_world_counter(interp), min_valid, max_valid)
             if xapplicable === false
                 add_remark!(interp, sv, "For one of the union split cases, too many methods matched")
-                return Any
+                return CallMeta(Any, false)
             end
+            push!(infos, MethodMatchInfo(xapplicable, ambig))
             append!(applicable, xapplicable)
             thisfullmatch = _any(match->match[4], xapplicable)
             found = false
@@ -76,23 +78,25 @@ function abstract_call_gf_by_type(interp::AbstractInterpreter, @nospecialize(f),
                 push!(fullmatch, thisfullmatch)
             end
         end
+        info = UnionSplitInfo(splitsigs, infos)
     else
         mt = ccall(:jl_method_table_for, Any, (Any,), atype)
         if mt === nothing
             add_remark!(interp, sv, "Could not identify method table for call")
-            return Any
+            return CallMeta(Any, false)
         end
         mt = mt::Core.MethodTable
-        applicable = matching_methods(atype, sv.matching_methods_cache, max_methods,
+        applicable, ambig = matching_methods(atype, sv.matching_methods_cache, max_methods,
                                       get_world_counter(interp), min_valid, max_valid)
         if applicable === false
             # this means too many methods matched
             # (assume this will always be true, so we don't compute / update valid age in this case)
             add_remark!(interp, sv, "Too many methods matched")
-            return Any
+            return CallMeta(Any, false)
         end
         push!(mts, mt)
         push!(fullmatch, _any(match->match[4], applicable))
+        info = MethodMatchInfo(applicable, ambig)
     end
     update_valid_age!(min_valid[1], max_valid[1], sv)
     applicable = applicable::Array{Any,1}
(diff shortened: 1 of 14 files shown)


Keno:
Also cc @martinholters who was complaining about the lack of such a mechanism over in #36169.

martinholters:
Should the `info` be a `Vector` (defaulting to `[]` instead of `nothing`) so that we can attach multiple different pieces of extra information?

base/compiler/ssair/inlining.jl
```diff
@@ -1006,6 +1006,22 @@ function process_simple!(ir::IRCode, idx::Int, params::OptimizationParams, world
     return (sig, invoke_data)
 end
 
+# This is not currently called in the regular course, but may be needed
+# if we ever want to re-run inlining again later in the pass pipeline after
+# additional type information was discovered.
+function recompute_method_matches(atype, sv)
```
JeffBezanson, line 1012:
```suggestion
function recompute_method_matches(@nospecialize(atype), sv::OptimizationState)
```

JeffBezanson:
Saving all the method matches at each call site seems to be pretty expensive; sysimg +4%. I think we should discard those by default when the optimizer finishes.

This will need to update the `deserialize` method for `CodeInfo`. Ideally it should be able to detect older versions that don't have the new field and handle them correctly.

Keno:
Yes, agree, I think we should discard these once inlining finished (by default, external optimizers can override) and have a special representation for vectors where all stmtinfo is `norhing`. I just didn't get around to implementing that yet.

base/compiler/abstractinterpretation.jl
```diff
@@ -703,118 +707,161 @@ function argtype_tail(argtypes::Vector{Any}, i::Int)
...
+            body = UnionAll(tv, body)
+        end
+        ret = canconst ? AbstractEvalConstant(body) : Type{body}
+        return ret
+    end
+    return Any
+end
+
+# call where the function is known exactly
+function abstract_call_known(interp::AbstractInterpreter, @nospecialize(f),
```
vtjnash, line 848: Allocating all these Tuple objects seems unfortunate. We've tried to work pretty hard to remove and minimize that pattern, as it seems that it can sap performance.

Keno:
> Should the `info` be a `Vector` (defaulting to `[]` instead of `nothing`) so that we can attach multiple different pieces of extra information?

I thought about it, but I think at least for the moment, there needs to be a very clear consumer for this side channel information, since it can have semantic information. If we ever get to the point where we want to store multiple independent pieces of information here, we can re-evaluate what the best format for storing it is.

base/compiler/abstractinterpretation.jl
Keno: Should we do a Pair{Any, Any} or a specialized struct instead. Seems like that may potentially be easier on the compiler.

Keno: Reminder to self: Whatever we do here, should also apply to e.g. typeinf_edge.

vtjnash:
I've had some feeling that the intent of "CodeInfo" corresponds to just the information we want to convey to codegen. While "IRCode" is that information plus the additional per-statement information computed by inference (with the intent that "InstructionStream" helps to manage it as a StructOfArrays view). And "InferenceState" is the additional state required to compute that.

To that perception, would it make sense to add these new columns only to IRCode, but to find a way to preserve the inference result (including extra metadata like this) using that struct as the vehicle in the "InferenceResult" part of the internal caches in the AbstractInterpreter? (possibly also renaming IRCode to "AnnotatedCodeInfo," or something like that, to better hint at that distinction)

Keno:
I think that's generally sensible, though for the AD case for example, I do also want to be able to cache the stmtinfo without ever running the optimizer. That's still fine for the moment, since I'm currently managing my own caches there, but it would be unfortunate to have a design where stmtinfo can never appear in the internal caches.

base/compiler/abstractinterpretation.jl
Keno: @vtjnash does Pair{Any,Any} sound reasonable to you?

Keno:
Alright, updated after some discussion with @JeffBezanson and @vtjnash earlier this week. Changes:
1. Rather than using tuples for (rt, info), there's now a dedicated `CallMeta` type. I verified that this codegens well and as expected without allocating the extra object.
2. I also sunk the assignment to the info array into abstract_eval_statement, right after abstract_call. I like having the information in the return value for `abstract_call`, because that gets called recursively (e.g. from abstract_iteration), so it's nice to bundle up all that information, but there's really no need to also forward propagate it from abstract_eval_statement. Also, this nicely separates out information about a call from information about other kinds of statements, and gives a good place to add additional information about calls.
3. stmtinfo got removed from CodeInfo and is not an inference/IRCode-only concept

Additionally, I added a fast-path in inlining, that short-circuits any additional work if the `info` is false (which is set by inference if it determines that doing the method lookup was worthless or inlining isn't warranted for other reasons). Interestingly, this change improved sysimg build time by about 5% over master. I didn't look at it too closely, but I suppose it prevents some method table queries that inference had a fast-reject for but inlining was still doing.

base/compiler/ssair/inlining.jl
```diff
@@ -1015,12 +1038,22 @@ function assemble_inline_todo!(ir::IRCode, sv::OptimizationState)
...
+            end
+            infos = [info]
             splits = Any[sig.atype]
         else
             splits = Any[]
             for union_sig in UnionSplitSignature(sig.atypes)
                 push!(splits, argtypes_to_type(union_sig))
             end
+            if !isa(info, UnionSplitInfo)
+                infos = fill(nothing, length(splits))
```
vtjnash, line 1052: wrong object type?

base/compiler/ssair/inlining.jl
```diff
@@ -994,6 +994,22 @@ function process_simple!(ir::IRCode, idx::Int, params::OptimizationParams, world
     return (sig, invoke_data)
 end
 
+# This is not currently called in the regular course, but may be needed
+# if we ever want to re-run inlining again later in the pass pipeline after
+# additional type information was discovered.
+function recompute_method_matches(atype, sv)
+    # Regular case: Retrieve matching methods from cache (or compute them)
+    # World age does not need to be taken into account in the cache
+    # because it is forwarded from type inference through `sv.params`
```
vtjnash, line 1003: This seems confusing to state. The world age _is_ taken into account, because that's part of what `sv.params` means, and is partly why the cache is part of the params.

vtjnash (approved):
Yes, this seems better to me. I still think it may be a complicated way to specify the info-lattice (this PR simply discards the info when it's inconvenient to merge), since it can't readily handle either of the existing cases (the expression type—which calls tmerge—or the statement's edges list—which calls union), though it would be conservatively correct for adding handling of other metadata columns (like pure), so it seems okay.

base/compiler/ssair/inlining.jl
Keno: That comment was there before ;)

Keno: Though I should probably update it, because it's not the regular case anymore.

base/compiler/ssair/inlining.jl
Keno: You mean it should be Vector{Any}? Yeah, let me do that.

PR #36512, opened 2020-07-02
fix doctests broken by 36134

simeonschaub:
Looks like these were missed in #36134 
cc: @KristofferC

### DIFF
--- a/doc/src/manual/constructors.md
+++ b/doc/src/manual/constructors.md
@@ -552,7 +552,8 @@ ERROR: MethodError: no method matching SummedArray(::Vector{Int32}, ::Int32)
 Closest candidates are:
   SummedArray(::Vector{T}) where T at none:4
 Stacktrace:
- [1] top-level scope at none:1
+ [1] top-level scope
+   @ none:1
 ```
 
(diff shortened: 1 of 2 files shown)


KristofferC:
The doctests ran and passed there. Perhaps something was merged in between.

simeonschaub:
Yes, that was my guess too.

doc/src/manual/constructors.md
```diff
@@ -552,7 +552,8 @@ ERROR: MethodError: no method matching SummedArray(::Vector{Int32}, ::Int32)
 Closest candidates are:
   SummedArray(::Vector{T}) where T at none:4
 Stacktrace:
- [1] top-level scope at none:1
```
KristofferC, line 555: Can we just change these to `[...]`.

simeonschaub: Should the
```
 Closest candidates are:
  SummedArray(::Vector{T}) where T at none:4 
```
still be included? But maybe also not that important.

KristofferC: I think that can be kept.

KristofferC:
Merging early so CI don't fail in other places.

PR #36515, opened 2020-07-02
JULIA_NUM_THREADS physical->logical

jmkuhn:


doc/src/manual/environment-variables.md
```diff
@@ -186,7 +186,7 @@ a master process to establish a connection before dying.
 
 An unsigned 64-bit integer (`uint64_t`) that sets the maximum number of threads
 available to Julia. If `$JULIA_NUM_THREADS` exceeds the number of available
-physical CPU cores, then the number of threads is set to the number of cores. If
+logical CPU cores, then the number of threads is set to the number of cores. If
```
JeffBezanson, line 189:
```suggestion
CPU threads (logical cores), then the number of threads is set to the number of CPU threads. If
```
plus replace cores with threads elsewhere.

Issue #34495, opened 2020-01-24
Taking Inno (Un-)Installer seriously


rapus95:
First I want to note that the default directory changed compared to the previous version (Julia 1.3).
previous:  `AppData/Local/Julia-####`
on 1.4: `AppData/Local/Programs/Julia/Julia-####`
Was that intended?

Now, that the installer was replaced we might finally add some options to the gui (will make the NEWS.md more satisfying 😁)
Ideally I'd envision an installer with the following options:
```
Path to install: <path>

[ ] Update PATH to point to the new version (replace previous references)*
    [ ] Uninstall those previous versions from disk

* This will allow Julia to be run by typing "julia" into the command prompt or
execute window (Win+R) without any further setup.
```

The first checkbox would search for paths which point to a Julia bin directory (& contain "julia" in their path string, to prevent unexpected side effects) and replace them with the path to the new /bin. We could even consider checking that box by default if a previous Julia already is on the PATH. (Since then, installing a new version seems like an upgrade.)
The second checkbox would call the uninstaller in all directories which were found in the previous step. Here it's rather controversy which option to take by default.

This would fit nicely into the less system oriented workflow of Windows users like me 😄 (=point-and-click-adventure-like installer)

As I've no clue how the Inno setup works, the following might help:
https://stackoverflow.com/questions/3304463/how-do-i-modify-the-path-environment-variable-when-running-an-inno-setup-install
https://stackoverflow.com/questions/2000296/inno-setup-how-to-automatically-uninstall-previous-installed-version

The uninstaller on the other hand should provide some checkbox for "delete configuration" which would also delete the .julia directory. Might be marked with some waarning to only check if no further usage is wanted. Reason: Optimally people won't need to know that .julia exists. (Well, once manual cleanup to fix bugs isn't needed anymore 😛) And thus we should cover that cleanup in some way aswell.

KristofferC:
It would be good to separate the concrete change

> First I want to note that the default directory changed compared to the previous version (Julia 1.3).

into its own issue and then the more speculative can be kept here.

musm:
All of those should be doable and improve the experience.
But I  do feel like the path update suggestions might be too cute. I think chocolatey kinda already solves that problem with upgrades and so forth, but I haven't tested it recently with Julia. 

Personally I just rename to julia install to install it to a folder names `Julia-1.x` and have`Julia-1.x/bin` added to path (actually I just have  a julia calling function in my powershell profile).

[...]

Issue #34552, opened 2020-01-28
Provide a julia.bat file in the installation directory of Julia on Windows


KristofferC:
Background issue: https://github.com/JuliaLang/www.julialang.org/issues/615.

Basically, to be able to start Julia from the terminal in Windows, one has to add the `Julia-1.x.y/bin` folder to `PATH`  which I have understood is a bad idea because Windows also look in `PATH` for shared libraries to load for all applications (like `LD_LIBRARY_PATH`).

In a similar vein to the `julia.bat` file we create when compiling Julia on Windows, we could also distribute a `julia.bat`-file that is in the Julia installation directory. It would then be enough to add that directory to `PATH` (which doesn't contain any shared libraries).

cc @musm, do you think that is a reasonable idea?

visr:
It would be good to have a directory that is safe to add to the Path. Would be even better if the installer offered to do it for you, just like for instance VS Code and Atom.

Note that I would currently not consider the installation root for safe, since uninstall.exe is also there.

For Anaconda, I now only add the 'condabin' directory to my path for safety, but this is not the standard yet.

KristofferC:
> Note that I would currently not consider the installation root for safe, since uninstall.exe is also there.

Good point.

musm:
I'm not sure I  agree with the suggestion here to ship a  `.bat` file that opens up  `julia.exe`.  

The problem with adding the `.bat` is that it launches the `exe`  from a  cmd session and leads to annoying issues like https://github.com/microsoft/terminal/issues/217 

Python also dumps everything in the same folder like we do, although the don't use a `bin` folder and just dump everything in the root folder. 

Installing Julia through Chocolatey  basically solves this problem and handles the upgrade issues  in https://github.com/JuliaLang/julia/issues/34495   through the use of exe 'shims' (the issues in https://github.com/JuliaLang/julia/issues/34495 are mainly s related to 'package' level configuration). We should still try to improve things for those who don't use chocolatey.

rapus95:
Well, I also could post that at the other issue (#34495) but I think it fits here more nicely as the other issue is more about automation of setup while this one is about better accessibility.
How about adding a new directory to the ~/.julia path, something like "exported_binaries" which will hold references(or executable scripts) to all binaries that should be accessible via the path. That'd have 2 advantages:
a) Once we fully arrive at shared libraries compiled from Julia it fits well into the workflow with artifacts and packages which are also placed in ~/.julia
[...]

PR #36518, opened 2020-07-02
Incorporate an add to PATH option in Windows Inno Setup installer

musm:
This adds an **unchecked by default option** to add the Julia bin directory to path on Windows for the installer.

The code handles adding the bin to path and also automatically removes the path upon uninstallation. 

image

Part of:  https://github.com/JuliaLang/julia/issues/34495

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -32,6 +32,10 @@ Multi-threading changes
 Build system changes
 --------------------
 
+* Windows Installer now has the option to 'Add Julia to Path'. To unselect this option
+  from the commandline simply remove the tasks you do not want to be installed: e.g.
+  `./julia-installer.exe /TASKS="desktopicon,startmenu,addtopath"`, adds a desktop
+  icon, a startmenu group icon, and adds Julia to system PATH.
 
 New library functions
 ---------------------
--- a/contrib/windows/build-installer.iss
+++ b/contrib/windows/build-installer.iss
@@ -73,15 +73,20 @@ SetupIconFile={#RepoDir}\contrib\windows\julia.ico
 UninstallDisplayName={#AppNameLong}
 UninstallDisplayIcon={app}\{#AppMainExeName}
 UninstallFilesDir={app}\uninstall
+ChangesEnvironment=true
 
 
 [Languages]
 Name: "english"; MessagesFile: "compiler:Default.isl"
 
+[CustomMessages]
+AdditionalIcons=Icons:
+Other=%nOther:
 
 [Tasks]
-Name: "desktopicon"; Description: "Create a Desktop shortcut"
-Name: "startmenu"; Description: "Create a Start Menu entry"
+Name: "desktopicon"; Description: "Create a Desktop shortcut"; GroupDescription: "{cm:AdditionalIcons}";
+Name: "startmenu"; Description: "Create a Start Menu entry"; GroupDescription: "{cm:AdditionalIcons}";
+Name: "addtopath"; Description: "Add {#AppName} to PATH"; GroupDescription: "{cm:Other}"; Flags: unchecked;
 
 
 [Files]
@@ -89,8 +94,8 @@ Source: "{#SourceDir}\*"; DestDir: "{app}"; Flags: ignoreversion recursesubdirs 
 
 
 [Icons]
-Name: "{autostartmenu}\{#AppNameLong}"; Filename: "{app}\{#AppMainExeName}"; WorkingDir: "{%USERPROFILE}"; Tasks: startmenu
 Name: "{autodesktop}\{#AppNameLong}"; Filename: "{app}\{#AppMainExeName}"; WorkingDir: "{%USERPROFILE}"; Tasks: desktopicon
+Name: "{autostartmenu}\{#AppNameLong}"; Filename: "{app}\{#AppMainExeName}"; WorkingDir: "{%USERPROFILE}"; Tasks: startmenu
 
 
 [Run]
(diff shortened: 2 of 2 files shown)


vtjnash:
I'm inclined to think it should be unchecked by default, to reduce the risk of breaking other software that uses Julia internally, but happy to hear what other languages do (looks like python is non-default, while nodejs is probably default checked):

musm:
Rust adds  to path a whole bunch of `.exe` files by default, same with  vs code. 

I'm inclined to leave it selected by default, because of beginners often struggle with 'adding to path'.  I'm also happy to hear others on this.

musm:
If there aren't any objections can I merge this?

rapus95:
One remaining question for me, what will it do if Julia already is on the path?

musm:
That's a separate issue, I'd like to tackle in a different PR. This PR only adds it to the end of the path list, and the corresponding uninstaller will remove it from path list. 

For  the "update" strategy.  Right now the simplest method that I got to work, is for the installer to check if Julia is already installed and if it is, the installer will then ask  you if you want to remove the previous version. Removing the previous version will then remove it from the path and then the new installation will add the new version to path.

musm:
I posted a comment regarding that in https://github.com/JuliaLang/julia/issues/34495#issuecomment-653806259

musm:
I just read over https://github.com/JuliaLang/julia/issues/34552#issuecomment-579619366

It's an interesting proposal. 

Can you explain in more detail how the `exported_binaries` directory would work?  Like, is the idea to create symlinks from the julia bin directory to the `~/.julia/exported_binares` folder? 

> How about adding a new directory to the ~/.julia path, something like "exported_binaries" which will hold references(or executable scripts) to all binaries that should be accessible via the path. That'd have 3 advantages:

rapus95:
> Can you explain in more detail how the `exported_binaries` directory would work? Like, is the idea to create symlinks from the julia bin directory to the `~/.julia/exported_binares` folder?

The idea was that that "exported_binaries" directory will contain all files that shall be in the reach of the path variables. So right now, there probably will only be a script which leads to the right executable. So right now the idea might sound weird and unnecessary. The idea is that it'll establish a place where we/the Julia compiler later can put the "exported as binary" programs to once binary export lands.

musm:
Unchecked by default now.

image


Is that text at the bottom helpful? I kinda want to remove it.

rapus95:
I kinda like it but it only hokds information for those people who also know what the path is anyway. So I'd rather replace it with a suggestion for those who don't know what it means

musm:
Doesn't that explain it well enough? What to do you think it should say, without being too lengthy. Looking at other languages e.g. Python, GIT they do a much worse or no description at all.

rapus95:
I mean thst I'm actually very sure that there are a lot of people who don't even know what the command prompt is. But those who do don't need an explanation.

musm:
Maybe replace command prompt with terminal

rapus95:
I have a salty moment right now which makes me think whether "if you don't know if you need it, ask the person who asked you to install it" would have an appropriate amount of humour 😇😂

musm:
True best to remove It. Merging this PR after updating it with an unchecked default.

rapus95:
i think this options only use would be for people who write julia code for other people since it helps setting up script based automation. that brings me to another idea:
How about adding some function to `System` which adds the own executable to the systems path. That would have 3 advantages.
1) we can leverage julia for it (including checking for its existence or replacing/updating old links)
2) you don't need to decide before the actual installation and it also automatically leads to a function to remove it at any time.
3) it's easy to explain to someone who doesn't know coding (just run the program and type `somecommand()`)

musm:
@rapus95 please look at https://github.com/JuliaLang/julia/issues/36672 it will be the future installer and basically with `juliaup` all the issues you have are fixed.


Right now this is an incremental improvement and it can help as a stop-gap solution for newbs who don't want to go through the hassle of adding to PATH and just listen to the instructor that tells them to select this option and get going.

rapus95:
> for newbs who don't want to go through the hassle of adding to PATH and just listen to the instructor that tells them to select this option and get going.

Erm now I'm confused, *thinking*
found out I confused myself 😂😂 somehow I thought you'd remove the path option when we were talking about removing the tooltip since anything beyond "ask your instructor" probably wouldn't help anyway... Sry for always coming up with new ideas which almost run contrary to the existing concept 😅🙈

musm:
No problem! Your input is appreciated

davidanthoff:
Should this really have been merged before https://github.com/JuliaLang/julia/pull/36588 is done? This now adds _a lot_ of DLLs onto the `PATH`, pretty much everything that is shipped with Julia, which I thought was always considered not a good idea?

I had always assumed that if something gets added to the `PATH`, it should be a folder that contains _only_ a `julia.exe` or `julia.cmd` or something like that.

I should say that I'm not super clear how important that policy actually is, but it was for a long time here and I'm just surprised that it was given up without some discussion of what has changed?

musm:
It's unchecked by default....

musm:
> Should this really have been merged before #36588 is done? This now adds a lot of DLLs onto the PATH, pretty much everything that is shipped with Julia, which I thought was always considered not a good idea?

I'm not sure why that should change whether this should get merged before or after. The option is by default unchecked. This is still early in 1.6, so after staticfloat's PR the situation will be improved w.r.t to the dlls getting added to path (granted I personally haven't had this issue, but understand it's possible). If the path that we need to add to PATH does indeed need to change, it will be trivial for me to update the location that get's added to Path. So no I don't find it an issue when this got merged.

rapus95:
Tbh I always just put julia/bin to PATH w/o caring about the other files... Now I at least can make that happen w/o having to manually change the environment variables. Whether that's a good idea IMO is out of scope of the capability to do it. So in my opinion that's just a feature for convenience which makes it independent from the other issue.

PR #36521, opened 2020-07-03
Allow PartialStruct on inference edges

Keno:
At the moment, we only allow types, or completely Const inference
results to propagate along inference edges. In particular PartialStruct
is disallowed. I noticed some inference sub-optimalities as a result
(e.g. in iterating `Pair`), but more generally, I would prefer if
adding an additional return value somewhere (i.e. returning a tuple
of two values, rather than a single value) would not degrade the
inference precision for the previously returned value. This fixes
the issue by allowing `PartialStruct` to propagate along inference
edges.

### DIFF
--- a/base/compiler/tfuncs.jl
+++ b/base/compiler/tfuncs.jl
@@ -1492,18 +1492,21 @@ function return_type_tfunc(interp::AbstractInterpreter, argtypes::Vector{Any}, v
                     if isa(rt, Const)
                         # output was computed to be constant
                         return Const(typeof(rt.val))
-                    elseif hasuniquerep(rt) || rt === Bottom
-                        # output type was known for certain
-                        return Const(rt)
-                    elseif (isa(tt, Const) || isconstType(tt)) &&
-                           (isa(aft, Const) || isconstType(aft))
-                        # input arguments were known for certain
-                        # XXX: this doesn't imply we know anything about rt
-                        return Const(rt)
-                    elseif isType(rt)
-                        return Type{rt}
                     else
-                        return Type{<:widenconst(rt)}
+                        rt = widenconst(rt)
+                        if hasuniquerep(rt) || rt === Bottom
+                            # output type was known for certain
+                            return Const(rt)
+                        elseif (isa(tt, Const) || isconstType(tt)) &&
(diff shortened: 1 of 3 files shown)


vtjnash:
As you seem to have noticed, we didn't do this because we didn't want the cost of preserving these in memory. Instead, my recollection is that we intended to recompute them if we guessed that it might be worthwhile to, but I'm not sure if we do that now for any cases, from a brief look at the code for handling ReturnNode.

Keno:
How can we recompute this without basically re-doing inference? Is the memory overhead really likely to be that big? Given that we already do it for `Const`, I wouldn't have expected the additional overhead to be that big, but I can look at sysimg size numbers. If it does turn out to be significant, perhaps we can have an element limit on the partial struct?

vtjnash:
We'd redo inference. There's times before when I'd found it faster to re-run inference than keeping the extra objects in memory (due to the load time and memory pressure effects), but we've regressed significantly since then.

Keno:
Seems like we could literally just measure how long something took to infer and how big the object is and then trade them off against each other. Doesn't seem specific to `PartialStruct` of course. `Const` can have the same issues.

Keno:
Might also want to have some counter of how many times a given CodeInstance was requested and if the answer is "a lot", then raise the caching threshold.

Keno:
Looked at sysimg size on this branch, which seems to be in the noise - +0.04% sysimage size total (+0.06% in the sysimage counterbalanced by slightly smaller native code).

vtjnash:
Yeah, for starters, what do you think about generating the count and list of methods in the sysimg/tests that are impacted by this?

Keno:
Here's the list for all affected methods and their corresponding PartialStruce edge during sysimage build: https://gist.github.com/Keno/7ac695b209b0364c2780d4a0900d127b

JeffBezanson:
Looking at build times, TTFP, and sysimg size the overhead of this seems pretty negligible to me.

Keno:
From discussion: Let's change the representation of PartialStruct to not serialize Core.Compiler object into the global cache.

vtjnash (commented):
In off-line discussions, decided that this seemed good, revealed a couple issues to fix before merging

PR #36524, opened 2020-07-03
Improve precision of abstract_iteration

Keno:
Currently abstract iteration works fine for length-1 iterators, but
fails over for other small length iterators, such as Pair{Int, Int},
or common patterns where `iterate` iterates over the fields of a
small struct. Other examples include StaticVectors and constant
iterators, which should all now unroll properly up to the
MAX_TUPLE_SPLAT limit. That said, MAX_TUPLE_SPLAT is quite high
at the moment, but because abstract_iteration isn't very precise,
it's unlikely to be limiting. With this increase in precision,
we may find that MAX_TUPLE_SPLAT is too high and we should lower it.

Also note that while this is a nice improvement, performance is still
not great, since we currently can't inline these apply calls. However,
fixing that is in progress in a parallel PR and with both changes
put together, the performance improvement of splatting of small
iterators is quite sizeable.

### DIFF
--- a/base/compiler/abstractinterpretation.jl
+++ b/base/compiler/abstractinterpretation.jl
@@ -223,9 +223,7 @@ function abstract_call_method_with_const_args(interp::AbstractInterpreter, @nosp
             end
         elseif istopfunction(f, :iterate)
             itrty = argtypes[2]
-            if itrty isa Type && !issingletontype(itrty)
-                return Any
-            elseif itrty ⊑ Array
+            if itrty ⊑ Array
                 return Any
             end
         end
@@ -569,37 +567,51 @@ function abstract_iteration(interp::AbstractInterpreter, @nospecialize(itft), @n
     stateordonet === Bottom && return Any[Bottom]
     valtype = statetype = Bottom
     ret = Any[]
-    stateordonet = widenconst(stateordonet)
-    while !(Nothing <: stateordonet) && length(ret) < InferenceParams(interp).MAX_TUPLE_SPLAT
-        if !isa(stateordonet, DataType) || !(stateordonet <: Tuple) || isvatuple(stateordonet) || length(stateordonet.parameters) != 2
+    # Try to unroll the iteration up to MAX_TUPLE_SPLAT, which covers any finite
+    # length iterators, or interesting prefix
+    while true
+        stateordonet_widened = widenconst(stateordonet)
+        if stateordonet_widened === Nothing
+            return ret
+        end
+        if Nothing <: stateordonet_widened || length(ret) >= InferenceParams(interp).MAX_TUPLE_SPLAT
             break
         end
-        if stateordonet.parameters[2] <: statetype
-            # infinite (or failing) iterator
+        if !isa(stateordonet_widened, DataType) || !(stateordonet_widened <: Tuple) || isvatuple(stateordonet_widened) || length(stateordonet_widened.parameters) != 2
+            break
+        end
+        nstatetype = getfield_tfunc(stateordonet, Const(2))
+        # If there's no new information in this statetype, don't bother continuing,
+        # the iterator won't be finite.
+        if nstatetype ⊑ statetype
             return Any[Bottom]
         end
-        valtype = stateordonet.parameters[1]
-        statetype = stateordonet.parameters[2]
+        valtype = getfield_tfunc(stateordonet, Const(1))
         push!(ret, valtype)
+        statetype = nstatetype
+        stateordonet = abstract_call_known(interp, iteratef, nothing, Any[Const(iteratef), itertype, statetype], vtypes, sv)
+    end
+    # From here on, we start asking for results on the widened types, rather than
+    # the precise (potentially const) state type
+    statetype = widenconst(statetype)
+    valtype = widenconst(valtype)
+    while valtype !== Any
         stateordonet = abstract_call_known(interp, iteratef, nothing, Any[Const(iteratef), itertype, statetype], vtypes, sv)
         stateordonet = widenconst(stateordonet)
-    end
-    if stateordonet === Nothing
-        return ret
-    end
-    while valtype !== Any
         nounion = typesubtract(stateordonet, Nothing)
         if !isa(nounion, DataType) || !(nounion <: Tuple) || isvatuple(nounion) || length(nounion.parameters) != 2
             valtype = Any
             break
         end
         if nounion.parameters[1] <: valtype && nounion.parameters[2] <: statetype
+            if typeintersect(stateordonet, Nothing) === Union{}
+                # Reached a fixpoint, but Nothing is not possible => iterator is infinite or failing
+                return Any[Bottom]
+            end
             break
         end
         valtype = tmerge(valtype, nounion.parameters[1])
         statetype = tmerge(statetype, nounion.parameters[2])
-        stateordonet = abstract_call_known(interp, iteratef, nothing, Any[Const(iteratef), itertype, statetype], vtypes, sv)
-        stateordonet = widenconst(stateordonet)
     end
     push!(ret, Vararg{valtype})
     return ret
--- a/test/compiler/inference.jl
+++ b/test/compiler/inference.jl
@@ -2674,3 +2674,7 @@ partial_return_1(x) = (x, 1)
 partial_return_2(x) = Val{partial_return_1(x)[2]}
 
 @test Base.return_types(partial_return_2, (Int,)) == Any[Type{Val{1}}]
+
+# Precision of abstract_iteration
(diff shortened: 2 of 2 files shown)


Keno:
> Also note that while this is a nice improvement, performance is still
> not great, since we currently can't inline these apply calls. However,
> fixing that is in progress in a parallel PR and with both changes
> put together, the performance improvement of splatting of small
> iterators is quite sizeable.

Some numbers on this:

Before
```
julia> f(x) = (x...,)

julia> @benchmark f(1=>2)
BenchmarkTools.Trial: 
  memory estimate:  96 bytes
  allocs estimate:  3
  --------------
  minimum time:     351.580 ns (0.00% GC)
  median time:      357.165 ns (0.00% GC)
  mean time:        373.367 ns (0.97% GC)
  maximum time:     9.967 μs (96.17% GC)
  --------------
  samples:          10000
  evals/sample:     212
```
After:
```
julia> @benchmark f(1=>2)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     2.026 ns (0.00% GC)
  median time:      2.032 ns (0.00% GC)
  mean time:        2.035 ns (0.00% GC)
  maximum time:     6.816 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000
```

martinholters:
Something I long had on my to do list, but a half-hearted attempt didn't work out, so it stayed there. Glad you tackled it!

I wouldn't worry too much about the high MAX_TUPLE_SPLAT: For those cases where this PR makes a difference, i.e. where `iterate` returns a const-inferable state, it is also probably quite simple and therefore cheap to infer/const-propagate. Depending on the context, the more precise inference result might make up for the extra work. The worst case is probably where MAX_TUPLE_SPLAT is exceeded: Extra work with little profit. Code splatting lots of large StaticArrays might take a hit? A middle ground might be to have a second, lower cut-off and start `widenconst`ing (i.e. like before this PR) once that is exceeded. But let's first see whether this causes any relevant latency issues as is.

base/compiler/abstractinterpretation.jl
```diff
@@ -592,6 +599,10 @@ function abstract_iteration(interp::AbstractInterpreter, @nospecialize(itft), @n
             break
         end
         if nounion.parameters[1] <: valtype && nounion.parameters[2] <: statetype
+            if typeintersect(stateordonet, Nothing) === Union{}
+                # Reached a fixpoint, but Nothing is not possible => iterator is infinite or failing
+                return Any[Bottom]
```
martinholters, line 604: In the first iteration of this loop, the `stateordonet` has been computed using possibly const arguments, so while it has been `widenconst`ed, the type may be misleading here. E.g. just `(1:33...,)` will be infered as `Union{}` due to this.

base/compiler/abstractinterpretation.jl
```diff
@@ -567,21 +565,30 @@ function abstract_iteration(interp::AbstractInterpreter, @nospecialize(itft), @n
     stateordonet === Bottom && return Any[Bottom]
     valtype = statetype = Bottom
     ret = Any[]
-    stateordonet = widenconst(stateordonet)
-    while !(Nothing <: stateordonet) && length(ret) < InferenceParams(interp).MAX_TUPLE_SPLAT
-        if !isa(stateordonet, DataType) || !(stateordonet <: Tuple) || isvatuple(stateordonet) || length(stateordonet.parameters) != 2
+    # Try to unroll the iteration up to MAX_TUPLE_SPLAT, which covers any finite
+    # length iterators, or interesting prefix
+    while true
+        stateordonet_widened = widenconst(stateordonet)
```
martinholters, line 571:
```suggestion
        stateordonet_widened = widenconst(stateordonet)
        if stateordonet_widened === Nothing
            return ret
        end  
```
So this test can be skipped below.
(Is there a way to suggest an addition only?)

base/compiler/abstractinterpretation.jl
```diff
@@ -567,21 +565,30 @@ function abstract_iteration(interp::AbstractInterpreter, @nospecialize(itft), @n
...
+        statetype = nstatetype
         stateordonet = abstract_call_known(interp, iteratef, nothing, Any[Const(iteratef), itertype, statetype], vtypes, sv)
-        stateordonet = widenconst(stateordonet)
     end
+    valtype = widenconst(valtype)
+    statetype = widenconst(statetype)
+    stateordonet = widenconst(stateordonet)
     if stateordonet === Nothing
         return ret
     end
```
martinholters, line 594:
```suggestion
    stateordonet = abstract_call_known(interp, iteratef, nothing, Any[Const(iteratef), itertype, statetype], vtypes, sv)
    stateordonet = widenconst(stateordonet)
```
Recompute `stateordonet` with widened state.

martinholters (changes requested):
I have a proposal for fixing the `(1:33...,)` case which also fixes the `abstractarray` test, see suggestions.

martinholters:
I've noticed something else funny going on here: After `include("compiler/compiler.jl")`, displaying the result of splatting into tuple produces
```
TypeError(func=:Tuple, context="parameter", expected=Type, got=Core.Compiler.Const(val=2, actual=false))
jl_type_error_rt at ../src/rtutils.c:118
jl_f_apply_type at ../src/builtins.c:962
jl_apply at ../src/julia.h:1712 [inlined]
do_apply at ../src/builtins.c:655
argtypes_to_type at ./compiler/typeutils.jl:48
abstract_call_known at ./compiler/abstractinterpretation.jl:931
abstract_call at ./compiler/abstractinterpretation.jl:955
abstract_call at ./compiler/abstractinterpretation.jl:939
abstract_eval_statement at ./compiler/abstractinterpretation.jl:1061
typeinf_local at ./compiler/abstractinterpretation.jl:1300
typeinf_nocycle at ./compiler/abstractinterpretation.jl:1371
typeinf at ./compiler/typeinfer.jl:12
abstract_call_method_with_const_args at ./compiler/abstractinterpretation.jl:289
unknown function (ip: 0x7f6dcb9b1af7)
abstract_call_gf_by_type at ./compiler/abstractinterpretation.jl:139
abstract_call_known at ./compiler/abstractinterpretation.jl:932
abstract_call at ./compiler/abstractinterpretation.jl:955
[...]
jl_compile_method_internal at ../src/gf.c:2074
jl_compile_method_internal at ../src/gf.c:2037 [inlined]
_jl_invoke at ../src/gf.c:2344 [inlined]
jl_apply_generic at ../src/gf.c:2535
display at ../stdlib/v1.6/REPL/src/REPL.jl:218
display at ../stdlib/v1.6/REPL/src/REPL.jl:232
display at ./multimedia.jl:328
```
E.g.
```julia
julia> f() = (()...,)

julia> f();

julia> f()
TypeError(func=:Tuple, context="parameter", expected=Type, got=Core.Compiler.Const(val=2, actual=false))
[...]
()

julia> @code_typed f() # also here
CodeInfo(
1 ─     return Internal error: encountered unexpected error in runtime:
TypeError(func=:Tuple, context="parameter", expected=Type, got=Core.Compiler.Const(val=2, actual=false))
[...]
()
) => Tuple{}

```

The `argtypes` given to `argtypes_to_type` in these cases are
```julia
Array{Any, (3,)}[
  Core.Compiler.Const(val=typeof(Base.iterate)(), actual=false),
  Core.Compiler.Const(val=("#= circular reference @-", 1, " =#"), actual=false),
  Core.Compiler.Const(val=2, actual=false)]
```
and get widened to
```julia
Array{Any, (3,)}[
  typeof(Base.iterate),
  Tuple{String, Int64, String},
  Core.Compiler.Const(val=2, actual=false)]
```
No idea what to make of this...

martinholters:
Ah, maybe the `Core.Compiler.Const(val=2, actual=false)` is a result cached from before reloading `compiler.jl` and hence is not recognized by `widenconst`? And this is to be expected and `include("compiler/compiler.jl")` is just a don't-do-that? If so sorry for the noise.

Keno:
> After `include("compiler/compiler.jl")`, displaying the result of splatting into tuple produces

I can't reproduce this. I usually just use `using Revise; Revise.track(Core.Compiler)` which works fine most of the time.

Keno:
Anyway, I rotated the second loop, which I think should address the bug you found.

martinholters (approved):
LGTM. Should we add an explicit test for `(1:N...,)` with `N > MAX_TUPLE_SPLAT`?

vtjnash (approved):

Keno:
> I've noticed something else funny going on here: After `include("compiler/compiler.jl")`, displaying the result of splatting into tuple produces

This will be fixed by #36570

PR #36525, opened 2020-07-03
Multiplication/division of Dates.CompoundPeriod and Real

sostock:
This enables multiplication and division of a `CompoundPeriod` by a real number. `zero` and `one` for `CompoundPeriod`s are added as well:
```julia
julia> Year(1) + Day(1)
1 year, 1 day

julia> 2 * ans
2 years, 2 days

julia> ans / 2
1 year, 1 day

julia> zero(Dates.CompoundPeriod)
empty period

julia> one(Dates.CompoundPeriod)
1
```


quinnj (approved):
Looks ok by me; @stevengj also helped w/ CompoundPeriod semantics, if he wouldn't mind taking a look

rafaqz:
Can we get this merged?

quinnj:
@stevengj, would you mind reviewing this real quick?

stdlib/Dates/src/periods.jl
```diff
@@ -351,6 +354,14 @@ Base.show(io::IO,x::CompoundPeriod) = print(io, string(x))
 (-)(x::CompoundPeriod) = CompoundPeriod(-x.periods)
 (-)(y::Union{Period, CompoundPeriod}, x::CompoundPeriod) = (-x) + y
 
+(*)(x::CompoundPeriod, y::Real) = iszero(y) ? CompoundPeriod() : CompoundPeriod(x.periods * y)
+(*)(y::Real, x::CompoundPeriod) = x * y
+
+(*)(A::CompoundPeriod, B::AbstractArray) = Broadcast.broadcast_preserving_zero_d(*, A, B)
+(*)(A::AbstractArray, B::CompoundPeriod) = Broadcast.broadcast_preserving_zero_d(*, A, B)
```
stevengj, line 361: Can't the second definition just be `B * A`?

stdlib/Dates/src/periods.jl
```diff
@@ -351,6 +354,14 @@ Base.show(io::IO,x::CompoundPeriod) = print(io, string(x))
...
+
+(/)(x::CompoundPeriod, y::Real) = isfinite(y) ? CompoundPeriod(x.periods / y) : CompoundPeriod()
```
stevengj, line 363: Do you mean `!isinf` instead of `isfinite`?  The `isfinite` function also returns `false` for `NaN`

stevengj: And do we also need a `(\)(y::Real, x::CompoundPeriod) = x / y` method?

stdlib/Dates/src/periods.jl
sostock: It’s the same definition as for `Period`s:
https://github.com/JuliaLang/julia/blob/24d9eab45632bdb3120c9e664503745eb58aa2d6/stdlib/Dates/src/periods.jl#L97-L98
But the same applies there, we could change them both.

stdlib/Dates/src/periods.jl
sostock: I didn’t think about the `NaN` case, I will change it to `!isinf`.

`\` currently isn’t implemented for `Period`s, so I didn’t add it for `CompoundPeriod`. But I think it would be good to add both.

sostock: I have fixed the `NaN` case. I would leave `\` for another PR, since it is currently not implemented for `Period`s.

stevengj (approved):

stdlib/Dates/src/periods.jl
```diff
@@ -371,8 +371,15 @@ Base.convert(::Type{T}, x::CompoundPeriod) where T<:Period =
 (-)(x::CompoundPeriod) = CompoundPeriod(-x.periods)
 (-)(y::Union{Period, CompoundPeriod}, x::CompoundPeriod) = (-x) + y
 
+(*)(x::CompoundPeriod, y::Real) = iszero(y) ? CompoundPeriod() : CompoundPeriod(x.periods * y)
+(*)(y::Real, x::CompoundPeriod) = x * y
+
+(/)(x::CompoundPeriod, y::Real) = isinf(y) ? CompoundPeriod() : CompoundPeriod(x.periods / y)
```
stevengj, line 377: I'm not 100% convinced by these `isinf` and `iszero` checks.  This corresponds to canonicalizing the result, no?  But why do we only canonicalize a single case?

sostock: The `CompoundPeriod` constructor canonicalizes the result anyway, these checks are just improving performance.

stevengj: It's not clear that the performance improvement (for a pretty unusual case) is worth the increase in code complexity here.

sostock: Okay, I have removed the special handling of the `isinf`/`iszero` cases.

DilumAluthge:
It looks like two committers approved this a long time (> 5 years) ago. Then it seems to have stalled.

@quinnj @stevengj Is this good to merge?

@sostock Can you rebase on the latest master, and fix the merge conflicts?

PR #30445, opened 2018-12-19
Add `signed(T)`, `unsigned(T)`.

NHDaly:
Converts Types into their corresponding signed/unsigned equivalents, maintaining storage size, but changing signed-ness.

Fixes https://github.com/JuliaLang/julia/issues/30444.

PR #36526, opened 2020-07-03
add `inttype`, `floattype` like `uinttype`

JeffreySarnoff:
`uinttype` is an unexported function that maps `IEEEFloat` types to their bitwidth-matched `Unsigned` types.
This PR adds `inttype` and `floattype` (also unexported).  `inttype` maps `IEEEFloats` to bitwidth-matched `Signed` types and `floattype` maps `uinttype(T), inttype(T) where {T<:IEEEFloat}` to the bitwidth-matched `IEEEFloat` type.

I use these functions often -- less elegantly defined (overloading `reinterpret` or `rem`). As they are unexported, I did not add NEWS.  Since there are no tests for `uinttype`, I did not add tests .. although I am happy to do so for all three functions should that be preferred.

### DIFF
--- a/base/float.jl
+++ b/base/float.jl
@@ -903,6 +903,17 @@ end
 uinttype(::Type{Float64}) = UInt64
 uinttype(::Type{Float32}) = UInt32
 uinttype(::Type{Float16}) = UInt16
+inttype(::Type{Float64}) = Int64
+inttype(::Type{Float32}) = Int32
+inttype(::Type{Float16}) = Int16
+# float size of integer
+floattype(::Type{UInt64}) = Float64
+floattype(::Type{UInt32}) = Float32
+floattype(::Type{UInt16}) = Float16
+floattype(::Type{Int64}) = Float64
+floattype(::Type{Int32}) = Float32
+floattype(::Type{Int16}) = Float16
+
 
 Base.iszero(x::Float16) = reinterpret(UInt16, x) & ~sign_mask(Float16) == 0x0000
 


kimikage:
[FixedPointNumbers](https://github.com/JuliaMath/FixedPointNumbers.jl) defines and exports [`floattype`](https://github.com/JuliaMath/FixedPointNumbers.jl/blob/2f2a34a42b067b45e05724632a8dca019a5a0eba/src/FixedPointNumbers.jl#L110-L151), which is used by packages related to graphic and image processing.
It's weird that the `floattype` is defined in FixedPointNumbers, and it's not bad to be defined in `Base`. However, the `floattype` in FixedPointNumbers is not consistent with `Base.float` due to performance concerns  (cf. https://github.com/JuliaMath/FixedPointNumbers.jl/issues/127).

Of course, this is just the problem in a specific ecosystem.

kimikage:
cf. #30445

oscardssmith:
bump. I think this would be good to have (at least internally).

kimikage:
I agree that it is not a problem because they are not exported, but I do not think it is good practice to use functions that are not exported.

BTW, we can find `inttype` in a surprising place. :smile:
```julia
julia> Base.Threads.inttype(Float32)
Int32
```

The root of the problem is that whether you associate the names `inttype` or `floatype` with reinterpretation or conversion depends on the context and personal mindset.

musm:
I've also needed these in the past. Since they're internal I find these definitions relatively harmless and are more for convenience.

Sans objections I plan on merging this in a day.

PR #36532, opened 2020-07-04
Fix #36531 - Error in abstract_iteration

### LINKED ISSUE #36531: Internal error in abstract_iteration
Happens on master and at least back to 1.4:
```
julia> f(args...) = Core._apply_iterate(Base.iterate, identity, args...)

julia> f(1,2,3)
TypeError(func=:Tuple, context="non-final parameter", expected=Type, got=Vararg{Int64, N} where N)
jl_type_error_rt at /home/keno/julia/src/rtutils.c:118
jl_f_apply_type at /home/keno/julia/src/builtins.c:959
jl_apply at /home/keno/julia/src/julia.h:1711 [inlined]
do_apply at /home/keno/julia/src/builtins.c:655
argtypes_to_type at ./compiler/typeutils.jl:46 [inlined]
abstract_call_known at ./compiler/abstractinterpretation.jl:908
abstract_call_known at ./compiler/abstractinterpretation.jl:698
abstract_iteration at ./compiler/abstractinterpretation.jl:589
precise_container_type at ./compiler/abstractinterpretation.jl:536
```

Keno:
The issue here is passing a `Vargarg` to `precise_container_type`, which
doesn't really make sense. Instead, we need to have the caller unwrap
the vararg, request the precise container type of the inner type and
then re-wrap the answer in a vararg.

### DIFF
--- a/test/compiler/inference.jl
+++ b/test/compiler/inference.jl
@@ -2660,3 +2660,8 @@ function symcmp36230(vec)
     return false
 end
 @test Base.return_types(symcmp36230, (Vector{Any},)) == Any[Bool]
+
+# Issue #36531, double varargs in abstract_iteration
+f36531(args...) = tuple((args...)...)
+@test @inferred(f36531(1,2,3)) == (1,2,3)
+@test code_typed(f36531, Tuple{Vararg{Int}}) isa Vector
--- a/base/compiler/abstractinterpretation.jl
+++ b/base/compiler/abstractinterpretation.jl
@@ -550,6 +550,7 @@ function abstract_iteration(interp::AbstractInterpreter, @nospecialize(itft), @n
     else
         return Any[Vararg{Any}]
     end
+    @assert !isvarargtype(itertype)
     stateordonet = abstract_call_known(interp, iteratef, nothing, Any[itft, itertype], vtypes, sv)
     # Return Bottom if this is not an iterator.
     # WARNING: Changes to the iteration protocol must be reflected here,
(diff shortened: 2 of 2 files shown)


test/compiler/inference.jl
```diff
@@ -2660,3 +2660,8 @@ function symcmp36230(vec)
     return false
 end
 @test Base.return_types(symcmp36230, (Vector{Any},)) == Any[Bool]
+
+# Issue #36531, double varargs in abstract_iteration
+splat36531(args...) = args
+f36531(args...) = Core._apply_iterate(Base.iterate, splat36531, args...)
```
martinholters, line 2666:
```suggestion
f36531(args...) = Core._apply_iterate(Base.iterate, tuple, args...)
```
or
```suggestion
f36531(args...) = tuple((args...)...)
```

martinholters (approved):
Good catch! Fix LGTM.

test/compiler/inference.jl
```diff
@@ -2660,3 +2660,8 @@ function symcmp36230(vec)
...
+@test @inferred(f36531(1,2,3)) == (1,2,3)
```
martinholters, line 2667: This tests seems to pass after printer the error message, which might go unnoticed if it regresses.
Maybe something like
```suggestion
@test code_typed(f36531, Tuple{Vararg{Int}}) isa Vector
```

Keno: I'll do both, run it with `@inferred` and the code typed to test correctness as well as absence of error.

PR #36533, opened 2020-07-04
Bump OpenBLAS BB build to provide aarch64 support

IanButterworth:
Improves aarch64 blas support 

OpenBLAS discussion https://github.com/xianyi/OpenBLAS/issues/2696
BB PR https://github.com/JuliaPackaging/Yggdrasil/pull/1280


giordano:
Note that's not enough: https://github.com/JuliaLang/julia/pull/36067#issuecomment-635765156

Keno (changes requested):
What @giordano said

ianshmean:
Updated with what @giordano said. Although, it didn't seem like `deps/blas.mk` needed updating for `DYNAMIC_ARCH=1`?

Keno:
Do we need to change anything in the makefile for a non-BB build?

ianshmean:
I don't believe so, as my understanding is that the `DYNAMIC_ARCH=1` is only advised when building for redistribution, and handling of it appears to be agnostic to arch in the makefile

ianshmean:
Perhaps, would you mind checking my assertion @giordano ?

giordano:
Yes, my understanding of https://github.com/JuliaLang/julia/blob/09eee44bf600db14c8f7dce7b88ad01b751041a1/deps/blas.mk#L35-L38 is that this is relevant only for the distributed Julia binary.  Users building locally would probably not be interested in enabling  this flag.

ianshmean:
@keno Merge?

PR #36538, opened 2020-07-04
FAQ about moving packages across Julia versions

heliosdrm:


kimikage:
I don't really like to document such unreliable methods. At least, the `X.Y` and `W.Z` should be `1.X` and `1.Y`.

doc/src/manual/faq.md
```diff
@@ -980,3 +980,11 @@ If this describes you, you may also be interested in reading our [guidelines for
 
 Links to each of these download types can be found on the download page at [https://julialang.org/downloads/](https://julialang.org/downloads/).
 Note that not all versions of Julia are available for all platforms.
+
+### How can I transfer the list of installed packages after updating my version of Julia?
+
+When you add a package, this is not really done for the version of Julia that is being used, but for a certain [environment](https://docs.julialang.org/en/v1/manual/code-loading/#Environments-1). However, each minor version of Julia has its own default environment, defined by the files `Project.toml` and `Manifest.toml` in the folder `.julia/environments/vX.Y`, where `X.Y` is the number of the version. The packages defined in those files are the ones that can normally be used when you start a session in that version of Julia.
```
Keno, line 986: The first sentence is a bit of a pronoun mess. How about starting with the fact about the environments and then elaborating on the implications:

```
By default each minor version of julia has its own default [environment](...). As a result, upon installing a new minor version of Julia, the packages you added using the previous minor version will not be available by default. The default environment for a given julia version is defined by ...
```

doc/src/manual/faq.md
```diff
@@ -980,3 +980,11 @@ If this describes you, you may also be interested in reading our [guidelines for
...
+
+If you install a new version of Julia, say `W.Z`, and want to use in its default environment the same packages as in `X.Y`, you can copy the contents of the file `Project.toml` from `X.Y` to `W.Z`. Then, in a session of the new Julia version, enter the "package management mode" by typing the key `]`, and run the command [`instantiate`].
```
Keno, line 988: Just state it as instructions, and use less variables (e.g. `1.x` and `1.x+1`), or even more concrete by choosing some concrete version (e.g. 1.3 and 1.4).

Keno:
I think this is fine to document, but if this is a common enough request maybe we should just have a Pkg REPL command for it (cc @StefanKarpinski @fredrikekre @KristofferC).

doc/src/manual/faq.md
heliosdrm: Done.

doc/src/manual/faq.md
heliosdrm: Done, thanks for the suggestion

StefanKarpinski (approved):

KristofferC:
Thanks, nicely written!

PR #36542, opened 2020-07-05
Introduce ArrayContainer in performance tips

blegat:
The docs starts talking about a type `ArrayContainer` that is never defined. I assume it is supposed to be the type of `x`.


blegat:
Bump

helgee (approved):
LGTM

PR #36543, opened 2020-07-05
Fix off-by-one dayofquarter() in leap years

lbilli:
`dayofquarter()` is off by one for leap years when `month > 3`.
Before:
```julia
julia> dayofquarter(Date(2019,4,1))
1
julia> dayofquarter(Date(2020,4,1))
2
```
After:
```julia
julia> dayofquarter(Date(2019,4,1))
1
julia> dayofquarter(Date(2020,4,1))
1
```


PallHaraldsson:
Since I pointed this out, I reviewd and seems ok. And since a trivial fix, backport to 1.5 and 1.0?

quinnj (approved):

PR #36552, opened 2020-07-06
test REPL package completion if package is already imported

simeonschaub:
Followup to @vtjnash's comment in https://github.com/JuliaLang/julia/pull/36506#discussion_r449968312. At least passes for me locally.


simeonschaub:
@vtjnash Are you still interested in this?

vtjnash:
Ah, this seems good. But why does it pass here if it fails when the module name is Main?

simeonschaub:
@twhitehead's comment in https://github.com/JuliaLang/julia/issues/32377#issuecomment-591632192 seems to attribute it to there actually being two REPL modules in Main, but I have no clue how that's even possible.

twhitehead:
Not two REPL modules in Main. Just two completions for REPL.

The one being the expected one from the test. The other being due to runtests.jl having imported REPL.

Issue #36557, opened 2020-07-07
Generalize `rand` to allow `AbstractUnitRanges`


goretkin:
If `OffsetArrays.jl` is imported, then `rand(Bool, 0:2)` could be made to work.

There is an ambiguity with the expression
`rand(Base.OneTo(3))`, which could either be equivalent to `rand(1:3)` or `rand(3)`. Backwards compatibility dictates it must be the former.

StefanKarpinski:
Yes, when the first argument is a non-integer collection, then it should be sampled from.

goretkin:
From [discourse](https://discourse.julialang.org/t/future-directions-of-julia/46269/77):

> You can’t write rand(-1:1, 0:4) because rand(-1:1) means something, but not sure why rand(Float64, -1:1, 0:4) can’t work.

PR #36558, opened 2020-07-07
Generalize `rand` for `OffsetArrays` e.g. `rand(Bool, 0:2)`

goretkin:
Address https://github.com/JuliaLang/julia/issues/36557
I think it might make sense to consider allowing

1. just `rand(Bool, (0:2,))`
2. also `rand(Bool, 0:2)`
3. furthermore `rand(Bool, 0:2, 0:2)`

1 is the most conservative change, and I can peel off a commit to revert to that situation.

### DIFF
--- a/stdlib/Random/test/runtests.jl
+++ b/stdlib/Random/test/runtests.jl
@@ -815,4 +815,15 @@ isdefined(Main, :OffsetArrays) || @eval Main include(joinpath($(BASE_TEST_PATH),
     @test axes(rand(Bool, Base.OneTo(3))) === (Base.OneTo(3), )
     @test axes(rand(Bool, 3)) === (Base.OneTo(3), )
     @test axes(rand(Bool, 0:2)) == (0:2, )
+
+    # axes splatted
+    @test axes(rand(Bool, Base.OneTo(3), Base.OneTo(3))) === (Base.OneTo(3), Base.OneTo(3))
+    @test axes(rand(Bool, Base.OneTo(3), 3)) === (Base.OneTo(3), Base.OneTo(3))
+    @test axes(rand(Bool, 0:2, 0:2)) == (0:2, 0:2)
+
+    # edge cases
+    @test rand(3) isa Array{Float64, 1}
+    @test rand(1:3) isa Integer
+    @test rand(1:3, 1:3) isa OffsetArrays.OffsetArray
+    @test axes(rand(1:3, 10)) === (Base.OneTo(10),)
 end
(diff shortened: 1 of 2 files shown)


stdlib/Random/test/runtests.jl
```diff
@@ -801,3 +801,29 @@ end
...
+
+    # axes splatted
+    @test axes(rand(Bool, Base.OneTo(3), Base.OneTo(3))) === (Base.OneTo(3), Base.OneTo(3))
+    @test axes(rand(Bool, Base.OneTo(3), 3)) === (Base.OneTo(3), Base.OneTo(3))
+    @test axes(rand(Bool, 0:2, 0:2)) == (0:2, 0:2)
+
+    # edge cases
+    @test rand(3) isa Array{Float64, 1}
+    @test rand(1:3) isa Integer
+    @test rand(1:3, 1:3) isa OffsetArrays.OffsetArray
```
johnnychen94, line 827: What would happen if `OffsetArrays` isn't imported?

johnnychen94, line 827: It looks good to me but I'm not sure how others look at this. This behavior seems not very compatible with the current usage of `rand`; it is reasonable but it would be a little unintuitive that `rand(3, 1:3)` returns an offset matrix.

goretkin: To check that I understand, what would be more intuitive for `rand(3, 1:3)`?

goretkin: That's the error I would expect, too. I agree it should be tested. I don't know how to test it since tests before this test runs could have loaded `OffsetArrays.jl`. Wrapping the test code in a new module would not hide the definition of `Base.similar`

johnnychen94: I was thinking the same you mentioned in #36557, this should be the correct behavior considering the backward compatibility.

goretkin: The example there was more like `rand(1:3)` than `rand(3, 1:3)`

JeffBezanson:
Triage thinks the right way to factor this is to pass a random sampler to an array constructor. `rand` can't know about every type of array you might want to create.

goretkin:
> `rand` can't know about every type of array you might want to create.

I think that is a good stance. I would say that `rand` doesn't need to know about every type of array, though. The meat-and-potatoes of the change is replacing `Array{X}(undef, dims)` with `similar(Array{X}, dims)`, which would allow future `dims` types to indicate future Array types (so long as the appropriate `similar` method is defined).

> pass a random sampler to an array constructor.
I wasn't sure if this is already something in place, even for `Array`, or if it's an idea for the future. e.g. is there a more orthogonal spelling, splitting up the `sp` and the `rand` part for `sprand`?

PR #36559, opened 2020-07-07
add doc for all(p,A,dims) and any(p,A,dims)

PaulSoderlind:
to reducedim.jl

### DIFF
--- a/base/reducedim.jl
+++ b/base/reducedim.jl
@@ -444,6 +444,31 @@ julia> sum(A, dims=2)
 sum(A::AbstractArray; dims)
 
 """
+    sum(f, A::AbstractArray; dims)
+
+Sum the results of calling function `f` on each element of an array over the given
+dimensions.
+
+# Examples
+```jldoctest
+julia> A = [1 2; 3 4]
+2×2 Matrix{Int64}:
+ 1  2
+ 3  4
+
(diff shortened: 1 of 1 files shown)


PaulSoderlind:
This PR suggests documentation of the `all(p,A,dims)` and `any(p,A,dims)` methods.  One of the tests (buildbot/tester_linuxaarch64) fails, but I must admit I do not really know why.

Keno:
Is it worth adding a section somewhere describing how all these `dims` keywords work, particularly in the presence of multiple dimensions, e.g. in `all(rand(Bool, 3, 3, 3); dims=(1,3))`. Maybe we have that somewhere already. If so, we should reference it.

PaulSoderlind:
@Keno A quick look suggests that there is no such section, but someone might know better.

I did, however, notice that also `sum, prod, maximum, minimum` lack documentation of the `f(p,A,,dims)` methods.

PaulSoderlind:
This PR now suggests documentation/examples of `fun(f,A;dims)`  versions of `any, all, sum, prod, maximum` and `minimum.` This just amounts to some more doc strings in `reducedim.jl`.

StefanKarpinski:
Thanks for the doc improvements!

PR #36563, opened 2020-07-07
Update floatfuncs.jl

likanzhan:
Level not correctly rendered (examples were incorrectly rendered as a separate chapter). Is it the right way to correct this?

### DIFF
--- a/base/floatfuncs.jl
+++ b/base/floatfuncs.jl
@@ -99,7 +99,7 @@ julia> round(357.913; sigdigits=4, base=2)
     value represented by `1.15` is actually *less* than 1.15, yet will be
     rounded to 1.2.
 
-    # Examples
+# Examples
     ```jldoctest; setup = :(using Printf)
     julia> x = 1.15
     1.15


base/floatfuncs.jl
```diff
@@ -99,7 +99,7 @@ julia> round(357.913; sigdigits=4, base=2)
     value represented by `1.15` is actually *less* than 1.15, yet will be
     rounded to 1.2.
 
-    # Examples
+# Examples
```
Keno, line 102: I think this example was supposed to apply to the interior of the note, but I'm not sure we have syntax for that. @fredrikekre do you know?

fredrikekre: Yea, it is correct as is: https://docs.julialang.org/en/v1/base/math/#Base.round-Tuple{Type,Any}

The rendering looks a bit funky but probably a Documenter CSS issue? @mortenpi

mortenpi: Yeah, I think it's an overlooked edge case (didn't think about headers in admonitions, so it gets the same styling as the main body).

mortenpi: A workaround would be to turn it into something like a L3/L4 heading or just bold text.

Issue #36037, opened 2020-05-26
Overwriting inner functions eliminates previous invocations


staticfloat:
It appears that overwriting a function defined within another function can cause strange ordering issues.  MWE:

```julia
function do_test()
    function innerfunc()
        println(stderr, "innerfunc")
    end
    println(stderr, "pre-innerfunc")
    innerfunc()
    println(stderr, "post-innerfunc")

    # Overwrite innerfunc
    innerfunc() = nothing
    println("innerfunc is now overwritten")
end

do_test()
```

Expected output:
```
pre-innerfunc
innerfunc
post-innerfunc
innerfunc is now overwritten
```

Actual output:
```
pre-innerfunc
post-innerfunc
innerfunc is now overwritten
```

It appears that the optimizer records the final value of `innerfunc()` and completely disregards the original value:
```
julia> @code_lowered do_test()
CodeInfo(
1 ─      innerfunc = %new(Main.:(var"#innerfunc#3"))
│        Main.println(Main.stderr, "pre-innerfunc")
│        (innerfunc)()
│        Main.println(Main.stderr, "post-innerfunc")
│   %5 = Main.println("innerfunc is now overwritten")
└──      return %5
)
```

This is reproducible on every release of from 1.0 to `master`, I didn't bother looking farther back than 1.0.

vtjnash:
Well that is very minimal 😜

staticfloat:
Something went wrong, hah.  Updated.

staticfloat:
I will note that if this is run at top-level, it works as expected.  It's only when you're redefining an inner function.  I ran into this when I was debugging a test suite and accidentally had `@test inner_func() = "foo"` instead of `@test inner_func() == "foo"`, and couldn't understand why this was seemingly skipping all sorts of side-effects that the original definition of `inner_func()` should have been causing earlier on in that test set.

KristofferC:
Looks like https://github.com/JuliaLang/julia/issues/15602?

KristofferC:
Looks like https://github.com/JuliaLang/julia/issues/15602?

staticfloat:
I don't really see how these are related?  That issue is about a duplicate method warning, but in this case there is no warning, and in that thread I don't really see people writing methods like mine, it's all about having methods in separate branches that conflict with eachother.

vtjnash:
Duplicate of #15602

vtjnash:
We should perhaps turn back on the method overwrite warning

Issue #15602, opened 2016-03-23
Bogus duplicate function definition warning


wbhart:
In Julia 0.5.0-dev+3171 the following code creates a bogus duplicate function definition warning.

```
function myfun(n::Int)
     if n < 3
        doit() = 3
     else
        doit() = 5
     end 
end
```

JeffBezanson:
This isn't allowed any more; I'll have to come up with a better error message. Now you need to do

```
function myfun(n::Int)
     if n < 3
        doit = ()->3
     else
        doit = ()->5
     end 
end
```

ChrisRackauckas:
Is there a reason why this is no longer allowed?

JeffBezanson:
Yes; this restriction makes it possible to hoist construction of `f`'s method table to the top level. A clearer example is

```
function myfun()
    f() = 0
    if rand(Bool)
        f(x::Int) = 1
    end
    f(...)
end
```

Here, `f` would sometimes have one method and sometimes have two methods. That would require rebuilding `f`'s method table on each call to `myfun`, which is too slow to be useful. The work-around is to use separate functions, as in my comment above, or

```
function myfun()
    f1() = 0
    f2() = f1()
    f2(x::Int) = 1
    f = rand(Bool) ? f2 : f1
    f(...)
end
```

ChrisRackauckas:
Yeah, I found those workarounds. Thanks for explaining it. Now that I know there's a reason it's easier to remember. 

I think that this should say something like `Warning: functions cannot be conditionally defined. Last method definition is chosen for each dispatch.` and then have a link to somewhere in documentation which elaborates more on it (I think there should be a whole page which elaborates on errors and warnings).

However, to make it a little more user friendly, my example

```julia
function g(a)
  if a
    f() = 2
  else
    f() = 3
  end
  return f
end
f = g(true)
f() # Returns 3
```

shouldn't error with `g(false)`, but should just have the same answer as `g(true)` and give the same warning.

ViralBShah:
Perhaps this should be documented in NEWS.md as @josefsachsconning pointed out.

josefsachsconning:
Documentation pretty crucial, since this is a breaking change.

josefsachsconning:
Would it not be better for this to generate an error instead of a warning?

JeffBezanson:
It's iffy; the general case is very complicated. For example

```
function f()
    if a
        f(x::A) = ...
    end
    if b
        f(x::B) = ...
    end
[...]

PR #23002, opened 2017-07-28
RFC: add command option to remove method redefinition warnings

KristofferC:
With #265 fixed, more and more workflow seems to move towards redefining methods while keeping Julia running, even methods that exist in modules, e.g. https://github.com/timholy/Revise.jl. Also in interactive sessions like IJulia, redefining methods  is very common.

Since this seems to be the way people will develop code, it is nice to be able to turn off the method redefinition warnings globally. This PR adds a command line option to turn them off.

PR #36565, opened 2020-07-07
method overwrite warning

daviehh:
Re: https://github.com/JuliaLang/julia/issues/15602 and https://github.com/JuliaLang/julia/issues/36037 (possibly others)

Currently, expressions like
```
function myfun(n::Int)
     if n < 3
        println("branch 1")
        doit() = 3
     else
        println("branch 1")
        doit() = 5
     end 
end
```

and

```
function do_test()
    function innerfunc()
        println(stderr, "innerfunc")
    end
    println(stderr, "pre-innerfunc")
    innerfunc()
    println(stderr, "post-innerfunc")

    # Overwrite innerfunc
    innerfunc() = nothing
    println("innerfunc is now overwritten")
end

do_test()
```

would "silently" overwrite the function, such that e.g. with the 1st example, `a = myfun(1)` would print `branch 1` but  `a()` returns `5` since method is silently overwritten, making such scenarios a bit difficult to debug. 

A long-term solution could be to error in such cases, meanwhile, this PR set the default setting to warn about such overwrites, and add a function to flip the warning setting (and a convenience macro) to temporarily disable the warning when such overwrite is intentional.

Example:
```
julia> f(x) = 1
julia> f(x) = x
WARNING: Method definition f(Any) in module Main at REPL[1]:1 overwritten at REPL[2]:1.
julia> Base.@overwrite f(x) = x * 3
julia> f(1)
 3
```

### DIFF
--- a/base/util.jl
+++ b/base/util.jl
@@ -502,6 +502,50 @@ function _kwdef!(blk, params_args, call_args)
     blk
 end
 
+
+"""
+    @overwrite
+
+This is a helper macro that temporarily disables the method overwrite warning.
+
+!!! compat "Julia 1.6"
+        `Base.@overwrite` requires at least Julia 1.6.
+
+# Examples
+```
+julia> f(x) = 1
+ f (generic function with 1 method)
+
+julia> f(x) = x
+WARNING: Method definition f(Any) in module Main at REPL[1]:1 overwritten at REPL[2]:1.
+ f (generic function with 1 method)
+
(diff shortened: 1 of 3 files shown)


KristofferC:
The whole reason this setting was introduced was to not show them by default (https://github.com/JuliaLang/julia/pull/23002).

mbauman:
I would much rather this be done in the lowering for local functions only.

daviehh:
got it, thanks! closing for now

PR #36570, opened 2020-07-08
Move Const/PartialStruct to Core

Keno:
The change in #36521 proposes to make PartialStruct legal in the
global code cache. However, `@vtjnash` points out that since
these structs are currently defined in `Core.Compiler`, this
would make loading a second copy of inference impossible. However,
loading a second copy of inference is desirable for development,
so we generally put any type that can appear in the global cache
into `Core` to make this possible. As such we have two options
here:
1. Change the representation of PartialStruct to only use objects
already in Core
2. Move Const/PartialStruct into Core

I originally thought that option 1 would be preferable, since I
though that the only meaningful information that could appear
in a PartialStruct field would be a Const or a nested PartialStruct.
However, I forgot that it sometimes also contains Type{...} or elements
of the type lattice more precise than what the widened type prescribes.
As such, building PartialStruct out of solely types defined in Core
would have gotten a bit messy, and it's much simpler just to move
both of these to Core themselves. `@JeffBezanson` also points out
that we may eventually want to look at these in codegen, so define
them in C while we're at it.

### DIFF
--- a/base/boot.jl
+++ b/base/boot.jl
@@ -399,9 +399,9 @@ eval(Core, :(CodeInstance(mi::MethodInstance, @nospecialize(rettype), @nospecial
                           min_world::UInt, max_world::UInt) =
                 ccall(:jl_new_codeinst, Ref{CodeInstance}, (Any, Any, Any, Any, Int32, UInt, UInt),
                     mi, rettype, inferred_const, inferred, const_flags, min_world, max_world)))
-eval(Core, :(Const(@nospecialize(v)) = $(Expr(:new, Const, :v, false))))
-eval(Core, :(Const(@nospecialize(v), actual::Bool) = $(Expr(:new, Const, :v, :actual))))
-eval(Core, :(PartialStruct(@nospecialize(typ), fields::Array{Any, 1}) = $(Expr(:new, PartialStruct, :typ, :fields))))
+eval(Core, :(Const(@nospecialize(v)) = $(Expr(:new, :Const, :v, false))))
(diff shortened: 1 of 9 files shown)


base/boot.jl
```diff
@@ -399,6 +399,9 @@ eval(Core, :(CodeInstance(mi::MethodInstance, @nospecialize(rettype), @nospecial
                           min_world::UInt, max_world::UInt) =
                 ccall(:jl_new_codeinst, Ref{CodeInstance}, (Any, Any, Any, Any, Int32, UInt, UInt),
                     mi, rettype, inferred_const, inferred, const_flags, min_world, max_world)))
+eval(Core, :(Const(@nospecialize(v)) = $(Expr(:new, Const, :v, false))))
+eval(Core, :(Const(@nospecialize(v), actual::Bool) = $(Expr(:new, Const, :v, :actual))))
+eval(Core, :(PartialStruct(@nospecialize(typ), fields::Array{Any, 1}) = $(Expr(:new, PartialStruct, :typ, :fields))))
```
vtjnash, line 404:
```suggestion
eval(Core, :(Const(@nospecialize(v)) = $(Expr(:new, :Const, :v, false))))
eval(Core, :(Const(@nospecialize(v), actual::Bool) = $(Expr(:new, :Const, :v, :actual))))
eval(Core, :(PartialStruct(@nospecialize(typ), fields::Array{Any, 1}) = $(Expr(:new, :PartialStruct, :typ, :fields))))
```
Seems to be what we usually do

JeffBezanson (approved):

PR #36582, opened 2020-07-08
Remove some references to build time environment from sysimg

yuyichao:
All of these are always re-populated at load time and
does not need to be in sysimg.

Similar to #36536.

I'm also carrying more patches to remove all build time path in the binary for a more reprodicuble build as well as being much more friendly to gdb by default similar to the `-f*-prefix-map` GCC options. Those patches are much dirtier currently though....

### DIFF
--- a/base/sysimg.jl
+++ b/base/sysimg.jl
@@ -101,4 +101,9 @@ let
     empty!(LOAD_PATH)
     empty!(DEPOT_PATH)
 end
+
+@eval Sys begin
+    BINDIR = ""
+    STDLIB = ""
+end
 end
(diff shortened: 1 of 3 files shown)


JeffBezanson:
The only issue I see with this is that generate_precompile is optional.

yuyichao:
Doing this in `sysimg.jl` is also possible, but then `Sys.BINDIR` and `Sys.STDLIB`, both contains reference to build dir would need to be repopulated before calling this. The current patch is good enough for my case but I can try making it more general...

yuyichao:
Should be cleared before generate_precompile as well now.

PR #36583, opened 2020-07-08
fix #36527, skipped marking of some items in excstack

### LINKED ISSUE #36527: Segfault on showerror in v1.0 - latest
I have tried to slim this down as much as possible, this is the best I could do:

## `runtests.jl`
```jl
using Test

"something"
function jl_is_runnable(path)
    🏡 = Core.eval(Main, :(module asdf end))
    try
        Core.eval(🏡, :(include($path)))
    catch ex
        println.(enumerate(readlines(path; keep=true)))

        showerror(stderr, ex, stacktrace(catch_backtrace()))
    end
end

@test jl_is_runnable("./worse.jl")
```

## `worse.jl`
```jl
@assert z == 10
```

The result is a [segfault](https://github.com/fonsp/reproduce-36527/runs/835490754?check_suite_focus=true#step:4:25), on Linux, macOS and Windows, `v1.0` until `v1.5`. 

(Deleting any line from `runtests.jl` fixes the problem, hence the silly MWE.)

JeffBezanson:
When moving to the next group of julia values in a trace the counter was not reset, so we didn't mark values after the first group.

fix #36527

### DIFF
--- a/src/gc.c
+++ b/src/gc.c
@@ -2230,6 +2230,7 @@ excstack: {
                         goto mark;
                     }
                 }
+                jlval_index = 0;
             }
             // The exception comes last - mark it
             new_obj = jl_excstack_exception(excstack, itr);
--- a/test/exceptions.jl
+++ b/test/exceptions.jl
@@ -367,3 +367,19 @@ end
         exc
     end == ErrorException("rethrow(exc) not allowed outside a catch block")
 end
+
+# issue #36527
+function f36527()
+    caught = false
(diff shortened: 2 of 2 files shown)


vtjnash (approved):

c42f (approved):
Good catch.  Nasty bug :grimacing: 

This bug was introduced while generalizing trace entries in #33277, so I think that means it will be in 1.4 and 1.5.

staticfloat:
I ran into this in the wild on 1.4.2 mere minutes before Jeff fixed this.  Amazing.

PR #36587, opened 2020-07-08
Fix compilation on LLVM 10

yuyichao:


vtjnash:
These seem like possibly just `MaybeAlign`?

yuyichao:
You are right, this errors for LLVM 10. However, it seems to have changed again for LLVM 11... https://github.com/llvm/llvm-project/commit/accc6b55450510577f9cfa0b434c97e45ab5d6c6 which is what I was reading when writing this code. So I guess on LLVM 11 the alignment will never be missing and we can just pass the result from `getAlign` to it?

In any case, my local compilation has just finished and I've updated this for LLVM 10... Someone with LLVM 11 from the past 2 months can check what's needed there...

vtjnash:
Ah, seems to be. After https://github.com/JuliaLang/julia/pull/36507/commits/2b510b417b83b3c08a069866ecf90e2f3a97fc2d merged (today), most of the Loads/Stores we emit should have alignment. Some were still missing though, and LLVM could have created others.

PR #36591, opened 2020-07-09
Consistent use of singleton types in manual.

### LINKED ISSUE #7135: Manual could be clearer about parametric singleton types
I was confused about the meaning of parametric singleton types such as `Type{Array{T}}`, and the manual section "Singleton Types" does not say anything about them.  It should be improved.

The only documentation section I could find that says anything about parametric singleton types is "Case Study: Rational Conversions", which is misleading because it defines methods for `convert{T<:Int}(::Type{Rational{T}}, x::...)` but not  `convert(::Type{Rational}, x::...)`.  At first glance it seems like `convert(Rational, x)` would invoke one of those `Type{Rational{T}}` methods just as `convert(Rational{Int8}, x)` would.  In fact `convert(Rational, x)` invokes a method which is not in this documentation section.  The convert methods for rational numbers in base/rational.jl are somewhat different from what the documentation says.

Here's a simple example: define `f{T}(::Type{Array{T}}) = 1` then `f(Array{Int})` invokes that method but `f(Array)` and `f(Array{Int,1})` do not.  This makes perfect sense when you think about it, but it isn't really obvious and some specific documentation would help.

BTW unlike the 0.2 version, in the latest version manual on docs.julialang.org the section "Singleton Types" does not show up in the table of contents at all.  I am not sure what the problem is.  It can be found by going to "Parametric Types" in the table of contents and then scrolling down a long way.

tpapp:
Fixes #7135, see discussion there.

### Changes

- separate discussion of singleton types (as in `Base.issingletontype`) and `Type{T}`

- move discussion about singleton types to a later section so that parametric singleton types can be discussed

- add example and discussion about parametric singleton types

- add usage hint about `Type` outside dispatch ~~(TL;DR: don't)~~ example: type sharpening (thanks @tkf)

### Open questions:

- [x] is there a terminology for `Type{T}` now that we don't confuse it with singleton types? If not, do we want to introduce one, eg call it "type selectors"? 

- [x] should I mention that abstract types can't be singleton types, or is it obvious from the context?

- [x] should I expunge the remaining use of 'singleton' for `Type{T}` from *Conversion and Promotion*?

### DIFF
--- a/doc/src/manual/conversion-and-promotion.md
+++ b/doc/src/manual/conversion-and-promotion.md
@@ -173,8 +173,7 @@ The type of the first argument of this method is a [`Type{T}`](@ref man-typet-ty
 when the first argument is the type value `MyType`. Notice the syntax used for the first
 argument: the argument name is omitted prior to the `::` symbol, and only the type is given.
 This is the syntax in Julia for a function argument whose type is specified but whose value
-does not need to be referenced by name. In this example, since the type is a singleton, we
-already know its value without referring to an argument name.
+does not need to be referenced by name.
 
 All instances of some abstract types are by default considered "sufficiently similar"
 that a universal `convert` definition is provided in Julia Base.
--- a/doc/src/manual/conversion-and-promotion.md
+++ b/doc/src/manual/conversion-and-promotion.md
@@ -168,7 +168,7 @@ Such a definition might look like this:
 convert(::Type{MyType}, x) = MyType(x)
 ```
 
-The type of the first argument of this method is a [singleton type](@ref man-singleton-types),
+The type of the first argument of this method is a [`Type{T}`](@ref man-typet-type),
 `Type{MyType}`, the only instance of which is `MyType`. Thus, this method is only invoked
 when the first argument is the type value `MyType`. Notice the syntax used for the first
 argument: the argument name is omitted prior to the `::` symbol, and only the type is given.
--- a/doc/src/manual/conversion-and-promotion.md
+++ b/doc/src/manual/conversion-and-promotion.md
@@ -168,8 +168,8 @@ Such a definition might look like this:
 convert(::Type{MyType}, x) = MyType(x)
 ```
 
-The type of the first argument of this method is a [`Type{T}`](@ref man-typet-type),
-`Type{MyType}`, the only instance of which is `MyType`. Thus, this method is only invoked
+The type of the first argument of this method is [`Type{MyType}`](@ref man-typet-type),
+the only instance of which is `MyType`. Thus, this method is only invoked
 when the first argument is the type value `MyType`. Notice the syntax used for the first
 argument: the argument name is omitted prior to the `::` symbol, and only the type is given.
 This is the syntax in Julia for a function argument whose type is specified but whose value
--- a/doc/src/manual/types.md
+++ b/doc/src/manual/types.md
@@ -458,7 +458,7 @@ To recap, two essential properties define immutability in Julia:
       functions as pointers to heap-allocated values except in cases where the compiler
       is sure that there's no way to tell that this is not what is happening.
 
-## Declared Types
+## [Declared Types](@id man-declared-types)
 
 The three kinds of types (abstract, primitive, composite) discussed in the previous
 sections are actually all closely related. They share the same key properties:
@@ -1061,8 +1061,13 @@ element type.
 ## [Singleton types](@id man-singleton-types)
 
 Immutable composite types with no fields are called *singletons*. Formally, if 
-`a isa T && b isa T` implies `a === b`, then `T` is a singleton type. [^2]
-[`Base.issingletontype`](@ref) can be used to check if a type is a singleton type.
+
+1. `T` is an immutable composite type (ie defined with `struct`),
+1. `a isa T && b isa T` implies `a === b`,
+
+then `T` is a singleton type.[^2] [`Base.issingletontype`](@ref) can be used to check if a
+type is a singleton type. [Abstract types](@ref man-abstract-types) cannot be singleton
+types by construction.
 
 From the definition, it follows that there can be only one instance of such types:
 
@@ -1101,7 +1106,7 @@ julia> NoFieldsParam{Int}() === NoFieldsParam{Int}()
 true
 ```
 
-## [`Type{T}` types](@id man-typet-type)
+## [`Type{T}` type selectors](@id man-typet-type)
 
 For each type `T`, the `Type{T}` is an abstract parametric type whose only instance is the
 object `T`. Until we discuss [Parametric Methods](@ref) and [conversions](@ref
@@ -1171,9 +1176,27 @@ julia> isa("foo", Type)
 false
 ```
 
-Finally, while `Type` is part of Julia's type hierarchy like any other abstract parametric
-type, it is not commonly used outside method signatures. If you find that you are frequently
-dealing with standalone `Type{T}` objects, there may be an opportunity to reorganize your code.
+While `Type` is part of Julia's type hierarchy like any other abstract parametric type, it
+is not commonly used outside method signatures except in some special cases. Another
+important use case for `Type` is sharpening field types which would otherwise be captured
+less precisely, eg as [`DataType`](@ref man-declared-types) in the example below where the
+default constuctor could lead to performance problems in code relying on the precise wrapped
+type (similarly to [abstract type parameters](@ref man-performance-abstract-container)).
+
+```jldoctest
+julia> struct WrapType{T}
+       value::T
+       end
+
+julia> WrapType(Float64) # default constructor, note DataType
+WrapType{DataType}(Float64)
+
+julia> WrapType(::Type{T}) where T = WrapType{Type{T}}(T)
+WrapType
+
+julia> WrapType(Float64) # sharpened constructor, note more precise Type{Float64}
+WrapType{Type{Float64}}(Float64)
+```
 
 ## Type Aliases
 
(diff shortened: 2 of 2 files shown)


tkf:
> add usage hint about `Type` outside dispatch (TL;DR: don't)

A usage (workaround) outside dispatch is to get a sharp field type.  For example, the type of `Fix1(Vector, undef)` is `Fix1{Type{Vector},UndefInitializer}` rather than `Fix1{DataType,UndefInitializer}`.  This is required for the field type to be inferable.  So, I think it'd be better to avoid stressing that `Type{T}` is _only_ for dispatch.  "`Type{T}` types" section in this PR seems to be fine as-is, though.

doc/src/manual/types.md
```diff
@@ -1127,6 +1058,123 @@ must always be specified in full, this is not especially helpful, but in Julia,
 to write just `Vector` for the abstract type including all one-dimensional dense arrays of any
 element type.
 
+## [Singleton types](@id man-singleton-types)
+
+Immutable composite types with no fields are called *singletons*. Formally, if 
+`a isa T && b isa T` implies `a === b`, then `T` is a singleton type. [^2]
```
tkf, line 1064: I think the spirit of this formalism is great, but if I have

```julia
julia> abstract type A end

julia> struct B <: A end
```

and I don't define any other subtype of `A`, would it imply that `A` is also a singleton type?

Maybe making this perfect requires `isconcretetype(T)`?  Although if you say "for any `a` and `b`" I guess it could mean "for any `a` and `b` of any implementable type."

> should I mention that abstract types can't be singleton types, or is it obvious from the context?

So yes, I think this helps to make this definition clearer.

tkf:
> is there a terminology for `Type{T}` now that we don't confuse it with singleton types? If not, do we want to introduce one, eg call it "type selectors"?

Yeah, calling `Type{T}` a singleton sounds a bit confusing since you can write something like `Type{<:Integer}`.  For the new name, "type selectors" sound reasonable to me.

> should I expunge the remaining use of 'singleton' for `Type{T}` from _Conversion and Promotion_?

:+1: from me :)

doc/src/manual/types.md
tpapp: Yes, I think we should stress that only concrete types can to be singletons.

tpapp:
@tkf: thanks for the review. Yes, sharp field types are another use case --- if you have a small, self-contained example, I would be happy to add it to the PR if that's OK with you.

I included the usage hint because I have seen users on Discourse trying to work with `Type{Type{T}}` etc constructs and other misunderstandings. `Type` is mysterious to a newcomer as is, and each common use case should serve as an example. So I would rather just include them if feasible.

tkf:
Yeah, including example is a good idea.  I think a minimal example is `Some`:

```julia
struct SomeBad{T}
    value::T
end

struct SomeGood{T}
    value::T
end
SomeGood(::Type{T}) where {T} = SomeGood{Type{T}}(T)

sbad = SomeBad(Symbol)
sgood = SomeGood(Symbol)

@assert sbad isa SomeBad{DataType}
@assert sgood isa SomeGood{Type{Symbol}}

getvalue(x) = x.value

@code_warntype getvalue(sbad)  # => Body::DataType
@code_warntype getvalue(sgood)  # => Body::Type{Symbol}
```

tpapp:
Thanks @tkf, I simplified the example somewhat as I did not want to deal with `@code_warntype` in a `jldoctest`. I also did the other changes above.

tpapp:
Friendly bump — if this needs to be improved, I am happy to implement any further suggested changes.

StefanKarpinski:
I've requested review by @JeffBezanson. I'll merge one way or another this week.

doc/src/manual/conversion-and-promotion.md
```diff
@@ -168,13 +168,12 @@ Such a definition might look like this:
 convert(::Type{MyType}, x) = MyType(x)
 ```
 
-The type of the first argument of this method is a [singleton type](@ref man-singleton-types),
+The type of the first argument of this method is a [`Type{T}`](@ref man-typet-type),
 `Type{MyType}`, the only instance of which is `MyType`. Thus, this method is only invoked
```
rfourquet, line 172: Why not put directly `Type{MyType}` within the `[]` brackets, where `Type{T}` is currently?

doc/src/manual/types.md
```diff
@@ -1127,6 +1058,146 @@ must always be specified in full, this is not especially helpful, but in Julia,
...
+julia> Base.issingletontype(NoFieldsParam{Int}) # parametrized, it is a singleton
+true
+
+julia> NoFieldsParam{Int}() === NoFieldsParam{Int}()
+true
+```
+
+## [`Type{T}` type selectors](@id man-typet-type)
+
+For each type `T`, the `Type{T}` is an abstract parametric type whose only instance is the
```
rfourquet, line 1111: "the `Type{T}` is ..." sounds weird, it reads "the type `T`", where in fact we mean "the `Type{T}` type". I would just remove "the":
```suggestion
For each type `T`, `Type{T}` is an abstract parametric type whose only instance is the
```

doc/src/manual/types.md
```diff
@@ -1127,6 +1058,146 @@ must always be specified in full, this is not especially helpful, but in Julia,
...
+false
+
+julia> isa(Real, Type{Real})
+true
+
+julia> isa(Float64, Type{Real})
+false
+```
+
+In other words, [`isa(A,Type{B})`](@ref) is true if and only if `A` and `B` are the same object
```
rfourquet, line 1135:
```suggestion
In other words, [`isa(A, Type{B})`](@ref) is true if and only if `A` and `B` are the same object
```

doc/src/manual/types.md
```diff
@@ -1127,6 +1058,146 @@ must always be specified in full, this is not especially helpful, but in Julia,
...
+and that object is a type.
+
+In particular, since parametric types are [invariant](@ref man-parametric-composite-types),
```
rfourquet, line 1138: Sounds like the sentence is missing a "connection" word, e.g. I would expect "we have:" or something like that at the end, before the example.

doc/src/manual/types.md
```diff
@@ -1127,6 +1058,146 @@ must always be specified in full, this is not especially helpful, but in Julia,
...
+false
+
+julia> isa("foo", Type)
+false
+```
+
+While `Type` is part of Julia's type hierarchy like any other abstract parametric type, it
+is not commonly used outside method signatures except in some special cases. Another
+important use case for `Type` is sharpening field types which would otherwise be captured
+less precisely, eg as [`DataType`](@ref man-declared-types) in the example below where the
```
rfourquet, line 1182:
```suggestion
less precisely, e.g. as [`DataType`](@ref man-declared-types) in the example below where the
```

tpapp:
@rfourquet, thanks for the thorough reading and the detailed suggestions, I applied all of them.

StefanKarpinski (approved):

StefanKarpinski:
Great clarification. Thank you! This looks ready to merge once CI passes.

doc/src/manual/types.md
```diff
@@ -1127,6 +1058,146 @@ must always be specified in full, this is not especially helpful, but in Julia,
 to write just `Vector` for the abstract type including all one-dimensional dense arrays of any
 element type.
 
+## [Singleton types](@id man-singleton-types)
+
+Immutable composite types with no fields are called *singletons*. Formally, if
```
dkarrasch, line 1063:
```suggestion
Immutable composite types with no fields are called *singletons*. Formally, if
```

doc/src/manual/types.md
```diff
@@ -1127,6 +1058,146 @@ must always be specified in full, this is not especially helpful, but in Julia,
...
+
+1. `T` is an immutable composite type (ie defined with `struct`),
+1. `a isa T && b isa T` implies `a === b`,
+
+then `T` is a singleton type.[^2] [`Base.issingletontype`](@ref) can be used to check if a
```
rfourquet, line 1068: Is there an official convention on whether the footnote sign should go before or after the period? I saw both in the manual.

doc/src/manual/types.md
```diff
@@ -1127,6 +1058,146 @@ must always be specified in full, this is not especially helpful, but in Julia,
 to write just `Vector` for the abstract type including all one-dimensional dense arrays of any
 element type.
 
+## [Singleton types](@id man-singleton-types)
+
+Immutable composite types with no fields are called *singletons*. Formally, if
+
+1. `T` is an immutable composite type (ie defined with `struct`),
```
rfourquet, line 1065:
```suggestion
1. `T` is an immutable composite type (i.e. defined with `struct`),
```

rfourquet (approved):
Yes nice change! Just one more possible change ("ie" -> "i.e.") if you also update this branch for something else.

doc/src/manual/types.md
tpapp: The [Chicago Manual of Style](https://www.chicagomanualofstyle.org/qanda/data/faq/topics/Punctuation/faq0020.html) suggests that it goes after the period.

tpapp: Incidentally, since these questions are in general important for high-quality documentation, I started some discussion about referring a style guide in #36723.

tpapp:
Another friendly bump: I think I addressed all comments.

I am happy to wait for more feedback if necessary, just don't want this PR to be forgotten.

tkf:
It has two approvals and the `merge me` label. I think it's more than enough :)

PR #36485, opened 2020-06-30
Update ARM feature and CPU detection (supersedes #36464)

IanButterworth:
I messed up #36464 with an accidental blank force push, so that PR can't be reopened. This is the replacement

Updates ARM feature and CPU detection

cc. @yuyichao

PR #36592, opened 2020-07-09
WIP: Some minimal support for Apple Silicon

Keno:
With this and some hacks to various dependencies, Julia builds for
Darwin ARM64. We don't get very far though, and quickly end up
jumping into a bit of uninitialized JIT memory, so there's clearly
more work to be done here. That said, let's take this one step at
a time, so here's the few pieces that at least make it build.

### DIFF
--- a/src/signals-mach.c
+++ b/src/signals-mach.c
@@ -112,13 +112,30 @@ static kern_return_t profiler_segv_handler
                  mach_msg_type_number_t                   code_count);
 #endif
 
+#if defined(_CPU_X86_64_)
+typedef x86_thread_state64_t host_thread_state_t;
+typedef x86_exception_state64_t host_exception_state_t;
+#define THREAD_STATE x86_THREAD_STATE64
+#define THREAD_STATE_COUNT x86_THREAD_STATE64_COUNT
+#define HOST_EXCEPTION_STATE x86_EXCEPTION_STATE64
+#define HOST_EXCEPTION_STATE_COUNT x86_EXCEPTION_STATE64_COUNT
+
 enum x86_trap_flags {
     USER_MODE = 0x4,
     WRITE_FAULT = 0x2,
     PAGE_PRESENT = 0x1
 };
 
-static void jl_call_in_state(jl_ptls_t ptls2, x86_thread_state64_t *state,
+#elif defined(_CPU_AARCH64_)
+typedef arm_thread_state64_t host_thread_state_t;
+typedef arm_exception_state64_t host_exception_state_t;
+#define THREAD_STATE ARM_THREAD_STATE64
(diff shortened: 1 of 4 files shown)


src/signals-mach.c
```diff
@@ -112,13 +112,30 @@ static kern_return_t profiler_segv_handler
...
     USER_MODE = 0x4,
     WRITE_FAULT = 0x2,
     PAGE_PRESENT = 0x1
 };
 
-static void jl_call_in_state(jl_ptls_t ptls2, x86_thread_state64_t *state,
+#elif defined(_CPU_AARCH64_)
+typedef arm_thread_state64_t host_thread_state_t;
+typedef arm_exception_state64_t host_exception_state_t;
+#define HOST_EXCEPTION_STATE ARM_EXCEPTION_STATE
```
Keno, line 134: Note to self: This should be ARM_EXCEPTION_STATE64

yuyichao: Is that a change from Apple that you are expecting? Or are you going to push a new version?

yuyichao (approved):
The change makes sense, though I’m just assuming the signal handling is correct....

A few questions,

1. Is it A64 only?
2. What’s the actual chip and features? What’s apple’s way of detecting the features?
    I’ve heard that it can boot Linux so checking it from there should be interesting (to see if https://github.com/JuliaLang/julia/pull/36485 detects it correctly).i assume it doesn’t have hwcap or at least not in the same way Linux does it. Does it at least support what Linux have with HWCAP_CPUID? Or is it like iOS where, as I read, you jus need to hard code the known hardware features of various Apple devices.

src/signals-mach.c
Keno: I'll push a new version, there's still some other issues with it also.

Keno:
> * Is it A64 only?

I believe so yes. Darwin has support for A32, but all recent iDevices are A64 and I believe there will never be an A32 mac.

> What’s the actual chip and features? What’s apple’s way of detecting the features?

The chip in the DTK is an Apple A12Z. I haven't figured out if there's any way to do feature detection. I need to look into that.

staticfloat (approved):

Keno:
Looks like we can feature detect via sysctl:
```
$ sysctl hw.optional
hw.optional.floatingpoint: 1
hw.optional.watchpoint: 4
hw.optional.breakpoint: 6
hw.optional.neon: 1
hw.optional.neon_hpfp: 1
hw.optional.neon_fp16: 1
hw.optional.armv8_1_atomics: 1
hw.optional.armv8_crc32: 1
hw.optional.armv8_2_fhm: 0
hw.optional.amx_version: 0
hw.optional.ucnormal_mem: 0
hw.optional.arm64: 1
```

PR #36596, opened 2020-07-09
some minor code quality / correctness improvements

vtjnash:
Digging around at a few things that seemed to show up re-compiled often

### DIFF
--- a/base/Base.jl
+++ b/base/Base.jl
@@ -84,6 +84,7 @@ include("range.jl")
 include("error.jl")
 
 # core numeric operations & types
+==(x, y) = x === y
 include("bool.jl")
 include("number.jl")
 include("int.jl")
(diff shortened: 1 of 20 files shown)


JeffBezanson:
Something in 5e86746b1774f5d32978ac427c2c93d2844fca14 is going haywire --- 4x longer to precompile Plots, +5% sysimg.

vtjnash:
Ah, I see what happened. The new lazy-ambiguity logic assumes that subtyping is transitive over specificity, and I accidentally formed a contradiction of that premise on the definition of `==` and returned too many matches. In particular, I observe the set of facts that:
```
!(a <: b)
a <: c
c morespecific b # ie. c < b

where a == c == Tuple{Type,Type}
and b == Tuple{T,T} where T
```
And I wasn't checking for this case, so in the results, it appeared that `b` was also a possible method match, which complicated the caching logic (made it slow).

PR #36597, opened 2020-07-09
Add option '--compile=min' to Julia man page.

derekmahar:
Add option '--compile=min' to Julia man page.


derekmahar:
Please ignore this PR.

Issue #26894, opened 2018-04-24
shell mode on windows sometimes corrupt the terminal state


KristofferC:
repo `;ls` in windows REPL.

![image](https://user-images.githubusercontent.com/1282691/39179602-0030c2f8-47b5-11e8-87a0-987d231157b9.png)

Seems to work on 0.6.2. Can anyone repro?

helgee:
Successfully reproduced, is this the `ls` from Git?

helgee:
Seems to happen with all the utils shipped with Git for Windows:

```
               _
   _       _ _(_)_     |  A fresh approach to technical computing
  (_)     | (_) (_)    |  Documentation: https://docs.julialang.org
   _ _   _| |_  __ _   |  Type "?help" for help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 0.7.0-DEV.4939 (2018-04-24 02:56 UTC)
 _/ |\__'_|_|_|\__'_|  |  Commit ef9168a6f8* (0 days old master)
|__/                   |  i686-w64-mingw32

shell> ..\\Programs\\Git\\usr\\bin\\echo.exe blob
blob

[7C[32m[1mjulia> [0m[0m
```

KristofferC:
Perhaps we should add a 

`write(terminal, "\033c")` (reset the terminal state)

in the beginning of

https://github.com/JuliaLang/julia/blob/ef9168a6f8a7eb016519649e0dd2cf49f8a81648/stdlib/REPL/src/LineEdit.jl#L1180-L1185

PetrKryslUCSD:
@KristofferC : This issue has been closed, but has the problem been solved? I can still see it happening with  Version 0.7.0-beta2.43...

KristofferC:
How is this closed?

PetrKryslUCSD:
Sorry, I meant #27782.

Issue #36584, opened 2020-07-08
Windows REPL cursor jumps to the start when hitting backspace


musm:
There's a  buggy behavior when hitting backspace after any key which forces the cursor to jump the start of the line

See:

![1](https://user-images.githubusercontent.com/4319522/86955382-76d76600-c125-11ea-97de-2b800c404c9b.gif)

vtjnash:
We're often redrawing the whole line, to keep it clean. But you're just seeing how slow performing Windows is.

davidanthoff:
I think this should stay open. Other programs, like Python, manage to have a solution on Windows that doesn't have such flickers, so it certainly seems technically feasible to have a better implementation on Windows.

I'm also seeing a similar flicker in the VS Code terminal.

Just out of curiosity: I think when I looked at this last, the way Julia behaved was that it had the VT sequences that Julia emits handled by libuv, right? And then libuv would be calling the old console APIs in Windows? I think for many reasons it would make a lot more sense to cut out the libuv part here and have the windows console itself handle the VT sequences directly if Julia is run on a reasonably recent Windows 10 version. Support for that has landed years ago in Windows 10. If I recall the situation correctly, then libuv actually supports that out-of-the-box for a while, but the libuv fork that Julia uses had this disabled? If I'm recalling all of this correctly, then we are going through _very many_ emulation layers here, which could easily explain this flicker. And maybe things are better with the current, up-to-date APIs that Windows provides for console stuff?

vtjnash:
None of those things are relevant here. Python doesn't flicker because it's REPL is too primitive to be useful for anything much. I don't know why VS Code doesn't implement a better renderer, though I suspect the OS might just causes enough performance issue, that it can't do better on Windows. Other platforms use the same code and don't have this issue. The "new" APIs work by adding extra emulation layers that translates the codes into commands that the conhost renderer understands, similar to libuv but with more layers of emulation.

musm:
Just some more notes for reference: 

For example, the Ipython  REPL (i.e. not the python)  is actually very  feature rich, and doesn't suffer the same problems  encountered in this issue.

In the original case, when hitting backspace or the left/right arrow, and using a patched Julia and libuv fork, some of the issues of the navigation keys causing buggy REPL behavior **are resolved.**  There's still a problem when you hit the UP arrow key, but the situation is improved with respect to left/right and backspace navigation. 


[...]

PR #36598, opened 2020-07-09
Enable tty, vt sequences on Windows  

### LINKED ISSUE #31491: Fix color support on Windows via VT sequences
This is a follow up on https://github.com/JuliaLang/julia/issues/7267#issuecomment-476935384. The issue is that right now that @KristofferC's https://github.com/KristofferC/Crayons.jl package doesn't print colors properly on Windows, because the julia process on Windows doesn't configure proper VT support on newer versions of Windows 10.

I'm sketching what I found so far, but don't have the time to fix this. I believe this needs to be fixed somewhere either in libuv or julia, not sure. But maybe what I found so far is helpful for doing that.

I started out with starting julia on Windows 10 via the start menu link. I then ran:
```julia
julia> hOutput = ccall(:GetStdHandle, Int32, (Int32,), -11)
88

julia> dwMode = Ref{UInt32}(0)
Base.RefValue{UInt32}(0x00000000)

julia> ccall(:GetConsoleMode, Int32, (Int32, Ptr{Nothing}), hOutput, dwMode)
1

julia> dwMode[]
0x00000003
```
The first thing to note is that by default the `ENABLE_VIRTUAL_TERMINAL_PROCESSING` flag is not set for this process, which means that none of the new VT support in Windows 10 is active.

Alright, lets set that flag (the value of it is 0x004):
```julia
julia> newMode = dwMode[] | 0x0004
0x00000007

julia> ccall(:SetConsoleMode, Int32, (Int32, Int32), hOutput, newMode)
1
```
If I now call
```julia
julia> using Crayons

julia> Crayons.test_256_colors(false)
```
things still don't look good. BUT, if I do the following, I get the correct output:
```julia
julia> buf = IOBuffer()

julia> Crayons.test_256_colors(buf, false)

julia> tr_buf = transcode(UInt16, String(take!(buf)))

julia> ccall(:WriteConsoleW, Int32, (Int32, Ptr{UInt8}, Int32, Ptr{Nothing}, Ptr{Nothing}), hOutput, tr_buf , length(tr_buf ), C_NULL, C_NULL)
```
I'm not sure what is going on, but it looks like somehow that `print` and friends don't send that string in such a way to the console that the VT sequences make it there. Maybe libuv is intercepting these sequences in some form? I have no idea, but clearly whatever `print` is doing is preventing things from working as they should in this case.

I believe `ENABLE_VIRTUAL_TERMINAL_INPUT` is also not set by default. I don't think that matters for the color example, but it probably means that some other VT feature that is in principle supported on Windows is not active when julia runs.

musm:
Enable tty, vt sequences on Windows.

Ultimately, we can't control what external processes assume on the console mode when launched within Julia. With the shared console mode that Julia utilizes, this is problematic and the terminal can be left in a corrupted state. As a result, it's safest to *reset the mode* before each prompt to ensure we enter back into the original console state.
In particular, this fixes problems like that in https://github.com/JuliaLang/julia/issues/26894#issuecomment-383901827.

Currently Julia has disabled vt support, as this can  help enhance compatibility. Note, however, we can  corrupt the terminal state even under the current status quo, which uniformly disables  vt support, and this can be demonstrated through examples. Disabling tty support, has been an annoyance on Windows and has lead to a lot of comments on how the Windows Julia terminal experience is crippled and that this is unfixable since Microsoft has not improved the terminal experience, leading to discussion on other terminal emulators such as mintty. In fact, I think  we can support a quality terminal experience on Windows if we carefully enable features that support an improved REPL experience.

This PR in tandem with https://github.com/JuliaLang/libuv/pull/11  enables tty support on Windows. 

Here's an example of Julia with these patches using the Windows Terminal:
image

For those that want to try out this PR built with the updated libuv fork that enables tty support on Windows, here's a x86-64 windows installer you can play around with: https://1drv.ms/u/s!AvuMkGVjog3NuaQswzrzMyFP8XlVNw?e=qbv749

 

closes https://github.com/JuliaLang/julia/issues/31491

### DIFF
--- a/src/support/libsupportinit.c
+++ b/src/support/libsupportinit.c
@@ -12,9 +12,7 @@ static int isInitialized = 0;
 void libsupport_init(void)
 {
     if (!isInitialized) {
-#ifdef _OS_WINDOWS_
-        SetConsoleCP(1252); // ANSI Latin1; Western European (Windows)
-#endif
+
         setlocale(LC_ALL, ""); // set to user locale
         setlocale(LC_NUMERIC, "C"); // use locale-independent numeric formats
 
--- a/stdlib/REPL/src/LineEdit.jl
+++ b/stdlib/REPL/src/LineEdit.jl
@@ -1277,7 +1277,7 @@ end
 # not leave the console mode in a corrupt state.
 if Sys.iswindows()
 function _console_mode()
-    hOutput = ccall(:GetStdHandle, Int32, (Int32,), -11) # STD_OUTPUT_HANDLE
+    hOutput = ccall(:GetStdHandle, stdcall, Int32, (Int32,), -11) # STD_OUTPUT_HANDLE
     dwMode = Ref{UInt32}(0)
     ccall(:GetConsoleMode, Int32, (Int32, Ptr{Nothing}), hOutput, dwMode)
     dwMode[]
@@ -1286,8 +1286,8 @@ const default_console_mode = _console_mode()
 function _reset_console_mode()
     mode = _console_mode()
     if mode !== default_console_mode
-        hOutput = ccall(:GetStdHandle, Int32, (Int32,), -11) # STD_OUTPUT_HANDLE
-        ccall(:SetConsoleMode, Int32, (Int32, Int32), hOutput, default_console_mode)
+        hOutput = ccall(:GetStdHandle, stdcall, Int32, (Int32,), -11) # STD_OUTPUT_HANDLE
+        ccall(:SetConsoleMode, stdcall, Int32, (Int32, Int32), hOutput, default_console_mode)
         nothing
     end
 end
--- a/stdlib/REPL/src/LineEdit.jl
+++ b/stdlib/REPL/src/LineEdit.jl
@@ -1272,8 +1272,29 @@ function write_prompt(terminal, p::Prompt)
     return width
 end
 
+# On Windows, when launching external processes, we cannot control what assumption they make on the
+# console mode. We thus forcibly reset the console mode at the start of the prompt to ensure they do
+# not leave the console mode in a corrupt state.
+if Sys.iswindows()
+function _console_mode()
+    hOutput = ccall(:GetStdHandle, Int32, (Int32,), -11)
+    dwMode = Ref{UInt32}(0)
+    ccall(:GetConsoleMode, Int32, (Int32, Ptr{Nothing}), hOutput, dwMode)
+    dwMode[]
+end
+const default_console_mode = _console_mode()
+function _reset_console_mode()
+    ENABLE_VIRTUAL_TERMINAL_PROCESSING = 0x004
+    newMode = default_console_mode | ENABLE_VIRTUAL_TERMINAL_PROCESSING
+    hOutput = ccall(:GetStdHandle, Int32, (Int32,), -11)
+    ccall(:SetConsoleMode, Int32, (Int32, Int32), hOutput, newMode)
+    nothing
+end
+end
+
 # returns the width of the written prompt
 function write_prompt(terminal, s::Union{AbstractString,Function})
+    @static Sys.iswindows() && _reset_console_mode()
     promptstr = prompt_string(s)
     write(terminal, promptstr)
     return textwidth(promptstr)
--- a/stdlib/REPL/src/LineEdit.jl
+++ b/stdlib/REPL/src/LineEdit.jl
@@ -1275,21 +1275,30 @@ end
 # On Windows, when launching external processes, we cannot control what assumption they make on the
 # console mode. We thus forcibly reset the console mode at the start of the prompt to ensure they do
 # not leave the console mode in a corrupt state.
+# FIXME: when pseudo-tty are used for child processes
 if Sys.iswindows()
 function _console_mode()
-    hOutput = ccall(:GetStdHandle, stdcall, Int32, (Int32,), -11) # STD_OUTPUT_HANDLE
-    dwMode = Ref{UInt32}(0)
-    ccall(:GetConsoleMode, stdcall, Int32, (Int32, Ptr{Nothing}), hOutput, dwMode)
-    dwMode[]
+    hOutput = ccall(:GetStdHandle, stdcall, Ptr{Cvoid}, (UInt32,), unsafe_trunc(UInt32,-11)) # STD_OUTPUT_HANDLE
+    dwMode = Ref{UInt32}()
+    ccall(:GetConsoleMode, stdcall, Int32, (Ref{Cvoid}, Ref{UInt32}), hOutput, dwMode)
+    return dwMode[]
 end
-const default_console_mode = _console_mode()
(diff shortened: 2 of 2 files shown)


davidanthoff:
I just verified that this PR also solves the flickering cursor problem described in https://github.com/JuliaLang/julia/issues/36584. I verified this for the old school Windows console, the new Windows Terminal and the VS Code built-in terminal: all three had the flickering without this PR, with this PR it is gone in all three.

musm:
Yes this PR fixes that issue for me, as well!

As it stands, this PR is a stop-gap solution, until libuv has proper pseudo-tty support. Given that we don't have a time frame on when and if such a feature will come, I think something similar to what is proposed in this PR is needed in the meantime.  Granted it is a stop-gap solution on already problematic scenario, being that we are forced to use a shared console between child processes, which can have different implicit assumptions on the console state and lead to corruption.

davidanthoff:
We should probably try this on the oldest officially supported Windows version, just to be sure. [This](https://julialang.org/downloads/#currently_supported_platforms) suggests that Windows 10 is the min version for Julia. There are of course very many different Windows 10 versions, and many of the older releases are no longer supported by MS. Looking at [this table](https://support.microsoft.com/en-us/help/13853/windows-lifecycle-fact-sheet) and assuming that versions that are still supported for enterprise and edu customers are also what Julia officially supports, is 1709 the oldest version that we need to test? Or should we go with 1803 right away, because any build that includes this PR here would come out after 10/13 in any case?

musm:
I updated this PR and I actually think it should now be compatible with any Windows version.  All that the new code does it reset the terminal mode back to the default mode, in case a shell application modifies it.

musm:
There's a mysterious bug where simply calling `hOutput = ccall(:GetStdHandle, Int32, (Int32,), -11)` segfaults Julia in 32-bit. Not sure what's causing this.

musm:
For reference here's the example that demonstrates the current way `run` works is actually  broken. This example is courtesy  @davidanthoff 

```julia
echo_cmd = `C:\\Program\ Files\\Git\\usr\\bin\\echo.exe hello`

 @async begin
    sleep(10)
    run(echo_cmd)
 end

run(`wsl`)

```
Without `echo_cmd` things, like `htop` work perfectly in the wsl started by `run`, but after `10` seconds, the `echo_cmd` command executes and this completely corrupts the console. 


This PR actually improves this pathological test case, because on master executing the above corrupts the console permanently, but with this PR resetting to the default states returns the REPL back to  useful state.

musm:
So on my end I'm happy with the current state of this RP. 

@vtjnash  Would love to get your review on this if you have the time 😃.

musm:
going to merge in a few days if no objections

stdlib/REPL/src/LineEdit.jl
```diff
@@ -1272,8 +1272,30 @@ function write_prompt(terminal, p::Prompt)
...
+# console mode. We thus forcibly reset the console mode at the start of the prompt to ensure they do
+# not leave the console mode in a corrupt state.
+if Sys.iswindows()
+function _console_mode()
+    hOutput = ccall(:GetStdHandle, stdcall, Int32, (Int32,), -11) # STD_OUTPUT_HANDLE
+    dwMode = Ref{UInt32}(0)
+    ccall(:GetConsoleMode, stdcall, Int32, (Int32, Ptr{Nothing}), hOutput, dwMode)
+    dwMode[]
+end
+const default_console_mode = _console_mode()
```
vtjnash, line 1285: should not be a build-time constant

stdlib/REPL/src/LineEdit.jl
```diff
@@ -1272,8 +1272,30 @@ function write_prompt(terminal, p::Prompt)
...
+function _reset_console_mode()
+    mode = _console_mode()
+    if mode !== default_console_mode
+        hOutput = ccall(:GetStdHandle, stdcall, Int32, (Int32,), -11) # STD_OUTPUT_HANDLE
```
vtjnash, line 1289: it's a OS_HANDLE, not an Int32

src/support/libsupportinit.c
```diff
@@ -12,9 +12,7 @@ static int isInitialized = 0;
 void libsupport_init(void)
 {
     if (!isInitialized) {
-#ifdef _OS_WINDOWS_
-        SetConsoleCP(1252); // ANSI Latin1; Western European (Windows)
-#endif
+
```
vtjnash, line 15: this feels slightly risky and unrelated? (reverts eac525c89f463b2721f348a5870a81f7b431d6be, but libuv is now unicode-aware, so perhaps that's not as bad now)

musm: We need to remove this or set the console to `SetConsoleCP(CP_UTF8)` or else the sequences don't get interpreted correctly

stdlib/REPL/src/LineEdit.jl
musm: Good point! I changed to be a reference that get's set  the first time the REPL is printed at runtime and then is cached perpetually.

stdlib/REPL/src/LineEdit.jl
musm: Fixed. Also updated to be more pedantic about DWORD being a a UInt32 and avoiding the implicit cast using an `unsafe_trunc`

src/support/libsupportinit.c
vtjnash: CP_UTF8 (65001) is not well supported. That's why we set it to 1252 (It even can break `dir`: https://github.com/nodejs/node-v0.x-archive/issues/4246#issuecomment-10142715)

musm: Is that comment  relevant? It looks like part of the issue was with using a raster front

musm: Looking at  https://github.com/JuliaLang/julia/issues/6795

```
print("测试")
测试

julia> print("\346\265\213\350\257\225")
测试
```

With this PR

musm: Looks like things probably got fixed internally in libuv?

musm: So I think this is safe.

musm:
Ok to merge soon?

src/support/libsupportinit.c
vtjnash: Yes, libuv Is fixed now to bypass this. The issue would be with other shared libraries that assume an ascii-compatible stream.

StefanKarpinski:
Unless there are objections or feedback, then yes, please do merge.

src/support/libsupportinit.c
musm: Good to hear. Is that issue more theoretical than practically relevant (?), I don't know what we could do about this pathological case without regressing the whole system to be dumber because of some library is forcing ascii.

musm:
Thanks @vtjnash  for the review! and also @davidanthoff for working to test this PR and his collaboration on trying to make this a reality.

PR #36606, opened 2020-07-10
Improve exception stack printing

### LINKED ISSUE #31257: Display of exception stacks could be better
To me the new exception stack printing (#30900) is not so clear, compare with e.g. this reduced example:
```
$ cat foo.jl 
function foo()
    error("hello")
end
foo()
```
which displays on master as
```
$ julia-master --color=yes foo.jl 
ERROR: Error while loading expression starting at /home/fredrik/foo.jl:4
caused by [exception 1]
hello
Stacktrace:
 [...]
 [8] _start() at ./client.jl:476
```
and on Julia 1 as
```
$ julia11 --color=yes foo.jl 
ERROR: LoadError: hello
Stacktrace:
 [...]
 [8] _start() at ./client.jl:436
in expression starting at /home/fredrik/foo.jl:4
```

I have two problems with the new printing:

1. From the output on master it took me quite a while to understand that the `hello` output actually came from the `ERROR:` two lines above. IMO it was more clear before what the error was, and I suggest this should be printed as
    ```
    $ julia-master --color=yes foo.jl 
    ERROR: hello
    while loading expression starting at /home/fredrik/foo.jl:4
    caused by [exception 1]
    Stacktrace:
     [...]
     [8] _start() at ./client.jl:476
    ```
   instead. It seems more important to print the message before the location.
  
2. When seeing `caused by [exception 1]` I would expect a list of exceptions somewhere where I can look up what `exception 1` is. What is `exception 1` in the case above? It is not clear to me.

c42f:
The expanded stacktrace printing introduced in #36134 and the fact that
we no longer print errors in red (#36015) makes it harder to distinguish
distinct exceptions in the stack.

Add a newline for this, and print the "caused by" in dark red. Also remove
exception numbering, which was arguably not very helpful.

Arguably this (along with previous changes over time) fixes #31257.

Before:
before

After:
after

### DIFF
--- a/base/errorshow.jl
+++ b/base/errorshow.jl
@@ -872,7 +872,7 @@ function show_exception_stack(io::IO, stack::Vector)
     nexc = length(stack)
     for i = nexc:-1:1
         if nexc != i
-            printstyled(io, "caused by [exception ", i, "]\n", color=:light_black)
+            printstyled(io, "\ncaused by:\n", color=error_color())
         end
         exc, bt = stack[i]
         showerror(io, exc, bt, backtrace = bt!==nothing)
--- a/test/client.jl
+++ b/test/client.jl
@@ -11,7 +11,8 @@ end
 nested_error_pattern = r"""
     ERROR: DivideError: integer division error
     Stacktrace:.*
-    caused by \[exception 1\]
+
+    caused by:
     UndefVarError: __not_a_binding__ not defined
     Stacktrace:.*
     """s


base/errorshow.jl
```diff
@@ -872,7 +872,7 @@ function show_exception_stack(io::IO, stack::Vector)
     nexc = length(stack)
     for i = nexc:-1:1
         if nexc != i
-            printstyled(io, "caused by [exception ", i, "]\n", color=:light_black)
+            printstyled(io, "\ncaused by:\n", color=:red)
```
KristofferC, line 875:
```suggestion
            printstyled(io, "\ncaused by:\n", color=error_color())
```

maybe?

JeffBezanson (approved):

vtjnash (approved):

base/errorshow.jl
c42f: Sure, fixed.

PR #29249, opened 2018-09-18
REPL: allow tweaking the implicit IOContext of the REPL

rfourquet:
This adds an `iocontext::Dict{Symbol,Any}` field to `LineEditREPL`,
which can be initialized with `atreplinit`, and updated
interactively, e.g. `Base.active_repl.iocontext[:compact] = true`.

Fixes #20509.

An alternative would be providing specific functions, e.g. `setiocontext(:key, value)`, `resetiocontext(:key)` or something. I think I prefer the dictionnary approach, because a `Dict` already has a convenient API for the needs here, and as an interactive feature, this can be changed relatively easily if needed.

I wonder whether we should think about a namespace policy to use this, as package `X` could say for example "use `iocontext[:format] = :bin` for printing with binary digits", while package `Y` could say have `iocontext[:format] = whatever` to control formatting of other stuff, and it becomes impossible to have control over both packages at the same time.
Another option would therefore to have a predetermined list of settable properties (`:compact`, `:limit` etc), in which case using a `mutable struct` with corresponding fields would probably be a more appropriate implementation than a `Dict`.

PR #36610, opened 2020-07-10
fix #36492, ensure output is limited even with basic REPLs

### LINKED ISSUE #36492: Truncation of displayed array entries in the REPL with TERM=dumb
Is the REPL broken in the latest stable version of Julia? In version 1.3.1 I obtained the (expected) truncation of arrays displayed in the REPL. For example, the command
randn(100,100)
displays a truncated set of values.

In Julia versions 1.4.1 and 1.4.2 the same command will attempt to display all 10000 values of this array. Of course, the situation is much worse for huge arrays.

I understand that I can suppress the output with a trailing semicolon, but I preferred the old behavior that displayed a small selection of array values.

displaysize(stdout) returns the value (24, 112), which looks correct to me.

mbauman:
This is simple enough; just unconditionally use an IOContext with `:limit=>true` and then allow the fancier REPLs to further customize.

### DIFF
--- a/stdlib/REPL/src/REPL.jl
+++ b/stdlib/REPL/src/REPL.jl
@@ -205,11 +205,11 @@ end
 
 function display(d::REPLDisplay, mime::MIME"text/plain", x)
     with_methodtable_hint(d.repl) do io
+        io = IOContext(io, :limit => true, :module => Main)
         get(io, :color, false) && write(io, answer_color(d.repl))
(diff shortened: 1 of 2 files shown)


mbauman:
I think this was an unintentional effect from #29249.

JeffBezanson (approved):

rfourquet (approved):

rfourquet:
> I think this was an unintentional effect from #29249.

Yes I totally overlooked that other terminal types existed.

PR #36614, opened 2020-07-10
Sharper field type for `NamedTuple`s with types in fields

tkf:
This PR tweaks the `NamedTuple{names}` constructor so that `namedtuple.field` is inferable when `field` is a type.

Before

```julia
julia> typeof((a=Int,))
NamedTuple{(:a,),Tuple{DataType}}
```

After

```julia
julia> typeof((a=Int,))
NamedTuple{(:a,),Tuple{Type{Int64}}}
```

### DIFF
--- a/base/boot.jl
+++ b/base/boot.jl
@@ -562,7 +562,24 @@ end
 
 NamedTuple() = NamedTuple{(),Tuple{}}(())
 
-NamedTuple{names}(args::Tuple) where {names} = NamedTuple{names,typeof(args)}(args)
+function safe_Typeof(@nospecialize(x))
+    if x isa DataType
+        if x.layout === Ptr{Nothing}(0)  # we don't have C_NULL yet
+            return DataType
+        end
+    end
+    return Typeof(x)
+end
+
+_nt_typeof(@nospecialize(x)) = safe_Typeof(x)
+_nt_typeof(@nospecialize(xs::Tuple)) = Tuple{_nt_typesof_impl(xs...)...}
+_nt_typesof_impl() = ()
+_nt_typesof_impl(@nospecialize(x), @nospecialize xs...) =
+    (safe_Typeof(x), _nt_typesof_impl(xs...)...)
+# It's non-recursive (using `safe_Typeof` instead of `_nt_typeof`) so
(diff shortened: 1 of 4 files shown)


base/boot.jl
```diff
@@ -562,7 +562,24 @@ end
 
 NamedTuple() = NamedTuple{(),Tuple{}}(())
 
-NamedTuple{names}(args::Tuple) where {names} = NamedTuple{names,typeof(args)}(args)
+function safe_Typeof(@nospecialize(x))
+    if x isa DataType
+        if x.layout === Ptr{Nothing}(0)  # we don't have C_NULL yet
```
quinnj, line 567: this looks.....scary? do we not have an official query function for this?

tkf: I guess not? `.layout == C_NULL` seems to be the common way in `Base`:

```console
$ rg '\.layout'
reflection.jl
336:    dt.layout == C_NULL && throw(UndefRefError())
337:    alignment = unsafe_load(convert(Ptr{DataTypeLayout}, dt.layout)).alignment
370:    dt.layout == C_NULL && throw(UndefRefError())
371:    flags = unsafe_load(convert(Ptr{DataTypeLayout}, dt.layout)).flags
383:    dt.layout == C_NULL && throw(UndefRefError())
384:    npointers = unsafe_load(convert(Ptr{DataTypeLayout}, dt.layout)).npointers
392:located at `(dt.layout + sizeof(DataTypeLayout))`.
399:    dt.layout == C_NULL && throw(UndefRefError())
400:    flags = unsafe_load(convert(Ptr{DataTypeLayout}, dt.layout)).flags
```

JeffBezanson: I'm afraid of the performance implications of this PR, but why not just use `Typeof` here?

tkf: By performance, do you mean compile-time?

I'm checking `.layout` as otherwise I get:

```julia
julia> NamedTuple{(:a,),Tuple{Type{Vector.body}}}((Vector.body,))
ERROR: UndefVarError: T not defined
```

DilumAluthge:

PR #36615, opened 2020-07-11
Update our vendored copy of config.sub

Keno:
In particular to pull in [1], which fixes configure for
several of our dependencies on Apple Silicon.

[1] http://git.savannah.gnu.org/cgit/config.git/commit/?id=3c20dd431e06337a8b49d8bf791a6a19523a34a4

Also cc @giordano @staticfloat, I believe we're providing this file in BB also,
so ideally we should pull in the update.

### DIFF
--- a/deps/patches/config.sub
+++ b/deps/patches/config.sub
@@ -1,8 +1,8 @@
 #! /bin/sh
 # Configuration validation subroutine script.
-#   Copyright 1992-2018 Free Software Foundation, Inc.
+#   Copyright 1992-2020 Free Software Foundation, Inc.
 
-timestamp='2018-12-31'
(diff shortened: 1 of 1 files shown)


giordano:
Pull request to update the scripts in the RootFS of BinaryBuilder: https://github.com/JuliaPackaging/Yggdrasil/pull/1305

Keno:
> Pull request to update the scripts in the RootFS of BinaryBuilder: [JuliaPackaging/Yggdrasil#1305](https://github.com/JuliaPackaging/Yggdrasil/pull/1305)

Perfect, thanks. Not that we have a BB shard yet anyway, but might as well get a head start.

PR #36622, opened 2020-07-11
Fix typeassert dropping PartialStruct information

Keno:
There was a complaint on discourse that Generator splats weren't
infering precisely enough [1]. I thought the recent improvement
to abstract iteration precision would have fixed that, but that
turns out not to be the case for a stupid reason. At the moment,
any type assert (even something silly like `x::Any`), will drop
`PartialStruct` information. Now, generators have an
`::Tuple{Any, Any}` typeassert, which prevented the new iteration
logic from working properly. This fixes typasserts, to not drop
this information. The first couple of conditions that just
return the original `PartialStruct` if the typeassert has no
effect are doing most of the work here. As an add-on, I also
threw in the logic to narrow the PartialStruct if the typeassert
actually introduces additional information. It's not a particularly
common code path, but might as well.

[1] https://discourse.julialang.org/t/tuples-iterators-splats-and-type-stability/42849

### DIFF
--- a/base/compiler/tfuncs.jl
+++ b/base/compiler/tfuncs.jl
@@ -485,18 +485,38 @@ add_tfunc(typeof, 1, 1, typeof_tfunc, 0)
 function typeassert_tfunc(@nospecialize(v), @nospecialize(t))
     t = instanceof_tfunc(t)[1]
     t === Any && return v
-    if isa(v, Const)
-        if !has_free_typevars(t) && !isa(v.val, t)
-            return Bottom
+    function typeassert_type_instance(@nospecialize(v), @nospecialize(t))
+        if isa(v, Const)
+            if !has_free_typevars(t) && !isa(v.val, t)
+                return Bottom
+            end
+            return v
+        elseif isa(v, PartialStruct)
+            has_free_typevars(t) && return v
+            widev = widenconst(v)
+            if widev <: t
+                return v
+            elseif typeintersect(widev, t) === Bottom
+                return Bottom
+            end
+            @assert widev <: Tuple
+            new_fields = Vector{Any}(undef, length(v.fields))
+            for i = 1:length(new_fields)
+                new_fields[i] = typeassert_type_instance(v.fields[i], getfield_tfunc(t, Const(i)))
+                if new_fields[i] === Bottom
+                    return Bottom
+                end
+            end
+            return tuple_tfunc(new_fields)
+        elseif isa(v, Conditional)
+            if !(Bool <: t)
+                return Bottom
+            end
+            return v
         end
-        return v
-    elseif isa(v, Conditional)
-        if !(Bool <: t)
-            return Bottom
-        end
-        return v
+        return typeintersect(widenconst(v), t)
     end
-    return typeintersect(widenconst(v), t)
+    return typeassert_type_instance(v, t)
 end
 add_tfunc(typeassert, 2, 2, typeassert_tfunc, 4)
 
--- a/test/compiler/inference.jl
+++ b/test/compiler/inference.jl
@@ -2716,3 +2716,21 @@ function f_apply_union_split(fs, x)
 end
 
 @test Base.return_types(f_apply_union_split, Tuple{Tuple{typeof(sqrt), typeof(abs)}, Int64}) == Any[Union{Int64, Float64}]
+
+# Precision of typeassert with PartialStruct
+function f_typ_assert(x::Int)
+    y = (x, 1)
+    y = y::Any
+    Val{y[2]}
+end
+@test Base.return_types(f_typ_assert, (Int,)) == Any[Type{Val{1}}]
+
+function f_typ_assert2(x::Any)
+    y = (x::Union{Int, Float64}, 1)
+    y = y::Tuple{Int, Any}
+    (y[1], Val{y[2]}())
+end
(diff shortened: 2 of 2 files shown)


base/compiler/tfuncs.jl
```diff
@@ -484,14 +484,27 @@ function typeof_tfunc(@nospecialize(t))
 end
 add_tfunc(typeof, 1, 1, typeof_tfunc, 0)
 
-function typeassert_tfunc(@nospecialize(v), @nospecialize(t))
-    t = instanceof_tfunc(t)[1]
-    t === Any && return v
+function typeassert_instance(@nospecialize(v), @nospecialize(t))
```
martinholters, line 487: I'm not too happy about this name, though I don't really have a better idea. Maybe at least drop a comment that `t` here is the one meant to be the "instance" (of a `Type{...}`), not `v`?

martinholters (approved):
Generally LGTM.

Without knowing that it's defined in the middle of `_tfunc`s, the name `typeassert_instance` might be quite misleading, but I don't want to stall this PR in naming bikeshedding.

JeffBezanson (approved):

base/compiler/tfuncs.jl
```diff
@@ -484,14 +484,27 @@ function typeof_tfunc(@nospecialize(t))
...
     if isa(v, Const)
         if !has_free_typevars(t) && !isa(v.val, t)
             return Bottom
         end
         return v
+    elseif isa(v, PartialStruct)
+        has_free_typevars(t) && return v
+        widev = widenconst(v)
+        newT = typeintersect(widev, t)
+        if newT === widev
```
vtjnash, line 497:
```suggestion
        if newT == widev
```

vtjnash: I know these seem almost the same (since `.typ` is usually concrete), but the type-system knows that too, and can handle the exceptions better

base/compiler/tfuncs.jl
```diff
@@ -484,14 +484,27 @@ function typeof_tfunc(@nospecialize(t))
...
+            return v
+        end
```
vtjnash, line 499:
```suggestion
        if widev <: t
            return v
        elseif typeintersect(widev, t) === Bottom
            return Bottom
        end
```

vtjnash: Calling `typeintersect` here would force us to call `abstract_apply` on `tuple_tfunc` over the result, to recompute `widev` from `newT`. That's awkward (sometimes more accurate, sometimes less), so let's avoid that.

base/compiler/tfuncs.jl
```diff
@@ -484,14 +484,27 @@ function typeof_tfunc(@nospecialize(t))
...
+        new_fields = Vector{Any}(undef, length(v.fields))
+        for i = 1:length(new_fields)
+            new_fields[i] = typeassert_instance(v.fields[i], getfield_tfunc(newT, Const(i)))
```
vtjnash, line 502:
```suggestion
            new_fields[i] = typeassert_instance(v.fields[i], getfield_tfunc(t, Const(i)))
```

This would be strictly more precise.

base/compiler/tfuncs.jl
Keno: Maybe just make it an inner function inside `typeassert_tfunc`?

base/compiler/tfuncs.jl
```diff
@@ -484,14 +484,27 @@ function typeof_tfunc(@nospecialize(t))
...
+            if new_fields[i] === Bottom
+                return Bottom
+            end
+        end
+        return PartialStruct(newT, new_fields)
```
vtjnash, line 507:
```suggestion
        return tuple_tfunc(new_fields)
```

This would be strictly more correct.

base/compiler/tfuncs.jl
```diff
@@ -500,6 +513,12 @@ function typeassert_tfunc(@nospecialize(v), @nospecialize(t))
     end
     return typeintersect(widenconst(v), t)
 end
+
+function typeassert_tfunc(@nospecialize(v), @nospecialize(t))
+    t = instanceof_tfunc(t)[1]
+    t === Any && return v
+    typeassert_instance(v, t)
```
vtjnash, line 520:
```suggestion
    return typeassert_instance(v, t)
```

I know a computer doesn't care about the different, but I treat it as a trivial comment for future human readers

vtjnash (changes requested):

base/compiler/tfuncs.jl
Keno: I figured the case where typeintersect would return something that's `==` to one of the inputs, but not `===` to one of the inputs would be pretty rare, and in the case where it doesn't, using `===` would bypass the useless sometype query that we don't really care about.

base/compiler/tfuncs.jl
Keno: `t` is not necessarily a tuple

base/compiler/tfuncs.jl
Keno: I don't understand what you're saying here.

Keno:
@vtjnash points out that currently if the type inside a PartialStruct is not a tuple, it must be concrete, so the fields parts of this didn't apply and just re-doing `tuple_tfunc` would be strictly more precise. Updated to do that.

vtjnash (approved):

PR #36623, opened 2020-07-11
Commute union splitting and MethodTable identification

Keno:
This allows inference of calls where union-splitting the function would
lead to an acceptable answer, e.g. in
```
function randapply(fs, x)
  i = rand(1:length(fs))
  f = fs[i]
  f(x)
end
@code_warntype randapply((sqrt, abs), 3)
```

Previously this only worked if an extra function boundary was introduced,
which would cause union splitting on some non-first argument,
e.g.
```
function randapply2(fs, x)
  xs = map(f -> f(x), fs)
  rand(xs)
end
@code_warntype randapply((sqrt, abs), 3)
```
which currently infers fine.

This patch simply moves the identification of which MethodTable we
looked at into the union splitting logic. This seems to fix the
case of interest, and I don't see any correctness problems.

As requested by @zenna on slack.

### DIFF
--- a/base/compiler/abstractinterpretation.jl
+++ b/base/compiler/abstractinterpretation.jl
@@ -39,20 +39,22 @@ function abstract_call_gf_by_type(interp::AbstractInterpreter, @nospecialize(f),
     if sv.currpc in sv.throw_blocks
         return Any
     end
-    mt = ccall(:jl_method_table_for, Any, (Any,), atype)
-    if mt === nothing
-        add_remark!(interp, sv, "Could not identify method table for call")
-        return Any
-    end
-    mt = mt::Core.MethodTable
     min_valid = UInt[typemin(UInt)]
     max_valid = UInt[typemax(UInt)]
     atype_params = unwrap_unionall(atype).parameters
     splitunions = 1 < countunionsplit(atype_params) <= InferenceParams(interp).MAX_UNION_SPLITTING
+    mts = Core.MethodTable[]
+    fullmatch = Bool[]
     if splitunions
         splitsigs = switchtupleunion(atype)
         applicable = Any[]
         for sig_n in splitsigs
+            mt = ccall(:jl_method_table_for, Any, (Any,), sig_n)
+            if mt === nothing
+                add_remark!(interp, sv, "Could not identify method table for call")
+                return Any
+            end
+            mt = mt::Core.MethodTable
             xapplicable = matching_methods(sig_n, sv.matching_methods_cache, max_methods,
                                            get_world_counter(interp), min_valid, max_valid)
             if xapplicable === false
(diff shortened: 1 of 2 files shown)


JeffreySarnoff:
I see how it is of benefit and there does not appear to be a credible downside.

base/compiler/abstractinterpretation.jl
```diff
@@ -161,18 +178,22 @@ function abstract_call_gf_by_type(interp::AbstractInterpreter, @nospecialize(f),
...
-            match = applicable[i]::SimpleVector
-            method = match[3]::Method
-            if atype <: method.sig
-                fullmatch = true
-                break
+        if length(mts) == 1
+            for i in napplicable:-1:1
+                match = applicable[i]::SimpleVector
+                method = match[3]::Method
+                if atype <: method.sig
```
vtjnash, line 185:
```suggestion
                if match[4]
```
(this is new, but might as well start using it)

base/compiler/abstractinterpretation.jl
```diff
@@ -34,34 +34,50 @@ function matching_methods(@nospecialize(atype), cache::IdDict{Any, Tuple{Any, UI
...
+            end
+            mt = mt::Core.MethodTable
             xapplicable = matching_methods(sig_n, sv.matching_methods_cache, max_methods,
                                            get_world_counter(interp), min_valid, max_valid)
             if xapplicable === false
                 add_remark!(interp, sv, "For one of the union split cases, too many methods matched")
                 return Any
             end
             append!(applicable, xapplicable)
+            push_unique!(mts, mt)
```
vtjnash, line 72:
```suggestion
            contains_is(mts, mt) || push!(mts, mt)
```
(this is the pattern we use elsewhere anyways)

vtjnash (approved):

base/compiler/abstractinterpretation.jl
Keno: `match[4]` or `!match[4]`? Which way around is this boolean?

Keno: I tried this change and it causes singificant sysimage build time regressions. Given that it is an orthogonal change to this PR, I'll drop this here and it can be in a separate PR where can debug why it's problematic for performance.

vtjnash: Oh, it’s probably getting confused by Union-splitting. Carry on.

base/compiler/abstractinterpretation.jl
```diff
@@ -161,18 +170,22 @@ function abstract_call_gf_by_type(interp::AbstractInterpreter, @nospecialize(f),
             add_backedge!(edge::MethodInstance, sv)
         end
         fullmatch = false
-        for i in napplicable:-1:1
-            match = applicable[i]::SimpleVector
-            method = match[3]::Method
-            if atype <: method.sig
-                fullmatch = true
-                break
+        if length(mts) == 1
```
JeffBezanson, line 173: When more than one MethodTable matches, it seems to me `fullmatch` should be set per MethodTable, based on the signatures that matched in that table.

Keno:
Alright, I think I was able to address both @JeffBezanson's and @vtjnash's latest comments.

vtjnash (approved):

PR #36624, opened 2020-07-11
Darwin/ARM64: Set default mcpu target

Keno:
And unconditionally require the CRC32 extension on this platform.
All current and future chips should have it.

### DIFF
--- a/Make.inc
+++ b/Make.inc
@@ -777,6 +777,11 @@ ifeq ($(ARCH),amd64)
 override ARCH := x86_64
 endif
 
+# We map arm64 (Apple spelling) to aarch64 to avoid having to deal with both spellings everywhere
+ifeq ($(ARCH),arm64)
+override ARCH := aarch64
+endif
+
 ifeq ($(ARCH),i386)
 BINARY:=32
 ISX86:=1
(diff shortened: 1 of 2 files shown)


Make.inc
```diff
@@ -853,6 +858,10 @@ endif
 ifneq (,$(findstring aarch64,$(ARCH)))
 OPENBLAS_DYNAMIC_ARCH:=0
 OPENBLAS_TARGET_ARCH:=ARMV8
+ifeq ($(OS),Darwin)
+# Apple Chips are all at least A12Z
+MCPU:=apple-a12
```
yuyichao, line 863: So the main issue I worry about this here is that this could override the setting of the compiler if it is ever changed to a different one.

It's certainly much less of an issue on apple platform where there are only going to be relatively few platforms to keep track of and it can still be overwritten if necessary by setting `MCPU` on the make command line so I guess this is OK...

src/crc32c.c
```diff
@@ -341,6 +345,8 @@ static crc32c_func_t crc32c_dispatch(unsigned long hwcap)
 // For ifdef detection below
 #    define crc32c_dispatch() crc32c_dispatch(getauxval(AT_HWCAP))
 #    define crc32c_dispatch_ifunc "crc32c_dispatch"
+#  else
+#  warning CRC32 feature detection not implemented for this OS. Falling back to software version.
```
yuyichao, line 349: Hmm, time to rebase my local crc32 branch for 32bit arm I guess... ;-p

yuyichao (approved):

src/crc32c.c
Keno: Isn't thins whole thing guarded by _CPU_AARCH64_? Not sure I understand the comment.

yuyichao: Yeah, I though thihs ifdef was one level up (see edit...)

PR #36625, opened 2020-07-12
Darwin/ARM64: Fix throwing of synchronous memory exceptions

Keno:
Needs decoding of the ARM64 ESR register. Additionally, the Aarch64 ABI requires 16 byte aligned stack
pointers on function entry, so we can't push the return pointer there. Whether we need to construct a
stack frame instead will depend on the unwinder (which doesn't work yet), but I don't believe so,
since the ABI mandates a frame pointer.

### DIFF
--- a/src/signals-mach.c
+++ b/src/signals-mach.c
@@ -133,6 +133,12 @@ typedef arm_exception_state64_t host_exception_state_t;
 #define THREAD_STATE_COUNT ARM_THREAD_STATE64_COUNT
 #define HOST_EXCEPTION_STATE ARM_EXCEPTION_STATE64
 #define HOST_EXCEPTION_STATE_COUNT ARM_EXCEPTION_STATE64_COUNT
+
+enum aarch64_esr_layout {
+    EC_MASK = ((uint32_t)0b111111) << 25,
+    EC_DATA_ABORT = ((uint32_t)0b100100) << 25,
+    ISR_DA_WnR = ((uint32_t)1) << 6
+};
 #endif
 
 static void jl_call_in_state(jl_ptls_t ptls2, host_thread_state_t *state,
@@ -141,11 +147,11 @@ static void jl_call_in_state(jl_ptls_t ptls2, host_thread_state_t *state,
     uint64_t rsp = (uint64_t)ptls2->signal_stack + sig_stack_size;
     assert(rsp % 16 == 0);
 
+#ifdef _CPU_X86_64_
     // push (null) $RIP onto the stack
     rsp -= sizeof(void*);
     *(void**)rsp = NULL;
 
-#ifdef _CPU_X86_64_
     state->__rsp = rsp; // set stack pointer
     state->__rip = (uint64_t)fptr; // "call" the function
 #else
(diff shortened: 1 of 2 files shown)


Keno:
Oh, also added a test for this code path, since I don't think we had one.

src/signals-mach.c
```diff
@@ -141,11 +147,11 @@ static void jl_call_in_state(jl_ptls_t ptls2, host_thread_state_t *state,
     uint64_t rsp = (uint64_t)ptls2->signal_stack + sig_stack_size;
     assert(rsp % 16 == 0);
 
+#ifdef _CPU_X86_64_
     // push (null) $RIP onto the stack
     rsp -= sizeof(void*);
     *(void**)rsp = NULL;
```
yuyichao, line 153: The corresponding aarch64 operation is setting `lr` to zero. https://github.com/JuliaLang/julia/blob/a23a4ff08da5b6d95e9a35eee96e3260a452c02b/src/signals-unix.c#L135

yuyichao: And is https://github.com/JuliaLang/julia/blob/a23a4ff08da5b6d95e9a35eee96e3260a452c02b/src/signals-unix.c#L162 used?

test/misc.jl
```diff
@@ -842,3 +842,15 @@ end
 @testset "fieldtypes Module" begin
     @test fieldtypes(Module) isa Tuple
 end
+
+# Test that read fault on a prot-none region does not incorrectly give
+# ReadOnlyMemoryEror, but rather crashes the program
+const MAP_ANONYMOUS_PRIVATE = Sys.isbsd() ? 0x1002 : 0x22
+let script = :(let ptr = Ptr{Cint}(ccall(:jl_mmap, Ptr{Cvoid},
```
yuyichao, line 849: Do we actually guarantee this everywhere else?

Keno: Yes, I think we have this logic on every platform.

src/signals-mach.c
Keno: Right, I was wondering if we should set lr to NULL, but I haven't looked at the unwinder yet. I guess it can't really hurt.

test/misc.jl
yuyichao: It throws readonlymmoryerror on linux for example. I don't really see why it's that important to differentiate reading `PROT_NONE` from writing `PROT_READ`. The linux one only check for permission violation and would tread violation of read/write/execution the same.

yuyichao (changes requested):
Either the test or the implementation needs to be updated.

src/signals-mach.c
yuyichao: Yeah it shoudn't hurt. And I think at least in one case I've seen it helps the unwinder (either ours or gdb) terminates the stack trace.

test/misc.jl
yuyichao: I imaging what you could do is to unmap the page before issueing the `unsafe_load`. Assuming the code is compiled it should make sure you are loading from a hole in the mapping and that should be a crash.

Keno: > I imaging what you could do is to unmap the page before issueing the `unsafe_load`. Assuming the code is compiled it should make sure you are loading from a hole in the mapping and that should be a crash.

That doesn't exercise this code path. We explicitly check that the mapping exists first.

Keno: I think we may just want to fix the linux behavior here to bring it in line with mac/linux. I don't see a good reason to throw ReadOnlyMemoryError here.

yuyichao: ... Sure, that's fine. It's technically breaking but it was an error before anyway...

Keno: Or it's a bug fix and brings linux in line with the other platforms ;).

yuyichao: Breaking for ppl (like me) who hardly ever use julia anywhere other than linux... Not that I've ever cought any read only memory error when reading things anyway.............

yuyichao: I do agree the name of the exception is very suggestive of what it **should** do...

PR #36630, opened 2020-07-12
Improve docstring for `baremodule`.

EricForgy:


base/docs/basedocs.jl
```diff
@@ -126,8 +126,22 @@ kw"__init__"
 """
     baremodule
 
-`baremodule` declares a module that does not contain `using Base`
-or a definition of [`eval`](@ref Base.eval). It does still import `Core`.
+`baremodule` declares a module that does not contain `using Base` or a definition of
```
simeonschaub, line 129: Trailing Whitespace:
```suggestion
`baremodule` declares a module that does not contain `using Base` or a definition of
```

EricForgy:
Thanks for suggestion @simeonschaub

base/docs/basedocs.jl
```diff
@@ -126,8 +126,22 @@ kw"__init__"
...
+```
+
+is equivalent to
+
+```julia
+baremodule Foo
+
+using Base
+
+eval(ex) = Core.eval(@__MODULE__, ex)
```
mcabbott, line 143: Could each of these perhaps have an `end`, a bit less confusing? 
```suggestion
    using Base
    eval(ex) = Core.eval(@__MODULE__, ex)
end
```

base/docs/basedocs.jl
```diff
@@ -126,8 +126,22 @@ kw"__init__"
 """
     baremodule
 
-`baremodule` declares a module that does not contain `using Base`
-or a definition of [`eval`](@ref Base.eval). It does still import `Core`.
+`baremodule` declares a module that does not contain `using Base` or a definition of
+[`eval`](@ref Base.eval). It does still import `Core`. In other words,
+
+```julia
+module Foo
```
mcabbott, line 133:
```suggestion
module Foo end
```

base/docs/basedocs.jl
```diff
@@ -126,8 +126,22 @@ kw"__init__"
...
+```
+
+is equivalent to
+
+```julia
+baremodule Foo
+
+using Base
+
```
KristofferC, line 142: I think there is also an `include` method being defined.

simeonschaub:
Fun fact: I just discovered this is already documented [here](https://docs.julialang.org/en/v1/manual/modules/#Default-top-level-definitions-and-bare-modules-1), but I don't think it would hurt, to have this in the do string as well.

EricForgy:
Thanks guys 👍 

Good catch @simeonschaub. I went ahead and updated it to be the same as the other section of docs (including the `end`) 😊

PR #36635, opened 2020-07-13
more DEPOT_PATH docs

jw3126:
From https://discourse.julialang.org/t/high-view-description-of-the-folders-in-julia/20506/2
I need a little guidance here. I feel like the `DEPOT_PATH` docstring is not the ideal place to explain this. Any better location where this info could live?

### DIFF
--- a/base/initdefs.jl
+++ b/base/initdefs.jl
@@ -71,7 +71,7 @@ environment variable if set.
 ## DEPOT_PATH contents
 
 Each entry in `DEPOT_PATH` is a path to a directory which contains subdirectories used by Julia for various purposes.
-of whats typically found there:
+Here is an overview of some of the subdirectories that may exist in a depot:
 
 * `clones`: Contains full clones of package repos. Maintained by `Pkg.jl` and used as a cache.
 * `compiled`: Contains precompiled `*.ji` files for packages. Maintained by Julia.
--- a/base/initdefs.jl
+++ b/base/initdefs.jl
@@ -70,7 +70,7 @@ environment variable if set.
 
 ## DEPOT_PATH contents
 
-The directory in `DEPOT_PATH` contains a set of subdirectories, used for different purposes. Here is an overview
+Each entry in `DEPOT_PATH` is a path to a directory which contains subdirectories used by Julia for various purposes.
 of whats typically found there:
 
 * `clones`: Contains full clones of package repos. Maintained by `Pkg.jl` and used as a cache.
--- a/base/initdefs.jl
+++ b/base/initdefs.jl
@@ -74,16 +74,12 @@ The directory in `DEPOT_PATH` contains a set of subdirectories, used for differe
 of whats typically found there:
 
 * `clones`: Contains full clones of package repos. Maintained by `Pkg.jl` and used as a cache.
-* `compiled`: Contains precompiled `*.ji` files for packages. Maintained by `Pkg.jl`.
+* `compiled`: Contains precompiled `*.ji` files for packages. Maintained by Julia.
 * `dev`: Default directory for `Pkg.develop`. Maintained by `Pkg.jl` and the user.
-* `environments`: Default package environments. For instance the global environment for a specific julia version. Maintained by `Pkg.jl`
-* `logs`: Contains logs of `Pkg` and `REPL` operations. Maintained by `Pkg.jl` and `Julia`. CHECK WHO ELSE LOGS HERE?
-* `packages`: Contains packages, some of which were explicitly installed and some which are implicit dependencies. Maintained by `Pkg.jl`
-* `registries`: Contains package registries. By default only `General`. Maintained by `Pkg.jl`
-
-Certain packages, also create and populate other subdirectories. For instance:
-* `conda`: Maintained by `Conda.jl`
-* `prefs`: Maintained by `IJulia.jl`
+* `environments`: Default package environments. For instance the global environment for a specific julia version. Maintained by `Pkg.jl`.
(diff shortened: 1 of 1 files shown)


base/initdefs.jl
```diff
@@ -68,6 +68,23 @@ registries, packages, etc. installed and managed by system administrators.
...
+## DEPOT_PATH contents
+
+The directory in `DEPOT_PATH` contains a set of subdirectories, used for different purposes. Here is an overview
+of whats typically found there:
+
+* `clones`: Contains full clones of package repos. Maintained by `Pkg.jl` and used as a cache.
+* `compiled`: Contains precompiled `*.ji` files for packages. Maintained by `Pkg.jl`.
+* `dev`: Default directory for `Pkg.develop`. Maintained by `Pkg.jl` and the user.
+* `environments`: Default package environments. For instance the global environment for a specific julia version. Maintained by `Pkg.jl`
+* `logs`: Contains logs of `Pkg` and `REPL` operations. Maintained by `Pkg.jl` and `Julia`. CHECK WHO ELSE LOGS HERE?
```
jw3126, line 80: Anything else logs here?

base/initdefs.jl
```diff
@@ -68,6 +68,23 @@ registries, packages, etc. installed and managed by system administrators.
 `DEPOT_PATH` is populated based on the [`JULIA_DEPOT_PATH`](@ref JULIA_DEPOT_PATH)
 environment variable if set.
 
+## DEPOT_PATH contents
+
+The directory in `DEPOT_PATH` contains a set of subdirectories, used for different purposes. Here is an overview
+of whats typically found there:
+
+* `clones`: Contains full clones of package repos. Maintained by `Pkg.jl` and used as a cache.
+* `compiled`: Contains precompiled `*.ji` files for packages. Maintained by `Pkg.jl`.
```
jw3126, line 77: Is `Pkg.jl` the only thing that create `.ji` files here?

base/initdefs.jl
```diff
@@ -68,6 +68,23 @@ registries, packages, etc. installed and managed by system administrators.
 `DEPOT_PATH` is populated based on the [`JULIA_DEPOT_PATH`](@ref JULIA_DEPOT_PATH)
 environment variable if set.
 
+## DEPOT_PATH contents
+
+The directory in `DEPOT_PATH` contains a set of subdirectories, used for different purposes. Here is an overview
+of whats typically found there:
+
+* `clones`: Contains full clones of package repos. Maintained by `Pkg.jl` and used as a cache.
```
jw3126, line 76: So this gets populated, whenever I do a `] add SomePackage#some_branch_etc`? Anything else?

base/initdefs.jl
```diff
@@ -68,6 +68,23 @@ registries, packages, etc. installed and managed by system administrators.
...
+* `dev`: Default directory for `Pkg.develop`. Maintained by `Pkg.jl` and the user.
+* `environments`: Default package environments. For instance the global environment for a specific julia version. Maintained by `Pkg.jl`
+* `logs`: Contains logs of `Pkg` and `REPL` operations. Maintained by `Pkg.jl` and `Julia`. CHECK WHO ELSE LOGS HERE?
+* `packages`: Contains packages, some of which were explicitly installed and some which are implicit dependencies. Maintained by `Pkg.jl`
+* `registries`: Contains package registries. By default only `General`. Maintained by `Pkg.jl`
+
+Certain packages, also create and populate other subdirectories. For instance:
+* `conda`: Maintained by `Conda.jl`
+* `prefs`: Maintained by `IJulia.jl`
+
```
jw3126, line 87: Anything else that I should list here?

base/initdefs.jl
fredrikekre: Pkg isn't involved in this.

base/initdefs.jl
fredrikekre: I don't think any "other" directories should be listed here. This is essentially just packages that have "gone rogue" (due to Pkg /Julia lacking a system for this so far).

base/initdefs.jl
jw3126: Okay thanks. I realize, that I have little knowledge about responsibilities of Pkg vs Julia. Pkg is responsible for finding out which package(versions) belong to an environment and hand these paths over to Julia? And then Julia takes over and is responsible for the rest (e.g. loading/creating `.ji` files)?

base/initdefs.jl
```diff
@@ -68,6 +68,19 @@ registries, packages, etc. installed and managed by system administrators.
 `DEPOT_PATH` is populated based on the [`JULIA_DEPOT_PATH`](@ref JULIA_DEPOT_PATH)
 environment variable if set.
 
+## DEPOT_PATH contents
+
+The directory in `DEPOT_PATH` contains a set of subdirectories, used for different purposes. Here is an overview
```
StefanKarpinski, line 73:
```suggestion
Each entry in `DEPOT_PATH` is a path to a directory which contains subdirectories used by Julia for various purposes.
```

base/initdefs.jl
```diff
@@ -68,6 +68,19 @@ registries, packages, etc. installed and managed by system administrators.
...
+of whats typically found there:
```
StefanKarpinski, line 74:
```suggestion
Here is an overview of some of the subdirectories that may exist in a depot:
```

base/initdefs.jl
StefanKarpinski: Not entirely sure these days but close enough.

StefanKarpinski (approved):

PR #36638, opened 2020-07-13
Pass force=true option to rename in case of loading failure in compilecache

### LINKED ISSUE #36621: ArgumentError: `C:\..\compiled\..\_.ji' exists.
I am getting this error frequently as I edit a development branch on github and then `pkg> add ErrorfreeArithmetic#fix` and then (in a new REPL) `using` it. The very good news is that the message is helpfully specific (see below) so I can remedy the situation.  The issue is that Julia appears to know what needs to be done and the only file in jeopardy is a `.ji` file which autoregenerates if missing.  

Why not rename (somehow move) that file to `_.ji.1` or whatever and proceed without involving me? If the old file is kept, then `Pkg.gc()` pbly should cull it.
  
```julia
julia> using ErrorfreeArithmetic
[ Info: Precompiling ErrorfreeArithmetic [90fa49ef-747e-5e6f-a989-263ba693cf1a]
ERROR: ArgumentError: 'C:\Users\jas\.julia\compiled\v1.6\ErrorfreeArithmetic\nMLcX_1H60c.ji' exists. `force=true` is required to remove 'C:\Users\jas\.julia\compiled\v1.6\ErrorfreeArithmetic\nMLcX_1H60c.ji' before copying.
Stacktrace:
 [1] checkfor_mv_cp_cptree(src::String, dst::String, txt::String; force::Bool)
   @ Base.Filesystem .\file.jl:308
 [2] cp(src::String, dst::String; force::Bool, follow_symlinks::Bool)
   @ Base.Filesystem .\file.jl:345
 [3] rename(src::String, dst::String)
   @ Base.Filesystem .\file.jl:898
 [4] compilecache(pkg::Base.PkgId, path::String)
   @ Base .\loading.jl:1341
 [5] _require(pkg::Base.PkgId)
   @ Base .\loading.jl:1030
 [6] require(uuidkey::Base.PkgId)
   @ Base .\loading.jl:928
 [7] require(into::Module, mod::Symbol)
   @ Base .\loading.jl:923
```

```juliah
julia> versioninfo()
Julia Version 1.6.0-DEV.378
Commit c76960224f (2020-07-06 08:44 UTC)
Platform Info:
  CPU: Intel(R) Xeon(R) E-2176M  CPU @ 2.70GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
Environment:
  JULIA_NUM_THREADS = 8
```

musm:
This fixes https://github.com/JuliaLang/julia/issues/36621 for me

What's happening is that the rename via `jl_fs_rename` fails and then we have a `cp` failure since `force=true` is not passed.

### DIFF
--- a/base/file.jl
+++ b/base/file.jl
@@ -890,12 +890,11 @@ function unlink(p::AbstractString)
 end
 
 # For move command
-function rename(src::AbstractString, dst::AbstractString)
+function rename(src::AbstractString, dst::AbstractString; force::Bool=false)
     err = ccall(:jl_fs_rename, Int32, (Cstring, Cstring), src, dst)
     # on error, default to cp && rm
     if err < 0
-        # force: is already done in the mv function
-        cp(src, dst; force=false, follow_symlinks=false)
+        cp(src, dst; force=force, follow_symlinks=false)
         rm(src; recursive=true)
     end
     nothing
--- a/base/loading.jl
+++ b/base/loading.jl
@@ -1338,7 +1338,7 @@ function compilecache(pkg::PkgId, path::String)
             chmod(tmppath, filemode(path) & 0o777)
 
             # this is atomic according to POSIX:
-            rename(tmppath, cachefile)
+            rename(tmppath, cachefile; force=true)
             return cachefile
         end
     finally


base/file.jl
```diff
@@ -890,12 +890,11 @@ function unlink(p::AbstractString)
 end
 
 # For move command
-function rename(src::AbstractString, dst::AbstractString)
+function rename(src::AbstractString, dst::AbstractString; force::Bool=false)
     err = ccall(:jl_fs_rename, Int32, (Cstring, Cstring), src, dst)
     # on error, default to cp && rm
     if err < 0
-        # force: is already done in the mv function
```
musm, line 897: I don't know what is meant by this comment? I think this looks like a very old vestige, when this function got split?

base/loading.jl
```diff
@@ -1338,7 +1338,7 @@ function compilecache(pkg::PkgId, path::String)
             chmod(tmppath, filemode(path) & 0o777)
 
             # this is atomic according to POSIX:
-            rename(tmppath, cachefile)
+            rename(tmppath, cachefile; force=true)
```
kpamnany, line 1341: This could simply be changed to `mv(tmppath, cachefile; force = true)`, and then the changes to `rename()` are unnecessary.

But I'm not too happy with this solution because the reason this was added in the first place was for distributed execution on a shared filesystem, so that processes wouldn't error out because of racing to create the cache. With this fix, every process will overwrite the cache file -- the intended behavior was for all the processes that lost the race to silently fail to rename their temporary cache to the primary cache.

I was thinking of adding the following before the `finally`:
```julia
catch e
    if !isa(e, ArgumentError)
        throw(e)
    end
```

Or something like that but it doesn't matter much I suppose; we don't save on the system calls with the intended way and `rename` is atomic so it shouldn't break anything.

Any thoughts on this @KristofferC?

musm: @kpamnany  Note the current error in the linked issue does not involve distributed execution. The error crops up upon *any* recompilation of  packages, which makes it impossible to use `Pkg` on master without manually deleting the procompilation files.

kpamnany: Yes, I understand. The swiftest fix is to change the `rename` to `mv` with `force = true` as I said.

musm: Got it, might be best if you could open a PR to discuss your proposed improvement over this stop-gap solution.

kpamnany: On inspection, `mv()` uses `checkfor_mv_cp_cptree()` _before_ `rename()` and that function does an `rm()`, thus destroying the atomicity of `rename()` which is what we wanted in the first place. So your fix here is good.

musm: I'm going to merge this in the meantime because I can't currently develop and use Julia packages on master with this bug.

It's probably worthwhile to explore something like in  https://github.com/JuliaLang/julia/pull/36638#discussion_r453820564 in  a separate PR

PR #36640, opened 2020-07-13
Makefile: don't install non-existing 'julia.svg' file

StefanKarpinski:


ViralBShah:
I wonder if we will lose a desktop icon in an installer.

nalimilan:
I had missed this. If we install a .desktop file, we should ship the icon with it. Otherwise, stop installing a .desktop file. But that file is required by the AppData file too.

Overall would rather have kept that .svg file, keeping it updated isn't so much work, and we have contrib/windows/julia.ico to keep in sync anyway.

ViralBShah:
We don't install a desktop file - since we only provide tarballs. But distro packagers maybe installing it.

nalimilan:
Yes, but the .desktop file is stored in this repo...

ViralBShah:
We should probably get rid of the .desktop file too.

ViralBShah:
In the ideal world, we would keep all the building of installers separate from this repo, which should only build a tarball or zip, and then have separate repos for installers for each platform.

ViralBShah:
If it breaks existing workflows, we should just add the svg back.

nalimilan:
I don't see what we would gain by moving installers to separate repo. It's easier to keep things in sync by using a single repo, and easier to discover for users and packagers.

The removal of the SVG required me to adjust my RPM .spec files. Not a big deal, but having it in the repo rather than in another, harder to find one seemed logical to me.

StefanKarpinski:
I don't really care if this is here or elsewhere—I deleted this from the Makefile because someone else had already deleted the SVG.

ViralBShah:
Ok, we should bring the logo back (we'll get the updated design), and we should visit the .ico as well and make sure all the logos are in sync. The Makefile can stay as is, but the files will be available for packagers.

PR #36641, opened 2020-07-13
Make `isempty(c::Channel)` a non-mutating operation

staticfloat:
Previously, `isempty(c::Channel)` would fall back to `iterate(c) ===
nothing`, which actually consumed a value from the channel.  Instead,
let's just define it in terms of `isready(c)` since that already exists.

### DIFF
--- a/base/channels.jl
+++ b/base/channels.jl
@@ -419,6 +419,7 @@ on a [`put!`](@ref).
 """
 isready(c::Channel) = n_avail(c) > 0
 n_avail(c::Channel) = isbuffered(c) ? length(c.data) : length(c.cond_put.waitq)
+isempty(c::Channel) = isbuffered(c) ? isempty(c.data) : isempty(c.cond_put.waitq)
 
 lock(c::Channel) = lock(c.cond_take)
 unlock(c::Channel) = unlock(c.cond_take)
--- a/test/channels.jl
+++ b/test/channels.jl
@@ -470,6 +470,45 @@ end
     @test_throws InvalidStateException Base.check_channel_state(c)
 end
 
+# PR #36641
+# Ensure that `isempty()` does not mutate a Channel's state:
+@testset "isempty(::Channel) mutation" begin
+    function isempty_timeout(c::Channel)
+        inner_c = Channel{Union{Bool,Nothing}}()
+        @async put!(inner_c, isempty(c))
+        @async begin
+            sleep(0.01)
+            if isopen(inner_c)
+                put!(inner_c, nothing)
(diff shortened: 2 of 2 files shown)


StefanKarpinski:
Seems like arguably a bugfix. I would not think that `isempty` should consume anything in general.

vtjnash (approved):
I wonder if we should also define this in terms of `isempty` on the internal objects (O(1)) instead of length (O(n)), but this is a good fix regardless

staticfloat:
I'm honestly not sure when this fallback is a good idea:

https://github.com/JuliaLang/julia/blob/110765a87af68120f2f9f4aa0bbc4054db491359/base/essentials.jl#L773

It seems a bad idea to me to ever call `iterate(itr)` to check if it's empty.

Keno:
> I'm honestly not sure when this fallback is a good idea:

We have very few iterators that are mutable, but they have a few extra requirements. We should probably document what those are ... or at least figure out what they are ;).

Keno:
That said, there is a bit of a problem with this kind of query in general, since `isempty` being true here doesn't mean that you can't get an object from it, it just means that it'll block. Obviously this definition is strictly better than the fallback we had before (since that answer would be both useless and maybe drop an object you care about), but I'm wondering if we should have it at all.

staticfloat:
> since isempty being true here doesn't mean that you can't get an object from it, it just means that it'll block

I think that's the difference between `isempty()` and `!isopen()`; just because something is empty doesn't mean it's permanently empty.

JeffBezanson:
I am quite sure we did not intend `isempty` to be used on Channels, since it doesn't really make sense to have both that and `isready`. I also wish `isready` did not exist, since it's just a race condition. So I'm curious, how did this come up?

staticfloat:
> how did this come up?

The usual method of applying verbs from other areas and seeing if they work; I had a `Channel` that I closed and wanted to drain, so I wrote:

```julia
close(c)
while !isempty(c)
    process(take!(c))
end
```

Which seemed to work, except then I noticed it was only processing every other item.  I would be fine with defining `isempty()` more narrowly and telling everyone to use `isready()`.  I agree that there's something of a mixture of concerns here; what my use case _really_ wants is for `eof(::Channel)` to be defined, which would be something like `!isopen(c) && !isready(c)`.  That way we aren't relying upon the caller to have already closed the channel to avoid raciness.

JeffBezanson:
I think the only thing that really works for that is something like `iterate`, where you either get a value or find out there aren't any more.

PR #36642, opened 2020-07-13
restrict last(v::AbstractArray, n::Integer) to AbstractVector

stevengj:
As explained in my comment in #34868, we should be consistent here.

… in general, this optimization only seems to make sense for arrays with linear indexing, but it's easiest to just use AbstractVector here and rely on the `Iterators.reverse` fallback for other cases.

cc @giordano


giordano:
I remember there were some cases where the specialised `first` method with a non-vector argument would have returned something unexpected, but I can't remember what exactly was the problem.  Instead, I didn't find any issue with `last`, but I agree it's probably be safer to be conservative and restrict also this `last` method to vectors only

PR #36643, opened 2020-07-13
add `intersect` for `CartesianIndices`

### LINKED ISSUE #33055: Intersection of CartesianIndices
Hey,
I recently stumbled across the fact, that intersection of CartesianIndices can be quite slow and scales with the size of the ranges that are used. As such I found out that there is no specialized version for the intersection of multiple CartesianIndices. Does this have any reason? If not, if you want and if you tell me where that line
```julia
Base.intersect(a::CartesianIndices, b::CartesianIndices) = CartesianIndices(intersect.(a.indices, b.indices))
```
 would go and what the proper naming conventions here are, I could do a pull request.
EDIT: How to handle intersect(IdentityUnitRange, IdentityUnitRange) and mixed intersect?

https://github.com/JuliaLang/julia/blob/master/base/multidimensional.jl#L385
And while looking into this file, why do you not use a short-circuiting `&&` in this line?

stev47:
fixes #33055

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -67,6 +67,7 @@ Standard library changes
 * `Libdl` has been moved to `Base.Libc.Libdl`, however it is still accessible as an stdlib ([#35628]).
 * `first` and `last` functions now accept an integer as second argument to get that many
   leading or trailing elements of any iterable ([#34868]).
+* `intersect` on `CartesianIndices` now returns `CartesianIndices` instead of `Vector{<:CartesianIndex}` ([#36643]).
 
 #### LinearAlgebra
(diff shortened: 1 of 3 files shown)


mbauman:
Nice, thanks!  I think we should do this, but it's perhaps worth noting that this changes the return type from a mutable `Vector{CartesianIndex{N}}` to an immutable `N`-dimensional `CartesianIndices`.  I think nearly everyone who is using this will find it to be much more generally useful, and the fact that it's several complexity classes faster doesn't hurt, either.

mbauman (approved):

mbauman:
It feels like a compat annotation would just make the docstring for `intersect` noisy, but I suppose it'd be nice to mention this in NEWS.

stev47:
Ok, I added the news entry. The timeout on aarch64 seems unrelated to me.

Issue #36569, opened 2020-07-07
Sparse lu factorization reuse regression


dwfmarchant:
We've found what we think is a regression in the 1.5.0-rc1 sparse `lu` function.  

```julia
using LinearAlgebra
using SparseArrays

N = 12000
A = sprand(N, N, 0.0003)
A = A'*A + sparse(I, N, N)
x = zeros(N)

println("Factorization")
@time Afact = lu(A)

rhs = randn(N)
println("Solve #1")
@time x .= Afact \ rhs

rhs = randn(N)
println("Solve #2")
@time x .= Afact \ rhs

rhs = randn(N)
println("Solve #3")
@time x .= Afact \ rhs
```

In 1.4.2 this produces:
```
Factorization
  3.795173 seconds (246.00 k allocations: 1.102 GiB, 0.58% gc time)
Solve #1
  0.163573 seconds (243.62 k allocations: 13.421 MiB)
Solve #2
  0.090107 seconds (8 allocations: 750.219 KiB)
Solve #3
  0.089527 seconds (8 allocations: 750.219 KiB)
```
The factorization takes more time, but then subsequent back-solves are preformed quickly.

In 1.5.0-rc1 this produces:
```
Factorization
  3.892322 seconds (217.34 k allocations: 1.088 GiB, 0.62% gc time)
Solve #1
  3.676875 seconds (234.65 k allocations: 1.081 GiB, 2.27% gc time)
Solve #2
  3.575544 seconds (43 allocations: 1.069 GiB, 0.05% gc time)
Solve #3
  3.685203 seconds (43 allocations: 1.069 GiB, 0.08% gc time)
```
Here it appears that the factorization is being repeated at each back-solve.

If I switch the call to lu to cholesky everything works as expected in 1.4.2 and 1.5.0-rc1

My coworker @Pbellive ran a bisect which identified #33738 as the problem but we haven't had a chance to look into that further.

dkarrasch:
Just a quick side remark: if you want to solve in-place of `x`, then you better call `ldiv!(x, Afact, rhs)`. That doesn't solve the issue though, which remains a mystery to me. That PR added a `lu!` method, but that method doesn't seem to get called by the backsolve...

ViralBShah:
Pinging @goggle here as well. Should hopefully not be too hard to fix.

KristofferC:
>  That PR added a lu! method, but that method doesn't seem to get called by the backsolve...

[...]

PR #36652, opened 2020-07-14
Potential fix for lu factorization regression? (#36569)

IanButterworth:
Given this seems to be a blocker for 1.5 I thought I'd try out a fix

### DIFF
--- a/stdlib/SuiteSparse/src/umfpack.jl
+++ b/stdlib/SuiteSparse/src/umfpack.jl
@@ -221,7 +221,7 @@ function lu!(F::UmfpackLU, S::SparseMatrixCSC{<:UMFVTypes,<:UMFITypes}; check::B
     F.rowval = zerobased ? copy(rowvals(S)) : decrement(rowvals(S))
     F.nzval = copy(nonzeros(S))
 
-    umfpack_numeric!(F)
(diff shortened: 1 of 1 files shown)


ianshmean:
```
julia> using LinearAlgebra

julia> using SparseArrays

julia> begin
       N = 12000
       A = sprand(N, N, 0.0003)
       A = A'*A + sparse(I, N, N)
       x = zeros(N)

       println("Factorization")
       @time Afact = lu(A)

       rhs = randn(N)
       println("Solve #1")
       @time x .= Afact \ rhs

       rhs = randn(N)
       println("Solve #2")
       @time x .= Afact \ rhs

       rhs = randn(N)
       println("Solve #3")
       @time x .= Afact \ rhs
       end
Factorization
  3.035685 seconds (188.71 k allocations: 1.027 GiB, 2.18% gc time)
Solve #1
  0.169768 seconds (241.44 k allocations: 14.804 MiB)
Solve #2
  0.097856 seconds (8 allocations: 750.219 KiB)
Solve #3
  0.098167 seconds (8 allocations: 750.219 KiB)
```

KristofferC:
Let's go with this since we really need to get out 1.5 RC2. If there is something bad with it, it can be fixed for 1.5 final.

PR #36658, opened 2020-07-14
stacktraces doc: fix typos

rfourquet:


doc/src/manual/stacktraces.md
```diff
@@ -88,18 +88,16 @@ Symbol("~/julia/usr/share/julia/stdlib/v0.7/REPL/src/REPL.jl")
...
-julia> top_frame.inlined
+julia> frame.inlined
 false
 
-julia> top_frame.from_c
+julia> frame.from_c
 false
-```
 
-```julia-repl
```
rfourquet, line 101: The block below was separated out in the past because it was the only one which wouldn't be doctested, but this difference doesn't exist anymore.

PR #36661, opened 2020-07-14
lower header level to make the content be rendered as one chapter

likanzhan:
Is it OK if we lower the level of the last header so that this document will be rendered as one chapter in the PDF version rather than two separate chapters?


KristofferC:
Please provide more descriptive titles to PRs

StefanKarpinski:
The commit message can be edited on squash/merge.

PR #36663, opened 2020-07-14
Update `@isdefined` to have clearer examples

pdeffebach:
In reference to the discourse thread [here](https://discourse.julialang.org/t/test-for-existence-of-a-variable/42845/25)


KristofferC:
How is this an improvement over the current example?

yuyichao:
Adding a global scope example is fine but it definitely should not replace the pure local variable one. In fact, since testing a global variable is possible with functions and such function exists (isdefined) it's more important to show the macro work on local variables.

compleathorseplayer:
Experts like yourselves undoubtedly have your ideas about the most important and 'correct' usage of commands like this, but I have a long experience of trying to motivate young people (and myself) to engage with the unfamiliar, and (in my opinion), the simplest possible examples that can illustrate usage or concept seem to work the best for this, even if it seems to oversimplify the usual case.

The basic idea here is whether x is there or it isn't  
[ the delicacy of context can be learned later.]

If x is there, then we get
julia> @isdefined x
true

 If x is not there, then
julia> @isdefined x
false

This is simple, and jumps out at the reader.   The fine print of where 'there' refers to can be figured out later.

Also, I think disallowing isdefined(x) with a single argument was unhelpful, and I spent an hour trying to understand what the two argument - version was for, and never quite got it (myself MIT-educated and a widely-published  author in the Mathematical Sciences) .
   
I have expressed opinions elsewhere that (IMO) creeping deprecation of the most intuitive aspects of Julia is disappointing.
Obviously, I can't do anything about that (far be it for me to question), but as a user, I can offer feedback on documentation.   Students (and perpetual newbies like myself) like to be told what not to worry about (such as extra arguments) initially.   If they (and I) see the value and get interested, then the rest can usually figured out.

yuyichao:
> Experts like yourselves undoubtedly have your ideas about the most important and 'correct' usage of commands like this, but I have a long experience of trying to motivate young people (and myself) to engage with the unfamiliar, and (in my opinion), the simplest possible examples that can illustrate usage or concept seem to work the best for this, even if it seems to oversimplify the usual case.

And a wrong example that seems simple is very misleading. Especially if you replaces the more representative usage of it.

> Also, I think disallowing isdefined(x) with a single argument was unhelpful, and I spent an hour trying to understand what the two argument - version was for, and never quite got it (myself MIT-educated and a widely-published author in the Mathematical Sciences) .

No. The single argument **function** never does what you think it does. `isdefined(:x)` never tell you if writing `x` will give you an error or not.

> I have expressed opinions elsewhere that (IMO) creeping deprecation of the most intuitive aspects of Julia is disappointing.
> Obviously, I can't do anything about that (far be it for me to question), but as a user, I can offer feedback on documentation. Students (and perpetual newbies like myself) like to be told what not to worry about (such as extra arguments) initially. If they (and I) see the value and get interested, then the rest can usually figured out.

Explaination like what you are trying to change the doc to in this case is exactly what can lead people to have a misunderstanding of `isdefined` and/or `@isdefined` like what you have right now. And that's precisely what I'm saying you must not replace the current example.

pdeffebach:
> Also, I think disallowing isdefined(x) with a single argument was unhelpful,

To clarify, the reason `isdefined(:x)` isn't useful is because the behavior would have to depend on the scope that the function is called in. I.e., for checking if a variable `x` is defined for a global scope of some module, we would have to rely on a global state inside the function call. This is not best practices. 

Additionally, it's quite hard to get this global behavior right, you can see from the source code that `isdefined` is not defined like other julia funtions. It has it's own Expression similar to `:vect`. 

```
macro isdefined(s::Symbol)
    return Expr(:isdefined, esc(s))
end
```

Additionally, your proposal

```
@isdefined x
```

is likely to always return `true`. I always have a variable called `x` in my environment, especially if I'm learning a new language. 

A better syntax might be 

```
@isdefined newvar_unique_123288
```

so that there will be no name collisions. However even this example will only work exactly once, which may lead to additional confusion.

StefanKarpinski:
I think it makes sense to demonstrate `@isdefined` in both global and local scope since in practice they are quite different. I'm not too concerned about people trying to evaluate the global example twice since I think they will fairly intuitively understand why it doesn't work the second time—because the variable is already defined.

StefanKarpinski (approved):

Issue #30703, opened 2019-01-12
No color in printstyled when pipe is used


ronisbr:
Hi guys!

I am trying to save a text file with the output of a Julia script but with the escape sequences so that I can render the colors. However, when I call Julia using pipe, I get no colors. Even if I have `--color=yes`. You can test by starting Julia as:

    julia --color=yes | tee

The logo will be gray (but `julia>` will be colored). Is there a way to provide another option, like `--color=force` so that `printstyled` always use the escape sequences?

ronisbr:
As a workaround, one can use:

```
screen -dm -L sh -c 'julia --color=yes teste.jl' && mv screenlog.0 output.txt
```

and then `output.txt` will have the colored output (with escape sequences).

PR #36671, opened 2020-07-15
Make `with_output_color` respect `have_color`

giordano:
Since v1.6 has to be the most colourful release yet, I thought it would be nice to have logging use colours in CI, which is currently never the case in GitHub Actions for example.

I'm not sure this is the right way to check whether the function should use colours since it ignores the stream, but I believe `iscolor` should be `true` when `--color=yes` is passed.  I figured I'd open a questionable PR to get the right answer :slightly_smiling_face:

### DIFF
--- a/base/util.jl
+++ b/base/util.jl
@@ -69,7 +69,7 @@ text_colors
 
 function with_output_color(@nospecialize(f::Function), color::Union{Int, Symbol}, io::IO, args...; bold::Bool = false)
     buf = IOBuffer()
-    iscolor = get(io, :color, false)::Bool
+    iscolor = get_have_color()
     try f(IOContext(buf, io), args...)
     finally
         str = String(take!(buf))


Keno:
Why doesn't the io object have the color flag set? I don't think checking the global flag is correct here.

giordano:
> Why doesn't the io object have the color flag set?

On GitHub Actions it's not clear.

> I don't think checking the global flag is correct here.

Fair enough, but at least checking whether `--color=yes` is set should be?

Keno:
--color=yes should just force the color flag to be set on stdout. It's supposed to be a force override for the tty detection.

giordano:
On GitHub Actions, `stdout` isn't a `TTY`:
```
stdout = Base.PipeEndpoint(RawFD(0x0000000d) open, 0 bytes waiting)
```
But ANSI escape sequences are rendered correctly as colours when forced.

Keno:
Right, so `--color=yes` should override wherever we add the color property to try objects and also do that for other kinds of io.

giordano:
Sorry, I'm not sure what you suggest to do.

tkf:
Maybe do this

https://github.com/JuliaLang/julia/blob/8320fcc7d90ad2523a8f3a3410e7e2d6ac1aa1bf/base/ttyhascolor.jl#L25

for `PipeEndpoint` as well?

tkf:
But I think it's bad how `tput` on stdout is reflected to `TTY` and `PipeEndpoint` of potentially irrelevant file descriptors.  It'd be nice if `TTY` and `PipeEndpoint` structs have `have_color` field (maybe that's what Keno was saying).

Keno:
Just wrap stdout in an iocontext when the io object is initialized and the color flag is set

tkf:
Doesn't it break `redirect_stdout` etc.?  Though maybe they should work with `IOContext`.

tkf:
```julia
julia> methods(redirect_stdout)
# 4 methods for generic function "redirect_stdout":
[1] redirect_stdout(handle::Union{IOStream, Base.LibuvStream}) in Base at stream.jl:1148
[2] redirect_stdout() in Base at stream.jl:1154
[3] redirect_stdout(::Base.DevNull) in Base at stream.jl:1160
[4] redirect_stdout(f::Function, stream) in Base at stream.jl:1213

To edit a specific method, type the corresponding number into the REPL and press Ctrl+Q

julia> redirect_stdout(IOContext(stdout, :color => false))
ERROR: MethodError: no method matching redirect_stdout(::IOContext{Base.TTY})
```

Keno:
Yeah, just fix those to make them work with an IOContext

kimikage:
cf #30703

giordano:
Yes, also with a pipe `stdout` is a `PipeEndpoint`:
```
% julia -e '@show stdout' | cat
stdout = Base.PipeEndpoint(RawFD(0x0000000e) open, 0 bytes waiting)
```

giordano:
I had a look at wrapping `stdout` in an `IOContext`: `stdout` is defined in https://github.com/JuliaLang/julia/blob/77487611fd9751c0f31ac3853072e6ca11efaf05/base/coreio.jl#L31 `IOContext` is defined in https://github.com/JuliaLang/julia/blob/77487611fd9751c0f31ac3853072e6ca11efaf05/base/show.jl#L222-L230 `get_have_color` is defined in https://github.com/JuliaLang/julia/blob/77487611fd9751c0f31ac3853072e6ca11efaf05/base/ttyhascolor.jl#L17-L21 `base/coreio.jl` is the very first file included in `base/Base.jl` https://github.com/JuliaLang/julia/blob/77487611fd9751c0f31ac3853072e6ca11efaf05/base/Base.jl#L36 so I don't see how it's possible to wrap `stdout` as suggested.

As an alternative, what about this change in this PR:
```diff
@@ -69,7 +69,7 @@ text_colors
 
 function with_output_color(@nospecialize(f::Function), color::Union{Int, Symbol}, io::IO, args...; bold::Bool = false)
     buf = IOBuffer()
-    iscolor = get(io, :color, false)::Bool
+    iscolor = io === stdout ? get_have_color() : get(io, :color, false)::Bool
     try f(IOContext(buf, io), args...)
     finally
         str = String(take!(buf))
```
`with_output_color` would use `get_have_color` only for `stdout`, as intended, and fallback to `get(io, :color)` in all other cases.  I ran `make testall` locally and all tests pass (apart from a couple in `Sockets` that seem related to DNS issues, not printing).  Also a couple of doctests that I saw failing here would be successful with this change.

Related to #30703, with this change:
```
$ julia -e 'printstyled("Hello world\n"; color=:cyan)' | cat -A
^[[36mHello world^[[39m$
$ julia --color=yes -e 'printstyled("Hello world\n"; color=:cyan)' | cat -A
^[[36mHello world^[[39m$
$ julia --color=no -e 'printstyled("Hello world\n"; color=:cyan)' | cat -A
Hello world$
```

JeffBezanson:
I think the main problem is that multiple places check `get(io, :color, false)`, and this would introduce a discrepancy. Also `get_have_color` calls `ttyhascolor`, which will colorize output even if it is redirected to a file. I also worry about command line options changing the output that a program sends to some random I/O stream. I think the right way to do this is to manually pass `:color=>true` when you know you have a non-TTY that wants color ouput --- in other words the testsuite code should call `redirect_stdout` to something wrapped in an IOContext.

giordano:
> I think the main problem is that multiple places check `get(io, :color, false)`, and this would introduce a discrepancy.

We maybe can find a good interface?

> Also `get_have_color` calls `ttyhascolor`, which will colorize output even if it is redirected to a file.

Not if `--color=no` like in the example above.

> in other words the testsuite code should call `redirect_stdout` to something wrapped in an IOContext.

It doesn't look like `redirect_stdout` works with `IOContext`?

JeffBezanson:
>  `stdout = Core.stdout`

Good news; it does not need to be set that early. We can re-set `stdout` while processing command line arguments. It should also be possible to implement the redirect functions for `IOContext`; I will experiment.

giordano:
Closing in favour of #36689 :slightly_smiling_face:

PR #36673, opened 2020-07-15
explain method signatures for `Any` arguments

heliosdrm:
Context: https://discourse.julialang.org/t/function-type-signature/43101/17?u=heliosdrm


tpapp:
Yes, a sentence mentioning this, maybe with an example, would be useful. Perhaps right at the beginning of the *Methods* chapter, before the example with constraints in the method signature.

JeffBezanson:
This failed doctests.

StefanKarpinski:
Ooops.

heliosdrm:
So, should I wait until https://github.com/JuliaDocs/Documenter.jl/pull/1369 is solved, and the redo this PR? Or do you have another suggestion to work around it? Thanks.

PR #36676, opened 2020-07-15
digits docstring no need to allocate

Wikunia:
`sum` works without allocating => this speeds up the computation.
I'm not sure whether I should run Documenter locally or whether this is done via Travis/GitHub Actions


mbauman (approved):

mbauman:
Welcome! This looks great.  There's no need to run documenter locally; that'll happen automatically assuming GitHub is actually working properly.

StefanKarpinski (approved):

PR #36687, opened 2020-07-16
Need `LinearAlgebra` to use `I`

likanzhan:
Need `LinearAlgebra` to use `I`

### DIFF
--- a/stdlib/SparseArrays/src/sparsematrix.jl
+++ b/stdlib/SparseArrays/src/sparsematrix.jl
@@ -98,6 +98,7 @@ Returns the number of stored (filled) elements in a sparse array.
 
 # Examples
 ```jldoctest
+julia> using LinearAlgebra
 julia> A = sparse(2I, 3, 3)
 3×3 SparseMatrixCSC{Int64,Int64} with 3 stored entries:
  2  ⋅  ⋅
@@ -130,6 +131,7 @@ modifications to the returned vector will mutate `A` as well. See
 
 # Examples
 ```jldoctest
+julia> using LinearAlgebra
 julia> A = sparse(2I, 3, 3)
 3×3 SparseMatrixCSC{Int64,Int64} with 3 stored entries:
  2  ⋅  ⋅
@@ -158,6 +160,7 @@ nonzero values. See also [`nonzeros`](@ref) and [`nzrange`](@ref).
 
 # Examples
 ```jldoctest
+julia> using LinearAlgebra
 julia> A = sparse(2I, 3, 3)
 3×3 SparseMatrixCSC{Int64,Int64} with 3 stored entries:
  2  ⋅  ⋅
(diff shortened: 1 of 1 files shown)


StefanKarpinski:
Can you add blank lines after each of these since that's how the REPL prints them? (Note that this breaks the doctests, so you may want to make sure it doesn't do that.)

stdlib/SparseArrays/src/sparsematrix.jl
```diff
@@ -98,6 +98,9 @@ Returns the number of stored (filled) elements in a sparse array.
 
 # Examples
 ```jldoctest
+
```
StefanKarpinski, line 101:
```suggestion
```

stdlib/SparseArrays/src/sparsematrix.jl
```diff
@@ -130,6 +133,9 @@ modifications to the returned vector will mutate `A` as well. See
...
```
StefanKarpinski, line 136:
```suggestion
```

stdlib/SparseArrays/src/sparsematrix.jl
```diff
@@ -158,6 +164,9 @@ nonzero values. See also [`nonzeros`](@ref) and [`nzrange`](@ref).
...
```
StefanKarpinski, line 167:
```suggestion
```

stdlib/SparseArrays/src/sparsematrix.jl
```diff
@@ -664,6 +673,9 @@ Convert an AbstractMatrix `A` into a sparse matrix.
...
```
StefanKarpinski, line 676:
```suggestion
```

stdlib/SparseArrays/src/sparsematrix.jl
```diff
@@ -3322,6 +3334,9 @@ Concatenate matrices block-diagonally. Currently only implemented for sparse mat
...
```
StefanKarpinski, line 3337:
```suggestion
```

StefanKarpinski (changes requested):
No blank lines above the first prompt. In general: take a look at how the surrounding code and/or docs look and emulate it.

Keno:
Would it be better to just have SparseArrays re-export `I`?

StefanKarpinski:
Let's see if tests pass these days.

KristofferC:
You need to rebase it since the buildbots don't run on the merge commit.

vtjnash:
We already do this, via `make.jl`, so if we think it should be in each example, we should remove it from the global config:
```
DocMeta.setdocmeta!(
    SparseArrays,
    :DocTestSetup,
    maybe_revise(:(using SparseArrays, LinearAlgebra));
    recursive=true, warn=false,
)
```

I'm inclined to instead reject this PR, but I'm not opposed to it if others feel it is reads better.

PR #36688, opened 2020-07-16
allow redirecting standard streams to `IOContext`

JeffBezanson:
This will be helpful for #36671, allowing a non-TTY stdout to be set to `:color=>true`.

PR #36689, opened 2020-07-16
fix some issues with color configuration

### LINKED ISSUE #30703: No color in printstyled when pipe is used

JeffBezanson:
The issues here were truly surprising. Try running julia with `--color=no` (i guess I never do). Most color goes away, but all prompts are still bold, and while other prompts are un-colored, the Pkg prompt is still blue. I almost abandoned hope when I saw that. Then there is #30703, where redirecting to a pipe (`julia | tee`) removes most color, except *all* prompts are colored and bold, and `--color=yes` doesn't re-enable color. So this does the following:

- If `--color` is specified and stdout or stderr is not a TTY, wrap it in an IOContext with the color setting.
- Consistently use `get` on an io stream everywhere to determine whether to use color.
- Thread the color setting through to `write_prompt` to disable bold.
- Disable beep color if `hascolor=false` is passed to the REPL.

Needs #36688 to work. Should fix or help #36671. Fixes #30703.

### DIFF
--- a/base/client.jl
+++ b/base/client.jl
@@ -371,13 +371,12 @@ function run_main_repl(interactive::Bool, quiet::Bool, banner::Bool, history_fil
         invokelatest(REPL_MODULE_REF[]) do REPL
             term_env = get(ENV, "TERM", @static Sys.iswindows() ? "" : "dumb")
             term = REPL.Terminals.TTYTerminal(term_env, stdin, stdout, stderr)
-            color_set || (global have_color = REPL.Terminals.hascolor(term))
             banner && Base.banner(term)
             if term.term_type == "dumb"
                 active_repl = REPL.BasicREPL(term)
                 quiet || @warn "Terminal not fully functional"
             else
-                active_repl = REPL.LineEditREPL(term, have_color, true)
+                active_repl = REPL.LineEditREPL(term, get(stdout, :color, false), true)
                 active_repl.history_file = history_file
             end
             # Make sure any displays pushed in .julia/config/startup.jl ends up above the
@@ -487,7 +486,7 @@ function _start()
         invokelatest(display_error, catch_stack())
         exit(1)
     end
-    if is_interactive && have_color === true
+    if is_interactive && get(stdout, :color, false)
         print(color_normal)
     end
 end
--- a/stdlib/REPL/src/LineEdit.jl
+++ b/stdlib/REPL/src/LineEdit.jl
@@ -395,7 +395,7 @@ function refresh_multi_line(termbuf::TerminalBuffer, terminal::UnixTerminal, buf
     regstart, regstop = region(buf)
     written = 0
     # Write out the prompt string
-    lindent = write_prompt(termbuf, prompt)
+    lindent = write_prompt(termbuf, prompt, hascolor(terminal))
     # Count the '\n' at the end of the line if the terminal emulator does (specific to DOS cmd prompt)
     miscountnl = @static Sys.iswindows() ? (isa(Terminals.pipe_reader(terminal), Base.TTY) && !Base.ispty(Terminals.pipe_reader(terminal))) : false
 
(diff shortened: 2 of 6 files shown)


stdlib/REPL/src/REPL.jl
```diff
@@ -849,7 +855,7 @@ end
 
 function reset(repl::LineEditREPL)
     raw!(repl.t, false)
-    print(repl.t, Base.text_colors[:normal])
+    hascolor(repl) && print(repl.t, Base.text_colors[:normal])
```
vtjnash, line 858: Possible odd to return a Boolean?

vtjnash (approved):
Epic!

stdlib/REPL/src/REPL.jl
```diff
@@ -436,9 +436,15 @@ mutable struct LineEditREPL <: AbstractREPL
     last_shown_line_infos::Vector{Tuple{String,Int}}
     interface::ModalInterface
     backendref::REPLBackendRef
-    LineEditREPL(t,hascolor,prompt_color,input_color,answer_color,shell_color,help_color,history_file,in_shell,in_help,envcolors) =
+    function LineEditREPL(t,hascolor,prompt_color,input_color,answer_color,shell_color,help_color,history_file,in_shell,in_help,envcolors)
+        opts = Options()
+        opts.hascolor = hascolor
+        if !hascolor
+            opts.beep_colors = [""]
+        end
```
rfourquet, line 444: Maybe this logic belongs in the `Options` constructor taking kwargs? e.g. in the body of that method, replace `beep_colors` by `hascolor ? beep_colors : [""]` (or change the default value by `hascolor ? ["\e[90m"] : [""]`).

stdlib/REPL/src/LineEdit.jl
```diff
@@ -1259,21 +1259,21 @@ refresh_line(s, termbuf) = refresh_multi_line(termbuf, s)
 default_completion_cb(::IOBuffer) = []
 default_enter_cb(_) = true
 
-write_prompt(terminal, s::PromptState) = write_prompt(terminal, s.p)
+write_prompt(terminal, s::PromptState, hascolor::Bool) = write_prompt(terminal, s.p, hascolor)
```
StefanKarpinski, line 1262: Maybe call this just `color`? It's typically only set if the terminal has color, but it could be called with anything.

Keno (approved):

Keno:
Modulo passing CI of course ;)

giordano:
With this PR now I get
```
$ ./julia --color=yes -e 'printstyled(stderr, "Hello world\n"; color=:cyan)' 2>&1 | cat -A
^[[36mHello world^[[39m$
$ ./julia --color=yes -e '@info "Hello world"' 2>&1 | cat -A
[ Info: Hello world$
```
:thinking:

giordano:
As far as I can see, the problem is that the stream of the global logger doesn't have the `color` flag, even though it should be `stderr`:
```
$ usr/bin/julia --color=yes -e 'using Logging; @show get(global_logger().stream, :color, nothing)' 2>&1 | cat -A
get((global_logger()).stream, :color, nothing) = nothing$
```

base/client.jl
```diff
@@ -210,6 +210,14 @@ function exec_options(opts)
     history_file          = (opts.historyfile != 0)
     color_set             = (opts.color != 0) # --color!=auto
     global have_color     = color_set ? (opts.color == 1) : nothing # --color=on
+    if color_set
+        if !isa(stdout, TTY)
+            global stdout = IOContext(stdout, :color => have_color::Bool)
+        end
+        if !isa(stderr, TTY)
+            global stderr = IOContext(stderr, :color => have_color::Bool)
```
tkf, line 218: I think fixing the issue @giordano observed https://github.com/JuliaLang/julia/pull/36689#issuecomment-659782477 requires calling `global_logger(ConsoleLogger(stderr))` again here?  This is because, at this point, `Logging.__init__` is already executed and `global_logger` is configured without `IOContext`.

giordano:
Seems to finally work
```
$ ./julia -e '@info "Hello world"' 2>&1 | cat -A
[ Info: Hello world$
$ ./julia --color=yes -e '@info "Hello world"' 2>&1 | cat -A
^[[36m^[[1m[ ^[[22m^[[39m^[[36m^[[1mInfo: ^[[22m^[[39mHello world$
$ ./julia --color=no -e '@info "Hello world"' 2>&1 | cat -A
[ Info: Hello world$
```
:tada:

ianshmean:
🌈

ianshmean:
Confirmed to work in github actions (by locally building this PR, then mounting as a github self-hosted runner)
image

giordano:
The failure in macOS seems unrelated?  Good to merge otherwise?

PR #36690, opened 2020-07-16
stop showing shortcut tips in REPL for every stacktrace/methodlist

### LINKED ISSUE #36575: "To edit a specific method, type the corresponding number into the REPL and press Ctrl+Q" is annoying
I think printing this message once is a good idea, but every time is overkill. Could this simply be printed once (with a message saying that this message will not be printed after this, so that users pay attention the first time)? Eg by writing a file in .julia.

KristofferC:
After reading this message probably a hundred times it is, unfortunately, starting to get quite annoying. I think showing "tooltip" style of messages during standard usage (non-error situations) will always have this weakness in that it leads to fatigue in seeing the (after reading it the first time) useless message over and over and over.

This PR also fixes a small bug with trying to open REPL frames (it should do nothing).

Fixes https://github.com/JuliaLang/julia/issues/36575


JeffBezanson:
This is broken:
```
Error showing value of type Base.MethodList:
ERROR: type #repl has no field last_shown_line_infos
Stacktrace:
  [1] setproperty!(x::Function, f::Symbol, v::Vector{Tuple{String,Int64}})
    @ Base ./Base.jl:34
  [2] display(d::REPL.REPLDisplay, mime::MIME{Symbol("text/plain")}, x::Any)
    @ REPL ~/src/julia/usr/share/julia/stdlib/v1.6/REPL/src/REPL.jl:217
  [3] display(d::REPL.REPLDisplay, x::Any)
    @ REPL ~/src/julia/usr/share/julia/stdlib/v1.6/REPL/src/REPL.jl:221
```

PR #36691, opened 2020-07-16
add `only` to docs of `Ref` object

Moelf:


StefanKarpinski (approved):

stevengj:
Why are we mentioning `only` here?  I get that it should be *possible*, but I would always recommend `[]` for containers that by definition contain exactly one element, like `Ref` or 0-dimensional arrays.

Moelf:
mainly because of https://discourse.julialang.org/t/what-does-mean-in-this-case/43175/ , `only` is more verbose than `[]` since `[]` in many lines of `[begin:blah]` or alike is quite 'invisible';

of course, the same problem exists for both `[]` and `only` is that you can't tell if it's a `Ref` or a `[x]`.

PR #36693, opened 2020-07-16
Add a dash in the Windows installer path between the version number

### LINKED ISSUE #35720: Would it be possible to remove the space in the name of the DEV Julia?
On Windows C:\Users\PKrysl\AppData\Local\Programs\Julia 1.5.0-DEV\bin 
is not as convenient as
C:\Users\PKrysl\AppData\Local\Programs\Julia-1.5.0-DEV\bin
In other words, no spaces is in the path. Would that be possible?

musm:
Close https://github.com/JuliaLang/julia/issues/35720

It seems like the space is causing some people issues with 'non-standard' escaping that might be tricky in some configurations.


StefanKarpinski (approved):

musm:
CI failures  unrelated

PR #36694, opened 2020-07-16
fix `hasfield` for unionall-wrapped DataTypes

JeffBezanson:
Other similar functions like fieldname and fieldnames work, so this should too.

### DIFF
--- a/base/reflection.jl
+++ b/base/reflection.jl
@@ -668,7 +668,7 @@ function fieldindex(T::DataType, name::Symbol, err::Bool=true)
     return Int(ccall(:jl_field_index, Cint, (Any, Any, Cint), T, name, err)+1)
 end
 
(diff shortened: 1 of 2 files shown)


base/reflection.jl
```diff
@@ -668,6 +668,8 @@ function fieldindex(T::DataType, name::Symbol, err::Bool=true)
     return Int(ccall(:jl_field_index, Cint, (Any, Any, Cint), T, name, err)+1)
 end
 
+fieldindex(t::UnionAll, name::Symbol, err::Bool=true) = fieldindex(unwrap_unionall(t), name, err)
```
vtjnash, line 671:
```suggestion
fieldindex(t::UnionAll, name::Symbol, err::Bool=true) = fieldindex(something(argument_datatype(t)), name, err)
```

PR #36696, opened 2020-07-16
Revert "explain method signatures for `Any` arguments"

StefanKarpinski:
Reverts JuliaLang/julia#36673

It seems that doctests can't handle the case where the output includes something that looks like a comment.

cc @mortenpi — could we support this in a doctest?


fredrikekre:
I think it just misses the tooltip part of the output and will be fixed by something like https://github.com/JuliaLang/julia/pull/36690 too.

fredrikekre:
Eh, nevermind; The line starts with # and is treated as a comment.

PR #36697, opened 2020-07-16
build: make libunwind linked dynamically

vtjnash:
Not sure why this was first linked statically, as the commit that
introduced this simply had the message "restoring stuff that seems to
have been clobbered by the revert of the unintended merge to master".

Nearly all other libraries that we use are linked dynamically.


ararslan:
> Nearly all other libraries that we use are linked dynamically.

Off topic here, but linking libuv dynamically would likely fix #34627. Generally :+1: to dynamically linking more dependencies. :slightly_smiling_face:

vtjnash:
This is now dependent on us building libosxunwind.dylib correctly (https://github.com/JuliaLang/libosxunwind/pull/16)

Keno:
IIRC the problem was that we would pick up libosxunwind from libSystem. Do your build system tweaks fix that? Also, separate from this PR, but we should switch our upstream to the LLVM copy to pick up aarch64 support.

vtjnash:
Possibly—our dylib had accidentally just been an empty shell re-exporting libSystem. Can’t say it will work now, but definitely wouldn’t have worked before. And agreed we should attempt to apply our diff on top of a newer release.

staticfloat (approved):
LGTM!  I'm happy to have more things used through a more regularized interface

PR #36701, opened 2020-07-17
add length function to Enum

blairdrummond:
Adds a `length` function that you can call like `length(Fruit)`.

It's equivalent to `length(instances(YourEnum))`.

```
julia> @enum Fruit apple=1 orange=2 kiwi=3
julia> length(Fruit)
3
```

I was motivated by the `len(YourEnum)` from Python. [Example of someone looking for this](https://discourse.julialang.org/t/the-length-of-a-enum-object-in-julia/19865/3)

*Please let me know if I made any mistakes in the documentation/testing*

### DIFF
--- a/base/Enums.jl
+++ b/base/Enums.jl
@@ -115,6 +115,13 @@ To list all the instances of an enum use `instances`, e.g.
 julia> instances(Fruit)
 (apple, orange, kiwi)
 ```
+
+To get the number of instances of an enum use `length`, e.g.
+
+```jldoctest fruitenum
+julia> length(Fruit)
+3
+```
 """
 macro enum(T, syms...)
     if isempty(syms)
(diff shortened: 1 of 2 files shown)


JeffBezanson:
Welcome, and thanks for noticing this!

Enum types are not themselves iterable or collections, so I'm not sure `length` makes sense. `length(instances(type))` seems like a good way to express this --- it's clear you're counting the instances.

blairdrummond:
Hmm, I suppose it's true that at the moment they aren't iterable... but perhaps they should be? It seems really natural to be able to enumerate/iterate over an enumerable, doesn't it?

blairdrummond:
```julia
Base.iterate(t::Type{<:Enum}) = iterate(instances(t))
Base.iterate(t::Type{<:Enum}, i::Int64) = iterate(instances(t), i)

@enum Week begin
    MONDAY
    TUESDAY
    WEDNESDAY
    THURSDAY
    FRIDAY
    SATURDAY
    SUNDAY
end

for day in Week
    println(day)
end

println(length(Week))
```

I think it seems intuitive; doesn't really take much work and I don't think there's ambiguity. But if you think it's a poor fit or that it could cause confusion then that's fair too. I'm pretty new to Julia so maybe I haven't absorbed the Julia-isms yet

KristofferC:
To me,  iterating on a type like that looks pretty strange. I also think that using `instances` works pretty well as is.

PR #36702, opened 2020-07-17
RFC: Give MethodMatch its own type

Keno:
Every time I see any code dealing with the svecs we get back from
the method match code, I think that it shouldn't really be an svec:
1. It's always the same length
2. Basically every use of it typeasserts the field type
3. Every use of it needs the same magic numbers to access the fields.

All put together this should just be a type. This basically does that,
but there's a few things missing, so before I go finish it, I figured,
I'd make sure that there isn't some reason not to do this.

TODO:
1. (deprecated) getindex/iterate for `MethodMatch` to avoid breaking packages that use the internal API
2. Make the bitfield a uint8_t for compatibility with compiler that use a different layout
3. Double check the gf.c changes. I think I fatfingered one of the mappings.

Doesn't need extensive review in this state, since it isn't done,
just a quick yay/nay on the direction.

### DIFF
--- a/src/jltypes.c
+++ b/src/jltypes.c
@@ -37,6 +37,7 @@ jl_datatype_t *jl_typedslot_type;
 jl_datatype_t *jl_argument_type;
 jl_datatype_t *jl_const_type;
 jl_datatype_t *jl_partial_struct_type;
+jl_datatype_t *jl_method_match_type;
 jl_datatype_t *jl_simplevector_type;
 jl_typename_t *jl_tuple_typename;
 jl_datatype_t *jl_anytuple_type;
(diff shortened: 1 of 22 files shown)


JeffBezanson:
Yes I definitely like this.

src/jltypes.c
```diff
@@ -2383,6 +2384,11 @@ void jl_init_types(void) JL_GC_DISABLED
                                        jl_perm_symsvec(2, "typ", "fields"),
                                        jl_svec2(jl_any_type, jl_array_any_type), 0, 0, 2);
 
+    jl_method_match_type = jl_new_datatype(jl_symbol("MethodMatch"), core, jl_any_type, jl_emptysvec,
+                                       jl_perm_symsvec(4, "metharg", "methsp", "method", "fully_covers"),
```
vtjnash, line 2388: Can we give these real words? (or at least the same mangling we do now):
`signature` (or `specTypes`), `env` (or `sparam_vals`), `method`, `<:` (or whatever this is)?

Keno: This is what they were called in inlining. I agree we can pick better names. How about just `specTypes`, `sparams`, `method`, `fully_covers`? I don't like using `<:` as a field name.

StefanKarpinski: How about not mixing snake case and camel case? 😬

Keno: `spec_types` then, though at some point we should also switch over `MethodInstance` then.

Keno:
Alright, this should be done now.

Issue #31202, opened 2019-02-28
Documenting all exports from the standard libraries


mbauman:
Adding docstrings to all exports from Base has been a great success (#26919)!  We are about 80% done there (and I just moved the goalposts — we had been close to 95% prior to that).  That's awesome — thanks everyone who participated.  Now let's move on to the standard libraries!

```julia
searchdocs(mod) = searchdocs!(Any[], mod)
function searchdocs!(objs, mod, seen=Set())
	mod in seen && return
	push!(seen, mod)
	for name in names(mod)
		isdefined(mod, name) || continue
		Base.isdeprecated(mod, name) && continue
		b = getfield(mod, name)
		docs = Base.Docs.doc(Base.Docs.Binding(mod, name))
		if startswith(string(docs), "No documentation found")
		    push!(objs, string(mod, '.', name))
		end
		b isa Module && searchdocs!(objs, b, seen)
	end
	objs
end
ms = []
for M in [filter(f->f[1] in 'A':'Z' && !occursin(r"[-.]", f) && f != "Makefile", readdir("stdlib")); "Pkg"]
    @eval import $(Symbol(M))
    @eval append!(ms, searchdocs($(Symbol(M))))
end
foreach(x->println("* [ ] `", x, '`'), ms)
```

* [x] `Base64.Base64` (#31332)
* [x] `Dates.Apr` (#31218)
* [x] `Dates.April` (#31218)
* [x] `Dates.Aug` (#31218)
* [x] `Dates.August` (#31218)
* [ ] `Dates.DatePeriod`
* [x] `Dates.Dec` (#31218)
* [x] `Dates.December` (#31218)
* [x] `Dates.Feb` (#31218)
* [x] `Dates.February` (#31218)
* [x] `Dates.Fri` (#31574)
* [x] `Dates.Friday` (#31574)
* [x] `Dates.ISODateFormat`
* [x] `Dates.ISODateTimeFormat`
* [x] `Dates.ISOTimeFormat`
* [x] `Dates.Jan`
* [x] `Dates.January`
* [x] `Dates.Jul` (#31218)
* [x] `Dates.July` (#31218)
* [x] `Dates.Jun` (#31218)
* [x] `Dates.June` (#31218)
* [x] `Dates.Mar` (#31218)
* [x] `Dates.March` (#31218)
* [x] `Dates.May` (#31218)
* [x] `Dates.Mon` (#31574)
* [x] `Dates.Monday` (#31574)
[...]

PR #36703, opened 2020-07-17
added docstrings for date exports

mrdofu:
See #31202


stdlib/Dates/src/types.jl
```diff
@@ -19,7 +19,32 @@ abstract type AbstractTime end
 `Period` types represent discrete, human representations of time.
 """
 abstract type Period     <: AbstractTime end
+
+"""
+    Year
```
JeffBezanson, line 24: I would put `DataPeriod` here, and list the subtypes in the description.

ViralBShah:
@quinnj Can you review if this should be merged?

quinnj (approved):
LGTM

PR #36704, opened 2020-07-17
gf lookup: avoid creating unintended specializations and cache misses

### LINKED ISSUE #36669: `@allocated` test in InteractiveUtils failing on CI
See https://build.julialang.org/#/builders/34/builds/1613 and https://build.julialang.org/#/builders/34/builds/1623. This seems new (within the last day or so).

vtjnash:
The typemap code was forgetting to handle the case where the user needed
to look for something complicated (like AbstractVec{Float64}) in the
method table cache (such as for ==).

edit: fixes #36669

### DIFF
--- a/base/compiler/typelattice.jl
+++ b/base/compiler/typelattice.jl
@@ -84,8 +84,10 @@ struct NotFound end
 
 const NOT_FOUND = NotFound()
 
-const CompilerTypes = Union{Type, MaybeUndef, Const, Conditional, NotFound, PartialStruct}
+const CompilerTypes = Union{MaybeUndef, Const, Conditional, NotFound, PartialStruct}
 ==(x::CompilerTypes, y::CompilerTypes) = x === y
+==(x::Type, y::CompilerTypes) = false
+==(x::CompilerTypes, y::Type) = false
 
 #################
 # lattice logic #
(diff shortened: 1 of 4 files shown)


base/compiler/typelattice.jl
```diff
@@ -84,8 +84,10 @@ struct NotFound end
 
 const NOT_FOUND = NotFound()
 
-const CompilerTypes = Union{Type, MaybeUndef, Const, Conditional, NotFound, PartialStruct}
-==(x::CompilerTypes, y::CompilerTypes) = x === y
+const CompilerTypes = Union{MaybeUndef, Const, Conditional, NotFound, PartialStruct}
+==(x::CompilerTypes, y::CompilerTypes) = false
```
JeffBezanson, line 88: I guess we never use this method, but this definition still seems a little hazardous?

Keno:
Did I remember correctly that this may help #36669?

vtjnash:
Yes, just waiting for this to finish CI (with Jeff's correction)

JeffBezanson (approved):

JeffBezanson:
Hmm, does the buildbot not have `pbpaste`?

Keno:
> Hmm, does the buildbot not have `pbpaste`?

Long story - don't ask. Should be fixed in the near future.

Issue #36714, opened 2020-07-17
Better default `sizeof(::AbstractArray)`?


mbauman:
By default, new `AbstractArray` subtypes fall back to the fully generic `sizeof(::Any)` implementation that simply gives the size of their fields.  Given that we know more about what an `AbstractArray` is — and given that we have a default `write` implementation, it seems like we should be able to do a better job of getting closer to the docstring of `sizeof`:

> Size, in bytes, of the canonical binary representation of the given `DataType` `T`, if any. Size, in bytes, of object `obj` if it  is not `DataType`.

For example:

```Julia
julia> write(IOBuffer(), rand(5,5))
200

julia> sizeof(rand(5,5))
200

julia> write(IOBuffer(), rand(5,5)')
200

julia> sizeof(rand(5,5)')
8
```

Even the "compressed" representations by default `write` their whole contents:

```
julia> using LinearAlgebra

julia> write(IOBuffer(), Diagonal(rand(5)))
200

julia> sizeof(Diagonal(rand(5)))
8
```

So it'd seem like a `sizeof(A::AbstractArray) = length(A) * sizeof(eltype(A))` definition would be better all around.  If your array has a specialized canonical representation, then you define `sizeof` to match when you implement `write`.

PR #36715, opened 2020-07-17
Fix `sizeof` for non-Array subarrays

mbauman:
Not all `AbstractArray`s define elsize.  Cf. #35900, #36714.


JeffBezanson:
Uhoh, now that I think about it this opens a bigger can of worms. When the dimension 1 stride is not 1, we write elements individually (no alignment padding), and otherwise in memory chunks (yes alignment padding). For SubArrays of Arrays, we should probably change that to always include the padding. I can open a separate PR for that.

Similarly, there might be many other types besides Array that need to use `elsize` here, so we should probably add a fallback for all AbstractArray types.

PR #36717, opened 2020-07-17
Add `haskey(::RegexMatch, ::Symbol)` to test for named capture groups

staticfloat:
Currently, the only way to test for a named capture group within
`RegexMatch` is to attempt to index into it and catch the error after it
fails; this adds a way to probe for a given key.


stevengj (approved):

staticfloat:
@StefanKarpinski I reverted one of your return code checks because I didn't want to rely on exceptions in order to figure out if a named capture group does or does not exist.  I hope that's okay; I checked that all call sites are properly dealing with that error code already.

StefanKarpinski:
Sure, that's probably a good change.

PR #36727, opened 2020-07-18
Fix correctness issues in sizeof tfunc

### LINKED ISSUE #36710: Broadcasting sizeof on Ptr returns wrong size
I'm on Julia v1.4.2 **32bit** so my pointers are only 4 bytes
Or so I thought... It seems `sizeof()` has some trouble broadcasting on a parametric `Ptr`.

```julia
julia> sizeof(Ptr)
4

julia> sizeof.((Ptr,))
(8,)

julia> sizeof.((Ptr{Cvoid},))
(4,)
```

Keno:
The `Core.sizeof` function can take either a value or a type,
which can be a bit confusing in the tfunc, because the tfunc
operates on the types of the values passed to the builtin.
In particular, we were incorrectly returning Const(16) (on 64bit)
for sizeof_tfunc(UnionAll), because it was treating it the asme
as `sizeof_tfunc(Const(UnionAll))`. Try to clean that up as well
as fixing a similar confusion in the nothrow version of this
function. Lastly, we had a similar fast path in codegen, which
would try to inline the actual size for a constant datatype argument.
For codegen, just rm that whole code path, since it should have no
more information than inference and figuring it out in inference
exposes strictly more optimization opportunities.

Fixes #36710

### DIFF
--- a/base/compiler/tfuncs.jl
+++ b/base/compiler/tfuncs.jl
@@ -302,16 +302,34 @@ function sizeof_nothrow(@nospecialize(x))
         end
     elseif isa(x, Conditional)
         return true
-    else
-        x = widenconst(x)
     end
     if isa(x, Union)
         return sizeof_nothrow(x.a) && sizeof_nothrow(x.b)
     end
-    isconstType(x) && (x = x.parameters[1]) # since sizeof(typeof(x)) == sizeof(x)
-    x === DataType && return false
-    return isconcretetype(x) || isprimitivetype(x)
+    t, exact = instanceof_tfunc(x)
+    if !exact
+        # Could always be bottom at runtime, which throws
+        return false
+    end
+    if t !== Bottom
+        t === DataType && return true
+        x = t
+        x = unwrap_unionall(x)
+        if isa(x, Union)
+            isinline, sz, _ = uniontype_layout(x)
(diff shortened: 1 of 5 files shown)


base/compiler/tfuncs.jl
```diff
@@ -302,16 +302,30 @@ function sizeof_nothrow(@nospecialize(x))
...
     end
-    isconstType(x) && (x = x.parameters[1]) # since sizeof(typeof(x)) == sizeof(x)
-    x === DataType && return false
+    t, exact = instanceof_tfunc(x)
+    if !exact
+        # Could always be bottom at runtime, which throws
+        return false
+    end
+    if t !== Bottom
+        # The value corresponding to `x` at runtime could be a type.
```
JeffBezanson, line 315: I think this still gives the wrong answer for e.g. `sizeof_nothrow(Type{Vector{Int}})` (or `Type{String}`). `sizeof(Vector{Int})` throws.

Keno: Good catch.

Keno:
@JeffBezanson I've updated this to more explicitly track the logic in the builtin. See what you think.

JeffBezanson (approved):

PR #36729, opened 2020-07-19
add operators `<--` and `<-->`

### LINKED ISSUE #36666: Feature request: additional plain text arrow symbols
Right now `-->` parses as a symbol, along with a fantastic number of unicode arrows. It would be nice if we could also have the reversed arrow, `<--`, and some kind of reversible plaintext arrow like `<->` or `<-->`. We use such notation in the DiffEqBiological DSL for specifying chemical reactions, i.e. `A + B --> C`, and have had several requests for some type of reversible arrow notation from several users now for whom using a unicode arrow was problematic.

Thanks!

JeffBezanson:
AFAICT these give parse errors currently, so there should be no problem. I'll separately PR a "minor change" to make `-->` a more normal operator.

fixes #36666

### DIFF
--- a/src/julia-parser.scm
+++ b/src/julia-parser.scm
@@ -15,7 +15,7 @@
 (define prec-conditional '(?))
 (define prec-arrow       (append!
                           '(-- -->)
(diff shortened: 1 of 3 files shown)


JeffBezanson:
Hmm, we might want to disallow `1<---2`, which this parses as `1 <-- -2`. Further `-`s are ok since you start getting an error about trying to use `--`.

src/julia-parser.scm
```diff
@@ -15,7 +15,7 @@
 (define prec-conditional '(?))
 (define prec-arrow       (append!
                           '(-- -->)
```
simeonschaub, line 17: Is there a reason, we don't have these with dots? If we have dotted versions of `<--` and `<-->`, it probably makes sense to have those as well.

JeffBezanson: That's part of the `-->` issue I referenced above.

simeonschaub: Ah, right! Didn't read Stefan's comment closely enough.

PR #36731, opened 2020-07-19
Added latex mimetype printing of complex and rational numbers

gustaphe:
`show(io::IO,mime::MIME"text/latex",x::Any)` now acts like the function with `mime:MIME"text/plain"` except 

- if `x::Complex` in which case it writes `i` instead of `im` and `\cdot` instead of `*`, and 
- if `x::Rational` in which case it writes `\frac{a}{b}` instead of `a//b`

### DIFF
--- a/base/complex.jl
+++ b/base/complex.jl
@@ -202,6 +202,27 @@ end
 show(io::IO, z::Complex{Bool}) =
     print(io, z == im ? "im" : "Complex($(z.re),$(z.im))")
 
+function show(io::IO, mime::MIME"text/latex",z::Complex)
+    # complex numbers are displayed with "i" instead of "im" in latex, and with "\cdot" instead of "*"
+    r,i = reim(z)
+    show(io, mime, r)
+    if signbit(i) && !isnan(i)
+        print(io," - ")
+        if isa(i,Signed) && i == typemin(typeof(i))
+            show(io, mime, -widen(i))
+        else
+            show(io, mime, -i)
+        end
+    else
+        print(io, " + ")
+        show(io, mime, i)
+    end
+    if !(isa(i,Integer) && !isa(i,Bool) || isa(i,AbstractFloat) && isfinite(i))
+        print(io, " \\cdot ")
+    end
+    print(io,"i")
+end
+
 function show_unquoted(io::IO, z::Complex, ::Int, prec::Int)
     if operator_precedence(:+) <= prec
         print(io, "(")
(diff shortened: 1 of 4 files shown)


gustaphe:
I misunderstood the testing procedure. `test-revise` worked fine. I'm trying to figure out why `test` fails, but I'm having trouble deciphering the error message. The only part of it that points to something I've edited is the line in `Base.jl` where `complex.jl` is imported, but the code inside runs fine when copied to REPL,

`make clean`, `make cleanall` and `git clean -fdx` all failed to improve the situation. Excuse a rookie contributor.

jmert:
The `MIME` type isn't defined yet when `complex.jl` and `rational.jl` are included — `MIME` is defined later in `multimedia.jl` — so that's probably error which is occurring. (You don't see an error at the REPL because by that point, all of `Base` is already defined, but when initially defining `Base`, you have to worry about the order in which the pieces are defined.)

gustaphe:
Thank you, that does indeed seem to have been the problem. I see now why the other `show` methods are defined in their own file.

I ran a `make clean testall` for this latest version, and it fails at `channels` and `FileWatching`. No clue what causes that, I don't see how my edits touch those tests at all. And how come `buildbot/tester_<architecture>` works for all architectures but `linux64`, is that normal?

JeffBezanson:
Welcome, @gustaphe, this is cool! The package Latexify.jl has some similar functionality; this should probably be incorporated there. It has a `latexify` function, but it doesn't seem to be hooked up to `show`, so these methods should probably by added to that package.

gustaphe:
@JeffBezanson Would that not constitute type piracy? Latexify.jl doesn't own `show`, `MIME"text/latex"`, `Rational`, `Complex` etc...

I can see how `show.jl` might be considered a bit cluttered if we were to take this to its logical conclusion (adding `3 \cdot 10^{15}` instead of `3e15` for `Floats` and conversion of unicode characters etc), but at least the generic method for `x::Any` would be good for expanding other packages (I've especially been thinking about Polynomials.jl and Unitful.jl). 

And the `MIME"text/html"` methods are defined in Base, so it would be consistent with that.

That said, I'm obviously not an expert, if you think this should be externalized I can go that route instead.

gustaphe:
@JeffBezanson Could you clarify, by "should be incorporated there" do you mean I should move these methods there, or I should fix this PR and have Latexify.jl use them?

stevengj:
I'm skeptical that we should have `show` methods for this, as opposed to having to call an explicit function like `latexify` to get LaTeX-math-formatted representations of objects.

Having these `show` methods would mean that complex numbers and rationals would display as LaTeX by default in IJulia; I'm not sure that this is what one would typically want, as opposed to the `repr` form that can by copy-pasted into other code.

gustaphe:
Okay, I have not used IJulia, and wasn't aware they were using a check like that. I'll go the long way around. Thank you for the feedback!

PR #36735, opened 2020-07-20
Update arrays.md

HelgavonLichtenstein:
Removed the comma between ``` x=rx y=ry]``` in ```A = [ F(x,y,...) for x=rx y=ry ... ]``` as it results in an ```syntax: unexpected "]"``` otherwise.

### DIFF
--- a/doc/src/manual/arrays.md
+++ b/doc/src/manual/arrays.md
@@ -226,7 +226,7 @@ Comprehensions provide a general and powerful way to construct arrays. Comprehen
 similar to set construction notation in mathematics:
 
 ```
-A = [ F(x,y,...) for x=rx, y=ry, ... ]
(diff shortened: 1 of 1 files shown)


rfourquet:
Can you show a concrete example with the error? Apparently there must not be a comma at the end of the range list (right before the `]`), but removing commas between range specifications as you suggest with this change is not valid.

HelgavonLichtenstein:
```
beginning = 2+30+1
ending = 2+30+50
[x+y for x=1:4, for y=beginning:ending]
```
This gives the `syntax: unexpected "]"` error, removing the comma makes it run.

rfourquet:
Yes after the comma no new `for` is allowed, try `[x+y for x=1:4, y=beginning:ending]`

HelgavonLichtenstein:
Ah, I somehow missed that. Thank you.

PR #36737, opened 2020-07-20
fix #36732: printing of single-entry matrix

### LINKED ISSUE #36732: wrong printing for single-element matrix
The following printing is wrong, as the output would produce a vector, not a matrix:
```julia
julia> show(IOContext(stdout, :compact=>true), fill(42, 1, 1))
[42]
```
Observed on master and 1.5.

simeonschaub:
I find it a bit odd, that we currently print every `AbstractMatrix` as if it were just a `Matrix`, but that's definitely a separate discussion. I tried to at least respect the `axes` of the matrix when printed as `fill(...)`, but if people deem that unnecessary, I could also just remove that.
fixes #36732

### DIFF
--- a/test/offsetarray.jl
+++ b/test/offsetarray.jl
@@ -223,7 +223,7 @@ targets1 = ["0-dimensional $OAs_name.OffsetArray{Float64,0,Array{Float64,0}}:\n1
             "1×1×1×1 $OAs_name.OffsetArray{Float64,4,Array{Float64,4}} with indices 2:2×3:3×4:4×5:5:\n[:, :, 4, 5] =\n 1.0"]
 targets2 = ["(fill(1.0), fill(1.0))",
             "([1.0], [1.0])",
-            "([1.0], [1.0])",
+            "(fill(1.0, Base.IdentityUnitRange(2:2), Base.IdentityUnitRange(3:3)), fill(1.0, Base.IdentityUnitRange(2:2), Base.IdentityUnitRange(3:3)))",
             "([1.0], [1.0])",
             "([1.0], [1.0])"]
 @testset "printing of OffsetArray with n=$n" for n = 0:4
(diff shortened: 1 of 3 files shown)


thofma:
Since when is `show` supposed to be `roundtripable`? Isn't this what `repr` is for?

simeonschaub:
`repr` is just the output of `show` as a string, we don't always guarantee roundtripability, but I think it's definitely confusing to print vectors and matrices the same here. We generally don't consider changing printing of types in `show` as very breaking, so I don't really see a problem with this PR in that respect.

test/offsetarray.jl
```diff
@@ -223,7 +223,7 @@ targets1 = ["0-dimensional $OAs_name.OffsetArray{Float64,0,Array{Float64,0}}:\n1
             "1×1×1×1 $OAs_name.OffsetArray{Float64,4,Array{Float64,4}} with indices 2:2×3:3×4:4×5:5:\n[:, :, 4, 5] =\n 1.0"]
 targets2 = ["(fill(1.0), fill(1.0))",
             "([1.0], [1.0])",
-            "([1.0], [1.0])",
+            "(fill(1.0, Base.IdentityUnitRange(2:2), Base.IdentityUnitRange(3:3)), fill(1.0, Base.IdentityUnitRange(2:2), Base.IdentityUnitRange(3:3)))",
```
simeonschaub, line 226: Maybe we should just ignore the `axes`?

mbauman:
This seems stranger than the status quo, IMO.

simeonschaub:
Do you mean just the `axes` bit or printing matrices with `fill` at all? I do think it's confusing not to be able to distinguish vectors from matrices if there is only a single element, especially since this is already a common source of confusion for people coming from Matlab.

mbauman:
I don't like that this introduces a very large discontinuity between `fill(0, (1,1))` and `fill(0, (2, 1))`, and I think newcomers will be more puzzled by a `fill` (and this discontinuity) than by the status quo.  Yes, they'll also see it for 0-dimensional arrays, but those are much rarer.

PR #36739, opened 2020-07-20
fix `elsize` and `write` for SubArrays of Arrays

JeffBezanson:
~~Alternative~~Followup to #36715.

### DIFF
--- a/base/io.jl
+++ b/base/io.jl
@@ -676,7 +676,7 @@ function write(s::IO, a::SubArray{T,N,<:Array}) where {T,N}
     if !isbitstype(T) || !isa(a, StridedArray)
         return invoke(write, Tuple{IO, AbstractArray}, s, a)
     end
-    elsz = sizeof(T)
+    elsz = elsize(a)
     colsz = size(a,1) * elsz
     GC.@preserve a if stride(a,1) != 1
         for idxs in CartesianIndices(size(a))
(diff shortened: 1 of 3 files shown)


base/abstractarray.jl
```diff
@@ -150,6 +150,7 @@ prevind(::AbstractArray, i::Integer) = Int(i)-1
 nextind(::AbstractArray, i::Integer) = Int(i)+1
 
 eltype(::Type{<:AbstractArray{E}}) where {E} = @isdefined(E) ? E : Any
+elsize(::Type{<:AbstractArray{T}}) where {T} = aligned_sizeof(T)
```
mbauman, line 153: This doesn't feel right to me, either.  The default `write(::AbstractArray)` just does:

```Julia
    for a in A
        nb += write(s, a)
    end
```

Which won't include alignment padding by default.

JeffBezanson: I guess that's right --- it seems we can't fix this more thoroughly without more drastic or breaking changes. For now we should probably do your PR + define `elsize` for `SubArray{<:Any,<:Any,<:Array}` + just the `write` change from here.

In the future we should consider writing alignment padding in `write(::AbstractArray)`, and possibly exporting `elsize`.

mbauman (approved):

PR #36743, opened 2020-07-20
Refactor compiler method lookup interface

Keno:
The primary motivation here is to clean up the notion of
"looking up a method in the method table" into a single
object that can be passed around. At the moment, it's
all a bit murky, with fairly large state objects being
passed around everywhere and implicit accesses to the
global environment. In my AD use case, I need to be a
bit careful to make sure the various inference and
optimization steps are looking things up in the correct
tables/caches, so being very explicit about where things
need to be looked up is quite helpful.

In particular, I would like to clean up the optimizer, to
not require the big `OptimizationState` which is currently
a bit of a mix of things that go into IRCode and information
needed for method lookup/edge tracking. That isn't part of
this PR, but will build on top of it.

More generally, with a bunch of the recent compiler work,
I've been trying to define more crisp boundaries between
the various components of the system, giving them clearer
interfaces, and at least a little bit of documentation.
The compiler is a very powerful bit of technology, but I
think people having been avoiding it, because the code
looks a bit scary. I'm hoping some of these cleanups will
make it easier for people to understand what's going on.
Here in particular, I'm using `findall(sig, table)` as
the predicate for method lookup. The idea being that
people familiar with the `findall(predicate, collection)`
idiom from regular julia will have a good intuitive
understanding of what's happening (a collection is searched
for a predicate), an array of matches is returned, etc.
Of course, it's not a perfect fit, but I think these
kinds of mental aids can be helpful in making it easier
for people to read compiler code (similar to how #35831
used `getindex` as the verb for cache lookup). While
I was at it, I also cleaned up the use of out-parameters
which leaked through too much of the underlying C API
and replaced them by a proper struct of results.

### DIFF
--- a/base/compiler/methodtable.jl
+++ b/base/compiler/methodtable.jl
@@ -1,0 +1,93 @@
+abstract type MethodTableView; end
+
+struct MethodLookupResult
+    # Really Vector{Core.MethodMatch}, but it's easier to represent this as
+    # and work with Vector{Any} on the C side.
+    matches::Vector{Any}
+    valid_worlds::WorldRange
+    ambig::Bool
+end
+length(result::MethodLookupResult) = length(result.matches)
+function iterate(result::MethodLookupResult, args...)
+    r = iterate(result.matches, args...)
+    r === nothing && return nothing
+    match, state = r
+    return (match::MethodMatch, state)
+end
+getindex(result::MethodLookupResult, idx::Int) = getindex(result.matches, idx)::MethodMatch
+
+"""
+    struct InternalMethodTable <: MethodTableView
+
+A struct representing the state of the internal method table at a
+particular world age.
+"""
+struct InternalMethodTable <: MethodTableView
+    world::UInt
+end
+
+"""
+    struct CachedMethodTable <: MethodTableView
+
+Overlays another method table view with an additional local fast path cache that
+can respond to repeated, identical queries faster than the original method table.
+"""
+struct CachedMethodTable{T} <: MethodTableView
+    cache::IdDict{Any, Union{Missing, MethodLookupResult}}
+    table::T
+end
+CachedMethodTable(table::T) where T =
+    CachedMethodTable{T}(IdDict{Any, Union{Missing, MethodLookupResult}}(),
+        table)
+
+"""
+    findall(sig::Type{<:Tuple}, view::MethodTableView; limit=typemax(Int))
+
+Find all methods in the given method table `view` that are applicable to the
+given signature `sig`. If no applicable methods are found, an empty result is
+returned. If the number of applicable methods exeeded the specified limit,
+`missing` is returned.
+"""
+function findall(@nospecialize(sig::Type{<:Tuple}), table::InternalMethodTable; limit::Int=typemax(Int))
+    _min_val = RefValue{UInt}(typemin(UInt))
+    _max_val = RefValue{UInt}(typemax(UInt))
+    _ambig = RefValue{Int32}(0)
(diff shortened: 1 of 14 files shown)


Keno:
Realized `jl_gf_invoke_lookup` also needs the method table. It wasn't quite clear to me what the correct verb should be for that query, but I ended up settling on `findsup`. In particular the method it looks for is the least upper bound/supremum among the methods that covers the requested signature (under the subtype/specificity relation) and in particular also the supremum of the subset of methods returned by `findall` so it seemed fitting. `findsup` isn't (currently) a verb we have in our find/search API, but `findmax` wasn't quite accurate, since the method found need not be the maximal element returned by `findall`. I considered keeping it as `which` or `invoke_lookup` or something, but that just seems circular to me "how is invoke implemented? "well, it looks up the method to invoke" ". I think `findsup` is fairly explicit about what the query actually is, which I think would be helpful to people reading the compiler source and trying to understand how things work. @JeffBezanson I'd like your opinion on this naming, since you've probably thought about the right names for these kinds of things most.

base/compiler/methodtable.jl
```diff
@@ -0,0 +1,93 @@
...
+    r === nothing && return nothing
+    match, state = r
+    return (match::MethodMatch, state)
+end
+getindex(result::MethodLookupResult, idx::Int) = getindex(result.matches, idx)::MethodMatch
+
+"""
+    struct InternalMethodTable <: MethodTableView
+
+A singleton struct representing the state of the internal method table at a
```
JeffBezanson, line 22:
```suggestion
A struct representing the state of the internal method table at a
```

base/compiler/methodtable.jl
```diff
@@ -0,0 +1,93 @@
...
+particular world age.
+"""
+struct InternalMethodTable <: MethodTableView
+    world::UInt
+end
+
+"""
+    struct InternalMethodTable <: MethodTableView
```
JeffBezanson, line 30:
```suggestion
    struct CachedMethodTable <: MethodTableView
```

base/compiler/methodtable.jl
Keno: Thanks. I originally had this a singleton struct with the world separately, but I think this makes more sense.

base/compiler/stmtinfo.jl
```diff
@@ -7,8 +7,7 @@ to re-consult the method table. This info is illegal on any statement that is
 not a call to a generic function.
 """
 struct MethodMatchInfo
```
JeffBezanson, line 9: Maybe this type should be combined with `MethodLookupResult`?

base/compiler/stmtinfo.jl
```diff
@@ -7,8 +7,7 @@ to re-consult the method table. This info is illegal on any statement that is
...
-    applicable::Any
-    ambig::Bool
+    results::Union{Missing, MethodLookupResult}
 end
```
JeffBezanson, line 11: The doc string below this says `MethodMatchInfo` again instead of `UnionSplitInfo`.

JeffBezanson (approved):
Ahh, abstraction! :heart_eyes: :smile:

base/compiler/stmtinfo.jl
Keno: I'm not quite sure yet. There's some additional information about method matches I may want to keep in the near future, that I think could go here nicely. My preference would be to keep this as is for now and if it turns out that information would be better somewhere else, we can always refactor this, since it's all internal.

base/compiler/stmtinfo.jl
Keno: Thanks, sorry. Will fix both instances.

PR #36747, opened 2020-07-21
manual: scope effect from using and import

fonsp:


musm:
This table was removed in https://github.com/JuliaLang/julia/pull/38271

PR #36754, opened 2020-07-21
fix #27429, propagate include path in `at-everywhere`

### LINKED ISSUE #27429: include() relative paths are different on worker processes
The behavior of `@everywhere include` seems inconsistent/broken. (I'm using Julia 0.6.3 on OSX)

Suppose you have a repository structure:

    repo/
       src/
           -MyMod.jl
           -script.jl

with “MyMod.jl” containing:

    module MyMod
        export f
        f(x) = 5 * x
    end

and “script.jl” containing:

    @everywhere include("MyMod.jl")
    using MyMod
    
    for p in workers()
        @show remotecall_fetch(f, p, p)
    end

From within `repo/src/` you can run `julia -p 4 script.jl` successfully, but from `repo/` you cannot run Julia with multiple processes (i.e. call `julia -p 4 src/script.jl`), since the worker processes attempt to load `repo/MyMod.jl` which doesn’t exist.

This seems to be because `include` interprets relative paths differently when run on the main process than it does when run on worker processes. On main "relative" is relative to the script/file, whereas on workers it is relative to the current working directory.

(Note the behavior of `include` allows you to run "script.jl" on a single process from anywhere.) 

As a workaround I've been using:

    function include_everywhere(filepath)
        include(filepath) # Load on Node 1 first, triggering any precompile
        if nprocs() > 1
            fullpath = joinpath(@__DIR__, filepath)
            @sync for p in workers()
                @async remotecall_wait(include, p, fullpath)
            end
        end
    end
    
    include_everywhere("MyMod.jl")
    using MyMod

But this feels like it ought to be the default behavior of `@everywhere include`.

JeffBezanson:
This is a bit ad-hoc but I think it's worth it. Definitely a noticeable bump in user-friendliness.

fix #27429

### DIFF
--- a/stdlib/Distributed/src/macros.jl
+++ b/stdlib/Distributed/src/macros.jl
@@ -187,7 +187,11 @@ Instead, local variables can be broadcast using interpolation:
 The optional argument `procs` allows specifying a subset of all
 processes to have execute the expression.
 
-Equivalent to calling `remotecall_eval(Main, procs, expr)`.
+Similar to calling `remotecall_eval(Main, procs, expr)`, but with two extra features:
+
+    - `using` and `import` statements run on the calling process first, to ensure
+      packages are precompiled.
+    - The current source file path used by `include` is propagated to other processes.
 """
 macro everywhere(ex)
     procs = GlobalRef(@__MODULE__, :procs)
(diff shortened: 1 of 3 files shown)


stdlib/Distributed/src/macros.jl
```diff
@@ -198,7 +202,8 @@ macro everywhere(procs, ex)
     imps = extract_imports(ex)
     return quote
         $(isempty(imps) ? nothing : Expr(:toplevel, imps...)) # run imports locally first
-        let ex = $(Expr(:quote, ex)), procs = $(esc(procs))
+        let ex = Expr(:toplevel, :(task_local_storage()[:SOURCE_PATH] = $(get(task_local_storage(), :SOURCE_PATH, nothing))), $(Expr(:quote, ex))),
```
vtjnash, line 205: doesn't this need to nest and restore the old value?

JeffBezanson: We know that `remotecall_eval` will evaluate the expression in a new task.

PR #36759, opened 2020-07-22
Add partially evaluated methods ===(x), !==(x), and isa(T)

### LINKED ISSUE #32018: Single argument `isa` function
It would be nice to have a single argument `isa` for cleaner filtering code.
```jl
filter(x -> !isa(x, AbstractFloat), [1, 2.0])
# could become
filter(!isa(AbstractFloat), [1, 2.0])
```
This would provide some consistency with other functions, such as `isequal`, but since `isa` is a builtin function, defining other variants doesn't seem trivial.
```jl
julia> Core.isa(::Type{T}) where {T} = (x) -> isa(x, T)
ERROR: cannot add methods to a builtin function
```

### LINKED ISSUE #36163: Add curried versions of === and !==
The other day I found myself wishing I could do `===(nothing)` in the same way as we can currently do `==(5)`. This isn't straightforward to add since `===` is a builtin function rather than generic. And I'm not sure if its desirable to make `===` generic since it's not something you really want people to override (related #32844).

tkf:
This PR adds `===(x)`, `!==(x)`, and `isa(T)` that creates a `Fix2` like `isequal` etc.  For `===(x)` and `isa(T)`, it looks like I needed to add tfuncs.

fix #36163
fix #32018

### DIFF
--- a/base/docs/basedocs.jl
+++ b/base/docs/basedocs.jl
@@ -1577,6 +1577,17 @@ Integer
 invoke
 
 """
+    ===(x)
+
+Create a function that compares its argument to `x` using [`===`](@ref), i.e.
+a function equivalent to `y -> y == x`.
+
+!!! compat "Julia 1.6"
+    The partially applied form `===(x)` requires Julia 1.6 or later.
+"""
+===(x)
+
+"""
     isa(x, type) -> Bool
 
 Determine whether `x` is of the given `type`. Can also be used as an infix operator, e.g.
@@ -1600,7 +1611,24 @@ julia> 1 isa Number
 true
 ```
 """
-isa
+isa(x, T)
+
+"""
+    isa(type)
+
+Create a function that compares its argument to `x` using [`isa`](@ref), i.e.
+a function equivalent to `type -> x isa type`.
+
+!!! compat "Julia 1.6"
+    The partially applied form `isa(type)` requires Julia 1.6 or later.
+
+# Examples
+```jldoctest
+julia> isa(Int)(1)
+true
+```
+"""
+isa(T)
 
 """
     DivideError()
--- a/src/builtin_proto.h
+++ b/src/builtin_proto.h
@@ -12,11 +12,11 @@ extern "C" {
 #ifdef DEFINE_BUILTIN_GLOBALS
 #define DECLARE_BUILTIN(name) \
     JL_CALLABLE(jl_f_##name); \
-    jl_value_t *jl_builtin_##name
+    jl_value_t *jl_builtin_##name JL_GLOBALLY_ROOTED
 #else
 #define DECLARE_BUILTIN(name) \
     JL_CALLABLE(jl_f_##name); \
-    extern jl_value_t *jl_builtin_##name
+    extern jl_value_t *jl_builtin_##name JL_GLOBALLY_ROOTED
 #endif
 
 DECLARE_BUILTIN(throw);      DECLARE_BUILTIN(is);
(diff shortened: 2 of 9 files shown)


src/builtin_proto.h
```diff
@@ -12,11 +12,11 @@ extern "C" {
 #ifdef DEFINE_BUILTIN_GLOBALS
 #define DECLARE_BUILTIN(name) \
     JL_CALLABLE(jl_f_##name); \
-    jl_value_t *jl_builtin_##name
+    jl_value_t *jl_builtin_##name JL_GLOBALLY_ROOTED
 #else
 #define DECLARE_BUILTIN(name) \
     JL_CALLABLE(jl_f_##name); \
-    extern jl_value_t *jl_builtin_##name
+    extern jl_value_t *jl_builtin_##name JL_GLOBALLY_ROOTED
```
tkf, line 19: This is a bit of a long shot for working around analyzegc https://build.julialang.org/#/builders/65/builds/1574/steps/4/logs/stdio

But I wonder if there is a reason why this is not used?  Maybe I'm doing something stupid here?

If this is not the right way to do it, maybe I need to do something like this?

```diff
diff --git a/src/builtins.c b/src/builtins.c
index 440b69629e..c2576d553d 100644
--- a/src/builtins.c
+++ b/src/builtins.c
@@ -369,6 +369,7 @@ JL_CALLABLE(jl_f_is)
         jl_function_t *fix2 = jl_get_function(jl_base_module, "Fix2");
         if (fix2 == NULL)
             jl_undefined_var_error(jl_symbol("Fix2"));
+        JL_GC_PROMISE_ROOTED(jl_builtin_is);
         return jl_call2(fix2, jl_builtin_is, args[0]);
     }
     if (args[0] == args[1])
@@ -440,6 +441,7 @@ JL_CALLABLE(jl_f_isa)
         jl_function_t *fix2 = jl_get_function(jl_base_module, "Fix2");
         if (fix2 == NULL)
             jl_undefined_var_error(jl_symbol("Fix2"));
+        JL_GC_PROMISE_ROOTED(jl_builtin_isa);
         return jl_call2(fix2, jl_builtin_isa, args[0]);
     }
     JL_TYPECHK(isa, type, args[1]);
```

base/docs/basedocs.jl
```diff
@@ -1576,6 +1576,17 @@ Integer
 """
 invoke
 
+"""
+    ===(x)
+
+Create a function that compares its argument to `x` using [`===`](@ref), i.e.
+a function equivalent to `y -> y == x`.
```
ericphanson, line 1583: Should be `y -> y === x`

JeffBezanson:
This is tricky. I suspect the only good way to do this is to make `===` generic (also `isa`), since otherwise it will need special inlining and/or codegen support. What I would try is

- Rename the current builtin `===` to something else
- Move Fix2 to Core (boot.jl) and define the two methods of `===` there
- Add a check to prevent adding more methods to `===`
- Inference can still handle `Core.:(===)` specially, since we know what it does
- We might need to still make `===` a `Builtin` by special arrangement?
- Check the impact on the compiler. Might need to e.g. avoid adding backedges to `===` (that would be a natural part of a general "sealed functions" feature).

vtjnash:
> Add a check to prevent adding more methods to ===

```julia
julia> setfield!(typeof(+).name.mt, fieldcount(Core.MethodTable), 0x01)
# this "seals" a method, and can trick it into thinking that `+` is now a subtype of `Core.Builtin`

julia> Base.:(+)(a) = 1
ERROR: cannot add methods to a builtin function
```

cossio:
Something holding this back?

PR #36763, opened 2020-07-22
Fix `pbpaste` error on MacOS buildbots

staticfloat:
We had a typo here for many a year; apparently the older buildbots did
not need `reattach-to-user-namespace` quite as desperately as we thought
they did.


Keno:
Can we just call reattach on the entire buildbot process? It seems a little odd to do that here (of course we should still merge this as soon as it passes, so we can fix CI).

staticfloat:
We can, but note that the primary use of this is for people who run inside of `tmux` on a normal OSX system, not for CI.  This was fixed in tmuxs 2.6+, but that's still new enough that it's nice to automatically fix it for users that have `reattach-to-user-namespace` installed.

PR #36765, opened 2020-07-22
RFC: mark these methods as eligible for the faster cache for novel lookups

vtjnash:
Some observed methods that we may specialize often

### DIFF
--- a/base/dict.jl
+++ b/base/dict.jl
@@ -367,6 +367,7 @@ end
         # > 3/4 deleted or > 2/3 full
         rehash!(h, h.count > 64000 ? h.count*2 : h.count*4)
     end
+    nothing
 end
 
 function setindex!(h::Dict{K,V}, v0, key0) where V where K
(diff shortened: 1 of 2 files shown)


base/dict.jl
```diff
@@ -351,7 +351,7 @@ function ht_keyindex2!(h::Dict{K,V}, key) where V where K
     return ht_keyindex2!(h, key)
 end
 
-@propagate_inbounds function _setindex!(h::Dict, v, key, index)
+@propagate_inbounds @inline function _setindex!(h::Dict, v, key, index)
```
simeonschaub, line 354: Doesn't `@propagate_inbounds` already do inlining?

vtjnash:
```suggestion
@propagate_inbounds function _setindex!(h::Dict, v, key, index)
```
Indeed, that is what the docs say.

base/dict.jl
```diff
@@ -392,6 +393,22 @@ function setindex!(h::Dict{K,V}, v0, key::K) where V where K
     return h
 end
 
+function setindex!(h::Dict{K,Any}, v, key::K) where K
```
oscardssmith, line 396: why not `setindex!(h::Dict{K,Any}, @nospecialize v, key::K)`?

vtjnash: I have options, so I use thehm

PR #36769, opened 2020-07-22
return tuple type manipulations to deprecated.jl

vtjnash:
Some important packages seem to use these.


JeffBezanson:
Ugh, just discovered `parameter_upper_bound` has users as well.

JeffBezanson:
Ugh, just discovered `parameter_upper_bound` has users as well.

vtjnash:
That's annoying. It's always been present here for exporting from Compat.jl, never from here.

PR #36771, opened 2020-07-23
Streamline bail path in inlining

Keno:
When inlining declines to inline something, it instead turns them into
:invoke statements. These are then turned into direct (non-inlined)
calls by codegen or otherwise receive a fast path at runtime. While
inlining has evolved quite a bit, this code has stayed much the same
since it was introduced four years ago and doesn't seem to make much
sense as is. In particular:

1. For the non-`invoke()` case we were doing an extra method look that
seems entirely superfluous, because we already had to do the very same
method lookup just to reach this point. The only thing this path was
doing at that point was creating a "compilable" specialization (which
might use a slightly different signature). We might as well do that
directly.

2. For the invoke case, we were pro-actively adding the specialization
to the `->invokes` dispatch cache. However, this doesn't make much sense
a priori either, because the bail path does not go through the runtime
`invoke()` code that uses that cache (it did many years ago when this
code was introduced, but hasn't in a long time). There does not seem
to be a good reason to believe that this signature will be any more
likely than any other to be invoked using the runtime mechanism.

This cleans up that path by getting rid of both the superfluous method
lookup and the superfluous addition to the `->invokes` cache. There
should be a slight performance improvement as well from avoiding
this superfluous work, but the bail path is less common than one
might expect (the vast majority of call sites are inlined) and in
measurements the effect seems to be in the noise. Nevertheless,
it seems like a nice simplification and is conceptually clearer.

### DIFF
--- a/base/compiler/utilities.jl
+++ b/base/compiler/utilities.jl
@@ -118,8 +118,24 @@ function retrieve_code_info(linfo::MethodInstance)
     end
 end
 
+# Get at the nonfunction_mt, which happens to be the mt of SimpleVector
+const nonfunction_mt = typename(SimpleVector).mt
+
+function get_compileable_sig(method::Method, @nospecialize(atypes), sparams::SimpleVector)
+    isa(atypes, DataType) || return Nothing
+    mt = ccall(:jl_method_table_for, Any, (Any,), atypes)
+    mt === nothing && return nothing
+    return ccall(:jl_normalize_to_compilable_sig, Any, (Any, Any, Any, Any),
+        mt, atypes, sparams, method)
+end
+
 # get a handle to the unique specialization object representing a particular instantiation of a call
-function specialize_method(method::Method, @nospecialize(atypes), sparams::SimpleVector, preexisting::Bool=false)
+function specialize_method(method::Method, @nospecialize(atypes), sparams::SimpleVector, preexisting::Bool=false, compilesig::Bool=false)
+    if compilesig
+        new_atypes = get_compileable_sig(method, atypes, sparams)
+        new_atypes === nothing && return nothing
+        atypes = new_atypes
+    end
     if preexisting
         # check cached specializations
         # for an existing result stored there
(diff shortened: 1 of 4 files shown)


vtjnash:
Many of these queries didn't previously exist, and `jl_compilation_sig` used to give the desired signature, not the actual one. They're now hopefully factored better, such that it should perhaps be possible now to do this.

Keno:
Updated to also drop the extra addition to `->invokes`.

vtjnash:
This is seeming like a good cleanup of cruft

base/compiler/utilities.jl
```diff
@@ -118,8 +118,34 @@ function retrieve_code_info(linfo::MethodInstance)
...
+# Get at the nonfunction_mt, which happens to be the mt of SimpleVector
+const nonfunction_mt = typename(SimpleVector).mt
+
+function get_compileable_sig(method::Method, @nospecialize(atypes), sparams::SimpleVector)
+    isa(atypes, DataType) || return Nothing
+    mt = ccall(:jl_method_table_for, Any, (Any,), atypes)
+    mt === nothing && return nothing
+    nspec = mt === typename(Type).mt || mt === nonfunction_mt ?
+        method.nargs + 1 : mt.max_args + 2
+    newparams = RefValue{SimpleVector}()
```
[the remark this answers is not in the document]
Keno: Ah, you're right, we can just change jl_compilation_sig to return newparams though.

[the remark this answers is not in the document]
Keno: I guess just doing it in C is fine also

src/gf.c
```diff
@@ -567,7 +567,7 @@ static jl_value_t *ml_matches(jl_methtable_t *mt, int offs,
                               size_t *min_valid, size_t *max_valid, int *has_ambiguity);
 
 // get the compilation signature specialization for this method
-static void jl_compilation_sig(
+JL_DLLEXPORT void jl_compilation_sig(
```
vtjnash, line 570:
```suggestion
static void jl_compilation_sig(
```

base/compiler/utilities.jl
```diff
@@ -118,8 +118,24 @@ function retrieve_code_info(linfo::MethodInstance)
...
 
+# Get at the nonfunction_mt, which happens to be the mt of SimpleVector
+const nonfunction_mt = typename(SimpleVector).mt
+
+function get_compileable_sig(method::Method, @nospecialize(atypes), sparams::SimpleVector)
+    isa(atypes, DataType) || return Nothing
+    mt = ccall(:jl_method_table_for, Any, (Any,), atypes)
+    mt === nothing && return nothing
+    return ccall(:jl_normalize_to_compilable_sig, Any, (Any, Any, Any, Any),
+        mt, atypes, sparams, method)::Union{Nothing, Type{<:Tuple}}
```
vtjnash, line 129: This is a pretty expensive looking type-assert, since it also conveys no useful information

Keno: Fair enough. I'll drop it.

PR #36776, opened 2020-07-23
update MPFR to v4.1.0

### LINKED ISSUE #35796: Compile MPFR and GMP to be thread safe
This issue stems from [issue #10040](https://github.com/JuliaLang/julia/issues/10040#issuecomment-625551668).

MPFR has a configuration option `--enable-thread-safe` that we should use by default.

MPFR has a configuration option`--enable-gmp-internals` that we should use iff library versioning works (or is irrelevant) for our binaries:

> --enable-gmp-internals  allows the MPFR build to use GMP's undocumented
functions (not from the public API). Note that library versioning is not guaranteed to work if this option is used. Thus it must not be used in binary distributions.

vtjnash:
Includes build fixes to ensure it is built thread-safe. Fixes #35796 regression.


JeffBezanson:
Backport?

vtjnash:
Seems like AArch64 is having problems computing pi with this update. I might need someone with access to a machine to investigate.

JeffBezanson:
Can we try 4.0.2 with threading enabled?

vtjnash:
I assume I just need to rerun the Yggdrasil builder in a branch

vtjnash:
Looking deeply into this, seems like something it going very badly. I see we're crashing on this copy/zero:
```
    if (MPFR_UNLIKELY (_destprec >= _srcprec))                              \
      {                                                                     \
        _srcs  = MPFR_PREC2LIMBS (_srcprec);                                \
        _dests = MPFR_PREC2LIMBS (_destprec) - _srcs;                       \
        MPN_COPY (_destp + _dests, srcp, _srcs);                            \
        MPN_ZERO (_destp, _dests);                                          \
        inexact = 0;                                                        \
      }                                                                     \
```

Where we know that `_destprec == 32` and `_srcprec == 32 + 63`!

getting called from
`MPFR_RNDRAW (inexact, a, tmp, bq + cq, rnd_mode, sign, ax2++); (mul.c:1086)`
getting called from
`mpfr_mul (bound_emin, bound_emin, e, MPFR_RNDD); (exp.c:119)`

vtjnash:
I'm almost certain this is a gcc (4.8.5) bug happening on some new code in mpfr-4.1.0 (prior code didn't have the cast to int)
```
   1084   /* b1 is 0 or 1 (most significant bit from the raw product) */
   1085   ax2 = ax + ((int) b1 - 1);
   1086   MPFR_RNDRAW (inexact, a, tmp, bq + cq, rnd_mode, sign, ax2++);
```
Where the round-down macro contains the `if (_destprec >= _srcprec) { bad } else { good }` I mention above.

The assembly though appears to have re-written that into:
```
if (b1 == 0) {
  ax2 = ax - 1;
  bad;
}
else {
  ax2 = ax + 0;
  good;
}
```

So either staying on 4.0.2 with the new flags or using a less ancient version of gcc (June 23, 2015) should resolve that.

giordano:
Is gcc v5 less ancient enough?

vtjnash:
I don't have gcc 4 installed to compare, but gcc 5 seems to be generating sensible code for me.

PR #36778, opened 2020-07-23
remove restriction on number of threads

### LINKED ISSUE #36469: Warn user about number of threads when greater than logical cores
Should a user be warned when the number of threads specified with `JULIA_NUM_THREADS` is greater than the number of logical cores? In this case Julia ignores the number provided by the user and uses the latter.

On a Intel Core i9 for example, the variable
```
JULIA_NUM_THREADS=16
```

is ignored and Julia is started with 8 threads.

vtjnash:
Closes #36469

### DIFF
--- a/src/threading.c
+++ b/src/threading.c
@@ -407,16 +407,13 @@ void jl_init_threading(void)
 #endif
 
     // how many threads available, usable
-    int max_threads = jl_cpu_threads();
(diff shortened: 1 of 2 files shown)


JeffBezanson:
Triage approves. But there should be an error if the number of threads is > cpu threads and JULIA_EXCLUSIVE is set.

PR #36780, opened 2020-07-23
view macro hygiene + comment

simonbyrne:
A comment on Slack asked why `@view(X[a])` expands to `true && view(X,a)`, which is not at all obvious, so added a comment. Also fixed a macro hygiene issue.

### DIFF
--- a/base/views.jl
+++ b/base/views.jl
@@ -113,11 +113,13 @@ macro view(ex)
     if Meta.isexpr(ex, :ref)
         ex = replace_ref_begin_end!(ex)
         if Meta.isexpr(ex, :ref)
-            ex = Expr(:call, view, ex.args...)
+            ex = Expr(:call, Base.view, ex.args...)
         else # ex replaced by let ...; foo[...]; end
             @assert Meta.isexpr(ex, :let) && Meta.isexpr(ex.args[2], :ref)
-            ex.args[2] = Expr(:call, view, ex.args[2].args...)
+            ex.args[2] = Expr(:call, Base.view, ex.args[2].args...)
         end
+        # prevent accidentally redefining `view` if used on the LHS, i.e.
+        #   @view(A[x]) = 2
         Expr(:&&, true, esc(ex))
     else
         throw(ArgumentError("Invalid use of @view macro: argument must be a reference expression A[...]."))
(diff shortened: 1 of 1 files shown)


Keno:
Would `let; view(X,a); end` be a more obvious expansion that has the same effect. It's a minor difference, but `&&` does introduce an extra basic block.

simonbyrne:
Yes, that would do the same thing. I can update the PR

Keno:
@JeffBezanson may have an opinion, but that sounds better to me.

base/views.jl
```diff
@@ -113,11 +113,13 @@ macro view(ex)
     if Meta.isexpr(ex, :ref)
         ex = replace_ref_begin_end!(ex)
         if Meta.isexpr(ex, :ref)
-            ex = Expr(:call, view, ex.args...)
+            ex = Expr(:call, Base.view, ex.args...)
```
mbauman, line 116: This may be intentional — and I wouldn't be surprised if folks somewhere are depending upon this.

Keno: It actually doesn't make a difference because it's putting in the `view` function as a value. You may probably either want `:view`, or :(Base.view) or  `GlobalRef(Base, :view)`.

simonbyrne: Is there a problem with leaving it as a value?

Keno: The AST probably serializes better without the explicit object, but it doesn't really matter.

mbauman: Oh duh, right.  I really just figured that SGJ wouldn't get this wrong.  Hah!

base/views.jl
```diff
@@ -112,13 +112,14 @@ julia> A
...
-            @assert Meta.isexpr(ex, :let) && Meta.isexpr(ex.args[2], :ref)
-            ex.args[2] = Expr(:call, view, ex.args[2].args...)
+        if !Meta.isexpr(ex, :let)
+            # wrap in a let block to prevent accidentally redefining `view` if used on the LHS, e.g.
+            #   @view(A[x]) = 2
+            ex = Expr(:let, Expr(:block), ex)
         end
-        Expr(:&&, true, esc(ex))
+        @assert Meta.isexpr(ex.args[2], :ref)
+        ex.args[2] = :(Base.view($(ex.args[2].args...))
```
mbauman, line 121: Ok, now this won't work.  We can actually shadow `Base` in local scopes:

```Julia
julia> let Base = 1
           @view (1:2)[:]
       end
ERROR: type Int64 has no field view
```

Also missing a parens at the end of the line)

Keno: It'll work if you only `esc` the arguments, not the whole expression.

simonbyrne: yeah, i was trying to avoid that since i didn't understand what the `let` block was doing.

PR #36785, opened 2020-07-24
Clean up atomic access to sleep_check_state

### LINKED ISSUE #36699: Deadlock  from simple repeated `@spawn` and `wait`
Simply repeating `@spawn` and `wait` causes a dead lock:

```julia
julia> for _ in 1:1000
           for _ in 1:1000
               ts = [(Threads.@spawn nothing) for _ in 1:Threads.nthreads()]
               foreach(wait, ts)
           end
           print('.')
           flush(stdout)
       end
............................................^CERROR: InterruptException:
Stacktrace:
 [1] poptask(W::Base.InvasiveLinkedListSynchronized{Task})
   @ Base ./task.jl:706
 [2] wait
   @ ./task.jl:714 [inlined]
 [3] wait(c::Base.GenericCondition{Base.Threads.SpinLock})
   @ Base ./condition.jl:106
 [4] _wait(t::Task)
   @ Base ./task.jl:240
 [5] wait
   @ ./task.jl:267 [inlined]
 [6] foreach(f::typeof(wait), itr::Vector{Task})
   @ Base ./abstractarray.jl:2072
 [7] top-level scope
   @ REPL[4]:4

To edit a specific method, type the corresponding number into the REPL and press Ctrl+Q

julia> versioninfo()
Julia Version 1.6.0-DEV.477
Commit b07594dac0 (2020-07-16 18:42 UTC)
Platform Info:
  CPU: Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz
  WORD_SIZE: 64
  LIBM: libopenlibm

julia> Threads.nthreads()
4
```

In the above session, I saw '.'s printed pretty rapidly until it hangs.  Then I terminated the evaluation with Ctrl-C.

I see this in 1.5.0-rc1.0 but not in 1.4.

(I tried `JULIA_RR_RECORD_ARGS='--chaos --num-cores=4' JULIA_NUM_THREADS=4 julia --startup-file=no --bug-report=rr` etc. to get a deadlock in `rr`. But so far I couldn't get it under `rr`.)

Keno:
Fixes #36699 I belive. Here's my best understanding of what's happening.
The relevant operations are:

Thread 1:
```
push!(W, task) # assign the task to the waitqueue, implicitly a non-atomic assign to .head of the queue protected by a lock
# Some time later
if (atomic_load(sleep_state) == sleeping) {
   if (atomic_exchange(...)) {
     wake
   }
}
```

Thread 2:
```
if isempty(W) {
   atomic_store(sleep_state)
   !isempty(W) && continue()
   sleep() 
}
```

Now, x86 has a very strong memory model, but one thing in particular that is allowed is reordering loads before previous stores to another memory location. In particular, it is legal for the `atomic_load()` of sleep state to be speculated before the addition of `task` to `W`. As a result, it is possible to have the following interleaving:

```
T1:
r = atomic_load(sleep_state) # r == not_sleeping

T2: 
   atomic_store(sleep_state)
   isempty(W) # true
   sleep()

T1:
push!(W, task) # assign the task to the waitqueue, implicitly a non-atomic assign to .head of the queue protected by a lock
# Some time later
if (r == sleeping #= true =#) {

}
```

Thus `T1` never wakes up `T2` the observed deadlock results.

This commit fixes that by dropping the `atomic_load` and using and atomic exchange instead, which may not be re-ordered in the same way. A different fix would be to place an explicit ReadWrite barrier before attempting to wake up the threads.
That said, I think we need to take a close look at this code in general with respect to atomic correctness. I suspect there are more demons.

### DIFF
--- a/src/partr.c
+++ b/src/partr.c
@@ -330,13 +330,10 @@ static int sleep_check_after_threshold(uint64_t *start_cycles)
 static void wake_thread(int16_t tid)
 {
     jl_ptls_t other = jl_all_tls_states[tid];
-    if (jl_atomic_load(&other->sleep_check_state) != not_sleeping) {
-        int16_t state = jl_atomic_exchange(&other->sleep_check_state, not_sleeping); // prohibit it from sleeping
-        if (state == sleeping) { // see if it was possibly sleeping before now
-            uv_mutex_lock(&other->sleep_lock);
-            uv_cond_signal(&other->wake_signal);
-            uv_mutex_unlock(&other->sleep_lock);
-        }
+    if (jl_atomic_bool_compare_exchange(&other->sleep_check_state, sleeping, not_sleeping)) {
+        uv_mutex_lock(&other->sleep_lock);
+        uv_cond_signal(&other->wake_signal);
+        uv_mutex_unlock(&other->sleep_lock);
     }
 }
 
@@ -358,7 +355,7 @@ JL_DLLEXPORT void jl_wakeup_thread(int16_t tid)
     JULIA_DEBUG_SLEEPWAKE( wakeup_enter = cycleclock() );
     if (tid == self || tid == -1) {
         // we're already awake, but make sure we'll exit uv_run
-        if (ptls->sleep_check_state != not_sleeping)
+        if (jl_atomic_load_relaxed(&ptls->sleep_check_state) == sleeping)
             jl_atomic_store(&ptls->sleep_check_state, not_sleeping);
         if (uvlock == system_self)
             uv_stop(jl_global_event_loop());
@@ -405,7 +402,10 @@ static jl_task_t *get_next_task(jl_value_t *trypoptask, jl_value_t *q)
 
 static int may_sleep(jl_ptls_t ptls)
 {
-    return ptls->sleep_check_state == sleeping;
+    // sleep_check_state is only transitioned from not_sleeping to sleeping
+    // by the thread itself. As a result, if this returns false, it will
+    // continue returning false. If it returns true, there are no guarantees.
+    return jl_atomic_load_relaxed(&ptls->sleep_check_state) == sleeping;
 }
 
 extern volatile unsigned _threadedregion;
@@ -476,8 +476,7 @@ JL_DLLEXPORT jl_task_t *jl_task_get_next(jl_value_t *trypoptask, jl_value_t *q)
                     JL_UV_UNLOCK();
                     // optimization: check again first if we may have work to do
                     if (!may_sleep(ptls)) {
-                        if (ptls->sleep_check_state != not_sleeping)
-                            jl_atomic_store(&ptls->sleep_check_state, not_sleeping); // let other threads know they don't need to wake us
+                        assert(ptls->sleep_check_state == not_sleeping);
                         start_cycles = 0;
                         continue;
                     }
@@ -495,7 +494,7 @@ JL_DLLEXPORT jl_task_t *jl_task_get_next(jl_value_t *trypoptask, jl_value_t *q)
                 if (!_threadedregion && active && ptls->tid == 0) {
                     // thread 0 is the only thread permitted to run the event loop
                     // so it needs to stay alive
-                    if (ptls->sleep_check_state != not_sleeping)
+                    if (jl_atomic_load_relaxed(&ptls->sleep_check_state) != not_sleeping)
                         jl_atomic_store(&ptls->sleep_check_state, not_sleeping); // let other threads know they don't need to wake us
                     start_cycles = 0;
                     continue;
@@ -510,8 +509,7 @@ JL_DLLEXPORT jl_task_t *jl_task_get_next(jl_value_t *trypoptask, jl_value_t *q)
                 uv_cond_wait(&ptls->wake_signal, &ptls->sleep_lock);
                 // TODO: help with gc work here, if applicable
             }
-            if (ptls->sleep_check_state != not_sleeping)
-                jl_atomic_store(&ptls->sleep_check_state, not_sleeping); // let other threads know they don't need to wake us
+            assert(ptls->sleep_check_state == not_sleeping);
             uv_mutex_unlock(&ptls->sleep_lock);
             JULIA_DEBUG_SLEEPWAKE( ptls->sleep_leave = cycleclock() );
             jl_gc_safe_leave(ptls, gc_state); // contains jl_gc_safepoint


tkf (approved):
Thanks! This fixes the problem for me. (I tried 10x longer version of #36699 a few times. No deadlock so far.)

vtjnash:
So the issue is just that may_sleep is getting hoisted (by the compiler, I assume)?

src/partr.c
```diff
@@ -476,8 +476,7 @@ JL_DLLEXPORT jl_task_t *jl_task_get_next(jl_value_t *trypoptask, jl_value_t *q)
                     JL_UV_UNLOCK();
                     // optimization: check again first if we may have work to do
                     if (!may_sleep(ptls)) {
-                        if (ptls->sleep_check_state != not_sleeping)
-                            jl_atomic_store(&ptls->sleep_check_state, not_sleeping); // let other threads know they don't need to wake us
+                        assert(ptls->sleep_check_state == not_sleeping);
```
c42f, line 479: Should we just make every access of `sleep_check_state` atomic as a matter of course?

```suggestion
                        assert(jl_atomic_load_relaxed(&ptls->sleep_check_state) == not_sleeping);
```

src/partr.c
```diff
@@ -405,7 +402,10 @@ static jl_task_t *get_next_task(jl_value_t *trypoptask, jl_value_t *q)
 
 static int may_sleep(jl_ptls_t ptls)
 {
-    return ptls->sleep_check_state == sleeping;
+    // sleep_check_state is only transitioned from not_sleeping to sleeping
+    // by the thread itself. As a result, if this returns false, it will
+    // continue returning false. If it returns true, there are no guarantees.
+    return jl_atomic_load_relaxed(&ptls->sleep_check_state) == sleeping;
```
c42f, line 408: Do we need something stronger than relaxed ordering here, because `sleep_check_state` is meant to ensure consistency of other loads/stores?

src/partr.c
```diff
@@ -358,7 +355,7 @@ JL_DLLEXPORT void jl_wakeup_thread(int16_t tid)
     JULIA_DEBUG_SLEEPWAKE( wakeup_enter = cycleclock() );
     if (tid == self || tid == -1) {
         // we're already awake, but make sure we'll exit uv_run
-        if (ptls->sleep_check_state != not_sleeping)
+        if (jl_atomic_load_relaxed(&ptls->sleep_check_state) == sleeping)
             jl_atomic_store(&ptls->sleep_check_state, not_sleeping);
```
c42f, line 359: So this pair of operations looks like a CAS, but done non-atomically. Can that be a problem?

c42f (commented):
I spent some time staring at this in confusion :-)

At a surface level I can see is that there's various worrying-looking patterns going on here in terms of threads loading from `sleep_check_state` without using atomic operations (not even relaxed loads) and also using multiple atomics to act on that state non atomically.

I noticed that there's some loads of `sleep_check_state` in this file that this patch doesn't cover: line 436 and 441 — they should be relaxed loads at least? Also if `sleep_check_state` is used to protect other state, do we need to be using acquire loads?

For the state transition `not_sleeping <--> sleeping`, this is done in various places, and as a pair of atomics rather than a CAS which might be fine but seems non-obvious. I think it may be clearer to factor the transition into a function in the spirit of `may_sleep`.

As to the exact cause of the deadlock I sure didn't see it by just eyeballing this diff! I look forward to the explanation.

JeffBezanson:
Comparing to 1.4, the test case in the issue (if you put it inside a function) goes from 3.45 seconds to 3.0 seconds, so that's also excellent :tada: (4 threads in my case)

src/partr.c
Keno: It's not a data race, because the only other writer is monotonic (makes the same change). In fact, we can probably relax the atomic store here.

src/partr.c
Keno: The actual condition variable is protected by a lock, so we don't depend on any memory access dependencies here, I don't think. That's why I picked relaxed. Do you disagree?

Keno:
> At a surface level I can see is that there's various worrying-looking patterns going on here in terms of threads loading from `sleep_check_state` without using atomic operations

Yes, I think this code could use a bit of a refactor. However, I would like to do that more systematically and also make use of tsan support to check these things under the test suite. This PR is just here to get the bug fixed.

src/partr.c
c42f: Right I think that makes sense. I wondered whether the store could be less strict.

Keno: Upon further reflection the atomic does guard the memory contents of the work queue, so we might want something stronger. I need to think about it.

src/partr.c
c42f: > Do you disagree?

I don't disagree, I'm just unsure.

The condition variable seems fine. But I noticed that `may_sleep` is also used in a branch related to the event loop which is protected with `jl_uv_mutex`.  So `may_sleep` doesn't always go along with the same lock. Which seems potentially confusing but perhaps harmless.

c42f:
> This PR is just here to get the bug fixed.

Fine by me, this is my first look into this code.

JeffBezanson:
We should add the test case from the issue, just to check that it doesn't hang.

Keno:
I've updated the description with my root cause analysis of this issue.

Keno:
> We should add the test case from the issue, just to check that it doesn't hang.

I think you're inviting hangs on non-x86 platforms that have an even weaker memory model. I would certainly recommend against backporting such a test.

src/partr.c
vtjnash: Yeah, this not reads like an CAS, but it's not. I'd suggest writing this as `!= sleeping`, so it's clearer that it's just a regular (relaxed) store.

src/partr.c
vtjnash: I agree. The change to use `assert` also makes sense (since that's currently the implementation of `may_sleep`), but assumes that we don't alter the implementation of `may_sleep` to possibly have disjoint conditions that we want to check (i.e. the original partr code had a global `sleep_check_state`, so that bulk operations were also attempted).

src/partr.c
c42f: I prefer the version which avoids the double negative. To be more explanatory it would be helpful to factor the uses of `sleep_check_state` into a minimum of locations and expand the comments describing which state/invariant it's meant to protect and how this implies the atomic ops which are used.

PR #36793, opened 2020-07-24
normalize lowering of `-->`

JeffBezanson:


JeffBezanson:
Ok, this is a version that leaves parsing alone for compatibility with existing macros, but adds lowering so you can use the operator normally.

vtjnash (approved):

PR #36794, opened 2020-07-24
add another deprecated internal function for backwards compat

JeffBezanson:
ColorTypes uses this, so it's needed to allow packages to work on master.

### DIFF
--- a/base/deprecated.jl
+++ b/base/deprecated.jl
@@ -229,5 +229,9 @@ function tuple_type_cons(::Type{S}, ::Type{T}) where T<:Tuple where S
     @_pure_meta
     Tuple{S, T.parameters...}
 end
+function parameter_upper_bound(t::UnionAll, idx)
+    @_pure_meta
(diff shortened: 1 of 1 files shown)


timholy:
xref https://github.com/JuliaGraphics/ColorTypes.jl/pull/209

JeffBezanson:
Great! If that turns out to be the only user we can just close this.

fredrikekre:
Some (older) versions of Compat uses this as well, so maybe we can merge this after all? Failure e.g. here: https://travis-ci.org/github/JuliaLang/Pkg.jl/jobs/716196188#L203

JeffBezanson:
This is used by more packages, so reopening.

fredrikekre:
I ended up fixing it for Compat (https://github.com/JuliaLang/Compat.jl/pull/717) but maybe there were more packages?

KristofferC:
It's also about packages that are stuck on older versions.

PR #36795, opened 2020-07-24
fix #36783, more specialization on called Type arguments

### LINKED ISSUE #36783: inference failure for Float64 constructor?
[This discrepancy](https://discourse.julialang.org/t/float64-is-typecasted-to-float16/43441/9?u=stevengj) doesn't seem right:
```jl
julia> x = rand(Float16, 10_000);

julia> @btime sum($Float64, $x);
  776.911 μs (29999 allocations: 468.73 KiB)

julia> @btime sum($(y -> Float64(y)), $x);
  42.417 μs (0 allocations: 0 bytes)
```

JeffBezanson:
fix #36783

### DIFF
--- a/src/gf.c
+++ b/src/gf.c
@@ -668,8 +668,12 @@ static void jl_compilation_sig(
                   also matching all other TypeConstructors. This means neither
                   Type{TC} nor TypeConstructor is more specific.
                 */
-                if (!*newparams) *newparams = jl_svec_copy(tt->parameters);
-                jl_svecset(*newparams, i, jl_typetype_type);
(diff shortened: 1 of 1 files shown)


src/gf.c
```diff
@@ -874,14 +878,15 @@ JL_DLLEXPORT int jl_isa_compileable_sig(
         }
```
vtjnash, line 878: I vague think the previous conditional (`jl_is_kind(elt)`) should `return 0` if `iscalled`, but there's a TODO there that it's possibly missing other cases already

vtjnash (approved):

PR #36805, opened 2020-07-26
aotcompile: Merge two continue's into one

zlatanvasovic:
A minor improvement.


KristofferC:
I personally don't really see this as an improvement. Why do you think this is better?

StefanKarpinski:
Agree, I think the current version is clearer. They're different cases that both happen to continue, but they're logically distinct so might as well leave them separate.

zdroid:
Well as I said, it's a minor improvement. It reduces the code duplication and maybe removes an instruction or two (but more likely it ends up compiling to the same code).

KristofferC:
Again, I think it would still be good to explain why this is an improvement at all. And yes, there is no performance effect of this.

Keno:
Yeah, this is contrary to most style guides which explicitly recommend to reduce the condition depth, unless the conditions are very closely linked.

zdroid:
@Keno I agree with that one, so I'm closing this. Thanks for the remarks.

PR #36806, opened 2020-07-26
CONTRIBUTING: mark C code properly

zlatanvasovic:
This makes the section way more readable.


StefanKarpinski (approved):
Looks good! Thanks for the formatting improvement. That was pretty unreadable before.

PR #36808, opened 2020-07-26
change import to using in Pkg error message

simeonschaub:
We generally recommend `using` over `import`, so I don't see any reason to suggest `import` here.

### DIFF
--- a/base/loading.jl
+++ b/base/loading.jl
@@ -892,7 +892,7 @@ function require(into::Module, mod::Symbol)
         if where.uuid === nothing
             throw(ArgumentError("""
                 Package $mod not found in current path:
-                - Run `import Pkg; Pkg.add($(repr(String(mod))))` to install the $mod package.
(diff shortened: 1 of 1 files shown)


KristofferC:
> We generally recommend using over import

I would argue the complete opposite.

simeonschaub:
The one in the manual doesn't give any recommendations here, but both YASGuide and BlueStyle recommend strongly againt using `import`.

KristofferC:
AFAIU, that is only for methods to not accidentally extend them. They also say to explicitly list the symbols used so the equivalent here would be `using Pkg: Pkg` but this is equivalent to `import Pkg`.

jonas-schulze:
It seems that Pkg doesn't export `add`, so `using Pkg` followed by the fully qualified `Pkg.add` does not look logical to me.

simeonschaub:
Ah, you're right. I somehow was under the impression that `import Pkg` also imported functions for extension. I guess `import` is more clear here then.

PR #36811, opened 2020-07-26
make Task `state` field more efficient

JeffBezanson:
This doesn't do much yet because of GC size classes, but it's silly to use 8 bytes for a 3-state flag, and we'll be able to remove a couple more fields later.

### DIFF
--- a/base/task.jl
+++ b/base/task.jl
@@ -121,6 +121,30 @@ Get the currently running [`Task`](@ref).
 """
 current_task() = ccall(:jl_get_current_task, Ref{Task}, ())
 
+# task states
+
+const task_state_runnable = UInt8(0)
+const task_state_done     = UInt8(1)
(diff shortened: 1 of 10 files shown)


base/task.jl
```diff
@@ -121,6 +121,29 @@ Get the currently running [`Task`](@ref).
...
 current_task() = ccall(:jl_get_current_task, Ref{Task}, ())
 
+# task states
+
+const task_state_runnable = UInt8(0)
+const task_state_done     = UInt8(1)
+const task_state_failed   = UInt8(2)
+
+@inline function getproperty(t::Task, field::Symbol)
+    if field === :state
```
vtjnash, line 131: Is the intent to add a depwarn here in v2? If so, perhaps a comment would be useful?

vtjnash (approved):

PR #36812, opened 2020-07-26
julia-config: avoid code repetition in main()

zlatanvasovic:


yuyichao:
Note that fewer line does not equal to better or clearer code. Also, there was no code duplication here. Not in a way more sigificant than having the same letter appearing in the same word more than ones.

zdroid:
@yuyichao I agree with your statement, but why do you consider the previous code without duplication?

Also note that
```julia
for flag in (:ldflags, :cflags, :ldlibs, :allflags)
  if args == "--$flag"
```
can be merged into one check (if args in string list), with some modifications. Is that something you're looking for?

yuyichao:
> but why do you consider the previous code without duplication?

Where **is** the duplication? A single word don't count. Even for just repeated word, the only ones I can see are,

1. `ldflags`, ... appearing twice.

    And the twice they each appear has absolutely nothing to do with each other. Because of that, the way you use to force them together uses bad coading style, it doesn't work (try run it) and is in general slower.

2. `println` and `doframework`

   These are more legitimacy duplication, as in the appearance of the same word actually mean the same thing. The way to remove such duplication is to have shared state. But again, it's a single thing on the input and output and here's basically no way you can reduce that duplication in any clearer than it already is.

> can be merged into one check (if args in string list), with some modifications. Is that something you're looking for?

No.

yuyichao:
Also note that you are making changes for the sole purpose of reducing repetition, without checking if the code works, has better performance, or easier to read/understand. But removing code duplication is **never** the goal. It's a heuristic that is sometimes good but in the end it's always only the tool to acheive somethine else (clearity, performance, etc.) and is not itself something you should aim for.

StefanKarpinski:
FWIW, I’m fine with this change and thing it’s slightly clearer although the need for eval gives me pause.

JeffBezanson:
This change isn't *bad*, but I also think it's pretty unnecessary. This structure is also less flexible if one of the arguments needs some different handling at some point in the future.

yuyichao:
The change **IS** bad in that it has scope issues due to eval so it doesn't actually work. i.e.

> Because of that, the way you use to force them together uses bad coading style, it doesn't work (try run it) and is in general slower.

zdroid:
I just proposed this as an idea, which obviously can be made more perfomant and clear (as I noted in the previous comment), but fine, I'll close this.

@yuyichao Repetition of the same code syntax is what people call code repetition, but alright, if you don't consider it such. To add another flag "--xflags", you could put it into the same logical flow by just adding (`:xflags`) to the tuple, instead of repeating the same syntax. Or if you wanted to change the printing way from `println()` to something else, you could just do that. I don't see how the current approach is flexible at all.

@StefanKarpinski What would be the best way to do that?

StefanKarpinski:
It seems like people aren't too into this compactification of the code, so this is moot, but you can avoid `eval` by just iterating over the functions instead of the symbols for the functions:
```jl
for flag in (ldflags, cflags, ldlibs, allflags)
    if args == "--$flag"
        println(flag(doframework))
        break
    end
end
```
This relies on the fact that interpolating a function into a string just produces the function's name, which matches the option name.

zdroid:
@StefanKarpinski Makes sense. I was thinking this way instead:
```julia
for args in ARGS
    args = lstrip(args, '-')
    if args in ("ldflags", "cflags", "ldlibs", "allflags")
        println(eval(Symbol(args))(doframework))
    end
end
```

StefanKarpinski:
That's a bit better but it still uses `eval` which can be avoided. Using `getfield` for this:
```jl
julia> foo = 123
123

julia> getfield(@__MODULE__, Symbol("foo"))
123
```
So the way to write this would be:
```jl
for arg in ARGS
    arg = lstrip(arg, '-')
    if arg in ("ldflags", "cflags", "ldlibs", "allflags")
        flags = getfield(@__MODULE__, Symbol(arg))
        println(flags(doframework))
    end
end
```
But again, I think that iterating over the function objects is clearer and simpler.

zdroid:
@StefanKarpinski I tried both `getfield` and iterating over the function objects, but both end up using 2 kB of memory compared to <500 bytes for original `if-elseif-elseif-...`, at the same performance.

KristofferC:
FWIW, this code is not performance critical.

StefanKarpinski:
Yeah, these will completely tank the compilers ability to understand what functions are called, so a lot more things need to be dispatched dynamically and heap allocated rather than just called and returned. But as @KristofferC said, the performance of this code doesn't matter at all.

bramtayl:
What about using something like `Dict("--ldflags" => ldflags)`? Definitely not performant but could make things prettier

PR #36814, opened 2020-07-26
README: turn directory structure into a table

zlatanvasovic:
More readable and properly displayed on mobile devices. (You can check this out in the rich diff provided by GitHub.)

Since GitHub tables must have a header, I've chosen "Directory" and "Contents" as appropriate descriptions. If you have something better in mind, edits are welcome.


zdroid:
Does this look good?

PR #36818, opened 2020-07-27
Fix C compiler warning in jl_get_safe_restore()

fingolfin:
This was accidentally introduced in PR #36064.


fingolfin:
@c42f sorry to bother you, but would you mind merging this (seeing as you are familiar with the PR in which I accidentally introduced this C compiler warning)

c42f:
Yes indeed, thanks for the fix.

PR #36819, opened 2020-07-27
Fix compact float printing when output contains exactly 6 digits

quinnj:
This was really just a regression when switching from grisu -> ryu
algorithm for float printing. The fix is just putting [@vtjnash's
code](https://github.com/JuliaLang/julia/commit/5d2a0ec06761ea2b445909757d7744b9f7a93072#diff-22819a3d47074bbe6530dfef9c94969cR75)
for the grisu fix into the ryu branch.

Can perhaps still be included in 1.5?

### DIFF
--- a/base/ryu/shortest.jl
+++ b/base/ryu/shortest.jl
@@ -330,7 +330,8 @@ end
     olength = decimallength(output)
     exp_form = true
     pt = nexp + olength
-    if -4 < pt <= (precision == -1 ? (T == Float16 ? 3 : 6) : precision)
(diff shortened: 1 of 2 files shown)


andreasnoack:
@quinnj What is the reason for switching to scientific notation when the first digit is non-zero?
```julia
julia> sprint(show, 666666.049, context=:compact => true)
"666666.0"

julia> sprint(show, 666666.05, context=:compact => true)
"6.66666e5"

julia> sprint(show, 666667.049, context=:compact => true)
"666667.0"
```

PR #36821, opened 2020-07-27
Fix function call to rec_backtrace_ctx when using USE_SYSTEM_LIBUNWIND=1

rbvermaa:
When building julia-1.5.0-rc1 from source, we ran into the following error. It looks like one call to `rec_backtrace_ctx` was not changed when an extra arg was added. 

Not 100% this patch is correct (whether to pass 1 for the lockless argument), but it does fix the compilation error, and is similar to the other branch of the `ifndef`.

```
In file included from /private/tmp/nix-build-julia-1.5.0-rc1-patched.drv-0/source/src/signal-handling.c:117:
In file included from ./signals-unix.c:218:
./signals-mach.c:462:135: error: too few arguments to function call, expected 5, have 4
                bt_size_cur += rec_backtrace_ctx((jl_bt_element_t*)bt_data_prof + bt_size_cur, bt_size_max - bt_size_cur - 1, uc, NULL);
                               ~~~~~~~~~~~~~~~~~                                                                                      ^
./julia_internal.h:797:1: note: 'rec_backtrace_ctx' declared here
size_t rec_backtrace_ctx(jl_bt_element_t *bt_data, size_t maxsize, bt_context_t *ctx,
^
1 error generated.
make[1]: *** [Makefile:161: signal-handling.o] Error 1
```

### DIFF
--- a/src/signals-mach.c
+++ b/src/signals-mach.c
@@ -501,7 +501,7 @@ void *mach_profile_listener(void *arg)
 
                 forceDwarf = -2;
 #else
-                bt_size_cur += rec_backtrace_ctx((jl_bt_element_t*)bt_data_prof + bt_size_cur, bt_size_max - bt_size_cur - 1, uc, NULL);
+                bt_size_cur += rec_backtrace_ctx((jl_bt_element_t*)bt_data_prof + bt_size_cur, bt_size_max - bt_size_cur - 1, uc, NULL, 1);
 #endif
 
                 // Mark the end of this block with 0


NHDaly:
Note: this also needs 1.5 backport.

Sacha0:
It looks like this patch is no longer correct after https://github.com/JuliaLang/julia/pull/37002. Whether the issue persists I do not know, but will try to find out :).

Sacha0:
This issue has been obviated on master. Perhaps close? :)

NHDaly:
@Sacha0 would it still make sense to backport this to 1.5 though?

Sacha0:
I'm operating under the assumption that with 1.6 inbound, there will be no more 1.5 releases. But if there are additional 1.5 release, agreed! :)

PR #36823, opened 2020-07-27
Add jl_active_task_stack to gcext API

fingolfin:
This is meant as a successor for `jl_task_stack_buffer`, which this patch marks as deprecated in a comment.
See also the discussion in PR #36064 for more information -- note that this PR does *not* (yet) attempt the more refined tracking of task stack sizes (in particular, there is no attempt to determine better stack bounds for tasks which are active on other threads). However, this PR should already allow us to simplify some code in GAP and in particular allow that code to not access the `copy_stacks` member of tasks. <s>This is somewhat urgent now for us, as https://github.com/JuliaLang/julia/pull/36802 is about to change the layout of `jl_task_t`.</s> Instead of adding an accessor function for `copy_tasks`, it seems better to just implement a first version of `jl_active_task_stack` now, and refine it later (the API would stay the same, just the returned bounds would be tighter).

@rbehrends please scrutinize this closely, I hacked this together rather quickly and would be surprised if I didn't screw up something somewhere ;-). 

I also have *not* yet tested this PR with GAP, so until that and until @rbehrends went over it, the Julia team should not feel obliged to waste time looking at this (that said, any feedback is of course highly welcome).

### DIFF
--- a/src/julia_gcext.h
+++ b/src/julia_gcext.h
@@ -125,7 +125,14 @@ JL_DLLEXPORT jl_value_t *jl_gc_internal_obj_base_ptr(void *p);
 // the size of that stack buffer upon return. Also, if task is a thread's
 // current task, that thread's id will be stored in *tid; otherwise,
 // *tid will be set to -1.
+//
+// DEPRECATED: use jl_active_task_stack() instead.
 JL_DLLEXPORT void *jl_task_stack_buffer(jl_task_t *task, size_t *size, int *tid);
+
+// Query the active stack range for the given task, and set *start and *end
+// accordingly. That is, the active part of the stack is between *start and *end.
+// However, these bounds may not be tight.
+JL_DLLEXPORT void jl_active_task_stack(jl_task_t *task, char **start, char **end);
 
 #ifdef __cplusplus
 }
--- a/src/julia_gcext.h
+++ b/src/julia_gcext.h
@@ -129,10 +129,13 @@ JL_DLLEXPORT jl_value_t *jl_gc_internal_obj_base_ptr(void *p);
 // DEPRECATED: use jl_active_task_stack() instead.
 JL_DLLEXPORT void *jl_task_stack_buffer(jl_task_t *task, size_t *size, int *tid);
 
-// Query the active stack range for the given task, and set *start and *end
-// accordingly. That is, the active part of the stack is between *start and *end.
-// However, these bounds may not be tight.
-JL_DLLEXPORT void jl_active_task_stack(jl_task_t *task, char **start, char **end);
+// Query the active and total stack range for the given task, and set
+// *active_start and *active_end respectively *total_start and *total_end
+// accordingly. The range for the active part is a best-effort approximation
+// and may not be tight.
+JL_DLLEXPORT void jl_active_task_stack(jl_task_t *task,
+                                       char **active_start, char **active_end,
+                                       char **total_start, char **total_end);
 
 #ifdef __cplusplus
 }
--- a/src/task.c
+++ b/src/task.c
@@ -247,6 +247,52 @@ JL_DLLEXPORT void *jl_task_stack_buffer(jl_task_t *task, size_t *size, int *tid)
     return (void *)((char *)task->stkbuf + off);
 }
 
+JL_DLLEXPORT void jl_active_task_stack(jl_task_t *task, char **start, char **end)
+{
+    if (!task->started) {
+        *start = *end = 0;
+        return;
+    }
+
+    int16_t tid = task->tid;
+    jl_ptls_t ptls2 = (tid != -1) ? jl_all_tls_states[tid] : 0;
+
+    if (task->copy_stack && ptls2 && task == ptls2->current_task) {
+        *start  = (char*)ptls2->stackbase - ptls2->stacksize;
+        *end = (char*)ptls2->stackbase;
+    }
+    else if (task->stkbuf) {
+        *start  = (char*)task->stkbuf;
+#ifdef COPY_STACKS
+        // save_stack stores the stack of an inactive task in stkbuf, and the
+        // actual number of used bytes in copy_stack.
+        if (task->copy_stack > 1)
+            *end = (char*)task->stkbuf + task->copy_stack;
+        else
+#endif
+            *end = (char*)task->stkbuf + task->bufsz;
+
+#ifndef _OS_WINDOWS_
+        if (jl_all_tls_states[0]->root_task == task) {
+            // See jl_init_root_task(). The root task of the main thread
+            // has its buffer enlarged by an artificial 3000000 bytes, but
+            // that means that the start of the buffer usually points to
+            // inaccessible memory. We need to correct for this.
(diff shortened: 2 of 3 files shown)


src/task.c
```diff
@@ -235,6 +235,52 @@ JL_DLLEXPORT void *jl_task_stack_buffer(jl_task_t *task, size_t *size, int *tid)
     return (void *)((char *)task->stkbuf + off);
 }
 
+JL_DLLEXPORT void jl_active_task_stack(jl_task_t *task, char **start, char **end)
+{
+    if (!task->started) {
```
fingolfin, line 240: We didn't handle this case in `jl_task_stack_buffer`, and I am not sure we need to handle it here? But I thought "better safe than sorry"...

src/task.c
```diff
@@ -235,6 +235,52 @@ JL_DLLEXPORT void *jl_task_stack_buffer(jl_task_t *task, size_t *size, int *tid)
...
+        *start = *end = 0;
+        return;
+    }
+
+    int16_t tid = task->tid;
+    jl_ptls_t ptls2 = (tid != -1) ? jl_all_tls_states[tid] : 0;
+
+    if (task->copy_stack && ptls2 && task == ptls2->current_task) {
```
fingolfin, line 248: This code is strongly inspired by `gc_scrub_task` with a few differences:
- we adjust for `ROOT_TASK_STACK_ADJUSTMENT`
- we adjust the stack end if `task->copy_stack > 1` and `COPY_STACKS` is #defined.

test/gcext/gcext.c
```diff
@@ -478,37 +478,27 @@ static int stack_grows_down(void) {
...
-            // only scan the live portion of the stack.
-            char *end_stack = (char *) stack + size;
-            if (lt_ptr(stack, &size) && lt_ptr(&size, (char *)stack + size)) {
-                if (stack_grows_down()) {
-                    size = end_stack - (char *)&size;
-                    stack = (void *)&size;
-                }
-                else {
-                    size = (char *) end_stack - (char *) &size;
-                }
```
fingolfin, line 500: This is all now handled by `jl_active_task_stack`

test/gcext/gcext.c
```diff
@@ -478,37 +478,27 @@ static int stack_grows_down(void) {
...
+    jl_active_task_stack(task, &start_stack, &end_stack);
+
+    // this is the live stack of a thread. Is it ours?
+    if (start_stack && task == (jl_task_t *)jl_get_current_task()) {
+        if (!(lt_ptr(start_stack, &var_on_frame) && lt_ptr(&var_on_frame, end_stack))) {
+            // error, current stack frame must be on the live stack.
+            jl_error("stack frame not part of the current task");
         }
-        else
-            stack = NULL;
```
fingolfin, line 507: This (removed) code puzzles me: it seems we didn't bother to scan the stacks of (the active tasks of) background threads at all? Perhaps to avoid running into guard pages? Perhaps we actually *should* scan those, and also make sure the test activates multiple threads and tasks, to stress test that (and then use `safe_restore` via the new APIs)?

src/task.c
```diff
@@ -235,6 +235,52 @@ JL_DLLEXPORT void *jl_task_stack_buffer(jl_task_t *task, size_t *size, int *tid)
...
+        *start  = (char*)ptls2->stackbase - ptls2->stacksize;
+        *end = (char*)ptls2->stackbase;
+    }
+    else if (task->stkbuf) {
+        *start  = (char*)task->stkbuf;
+#ifdef COPY_STACKS
+        // save_stack stores the stack of an inactive task in stkbuf, and the
+        // actual number of used bytes in copy_stack.
+        if (task->copy_stack > 1)
+            *end = (char*)task->stkbuf + task->copy_stack;
```
rbehrends, line 258: `task->copy_stack` counts bytes from the high address (see `save_stack()`), so it should probably be something like:

```
*end = (char*)task->stkbuf + task->bufsz;
*start = *end - task->copy_stack;
```

src/julia_gcext.h
```diff
@@ -125,8 +125,15 @@ JL_DLLEXPORT jl_value_t *jl_gc_internal_obj_base_ptr(void *p);
...
 // current task, that thread's id will be stored in *tid; otherwise,
 // *tid will be set to -1.
+//
+// DEPRECATED: use jl_active_task_stack() instead.
 JL_DLLEXPORT void *jl_task_stack_buffer(jl_task_t *task, size_t *size, int *tid);
 
+// Query the active stack range for the given task, and set *start and *end
+// accordingly. That is, the active part of the stack is between *start and *end.
+// However, these bounds may not be tight.
+JL_DLLEXPORT void jl_active_task_stack(jl_task_t *task, char **start, char **end);
```
rbehrends, line 135: I wonder if it might be useful to return the start of the stack buffer along with its end and a lower bound for the current sp. While this may not currently have an application, it might be useful to have a more comprehensive interface for future use cases (especially debugging) and this is not  function that is going to be used on a hot path.

fingolfin: So you are suggesting that besides providing the "active" range (resp. the best approximation it can determine for that), it should also provide the total range, i.e., more or less what `jl_task_stack_buffer` produces? I.e. something like
```suggestion
JL_DLLEXPORT void jl_active_task_stack(jl_task_t *task, char **active_start, char **active_end, char **total_start, char **total_end);
```
or so?

src/task.c
fingolfin: I don't follow. Looking at `save_stack`, I see this:
```C
...
    if (lastt->bufsz < nb) {
        buf = (void*)jl_gc_alloc_buf(ptls, nb);
        lastt->stkbuf = buf;
        lastt->bufsz = nb;
    }
    else {
        buf = lastt->stkbuf;
    }
    *pt = NULL; // clear the gc-root for the target task before copying the stack for saving
    lastt->copy_stack = nb;
    lastt->sticky = 1;
    memcpy_a16((uint64_t*)buf, (uint64_t*)frame_addr, nb);
...
```
So data starts at `lastt->stkbuf` and its length is `lastt->copy_stack`. What am I missing?

rbehrends: Oh, I just realized that I confused the live stack with the saved stack. The copy is from `frame_addr`, but is put at the beginning of the buffer. Nevermind.

src/julia_gcext.h
rbehrends: Roughly speaking, yes. I.e. it might be useful to know the stack buffer along with the part of the stack buffer (or an approximation thereof) that is currently in use, e.g. for debugging and assertions.

JeffBezanson (approved):

JeffBezanson:
Looks fine to me if this is ready to merge.

rbehrends:
The `total_start/total_end` changes look good to me.

fingolfin:
@JeffBezanson should be ready now (be advised that I rebased and added a second commit -- so you may wish to re-review that bit)

fingolfin:
@JeffBezanson OK to merge this?

fingolfin:
@JeffBezanson ping? I've once again rebased this, to verify that the CI tests still pass.

PR #36824, opened 2020-07-27
Only values, not variables, have types...

josePereiro:
The manual says "Only values, not variables, have types -- variables are simply names bound to values..." but just a few lines after "it declares the variable to always have the specified type...".

### DIFF
--- a/doc/src/manual/types.md
+++ b/doc/src/manual/types.md
@@ -40,7 +40,7 @@ up front are:
   * There is no meaningful concept of a "compile-time type": the only type a value has is its actual
     type when the program is running. This is called a "run-time type" in object-oriented languages
     where the combination of static compilation with polymorphism makes this distinction significant.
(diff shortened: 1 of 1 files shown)


doc/src/manual/types.md
```diff
@@ -40,7 +40,8 @@ up front are:
   * There is no meaningful concept of a "compile-time type": the only type a value has is its actual
     type when the program is running. This is called a "run-time type" in object-oriented languages
     where the combination of static compilation with polymorphism makes this distinction significant.
-  * Only values, not variables, have types -- variables are simply names bound to values.
+  * Only values, not variables, have types -- variables are simply names bound to values, although for 
+    simplicity we often say "type of a variable".
```
StefanKarpinski, line 44:
```suggestion
    simplicity we may say "type of a variable" as shorthand for "type of the value to which a variable refers."
```

StefanKarpinski:
This has trailing whitespace that needs to be removed.

Issue #36786, opened 2020-07-24
Revisit erroring behaviour of special functions


MasonProtter:
Obviously, this can't be changed until 2.0, but I think we should revisit the design of special functions such as `sqrt`, `sin`, etc. to make them pure functions which cannot error. The fact that `sqrt(-1)` or `sin(Inf)`  produce errors make many important optimizations impossible, or harder for the compiler to perform. We can solve these things with special cases in the compiler, but that'll quickly turn into optimization whack-a-mole.  

The IEEE floating point specification (and computer hardware) already has a great solution for this: `NaN`. One can argue that producing `NaN` for these functions *is* the correct thing to do according to the IEEE spec.

The main problem that I forsee with returning `NaN` rather than an error is that `NaN`s are notoriously  difficult to track down, but I feel that with the new compiler pass infrastructure (or even the current Cassette.jl / IRTools.jl), making better tooling for detecting `NaN` production would solve this potential complaint and allow us to make more principled optimizations for these functions.

musm:
Just use https://github.com/mlubin/NaNMath.jl ?

musm:
Dupe of https://github.com/JuliaLang/julia/issues/5234

MasonProtter:
I've been told to use NaNMath several times, but first of all NaNMath.jl does **not** solve the problem with optimizations here because it uses `ccall`s that the compiler can't prove are pure, so it has the same issue (in addition to being slower than Base):
```julia
julia> using StaticArrays, NaNMath, BenchmarkTools

julia> let v = @SVector(rand(20)), x = 5.0
           sinx = NaNMath.sin(x)
           @btime $v .+ NaNMath.sin($x)
           @btime $v .+ $sinx
           @btime $v .+ sin($x)
       end;
  13.175 ns (0 allocations: 0 bytes)
  0.019 ns (0 allocations: 0 bytes)
  10.801 ns (0 allocations: 0 bytes)
```

Second of all, I think if Base is going to provide these functions, it's worth having Base do them right rather than rely on external packages to patch performance problems.

musm:
There is an 10-15% speed increase  in performance between `NaNMath.sin` and `Base.sin` due to the fact that `Base.sin` is completely implemented in julia based on the openlibm code, while `NaNMath.sin` is calling the openlibm library directly, and the Julia implementation runs faster. *This entirely explains the performance difference between `NaNMath.sin` and `Base.sin`.*

[...]

PR #33730, opened 2019-10-31
updating domtrees dynamically, removing all unreachable blocks

yhls:
This all started with #29107, which (is at least one thing that) stands in the way of turning DCE back on globally (it was disabled in #29265). The problem is that when there is only one predecessor, replacing a phi node with its only possible value can result in a forward reference when this phi node is in an unreachable loop. This can cause problems with passes that process all IR, whether or not the code is dead.

We can remove dead loops by removing all unreachable blocks, instead of just removing blocks with no predecessors, which is what we currently do. To determine reachability, all that would be required is a depth-first traversal of the CFG, but as suggested by #29140, I have implemented an algorithm for updating the dominator tree of a CFG dynamically. I hear that this would be useful to have anyway. Just to be safe, I’ve also prevented a phi node from being replaced with its value if that value is defined after the phi node is.

The dynamic domtree algorithm basically works by figuring out if we can get away with doing nothing or recomputing only some of the immediate dominators when we add or delete an edge in the CFG, or if we have to recompute the whole domtree from scratch.

As this PR stands, domtrees are not updated dynamically yet, even though the dynamic algorithm has been implemented. First, I tried just computing the domtree from scratch when we decide which blocks are dead, which happens when an `IncrementalCompact` is created. I tried to get an idea of the performance impact by comparing how long Base and Stdlibs took to compile, but the compile times were pretty noisy and I couldn’t confidently notice any difference. Next, I tried adding a `DomTree` to every `CFG` and computing it every time a `CFG` is constructed. This appears to lengthen compile times by about 5% compared with master, and 3% compared with master with DCE on. The only times a new `CFG` is made from an old one are when we finish iterating over an `IncrementalCompact`, and in `domsort_ssa!`. Presumably we would be able to recover some performance by updating the domtree dynamically in these places.

I’ve also addressed a bug wherein references to dead blocks that have been removed remain in phi nodes.

PR #32368, opened 2019-06-19
codegen: propagate at-pure macro to llvm

vtjnash:
Adds ReadNone and NoUnwind (and thunk) to call sites marked `@pure` in the source code. This allows LLVM to perform LICM and hoisting on these functions, which can be highly profitable.

PR #36832, opened 2020-07-28
add LoopInfo + LICM

vchuravy:
In recent discussion as well as in the past the topic came up that LLVM LICM pass is currently not effectual for calls to Julia code and Julia specific constructs. In https://github.com/JuliaLang/julia/issues/36786#issuecomment-664247514 I explained my understanding that we don't need to change the semantics of math functions to enable LICM. In general it seems that informing LLVM about the pureness of Julia functions might not be valid (https://github.com/JuliaLang/julia/pull/29566#issuecomment-449604059, #32368) since Julia and LLVM have differing definitions, this will fixe #29285

We could also do #36809 (Custom-LICM pass for `GC.@preserve` on the LLVM level) as part of this pass.

In order to make this more useful we might want to track more properties of Julia methods. For simplicity for now I have limited myself to `@pure` functions.

TODO:
- [ ] Create a pre-header properly, this is a CFG manipulation and might need #33730 to update the domtree dynamically
- [ ] Iterate over nested loops correctly, innermost to outermost, while maintaining LoopInfo
- [ ] Various todos in the code

cc: @yhls, @jpsamaroo 

@timholy Together with a Loop Induction Variables deduction pass the loopinfo pass could be used to implement #35074 without special casing `CartesianIndices`

I am putting this up mostly to get help with figuring out the CFG manipulation part. I looked through inlining, but it seems we could make the insertion of BBs an independent util.


vchuravy:
With this PR:

```
Base.@pure @noinline mysin(x) = sin(x)
function example(A, x)
           for i in 1:length(A)
               A[i] = A[i] + (mysin(x) + mysin(x))
           end
       end
```

turns into:
```
julia> @code_typed example(zeros(3), 3.0)
CodeInfo(
1 ── %1  = Base.arraylen(A)::Int64
│    %2  = Base.sle_int(1, %1)::Bool
│    %3  = Base.ifelse(%2, %1, 0)::Int64
│    %4  = Base.slt_int(%3, 1)::Bool
└───       goto #3 if not %4
2 ──       goto #4
3 ──       goto #4
4 ┄─ %8  = φ (#2 => true, #3 => false)::Bool
│    %9  = φ (#3 => 1)::Int64
│    %10 = φ (#3 => 1)::Int64
│    %11 = Base.not_int(%8)::Bool
│    %12 = invoke Main.mysin(_3::Float64)::Float64
│    %13 = invoke Main.mysin(_3::Float64)::Float64
│    %14 = Base.add_float(%12, %13)::Float64
└───       goto #10 if not %11
5 ┄─ %16 = φ (#4 => %9, #9 => %26)::Int64
│    %17 = φ (#4 => %10, #9 => %27)::Int64
│    %18 = Base.arrayref(true, A, %16)::Float64
│    %19 = Base.add_float(%18, %14)::Float64
│          Base.arrayset(true, A, %19, %16)::Vector{Float64}
│    %21 = (%17 === %3)::Bool
└───       goto #7 if not %21
6 ──       goto #8
7 ── %24 = Base.add_int(%17, 1)::Int64
└───       goto #8
8 ┄─ %26 = φ (#7 => %24)::Int64
│    %27 = φ (#7 => %24)::Int64
│    %28 = φ (#6 => true, #7 => false)::Bool
│    %29 = Base.not_int(%28)::Bool
└───       goto #10 if not %29
9 ──       goto #5
10 ┄       return nothing
) => Nothing
````

Instead of:
```
CodeInfo(
1 ── %1  = Base.arraylen(A)::Int64
│    %2  = Base.sle_int(1, %1)::Bool
│    %3  = Base.ifelse(%2, %1, 0)::Int64
│    %4  = Base.slt_int(%3, 1)::Bool
└───       goto #3 if not %4
2 ──       goto #4
3 ──       goto #4
4 ┄─ %8  = φ (#2 => true, #3 => false)::Bool
│    %9  = φ (#3 => 1)::Int64
│    %10 = φ (#3 => 1)::Int64
│    %11 = Base.not_int(%8)::Bool
└───       goto #10 if not %11
5 ┄─ %13 = φ (#4 => %9, #9 => %26)::Int64
│    %14 = φ (#4 => %10, #9 => %27)::Int64
│    %15 = Base.arrayref(true, A, %13)::Float64
│    %16 = invoke Main.mysin(_3::Float64)::Float64
│    %17 = invoke Main.mysin(_3::Float64)::Float64
│    %18 = Base.add_float(%16, %17)::Float64
│    %19 = Base.add_float(%15, %18)::Float64
│          Base.arrayset(true, A, %19, %13)::Vector{Float64}
│    %21 = (%14 === %3)::Bool
└───       goto #7 if not %21
6 ──       goto #8
7 ── %24 = Base.add_int(%14, 1)::Int64
└───       goto #8
8 ┄─ %26 = φ (#7 => %24)::Int64
│    %27 = φ (#7 => %24)::Int64
│    %28 = φ (#6 => true, #7 => false)::Bool
│    %29 = Base.not_int(%28)::Bool
└───       goto #10 if not %29
9 ──       goto #5
10 ┄       return nothing
) => Nothing
```

yuyichao:
> We could also do #36809 (Custom-LICM pass for GC.@preserve on the LLVM level) as part of this pass.

For sure. Though I wouldn't use one to replace the other as long as we still uses any LLVM loop passes. In the llvm pass, I'm counting on LLVM LICM to hoist some generation of the pointer value out of the loop (e.g. loading from a array/object that is loop invariant).

Also, there are some julia specific LICM that can only be done at LLVM level, including everything GC/allocation related. It can certainly be done in julia code in the future but it'll require a much lower level IR.

yuyichao:
Also, I feel like it wouldn't hurt to mark some functions that are believed to be "pure"/LICM safe with some custom metadata so that our own LLVM pass can pick it up.

vtjnash:
> In general it seems that informing LLVM about the pureness of Julia functions might not be valid (#29566 (comment), #32368) since Julia and LLVM have differing definitions, this will fixe #29285

FWIW, the issue was that many of our uses of `@pure` aren't valid to LICM, not that it was bad to set the attribute in LLVM.

yuyichao:
I’m not really commenting on what is safe to LICM, I just mean that whatever identified  here as safe should also be safe to do so in llvm. I assume there isn't a case where the code is safe to licm in Julia but not in llvm?

vchuravy:
Now that we have effect analysis this is now feasible and thanks #45305 added the gnarly details of how to do CFG manipulations.

Still some details to figure out, in particular how to deal with loop-nests and multiple loops in a function since the act of mutating the CFG invalidates the collect loop information.

Also need to check multiple invariant statements doing the right thing.

Developed this out-of-tree in https://github.com/vchuravy/Loops.jl and https://vchuravy.dev/talks/licm/

Will fix the long-standing #29285

PR #36834, opened 2020-07-29
add docstring for return_type

jw3126:


base/compiler/typeinfer.jl
```diff
@@ -659,6 +659,19 @@ function typeinf_ext_toplevel(interp::AbstractInterpreter, linfo::MethodInstance
...
+    return_type(f, ArgType)
+
+Estimate the return type of `f` given arguments of type `ArgType`. In other words this gives an upper bound for `typeof(f((args::ArgType)...))`.
+
+!!! warning
+
+    This function is not part of the stable API. It is an implementation detail of the compiler,
+    that can and will change any time. For instance it might give sharper bounds as
+    inference improves or looser bounds to improve latency.
+    Using `return_type` is very likely to lead to undefined behaviour.
```
jw3126, line 673: Are there more reasons to avoid `return_type`? I vaguely remember, that it hurts inference, is that true?

JeffBezanson: "Undefined behavior" has a very specific meaning in programming languages. I would say "unpredictable behavior" instead.

This pretty much covers the reasons not to use it, but you could also add that it makes programs less robust since this function is likely to do what you want only for certain types and not all types.

base/Base.jl
```diff
@@ -45,6 +45,9 @@ if isdefined(Core, :Compiler)
     Core.atdoc!(CoreDocs.docm)
 end
 
+# TODO better place for this?
```
JeffBezanson, line 48: How about base/docs/basedocs.jl?

jw3126:
Can this be merged?

StefanKarpinski:
@JeffBezanson, your call.

ViralBShah:
If we like it - let's rebase it - otherwise we should close.

tkf:
I believe we should make `return_type` a public API and explain all the nuances. For my suggestion, see #44340

jw3126:
I close, if we want to document this, @tkf s version is much better than this PR!

PR #36836, opened 2020-07-29
Add specialized unique(::AbstractRange) method

sostock:
Currently, there is no specialized `unique` implementation for ranges, so `unique` falls back to the default implementation for iterators, which seems wasteful.

This PR implements `unique(::AbstractRange)` and simplifies `allunique(::AbstractRange)`.

`unique(::AbstractRange)` now returns a range instead of a `Vector`. Is it okay that `unique` does not return a mutable collection, or should I change it to return a `Vector` like before?

### DIFF
--- a/base/set.jl
+++ b/base/set.jl
@@ -166,6 +166,8 @@ _unique_from(itr, out, seen, i) = unique_from(itr, out, seen, i)
     return out
 end
 
+unique(r::AbstractRange) = allunique(r) ? r : oftype(r, r[begin:begin])
+
 """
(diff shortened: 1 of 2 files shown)


StefanKarpinski:
I think this is ok to change in a minor release. This does seem like a good optimization to have.

sostock:
The `allunique` simplification is actually a bugfix (cf. #36849), so I will split this PR into two, one containing the `allunique` bugfix and one containing the new `unique` method.

PR #36841, opened 2020-07-29
Diagonal: Check dimensions when multiplying with vector

barucden:
I have added a dimension check for the specific multiplication of diagonal matrix and vector, which was missing as reported by JuliaLang/LinearAlgebra.jl#750.

### DIFF
--- a/stdlib/LinearAlgebra/src/diagonal.jl
+++ b/stdlib/LinearAlgebra/src/diagonal.jl
@@ -174,7 +174,15 @@ end
 (*)(x::Number, D::Diagonal) = Diagonal(x * D.diag)
 (*)(D::Diagonal, x::Number) = Diagonal(D.diag * x)
 (/)(D::Diagonal, x::Number) = Diagonal(D.diag / x)
-(*)(Da::Diagonal, Db::Diagonal) = Diagonal(Da.diag .* Db.diag)
+
+function (*)(Da::Diagonal, Db::Diagonal)
+    nDa, mDb = size(Da, 2), size(Db, 1)
+    if nDa != mDb
+        throw(DimensionMismatch("second dimension of Da, $nDa, does not match first dimension of Db, $mDb"))
+    end
+    return Diagonal(Da.diag .* Db.diag)
+end
+
 function (*)(D::Diagonal, V::AbstractVector)
     nD = size(D, 2)
     if nD != length(V)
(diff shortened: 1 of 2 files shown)


stdlib/LinearAlgebra/src/diagonal.jl
```diff
@@ -175,7 +175,13 @@ end
 (*)(D::Diagonal, x::Number) = Diagonal(D.diag * x)
 (/)(D::Diagonal, x::Number) = Diagonal(D.diag / x)
 (*)(Da::Diagonal, Db::Diagonal) = Diagonal(Da.diag .* Db.diag)
```
fredrikekre, line 177: This method needs the same treatment -- can you add that too?

barucden: Sure, I pushed another commit

mcognetta: I don't think that Diagonal*Diagonal needs it. It falls back to the element wise product of vectors which has a dimensionality check.

barucden: It seems that a single-element vector behaves as a scalar when doing the element-wise product:
```
julia> randn(1) .* randn(5)
5-element Vector{Float64}:
 -0.09149628486880286
 -0.20889567435214648
  0.09612350337701256
  0.08092068173428073
  0.06652421890696196
```

fredrikekre (approved):
Thanks, and welcome as a contributor to Julia!

stdlib/LinearAlgebra/src/diagonal.jl
mcognetta: Ah, nice catch.

barucden:
Thank you!

KristofferC:
This makes a package fail when backported. Even though it is a bugfix, I'll leave it out for now.

fredrikekre:
Which package?

barucden:
[SparseGaussianProcesses](https://github.com/JuliaCI/NanosoldierReports/blob/943384554f9d9b8edc48f4ca1a6563fdb27f9cde/pkgeval/by_hash/317b2cf_vs_f0b8c76/logs/SparseGaussianProcesses/1.5.1-pre-9d1f8db49b.log), I believe.

PR #36850, opened 2020-07-30
Fix allunique(::StepRangeLen)

### LINKED ISSUE #36849: allunique(::StepRangeLen) regression
Julia 1.4:
```julia
julia> allunique(range(0.0, stop=0.0, length=5))
false
```

Julia 1.5-rc2 and master:
```julia
julia> allunique(range(0.0, stop=0.0, length=5))
true
```

This was introduced by #35008. The regression would be fixed by #36836. However, that PR also introduces a minor change. I will split it into two PRs, one containing the `allunique` bugfix and another one with the minor change.

sostock:
Fixes #36849.

### DIFF
--- a/base/set.jl
+++ b/base/set.jl
@@ -393,9 +393,7 @@ end
 
 allunique(::Union{AbstractSet,AbstractDict}) = true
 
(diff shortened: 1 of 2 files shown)


sostock:
This doesn’t need backporting to 1.0, only to 1.5.

PR #36851, opened 2020-07-30
 Add specialized unique(::AbstractRange) method

sostock:
This was formerly #36836, which I apparently cannot reopen because I force-pushed it. This implements the “minor change” part of that PR on top of #36850 and should be discussed at triage.


sostock:
Since this is a minor change, should it appear in NEWS.md?

stevengj:
It's not an API change, just an optimization, so no NEWS required.

stevengj (approved):

PR #36853, opened 2020-07-30
symbol.c: merge sum calculations

zlatanvasovic:
This should be fine unless those add extra clarity to the code.
(But I don't think so, e.g. `... len + 1 + 7) & -8` doesn't add any clarity).

### DIFF
--- a/src/symbol.c
+++ b/src/symbol.c
@@ -24,7 +24,7 @@ static uintptr_t hash_symbol(const char *str, size_t len) JL_NOTSAFEPOINT
 
 static size_t symbol_nbytes(size_t len) JL_NOTSAFEPOINT
 {
-    return (sizeof(jl_taggedvalue_t) + sizeof(jl_sym_t) + len + 1 + 7) & -8;
+    return (sizeof(jl_taggedvalue_t) + sizeof(jl_sym_t) + len + 8) & -8;
 }
 
 static jl_sym_t *mk_symbol(const char *str, size_t len) JL_NOTSAFEPOINT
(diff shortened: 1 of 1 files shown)


yuyichao:
The code **DOES** add extra clearity. `(+ 7) & -8` is a standard pattern for rounding and `len + 3 + sizeof(gs_name)` is a value with meaning in the computation.

zdroid:
@yuyichao Right, I missed that. I see `sizeof(gs_name) + len + 3` is used previously. Would it be any good if that value was extracted into a seperate variable?

 Also, you don't need to **SCREAM** at me to get your point across, I understand you very well. :smile:

KristofferC:
As a side note, I am not sure working on PRs that do these types of trivial refactorings (this PR, https://github.com/JuliaLang/julia/pull/36812, https://github.com/JuliaLang/julia/pull/36805 etc) on parts of the codebase that one previously haven't worked on is such a good idea. You don't have the context in which that code was written so assuming that the code is written the way it is for a reason might make more send. It's unclear to me that hunting around the code base for these types of refactorings/compressions is time well spent, both for the PR author and reviewers.

yuyichao:
> Right, I missed that. I see `sizeof(gs_name) + len + 3` is used previously. Would it be any good if that value was extracted into a seperate variable?

No.

zdroid:
I'm just exploring the codebase so those minor things come at a glance. But of course, if you are against me doing this, I won't do it. Sorry for incovenience.

StefanKarpinski:
It's great that you're going through the code base like this—please don't be discouraged by the reactions of these cantankerous compiler hackers. They're a grumpy lot 😝. But please keep in mind that you're in "their home" so to speak—these people live with this code every day, so while improvements are welcome, consider first:

* Is this the way it is for a reason? In this case, why would the code not already combine obvious addition of constants? The answer is that keeping them separate clarifies the logic.
* Don't just change something because you have a slight preference for writing it another way. Someone who works on this code wrote it that way in the first place.

That said, this code is far from perfect, but it seems like the threshold for tweaks should probably be higher.

PR #36862, opened 2020-07-30
speedup norm on AbstractArray{<:Number}

jw3126:
This gives a nice speedup for generic arrays.
## PR
```julia
julia> arr = randn(10^5); oarr = OffsetArray(arr, eachindex(arr)); @btime norm(arr); @btime norm(oarr)
  29.836 μs (1 allocation: 16 bytes)
  73.266 μs (1 allocation: 16 bytes)
```
## Baseline
```julia
julia> arr = randn(10^5); oarr = OffsetArray(arr, eachindex(arr)); @btime norm(arr); @btime norm(oarr)

  29.833 μs (1 allocation: 16 bytes)
  254.549 μs (1 allocation: 16 bytes)
```

### DIFF
--- a/stdlib/LinearAlgebra/src/generic.jl
+++ b/stdlib/LinearAlgebra/src/generic.jl
@@ -532,12 +532,59 @@ function generic_normp(x, p)
     end
 end
 
+
 normMinusInf(x) = generic_normMinusInf(x)
 normInf(x) = generic_normInf(x)
 norm1(x) = generic_norm1(x)
 norm2(x) = generic_norm2(x)
 normp(x, p) = generic_normp(x, p)
+function norm0(itr)
+    T = typeof(float(norm(first(itr))))
+    return T(count(!iszero, itr))
+end
 
+@inline strip_val(p) = p
+@inline strip_val(::Val{p}) where {p} = p
+@inline abspow(x, p) = norm(x)^p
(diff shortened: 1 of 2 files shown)


stdlib/LinearAlgebra/src/generic.jl
```diff
@@ -582,14 +635,14 @@ julia> norm(v, 1)
...
+julia> norm([1 2 3; 4 5 6; 7 8 9]) ≈ sqrt(sum(abs2, 1:9))
+true
 
-julia> norm([1 2 3 4 5 6 7 8 9])
-16.881943016134134
+julia> norm([1 2 3 4 5 6 7 8 9]) ≈ sqrt(sum(abs2, 1:9))
+true
 
-julia> norm(1:9)
-16.881943016134134
```
jw3126, line 592: Was off by one epsilon. This is expected. Reason is that somewhere we optimized `a/b` into `binv = inv(b); a * binv`. Also, SIMD is used now (not in this specific example), which can also give slightly different results.

DilumAluthge:
We have moved the LinearAlgebra stdlib to an external repo: https://github.com/JuliaLang/LinearAlgebra.jl

@jw3126 If you think that this PR is still relevant, please open a new PR on the [LinearAlgebra.jl](https://github.com/JuliaLang/LinearAlgebra.jl) repo.

PR #36733, opened 2020-07-19
avoid invalidations when it doesn't resolve an MethodError

vtjnash:
In these cases, the new method would not be called because the
call would still be an ambiguity error instead.

We also might observe that the method doesn't resolve an old MethodError
because it was previously covered by a different method.

(replaces #35877)

PR #36863, opened 2020-07-30
add a constructor for Union{}

vtjnash:
This often shows up accidentally in method intersections, which can force recompilation of completely unrelated code. Define this explicitly so that the compiler can always ignore this dispatch possibility.


timholy:
StaticArrays goes from 1200 invalidations to just 31. Whether we do it this way or some other way, this is a great outcome! :fireworks:

timholy:
To those watching, worth noting that this works thanks to #36733.

JeffBezanson (approved):

PR #36870, opened 2020-07-31
Runtime event trace visualization

kpamnany:
This adds a `bpftrace` script containing handlers for many of the probes in the Julia runtime, and also adds some probes. The various handlers produce JSON output that can be read by [Perfetto](https://ui.perfetto.dev/).

Here's what this looks like for parallel Fibonacci with 4 threads:
image

BUILD INSTRUCTIONS:
- Create `Make.user` containing `WITH_DTRACE = 1`.

TODO:
- [X] Determine why there are misplaced end events and fix.
- [X] Cannot use task pointer as task ID as tasks can be reused.
- [X] Add more probes (start+finish inference, start+finish codegen, others?).

Cc: @vtjnash, @JeffBezanson

### DIFF
--- a/contrib/bpftrace/rt_perfetto.bt
+++ b/contrib/bpftrace/rt_perfetto.bt
@@ -3,11 +3,13 @@
 BEGIN
 {
     @gc_num = 1;
+    @ti_num = 1;
     printf("{\"displayTimeUnit\":\"ns\",\"traceEvents\":[\n");
 }
 
 END
 {
+    clear(@ti_num);
     clear(@gc_num);
     printf("]}");
 }
@@ -89,6 +91,18 @@ usdt:usr/lib/libjulia-internal.so:julia:rt__pause__task
     printf("{\"name\":\"TaskRun\",\"cat\":\"task\",\"id\":\"0x%x\",\"ph\":\"E\",\"pid\":%-d,\"tid\":%-d,\"ts\":%lu},\n", *$ptr0, pid, tid, elapsed);
 }
 
+// Type inference events
+usdt:usr/lib/libjulia-internal.so:julia:rt__start__typeinf
+{
+    printf("{\"name\":\"TypeInference\",\"cat\":\"compiler\",\"id\":%-d,\"ph\":\"B\",\"pid\":%-d,\"tid\":%-d,\"ts\":%lu},\n", @ti_num, pid, tid, elapsed);
+}
+
+usdt:usr/lib/libjulia-internal.so:julia:rt__finish__typeinf
+{
+    printf("{\"name\":\"TypeInference\",\"cat\":\"compiler\",\"id\":%-d,\"ph\":\"E\",\"pid\":%-d,\"tid\":%-d,\"ts\":%lu},\n", @ti_num, pid, tid, elapsed);
+    @ti_num++;
+}
+
 /*
 // IO loop events: MANY events
 usdt:usr/lib/libjulia-internal.so:julia:rt__start__process__events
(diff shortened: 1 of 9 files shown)


src/trace.c
```diff
@@ -0,0 +1,115 @@
...
+    uint64_t timestamp;
+    uint64_t event;
+
+} jl_tracepkt_t;
+
+static struct sockaddr_in send_addr;
+static uv_udp_t *send_socket;
+static uv_udp_send_t **send_reqs;
+
+static jl_mutex_t flush_traces_lock;
```
vtjnash, line 27: This should be the iolock, since you're accessing the jl_io_loop

src/trace.c
```diff
@@ -0,0 +1,115 @@
...
+#ifdef ENABLE_TRACING
+
+#include <stdlib.h>
+#include "threading.h"
+
+typedef struct _jl_tracepkt_t {
+    int16_t tid;
+    uint64_t timestamp;
+    uint64_t event;
+
```
vtjnash, line 20: You could pack more into a single UDP packet by coalescing the common data:
```suggestion
    uint16_t tid;
    uint32_t pid;
    uint64_t timestamp0;
    struct {
        uint16_t timestamp;
        uint16_t event;
    } data[(1472 - 128) / 32];
```

vilterp:
May be worth trying to feed this into [Catapult](https://github.com/catapult-project/catapult), the tracing UI built into Chrome, so we don't have to write our own. The format is documented here: https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/preview

The Go runtime uses this to visualize scheduling and GC events. Here's an example debugging session: https://about.sourcegraph.com/go/an-introduction-to-go-tool-trace-rhys-hiltner/

kpamnany:
With the current set of `jl_trace()` calls, the trace file generated from `julia --threads=4 --trace-file=julia_startup.xml` looks like:

image

kpamnany:
After some discussions, we decided that this will be reworked on top of #43453 as it will be possible to completely avoid overhead when not interested in tracing (without recompiling).

kpamnany:
For the added probes:
- Type inference [begin](https://github.com/JuliaLang/julia/pull/36870/files#diff-9a718f382a240acdbe9d0bdbc3ffb521f67af0e62b2ba8c8594b8faa65c78014R3305) and [end](https://github.com/JuliaLang/julia/pull/36870/files#diff-9a718f382a240acdbe9d0bdbc3ffb521f67af0e62b2ba8c8594b8faa65c78014R3320) seem to properly demarcate inference but no contextual information is available at these probe points (i.e. what is being inferred).
- Compilation [begin](https://github.com/JuliaLang/julia/pull/36870/files#diff-41aaf4203c5facc33e59fccd8a09a61fe335d834b35a2679a9f2a2d0514166abR108) and [end](https://github.com/JuliaLang/julia/pull/36870/files#diff-41aaf4203c5facc33e59fccd8a09a61fe335d834b35a2679a9f2a2d0514166abR215) seem to include codegen ([begin](https://github.com/JuliaLang/julia/pull/36870/files#diff-62cfb2606c6a323a7f26a3eddfa0bf2b819fa33e094561fee09daeb328e3a1e7R7928) and [end](https://github.com/JuliaLang/julia/pull/36870/files#diff-62cfb2606c6a323a7f26a3eddfa0bf2b819fa33e094561fee09daeb328e3a1e7R7959)) and LLVM optimization ([begin](https://github.com/JuliaLang/julia/pull/36870/files#diff-41aaf4203c5facc33e59fccd8a09a61fe335d834b35a2679a9f2a2d0514166abR489) and [end](https://github.com/JuliaLang/julia/pull/36870/files#diff-41aaf4203c5facc33e59fccd8a09a61fe335d834b35a2679a9f2a2d0514166abR538)) but from looking at the visualization, there is more than these two things happening during compilation.

I've also switched to using `jl_object_id(task)` for a unique task identifier (since the task pointer itself is not unique).

Hope to get suggestions for additional useful probe points, new/more useful probe parameters or any other changes. Otherwise, this is good to go.

kpamnany:
Here's a zoomed-in section of the Perfetto visualization for parallel fib with 4 threads, showing the newly added sections:

image

PR #20150, opened 2017-01-20
use a counter for passed tests instead of storing the Result object

KristofferC:
Should fix #20137. cc @Keno

PR #36879, opened 2020-08-02
RFC: Add information to passing tests

### LINKED ISSUE #25483: Don't discard all information about passing tests immidately
Right now the testing infrastructure itself discards all information for passing tests.
The `Pass` result type has fields for all the information, and a show method that would display it, but they are all blanked out with `nothing` etc as their values.
https://github.com/JuliaLang/julia/blob/6a45619a22c668237fac959225a5700a31b7da5f/stdlib/Test/src/Test.jl#L432

Determining whether or not to do this should not be done there.
It should be done in the `record` method of the testset.
That way it can be configured.

`record(::DefaultTestSet, ::Pass)` already discards the pass result entirely anyway.

I don't even think doing it later will cost memory, all that information is being captured anyway so that it can be reported for Failure/

mmiller-max:
Closes #25483.

### Summary

With this PR, the information (tested expression and the value) of a succfessul test will be stored in the `Pass` `Result` rather than these fields being set to `nothing`. This will be the case for both `@test` and `@test_throws`. There will be no change to `DefaultTestSet` behaviour, as the current behaviour of `record(ts:DefaultTestSet, t::Pass)` discards any information in the `Pass` result and just increments the number of successful tests. Storing the information allows other testsets to use this information, for example in test report generation.

### More info

Following the discussion in the issue, it seems like the changes made by #20150 mean that information can be stored when tests `Pass` as the whole result is discarded by `DefaultTestSet`s.

This allows other `CustomTestSet`s to use this information, which is important if this information is required, for example by a test case management suite.

### DIFF
--- a/stdlib/Test/src/Test.jl
+++ b/stdlib/Test/src/Test.jl
@@ -85,8 +85,9 @@ struct Pass <: Result
     orig_expr
     data
     value
-    function Pass(test_type::Symbol, orig_expr, data, thrown)
-        return new(test_type, orig_expr, data, thrown isa String ? "String" : thrown)
+    source::Union{Nothing,LineNumberNode}
+    function Pass(test_type::Symbol, orig_expr, data, thrown, source)
+        return new(test_type, orig_expr, data, thrown isa String ? "String" : thrown, source)
     end
 end
 
@@ -236,6 +237,7 @@ function Serialization.serialize(s::Serialization.AbstractSerializer, t::Pass)
     Serialization.serialize(s, t.orig_expr === nothing ? nothing : string(t.orig_expr))
     Serialization.serialize(s, t.data === nothing ? nothing : string(t.data))
     Serialization.serialize(s, string(t.value))
+    Serialization.serialize(s, t.source === nothing ? nothing : t.source)
     nothing
 end
 
@@ -353,9 +355,12 @@ Returns a `Pass` `Result` if it does, a `Fail` `Result` if it is
 ```jldoctest
 julia> @test true
 Test Passed
+  Expression: true
 
 julia> @test [1, 2] + [2, 1] == [3, 3]
 Test Passed
+  Expression: [1, 2] + [2, 1] == [3, 3]
+   Evaluated: [3, 3] == [3, 3]
 ```
 
 The `@test f(args...) key=val...` form is equivalent to writing
@@ -365,6 +370,8 @@ is a call using infix syntax such as approximate comparisons:
 ```jldoctest
 julia> @test π ≈ 3.14 atol=0.01
 Test Passed
+  Expression: ≈(π, 3.14, atol = 0.01)
+   Evaluated: ≈(π, 3.14; atol = 0.01)
 ```
 
 This is equivalent to the uglier test `@test ≈(π, 3.14, atol=0.01)`.
@@ -393,6 +400,8 @@ Test Broken
 
 julia> @test 2 + 2 ≈ 5 atol=1 broken=false
 Test Passed
+  Expression: ≈(2 + 2, 5, atol = 1)
+   Evaluated: ≈(4, 5; atol = 1)
 
 julia> @test 2 + 2 == 5 skip=true
 Test Broken
@@ -400,6 +409,8 @@ Test Broken
 
 julia> @test 2 + 2 == 4 skip=false
 Test Passed
+  Expression: 2 + 2 == 4
+   Evaluated: 4 == 4
 ```
 
 !!! compat "Julia 1.7"
@@ -604,7 +615,7 @@ function do_test(result::ExecutionResult, orig_expr)
         value = result.value
         testres = if isa(value, Bool)
             # a true value Passes
-            value ? Pass(:test, nothing, nothing, value) :
+            value ? Pass(:test, orig_expr, result.data, value, result.source) :
                     Fail(:test, orig_expr, result.data, value, result.source)
         else
             # If the result is non-Boolean, this counts as an Error
@@ -646,10 +657,12 @@ Note that `@test_throws` does not support a trailing keyword form.
 ```jldoctest
 julia> @test_throws BoundsError [1, 2, 3][4]
 Test Passed
+  Expression: ([1, 2, 3])[4]
       Thrown: BoundsError
 
 julia> @test_throws DimensionMismatch [1, 2, 3] + [1, 2]
 Test Passed
+  Expression: [1, 2, 3] + [1, 2]
       Thrown: DimensionMismatch
 ```
 """
@@ -707,7 +720,7 @@ function do_test_throws(result::ExecutionResult, orig_expr, extype)
             end
         end
         if success
-            testres = Pass(:test_throws, nothing, nothing, exc)
+            testres = Pass(:test_throws, orig_expr, extype, exc, result.source)
         else
             testres = Fail(:test_throws_wrong, orig_expr, extype, exc, result.source)
         end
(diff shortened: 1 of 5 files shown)


mmiller-max:
To retain the same output for a `Pass` as before this PR, `Base.show` needed to be modified. I _think_ the code that I removed from `Base.show` was never being used (as those fields were always `nothing`) but this needs confirming by someone with a wider knowledge than me.

This PR is ready for a code review.

mmiller-max:
Please can someone review this? Cheers

StefanKarpinski:
Giving a summary of the effect of this would be helpful. Who would be good to review? GitHub suggests @JeffBezanson, but that may just be because the did some sweeping change that touched this code.

timholy:
Maybe @oxinabox who filed the issue.

Would also be great to see some numbers on whether/how this affects the time to run Julia's own tests.

mmiller-max:
Thanks both for your comments.

I'll run a comparison on `make test-all` (if that's the best way to do it?) tonight (Australian time), and here's a brief summary of the effect:

With this PR, the information (tested expression and the value) of a succfessul test will be stored in the `Pass` `Result` rather than these fields being set to `nothing`. This will be the case for both `@test` and `@test_throws`. There will be no change to `DefaultTestSet`, as the current behaviour of `record(ts:DefaultTestSet, t::Pass)` discards any information in the `Pass` result and just increments the number of successful tests. Storing the information allows other testsets to use this information, for example in test report generation.

The storing of information in the `Pass` object means that `Base.show(io::IO, t::Pass)` has to be modified so that the output when a test passes is unchanged. The removed lines from this function print the Pass information if not `nothing`, which on master it always is and therefore the printing was never being done. I think removing these lines should have no effect.

mmiller-max:
Julia test times averaged across two runs:

With this PR: 42.5 mins
On a master commit (7c0cb30f5bc3bc2e0c520bcc302f713c5738de9a): 44.5 mins

The difference could be just what else was running on my machine at the same time, it would be useful if someone else could repeat the test so the results don't just rely on one laptop.

I have been struggling to actually run Julia's tests as the generating precompile statements step hangs (>1 hr) whilst building, so have had to resort to running `make test-all` inside a container. This in turns means the `sockets` and `file` tests fail, but hopefully its still representative. The same failures occur with and without the changes.

oxinabox:
>  Maybe @oxinabox who filed the issue.

Note that @mmiller-max is now working on the package that I was working on when I filed that issue.
(TestReports.jl)

I will review this PR presently

stdlib/Test/src/Test.jl
```diff
@@ -89,16 +89,9 @@ struct Pass <: Result
...
-    if !(t.orig_expr === nothing)
-        print(io, "\n  Expression: ", t.orig_expr)
-    end
     if t.test_type === :test_throws
         # The correct type of exception was thrown
         print(io, "\n      Thrown: ", typeof(t.value))
-    elseif t.test_type === :test && t.data !== nothing
-        # The test was an expression, so display the term-by-term
-        # evaluated version as well
-        print(io, "\n   Evaluated: ", t.data)
```
oxinabox, line 101: Why don't we want to display this?
Given someone has created a test result object, why not show its fields?

stdlib/Test/src/Test.jl
```diff
@@ -505,7 +498,7 @@ function do_test(result::ExecutionResult, orig_expr)
         value = result.value
         testres = if isa(value, Bool)
             # a true value Passes
-            value ? Pass(:test, nothing, nothing, value) :
+            value ? Pass(:test, orig_expr, result.data, value) :
```
oxinabox, line 501: I wonder if we would also like to store the `result.source` in `Pass`.
Since presumably the justification for why you would not want it is that it wasn't a failure.
But that is also a justification for why you wouldn't want the `orig_expr` and the `result.data`.
Which we do infact want.

oxinabox (commented):
I think this needs tests that it actually is storing the correct thing.

stdlib/Test/src/Test.jl
mmiller-max: I can see it being argued either way.

One the one hand the information is there so why not display it to give the user more information, but on the other hand if a test has passed the user may not want the details of the passing of that test to fill up the REPL. I typically only see the `Test Passed` output when manually stepping through tests to find the failing ones, and wouldn't be interested in knowing the details of the ones that passed. Normally the tests are in `testsets` and therefore the pass information isn't shown either way.

I kept it to the latter as that's what was originally happening before the changes here, albeit because that information wasn't stored so couldn't be displayed anyway.

I'm leaning towards not displaying it, but its definitely worth a discussion.

stdlib/Test/src/Test.jl
mmiller-max: Yes good point, I'll add that in

stdlib/Test/src/Test.jl
oxinabox: Fair enough. I am not sure when this is displayed, except if `@test` is used on its out, in the REPL.
 do `@testsets` call display on the tests with in them?

mmiller-max:
Bump.

I think the following two points need addressing to move this PR forward:

1. Deciding on the test output, as per [this](https://github.com/JuliaLang/julia/pull/36879#discussion_r468892636) discussion. Who needs to be involved with the decision?
2. Getting someone else to do a benchmark to check performance of Julia's tests.

Cheers

mmiller-max:
Bump

Just wondering if this has been discussed on a triage call or if that's still pending?

vtjnash:
This looks fine to me. We needed to make sure not to store the information (by default), but it didn't need to be removed this early. Please ping me when you rebase for review. For the questions:

1. Please remove the printing changes, as those aren't applicable to the rest of the PR and will let us merge this without issue.
2. If it passes on CI, that should be good enough for confirming performance is good.

mmiller-max:
@vtjnash thanks for following up. Rebased and ready for another review. Without the changes for printing there are quite a few doctest changes due to `Pass`es now containing information that can be printed.

vtjnash (approved):
lgtm

stdlib/Test/src/Test.jl
```diff
@@ -85,8 +85,9 @@ struct Pass <: Result
     orig_expr
     data
     value
-    function Pass(test_type::Symbol, orig_expr, data, thrown)
-        return new(test_type, orig_expr, data, thrown isa String ? "String" : thrown)
+    source::Union{Nothing,LineNumberNode}
+    function Pass(test_type::Symbol, orig_expr, data, thrown, source)
```
vtjnash, line 89: A package (Jute) was using the constructor directly in its tests. Should we add a default value (nothing)?

mmiller-max: Something like this?

```Julia
Pass(test_type::Symbol, orig_expr, data, thrown) = Pass(test_type, orig_expr, data, thrown, nothing)
```

Does it make more sense for Jute to update its use of the constructor if it's an internal API? Happy either way just wanted to check.

vtjnash: I think that makes sense, we generally try not to break packages unnecessarily. That can equivalently be written more concisely as:
```
function Pass(test_type::Symbol, orig_expr, data, thrown, source=nothing)
```

mmiller-max: Sounds good - I won't be able to get to it for a few weeks as I'm away from my computer but will do it then.

I'm not quite sure how your suggestions works, doesn't that need an `end` to work and you'd need to pass the arguments to `Pass`?

mmiller-max: finally in a position to look at this (sorry for delay) but think its been done now (#41424 ), so will resolve this comment.

PR #36880, opened 2020-08-02
fix formatting issue in docs

### LINKED ISSUE #36878: incorrect formatting on primitive types section, first warning label
here: https://docs.julialang.org/en/v1/manual/types/#Primitive-Types-1

OkonSamuel:
closes #36878

### DIFF
--- a/doc/src/manual/types.md
+++ b/doc/src/manual/types.md
@@ -238,6 +238,7 @@ of function arguments that are containers of abstract types; see [Performance Ti
 ## Primitive Types
 
 !!! warning
+
     It is almost always preferable to wrap an existing primitive type in a new
     composite type than to define your own primitive type.
 


StefanKarpinski:
Does this fix it? I would think that adding a line here would not help attach the warning to the text.

OkonSamuel:
> Does this fix it? I would think that adding a line here would not help attach the warning to the text.

Yes. It does. I faced something similar when generating https://github.com/cstjean/ScikitLearn.jl/edit/master/docs/src/man/models.md. 
You can also confirm this by comparing https://github.com/JuliaLang/julia/blob/master/doc/src/manual/types.md#primitive-types with https://github.com/OkonSamuel/julia/blob/master/doc/src/manual/types.md#primitive-types.

StefanKarpinski:
@mortenpi, is this a Documenter bug? I'm a bit confused by what the right way to format/fix this is.

mortenpi:
No, it's an indentation problem on the 1.5-branch. It's already fixed on master: https://docs.julialang.org/en/v1.6-dev/manual/types/#Primitive-Types-1

The correct fix would be backport #36254 to 1.5.

For what it's worth, as a style matter, I personally prefer having an empty line between the `!!!` line and the admonition body, so we _could_ merge this, even if it doesn't actually do anything for #36878.

Issue #36773, opened 2020-07-23
Mention flag in nonstandard string literals docs


jakobnissen:
You can define code like this:
```
macro foo_str(str, flag)
    # do stuff
end
```
And "call" this macro with the following syntax: `foo"str"flag`. The feature of the `flag` is not documented in the [section of the docs mentioning string literals](https://docs.julialang.org/en/v1.6-dev/manual/metaprogramming/#Non-Standard-String-Literals-1). But it would be nice if it was.

pret3nti0u5:
Hey, I would like to work on this issue. I am new to contributing to Open Source projects some guidance would be helpful.

jakobnissen:
@pret3nti0u5 That's great! There is a good tutorial for making a PR to Julia here: https://kshyatt.github.io/post/firstjuliapr/ . The tutorial covers adding new tests to Julia, but amending the docs is very similar. The file you want to change is `doc/src/manual/metaprogramming.md`.

Please ping me here on GitHub or PM me on the Julia slack if you need help :)

pret3nti0u5:
Here's the [PR](https://github.com/JuliaLang/julia/pull/36881).

JeffBezanson:
Fixed by #36881

PR #36881, opened 2020-08-02
Mentioned flag in nonstandard string literal docs

pret3nti0u5:
Added documentation for flags in nonstandard sting literal docs, pertaining to issue #36773

### DIFF
--- a/doc/src/manual/metaprogramming.md
+++ b/doc/src/manual/metaprogramming.md
@@ -1077,6 +1077,21 @@ but they are just functions, written entirely in Julia. You can read their sourc
 what they do -- and all they do is construct expression objects to be inserted into your program's
 syntax tree.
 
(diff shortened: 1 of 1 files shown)


doc/src/manual/metaprogramming.md
```diff
@@ -1077,6 +1077,21 @@ but they are just functions, written entirely in Julia. You can read their sourc
...
+    # do stuff
+end
+```
+This macro can then be called with the following syntax:
+
+```julia
+foo"str"flag
+```
+
+The type of flag in the above mentioned syntax would be a String with contents of whatever trails after the string literal.
```
StefanKarpinski, line 1093: Please quote `String` here.

StefanKarpinski (changes requested):

StefanKarpinski (approved):

PR #36887, opened 2020-08-03
Cleanup of warnings in documentation admonitions

Seelengrab:
https://github.com/JuliaLang/julia/pull/34670#issuecomment-667869199

Seems like they were missed, either when changing doc systems or when changing the keyword in the markdown parser.

If there are other places that I missed or other keywords that have not been changed, we can update those here as well.


Seelengrab:
This doesn't really need CI - is there a way I can mark this PR (or future ones, if necessary) not to run CI? I had a quick look through CONTRIBUTING.md, but didn't find anything.

PR #36888, opened 2020-08-03
various DCE-related bugfixes

yhls:
Various DCE-related bugfixes originally from #33730

@Keno @vchuravy


Keno (approved):

base/compiler/ssair/passes.jl
```diff
@@ -1096,41 +1125,32 @@ function cfg_simplify!(ir::IRCode)
...
     bb_starts[1] = 1
     for i = 1:length(result_bbs_lengths)
         bb_starts[i+1] = bb_starts[i] + result_bbs_lengths[i]
     end
-    # Figure out the pred and succ lists for each basic block in our merged result
-    cresult_bbs = let result_bbs = result_bbs,
-                      merged_succ = merged_succ,
-                      merge_into = merge_into,
-                      bbs = bbs,
-                      bb_rename_succ = bb_rename_succ
```
vtjnash, line 1109: Removing the `let` block prohibits inference of the types of the captured references, which could be a large performance penalty on the inner loops here.

Keno: Good catch, I remember adding this let block for that very reason

PR #31632, opened 2019-04-06
Partially revert "Remove specialized sort! method for PartialQuickSort{Int} (fixes #12833)"

kmsquire:
This reverts commit 9e8fcd30e9d7d4e88e94cfebec59778a71387b97, except for the
test.  The underlying issue(s) causing #12833 seem to have been fixed with recent
changes in inference.

PR #36896, opened 2020-08-03
Remove specialized PartialQuickSort{<:Integer} method

### LINKED ISSUE #34233: partialsort slower with single index than with range
When computing the value at an index near the end of a vector, `partialsort(v, length(v)-n:length(v))` is much faster than `partialsort(v, length(v)-n)`. I would have expected similar timings for both operations.
```julia
julia> using BenchmarkTools

julia> v = rand(10_000);

julia> @btime partialsort(v, 9_000:10_000);
  113.573 μs (3 allocations: 78.25 KiB)

julia> @btime partialsort(v, 9_000);
  891.562 μs (3 allocations: 78.22 KiB)
```

There doesn't seem to be any type instability (contrary to https://github.com/JuliaLang/julia/pull/12838), but looking at how the two methods progress, it appears that the range method finishes in 26 iterations, as it raises the lower bound (`lo`) after each call to `selectpivot!` since it has a branch for the case when the pivot is smaller than the beginning of the range:
https://github.com/JuliaLang/julia/blob/13d1bd4689e2afc12ddf5f48ad22efd13f01782a/base/sort.jl#L622-L623
That branch is taken for all but the 3 last iterations.
OTC the integer method doesn't have that branch, and it raises the lower bound much more slowly (even in the last iteration it's still at 2874). So it takes 38 iterations, each with a wide low-high range. Is a branch missing?

Cc: @kmsquire

nalimilan:
This specialized method is actually considerably slower than the one that also works for `PartialQuickSort{<:OrdinalRange}`. It had been removed before in 9e8fcd3 due to an inference issue, and reintroduced by c41d5a3 (#31632) once that issue got fixed, but apparently without benchmarks. It turns out that saving one comparison per iteration isn't worth it. The gain is especially large when looking for a value among the largest in the array, as the specialized method always sorted all values lower than the requested one (fixes #34233).

I must say I'm not completely sure why the specialized method is so slow. At least it's not due to a type instability.

Here are some benchmarks. I could only find one case (the 3rd one) where this PR is very slightly slower. In all other cases it's much faster (the gain could even be higher if looking for one of the few largest values in the array). I can add benchmarks to BaseBenchmarks if the PR sounds OK.
```
using BenchmarkTools
for len in (100, 10_000, 10_000_000),
    k in (len ÷ 9, len  ÷ 2, len - (len ÷ 9))
    @show len, k
    @btime partialsort!(x, $k) setup=(x=rand($len));
    @btime partialsort!(x, $k) setup=(x=rand(UInt8, $len));
    @btime partialsort!(x, $k) setup=(x=big.(rand($len)));
end

# master
(len, k) = (100, 11)
  558.107 ns (0 allocations: 0 bytes)
  91.105 ns (0 allocations: 0 bytes)
  3.441 μs (0 allocations: 0 bytes)
(len, k) = (100, 50)
  762.342 ns (0 allocations: 0 bytes)
  199.935 ns (0 allocations: 0 bytes)
  5.313 μs (0 allocations: 0 bytes)
(len, k) = (100, 89)
  1.537 μs (0 allocations: 0 bytes)
  327.360 ns (0 allocations: 0 bytes)
  9.592 μs (0 allocations: 0 bytes)
(len, k) = (10000, 1111)
  121.428 μs (0 allocations: 0 bytes)
  52.193 μs (0 allocations: 0 bytes)
  571.665 μs (0 allocations: 0 bytes)
(len, k) = (10000, 5000)
  517.929 μs (0 allocations: 0 bytes)
  240.547 μs (0 allocations: 0 bytes)
  2.030 ms (0 allocations: 0 bytes)
(len, k) = (10000, 8889)
  877.613 μs (0 allocations: 0 bytes)
  411.764 μs (0 allocations: 0 bytes)
  3.458 ms (0 allocations: 0 bytes)
(len, k) = (10000000, 1111111)
  242.663 ms (0 allocations: 0 bytes)
  83.886 ms (0 allocations: 0 bytes)
  3.427 s (0 allocations: 0 bytes)
(len, k) = (10000000, 5000000)
  981.472 ms (0 allocations: 0 bytes)
  302.894 ms (0 allocations: 0 bytes)
  11.731 s (0 allocations: 0 bytes)
(len, k) = (10000000, 8888889)
  1.605 s (0 allocations: 0 bytes)
  495.437 ms (0 allocations: 0 bytes)
  20.273 s (0 allocations: 0 bytes)

# This PR
(len, k) = (100, 11)
  555.332 ns (0 allocations: 0 bytes)
  87.643 ns (0 allocations: 0 bytes)
  3.536 μs (0 allocations: 0 bytes)
(len, k) = (100, 50)
  558.176 ns (0 allocations: 0 bytes)
  88.413 ns (0 allocations: 0 bytes)
  3.564 μs (0 allocations: 0 bytes)
(len, k) = (100, 89)
  563.253 ns (0 allocations: 0 bytes)
  62.689 ns (0 allocations: 0 bytes)
  3.081 μs (0 allocations: 0 bytes)
(len, k) = (10000, 1111)
  47.266 μs (0 allocations: 0 bytes)
  18.533 μs (0 allocations: 0 bytes)
  265.108 μs (0 allocations: 0 bytes)
(len, k) = (10000, 5000)
  93.967 μs (0 allocations: 0 bytes)
  17.005 μs (0 allocations: 0 bytes)
  434.462 μs (0 allocations: 0 bytes)
(len, k) = (10000, 8889)
  47.566 μs (0 allocations: 0 bytes)
  19.584 μs (0 allocations: 0 bytes)
  326.543 μs (0 allocations: 0 bytes)
(len, k) = (10000000, 1111111)
  65.597 ms (0 allocations: 0 bytes)
  32.917 ms (0 allocations: 0 bytes)
  960.879 ms (0 allocations: 0 bytes)
(len, k) = (10000000, 5000000)
  144.927 ms (0 allocations: 0 bytes)
  79.717 ms (0 allocations: 0 bytes)
  1.897 s (0 allocations: 0 bytes)
(len, k) = (10000000, 8888889)
  80.506 ms (0 allocations: 0 bytes)
  29.989 ms (0 allocations: 0 bytes)
  1.356 s (0 allocations: 0 bytes)
```

### DIFF
--- a/base/sort.jl
+++ b/base/sort.jl
@@ -616,31 +616,8 @@ function sort!(v::AbstractVector, lo::Integer, hi::Integer, a::MergeSortAlg, o::
     return v
 end
 
-function sort!(v::AbstractVector, lo::Integer, hi::Integer, a::PartialQuickSort{<:Integer},
+function sort!(v::AbstractVector, lo::Integer, hi::Integer, a::PartialQuickSort,
                o::Ordering)
-    @inbounds while lo < hi
-        hi-lo <= SMALL_THRESHOLD && return sort!(v, lo, hi, SMALL_ALGORITHM, o)
-        j = partition!(v, lo, hi, o)
-        if j >= a.k
-            # we don't need to sort anything bigger than j
-            hi = j-1
-        elseif j-lo < hi-j
-            # recurse on the smaller chunk
-            # this is necessary to preserve O(log(n))
-            # stack space in the worst case (rather than O(n))
-            lo < (j-1) && sort!(v, lo, j-1, a, o)
-            lo = j+1
-        else
-            (j+1) < hi && sort!(v, j+1, hi, a, o)
-            hi = j-1
-        end
-    end
-    return v
-end
-
-
-function sort!(v::AbstractVector, lo::Integer, hi::Integer, a::PartialQuickSort{T},
-               o::Ordering) where T<:OrdinalRange
     @inbounds while lo < hi
         hi-lo <= SMALL_THRESHOLD && return sort!(v, lo, hi, SMALL_ALGORITHM, o)
         j = partition!(v, lo, hi, o)
@@ -650,6 +627,9 @@ function sort!(v::AbstractVector, lo::Integer, hi::Integer, a::PartialQuickSort{
         elseif j >= last(a.k)
             hi = j-1
         else
+            # recurse on the smaller chunk
+            # this is necessary to preserve O(log(n))
+            # stack space in the worst case (rather than O(n))
             if j-lo < hi-j
                 lo < (j-1) && sort!(v, lo, j-1, a, o)
                 lo = j+1
--- a/test/sorting.jl
+++ b/test/sorting.jl
@@ -289,27 +289,40 @@ Base.step(r::ConstantRange) = 0
     end
 
     @testset "unstable algorithms" begin
-        for alg in [QuickSort, PartialQuickSort(length(a))]
-            b = sort(a, alg=alg)
-            @test issorted(b)
-            b = sort(a, alg=alg, rev=true)
-            @test issorted(b, rev=true)
-            b = sort(a, alg=alg, by=x->1/x)
-            @test issorted(b, by=x->1/x)
-        end
+        b = sort(a, alg=QuickSort)
+        @test issorted(b)
+        @test last(b) == last(sort(a, alg=PartialQuickSort(length(a))))
+        b = sort(a, alg=QuickSort, rev=true)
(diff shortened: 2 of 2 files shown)


nalimilan:
Bump.

kmsquire (commented):
Late review, but LGTM.

nalimilan:
@kmsquire Do you have any idea why there used to be a specialize method which was actually slower? Was there a mistake in the algorithm?

kmsquire:
I’m not sure. I’ll try to look at it a bit closer tomorrow.

PR #36900, opened 2020-08-03
Add `parse()` and `tryparse()` for `VersionNumber`s

staticfloat:
This allows for a more consistent approach with `VersionNumbers` as
compared to other types that we may be parsing.

### DIFF
--- a/base/version.jl
+++ b/base/version.jl
@@ -121,6 +121,18 @@ function VersionNumber(v::AbstractString)
     return VersionNumber(major, minor, patch, prerl, build)
 end
 
+parse(::Type{VersionNumber}, v::AbstractString) = VersionNumber(v)
+function tryparse(::Type{VersionNumber}, v::AbstractString)
+    try
+        return VersionNumber(v)
+    catch e
+        if isa(e, InterruptException)
+            rethrow(e)
+        end
+        return nothing
+    end
+end
+
 """
     @v_str
 
--- a/test/version.jl
+++ b/test/version.jl
@@ -79,6 +79,12 @@ using Random
 
 @test_throws ArgumentError VersionNumber(4, 3, 2, (), ("", 1))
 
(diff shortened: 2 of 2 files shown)


base/version.jl
```diff
@@ -121,6 +121,18 @@ function VersionNumber(v::AbstractString)
     return VersionNumber(major, minor, patch, prerl, build)
 end
 
+parse(::Type{VersionNumber}, v::AbstractString) = VersionNumber(v)
+function tryparse(::Type{VersionNumber}, v::AbstractString)
```
fredrikekre, line 125: Somewhat OT, but maybe we could have a fallback `tryparse` method. I have found times where only `parse` is implemented but I would like to `tryparse`.
```
function tryparse(::Type{T}, s::AbstractString) where T
    try
        return parse(T, s)
    catch e
        return nothing
    end
end
```

staticfloat: Agreed; although we should still handle `InterruptException` there.

JeffBezanson: I would make it as narrow as possible, and only handle `ArgumentError` and rethrow everything else.

KristofferC: I don't think there such be a fallback method that relies on `try catch` like that. Even by only handling `ArgumentError`, you don't know that it's the right one.

staticfloat: Without more structure imposed upon our exceptions handling, I don't think it's possible to really determine what a `parse()` method would or would not throw, so perhaps a generic `tryparse()` is not the right choice.  If we do have one, I think there's an argument to be made for swallowing all exceptions that could conceivably be thrown from within it (`InterruptException` is not one of those, hence why I suggest rethrowing it) but in any case, I'm not very satisfied by that either.  One would hope that `parse()` methods wouldn't be doing anything too complicated inside of them, but sadly we don't have any way to enforce that.

PR #36917, opened 2020-08-04
Add micro editor support

jakobnissen:
It's not clear to me that Base should offer specific support for all kinds of terminal-based editors, but as long at it does, it should at least support $MY_FAVORITE_EDITOR :)

This PR allows the `@edit` macro to work with the [Micro editor](https://micro-editor.github.io/) out of the box.


StefanKarpinski:
Why not?

image

Issue #35485, opened 2020-04-15
isequal and hash are incompatible for UniformScaling


tkf:
```julia
julia> isequal(diagm(ones(3)), I)
true

julia> isequal(diagm(ones(4)), I)
true

julia> hash(I)
0x6bd2535d045989b2

julia> hash(diagm(ones(4)))
0x6ba2635b5ca237d8

julia> hash(diagm(ones(3)))
0xf728e5a95d43274d
```

This is because `isequal` on `UniformScaling` is not defined and dispatches to the generic method

```julia
julia> @which isequal(diagm(ones(3)), I)
isequal(x, y) in Base at operators.jl:123
```

which calls `==`:

https://github.com/JuliaLang/julia/blob/b37cc3d7c03d0496c65431d621f5d6b7939f8089/base/operators.jl#L123

It seems `isequal` on `UniformScaling` has to be defined manually.

A more fundamental solution may be to define `isequal` in terms of the identity `===` by default.  It seems brittle to define a function with strict semantics (`isequal`) in terms of the one with weaker semantics (`==`).  (Though this is probably impossible to do within 1.x.)

StefanKarpinski:
The fundamental issue is that `==` is non-transitive in the presence of `I` which is a bad situation:
```jl
julia> I(3) == I == I(4)
true

julia> I(3) == I(4)
false
```
The best solution is unclear, however. It seems that it's either this or make `I(n) == I` false.

StefanKarpinski:
I want to slate this for fixing in 1.6 since this is the only example of equality being non-transitive that I'm aware of.

StefanKarpinski:
Triage agrees that `I(n) == I` should not be true, so now we "just" need a PR to change this.

StefanKarpinski:
Once there's a PR, we'll need a PkgEval run.

andreasnoack:
While I think it's probably right to change the behavior here, the discussion reminds me that there are some more general challenges associated with the "`x==y` implies `hash(x) == hash(y)`" orthodoxy https://github.com/JuliaParallel/DistributedArrays.jl/issues/4#issuecomment-89590255

PR #36918, opened 2020-08-04
Make UnformScaling inequal to matrices

jakobnissen:
See issue JuliaLang/LinearAlgebra.jl#716 

This simply deletes the specialized code that previously made `UniformScaling` equal to identity matrices, such that it now just defaults back to the default implementation of `==`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,6 +11,9 @@ New language features
 
 Language changes
 ----------------
+* Instances of `UniformScaling` can no longer compare equal to diagonal matrices.
+  Previous behavior violated the transitivity of `==` since previously,
+  `UniformScaling` compared equal to diagonal matrices of different sizes.
 
 
 Compiler/Runtime improvements
--- a/stdlib/LinearAlgebra/src/uniformscaling.jl
+++ b/stdlib/LinearAlgebra/src/uniformscaling.jl
@@ -305,25 +305,6 @@ end
 
 ==(J1::UniformScaling,J2::UniformScaling) = (J1.λ == J2.λ)
 
-## equality comparison with UniformScaling
-==(J::UniformScaling, A::AbstractMatrix) = A == J
-function ==(A::AbstractMatrix, J::UniformScaling)
-    require_one_based_indexing(A)
-    size(A, 1) == size(A, 2) || return false
-    iszero(J.λ) && return iszero(A)
-    isone(J.λ) && return isone(A)
-    return A == J.λ*one(A)
-end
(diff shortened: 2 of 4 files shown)


mbauman:
I think it'd be worth putting a compat annotation on the `UniformScaling` docstring itself.

andreasnoack:
I'd expect this to be pretty breaking so let's not forget https://github.com/JuliaLang/julia/issues/35485#issuecomment-659604677

mbauman:

nanosoldier:
[Your package evaluation job](https://github.com/JuliaLang/julia/pull/36918#issuecomment-670553186) has completed - possible new issues were detected. A full report can be found [here](https://github.com/JuliaCI/NanosoldierReports/blob/fb40dd23b5d8962295ec32d3d4920fc0d3b673e0/pkgeval/by_hash/77cab85_vs_7c0cb30/report.md). cc @maleadt

jakobnissen:
I'm not sure what else needs to be done here for this to be merged.

mbauman:
Nearly all of those PkgEval failures are indeed related — and unfortunately it's not just testing code.  DiffEq looks for `==I` in detecting mass matrices in its src.  Here's an incomplete regex search: [`==\s*\d*\*?I\b`](https://juliahub.com/ui/RepoSearch?q=%3D%3D%5Cs%2a%5Cd%2a%5C%2a%3FI%5Cb&r=true)

The question now is simply: are those breakages small enough for a minor change?  We can of course help the open source packages out, but they're representative of closed-source codebases, too.

StefanKarpinski:
That does, unfortunately, seem like a lot of breakage for a 1.x release. Here's an intermediate change we probably could make, however:

* Leave `I(3) == I` alone
* Change `isequal(I(3), I)` to be false to match hashing

That way at least we fix https://github.com/JuliaLang/julia/issues/35485, which cannot otherwise be fixed without making all identity matrices equal to each other, which seems very bad. It seems unlikely that people are using `isequal` to compare `I` with matrices like this or putting `I` into sets of dictionaries. Then we can slate changing `==` to agree with `isequal` for being a 2.0 change.

StefanKarpinski:
Triage agrees with my plan to leave `==` alone and change `isequal` for now. The full change can be done in 2.0.

JeffBezanson:
Agree with just changing `isequal`; that's the one that needs to be as well-behaved as possible.

jakobnissen:
Closed in #37525

PR #36920, opened 2020-08-04
Some cleanup and comment for addrspace tracking in codegen

yuyichao:
The first two commits should be purely spelling change/clean up.

* NULL pointer in the tracked address space appears to be used more often than the one in default addrspace.
* Remove implicit parameters from a few related functions that used macro "hack". Those are leftover from when the context is a global state AFAICT.

The last commit deals exclusively with `maybe_decay_tracked`.

1. Unify and document the returned LLVM type for a few common functions.
2. Require `boxed` and the `.Vboxed` fields to be `T_prjlvalue`.

    AFAICT many places are assuming this already. So what I've changed is to document this and add a few assertions to make sure this invariance hold when it's unclear from just local information.

3. Remove the "maybe"-ness of the cast in most places.

    Turns out that almost all the usecases have known input types.

4. Cleanup `maybe_decay_tracked` itself.

    AFAICT, there's no use of the function that uses the `T_pprjlvalue` branch. It should also be nice to reduce reliance on the eltype of pointers.

### DIFF
--- a/src/intrinsics.cpp
+++ b/src/intrinsics.cpp
@@ -861,6 +861,7 @@ static jl_cgval_t emit_ifelse(jl_codectx_t &ctx, jl_cgval_t c, jl_cgval_t x, jl_
                 if (!y_vboxed)
                     y_vboxed = ConstantPointerNull::get(cast<PointerType>(x_vboxed->getType()));
                 ret.Vboxed = ctx.builder.CreateSelect(isfalse, y_vboxed, x_vboxed);
+                assert(ret.Vboxed->getType() == T_prjlvalue);
             }
             return ret;
         }
--- a/src/intrinsics.cpp
+++ b/src/intrinsics.cpp
@@ -804,8 +804,8 @@ static jl_cgval_t emit_ifelse(jl_codectx_t &ctx, jl_cgval_t c, jl_cgval_t x, jl_
                 ifelse_tbaa = x.tbaa;
             }
             else {
-                x_ptr = decay_derived(x_ptr);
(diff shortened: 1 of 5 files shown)


src/intrinsics.cpp
```diff
@@ -499,7 +499,7 @@ static jl_cgval_t generic_bitcast(jl_codectx_t &ctx, const jl_cgval_t *argv)
         return mark_julia_type(ctx, vx, false, bt);
     }
     else {
-        Value *box = emit_allocobj(ctx, nb, boxed(ctx, bt_value));
+        Value *box = emit_allocobj(ctx, nb, literal_pointer_val(ctx, (jl_value_t*)bt));
```
vtjnash, line 502: `bt` isn't a type of the object (it's a partial supertype)

vtjnash (approved):

src/intrinsics.cpp
yuyichao: Reverted. I thought the primitive check above filters out ones with free parameters but seems not (and I missed the branch right above).

The change was during one of my attempt to eliminate one of the last few maybe's which I subsequently gave up so it's not necessary here....

Keno (approved):

yuyichao:
Any use of v_rnull can just be replaced with how it is defined here, or in case it was changed from a equivalent value just leave it unchanged.

KristofferC:
This I got them: https://github.com/JuliaLang/julia/compare/b3ba2340298..dbf74365f0d. Feel free to look it over.

PR #36925, opened 2020-08-05
use more === in tfuncs

oxinabox:
I don't know that this will have much effect on performance.
In general compiler reasons better about `===` but since this whole this in in `@nospecialize` idk about that.
Also it avoids potential invalidations, but since it is `Core.==` it is probably not invalidate.
But still it is nice for consistency.
We mostly use `===` throughout this file, this changed to almost entirely use that.

It also tweaks a few comments,
and makes a bit of spacing more consistent with the rest of the file.

(I was just reading though the file looking to understand it better)

### DIFF
--- a/base/compiler/tfuncs.jl
+++ b/base/compiler/tfuncs.jl
@@ -242,7 +242,7 @@ end
 add_tfunc(===, 2, 2, egal_tfunc, 1)
 
 function isdefined_nothrow(argtypes::Array{Any, 1})
-    length(argtypes) == 2 || return false
+    length(argtypes) === 2 || return false
     return typeintersect(widenconst(argtypes[1]), Module) === Union{} ?
         (argtypes[2] ⊑ Symbol || argtypes[2] ⊑ Int) :
          argtypes[2] ⊑ Symbol
@@ -320,8 +320,8 @@ function sizeof_nothrow(@nospecialize(x))
             return isinline
         end
         isa(x, DataType) || return false
-        x.layout == C_NULL && return false
-        (datatype_nfields(x) == 0 && !datatype_pointerfree(x)) && return false
+        x.layout === C_NULL && return false
+        (datatype_nfields(x) === 0 && !datatype_pointerfree(x)) && return false
         return true
     else
         x = widenconst(x)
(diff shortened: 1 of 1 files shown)


base/compiler/tfuncs.jl
```diff
@@ -717,7 +718,7 @@ function getfield_nothrow(@nospecialize(s00), @nospecialize(name), @nospecialize
         s.abstract && return false
         # If all fields are always initialized, and bounds check is disabled, we can assume
         # we don't throw
-        if bounds_check_disabled && !isvatuple(s) && s.name !== NamedTuple.body.body.name && fieldcount(s) == s.ninitialized
+        if bounds_check_disabled && !isvatuple(s) && s.name !== NamedTuple.body.body.name && fieldcount(s) === s.ninitialized
```
vtjnash, line 721: This makes it more brittle:
```
julia> typeof(Int).ninitialized |> typeof
Int32

julia> fieldcount(Int) |> typeof
Int64
```

base/compiler/tfuncs.jl
```diff
@@ -320,8 +320,8 @@ function sizeof_nothrow(@nospecialize(x))
             return isinline
         end
         isa(x, DataType) || return false
-        x.layout == C_NULL && return false
-        (datatype_nfields(x) == 0 && !datatype_pointerfree(x)) && return false
+        x.layout === C_NULL && return false
+        (datatype_nfields(x) === 0 && !datatype_pointerfree(x)) && return false
```
vtjnash, line 324: Same here:
```
julia> Base.datatype_nfields(Int) |> typeof
UInt32
```

base/compiler/tfuncs.jl
oxinabox: Huh, TIL.
i thought both of these were certain to be `Int`

Keno:
I don't think there is much point to changing all the integer comparisons to `===`. The fact that it's an integer on either side should be well inferred.

vtjnash:
Hopefully nobody is redefining `Core.Compiler.:(+)(::Int, ::Int)`, so while this isn't wrong, it's just not really an improvement either.

PR #36926, opened 2020-08-05
Avoid constructing an array only to splat it to make a tuple in tuple_tfunc

oxinabox:


base/compiler/tfuncs.jl
```diff
@@ -1262,7 +1262,7 @@ function tuple_tfunc(atypes::Vector{Any})
         end
     end
     if all_are_const
-        return Const(tuple(Any[atypes[i].val for i in 1:length(atypes)]...))
+        return Const(Tuple(atypes[i].val for i in 1:length(atypes)))
```
oxinabox, line 1265: alternatively:
```suggestion
        return Const(ntuple(i -> atypes[i].val, length(atypes)))
```

vtjnash:
Those both still construct it (can't avoid that as it's needed for dispatch), they would just also make a closure and allocate some other objects too.

oxinabox:
I made a similar case to time in the REPL, not sure that it is completely valid, but it seems like it is much faster to use `ntuple`.
```julia
julia> foo_vecsplat(x) = tuple(Any[x[i] for i in 1:length(x)]...)
foo_vecsplat (generic function with 1 method)

julia> foo_ntuple(x) = ntuple(i->x[i], length(x))
foo_ntuple (generic function with 1 method)

julia> foo_generator(x) = Tuple(x[i] for i in 1:length(x))
foo_generated (generic function with 1 method)

julia> @btime foo_vecsplat([1,2,3,4,5,6,7,8]);
  244.023 ns (3 allocations: 368 bytes)

julia> @btime foo_ntuple([1,2,3,4,5,6,7,8]);
  32.596 ns (2 allocations: 224 bytes)

julia> @btime foo_generator([1,2,3,4,5,6,7,8]);
  286.795 ns (3 allocations: 368 bytes)
```

It seems like it still holds with `@nospecialize`
```julia
julia> bar_ntuple(@nospecialize x) = ntuple(i->x[i], length(x))
bar_ntuple (generic function with 1 method)

julia> @btime bar_ntuple([1,2,3,4,5,6,7,8]);
  32.676 ns (2 allocations: 224 bytes)
```

vtjnash:
That's not quite valid. Using `ntuple` is a bit faster, but not by quite that much (it's the same implementation as vecsplat, but additionally has optimizations for less than 10 elements):
```
julia> @btime foo_ntuple($(Any[1,2,3,4,5,6,7,8]));
  157.234 ns (1 allocation: 80 bytes)
julia> @btime foo_vecsplat($(Any[1,2,3,4,5,6,7,8]));
  304.480 ns (2 allocations: 224 bytes)
```

oxinabox:
> ... but additionally has optimizations for less than 10 elements ...

Are those optimisations relevent here?
Trying to familarise myself more with this code

JeffBezanson:
Tuples are usually small, so that's the case we're likely to hit here. But, this code probably does not run often enough in the compiler to have an impact.

vtjnash (approved):
Yep. But since you've done the work to identify it, seems sensible to merge.

PR #36929, opened 2020-08-05
Correct TSAN tasking integration

Keno:
In order for tsan to work, any setjmp/longjmp must be executed while
that task's tsan state is active. As a result, we cannot switch the
tsan state until said setjmp is completed, but need to do it
(as basically the only thing happening) between the setjmp and the
subsequent longjmp. To facilitate this without too much disruption,
move the tsan state into the jl_ucontext_t, which seems appropriate
since it's additional state that needs to be restored on context
switch.

Also forbid using TSAN from Clang < 11, where the runtime library
has bugs that cause us to exhaust the maximum number of allowed
mappings.

With this, TSAN works, but obviously has a fairly large number of complaints.
For reference, here is my Make.user:
```
USE_BINARYBUILDER_LLVM=0
USECLANG=1
SANITIZE=1
SANITIZE_THREAD=1
JULIA_PRECOMPILE=0
override CC=/home/keno/llvm-project-build/bin/clang
override CXX=/home/keno/llvm-project-build/bin/clang++
```


Make.inc
```diff
@@ -431,6 +431,10 @@ CXX_DISABLE_ASSERTION := -DJL_NDEBUG
 DISABLE_ASSERTIONS := -DNDEBUG -DJL_NDEBUG
 endif
 
+ifeq ($(LLVM_ASSERTIONS),0)
+CXX_DISABLE_ASSERTION += -DNDEBUG
+endif
+
```
vtjnash, line 437: As this variable does not have the actual runtime value (which should get set from llvm-config), we have a header (julia_assert.h) that is supposed to already deal with this.

src/gc-stacks.c
```diff
@@ -226,6 +226,12 @@ void sweep_stack_pools(void)
                     t->stkbuf = NULL;
                     _jl_free_stack(ptls2, stkbuf, bufsz);
                 }
+#ifdef JL_TSAN_ENABLED
+                if (t->ctx.tsan_state) {
+                    t->ctx.tsan_state = NULL;
+                    __tsan_destroy_fiber(t->ctx.tsan_state);
```
vtjnash, line 232: This is a NULL pointer?
```suggestion
                    __tsan_destroy_fiber(NULL);
```

src/task.c
```diff
@@ -279,6 +277,10 @@ JL_DLLEXPORT jl_task_t *jl_get_next_task(void)
     return ptls->current_task;
 }
 
+#ifdef JL_TSAN_ENABLED
+char tsan_state_corruption[] = "TSAN state corrupted. Exiting HARD!\n";
```
vtjnash, line 281:
```suggestion
const char tsan_state_corruption[] = "TSAN state corrupted. Exiting HARD!\n";
```

src/task.c
```diff
@@ -290,6 +292,17 @@ static void ctx_switch(jl_ptls_t ptls)
     // none of these locks should be held across a task switch
     assert(ptls->locks.len == 0);
 
+#ifdef JL_TSAN_ENABLED
+    if (lastt->ctx.tsan_state != __tsan_get_current_fiber()) {
+        // Something went really wrong - don't even assume that we can
+        // use assert/abort which involve lots of signal handling that
+        // looks at the tsan state.
+        write(STDERR_FILENO, tsan_state_corruption, sizeof(tsan_state_corruption));
```
vtjnash, line 300:
```suggestion
        write(STDERR_FILENO, tsan_state_corruption, sizeof(tsan_state_corruption) - 1);
```

vtjnash (approved):

Make.inc
Keno: Looks like `llvm-config` no longer sets `NDEBUG`, but it still needs to match what LLVM expects for ABI reasons.

src/task.c
```diff
@@ -343,50 +356,75 @@ static void ctx_switch(jl_ptls_t ptls)
...
+        {
             sanitizer_start_switch_fiber(t->stkbuf, t->bufsz);
-            jl_swap_fiber(lastt_ctx, &t->ctx);
-            sanitizer_finish_switch_fiber();
+            if (killed) {
+                tsan_switch_to_ctx(&t->ctx);
+                tsan_destroy_ctx(ptls, &lastt->ctx);
+                jl_set_fiber(&t->ctx); // (doesn't return)
+                abort(); // unreachable
+            } else {
```
JeffBezanson, line 386: Style: line break between `}` and `else`

PR #36942, opened 2020-08-06
Make +(::AbstractArray) return a copy

### LINKED ISSUE #33271: Inconsistency with unary `+` for sparse arrays
For dense arrays, the unary `+` operator creates a copy, for sparse arrays it doesn’t:

```julia
julia> using SparseArrays

julia> A = rand(5,5);

julia> B = sprand(5,5,0.5);

julia> +A === A
false

julia> +B === B
true
```
I think this should be made consistent. I prefer the copying behavior, since it matches the behavior of the unary `-` (which obviously can’t return the same object).

sostock:
Fixes #33271.

### DIFF
--- a/base/abstractarraymath.jl
+++ b/base/abstractarraymath.jl
@@ -94,7 +94,7 @@ conj!(x::AbstractArray{<:Real}) = x
 real(x::AbstractArray{<:Real}) = x
 imag(x::AbstractArray{<:Real}) = zero(x)
 
-+(x::AbstractArray{<:Number}) = x
++(x::AbstractArray{<:Number}) = copy(x)
 *(x::AbstractArray{<:Number,2}) = x
 
 # index A[:,:,...,i,:,:,...] where "i" is in dimension "d"
--- a/stdlib/LinearAlgebra/src/adjtrans.jl
+++ b/stdlib/LinearAlgebra/src/adjtrans.jl
@@ -262,7 +262,10 @@ Broadcast.broadcast_preserving_zero_d(f, tvs::Union{Number,TransposeAbsVec}...) 
 
 ### linear algebra
 
-(-)(A::Adjoint)   = Adjoint(  -A.parent)
+(+)(A::Adjoint) = Adjoint(+A.parent)
+(-)(A::Adjoint) = Adjoint(-A.parent)
+
+(+)(A::Transpose) = Transpose(+A.parent)
 (-)(A::Transpose) = Transpose(-A.parent)
 
 ## multiplication *
--- a/stdlib/LinearAlgebra/test/adjtrans.jl
+++ b/stdlib/LinearAlgebra/test/adjtrans.jl
@@ -557,4 +557,14 @@ end
     @test transpose(Int[]) * Int[] == 0
 end
 
+@testset "unary plus creates a copy (issue #33271)" begin
+    P = rand(3,3)
+    for A = (adjoint(P), transpose(P))
+        pA = +A
+        @test typeof(pA) === typeof(A)
+        @test pA == A
+        @test pA !== A
+    end
+end
+
 end # module TestAdjointTranspose
--- a/stdlib/LinearAlgebra/test/bidiag.jl
+++ b/stdlib/LinearAlgebra/test/bidiag.jl
@@ -375,6 +375,13 @@ Random.seed!(1)
             @test typeof(Diag*SymTri)  <: Tridiagonal
         end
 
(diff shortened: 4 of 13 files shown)


JeffBezanson:
This is needlessly inefficient, and adds a new dependency on `copy` that might break some array types. If this really must make a copy (but do you really want people writing `+A` as a "clever" way to get a copy?), I think a good way to do it is to implement it the same way as unary `-` everywhere. For example we have this in arraymath.jl:

```
for f in (:-, :conj, :real, :imag)
    @eval ($f)(A::AbstractArray) = broadcast_preserving_zero_d($f, A)
end
```

`:+` could be added there, which should hopefully make it more generic.

Interestingly, the reason `+(::Array)` makes a copy now is this:
```
for f in (:+, :-)
    @eval function ($f)(A::AbstractArray, B::AbstractArray)
        promote_shape(A, B) # check size compatibility
        broadcast_preserving_zero_d($f, A, B)
    end
end

function +(A::Array, Bs::Array...)
    for B in Bs
        promote_shape(A, B) # check size compatibility
    end
    broadcast_preserving_zero_d(+, A, Bs...)
end
```

The `Array` method there matches one argument, and it's not clear to me whether that was intentional. I also wonder whether the `AbstractArray` methods could be extended to multiple arguments instead. cc @mbauman @martinholters

sostock:
Initially, I wanted to implement this by adding `:+` to the list above, but this would also change the return type of `+(::AbstractArray{Bool})` (since `+(::Bool)` returns an `Int`). I didn’t want to make both changes in the same PR.

I am also fine with the non-copying behavior, I just think it should be consistent.

martinholters:
I don't think `+(::Array)` making a copy was intentional per se, and I don't think making `+(::AbstractArray)` always return a copy just for the sake of it is a good move. Generally, I'd expect `+x` to be a no-op.

However, given that `+x` is not a no-op if `x::Bool`, I weakly favor `+(::AbstractArray{Bool})` to return an `AbstractArray{Int}`. But then, why should `Bool` be special-cased here? How to figure out whether `+(::MyType)` is a no-op and consequently, whether `+(x::AbstractArray{MyType})` can just return `x`? E.g. If we accept `+(::Array{Bool})::Array{Int}`, I think we really also want `+(::Array{Array{Bool}})::Array{Array{Int}}`. And then there is the question of abstract element types: `+Number[true]` currently returns `Int[1]`, while `+Number[true, 1.0]` returns a `Vector{Real}`, and I think using the broadcast rules for determining the return type here is the only way if we accept `+x` not to be a no-op.

So ideally, I'd like the following to be true: `+x` and `.+x` are equivalent except that `x === +x` (i.e. avoid copying) if possible. And while it might be possible to codify that rule, I'm not certain it is possible to do so while making the compiler recognize the no-op cases as such.

So just always returning the argument might be the best compromise after all.

martinholters:
> And while it might be possible to codify that rule, I'm not certain it is possible to do so while making the compiler recognize the no-op cases as such.

Actually, there might be hope:
```julia
function foo(x)
    # empty x would need a special case, too
    T = @inbounds typeof(+x[firstindex(x)])
    for xi in x
        if xi !== +xi
            return .+x
        end
        T = Base.promote_typejoin(T, typeof(xi))
    end
    if T != eltype(x)
        return .+x
    end
    return x
end
```
encodes what I have in mind and...
```julia
julia> @code_native foo(zeros(4))
	.text
; ┌ @ REPL[56]:1 within `foo'
	movq	%rsi, -8(%rsp)
	movq	(%rsi), %rax
; └
; ┌ @ REPL[56]:5 within `foo'
	retq
; └
```
Of course, the `return .+x` should be replaced with something that does not apply `+` a second time to the elements already processed, so this needs more tweaking and I'm not sure it's worth it, but it might be doable.

JeffBezanson:
Ah, the `Bool` case is helpful.

> and I think using the broadcast rules for determining the return type here is the only way if we accept `+x` not to be a no-op.

That seems like the best option to me. We should just implement unary `+` the same way as `-`. Even though it's a more significant change, this inconsistency should probably be fixed:
```
julia> +[true]
1-element Array{Int64,1}:
 1

julia> .+Diagonal([true])
1×1 Diagonal{Int64,Array{Int64,1}}:
 1

julia> +Diagonal([true])
1×1 Diagonal{Bool,Array{Bool,1}}:
 1
```

DilumAluthge:

@sostock If you're still interested in working on this PR, let me know, and I can re-open it.

PR #36948, opened 2020-08-06
compiler: fixes problematic abstract eval passes:

aviatesk:
- `abstract_eval_value_expr` will no longer return `nothing`
- fix problematic `isdefined` check within `abstract_eval_statement`

@vtjnash could you have a review on this ?

### DIFF
--- a/base/compiler/abstractinterpretation.jl
+++ b/base/compiler/abstractinterpretation.jl
@@ -1071,6 +1071,7 @@ function abstract_eval_value_expr(interp::AbstractInterpreter, e::Expr, vtypes::
         if 1 <= n <= length(sv.sptypes)
             t = sv.sptypes[n]
         end
+        return t
     elseif e.head === :boundscheck
         return Bool
     else
@@ -1202,7 +1203,7 @@ function abstract_eval_statement(interp::AbstractInterpreter, @nospecialize(e), 
                 t = Const(true) # definitely assigned previously
             end
         elseif isa(sym, Symbol)
-            if isdefined(sv.mod, sym.name)
+            if isdefined(sv.mod, sym)
                 t = Const(true)
             end
         elseif isa(sym, GlobalRef)


base/compiler/abstractinterpretation.jl
```diff
@@ -1202,7 +1203,7 @@ function abstract_eval_statement(interp::AbstractInterpreter, @nospecialize(e),
                 t = Const(true) # definitely assigned previously
             end
         elseif isa(sym, Symbol)
-            if isdefined(sv.mod, sym.name)
```
aviatesk, line 1205: This seems to be an obvious bug, but maybe the current abstract interpretation won't touch this pass and so any bug from this happens so far ? Well I believe this is better to be fixed anyway.

base/compiler/abstractinterpretation.jl
```diff
@@ -1071,6 +1071,7 @@ function abstract_eval_value_expr(interp::AbstractInterpreter, e::Expr, vtypes::
         if 1 <= n <= length(sv.sptypes)
             t = sv.sptypes[n]
         end
+        return t
```
aviatesk, line 1074: Currrently `abstract_eval_value_expr` will return `nothing` when the `if` condition above is `false`; I guess this function is better to return types instead of that, right ? Not sure what will actually happen when this returns `nothing` though.

base/compiler/abstractinterpretation.jl
Keno: I don't think we allow raw Symbols anymore to refer to globals. Only `GlobalRef`. Should probably delete branch entirely. @JeffBezanson?

base/compiler/abstractinterpretation.jl
Keno: Whoops, thanks!

Keno:
Quick tests shows that we are still using the symbol form, so LGTM.

base/compiler/abstractinterpretation.jl
JeffBezanson: We do still use this form, but we always convert it to `GlobalRef` if we're going to call inference.

aviatesk:
> Quick tests shows that we are still using the symbol form, so LGTM.
 
Would you mind sharing an example to test that ? I couldn't make a MRE for that pass.

Keno:
> Would you mind sharing an example to test that ? I couldn't make a MRE for that pass.

I tried the following diff, which fails tests:
```
diff --git a/base/compiler/abstractinterpretation.jl b/base/compiler/abstractinterpretation.jl
index 9461effc93..5d86cf19b9 100644
@@ -1201,10 +1201,6 @@ function abstract_eval_statement(interp::AbstractInterpreter, @nospecialize(e),
             elseif !vtyp.undef
                 t = Const(true) # definitely assigned previously
             end
-        elseif isa(sym, Symbol)
-            if isdefined(sv.mod, sym.name)
-                t = Const(true)
-            end
         elseif isa(sym, GlobalRef)
             if isdefined(sym.mod, sym.name)
                 t = Const(true)
diff --git a/src/codegen.cpp b/src/codegen.cpp
index dbce9f8c11..a64d02a8b3 100644
--- a/src/codegen.cpp
+++ b/src/codegen.cpp
@@ -3613,15 +3613,9 @@ static jl_cgval_t emit_isdefined(jl_codectx_t &ctx, jl_value_t *sym)
     else {
         jl_module_t *modu;
         jl_sym_t *name;
-        if (jl_is_globalref(sym)) {
-            modu = jl_globalref_mod(sym);
-            name = jl_globalref_name(sym);
-        }
-        else {
-            assert(jl_is_symbol(sym) && "malformed isdefined expression");
-            modu = ctx.module;
-            name = (jl_sym_t*)sym;
-        }
+        assert(jl_is_globalref(sym));
+        modu = jl_globalref_mod(sym);
+        name = jl_globalref_name(sym);
         jl_binding_t *bnd = jl_get_binding(modu, name);
         if (bnd) {
             if (bnd->value != NULL)
diff --git a/src/interpreter.c b/src/interpreter.c
index 0f753346e0..aa244d70d9 100644
--- a/src/interpreter.c
+++ b/src/interpreter.c
@@ -220,9 +220,6 @@ static jl_value_t *eval_value(jl_value_t *e, interpreter_state *s)
         else if (jl_is_globalref(sym)) {
             defined = jl_boundp(jl_globalref_mod(sym), jl_globalref_name(sym));
         }
-        else if (jl_is_symbol(sym)) {
-            defined = jl_boundp(s->module, (jl_sym_t*)sym);
-        }
         else if (jl_is_expr(sym) && ((jl_expr_t*)sym)->head == static_parameter_sym) {
             ssize_t n = jl_unbox_long(jl_exprarg(sym, 0));
             assert(n > 0);
```
but per @JeffBezanson's comment above, it might be ok to remove that path from inference anyway, because it'll get converted to GlobalRef.

aviatesk:
@Keno thx for sharing that ;)

Issue #36951, opened 2020-08-06
suboptimal type-intersections


vtjnash:
```
julia> typeintersect(Type{T} where T>:Missing, Type{Some{T}} where T)
Type{Some{T}} where T

julia> typeintersect(Type{Union{Missing,T}} where T, Type{Some{T}} where T)
Union{}

julia> (Type{T} where {T>:Missing}) == Type{Union{Missing,T}} where T
true
```

```
julia> typeintersect(Tuple{typeof(convert),Type{T},T} where T, Tuple{typeof(convert),Type{T},Any} where T>:Nothing)
Tuple{typeof(convert),Type{T},Any} where T>:Nothing
# preferred: Tuple{typeof(convert),Type{T},T} where T>:Nothing
```

vtjnash:
Another similar one:
```
julia> typeintersect(
  Tuple{typeof(vcat),Vararg{Union{Vector{T}, Matrix{T}, LinearAlgebra.Adjoint{T,Vector{T}}, LinearAlgebra.Transpose{T,Vector{T}}, LinearAlgebra.AbstractTriangular{T,A} where A<:(Matrix{T} where T), LinearAlgebra.Hermitian{T,A} where A<:(Matrix{T} where T), LinearAlgebra.Symmetric{T,A} where A<:(Matrix{T} where T)},N} where N} where T,
  Tuple{typeof(vcat),Vararg{Number,N} where N})
Tuple{typeof(vcat),Vararg{Number,N} where N} # returned LHS
# preferred: Tuple{typeof(vcat),Vararg{Union{}}}
```

vtjnash:
Another similar one:
```
julia> typeintersect(
  Tuple{typeof(LinearAlgebra.ldiv!),LinearAlgebra.Adjoint{var"#s812",var"#s811"} where var"#s811"<:(LinearAlgebra.LowerTriangular{T,var"#s810"} where var"#s810"<:(StridedMatrix{T} where T)) where var"#s812",StridedVecOrMat{T}} where T<:Union{ComplexF32, ComplexF64},
  Tuple{typeof(LinearAlgebra.ldiv!),LinearAlgebra.Adjoint{var"#s812",var"#s811"} where var"#s811"<:(LinearAlgebra.UnitUpperTriangular{T,var"#s810"} where var"#s810"<:(StridedMatrix{T} where T)) where var"#s812",StridedVecOrMat{T}} where T<:Union{ComplexF32, ComplexF64})
[...]

PR #36962, opened 2020-08-07
ml-matches: update and improve ambiguity computation

vtjnash:
This uses the much-enhanced ambiguity computing algorithm in ml-matches, and implements this all better. The failures here are actually all type-intersection problems (https://github.com/JuliaLang/julia/issues/36951), as far as I can tell.


stdlib/LinearAlgebra/test/ambiguous_exec.jl
```diff
@@ -1,4 +1,180 @@
...
 
-using Test, LinearAlgebra
-@test detect_ambiguities(LinearAlgebra; imported=true, recursive=true) == []
+using Test, LinearAlgebra, SparseArrays, SuiteSparse
+let ambig = detect_ambiguities(LinearAlgebra; recursive=true)
+    @test_broken isempty(ambig)
+    ambig = Set{Any}(((m1.sig, m2.sig) for (m1, m2) in ambig))
+    expect = []
+
+    Sys.iswindows() || push!(expect, (Tuple{typeof(\),Factorization{T},Union{Array{Complex{T},1}, Array{Complex{T},2}}} where T<:Union{Float32, Float64}, Tuple{typeof(\),LQ{TA,S} where S<:AbstractMatrix{TA},StridedVecOrMat{TB}} where TB where TA))
```
vtjnash, line 9: this one seems possibly more real though, as the intersection can be either:
```
julia> t = typeintersect(Tuple{typeof(\),Factorization{T},Union{Array{Complex{T},2}, Array{Complex{T},1}}} where T<:Union{Float32, Float64}, Tuple{typeof(\),LQ{TA,S} where S<:AbstractMatrix{TA},StridedVecOrMat{TB}} where TB where TA)
Tuple{typeof(\),
LQ{T,S} where S<:AbstractMatrix{T},
Union{Array{Complex{T},2}, Array{Complex{T},1}}} where T<:Union{Float32, Float64}
```
or
```
julia> t = typeintersect(Tuple{typeof(\),LQ{TA,S} where S<:AbstractMatrix{TA},StridedVecOrMat{TB}} where TB where TA, Tuple{typeof(\),Factorization{T},Union{Array{Complex{T},2}, Array{Complex{T},1}}} where T<:Union{Float32, Float64})
Tuple{typeof(\),
Union{LQ{TA,S} where S<:AbstractMatrix{TA}, LQ{TA1,S} where S<:AbstractMatrix{TA1}},
Union{Array{Complex{TA},1}, Array{Complex{TA1},2}}} where TA1<:Union{Float32, Float64} where TA<:Union{Float32, Float64}
```

which end up matching slightly different sets of methods:
```
1-element Vector{Any}:
 Core.MethodMatch(Tuple{typeof(\),Union{LQ{T,S} where S<:AbstractMatrix{T}, LQ{T,S} where S<:AbstractMatrix{T}},Union{Array{Complex{T},2}, Array{Complex{T},1}}} where T<:Union{Float32, Float64}, svec(T<:Union{Float32, Float64}), \(F::LQ{T,S} where S<:AbstractMatrix{T}, B::Union{Array{Complex{T},1}, Array{Complex{T},2}}) where T<:Union{Float32, Float64} in LinearAlgebra at /data/vtjnash/julia/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/lq.jl:322, true)
```
vs
```
2-element Vector{Any}:
 Core.MethodMatch(Tuple{typeof(\),LQ{T,S} where S<:AbstractMatrix{T},Union{Array{Complex{T},1}, Array{Complex{T},2}}} where T<:Union{Float32, Float64}, svec(T<:Union{Float32, Float64}), \(F::LQ{T,S} where S<:AbstractMatrix{T}, B::Union{Array{Complex{T},1}, Array{Complex{T},2}}) where T<:Union{Float32, Float64} in LinearAlgebra at /data/vtjnash/julia/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/lq.jl:322, false)
 Core.MethodMatch(Tuple{typeof(\),Union{LQ{TA,S} where S<:AbstractMatrix{TA}, LQ{TA1,S} where S<:AbstractMatrix{TA1}},Union{Array{Complex{TA},1}, Array{Complex{TA1},2}}} where TA1<:Union{Float32, Float64} where TA<:Union{Float32, Float64}, svec(TA<:Union{Float32, Float64}, TB<:(Complex{TA} where TA<:Union{Float32, Float64})), \(A::LQ{TA,S} where S<:AbstractMatrix{TA}, B::StridedVecOrMat{TB}) where {TA, TB} in LinearAlgebra at /data/vtjnash/julia/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/lq.jl:310, true)
```

timholy:
Sorry, I should have looked at this earlier. While you've carefully carved out exemptions for ambiguity tests that run within Julia itself, what should packages do? I have a lot of packages where the tests include `detect_ambiguities`, and as of today we're seeing new failures. For example, https://travis-ci.org/github/JuliaDebug/JuliaInterpreter.jl/jobs/723397378.

I'm not arguing that the new algorithm is wrong (I haven't checked), but just wondering whether there's a plan for moving forward.

vtjnash:
There’s no reason anymore to include Base and Core in your list. List just the modules you care about.

timholy:
Is that true even if we add methods to functions owned by Base?

I'll mimic the style of tests here. I am adding a "needs news" tag.

test/ambiguous.jl
```diff
@@ -149,21 +147,61 @@ end
...
-# some ambiguities involving Union{} type parameters are expected, but not required
-@test !isempty(detect_ambiguities(Core, Base; imported=true, ambiguous_bottom=true))
+@testset "detect_ambiguities" begin
+    let ambig = Set{Any}(((m1.sig, m2.sig) for (m1, m2) in detect_ambiguities(Core, Base; recursive=true, ambiguous_bottom=false)))
+        @test_broken isempty(ambig)
+
+        # this is basically just a list of everything for which typeintersection is too conservative,
+        # and ends up matching some methods that don't actually have an interesting intersection
+        expect = []
+        push!(expect, (Tuple{typeof(convert),Type{T},Any} where T>:Union{Missing, Nothing}, Tuple{typeof(convert),Type{T},T} where T>:Nothing))
```
tkf, line 162: I try to understand this list of methods enumerated in the `expect` vector. Since this PR is an improvement, they are previously false-negative and now true-positive, right? If so, is it correct to assume that I should be able to invoke method error with some combination? However, I couldn't get the run-time ambiguity error with a few examples for this first item of `expect`. What am I doing wrong here?

```julia
julia> VERSION
v"1.6.0-DEV.819"

julia> convert(Union{Missing, Nothing, Int}, 1)
1

julia> convert(Union{Missing, Nothing}, missing)
missing

julia> convert(Union{Missing, Nothing}, nothing)

julia> convert(Union{Missing, Nothing, Int}, missing)
missing

julia> convert(Union{Missing, Nothing, Int}, nothing)

julia> convert(Union{Missing, Nothing, Int}, 1)
1
```

vtjnash: Not necessarily—the type system might claim that such an ambiguity exists, while no counterfactual actually does. I posted a link above that references this particular failure in the type system (https://github.com/JuliaLang/julia/issues/36951).

tkf: What is the API of `detect_ambiguities`?  I've been assuming that it returns a non-empty vector if and only if there exist concrete argument types that would cause a method error.  Could you document the exact guarantee?  For example, it would be helpful to document that the users should expect false positive or negative.

If providing a guarantee is not possible or not appropriate (e.g., to allow the type system to evolve), can we demote `detect_ambiguities` (and `detect_unbound_args`?) to experimental API and explicitly mention as such.

tkf:
I noted above https://github.com/JuliaLang/julia/pull/36962#discussion_r483285985, I think we need more documentation. So, I'm also adding `needs docs`

I'm also adding `breaking` because:

1. It breaks existing user code bases.
2. If we retrofit a more specific API definition by introducing new documentation, it could be different from the users' mental models.  It is not ideal to change the behavior of a 4-year-old API (https://github.com/JuliaLang/julia/commit/be8679e965d89baf10161112e44f594339cd7daf) and claim a change in observable behavior as a bug fix or improvement when there is no documentation to ensure that the old and new behaviors are compatible in some ways.

test/ambiguous.jl
vtjnash: Aside from problems with the type system, that is what this did before and is still what it does. Hence the new bug report number also. There was a bug in v1.0 where someone got a conditional backwards and accidentally additionally filtered out any ambiguity resulting from a failure of type-intersection. Hence this PR. That's now fixed.

tkf: > that is what this did before and is still what it does

What is _that_?  Could you explain what `detect_ambiguities` does now?  Could you clarify what kind of "bound" users can assume with respect to the run-time behavior (method error)? No false positive (which is apparently not the case from your previous comment?)? No false negative (i.e., if `detect_ambiguities` returns an empty vector, there will be no method error; given there is no bug, of course)? Both kinds of failures are possible (i.e., it's just a hint)?

JeffBezanson:
This seems to cause some significant latency regressions.
Current master:
```
Base  ─────────── 20.674068 seconds
Stdlibs: ────  31.106509 seconds 60.0718%

julia> @time using Plots
 94.738177 seconds (8.06 M allocations: 545.291 MiB, 0.12% gc time)

  7.265292 seconds (10.40 M allocations: 601.608 MiB, 3.82% gc time)

julia> @time using Plots
  4.513663 seconds (7.97 M allocations: 539.723 MiB, 2.35% gc time)
```

with this reverted:
```
Base  ─────────── 20.185031 seconds
Stdlibs: ────  29.825570 seconds 59.6366%

julia> @time using Plots
 85.283090 seconds (6.21 M allocations: 443.919 MiB, 0.12% gc time)

  6.373428 seconds (8.95 M allocations: 514.353 MiB, 4.12% gc time)

julia> @time using Plots
  3.221717 seconds (6.12 M allocations: 438.344 MiB, 3.30% gc time)
```

So I don't think this change is worth it.

vtjnash:
Yeah, I don’t think inference should be asking us to do the more expensive query (eliminate ambiguous methods) that this makes even worse. Need to first get data on what patterns this is affecting. IIUC, this is revealing problems with type-intersection that are stealing away a lot of performance by making it slightly worse in those cases.

timholy:
Ah, I see from https://github.com/JuliaLang/julia/issues/37458#issuecomment-689255214 that this seems likely to be reverted.

vtjnash:
It should probably be surgically disabled again (as this code has been), rather than undoing the bug fixes entirely, but I can't get to that till next week.

Issue #36963, opened 2020-08-07
`findnext(A, start)` and `findnext(testf::Function, A, start)` behave differently


goretkin:
e.g.

```julia
using InfiniteArrays

# works
findfirst(>(5), 1:InfiniteArrays.Infinity())

# does not
findfirst(1:InfiniteArrays.Infinity() .> 5)
```

Divergence at https://github.com/JuliaLang/julia/pull/35965

dlfivefifty:
Just needs

https://github.com/JuliaLang/julia/blob/dffc8895b897a78a8dca64c784962675df49c344/base/array.jl#L1725

To become
```julia
i = oftype(first(keys(A)), start)
```
Shall I make a PR?

nalimilan:
Probably better convert to `eltype(keys(A))` instead, to avoid special-casing the last index.

goretkin:
@dlfivefifty I would prefer to get #36965 in as a solution. And then we could follow @nalimilan's suggestion, and we'll only have to fix it in half as many places.

vtjnash:
Addressed by #36965, IIUC

Issue #36938, opened 2020-08-06
`findnext` and degenerate multidimensional arrays


goretkin:
This behavior strikes me as inconsistent:

```julia
julia> x1 = zeros(Bool, (Base.OneTo(2), Base.OneTo(0)))
2×0 Array{Bool,2}

julia> x2 = zeros(Bool, (Base.OneTo(0), Base.OneTo(2)))
0×2 Array{Bool,2}

julia> first(keys(x1))
CartesianIndex(1, 1)

julia> first(keys(x2))
CartesianIndex(1, 1)

julia> last(keys(x1))
CartesianIndex(2, 0)

julia> last(keys(x2))
CartesianIndex(0, 2)

julia> findnext(x1, first(keys(x1)))

julia> findnext(x2, first(keys(x2)))
ERROR: BoundsError: attempt to access 0×2 Array{Bool,2} at index [1, 1]
Stacktrace:
 [1] getindex at ./array.jl:810 [inlined]
 [2] getindex at ./multidimensional.jl:557 [inlined]
 [3] findnext(::Array{Bool,2}, ::CartesianIndex{2}) at ./array.jl:1728
 [4] top-level scope at REPL[32]:1
```

I had expected both calls to `findnext` to return `nothing`.

Honing in, the inconsistency arises because of

```julia
julia> CartesianIndex(2, 0) > CartesianIndex(1, 1)
false

julia> CartesianIndex(0, 2) > CartesianIndex(1, 1)
true
```

and so the check at https://github.com/JuliaLang/julia/blob/4c1e3c039775e56c72a4a638db0a8a20f4972dd3/base/array.jl#L1726 passes through.

I believe the intention in `findnext` is to compare `CartesianIndex`s in a way that agrees with the corresponding comparison on the `LinearIndex` into the array. For example:

```julia
julia> Base._sub2ind(x1, Tuple(first(keys(x1)))...)
1

julia> Base._sub2ind(x1, Tuple(last(keys(x1)))...)
0

julia> Base._sub2ind(x2, Tuple(first(keys(x2)))...)
1

julia> Base._sub2ind(x2, Tuple(last(keys(x2)))...)
0
```

(I tried on Julia 1.4 and Julia 1.5)

nalimilan:
Good catch. I guess the reason for the current behavior is to have first(keys(x)):last(keys(x)) == keys(x)`. But it's not great that `first(keys(x)) < last(keys(x))` even when `x` is empty.

[...]

PR #25577, opened 2018-01-15
Change findfirst/findlast/findnext/findprev to return the same index type as keys()

nalimilan:
This is consistent with `find`. Also add needed `last(::EachStringIndex)` and `prevind(::AbstractArray, ::CartesianIndex)` methods.

Part of https://github.com/JuliaLang/julia/issues/10593.

`findnext`/`findprev` are not as generic as `find`/`findfirst`/`findlast` since their interface assumes that indices are ordered (comparable with `<`) and that they are an efficient way of keeping track of the iteration state. This works for arrays, tuples and strings, but not e.g. for dicts (and not even e.g. for `OrderedDict`, since the ordering of keys isn't reflected by `<`). Only a `findeach` function returning an iterator would allow that (we can always introduce it later).

It would be possible to support passing linear indices to `findnext`/`findprev`, and have them return linear indices in that case. This behavior would probably have to be restricted to `AbstractArray`. That would limit breakage compared with 0.6 a bit (though `findfirst` and `findlast` would still break). Should I add it?

PR #36965, opened 2020-08-07
Reduce code duplication in find* functions

goretkin:
Try to address https://github.com/JuliaLang/julia/issues/36963

Also, whatever fix arises from https://github.com/JuliaLang/julia/issues/36938 will only need to be applied in a fewer number of places.

I expect the compiler can certainly specialize on `identity`. Is it necessary to force inlining?

### DIFF
--- a/base/array.jl
+++ b/base/array.jl
@@ -1720,18 +1720,7 @@ julia> findnext(A, CartesianIndex(1, 1))
 CartesianIndex(2, 1)
 ```
 """
-function findnext(A, start)
-    l = last(keys(A))
-    i = oftype(l, start)
-    i > l && return nothing
-    while true
-        A[i] && return i
-        i == l && break
-        # nextind(A, l) can throw/overflow
-        i = nextind(A, i)
-    end
-    return nothing
-end
+findnext(A, start) = findnext(identity, A, start)
 
 """
     findfirst(A)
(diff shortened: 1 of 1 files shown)


goretkin:
predicate and non-predicate versions were introduced by https://github.com/JuliaLang/julia/pull/25577

nalimilan:
Sounds good, but can you run some benchmarks and look at the generated code to check that this doesn't introduce regressions?

goretkin:
Yeah, it would be good to have a sense that there aren't performance regressions. Can nanosoldier help here?

nalimilan:
Yes, if there are tests in BaseBenchmarks for this (haven't checked) or if you make a PR to add them. But a quick local check would be enough IMHO. You can also check that the generated code is the same (which is likely).

goretkin:
Looks like there are some: https://github.com/JuliaCI/BaseBenchmarks.jl/blob/ce11d8badcb0e9f6d8b0518ffd57a472128af930/src/find/FindBenchmarks.jl#L39

but I'll take a look at the generated code, since you're probably right that it didn't change.

goretkin:
FWIW, I wanted a way to programmatically compare the generated code, and I didn't know how to do that since all the tools seem to print to `stdout` directly. Perfect was the enemy of good, but I'm glad to see this merged!

PR #36966, opened 2020-08-07
Add `Scratch` stdlib

staticfloat:
This implements functionality and tests for a new `Scratch` subsystem,
providing an abstraction for a mutable datastore that can be explicitly
lifecycled to an owning package, or shared among multiple packages.

Note that garbage collecting these scratch spaces is relegated to
`Pkg.gc()`, included in the next commit that bumps Pkg version.


fredrikekre:
> Note that garbage collecting these scratch spaces is relegated to
Pkg.gc(), included in the next commit that bumps Pkg version.

Is this the only reason Scratch needs to be a stdlib?

stdlib/Scratch/docs/index.md
```diff
@@ -0,0 +1,145 @@
...
+> Can I create a scratch space that is not shared across versions of my package?
+
+Yes!  Make use of the `key` parameter and Pkg's ability to look up the current version of your package at compile-time:
+
+```julia
+module VersionSpecificExample
+using Pkg.TOML, Scratch
+
+# Get the current version at compile-time, that's fine it's not going to change. ;)
+function get_version()
```
thofma, line 81: This is a very useful function in its own. By putting in the official docs, does this become the official way to query for the own package version at compile-time?

stdlib/Scratch/src/Scratch.jl
```diff
@@ -0,0 +1,240 @@
...
+function clear_scratchspaces!(m::Module)
+    uuid = Base.PkgId(__module__).uuid
+    if uuid === nothing
+        throw(ArgumentError("Cannot find owning package for Module $(m)"))
+    end
+    return clear_scratchspaces!(uuid)
+end
+
+"""
+    @get_scratch!(key)
```
laborg, line 224: Sorry for the drive by commenting: I really think it would be better not to have this macro in the API. As a package developer you wouldn't call `@get_scratch!` very often, so having to memorize a name once as a function and once as a macro is cumbersome (e.g. there has even been confusion about the often used `error()` and `@error`). At least it is harder to remember than the omnipresent `@__MODULE__`.

staticfloat:
@KristofferC Is it possible there is a TOML parser bug here?  I am running `cat scratch_usage.toml` first and getting the following contents:

```
      From worker 2:	[["C:\Windows\TEMP\jl_kBPAAU\scratchspaces\93485645-17f1-6f3b-45bc-419db53815ea\1.0.0"]]
      From worker 2:	time = 2020-08-10T08:35:46.381Z
      From worker 2:	parent_projects = ["C:\Windows\TEMP\jl_MbzW2M\ScratchUsage\Project.toml"]
      From worker 2:	[["C:\Windows\TEMP\jl_kBPAAU\scratchspaces\93485645-17f1-6f3b-45bc-419db53815ea\1"]]
      From worker 2:	time = 2020-08-10T08:35:46.475Z
      From worker 2:	parent_projects = ["C:\Windows\TEMP\jl_MbzW2M\ScratchUsage\Project.toml"]
      From worker 2:	[["C:\Windows\TEMP\jl_kBPAAU\scratchspaces\00000000-0000-0000-0000-000000000000\GlobalSpace"]]
      From worker 2:	time = 2020-08-10T08:35:46.475Z
      From worker 2:	parent_projects = ["C:\Windows\TEMP\jl_kBPAAU\environments\v1.6\Project.toml"]
```

But then immediately afterward, attempting to parse it gets the following:
```
Got exception outside of a @test
  Pkg.TOML.ParserError(7, 8, "unknown string escape: `\\x57`")
```

And this is only happening on Windows.

KristofferC:
> @KristofferC Is it possible there is a TOML parser bug here? 

Possibly, do you have a file that reproduces it? Would also be interesting to try it with the parser in https://github.com/JuliaLang/julia/pull/36018

staticfloat:
> Would also be interesting to try it with the parser in #36018

It also happens with the parser in #36018.  I must be doing something stupid.

Here's the `.toml` file: https://gist.github.com/staticfloat/bae105297ca393fb78d2f137e59844b8

*EDIT*: After pasting it into GitHub, I notice the TOML syntax highlighting is showing that TOML interprets backslash escape codes!  That could cause some issues on windows, of course, although it seems odd that the Pkg TOML parser complaints about `\\x57` (which looks like UTF-8 mojibake) but I suppose that's just a `\W`.

StefanKarpinski:
Yeah, those need to be double escaped.

staticfloat:
Is there a good way to silence all the Pkg output during the scratch tests?

stdlib/Scratch/docs/index.md
staticfloat: I think you'd have to ask Fredrik or Kristoffer.  I had some vague ideas on how to do this properly, but it seems there are edge cases I didn't take into account.  The text above this assumed that my previous Pkg PR got merged, which it didn't, so I'll fix that now.

stdlib/Scratch/src/Scratch.jl
staticfloat: I disagree; I think making it easy for the user to not have to look up their own UUID or pass in `@__MODULE__` all the time is worthwhile.

staticfloat:
Kristoffer and Fredrik convinced me that we should just ship this as a normal package, as the only thing tying it to a Julia release is `Pkg.gc()`'s ability to clean up after it.  This seems reasonable to me (and gives us the ability to ship fixes independent of Julia version) so I'm going to just do that.

StefanKarpinski:
Maybe we need to provide a generic way to hook into Pkg.gc?

PR #36967, opened 2020-08-07
[WIP] `find*` bounds checking

### LINKED ISSUE #36938: `findnext` and degenerate multidimensional arrays

goretkin:
Attempt to [identify and] fix https://github.com/JuliaLang/julia/issues/36938

~(dovetails with https://github.com/JuliaLang/julia/pull/36965)~ (already merged)

### DIFF
--- a/test/arrayops.jl
+++ b/test/arrayops.jl
@@ -594,6 +594,33 @@ end
         @test findprev(b, T(1)) isa keytype(b)
         @test findprev(b, T(2)) isa keytype(b)
     end
+
+    @testset "issue #36938" begin
+        # 👎 denotes tests that are broken.
+        @test findfirst(fill(true, (0, 2))) === nothing     # 👎
+        @test findfirst(fill(true, (2, 0))) === nothing
+        @test findfirst(fill(true, (0, 0))) === nothing
+
+        @test findlast(fill(true, (0, 2))) === nothing      # 👎
+        @test findlast(fill(true, (2, 0))) === nothing
+        @test findlast(fill(true, (0, 0))) === nothing
+
+        @test findnext(fill(true, (2, 2)), CartesianIndex(1, 3)) === nothing
+        @test findnext(fill(true, (2, 2)), CartesianIndex(3, 1)) === nothing    # 👎
+        @test findnext(fill(true, (2, 2)), CartesianIndex(3, 3)) === nothing
+
+        @test_throws BoundsError findnext(fill(true, (2, 2)), CartesianIndex(1, 0))
+        @test_throws BoundsError findnext(fill(true, (2, 2)), CartesianIndex(0, 1))
+        @test_throws BoundsError findnext(fill(true, (2, 2)), CartesianIndex(0, 0))
+
+        @test_throws BoundsError findprev(fill(true, (2, 2)), CartesianIndex(1, 3))
+        @test_throws BoundsError findprev(fill(true, (2, 2)), CartesianIndex(3, 1))
+        @test_throws BoundsError findprev(fill(true, (2, 2)), CartesianIndex(3, 3))
+
+        @test findprev(fill(true, (2, 2)), CartesianIndex(1, 0)) === nothing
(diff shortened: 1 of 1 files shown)


vtjnash:
rebase, for a new CI run?

goretkin:
I rebased.

test/arrayops.jl
```diff
@@ -594,6 +594,33 @@ end
         @test findprev(b, T(1)) isa keytype(b)
         @test findprev(b, T(2)) isa keytype(b)
     end
+
+    @testset "issue #36938" begin
+        # 👎 denotes tests that are broken.
+        @test findfirst(fill(true, (0, 2))) === nothing     # 👎
```
KristofferC, line 600: Shouldn't these use `@test_broken` with the correct answer then?

vtjnash: Looks like all lines with :-1: are currently throwing a BoundsError so this PR is failing CI

goretkin: I've lost track of what needs to happen here. I think my goal was to write tests that should pass, but I don't remember if the current (inconsistent) behavior can be fixed before Julia 2.0, e.g.:

```julia
julia> findfirst(fill(true, (0,)))

julia> findfirst(fill(true, (1, 0)))

julia> findfirst(fill(true, (0, 1)))

julia> findfirst(fill(true, (2, 0)))

julia> findfirst(fill(true, (0, 2)))
ERROR: BoundsError: attempt to access 0×2 Matrix{Bool} at index [1, 1]
```

Obviously failing CI does not help. 

I've long wanted a macro like the one below, then I could write these as `@test_broken`:

```
julia> macro error_as_value(ex)
           quote
               try
                   return Some($(ex))
               catch e
                   return e
               end
           end
       end
@error_as_value (macro with 1 method)

julia> @error_as_value sqrt(4)
Some(2.0)

julia> @error_as_value sqrt(-4)
DomainError(-4.0, "sqrt will only return a complex result if called with a complex argument. Try sqrt(Complex(x)).")
```

DilumAluthge:

@goretkin If you're still interested in working on this PR, let me know, and I can re-open it.

PR #36971, opened 2020-08-08
Mention export from InteractiveUtils in docs

ConnectedSystems:
I recently had a frustrating experience where seemingly valid code that worked in the REPL did not work in my own packages.

Specifically, it is not mentioned in the documentation that `subtypes` was being automatically exported from `InteractiveUtils` when using the REPL.
Thanks to @oheil for pointing this out (see exchange [here](https://discourse.julialang.org/t/error-with-base-generic-function-calls/44174)).

Although in my case I figured out that `subtypes` wasn't exactly what I needed, I still feel the documentation should be explicit about what functions are in Base and what functions aren't to avoid the kind of off-putting experience I had.

The change here isn't exhaustive but if there is agreement I'm willing to help update the documentation where needed.

### DIFF
--- a/doc/src/base/reflection.md
+++ b/doc/src/base/reflection.md
@@ -63,8 +63,8 @@ julia> InteractiveUtils.subtypes(AbstractFloat)
 Any abstract subtype will also be included in this list, but further subtypes thereof will not;
 recursive application of [`subtypes`](@ref) may be used to inspect the full type tree.
 
-Note that [`subtypes`](@ref) is located inside [`InteractiveUtils`](@ref) but is automatically
-exported when using the REPL.
+Note that [`subtypes`](@ref) is located inside [`InteractiveUtils`](@ref man-interactive-utils) but
+is automatically exported when using the REPL.
 
 ## DataType layout
 
--- a/doc/src/base/reflection.md
+++ b/doc/src/base/reflection.md
@@ -51,7 +51,7 @@ The *direct* subtypes of any `DataType` may be listed using [`subtypes`](@ref). 
 the abstract `DataType` [`AbstractFloat`](@ref) has four (concrete) subtypes:
 
 ```jldoctest; setup = :(using InteractiveUtils)
-julia> subtypes(AbstractFloat)
+julia> InteractiveUtils.subtypes(AbstractFloat)
 5-element Vector{Any}:
  BigFloat
  Core.BFloat16
@@ -86,7 +86,7 @@ the unquoted and interpolated expression ([`Expr`](@ref)) form for a given macro
 be passed instead!). For example:
 
 ```jldoctest; setup = :(using InteractiveUtils)
-julia> macroexpand(@__MODULE__, :(@edit println("")) )
+julia> InteractiveUtils.macroexpand(@__MODULE__, :(@edit println("")) )
 :(InteractiveUtils.edit(println, (Base.typesof)("")))
 ```
 
@@ -143,8 +143,8 @@ For more information see [`@code_lowered`](@ref), [`@code_typed`](@ref), [`@code
 The aforementioned functions and macros take the keyword argument `debuginfo` that controls the level
 debug information printed.
 
-```julia-repl
-julia> @code_typed debuginfo=:source +(1,1)
+```jldoctest; setup = :(using InteractiveUtils)
+julia> InteractiveUtils.@code_typed debuginfo=:source +(1,1)
 CodeInfo(
     @ int.jl:53 within `+'
 1 ─ %1 = Base.add_int(x, y)::Int64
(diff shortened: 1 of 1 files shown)


fredrikekre:
I am not sure this is the best place for this -- should we then add this to every mention of methods from InteractiveUtils? A suggestion would be to add an FAQ entry with something like "What are the differences between running code in the REPL and in a script?".

ConnectedSystems:
@fredrikekre Others might argue that each example should be able to stand on its own but for myself, I find your suggestion of an FAQ entry would be preferable to how things are currently provided that its existence is made clear.

I guess the larger complaints here are that it is not immediately clear, to a newcomer, what is in Base Julia and that the language behavior can change between the REPL vs a script. The recent change to variable scoping is one example that people seem to have an issue with (at least anecdotally).

I would suggest that, in the short term at least, relevant pages have a clear link to this FAQ at the top.

fredrikekre:
How about something like
````diff
diff --git a/doc/src/devdocs/reflection.md b/doc/src/devdocs/reflection.md
index fbf0fd5..acad9d6 100644
--- a/doc/src/devdocs/reflection.md
+++ b/doc/src/devdocs/reflection.md
@@ -50,7 +50,9 @@ of these fields is the `types` field observed in the example above.
 The *direct* subtypes of any `DataType` may be listed using [`subtypes`](@ref). For example,
 the abstract `DataType` [`AbstractFloat`](@ref) has four (concrete) subtypes:
 
-```jldoctest; setup = :(using InteractiveUtils)
+```jldoctest
+julia> using InteractiveUtils # loaded automatically in the REPL
+
 julia> subtypes(AbstractFloat)
 4-element Vector{Any}:
  BigFloat
````
for this particular case, and then someone can add the FAQ in another PR?

**Comment by vtjnash** [comment]
How about something in between, where we keep the text unchanged, but fully-qualify the usage:
````
```jldoctest; setup = :(using InteractiveUtils)
julia> InteractiveUtils.subtypes(AbstractFloat)
...
````

**Comment by ConnectedSystems** [comment]
I like this compromise @vtjnash !

**Comment by vtjnash** [comment]
Are you familiar with updating PRs? I can merge once that is done.

**Comment by ConnectedSystems** [comment]
Yes, of course.

I'll try to find the time to do so soon.

**Comment by simeonschaub** [comment]
It seems odd to me to explicitly qualify exports of `InteractiveUtils` here when it's already clear that the examples are run in the REPL. @fredrikekre's proposal sounds clearer to me personally.

**Comment by ViralBShah** [comment]
Do we have consensus on what we want here?

**Comment by DilumAluthge** [comment]
The build is failing due to an error when building the docs.

File: doc/src/base/reflection.md [hunk 1]
```diff
@@ -140,8 +143,8 @@ For more information see [`@code_lowered`](@ref), [`@code_typed`](@ref), [`@code
 The aforementioned functions and macros take the keyword argument `debuginfo` that controls the level
 debug information printed.
 
-```julia-repl
-julia> @code_typed debuginfo=:source +(1,1)
+```jldoctest; setup = :(using InteractiveUtils)
+julia> InteractiveUtils.@code_typed debuginfo=:source +(1,1)
 CodeInfo(
     @ int.jl:53 within `+'
```
giordano, line 149: Doctest is now failing, if I got it correctly it should be
```suggestion
    @ int.jl:87 within `+'
```

doc/src/base/reflection.md
```diff
@@ -140,10 +143,10 @@ For more information see [`@code_lowered`](@ref), [`@code_typed`](@ref), [`@code
 The aforementioned functions and macros take the keyword argument `debuginfo` that controls the level
 debug information printed.
 
-```julia-repl
-julia> @code_typed debuginfo=:source +(1,1)
+```jldoctest; setup = :(using InteractiveUtils)
```
vtjnash, line 146:
```suggestion
```jldoctest; setup = :(using InteractiveUtils) filter=r"int.jl:\d+"
```

doc/src/base/reflection.md
```diff
@@ -140,10 +143,10 @@ For more information see [`@code_lowered`](@ref), [`@code_typed`](@ref), [`@code
 The aforementioned functions and macros take the keyword argument `debuginfo` that controls the level
 debug information printed.
 
-```julia-repl
-julia> @code_typed debuginfo=:source +(1,1)
+```jldoctest; setup = :(using InteractiveUtils), filter = r"int.jl:\d+"
+julia> InteractiveUtils.@code_typed debuginfo=:source +(1,1)
 CodeInfo(
-    @ int.jl:53 within `+'
+    @ int.jl:87 within `+'
```
vtjnash, line 149:
```suggestion
    @ int.jl:87 within `+`
```

PR #36984, opened 2020-08-10
Add support for syntax highlighting for `code_llvm` and `code_native`

### LINKED ISSUE #36634: Colored printing of `code_llvm` and  `code_native`
cf. https://discourse.julialang.org/t/why-are-code-llvm-and-code-native-not-displayed-in-color/42959

Recently, the display of stack traces has become colorful (cf. issue #36026, PR #36134). I also think it is helpful to make the output of `code_llvm` and `code_native` colorful.

Unregistered package [ColorfulCodeGen.jl](https://github.com/tkf/ColorfulCodeGen.jl) provides the functionality (thanks to @tkf). However, the ColorfulCodeGen.jl uses an external tool [Pygments](https://pygments.org/).

I wondered if huge lists of keywords and/or complicated parsers were needed to color the automatically generated LLVM IR or native assembler codes. So I tried to implement it. You can copy & paste the following PoC to your REPL (Julia v1.4 or later).
**Edit:** You can get the slightly improved version and `code_native` version via [ColoredLLCodes.jl](https://github.com/kimikage/ColoredLLCodes.jl) package now.

<details>
<summary>PoC for `code_llvm`</summary>

```julia
import InteractiveUtils: _dump_function, code_llvm, code_native

const IO_ = Union{Base.AbstractPipe, Base.LibuvStream} # avoid overwriting

llstyle = Dict{Symbol, Tuple{Bool, Union{Symbol, Int}}}(
    :default     => (false, :light_black),
    :comment     => (false, :green),
    :label       => (false, :light_red),
    :instruction => ( true, :light_cyan),
    :type        => (false, :cyan),
    :number      => (false, :yellow),
    :bracket     => (false, :yellow),
    :variable    => (false, :normal),
    :keyword     => (false, :light_magenta),
    :funcname    => (false, :light_yellow),
)

const num_regex = r"^(?:\$?-?\d+|0x[0-9A-F]+|-?(?:\d+\.?\d*|\.\d+)(?:[eE][+-]?\d+)?)$"

function printstyled_ll(io::IO, x, s::Symbol, trailing_spaces="")
    isempty(x) || printstyled(io, x, bold=llstyle[s][1], color=llstyle[s][2])
    print(io, trailing_spaces)
end

function code_llvm(io::IO_, @nospecialize(f), @nospecialize(types),
                   raw::Bool, dump_module::Bool=false, optimize::Bool=true,
                   debuginfo::Symbol=:default)
    d = _dump_function(f, types, false, false, !raw, dump_module, :att, optimize, debuginfo)
    if get(io, :color, false)
        print_llvm(io, d)
    else
        print(io, d)
    end
end

function print_llvm(io::IO, code::String)
    buf = IOBuffer(code)
    for line in eachline(buf)
        m = match(r"^(\s*)((?:[^;]|;\")*)(.*)$", line)
        m === nothing && continue
        indent, tokens, comment = m.captures
        print(io, indent)
        print_llvm_tokens(io, tokens)
        printstyled_ll(io, comment, :comment)
        println(io)
    end
end

const llvm_types =
    r"^(?:void|half|float|double|x86_\w+|ppc_\w+|label|metadata|type|opaque|token|i\d+)$"
const llvm_cond = r"^(?:[ou]?eq|[ou]?ne|[uso][gl][te]|ord|uno|true|false)$"

function print_llvm_tokens(io, line)
    tokens = line
    m = match(r"^((?:[^\s:]+:)?)(\s*)(.*)", tokens)
    if m !== nothing
        label, spaces, tokens = m.captures
        printstyled_ll(io, label, :label, spaces)
    end
    m = match(r"^(%[^\s=]+)(\s*)=(\s*)(.*)", tokens)
    if m !== nothing
        result, spaces, spaces2, tokens = m.captures
        printstyled_ll(io, result, :variable, spaces)
        printstyled_ll(io, '=', :default, spaces2)
    end
    m = match(r"^([a-z]\w*)(\s*)(.*)", tokens)
    if m !== nothing
        inst, spaces, tokens = m.captures
        printstyled_ll(io, inst, inst == "define" ? :keyword : :instruction, spaces)
    end

    print_llvm_operands(io, tokens)
end

function print_llvm_operands(io, tokens)
    while !isempty(tokens)
        tokens = print_llvm_operand(io, tokens)
    end
    return tokens
end

function print_llvm_operand(io, tokens)
    islabel = false
    while !isempty(tokens)
        m = match(r"^,(\s*)(.*)", tokens)
        if m !== nothing
            spaces, tokens = m.captures
            printstyled_ll(io, ',', :default, spaces)
            break
        end
        m = match(r"^(\*+)(\s*)(.*)", tokens)
        if m !== nothing
            asterisks, spaces, tokens = m.captures
            printstyled_ll(io, asterisks, :default, spaces)
            continue
        end
        m = match(r"^([({\[<])(\s*)(.*)", tokens)
        if m !== nothing
            bracket, spaces, tokens = m.captures
            printstyled_ll(io, bracket, :bracket, spaces)
            tokens = print_llvm_operands(io, tokens) # enter
            continue
        end
        m = match(r"^([)}\]>])(\s*)(.*)", tokens)
        if m !== nothing
            bracket, spaces, tokens = m.captures
            printstyled_ll(io, bracket, :bracket, spaces)
            break # leave
        end

        m = match(r"^([^\s,*(){}\[\]<>]+)(\s*)(.*)", tokens)
        m === nothing && break
        token, spaces, tokens = m.captures
        if occursin(llvm_types, token)
            printstyled_ll(io, token, :type)
            islabel = token == "label"
        elseif occursin(llvm_cond, token) # condition code is instruction-level
            printstyled_ll(io, token, :instruction)
        elseif occursin(num_regex, token)
            printstyled_ll(io, token, :number)
        elseif occursin(r"^@.+$", token)
            printstyled_ll(io, token, :funcname)
        elseif occursin(r"^%.+$", token)
            printstyled_ll(io, token, islabel ? :label : :variable)
            islabel = false
        elseif occursin(r"^[a-z]\w+$", token)
            printstyled_ll(io, token, :keyword)
        else
            printstyled_ll(io, token, :default)
        end
        print(io, spaces)
    end
    return tokens
end
```

</details>

code_llvm

Note that the `code_llvm` and `code_native` are defined in `InteractiveUtils` (stdlib), not `Base`. Also, please do not start a bicycle shed discussion on color schemes before we have a firm idea of ​​where and how to implement this feature. :sweat_smile:

I can continue to improve the PoC in a separate package, but the value of such a heuristic approach (i.e. the motivation) will be diminished.

kimikage:
This adds the support for syntax highlighting for `code_llvm` and `code_native` in `InteractiveUtils`.

This code comes from the newly registered package [`ColoredLLCodes.jl`](https://github.com/kimikage/ColoredLLCodes.jl)．With that package, you can try out the syntax highlighting feature.

I have no intention of abandoning the maintenance of `ColoredLLCodes.jl` in order to use the syntax highlighting within older versions of Julia. Therefore, there is no need to implement this feature in `InteractiveUtils`, and in fact, there are both opinions as to whether or not it should be implemented in `InteractiveUtils`.

Also, this is just a cosmetic change, i.e. this only changes the colors. (cf. https://github.com/JuliaLang/julia/issues/36634#issuecomment-659084346)

Closes #36634

### DIFF
--- a/stdlib/InteractiveUtils/src/codeview.jl
+++ b/stdlib/InteractiveUtils/src/codeview.jl
@@ -1,17 +1,46 @@
 # This file is a part of Julia. License is MIT: https://julialang.org/license
 
+# highlighting settings
+highlighting = Dict{Symbol, Bool}(
+    :warntype => true,
+    :llvm => true,
+    :native => true,
+)
+
+llstyle = Dict{Symbol, Tuple{Bool, Union{Symbol, Int}}}(
+    :default     => (false, :light_black), # e.g. comma, equal sign, unknown token
+    :comment     => (false, :green),
+    :label       => (false, :light_red),
+    :instruction => ( true, :light_cyan),
+    :type        => (false, :cyan),
+    :number      => (false, :yellow),
+    :bracket     => (false, :yellow),
+    :variable    => (false, :normal), # e.g. variable, register
+    :keyword     => (false, :light_magenta),
+    :funcname    => (false, :light_yellow),
+)
+
+function printstyled_ll(io::IO, x, s::Symbol, trailing_spaces="")
+    printstyled(io, x, bold=llstyle[s][1], color=llstyle[s][2])
+    print(io, trailing_spaces)
+end
+
 # displaying type warnings
 
 function warntype_type_printer(io::IO, @nospecialize(ty), used::Bool)
     used || return
     if ty isa Type && (!Base.isdispatchelem(ty) || ty == Core.Box)
-        if ty isa Union && Base.is_expected_union(ty)
+        if highlighting[:warntype] && ty isa Union && Base.is_expected_union(ty)
             Base.emphasize(io, "::$ty", Base.warn_color()) # more mild user notification
         else
             Base.emphasize(io, "::$ty")
         end
     else
-        Base.printstyled(io, "::$ty", color=:cyan) # show the "good" type
+        if highlighting[:warntype]
+            Base.printstyled(io, "::$ty", color=:cyan) # show the "good" type
+        else
+            Base.print(io, "::$ty")
+        end
     end
     nothing
 end
@@ -135,8 +164,15 @@ All metadata and dbg.* calls are removed from the printed bitcode. For the full 
 To dump the entire module that encapsulates the function (with declarations), set the `dump_module` keyword to true.
 Keyword argument `debuginfo` may be one of source (default) or none, to specify the verbosity of code comments.
 """
-code_llvm(io::IO, @nospecialize(f), @nospecialize(types), raw::Bool, dump_module::Bool=false, optimize::Bool=true, debuginfo::Symbol=:default) =
-    print(io, _dump_function(f, types, false, false, !raw, dump_module, :att, optimize, debuginfo))
+function code_llvm(io::IO, @nospecialize(f), @nospecialize(types), raw::Bool,
+                   dump_module::Bool=false, optimize::Bool=true, debuginfo::Symbol=:default)
+    d = _dump_function(f, types, false, false, !raw, dump_module, :att, optimize, debuginfo)
+    if highlighting[:llvm] && get(io, :color, false)
+        print_llvm(io, d)
+    else
+        print(io, d)
+    end
+end
 code_llvm(io::IO, @nospecialize(f), @nospecialize(types=Tuple); raw::Bool=false, dump_module::Bool=false, optimize::Bool=true, debuginfo::Symbol=:default) =
     code_llvm(io, f, types, raw, dump_module, optimize, debuginfo)
 code_llvm(@nospecialize(f), @nospecialize(types=Tuple); raw=false, dump_module=false, optimize=true, debuginfo::Symbol=:default) =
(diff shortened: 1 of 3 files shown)


kimikage:
I'll close this, but you can re-open this if you want.

musm:
Are we just looking for someone to review this?

musm:
I think there's  a question whether syntax highlighting belongs here.  I know we don't highlight Julia code, but this is less benign and doesn't rely on a parser to determine highlighting.  I believe the code is simple enough and strongly increases usability of these functions  to be considered in `InteractiveUtils`.

StefanKarpinski:
I agree that this seems worth having. It doesn't add any external dependencies and these formats are pretty stable. And whenever we switch to using a better parser for Julia itself, we'll probably add syntax highlighting for Julia code in the REPL too.

stdlib/InteractiveUtils/src/codeview.jl
```diff
@@ -1,17 +1,46 @@
 # This file is a part of Julia. License is MIT: https://julialang.org/license
 
+# highlighting settings
+highlighting = Dict{Symbol, Bool}(
```
musm, line 4: should these dictionaries be `const`?

kimikage: At the moment, there is no public API to change the configuration, so I think specifying `const` would only make them less convenient.

Of course, I'm in favor of adding the public API. However, it should be designed to be consistent with the design for customization APIs of other features (e.g. stack traces). I think that requires a separate discussion.

musm:
I added the triage label, to discuss if this is ok to merge.

IMO I think it is and as you mention Stefan, if we had a better/faster Julia parser in Julia we would also syntax highlight Julia code in the REPL.

JeffBezanson:
:+1:

kimikage:
Thanks for the reviews and merging.

Perhaps we'll get some feedback from the nightly build users. Suggestions for improvement are welcome.

PR #36990, opened 2020-08-10
Add an option to force codegen in imaging mode

yuyichao:
This allows much easier debugging of imaging mode codegen
without hacks like specifying a dummy output and manually running
module init functions.

Also allows specializing codegen to give result more similar to
code in actual sysimg in this debugging mode. Ref #36974.

### DIFF
--- a/base/options.jl
+++ b/base/options.jl
@@ -44,6 +44,7 @@ struct JLOptions
     incremental::Int8
     image_file_specified::Int8
     warn_scope::Int8
+    image_codegen::Int8
 end
 
 # This runs early in the sysimage != is not defined yet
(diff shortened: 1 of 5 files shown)


vtjnash:
Makes some sense. I'd like to eventually eliminate non-imaging-mode branches from codegen.cpp (just using functions like `jl_link_global` to fix it up or optimize later, when desired)

yuyichao:
Sure. In any case there'll still be different LLVM IR and native code generated so there's still need for an option to force a non-default mode for debugging. The only difference would be whether such option is ready by codegen, aotcompile, or our llvm passes...

PR #36992, opened 2020-08-10
Allow LLVM to optimize GC write barriers for new objects

yuyichao:
Make alignment of tag more obvious for LLVM and
add a few optimization passes after GC lowering to let LLVM
delete some write barriers.

This fixes a regression from LLVM < 5 where the pass ordering makes this
happen automatically.

On top of https://github.com/JuliaLang/julia/pull/36970 and https://github.com/JuliaLang/julia/pull/36991, this reduces system image code size by 0.4%.

Testing this for sysimg code was the motivation for https://github.com/JuliaLang/julia/pull/36990 .

### DIFF
--- a/src/cgutils.cpp
+++ b/src/cgutils.cpp
@@ -335,6 +335,11 @@ static unsigned julia_alignment(jl_value_t *jt)
         // Array always has this alignment
         return JL_SMALL_BYTE_ALIGNMENT;
     }
+    if (jt == (jl_value_t*)jl_datatype_type) {
+        // types are never allocated in julia code/on the stack
+        // and this is the guarantee we have for the GC bits
+        return 16;
+    }
     assert(jl_is_datatype(jt) && ((jl_datatype_t*)jt)->layout);
     unsigned alignment = jl_datatype_align(jt);
     if (alignment > JL_HEAP_ALIGNMENT)
--- a/src/llvm-late-gc-lowering.cpp
+++ b/src/llvm-late-gc-lowering.cpp
@@ -1133,6 +1133,14 @@ static bool isLoadFromConstGV(LoadInst *LI)
     return false;
 }
 
+static uint64_t getLoadValueAlign(LoadInst *LI)
+{
+    MDNode *md = LI->getMetadata(LLVMContext::MD_align);
+    if (!md)
+        return 1;
+    return mdconst::extract<ConstantInt>(md->getOperand(0))->getLimitedValue();
+}
+
 static bool LooksLikeFrameRef(Value *V) {
     if (isSpecialPtr(V->getType()))
         return false;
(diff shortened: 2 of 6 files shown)


src/cgutils.cpp
```diff
@@ -330,6 +330,11 @@ static unsigned julia_alignment(jl_value_t *jt)
         // Array always has this alignment
         return JL_SMALL_BYTE_ALIGNMENT;
     }
+    if (jt == (jl_value_t*)jl_datatype_type) {
```
vtjnash, line 333:
```suggestion
    if (jl_is_mutable(jt)) {
```

Assuming llvm-alloc-opt is willing to commit to this also?

yuyichao: I don't think that's currently the case even for heap allocation. `Ref{Int8}()` on 32bit arch is only 8 byte aligned.

For stack allocation, I don't want to force any strong alignment guarantee than necessary (especially since 16byte alignment on entry isn't guaranteed everywhere). I assume it'll do a good job if aligning the allocation is benefitial. Cases that are handled in llvm-alloc-opt should be fairly easy for llvm to reason about.

For known heap address, we could certainly let llvm know when better alignment is available. However, that's also not what this function is used for. I'm pretty sure in general this function is also used for alignment of inline storage, where we certainly don't always have 16 byte of alignment.

src/llvm-late-gc-lowering.cpp
```diff
@@ -2105,11 +2113,54 @@ bool LateLowerGCFrame::CleanupIR(Function &F, State *S) {
...
+                    }
+                }
+                else if (auto LI = dyn_cast<LoadInst>(tag)) {
+                    // Make sure the load is correctly marked as aligned
+                    // since LLVM might have removed them.
+                    // We can't do this in general since the load might not be
+                    // a type in other branches.
+                    // However, it should be safe for us to do this on const globals
+                    // which should be the important cases as well.
+                    if (isLoadFromConstGV(LI) && getLoadValueAlign(LI) < 16) {
```
vtjnash, line 2137: This feels like it might not apply in imaging mode? I'm not really sure how they're represented by then here.

yuyichao: What do you mean? Most of what I see here are still loads of global variables much like how they are emitted by codegen though some have their alignment bit stripped off by LLVM (mostly LICM). There are also some calls but those are handled by the return alignment attribute and doesn't really get stripped off that easily... Since for the most important case (allocation of object of known type) it's really just a constant I don't think there's much else LLVM want to do with these other than moving them around.

And this is really an optimization to make convey the same info to LLVM while keeping the IR as simple as possible. As long as the information is correct, any cases that aren't captured by this will just use the fallback alignment assumption below. Given how far it progresses when I add an abort in the fallback assume branch below I think most of the cases are already captured here. The main (only?) exception is some phi node, which are probably either directly or indirectly caused by union splitting and that's the main reason I'm still keeping that branch...

PR #37002, opened 2020-08-11
win32: fix in_stackwalk mutex

### LINKED ISSUE #35117: make stacktrace lookup and unwinding more threadsafe
in particular, this is broken on Windows due to the `jl_in_stackwalk` global variable

vtjnash:
This should make this previously non-atomic mutex threadsafe.

Fixes #35117

### DIFF
--- a/doc/src/devdocs/locks.md
+++ b/doc/src/devdocs/locks.md
@@ -27,6 +27,7 @@ The following are definitely leaf locks (level 1), and must not try to acquire a
 >   * pagealloc
 >   * gc_perm_lock
 >   * flisp
+>   * jl_in_stackwalk (Win32)
 >
 >     > flisp itself is already threadsafe, this lock only protects the `jl_ast_context_list_t` pool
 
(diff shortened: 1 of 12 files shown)


Keno:
This is supposed to lock out doing a backtrace while adding new symbols? What happens when the profiler fires while we're in that code? Shouldn't there be a `trylock` somewhere that bails if we're trying to stackwalk while in symbol registration?

vtjnash:
I intended to have the profiler get the lock before stopping any threads. I considered alternatively using trylock, but hoped it shouldn't add too much noise (beyond the problems already seen with profiling thread scaling in the presence of locks #9224). I think I may have forgotten to finish with doing that part of the PR however (cf. your past work in https://github.com/JuliaLang/julia/pull/34454 and open issue #13294)

Keno:
Hmm, I see, I guess that works on win/mac since the profiler starts on a separate thread. This LGTM then.

Keno (approved):

Issue #37009, opened 2020-08-12
`--` is removed from ARGS


xitology:
Create an executable Julia script that shows its arguments:
```console
$ cat >showargs.jl <<EOF
#!/usr/bin/env julia
@show ARGS
EOF
$ chmod a+x showargs.jl
```
Now try to pass `--` as a command-line argument:
```console
$ ./showargs.jl --foo --bar
ARGS = ["--foo", "--bar"]
$ ./showargs.jl -- --foo --bar
ARGS = ["--foo", "--bar"]
$ ./showargs.jl --foo -- --bar
ARGS = ["--foo", "--bar"]
$ ./showargs.jl -- --foo -- --bar
ARGS = ["--foo", "--", "--bar"]
```
See how the first `--` does not appear in the list.

Note that all `--` are preserved if the script is invoked with `julia -- showargs.jl`:
```console
$ julia -- showargs.jl -- --foo -- --bar
ARGS = ["--", "--foo", "--", "--bar"]
```

The expected behaviour for `showargs.jl` is to display its arguments as is:
```console
$ ./showargs.jl --foo --bar
ARGS = ["--foo", "--bar"]
$ ./showargs.jl -- --foo --bar
ARGS = ["--", "--foo", "--bar"]
$ ./showargs.jl --foo -- --bar
ARGS = ["--foo", "--", "--bar"]
$ ./showargs.jl -- --foo -- --bar
ARGS = ["--", "--foo", "--", "--bar"]
$ julia -- showargs.jl -- --foo -- --bar
ARGS = ["--", "--foo", "--", "--bar"]
```
Indeed, a Julia script may need to implement its own command-line processing and use `--` as a customary separator between options and regular arguments, which is impossible if `--` is removed from `ARGS`.

It also doesn't match how other interpreters handle `--`. For instance, Python:
```console
$ cat >showargs.py <<EOF
#!/usr/bin/env python3
import sys
print(sys.argv[1:])
EOF
$ chmod a+x showargs.py
$ ./showargs.py -- --foo -- --bar
['--', '--foo', '--', '--bar']
$ python -- showargs.py -- --foo -- --bar
['--', '--foo', '--', '--bar']
```
I also checked Bash, Perl, and Ruby, and they all preserve `--`.

It appears that Julia tries to bypass `getopt()` and handle `--` manually since #10226. But this is unnecessary because `getopt()` with `+` in optstring does the right thing.

brenhinkeller:
[...]

PR #37010, opened 2020-08-12
let getopt() handle `--`

xitology:
This is a proposed fix for #37009. It removes manual processing of `--` and relies on `getopt()` to handle it correctly. The expected behavior:
1. If `--` appears before the first non-option argument, it forces the end of option processing and does not appear in `ARGS`.
2. If `--` appears after a non-option argument, it is preserved in `ARGS`. This `--` has no effect on option processing since it's already stopped on the first non-option argument.

### DIFF
--- a/base/client.jl
+++ b/base/client.jl
@@ -201,10 +201,6 @@ end
 include_ifexists(mod::Module, path::AbstractString) = isfile(path) && include(mod, path)
 
 function exec_options(opts)
-    if !isempty(ARGS)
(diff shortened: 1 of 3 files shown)


KristofferC:
Rebased version in https://github.com/JuliaLang/julia/pull/37010 since this branch does not allow edits from maintainers.

oscardssmith:
I believe you mean #45335

KristofferC:
Oops, yes. Thanks

PR #37017, opened 2020-08-12
Distributed: AbstractString -> String in field types

timholy:
Now that Pkg and REPL are in better shape, Distributed is turning up as the next-most-likely source of invalidations. A lot of these stem from the fact that fields in a couple of types are `AbstractString`. Is this necessary? This PR proposes to make them all concrete Strings, which will force conversion when the objects are created if necessary but will ensure that subsequent operations will be fast. Consequently, this may improve performance (but I haven't tested).

Not quite sure who is the proper person to review this, but at least we'll learn whether it passes tests this way.


StefanKarpinski:
This very much seems like the right thing to me. In structs you generally want to store non-view, plain old string.

PR #37019, opened 2020-08-12
Preserve non-concrete types in promote_typejoin

### LINKED ISSUE #35504: promote_typejoin with abstract types
Currently, `promote_typejoin(Missing, T)` returns `Any` when `T` isn't concrete. This can be annoying in cases where the type information is necessary to perform some operation (and not just for performance), like computing `zero(T)`:
```julia
julia> nt1 = NamedTuple{(:x, :y),Tuple{Union{Missing, Int},Union{Missing, Int}}}((missing, missing));

julia> eltype(nt1)
Union{Missing, Int64}

julia> sum(skipmissing(nt1))
0

julia> nt2 = NamedTuple{(:x, :y),Tuple{Union{Missing, Int},Union{Missing, Float64}}}((missing, missing));

julia> eltype(nt2)
Any

julia> sum(skipmissing(nt2))
ERROR: MethodError: no method matching zero(::Type{Any})
[...]
```
This would work if `promote_typejoin` returned `Union{Missing, Real}` since `zero` is defined for that type.

This situation happens in practice, for example when a row or group of rows happens to contain only missing values (see https://github.com/JuliaData/DataFrames.jl/pull/2196).

Would it make sense to change `promote_typejoin` to stop special-casing concrete types?

Cc: @JeffBezanson @vtjnash

vtjnash:
It is useful to have `promote_typejoin(Union{Missing, Int}, Float64}` return
`Union{Missing, Real}` instead of `Any`, in particular because `zero` is defined
on the former but not on the latter. This allows `sum(skipmissing(::NamedTuple))`
to work even when it contains only missing values.

This function, in theory, in the past was overloadable, but coverage of input types was actually rather poor previously already (one of the new tests in #36939 seemed wrong for `Base.promote_typejoin(Union{Int, U}, T) === Union{Union{Int, U}, T}`) despite an exponential increase in methods required for each new type overloaded (thus making it infeasible to correctly ever overload this function).

Fixes #35504, Closes #36939

Co-authored-by: Milan Bouchet-Valat <nalimilan@club.fr>

### DIFF
--- a/test/core.jl
+++ b/test/core.jl
@@ -168,11 +168,21 @@ for T in (Nothing, Missing)
     @test Base.promote_typejoin(Int, T) === Union{Int, T}
     @test Base.promote_typejoin(T, String) === Union{T, String}
     @test Base.promote_typejoin(Vector{Int}, T) === Union{Vector{Int}, T}
-    @test Base.promote_typejoin(Vector, T) === Any
-    @test Base.promote_typejoin(Real, T) === Any
-    @test Base.promote_typejoin(Int, String) === Any
-    @test Base.promote_typejoin(Int, Union{Float64, T}) === Any
-    @test Base.promote_typejoin(Int, Union{String, T}) === Any
+    @test Base.promote_typejoin(Vector, T) === Union{Vector, T}
+    @test Base.promote_typejoin(Real, T) === Union{Real, T}
+    for U in (String, Float64)
+        @test Base.promote_typejoin(Int, U) === typejoin(Int, U)
+        @test Base.promote_typejoin(Int, Union{U, T}) === Union{typejoin(Int, U), T}
+        @test Base.promote_typejoin(Union{Int, U}, T) === Union{Union{Int, U}, T}
+        @test Base.promote_typejoin(Union{T, U}, Int) === Union{typejoin(Int, U), T}
+        @test Base.promote_typejoin(Union{T, U}, Union{T, Int}) === Union{typejoin(Int, U), T}
+        @test Base.promote_typejoin(Union{T, U}, Union{Missing, Int}) ===
+            Union{typejoin(Int, U), T, Missing}
+        @test Base.promote_typejoin(Union{T, U}, Union{Nothing, Int}) ===
+            Union{typejoin(Int, U), T, Nothing}
+        @test Base.promote_typejoin(Union{T, Nothing, U}, Union{Nothing, Missing, Int}) ===
+            Union{typejoin(Int, U), T, Nothing, Missing}
+    end
     @test Base.promote_typejoin(T, Union{}) === T
     @test Base.promote_typejoin(Union{}, T) === T
 end
--- a/base/promotion.jl
+++ b/base/promotion.jl
@@ -128,20 +128,12 @@ Compute a type that contains both `T` and `S`, which could be
 either a parent of both types, or a `Union` if appropriate.
 Falls back to [`typejoin`](@ref).
 """
-promote_typejoin(@nospecialize(a), @nospecialize(b)) = _promote_typejoin(a, b)::Type
-_promote_typejoin(@nospecialize(a), @nospecialize(b)) = typejoin(a, b)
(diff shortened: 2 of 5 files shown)


test/core.jl
```diff
@@ -168,11 +168,29 @@ for T in (Nothing, Missing)
...
-    @test Base.promote_typejoin(Int, String) === Any
-    @test Base.promote_typejoin(Int, Union{Float64, T}) === Any
-    @test Base.promote_typejoin(Int, Union{String, T}) === Any
+    @test Base.promote_typejoin(Vector, T) === Union{Vector, T}
+    @test Base.promote_typejoin(Real, T) === Union{Real, T}
+    for U in (String, Float64)
+        @test Base.promote_typejoin(Int, U) === typejoin(Int, U)
+        @test Base.promote_typejoin(Int, Union{U, T}) === Union{typejoin(Int, U), T}
+        @test Base.promote_typejoin(Union{Int, U}, T) === Union{Union{Int, U}, T}
+        @test Base.promote_typejoin(Union{T, U}, Int64) === Union{typejoin(Int, U), T}
```
martinholters, line 177: Although inconsequential, this mix of `Int` and `Int64` (repeated a few times below) gave me a pause. I assume it is unintended?

test/core.jl
```diff
@@ -168,11 +168,29 @@ for T in (Nothing, Missing)
...
+        @test Base.promote_typejoin(Union{T, Nothing, U}, Union{Nothing, Missing, Int64}) ===
+            Union{typejoin(Int, U), T, Nothing, Missing}
+        @test Base.promote_typejoin(Union{T, Missing, U}, Union{Nothing, Missing, Int64}) ===
+            Union{typejoin(Int, U), T, Nothing, Missing}
+        @test Base.promote_typejoin(Union{Nothing, Missing, U},
+                                    Union{Nothing, Missing, Int64}) ===
+            Union{typejoin(Int, U), Nothing, Missing}
+        @test Base.promote_typejoin(Union{Nothing, Missing, U},
+                                    Union{Nothing, Missing, Int64}) ===
+            Union{typejoin(Int, U), Nothing, Missing}
```
martinholters, line 192: Duplicate of the test above?

test/core.jl
```diff
@@ -168,11 +168,29 @@ for T in (Nothing, Missing)
...
+            Union{typejoin(Int, U), T, Missing}
+        @test Base.promote_typejoin(Union{T, U}, Union{Nothing, Int64}) ===
+            Union{typejoin(Int, U), T, Nothing}
+        @test Base.promote_typejoin(Union{T, Nothing, U}, Union{Nothing, Missing, Int64}) ===
+            Union{typejoin(Int, U), T, Nothing, Missing}
+        @test Base.promote_typejoin(Union{T, Missing, U}, Union{Nothing, Missing, Int64}) ===
+            Union{typejoin(Int, U), T, Nothing, Missing}
+        @test Base.promote_typejoin(Union{Nothing, Missing, U},
+                                    Union{Nothing, Missing, Int64}) ===
+            Union{typejoin(Int, U), Nothing, Missing}
```
martinholters, line 189: Already tested above (for `T === Nothing`).

martinholters (commented):
I can't really judge the implications of this change, but it looks reasonable and the implementation seems sound. The tests could use a minor revision, though.

nalimilan (commented):
Thanks for taking this up! I don’t have much to say about the implementation as long as tests pass.

JeffBezanson (approved):

cstjean:
Before this PR, we could implement `_promote_typejoin` so that unions of our own types didn't collapse to `Any`

```julia
 Base._promote_typejoin(::Type{MyType{A}}, ::Type{T}) where {T, A} = 
     isconcretetype(T) || T === Union{} ? Union{T, MyType{A}} : Any 
 Base._promote_typejoin(::Type{T}, ::Type{MyType{A}}) where {T, A} = 
     isconcretetype(T) || T === Union{} ? Union{T, MyType{A}} : Any 
```

We relied on that in our code, is there any work-around on 1.6? Should we overwrite the `promote_typesubtract` method?

Issue #36106, opened 2020-06-01
map, collect and similar for custom arrays


nalimilan:
The current design of `map`, `collect` and `similar` seems problematic for custom arrays which are associated with a custom element type, like `CategoricalArray` and its `CategoricalValue` type.

In CategoricalArrays.jl, I would like `map(f, ::CategoricalArray)` to return a `CategoricalArray` if `f` returns only `Union{CategoricalValue, Missing}` values. This is natural in particular so that `map(identity,  ::CategoricalArray)` returns a `CategoricalArray`.

To achieve this, I could define `Base.map(f, ::CategoricalArray)`, but that would require duplicating tricky code from Base since it needs to handle eltype widening and so on. So I tried to define `similar` so that `collect`, that `map` uses under the hood, returns a `CategoricalArray` when appropriate. But `collect` uses `similar(1:1, T, axes(itr))`, so I have to override `similar(::AbstractRange, ::Type{<:Union{CategoricalValue, Missing}})`. For consistency I also have to define similar methods for `AbstractArray`, `Array`, `Vector` and `Matrix` (due to ambiguities).

Doing that has two consequences:
- The first is that `collect(::CategoricalArray)` always returns a `CategoricalArray`. This makes sense actually since `Array{CategoricalValue}` is an inefficient type. But that seems to go against the docstring for `collect` which says that it returns an `Array`.
- The second, more serious issue is that `getindex(::Array{<:CategoricalValue}, ::Array)` also returns a `CategoricalArray`. This doesn't sound correct.

This leads me to raise two questions/proposals:
1. Should the `collect` docstring be made less strict? It sounds useful to be able to collect the contents of an interator into the most natural/efficient array type. If one really wants an `Array` better do `Array(itr)` -- otherwise `collect` is redundant.
2. Should `collect` use a new system like `similar(AbstractArray, T, axes(itr))` instead of `similar(1:1, T, axes(itr))`? That would allow specifying that the most appropriate `AbstractArray{<:T}` is requested rather than `Array`. That would differ from `getindex` which really wants the type of the input array. This could be introduced without breakage by having it fall back to `similar(Array, ...)`.

mbauman:
Ref:

* #18161: similar and traits
* #16029: `collect()` into `Vector()`? 
* #34478: Standard API for push_widen! pattern 


[...]

PR #37028, opened 2020-08-13
Fix type of allocated array when broadcasting type unstable function

nalimilan:
We need to call similar on the `Broadcasted` object rather than on dest array. Otherwise the `BroadcastStyle` isn't taken into account when allocating new array due to function returning elements of different types.

See https://github.com/JuliaLang/julia/issues/36106#issuecomment-637022450.

### DIFF
--- a/test/broadcast.jl
+++ b/test/broadcast.jl
@@ -510,13 +510,17 @@ Base.BroadcastStyle(::Type{T}) where {T<:AD2Dim} = AD2DimStyle()
     aa = Array19745(a)
     fadd(aa) = aa .+ 1
     fadd2(aa) = aa .+ 1 .* 2
+    fadd3(aa) = aa .+ [missing; 1:9]
     fprod(aa) = aa .* aa'
     @test a .+ 1  == @inferred(fadd(aa))
     @test a .+ 1 .* 2  == @inferred(fadd2(aa))
     @test a .* a' == @inferred(fprod(aa))
+    @test isequal(a .+ [missing; 1:9], fadd3(aa))
+    @test_broken Core.Compiler.return_type(fadd3, (typeof(aa),)) <: Array19745{<:Union{Float64, Missing}}
     @test isa(aa .+ 1, Array19745)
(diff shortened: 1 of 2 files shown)


test/broadcast.jl
```diff
@@ -510,13 +510,17 @@ Base.BroadcastStyle(::Type{T}) where {T<:AD2Dim} = AD2DimStyle()
     aa = Array19745(a)
     fadd(aa) = aa .+ 1
     fadd2(aa) = aa .+ 1 .* 2
+    fadd3(aa) = aa .+ [missing; 1:9]
     fprod(aa) = aa .* aa'
     @test a .+ 1  == @inferred(fadd(aa))
     @test a .+ 1 .* 2  == @inferred(fadd2(aa))
     @test a .* a' == @inferred(fprod(aa))
+    @test isequal(a .+ [missing; 1:9], fadd3(aa))
+    @test_broken Core.Compiler.return_type(fadd3, (typeof(aa),)) <: Array19745{<:Union{Float64, Missing}}
```
nalimilan, line 519: This is https://github.com/JuliaLang/julia/pull/30485.

mbauman:
Ah, this makes sense.  So the destination is initially created with:

```
dest = similar(bc, T)
```

Subsequent widenings had previously been

```
dest = similar(dest, T′)
```

but that privileges the initial `similar(bc, T)` and makes things order-dependent.  Always going back to the `bc` object here definitely seems to be the right thing to do.

mbauman (approved):

JeffBezanson:
Any chance somebody could trip over this? Maybe a NEWS item?

nalimilan:
I don't think it should surprise anybody. If this changes the behavior for you, then you obtained different array (container) types depending on whether your function returned elements of the same type or not. Package authors that experienced this are more likely to have worked around this by defining `similar` methods on the array type in addition to methods on `Broadcasted` (like I did on CategoricalArrays, among other reasons). Also it's so technical that probably nobody will understand the meaning the a NEWS.md entry mentioning it. :-p

PR #37034, opened 2020-08-13
add a TOML standard library to Julia

KristofferC:
This adds a TOML standard library for parsing and writing TOML files. ~~The source files currently reside in https://github.com/KristofferC/TOML.jl. The reason for this is that there is already a TOML repository in the JuliaLang organization which should probably be moved.~~

The current docs can be found at https://kristofferc.github.io/TOML.jl/dev/.

The parser part is the same as in https://github.com/JuliaLang/julia/pull/36018 (with the review comments applied).

After this is merged, we should move the parser part into Base and rebase https://github.com/JuliaLang/julia/pull/36018  on top of it.

~~Since the parser part is going to be in Base, it isn't clear to me whether all files for the stdlib should be in this repo or if it still makes sense to have a separate repo for e.g. issues.~~

cc @StefanKarpinski, @staticfloat


staticfloat:
If it's going to have tight integration with code in `Base` (like the `Artifacts` stdlib will) I think it's better to have the code live here in the julia repository.

It looks like the windows tests are failing because it's trying to import JSON and can't?

KristofferC:
Only as tight of an integration that code loading can actually use it. It's pretty nice to have all issues in one place. Hm.

staticfloat:
I think stdlib tests don't get their `extras` targets populated.  :(

KristofferC:
Yeah, they don't run with Pkg.test.

PR #37036, opened 2020-08-13
Allow expressions as logging keys

simonbyrne:
These will be converted to Symbols, similar to how positional args are handled


c42f (approved):
This makes sense to me.  It's a natural generalization of our existing handling of non-assignment expressions which turns them into key value pairs, as in

```julia
julia> @info "hi" sum([1,2])
┌ Info: hi
└   sum([1, 2]) = 3
```

PR #37048, opened 2020-08-14
Fix incorrect source info in docstrings

### LINKED ISSUE #36906: Bad docstring path/linenumber if in macro call
The filename and line numbers for a docstring are not properly stored if the definition is wrapped in a macro call, e.g. in the following case:

```julia
"""
    struct A
"""
Base.@kwdef struct A
  a::Int = 1
end
```

For an MWE, `include`-ing this:

```julia
"""
    struct A
"""
Base.@kwdef struct A
  a::Int = 1
end

"""
    struct B
"""
struct B
  a::Int
end

Base.@kwdef struct C
  a::Int = 1
end

@doc """
    struct C
""" C
```

will lead to this when fetching the docstring metadata (note the missing `:path` and `:linenumber` for `A`):

```
julia> @doc(A).meta[:results][1].data
Dict{Symbol,Any} with 6 entries:
  :typesig    => Union{}
  :module     => Main
  :linenumber => 0
  :binding    => A
  :path       => ""
  :fields     => Dict{Any,Any}()

julia> @doc(B).meta[:results][1].data
Dict{Symbol,Any} with 6 entries:
  :typesig    => Union{}
  :module     => Main
  :linenumber => 14
  :binding    => B
  :path       => "/home/mortenpi/Julia/test/test.jl"
  :fields     => Dict{Any,Any}()

julia> @doc(C).meta[:results][1].data
Dict{Symbol,Any} with 5 entries:
  :typesig    => Union{}
  :module     => Main
  :linenumber => 25
  :binding    => C
  :path       => "/home/mortenpi/Julia/test/test.jl"
```

X-ref: JuliaDocs/Documenter.jl#1374

MichaelHatherly:
Fixes #36906, where source info wasn't being passed through `@doc` when it was applied to definitions defined within macros.


fredrikekre:
@mortenpi can you review perhaps? LGTM but I am not too familiar with this code.

base/docs/Docs.jl
```diff
@@ -509,6 +509,8 @@ function docm(source::LineNumberNode, mod::Module, ex)
         return REPL.lookup_doc(ex)
     end
 end
+# Drop incorrect line numbers produced by nested macro calls.
+docm(source::LineNumberNode, mod::Module, _, _, x...) = docm(source, mod, x...)
```
mortenpi, line 513: Just out of curiousity: when and where does this signature actually get called?

test/docs.jl
```diff
@@ -464,10 +471,18 @@ let md = meta(MacroGenerated)[@var(MacroGenerated.g)]
...
+    @test md.docs[Tuple{Any}].data[:linenumber] == MacroGenerated.LINE_NUMBER_G
 end
 
 @test isdefined(MacroGenerated, :_g)
 
+let md = meta(MacroGenerated)[@var(MacroGenerated.T)]
+    @test md.order == Type[Union{}]
+    @test docstrings_equal(md.docs[Union{}], doc"T")
+    @test md.docs[Union{}].data[:linenumber] == MacroGenerated.LINE_NUMBER_T
+    @test md.docs[Union{}].data[:path] == @__FILE__()
```
mortenpi, line 483: Might be worth padding the `:path` tests for the two previous ones too?

mortenpi (approved):
LGTM as well. As far as I can tell, we're now correctly passing on the line number of the macro call to the docstring, and the tests cover this case now.

base/docs/Docs.jl
MichaelHatherly: It's called indirectly by the `__doc__!` method above since it generates macro calls to `@doc` with two sets of source/module info, this strips out the incorrect ones and passes back to `docm`. Took me quite a while to debug my way through it initially.

MichaelHatherly:
Should be good to go once CI finishes.

PR #37052, opened 2020-08-14
allow at-a"..." macro syntax

### LINKED ISSUE #37024: parse @a"..." as @a("...")
I think this would be consistent with what we currently do for square and curly braces. At first, this may seem redundant, because we already have custom string literals, but the difference would be that with this syntax, string interpolation would get expanded at parse time, just like for regular non-raw strings, whereas custom literals are always parsed as raw literals and if a custom string wants interpolation, it has to call `Meta.parse` itself at macro expansion time, which has some caveats. This is currently a syntax error, so this should be fine to change in 1.x.

simeonschaub:
This lets you avoid the space in macro calls after the macro name for string as well as command literals. It also consumes a literal suffix and passes that to the macro as a second argument.
closes #37024

### DIFF
--- a/src/julia-parser.scm
+++ b/src/julia-parser.scm
@@ -1254,6 +1254,17 @@
              (if (and (or (symbol? ex) (valid-modref? ex))
                       (not (operator? ex))
                       (not (ts:space? s)))
+               (if macrocall?
+                 ;; @a"str" => @a("str")
+                 (let* ((macstr (parse-atom s))
+                        (nxt (peek-token s))
+                        (macname ex))
+                   (if (and (symbol? nxt) (not (operator? nxt))
+                            (not (ts:space? s)))
+                       ;; string literal suffix, "s"x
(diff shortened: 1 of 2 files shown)


StefanKarpinski:
I'm ok with allowing this syntax, but I don't love that it's slightly subtly different from non-standard string macros.

fredrikekre:
Wouldn't this be a better syntax for string macros instead?

JeffBezanson:
I find this a bit too fiddly. It's especially confusing that it allows `@a"str"suffix` even though it does not call a string macro. But even without that the ambiguity between `@a"str"`, `@a "str"`, and `a"str"` is too confusing.

StefanKarpinski:
The current error is a bit inscrutable:
```jl
julia> :(@a"str")
ERROR: syntax: invalid macro usage "@(@a_str # REPL[54], line 1 "str")"
```

PR #37054, opened 2020-08-14
add at-__SOURCE__ macro

simeonschaub:
I recently wanted this and it seems generally useful.

### DIFF
--- a/base/essentials.jl
+++ b/base/essentials.jl
@@ -855,6 +855,20 @@ julia> peek(b, Char)
 function peek end
 
 """
+    @__SOURCE__ -> LineNumberNode
+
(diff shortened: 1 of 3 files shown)


StefanKarpinski:
I was going to comment that I wasn't too keen on exposing `LineNumberNode` like this, but it turns out that it's exported from `Base` (why??) so I guess we're not exposing it any more than before. Still, I'm not sure this is necessary given that we have both `@__LINE__` and `@__FILE__` which already give you both parts.

JeffBezanson:
-1 from me too, it's redundant.



> why??

You hit these when writing macros, and sometimes people want to e.g. strip out or replace them. Doesn't mean it *needs* to be exported, but it does have a use.

PR #37055, opened 2020-08-14
add compat entry for sincospi

simeonschaub:
Another omission from #35816...

### DIFF
--- a/base/special/trig.jl
+++ b/base/special/trig.jl
@@ -864,6 +864,9 @@ end
     sincospi(x)
 
 Simultaneously compute `sinpi(x)` and `cospi(x)`, where the `x` is in radians.
+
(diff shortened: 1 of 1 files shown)


base/special/trig.jl
```diff
@@ -864,6 +864,9 @@ end
     sincospi(x)
 
 Simultaneously compute `sinpi(x)` and `cospi(x)`, where the `x` is in radians.
+
+!!! compat "Julia 1.6"
+    The method which accepts a type requires Julia 1.6 or later.
```
kimikage, line 869: "accepts a type"?

simeonschaub: Oops... Copy and paste error.

Issue #36551, opened 2020-07-06
Dense * Tridiagonal = Sparse?


ChrisRackauckas:
```julia
julia> using LinearAlgebra

julia> N = 8
8

julia> Mx = Tridiagonal([1.0 for i in 1:N-1],[-2.0 for i in 1:N],[1.0 for i in 1:N-1])
8×8 Tridiagonal{Float64,Array{Float64,1}}:
 -2.0   1.0    ⋅     ⋅     ⋅     ⋅     ⋅     ⋅
  1.0  -2.0   1.0    ⋅     ⋅     ⋅     ⋅     ⋅
   ⋅    1.0  -2.0   1.0    ⋅     ⋅     ⋅     ⋅
   ⋅     ⋅    1.0  -2.0   1.0    ⋅     ⋅     ⋅
   ⋅     ⋅     ⋅    1.0  -2.0   1.0    ⋅     ⋅
   ⋅     ⋅     ⋅     ⋅    1.0  -2.0   1.0    ⋅
   ⋅     ⋅     ⋅     ⋅     ⋅    1.0  -2.0   1.0
   ⋅     ⋅     ⋅     ⋅     ⋅     ⋅    1.0  -2.0

julia> A = rand(N,N)
8×8 Array{Float64,2}:
 0.52234   0.075572   0.492018  0.534107  …  0.11925    0.933859   0.927217
 0.554405  0.674733   0.322482  0.895656     0.025754   0.842152   0.810342
 0.116686  0.603084   0.130214  0.475674     0.0697535  0.97639    0.713087
 0.717719  0.211584   0.568428  0.294381     0.643912   0.270161   0.337518
 0.251835  0.569177   0.554402  0.577696     0.535705   0.854619   0.145042
 0.267439  0.0778166  0.480589  0.133326  …  0.365856   0.543645   0.157496
 0.366063  0.68394    0.365163  0.443594     0.75635    0.0335027  0.0916263
 0.423794  0.728052   0.885349  0.717036     0.24141    0.017163   0.651004

julia> AMx = A*Mx
8×8 SparseArrays.SparseMatrixCSC{Float64,Int64} with 64 stored entries:
  [1, 1]  =  -0.969108
  [2, 1]  =  -0.434078
  [3, 1]  =  0.369713
  [4, 1]  =  -1.22385
  [5, 1]  =  0.0655065
  [6, 1]  =  -0.457062
  [7, 1]  =  -0.0481849
[...]

PR #37065, opened 2020-08-16
Let `muladd` accept arrays

mcabbott:
This extends `Base.muladd` to work on arrays, using `mul!`, which makes some combinations more efficient:
```julia
julia> A = rand(20,5); x = rand(5, 1000); b = rand(20);

julia> @btime $A * $x .+ $b;
  29.787 μs (4 allocations: 312.66 KiB)

julia> @btime muladd($A, $x, $b);
  22.322 μs (2 allocations: 156.33 KiB)
```
Motivated by things like https://github.com/FluxML/Flux.jl/pull/1272, where it would be convenient to fuse these two, rather than fusing addition & function application `tanh.((W*x) .+ b)` with broadcasting.

~~Needs tests, and thinking about `muladd(::Adjoint, ...)` & `muladd(..., ::Number)` cases,~~ if anyone thinks this is a good idea.

### DIFF
--- a/stdlib/LinearAlgebra/src/matmul.jl
+++ b/stdlib/LinearAlgebra/src/matmul.jl
@@ -183,6 +183,37 @@ for elty in (Float32,Float64)
 end
 
 """
+    muladd(A, y, z)
+
+Combined multiply-add, `A*y .+ z`, for either matrix-matrix multiplication or
+matrix-vector multiplication. The result is always the size of `A*y`, although
+`z` have fewer dimensions.
+
+# Examples
+```jldoctest
+julia> A=[1.0 2.0; 3.0 4.0]; B=[1.0 1.0; 1.0 1.0]; C=[0, 100];
+
+julia> muladd(A, B, C)
+2×2 Matrix{Float64}:
+   3.0    3.0
+ 107.0  107.0
+```
+"""
+function Base.muladd(A::AbstractMatrix{TA}, y::AbstractVector{Ty}, z::AbstractVector{Tz}) where {TA, Ty, Tz}
+    T = promote_type(TA, Ty, Tz)
+    C = similar(A, T, size(A,1))
+    C .= z
+    mul!(C, A, y, true, true)
+end
+
+function Base.muladd(A::AbstractMatrix{TA}, y::AbstractMatrix{Ty}, z::AbstractVecOrMat{Tz}) where {TA, Ty, Tz}
+    T = promote_type(TA, Ty, Tz)
+    C = similar(A, T, size(A,1), size(y,2))
+    C .= z
+    mul!(C, A, y, true, true)
+end
+
+"""
     mul!(Y, A, B) -> Y
 
 Calculates the matrix-matrix or matrix-vector product ``AB`` and stores the result in `Y`,
--- a/stdlib/LinearAlgebra/src/matmul.jl
+++ b/stdlib/LinearAlgebra/src/matmul.jl
@@ -185,9 +185,8 @@ end
 """
     muladd(A, y, z)
 
-Combined multiply-add, `A*y .+ z`, for either matrix-matrix multiplication or
-matrix-vector multiplication. The result is always the size of `A*y`, although
-`z` may have fewer dimensions.
+Combined multiply-add, `A*y .+ z`, for matrix-matrix or matrix-vector multiplication.
+The result is always the size of `A*y`, but `z` may be smaller, or a scalar.
 
 !!! compat "Julia 1.6"
      These methods require Julia 1.6 or later.
@@ -220,18 +219,13 @@ Base.muladd(x::AdjointAbsVec, A::AbstractMatrix, z) = muladd(A', x', z')'
 Base.muladd(x::TransposeAbsVec, A::AbstractMatrix, z) = transpose(muladd(transpose(A), transpose(x), transpose(z)))
 
 function Base.muladd(u::AbstractVector, v::AdjOrTransAbsVec, z)
-    if z isa AbstractArray
-        ndims(z) > 2 && throw(DimensionMismatch("cannot broadcast array to have fewer dimensions"))
-    end
+    ndims(z) > 2 && throw(DimensionMismatch("cannot broadcast array to have fewer dimensions"))
     (u .* v) .+ z
 end
 
 function Base.muladd(u::AdjOrTransAbsVec, v::AbstractVector, z)
     uv = _dot_nonrecursive(u, v)
-    if z isa AbstractArray
-        uv isa AbstractArray && ndims(uv) <= ndims(z) ||
-            throw(DimensionMismatch("cannot broadcast array to have fewer dimensions"))
-    end
+    ndims(z) > ndims(uv) && throw(DimensionMismatch("cannot broadcast array to have fewer dimensions"))
     uv .+ z
 end
 
--- a/stdlib/LinearAlgebra/src/matmul.jl
+++ b/stdlib/LinearAlgebra/src/matmul.jl
@@ -186,16 +186,16 @@ end
     muladd(A, y, z)
 
 Combined multiply-add, `A*y .+ z`, for matrix-matrix or matrix-vector multiplication.
-The result is always the size of `A*y`, but `z` may be smaller, or a scalar.
+The result is always the same size as `A*y`, but `z` may be smaller, or a scalar.
 
 !!! compat "Julia 1.6"
      These methods require Julia 1.6 or later.
 
 # Examples
 ```jldoctest
-julia> A=[1.0 2.0; 3.0 4.0]; B=[1.0 1.0; 1.0 1.0]; C=[0, 100];
+julia> A=[1.0 2.0; 3.0 4.0]; B=[1.0 1.0; 1.0 1.0]; z=[0, 100];
 
-julia> muladd(A, B, C)
+julia> muladd(A, B, z)
 2×2 Matrix{Float64}:
    3.0    3.0
  107.0  107.0
--- a/stdlib/LinearAlgebra/src/matmul.jl
+++ b/stdlib/LinearAlgebra/src/matmul.jl
@@ -203,14 +203,14 @@ julia> muladd(A, B, C)
 ```
 """
 function Base.muladd(A::AbstractMatrix{TA}, y::AbstractVector{Ty}, z) where {TA, Ty}
-    T = promote_type(TA, Ty, z isa AbstractArray ? eltype(z) : typeof(z))
+    T = promote_type(TA, Ty, eltype(z))
     C = similar(A, T, axes(A,1))
     C .= z
     mul!(C, A, y, true, true)
 end
 
 function Base.muladd(A::AbstractMatrix{TA}, B::AbstractMatrix{TB}, z) where {TA, TB}
-    T = promote_type(TA, TB, z isa AbstractArray ? eltype(z) : typeof(z))
+    T = promote_type(TA, TB, eltype(z))
     C = similar(A, T, axes(A,1), axes(B,2))
     C .= z
     mul!(C, A, B, true, true)
--- a/stdlib/LinearAlgebra/src/matmul.jl
+++ b/stdlib/LinearAlgebra/src/matmul.jl
@@ -187,7 +187,10 @@ end
 
 Combined multiply-add, `A*y .+ z`, for either matrix-matrix multiplication or
 matrix-vector multiplication. The result is always the size of `A*y`, although
-`z` have fewer dimensions.
+`z` may have fewer dimensions.
+
+!!! compat "Julia 1.6"
+     These methods require Julia 1.6 or later.
 
 # Examples
 ```jldoctest
--- a/stdlib/LinearAlgebra/src/matmul.jl
+++ b/stdlib/LinearAlgebra/src/matmul.jl
@@ -202,18 +202,37 @@ julia> muladd(A, B, C)
  107.0  107.0
 ```
 """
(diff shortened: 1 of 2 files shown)


mcabbott:
Now has tests, and `muladd(A, x, b)` adds any `b` for which `C = A*x; C = C .+ b` would work. Behaves the obvious way when A is an Adjoint vector.

stdlib/LinearAlgebra/src/matmul.jl
```diff
@@ -182,6 +182,59 @@ for elty in (Float32,Float64)
...
+julia> A=[1.0 2.0; 3.0 4.0]; B=[1.0 1.0; 1.0 1.0]; C=[0, 100];
+
+julia> muladd(A, B, C)
+2×2 Matrix{Float64}:
+   3.0    3.0
+ 107.0  107.0
+```
+"""
+function Base.muladd(A::AbstractMatrix{TA}, y::AbstractVector{Ty}, z) where {TA, Ty}
+    T = promote_type(TA, Ty, z isa AbstractArray ? eltype(z) : typeof(z))
```
dkarrasch, line 206: Which cases do you have in mind for which `eltype` doesn't work? For `Number`s, at least, it does.

stdlib/LinearAlgebra/src/matmul.jl
```diff
@@ -182,6 +182,59 @@ for elty in (Float32,Float64)
...
+    mul!(C, A, B, true, true)
+end
+
+Base.muladd(x::AdjointAbsVec, A::AbstractMatrix, z) = muladd(A', x', z')'
+Base.muladd(x::TransposeAbsVec, A::AbstractMatrix, z) = transpose(muladd(transpose(A), transpose(x), transpose(z)))
+
+function Base.muladd(u::AbstractVector, v::AdjOrTransAbsVec, z)
+    if z isa AbstractArray
+        ndims(z) > 2 && throw(DimensionMismatch("cannot broadcast array to have fewer dimensions"))
+    end
```
dkarrasch, line 225: Wouldn't the next line complain anyway in that case?

stdlib/LinearAlgebra/src/matmul.jl
mcabbott: I was imagining arrays of matrices, e.g. I think many things in LinearAlgebra work if, instead of complex numbers, you use as elements 2x2 matrices.

stdlib/LinearAlgebra/src/matmul.jl
mcabbott: The next line would permit a higher-dimensional `z`, such as `muladd(1:3, ones(2)', ones(3,2,4))`. I thought that this should be forbidden, for consistency with the methods which use `mul!`.

stdlib/LinearAlgebra/src/matmul.jl
dkarrasch: Wait, but arrays of matrices fall in the first (i.e., the `eltype`) category, right? I was wondering if you can't simply call `eltype` on any type of `z` you can imagine. Or conversely, for which type of `z` do you think you need `typeof(z)`?

mcabbott: Oh you might be right, sorry. My pauli matrices would go wrong in `A*x .+ z` which `muladd` is supposed to mimmic, an on the next line’s broadcast too.

dkarrasch: In any case, if you imagine such a use case, I think we should nail the intended behavior with a test.

mcabbott: I've now added some vector-of-matrices tests. One fails I think because of a 5-arg mul problem but I need to look a bit more. 

Besides `eltype(::Number)`, it's safe to use `ndims(::Number)==0` right?

stdlib/LinearAlgebra/src/matmul.jl
mcabbott: Now added a test of this error, too.

stdlib/LinearAlgebra/src/matmul.jl
dkarrasch: Yes, there is
```julia
julia> @which ndims(3.)
ndims(x::Number) in Base at number.jl:67
```

stdlib/LinearAlgebra/src/matmul.jl
```diff
@@ -182,6 +182,53 @@ for elty in (Float32,Float64)
     end
 end
 
+"""
+    muladd(A, y, z)
+
+Combined multiply-add, `A*y .+ z`, for matrix-matrix or matrix-vector multiplication.
+The result is always the size of `A*y`, but `z` may be smaller, or a scalar.
```
dkarrasch, line 189: I'm not a native speaker, but shouldn't it be
```suggestion
The result is always of the size of `A*y`, but `z` may be smaller, or a scalar.
```
? Otherwise, one may expect a tuple of `Int`s.

stdlib/LinearAlgebra/src/matmul.jl
```diff
@@ -182,6 +182,53 @@ for elty in (Float32,Float64)
...
+Base.muladd(x::TransposeAbsVec, A::AbstractMatrix, z) = transpose(muladd(transpose(A), transpose(x), transpose(z)))
+
+function Base.muladd(u::AbstractVector, v::AdjOrTransAbsVec, z)
+    ndims(z) > 2 && throw(DimensionMismatch("cannot broadcast array to have fewer dimensions"))
+    (u .* v) .+ z
+end
+
+function Base.muladd(u::AdjOrTransAbsVec, v::AbstractVector, z)
+    uv = _dot_nonrecursive(u, v)
+    ndims(z) > ndims(uv) && throw(DimensionMismatch("cannot broadcast array to have fewer dimensions"))
```
dkarrasch, line 228: Is there any way one could have this test before computing `_dot_nonrecursive`?

stdlib/LinearAlgebra/src/matmul.jl
mcabbott: Thanks I didn't spot that reading. How about "The result is always the same size as `A*y`,"?

stdlib/LinearAlgebra/src/matmul.jl
mcabbott: This is the only multiplication which produces a scalar (on ordinary arrays) and would then expect z nothing more than a scalar. 

On arrays of arrays, it may produce one sub-array, and in this case `z` is allowed to be the same as that. One could try to infer whether this will happen before  `_dot_nonrecursive`, but what would we gain from the extra complexity?

dkarrasch: I just thought that one may wish to throw before doing "heavy" computations, and not first work hard just to eventually realize that it wouldn't work.

If it's a mess to figure that out beforehand, I think it's fine.

mcabbott: Worst case would be something like `Any[ones(2,1) for _ in 1:3]' * Any[ones(2) for _ in 1:3] .+ [1]` where you can't tell from the types that it's safe. 

I guess an inner product will seldom be the biggest step in the your calculation, too, so hopefully not a big penalty to wait for the error.

mcabbott:
Bump?

mateuszbaran:
I think this change broke the following:
On 1.5.2:
```julia
julia> using LinearAlgebra

julia> muladd([2 2; 1 2], [2 1; 1 2], I)
2×2 Array{Int64,2}:
 7  6
 4  6
```
while currently on master:
```julia
julia> using LinearAlgebra

julia> muladd([2 2; 1 2], [2 1; 1 2], I)
ERROR: MethodError: no method matching length(::UniformScaling{Bool})
Closest candidates are:
  length(::Union{Base.KeySet, Base.ValueIterator}) at abstractdict.jl:54
  length(::Union{Adjoint{T, S}, Transpose{T, S}} where S where T) at /home/mateusz/repos/julia/julia/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/adjtrans.jl:194
  length(::IdDict) at iddict.jl:136
  ...
Stacktrace:
 [1] _similar_for(c::UnitRange{Int64}, #unused#::Type{Bool}, itr::UniformScaling{Bool}, #unused#::Base.HasLength)
   @ Base ./array.jl:575
 [2] _collect(cont::UnitRange{Int64}, itr::UniformScaling{Bool}, #unused#::Base.HasEltype, isz::Base.HasLength)
   @ Base ./array.jl:608
 [3] collect(itr::UniformScaling{Bool})
   @ Base ./array.jl:602
 [4] broadcastable(x::UniformScaling{Bool})
   @ Base.Broadcast ./broadcast.jl:682
 [5] broadcasted
   @ ./broadcast.jl:1303 [inlined]
 [6] muladd(A::Matrix{Int64}, B::Matrix{Int64}, z::UniformScaling{Bool})
   @ LinearAlgebra ~/repos/julia/julia/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/matmul.jl:214
 [7] top-level scope
   @ REPL[12]:1

julia> versioninfo()
Julia Version 1.6.0-DEV.1384
Commit 4711fc3cdf (2020-10-30 14:50 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: Intel(R) Core(TM) i7-4800MQ CPU @ 2.70GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.0 (ORCJIT, haswell)
```

mcabbott:
Oops, I was working on the assumption that the sole existing use for `muladd` was with numbers. I left `z` untyped just for simplicity but it could be constrained as you suggest.

The error here is from `[6 6; 4 5] .+ I`, I'm somewhat surprised that doesn't work. If it did work, then this method would be more efficient than the fallback. Might it be worth fixing? 

However, if there are other existing uses of `muladd` with matrices, then this whole thing may need to tread more carefully. For instance, on master this hits this PR's `mul!(similar(...` method:
```julia
julia> muladd(Diagonal([1,2]), [3,4], [5,6])
2-element SparseArrays.SparseVector{Int64, Int64} with 2 stored entries:
  [1]  =  8
  [2]  =  14

julia> muladd(Diagonal([1,2]), [3,4], [5,6]) # on Julia 1.5
2-element Array{Int64,1}:
  8
 14
```

dkarrasch:
I was thinking whether there is some way to deduce the _type_ of `A*B` and have `C` `similar` to that to do `mul!(C, A, B, ...)`. I tried a little, but I don't think I found a good solution. At times, even that return type happens to be suboptimal. I don't remember the issue, but there was a difference whether the structured matrix was the left or the right factor, because we usually (always?) take the right factor and make `C` similar to it.

mcabbott:
I think quite a lot of the structured types returned by `A*B` are destroyed by `.+z`. There are a few exceptions:
```julia
di = Diagonal(1:3)
ut = UpperTriangular(rand(1:9, 3,3))

di * di .+ di isa Diagonal
ut * ut .+ ut isa UpperTriangular
ut * ut .+ di isa UpperTriangular
```

```

**Comment by mateuszbaran** [comment]
I think there may be a lot more patterns where `A*B` is sparse but `z` is dense. `Bidiagonal`, `Tridiagonal`, `SymTridiagonal`, `SparseMatrixCSC`, etc. Determining the correct result type of `Base.muladd(A::AbstractMatrix{TA}, B::AbstractMatrix{TB}, z) where {TA, TB}` would be very complex -- the signature is very broad. Maybe for now you could just implement this `muladd` only for either `A` or `B` being a `StridedVecOrMat`?

**Comment by mcabbott** [comment]
Sparse matrices appear to be handled correctly, in that `similar(sm)` gives the same type as the naiive `A*y .+ z` version:
```julia
m = rand(3,3); sm = sparse(m)
v = rand(3); sv = sparse(v);

sm * sm .+ 1 isa SparseMatrixCSC
sm * sm .+ v isa SparseMatrixCSC
sm * sm .+ m isa SparseMatrixCSC

sm * sm + 1 # error
sm * sm + v # error
sm * sm + m isa Matrix
```
The snippet above should handle dense triangular cases. Acting with `similar` on `Bidiagonal` or `SymTridiagonal` gives a sparse matrix, matching what `*` does. 

Still hoping a little bit that it won't be necessary to go all the way down to StridedArray, but we may end up there.

mateuszbaran:
I don't quite understand. Right now in `Base.muladd(A::AbstractMatrix{TA}, B::AbstractMatrix{TB}, z) where {TA, TB}` you allocate the result based on the type of `A`. But if either `B` or `z` is dense, the result should also be dense. How are you going to achieve that?

mcabbott:
Sure, matching the existing choices for every possible trio of types probably can't be done with some simple rule. Besides the `sparse .+ dense` examples, I don't know why these are the way they are:
```julia
SymTridiagonal(m+m') * m isa Matrix
m * SymTridiagonal(m+m') isa SparseMatrixCSC
```
I hope nobody's relying on that. But more importantly, someone may be relying on this, which ought not to go via `mul!`:
```julia
using StaticArrays
muladd(SA[1 2; 3 4], SA[5,6], SA[7,8]) 
```
Which argues for making it much more opt-in.

dkarrasch:
> I don't know why these are the way they are:

They are the way they are because, as I said, we call `similar` on the right factor, not on the left, see #36551.

mcabbott:
Thanks I hadn't seen that discussion. This was my failed attempt at an exotic case...

mcabbott:
Not quite done but this is a branch:

https://github.com/JuliaLang/julia/compare/master...mcabbott:mulladd2

This supports triangular & diagonal matrices. It also supports `UniformScaling` by itself, mixed with `Diagonal`, and perhaps weirdly as the 3rd argument with two matrices.

mcabbott:
Now at #38250.

PR #37069, opened 2020-08-17
Add `get-task-allow` entitlement to allow debuggers to attach to codesigned Julia

staticfloat:
Without this entitlement, users that wish to attach a debugger to the
codesign Julia executable they receive from `julialang.org` must strip
the codesigning signature from the Julia executable (e.g. via `codesign
--remove-signature Julia-1.5.app/Contents/Resources/bin/julia`).  This
has its disadvantages, of course, so much better to simply declare to
the OS that it's alright for other processes to attach to this process.

### DIFF
--- a/contrib/mac/app/Entitlements.plist
+++ b/contrib/mac/app/Entitlements.plist
@@ -4,6 +4,8 @@
 <dict>
 	<key>com.apple.security.automation.apple-events</key>
 	<true/>
+	<key>com.apple.security.cs.get-task-allow</key>
+	<true/>
 	<key>com.apple.security.cs.allow-dyld-environment-variables</key>
 	<true/>
 	<key>com.apple.security.cs.allow-jit</key>


contrib/mac/app/Entitlements.plist
```diff
@@ -4,6 +4,8 @@
 <dict>
 	<key>com.apple.security.automation.apple-events</key>
 	<true/>
+	<key>com.apple.security.cs.get-task-allow</key>
```
notinaboat, line 7: It should be `com.apple.security.get-task-allow` (delete the `.cs`).

KristofferC: Perhaps you can open a PR with the fix?

notinaboat:
I'm not sure I know how to test the fix. I really know nothing about code signing other than googling the error message led me to here and I noticed the miss-spelling.
I've done `--remove-signature` to work around the problem on my mac for now.

staticfloat:
Thanks for the heads up `notinaboat`!  This will be fixed with https://github.com/JuliaLang/julia/pull/41559, so you can see the change is quite simple.  :)

PR #37075, opened 2020-08-17
faster digits(::BigInt)

stevengj:
Since this [is a blocking issue for Julia 1.6](https://discourse.julialang.org/t/why-julia-is-so-much-slower-than-python-invitation-to-euler-project/45029/55?u=stevengj), this is about 50x faster for `digits(factorial(big(100)))`.

### DIFF
--- a/base/gmp.jl
+++ b/base/gmp.jl
@@ -694,6 +694,8 @@ function string(n::BigInt; base::Integer = 10, pad::Integer = 1)
     String(sv)
 end
 
+digits(n::BigInt; base::Integer = 10, pad::Integer = 1) = reverse!(map(x -> Int(x-0x30), codeunits(string(n; base, pad))))
+
 function ndigits0zpb(x::BigInt, b::Integer)
     b < 2 && throw(DomainError(b, "`b` cannot be less than 2."))
     x.size == 0 && return 0 # for consistency with other ndigits0z methods
--- a/base/gmp.jl
+++ b/base/gmp.jl
@@ -695,14 +695,11 @@ function string(n::BigInt; base::Integer = 10, pad::Integer = 1)
 end
 
 function digits(n::BigInt; base::Integer = 10, pad::Integer = 1)
-  if base == 10
-    dig = map(x -> Int(x-0x30), codeunits(string(n; base, pad)))
-  elseif base ≤ 62
-    dig = map(x -> Int(x > 0x39 ? x-0x41 : x-0x30), codeunits(string(n; base, pad)))
+  if base ≤ 62
+    return reverse!(map(x -> Int(x > 0x39 ? x-0x41 : x-0x30), codeunits(string(n; base, pad))))
   else
     return invoke(digits, (Integer,), n; base, pad) # slow generic method
   end
-  return reverse!(dig)
 end
 
 function ndigits0zpb(x::BigInt, b::Integer)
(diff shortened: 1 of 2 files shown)


base/gmp.jl
```diff
@@ -694,6 +694,14 @@ function string(n::BigInt; base::Integer = 10, pad::Integer = 1)
     String(sv)
 end
 
+function digits(n::BigInt; base::Integer = 10, pad::Integer = 1)
+  if base ≤ 62
```
StefanKarpinski, line 698: Should this also check that `2 ≤ base`?

stevengj: Done.

StefanKarpinski (commented):
Should use 4-space indentation.

base/gmp.jl
rfourquet: Might it work for `-62 <= base <= -2` ? (I'm asking just for curiosity, implementing fast `digits` for negative bases isn't blocking a release).

rfourquet:
Could this be implemented as a method of `digits!` ? (so both `digits!` and `digits` benefit from the improvement).

stevengj:
Oh, I didn't even realize that we had a `digits!` function, will do.

stevengj:
Switched to `digits!`.  This slows down `digits` by about 25% for base 10, possibly because it ends up calling `ndigits` twice, but since this function is not likely to be performance critical in real applications I doubt it is worth the code duplication to add a specialized `digits` as well as `digits!`.

StefanKarpinski:
I'm reluctant to get even more into the weeds here, but this is a release blocker, so perhaps it's warranted. The double allocation feels unfortunate, but of course, as long as the output here is `Vector{Int}` while `MPZ.get_str!` can only write into a `Vector{UInt8}` there's nothing we can do about that. If, however, the contract for `digits` was changed so that it could return `Vector{Int8}`, then we could call `MPZ.get_str! `the way that `string` does:

https://github.com/JuliaLang/julia/blob/613af3c04f61961449901112d50ffdaf34d2d069/base/gmp.jl#L682-L695

That would also avoid the multiple calls to `ndigits`. The approach would be to print into a `Vector{Int8}` and then adjust the values from ASCII digits down to Int8 values. Not sure it's worth it, however.

stevengj:
Well, `digits(Int8,…)` returns a `Vector{Int8}`… but as I commented above it mainly makes sense to optimize `digits!` rather than `digits`.

Of course, even with a wider integer type you could `reinterpret` into an `Int8` array and then shuffle things around in-place to avoid the extra allocation.

stevengj:
I played around with a bit, and it seems like getting a performance boost from avoiding the extra `String` allocation is more trouble than it's worth…

StefanKarpinski (approved):

test/gmp.jl
```diff
@@ -293,6 +293,16 @@ let s, n = bigfib(1000001)
     @test startswith(s, "316047687386689")
 end
 
+@testset "digits" begin
+    n = Int64(2080310129088201558)
+    N = big(n)
+    for base in (2,7,10,11,16,30,50,62,64,100), pad in (0,1,10,100)c
```
JeffBezanson, line 299:
```suggestion
    for base in (2,7,10,11,16,30,50,62,64,100), pad in (0,1,10,100)
```

stevengj: fixed

PR #37078, opened 2020-08-17
Improve inference for PhiNode.edges

timholy:
We have
```julia
julia> fieldtype(Core.PhiNode, :edges)
Vector{Any} = Array{Any,1}
```
despite the fact that all entries are `Int`. Even the [docs](https://docs.julialang.org/en/latest/devdocs/ssair/#Phi-nodes-and-Pi-nodes-1) say this field is a `Vector{Int}`. I spent a little bit of time trying to change that at the level of the C code but kept running into serialization-related problems. So this circumvents the problem by type-assertion. That said, I'd be happy with having someone close this PR and change the field type to `Vector{Int}` (or perhaps `Vector{Int32}` since there's already a corresponding C type declared).


JeffBezanson:
I'll take a look. We probably just need to add a compatibility shim to Serialization.

JeffBezanson:
Replaced by #37108

PR #37079, opened 2020-08-17
Solve inference problems in IRShow

timholy:
While not performance-sensitive, inference problems here make this code vulnerable to invalidation. (IRShow is in Base so is not subject to the protections of Core.Compiler.)

### DIFF
--- a/base/compiler/ssair/show.jl
+++ b/base/compiler/ssair/show.jl
@@ -38,9 +38,10 @@ function print_stmt(io::IO, idx::Int, @nospecialize(stmt), used::BitSet, maxleng
         print(io, stmt.typ)
         print(io, ")")
     elseif isexpr(stmt, :invoke)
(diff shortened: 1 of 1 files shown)


base/compiler/ssair/show.jl
```diff
@@ -432,28 +435,30 @@ function DILineInfoPrinter(linetable::Vector, showtypes::Bool=false)
...
-                    print(io, " within `", method_name(frame), "'")
-                    if collapse
-                        method = method_name(frame)
-                        while nctx < nframes
-                            frame = DI[nframes - nctx]
-                            method_name(frame) === method || break
-                            nctx += 1
-                            push!(context, frame)
-                            print(io, " @ ", frame.file, ":", frame.line)
+                let started=started
```
JeffBezanson, line 438: With the type declaration added to `started`, this should not be necessary.

PR #37080, opened 2020-08-17
Improve shape inference in the `cat` pipeline 

timholy:
`cat` is frequently called with poor inference, since one only has to concatenate a couple of different container types before inference punts on the result type. While this does not make the return type in mixed container types fully inferrable, it does improve the analysis of the shape.

Because it doesn't fully fix the inference problems, it's not obvious this is a testable improvement except at the aggregate level of backedges. Given that the issues are not fully fixed, the `T::Type` to `::Type{T}` changes don't really seem to be necessary, and I could back those out if desired.

### DIFF
--- a/base/abstractarray.jl
+++ b/base/abstractarray.jl
@@ -1437,15 +1437,15 @@ vcat(V::AbstractVector{T}...) where {T} = typed_vcat(T, V...)
 AbstractVecOrTuple{T} = Union{AbstractVector{<:T}, Tuple{Vararg{T}}}
 
 function _typed_vcat(::Type{T}, V::AbstractVecOrTuple{AbstractVector}) where T
-    n::Int = 0
+    n = 0
     for Vk in V
-        n += length(Vk)
+        n += Int(length(Vk))::Int
     end
     a = similar(V[1], T, n)
     pos = 1
-    for k=1:length(V)
+    for k=1:Int(length(V))::Int
         Vk = V[k]
-        p1 = pos+length(Vk)-1
+        p1 = pos + Int(length(Vk))::Int - 1
         a[pos:p1] = Vk
         pos = p1+1
     end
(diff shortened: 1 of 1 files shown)


base/abstractarray.jl
```diff
@@ -1436,15 +1436,15 @@ vcat(V::AbstractVector{T}...) where {T} = typed_vcat(T, V...)
 AbstractVecOrTuple{T} = Union{AbstractVector{<:T}, Tuple{Vararg{T}}}
 
 function _typed_vcat(::Type{T}, V::AbstractVecOrTuple{AbstractVector}) where T
-    n::Int = 0
+    n = 0
     for Vk in V
-        n += length(Vk)
+        n += length(Vk)::Union{Int,UInt}
```
JeffBezanson, line 1441: I'm not sure this is safe to assume.

timholy: I'll change it to `Integer`. That will leave some `promote_rule` invalidations but it maybe be better than nothing.

timholy: Or do you think we can safely convert-and-assert? Are you wondering about supporting arrays with more elements than `typemax(Int)`?

base/abstractarray.jl
```diff
@@ -1584,17 +1584,18 @@ end
 _cat(dims, X...) = cat_t(promote_eltypeof(X...), X...; dims=dims)
 
 @inline cat_t(::Type{T}, X...; dims) where {T} = _cat_t(dims, T, X...)
-@inline function _cat_t(dims, T::Type, X...)
+@inline function _cat_t(dims, ::Type{T}, X...) where {T}
     catdims = dims2cat(dims)
-    shape = cat_shape(catdims, map(cat_size, X))
+    shape = cat_shape(catdims, map(cat_size, X)::Tuple{Vararg{Union{Int,Dims}}})::Dims
```
pabloferz, line 1589: Looking at `cat_size` it seems to me that it would be just `::Union{Int,Dims}`, but maybe I'm missing something.

timholy: It may not be flattened (yet).

timholy: Oh, I get your point. Yes, that's essentially what this type-assert says: each `cat_size(X[i])::Union{Int,Dims}`.

pabloferz: Got it, I somehow missed on the `map` part. LGTM

KristofferC (approved):

PR #26930, opened 2018-04-29
Support context module in replcompletions

jamii:
RE: https://discourse.julialang.org/t/networked-repl/2085/15?u=jamii

@pfitzseb as promised :)

This just adds the context module. 

For structured completions, I could:

* Change the return type of completions and fix the usage in REPL. Then IJulia and co have to change their usage code.
* Make a new `structured_completions` function that does all the work and `completions` calls that and converts the results to strings. Avoids breaking downstream stuff, but leaves a gross api wart.

Which is preferable?

PR #37081, opened 2020-08-17
Improve inference broadly throughout REPL

timholy:
This attempts to improve inference in REPL. There are several pieces:

- type-stable interfaces have beend added for all abstract types ~`REPLCompletions.Completion` and  `TerminalMenus.AbstractMenu`~
- by defining the `Options` struct prior to creating the `LineEdit` module, we gain the ability to improve many inference results
- several methods assume the presence of fields that are not guaranteed for all subtypes. Restrict to subtypes that have the requisite fields.
- because REPL is compiled with low optimization settings, to prevent loss of type information it's helpful to declare more types for input arguments on internal methods.
- in a couple of places, avoiding unnecessary Unions makes later steps easier
- many implementations were improved. In most or perhaps all places I removed return-type annotations (left from previous attempts to enforce a minimal level of inference accuracy) since they are no longer necessary.

~This does a lot to make REPL more inferrable and less vulnerable to invalidation, but if you check out the precompile statements that get issued during bootstrap, REPL is still the standout "bad actor" with regards to non-inferrable code. (Especially after some of my other PRs from today.)~

### DIFF
--- a/base/io.jl
+++ b/base/io.jl
@@ -358,7 +358,7 @@ function pipe_reader end
 function pipe_writer end
 
 write(io::AbstractPipe, byte::UInt8) = write(pipe_writer(io)::IO, byte)
-unsafe_write(io::AbstractPipe, p::Ptr{UInt8}, nb::UInt) = unsafe_write(pipe_writer(io)::IO, p, nb)
+unsafe_write(io::AbstractPipe, p::Ptr{UInt8}, nb::UInt) = unsafe_write(pipe_writer(io)::IO, p, nb)::Union{Int,UInt}
 buffer_writes(io::AbstractPipe, args...) = buffer_writes(pipe_writer(io)::IO, args...)
 flush(io::AbstractPipe) = flush(pipe_writer(io)::IO)
 
--- a/base/iobuffer.jl
+++ b/base/iobuffer.jl
@@ -418,7 +418,7 @@ end
 function unsafe_write(to::GenericIOBuffer, p::Ptr{UInt8}, nb::UInt)
     ensureroom(to, nb)
     ptr = (to.append ? to.size+1 : to.ptr)
-    written = Int(min(nb, length(to.data) - ptr + 1))
+    written = Int(min(nb, Int(length(to.data))::Int - ptr + 1))
     towrite = written
     d = to.data
     while towrite > 0
--- a/base/strings/io.jl
+++ b/base/strings/io.jl
@@ -178,12 +178,12 @@ string(a::Symbol) = String(a)
 # note: print uses an encoding determined by `io` (defaults to UTF-8), whereas
 #       write uses an encoding determined by `s` (UTF-8 for `String`)
 print(io::IO, s::AbstractString) = for c in s; print(io, c); end
-write(io::IO, s::AbstractString) = (len = 0; for c in s; len += write(io, c); end; len)
+write(io::IO, s::AbstractString) = (len = 0; for c in s; len += Int(write(io, c))::Int; end; len)
 show(io::IO, s::AbstractString) = print_quoted(io, s)
 
 # optimized methods to avoid iterating over chars
 write(io::IO, s::Union{String,SubString{String}}) =
-    GC.@preserve s unsafe_write(io, pointer(s), reinterpret(UInt, sizeof(s)))
+    GC.@preserve s Int(unsafe_write(io, pointer(s), reinterpret(UInt, sizeof(s))))::Int
 print(io::IO, s::Union{String,SubString{String}}) = (write(io, s); nothing)
 
 ## printing literal quoted string data ##
--- a/stdlib/REPL/src/LineEdit.jl
+++ b/stdlib/REPL/src/LineEdit.jl
@@ -1407,10 +1407,10 @@ struct KeyAlias
     KeyAlias(seq) = new(normalize_key(seq))
 end
 
-function match_input(f::Function, s, term, cs, keymap)
+function match_input(f::Function, s::Union{Nothing,MIState}, term, cs::Vector{Char}, keymap)
     update_key_repeats(s, cs)
     c = String(cs)
-    return function (s, p)
+    return function (s, p)  # s::Union{Nothing,MIState}; p can be (at least) a LineEditREPL, PrefixSearchState, Nothing
         r = Base.invokelatest(f, s, p, c)
         if isa(r, Symbol)
             return r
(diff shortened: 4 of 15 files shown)


timholy:
OK, this grew into a much bigger deal. This should fix inference throughout essentially all of REPL.

base/io.jl
```diff
@@ -358,7 +358,7 @@ function pipe_reader end
 function pipe_writer end
 
 write(io::AbstractPipe, byte::UInt8) = write(pipe_writer(io)::IO, byte)
-unsafe_write(io::AbstractPipe, p::Ptr{UInt8}, nb::UInt) = unsafe_write(pipe_writer(io)::IO, p, nb)
+unsafe_write(io::AbstractPipe, p::Ptr{UInt8}, nb::UInt) = unsafe_write(pipe_writer(io)::IO, p, nb)::Union{Int,UInt}
```
timholy, line 361: Methods for [`Union{Core.CoreSTDERR,Core.CoreSTDOUT}`](https://github.com/JuliaLang/julia/blob/72ffbf1758946b7b1a3f8c666e27c8d89dbd406f/base/coreio.jl#L27) and [`Base.FileSystem.File`](https://github.com/JuliaLang/julia/blob/72ffbf1758946b7b1a3f8c666e27c8d89dbd406f/base/filesystem.jl#L128-L134) return `UInt`, the rest (I think) return `Int`. Should we make them all one or the other?

timholy: Not to derail the main point too much, but for overall context:
```julia
julia> for (m, rt) in zip(methods(write, (IO, Any)), Base.return_types(write, (IO, Any)))
           println(m => rt)
       end
write(io::Union{Core.CoreSTDERR, Core.CoreSTDOUT}, x::UInt8) in Base at coreio.jl:26 => Int64
write(s::IO, x::Union{Float16, Float32, Float64, Int128, Int16, Int32, Int64, UInt128, UInt16, UInt32, UInt64}) in Base at io.jl:649 => Any
write(io::IO, s::Union{SubString{String}, String}) in Base at strings/io.jl:185 => Int64
write(io::IO, s::Base.CodeUnits) in Base at strings/basic.jl:744 => Int64
write(to::IO, p::Ptr) in Base at io.jl:654 => Any
write(io::IO, x::Enum{T}) where T<:Integer in Base.Enums at Enums.jl:21 => Any
write(s::IO, a::SubArray{T,N,var"#s68",I,L} where L where I where var"#s68"<:Array) where {T, N} in Base at io.jl:675 => Any
write(s::IO, z::Rational) in Base at rational.jl:94 => Int64
write(io::IO, s::AbstractString) in Base at strings/io.jl:181 => Int64
write(s::IO, a::Array) in Base at io.jl:667 => Any
write(s::IO, x::Ref{T}) where T in Base at io.jl:647 => Any
write(s::IO, z::Complex) in Base at complex.jl:220 => Int64
write(s::IO, B::BitArray) in Base at bitarray.jl:1823 => Any
write(s::IO, A::AbstractArray) in Base at io.jl:656 => Any
write(s::IO, x::Bool) in Base at io.jl:653 => Any
write(io::IO, s::Symbol) in Base at io.jl:708 => Any
write(io::IO, c::Char) in Base at io.jl:696 => Int64
write(io::IO, cred::LibGit2.GitCredential) in LibGit2 at /home/tim/src/julia-master/usr/share/julia/stdlib/v1.6/LibGit2/src/gitcredential.jl:94 => Nothing
write(io::IO, s::Base.SecretBuffer) in Base at secretbuffer.jl:126 => Any
write(s::IO, x::Int8) in Base at io.jl:648 => Any
write(io::Base.AbstractPipe, byte::UInt8) in Base at io.jl:360 => Any
write(to::Base.GenericIOBuffer, from::Base.GenericIOBuffer) in Base at iobuffer.jl:408 => Int64
write(to::IO, from::IO) in Base at io.jl:713 => Any
write(to::Base.GenericIOBuffer, a::UInt8) in Base at iobuffer.jl:445 => Int64
write(pipe::Base64.Base64EncodePipe, x::UInt8) in Base64 at /home/tim/src/julia-master/usr/share/julia/stdlib/v1.6/Base64/src/encode.jl:98 => Int64
write(io::Base.SecretBuffer, b::UInt8) in Base at secretbuffer.jl:112 => Int64
write(s::Base.BufferStream, b::UInt8) in Base at stream.jl:1351 => Int64
write(s::Base.LibuvStream, b::UInt8) in Base at stream.jl:1091 => Int64
write(f::Base.Filesystem.File, c::UInt8) in Base.Filesystem at filesystem.jl:136 => UInt64
write(::Base.DevNull, ::UInt8) in Base at coreio.jl:16 => Int64
write(s::IOStream, b::UInt8) in Base at iostream.jl:366 => Int64
write(s::IO, x::UInt8) in Base at io.jl:224 => Union{}
write(io::IO, x) in Base at io.jl:635 => Union{}
```

timholy:
I expect this version should finally pass tests, as on the previous run only 32-bit was failing. Here, for completion methods I've changed `typeof(ret[2])` from `UnitRange{Int64}` to `UnitRange{Int}`. That choice in #26930 did not receive any discussion that I can see, so I'm going with the view that it wasn't particularly deliberate. Really making it work properly with `Int64` (which we do for some file-position methods even on 32-bit) would require more extensive changes, and I doubt it's relevant for REPL completions.

timholy:
I'm aware that this is a monster to review and that there aren't that many people who work on the REPL, so if necessary I'm happy to take responsibility for problems this causes.

KristofferC (approved):
I am not super familiar with the REPL code but looks good to me from what I can see.

aviatesk:
As for completions part: the changes looks good (I gave it a quick try and even made [equivalent changes for FuzzyCompletions.jl](https://github.com/JunoLab/FuzzyCompletions.jl/pull/4), and it seems to work great)

timholy:
@ChrisRackauckas it's really fun to be able to load DifferentialEquations and have your next tab-complete not have a 0.25s latency.

ChrisRackauckas:
you're the hero we needed.

timholy:
Also fun: Julia 1.5:
```julia
julia> @time using DifferentialEquations
 22.845265 seconds (61.37 M allocations: 3.389 GiB, 4.10% gc time)
```
Today's master branch:
```julia
julia> @time using DifferentialEquations
  6.944444 seconds (21.22 M allocations: 1.442 GiB, 5.81% gc time)
```

PR #37082, opened 2020-08-17
Ensure that BitArrays are inferrable for arbitrary `Integer`

timholy:
BitArray operations are frequently performed with partial inference information (particularly from Pkg but others as well). Currently the internal operations are a source of invalidation for common arithmetic and bitwise operators. There seems to be no need to live with this, as all the inference problems are fixed by converting `Integer`s to `Int`s. This is done elsewhere already in BitArray code, so it only seems to be continuing a process that is already underway.

Because of the multiplicity of methods `BitArray{N}(a::Any)` is not fully inferrable but there are at least limited victories, and the real benefits are in the elimination of instabilities in the internal operations.

### DIFF
--- a/base/bitarray.jl
+++ b/base/bitarray.jl
@@ -458,7 +458,9 @@ function unsafe_copyto!(dest::BitArray, doffs::Integer, src::Union{BitArray,Arra
     return dest
 end
 
-function copyto!(dest::BitArray, doffs::Integer, src::Array, soffs::Integer, n::Integer)
+copyto!(dest::BitArray, doffs::Integer, src::Array, soffs::Integer, n::Integer) =
+    _copyto_int!(dest, Int(doffs), src, Int(soffs), Int(n))
+function _copyto_int!(dest::BitArray, doffs::Int, src::Array, soffs::Int, n::Int)
     n == 0 && return dest
     soffs < 1 && throw(BoundsError(src, soffs))
     doffs < 1 && throw(BoundsError(dest, doffs))
@@ -803,8 +805,8 @@ function sizehint!(B::BitVector, sz::Integer)
     return B
 end
 
-function resize!(B::BitVector, n::Integer)
-    n = Int(n)
+resize!(B::BitVector, n::Integer) = _resize_int!(B, Int(n))
+function _resize_int!(B::BitVector, n::Int)
     n0 = length(B)
     n == n0 && return B
     n >= 0 || throw(BoundsError(B, n))
@@ -880,7 +882,8 @@ function popfirst!(B::BitVector)
     return item
 end
 
-function insert!(B::BitVector, i::Integer, item)
+insert!(B::BitVector, i::Integer, item) = _insert_int!(B, Int(i), item)
+function _insert_int!(B::BitVector, i::Int, item)
     i = Int(i)
     n = length(B)
     1 <= i <= n+1 || throw(BoundsError(B, i))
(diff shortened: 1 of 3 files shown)


base/bitarray.jl
```diff
@@ -94,7 +94,7 @@ julia> BitVector(nt)
 ```
 """
 function BitVector(nt::Tuple{Vararg{Bool}})
-    bv = BitVector(undef, length(nt))
+    bv = BitVector(undef, length(nt)::Int)
```
martinholters, line 97: Why is this one needed? Are there circumstances under which the `length` call here cannot be inferred as `::Int`? Then maybe that should be looked at.

Certainly we get
```julia
julia> code_typed(length, Tuple{Tuple{Vararg{Bool}}})
1-element Vector{Any}:
 CodeInfo(
1 ─ %1 = Base.nfields(t)::Int64
└──      return %1
) => Int64
```
So it can't be too bad.

base/bitarray.jl
```diff
@@ -530,14 +530,15 @@ function BitArray{N}(A::Array{Bool,N}) where N
...
+    return B::BitArray{N}
 end
 
 reinterpret(::Type{Bool}, B::BitArray, dims::NTuple{N,Int}) where {N} = reinterpret(B, dims)
 reinterpret(B::BitArray, dims::NTuple{N,Int}) where {N} = reshape(B, dims)
 
 if nameof(@__MODULE__) === :Base  # avoid method overwrite
-(::Type{T})(x::T) where {T<:BitArray} = copy(x)
+(::Type{T})(x::T) where {T<:BitArray} = copy(x)::T
+BitArray{N}(x::BitArray) where {N} = copy(x)::BitArray{N}
```
martinholters, line 541: Looks like this method will always fail with a `TypeError`? (If `x::BitArray{N}`, the method above is hit, otherwise, the type assertion on `copy(x)` will fail.) So maybe just throw a descriptive error explicitly?

base/bitarray.jl
```diff
@@ -500,7 +500,7 @@ end
 
 BitArray(A::AbstractArray{<:Any,N}) where {N} = BitArray{N}(A)
 function BitArray{N}(A::AbstractArray{T,N}) where N where T
-    B = BitArray(undef, size(A))
+    B = BitArray(undef, size(A)::Dims{N})
```
JeffBezanson, line 503: This probably needs to be a convert-plus-assert rather than just an assert.

base/bitarray.jl
timholy: Good point, I think we can just delete that method, it shouldn't ever be called.

base/bitarray.jl
```diff
@@ -797,6 +797,7 @@ function sizehint!(B::BitVector, sz::Integer)
 end
 
 function resize!(B::BitVector, n::Integer)
+    n = Int(n)
```
martinholters, line 800: I'm pretty sure in other places we do this kind of thing as
```julia
resize!(B::BitVector, n::Integer) = resize!(B,  Int(n))
function resize!(B::BitVector, n::Int)
   ...
```
The rationale being that this way, only the trivial `resize!(::BitVector, ::Integer)` needs to be specialized on different types (unless it is inlined away anyway), while only one specialization is needed for the actual implementation. Is that not a concern here? Or has the general landscape shifted such that that paradigm is obsolete?

timholy: Tough call. There are a few advantages in keeping the number of methods small, as Julia's compiler does "world-splitting" (like union-splitting except making decisions based on existing methods rather than an explicit `Union`) when there are 3 or fewer methods.

We could add a `@nospecialize n` but I'm not sure I like that either.

martinholters: If the number of methods is an issue then
```julia
resize!(B::BitVector, n::Integer) = _resize!(B,  Int(n))
function _resize!(B::BitVector, n::Int)
   ...
```
would be the obvious variation to keep that in check.

base/bitarray.jl
```diff
@@ -933,6 +947,7 @@ function _deleteat!(B::BitVector, i::Integer)
 end
 
 function deleteat!(B::BitVector, i::Integer)
+    i = Int(i)
```
KristofferC, line 950: Could do the "split out the `Int` conversion part"-dance here as well.

KristofferC (approved):
There are a few more functions that could split out the `Int` conversion but maybe they are considered small so it doesn't really matter.

base/bitarray.jl
timholy: My feeling is the bit of bounds-checking isn't worth worrying about.

PR #37084, opened 2020-08-17
Improve partial inference of key container operations

timholy:
`resize!` and `sizehint!` are big targets for invalidation, and quite a lot of code was calling these methods with poor inference about the size. Since the size is just `Int`, asserting that eliminates numerous vulnerabilities. Likewise, bounds-checking always works with `Bool`. Finally, add a short-circuit for `eltype(::Tuple{T})`.

### DIFF
--- a/base/abstractset.jl
+++ b/base/abstractset.jl
@@ -86,7 +86,7 @@ max_values(::Type{Bool}) = 2
 max_values(::Type{Nothing}) = 1
 
 function union!(s::AbstractSet{T}, itr) where T
-    haslength(itr) && sizehint!(s, length(s) + length(itr))
+    haslength(itr) && sizehint!(s, length(s) + Int(length(itr))::Int)
     for x in itr
         push!(s, x)
         length(s) == max_values(T) && break
--- a/base/array.jl
+++ b/base/array.jl
@@ -984,7 +984,7 @@ push!(a::AbstractVector, iter...) = append!(a, iter)
 function _append!(a, ::Union{HasLength,HasShape}, iter)
     n = length(a)
     i = lastindex(a)
-    resize!(a, n+length(iter))
+    resize!(a, n+Int(length(iter))::Int)
     @inbounds for (i, item) in zip(i+1:lastindex(a), iter)
         a[i] = item
     end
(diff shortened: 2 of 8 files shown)


base/abstractset.jl
```diff
@@ -86,7 +86,7 @@ max_values(::Type{Bool}) = 2
 max_values(::Type{Nothing}) = 1
 
 function union!(s::AbstractSet{T}, itr) where T
-    haslength(itr) && sizehint!(s, length(s) + length(itr))
+    haslength(itr) && sizehint!(s, length(s) + length(itr)::Int)
```
JeffBezanson, line 89: Probably not safe.

base/iddict.jl
```diff
@@ -162,6 +162,7 @@ function get!(default::Callable, d::IdDict{K,V}, @nospecialize(key)) where {K, V
         if !isa(val, V)
             val = convert(V, val)
         end
+        val::V
```
JeffBezanson, line 165: Does it still work if you put this on the `convert(V, val)` call instead? That seems better.

base/dict.jl
```diff
@@ -108,7 +108,7 @@ end
 Dict{K,V}(p::Pair) where {K,V} = setindex!(Dict{K,V}(), p.second, p.first)
 function Dict{K,V}(ps::Pair...) where V where K
     h = Dict{K,V}()
-    sizehint!(h, length(ps))
+    sizehint!(h, length(ps)::Int)
```
JeffBezanson, line 111: This is surprising --- can we really not infer this?

base/array.jl
```diff
@@ -984,7 +984,7 @@ push!(a::AbstractVector, iter...) = append!(a, iter)
 function _append!(a, ::Union{HasLength,HasShape}, iter)
     n = length(a)
     i = lastindex(a)
-    resize!(a, n+length(iter))
+    resize!(a, n+length(iter)::Int)
```
JeffBezanson, line 987: Also this and the next one; arbitrary iterators might not have `Int` length.

base/dict.jl
timholy: I can't recall where this came up. I might have just been routinely annotating all `sizehint!` calls. I can't find any evidence that we need this.

base/array.jl
timholy: You don't mind converting to Int and then asserting?

timholy:
I threw in a random `show_unquoted` fix that I failed to notice earlier.

KristofferC (approved):
LGTM (I didn't understand the `base/tuple.jl` part though). Could maybe use comments there?

PR #36107, opened 2020-06-01
RFC: print all types using equivalent exported aliases

vtjnash:
Given triage's interest in https://github.com/JuliaLang/julia/pull/36032, this works to generalize those printing improvements to all types. While type-intersection failures can be rather unexpected (i.e. handling difficult cases, but not simplifications of them), it often does much better than master anyways:

```julia
julia> T = Union{Nothing, StridedVector{S}} where S
Union{Nothing, StridedArray{S,1}} where S

julia> T = Union{Nothing, StridedVecOrMat{S}} where S<:Integer
Union{Nothing, StridedMatrix{S}, StridedVector{S}} where S<:Integer

julia> T = Union{Nothing, StridedVecOrMat{S}, S} where S<:Integer
Union{Nothing, S, StridedMatrix{S}, StridedVector{S}} where S<:Integer

julia> StridedVecOrMat
[...]

PR #25826, opened 2018-01-31
don't print a space in Union printing

fredrikekre:
This:
```julia
julia> Union{Int,Missing}
Union{Missing,Int64}

julia> Vector{Union{Int,Missing}}(missing, 2)
2-element Array{Union{Missing,Int64},1}:
 missing
 missing
```
Instead of: 
```julia
julia> Union{Int,Missing}
Union{Missing, Int64}

julia> Vector{Union{Int,Missing}}(missing, 2)
2-element Array{Union{Missing, Int64},1}:
 missing
 missing
```
This is how `Tuple` is printed, thought it would be nice to be consistent:
```julia
julia> Tuple{Int,Missing}
Tuple{Int64,Missing}
```

PR #37085, opened 2020-08-17
add spaces between type parameters and Dict entries

JeffBezanson:
I expect this to be my most dramatic PR ever. I think this situation has gotten a bit silly:

```
julia> Tuple{typeof(isequal), Union{Nothing, Float64}, Union{Nothing, Float64}}
Tuple{typeof(isequal),Union{Nothing, Float64},Union{Nothing, Float64}}

julia> :(Tuple{Int,Int})
:(Tuple{Int, Int})

julia> Tuple{Int,Int}
Tuple{Int64,Int64}
```

Now that the `Vector` and `Matrix` (etc.) type printing change is done, many of the tests this would break are already fixed. I'll fix the remainder if there is agreement.


fredrikekre:
RIP every doctest in the ecosystem :smile: 

If people are running doctests on more than one Julia version, the following doctestfilters (https://juliadocs.github.io/Documenter.jl/stable/man/doctests/#Filtering-Doctests) can be applied to make it pass:
```
doctestfilters = [
    r"{([a-zA-Z0-9]+,\s?)+[a-zA-Z0-9]+}",
    r"(Array{[a-zA-Z0-9]+,\s?1}|Vector{[a-zA-Z0-9]+})",
    r"(Array{[a-zA-Z0-9]+,\s?2}|Matrix{[a-zA-Z0-9]+})",
]
```
which should deal with this and also the two most common failures from #36107 (`Array{X,1}` -> `Vector{X,1}` and `Array{X,2}` -> `Matrix{X}`).

fredrikekre:
Also interesting to note that in the codebase itself we seem to prefer no space:
```
$ grep --include \*.jl --include \*.md -roP '{([a-zA-Z0-9]+,\s)+[a-zA-Z0-9]+}' . | wc -l
2263

$ grep --include \*.jl --include \*.md -roP '{([a-zA-Z0-9]+,)+[a-zA-Z0-9]+}' . | wc -l
5101
```

KristofferC:
Crazy idea, but maybe Documenter should apply those filters by default..?

fredrikekre:
Since doctests are pretty sensitive to printing like this, new random numbers streams etc, I think most packages only build and doctests on a single Julia version though.

JeffBezanson:
> Edit: And same in packages on my disk, no space is more common:

So? Should we go the other way and remove all the spaces? Or do we currently have an ideal compromise, using spaces half the time? :smile:

fredrikekre:
> So?

I just wanted to point out that most Julia programmers prefer no space in type parameters. That is also the recommended style according to e.g. YASGuide (https://github.com/jrevels/YASGuide#linealignmentspacing-guidelines).


> Should we go the other way and remove all the spaces?

Yes: https://github.com/JuliaLang/julia/pull/25826 :smile:

JeffBezanson:
> Use a single space after commas and semicolons. The exception is within type parameter lists, where no spaces should be added after commas.

Seems odd to me. I would almost guess it says that just to follow how we were printing them. @jrevels is that the case?

jrevels:
> I would almost guess it says that just to follow how we were printing them. @jrevels is that the case?

Originally the YASGuide did not include the type parameter list exception in its "spaces-after-commas mandate", but we added it after it was brought up that most code in the wild does not include spaces after commas in type parameter lists (https://github.com/jrevels/YASGuide/issues/3). So this exception was introduced purely to follow existing "convention" (or at least our perception of it, which probably was indeed influenced by the existing printing style).

timholy:
Even if we like it better, I'm not sure that changing away from no spaces is worth it. In the JuliaImages organization we had a half-dozen PRs like [this one](https://github.com/JuliaImages/ImageTransformations.jl/pull/98); that was a change that seems unambiguously better, yet despite that one can't help feeling like it was sort of a waste of a couple of hours (across the multiple packages). It's not always completely trivial to write these in a way that works across multiple Julia versions with different printing conventions and yet doesn't abandon all sense that you're actually testing the output in some meaningful way. I'm not saying it's hard, but it is the kind of annoyance that we should think twice before inflicting on folks.

mateuszbaran:
If you are adding spaces after commas, maybe `show_typealias` also should have them? On the current master I have:
```julia
julia> SVector{2}
SVector{2,T} where T = SArray{Tuple{2}, T, 1, 2} where T
```
which is still inconsistent.

JeffBezanson:
Thanks for catching that, will fix.

PR #37086, opened 2020-08-17
Improve inference on some type-, Expr-, and error operations

timholy:
This aims to reduce our backedges to operations like `+(::Int, ::Any)`. One noteworthy change is the eltype of `CodeInfo.codelocs`.


base/error.jl
```diff
@@ -68,6 +68,8 @@ end
 # convert dual arrays (raw bt buffer, array of GC managed values) to a single
 # array of locations
 function _reformat_bt(bt, bt2)
+    bt::Vector
+    bt2::Vector
```
JeffBezanson, line 72: Any reason not to put `bt::Vector{Ptr{Cvoid}}, bt2::Vector{Any}` in the signature instead?

base/errorshow.jl
```diff
@@ -234,16 +234,16 @@ function showerror(io::IO, ex::MethodError)
     arg_types_param::SimpleVector = arg_types.parameters
     print(io, "MethodError: ")
     ft = typeof(f)
-    name = ft.name.mt.name
+    name = (ft.name.mt::Core.MethodTable).name
```
JeffBezanson, line 237: I think we could probably declare the type of the `mt` field.

src/jltypes.c
```diff
@@ -2269,7 +2269,7 @@ void jl_init_types(void) JL_GC_DISABLED
                             "pure"),
                         jl_svec(18,
                             jl_array_any_type,
-                            jl_any_type,
+                            jl_array_int32_type,
```
JeffBezanson, line 2272: There is a line in method.c that sets this to `jl_nothing`. Should change that to NULL so nobody gets confused. There is also an assertion `src.codelocs !== nothing` in show.jl that can be removed.

base/error.jl
timholy: Oops, leftovers from a Revise/Cthulhu experiment that I forgot to finalize.

timholy: Oh, wait, no, there was a reason: `Vector` isn't defined at this point in bootstrap yet.

base/error.jl
```diff
@@ -67,7 +67,7 @@ end
 
 # convert dual arrays (raw bt buffer, array of GC managed values) to a single
 # array of locations
-function _reformat_bt(bt, bt2)
+function _reformat_bt(bt::Array{<:Any,1}, bt2::Array{<:Any,1})
```
JeffBezanson, line 70: I'm pretty sure this is always called on `Array{Ptr{Cvoid},1}` and `Array{Any,1}` exactly.

JeffBezanson (approved):

PR #37087, opened 2020-08-17
Improve inference in Docs and Markdown

timholy:
Because Docs is set to `@nospecialize` for the entire module, declaring types on input arguments makes a large difference in what
methods get inferred. This heads off a number of completely impossible combinations.

It's worth noting that enough methods have `@nospecialize` arguments that it might be worth considering removing the module-global `@nospecialize`. I understand the convenience of the module-global version, but it made for some really bizarre headscratching for why we were getting a `signature!(::Any, ::Expr)` MethodInstance.

### DIFF
--- a/base/docs/Docs.jl
+++ b/base/docs/Docs.jl
@@ -83,7 +83,7 @@ function initmeta(m::Module)
     nothing
 end
 
-function signature!(tv, expr::Expr)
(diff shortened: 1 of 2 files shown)


KristofferC (approved):

KristofferC (approved):
But seems to be some new ambiguities this hits.

PR #37089, opened 2020-08-17
Improve inference for a few Distributed operations

timholy:
Distributed is still a challenge wrt inference, but this improves at least a few aspects. More comprehensive fixes would probably require a more serious look than I've taken.

### DIFF
--- a/base/asyncmap.jl
+++ b/base/asyncmap.jl
@@ -190,13 +190,13 @@ end
 
 function setup_chnl_and_tasks(exec_func, ntasks, batch_size=nothing)
     if isa(ntasks, Function)
-        nt = ntasks()
+        nt = ntasks()::Int
         # start at least one worker task.
         if nt == 0
             nt = 1
         end
     else
-        nt = ntasks
+        nt = ntasks::Int
     end
 
     # Use an unbuffered channel for communicating with the worker tasks. In the event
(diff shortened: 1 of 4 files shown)


base/asyncmap.jl
```diff
@@ -184,19 +184,19 @@ function maptwice(wrapped_f, chnl, worker_tasks, c...)
         return asyncrun.x
     else
         # second run, extract values from the Refs and return
-        return map(ref->ref.x, asyncrun)
+        return map(ref->ref.x, asyncrun)::Vector
```
JeffBezanson, line 187: Looks like this is not correct.

stdlib/Distributed/src/pmap.jl
```diff
@@ -213,15 +213,17 @@ function process_batch_errors!(p, f, results, on_error, retry_delays, retry_chec
...
-                    on_error = on_error, retry_delays = collect(retry_delays)[2:end], retry_check = retry_check)
-        elseif on_error !== nothing
-            error_processed = map(on_error, exceptions)
-        else
-            throw(CompositeException(exceptions))
+        let state=state
+            if (length(retry_delays)::Int > 0) &&
+                    (retry_check === nothing || all([retry_check(state,ex)[2] for ex in exceptions]))
+                # BatchProcessingError.data is a tuple of original args
+                error_processed = pmap(x->f(x...), p, [x.data for x in errors];
```
JeffBezanson, line 220: Needs to be declared outside the let.

PR #37091, opened 2020-08-17
Add an interface for abstract SHA types

timholy:
This improves inference greatly in circumstances of partial inference.

### DIFF
--- a/stdlib/SHA/src/types.jl
+++ b/stdlib/SHA/src/types.jl
@@ -99,7 +99,7 @@ function Base.getproperty(ctx::SHA3_CTX, fieldname::Symbol)
     elseif fieldname === :bc
         return getfield(ctx, :bc)::Vector{UInt64}
     else
-        error("SHA3_CTX has no field ", fieldname)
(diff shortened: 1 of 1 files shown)


stdlib/SHA/src/types.jl
```diff
@@ -74,6 +89,20 @@ mutable struct SHA3_512_CTX <: SHA3_CTX
...
+    if fieldname === :state
+        return getfield(ctx, :state)::Vector{UInt64}
+    elseif fieldname === :bytecount
+        return getfield(ctx, :bytecount)::UInt128
+    elseif fieldname === :buffer
+        return getfield(ctx, :buffer)::Vector{UInt8}
+    elseif fieldname === :bc
+        return getfield(ctx, :bc)::Vector{UInt64}
+    else
+        error("SHA3_CTX has no field ", fieldname)
```
KristofferC, line 102: If we want to exactly reproduce the normal error it would be

```suggestion
        error("type ", typeof(ctx), " has no field ", fieldname)
```

but perhaps it doesn't matter.

KristofferC (approved):

PR #37092, opened 2020-08-17
Various inference improvements

timholy:
A fairly random assortment of improvements. Perhaps most notable are the improvements to commands that affect code-loading, since invalidations there have a big effect on package load times.

### DIFF
--- a/base/loading.jl
+++ b/base/loading.jl
@@ -131,10 +131,11 @@ end
 const slug_chars = String(['A':'Z'; 'a':'z'; '0':'9'])
 
 function slug(x::UInt32, p::Int)
+    y::UInt32 = x
     sprint(sizehint=p) do io
         n = length(slug_chars)
         for i = 1:p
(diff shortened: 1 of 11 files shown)


base/loading.jl
```diff
@@ -130,7 +130,9 @@ end
 
 const slug_chars = String(['A':'Z'; 'a':'z'; '0':'9'])
 
-function slug(x::UInt32, p::Int)
+slug(x::UInt32, p::Int) = _slug(unsigned(Int)(x), p)
+function _slug(x::T, p::Int) where T<:Unsigned
+    x::T = x
```
JeffBezanson, line 135: I don't understand this one --- why isn't it better just to have `x::UInt32`?

timholy: I was letting it go to what happens "naturally" with `divrem`, but keeping it `UInt32` is surely better.

Issue #37020, opened 2020-08-12
error printing latency regression


JeffBezanson:
I noticed that printing errors in the repl seemed sluggish.
Master:
```
$ time usr/bin/julia -e '1+""'
ERROR: MethodError: no method matching +(::Int64, ::String)
...
ROOT                      :  4.67 %   166096280
GC                        :  1.66 %   58897114
LOWERING                  :  0.02 %   600542
PARSING                   :  0.01 %   217937
INFERENCE                 : 16.61 %   590820880
CODEGEN                   :  2.31 %   81987111
METHOD_LOOKUP_SLOW        :  0.03 %   1200141
METHOD_LOOKUP_FAST        :  2.97 %   105474981
LLVM_OPT                  : 58.10 %   2066395982
LLVM_MODULE_FINISH        :  0.84 %   29807302
LLVM_EMIT                 :  0.01 %   200406
METHOD_MATCH              :  5.46 %   194225400

real	0m2.427s
```
v1.5:
```
$ time usr/bin/julia -e '1+""'
ERROR: MethodError: no method matching +(::Int64, ::String)
...
ROOT                      :  5.31 %   87134552
GC                        :  1.96 %   32241752
LOWERING                  :  0.07 %   1134365
PARSING                   :  0.02 %   281402
INFERENCE                 : 27.24 %   447034252
CODEGEN                   :  1.70 %   27962001
METHOD_LOOKUP_SLOW        :  0.28 %   4537017
METHOD_LOOKUP_FAST        :  4.74 %   77857820
LLVM_OPT                  : 42.52 %   697836048
LLVM_MODULE_FINISH        :  0.86 %   14092778
LLVM_EMIT                 :  0.02 %   297361
METHOD_LOOKUP_COMPILE     :  0.07 %   1067969
METHOD_MATCH              :  3.40 %   55836517

real	0m1.143s
```
Seems to be mostly llvm time.

JeffBezanson:
Interestingly, seems to have been triggered by c974bc65f964cf838f1b6764dfc10e5f9545e1ed.

KristofferC:
Running with `--trace-compile=stderr` shows much more being compiled on master.

JeffBezanson:
It looks like it could involve printing to an `IOContext{TTY}` instead of to an `IOContext{IOBuffer}` (what `with_output_color` does). Is it possible the precompile script isn't catching that case?

KristofferC:
It seems the precompile script is compiling for:

[...]

PR #37094, opened 2020-08-17
add nospecialize to some error IO

KristofferC:
Takes ` julia -e '1+""'` from 2.3s to 1.1s.

Ref https://github.com/JuliaLang/julia/issues/37020.

As that issue states, there is also the potential of collecting many of these things into a module and apply module level `@nospecialize` and `@optlevel`. I could change to doing that, but it is a bit annoying because it changes the namespace and then a lot of things have to be `imported` / prefixed.

### DIFF
--- a/base/errorshow.jl
+++ b/base/errorshow.jl
@@ -12,7 +12,7 @@ julia> struct MyException <: Exception
            msg::AbstractString
        end
 
-julia> function Base.showerror(io::IO, err::MyException)
+julia> function Base.showerror(@nospecialize(io::IO), err::MyException)
            print(io, "MyException: ")
            print(io, err.msg)
        end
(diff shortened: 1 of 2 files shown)


base/errorshow.jl
```diff
@@ -187,7 +187,7 @@ function print_with_compare(io::IO, @nospecialize(a::DataType), @nospecialize(b:
             if i > length(b.parameters)
                 printstyled(io, a.parameters[i], color=color)
             else
-                print_with_compare(io::IO, a.parameters[i], b.parameters[i], color)
+                print_with_compare(@nospecialize(io::IO), a.parameters[i], b.parameters[i], color)
```
timholy, line 190: You don't need this on the call, right?

timholy (approved):
Looks good. This doesn't affect the performance of our StackOverflowError stacktrace compressor, right? That's one place where specialization may be helpful.

fredrikekre:
> Takes julia -e '1+""' from 2.3s to 1.1s.

Ah, the classic benchmark for comparing languages :trollface:

KristofferC:
> . This doesn't affect the performance of our StackOverflowError stacktrace compressor, right? That's one place where specialization may be helpful

It's only about specialization for the `io` argument so I wouldn't think so. Can try time some stuff.

JeffBezanson:
I guess there's nothing wrong with this, but it would be good to make the precompiled types match as well.

KristofferC:
> I guess there's nothing wrong with this, but it would be good to make the precompiled types match as well.

Well, it's a different `IO` type depending on whether we use the REPL or write directly to a terminal with `-e`. When we record the precompile statements we use the `REPL` one. So I don't really see what "making the precompiled types match" mean.

JeffBezanson:
Maybe we could rearrange the REPL to use stdout directly more often instead of writing to a TTYTerminal.

KristofferC:
Subsumed by https://github.com/JuliaLang/julia/pull/37127. I'm not sure we specialize this for a lot of IO types so perhaps this PR can just be closed then.

PR #37096, opened 2020-08-17
Update memory allocs for using views

JeffFessler:
Julia 1.5 wonderfully allocates views on the stack (as you know.)
This PR updates the memory values reported to match 1.5.
I did not update the times, because those are system dependent.
If I am making a PR for something that is auto-generated by make docs, then apologies in advance...


KristofferC:
Thanks!

PR #37101, opened 2020-08-18
make Profile more thread/signal-safe

### LINKED ISSUE #13294: Profiler can deadlock on OS X

### LINKED ISSUE #35117: make stacktrace lookup and unwinding more threadsafe

vtjnash:
Internally these codepaths contain an rwlock on every platform (or are specified that you should hold one). So this adds that lock, and additionally blocks signals on unix for extra protection against deadlock here.

Fixes #35117
Fixes #13294

### DIFF
--- a/base/process.jl
+++ b/base/process.jl
@@ -546,7 +546,7 @@ Returns successfully if the process has already exited, but throws an
 error if killing the process failed for other reasons (e.g. insufficient
 permissions).
 """
-function kill(p::Process, signum::Integer)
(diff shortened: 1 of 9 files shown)


vtjnash:
Investigated the macOS hang overnight and tracked down the version of macOS that introduced that bug—let's see how this does now.

PR #37104, opened 2020-08-18
Add QR multiplication by UniformScaling

oscardssmith:
before it had an `O(n^4)` fallback.


KristofferC:
Done really need to add this specialization for all different exotic array types.. It seems a bit excessive.

oscardssmith:
Would it be better to have a more general fallback that first converts to a matrix and then multiplies? For matrices without O(1) indexing, this often should be faster.

stevengj:
Shouldn’t the fallback first drop down to a scalar-matrix multiplication?

PR #37108, opened 2020-08-18
declare PhiNode.edges::Vector{Int32}

JeffBezanson:
This is pretty clearly the right type to use, but it does come with some annoyance of needing to convert to `Int` several places in the compiler. Seems worth it to me.

fixes #37078

### DIFF
--- a/base/boot.jl
+++ b/base/boot.jl
@@ -118,7 +118,7 @@
 #end
 
 #struct PhiNode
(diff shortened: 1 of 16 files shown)


timholy:
I was missing a fair bit on the C side, so thanks for doing this!

timholy (approved):

PR #37110, opened 2020-08-18
Use Documenter 0.25.2

mortenpi:
This bumps Documenter to 0.25.2.

- 0.25.2 fixes the handing of `#`s in doctest output, so this re-applies the documentation changes from #36673 by reverting #36696.


fredrikekre:
https://github.com/JuliaLang/julia/pull/37110/files#diff-619ea7b9ee7c2fa87a3720c1361ba46cR168 should also work as a doctest now then.

mortenpi:
Yep, it looks like it does. Enabled it now.

PR #37111, opened 2020-08-19
Update integers-and-floating-point-numbers.md

uniment:
Corrected rule for ambiguity resolution for parsing engineering notation and floating-point literals

### DIFF
--- a/doc/src/manual/integers-and-floating-point-numbers.md
+++ b/doc/src/manual/integers-and-floating-point-numbers.md
@@ -694,8 +694,8 @@ where syntactic conflicts arise:
 In all cases the ambiguity is resolved in favor of interpretation as numeric literals:
 
   * Expressions starting with `0x` are always hexadecimal literals.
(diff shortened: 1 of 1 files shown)


StefanKarpinski:
The context is that if something could be parsed as either a floating point literal or an integer juxtaposed with a variable, then these rules apply, which implies what you're trying to say here. This edit also isn't correct since `1e2`, for example, is in no interpretation an integer literal followed by `e` followed by another integer literal: it is either a single floating-point literal or the integer literal `1` followed by the variable `e2`. I don't think we want to get into the business of spelling out the entire format of these literals here since it's somewhat involved.

uniment:
Perhaps I'm using the wrong vocabulary, but this is what I'm trying to say:

```
e5=4
2e5 # returns 200000.0
e=4
2e # returns 8
```

Currently the wording implies that `2e` will be parsed as a floating-point literal, which is not the case.

Cheers!
Zeke

uniment:
By the way I'm really loving Julia; thanks for all your excellent work!

PR #37114, opened 2020-08-19
Improve optimizations on NULL pointer check

yuyichao:
Mark more functions and calls as non-null and use inbounds gep instead of integer arithmetics
so that LLVM can more easily infer the non-null-ness of pointers. (Partially revert 2bb430e383809dcf5f73605a65f408c7e5bd22d6)

Also add inline optimization for `ccall` on `jl_symbol_name` and makes
sure the return value is treated as non-null by LLVM.

### DIFF
--- a/src/codegen.cpp
+++ b/src/codegen.cpp
@@ -361,7 +361,7 @@ static AttributeList get_attrs_sext(LLVMContext &C)
 {
     return AttributeList::get(C,
                 AttributeSet(),
-                AttributeSet(),
+                Attributes(C, {Attribute::NonNull}),
                 {Attributes(C, {Attribute::SExt})});
 }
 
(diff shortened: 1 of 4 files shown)


src/codegen.cpp
```diff
@@ -614,7 +614,7 @@ static const auto jl_newbits_func = new JuliaFunction{
                 {T_prjlvalue, T_pint8}, false); },
     [](LLVMContext &C) { return AttributeList::get(C,
             AttributeSet(),
-            Attributes(C, {Attribute::NoAlias, Attribute::NonNull}),
+            Attributes(C, {Attribute::NonNull}),
```
vtjnash, line 617: backport?

yuyichao: Oh yeah, this line of change is in principle a bug fix..... (commit was from a week ago and I've forgot about that...............)

vtjnash: Yeah, I think I missed the second half of the langref documentation for this attribute. (for arguments, it means that it doesn't alias an mutable memory, but for return values, it also means the pointer itself is unique à la malloc)

KristofferC:
@yuyichao https://github.com/JuliaLang/julia/pull/37210/commits/f59a4fffd7b38feb56b68f687888b2d1e34a516b

PR #37119, opened 2020-08-19
Add Iterators.first and Iterators.last which don't throw for empty collections

non-Jedi:
EDIT: This PR now implements option 2 discussed at the bottom of this post rather than option 1 which it originally implemented.

On zulip, there was a thread about the fact that there's not a good
way of finding the first item matching a predicate in a non-indexable iterable without
manually writing out the `for`-loop
(<https://julialang.zulipchat.com/#narrow/stream/225540-gripes/topic/.60find.60.20in.20iterables>). One
possibility is using the existing `first` function composed with
`Iterators.filter`:

    first(Iterators.filter(>(5), 1:10))

The problem with this approach is that it requires wrapping in a `try`
block if the predicate not being satisfied is a
possibility (`Iterators.filter` returns an empty collection). There
should be a way of expressing this without using exception-handling as
control flow or writing the `for`-loop manually.

Ideally, I think first and last should just return
`Union{Some{T},Nothing}` always, but that would be a breaking
change. So I see two obvious alternatives:

1. This commit adds an optional predicate argument to `first` and `last`
   which is mapped over the collection and the first/last value
   satisfying predicate is returned wrapped in `Some`.
2. Add a `Iterators.first`/`Iterators.last` which are identical to the
   versions in Base except that when successful they return `Some{T}`
   and when the collection is empty return `Nothing`

### DIFF
--- a/base/abstractarray.jl
+++ b/base/abstractarray.jl
@@ -346,7 +346,7 @@ end
 """
     first(predicate, coll)
 
-Get the first element of `coll` satisfying `predicate` wrapped in `Some`.
+Get the first element of `coll` satisfying `predicate` wrapped in [`Some`](@ref).
 
 If no element of `coll` satisfies `predicate`, return `nothing`.
 
@@ -355,10 +355,10 @@ If no element of `coll` satisfies `predicate`, return `nothing`.
 julia> first(>(5), 1:10)
 Some(6)
 
-julia> first(isodd, 2:2:10) |> isnothing
+julia> isnothing(first(isodd, 2:2:10))
 true
 
-julia> first(iseven, [5; 3; 4; 2; 6; 8]) |> something
+julia> something(first(iseven, [5, 3, 4, 2, 6, 8]))
 4
 
 julia> something(first(>(10), 1:10), 0)
@@ -420,7 +420,7 @@ last(a) = a[end]
 """
     last(predicate, coll)
 
-Get the last element of `coll` satisfying `predicate` wrapped in `Some`.
+Get the last element of `coll` satisfying `predicate` wrapped in [`Some`](@ref).
 
 If no element of `coll` satisfies `predicate`, return `nothing`.
 
@@ -429,10 +429,10 @@ If no element of `coll` satisfies `predicate`, return `nothing`.
 julia> last(<(5), 1:10)
 Some(4)
 
-julia> last(isodd, 2:2:10) |> isnothing
+julia> isnothing(last(isodd, 2:2:10))
 true
 
-julia> last(iseven, [5; 3; 4; 2; 6; 9]) |> something
+julia> something(last(iseven, [5; 3; 4; 2; 6; 9]))
 6
 
 julia> something(last(>(10), 1:10), 0)
--- a/base/abstractarray.jl
+++ b/base/abstractarray.jl
@@ -369,6 +369,7 @@ function first(predicate, itr)
     for x in itr
         predicate(x) && return Some(x)
     end
+    return nothing
 end
 
 """
@@ -443,7 +444,7 @@ function last(predicate, itr)
     for x in itr
         out = ifelse(predicate(x), Some(x), out)
     end
-    out
+    return out
 end
 
 # faster version for arrays
@@ -451,6 +452,7 @@ function last(predicate, a::AbstractArray)
     @inbounds for i in reverse(eachindex(a))
         predicate(a[i]) && return Some(a[i])
     end
+    return nothing
 end
 
 """
--- a/base/abstractarray.jl
+++ b/base/abstractarray.jl
@@ -350,6 +350,9 @@ Get the first element of `coll` satisfying `predicate` wrapped in [`Some`](@ref)
 
 If no element of `coll` satisfies `predicate`, return `nothing`.
 
+!!! compat "Julia 1.6"
+    This method was added in Julia 1.6.
+
 # Examples
 ```jldoctest
 julia> first(>(5), 1:10)
@@ -424,6 +427,9 @@ Get the last element of `coll` satisfying `predicate` wrapped in [`Some`](@ref).
 
 If no element of `coll` satisfies `predicate`, return `nothing`.
 
+!!! compat "Julia 1.6"
+    This method was added in Julia 1.6.
+    
 # Examples
 ```jldoctest
 julia> last(<(5), 1:10)
--- a/base/abstractarray.jl
+++ b/base/abstractarray.jl
@@ -344,6 +344,34 @@ function first(itr)
 end
 
 """
+    first(predicate, coll)
+
+Get the first element of `coll` satisfying `predicate` wrapped in `Some`.
+
+If no element of `coll` satisfies `predicate`, return `nothing`.
+
+# Examples
+```jldoctest
+julia> first(>(5), 1:10)
+Some(6)
+
+julia> first(isodd, 2:2:10) |> isnothing
+true
+
+julia> first(iseven, [5; 3; 4; 2; 6; 8]) |> something
+4
+
+julia> something(first(>(10), 1:10), 0)
+0
+```
+"""
+function first(predicate, itr)
+    for x in itr
+        predicate(x) && return Some(x)
+    end
+end
+
+"""
     first(itr, n::Integer)
 
 Get the first `n` elements of the iterable collection `itr`, or fewer elements if `v` is not
@@ -387,6 +415,43 @@ julia> last([1; 2; 3; 4])
 ```
 """
 last(a) = a[end]
+
+"""
+    last(predicate, coll)
+
+Get the last element of `coll` satisfying `predicate` wrapped in `Some`.
+
+If no element of `coll` satisfies `predicate`, return `nothing`.
+
+# Examples
+```jldoctest
+julia> last(<(5), 1:10)
+Some(4)
+
+julia> last(isodd, 2:2:10) |> isnothing
+true
+
+julia> last(iseven, [5; 3; 4; 2; 6; 9]) |> something
+6
+
+julia> something(last(>(10), 1:10), 0)
+0
+```
+"""
+function last(predicate, itr)
+    out = nothing
+    for x in itr
+        out = ifelse(predicate(x), Some(x), out)
+    end
+    out
+end
+
+# faster version for arrays
+function last(predicate, a::AbstractArray)
+    @inbounds for i in reverse(eachindex(a))
+        predicate(a[i]) && return Some(a[i])
+    end
+end
 
 """
     last(itr, n::Integer)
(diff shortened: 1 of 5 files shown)


base/abstractarray.jl
```diff
@@ -343,6 +343,34 @@ function first(itr)
...
+4
+
+julia> something(first(>(10), 1:10), 0)
+0
+```
+"""
+function first(predicate, itr)
+    for x in itr
+        predicate(x) && return Some(x)
+    end
```
KristofferC, line 371:
```suggestion
    end
    return nothing
```

base/abstractarray.jl
```diff
@@ -388,6 +416,43 @@ julia> last([1; 2; 3; 4])
...
+julia> something(last(>(10), 1:10), 0)
+0
+```
+"""
+function last(predicate, itr)
+    out = nothing
+    for x in itr
+        out = ifelse(predicate(x), Some(x), out)
+    end
+    out
```
KristofferC, line 446:
```suggestion
    return out
```

base/abstractarray.jl
```diff
@@ -388,6 +416,43 @@ julia> last([1; 2; 3; 4])
...
+end
+
+# faster version for arrays
+function last(predicate, a::AbstractArray)
+    @inbounds for i in reverse(eachindex(a))
+        predicate(a[i]) && return Some(a[i])
+    end
```
KristofferC, line 453:
```suggestion
    end
    return nothing
```

base/abstractarray.jl
non-Jedi: Is there a style guide for Julia base somewhere? I didn't know it was preferred to always use `return`, and I don't see it listed in https://docs.julialang.org/en/v1/manual/style-guide/

ararslan: It's just a bit iffy to rely on the fact that `for` loops return `nothing`, so it's better to be explicit in the case that the loop completes.

base/abstractarray.jl
```diff
@@ -388,6 +417,44 @@ julia> last([1; 2; 3; 4])
 """
 last(a) = a[end]
 
+"""
+    last(predicate, coll)
+
+Get the last element of `coll` satisfying `predicate` wrapped in `Some`.
```
ararslan, line 423:
```suggestion
Get the last element of `coll` satisfying `predicate` wrapped in [`Some`](@ref).
```

base/abstractarray.jl
```diff
@@ -388,6 +417,44 @@ julia> last([1; 2; 3; 4])
...
+
+If no element of `coll` satisfies `predicate`, return `nothing`.
+
+# Examples
+```jldoctest
+julia> last(<(5), 1:10)
+Some(4)
+
+julia> last(isodd, 2:2:10) |> isnothing
```
ararslan, line 432:
```suggestion
julia> isnothing(last(isodd, 2:2:10))
```

base/abstractarray.jl
```diff
@@ -388,6 +417,44 @@ julia> last([1; 2; 3; 4])
...
+true
+
+julia> last(iseven, [5; 3; 4; 2; 6; 9]) |> something
```
ararslan, line 435:
```suggestion
julia> something(last(iseven, [5; 3; 4; 2; 6; 9]))
```

base/abstractarray.jl
```diff
@@ -343,6 +343,35 @@ function first(itr)
     x[1]
 end
 
+"""
+    first(predicate, coll)
+
+Get the first element of `coll` satisfying `predicate` wrapped in [`Some`](@ref).
+
+If no element of `coll` satisfies `predicate`, return `nothing`.
+
```
non-Jedi, line 352:
```suggestion

!!! compat "Julia 1.6"
    This method was added in Julia 1.6.

```

base/abstractarray.jl
```diff
@@ -388,6 +417,44 @@ julia> last([1; 2; 3; 4])
 """
 last(a) = a[end]
 
+"""
+    last(predicate, coll)
+
+Get the last element of `coll` satisfying `predicate` wrapped in [`Some`](@ref).
+
+If no element of `coll` satisfies `predicate`, return `nothing`.
+
```
non-Jedi, line 426:
```suggestion

!!! compat "Julia 1.6"
    This method was added in Julia 1.6.
    
```

non-Jedi (commented):
Should the docstrings instead be combined with the existing docstring for `first(coll)` as e.g.:

```julia
"""
    first([predicate=_ -> true,] coll)

...
"""
```

Also, I really appreciate the detailed code review. It would be great to get some feedback on the overall concept as well.

base/abstractarray.jl
```diff
@@ -343,6 +343,38 @@ function first(itr)
     x[1]
 end
 
+"""
+    first(predicate, coll)
+
+Get the first element of `coll` satisfying `predicate` wrapped in [`Some`](@ref).
+
+If no element of `coll` satisfies `predicate`, return `nothing`.
```
mbauman, line 351: This is interesting — and deviates from the existing `first` which would throw an error.  This gives me a bit of pause, but at a minimum it means that we cannot combine these docstrings:

```Julia
julia> first(_->true, [])

julia> first([])
```

base/abstractarray.jl
```diff
@@ -343,6 +343,38 @@ function first(itr)
...
+julia> something(first(iseven, [5, 3, 4, 2, 6, 8]))
+4
+
+julia> something(first(>(10), 1:10), 0)
+0
+```
+"""
+function first(predicate, itr)
+    for x in itr
+        predicate(x) && return Some(x)
```
mbauman, line 373: I'm not thrilled about `Some` here — I don't think we have a single API in base that returns `Some`.  That said, this is unlike the other find/search APIs in that they return indices or `RegexMatch`es or values from some other limited type set that does not include `Nothing`.

base/abstractarray.jl
```diff
@@ -388,6 +420,47 @@ julia> last([1; 2; 3; 4])
...
+julia> something(last(iseven, [5; 3; 4; 2; 6; 9]))
+6
+
+julia> something(last(>(10), 1:10), 0)
+0
+```
+"""
+function last(predicate, itr)
+    out = nothing
+    for x in itr
```
mbauman, line 450: I'd be inclined to not include this O(n) implementation  — I'd rather only support iterators that support `Iterators.Reverse` iteration.

mbauman (commented):
I'm not fully convinced about the functionality, but could be persuaded.

base/abstractarray.jl
non-Jedi: Good point on the docstrings.

base/abstractarray.jl
non-Jedi: Agreed. We don't include an O(n) implementation for regular `last` either.

base/abstractarray.jl
non-Jedi: I think I've convinced myself that at the very least this should be a separate function from `first`/`last`. Does `Iterators.first` (as suggested in OP) sound like a good name to you? In that case, we wouldn't have to create a predicate accepting version as it would compose with `Iterators.filter`.

In reality, I think all APIs that generically return an element of a collection but throw if the element isn't present should instead return `Union{Some{T},Nothing}`. This includes `first`, `last`, `getindex`, `reduce`, `foldl`, etc. This is obviously breaking; I should probably create a "Taking something seriously" issue to get feedback on that idea for julia 2.0.

non-Jedi:
Could someone educate me on the history of `Some`? @mbauman correctly noted that no base functions currently return values wrapped in `Some`. It seems odd to have an exported type in base that's basically unused (except for a few internal reduction functions). Was it added to base to be a point of coordination for external packages? Is remnant of the days of `Nullable`? Is it purely so that `something(Some(_->nothing))` can return `nothing`?

non-Jedi:
I'd like to bump this.

The core functionality I'm trying to get into `Base` here is some function that allows getting the first element of a lazily filtered (possibly empty) collection without wrapping in a `try` block; a function returning `Union{Some{T},Nothing}` seems like the obvious way of doing this. I'm relatively ambivalent on how that should be spelled.

If the decision is that a function for getting the first element of a lazily filtered collection isn't necessary in `Base`, I'm happy to close this. If a different approach would be better, I'd be happy to work on this further.

rfourquet:
I like trying to solve the stated problem

>   there's not a good way of finding the first item matching a predicate in a non-indexable iterable [...]

* So my first comment: while having a non-throwing `first` seems like a good general solution, what about designing a function which solves directly the problem? e.g. `Iterators.findfirst(f, iter)` (which could for example return the "index" (a la `enumerate`) together with the found object).
 
* Regarding using `Some`, it looks like an appropriate use here even if it's currently seldom used elsewhere. Note this other [PR](#34821) implementing "2-argument `get` returning `Union{Nothing,Some}`".

* I don't like putting this functionality in `Iterators.first`. As far as I am aware, other public functions in `Iterators` create an iterator. Naming this one `Iterators.first` feels just like adding a random namespace in front of `first` just because `first` is already taken. What about `tryfirst` (by analogy to `tryparse` which doesn't throw (unlike `parse`) ?

non-Jedi:
> * So my first comment: while having a non-throwing `first` seems like a good general solution, what about designing a function which solves directly the problem? e.g. `Iterators.findfirst(f, iter)` (which could for example return the "index" (a la `enumerate`) together with the found object).

Not all iterables have meaningful indices (e.g. `readeach`), or is the idea that it would work exactly like `enumerate` and count the number of elements?

> * I don't like putting this functionality in `Iterators.first`. As far as I am aware, other public functions in `Iterators` create an iterator. Naming this one `Iterators.first` feels just like adding a random namespace in front of `first` just because `first` is already taken. What about `tryfirst` (by analogy to `tryparse` which doesn't throw (unlike `parse`) ?

I agree on the unsuitability of the `Iterators` name; mostly I was trying to put it next to `Iterators.filter` where I feel it's most useful. `tryfirst` is a pretty good name given the precedent of `tryparse`. Another possibility would be to introduce a new namespace/module where this could live as `first` alongside `Union{Some{T},Nothing}`-returning versions of `getindex`, `reduce`, etc.

non-Jedi:
So revisiting this in 2023, I think what I really want is for [Maybe.jl](https://github.com/JuliaPreludes/Maybe.jl) ([which didn't have `Maybe.first` at the time I authored this PR](https://github.com/JuliaPreludes/Maybe.jl/pull/32)) to be a standard library.

If there's not appetite for a change like this, I'm happy to update this PR to rename the function `tryfirst` as suggested by @rfourquet above. I think the specific case of `first` without throwing on empty is more important than the generalization Maybe.jl gives.

DilumAluthge:

@non-Jedi If you're still interested in working on this PR, let me know, and I can re-open it.

PR #37121, opened 2020-08-19
build: drop explicit support for pre-Mavericks OS in the default binaries

vtjnash:
Hypothetically still supported in the build system back to 10.6 or so,
though I doubt it has been tested in a long time on such a system.

This lets us avoid needing to specify the libc++ explicitly, which makes
the build a bit simpler, and can help with building `clang` in our
`deps` folder (which may try to to build support libraries for foreign
targets, and get confused by these extra flags).

### DIFF
--- a/Make.inc
+++ b/Make.inc
@@ -83,9 +83,6 @@ USE_OPROFILE_JITEVENTS ?= 0
 # Set to 1 to enable profiling with perf
 USE_PERF_JITEVENTS ?= 0
 
-# libc++ is standard on OS X 10.9, but not for earlier releases
-USE_LIBCPP := 0
-
 # assume we don't have LIBSSP support in our compiler, will enable later if likely true
 HAVE_SSP := 0
 
@@ -463,8 +460,6 @@ else
 FC := $(CROSS_COMPILE)gfortran
 endif
 
-STDLIBCPP_FLAG :=
-
 ifeq ($(OS), FreeBSD)
 USEGCC := 0
 USECLANG := 1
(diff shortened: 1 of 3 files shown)


Make.inc
```diff
@@ -463,38 +460,35 @@ else
...
-ifeq ($(DARWINVER_GTE13),1)
-USE_LIBCPP := 1
-STDLIBCPP_FLAG := -stdlib=libstdc++
-else
-USE_LIBCPP := 0
-endif
+ifeq ($(DARWINVER_GTE13),0) # Lion / Mountain Lion specific configuration
+USEGCC := 0
+USECLANG := 1
+MACOSX_VERSION_MIN := 10.6
```
staticfloat, line 483: I guess there's not much point it setting `MACOSX_VERSION_MIN := 10.7` instead here, is there?  Better to be consistent and just have an "old MacOS" mode for `MACOSX_VERSION_MIN` and always set it to `10.6` or `10.9`.

deps/llvm.mk
```diff
@@ -88,6 +88,7 @@ LLVM_CPPFLAGS += $(CPPFLAGS)
 LLVM_LDFLAGS += $(LDFLAGS)
 LLVM_CMAKE += -DLLVM_TARGETS_TO_BUILD:STRING="$(LLVM_TARGETS)" -DCMAKE_BUILD_TYPE="$(LLVM_CMAKE_BUILDTYPE)"
 LLVM_CMAKE += -DLLVM_ENABLE_ZLIB=OFF -DLLVM_ENABLE_LIBXML2=OFF -DLLVM_HOST_TRIPLE="$(or $(XC_HOST),$(BUILD_MACHINE))"
+LLVM_CMAKE += -DCOMPILER_RT_ENABLE_IOS=OFF -DCOMPILER_RT_ENABLE_WATCHOS=OFF -DCOMPILER_RT_ENABLE_TVOS=OFF
```
staticfloat, line 91: Do we need to start including these in Yggdrasil's LLVM builds as well?

staticfloat (approved):

deps/llvm.mk
vtjnash: Not really. This just simplifies a bit what could go wrong, by avoiding building the support libraries for as many cross-compilation targets.

Make.inc
vtjnash: No point at all

ViralBShah: I think we should push to 10.9.

staticfloat: That is the default build configuration, look a little lower.

Issue #37046, opened 2020-08-14
Missing backedges


timholy:
Julia has a `MethodInstance` for `|(::Bool, ::Integer)`, but it has no backedges:
```julia
julia> methods(|, (Bool, Integer))
# 3 methods for generic function "|":
[1] |(x::Bool, y::Bool) in Base at bool.jl:37
[2] |(x::T, y::T) where T<:Integer in Base at promotion.jl:395
[3] |(a::Integer, b::Integer) in Base at int.jl:919

To edit a specific method, type the corresponding number into the REPL and press Ctrl+Q

julia> m = methods(|, (Bool, Integer)).ms[3]
|(a::Integer, b::Integer) in Base at int.jl:919

julia> m.specializations
svec(MethodInstance for |(::Int64, ::UInt16), MethodInstance for |(::UInt32, ::UInt8), MethodInstance for |(::UInt128, ::UInt32), MethodInstance for |(::Int64, ::UInt64), MethodInstance for |(::UInt128, ::UInt64), MethodInstance for |(::UInt8, ::UInt16), MethodInstance for |(::UInt16, ::UInt8), MethodInstance for |(::Bool, ::Integer), MethodInstance for |(::Int64, ::UInt32), MethodInstance for |(::Integer, ::Bool), MethodInstance for |(::UInt64, ::Integer), MethodInstance for |(::Integer, ::Int64), #undef, #undef, #undef, #undef, #undef, #undef)

julia> mi = m.specializations[8]
MethodInstance for |(::Bool, ::Integer)

julia> mi.backedges
ERROR: UndefRefError: access to undefined reference
Stacktrace:
 [1] getproperty(x::Core.MethodInstance, f::Symbol)
   @ Base ./Base.jl:33
 [2] top-level scope
   @ REPL[5]:1

To edit a specific method, type the corresponding number into the REPL and press Ctrl+Q
```
By inserting debugging code into the compiler:
```diff
diff --git a/base/compiler/abstractinterpretation.jl b/base/compiler/abstractinterpretation.jl
index 53a4f0f46d..144733d0b3 100644
--- a/base/compiler/abstractinterpretation.jl
+++ b/base/compiler/abstractinterpretation.jl
@@ -1260,6 +1260,14 @@ end
 # make as much progress on `frame` as possible (without handling cycles)
 function typeinf_local(interp::AbstractInterpreter, frame::InferenceState)
     @assert !frame.inferred
[...]

PR #37125, opened 2020-08-19
Avoid several bad `show_default` invalidations

### LINKED ISSUE #36005: Second "using" is slow (slower than third, can even be slower than the first)

timholy:
This fixes 53 quite critical invalidations from loading StaticArrays. At least by our old standards, 53 invalidations would be considered a modest number, but in this case they manage to hit the package-loading code, the docstring code, and the `MethodError` display code. Since particularly the first two are used by *everything*, these invalidations are about as bad as they could be.

The invalidation of the package-loading code is due to a rather hilarious chain of events. [This method definition](https://github.com/JuliaArrays/StaticArrays.jl/blob/ad583c99768f3a381ba20b1a0782c9ca50890b50/src/SArray.jl#L125-L126)
ends up invalidating [`show_default`](https://github.com/JuliaLang/julia/blob/89d6b469b633dde8f92b4d245ca3d8e5606e229c/base/show.jl#L417), which got called because `stale_cachefile` used an outdated variable name in a [`@debug` macro](https://github.com/JuliaLang/julia/blob/89d6b469b633dde8f92b4d245ca3d8e5606e229c/base/loading.jl#L1518) that ended up grabbing the global `mod` function. Since `show(::IO, ::Function)` can call `show_default`, this ended up invalidating basically all the package-loading code.

The docstring code got invalidated because `show(::IO, ::Pair{Symbol,Any})` [calls `show_default`](https://github.com/JuliaLang/julia/blob/89d6b469b633dde8f92b4d245ca3d8e5606e229c/base/show.jl#L907). This PR provides a dedicated path that circumvents `show_default`.

Finally, an innocuous-seeming function-print in `showerror(::IO, ::MethodError)` also led to quite a chain of invalidation.

EDIT: as pointed out below by @KristofferC, this fixes #36005. (~If master hadn't already fixed it, I haven't checked.~)

### DIFF
--- a/base/show.jl
+++ b/base/show.jl
@@ -904,7 +904,7 @@ function gettypeinfos(io::IO, p::Pair)
 end
 
 function show(io::IO, p::Pair)
-    isdelimited(io, p) && return show_default(io, p)
+    isdelimited(io, p) && return show_pairtyped(io, p)
     typeinfos = gettypeinfos(io, p)
     for i = (1, 2)
         io_i = IOContext(io, :typeinfo => typeinfos[i])
@@ -913,6 +913,11 @@ function show(io::IO, p::Pair)
         isdelimited(io_i, p[i]) || print(io, ")")
         i == 1 && print(io, get(io, :compact, false) ? "=>" : " => ")
     end
+end
+
+function show_pairtyped(io::IO, p::Pair{K,V}) where {K,V}
+    print(io, "Pair{", K, ',', V, '}')
+    show(io, (p.first, p.second))
 end
 
 function show(io::IO, m::Module)
@@ -2401,6 +2406,7 @@ end
 
 const undef_ref_str = "#undef"
 
+show(io::IO, ::UndefInitializer) = print(io, "UndefInitializer()")
 
 """
     summary(io::IO, x)
--- a/base/show.jl
+++ b/base/show.jl
@@ -916,7 +916,7 @@ function show(io::IO, p::Pair)
 end
 
 function show_pairtyped(io::IO, p::Pair{K,V}) where {K,V}
-    print(io, "Pair{", K, ',', V, '}')
+    show(io, typeof(p))
     show(io, (p.first, p.second))
 end
 
--- a/base/errorshow.jl
+++ b/base/errorshow.jl
@@ -287,7 +287,8 @@ function showerror(io::IO, ex::MethodError)
             )
             varnames = ("scalar", "array")
             first, second = arg_types_param[1] <: Number ? varnames : reverse(varnames)
-            print(io, "\nFor element-wise $(nouns[f]), use broadcasting with dot syntax: $first .$f $second")
+            fstring = f === Base.:+ ? "+" : "-"  # avoid depending on show_default for functions (invalidation)
+            print(io, "\nFor element-wise $(nouns[f]), use broadcasting with dot syntax: $first .$fstring $second")
         end
     end
     if ft <: AbstractArray
--- a/base/loading.jl
+++ b/base/loading.jl
@@ -1515,7 +1515,7 @@ function stale_cachefile(modpath::String, cachefile::String)
                     skip_timecheck = true
                     break
                 end
-                @debug "Rejecting cache file $cachefile because it provides the wrong uuid (got $build_id) for $mod (want $req_build_id)"
+                @debug "Rejecting cache file $cachefile because it provides the wrong uuid (got $build_id) for $req_key (want $req_build_id)"
                 return true # cachefile doesn't provide the required version of the dependency
             end
         end
--- a/stdlib/Sockets/src/addrinfo.jl
+++ b/stdlib/Sockets/src/addrinfo.jl
@@ -140,7 +140,7 @@ getaddrinfo(host::AbstractString, T::Type{<:IPAddr}) = getaddrinfo(String(host),
 function getaddrinfo(host::AbstractString)
     addrs = getalladdrinfo(String(host))
     if !isempty(addrs)
-        return addrs[begin]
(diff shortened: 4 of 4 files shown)


timholy:
Just to show the overall consequences: on this branch,
```julia
julia> using StaticArrays

julia> @time using Example
  0.002901 seconds (2.90 k allocations: 185.891 KiB)
```

On Julia 1.5,
```julia
julia> using StaticArrays

julia> @time using Example
  0.656540 seconds (1.33 M allocations: 66.060 MiB, 7.79% gc time)
```

That's gonna make a difference in package loading times.

There are a couple of `show_default` stragglers:
- we lack a `show` method for generic `IPAddr`: we have them for all subtypes but when inference is lost we end up inferring a call to `show_default`.
- something is calling `print(::ProcessException)` (we have a `showerror` method but not a `show` method)
- of course, `show(::IO, ::Function)` gets invalidated

For all of these, the trail of backedges truncates (#37046) so it is not easy to figure out where these are coming from.

KristofferC:
Fixes https://github.com/JuliaLang/julia/issues/36005?

timholy:
Heck yeah!

```julia
julia> @time using Plots
  3.217109 seconds (7.83 M allocations: 544.377 MiB, 4.78% gc time)

julia> @time using Plots
  0.000421 seconds (717 allocations: 38.922 KiB)
```

That 3.2s load time is pretty awesome (CCing @daschw, @BeastyBlacksmith). For reference, here's what I get on Julia 1.5:

```julia
julia> @time using Plots
  8.466600 seconds (18.09 M allocations: 1021.804 MiB, 3.44% gc time)

julia> @time using Plots
  0.685273 seconds (1.43 M allocations: 70.920 MiB, 11.50% gc time)

julia> @time using Plots
  0.000393 seconds (761 allocations: 44.250 KiB)
```

Both with Plots 1.6.0, FWIW. My guess is that the loading code was being invalidated repeatedly by several different dependencies of Plots (that extra time for the second `using` is just the most recent of these invalidations).

I also pushed a proposed fix for the `IPAddr` issue described above, although it has the downside of blocking any additional `IPAddr` subtypes. Such types would surely require other changes to the Sockets library so on balance I think it's a worthy fix, but let's see what others think. (It's also not a big deal either way, it only affects 2 additional invalidations.)

daschw:
Wow, that's awesome!

KristofferC:
> That 3.2s load time is pretty awesome (CCing @daschw, @BeastyBlacksmith). For reference, here's what I get on Julia 1.5:

>  FWIW. My guess is that the loading code was being invalidated repeatedly by several different dependencies of Plots (that extra time for the second using is just the most recent of these invalidations).

Out of curiosity, what do you get on master without this PR?

giordano:
@timholy what version of `FFMPEG.jl` are you using? How long does it take to load?

timholy:
@KristofferC, here's master (which of course has a lot of relevant recent work from you, Jeff, Jameson, Elliot, me, and likely others I'm not naming):
```julia
julia> @time using Plots
  3.370194 seconds (7.59 M allocations: 528.619 MiB, 6.43% gc time)

julia> @time using Plots
  0.377861 seconds (588.13 k allocations: 33.855 MiB, 4.82% gc time)

julia> @time using Plots
  0.000394 seconds (717 allocations: 38.922 KiB)
```

So most of the benefit over 1.5 is due to things we've recently added, but this PR adds further progress on the scale of ~0.5s if you're loading more than just Plots.

@giordano, here's what I get on master:
```julia
julia> @time using FFMPEG
  0.457856 seconds (994.93 k allocations: 63.036 MiB, 3.36% gc time)
```
I doubt this PR changes that (loading FFMPEG produces zero invalidations).

KristofferC:
> I doubt this PR changes that (loading FFMPEG produces zero invalidations).

@ElliotSaba has work in progress that will reduce this time quite significantly though.

giordano:
What version of FFMPEG.jl is that? Does it use FFMPEG_jll?

timholy:
```julia
(@v1.6) pkg> st FFMPEG
Status `~/.julia/environments/v1.6/Project.toml`
  [c87230d0] FFMPEG v0.3.0
```

base/show.jl
```diff
@@ -915,6 +915,11 @@ function show(io::IO, p::Pair)
     end
 end
 
+function show_pairtyped(io::IO, p::Pair{K,V}) where {K,V}
+    print(io, "Pair{", K, ',', V, '}')
```
JeffBezanson, line 919:
```suggestion
    show(io, typeof(p))
```

Issue #36946, opened 2020-08-06
String interpolation conflicts with markdown inline LaTeX syntax


jowch:
Writing an interpolated expression followed by second one in parentheses cause the content between the `$`s to be parsed as LaTeX.
```julia
julia> md"$(1) ($(2))"
  (1) (

  (2))
```
![Screenshot from 2020-08-06 10-10-59](https://user-images.githubusercontent.com/19571703/89561586-8d78e780-d7cd-11ea-80f2-0287061a1b4a.png)
Text parsed as LaTeX is displayed in purple.

Adding a space between the `(` and `$` triggers the second interpolation, but obviously leaves an unwanted space in the string.
```julia
julia> md"$(1) ( $(2))"
  (1) (

  (2))
```
![Screenshot from 2020-08-06 10-11-11](https://user-images.githubusercontent.com/19571703/89561694-b9946880-d7cd-11ea-96a1-e5b36e0d331b.png)

The insertion of a newline after an interpolation if it is the first thing in a markdown string is probably also a bug. This doesn't happen for a regular string.

My system information:
```
Julia Version 1.5.0
Commit 96786e22cc (2020-08-01 23:44 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: AMD Ryzen 5 3600 6-Core Processor
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-9.0.1 (ORCJIT, znver1)
```

PR #37130, opened 2020-08-20
Fix string interpolation into Markdown.Table and Markdown.List

### LINKED ISSUE #16194: Markdown strings and interpolation
How to do interpolation in markdown strings? Is the intended way to do it using dollars like in ordinary strings?
If so it seems that interpolation dollars and latex dollars are sometimes confused in markdown strings. For example the following snipped works:

```
x = "3"
md"""This works $(x) 

* a
* b
"""
```

This works 3 
- a
- b

While

```
md"""This works not 

* a $(x)
* b
"""
```

Throws an error:

```
MethodError: `latexinline` has no method matching latexinline(::Base.AbstractIOBuffer{Array{UInt8,1}}, ::Symbol)
Closest candidates are:
  latexinline(::IO, !Matched::Base.Markdown.LaTeX)
  latexinline(::IO, !Matched::Base.Markdown.Code)
  latexinline(::IO, !Matched::Array{T,1})
  ...
```

fredrikekre:
Fix string interpolation into `Markdown.Table` and `Markdown.List` in `md"..."` strings, fixes #16194.

### DIFF
--- a/stdlib/Markdown/src/Julia/interp.jl
+++ b/stdlib/Markdown/src/Julia/interp.jl
@@ -11,9 +11,13 @@ function interpinner(stream::IO, greedy = false)
     startswith(stream, '$') || return
     (eof(stream) || peek(stream, Char) in whitespace) && return
     try
-        return _parse(stream::IOBuffer, greedy = greedy)
+        return _parse(stream, greedy = greedy)
     catch e
-        return
+        if isa(e, Meta.ParseError)
+            return nothing
+        else
+            rethrow()
+        end
     end
 end
 
@@ -39,10 +43,19 @@ end
 
 toexpr(x) = x
 
-toexpr(xs::Vector{Any}) = Expr(:call, GlobalRef(Base,:getindex), Any, mapany(toexpr, xs)...)
+toexpr(xs::Union{Vector{Any},Vector{Vector{Any}}}) =
+    Expr(:call, GlobalRef(Base,:getindex), Any, mapany(toexpr, xs)...)
 
 for T in Any[MD, Paragraph, Header, Link, Bold, Italic]
     @eval function toexpr(md::$T)
         Expr(:call, typeof(md), $(map(x->:(toexpr(md.$x)), fieldnames(Base.unwrap_unionall(T)))...))
     end
+end
+
+function toexpr(md::Table)
+    Expr(:call, Table, toexpr(md.rows), md.align)
+end
+
+function toexpr(md::List)
+    Expr(:call, List, toexpr(md.items), md.ordered, md.loose)
 end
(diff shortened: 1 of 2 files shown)


stdlib/Markdown/src/Julia/interp.jl
```diff
@@ -11,7 +11,7 @@ function interpinner(stream::IO, greedy = false)
     startswith(stream, '$') || return
     (eof(stream) || peek(stream, Char) in whitespace) && return
     try
-        return _parse(stream::IOBuffer, greedy = greedy)
```
KristofferC, line 14: 😭 

Can we also check the exception type?

stdlib/Markdown/src/Julia/interp.jl
```diff
@@ -11,7 +11,7 @@ function interpinner(stream::IO, greedy = false)
...
+        return _parse(stream, greedy = greedy)
```
fredrikekre, line 14: This type-assert in `try-catch` threw sometimes since in some cases this function is passed a `GenericIOBuffer` instead of an `IOBuffer`....

stdlib/Markdown/src/Julia/interp.jl
fredrikekre: I don't quite know what exception type to look for, tests pass even if I comment out the try-catch so perhaps we can just do that?

thofma:
Does this by accident also fix https://github.com/JuliaLang/julia/issues/36946?

fredrikekre:
No.

stdlib/Markdown/src/Julia/interp.jl
fredrikekre:
````
julia> using Markdown, MethodAnalysis

julia> methodinstances(Markdown.interpinner)
2-element Vector{Core.MethodInstance}:
 MethodInstance for interpinner(::IOBuffer)
 MethodInstance for interpinner(::IOBuffer, ::Bool)

julia> begin
           x = "X"
           contains_X(md) = occursin(x, sprint(show, MIME("text/plain"), md))
           @assert contains_X(md"# $x") # H1
           @assert contains_X(md"## $x") # H2
           @assert contains_X(md"### $x") # H3
           @assert contains_X(md"x = $x") # Paragraph
           @assert contains_X(md"- $x") # List
           @assert contains_X(md"[$x](..)") # Link
           @assert contains_X(md"**$x**") # Bold
           @assert contains_X(md"*$x*") # Italic
           @assert contains_X( # Table
               md"""
               | name |
               |------|
               |  $x  |
               """)
       end

julia> methodinstances(Markdown.interpinner)
4-element Vector{Core.MethodInstance}:
 MethodInstance for interpinner(::IOBuffer)
 MethodInstance for interpinner(::Base.GenericIOBuffer{SubArray{UInt8,1,Vector{UInt8},Tuple{UnitRange{Int64}},true}})
 MethodInstance for interpinner(::IOBuffer, ::Bool)
 MethodInstance for interpinner(::Base.GenericIOBuffer{SubArray{UInt8,1,Vector{UInt8},Tuple{UnitRange{Int64}},true}}, ::Bool)
````

timholy (approved):

PR #31616, opened 2019-04-04
WIP: instrument julia to get per-method-instance llvm optimization timings

NHDaly:
This is a WIP PR that adds Dtrace timings to julia to allow printing timing information per method instance.

This was pair-programmed with @vchuravy 😊

We're currently trying to measure the `llvm` optimization times, but long-term it'd be nice to do lots of different sections.
Also, `dtrace` might only work on macOS/solaris, but there's an equivalent tracing tool for linux that should work with the same configuration files.

PR #37136, opened 2020-08-20
Add ccalls to enable dumping LLVM_OPT time per method instance.

NHDaly:
Adds these functions to toggle the output:
  - `ccall(:jl_dump_emitted_mi_name, Nothing, (Ptr{Nothing},), e_io.handle)`
  - `ccall(:jl_dump_llvm_opt, Nothing, (Ptr{Nothing},), l_io.handle)`

This will then be controlled from `SnoopCompileCore`, via https://github.com/timholy/SnoopCompile.jl/pull/135.

This logs a CSV file with mappings from C function name to Julia
MethodInstance name, and a YAML file with timings statistics about the
LLVM module being optimized, including:
 - Total time to optimize the llvm module
 - The `optlevel` used for the llvm module
 - The name of each function in the module, before and after
 - Number of instructions for each function, before and after
 - Number of basic blocks for each function, before and after

Co-Authored-By: @Sacha0 @vchuravy 
Fixes #31616

### DIFF
--- a/src/codegen.cpp
+++ b/src/codegen.cpp
@@ -95,6 +95,13 @@ typedef Instruction TerminatorInst;
 #include "processor.h"
 #include "julia_assert.h"
 
+JL_STREAM *dump_emitted_mi_name_stream = NULL;
+extern "C" JL_DLLEXPORT
+void jl_dump_emitted_mi_name(void *s)
+{
+    dump_emitted_mi_name_stream = (JL_STREAM*)s;
+}
+
 extern "C" {
 
 #include "builtin_proto.h"
@@ -7149,6 +7156,16 @@ jl_compile_result_t jl_emit_code(
         "functions compiled with custom codegen params must not be cached");
     JL_TRY {
         std::tie(m, decls) = emit_function(li, src, jlrettype, params);
+        if (dump_emitted_mi_name_stream != NULL) {
+            jl_printf(dump_emitted_mi_name_stream, "%s\t", decls.specFunctionObject.c_str());
+            // NOTE: We print the Type Tuple without surrounding quotes, because the quotes
+            // break CSV parsing if there are any internal quotes in the Type name (e.g. in
+            // Symbol("...")). The \t delineator should be enough to ensure whitespace is
+            // handled correctly. (And we don't need to worry about any tabs in the printed
+            // string, because tabs are printed as "\t" by `show`.)
+            jl_static_show(dump_emitted_mi_name_stream, li->specTypes);
+            jl_printf(dump_emitted_mi_name_stream, "\n");
+        }
     }
     JL_CATCH {
         // Something failed! This is very, very bad.
--- a/src/jitlayers.cpp
+++ b/src/jitlayers.cpp
@@ -38,6 +38,12 @@ void jl_dump_compiles(void *s)
 {
     dump_compiles_stream = (JL_STREAM*)s;
 }
+JL_STREAM *dump_llvm_opt_stream = NULL;
+extern "C" JL_DLLEXPORT
+void jl_dump_llvm_opt(void *s)
+{
+    dump_llvm_opt_stream = (JL_STREAM*)s;
+}
 
 static void jl_add_to_ee(std::unique_ptr<Module> m);
 static void jl_add_to_ee(std::unique_ptr<Module> &M, StringMap<std::unique_ptr<Module>*> &NewExports);
@@ -523,7 +529,36 @@ static void addPassesForOptLevel(legacy::PassManager &PM, TargetMachine &TM, raw
 
 CompilerResultT JuliaOJIT::CompilerT::operator()(Module &M)
 {
+    uint64_t start_time = 0;
+    if (dump_llvm_opt_stream != NULL) {
+        // Print LLVM function statistics _before_ optimization
+        // Print all the information about this invocation as a YAML object
+        jl_printf(dump_llvm_opt_stream, "- \n");
+        // We print the name and some statistics for each function in the module, both
+        // before optimization and again afterwards.
+        jl_printf(dump_llvm_opt_stream, "  before: \n");
+        for (auto &F : M.functions()) {
+            if (F.isDeclaration() || F.getName().startswith("jfptr_")) {
+                continue;
+            }
+            // Count number of Basic Blocks
+            int bbs = 0;
+            for (auto &B : F.getBasicBlockList()) {
+                std::ignore = B;
+                ++bbs;
+            }
+
+            // Each function is printed as a YAML object with several attributes
+            jl_printf(dump_llvm_opt_stream, "    \"%s\":\n", F.getName().str().c_str());
+            jl_printf(dump_llvm_opt_stream, "        instructions: %u\n", F.getInstructionCount());
+            jl_printf(dump_llvm_opt_stream, "        basicblocks: %u\n", bbs);
+        }
+
+        start_time = jl_hrtime();
+    }
+
     JL_TIMING(LLVM_OPT);
+
     int optlevel;
     if (jl_generating_output()) {
         optlevel = 0;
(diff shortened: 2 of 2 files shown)


NHDaly:
CC: @Sacha0 @vchuravy

NHDaly:
(I've opened this branch which just applies these same changes back to julia 1.5, so we can test our codebase which currently doesn't support 1.6 because of some of our package dependencies):
https://github.com/JuliaLang/julia/compare/v1.5.0...NHDaly:nhd-emit-per-method-llvm_timings--1.5

src/jitlayers.cpp
```diff
@@ -497,7 +503,36 @@ static void addPassesForOptLevel(legacy::PassManager &PM, TargetMachine &TM, raw
...
+            }
+            // Count number of Basic Blocks
+            int bbs = 0;
+            for (auto &B : F.getBasicBlockList()) {
+                std::ignore = B;
+                ++bbs;
+            }
+
+            // Each function is printed as a YAML object with several attributes
+            jl_printf(dump_llvm_opt_stream, "    %s:\n", F.getName().str().c_str());
```
vtjnash, line 526:
```suggestion
            jl_printf(dump_llvm_opt_stream, "    \"%s\":\n", F.getName().str().c_str());
```
(Is this necessary? I'm not well-versed in YAML)

src/jitlayers.cpp
```diff
@@ -539,6 +574,31 @@ CompilerResultT JuliaOJIT::CompilerT::operator()(Module &M)
                                  "The module's content was printed above. Please file a bug report");
     }
 
+    uint64_t end_time = 0;
+    if (dump_llvm_opt_stream != NULL) {
+        end_time = jl_hrtime();
+        jl_printf(dump_llvm_opt_stream, "  time_ns: %" PRIu64 "\n", end_time - start_time);
+
```
vtjnash, line 581:
```suggestion
        jl_printf(dump_llvm_opt_stream, "  optlevel: %d\n", optlevel);

```

src/jitlayers.cpp
```diff
@@ -497,7 +503,36 @@ static void addPassesForOptLevel(legacy::PassManager &PM, TargetMachine &TM, raw
...
+    uint64_t start_time = 0;
+    if (dump_llvm_opt_stream != NULL) {
+        // Print LLVM function statistics _before_ optimization
+        // Print all the information about this invocation as a YAML object
+        jl_printf(dump_llvm_opt_stream, "- \n");
+        // We print the name and some statistics for each function in the module, both
+        // before optimization and again afterwards.
+        jl_printf(dump_llvm_opt_stream, "  before: \n");
+        for (auto &F : M.functions()) {
+            if (F.getInstructionCount() == 0 || F.getName().str().rfind("jfptr_", 0) == 0) {
```
vtjnash, line 515:
```suggestion
            if (F.isDeclaration()) {
```

I'm not sure why the other check matters / makes sense (but seems like it should at least use [`startsWith`](https://llvm.org/doxygen/classllvm_1_1StringRef.html#aeb540aeadd339af860f5c217a85363b7))?

vtjnash (commented):
SGTM

src/jitlayers.cpp
NHDaly: Can't hurt, I agree.

I think I left it off because in previous commits I was using CSVs and i didn't know what would happen, and it occurred to me that the c function names can't have any whitespace anyway..

BUT still, i agree it seems nicer to include. Thanks. :)

src/jitlayers.cpp
NHDaly: Oooh, love it. Thanks, great suggestion! :)

src/jitlayers.cpp
NHDaly: We were leaving off the `jfptr_` functions, because they seem to be auto-generated boxing/unboxing functions, and just add noise to the output. I don't really remember exactly why we left them out... 

I can experiment with putting them back in, and see what happens? (We've been so focused on inference times the last three weeks, I've forgotten what some of this code does already)

NHDaly (commented):
Thanks for the review, @vtjnash! Sorry for my delay in responding.

As a short-term bandaid, `-O1` on julia 1.5 made a big improvement, so we stopped looking at this. But now that we're working through the inference timings, we'd like to go back to `-O2`, so we'll pick up this investigation again soon.

I'm sorry I let this languish for a bit, but do you think we'll be able to get this merged into 1.6? It would be really great to have this there so that our engineers don't need a custom build of julia to debug compilation times.

NHDaly:
And, similarly, I'm going to try to wrap up the per-method-instance _inference_ timings PR as well: https://github.com/JuliaLang/julia/pull/37749.

Ideally we'll be able to merge both of these this week so that they can make it into the v1.6 release! 😊 Does that seem possible to you? Sorry to be coming in so last-minute! ❤️ thanks for your help

src/codegen.cpp
```diff
@@ -6985,6 +6992,16 @@ jl_compile_result_t jl_emit_code(
         "functions compiled with custom codegen params must not be cached");
     JL_TRY {
         std::tie(m, decls) = emit_function(li, src, jlrettype, params);
+        if (dump_emitted_mi_name_stream != NULL) {
+            jl_printf(dump_emitted_mi_name_stream, "%s\t", decls.specFunctionObject.c_str());
+            // The quotes break CSV parsing if there are any internal quotes in the Type
+            // name (e.g.in Symbol("...")). The \t delineator should be enough.
+            // (And there can't be any tabs in it because the tabs are printed as "\t")
+            //jl_printf(dump_emitted_mi_name_stream, "\"");
```
Sacha0, line 7000: Perhaps drop this commented quote printing line and that two lines below in accord with the (great!) immediately preceding comment, and touch up the comment? :)

src/jitlayers.cpp
Sacha0: > We were leaving off the jfptr_ functions, because they seem to be auto-generated boxing/unboxing functions, and just add noise to the output. I don't really remember exactly why we left them out...

My memory of the `F.getName().str().rfind("jfptr_", 0)` check is hazy by this point as well, but what memory I do have is consistent with your recollection, i.e. the `jfptr_`-related output practically seemed like noise.

The `isDeclaration` and `startsWith` rewrites sound great! :)

src/jitlayers.cpp
```diff
@@ -539,6 +574,32 @@ CompilerResultT JuliaOJIT::CompilerT::operator()(Module &M)
...
+    uint64_t end_time = 0;
+    if (dump_llvm_opt_stream != NULL) {
+        end_time = jl_hrtime();
+        jl_printf(dump_llvm_opt_stream, "  time_ns: %" PRIu64 "\n", end_time - start_time);
+        jl_printf(dump_llvm_opt_stream, "  optlevel: %d\n", optlevel);
+
+        // Print LLVM function statistics _after_ optimization
+        jl_printf(dump_llvm_opt_stream, "  after: \n");
+        for (auto &F : M.functions()) {
+            if (F.getInstructionCount() == 0 || F.getName().str().rfind("jfptr_", 0) == 0) {
```
Sacha0, line 586: I suppose the `isDeclaration` and `startsWith` rewrites apply here as well? :)

Sacha0 (commented):
This patch looks great! Thanks @NHDaly! :)

It looks like one comment from Jameson remains outstanding (rewriting a conditional in terms of `isDeclaration` and `startsWith`), which I noted applies to similar code later in the patch as well. (Perhaps the duplicated code snippet should be factored into a little helper function? (It's sufficiently little duplication that I don't mind either way.) Otherwise this patch looks ready to go! :)

src/codegen.cpp
NHDaly: Thanks. done!

src/jitlayers.cpp
NHDaly: Done. I've rewritten this as `if (F.isDeclaration() || F.getName().startswith("jfptr_"))`.

src/jitlayers.cpp
NHDaly: yep; done. thanks

src/jitlayers.cpp
NHDaly: And after reflecting over this a bit, i think we skipped them simply because they don't add anything to the user.

The question one wants to answer is how much LLVM time did each julia function contribute to overall compilation times. we don't have any julia function to associate with the `jfptr_*` functions, so they will remain C function names, and from what i understand, every function will get on of them generated for it, and they will all be equally expensive to compile. So i don't think reporting them adds anything. They'll contribute noise in the downstream analyses, so I'm in favor of continuing to leave them off. :)

Does that make sense? Thanks!

NHDaly:
Thanks @Sacha0! Addressed comments, rebased on top of master, and pushed.

Sacha0 (approved):
Looks great! Thanks @NHDaly! :)

vchuravy (approved):

vchuravy:
PR needs squashing.

NHDaly:
Done. Thanks! :)

PR #37149, opened 2020-08-21
Symbol clean up and strlen fixes

yuyichao:
* Require the symbol to fit in ssize_t

  This is the limit we have on arrays and I don't think anyone is using
  symbols that is as long as half the full address space.

* Remove some unnecessary embeded NUL byte checks when constructing symbols

  Almost all use of `jl_symbol_n` have the caller checked for embedded NUL byte already.
  This technically introduces a C API/ABI breakage but it shouldn't matter on all platforms we support.

* Fix ccall of `strlen`.

------------------------

The change to `base/secretbuffer.jl` should be backported. None of the others should. It's funny that the only bug that can cause silent error that I found is in security related code = = ...........

### DIFF
--- a/base/expr.jl
+++ b/base/expr.jl
@@ -9,11 +9,10 @@ Generates a symbol which will not conflict with other variable names.
 """
 gensym() = ccall(:jl_gensym, Ref{Symbol}, ())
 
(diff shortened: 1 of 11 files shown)


KristofferC:
@yuyichao https://github.com/JuliaLang/julia/pull/37210/commits/9f1cc7ec7b7b6342b1256a34bd3bebe7aa94ae2b

PR #37151, opened 2020-08-22
[NFC] Two minor clean ups in gc lowering pass

yuyichao:
Should not affect anything... Just a bit confusing to read...

Split out from other functional changes...


Sacha0 (approved):
Reads better! :)

PR #37152, opened 2020-08-22
Add missing docstring for `zero` function which input is type

AtsushiSakai:
I found `zero` function can accept a type, not only a variable:

```
julia> zero(1.0)
0.0

julia> zero(Float64)
0.0
```

But, its docstring looks like missing later one, so I added it.

### DIFF
--- a/base/number.jl
+++ b/base/number.jl
@@ -223,7 +223,7 @@ map(f, x::Number, ys::Number...) = f(x, ys...)
 
 """
     zero(x)
-    zero(T::type)
+    zero(::Type)
 
 Get the additive identity element for the type of `x` (`x` can also specify the type itself).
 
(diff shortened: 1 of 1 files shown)


base/number.jl
```diff
@@ -223,6 +223,7 @@ map(f, x::Number, ys::Number...) = f(x, ys...)
 
 """
     zero(x)
+    zero(T::type)
```
staticfloat, line 226: I think it's a little clearer if this is labeled as `zero(::Type{<:Number})`

sostock: There are types which are not `<:Number` but support `zero`, e.g. `Dates.Day`.

staticfloat: True, but those are not hitting this particular method.  It's probably okay to just call it `::Type` then though.

AtsushiSakai: OK. I change it to `::Type`

staticfloat:
Thanks @AtsushiSakai!

PR #37153, opened 2020-08-22
Support LLVM 10 in GC checker

yuyichao:
Ref https://github.com/llvm/llvm-project/commit/2f169e7cdd9973d2aa4cba6b0a09126a5e7268ec
Ref https://github.com/llvm/llvm-project/commit/6b85f8e99b38aed5bb4eeb4ca8d7ccdd5a1bf48c

Also handle noreturn function call in GC frame counting,
and remove some old code for LLVM <= 7

### DIFF
--- a/src/clangsa/GCChecker.cpp
+++ b/src/clangsa/GCChecker.cpp
@@ -12,11 +12,7 @@
 #include "clang/StaticAnalyzer/Frontend/FrontendActions.h"
 #include "clang/Tooling/CompilationDatabase.h"
 #include "clang/Tooling/Tooling.h"
-#if LLVM_VERSION_MAJOR >= 8
 #include "clang/StaticAnalyzer/Frontend/CheckerRegistry.h"
-#else
-#include "clang/StaticAnalyzer/Core/CheckerRegistry.h"
-#endif
 
 #include <iostream>
 #include <memory>
@@ -34,6 +30,7 @@
 using std::make_unique;
 #else
 using llvm::make_unique;
+#define PathSensitiveBugReport BugReport
 #endif
 
 namespace {
@@ -42,6 +39,16 @@ using namespace ento;
 
 #define PDP std::shared_ptr<PathDiagnosticPiece>
 #define MakePDP make_unique<PathDiagnosticEventPiece>
+
+static const Stmt *getStmtForDiagnostics(const ExplodedNode *N)
+{
+#if LLVM_VERSION_MAJOR >= 10
+    return N->getStmtForDiagnostics();
+#else
+    return PathDiagnosticLocation::getStmt(N);
+#endif
+}
+
 
 class GCChecker
     : public Checker<
@@ -60,7 +67,7 @@ class GCChecker
   template <typename callback>
   void report_error(callback f, CheckerContext &C, const char *message) const;
   void report_error(CheckerContext &C, const char *message) const {
-    return report_error([](BugReport *) {}, C, message);
+    return report_error([](PathSensitiveBugReport *) {}, C, message);
   }
   void
   report_value_error(CheckerContext &C, SymbolRef Sym, const char *message,
@@ -227,11 +234,7 @@ private:
 
 public:
   void checkBeginFunction(CheckerContext &Ctx) const;
-#if LLVM_VERSION_MAJOR >= 7
   void checkEndFunction(const clang::ReturnStmt *RS, CheckerContext &Ctx) const;
-#else
-  void checkEndFunction(CheckerContext &Ctx) const;
-#endif
 #if LLVM_VERSION_MAJOR >= 9
   bool evalCall(const CallEvent &Call, CheckerContext &C) const;
 #else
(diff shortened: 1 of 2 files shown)


vchuravy:
Running with the core checker excluded I also get the following failure:

```
/home/vchuravy/src/julia/src/jitlayers.cpp:358:1: error: Non-popped GC frame present at end of function
}
^
/home/vchuravy/src/julia/src/jitlayers.cpp:328:9: note: Assuming field 'invoke' is equal to NULL
    if (unspec->invoke != NULL)
        ^~~~~~~~~~~~~~~~~~~~~~
/home/vchuravy/src/julia/src/jitlayers.cpp:328:5: note: Taking false branch
    if (unspec->invoke != NULL)
    ^
/home/vchuravy/src/julia/src/jitlayers.cpp:331:17: note: Field 'invoke' is equal to NULL
    if (unspec->invoke == NULL) {
                ^
/home/vchuravy/src/julia/src/jitlayers.cpp:331:5: note: Taking true branch
    if (unspec->invoke == NULL) {
    ^
/home/vchuravy/src/julia/src/jitlayers.cpp:333:9: note: GC frame changed here
        JL_GC_PUSH1(&src);
        ^~~~~~~~~~~~~~~~~
/home/vchuravy/src/julia/src/jitlayers.cpp:335:13: note: Assuming the condition is false
        if (jl_is_method(def)) {
            ^~~~~~~~~~~~~~~~~
/home/vchuravy/src/julia/src/julia.h:1136:30: note: expanded from macro 'jl_is_method'
#define jl_is_method(v)      jl_typeis(v,jl_method_type)
                             ^~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/vchuravy/src/julia/src/julia.h:144:25: note: expanded from macro 'jl_typeis'
#define jl_typeis(v,t) (jl_typeof(v)==(jl_value_t*)(t))
                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/vchuravy/src/julia/src/julia.h:129:22: note: expanded from macro 'jl_typeof'
#define jl_typeof(v) _jl_typeof((jl_value_t*)(v))
                     ^
/home/vchuravy/src/julia/src/jitlayers.cpp:335:9: note: Taking false branch
        if (jl_is_method(def)) {
        ^
/home/vchuravy/src/julia/src/jitlayers.cpp:349:16: note: Assuming 'src' is null
        assert(src && jl_is_code_info(src));
               ^~~
/usr/include/assert.h:90:27: note: expanded from macro 'assert'
     (static_cast <bool> (expr)                                         \
                          ^~~~
/home/vchuravy/src/julia/src/jitlayers.cpp:349:20: note: Left side of '&&' is false
        assert(src && jl_is_code_info(src));
                   ^
/home/vchuravy/src/julia/src/jitlayers.cpp:349:9: note: '?' condition is false
        assert(src && jl_is_code_info(src));
        ^
/usr/include/assert.h:90:7: note: expanded from macro 'assert'
     (static_cast <bool> (expr)                                         \
      ^
/home/vchuravy/src/julia/src/jitlayers.cpp:358:1: note: Non-popped GC frame present at end of function
}
^
```

yuyichao:
It thinks the `assert` returns? Does your `assert.h` have that?

In `checkEndFunction` there doesn't seem to be a check for noreturn function and there's a return statement passed in. Any way for you to check what's going on there?

yuyichao:
Actually I can reproduce it too if core check is disabled. It seems that the return statement was `NULL` which probably mean it was end of function due to a noreturn function call. I don't know why removing the core check cause this to happen but

```patch
diff --git a/src/clangsa/GCChecker.cpp b/src/clangsa/GCChecker.cpp
index 4e7d708a67..8bba1d37d2 100644
--- a/src/clangsa/GCChecker.cpp
+++ b/src/clangsa/GCChecker.cpp
@@ -727,7 +727,7 @@ void GCChecker::checkEndFunction(const clang::ReturnStmt *RS,
     C.addTransition(State);
   if (!C.inTopFrame())
     return;
-  if (C.getState()->get<GCDepth>() > 0)
+  if (RS && C.getState()->get<GCDepth>() > 0)
     report_error(C, "Non-popped GC frame present at end of function");
 }
```

seems to fix it and also seems correct in general (there are other cases where a NULL is passed here and this just seem to be the only case where it actually trigger an error...)

yuyichao:
This is not user facing code and I believe there shouldn't be regression on LLVM 9 at least so I'll merge soon to make it easier for me to test a few other changes.

I also removed some old code for LLVM <= 7 support.

It also seems that the core analyzer might be responsible for handling no-return function in general. Even with the NULL return statement case factored out there are other related issues from the caller of the function. Still not sure why the core analyzer complains what it does. I have a non-asserting and release build of LLVM maybe that makes a difference?

vchuravy:
I am just running the Yggdrasil build, so that is also non-asserting and release. It's weird that we are also seeing it on CI otherwise I might blame a header difference.

PR #37156, opened 2020-08-22
Make parser link work in Documenter

zlatanvasovic:
It currently doesn't work, as you can see on https://docs.julialang.org/en/v1/base/math/#Customizable-binary-operators-1.

### DIFF
--- a/doc/src/base/math.md
+++ b/doc/src/base/math.md
@@ -198,7 +198,7 @@ and
 \oplus `⊕`
 
 The complete list is in the parser code:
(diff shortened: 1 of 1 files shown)


StefanKarpinski (approved):

zdroid:
@StefanKarpinski can this be merged before 1.5.1?

fredrikekre:
1.5.1 is done some weeks ago but it will be in 1.6

StefanKarpinski:
I think it would be fine to backport documentation fixes, but I'm not sure what our policy on that is these days.

PR #37088, opened 2020-08-17
Reduce reliance on promotion operations via container typing

timholy:
While building Julia, we ~have~had a *lot* of calls to promote, often with abstract types. This short-circuits most of these by declaring the container type at the outset.

This has one user-visible outcome, improved inference for `walkdir`.

After my complete PR series is submitted, I'll post some benchmarks here on the impact on building Julia itself. Obviously no one PR accounts for the whole thing, but this one may be as good a place as any to deposit this info.

Issue #36892, opened 2020-08-03
Possible missing backedges and spurious (?) invalidation triggered by UnionAll


timholy:
I've been poking at this for a while now and still can't quite isolate this the way I want, so let me show you what I have. I'm trying to figure out how to set up a test for a fix for invalidations from loading ColorTypes.

The invalidations are triggered by adding new `eltype(::Type)` specializations. The ColorTypes issue is fixed just by adding
```julia
eltype(::Type{<:NamedTuple}) = Any
```
but I wonder if there's a more systematic fix? For the record we already have a definition
```julia
eltype(::Type{NamedTuple{names,T}} where names) where {T} = eltype(T)
```
so this is just about handling the case where the type is not fully specified.

Here is a prototype and analysis:
```julia
julia> f(::Type{T}) where T = eltype(T)
f (generic function with 1 method)

julia> function g()
           while true  # block interpreter
               return f(T where T<:NamedTuple{(:indent,)})
           end
       end
g (generic function with 1 method)

julia> g()
Any

julia> using SnoopCompileCore

julia> abstract type Container{T} end

julia> invlist = @snoopr Base.eltype(::Type{C}) where {T,C<:Container{T}} = T;

julia> using SnoopCompile

julia> trees = invalidation_trees(invlist)
1-element Vector{SnoopCompile.MethodInvalidations}:
 inserting eltype(::Type{C}) where {T, C<:Container{T}} in Main at REPL[6]:1 invalidated:
   backedges: 1: superseding eltype(::Type) in Base at array.jl:123 with MethodInstance for eltype(::Type{A} where A<:(NamedTuple{(:io,),_A} where _A<:Tuple{IO})) (3 children)
              2: superseding eltype(::Type) in Base at array.jl:123 with MethodInstance for eltype(::Type{A} where A<:(NamedTuple{(:init,),_A} where _A<:Tuple{IOContext})) (4 children)
              3: superseding eltype(::Type) in Base at array.jl:123 with MethodInstance for eltype(::Type{A} where A<:(NamedTuple{(:indent, :first_block, :sorted, :by),_A} where _A<:Tuple{Int64,Bool,Bool,Function})) (4 children)
              4: superseding eltype(::Type) in Base at array.jl:123 with MethodInstance for eltype(::Type{A} where A<:(NamedTuple{(:indent, :sorted, :by),_A} where _A<:Tuple{Int64,Bool,Function})) (5 children)
[...]

PR #37163, opened 2020-08-23
Improve inference for paths leading to `similar`

timholy:
For many packages that extend array functionality, `similar` now serves as the single biggest source of invalidation. These invalidations have been hard to diagnose using my SnoopCompile machinery, because runtime dispatch and a lack of "narrowing" of types lead to an absence of backedges (#36892). Without backedges, it's not easy to discover the callers, which is where you'd have to make changes to improve the quality of inference. This motivated me to implement a new general-purpose tool, `findcallers` in the [MethodAnalysis](https://github.com/timholy/MethodAnalysis.jl) package, that lets you search the type-inferred code of *every* `MethodInstance` for calls of a particular function with particular signature. (If you're already a MethodAnalysis user, note that 0.4.0 is the first release where `findcallers` is reasonably reliable, and even then it has some significant limitations as described in the docs).

I was able to use this tool to discover that the issue is not with the implementation of `similar` *per se*, but that we had many cases of comprehensions, `map`, broadcasting, and concatenation with poor inferrability. This PR, together with a companion to Pkg that I'll submit shortly, fixes almost all of them. Some of the patterns were a bit surprising, for example drilling down into
```julia
using Cthulhu
@descend vcat([1,2,3], 4)
```
you will discover that you get down into `_cat_t(::Val{1},::DataType,::Vector{Int64},::Vararg{Any,N} where N)::Any` and you end up calling `similar` with poor type information. (Conversely, `push!(copy([1,2,3]), 4)` has no such problems.)

An earlier version of this PR fixed all cases with poor container-type inference---we still had some with poor element-type inference from cases where `collect(::Generator{I,F})` had stopped specializing on the function type `F`, but since `similar` tends to be specialized on `I` this isn't really that big of a deal. However, I've at least temporarily abandoned the attempt to fix every last instance because of a single function, [maptwice](https://github.com/JuliaLang/julia/blob/0dab9e3f9f569a6462a0771d92c9e54c107164ea/base/asyncmap.jl#L156-L189), heavily used by `pmap`. If we could use `mapany` there then our inference problems there too would go away, but we have tests that explicitly check for element-type narrowing so this seems to be a no-go. It's a little odd given that Distributed has rampant inference problems to insist on the element narrowing, but perhaps that's essentially a way of compensating for Distributed's challenges in this regard. (And gettting good inference does seem in principle like it could be harder in a distributed context.)

Even having reverted those two `mapany`s in `maptwice`, this PR cuts the number of remaining invalidations from loading StaticArrays in half, compared to a local branch that includes all my still-open invalidation PRs. That leaves us with just 24 invalidations from StaticArrays, which is rather nice given that on Julia 1.5 it's more than 2000. Even for CategoricalArrays, `similar` now (locally, i.e., with all my modifications) only contributes 30 unique invalidations out of more than 2000 total, resolving a concern that @nalimilan expressed in https://github.com/JuliaData/CategoricalArrays.jl/issues/177.

### Noteworthy details (particularly directed at reviewers of this PR)

First, this changes field types of a few `struct`s, including `Cmd` which is exported. I don't think `Cmd` needs to support multidmensional arrays of Strings, but perhaps I'm mistaken.

Reviewers, also note the extra pickiness about arg types in REPL keymap functions. Since `setup_interface` converts all user-supplied keymaps into `Dict{Any,Any}` and none of the other functions are part of REPL's official interface, I don't think this would be considered breaking, but I did have to change some of REPL's tests (some of which directly call internal functions) in order to make them pass. As a reminder, the entire LineEdit module is under `@nospecialize`, so unless method arguments provide type information then inference uses `Any` no matter what the argument type actually is.

A final item of note concerns Julia's own build process. Together with my other open PRs, the total number of precompile statements is down to about 1550, a drop of about 400 compared to a couple months ago. This is getting close to the point where we'll actually have to change [this build-time assertion](https://github.com/JuliaLang/julia/blob/01fd807d5e9e8c4b66d92a598096406f4e3b1500/contrib/generate_precompile.jl#L197). The system image is also the smallest I've seen it in a long time:
```julia
tim@diva:~/src/julia-master$ ls -l usr/lib/julia/sys.so 
-rwxr-xr-x 1 tim holy 141802056 Aug 23 08:05 usr/lib/julia/sys.so
```
(Reference https://github.com/JuliaLang/julia/pull/37088#issuecomment-675106241 for where I was relatively recently.) While my general experience has been that the precompilation phase has gone slightly faster with fewer `MethodInstance`s, this PR (or perhaps more likely, the companion Pkg PR I'll also submit) violates that: now I'm getting
```julia
Generating precompile statements... 30/30
Executing precompile statements... 1543/1554
Precompilation complete. Summary:
Total ─────── 141.154365 seconds
Generation ── 118.166039 seconds 83.7141%
Execution ───  22.988326 seconds 16.2859%
    LINK usr/lib/julia/sys.so
```
which is a big slowdown. I'm guessing it's because inference stops "punting" so much, but I haven't made any attempt to investigate this.

### DIFF
--- a/base/Enums.jl
+++ b/base/Enums.jl
@@ -116,7 +116,7 @@ julia> instances(Fruit)
 (apple, orange, kiwi)
 ```
 """
-macro enum(T, syms...)
+macro enum(T::Union{Symbol,Expr}, syms...)
     if isempty(syms)
         throw(ArgumentError("no arguments given for Enum $T"))
     end
@@ -131,7 +131,7 @@ macro enum(T, syms...)
     elseif !isa(T, Symbol)
         throw(ArgumentError("invalid type expression for enum $T"))
     end
-    values = basetype[]
+    values = basetype[]::Vector
     seen = Set{Symbol}()
     namemap = Dict{basetype,Symbol}()
     lo = hi = 0
--- a/base/Enums.jl
+++ b/base/Enums.jl
@@ -131,7 +131,7 @@ macro enum(T::Union{Symbol,Expr}, syms...)
     elseif !isa(T, Symbol)
         throw(ArgumentError("invalid type expression for enum $T"))
     end
-    values = basetype[]::Vector
+    values = Vector{basetype}()
     seen = Set{Symbol}()
     namemap = Dict{basetype,Symbol}()
     lo = hi = 0
--- a/base/abstractarray.jl
+++ b/base/abstractarray.jl
@@ -211,11 +211,12 @@ eachindex(itrs...) = keys(itrs...)
 # eachindex iterates over all indices. IndexCartesian definitions are later.
 eachindex(A::AbstractVector) = (@_inline_meta(); axes1(A))
 
-@noinline function throw_eachindex_mismatch(::IndexLinear, A...)
-    throw(DimensionMismatch("all inputs to eachindex must have the same indices, got $(join(eachindex.(A), ", ", " and "))"))
+
+@noinline function throw_eachindex_mismatch_indices(::IndexLinear, inds...)
+    throw(DimensionMismatch("all inputs to eachindex must have the same indices, got $(join(inds, ", ", " and "))"))
 end
-@noinline function throw_eachindex_mismatch(::IndexCartesian, A...)
-    throw(DimensionMismatch("all inputs to eachindex must have the same axes, got $(join(axes.(A), ", ", " and "))"))
+@noinline function throw_eachindex_mismatch_indices(::IndexCartesian, inds...)
+    throw(DimensionMismatch("all inputs to eachindex must have the same axes, got $(join(inds, ", ", " and "))"))
 end
 
 """
@@ -269,7 +270,7 @@ function eachindex(::IndexLinear, A::AbstractArray, B::AbstractArray...)
     @_inline_meta
     indsA = eachindex(IndexLinear(), A)
     _all_match_first(X->eachindex(IndexLinear(), X), indsA, B...) ||
-        throw_eachindex_mismatch(IndexLinear(), A, B...)
+        throw_eachindex_mismatch_indices(IndexLinear(), eachindex(A), eachindex.(B)...)
     indsA
 end
 function _all_match_first(f::F, inds, A, B...) where F<:Function
--- a/base/abstractarray.jl
+++ b/base/abstractarray.jl
@@ -2232,13 +2232,7 @@ end
 map(f, A::AbstractArray) = collect_similar(A, Generator(f,A))
 
 mapany(f, A::AbstractArray) = map!(f, Vector{Any}(undef, length(A)), A)
-function mapany(f, itr)
-    ret = []
-    for x in itr
-        push!(ret, f(x))
-    end
-    return ret
-end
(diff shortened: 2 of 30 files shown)


timholy:
While it's basically incidental to the thrust of this PR, I added a `@nospecialize` to `_str_sizehint`, which had 56 `MethodInstance` specializations. But perhaps that was a bad idea since `sizehint` is typically a performance optimization. It's just that the way `_str_sizehint` is written makes it seem like it wants to make decisions at runtime.

I should also ask about `Base.print_to_string`---it has 222 specializations, which seems like a lot.

base/abstractarray.jl
```diff
@@ -2229,7 +2230,14 @@ end
...
 
-mapany(f, itr) = map!(f, Vector{Any}(undef, length(itr)::Int), itr)  # convenient for Expr.args
+mapany(f, A::AbstractArray) = map!(f, Vector{Any}(undef, length(A)), A)
+function mapany(f, itr)
+    ret = []
+    for x in itr
+        push!(ret, f(x))
+    end
+    return ret
+end
```
pabloferz, line 2240: Maybe,
```suggestion
mapany(f, itr) = Any[f(x) for x in itr]
```
?

base/Enums.jl
```diff
@@ -131,7 +131,7 @@ macro enum(T, syms...)
     elseif !isa(T, Symbol)
         throw(ArgumentError("invalid type expression for enum $T"))
     end
-    values = basetype[]
+    values = basetype[]::Vector
```
pabloferz, line 134:
```suggestion
    values = Vector{basetype}()
```
?

base/abstractarray.jl
timholy: Yes, that seems safe. I wasn't certain originally but I've verified that never calls `similar`.

KristofferC:
I tried master vs this for building julia sysimage and with this PR it was 4 seconds faster so I don't see the 

> which is a big slowdown. 

effect, at least.

base/errorshow.jl
```diff
@@ -781,7 +781,7 @@ function show_backtrace(io::IO, t::Vector)
 
     try invokelatest(update_stackframes_callback[], filtered) catch end
     # process_backtrace returns a Vector{Tuple{Frame, Int}}
-    frames = (first.(filtered))::Vector{StackFrame}
+    frames = map(x->first(x)::StackFrame, filtered)
```
pabloferz, line 784:
```suggestion
    frames = StackFrame[first(x) for x in filtered]
```

pabloferz:
The only thing that makes me uneasy here is the `_str_sizehint`, I feel it could have performance impacts, so I'd be inclined to leave that out.

base/errorshow.jl
timholy: Actually these two are different. My version asserts that `first(x)` is already a `StackFrame`; your version will call `convert(StackFrame, x::Any)::StackFrame`. Given that we know `x` is already a `StackFrame`, that makes my version less vulnerable to `convert` invalidations.

timholy:
I dropped the `_str_sizehint` commit.

KristofferC (approved):

Issue #37164, opened 2020-08-23
Warn that at least one deprecation warning emitted, default depwarn=yes again


BioTurboNick:
Based on the discussion here: #35362 and having run into this silent failure mode, I suggest a compromise:

- Restore default to `depwarn=yes`
- Add `depwarn=verbose`
- Under `yes`, Julia tracks deprecations received but does not print them. On receipt, it just prints a single warning shown below, with details accessible by a function call.
- Under `verbose`, all warnings are printed as they are received, which is the current behavior for `yes`
- Under `no`, no deprecations are tracked.

The single warning under `yes` would be something like:
```
┌ Warning: The running code contains deprecations at REPL[1]:1.
│ Use `depwarns()` to print the list of warnings, or start Julia with --depwarns=verbose to show warnings when they happen.
│ Deprecated code will still execute correctly. This warning will not be displayed again during this session.
└ @ REPL[1]:1
```

This could even just be a boolean switch that gets flipped on for the first deprecation received and nothing else, just to inform the user that they can take an action to see the warnings. More usefully, it could store the deprecation messages received in a set and display the specific ones received when a function is called, as in the example message above.

I can understand that it might still bother users of big packages to see a warning at all, but I think the tradeoff would be fair?

BioTurboNick:
The actual problem is that warnings are useful and important information that should not be completely silent by default, or else they wouldn't be warnings. There must be a default pathway leading the user from ignorance to knowledge that doesn't involve hunting around or stumbling on a problem unawares.

BioTurboNick:
I'm just suggesting an improvement that keeps the benefits of warnings while addressing the slowdowns caused by tons of warnings being emitted.

StefanKarpinski:
I think it may make sense to have soft & hard deprecations: soft ones are silent by default whereas hard ones warn by default. Then you can opt to silence all deprecations, print all deprecations and/or make deprecations fatal.

StefanKarpinski:
The problem here is that some people (e.g. Base) want to introduce deprecations that are silent by default whereas other people, e.g. LightGraphs, want to use the same deprecation mechanisms to introduce deprecations that are on by default so that they can guide people to change their code while allowing it to keep running like we did pre-1.0.

BioTurboNick:
I see, thanks @StefanKarpinski.

Would I be correct in presuming that among Base there is a desire to avoid adding overhead without a critical reason, and tracking warnings would do that?

[...]

PR #33697, opened 2019-10-28
Syntax for multidimensional arrays

BioTurboNick:
Use multiple semicolons to denote >2 dimensions in an inline N-dimensional array specification.

**Summary of all changes in PR and how it works: https://github.com/JuliaLang/julia/pull/33697#issuecomment-797029617**

Breaking changes: Extra semicolons in matrix expression currently ignored. However, use is likely to be rare and unintentional. ~~#37168 adds deprecation warning for this.~~ That PR was merged.

Related work:
- [x] Typed concatenation methods
- [x] Matching inline display (as used in print([1 2;3 4;;5 6;7 8;;9 0;2 3]); necessary for printing of vectors containing multidimensional arrays)
- [x] Performance checks / optimization
- [x] Documentation
- [x] Fix inline display for arrays too long to display
- [x] Array concatenation
- [x] General case (`Any` argument)
- [x] Handling for concatenating elements and arrays together
- [x] Implement ;; as second dimension as discussed
- [x] ~~Extend terminal semicolons to vector expression e.g. `[1,2,3;;;]`~~ Removed
- [x] Check print() methods e.g. `print(rand(Int8, 2,2,2))` (temporary)
- [x] Move to proper, reversible AST

I did this largely as an exercise for myself, to partly address #30467 . ~~Unsure how much I can do on the rest but if I get pointers I may be able to. If this approach is deemed worth continuing.~~

```
julia> [1 2;3 4;;5 6;7 8;;9 0;2 3]
2×2×3 Array{Int64,3}:
[:, :, 1] =
 1  2
 3  4

[:, :, 2] =
 5  6
 7  8

[:, :, 3] =
 9  0
 2  3

julia> [1;3;;5;7;;9;2]
2×1×3 Array{Int64,3}:
[:, :, 1] =
 1
 3

[:, :, 2] =
 5
 7

[:, :, 3] =
 9
 2

julia> [1;;5;;9]
1×1×3 Array{Int64,3}:
[:, :, 1] =
 1

[:, :, 2] =
 5

[:, :, 3] =
 9
```

PR #37168, opened 2020-08-23
Add deprecation warning for multiple successive semicolons

BioTurboNick:
In preparation for future use of multiple semicolons in an array expression to indicate multi-dimensional concatenation (e.g. #33697), this PR implements a deprecation warning. This syntax is currently allowed but has no effect. Trailing semicolons are still ignored.

```
[1 ; 2] # allowed
[1 ;; 2] == [1 ; 2] # deprecated
```

But, no one would currently see it due to #35648, unless my suggestion #37164 is implemented.


src/julia-parser.scm
```diff
@@ -1831,7 +1831,15 @@
                          (fix 'vect vec)     ; [x]   => (vect x)
                          (fix 'hcat vec))))  ; [x y] => (hcat x y)
           (case t
-            ((#\; #\newline)
+            ((#\;)
+             (take-token s)
+             (if (eqv? (peek-token s) #\;)
+               (parser-depwarn s (string "Multiple semicolons in an array concatenation expression currently have no effect, "
+                                  "but will map to higher-dimensional concatenation in a future version of Julia.")
```
JeffBezanson, line 1838:
```suggestion
                                  "but may have a new meaning in a future version of Julia.")
```

JeffBezanson:
Thanks, looks good to me. This should have a NEWS entry.

BioTurboNick:
@JeffBezanson just saw your suggestion and implemented it. Also added a draft NEWS entry.

BioTurboNick:
I rebased onto the more recent master. Was hoping it would get all tests to pass. I don't think the failures are related though?

PR #37173, opened 2020-08-24
Update variables-and-scoping.md

briochemc:
Replace `sum_to′` with `verbose_sum_to` because the `′` (`\prime`) character looks like the `'` (single quote) character in the deployed docs, which can be confusing.


StefanKarpinski:
Calling it verbose is a bit weird since it doesn't print anything, it just returns extra information. It's fairly common to use prime in Julia code like this (at least I do it quite often). Perhaps it would be sufficient to have a note explaining that it's a prime, not an apostrophe.

briochemc:
Sure, I agree, this was just a suggestion! Another suggestion would be to just rewrite the function, i.e., keep the same `sum_to` name. 

FWIW, I love using (and abusing) the Unicode characters, but I felt like the Julia documentation should reduce possible confusion as much as possible, at least until Documenter can enforce a default font (JuliaMono?) that displays the characters correctly.

FYI, I suggested this change because someone interpreted this `′` (`\prime`) as a `'` single quote and was asking about it on Slack. Also, in my browser (Firefox, on OSX), the Julia docs display the `′` exactly like it would display `'`, so they are indistinguishable by eye... Maybe they are displayed differently for other people?

mortenpi:
Having scanned through the text, I think just redefining the `sum_to` function would be fine.

Just my two cents though: I think it should be changed away from `\prime`. I don't see any justification of using `\prime` in this context, and I think it's better to avoid that character outside of very limited cases (e.g. specific math notation) because it is bound to cause confusion.

StefanKarpinski:
Maybe call it `sum_to_def` since it computes the sum up to `n` and whether `t` is defined?

PR #37185, opened 2020-08-25
[NFC] C code clearn up

yuyichao:
* Remove unused functions that are not obviously kept for debugging
* Make functions/variables not used in other files static
* Move non-static declarations of function and variables to headers
* Remove non-exported API and it's users from `julia.h` (moved to `julia_internal.h`)

  This removes `jl_vararg_kind_t` from public header. It's technically breaking.
  But since no one could actually use its value due to the function that produces this being not callable,
  this shouldn't break any actual/useful code.

-----

For reviewing, the main things to check are,

* The moving of vararg related functions from `julia.h` to `julia_internal.h`
* No actual code change otherwise
* No removal of `JL_DLLEXPORT` definition.
* No removal of `JL_DLLEXPORT` declaration and static/inline definition from `julia.h`

### DIFF
--- a/src/timing.h
+++ b/src/timing.h
@@ -3,6 +3,15 @@
 #ifndef JL_TIMING_H
 #define JL_TIMING_H
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+void jl_init_timing(void);
+void jl_destroy_timing(void);
+#ifdef __cplusplus
+}
+#endif
+
 #ifndef ENABLE_TIMINGS
 #define JL_TIMING(owner)
 #else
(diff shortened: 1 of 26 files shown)


src/timing.h
```diff
@@ -5,6 +5,14 @@
 
 #ifndef ENABLE_TIMINGS
 #define JL_TIMING(owner)
+#ifdef __cplusplus
+extern "C" {
+#endif
+void jl_init_timing(void);
+void jl_destroy_timing(void);
```
JeffBezanson, line 12: These functions are now only declared when ENABLE_TIMINGS is not defined.

PR #37186, opened 2020-08-25
Let `LinearIndices(::Broadcasted)` return `LinearIndices`

martinholters:
Ref https://github.com/JuliaLang/julia/pull/31020/files#r475669025. Not that it has caused me any trouble, but a constructor that returns an instance of another type than requested does have the potential for nasty surprises. And here, I don't think there is any advantage in not returning a `LinearIndices` ... but I don't know my ways around the broadcast code, so I might be wrong. ~While at it, this generalizes the definition to arbitrary `Broadcasted`.~

### DIFF
--- a/base/broadcast.jl
+++ b/base/broadcast.jl
@@ -227,7 +227,7 @@ Base.IndexStyle(bc::Broadcasted) = IndexStyle(typeof(bc))
 Base.IndexStyle(::Type{<:Broadcasted{<:Any,<:Tuple{Any}}}) = IndexLinear()
 Base.IndexStyle(::Type{<:Broadcasted{<:Any}}) = IndexCartesian()
 
-Base.LinearIndices(bc::Broadcasted{<:Any,<:Tuple{Any}}) = axes(bc)[1]
+Base.LinearIndices(bc::Broadcasted{<:Any,<:Tuple{Any}}) = LinearIndices(axes(bc))::LinearIndices{1}
 
 Base.ndims(::Broadcasted{<:Any,<:NTuple{N,Any}}) where {N} = N
 Base.ndims(::Type{<:Broadcasted{<:Any,<:NTuple{N,Any}}}) where {N} = N


base/broadcast.jl
```diff
@@ -227,7 +227,7 @@ Base.IndexStyle(bc::Broadcasted) = IndexStyle(typeof(bc))
 Base.IndexStyle(::Type{<:Broadcasted{<:Any,<:Tuple{Any}}}) = IndexLinear()
 Base.IndexStyle(::Type{<:Broadcasted{<:Any}}) = IndexCartesian()
 
-Base.LinearIndices(bc::Broadcasted{<:Any,<:Tuple{Any}}) = axes(bc)[1]
```
tkf, line 230: Removing `{<:Any,<:Tuple{Any}}` makes me a bit nervous as I don't know all the details of `Broadcasted` indexing.  In particular, do we preserve the property that

```julia
for i in LinearIndices(bc)
    f(@inbouns bc[i])
end
```

never segfaults (although it may throw other recoverable errors) and `f` sees the same elements as

```julia
ys = copy(bc)
for i in LinearIndices(ys)
    f(@inbouns ys[i])
end
```

?

This example indicates that this is not the case:

```julia
julia> bc = Broadcast.instantiate(Broadcast.broadcasted(+, reshape(1:6, (2, 3)), 1:2))
Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{2}}(+, ([1 3 5; 2 4 6], 1:2))

julia> unsafe_getindex(x, I...) = @inbounds x[I...]
unsafe_getindex (generic function with 1 method)

julia> unsafe_getindex(bc, 3)  # should be the same as the following two
6

julia> unsafe_getindex(bc, 1, 2)
4

julia> copy(bc)[3]
4
```

tkf (commented):
Thanks!  I think making sure that `LinearIndices` returns `LinearIndices` is definitely an improvement.

But I think it'd be better to be conservative here and keep `{<:Any,<:Tuple{Any}}` because of the reason I mentioned.

base/broadcast.jl
martinholters: Ah, true, that's bad. I'll revert widening the signature. I guess the alternative would be to make sure `Broadcasted` supports proper linear indexing, but I'm not sure that is worth the effort and definitely beyond the scope of this PR.

mbauman: Yes, it was an intentional choice that multidimensional broadcasteds don't support linear indices; they're not arrays and it felt strange to intentionally support a tarpit if it wasn't needed.

mbauman (approved):

PR #37188, opened 2020-08-25
Fix `builtin_tfunction` for `Core._expr`

martinholters:
This had no effect before, as `Expr` is not a `Builtin`, but `Core._expr` is.

Before:
```julia
julia> code_typed(Expr)
1-element Vector{Any}:
 CodeInfo(
1 ─ %1 = Core._apply_iterate(Core.iterate, Core._expr, args)::Any
└──      return %1
) => Any
```
After:
```julia
julia> code_typed(Expr)
1-element Vector{Any}:
 CodeInfo(
1 ─ %1 = Core._apply_iterate(Core.iterate, Core._expr, args)::Expr
└──      return %1
) => Expr
```

### DIFF
--- a/base/compiler/tfuncs.jl
+++ b/base/compiler/tfuncs.jl
@@ -1412,7 +1412,7 @@ function builtin_tfunction(interp::AbstractInterpreter, @nospecialize(f), argtyp
             end
         end
         return Any
-    elseif f === Expr
+    elseif f === Core._expr
         if length(argtypes) < 1 && !isva
             return Bottom
         end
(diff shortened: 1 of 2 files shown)


martinholters:
(Some interesting(?) archeology: I think this was missed in bec065f10879bbf67327fe2d06c7cc01e8c52b76 (#8712), so this is actually a fix to a regression ... between 0.3 and 0.4 :smile:)

JeffBezanson:
The `if isva; return Any` check in there is also pretty pessimistic, e.g.
```
julia> f(a...) = ===(a...)

julia> code_typed(f)
1-element Vector{Any}:
 CodeInfo(
1 ─ %1 = Core._apply_iterate(Base.iterate, Main.:(===), a)::Any
└──      return %1
) => Any
```

martinholters:
Yes, I had considered going further here, but this change was so obvious and seemed sufficiently beneficial considering `Expr(:head, args...)`-like constructs that I wanted to get it in before getting lost in some larger rework. But I now do have this on my mind...

PR #37191, opened 2020-08-25
RFC: Drop free_typevars check when simplifying in `jl_type_union`

### LINKED ISSUE #37180: type-intersection / union missed optimization
Seeing this surprised me a bit, since in that first Union in the out, `a === b`, so it's trivially degenerate:

```
julia> typeintersect(Tuple{typeof(\),Factorization{T},Union{Array{Complex{T},2}, Array{Complex{T},1}}} where T<:Union{Float32, Float64}, Tuple{typeof(\),LQ{TA,S} where S<:AbstractMatrix{TA},StridedVecOrMat{TB}} where TB where TA)
Tuple{typeof(\),Union{LQ{T,S} where S<:AbstractMatrix{T}, LQ{T,S} where S<:AbstractMatrix{T}},Union{Array{Complex{T},2}, Array{Complex{T},1}}} where T<:Union{Float32, Float64}

julia> ans.body.parameters[2].a === ans.body.parameters[2].b
true
```

martinholters:
Types with free typevars make me a but uncomfortable, but here, I think we can drop the check. The individual union elements would be enclosed in a single `UnionAll` afterwards anyway. At least, this change doesn't make everything explode immediately and it fixes #37180. It may well be that I missed something and this is invalid (hence RFC), but in that case, it would be nice to add a counter example to the subtype tests.

### DIFF
--- a/src/jltypes.c
+++ b/src/jltypes.c
@@ -494,13 +494,11 @@ JL_DLLEXPORT jl_value_t *jl_type_union(jl_value_t **ts, size_t n)
     assert(count == nt);
     size_t j;
     for(i=0; i < nt; i++) {
-        int has_free = temp[i]!=NULL && jl_has_free_typevars(temp[i]);
         for(j=0; j < nt; j++) {
             if (j != i && temp[i] && temp[j]) {
                 if (temp[i] == temp[j] || temp[i] == jl_bottom_type ||
                     temp[j] == (jl_value_t*)jl_any_type ||
-                    (!has_free && !jl_has_free_typevars(temp[j]) &&
-                     jl_subtype(temp[i], temp[j]))) {
+                    jl_subtype(temp[i], temp[j])) {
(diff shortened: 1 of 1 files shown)


vtjnash:
subtyping just isn't precise in the presence of malformed types with missing UnionAll bounds. We can add `jl_egal` to the checks though.

martinholters:
Do you have a counter-example where this fails? The failing tests are not helpful here. The failed core test is due to
```julia
julia> Union{Integer,T} where T <: Unsigned
Union{Integer, T} where T<:Unsigned # master
Integer # this PR
```
Given that `Integer == Union{Integer,T} where T <: Unsigned`, that looks ok. The failed docs test looks more like a real regression, but is too indirect for me to get the relation to this change.

Or this more a "may work now, but we give no guarantees, so don't do it"-thing?

Issue #37135, opened 2020-08-20
Optimizer fails to DCE typeassert with ::Type{<:Union} (string performance)


timholy:
In eliminating invalidations from `AbstractString` methods I added type-asserts like [these](https://github.com/JuliaLang/julia/blob/ec63d88148a6215cd26912cd11f402532db0194c/base/strings/substring.jl#L65). I assumed that these would be elided in cases where the type was known concretely, but that assumption appears to be wrong. Here's a demo:

```julia
julia> code_typed(sizeof, (SubString{String},))
1-element Vector{Any}:
 CodeInfo(
1 ─ %1 = Base.getfield(s, :ncodeunits)::Int64
│        invoke Core.TypeVar(Symbol("#s68")::Symbol, Union{UInt16, UInt32, UInt8}::Any)::TypeVar
│        invoke Core.TypeVar(Symbol("#s68")::Symbol, Union{UInt16, UInt32, UInt8}::Any)::Core.Compiler.PartialTypeVar(var"#s68"<:Union{UInt16, UInt32, UInt8}, true, true)
│   %4 = Base.mul_int(%1, 1)::Int64
└──      return %4
) => Int64
```
You can see that statements 2 and 3 aren't used by any subsequent code, yet they are still there. This is further confirmed by 

```julia
julia> code_llvm(sizeof, (SubString{String},))

;  @ strings/basic.jl:177 within `sizeof'
define i64 @julia_sizeof_3256({ {}*, i64, i64 }* nocapture nonnull readonly align 8 dereferenceable(24)) {
top:
  %1 = alloca [2 x {}*], align 8
  %.sub = getelementptr inbounds [2 x {}*], [2 x {}*]* %1, i64 0, i64 0
; ┌ @ strings/substring.jl:64 within `ncodeunits'
; │┌ @ Base.jl:33 within `getproperty'
    %2 = getelementptr inbounds { {}*, i64, i64 }, { {}*, i64, i64 }* %0, i64 0, i32 2
; └└
; ┌ @ strings/substring.jl:65 within `codeunit'
   store {}* inttoptr (i64 140505618344944 to {}*), {}** %.sub, align 8
   %3 = getelementptr inbounds [2 x {}*], [2 x {}*]* %1, i64 0, i64 1
   store {}* inttoptr (i64 140505642590608 to {}*), {}** %3, align 8
[...]

PR #37192, opened 2020-08-25
Use a codeunit type-assert that is easier for inference to elide

timholy:
Circumvents #37135

### DIFF
--- a/base/regex.jl
+++ b/base/regex.jl
@@ -428,7 +428,7 @@ struct SubstitutionString{T<:AbstractString} <: AbstractString
 end
 
 ncodeunits(s::SubstitutionString) = ncodeunits(s.string)::Int
-codeunit(s::SubstitutionString) = codeunit(s.string)::Type{<:Union{UInt8, UInt16, UInt32}}
+codeunit(s::SubstitutionString) = codeunit(s.string)::CodeunitType
 codeunit(s::SubstitutionString, i::Integer) = codeunit(s.string, i)::Union{UInt8, UInt16, UInt32}
 isvalid(s::SubstitutionString, i::Integer) = isvalid(s.string, i)::Bool
 iterate(s::SubstitutionString, i::Integer...) = iterate(s.string, i...)::Union{Nothing,Tuple{AbstractChar,Int}}
--- a/base/strings/basic.jl
+++ b/base/strings/basic.jl
@@ -81,6 +81,8 @@ See also: [`ncodeunits`](@ref)
 """
 codeunit(s::AbstractString)
 
+const CodeunitType = Union{Type{UInt8},Type{UInt16},Type{UInt32}}
+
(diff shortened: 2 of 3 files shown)


JeffBezanson (approved):

omus:
I was working on a performance improvement for `==(::AbstractString, ::AbstractString)` but I discovered this PR addresses the performance issue I was noticing (mostly allocations):

Before this PR (6aee988223):

```julia
julia> str = "A" ^ 1000;

julia> @btime invoke(==, Tuple{AbstractString, AbstractString}, $"ABC", $"ABC")
  4.357 ns (0 allocations: 0 bytes)
true

julia> @btime $(SubString("ABC")) == $"ABCD"
  29.574 ns (2 allocations: 64 bytes)
false

julia> @btime $(SubString("ABC")) == $"ABC"
  32.492 ns (2 allocations: 64 bytes)
true

julia> @btime $(SubString("ABC")) == $"DEF"
  31.985 ns (2 allocations: 64 bytes)
false

julia> @btime $(SubString(str)) == $str
  65.086 ns (2 allocations: 64 bytes)
true
```

With this PR (dbf5e005d9):

```julia
julia> str = "A" ^ 1000;

julia> @btime invoke(==, Tuple{AbstractString, AbstractString}, $"ABC", $"ABC")
  3.907 ns (0 allocations: 0 bytes)
true

julia> @btime $(SubString("ABC")) == $"ABCD"
  0.033 ns (0 allocations: 0 bytes)
false

julia> @btime $(SubString("ABC")) == $"ABC"
  4.003 ns (0 allocations: 0 bytes)
true

julia> @btime $(SubString("ABC")) == $"DEF"
  3.282 ns (0 allocations: 0 bytes)
false

julia> @btime $(SubString(str)) == $str
  38.978 ns (0 allocations: 0 bytes)
true
```

I'll be adding some benchmarks for this.

KristofferC:
Is this v1.5? Because AFAIU this just fixed an earlier regression on nightly. But good to have benchmarks anyway.

omus:
I noticed similar slow performance on Julia 1.3.1, 1.4.2, and 1.5.1. All of these versions were substantially worse with the final test:
```julia
julia> @btime $(SubString(str)) == $str
  4.043 μs (2 allocations: 112 bytes)
true
```

KristofferC:
Seems more likely that the improvement here was because of https://github.com/JuliaLang/julia/pull/35973

omus:
You are correct that https://github.com/JuliaLang/julia/pull/35973 is the main driver behind the performance improvement. I'll update my benchmarks to reflect this.

omus:
I've also opened https://github.com/JuliaLang/julia/pull/37467 which improves performance of the fallback `==(::AbstractString, ::AbstractString)` call.

PR #37193, opened 2020-08-25
Add a generic code-inferrability test

### LINKED ISSUE #36393: Julep: attempting to measure (& test?) "absolute" risk of invalidation

timholy:
This PR adds an overall "code quality assurance" test, checking elements of inferrability of Base and the stdlibs. It largely tests just the precompiled MethodInstances, so does not offer comprehensive coverage, but these are connected to invalidations so this offers a form of protection against regression of our recent progress.

There are several elements of the test:

- it checks whether `Base.require` is "bullet-proof" against invalidation. Anytime `Base.require` gets invalidated, it's more than half a second before you can start loading the next package in a sequence, so invalidations here are a big problem from the standpoint of package-loading performance. Currently this is marked `@test_broken`, because we still have some string-processing code with inference problems that could invalidate these methods, but a goal will be to turn these into `@test`s.
- it tests the overall number of "worrisome" MethodInstances (`length(badcounts)` in the test code), as well as the average number of total dependents (backedges, and the backedges of their backedges, etc) of these methods (`meancounts` in the test code). Currently this PR sets the bar at 1250 worrisome MethodInstances and 33 average dependents. If/when my current unmerged PRs make it in, these can drop to 1000 and 32, respectively. For reference, Julia 1.5 has 2030 such MethodInstances with an average of 38.5 dependents. Most of what's left to fix appears to be in the area of string processing, so with a bit more effort we may be able to drop these numbers quite dramatically.
- it checks inferrability of specific functions (mostly the `is*` functions which should generally return `Bool`), and whether specific unfortunate variants of certain other methods (`==`, `<`, etc) have been inferred. The absence of such MethodInstances indicates that all code that calls them is well-inferred.

It's worth acknowledging that this may put a new burden on contributors: not only do they need to contribute the functionality, but they need to make sure it's not dramatically worsening inference. I think this is worth having. After all, we check whitespace, and whitespace is probably not nearly as important to the overall user experience as having well-inferred, invalidation-resistant code. Nevertheless, to mitigate the extra hurdles I've tried to put some hints near places where tests might fail to help developers find resources to help them.

Closes #36393

### DIFF
--- a/base/reflection.jl
+++ b/base/reflection.jl
@@ -753,6 +753,13 @@ function signature_type(@nospecialize(f), @nospecialize(args))
     end
 end
 
+function call_type(@nospecialize(tt))
+    ft = tt.parameters[1]
+    argt = Tuple{tt.parameters[2:end]...}
+    name = Symbol(String(ft.name.name)[2:end])  # strip off leading '#'
+    return (getfield(ft.name.module, name), argt)
+end
+
 """
     code_lowered(f, types; generated=true, debuginfo=:default)
 
@@ -926,6 +933,14 @@ function visit(f, d::Core.TypeMapEntry)
     while d !== nothing
         f(d.func)
         d = d.next
+    end
+    nothing
+end
+function visit(f, d::SimpleVector)
+    for i = 1:length(d)
+        if isassigned(d, i)
+            f(d[i])
+        end
     end
     nothing
 end
--- a/test/choosetests.jl
+++ b/test/choosetests.jl
@@ -55,7 +55,7 @@ function choosetests(choices = [])
         "boundscheck", "error", "ambiguous", "cartesian", "osutils",
         "channels", "iostream", "secretbuffer", "specificity",
         "reinterpretarray", "syntax", "corelogging", "missing", "asyncmap",
-        "smallarrayshrink"
+        "smallarrayshrink", "inference_qa"
     ]
 
     tests = []
--- a/test/inference_qa.jl
+++ b/test/inference_qa.jl
@@ -1,0 +1,216 @@
+# This file is a part of Julia. License is MIT: https://julialang.org/license
+
+# The aim of these tests is to enforce a minimum quality of inferrability of Julia's
+# Base and specifically its precompiled methods. Passing these tests does not
+# indicate that Julia has no inference problems, but they are designed to catch some
+# inadvertent problems. While `@inferred` only tests the return type, the tests
+# in this file are designed to check the overall inference quality of method *internals*.
+
+# If you fail tests here, you can usually fix problems using `@code_warntype` or Cthulhu.jl's `@descend`.
+
+using Test
+
+isdefined(@__MODULE__, :is_atrisk_type) || include("testhelpers/methodanalysis.jl")
+
+## Test the testing tools
+
+@testset "is_atrisk_type" begin
+    @test  is_atrisk_type(Tuple{typeof(==),Any,Any})
+    @test  is_atrisk_type(Tuple{typeof(==),Symbol,Any})
+    @test  is_atrisk_type(Tuple{typeof(==),Any,Symbol})
+    @test !is_atrisk_type(Tuple{typeof(==),Symbol,Symbol})
+    @test !is_atrisk_type(Tuple{typeof(convert),Type{Any},Any})
+    @test !is_atrisk_type(Tuple{typeof(convert),Type{AbstractString},AbstractString})
+    @test !is_atrisk_type(Tuple{typeof(convert),Type{AbstractString},String})
+    @test  is_atrisk_type(Tuple{typeof(convert),Type{String},AbstractString})
+    @test !is_atrisk_type(Tuple{typeof(convert),Type{Union{Int,Float32}},Int})
+    @test !is_atrisk_type(Tuple{typeof(convert),Type{Union{Int,Float32}},Int32})
+    @test  is_atrisk_type(Tuple{typeof(convert),Type{Union{Int,Float32}},Integer})
+    @test !is_atrisk_type(Tuple{typeof(convert),Type{T} where T<:Union{Int,Float32},Int})
+    @test !is_atrisk_type(Tuple{typeof(map),Function,Vector{Any}})
+    @test !is_atrisk_type(Tuple{typeof(getindex),Dict{Union{String,Int},Any},Union{String,Int}})
+    @test  is_atrisk_type(Tuple{typeof(getindex),Dict{Union{String,Int},Any},Any})
+    @test !is_atrisk_type(Tuple{Type{BoundsError},Any,Any})
+    @test  is_atrisk_type(Tuple{typeof(sin),Any})
+    @test !is_atrisk_type(Tuple{typeof(length),Tuple{Any,Any}})
+    @test  is_atrisk_type(Tuple{typeof(setindex!),Vector{Int},Any,Int})
+    @test !is_atrisk_type(Tuple{typeof(setindex!),Vector{Any},Any,Int})
+    @test  is_atrisk_type(Tuple{typeof(push!),Vector{Int},Any})
+    @test !is_atrisk_type(Tuple{typeof(push!),Vector{Any},Any})
+    @test !is_atrisk_type(Tuple{typeof(push!),Vector{T} where T,Any})
+end
+
+## Prepwork
+
+# Count # of backedges for MethodInstances with abstract types
+
+function get_atrisk_methodinstances()
+    mivulnerabilities = Pair{MethodInstance,Int}[]
+    for mi in methodinstances()
+        if !fromcc(mi.def.module)
+            if is_atrisk_type(mi.specTypes)
+                push!(mivulnerabilities, mi => length(all_backedges(mi)))
+            end
+        end
+    end
+    return sort!(mivulnerabilities; by=last)
+end
+mivulnerabilities = get_atrisk_methodinstances()
+
+# Get all the things that depends on these at-risk methodinstances
+const atrisk_methodinstances = Set{MethodInstance}()
+for (mi, c) in mivulnerabilities
+    isdefined(mi, :backedges) && all_backedges!(atrisk_methodinstances, mi)
+    push!(atrisk_methodinstances, mi)
+end
+
+# Split into exported & private functions
+mivulnerabilities_exported = Pair{MethodInstance,Int}[]
+mivulnerabilities_private = similar(mivulnerabilities_exported)
+for (mi, c) in mivulnerabilities
+    if isexported(mi)
+        push!(mivulnerabilities_exported, mi=>c)
+    else
+        push!(mivulnerabilities_private, mi=>c)
+    end
+end
+
+## Tests
+
+@testset "Base.require vulnerabilities" begin
+    # Invalidating the code that loads packages leads to major slowdowns, especially if it happens repeatedly
+    # in a dependent chain of package loads. Ideally, we'd make this code-path "bulletproof".
+    for m in methods(Base.require)
+        @test isempty(remove_unlikely_methodinstances(atrisk_triggers(m, first.(mivulnerabilities_exported))))
+
+        # It's far less important to protect against invalidation of private functions,
+        # since generally packages should not extend these functions. Nevertheless it wouldn't
+        # be a bad thing.
+        @test_broken isempty(remove_unlikely_methodinstances(atrisk_triggers(m, first.(mivulnerabilities_private))))
+    end
+end
+
+# If you fail these tests, it may or may not be your fault---you may just be the
+# one that pushed one of these tests over the edge. Check `badcounts` and `meancounts`
+# before and after your changes; if the change is quite small, it's unlikely that your changes
+# are the root cause. In such cases, failures here should be investigated or reported,
+# but if inference problems can be fixed elsewhere you may not have to change your PR.
+# Conversely, if your changes *do* increase these numbers substantially, your changes have likely
+# triggered widespread inference problems---you should fix them before merging your PR.
+#
+# Never increase the thresholds here without public discussion. Indeed, the goal is to
+# shrink them as much as possible.
+@testset "Aggregate at-risk methodinstances" begin
+    # Test overall number of atrisk MethodInstances and their average number of backedges
+    badexp = Set(remove_unlikely_methodinstances(first.(mivulnerabilities_exported)))
+    badcounts = filter(pr->pr.first ∈ badexp, mivulnerabilities_exported)
+    @test length(badcounts) < 1250 # 1000
+    if length(badcounts) < 800
+        @info "There are now only $(length(badcounts)) at-risk specializations of exported methods, consider dropping the threshold"
+    end
+    meancounts = sum(last.(badcounts))/length(badcounts)
+    @test meancounts < 33 # 32
+    if meancounts < 24
+        @info "The mean number of at-risk backedges is now only $meancounts, consider dropping the threshold"
+    end
+end
+
+@testset "Specific return types" begin
+    # All the is* functions
+    # Not all of the broken cases have been checked carefully; it's possible some of these should return
+    # `Union{Bool,Missing}` or something.
+    @test        function_returns(isabspath, Bool)
+    @test        function_returns(isabstracttype, Bool)
+    @test_broken function_returns(isapprox, Bool)
+    @test_broken function_returns(isascii, Bool)
+    # @test function_returns(isassigned, Bool)
+    @test        function_returns(isbits, Bool)
+    @test        function_returns(isbitstype, Bool)
+    @test        function_returns(isblockdev, Bool)
+    @test        function_returns(ischardev, Bool)
+    @test        function_returns(iscntrl, Bool)
+    @test        function_returns(isconcretetype, Bool)
+    @test        function_returns(isconst, Bool)
+    @test        function_returns(isdefined, Bool)
+    @test        function_returns(isdigit, Bool)
+    @test        function_returns(isdir, Bool)
+    @test        function_returns(isdirpath, Bool)
+    @test_broken function_returns(isdisjoint, Bool)
+    @test        function_returns(isdispatchtuple, Bool)
+    @test_broken function_returns(isempty, Bool)
+    @test_broken function_returns(isequal, Bool; minargs=2)
+    @test_broken function_returns(iseven, Bool)
+    @test        function_returns(isexported, Bool)
+    @test        function_returns(isfifo, Bool)
+    @test        function_returns(isfile, Bool)
+    @test_broken function_returns(isfinite, Bool)
+    @test_broken function_returns(isinf, Bool)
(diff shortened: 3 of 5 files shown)


KristofferC:
> It's worth acknowledging that this may put a new burden on contributors: not only do they need to contribute the functionality, but they need to make sure it's not dramatically worsening inference.

Because of this, I am not sure this is the best way to do this. And with "this way" I mean, setting an arbitrary cap and then hard erroring CI when that is reached. This would be a bit similar to setting a hard cap on sysimage build time and failing CI when that is reached (assuming that it is completely deterministic). When that point is reached, it is likely a lot of CI will start failing and the whole thing grinds to a halt until someone with the skill goes in and pushes to number of invalidations down again.

Instead, (just like compilation time), I feel like this should be tracked via some time-series functionality where it can be monitored and when some threshold is reached it might start blinking red or something but hard failing CI because someone does a small fix that pushes the number of backedges for 2312 to 2319 seems too harsh.

> After all, we check whitespace, and whitespace is probably not nearly as important to the overall user experience as having well-inferred, invalidation-resistant code. 

While true, the difficulty of fixing the issue also needs to be weighed in which in one case is trivial and in the other very advanced.

timholy:
Excellent points as usual. Maybe a separate buildbot pass then? I'm clueless about how one goes about implementing this.

KristofferC:
The CUDA people have a nice benchmark site based on codespeed that shows timelines for various benchmarks (https://speed.juliagpu.org/timeline/#/?exe=5,3,4&base=3+74&ben=grid&env=1&revs=50&equid=off&quarts=on&extr=on).

It would be nice to hookup the daily Nanosoldier benchmarks to an instance of that and at the same time we could maybe add some compilation time benchmarks (https://github.com/JuliaCI/BaseBenchmarks.jl/pull/252) but also some "static" info like number of backedges etc so that could also be tracked over time.

timholy:
That makes sense, but I worry about how many people actually look at those numbers. I don't check Julia's performance routinely. OTOH, as long as it's checked before the release perhaps that would be enough.

OTOH, if we had a separate build pass, there's no way you'd miss a change although one could decide to go ahead and merge anyway. It does make sense to separate it out from "standard" tests of course.

KristofferC:
>  I don't check Julia's performance routinely. OTOH, as long as it's checked before the release perhaps that would be enough.

Maybe that is because we have no way of nicely visualizing it right now. 

> OTOH, if we had a separate build pass, there's no way you'd miss a change although one could decide to go ahead and merge anyway. It does make sense to separate it out from "standard" tests of course.

Well, that build pass is going to be green and ignored until it turns red and then it will still mostly be ignored. And when someone (you?) see it, then what do you do? You probably want to try to find the place where bad stuff happened but since you don't really have a pass / fail criteria the best thing is probably at that point to start building a time series to see if there are points where things suddenly got worse.

timholy:
Fair enough. I'm not sure how one gets started on that though.

KristofferC:
Hm, @staticfloat how long do we save the logs from the buildbot? If we add something that prints to the build log, would it be easy to build up a timeseries from that?

For now, perhaps we can:
- Keep the test error for the "non-fuzzy" things (like making sure `Base.require` is solid). This is a yes or no thing and a PR that breaks it can 
- Change the error for the fuzzy tests to warnings or something along those lines.

That means we can still merge the PR and the code gets regularly run so it doesn't bit rot and then we can think a bit more about how exactly the "holistic" measurement should be presented / checked?

martinholters:
Could we do a before/after comparison like is done e.g. for code coverage? "This PR increases invalidation susceptibility by 2.3%." or whatever.

staticfloat:
> Hm, @staticfloat how long do we save the logs from the buildbot? If we add something that prints to the build log, would it be easy to build up a timeseries from that?

The buildbots [clean things up after 120 days](https://github.com/JuliaCI/julia-buildbot/blob/master/master/master.cfg#L32-L33).  You can do this if you want to, but note that `buildbot` is far from a performant database.  Expect to wait like 0.5s for each datapoint you want to read.

You are the second or third person in the last month who has wanted to store timeseries data for packages.  I've been thinking about hosting a timeseries database like influxdb or timescaledb for the (authenticated) Julia community.  Combined with a small wrapper over HTTP (like [this abandoned package](https://github.com/staticfloat/InfluxDB.jl)), it would allow us to `POST` timeseries datapoints and then query timeseries back out with `HTTP` or even fancy things like Grafana.  Interested, or overkill?

timholy:
I do like the idea of a time series, but I think the more important one is to be notified when PRs add a lot of poorly-inferred calls. What I had in mind was best summarized by @martinholters (https://github.com/JuliaLang/julia/pull/37193#issuecomment-682444038).

timholy:
xref https://github.com/julia-actions/julia-invalidations, thanks @ianshmean!

timholy:
Is there any hope of doing something like this for Julia 1.6? We don't seem to have regressed, except there seem to be more ways to invalidate `Base.require` than I remember previously. At least one chunk of problems come from the preferences work (e.g., #38044, CC @staticfloat).

KristofferC:
@timholy could you check if https://github.com/JuliaLang/julia/tree/kc/stability_preferences helps with that?

KristofferC:
> Is there any hope of doing something like this for Julia 1.6?

I'm personally ok with putting in whatever you think would be helpful. I still worry a bit about a potential future PR "deadlock" when this starts to erroring and new contributors get confused about what they did wrong. But we can always disable the tests if they become disruptive so 🤷

timholy:
How about we aim to eliminate vulnerabilities in paths leading to `Base.require` and make that a hard-fail, and use warnings for the rest?

timholy:
OK, this seems to be working (test failure seems incidental). Ready for review.

I am not terribly attached to the `is*` tests, so we could delete those depending on how others feel about them. For quite a few of them there are more than 3 methods, so in cases of poor inference their caller is going to have to annotate at the call site anyway. This may still have some value since it can be nice to check what all methods do return, and if they're all `Bool` (as opposed to, say, `Union{Bool,Missing}`) then one can feel more confident about asserting that.

vchuravy:
I am seeing tests added in this PR failing in #38437

KristofferC:
Did CI run last time 7 days ago? Maybe it had time to regress in that time?

timholy:
Hmm, I should have re-run CI, although it seems broken right now.

Indeed, something has changed it. Specifically, the test that I think is the most important, the impossibility of invalidating `Base.require` without committing piracy, is now failing.
```julia
julia> mis = remove_unlikely_methodinstances(atrisk_triggers(m, first.(mivulnerabilities_exported)))
2-element Vector{MethodInstance}:
 MethodInstance for convert(::Type{Nothing}, ::Any)
 MethodInstance for convert(::Type{Union{}}, ::Any)

julia> using Cthulhu

julia> ascend(mis[1])
Choose a call for analysis (q to quit):
     convert(::Type{Nothing}, ::Any)
       setindex!(::IdDict{Dict{String, Any}, Nothing}, ::Any, ::Any)
   +     push!(::IdSet{Dict{String, Any}}, ::Dict{String, Any})
   +   setindex!(::IdDict{Any, Nothing}, ::Any, ::Any)
       setindex!(::IdDict{MethodInstance, Nothing}, ::Any, ::Any)
   +     push!(::IdSet{MethodInstance}, ::MethodInstance)
         push!(::IdSet{MethodInstance}, ::Any)
 >         treelist!(::FoldingTrees.Node{Cthulhu.Data{MethodInstance}}, ::IOBuffer, ::Any, ::String, ::IdSet{MethodInstance})
```

I'm showing the tree in collapsed form, folding everything below a "reasonable" call. It's basically all IdDict/IdSet-related. The offending `convert` call is https://github.com/JuliaLang/julia/blob/110f1256ac88ee0f12885289e807229acf9c44a3/base/iddict.jl#L74, which has been with us at least since 8b6a8c2bb463724b8cacf422acf330e9aebeab0d (February), so it's hard to believe that this directly changed. Seems like it might be something else, like perhaps an improvement elsewhere?

There's a fix in #38455.

PR #26980, opened 2018-05-04
Deprecate map(f, ::AbstractSet)

mbauman:
`map` on sets previously returned a `Set`, possibly changing the order or number of elements. This
behavior is deprecated and in the future `map` will preserve order and number of elements.

Fixes #26359.

PR #37195, opened 2020-08-25
Add `filter(f, itr) = collect(Iterators.filter(f, itr))`

mcabbott:
Closes #37146

### DIFF
--- a/base/abstractarray.jl
+++ b/base/abstractarray.jl
@@ -2317,6 +2317,9 @@ end
 map(f) = f()
 map(f, iters...) = collect(Generator(f, iters...))
 
+# since map(f, itr) is defined here:
+filter(f, itr) = collect(Iterators.filter(f, itr))
+
 # multi-item push!, pushfirst! (built on top of type-specific 1-item version)
 # (note: must not cause a dispatch loop when 1-item case is not defined)
 push!(A, a, b) = push!(push!(A, a), b)
--- a/test/abstractarray.jl
+++ b/test/abstractarray.jl
@@ -646,6 +646,11 @@ end
     @test map(()->1234) === 1234
 end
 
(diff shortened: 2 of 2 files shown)


bramtayl:
Unrelated but maybe we could change the map method to

`map(f, iters...) = collect(Iterators.map(f, iters...))`

Another thing worth thinking about eventually is whether we need auto-collecting versions of iterators in the first place?

mcabbott:
Aren't those identical?
```
julia> Iterators.map(sqrt, 1:10) === Base.Generator(sqrt, 1:10)
true
```

bramtayl:
Yes, but I think `Iterators.map` is intended to be a public and overloadable interface, whereas `Base.Generator` is an internal, I think.

mcabbott:
Bump? 

cc. @JeffBezanson since he said :+1: earlier...

mcabbott:
October bump?

Gyoshi:
`map` has been explicitly deprecated for Sets and Dicts (#26980):
```julia
map(f, ::AbstractDict) = error("map is not defined on dictionaries")
map(f, ::AbstractSet) = error("map is not defined on sets")
```
Maybe this should also hold for `filter`?

JeffBezanson:
No; `map` and `filter` are different in important ways. `map` changes values, so it's less clear what it should do for structures that impose invariants on the values they hold (e.g. that they are unique). `filter` just removes some elements.

I'm not sure we should add this. I think the fallback `map` definition is kind of a misfeature. We shouldn't encourage collecting all elements into an array until it's really needed. Anyway there's no conclusion from triage posted here, so let's wait for that.

JeffBezanson:
Nobody on triage this week seems to feel strongly positive about this.

PR #37197, opened 2020-08-25
Refine out callee rooted values from live set at the call site

yuyichao:
AFAICT the callee rooted address space is otherwise not used by any optimization...

0.9% reduction in sysimg code size. Mostly helps for the error throwing function and avoids some gc frame generation due to the error path.


yuyichao:
And this is the function change https://github.com/JuliaLang/julia/pull/37151 and https://github.com/JuliaLang/julia/pull/37155 were factored out from...

PR #37206, opened 2020-08-26
[mktempdir()]: reset permissions before attempting to delete

staticfloat:
Because users may do something weird like create directories without
write permissions for the owner, it is best to do everything we can to
delete the files created within `mktempdir()`.

Take two of this, only instead of making it the case for all `rm()`, only do it for `mktempdir()` as it's much more certain that users don't actually want those directories to persist.

### DIFF
--- a/base/file.jl
+++ b/base/file.jl
@@ -458,6 +458,32 @@ function tempdir()
     end
 end
 
+"""
+    prepare_for_deletion(path::AbstractString)
+
+Prepares the given `path` for deletion by ensuring that all directories within that
+`path` have write permissions, so that files can be removed from them.  This is
+automatically invoked by methods such as `mktempdir()` to ensure that no matter what
+weird permissions a user may have created directories with within the temporary prefix,
+it will always be deleted.
+"""
+function prepare_for_deletion(path::AbstractString)
+    # Nothing to do for non-directories
+    if !isdir(path)
+        return
+    end
+
+    try chmod(path, filemode(path) | 0o333)
+    catch; end
+    for (root, dirs, files) in walkdir(path)
+        for dir in dirs
+            dpath = joinpath(root, dir)
+            try chmod(dpath, filemode(dpath) | 0o333)
+            catch; end
+        end
+    end
+end
+
 const TEMP_CLEANUP_MIN = Ref(1024)
 const TEMP_CLEANUP_MAX = Ref(1024)
 const TEMP_CLEANUP = Dict{String,Bool}()
@@ -484,6 +510,7 @@ function temp_cleanup_purge(; force::Bool=false)
             if (force || asap) && ispath(path)
                 need_gc && GC.gc(true)
                 need_gc = false
+                prepare_for_deletion(path)
                 rm(path, recursive=true, force=true)
             end
             !ispath(path) && delete!(TEMP_CLEANUP, path)
@@ -682,7 +709,10 @@ function mktempdir(fn::Function, parent::AbstractString=tempdir();
         fn(tmpdir)
     finally
         try
-            ispath(tmpdir) && rm(tmpdir, recursive=true)
+            if ispath(tmpdir)
+                prepare_for_deletion(tmpdir)
+                rm(tmpdir, recursive=true)
+            end
         catch ex
             @error "mktempdir cleanup" _group=:file exception=(ex, catch_backtrace())
             # might be possible to remove later
--- a/test/file.jl
+++ b/test/file.jl
@@ -268,6 +268,20 @@ no_error_logging(f::Function) =
     end
 end
 
+@testset "mktempdir() permissions correction" begin
+    # Test that `mktempdir()` with children with weird permissions get cleared
+    # before we would delete the directory
+    local temp_dir_path
+    mktempdir() do dir
+        temp_dir_path = dir
+        mkdir(joinpath(dir, "foo"))
+        touch(joinpath(dir, "foo", "bar"))
+        chmod(joinpath(dir, "foo"), 0o444)
+        @test isdir(temp_dir_path)
+    end
+    @test !isdir(temp_dir_path)
+end
+
 #######################################################################
 # This section tests some of the features of the stat-based file info #
 #######################################################################


base/file.jl
```diff
@@ -484,6 +484,7 @@ function temp_cleanup_purge(; force::Bool=false)
             if (force || asap) && ispath(path)
                 need_gc && GC.gc(true)
                 need_gc = false
+                chmod(path, 0o777; recursive=true)
```
StefanKarpinski, line 487: I've been changing to `0o700` instead so that we don't accidentally expose these files to being read by an eavesdropper. I guess the issue with that is that the user might not own the files but have write permissions and take away their own write permissions by doing this. But I think it the user doesn't own the files, that's a corner case where we can just let this fail.

StefanKarpinski: The fully correct way to do this would be to add an option to `rm` that lets you try harder and changes the permissions precisely as necessary to allow deletion. So that would entail checking who owns the file and and picking the permissions based on that.

staticfloat: What about just setting all directories to `0o770`?  I was being a little lazy by using `chmod(; recursive=true)` here, we really only care about the directories.

StefanKarpinski: Making something group-readable is not really safer than world. What about `0o722`? That makes it writeable by anyone, which is normally a security hole but in this case since we're about to delete it, it should be safe.

staticfloat: Okay the actual change I made was to set it to `filemode(path) | 0o220`.  That seems the best option.

StefanKarpinski: I don't think there's a meaningful distinction between group and world on most systems: the default group is usually the group that everyone belongs to (except for root). So what about `0o222`? Giving someone else write permissions is acceptable since we're about to delete the file and so writing to it can't cause trouble.

base/file.jl
```diff
@@ -458,6 +458,32 @@ function tempdir()
...
+it will always be deleted.
+"""
+function prepare_for_deletion(path::AbstractString)
+    # Nothing to do for non-directories
+    if !isdir(path)
+        return
+    end
+
+    try chmod(path, filemode(path) | 0o333)
+    catch; end
```
vtjnash, line 477:
```suggestion
    try
        chmod(path, filemode(path) | 0o333)
    catch ex
        ex isa IOError || rethrow()
    end
```

The `catch` symbol is required, because you're required to do something with it.

base/file.jl
```diff
@@ -458,6 +458,32 @@ function tempdir()
...
+    for (root, dirs, files) in walkdir(path)
+        for dir in dirs
+            dpath = joinpath(root, dir)
+            try chmod(dpath, filemode(dpath) | 0o333)
```
vtjnash, line 481: Wouldn't it be much safer only to add permissions to the user? As it is written, it seems that this could sometimes be tricked into opening up write access to a system file and thus create a security vulnerability.
```suggestion
    chmod(dpath, filemode(dpath) | 0o300)
```

base/file.jl
```diff
@@ -458,6 +458,32 @@ function tempdir()
     end
 end
 
+"""
+    prepare_for_deletion(path::AbstractString)
+
+Prepares the given `path` for deletion by ensuring that all directories within that
+`path` have write permissions, so that files can be removed from them.  This is
```
vtjnash, line 465: write _and traversal_ permissions

Do we not need read permissions?

base/file.jl
```diff
@@ -682,7 +709,10 @@ function mktempdir(fn::Function, parent::AbstractString=tempdir();
         fn(tmpdir)
     finally
         try
-            ispath(tmpdir) && rm(tmpdir, recursive=true)
+            if ispath(tmpdir)
+                prepare_for_deletion(tmpdir)
+                rm(tmpdir, recursive=true)
```
vtjnash, line 714: Is there a reason we weren't passing `force=true` here, which already includes some logic to deal with this failure case? (and possibly extend that logic to cover more cases)

base/file.jl
StefanKarpinski: Agree. I believe I made a similar comment on some version of this.

vtjnash: Yeah, this code (necessarily, intentionally) has a TOCTOU vulnerability risk, so I think we need to be careful here with your earlier comments (which I didn't read until you mentioned it). With adding world writable, I believe someone can readily convince this code to chmod the entire system to +wX (by taking advantage of the enormous delay between `readdir`, `stat`/`islink`/`isdir`, and the `chmod`/`rm` post-processing operations). While that doesn't directly lead to leaking secrets, it does mean the attacker can replace any file with a malicious one that then exfiltrates the data (thereby gaining the missing read permissions).

base/file.jl
staticfloat: I don't know why we weren't passing `force=true`.

base/file.jl
staticfloat: I think we've just never encountered a directory without read permissions in practice.  Sounds reasonable to me though.

StefanKarpinski:
I'm starting to think that we should, instead of this approach which seems fundamentally flawed, change `rm` with `force=true` to change the permissions if necessary right before deleting each item. That seems safer since there's a much smaller time between the permission change and the deletion. We can also probably be a little more specific about changing the permissions of various kinds of inodes.

vtjnash:
Note that in my recollection of the attacks that I've seen, the attackers have plenty of tricks to make those operations take arbitrarily long amounts of time. Some possible workarounds are only changing permissions for the owner (since only the owner and root can chmod anyways, this isn't a security hole), and working with fd instead of  path.

StefanKarpinski:
Ah, that's a good point. I was thinking that we might not be the owner but if we're not the owner then we can't chmod anyway.

staticfloat:
> I'm starting to think that we should, instead of this approach which seems fundamentally flawed, change rm with force=true to change the permissions if necessary right before deleting each item.

That was my first idea, it was shot down here: https://github.com/JuliaLang/julia/pull/37160

StefanKarpinski:
That comment is incorrect in light of @vtjnash's point: if the right bit is not `u` then we can't chmod anyway.

staticfloat:
I think Yichao's point still stands; if the user is actually using `0555` on their directories to prevent `rm -rf` from affecting them, we probably want to honor that.  That's why I changed it to only apply to `mktempdir()`, since it's much easier to argue that incorrect permissions are a mistake there.

StefanKarpinski:
It does seem like tarballs sometimes contain permissions that make them impossible to delete later. Fortunately, Tar.jl ignores those and uses the git-style permissions (i.e. 755 for directories, symlinks and executables; 644 for plain files, moderated by the umask), so that shouldn't be a problem for us anymore. I "fixed" the documentation of Tar.jl to document that we respect tarball permissions, but later realized that we don't, so I should unfix that.

Issue #9909, opened 2015-01-24
`sizehint!(::Dict, n)` doesn't grow Dicts enough


garborg:
It seems like `sizehint!(d, n)` should grow `d` so that it could hold `n` pairs before rehashing and `resize!`ing its fields again. Currently `d` is only ready to hold `floor(n * 2/3)` pairs after resizing.

Other methods in Base use it assuming the the former behavior, and allocate more than they could. Simple change, but first, is there any any reason not make it?

mlubin:
The current behavior seems bizarre.

ivarne:
The relevant line is in [/base/dict.jl#L482](https://github.com/JuliaLang/julia/blob/08663d4bb05c5b8805a57f46f4feacb07c7f2564/base/dict.jl#L482)

JeffBezanson:
Even more relevant: https://github.com/JuliaLang/julia/blob/08663d4bb05c5b8805a57f46f4feacb07c7f2564/base/dict.jl#L569

However, resizing behavior also depends on the number of collisions, so it is not really possible to guarantee holding `n` items without rehashing, unless we allow lookup performance to degrade all the way to `O(n)`.

garborg:
I was thinking switching `function sizehint!(d::Dict, newsz)` to 
```julia
function sizehint!(d::Dict, newsz)
    newsz = ceil(Int, 1.5 * sz)
```
would be good enough.
It improves performance on any sizes I've tested and brings use (if not guarantee) in line with other `sizehint!` methods, at the cost of surface inconsistency with `rehash!`, but everything that uses the unexported `rehash!` is already reaching into `Dict` implementation and knows what `length`s it changes, and the name change above seems to clear up the surface inconsistency.

JeffBezanson:
Ok, that's fine. I wonder if we can bring the max load factor up to 3/4 or 4/5 (and thus lower that constant to 1.33 or 1.25). It seems to me the collision chain length is much more important than the fraction of space used. After all an ideal table would be 100% full with no collisions!

kshyatt:
Is this still a problem?

garborg:
It looks both changes are still todos: making sizehint! aware of the load factor and either increasing the load factor or changing how to decide when to resize.

StefanKarpinski:
[...]

PR #37208, opened 2020-08-26
improve speed of allunique

bkamins:
I propose a bit faster implementation of `allunique`. Here is the benchmark on `x = collect(1:10^8)`.

Old:
```
julia> @time allunique(x)
 17.072063 seconds (82 allocations: 4.499 GiB, 0.65% gc time)
true
```

New:
```
julia> @time allunique2(x)
 15.096852 seconds (82 allocations: 4.499 GiB, 0.73% gc time)
true
```

We can make it even faster with this implementation:
```
function allunique(C)
    seen = Dict{eltype(C), Nothing}()
    for x in C
        idx = ht_keyindex2!(seen, x)
        idx > 0 && return false
        _setindex!(seen, nothing, x, -idx)
    end
    return true
end
```
here is a benchmark:
```
julia> @time allunique(x)
 14.542865 seconds (82 allocations: 4.499 GiB, 0.78% gc time)
true
```
but I was not sure if it is a best idea to add a dependency on low level functions like `ht_keyindex2!` and `_setindex!` in a high level function like `allunique`.

### DIFF
--- a/base/set.jl
+++ b/base/set.jl
@@ -383,6 +383,7 @@ false
 """
 function allunique(C)
     seen = Set{eltype(C)}()
+    haslength(C) && sizehint!(seen, length(C))
     for (i, x) in enumerate(C)
         push!(seen, x)
         i > length(seen) && return false
--- a/base/set.jl
+++ b/base/set.jl
@@ -397,7 +397,6 @@ function allunique(C)
         end
     end
 
-    x === nothing && return true
     haslength(C) && sizehint!(seen, length(C))
 
     while x !== nothing
@@ -406,7 +405,6 @@ function allunique(C)
         idx > 0 && return false
         _setindex!(seen, nothing, v, -idx)
         x = iterate(C, s)
-    end
     end
     return true
 end
--- a/base/set.jl
+++ b/base/set.jl
@@ -383,10 +383,30 @@ false
 """
 function allunique(C)
     seen = Set{eltype(C)}()
+
+    x = iterate(C)
+    for i in OneTo(1000)
+        if x === nothing
+            return true
+        else
+            v, s = x
+            idx = ht_keyindex2!(seen, v)
+            idx > 0 && return false
+            _setindex!(seen, nothing, v, -idx)
+            x = iterate(C, s)
+        end
+    end
+
+    x === nothing && return true
     haslength(C) && sizehint!(seen, length(C))
-    for (i, x) in enumerate(C)
-        push!(seen, x)
-        i > length(seen) && return false
+
+    while x !== nothing
+        v, s = x
+        idx = ht_keyindex2!(seen, v)
+        idx > 0 && return false
+        _setindex!(seen, nothing, v, -idx)
+        x = iterate(C, s)
+    end
     end
     return true
 end
--- a/base/set.jl
+++ b/base/set.jl
@@ -383,14 +383,11 @@ false
 """
 function allunique(C)
     seen = Set{eltype(C)}()
-    for x in C
-        if in(x, seen)
-            return false
-        else
-            push!(seen, x)
-        end
+    for (i, x) in enumerate(C)
+        push!(seen, x)
+        i > length(seen) && return false
     end
-    true
+    return true
 end
 
 allunique(::Union{AbstractSet,AbstractDict}) = true
--- a/base/set.jl
+++ b/base/set.jl
@@ -382,7 +382,7 @@ false
 ```
 """
 function allunique(C)
-    seen = Set{eltype(C)}()
+    seen = Dict{eltype(C),Nothing}()
 
     x = iterate(C)
     for i in OneTo(1000)
--- a/base/set.jl
+++ b/base/set.jl
@@ -382,23 +382,18 @@ false
 ```
 """
 function allunique(C)
-    seen = Dict{eltype(C),Nothing}()
-
+    seen = Dict{eltype(C), Nothing}()
     x = iterate(C)
-    for i in OneTo(1000)
-        if x === nothing
-            return true
-        else
+    if haslength(C) && length(C) > 1000
+        for i in OneTo(1000)
             v, s = x
             idx = ht_keyindex2!(seen, v)
             idx > 0 && return false
             _setindex!(seen, nothing, v, -idx)
             x = iterate(C, s)
         end
+        sizehint!(seen, length(C))
     end
-
-    haslength(C) && sizehint!(seen, length(C))
-
     while x !== nothing
         v, s = x
         idx = ht_keyindex2!(seen, v)


chriselrod:
Thoughts on adding `sizehint!(seen, length(C))`?
This shaves off a couple seconds for me when `x = collect(1:10^8)`, but takes about a tenth of a second longer when `x = fill(0, 10^8)`.
Whether that helps on average depends on how long it takes to find the non-unique element. I don't know about most use cases, but presumably the arrays are not mostly homogeneous, so that `sizehint!` will help on average.

bkamins:
This is a good point as following

> but presumably the arrays are not mostly homogeneous

I would assume that when someone calls `allunique` one assumes that this is `true` and just wants to make sure it is indeed the case. Also a common case is that you run `allunique` and have two paths of the algorithm that are fast (when it is `true`) and slow (when it is `false`) and it is preferable to maximize the speed of optimistic branch.

The drawback is that when we get `false` fast `allunique` will have a larger memory footprint, but I think it is not a problem (if someone calls `allunique` one probably accepts that in the worst case maximum memory will be used).

base/set.jl
```diff
@@ -383,14 +383,11 @@ false
 """
 function allunique(C)
     seen = Set{eltype(C)}()
-    for x in C
-        if in(x, seen)
-            return false
-        else
-            push!(seen, x)
-        end
+    for (i, x) in enumerate(C)
```
bkamins, line 386:
```suggestion
    haslength(C) && sizehint!(seen, length(C))
    for (i, x) in enumerate(C)
```

StefanKarpinski:
What about calling `sizehint!(length(a))` after checking some threshold of elements? That way both extreme cases are optimized: of you find duplicates right away you don't have to allocate much memory, of there are no duplicates, you only reallocate O(1) times.

StefanKarpinski:
Oh, also, calling internal bashing methods in Base code is fine. If we break it, we'll fix it.

base/set.jl
```diff
@@ -383,14 +383,12 @@ false
...
-    for x in C
-        if in(x, seen)
-            return false
-        else
-            push!(seen, x)
-        end
+    haslength(C) && sizehint!(seen, length(C))
+    for (i, x) in enumerate(C)
+        push!(seen, x)
+        i > length(seen) && return false
```
bkamins, line 389:
```suggestion

    x = iterate(C)
    for i in Base.OneTo(1000)
        if x === nothing
            return true
        else
            v, s = x
            idx = Base.ht_keyindex2!(seen, v)
            idx > 0 && return false
            Base._setindex!(seen, nothing, v, -idx)
            x = iterate(C, s)
        end
    end

    x === nothing && return true
    Base.haslength(C) && sizehint!(seen, length(C))

    while x !== nothing
        v, s = x
        idx = Base.ht_keyindex2!(seen, v)
        idx > 0 && return false
        Base._setindex!(seen, nothing, v, -idx)
        x = iterate(C, s)
    end
```

bkamins:
I have applied your suggestion with 1000 as a threshold.

base/set.jl
```diff
@@ -383,14 +383,32 @@ false
...
+    haslength(C) && sizehint!(seen, length(C))
+
+    while x !== nothing
+        v, s = x
+        idx = ht_keyindex2!(seen, v)
+        idx > 0 && return false
+        _setindex!(seen, nothing, v, -idx)
+        x = iterate(C, s)
+    end
+    end
```
bkamins, line 410:
```suggestion
    haslength(C) && sizehint!(seen, length(C))

    while x !== nothing
        v, s = x
        idx = ht_keyindex2!(seen, v)
        idx > 0 && return false
        _setindex!(seen, nothing, v, -idx)
        x = iterate(C, s)
    end
```

base/set.jl
```diff
@@ -383,14 +383,30 @@ false
 """
 function allunique(C)
     seen = Set{eltype(C)}()
-    for x in C
-        if in(x, seen)
-            return false
+
+    x = iterate(C)
+    for i in OneTo(1000)
+        if x === nothing
```
fredrikekre, line 389: Seems like this `if-else` block can be removed by just adding `x === nothing && return true` above the `for` loop?

bkamins: I think we need this check inside the loop for cases when the iterator has less than 1000 elements.

fredrikekre: Ah, obviously.

base/set.jl
```diff
@@ -383,14 +383,30 @@ false
 """
 function allunique(C)
     seen = Set{eltype(C)}()
```
bkamins, line 385:
```suggestion
    seen = Dict{eltype(C),Nothing}()
```

StefanKarpinski:
Does that seem to give good performance for both cases?

bkamins:
I have rewritten the 2-pass approach a bit to make it faster.

The benchmarks are below. The general conclusion is I think to use what I propose now or not use `sizehint!` at all (the difference is essentially removing first `if` statement in the code).

The issue is that just doing `push!` triggers reallocation of `Dict` to different sizes than when using `sizehint!` so depending on the actual size of the dict ether one or the other is faster. You can see this for `length(x) = 32768` case - `sizehint!` actually allocates more than just by doing `push!` 32768 times and thus is slower.

```
julia> for x in [[1:8^i;] for i in 0:8]
       length(x) > 6 && (x[7] = 1)
       @show length(x)
       @btime allunique($x)
       @btime allunique2($x) # current proposal
       @btime allunique3($x) # proposal without sizehint!
       @btime allunique4($x) # proposal with sizehint! always
       end
length(x) = 1
  98.520 ns (4 allocations: 480 bytes)
  93.486 ns (4 allocations: 480 bytes)
  94.117 ns (4 allocations: 480 bytes)
  94.537 ns (4 allocations: 480 bytes)
length(x) = 8
  162.370 ns (4 allocations: 480 bytes)
  131.782 ns (4 allocations: 480 bytes)
  131.088 ns (4 allocations: 480 bytes)
  130.639 ns (4 allocations: 480 bytes)
length(x) = 64
  161.985 ns (4 allocations: 480 bytes)
  131.663 ns (4 allocations: 480 bytes)
  131.554 ns (4 allocations: 480 bytes)
  207.500 ns (7 allocations: 1.09 KiB)
length(x) = 512
  162.055 ns (4 allocations: 480 bytes)
  131.777 ns (4 allocations: 480 bytes)
  131.555 ns (4 allocations: 480 bytes)
  294.693 ns (7 allocations: 5.07 KiB)
length(x) = 4096
  161.675 ns (4 allocations: 480 bytes)
  134.365 ns (4 allocations: 480 bytes)
  131.670 ns (4 allocations: 480 bytes)
  754.455 ns (7 allocations: 36.66 KiB)
length(x) = 32768
  164.145 ns (4 allocations: 480 bytes)
  134.708 ns (4 allocations: 480 bytes)
  131.874 ns (4 allocations: 480 bytes)
  3.038 μs (7 allocations: 288.66 KiB)
length(x) = 262144
  162.113 ns (4 allocations: 480 bytes)
  134.708 ns (4 allocations: 480 bytes)
  131.603 ns (4 allocations: 480 bytes)
  29.499 μs (7 allocations: 2.25 MiB)
length(x) = 2097152
  161.983 ns (4 allocations: 480 bytes)
  134.136 ns (4 allocations: 480 bytes)
  131.777 ns (4 allocations: 480 bytes)
  753.400 μs (7 allocations: 18.00 MiB)
length(x) = 16777216
  163.104 ns (4 allocations: 480 bytes)
  135.632 ns (4 allocations: 480 bytes)
  133.143 ns (4 allocations: 480 bytes)
  7.493 ms (7 allocations: 144.00 MiB)

julia> for x in [[1:8^i;] for i in 0:8]
       @show length(x)
       @btime allunique($x)
       @btime allunique2($x) # current proposal
       @btime allunique3($x) # proposal without sizehint!
       @btime allunique4($x) # proposal with sizehint! always
       end
length(x) = 1
  95.058 ns (4 allocations: 480 bytes)
  91.518 ns (4 allocations: 480 bytes)
  90.691 ns (4 allocations: 480 bytes)
  90.900 ns (4 allocations: 480 bytes)
length(x) = 8
  179.001 ns (4 allocations: 480 bytes)
  140.467 ns (4 allocations: 480 bytes)
  137.979 ns (4 allocations: 480 bytes)
  137.616 ns (4 allocations: 480 bytes)
length(x) = 64
  1.380 μs (10 allocations: 3.88 KiB)
  979.900 ns (10 allocations: 3.88 KiB)
  969.900 ns (10 allocations: 3.88 KiB)
  925.000 ns (10 allocations: 3.65 KiB)
length(x) = 512
  9.500 μs (13 allocations: 13.22 KiB)
  5.940 μs (13 allocations: 13.22 KiB)
  5.900 μs (13 allocations: 13.22 KiB)
  6.400 μs (10 allocations: 23.46 KiB)
length(x) = 4096
  107.699 μs (22 allocations: 193.86 KiB)
  84.400 μs (22 allocations: 193.86 KiB)
  83.401 μs (22 allocations: 193.86 KiB)
  70.600 μs (12 allocations: 180.96 KiB)
length(x) = 32768
  832.599 μs (27 allocations: 770.16 KiB)
  958.200 μs (27 allocations: 1.46 MiB)
  642.000 μs (27 allocations: 770.16 KiB)
  924.801 μs (12 allocations: 1.41 MiB)
length(x) = 262144
  12.001 ms (37 allocations: 7.50 MiB)
  8.920 ms (27 allocations: 6.80 MiB)
  9.678 ms (37 allocations: 7.50 MiB)
  8.887 ms (12 allocations: 6.75 MiB)
length(x) = 2097152
  162.217 ms (52 allocations: 70.50 MiB)
  113.774 ms (27 allocations: 54.05 MiB)
  135.403 ms (52 allocations: 70.50 MiB)
  112.357 ms (12 allocations: 54.00 MiB)
length(x) = 16777216
  2.091 s (67 allocations: 574.50 MiB)
  1.587 s (27 allocations: 432.05 MiB)
  1.699 s (67 allocations: 574.50 MiB)
  1.564 s (12 allocations: 432.00 MiB)
```

JeffBezanson:
This version looks good to me. Additionally I would call this a performance bug in Dict --- for the exact same workload, `sizehint!` should always be a strict improvement. In this case sizehint is leading to a larger final allocated size of the Dict, which is a bit crazy. I'm not sure how that's happening but we should figure out how to tune it to fix that.

KristofferC:
> This version looks good to me. Additionally I would call this a performance bug in Dict --- for the exact same workload, sizehint! should always be a strict improvement. 

https://github.com/JuliaLang/julia/issues/9909 might be related

Issue #31162, opened 2019-02-23
Julep: Expr provenance when lowering


timholy:
When lowering code we have line-number provenance, but I want expression-provenance too.
I'd love it if
```julia
ex = quote
    if Sys.islinux()
        f() = 1; g() = 2
    else
        g() = 3
    end
end
thunk, src = Meta.lower_logged(m, ex)
```
resulted in something along the lines of 
```
# Lowered result                                           Corresponding src entry
:($(Expr(:thunk, CodeInfo(
    @ REPL[3]:2 within `top-level scope'
1 ─ %1  = Base.getproperty(Sys, :islinux)                  :(Sys.islinux)
│   %2  = (%1)()                                           :(Sys.islinux())
└──       goto #3 if not %2                                ex
    @ REPL[3]:3 within `top-level scope'
2 ─       $(Expr(:thunk, CodeInfo(                         :(f() = 1)
    @ none within `top-level scope'
1 ─     return $(Expr(:method, :f))
)))
│         $(Expr(:method, :f))                             :(f() = 1)
│   %6  = Core.Typeof(f)                                   :(f)
│   %7  = Core.svec(%6)                                    :(f())
│   %8  = Core.svec()                                      :(f())
│   %9  = Core.svec(%7, %8)                                :(f())
│         $(Expr(:method, :f, :(%9), CodeInfo(quote        :(f() = 1)
    return 1
end)))
│         $(Expr(:thunk, CodeInfo(                         :(g() = 2)
    @ none within `top-level scope'
1 ─     return $(Expr(:method, :g))
)))
│         $(Expr(:method, :g))                             :(g() = 2)
│   %13 = Core.Typeof(g)                                   :(g)
│   %14 = Core.svec(%13)                                   :(g())
│   %15 = Core.svec()                                      :(g() )
│   %16 = Core.svec(%14, %15)                              :(g())
│         $(Expr(:method, :g, :(%16), CodeInfo(quote       :(g() = 2)
    return 2
end)))
└──       return g                                         :(g)
    @ REPL[3]:5 within `top-level scope'
3 ─       $(Expr(:method, :g))                             :(g() = 3)
│   %20 = Core.Typeof(g)                                   :(g)
│   %21 = Core.svec(%20)                                   :(g())
│   %22 = Core.svec()                                      :(g())
│   %23 = Core.svec(%21, %22)                              :(g())
│         $(Expr(:method, :g, :(%23), CodeInfo(quote       :(g() = 3)
[...]

PR #37213, opened 2020-08-26
Make constructors for basic number types inferable

martinholters:
Starting from the assumption that a constructor should return an instance of the correct type, i.e. `T(...)::T` where `T isa Type`, I wondered: Does that hold for all types in julia proper? (Answer: no, e.g. ref. #37186 for an accidental violation I stumbled upon.) And can inference figure it out?

This PR makes the answer to the second question "yes" for the basic number types, see the test:
https://github.com/JuliaLang/julia/blob/dca67c9901e5962f3dc0cc2ba3b1e76937697947/test/numbers.jl#L2696-L2700
(There are two deprecated `BigFloat` methods which I haven't touched that would fail this test.)

However, while that holds for the individual constructor methods, for unknown argument types, usually too many methods match, so that e.g. `Float64(x::Any)` (26 possible methods) will still only be inferred as `Any`. So while I don't expect these changes to be wrong, I am not really sure how  useful they are. Maybe @timholy has some intuition about it from all his work on invalidations where he looked at a lot code with incomplete type information?

### DIFF
--- a/test/numbers.jl
+++ b/test/numbers.jl
@@ -2692,3 +2692,15 @@ end
         end
     end
 end
+
+@testset "constructor inferability for $T" for T in [AbstractFloat, #=BigFloat,=# Float16,
+        Float32, Float64, Integer, Bool, Signed, BigInt, Int128, Int16, Int32, Int64, Int8,
+        Unsigned, UInt128, UInt16, UInt32, UInt64, UInt8]
+    @test all(R -> R<:T, Base.return_types(T))
+end
+@testset "constructor inferability for BigFloat" begin
+    T = BigFloat
+    @test_broken all(R -> R<:T, Base.return_types(T))
+    @test all(m -> m.file == Symbol("deprecated.jl"),
+        collect(methods(T))[findall(R -> !(R<:T), Base.return_types(T))])
+end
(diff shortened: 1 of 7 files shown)


timholy (approved):

timholy:
I think these are worth having even if they don't solve the problem for cases of abstract inference. But yes, I think we'll still have to annotate the return type in a few cases.

Some day it might be helpful to have a tool that goes through the source code and suggests type-annotations to remove, because all inference problems have been fixed. It would be much nicer to write such a tool if we had #31162 and similar links so that we could transition seamlessly from the type-inferred code to the source code, but there are other ways of achieving similar ends.

test/numbers.jl
```diff
@@ -2692,3 +2692,14 @@ end
...
     end
 end
+
+@testset "constructor inferability for $T" for T in [AbstractFloat, #=BigFloat,=# Float16,
+        Float32, Float64, Integer, Bool, Signed, BigInt, Int128, Int16, Int32, Int64, Int8,
+        Unsigned, UInt128, UInt16, UInt32, UInt64, UInt8]
+    @test all(R -> R<:T, Base.return_types(T))
+end
+@testset "constructor inferability for BigFloat" begin
+    @test_broken all(R -> R<:T, Base.return_types(T))
```
JeffBezanson, line 2702: I think `T` is not defined here?

martinholters: D'Oh. I certainly did not intend the test to be _that_ broken. Fix incoming.

PR #37215, opened 2020-08-26
address some threading races

vtjnash:
Add annotations in many places (often relaxed, sometimes release) detected by TSAN as being data races. This additionally upgrades `jl_gc_state_set` to be a release store, as TSAN otherwise detects it as being lacking and gets rather noisy about it missing.

### DIFF
--- a/src/gc.c
+++ b/src/gc.c
@@ -193,17 +193,16 @@ static void jl_gc_wait_for_the_world(void)
 {
     if (jl_n_threads > 1)
         jl_wake_libuv();
-    for (int i = 0;i < jl_n_threads;i++) {
+    for (int i = 0; i < jl_n_threads; i++) {
         jl_ptls_t ptls2 = jl_all_tls_states[i];
-        // FIXME: The acquire load pairs with the release stores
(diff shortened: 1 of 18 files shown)


src/gc.c
```diff
@@ -3084,7 +3083,7 @@ JL_DLLEXPORT void jl_gc_collect(jl_gc_collection_t collection)
     }
     gc_debug_print();
 
-    int8_t old_state = jl_gc_state(ptls);
+    int8_t old_state = ptls->gc_state;
     ptls->gc_state = JL_GC_STATE_WAITING;
```
JeffBezanson, line 3087: Why don't these need to be atomic?

vtjnash:
```suggestion
    jl_atomic_store_release(&ptls->gc_state, JL_GC_STATE_WAITING);
```

The read doesn't need to be atomic because it's only ever written from the thread itself. I agree this appears to need a release store. I'm slightly surprised TSAN didn't notice that too, though perhaps the race window was just very small, since `jl_safepoint_start_gc` is an acquire-release fence.

PR #37219, opened 2020-08-26
add some needed `local` declarations to macros

JeffBezanson:
to avoid leaking globals


base/logging.jl
```diff
@@ -278,19 +278,19 @@ default_group(file) = Symbol(splitext(basename(file))[1])
...
             if !(logger === nothing)
-                id = $(log_data._id)
+                local id = $(log_data._id)
                 # Second chance at an early bail-out (before computing the message),
                 # based on arbitrary logger-specific logic.
                 if _invoked_shouldlog(logger, level, _module, group, id)
-                    file = $(log_data._file)
-                    line = $(log_data._line)
+                    local file = $(log_data._file)
+                    local line = $(log_data._line)
```
aviatesk, line 293: guess we can wrap the entire block into `let` instead of adding bunch of `local`s ? I hope it would look simpler

base/c.jl
```diff
@@ -643,8 +643,8 @@ function ccall_macro_lower(convention, func, rettype, types, args, nreq)
         sym = Symbol(string("arg", i, "root"))
         sym2 = Symbol(string("arg", i, ))
         earg, etype = esc(arg), esc(type)
-        push!(lowering, :($sym = Base.cconvert($etype, $earg)))
-        push!(lowering, :($sym2 = Base.unsafe_convert($etype, $sym)))
+        push!(lowering, :(local $sym = $(GlobalRef(Base, :cconvert))($etype, $earg)))
+        push!(lowering, :(local $sym2 = $(GlobalRef(Base, :unsafe_convert))($etype, $sym)))
```
aviatesk, line 647: another nit pick, why can't we directly interpolate functions ? 
```suggestion
        push!(lowering, :(local $sym = $(Base.cconvert)($etype, $earg)))
        push!(lowering, :(local $sym2 = $(Base.unsafe_convert)($etype, $sym)))
```

aviatesk (approved):

base/c.jl
JeffBezanson: While we allow that, it's a bit of a misfeature. It's better for code to describe where an object comes from, or how it is computed, than to contain the object itself. One way to think of it is that code is a textual description, so it should only contain text.

PR #37222, opened 2020-08-26
Add missing support for addition and subtraction

### LINKED ISSUE #28570: Date and DateTime support for Missing is missing
Currently, there is no support for `missing` values when it comes to working with `Date` and `DateTime`.

For example, the `Date` constructor throws an error when called on a `missing` input value:

```
Date(missing, "YY-mm-dd")

MethodError: Cannot `convert` an object of type Missings.Missing to an object of type Int64
This may have arisen from a call to the constructor Int64(...),
since type constructors fall back to convert methods.

Stacktrace:
 [1] Date(::Missings.Missing, ::String) at ./dates/types.jl:312
 [2] include_string(::String, ::String) at ./loading.jl:522
```

Similarly, computing the difference between two dates fails of one of them is missing:

```
Date(2018,8,10) - missing

MethodError: no method matching -(::Date, ::Missings.Missing)
Closest candidates are:
  -(::DateTime, ::Missings.Missing) at In[112]:8
  -(::Date, ::Base.Dates.Day) at dates/arithmetic.jl:78
  -(::Date, ::Base.Dates.Week) at dates/arithmetic.jl:76
  ...

Stacktrace:
 [1] include_string(::String, ::String) at ./loading.jl:522
```

Presumably, all unary and binary operators need to be overloaded, either in `Base` or in `Dates`. Any ideas and suggestions for how to address this issue would be more than welcome.

nilshg:
Inspired by the discussion in #28570.

Tests pass locally with this change. 

There are no additional tests for `missing` support in this PR, it seems to me the methods are so simple that I'm not sure what would be tested here (although I have a feeling saying things like this is why I'm not a software developer...)

### DIFF
--- a/stdlib/Dates/src/arithmetic.jl
+++ b/stdlib/Dates/src/arithmetic.jl
@@ -85,10 +85,10 @@ end
 (+)(y::Period, x::TimeType) = x + y
 
 # Missing support
-(+)(x::T, y::Missing) where {T<:AbstractTime} = missing
-(+)(x::Missing, y::T) where {T<:AbstractTime} = missing
-(-)(x::T, y::Missing) where {T<:AbstractTime} = missing
-(-)(x::Missing, y::T) where {T<:AbstractTime} = missing
+(+)(x::AbstractTime, y::Missing) = missing
+(+)(x::Missing, y::AbstractTime) = missing
+(-)(x::AbstractTime, y::Missing) = missing
+(-)(x::Missing, y::AbstractTime) = missing
     
 # AbstractArray{TimeType}, AbstractArray{TimeType}
 (-)(x::OrdinalRange{T}, y::OrdinalRange{T}) where {T<:TimeType} = Vector(x) - Vector(y)
--- a/stdlib/Dates/src/arithmetic.jl
+++ b/stdlib/Dates/src/arithmetic.jl
@@ -89,7 +89,7 @@ end
 (+)(x::Missing, y::AbstractTime) = missing
 (-)(x::AbstractTime, y::Missing) = missing
 (-)(x::Missing, y::AbstractTime) = missing
-    
+
 # AbstractArray{TimeType}, AbstractArray{TimeType}
 (-)(x::OrdinalRange{T}, y::OrdinalRange{T}) where {T<:TimeType} = Vector(x) - Vector(y)
 (-)(x::AbstractRange{T}, y::AbstractRange{T}) where {T<:TimeType} = Vector(x) - Vector(y)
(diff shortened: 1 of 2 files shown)


stdlib/Dates/src/arithmetic.jl
```diff
@@ -84,6 +84,12 @@ end
 (-)(x::Time, y::TimePeriod) = return Time(Nanosecond(value(x) - tons(y)))
 (+)(y::Period, x::TimeType) = x + y
 
+# Missing support
+(+)(x::T, y::Missing) where {T<:AbstractTime} = missing
+(+)(x::Missing, y::T) where {T<:AbstractTime} = missing
+(-)(x::T, y::Missing) where {T<:AbstractTime} = missing
+(-)(x::Missing, y::T) where {T<:AbstractTime} = missing
```
nalimilan, line 91: Even simpler:
```suggestion
# Missing support
(+)(x::AbstractTime, y::Missing) = missing
(+)(x::Missing, y::AbstractTime) = missing
(-)(x::AbstractTime, y::Missing) = missing
(-)(x::Missing, y::AbstractTime) = missing
```

pdeffebach:
You need `<`, `>`, `<=`, and `>=` as well, right?

nilshg:
> 
> 
> You need `<`, `>`, `<=`, and `>=` as well, right?

Those are covered in `missings.jl` in Base as per discussion on Slack, right?

pdeffebach:
Yes, no need to add them.

stdlib/Dates/test/arithmetic.jl
```diff
@@ -500,7 +500,7 @@ end
 end
 
 @testset "Missing arithmetic" begin
-    for t ∈ [Date; Time; subtypes(DatePeriod); subtypes(TimePeriod)]
+    for t ∈ [Date Time Day Month Week Year Hour Microsecond Millisecond Minute Nanosecond Second]
```
nalimilan, line 503: It is more natural to use a vector rather than a matrix even if the result is the same:
```suggestion
    for t ∈ [Date, Time, Day, Month, Week, Year, Hour, Microsecond, Millisecond, Minute, Nanosecond, Second]
```

nalimilan (approved):
Thanks, looks good to me.

As discussed at #28570, I think this is justified as all types that support `+` and `-` should propagate `missing`. This is different from adding new `missing`-propagating *functions*, which I agree is problematic as it's not clear where to stop. Here the limit is very clear: add *methods* to all functions that already propagate missing values to ensure consistency.

Fun fact: this missing method was also [spotted](https://github.com/pandas-dev/pandas/issues/28095#issuecomment-537743228) by pandas developers, who have recently worked on improving their missing values handling and took Julia as a reference.

nilshg:
> 
> 
> Thanks, looks good to me.
> 

Thanks for your help!

quinnj (approved):

PR #37233, opened 2020-08-27
Better Vararg handling in tfuncs

martinholters:
Ref. https://github.com/JuliaLang/julia/pull/37188#issuecomment-680095700

I've split this into three commits in case I should back-out part of it. The first commit special-cases tfuncs which take a fixed number of arguments and expands a `Vararg` into the needed number of arguments. The second commit expands a `Vararg` to obtain at least `minarg` arguments and chops off the `Vararg` if there are already `maxarg` "normal" arguments. Thus, tfuncs which take a variable number of arguments will have to deal with `Vararg` arguments, but not for the first `minarg` ones. Interestingly, only `apply_type_tfunc` needed an adjustment here. But that's where reviewers might want to take an extra careful look. The tfuncs to look at are the ones for (before the third commit):
```julia
julia> Core.Compiler.T_FFUNC_KEY[findall(tf -> tf[1] != tf[2], Core.Compiler.T_FFUNC_VAL)]
6-element Array{Any,1}:
 isdefined (built-in function)
 _expr (built-in function)
 applicable (built-in function)
 getfield (built-in function)
 fieldtype (built-in function)
 apply_type (built-in function)

julia> reinterpret.(Core.IntrinsicFunction, Int32.(findall(tf -> tf[1] != tf[2], Core.Compiler.T_IFUNC)))
2-element Array{Core.IntrinsicFunction,1}:
 llvmcall (intrinsic function #83)
 arraylen (intrinsic function #84) 
```
Finally, the third commit outlines the handling of `svec`, `arrayset`, `arrayref`, and `const_arrayref` into separate tfuncs. I assume they only received their special treatment in `builtin_tfunction` to deal with `Vararg`s which is not needed anymore now. I've left the cost at the default of 20, but of course, we'd now have the flexibility to adjust that.
There's still special handling for `tuple` and `invoke`, which have associated `tuple_tfunc` and `invoke_tfunc`, but these take different arguments, so I've left it at that for now.

### DIFF
--- a/base/compiler/tfuncs.jl
+++ b/base/compiler/tfuncs.jl
@@ -1093,6 +1093,9 @@ function apply_type_tfunc(@nospecialize(headtypetype), @nospecialize args...)
     else
         return Type
     end
+    if !isempty(args) && isvarargtype(args[end])
+        return Type
+    end
     largs = length(args)
     if headtype === Union
         largs == 0 && return Const(Bottom)
@@ -1417,11 +1420,6 @@ function builtin_tfunction(interp::AbstractInterpreter, @nospecialize(f), argtyp
             end
         end
         return Any
-    elseif f === Core._expr
-        if length(argtypes) < 1 && !isva
-            return Bottom
-        end
-        return Expr
     elseif f === invoke
         if length(argtypes) > 1 && sv !== nothing && (isa(argtypes[1], Const) || isa(argtypes[1], Type))
             if isa(argtypes[1], Const)
@@ -1471,15 +1469,17 @@ function builtin_tfunction(interp::AbstractInterpreter, @nospecialize(f), argtyp
             # definitely too many arguments
             return Bottom
         end
-        if tf[1] == tf[2] || length(argtypes) - 1 == tf[2]
-            # expand Vararg to required number of arguments
-            vatype = unwrapva(argtypes[end])
+        if length(argtypes) - 1 == tf[2]
             argtypes = argtypes[1:end-1]
-            while length(argtypes) < tf[2]
-                push!(argtypes, vatype)
-            end
         else
-            return Any
+            vatype = argtypes[end]
+            argtypes = argtypes[1:end-1]
+            while length(argtypes) < tf[1]
+                push!(argtypes, unwrapva(vatype))
+            end
+            if length(argtypes) < tf[2]
+                push!(argtypes, unconstrain_vararg_length(vatype))
+            end
         end
     elseif !(tf[1] <= length(argtypes) <= tf[2])
         # wrong # of args
--- a/base/compiler/tfuncs.jl
+++ b/base/compiler/tfuncs.jl
@@ -390,6 +390,7 @@ function nfields_tfunc(@nospecialize(x))
 end
 add_tfunc(nfields, 1, 1, nfields_tfunc, 1)
 add_tfunc(Core._expr, 1, INT_INF, (@nospecialize args...)->Expr, 100)
+add_tfunc(svec, 0, INT_INF, (@nospecialize args...)->SimpleVector, 20)
 function typevar_tfunc(@nospecialize(n), @nospecialize(lb_arg), @nospecialize(ub_arg))
     lb = Union{}
     ub = Any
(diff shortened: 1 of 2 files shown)


martinholters:
Bump. @JeffBezanson do I interpret your :+1: correctly as sympathetic to the idea/general approach, but not reviewed properly, yet?

test/compiler/inference.jl
```diff
@@ -2743,3 +2743,47 @@ f_generator_splat(t::Tuple) = tuple((identity(l) for l in t)...)
...
+@test Core.Compiler.return_type(apply26826, Tuple{typeof(fieldtype), Any, Any, Any, Any, Vararg}) == Union{}
+@test Core.Compiler.return_type(apply26826, Tuple{typeof(Core.apply_type), Vararg}) == Type
+@test Core.Compiler.return_type(apply26826, Tuple{typeof(Core.apply_type), Any, Vararg}) == Type
+@test Core.Compiler.return_type(apply26826, Tuple{typeof(Core.apply_type), Any, Any, Vararg}) == Type
+f_apply_cglobal(args...) = cglobal(args...)
+@test Core.Compiler.return_type(f_apply_cglobal, Tuple{Vararg{Type{Int}}}) == Ptr
+@test Core.Compiler.return_type(f_apply_cglobal, Tuple{Any, Vararg{Type{Int}}}) == Ptr
+@test Core.Compiler.return_type(f_apply_cglobal, Tuple{Any, Type{Int}, Vararg{Type{Int}}}) == Ptr{Int}
+@test Core.Compiler.return_type(f_apply_cglobal, Tuple{Any, Type{Int}, Type{Int}, Vararg{Type{Int}}}) == Union{}
+f_apply_llvmcall(args...) = Core.Intrinsics.llvmcall(args...)
```
vtjnash, line 2783: This isn't well-formed IR (llvmcall is weird and special)

base/compiler/tfuncs.jl
```diff
@@ -1464,7 +1452,24 @@ function builtin_tfunction(interp::AbstractInterpreter, @nospecialize(f), argtyp
...
+        if length(argtypes) - 1 == tf[2]
+            argtypes = argtypes[1:end-1]
+        else
+            vatype = argtypes[end]
+            argtypes = argtypes[1:end-1]
+            while length(argtypes) < tf[1]
+                push!(argtypes, unwrapva(vatype))
+            end
+            if length(argtypes) < tf[2]
+                push!(argtypes, vatype)
```
vtjnash, line 1469: possibly might also need to adjust a length constraint on vatype (though I don't know if we'd ever form this)? for example, if it was `Vararg{NTuple{N}, N} where N`. This is probably already broken elsewhere, so not to worry now

vtjnash (approved):

test/compiler/inference.jl
martinholters: `llvmcall` may not actually appear in `_apply_iterate` when compiling, but it's possible to construct cases where inference first sees a `Vararg` but then later it's compiled for a fixed type, e.g.:
```julia
foo(x...) = f_apply_llvmcall("ret i64 %0", Int, Tuple{Int}, x...)
bar() = foo(Any[4][1]...) # inferred as Any on master, Int on this branch
```
Not that this is code of any practical relevance, but `bar()` run just fine.

base/compiler/tfuncs.jl
martinholters: Good catch! Even if broken elsewhere, no excuse for not handling it correctly here, even if it doesn't make much difference now. As long as it's not overly complicated, that is, but I think this should do it:
```suggestion
                push!(argtypes, unconstrain_vararg_length(vatype))
```
There may be corner case where we loose a little bit of precision due to this, but that's better than giving a wrong result. Or am I missing something and this simple fix won't do it?

martinholters:
As I had to rebase anyway, I included the `unconstrain_vararg_length` (ref. https://github.com/JuliaLang/julia/pull/37233#discussion_r487690730). Unless you see any problems with that @vtjnash I think this should be good to go.

test/compiler/inference.jl
vtjnash: Yeah, llvmcall is pretty buggy and undefined with handling edge cases

vtjnash:
aside from the tests depending on undefined behavior (llvmcall), this lgtm

martinholters:
Ok, I've removed the `llvmcall` test. If there is a non-empty intersection of defined behavior for `llvmcall` and `Vararg` arguments I could test here, let me know. Otherwise, I'll merge in a day or so (assuming CI comes out green again).

Issue #36414, opened 2020-06-24
Potential Performance regression between Julia 1.4.2 and 1.5 Beta 1


lungben:
Hi,

I observed a strange effect comparing Timings for an Asian Pricer implementation in Julia 1.4.2 and 1.5 Beta 1.
For the vectorized implementation, Julia 1.5 is significantly faster than Julia 1.4.2 (which is great!).
However, for the loop implementation, it is the other way around - Julia 1.5 took nearly twice as long.

Could you please check? Or am I missing something obvious?

Thanks and Best Regards
Benjamin

```
using BenchmarkTools
using Random
using Statistics 

T, r, K, s, Xo = promote(1.0, 0.05, 55.0, 0.3, 50) # align types for performance
m, mc = 1000, 10_000

# Source: https://gist.github.com/bkamins/7ecb8bf493cac8d6f6f3c141eedbaf4b

function asian_sample_julia(T, r, K, σ, X₀, m::Integer)
    X = X₀
    x̂ = zero(X)
    Δ = T/m
    for i in 1:m
        X *= exp((r-σ^2/2)*Δ + σ*√Δ*randn())
        x̂ += X
    end
    exp(-r*T)*max(x̂/m - K, 0)
end

function asian_sample_vec_julia(T, r, K, σ, X₀, m::Integer; rng=randn)
    Δ = T/m
    X = rng(typeof(X₀), m)
    X .*= σ*√Δ
    X .+= (r-σ^2/2)*Δ
    X .= exp.(cumsum!(X, X))
    exp(-r*T)*max(mean(X)*X₀ - K, 0)
end

asian_loop_julia(T, r, K, σ, X₀, m::Integer, mc:: Integer, f; kwargs...) = mean(f(T, r, K, σ, X₀, m; kwargs...) for _ in 1:mc)

versioninfo()

println("Loop implementation:")
@btime asian_loop_julia($T, $r, $K, $s, $Xo, $m, $mc, $asian_sample_julia)

println("Vectorized implementation:")
@btime asian_loop_julia($T, $r, $K, $s, $Xo, $m, $mc, $asian_sample_vec_julia)
```

Output:
```
Julia Version 1.4.2
Commit 44fa15b150* (2020-05-23 18:35 UTC)
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: Intel(R) Core(TM) i7-8650U CPU @ 1.90GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
[...]

PR #37234, opened 2020-08-27
Performance regression of scalar randn() between Julia 1.4 and 1.5

lungben:
The scalar `randn()` function is significantly slower in Julia 1.5 (and master) than it was in Julia 1.4. This was a side-effect of the drastic speed-up implemented for randn vectors.

See https://github.com/JuliaLang/julia/issues/36414

The underlying issue is that a function call is not inlined.
In this PR, this issue is avoided by redefining the scalar version of `randn()`, using the Julia 1.4 code.
This brings back the V1.4 performance for `randn()`, but is not the nicest solution. But unfortunately I have not found a better one yet.

### DIFF
--- a/stdlib/Random/src/normal.jl
+++ b/stdlib/Random/src/normal.jl
@@ -35,9 +35,18 @@ julia> randn(rng, ComplexF32, (2, 3))
   0.611224+1.56403im   0.355204-0.365563im  0.0905552+1.31012im
 ```
 """
+#=
+When defining 
+`@inline randn(rng::AbstractRNG=default_rng()) = _randn(rng, rand(rng, UInt52Raw()))`
+the function call to `_randn` is currently not inlined, resulting in slightly worse
+performance for scalar random normal numbers than repeating the code of `_randn`
+inside the following function.
+=#
 @inline function randn(rng::AbstractRNG=default_rng())
     @inbounds begin
         r = rand(rng, UInt52Raw())
+        
+        # the following code is identical to the one in `_randn(rng::AbstractRNG, r::UInt64)`
         r &= 0x000fffffffffffff
         rabs = Int64(r>>1) # One bit for the sign
         idx = rabs & 0xFF
--- a/stdlib/Random/src/normal.jl
+++ b/stdlib/Random/src/normal.jl
@@ -35,7 +35,17 @@ julia> randn(rng, ComplexF32, (2, 3))
   0.611224+1.56403im   0.355204-0.365563im  0.0905552+1.31012im
 ```
 """
-@inline randn(rng::AbstractRNG=default_rng()) = _randn(rng, rand(rng, UInt52Raw()))
+@inline function randn(rng::AbstractRNG=default_rng())
+    @inbounds begin
+        r = rand(rng, UInt52Raw())
+        r &= 0x000fffffffffffff
+        rabs = Int64(r>>1) # One bit for the sign
+        idx = rabs & 0xFF
+        x = ifelse(r % Bool, -rabs, rabs)*wi[idx+1]
+        rabs < ki[idx+1] && return x # 99.3% of the time we return here 1st try
+        return randn_unlikely(rng, idx, rabs, x)
+    end
+end
 
 @inline function _randn(rng::AbstractRNG, r::UInt64)
     @inbounds begin
--- a/stdlib/Random/src/normal.jl
+++ b/stdlib/Random/src/normal.jl
@@ -35,14 +35,14 @@ julia> randn(rng, ComplexF32, (2, 3))
   0.611224+1.56403im   0.355204-0.365563im  0.0905552+1.31012im
 ```
 """
-#=
-When defining 
-`@inline randn(rng::AbstractRNG=default_rng()) = _randn(rng, rand(rng, UInt52Raw()))`
-the function call to `_randn` is currently not inlined, resulting in slightly worse
-performance for scalar random normal numbers than repeating the code of `_randn`
-inside the following function.
-=#
 @inline function randn(rng::AbstractRNG=default_rng())
+    #=
+    When defining 
+    `@inline randn(rng::AbstractRNG=default_rng()) = _randn(rng, rand(rng, UInt52Raw()))`
+    the function call to `_randn` is currently not inlined, resulting in slightly worse
+    performance for scalar random normal numbers than repeating the code of `_randn`
+    inside the following function.
+    =#
     @inbounds begin
         r = rand(rng, UInt52Raw())
         


KristofferC:
Should probably at least add a comment to both places about this duplication.

lungben:
> Should probably at least add a comment to both places about this duplication.

Good point, I added some comments.

stdlib/Random/src/normal.jl
```diff
@@ -35,7 +35,26 @@ julia> randn(rng, ComplexF32, (2, 3))
   0.611224+1.56403im   0.355204-0.365563im  0.0905552+1.31012im
 ```
 """
-@inline randn(rng::AbstractRNG=default_rng()) = _randn(rng, rand(rng, UInt52Raw()))
+@inline function randn(rng::AbstractRNG=default_rng())
+    #=
+    When defining 
+    `@inline randn(rng::AbstractRNG=default_rng()) = _randn(rng, rand(rng, UInt52Raw()))`
+    the function call to `_randn` is currently not inlined, resulting in slightly worse
```
rfourquet, line 42: I was wondering whether it's really `_randn` which fails to be inlined... Looking rapidly at `@code_typed randn(MersenneTwister())` doesn't show a call to `_randn` if I'm not mistaken. How did you conclude that `randn_` is not inlined?

rfourquet:
This patch looks fine as a last resort, but I would rather postpone merging it until later before the 1.6 release, to give us a chance to understand better what happens exactly. If `_randn` is not inlined, then it should because of the `@inline` annotation, so the issue might be fixed by finding a way to force the compile to inline.

In case this is considered for backport to the 1.5 release, then this commit could be applied directly to 1.5 for the next 1.5. release, merged for 1.6 later only if we don't find better.

stdlib/Random/src/normal.jl
lungben: I deduced it from the `@code_llvm` (but I might be wrong here - I do not have much experience with these topics):
In the Julia 1.5 version, there is the following call:
`%30 = call double @j__randn_3665(%jl_value_t* nonnull %0, i64 %29) #0`
In the Julia 1.4 code, this call is missing, but there are ca. 60 additional lines of LLVM code. I suppose this is the inlined function, but am not sure about it.

rfourquet:
So by looking again at the output of `@code_typed randn(MersenneTwister())`, the output is AFAICT exactly the same between the version on master and with your patch, except for type annotations on the global arrays `wi` and `ki`: on master it looks like
```
 %48 = Random.ki::Vector{UInt64}
```
whereas with your patch, it looks like
```
%48 = Random.ki::Core.Const(UInt64[0x0007799ec012f7b2, 0x0000000000000000, ... ]
```
I don't really know what it means, but it looks like inference is capable to infer "constness" of these arrays, leading to better perfs. Would someone know what might cause the difference and whether it's possible to help inference on the master version without duplicating the code? (maybe @Keno ?)

If someone wants to try this patch, just do
```
@eval Random @inline function randn(rng::AbstractRNG=default_rng())
          @inbounds begin
              r = rand(rng, UInt52Raw())

              # the following code is identical to the one in `_randn(rng::AbstractRNG, r::UInt64)`
              r &= 0x000fffffffffffff
              rabs = Int64(r>>1) # One bit for the sign
              idx = rabs & 0xFF
              x = ifelse(r % Bool, -rabs, rabs)*wi[idx+1]
              rabs < ki[idx+1] && return x # 99.3% of the time we return here 1st try
              return randn_unlikely(rng, idx, rabs, x)
          end
      end
```

KristofferC:
What happens if you make these arrays tuples? Tuples are immutable and might be handled better.

rfourquet:
> What happens if you make those arrays tuples? Tuples are immutable and might be handled better.

Good idea, this seem to improve performance a bit on master. But doing the same trick atop this PR also improves things, again thanks to a `Const` annotation:
```
%61 = Random.ki_::NTuple{256, UInt64}
# vs
%61 = Random.ki_::Core.Const((0x0007799ec012f7b2, 0x0000000000000000, ...
```

stdlib/Random/src/normal.jl
vtjnash: Since `_randn` is defined as calling `randn`, the inliner decides to stop. This is reflected in `code_llvm`, but due to the particular way that `code_typed` works and is cached, it won't be represented there.

chriselrod:
Perhaps interesting, Julia 1.4:
```julia
julia> using Random, BenchmarkTools

julia> const DEFAULTRNG = Random.default_rng();

julia> @benchmark Random.default_rng()
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     2.850 ns (0.00% GC)
  median time:      2.857 ns (0.00% GC)
  mean time:        2.995 ns (0.00% GC)
  maximum time:     5.182 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000

julia> @benchmark randn(DEFAULTRNG)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     3.592 ns (0.00% GC)
  median time:      4.061 ns (0.00% GC)
  mean time:        4.077 ns (0.00% GC)
  maximum time:     6.759 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000

julia> @benchmark randn()
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     5.309 ns (0.00% GC)
  median time:      5.769 ns (0.00% GC)
  mean time:        5.792 ns (0.00% GC)
  maximum time:     19.508 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000
```
Julia master:
```julia
julia> using Random, BenchmarkTools

julia> const DEFAULTRNG = Random.default_rng();

julia> @benchmark Random.default_rng()
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     2.844 ns (0.00% GC)
  median time:      2.848 ns (0.00% GC)
  mean time:        2.856 ns (0.00% GC)
  maximum time:     12.756 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000

julia> @benchmark randn(DEFAULTRNG)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     3.864 ns (0.00% GC)
  median time:      4.233 ns (0.00% GC)
  mean time:        4.255 ns (0.00% GC)
  maximum time:     13.951 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000

julia> @benchmark randn()
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     6.842 ns (0.00% GC)
  median time:      7.298 ns (0.00% GC)
  mean time:        7.322 ns (0.00% GC)
  maximum time:     11.818 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000
```

stdlib/Random/src/normal.jl
JeffBezanson: `randn_unlikely` is the noinline slow path though. If we moved the code we'd have to inline everything?

stdlib/Random/src/normal.jl
```diff
@@ -35,7 +35,26 @@ julia> randn(rng, ComplexF32, (2, 3))
   0.611224+1.56403im   0.355204-0.365563im  0.0905552+1.31012im
 ```
 """
-@inline randn(rng::AbstractRNG=default_rng()) = _randn(rng, rand(rng, UInt52Raw()))
+@inline function randn(rng::AbstractRNG=default_rng())
+    #=
+    When defining
```
JeffBezanson, line 40: Trailing whitespace here.

stdlib/Random/src/normal.jl
```diff
@@ -35,7 +35,26 @@ julia> randn(rng, ComplexF32, (2, 3))
...
+    #=
+    When defining 
+    `@inline randn(rng::AbstractRNG=default_rng()) = _randn(rng, rand(rng, UInt52Raw()))`
+    the function call to `_randn` is currently not inlined, resulting in slightly worse
+    performance for scalar random normal numbers than repeating the code of `_randn`
+    inside the following function.
+    =#
+    @inbounds begin
+        r = rand(rng, UInt52Raw())
+
```
JeffBezanson, line 48: and here

KristofferC:
Rebased in https://github.com/JuliaLang/julia/pull/39319

PR #37236, opened 2020-08-27
Add INNO_ARGS makefile var for signtool

musm:
@staticfloat 

Signing now works when I add the following to my `Make.user`

```
override INNO_ARGS=/Dsign=true "/Smysigntool=\$$q`cygpath -w /cygdrive/c/Program\ Files\ \(x86\)/Windows\ Kits/10/bin/10.0.19041.0/x64/signtool.exe`\$$q sign /f \$$q`cygpath -w /cygdrive/c/Users/Mus/Code/julia/mycert.pfx`\$$q /p Pass123 \$$f"
```

In simplified terms

```
override INNO_ARGS=/Dsign=true "/Smysigntool=\$$q`cygpath -w {{signtool path}}`\$$q sign /f \$$q`cygpath -w {{certificate path}}`\$$q /p {{certificate password}} \$$f"
```

Yes it's kind of a total mess, but we can't help it since this is going through like 3 layers of path/filename translations: makefiles, cygwin, and inno....

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -468,7 +468,7 @@ endif
 
 exe:
 	# run Inno Setup to compile installer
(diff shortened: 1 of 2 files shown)


staticfloat:
Great!  This is perfect.

musm:
Great. BTW do these makefile vars default to empty strings if not defined? I was contemplating whether we needed to define them in Make.inc

staticfloat:
Yes, they default to empty.

PR #37217, opened 2020-08-26
libuv upgrade to v2-1.39.0

vtjnash:
~~Needs https://github.com/libuv/libuv/issues/2838 still, but I don't have a clean version of that change to apply. @staticfloat can you make a PR to libuv that I can use as the reference patch?~~
```
* 87d3a78a - Remove incorrect constant in chmod (#12) (5 weeks ago) <Elliot Saba>
*   2d109120 - Merge pull request #10 from JuliaLang/sf/inheritance (4 months ago) <Elliot Saba>
|\  
| * 2b839f1c - Alter permissions given in chmod (4 months ago) <Elliot Saba>
|/  
*   183e5558 - Merge pull request #9 from JuliaLang/sf/readonly_chmod (4 months ago) <Elliot Saba>
|\  
| * 3e3e203c - Propagate permissions, and fix readonly permission ordering (4 months ago) <Elliot Saba>
* | b5bf0b87 - Merge pull request #8 from JuliaLang/sf/readonly_chmod (4 months ago) <Elliot Saba>
|\| 
| * 62808ca2 - win,fs: Continue to set `FILE_ATTRIBUTE_READONLY` (4 months ago) <Elliot Saba>
|/  
*   659812b7 - Merge pull request #6 from JuliaLang/sf/win_acl (4 months ago) <Elliot Saba>
|\  
| * 173a9acb - Fix c89 pedantic warnings (4 months ago) <Elliot Saba>
| * 3ce845cd - win,fs: Add more complete `fs__chmod()` implementation (4 months ago) <Elliot Saba>
| * ce539ac6 - win,fs: Upgrade `fs__access()` to respond to `X_OK`. (8 months ago) <Elliot Saba>
```

PR #37242, opened 2020-08-27
win: Add long path aware manifest

musm:
This enables Julia to use the new long path functionality of Win10 1607.

Along with the Julia libuv upgrade that @vtjnash  is working on in  https://github.com/JuliaLang/julia/pull/37217

Reference: https://docs.microsoft.com/en-us/windows/win32/sbscs/application-manifests#longpathaware


vtjnash (approved):
(pending #37217)

musm:
Verified long paths are working with Julia's `cd` with this patch.

PR #37243, opened 2020-08-27
code coverage fixes

vtjnash:


IanButterworth:
How were the changes to the the good test file `coverage_file.info` generated here? Was it updated by hand to be an ideal target?

Generally, I'm trying to get a handle on what's known about the reason for the misses, to figure out https://github.com/JuliaLang/julia/issues/44593

Issue #37070, opened 2020-08-17
Fix Cmd interpolation with environments


staticfloat:
I think we can make some improvements to `Cmd` and environment handling.

Problems:
* `setenv(cmd, "FOO" => "BAR")` will unconditionally overwrite all environment settings already set within `cmd`.

* ``` `$foo $bar` ``` where `foo` and `bar` are both `cmd` objects with environment variables embedded throws an error.

I propose the following:

* Change `Cmd(cmd::Cmd, env=env)` and `setenv(cmd::Cmd, env)` to merge environments rather than replace them.  Functionally, merging can be as simple as `vcat(old_env, new_env)`, since the OS will replace overlapping environments properly.  This would act a little strangely with hashing, though, as `hash(c1.env)` != `hash(c2.env)` even though dumping the environment of the eventual process would show the same environment.  This might be a feature though, as perhaps there are OS bugs where the environment block is not properly parsed or something.  We probably shouldn't pretend that a `cmd.env` is a dictionary when it really isn't, despite most users using it like one.

* Change `cmd_gen()` (the `Cmd` interpolation argument processor) to no longer special-case the first `Cmd` object it finds, and to instead concatenate environment blocks from all `Cmd` objects.  This would allow you to ``` `$foo $bar` ``` with two different sets of environment variables set.  Of course, if `foo` sets `PATH` to something, and `bar` sets it to something else, it would not intelligently combine the two, you'd just get `bar`'s `PATH`, but that is only to be expected.

These changes will allow us to stop using `withenv()` in JLL packages, which is thread-unsafe.  Right now, the `ENV` block is the only realistic method to cooperatively layer environments, as it allows a natural way to set default environment values that can be overridden by the user.  However, due to its thread-unsafety, it's ultimately undesirable.  It would be much better to change the current API from:

```julia
# Git_jll.git() sneakily wraps all user code in a `with_env()` so that it has the proper PATH, LD_LIBRARY_PATH, etc...
Git_jll.git() do git_path
    run(setenv(`$git_path ...`, "USER_SPECIFIC_GIT_OPTION" => "1")
end
```

to something more like:
```julia
run(setenv(`$(Git_jll.git()) ...`, "USER_SPECIFIC_GIT_OPTION" => "1")
```

where the `git()` function would return a `Cmd` object with all necessary environs set within it.  Without solving this issue, the second syntax will result in an error, as `git` will be unable to find its dependent libraries.

StefanKarpinski:
[...]

PR #37244, opened 2020-08-27
Add `addenv()`; a way to add new mappings into `Cmd` environment blocks.

staticfloat:
This addresses one-half of https://github.com/JuliaLang/julia/issues/37070 as discussed during triage today.  One slightly awkward part of this is that `setenv()` and `addenv()` don't actually mutate the `cmd` object passed in, so you must make sure to `cmd = addenv(cmd, "FOO" => "bar")`.  I have made the docs as explicit as possible about this.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -71,6 +71,7 @@ New library functions
   for writing `(f(args...) for args in zip(iterators...))`, i.e. a lazy `map` ([#34352]).
 * New function `sincospi` for simultaneously computing `sinpi(x)` and `cospi(x)` more
   efficiently ([#35816]).
+* New function `addenv` for adding environment mappings into a `Cmd` object, returning the new `Cmd` object.
 
 New library features
 --------------------
--- a/base/cmd.jl
+++ b/base/cmd.jl
@@ -60,9 +60,10 @@ while changing the settings of the optional keyword arguments:
   already open or on non-Windows systems.
 * `env`: Set environment variables to use when running the `Cmd`. `env` is either a
   dictionary mapping strings to strings, an array of strings of the form `"var=val"`, an
-  array or tuple of `"var"=>val` pairs, or `nothing`. In order to modify (rather than
-  replace) the existing environment, create `env` by `copy(ENV)` and then set
-  `env["var"]=val` as desired.
+  array or tuple of `"var"=>val` pairs. In order to modify (rather than replace) the
+  existing environment, initialize `env` with `copy(ENV)` and then set `env["var"]=val` as
+  desired.  To add to an environment block within a `Cmd` object without replacing all
+  elements, use `addenv()` which will return a `Cmd` object with the updated environment.
 * `dir::AbstractString`: Specify a working directory for the command (instead
   of the current directory).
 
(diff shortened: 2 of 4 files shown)


DilumAluthge (approved):
Looks good to me.

If you wanted to be extra clear, you could explicitly say in the docstring "the original Cmd object will not be mutated".

NEWS.md
```diff
@@ -60,6 +60,7 @@ New library functions
   for writing `(f(args...) for args in zip(iterators...))`, i.e. a lazy `map` ([#34352]).
 * New function `sincospi` for simultaneously computing `sinpi(x)` and `cospi(x)` more
   efficiently ([#35816]).
+* New function `Base.addenv` for adding environment mappings into a `Cmd` object, returning the new `Cmd` object.
```
fredrikekre, line 63:
```suggestion
* New function `addenv` for adding environment mappings into a `Cmd` object, returning the new `Cmd` object.
```
since it is exported.

base/cmd.jl
```diff
@@ -242,6 +243,29 @@ setenv(cmd::Cmd, env::Pair{<:AbstractString}...; dir="") =
     setenv(cmd, env; dir=dir)
 setenv(cmd::Cmd; dir="") = Cmd(cmd; dir=dir)
 
+"""
+    addenv(command::Cmd, env...)
+
+Merge new environment mappings into the given `Cmd` object, returning a new `Cmd` object.
+Duplicate keys are replaced.
+"""
+function addenv(cmd::Cmd, env::Pair{<:AbstractString}...)
```
fredrikekre, line 252: Maybe make the core method accept `Dict` just like `setenv`
```suggestion
function addenv(cmd::Cmd, env::Dict)
```

base/cmd.jl
```diff
@@ -242,6 +243,29 @@ setenv(cmd::Cmd, env::Pair{<:AbstractString}...; dir="") =
...
+        end
+    end
+    for (k, v) in env
+        new_env[k] = v
+    end
+    return setenv(cmd, new_env)
+end
+
+function addenv(cmd::Cmd, env::Vector{<:AbstractString})
+    return addenv(cmd, [k => v for (k, v) in split.(env, "=")]...)
```
fredrikekre, line 266: With the suggestion above you don't need this splatting
```suggestion
    return addenv(cmd, Dict{String,String}(k => v for (k, v) in split.(env, "=")))
```

base/cmd.jl
```diff
@@ -242,6 +243,29 @@ setenv(cmd::Cmd, env::Pair{<:AbstractString}...; dir="") =
...
 setenv(cmd::Cmd; dir="") = Cmd(cmd; dir=dir)
 
+"""
+    addenv(command::Cmd, env...)
+
+Merge new environment mappings into the given `Cmd` object, returning a new `Cmd` object.
+Duplicate keys are replaced.
+"""
+function addenv(cmd::Cmd, env::Pair{<:AbstractString}...)
+    new_env = Dict{AbstractString,AbstractString}()
```
fredrikekre, line 253:
```suggestion
    new_env = Dict{String,String}()
```
maybe?

omus:
I'll mention the docstring for `setenv` mentions using `withenv`. Maybe we should add a method to `withenv` for this instead of defining a new function? Could be something like
```julia
withenv(cmd::Cmd, env::Pair{<:AbstractString}...) -> Cmd
```

PR #37247, opened 2020-08-27
allow unicode modifiers after ' 

### LINKED ISSUE #34507: Unicode modifiers for adjoint operator

simeonschaub:
As discussed in the triage call, this is technically breaking because we currently parse `a'ᵀ` as `a' * ᵀ` and `a'ᵀb` as `a' * ᵀb`, but it doesn't look like this is actually [used in the wild](https://juliahub.com/ui/RepoSearch?q=.%2a%27%5B%5Cp%7BLm%7D%5D%28%3F%21%27%29.%2a&r=true). Still needs news.

fixes #34507

### DIFF
--- a/base/show.jl
+++ b/base/show.jl
@@ -1131,6 +1131,10 @@ function isidentifier(s::AbstractString)
 end
 isidentifier(s::Symbol) = isidentifier(string(s))
 
+is_op_suffix_char(c::AbstractChar) = ccall(:jl_op_suffix_char, Cint, (UInt32,), c) != 0
+
+_isoperator(s) = ccall(:jl_is_operator, Cint, (Cstring,), s) != 0
+
 """
     isoperator(s::Symbol)
 
@@ -1142,7 +1146,7 @@ julia> Base.isoperator(:+), Base.isoperator(:f)
 (true, false)
 ```
 """
-isoperator(s::Union{Symbol,AbstractString}) = ccall(:jl_is_operator, Cint, (Cstring,), s) != 0
+isoperator(s::Union{Symbol,AbstractString}) = _isoperator(s) || ispostfixoperator(s)
 
 """
     isunaryoperator(s::Symbol)
@@ -1169,7 +1173,26 @@ julia> Base.isbinaryoperator(:-), Base.isbinaryoperator(:√), Base.isbinaryoper
 (true, false, false)
 ```
 """
-isbinaryoperator(s::Symbol) = isoperator(s) && (!isunaryoperator(s) || is_unary_and_binary_operator(s))
+function isbinaryoperator(s::Symbol)
+    return _isoperator(s) && (!isunaryoperator(s) || is_unary_and_binary_operator(s)) &&
+        s !== Symbol("'")
+end
+
+"""
+    ispostfixoperator(s::Union{Symbol,AbstractString})
+
+Return `true` if the symbol can be used as a postfix operator, `false` otherwise.
+
+# Examples
+```jldoctest
+julia> Base.ispostfixoperator(Symbol("'")), Base.ispostfixoperator(Symbol("'ᵀ")), Base.ispostfixoperator(:-)
+(true, true, false)
+```
+"""
+function ispostfixoperator(s::Union{Symbol,AbstractString})
+    s = String(s)
+    return startswith(s, '\'') && all(is_op_suffix_char, SubString(s, 2))
+end
 
 """
     operator_precedence(s::Symbol)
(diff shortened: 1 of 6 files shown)


base/show.jl
```diff
@@ -1618,6 +1618,16 @@ function show_unquoted(io::IO, ex::Expr, indent::Int, prec::Int, quote_level::In
                 show_unquoted(io, arg1, indent, func_prec, quote_level)
             end
 
+        # ' with unicode suffix
+        elseif isa(func,Symbol) && length(func_args) == 1 && occursin(r"^'[\p{Lm}]+$", string(func))
```
JeffBezanson, line 1622: This should share an implementation with the case for the `'` head. You version is better since it handles operators (`(+)'`) and the existing code doesn't.

src/julia-parser.scm
```diff
@@ -1241,7 +1241,18 @@
              (if (ts:space? s)
                  (error (string "space not allowed before \"" t "\"")))
              (take-token s)
-             (loop (list t ex)))
+             (let* ((port (ts:port s))
```
JeffBezanson, line 1244: This should be handled in the lexer, by removing `'` from the no-suffix list, then the parser test needs to be changed to use a `SuffSet`.

simeonschaub: Won't this cause any issues with character literals? Or is that handled differently?

simeonschaub: I think we could use `read-operator` directly for this though.

simeonschaub: > This should be handled in the lexer, by removing ' from the no-suffix list, then the parser test needs to be changed to use a SuffSet.

When I remove `ctrans-op` form the no-suffix list, `'ᵀ'` parses as `(|'| |'ᵀ|)`, so there needs to be an additional argument to `read-operator` when parsing a call chain vs when starting to parse a new atom.

JeffBezanson: Oops, you're right.

JeffBezanson (approved):

JeffBezanson:
One thing to think about is whether `isoperator` should return true for these.

simeonschaub:
That's a good point! I think it should return true, because to the user, it acts just like `'` does. I would say whether `(operator? t)` is true is mostly an implementation detail of the parser, but that for all practical purposes `'ᵀ` is also an operator.

simeonschaub:
Should I just change the Julia function `isoperator`, or should I add a check to the C function `jl_is_operator``to return true here?

base/show.jl
```diff
@@ -1128,6 +1128,8 @@ function isidentifier(s::AbstractString)
 end
 isidentifier(s::Symbol) = isidentifier(string(s))
 
+is_op_suffix_char(c::AbstractChar) = ccall(:jl_op_suffix_char, Cint, (UInt32,), c) != 0
```
simeonschaub, line 1131: This function isn't annotated with `JL_DLLEXPORT`: https://github.com/JuliaLang/julia/blob/7ecd1f9e0da5d07e7dd1733093b246a37bc118ff/src/flisp/julia_extensions.c#L152. Do I need to add that or is this ok?

simeonschaub (commented):
Other than my comment I think this should be good to go.

base/show.jl
simeonschaub: As @giordano pointed out on Slack, this is not strictly needed, because `jl_op_suffix_char` is exported in `flisp.h` already, but for consistency with `jl_id_char` and `jl_id_start_char`, I think it makes sense to add this here as well.

simeonschaub:
@JeffBezanson Sorry for bothering you, but would you be willing to take another look and merge if ok?

simeonschaub:
Bump. :slightly_smiling_face:

simeonschaub:
Ok?

PR #37251, opened 2020-08-28
Type-stable `intersect` for ranges

### LINKED ISSUE #32410: `intersect` is type-unstable for some types of ranges
In the following examples, `intersect` is not type-stable:
```julia
julia> @code_warntype intersect(1:2:5, 1:5)
Body::Union{StepRange{Int64,Int64}, UnitRange{Int64}}
[...]

julia> @code_warntype intersect(1:2:5, 1//1:1:5//1)
Body::Union{StepRange{Rational{Int64},Int64}, StepRange{Int64,Int64}}
[...]

julia> @code_warntype intersect(1, big(2):big(5))
Body::Union{UnitRange{Int64}, UnitRange{BigInt}}
[...]
```
The type-instabilities also occur if the positions of the arguments are switched.

Edit: added another example

sostock:
Fixes #32410.

### DIFF
--- a/base/range.jl
+++ b/base/range.jl
@@ -785,17 +785,15 @@ union(r::OneTo, s::OneTo) = OneTo(max(r.stop,s.stop))
 
 intersect(r::AbstractUnitRange{<:Integer}, s::AbstractUnitRange{<:Integer}) = max(first(r),first(s)):min(last(r),last(s))
 
-intersect(i::Integer, r::AbstractUnitRange{<:Integer}) =
+intersect(i::Integer, r::AbstractUnitRange{<:Integer})::UnitRange{promote_type(typeof(i),eltype(r))} =
     i < first(r) ? (first(r):i) :
     i > last(r)  ? (i:last(r))  : (i:i)
 
 intersect(r::AbstractUnitRange{<:Integer}, i::Integer) = intersect(i, r)
 
-function intersect(r::AbstractUnitRange{<:Integer}, s::StepRange{<:Integer})
+function intersect(r::AbstractUnitRange{T1}, s::StepRange{T2,S})::StepRange{promote_type(T1,T2),S} where {T1<:Integer,T2<:Integer,S}
     if isempty(s)
         range(first(r), length=0)
-    elseif step(s) == 0
-        intersect(first(s), r)
     elseif step(s) < 0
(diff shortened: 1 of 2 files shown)


timholy:
Just a general comment, it's better to fix the implementation than to put a bandaid over it by adding a `convert` on the returned value (which is what a return-type assertion does). https://github.com/JuliaLang/julia/issues/32410#issuecomment-505413633 indicates that the fix might be to make sure that all branches return a step range. I.e., replace `f:l` with `f:1:l`.

sostock:
Okay, I will do that.

sostock:
I now changed the implementation. However, in two of the three methods, the return type is still calculated explicitly (but not via return-type annotations).

Should I change that as well, i.e., use `T(a):S(b):T(c)` instead of `StepRange{T,S}(a,b,c)` for the return values? Note that these are not equivalent: the former will never create a `StepRange{Int,Rational{Int}}`.

base/range.jl
```diff
@@ -253,6 +253,8 @@ end
 # For types where x+oneunit(x) may not be well-defined
 steprange_last_empty(start, step, stop) = start - step
 
+StepRange{T,S}(start, step, stop) where {T,S} = StepRange{T,S}(T(start), S(step), T(stop))
```
timholy, line 256: Nice. But I think there's an even better solution: take the type annotations off the arguments in https://github.com/JuliaLang/julia/blob/5f0469b2273c8de57a26aec88468e5b5eed3a61b/base/range.jl#L204

Anyone who calls the constructor with specific type-parameters is guaranteeing what the result type should be, and there aren't multiple inner constructors, so I think there's no need to restrict the types of the arguments.

timholy (approved):
This implementation seems fine to me. Thanks!

base/range.jl
sostock: Done! The values are still converted explicitly for `steprange_last`, so that one can do `StepRange{Int,Int}(1.0, 2.0, 3.0)` without getting an error.

timholy: Test failures seem related. Perhaps my suggestion led you in the wrong direction?

sostock: Should be fixed now.

timholy:
Thanks for an elegant improvement!

PR #37254, opened 2020-08-28
Let spdiagm preserve sparsity

### LINKED ISSUE #37174: Unexpected behaviours of `spdiagm`
Say we have

```julia
julia> y = sprand(10, 0.2)
10-element SparseVector{Float64,Int64} with 2 stored entries:
  [2 ]  =  0.804026
  [3 ]  =  0.0159904
```

Then, firstly, there is no method that works implicitly on the main diagonal, like `diagm(y)` does -- you have to explicitely say `spdiagm(0 => y)`.  Secondly, and more importantly, this method does not preserve sparsity of the input:

```julia
julia> spdiagm(0 => y)
10×10 SparseMatrixCSC{Float64,Int64} with 10 stored entries:
  [1 ,  1]  =  0.0
  [2 ,  2]  =  0.804026
  [3 ,  3]  =  0.0159904
  [4 ,  4]  =  0.0
  [5 ,  5]  =  0.0
  [6 ,  6]  =  0.0
  [7 ,  7]  =  0.0
  [8 ,  8]  =  0.0
  [9 ,  9]  =  0.0
  [10, 10]  =  0.0
```

Where I really had expected the result to be

```julia
julia> spdiagm(0 => y)
10×10 SparseMatrixCSC{Float64,Int64} with 2 stored entries:
  [2 ,  2]  =  0.804026
  [3 ,  3]  =  0.0159904
```

(That's on 1.3.1, sorry for not having updated yet :))

dkarrasch:
Closes #37174.

With this PR:
```julia
julia> y = sprand(10, 0.2)
10-element SparseVector{Float64,Int64} with 4 stored entries:
  [1 ]  =  0.290569
  [2 ]  =  0.454582
  [6 ]  =  0.578091
  [9 ]  =  0.822982

julia> spdiagm(0 => y)
10×10 SparseMatrixCSC{Float64,Int64} with 4 stored entries:
 0.290569   ⋅         ⋅    ⋅    ⋅    ⋅         ⋅    ⋅    ⋅         ⋅ 
  ⋅        0.454582   ⋅    ⋅    ⋅    ⋅         ⋅    ⋅    ⋅         ⋅ 
  ⋅         ⋅         ⋅    ⋅    ⋅    ⋅         ⋅    ⋅    ⋅         ⋅ 
  ⋅         ⋅         ⋅    ⋅    ⋅    ⋅         ⋅    ⋅    ⋅         ⋅ 
  ⋅         ⋅         ⋅    ⋅    ⋅    ⋅         ⋅    ⋅    ⋅         ⋅ 
  ⋅         ⋅         ⋅    ⋅    ⋅   0.578091   ⋅    ⋅    ⋅         ⋅ 
  ⋅         ⋅         ⋅    ⋅    ⋅    ⋅         ⋅    ⋅    ⋅         ⋅ 
  ⋅         ⋅         ⋅    ⋅    ⋅    ⋅         ⋅    ⋅    ⋅         ⋅ 
  ⋅         ⋅         ⋅    ⋅    ⋅    ⋅         ⋅    ⋅   0.822982   ⋅ 
  ⋅         ⋅         ⋅    ⋅    ⋅    ⋅         ⋅    ⋅    ⋅         ⋅ 
```


barucden:
Have you considered using dispatching over `isa`? For example, the methods for determining the number of (non-zero) coefficients
```julia
_numel(vect::SparseVector) = nnz(vect)
_numel(vect::AbstractArray) = length(vect)
```
instead of
```julia
p.second isa AbstractSparseVector ? nnz(p.second) : length(p.second)
```
?

Imho, it is more readable and/or extensible.

stdlib/SparseArrays/src/sparsematrix.jl
```diff
@@ -3516,12 +3521,20 @@ function spdiagm_internal(kv::Pair{<:Integer,<:AbstractVector}...)
             col = 0
         end
         r = 1+i:numel+i
-        I[r] = row+1:row+numel
-        J[r] = col+1:col+numel
-        copyto!(view(V, r), vect)
+        if vect isa AbstractSparseVector
```
barucden, line 3524: Maybe it would be possible to translate this part into two modifying methods, e.g.,
```julia
_copytodiag!((I, J, V), vect::SparseVector, row, col)
_copytodiag!((I, J, V), vect::AbstractVector, row, col)
```
 as well?

dkarrasch: One can do that. One issue is that `AbstractSparseVector` is not available at this point, so the method needs to be defined in a different file, same as the `_numel` method for the sparse vectors. It seems to me that it might be hard to understand what this method is good for, because it's not called within `sparsevector.jl`. So I'm not sure, but wouldn't mind either.

barucden (commented):
`AbstractSparseVector` (unlike `SparseVector`) should be available since it is defined in `abstractsparse.jl`, which is included prior to `sparsematrix.jl` in `SparseArrays.jl`. I guess your second point is valid. I just think that dispatching is clearer than `isa` branching. Perhaps, someone more experienced will join the discussion.

stdlib/SparseArrays/src/sparsematrix.jl
dkarrasch: Hm, one also needs to pass `numinds` and `numel`/`r`. I actually considered that earlier, but found that one would need to pass what feels like dozens of variables. 😄

barucden: Yeah, true. I did not realize that. Well, we could do something like
```julia
_numel(vect::AbstractSparseVector) = nnz(vect)
_numel(vect::AbstractVector) = length(vect)

function _elem_ix(vect::AbstractSparseVector, row, col)
    ix = nonzeroinds(vect)
    return (row .+ ix, col .+ ix)
end
function _elem_ix(vect::AbstractVector, row, col)
    numinds = length(vect)
    return (row+1:row+numinds, col+1:col+numinds)
end

_elems(vect::AbstractSparseVector) = nonzeros(vect)
_elems(vect::AbstractVector) = vect

function spdiagm_internal(kv::Pair{<:Integer,<:AbstractVector}...)
    ncoeffs = 0
    for p in kv
        ncoeffs += _numel(p.second)
    end
    I = Vector{Int}(undef, ncoeffs)
    J = Vector{Int}(undef, ncoeffs)
    V = Vector{promote_type(map(x -> eltype(x.second), kv)...)}(undef, ncoeffs)
    i = 0
    m = 0
    n = 0
    for p in kv
        dia = p.first
        vect = p.second
        numel = _numel(vect)
        if dia < 0
            row = -dia
            col = 0
        elseif dia > 0
            row = 0
            col = dia
        else
            row = 0
            col = 0
        end
        r = 1+i:numel+i
        I[r], J[r] = _elem_ix(vect, row, col)
        copyto!(view(V, r), _elems(vect))

        numinds = length(vect)
        m = max(m, row + numinds)
        n = max(n, col + numinds)
        i += numel
    end
    return I, J, V, m, n
end
```
But this approach introduces two additional "helper" functions so I am not sure if that is a good design...

dkarrasch: Basically, I went with your suggestion of using more helper functions. Thank you!

andreasnoack:
Which part of #37174 isn't addressed here?

dkarrasch:
> Which part of #37174 isn't addressed here?

The wish for a default `spdiagm(v::AbstractVector) = spdiagm(0 => v)`. It's easy to add, actually, but I'm not sure if there is some hidden reason why that hasn't been done earlier.

dkarrasch:
I included the convenience constructor. Originally, without much thinking, I was under the impression that this might be somewhat controversial, but the analogy to `diagm` is straightforward, and I don't think there is any conflict with existing functionality. Is this something we need to announce in NEWS? Seems pretty minor.

andreasnoack:
> The wish for a default `spdiagm(v::AbstractVector) = spdiagm(0 => v)`.

Thanks. I missed that part. I guess the only reason not to have it would be to keep the code slightly simpler by only having a single interface for `spdiagm`. However, I suspect most people would intuitively try calling `spdiagm(v::AbstractVector)` as their first attempt so it might be friendly to include that method.

andreasnoack (approved):

ViralBShah:
Good stuff - Since @andreasnoack has already reviewed, I am happy (without my own review).

Off-topic - I wish we could pick better names (and even move sparse matrices out of stdlib sooner than 2.0)

dkarrasch:
Something went horribly wrong with rebasing. I'll try again with #37684.

PR #37268, opened 2020-08-28
RFC: Change lowering of destructuring to avoid const prop dependence

Keno:
I'm currently doing some work with inference passes that have const
prop (temporarily) disabled and I noticed we actually rely on it
quite a bit for basic things. That's not terrible - const prop works
pretty well after all, but it still imposes a cost and while I want
to support it in my AD use case also, it makes destructuring quite
expensive, because everything needs to be inferred twice. This PR
is an experiment in changing the lowering to avoid having to const
prop the index. Rather than lowering `(a,b,c) = foo()` as:

```
it = foo()
a, s = indexed_iterate(it, 1)
b, s = indexed_iterate(it, 2, s)
c, s = indexed_iterate(it, 3, s)
```

we lower as:

```
it = foo()
iterate, index = index_and_itereate(it)
x = iterate(it)
a = index(x, 1)
y = iterate(it, x)
b = index(y, 2)
z = iterate(it, z)
c = index(z, 3)
```

For tuples `iterate` would simply return the first argument and
`index` would be `getfield`. That way, there is no const prop,
since `getfield` is called directly and inference can directly
use its tfunc. For the fallback case `iterate` is basically
just `Base.iterate`, with just a slight tweak to give an intelligent
error for short iterables.

On simple functions, there isn't much of a difference in execution
time, but benchmarking something more complicated like:
```
function g()
  a, = getfield(((1,),(2.0,3),("x",),(:x,)), Base.inferencebarrier(1))
  nothing
end
```

shows about a 20% improvement in end-to-end inference/optimize time,
which is substantial.


stdlib/Serialization/src/Serialization.jl
```diff
@@ -66,7 +66,7 @@ const TAGS = Any[
     :(=), :(==), :(===), :gotoifnot, :A, :B, :C, :M, :N, :T, :S, :X, :Y, :a, :b, :c, :d, :e, :f,
     :g, :h, :i, :j, :k, :l, :m, :n, :o, :p, :q, :r, :s, :t, :u, :v, :w, :x, :y, :z, :add_int,
     :sub_int, :mul_int, :add_float, :sub_float, :new, :mul_float, :bitcast, :start, :done, :next,
-    :indexed_iterate, :getfield, :meta, :eq_int, :slt_int, :sle_int, :ne_int, :push_loc, :pop_loc,
+    :index_and_iterate, :getfield, :meta, :eq_int, :slt_int, :sle_int, :ne_int, :push_loc, :pop_loc,
```
JeffBezanson, line 69: To be super conservative, we should leave this here and replace a reserved slot with the new symbol instead. Of course, loading code with `indexed_iterate` won't work anyway, but at least it's slightly more compatible.

JeffBezanson:
:+1: to the concept. Shouldn't it be called `iterate_and_index`? :smile:

base/missing.jl
```diff
@@ -69,6 +69,7 @@ convert(::Type{T}, x::T) where {T>:Union{Missing, Nothing}} = x
 convert(::Type{T}, x) where {T>:Missing} = convert(nonmissingtype_checked(T), x)
 convert(::Type{T}, x) where {T>:Union{Missing, Nothing}} = convert(nonmissingtype_checked(nonnothingtype_checked(T)), x)
 
+index_and_iterate(::Missing) = throw(MethodError(iterate, (missing,)))
```
simeonschaub, line 72: I assume this is there for the same reason as the one for `nothing`, but would it make sense to add a comment here as well?

base/namedtuple.jl
```diff
@@ -112,10 +112,10 @@ firstindex(t::NamedTuple) = 1
 lastindex(t::NamedTuple) = nfields(t)
 getindex(t::NamedTuple, i::Int) = getfield(t, i)
 getindex(t::NamedTuple, i::Symbol) = getfield(t, i)
-indexed_iterate(t::NamedTuple, i::Int, state=1) = (getfield(t, i), i+1)
 isempty(::NamedTuple{()}) = true
 isempty(::NamedTuple) = false
 empty(::NamedTuple) = NamedTuple()
+index_and_iterate(t::NamedTuple) = (arg1, getfield)
```
cirocavani, line 118:
```suggestion
iterate_and_index(t::NamedTuple) = (arg1, getfield)
```

simeonschaub:
Bump. Is this good to go?

simeonschaub:
Just bumping this again. :slightly_smiling_face:

Keno:
We decided we may want to get rid of special lowering for destructuring, so we decided against this.

PR #37269, opened 2020-08-28
fix exception stack pop for UndefVarError in catch block

JeffBezanson:
Before:
```
julia> try; error(0); catch; x; end
ERROR: UndefVarError: x not defined
Stacktrace:
```
after:
```
julia> try; error(0); catch; x; end
ERROR: UndefVarError: x not defined
Stacktrace:
 [1] top-level scope
   @ REPL[1]:1

caused by:
0
Stacktrace:
 [1] error(s::Int64)
   @ Base ./error.jl:42
 [2] top-level scope
   @ REPL[1]:1
```


c42f:
Looks good to me.

PR #37273, opened 2020-08-29
more accurate and type-stable sinc/cosc

### LINKED ISSUE #37227: sinc and cosc inaccuracy and type-instability
The main motivation to use sinc function is its nice and smooth behavior around zero. However, the current implementation can produce results greater than 1 close to zero, and the results are not even monotonous. 

``` julia
julia> sinc(1e-20)
1.0
julia> sinc(1e-30)
1.0
julia> sinc(1e-40)
1.0000000000000002
julia> sinc(1e-50)
1.0
julia> sinc(1e-60)
1.0000000000000002
julia> sinc(1e-70)
1.0
julia> sinc(1e-80)
1.0000000000000002
julia> sinc(1e-90)
1.0
```

Possible solution is simple: 

``` julia
sinc(x::Real) = abs(x) < 1e-10 ? one(x) : isinf(x) ? zero(x) : sinpi(x)/(pi*x)
```

However, this is not general enough for various data types, because the constant 1e-10 does not work for more precise data types.

```
Julia Version 1.5.0
Commit 96786e22cc (2020-08-01 23:44 UTC)
Platform Info:
  CPU: Intel(R) Core(TM) i5-8250U CPU @ 1.60GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
```

stevengj:
Fixes #37227.

(Replaces original algorithm from 2010! 1917197353051c65929126b7afd2b0c38dbb93ea)

Backport?

### DIFF
--- a/base/special/trig.jl
+++ b/base/special/trig.jl
@@ -1064,6 +1064,11 @@ than requiring `hypot`.
 fastabs(x::Number) = abs(x)
 fastabs(z::Complex) = abs(real(z)) + abs(imag(z))
 
+# sinc and cosc are zero if the real part is Inf and imag is finite
+isinf_real(x::Real) = isinf(x)
+isinf_real(x::Complex) = isinf(real(x)) && isfinite(imag(x))
+isinf_real(x::Number) = false
+
 """
     sinc(x)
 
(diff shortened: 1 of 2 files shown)


JeffBezanson (approved):

stevengj:
MacOS test failure seems unrelated.

stevengj:
@JeffBezanson, as I mentioned [in the issue](https://github.com/JuliaLang/julia/issues/37227#issuecomment-683295820), I'm curious where you found this definition of the `cosc` function in 2010.  It doesn't seem to be very common; was there a textbook or a paper you were using?

oscardssmith:
cosc definitely seems like one that should have moved out of base for 1.0 (maybe sinc too honestly).

stevengj:
I [added it to ChainRules.jl](JuliaDiff/ChainRules.jl/pull/255) to differentiate `sinc`, so that we'll have to maintain it forever. 😉

PR #37275, opened 2020-08-29
Add a statement-cost printer for analyzing inlining decisions

timholy:
While it now has a doctest, https://docs.julialang.org/en/latest/devdocs/inference/#The-inlining-algorithm-(inline_worthy)-1 has bitrotted more than once. And it's sufficiently laborious that it discourages rapid investigation. This makes printing the costs easy.

Here's a demo:
```julia
julia> Base.print_statement_costs(map, (typeof(sqrt), Tuple{Int}))
map(f, t::Tuple{Any}) in Base at tuple.jl:169
  0 1 ─ %1  = Base.getfield(_3, 1, true)::Int64
  1 │   %2  = Base.sitofp(Float64, %1)::Float64
  2 │   %3  = Base.lt_float(%2, 0.0)::Bool
  0 └──       goto #3 if not %3
  0 2 ─       Base.Math.throw_complex_domainerror(:sqrt, %2)::Union{}
  0 └──       unreachable
 20 3 ─ %7  = Base.Math.sqrt_llvm(%2)::Float64
  0 └──       goto #4
  0 4 ─       goto #5
  0 5 ─ %10 = Core.tuple(%7)::Tuple{Float64}
  0 └──       return %10
```
The numbers in the first column are the inliner's assigned cost.

~~If folks like this, I can add a test to make sure this stays working.~~ (done)

### DIFF
--- a/base/compiler/optimize.jl
+++ b/base/compiler/optimize.jl
@@ -362,31 +362,47 @@ function statement_cost(ex::Expr, line::Int, src::CodeInfo, sptypes::Vector{Any}
     return 0
 end
 
+function statement_or_branch_cost(@nospecialize(stmt), line::Int, src::CodeInfo, sptypes::Vector{Any}, slottypes::Vector{Any}, params::OptimizationParams, throw_blocks::Union{Nothing,BitSet})
+    thiscost = 0
+    if stmt isa Expr
+        thiscost = statement_cost(stmt, line, src, sptypes, slottypes, params,
+                                  params.unoptimize_throw_blocks && line in throw_blocks)::Int
+    elseif stmt isa GotoNode
+        # loops are generally always expensive
+        # but assume that forward jumps are already counted for from
+        # summing the cost of the not-taken branch
+        thiscost = stmt.label < line ? 40 : 0
+    elseif stmt isa GotoIfNot
+        thiscost = stmt.dest < line ? 40 : 0
+    end
+    return thiscost
+end
+
 function inline_worthy(body::Array{Any,1}, src::CodeInfo, sptypes::Vector{Any}, slottypes::Vector{Any},
                        params::OptimizationParams, cost_threshold::Integer=params.inline_cost_threshold)
     bodycost::Int = 0
-    if params.unoptimize_throw_blocks
-        throw_blocks = find_throw_blocks(body)
-    end
+    throw_blocks = params.unoptimize_throw_blocks ? find_throw_blocks(body) : nothing
     for line = 1:length(body)
         stmt = body[line]
-        if stmt isa Expr
-            thiscost = statement_cost(stmt, line, src, sptypes, slottypes, params,
-                                      params.unoptimize_throw_blocks && line in throw_blocks)::Int
-        elseif stmt isa GotoNode
-            # loops are generally always expensive
-            # but assume that forward jumps are already counted for from
-            # summing the cost of the not-taken branch
-            thiscost = stmt.label < line ? 40 : 0
-        elseif stmt isa GotoIfNot
-            thiscost = stmt.dest < line ? 40 : 0
-        else
-            continue
-        end
+        thiscost = statement_or_branch_cost(stmt, line, src, sptypes, slottypes, params, throw_blocks)
         bodycost = plus_saturate(bodycost, thiscost)
         bodycost > cost_threshold && return false
     end
     return true
+end
+
+function statement_costs!(cost::Vector{Int}, body::Vector{Any}, src::CodeInfo, sptypes::Vector{Any}, params::OptimizationParams)
+    throw_blocks = params.unoptimize_throw_blocks ? find_throw_blocks(body) : nothing
+    maxcost = 0
+    for line = 1:length(body)
+        stmt = body[line]
+        thiscost = statement_or_branch_cost(stmt, line, src, sptypes, src.slottypes, params, throw_blocks)
+        cost[line] = thiscost
+        if thiscost > maxcost
+            maxcost = thiscost
+        end
+    end
+    return maxcost
 end
 
 function is_known_call(e::Expr, @nospecialize(func), src, sptypes::Vector{Any}, slottypes::Vector{Any} = empty_slottypes)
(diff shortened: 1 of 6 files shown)


base/compiler/optimize.jl
```diff
@@ -386,6 +386,31 @@ function inline_worthy(body::Array{Any,1}, src::CodeInfo, sptypes::Vector{Any},
...
 end
 
+# Analyzing costs (keep in sync with the above)
+function statement_costs!(cost::Vector{Int}, body::Vector{Any}, src::CodeInfo, sptypes::Vector{Any}, params::OptimizationParams)
+    throw_blocks = find_throw_blocks(body)
+    maxcost = 0
+    for line = 1:length(body)
+        stmt = body[line]
+        thiscost = 0
+        if stmt isa Expr
```
KristofferC, line 396: The reason we can't just factor this whole thing into a function and use it in both places is to avoid dynamic dispatch?

base/reflection.jl
```diff
@@ -1139,6 +1139,45 @@ function return_types(@nospecialize(f), @nospecialize(types=Tuple), interp=Core.
     return rt
 end
 
+"""
+    print_statement_costs(io::IO, f, types)
+
+Print type-inferred and optimized code for `f` given argument types `types`,
+prepending each line with its cost as estimated by the compiler's inlining engine.
+"""
+function print_statement_costs(io::IO, @nospecialize(f), @nospecialize(t); kwargs...)
```
KristofferC, line 1148: Maybe it makes sense to add a `@code_` version of this to InteractiveUtils. Could always be done later of course.

base/compiler/optimize.jl
timholy: Mostly I thought it would be ugly...either you're storing into an array or summing integers. But I agree that for keeping them synced it would be better to combine them, and I guess a couple of helper functions would make it less ugly.

base/reflection.jl
kimikage: No way, is that `@code_costs`?:joy:

KristofferC: Or an argument to one of the others (`cost = true`)

kimikage: FYI, I'm providing [`@code_costs`](https://github.com/kimikage/CodeCosts.jl) and it's exactly @timholy who gave me the motivation to do so. :smile:

I have some objections about the display style, but I agree that Julia supports that feature.

KristofferC: Oh, nice!

base/compiler/optimize.jl
KristofferC: Maybe, it would avoid duplicating the hardcoded 40's though.

base/reflection.jl
timholy: I like your printing quite a lot better. But if you do it as separate code-display (`print(io, src); lines = readlines(io)`) then one risks having the two get misaligned, whereas this version ensures that the two will always stay aligned.

Demo:
```julia
julia> function buildexpr()
           items = [7, 3]
           ex = quote
               X = $items
               for x in X
                   println(x)
               end
           end
           return ex
       end
buildexpr (generic function with 1 method)

julia> @code_typed buildexpr()
CodeInfo(
1 ─ %1  = Core.tuple(7, 3)::Core.Const((7, 3), false)
│   %2  = $(Expr(:foreigncall, :(:jl_alloc_array_1d), Vector{Int64}, svec(Any, Int64), 0, :(:ccall), Vector{Int64}, 2, 2))::Vector{Int64}
│         Base.arraysize(%2, 1)::Int64
└──       goto #7 if not true
2 ┄ %5  = φ (#1 => 1, #6 => %10)::Int64
│   %6  = φ (#1 => 1, #6 => %16)::Int64
│   %7  = φ (#1 => 1, #6 => %17)::Int64
│   %8  = Base.getfield(%1, %6, true)::Int64
│         Base.arrayset(false, %2, %8, %5)::Vector{Int64}
│   %10 = Base.add_int(%5, 1)::Int64
│   %11 = (%7 === 2)::Bool
└──       goto #4 if not %11
3 ─       goto #5
4 ─ %14 = Base.add_int(%7, 1)::Int64
└──       goto #5
5 ┄ %16 = φ (#4 => %14)::Int64
│   %17 = φ (#4 => %14)::Int64
│   %18 = φ (#3 => true, #4 => false)::Bool
│   %19 = Base.not_int(%18)::Bool
└──       goto #7 if not %19
6 ─       goto #2
7 ┄       goto #8
8 ─ %23 = Core._expr(:(=), :X, %2)::Expr
│   %24 = $(Expr(:copyast, :($(QuoteNode(:(for x = X
      #= REPL[4]:6 =#
      println(x)
  end))))))::Expr
│   %25 = Core._expr(:block, $(QuoteNode(:(#= REPL[4]:4 =#))), %23, $(QuoteNode(:(#= REPL[4]:5 =#))), %24)::Expr
└──       return %25
) => Expr
```
You can see statement `%24` took more than one line to print.

kimikage: I just made a change that will make it easier to fix that problem. Perhaps that will be fixed in a few hours. :smiley:

timholy: I switched the order of the printing so the cost comes first...I just needed to look a little harder at how these functions were working. I edited the top post to show the current output.

kimikage: FYI, the unicode characters used in the block brackets may have different widths from the space (U+0020), even if the font is designed as "so-called" monospaced. :scream:

alignment

That' s not just a display issue. For example, when you want to extract an array of the cost numbers from the text, it gets annoying if they are inside the bracket.

So, I like the modified style :heart:

kimikage:
> ```julia
>   0 2 ─       Base.Math.throw_complex_domainerror(:sqrt, %2)::Union{}
> ```
Is this some kind of magic?

timholy:
#35982 and #30222 (the latter should be closed once I split out a couple of nuggets into a separate PR)

kimikage:
What I'm suggesting is the cost "zero". :sweat_smile:

timholy:
I'm not sure I understand your point, but the zero for the `throw` comes from https://github.com/JuliaLang/julia/blob/3285ee09f3a4758e9547c4ed76f0214991e73d94/base/compiler/optimize.jl#L329-L331

The old demo at https://docs.julialang.org/en/v1.6-dev/devdocs/inference/ set `line` to -1 (because the simple approach using `map` doesn't allow it), so it didn't return 0.

PR #37278, opened 2020-08-29
Implement TOML.parse([::Parser,] ::IO)

fredrikekre:
Implement `TOML.parse([::Parser,] ::IO)` and add more tests for `TOML.(try)parse(file)` entrypoints.

### DIFF
--- a/stdlib/TOML/src/TOML.jl
+++ b/stdlib/TOML/src/TOML.jl
@@ -30,10 +30,10 @@ const Parser = Internals.Parser
     parsefile(f::AbstractString)
     parsefile(p::Parser, f::AbstractString)
 
-Parses a file `f` and returns the resulting table (dictionary). Throws a
+Parse file `f` and return the resulting table (dictionary). Throw a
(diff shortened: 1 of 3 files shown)


stdlib/TOML/src/TOML.jl
```diff
@@ -56,29 +56,33 @@ tryparsefile(p::Parser, f::AbstractString) =
...
-[`ParserError`](@ref) upon failure.
+Parse the string `str`, or stream `io`, and return the resulting table (dictionary).
+Return a [`ParserError`](@ref) upon failure.
 
-See also [`TOML.parse`](@ref)
+See also: [`TOML.parse`](@ref)
 """
 tryparse(str::AbstractString) =
     Internals.tryparse(Parser(String(str)))
+tryparse(io::IO) = parse(read(io, String))
```
KristofferC, line 85: `tryparse` on RHS?

fredrikekre: Added tests :)

PR #18777, opened 2016-10-03
Make LinSpace generic

timholy:
This is my proposed alternative to #18492 for fixing the issues with LinSpace. The main attractions of this PR are:
- closes #14420 
- you can now use `linspace` with any type `T` for which `(1-t)*a::T + t*b::T` makes sense.

Given the large amount of work & discussion that has gone into `linspace` and friends with respect to numerical precision (at least #2333, #9637, #18492), it would be remiss not to comment (extensively :wink:) on how this does and does not address previous concerns or implementations. Let me address the issues one-by-one.
## Mind-reading

First and foremost, this implementation does not try to read your mind, whereas the implementation currently in master does. The easiest way to see what I mean is to compare master's behavior:

``` jl
julia> r = linspace(0.0, 0.3, 4)
4-element LinSpace{Float64}:
 0.0,0.1,0.2,0.3

julia> r[2]
0.1
```

with the behavior of this PR:

``` jl
julia> r = linspace(0, 0.3, 4)
4-element LinSpace{Float64}:
 0.0,0.1,0.2,0.3

julia> r[2]
0.09999999999999999
```

Note that, **given the specified endpoints**, that's the correct answer:

``` jl
julia> 0.3/3
0.09999999999999999

julia> Float64(BigFloat(0.3)/3)
0.09999999999999999
```

The version in master returns 0.1 essentially because it interprets the right endpoint 0.3 as `3//10`. This is what I refer to as mind-reading, and while it has some undeniable attractions, it's (1) only applicable to special types (e.g., Float32/64), (2) fairly fragile (it's the origin of #14420), and (3) arguably wrong even in principle (the user said 0.3, not `3//10`).

Note that, now, people who want mind reading can be told to use `Rational`s directly:

``` jl
julia> r = linspace(0, 3//10, 4)
4-element LinSpace{Rational{Int64}}:
 0//1,1//10,1//5,3//10

julia> r[2]
1//10
```
## Precision

All that said, there's still a valid concern about precision: linear interpolation is famously subject to roundoff error. **This implementation returns results with full numeric precision** by using two calls to `fma`. The key function (which other types can specialize) is:

``` jl
# High-precision interpolation. Accurate for t ∈ [0.5,1], so that 1-t is exact.
function lerpi{T}(j::Integer, d::Integer, a::T, b::T)
    t = j/d
    # computes (1-t)*a + t*b
    T(fma(t, b, fma(-t, a, a)))
end
```

[...]

PR #37281, opened 2020-08-29
More accurate Base.lerpi for higher-precision numbers

### LINKED ISSUE #37276: Inaccuracies for `LinRange`
This is a follow-up to discourse thread ["Bug in floating-point range?"](https://discourse.julialang.org/t/bug-in-floating-point-range/45706?u=ffevotte).

The most minimal example I could come up with to evidence the issue is the following one:
```julia
function test_range(T)
    @show r = range(T(0), stop=T(1), length=11)
    @show val = r[2]
    ref = 1//10
    err = (Rational{BigInt}(val) - ref) / ref  |> abs |> Float64
    @show err
end
```

Indexing into a `LinRange{BigFloat}` gives very accurate results, with errors in the order of ulp(BigFloat):
```julia
julia> test_range(BigFloat);
r = range(T(0), stop = T(1), length = 11) = range(0.0, stop=1.0, length=11)
val = r[2] = 0.1000000000000000000000000000000000000000000000000000000000000000000000000000002
err = 2.1590421387736112e-78
```

However, the same calculation performed on a `LinRange{Complex{BigFloat}}` is much less accurate, with errors in the same order of magnitude as if 64-bit floats had been used:
```julia
julia> test_range(Complex{BigFloat});
r = range(T(0), stop = T(1), length = 11) = range(0.0 + 0.0im, stop=1.0 + 0.0im, length=11)
val = r[2] = 0.1000000000000000055511151231257827021181583404541015625 + 0.0im
err = 5.551115123125783e-17

julia> test_range(Complex{Float64});
r = range(T(0), stop = T(1), length = 11) = range(0.0 + 0.0im, stop=1.0 + 0.0im, length=11)
val = r[2] = 0.1 + 0.0im
err = 5.551115123125783e-17
```

<br/>

As mentioned in the discourse thread above, I think I have identified `Base.lerpi` as the origin of the issue:
```julia
julia> Base.lerpi(1, 10, big"0.", big"1.")
0.1000000000000000000000000000000000000000000000000000000000000000000000000000002

julia> Base.lerpi(1, 10, Complex(big"0."), Complex(big"1.")) |> real
0.1000000000000000055511151231257827021181583404541015625
```
and I think I have a fix for this. I'll try and propose a PR as soon as I can.

ffevotte:
This should fix #37276.

Examples of things that are more accurate with this PR include:
```
julia> LinRange(Complex{BigFloat}(0), Complex{BigFloat}(1), 11)[2]
0.1000000000000000000000000000000000000000000000000000000000000000000000000000002 + 0.0im

julia> LinRange([0//1], [1//1], 11)[2]
1-element Vector{Rational{Int64}}:
 1//10
```
in comparison to the current master:
```
julia> LinRange(Complex{BigFloat}(0), Complex{BigFloat}(1), 11)[2]
0.1000000000000000055511151231257827021181583404541015625 + 0.0im

julia> LinRange([0//1], [1//1], 11)[2]
1-element Array{Rational{Int64},1}:
 3602879701896397//36028797018963968
```

I think that the proposed implementation of a generic `Base.lerpi` makes it less useful to have a specific method for rationals (such as defined in [rational.jl:467](https://github.com/JuliaLang/julia/blob/master/base/rational.jl#L467)).

One potential downside to the way things are done in this PR, is that there could potentially be a loss of accuracy w.r.t how things are currently done, when working on ranges of small (i.e. less than 64 bits) floating-point numbers. I'm not sure whether that was intended, but in the current implementation, the `j/d` quotient usually results in `t` being a `Float64`, meaning that computations involving `a` and `b` will be promoted to `Float64` too if the end points are smaller FP numbers. With the current implementation, all calculations use the same precision as the range end points.

It would not be too difficult to restore the current master behavior for smaller FP number types, but I'm not sure whether it is worth making the implementation more complicated. In particular, I have no idea how many users expect and rely on the fact that points in a `LinRange{Float16}` are computed in double precision before being rounded to half precision. Would somebody have some insight on this?

### DIFF
--- a/base/range.jl
+++ b/base/range.jl
@@ -685,7 +685,7 @@ end
 
 function lerpi(j::Integer, d::Integer, a::T, b::T) where T
     @_inline_meta
-    t = j/d
+    t = eltype(T)(j)/d
     T((1-t)*a + t*b)
 end
 
--- a/test/ranges.jl
+++ b/test/ranges.jl
@@ -1326,6 +1326,15 @@ end
     @test r[4] === 3.0+im
 end
 
+@testset "issue #37276" begin
+    r = range(Complex{BigFloat}(0), stop=Complex{BigFloat}(1), length=4)
+    @test isa(@inferred(r[2]), Complex{BigFloat})
+    @test r[1] == 0.
+    @test r[2] ≈ big"1"/3   rtol=10eps(BigFloat)
+    @test r[3] ≈ big"2"/3   rtol=10eps(BigFloat)
+    @test r[4] == 1.
+end
+
 # ambiguity between (:) methods (#20988)
 struct NotReal; val; end
 Base.:+(x, y::NotReal) = x + y.val


test/ranges.jl
```diff
@@ -1326,6 +1326,15 @@ end
     @test r[4] === 3.0+im
 end
 
+@testset "issue #37276" begin
+    r = range(Complex{BigFloat}(0), stop=Complex{BigFloat}(1), length=4)
+    @test isa(@inferred(r[2]), Complex{BigFloat})
+    @test r[1] == 0.
+    @test r[2] ≈ big"1"/3
```
simeonschaub, line 1333: Is this perhaps a bit too loose? Would it make sense to set a lower tolerance here?

ffevotte: Are you referring to the tolerance of `isapprox`?
AFAIU, the default tolerance used by `isapprox` is given by the square root of `eps` of the types of the operands, which we know to be `BigFloat`s here. In this case, this means a relative tolerance of approximately 4e-39 (which makes everything fail if any part of the computation is performed in double precision).

But you're right that it is much higher than the few ulps that we could expect; we could very well set this to something like 1e-75 and still be safe. TBH, the reason why I kept the default tolerance parameters is because this seems to be how it's done for all other approximate equality tests in `test/ranges.jl`.

TL;DR: I don't have any strong opinion about this. I can very well change this tolerance to any value deemed more appropriate.

simeonschaub: Yes, that's what I meant. I think it's better to be more conservative with tests like these, because it's always better to discover regressions early. Perhaps the other tests should be more strict as well, but that's probably a separate discussion.

base/range.jl
```diff
@@ -685,7 +685,7 @@ end
 
 function lerpi(j::Integer, d::Integer, a::T, b::T) where T
     @_inline_meta
-    t = j/d
+    t = eltype(T)(j)/d
```
JeffBezanson, line 688: Why does this use `eltype` instead of just `T(j)`?

ffevotte: This was required to satisfy the tests at https://github.com/JuliaLang/julia/blob/master/test/ranges.jl#L1283-L1290
In short, it ensures that things like this work:
```
julia> LinRange([0., 10.], [1., 12.], 4)
4-element LinRange{Array{Float64,1}}:
 [0.0, 10.0],[0.333333, 10.666667],[0.666667, 11.33333],[1.0, 12.0]
```

(I actually discovered such uses were possible when tests failed with an earlier version of my fix!)

test/ranges.jl
ffevotte: OK. I agree that more strictness is always a good thing in tests. I amended to commit to include a relative tolerance of `10eps(BigFloat)`, which I guess makes the intent explicit (although in this particular case results are exactly rounded, I think it is more future-proof to not rely on this and allow an error of a few ulps, which should correspond to the expectation users should have in the general case)

StefanKarpinski:
I believe @timholy originally introduced this for accurate linear interpolation of ranges, so he may have some idea.

ffevotte:
> I believe @timholy originally introduced this for accurate linear interpolation of ranges, so he may have some idea.

Thanks. I indeed found #18777 where `Base.lerpi` seems to have appeared. The entire discussion is well worth a read and very relevant here but, although a few (sophisticated) implementations of `lerpi` are discussed there, I haven't found much about the generic implementation that was chosen in the end.

I'd love to hear about the details if someone happens to remember them.

base/range.jl
timholy: A problem is that the current implementation works even for objects that don't define `eltype(T)`: basically it just needs an algebra supporting addition and multiplication by reals.

One option would be `t = j // d` but I will wager you that there will be overflow test failures somewhere. Another option would be `t = oftype(one(T), j/d)` but amazingly even `one(::Type{Vector{T}}) where T = one(T)` isn't defined. (Why not? That seems crazy.)

timholy: On balance perhaps the rational approach makes the most sense.

timholy:
> In particular, I have no idea how many users expect and rely on the fact that points in a LinRange{Float16} are computed in double precision before being rounded to half precision. Would somebody have some insight on this?

I don't have any useful insight. I doubt it would be a problem but I am not sure.

base/range.jl
ffevotte: > A problem is that the current implementation works even for objects that don't define `eltype(T)`: basically it just needs an algebra supporting addition and multiplication by reals.

That's true. Are there any examples of such types in the stdlib? If we want to support such uses, I'd like to add a specific test case for them (but I'd like to avoid defining a new type only for this purpose if possible).

ffevotte: > One option would be `t = j // d` but I will wager you that there will be overflow test failures somewhere.

What kind of issues are you expecting? Overflows when `j` or `d` are huge?
It looks to me like, in your proposal, the only arithmetic operation ever performed with rational is `(1-t)`, which is merely `(d-j)//d`. I might be missing something, but I wouldn't expect any issues with that.

There might be a small performance impact though, since `lerpi` would now perform two divisions: one to compute `t`, and one to compute `1-t`. Is that a problem? Would you like me to benchmark this?

In any case, `t = j // d` seems to pass all tests (but maybe we're missing specific test cases that would trigger issues with this implementation, which is why I'm asking about anticipated issues)

timholy: @StefanKarpinski has this annoying tendency to write tests that include the extreme ranges (joking, if he doesn't do that you can be sure that some prankster will report it as an issue).
```julia
julia> start, stop = 1, typemax(Int)
(1, 9223372036854775807)

julia> t = 3//4
3//4

julia> (1 - t) * start + t * stop
ERROR: OverflowError: 3 * 9223372036854775807 overflowed for type Int64
Stacktrace:
 [1] throw_overflowerr_binaryop(op::Symbol, x::Int64, y::Int64)
   @ Base.Checked ./checked.jl:154
 [2] checked_mul
   @ ./checked.jl:288 [inlined]
 [3] *(x::Rational{Int64}, y::Int64)
   @ Base ./rational.jl:314
 [4] top-level scope
   @ REPL[3]:1
```

timholy: Interestingly, though, if we reversed the order of operation (divide first, multiply second) then it would be fine---though then you should check underflow for ranges like one going from `eps(0.0)` to `3*eps(0.0)`. If you can find a robust solution, it almost seems worth creating a `SmallRatio` type for this (example: https://github.com/timholy/Ratios.jl).

ffevotte: Thanks, I was so focused on FP issues that hadn't thought about integer ranges.

One thing that we're able to handle in the current master is the following (which hits overflow issues with the `j//d` implementation as you mentioned):
```
julia> LinRange{Int}(1, 1+2^62, 5)
5-element LinRange{Int64}:
 1,1152921504606846976,2305843009213693952,3458764513820540928,4611686018427387904
```

(I'll note that it's currently not possible to handle `LinRange{Int}(1, typemax(Int), 3)` because `lerpi` uses FP numbers and `typemax(Int)` is not representable as a `Float64`)

<br/>

In any case, in light of this discussion, I'll try and think a bit more before proposing a new implementation that hopefully meets all envisioned use cases.

timholy: One last issue to pay attention to is performance. I think it will be slower the more divisions you do.

ffevotte: Yes, that's what I meant to say in my message above:

> There might be a small performance impact though, since `lerpi` would now perform two divisions: one to compute `t`, and one to compute `1-t`. Is that a problem? Would you like me to benchmark this?

I'll try and find an adequate compromise between all concerns involved in this.

DilumAluthge:

@ffevotte If you're still interested in working on this PR, let me know, and I can re-open it.

PR #37283, opened 2020-08-30
expose findfirst findnext for UInt8 vector

### LINKED ISSUE #37280: expose findfirst(pattern::AbstractVector{T}, A::AbstractVector{T}) where {T<:Union{Int8,UInt8}}
We already have a function `_searchindex` to [search for subsequences of byte arrays](https://github.com/JuliaLang/julia/blob/3f7b556bb3b45cfcb2c718e3e5e0f1d912e4272d/base/strings/search.jl#L177-L236).  It would be nice to expose this by adding corresponding methods to `findfirst` and `findnext`.

Although the current `_seardhindex` implementation only accepts `Vector` arguments, it looks like the implementation doesn't depend on the storage format, so the function signature could be easily widened to `AbstractVector{<:Union{Int8,UInt8}}` arguments.    (You'll need to replace the `sizeof` calls with `length`, and maybe use `firstindex` to support non 1-based indexing.)

(See also the [discussion on discourse](https://discourse.julialang.org/t/finding-position-of-a-sequence-in-a-uint8-array/45757).)

Moelf:
fix #37280


KristofferC:
Should these be moved away from the string file now?

base/strings/search.jl
```diff
@@ -174,9 +194,12 @@ function _searchindex(s::String, t::String, i::Integer)
     _searchindex(unsafe_wrap(Vector{UInt8},s), unsafe_wrap(Vector{UInt8},t), i)
 end
 
-function _searchindex(s::ByteArray, t::ByteArray, i::Integer)
-    n = sizeof(t)
-    m = sizeof(s)
+function _searchindex(s::AbstractVector{T},
+                      t::AbstractVector{T},
+                      i::Integer) where T <:Union{Int8,UInt8}
+
```
nalimilan, line 200:
```suggestion
```

test/strings/search.jl
```diff
@@ -408,3 +408,13 @@ for T = (UInt, BigInt)
...
+
+# issue 37280
+let A = [0x40, 0x52, 0x62, 0x52, 0x62]
+    @test findfirst([0x99], A) === nothing
+    @test findfirst([0x52], A) == 2:2
+    @test findfirst([0x52, 0x62], A) == 2:3
+    @test findnext([0x52, 0x62], A, 2) == 2:3
+    @test findnext([0x52, 0x62], A, 3) == 4:5
+    @test findnext([0x52, 0x62], A, 5) === nothing
+end
```
nalimilan, line 420: Can you also test passing out of bounds indices, both too small and too large? These are easy to get wrong.

Please also check what happens with empty arguments (both).

test/strings/search.jl
```diff
@@ -408,3 +408,13 @@ for T = (UInt, BigInt)
...
     end
 end
+
+# issue 37280
+let A = [0x40, 0x52, 0x62, 0x52, 0x62]
+    @test findfirst([0x99], A) === nothing
+    @test findfirst([0x52], A) == 2:2
+    @test findfirst([0x52, 0x62], A) == 2:3
+    @test findnext([0x52, 0x62], A, 2) == 2:3
+    @test findnext([0x52, 0x62], A, 3) == 4:5
```
nalimilan, line 418: This is better for ranges as it allows testing the bounds of empty ranges too:
```suggestion
    @test findfirst([0x52], A) === 2:2
    @test findfirst([0x52, 0x62], A) === 2:3
    @test findnext([0x52, 0x62], A, 2) === 2:3
    @test findnext([0x52, 0x62], A, 3) === 4:5
```

test/strings/search.jl
```diff
@@ -408,3 +408,13 @@ for T = (UInt, BigInt)
         @test findprev(isletter, astr, T(x)) isa Int
     end
 end
+
+# issue 37280
```
nalimilan, line 412: Better create a `@testset` and give it an explicit name (this is a feature, not just a bug). Also use a loop to test both `Int8` and `UInt8`, including mixing the two in each argument.

nalimilan (commented):
Thanks! This makes sense. I have a few remarks:
- It sounds like this method should eventually work for any `AbstractVector`. The fact that only `UInt8`/`Int8` are supported is due to an implementation detail. It's probably OK to start with a restricted method though, and generalize it later.
- We should measure the consequences of this addition for the API if we want to generalize this.  With this PR, `findfirst(pattern, collection)` will mean "look for the *sequence* `pattern` in `collection`". Currently, we have `findfirst(pred::Function, A)`, `findfirst(ch::AbstractChar, string::AbstractString)` and `findfirst(pattern::AbstractString, string::AbstractString)`. The methods added by this PR are consistent with the third method, i.e. the two arguments collections, and we are looking for a sequence. Do note that it is at odds with the second one, where the first argument is a single element. This is OK for strings as there's no ambiguity, but in general with `findfirst(x::AbstractArray, A::AbstractArray)` you cannot know whether `x` is supposed to be a sequence or an element since `A` could be an `AbstractArray{<:AbstractArray}`. So if we go with this PR, we will renounce to the possibility to treat `x` as a single element, e.g. we'll have to keep writing `findfirst(==(1), [1, 2])`. I think that's OK since the syntax is relatively short, but let's be aware of this decision.
- Finally, `findlast` and `findprev` should probably be implemented in this PR for consistency (I think `_rsearch` allows doing this).

base/strings/search.jl
```diff
@@ -194,7 +217,7 @@ function _searchindex(s::ByteArray, t::ByteArray, i::Integer)
     bloom_mask = UInt64(0)
     skip = n - 1
     tlast = _nthbyte(t,n)
```
stevengj, line 219: The `_nthbyte(t,index)` calls can just be `t[index]` — the `_nthbyte` function is only needed if `t` might be a string instead of an array.

base/strings/search.jl
```diff
@@ -194,7 +217,7 @@ function _searchindex(s::ByteArray, t::ByteArray, i::Integer)
...
-    for j in 1:n
+    for j in firstindex(s):n
```
stevengj, line 220: This doesn't look right: `n` may not be the last index.  Make sure you test this function where `s` and/or `t` are offset arrays.

base/strings/search.jl
```diff
@@ -174,9 +194,12 @@ function _searchindex(s::String, t::String, i::Integer)
     _searchindex(unsafe_wrap(Vector{UInt8},s), unsafe_wrap(Vector{UInt8},t), i)
 end
 
-function _searchindex(s::ByteArray, t::ByteArray, i::Integer)
-    n = sizeof(t)
-    m = sizeof(s)
+function _searchindex(s::AbstractVector{T},
+                      t::AbstractVector{T},
```
stevengj, line 198: Probably `AbstractVector{<:Union{Int8,UInt8}}`, so that `s` and `t` don't have to have the same element type.

base/strings/search.jl
```diff
@@ -296,6 +319,28 @@ julia> findnext('o', "Hello to the world", 6)
...
+```jldoctest
+julia> findnext([0x52, 0x62], [0x52, 0x62, 0x72], 5) === nothing
+true
+
+julia> findnext([0x52, 0x62], [0x40, 0x52, 0x62, 0x52, 0x62], 3)
+4:5
+```
+"""
+function findnext(pattern::AbstractVector{T},
+                  A::AbstractVector{T}, ind::Integer) where T<:Union{Int8,UInt8}
```
stevengj, line 340: Maybe call it `start` rather than `ind`?

base/strings/search.jl
```diff
@@ -296,6 +319,28 @@ julia> findnext('o', "Hello to the world", 6)
...
+# Examples
+```jldoctest
+julia> findnext([0x52, 0x62], [0x52, 0x62, 0x72], 5) === nothing
+true
+
+julia> findnext([0x52, 0x62], [0x40, 0x52, 0x62, 0x52, 0x62], 3)
+4:5
+```
+"""
+function findnext(pattern::AbstractVector{T},
```
stevengj, line 339: `AbstractVector{<:Union{Int8,UInt8}}`, similar to above.

stevengj:
> So if we go with this PR, we will renounce to the possibility to treat x as a single element

Not as long as this is restricted to a small set of element types.    I agree that this will come up in a hypothetical future PR where we generalize this to `find(pattern::AbstractArray, a::AbstractArray)`, however, but we could discuss that question in a future issue.

base/strings/search.jl
stevengj: Alternatively, we could define:
```
_nthbyte(t::AbstractVector,index) = t[index + (1 - firstindex(t))]
```
to allow us to use 1-based indexing (get rid of `firstindex` below and disregard my comment about `lastindex`).  This might be simpler, and there should be no runtime cost for `Vector` since the compiler should be able to eliminate the ` + (1 - firstindex(t))` operation.

test/strings/search.jl
Moelf: too small fails in other situation:
```
julia> findnext('z', "Hello to the world", -3)
ERROR: BoundsError: attempt to access String
  at index [-3]
```
I will add the > out of bounds test

nalimilan: Yeah but you need to test that error to be sure it's there. :-)

Moelf:
As suggested by @stevengj , shrieking original `_nthbyte` and define a general version for `AbstractVector` to minimize the changes for `_searchindex`. Added tests for `OffsetArray`. @KristofferC would you suggest a proper place for these code since it's no longer just `String`, I'm thinking `base/array.jl` maybe

base/strings/search.jl
```diff
@@ -166,35 +188,40 @@ function _search_bloom_mask(c)
...
+    if w < 0 || i - f_s > w
         return 0
     end
 
     bloom_mask = UInt64(0)
-    skip = n - 1
+    skip = n - f_s
     tlast = _nthbyte(t,n)
-    for j in 1:n
+    for j in eachindex(t)
```
stevengj, line 224: This isn't enough, because the `n - j - 1` computation below (and similar, like `if j == n-1`) assumes 1-based indexing.

You'll need to check all of the index calculations and fix them to use firstindex and lastindex.   Alternatively, shift the indexing calculations in nthbyte as I suggested in another comment.

Moelf: I have changed the `_nthbyte` definition thus only need to consider offset at the start and when returning the result

Moelf:
@nalimilan I have added `findlast` and `findprev` and their tests now, thanks for the comment. Yes, I was surprised when I first encountered that, for example, `findfirst(3, [1,2,3)` doesn't work. But I think it is fine since `==(x)` is a nice and explicit way of saying "treat `x` as a single element".

base/strings/search.jl
stevengj: In that case you should loop for `j = 1:n`

base/strings/search.jl
```diff
@@ -166,35 +188,40 @@ function _search_bloom_mask(c)
 end
 
 _nthbyte(s::String, i) = codeunit(s, i)
-_nthbyte(a::Union{AbstractVector{UInt8},AbstractVector{Int8}}, i) = a[i]
+_nthbyte(a::Union{Vector{UInt8},Vector{Int8}}, i) = a[i]
```
stevengj, line 191: You can eliminate this definition — it is redundant with the one below based on `firstindex(t)`, and LLVM is smart enough to eliminate the extra additions when `firstindex(t)` is inlined to `1` at compile-time.

base/strings/search.jl
```diff
@@ -166,35 +188,40 @@ function _search_bloom_mask(c)
...
+_nthbyte(t::AbstractVector, index) = t[firstindex(t) + (index-1)]
```
stevengj, line 192: I would normally write this as
```jl
t[index + (firstindex(t)-1)]
```
Either way, if `firstindex(t)` is a compile-time constant (as is the case for typical Julia arrays), then LLVM will simplify the indexing calculations.    But if `firstindex(t)` is *not* a compile-time constant, then separating out the index-independent calculation `firstindex(t)-1` might make it easier for the compiler to perform common-subexpression elimination.

base/strings/search.jl
```diff
@@ -166,35 +188,40 @@ function _search_bloom_mask(c)
...
-function _searchindex(s::ByteArray, t::ByteArray, i::Integer)
-    n = sizeof(t)
-    m = sizeof(s)
+function _searchindex(s::AbstractVector{<:Union{Int8,UInt8}},
+                      t::AbstractVector{<:Union{Int8,UInt8}},
+                      i::Integer)
+    n = length(t)
+    m = length(s)
+    f_s = firstindex(s)
+    i < f_s && throw(BoundsError(s, i))
```
stevengj, line 206: Since you are doing 1-based index calculations below, you should shift `i` to a 1-based index.    I would declare the parameter as `_i::Integer`, and then do:
```jl
i = Int(_i) - (firstindex(s) - 1)
i < 1 && throw(BoundsError(s, _i))
```
to shift `i` to a `1`-based index for use with `nthbyte` and your other 1-based calculations.

base/strings/search.jl
```diff
@@ -166,35 +188,40 @@ function _search_bloom_mask(c)
...
 
     w = m - n
-    if w < 0 || i - 1 > w
+    if w < 0 || i - f_s > w
         return 0
     end
 
     bloom_mask = UInt64(0)
-    skip = n - 1
+    skip = n - f_s
```
stevengj, line 222: Should stay as `n - 1` since you are doing 1-based index calculations.

base/strings/search.jl
```diff
@@ -215,7 +242,7 @@ function _searchindex(s::ByteArray, t::ByteArray, i::Integer)
 
             # match found
             if j == n - 1
-                return i+1
+                return i+f_s
```
stevengj, line 245: should stay as `i+1` since you are doing 1-based index calculations.

base/strings/search.jl
```diff
@@ -387,27 +452,29 @@ function _rsearchindex(s::String, t::String, i::Integer)
...
 end
 
-function _rsearchindex(s::ByteArray, t::ByteArray, k::Integer)
-    n = sizeof(t)
-    m = sizeof(s)
+function _rsearchindex(s::AbstractVector{<:Union{Int8,UInt8}}, t::AbstractVector{<:Union{Int8,UInt8}}, k::Integer)
+    n = length(t)
+    m = length(s)
+    f_s = firstindex(s)
+    k < f_s && throw(BoundsError(s, k))
```
stevengj, line 459: Same as above: since you are using `_nthbyte` to do 1-based index calculations, you should just shift `k` by `firstindex(s)` and then leave the rest of the calculations unchanged.

i.e. get rid of all the `f_s` changes.

base/strings/search.jl
```diff
@@ -387,27 +452,29 @@ function _rsearchindex(s::String, t::String, i::Integer)
...
-    if w < 0 || k <= 0
+    if w < 0 || k <= f_s
         return 0
     end
 
     bloom_mask = UInt64(0)
     skip = n - 1
     tfirst = _nthbyte(t,1)
-    for j in n:-1:1
+    for j in reverse(eachindex(t))
```
stevengj, line 477: Should stay as `n:-1:1` since you are doing 1-based indexing with `_nthbyte`.

base/strings/search.jl
```diff
@@ -428,7 +495,7 @@ function _rsearchindex(s::ByteArray, t::ByteArray, k::Integer)
 
             # match found
             if j == n
-                return i
+                return i + f_s - 1
```
stevengj, line 498: stays as `return i`

base/strings/search.jl
Moelf: not I think the returned index here refers to the actual index you want to use for a offset array

test/strings/search.jl
```diff
@@ -390,6 +390,34 @@ s_18109 = "fooα🐨βcd3"
...
+        @test findlast(VT[0x52], A) === 4:4
+
+        pattern = VT[0x52, 0x62]
+
+        @test findfirst(pattern, A) === 2:3
+        @test findnext(pattern, A, 2) === 2:3
+        @test findnext(pattern, A, 3) === 4:5
+        @test findnext(pattern, A, 5) === nothing
+        @test findnext(pattern, A, 99) === nothing
+        @test_throws BoundsError findnext(pattern, A, -3)
```
stevengj, line 410: We should also check for `BoundsError` if the starting index is `> length(A) + 1`.   (*not* `> length(A)` … a starting index 1 past the end of the array seems to be allowed in `findnext`.)

Moelf:
Took the suggested approach by @stevengj : shifting `_i` by `i = Int(_i) - (firstindex(s) - 1)` inside `_searchindex` and `_rsearchindex` to keep most of the logic unchanged. In any case please let me know if more edge case tests are needed.

base/strings/search.jl
```diff
@@ -166,17 +188,21 @@ function _search_bloom_mask(c)
...
-function _searchindex(s::ByteArray, t::ByteArray, i::Integer)
-    n = sizeof(t)
-    m = sizeof(s)
+function _searchindex(s::AbstractVector{<:Union{Int8,UInt8}},
+                      t::AbstractVector{<:Union{Int8,UInt8}},
+                      _i::Integer)
+    n = length(t)
+    m = length(s)
+    i = Int(_i) - (firstindex(s) - 1)
+    i < 1 && throw(BoundsError(s, _i))
```
stevengj, line 205: Should this be `(i < 1 || i > m+1)`?

base/strings/search.jl
```diff
@@ -293,8 +320,33 @@ julia> findnext('o', "Hello to the world", 6)
...
+
+julia> findnext([0x52, 0x62], [0x40, 0x52, 0x62, 0x52, 0x62], 3)
+4:5
+```
+"""
+function findnext(pattern::AbstractVector{<:Union{Int8,UInt8}},
+                  A::AbstractVector{<:Union{Int8,UInt8}},
+                  start::Integer)
+    (start == (lastindex(A)+1)) && return nothing
+    (start > (lastindex(A)+1)) && throw(BoundsError(A, start))
```
stevengj, line 347: Maybe this check should be in `_searchindex` along with the `i < 1` check?

base/strings/search.jl
```diff
@@ -428,7 +499,7 @@ function _rsearchindex(s::ByteArray, t::ByteArray, k::Integer)
 
             # match found
             if j == n
-                return i
+                return i - 1 + firstindex(s)
```
stevengj, line 502: `i + (firstindex(s) - 1)` just to be stylistically consistent

base/strings/search.jl
stevengj: See also below.

Presumably there are also bounds checks in the `String` case; it would be nice to combine the two bounds checks into `_searchindex`  if possible.

Moelf: the difficulty is that when `String` is not `ASCII`, the bounds is difficult to check but I will give it a try

stevengj: I don't think it matters — this is used for `String`, where the indices are code units (i.e. bytes), and the current behavior is for `findnext` is to throw `BoundsError` for indices `> m+1` (i.e. the indices can be one byte past the end of the UTF-8 data but not farther).

e.g. `findnext("a", "α", 3)` is allowed but `findnext("a", "α", 4)` throws an error.

So `(i < 1 || i > m+1)` should be correct for `String` searches as well.

Moelf: ah, not quite, because eventually the string version boils down to:
https://github.com/JuliaLang/julia/blob/41e603e5836bebf447f14c82fb2671b3567fc3bf/base/strings/search.jl#L85

I can add a check there maybe?

Moelf:
```
julia> findprev("dha", "askdha", 100)
4:6
```
found out this is actually valid......time to re-write test

Moelf:
actually run into a bit problem here, if we want to support `OffsetArray`, we need to stop using `0` as sentinel value in `_searchindex` and `_rsearchindex`, I tried to replace it with `nothing` but then I get error during compilation:
```
error during bootstrap:
LoadError("sysimg.jl", 19, LoadError("/home/akako/Documents/github/julia/usr/share/julia/stdlib/v1.6/Pkg/src/Pkg.jl", 3, LoadError("/home/akako/Documents/github/julia/usr/share/julia/stdlib/v1.6/Pkg/src/BinaryPlatforms.jl", 1, ArgumentError("Unsupported libgfortran '1.0.0'"))))
```

Moelf:
after some struggling I realized it's impossible to handle `OffsetArray` because `_searchindex` uses `0` as sentinel value and it cannot be handled in the same function that handles string. For now I have made it `require_one_based_indexing`.

base/strings/search.jl
nalimilan: @goretkin Does this sound right to you given what we discussed at #36967?

base/strings/search.jl
```diff
@@ -123,6 +123,27 @@ true
 """
 findfirst(ch::AbstractChar, string::AbstractString) = findfirst(==(ch), string)
 
+"""
+    findfirst(pattern::AbstractVector{<:Union{Int8,UInt8}},
+    A::AbstractVector{<:Union{Int8,UInt8}})
```
nalimilan, line 128: Align with `pattern`.

base/strings/search.jl
```diff
@@ -123,6 +123,27 @@ true
...
+```jldoctest
+julia> findfirst([0x52, 0x62], [0x40, 0x52, 0x62, 0x63])
+2:3
+```
+"""
+findfirst(pattern::AbstractVector{<:Union{Int8,UInt8}},
+          A::AbstractVector{<:Union{Int8,UInt8}})=
+    _search(A, pattern, firstindex(A))
+
+
```
nalimilan, line 145:
```suggestion
```

base/strings/search.jl
```diff
@@ -293,8 +320,30 @@ julia> findnext('o', "Hello to the world", 6)
 8
 ```
 """
-findnext(ch::AbstractChar, string::AbstractString, ind::Integer) =
-    findnext(==(ch), string, ind)
+findnext(ch::AbstractChar, string::AbstractString, start::Integer) =
+    findnext(==(ch), string, start)
+
+"""
+    findnext(pattern::AbstractVector{<:Union{Int8,UInt8}}, A::AbstractVector{<:Union{Int8,UInt8}}, start::Integer)
```
nalimilan, line 327:
```suggestion
    findnext(pattern::AbstractVector{<:Union{Int8,UInt8}},
             A::AbstractVector{<:Union{Int8,UInt8}},
             start::Integer)
```

base/strings/search.jl
```diff
@@ -314,6 +363,23 @@ julia> findfirst("Julia", "JuliaLang")
 findlast(pattern::AbstractString, string::AbstractString) =
     findprev(pattern, string, lastindex(string))
 
+"""
+    findlast(pattern::AbstractVector{<:Union{Int8,UInt8}},
+    A::AbstractVector{<:Union{Int8,UInt8}})
```
nalimilan, line 368: Same.

base/strings/search.jl
```diff
@@ -314,6 +363,23 @@ julia> findfirst("Julia", "JuliaLang")
...
+[`findprev(pattern, A, lastindex(A))`](@ref).
+
+# Examples
+```jldoctest
+julia> findlast([0x52, 0x62], [0x52, 0x62, 0x52, 0x62])
+3:4
+```
+"""
+findlast(pattern::AbstractVector{<:Union{Int8,UInt8}},
+         A::AbstractVector{<:Union{Int8,UInt8}})=
```
nalimilan, line 380:
```suggestion
         A::AbstractVector{<:Union{Int8,UInt8}}) =
```

base/strings/search.jl
```diff
@@ -293,8 +320,30 @@ julia> findnext('o', "Hello to the world", 6)
...
+```jldoctest
+julia> findnext([0x52, 0x62], [0x52, 0x62, 0x72], 3) === nothing
+true
+
+julia> findnext([0x52, 0x62], [0x40, 0x52, 0x62, 0x52, 0x62], 3)
+4:5
+```
+"""
+findnext(pattern::AbstractVector{<:Union{Int8,UInt8}},
+                  A::AbstractVector{<:Union{Int8,UInt8}},
```
nalimilan, line 344: Same. Also add space before `=`.

base/strings/search.jl
```diff
@@ -448,9 +517,9 @@ function _rsearchindex(s::ByteArray, t::ByteArray, k::Integer)
     0
 end
 
-function _rsearch(s::Union{AbstractString,ByteArray},
-                  t::Union{AbstractString,AbstractChar,Int8,UInt8},
-                  i::Integer)
+function _rsearch(s::Union{AbstractString,AbstractVector{<:Union{Int8,UInt8}}},
+                 t::Union{AbstractString,AbstractChar,AbstractVector{<:Union{Int8,UInt8}}},
```
nalimilan, line 521: Same.

base/strings/search.jl
```diff
@@ -503,9 +572,27 @@ julia> findprev('o', "Hello to the world", 18)
 15
 ```
 """
-findprev(ch::AbstractChar, string::AbstractString, ind::Integer) =
-    findprev(==(ch), string, ind)
+findprev(ch::AbstractChar, string::AbstractString, start::Integer) =
+    findprev(==(ch), string, start)
+
+"""
+    findprev(pattern::AbstractVector{<:Union{Int8,UInt8}}, A::AbstractVector{<:Union{Int8,UInt8}}, start::Integer)
```
nalimilan, line 579:
```suggestion
    findprev(pattern::AbstractVector{<:Union{Int8,UInt8}},
             A::AbstractVector{<:Union{Int8,UInt8}},
             start::Integer)
```

test/strings/search.jl
```diff
@@ -390,6 +390,36 @@ s_18109 = "fooα🐨βcd3"
     @test findall("aa", "aaaaaa", overlap=true) == [1:2, 2:3, 3:4, 4:5, 5:6]
 end
 
+# issue 37280
+@testset "UInt8, Int8 vector" begin
+    for VT in [Int8, UInt8]
+        A = VT[0x40, 0x52, 0x62, 0x52, 0x62]
```
nalimilan, line 396: This only tests the case where the pattern and array have the same eltype. It would be nice to test when they have different types, using
```suggestion
    for T in [Int8, UInt8], VT in [Int8, UInt8]
        A = T[0x40, 0x52, 0x62, 0x52, 0x62]
```

nalimilan:
Maybe you could use `firstindex(A)-1` as the sentinel? But I think it's fine to require 1-based indexing for now.

base/strings/search.jl
```diff
@@ -166,17 +185,22 @@ function _search_bloom_mask(c)
...
     _searchindex(unsafe_wrap(Vector{UInt8},s), unsafe_wrap(Vector{UInt8},t), i)
 end
 
-function _searchindex(s::ByteArray, t::ByteArray, i::Integer)
-    n = sizeof(t)
-    m = sizeof(s)
+function _searchindex(s::AbstractVector{<:Union{Int8,UInt8}},
+                      t::AbstractVector{<:Union{Int8,UInt8}},
+                      _i::Integer)
+    require_one_based_indexing(s)
```
stevengj, line 199: Why do we need to require this, since `_nthbyte` handles non 1-based indices?

stevengj: Aren't you testing with an offset array?  Why didn't it throw an error here?

Moelf: Because I wasn't testing offset array that has `firstindex` < 0; the reason this won't work is: https://github.com/JuliaLang/julia/blob/0e082ee1f173bb5992e02496b2de12e5ec98a2e5/base/strings/search.jl#L244 . In short, we can't tell the difference between "not found" (return `0`) and "found in offset array with idx `0`".

stevengj: We should definitely support an arbitrary `firstindex` here.   We could just return `typemin(Int)` as the sentinel value.

Moelf:
not sure what happened in that tester but this is ready except maybe we want to move the code somewhere other than `string/search`

base/strings/search.jl
```diff
@@ -266,7 +266,7 @@ function _search(s::Union{AbstractString,AbstractVector{<:Union{Int8,UInt8}}},
     idx = _searchindex(s,t,i)
     if isempty(t)
         idx:idx-1
-    elseif idx > 0
+    elseif idx > firstindex(s) - 1
```
nalimilan, line 269:
```suggestion
    elseif idx >= firstindex(s)
```

base/strings/search.jl
```diff
@@ -523,7 +523,7 @@ function _rsearch(s::Union{AbstractString,AbstractVector{<:Union{Int8,UInt8}}},
     idx = _rsearchindex(s,t,i)
     if isempty(t)
         idx:idx-1
-    elseif idx > 0
+    elseif idx > firstindex(s) - 1
```
nalimilan, line 526:
```suggestion
    elseif idx >= firstindex(s)
```

nalimilan (approved):
Thanks, looks good to me apart from two small stylistic suggestions.

If `firstindex(A) == typemin(Int)` an underflow will happen, but AFAICT that's OK as a `BoundsError` will be thrown at the top of the function.

Moelf:
I wanted to know what's behavior when offsetting an array close to `typemin/max` of Int: https://github.com/JuliaArrays/OffsetArrays.jl/issues/136 looks like a bug

base/strings/search.jl
stevengj: (Since `_nthbyte` also computes `firstindex(s) - 1`, leaving it as `idx > firstindex(s) - 1` preserves the common subexpression.  Not that it should matter much, since in most cases `firstindex(s)` should be a constant and the compiler will turn both variants into exactly the same code.)

stevengj (approved):

Moelf:
can be merged maybe?

stevengj:
Could use a NEWS item.

stevengj (approved):

stevengj:
Merge conflict

Moelf:
I believe this can be merged now (before another news merge conflict occur)

stevengj:
Thanks!

vtjnash:
When squashing and merging these commits, please try to fix up the commit message to contain only the relevant information. This one just had a rather long list of repeated:
```
    * Update base/strings/search.jl
    
    Co-authored-by: Milan Bouchet-Valat <nalimilan@club.fr>
```
and it makes `git log` a slight bit harder to scroll past.

PR #37289, opened 2020-08-30
RFC: Also change lowering of getproperty

Keno:
So after thinking about #37268, I was wondering what would happen
if one applied the same trick to the lowering of `getproperty`
as well. This PR starts at that, but is mostly a placeholder for
discussion of whether this is something we want to do. In particular,
this PR lowers `a.b` to:

```
getproperty(a)(a, :b)
```

with a default implementation of

```
getproperty(a) = getfield
```

The inference benefit is as expected. Timing end-to-end inference&optimize
time of:

```
struct Foo; end
f(a::Foo) = a.x
```

shows that time to infer `f` drops by about a third (because it doesn't need to
infer getproperty twice). Overall build time of the system image drops
by 6% for me.

If we do decide to go this way, things that would still need to be fixed:

- [ ] Do the same for `setproperty!`
- [ ] `ccall` doesn't like this lowering. Didn't look into it too closely
- [ ] Needs a fallback to backwards compatibility. I'm thinking something
along the lines of

```
getproperty(x) = hasmethod(getproperty, Tuple{typeof(x), Symbol}) ? getproperty : getfield
```

but obviously it would need inference integration to be fast.

### DIFF
--- a/base/Base.jl
+++ b/base/Base.jl
@@ -20,6 +20,8 @@ include(path::String) = include(Base, path)
 const is_primary_base_module = ccall(:jl_module_parent, Ref{Module}, (Any,), Base) === Core.Main
 ccall(:jl_set_istopmod, Cvoid, (Any, Bool), Base, is_primary_base_module)
 
+getproperty(x) = getfield
+
 # Try to help prevent users from shooting them-selves in the foot
 # with ambiguities by defining a few common and critical operations
 # (and these don't need the extra convert code)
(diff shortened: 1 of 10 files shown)


Keno:
In discussion with @JeffBezanson and @vtjnash we decided this was at least a worthwhile direction to explore, so we'll get this PR in shape. @JeffBezanson wanted to look at the ccall lowering issue and then I can finish this up.

vtjnash:
We had a similar discussion for `dotgetfield`, and felt this might not be useful. Do we still want to pursue this? That performance improvement is substantial (and I could see this might affect current work to add atomics).

Keno:
I don't think this is happening.

PR #37293, opened 2020-08-31
Simplify getfield and isdefined tfuncs by treating `DataType` as immutable

martinholters:
And a little cleanup: Neither `UnionAll` nor `SimpleVector` should need special-casing. The former is immutable and gets handled as well as any other immutable, the latter has no fields.

Technically, `DataType` can be mutated, but actually doing that is looking for more trouble than just wrong constant propagation, and we've been doing this for some fields of it anyway. But maybe we can throw in the `setfield` builtin when trying to modify a `DataType`? Or is it even possible to declare `DataType` immutable?

### DIFF
--- a/base/compiler/tfuncs.jl
+++ b/base/compiler/tfuncs.jl
@@ -24,6 +24,8 @@ function find_tfunc(@nospecialize f)
     end
 end
 
+const DATATYPE_TYPES_FIELDINDEX = fieldindex(DataType, :types)
+
 const TYPENAME_NAME_FIELDINDEX = fieldindex(Core.TypeName, :name)
 const TYPENAME_MODULE_FIELDINDEX = fieldindex(Core.TypeName, :module)
 const TYPENAME_WRAPPER_FIELDINDEX = fieldindex(Core.TypeName, :wrapper)
@@ -275,7 +277,11 @@ function isdefined_tfunc(@nospecialize(arg1), @nospecialize(sym))
                 return Const(true)
             elseif isa(arg1, Const)
                 arg1v = (arg1::Const).val
-                if !ismutable(arg1v) || isdefined(arg1v, idx) || isa(arg1v, DataType)
+                if isa(arg1v, DataType)
+                    if !is_dt_nonconst_field(idx)
+                        return Const(isdefined(arg1v, idx))
+                    end
+                elseif !ismutable(arg1v) || isdefined(arg1v, idx)
                     return Const(isdefined(arg1v, idx))
                 end
             end
@@ -599,6 +605,8 @@ function subtype_tfunc(@nospecialize(a), @nospecialize(b))
 end
 add_tfunc(<:, 2, 2, subtype_tfunc, 0)
 
+is_dt_nonconst_field(fld::Int) = fld == DATATYPE_TYPES_FIELDINDEX
+
 function fieldcount_noerror(@nospecialize t)
     if t isa UnionAll || t isa Union
         t = argument_datatype(t)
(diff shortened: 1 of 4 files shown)


base/compiler/tfuncs.jl
```diff
@@ -282,7 +275,7 @@ function isdefined_tfunc(@nospecialize(arg1), @nospecialize(sym))
                 return Const(true)
             elseif isa(arg1, Const)
                 arg1v = (arg1::Const).val
-                if !ismutable(arg1v) || isdefined(arg1v, idx) || (isa(arg1v, DataType) && is_dt_const_field(idx))
+                if !ismutable(arg1v) || isdefined(arg1v, idx) || isa(arg1v, DataType)
```
JeffBezanson, line 278: The `types` field can be undefined and then get set later.

martinholters: Ah. That's one of those where `is_dt_const_field` returned `true`, though.
Could we by any chance just trigger setting it here?

vtjnash: It can also get cleared later

martinholters: So you're saying that
```julia
julia> foo() = isdefined(VersionNumber, :types)

julia> @code_typed foo()
CodeInfo(
1 ─     return true
) => Bool
```
is bad because `foo` won't get recompiled when `VersionNumber.types` gets cleared and `foo()` will then wrongly still return `true`? That is bad, of course, but nothing this PR introduces. It's the same on 1.0, 1.5, and master. I'm not sure how big an undertaking fixing that would be, considering that some code (e.g. `fieldcount`) might rely on this (and `getfield`) being const-inferred.

martinholters: Okay, I managed to construct a test case for the "`types` may get set later" case:
```julia
g37293(::Type{T}) where T = isdefined(T, :types)
struct T37293; x::Val{g37293(T37293)}; end # call g37293 on incomplete type with not-yet-set types
@test g37293(T37293) == isdefined(T37293, :types)
```
This fails on master, but I have an update to this PR in the working that will fix it.

@vtjnash how can I provoke clearing the `types` field?

martinholters:
While at it, what about the fields of `TypeName`? Currently we treat only `name`, `module`, and `wrapper` as constant. These look ok? Is `names` constant? That would be helpful for `fieldcount` now that `(::DataType).types` cannot be treated as constant anymore. What about `hash`?

JeffBezanson:
I believe `TypeName.names` is either constant or undefined.

martinholters:
> I believe `TypeName.names` is either constant or undefined.

But it can become set later if undefined now? So if undefined infer `getfield` as `SimpleVector` (not `Bottom`), if set to `n`, infer as `Const(n)` would be ok?

Coming back to `(::DataType).types` once more. Even if it can get cleared, as long as it cannot be set to a different value than before, infering `getfield` on it as `Const` (if not undefined) would still be ok, as that does not automatically imply nothrow. Would that be valid? (I've tried figuring out how/when `types` gets populated, but I couldn't even figure out where it would be cleared in a way visible to Julia code.)

So to summarize:
|field|status|`isdefined`|`getfield`|remarks|
|---|---|---|---|---|
|`(::DataType).types`|undef|`Bool`|`SimpleVector`||
|  |set|`Bool`|`Const`|assuming can get cleared, but not be set differently
|`(::TypeName).names`|undef|`Bool`|`SimpleVector`|assuming can be set|
| |set|`Const(true)`|`Const`|...but neither cleared nor set differently|

Does that look ok? Then I'll implement it in this PR.

vtjnash:
Replaced by better effects modeling and `const` fields of mutables

PR #37296, opened 2020-08-31
Allow packages to force depwarns

### LINKED ISSUE #37164: Warn that at least one deprecation warning emitted, default depwarn=yes again

timholy:
Fixes #37164 in a gentler way than proposed in the OP, by allowing packages to call `depwarn(args...; force=true)`. No macro equivalent is provided in this PR.

### DIFF
--- a/base/deprecated.jl
+++ b/base/deprecated.jl
@@ -76,12 +76,12 @@ macro deprecate(old, new, ex=true)
     end
 end
 
-function depwarn(msg, funcsym)
+function depwarn(msg, funcsym; force::Bool=false)
     opts = JLOptions()
     if opts.depwarn == 2
         throw(ErrorException(msg))
     end
-    deplevel = opts.depwarn == 1 ? CoreLogging.Warn : CoreLogging.BelowMinLevel
+    deplevel = force || opts.depwarn == 1 ? CoreLogging.Warn : CoreLogging.BelowMinLevel
     @logmsg(
         deplevel,
         msg,


fredrikekre:
Is there an advantage of using `depwarn` rather than just `@warn` if you want to show them unconditionally? I guess the ability to turn them into hard errors with `--depwarn=error`?

timholy:
`@warn` shows the line containing the `@warn` statement itself; `depwarn` shows the caller. With https://github.com/JuliaImages/ImageCore.jl/pull/140, `julia test/deprecated.jl` yields
```julia
julia> include("deprecated.jl")
[ Info: Precompiling ImageCore [a09fc81d-aa75-5fe9-8630-4744c3626534]
[ Info: Deprecation warnings are expected
┌ Warning: `convert(Array{BGR}, img)` will soon switch to returning an array with non-concrete element type, which adds flexibility but
│ with great cost to performance. To maintain the current behavior, use `BGR.(img)` instead.
└ @ ImageCore ~/.julia/dev/ImageCore/src/deprecations.jl:6
┌ Info: It is recommended that you fix this now to avoid breakage when a new version is released and this warning is removed.
└ Tip: to see all deprecation warnings together with code locations, launch Julia with `--depwarn=yes` and rerun your code.
```

`ImageCore/src/deprecations.jl:6` is [this line](https://github.com/JuliaImages/ImageCore.jl/blob/5ecd65eb23d0e8d0d7bbb835cb566ac2f4592814/src/deprecations.jl#L6), which is not very useful because what the user wants to know is where in *my* code did I call a deprecated method from?

Whereas `julia --depwarn=yes test/deprecated.jl` yields
```julia
julia> include("deprecated.jl")
[ Info: Deprecation warnings are expected
┌ Warning: `convert(Array{BGR}, img)` will soon switch to returning an array with non-concrete element type, which adds flexibility but
│ with great cost to performance. To maintain the current behavior, use `BGR.(img)` instead.
│   caller = macro expansion at deprecated.jl:8 [inlined]
└ @ Core ~/.julia/dev/ImageCore/test/deprecated.jl:8
┌ Warning: `convert(Array{Lab}, img)` will soon switch to returning an array with non-concrete element type, which adds flexibility but
│ with great cost to performance. To maintain the current behavior, use `Lab.(img)` instead.
│   caller = macro expansion at deprecated.jl:12 [inlined]
└ @ Core ~/.julia/dev/ImageCore/test/deprecated.jl:12
```
Ignore the 2-vs-1 distinction and focus on the fact that now the file & line number indicates the caller.

johnnychen94:
> Is there an advantage of using depwarn rather than just `@warn` if you want to show them unconditionally?

Another advantage is to not showing thousands of warnings for the same function call 😆

fredrikekre:
> Another advantage is to not show thousands of warnings for the same function call

For that you can use `maxlog`
```
julia> function f()
           for i in 1:10
               @warn "hello, world" maxlog=1
           end
       end

julia> f()
┌ Warning: hello, world
└ @ Main REPL[2]:3
```
but yea, the wrong source of the warning is indeed annoying.

PR #37299, opened 2020-08-31
define keys on RegexMatch

oxinabox:
This came up in discussion in  https://github.com/JuliaLang/julia/pull/34355
Seperating it out from that PR because it is a far less clear question/answer.

In general if we define `getindex` we should define `keys`.

Some things in this PR you might take issue with:

<s>
### 1. It returns a tuple.
`keys`  only promises that it will return a iterator. `keys` on a `NamedTuple` returns a `Tuple`, `keys` on a `Vector` returns a `OneTo`.
A `Tuple` just seems like the simplest thing to do this will normally be small so i didn't think some kind of `KeySet` would be appropriate
also the keys *are* infact ordered, just like the are for a `NamedTuple`.
</s>

### 2. It returns a mix of named and number keys.
We can't return all named since they might not all be named.
We could return all numbered, I would be down with that.
Very easy PR. `keys(m::RegexMatch = keys(m.captures)`.

It seems like this is what we want since this is what `show` currently displays.

### DIFF
--- a/base/regex.jl
+++ b/base/regex.jl
@@ -149,16 +149,21 @@ struct RegexMatch <: AbstractMatch
     regex::Regex
 end
 
+function keys(m::RegexMatch)
+    idx_to_capture_name = PCRE.capture_names(m.regex.regex)
+    return map(eachindex(m.captures)) do i
+        # If the capture group is named, return it's name, else return it's index
+        get(idx_to_capture_name, i, i)
+    end
+end
+
 function show(io::IO, m::RegexMatch)
     print(io, "RegexMatch(")
     show(io, m.match)
-    idx_to_capture_name = PCRE.capture_names(m.regex.regex)
-    if !isempty(m.captures)
+    capture_keys = keys(m)
+    if !isempty(capture_keys)
         print(io, ", ")
-        for i = 1:length(m.captures)
-            # If the capture group is named, show the name.
-            # Otherwise show its index.
-            capture_name = get(idx_to_capture_name, i, i)
+        for (i, capture_name) in enumerate(capture_keys)
             print(io, capture_name, "=")
             show(io, m.captures[i])
             if i < length(m.captures)
--- a/test/regex.jl
+++ b/test/regex.jl
@@ -98,6 +98,7 @@
         @test !haskey(m, "foo")
         @test (m[:a], m[2], m["b"]) == ("x", "y", "z")
         @test sprint(show, m) == "RegexMatch(\"xyz\", a=\"x\", 2=\"y\", b=\"z\")"
+        @test keys(m) == ["a", 2, "b"]
     end
 
     # Backcapture reference in substitution string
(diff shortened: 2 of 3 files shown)


oxinabox:
If combined with #37300  we do end up with a similar problem to what we currently have for `NamedTuples` where `haskey` and `keys` disagree with each other.
`haskey(nt, 2)==true`  but `1 in keys(nt)==false` (if the first capture group was named)

One way to resolve this would be with a special object that overloads `in` to accept both.
Should that be done?
Doing that would make us less consistent with `NamedTuple`s however.

KristofferC:
> >  Tuple just seems like the simplest thing to do this will normally be small so i didn't think some kind of KeySet would be appropriate

Is it really simpler than e.g. a `Vector{Union{Int, Symbol}}`? If you have a function that you send these keys into, it will specialize for all permutations of `Symbol` and `Int` you pass in, which is perhaps not what you want.

oxinabox:
> Is it really simpler than e.g. a Vector{Union{Int, Symbol}}?

No it isn't.
I meant that it is simpler than defining a new  `KeySet` like type.
Definately could change to returning a `Vector`.

KristofferC:
>  I meant that it is simpler than defining a new KeySet like type.

I guess the advantage is that with a `KeySet` like type you don't need to allocate?

JeffBezanson:
This is a good feature, but I definitely think it should not return a tuple. A Vector would be fine. The tuple version is type-unstable, and I doubt you would ever want code specialized on this.

base/regex.jl
```diff
@@ -139,16 +139,21 @@ struct RegexMatch
     regex::Regex
 end
 
+function keys(m::RegexMatch)
+    idx_to_capture_name = PCRE.capture_names(m.regex.regex)
+    return ntuple(length(m.captures)) do i
```
oxinabox, line 144:
```suggestion
    return map(eachindex(m.captures)) do i
```

test/regex.jl
```diff
@@ -71,6 +71,7 @@
         @test !haskey(m, "foo")
         @test (m[:a], m[2], m["b"]) == ("x", "y", "z")
         @test sprint(show, m) == "RegexMatch(\"xyz\", a=\"x\", 2=\"y\", b=\"z\")"
+        @test keys(m) = (:a, 2, :b)
```
oxinabox, line 74:
```suggestion
        @test keys(m) == [:a, 2, :b]
```

JeffBezanson:
Please fix test and rebase.

test/regex.jl
```diff
@@ -81,6 +81,7 @@
         @test !haskey(m, "foo")
         @test (m[:a], m[2], m["b"]) == ("x", "y", "z")
         @test sprint(show, m) == "RegexMatch(\"xyz\", a=\"x\", 2=\"y\", b=\"z\")"
+        @test keys(m) == [:a, 2, :b]
```
oxinabox, line 84:
```suggestion
        @test keys(m) == ["a", 2, "b"]
```

oxinabox:
I thought this was merged ages ago.
Can we merge it so it stops accumulating conflicts?

oxinabox:
Bump

oxinabox:
Bump.
As far as I know noone has voiced any objection to this feature in the 6 months this PR has been open.
Is there anything at all i need to do or see resolved?

PR #37300, opened 2020-08-31
make haskey with Integers work with RegexMatch

oxinabox:
This is a follow up to https://github.com/JuliaLang/julia/pull/36717
Since we define `getindex`, and `haskey(::RegexMatch, ::Symbol)` we should keep going down the road towards being like a `NamedTuple` and also make `haskey(::RegexMatch, ::Integer)` work. 

See also my other PRs in this set to make it more NamedTupley #34355  and #37299

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -83,6 +83,7 @@ Standard library changes
 * `intersect` on `CartesianIndices` now returns `CartesianIndices` instead of `Vector{<:CartesianIndex}` ([#36643]).
 * `push!(c::Channel, v)` now returns channel `c`. Previously, it returned the pushed value `v` ([#34202]).
 * `RegexMatch` objects can now be probed for whether a named capture group exists within it through `haskey()` ([#36717]).
+* For consistency `RegexMatch` objects can also probed by `haskey` with an integer index for if a capture group exists within it ([#37300]).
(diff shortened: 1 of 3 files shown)


NEWS.md
```diff
@@ -83,6 +83,7 @@ Standard library changes
 * `intersect` on `CartesianIndices` now returns `CartesianIndices` instead of `Vector{<:CartesianIndex}` ([#36643]).
 * `push!(c::Channel, v)` now returns channel `c`. Previously, it returned the pushed value `v` ([#34202]).
 * `RegexMatch` objects can now be probed for whether a named capture group exists within it through `haskey()` ([#36717]).
+* For consistency `RegexMatch` objects can also probed by `haskey` with an integer index for if a capture group exists within it ([#37300]).
```
KristofferC, line 86:
```suggestion
* `haskey(r::RegexMatch, i::Int) has been added and returns if the capture group for `i` exists ([#37300]).
```

maybe? "probed by `haskey`" got me a bit confused. Sometimes code can be easier to understand than words.

oxinabox:
```suggestion
* For consistency `haskey(r::RegexMatch, i::Integer) has also been added and returns if the capture group for `i` exists ([#37300]).
```

PR #37303, opened 2020-08-31
Optimize codegen of access functions for SimpleVector

yuyichao:
Better alias info and GC root tracking.

This is the reason I made the observation in https://github.com/JuliaLang/julia/pull/37230#issuecomment-683287984

### DIFF
--- a/base/essentials.jl
+++ b/base/essentials.jl
@@ -590,19 +590,11 @@ function getindex(v::SimpleVector, i::Int)
     @boundscheck if !(1 <= i <= length(v))
         throw(BoundsError(v,i))
     end
-    t = @_gc_preserve_begin v
-    x = unsafe_load(convert(Ptr{Ptr{Cvoid}},pointer_from_objref(v)) + i*sizeof(Ptr))
-    x == C_NULL && throw(UndefRefError())
-    o = unsafe_pointer_to_objref(x)
-    @_gc_preserve_end t
-    return o
+    return ccall(:jl_svec_ref, Any, (Any, Int), v, i - 1)
 end
 
 function length(v::SimpleVector)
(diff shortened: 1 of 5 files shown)


martinholters:
Can you give before/after of generated code, possibly with
```julia
@@ -982,7 +982,7 @@ function abstract_call_known(interp::AbstractInterpreter, @nospecialize(f),
         return CallMeta(abstract_call_known(interp, <:, fargs, argtypes, sv).rt, false)
     elseif la == 2 && isa(argtypes[2], Const) && isa(argtypes[2].val, SimpleVector) && istopfunction(f, :length)
         # mark length(::SimpleVector) as @pure
-        return CallMeta(Const(length(argtypes[2].val)), false)
+        return CallMeta(Const(length(argtypes[2].val)), nothing)
     elseif la == 3 && isa(argtypes[2], Const) && isa(argtypes[3], Const) &&
             isa(argtypes[2].val, SimpleVector) && isa(argtypes[3].val, Int) && istopfunction(f, :getindex)
         # mark getindex(::SimpleVector, i::Int) as @pure
```
also applied?

yuyichao:
The final LLVM IR is locally identical, the only change is a simpler julia IR and better code when interacting with other code. AFAICT, the old code inlines just fine and this does not really make inlining easier in any significant way, at least not for any of the functions I touched directly. They inline just fine for non-constant arguments and causes a runtime dispatch for constant arguments for both `length` and `getindex`.

I just pushed another update with better type check (only applies if someone ccall these directly) and I included a constant evaluation for `length` to be consistent with `sizeof`. With your patch that causes the original code to be inlined it does now give a constant in the LLVM IR but I think it should still be fixed in inference instead...

JeffBezanson (approved):

PR #37308, opened 2020-08-31
fix #37282, printing juxtaposition of 0 confusion with hex,bin,oct literals

### LINKED ISSUE #37282: Ambiguity between the representation of Expr containing hexadecimal literals and juxtaposing of a variable starting with 'x'
The two Expr "a" and "b" have two very different meanings but are represented in the same way by the "string" function.
The problem is fortunately only of the representation since both of the Expr get evaluated correctly.
```julia
x01 = 5
#output: 
# 5
a = :((0)x01 + 9)
#output: 
# :(0x01 + 9)
b = :(0x01 + 9)
#output: 
# :(0x01 + 9)
@eval $a
#output: 
# 9
@eval $b
#output: 
# 10

#=
Julia Version 1.5.1
Commit 697e782ab8 (2020-08-25 20:08 UTC)
Platform Info:
  CPU: AMD Ryzen 5 3600 6-Core Processor
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-9.0.1 (ORCJIT, znver2)
=#
```

JeffBezanson:
fix #37282

### DIFF
--- a/base/show.jl
+++ b/base/show.jl
@@ -1601,7 +1601,10 @@ function show_unquoted(io::IO, ex::Expr, indent::Int, prec::Int, quote_level::In
         # scalar multiplication (i.e. "100x")
         elseif (func === :* &&
             length(func_args) == 2 && isa(func_args[1], Union{Int, Int64, Float32, Float64}) &&
-            isa(func_args[2], Symbol) && !in(string(func_args[2]::Symbol)[1], ('e', 'E', 'f')))
+            isa(func_args[2], Symbol) &&
+            !in(string(func_args[2]::Symbol)[1], ('e', 'E', 'f', (func_args[1] == 0 && func_args[1] isa Integer ?
+                                                                  # don't juxtapose 0 with b, o, x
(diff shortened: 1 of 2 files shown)


Keno (approved):

base/show.jl
```diff
@@ -1601,7 +1601,10 @@ function show_unquoted(io::IO, ex::Expr, indent::Int, prec::Int, quote_level::In
         # scalar multiplication (i.e. "100x")
         elseif (func === :* &&
             length(func_args) == 2 && isa(func_args[1], Union{Int, Int64, Float32, Float64}) &&
```
vtjnash, line 1603:
```suggestion
            length(func_args) == 2 && isa(func_args[1], Union{Int32, Int64, Float32, Float64}) &&
```

vtjnash: Nvm, I assume this might have been intentional

vtjnash (approved):

PR #37311, opened 2020-08-31
Pass the full mapped length to msync/FlushViewOfFile

jmert:
While looking at the details of the `Mmap` module, I noticed that the `sync!` command seems to pass the wrong length to the underlying OS functions. On Windows, the [length argument](https://docs.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-flushviewoffile) takes the number of bytes to sync, but it was being passed the number of elements of the array. Then on both Windows and Unix systems, the offset of the array's start address with respect to a `PAGESIZE` boundary was left out of the total length calculation.

Unfortunately, I haven't been able to construct a test (on Linux) where I can observe a partial sync — even writing `Int64` vectors (so that the length and sizeof differ) to a new, large (~400 MB) file living on a non-`tmpfs` filesystem where the memory map starts mid-page, I always read back in the array as expected (including comparing before and after the sync and after closing the mmap+file completely).


vtjnash:
I think you'd have to crash the machine, or observe it over a network. Otherwise, I think it's promised to keep the in-memory views of the file in sync even without this.

PR #37320, opened 2020-09-01
Move Artifacts and BinaryPlatforms into Base

staticfloat:
This is the fundamental infrastructure needed for the 1.6 `Artifacts` rework.

* Imports `Pkg.BinaryPlatforms` to `Base.BinaryPlatforms`.  I didn't think it was useful enough to be exported from `Base`, so it's currently usable through `using Base.BinaryPlatforms`.  It has been more or less completely rewritten to be more flexible, more compiler-friendly and to power us into the age of https://github.com/JuliaLang/Pkg.jl/issues/1780

* Imports `Pkg.Artifacts` to its own stdlib, `Artifacts`.  Includes locating pieces of `Pkg.Artifacts`, not the installation pieces, which are still within `Pkg`.  Scripts that want to generate artifacts should still `using Pkg.Artifacts`, but scripts that just want to use artifacts that have already been generated are free to simply `using Artifacts`, which should cut down on time to first load by a nice chunk.  If a package attempts to `@artifact"foo"` on an artifact that does not exist, the `Artifacts` module will dynamically load `Pkg` and invoke the downloading routines through an inference barrier.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -102,6 +102,9 @@ Standard library changes
 * The `Pkg.BinaryPlatforms` module has been moved into `Base` as `Base.BinaryPlatforms` and heavily reworked.
   Applications that want to be compatible with the old API should continue to import `Pkg.BinaryPlatforms`,
   however new users should use `Base.BinaryPlatforms` directly. ([#37320])
+* The `Pkg.Artifacts` module has been imported as a separate standard library.  It is still available as
+  `Pkg.Artifacts`, however starting from Julia v1.6+, packages may import simply `Artifacts` without importing
+  all of `Pkg` alongside. ([#37320])
 
 #### LinearAlgebra
 * New method `LinearAlgebra.issuccess(::CholeskyPivoted)` for checking whether pivoted Cholesky factorization was successful ([#36002]).
--- a/NEWS.md
+++ b/NEWS.md
@@ -99,6 +99,9 @@ Standard library changes
 * `RegexMatch` objects can now be probed for whether a named capture group exists within it through `haskey()` ([#36717]).
 * For consistency `haskey(r::RegexMatch, i::Integer)` has also been added and returns if the capture group for `i` exists ([#37300]).
 * A new standard library `TOML` has been added for parsing and printing [TOML files](https://toml.io) ([#37034]).
+* The `Pkg.BinaryPlatforms` module has been moved into `Base` as `Base.BinaryPlatforms` and heavily reworked.
+  Applications that want to be compatible with the old API should continue to import `Pkg.BinaryPlatforms`,
+  however new users should use `Base.BinaryPlatforms` directly. ([#37320])
 
 #### LinearAlgebra
 * New method `LinearAlgebra.issuccess(::CholeskyPivoted)` for checking whether pivoted Cholesky factorization was successful ([#36002]).
--- a/base/binaryplatforms.jl
+++ b/base/binaryplatforms.jl
@@ -1,0 +1,1010 @@
+module BinaryPlatforms
+
+export AbstractPlatform, Platform, HostPlatform, platform_dlext, tags, arch, os,
+       os_version, libc, compiler_abi, libgfortran_version, libstdcxx_version,
+       cxxstring_abi, parse_dl_name_version, detect_libgfortran_version,
+       detect_libstdcxx_version, detect_cxxstring_abi, call_abi, wordsize, triplet,
+       select_platform, platforms_match, platform_name
+import .Libc.Libdl
+
+# This exists to ease compatibility with old-style Platform objects
+abstract type AbstractPlatform; end
+
+"""
+    Platform
+
+A `Platform` represents all relevant pieces of information that a julia process may need
+to know about its execution environment, such as the processor architecture, operating
+system, libc implementation, etc...  It is, at its heart, a key-value mapping of tags
+(such as `arch`, `os`, `libc`, etc...) to values (such as `"arch" => "x86_64"`, or
+`"os" => "windows"`, etc...).  `Platform` objects are extensible in that the tag mapping
+is open for users to add their own mappings to, as long as the mappings do not conflict
+with the set of reserved tags: `arch`, `os`, `os_version`, `libc`, `call_abi`,
+`libgfortran_version`, `libstdcxx_version`, `cxxstring_abi` and `julia_version`.
+
+Valid tags and values are composed of alphanumeric and period characters.  All tags and
+values will be lowercased when stored to reduce variation.
+
+Example:
+
+    Platform("x86_64", "windows"; cuda = "10.1")
+"""
+struct Platform <: AbstractPlatform
+    tags::Dict{String,String}
+    # The "compare strategy" allows selective overriding on how a tag is compared
+    compare_strategies::Dict{String,Function}
+
+    function Platform(arch::String, os::String;
+                      validate_strict::Bool = false,
+                      compare_strategies::Dict{String,<:Function} = Dict{String,Function}(),
+                      kwargs...)
+        # A wee bit of normalization
+        os = lowercase(os)
+        arch = lowercase(arch)
+        if arch ∈ ("amd64",)
+            arch = "x86_64"
+        elseif arch ∈ ("i386", "i586")
+            arch = "i686"
+        elseif arch ∈ ("arm",)
+            arch = "armv7l"
+        elseif arch ∈ ("ppc64le",)
+            arch = "powerpc64le"
+        end
+
+        tags = Dict{String,String}(
+            "arch" => arch,
+            "os" => os,
+        )
+        for (tag, value) in kwargs
+            tag = lowercase(string(tag))
+            if tag ∈ ("arch", "os")
+                throw(ArgumentError("Cannot double-pass key $(tag)"))
+            end
+
+            # Drop `nothing` values; this means feature is not present or use default value.
+            if value === nothing
+                continue
+            end
+
+            # Normalize things that are known to be version numbers so that comparisons are easy.
+            # Note that in our effort to be extremely compatible, we actually allow something that
+            # doesn't parse nicely into a VersionNumber to persist, but if `validate_strict` is
+            # set to `true`, it will cause an error later on.
+            if tag ∈ ("libgfortran_version", "libstdcxx_version", "os_version")
+                normver(x::VersionNumber) = string(x)
+                function normver(str::AbstractString)
+                    v = tryparse(VersionNumber, str)
+                    if v === nothing
+                        # If this couldn't be parsed as a VersionNumber, return the original.
+                        return str
+                    end
+                    # Otherwise, return the `string(VersionNumber(str))` version.
+                    return normver(v)
+                end
+                value = normver(value)
+            end
+
+            # I know we said only alphanumeric and dots, but let's be generous so that we can expand
+            # our support in the future while remaining as backwards-compatible as possible.  The
+            # only characters that are absolutely disallowed right now are `-`, `+`, ` ` and things
+            # that are illegal in filenames:
+            nonos = raw"""+- /<>:"'\|?*"""
+            if any(occursin(nono, tag) for nono in nonos)
+                throw(ArgumentError("Invalid character in tag name \"$(tag)\"!"))
+            end
+
+            # Normalize and reject nonos
+            value = lowercase(string(value))
+            if any(occursin(nono, value) for nono in nonos)
+                throw(ArgumentError("Invalid character in tag value \"$(value)\"!"))
+            end
+            tags[tag] = value
+        end
+
+        # Auto-map call_abi and libc where necessary:
+        if os == "linux" && !haskey(tags, "libc")
+            # Default to `glibc` on Linux
+            tags["libc"] = "glibc"
+        end
+        if os == "linux" && arch ∈ ("armv7l", "armv6l") && "call_abi" ∉ keys(tags)
+            # default `call_abi` to `eabihf` on 32-bit ARM
+            tags["call_abi"] = "eabihf"
+        end
+
+        # If the user is asking for strict validation, do so.
+        if validate_strict
+            validate_tags(tags)
+        end
+
+        # By default, we compare julia_version only against major and minor versions:
+        if haskey(tags, "julia_version") && !haskey(compare_strategies, "julia_version")
+            compare_strategies["julia_version"] = (a::String, b::String, a_comparator, b_comparator) -> begin
+                a = VersionNumber(a)
+                b = VersionNumber(b)
+                return a.major == b.major && a.minor == b.minor
+            end
+        end
+
+        return new(tags, compare_strategies)
+    end
+end
+
+# Other `Platform` types can override this (I'm looking at you, `AnyPlatform`)
+tags(p::Platform) = p.tags
+Base.haskey(p::AbstractPlatform, k::String) = haskey(tags(p), k)
+
+# Allow us to easily serialize Platform objects
+function Base.repr(p::Platform; context=nothing)
+    str = string(
+        "Platform(",
+        repr(arch(p)),
+        ", ",
+        repr(os(p)),
+        "; ",
+        join(("$(k) = $(repr(v))" for (k, v) in tags(p) if k ∉ ("arch", "os")), ", "),
+        ")",
+    )
+end
+
+# Simple equality definition; for compatibility testing, use `platforms_match()`
+Base.:(==)(a::AbstractPlatform, b::AbstractPlatform) = tags(a) == tags(b)
+
+const ARCHITECTURE_FLAGS = Dict(
+    "x86_64" => ["x86_64", "avx", "avx2", "avx512"],
+    "i686" => ["prescott"],
+    "armv7l" => ["armv7l", "neon", "vfp4"],
+    "armv6l" => ["generic"],
+    "aarch64" => ["armv8", "thunderx2", "carmel"],
+    "powerpc64le" => ["generic"],
+)
+function validate_tags(tags::Dict)
+    throw_invalid_key(k) = throw(ArgumentError("Key \"$(k)\" cannot have value \"$(tags[k])\""))
+    # Validate `arch`
+    if tags["arch"] ∉ keys(ARCHITECTURE_FLAGS)
+        throw_invalid_key("arch")
+    end
+    # Validate `os`
+    if tags["os"] ∉ ("linux", "macos", "freebsd", "windows")
+        throw_invalid_key("os")
+    end
+    # Validate `os`/`arch` combination
+    throw_os_mismatch() = throw(ArgumentError("Invalid os/arch combination: $(tags["os"])/$(tags["arch"])"))
+    if tags["os"] == "windows" && tags["arch"] ∉ ("x86_64", "i686", "armv7l", "aarch64")
+        throw_os_mismatch()
+    end
+    if tags["os"] == "macos" && tags["arch"] ∉ ("x86_64", "aarch64")
+        throw_os_mismatch()
+    end
+
+    # Validate `os`/`libc` combination
+    throw_libc_mismatch() = throw(ArgumentError("Invalid os/libc combination: $(tags["os"])/$(tags["libc"])"))
+    if tags["os"] == "linux"
+        # Linux always has a `libc` entry
+        if tags["libc"] ∉ ("glibc", "musl")
+            throw_libc_mismatch()
+        end
+    else
+        # Nothing else is allowed to have a `libc` entry
+        if haskey(tags, "libc")
+            throw_libc_mismatch()
+        end
+    end
+
+    # Validate `os`/`arch`/`call_abi` combination
+    throw_call_abi_mismatch() = throw(ArgumentError("Invalid os/arch/call_abi combination: $(tags["os"])/$(tags["arch"])/$(tags["call_abi"])"))
+    if tags["os"] == "linux" && tags["arch"] ∈ ("armv7l", "armv6l")
+        # If an ARM linux has does not have `call_abi` set to something valid, be sad.
+        if !haskey(tags, "call_abi") || tags["call_abi"] ∉ ("eabihf", "eabi")
+            throw_call_abi_mismatch()
+        end
+    else
+        # Nothing else should have a `call_abi`.
+        if haskey(tags, "call_abi")
+            throw_call_abi_mismatch()
+        end
+    end
+
+    # Validate `libgfortran_version` is a parsable `VersionNumber`
+    throw_version_number(k) = throw(ArgumentError("\"$(k)\" cannot have value \"$(tags[k])\", must be a valid VersionNumber"))
+    if "libgfortran_version" in keys(tags) && tryparse(VersionNumber, tags["libgfortran_version"]) === nothing
+        throw_version_number("libgfortran_version")
+    end
+
+    # Validate `libstdcxx_version` is a parsable `VersionNumber`
+    if "libstdcxx_version" in keys(tags) && tryparse(VersionNumber, tags["libstdcxx_version"]) === nothing
+        throw_version_number("libstdcxx_version")
+    end
+
+    # Validate `cxxstring_abi` is one of the two valid options:
+    if "cxxstring_abi" in keys(tags) && tags["cxxstring_abi"] ∉ ("cxx03", "cxx11")
+        throw_invalid_key("cxxstring_abi")
+    end
+
+    # Validate `march` is one of our recognized microarchitectures for the architecture we're advertising
+    if "march" in keys(tags) && tags["march"] ∉ ARCHITECTURE_FLAGS[tags["arch"]]
+        throw(ArgumentError("\"march\" cannot have value \"$(tags["march"])\" for arch $(tags["arch"])"))
+    end
+
+    # Validate `cuda` is a parsable `VersionNumber`
+    if "cuda" in keys(tags) && tryparse(VersionNumber, tags["cuda"]) === nothing
+        throw_version_number("cuda")
+    end
+end
+
+function set_compare_strategy!(p::Platform, key::String, f::Function)
+    if !haskey(p.tags, key)
+        throw(ArgumentError("Cannot set comparison strategy for nonexistant tag $(key)!"))
+    end
+    p.compare_strategies[key] = f
+end
+
+function get_compare_strategy(p::Platform, key::String, default = compare_default)
+    if !haskey(p.tags, key)
+        throw(ArgumentError("Cannot get comparison strategy for nonexistant tag $(key)!"))
+    end
+    return get(p.compare_strategies, key, default)
+end
+get_compare_strategy(p::AbstractPlatform, key::String, default = compare_default) = default
+
+
+
+"""
+    compare_default(a::String, b::String, a_requested::Bool, b_requested::Bool)
+
+Default comparison strategy that falls back to `a == b`.  This only ever happens if both
+`a` and `b` request this strategy, as any other strategy is preferrable to this one.
+"""
+function compare_default(a::String, b::String, a_requested::Bool, b_requested::Bool)
+    return a == b
+end
+
+"""
+    compare_version_cap(a::String, b::String, a_comparator, b_comparator)
+
+Example comparison strategy for `set_comparison_strategy!()` that implements a version
+cap for host platforms that support _up to_ a particular version number.  As an example,
+if an artifact is built for macOS 10.9, it can run on macOS 10.11, however if it were
+built for macOS 10.12, it could not.  Therefore, the host platform of macOS 10.11 has a
+version cap at `v"10.11"`.
+
+Note that because both hosts and artifacts are represented with `Platform` objects it
+is possible to call `platforms_match()` with two artifacts, a host and an artifact, an
+artifact and a host, and even two hosts.  We attempt to do something intelligent for all
+cases, but in the case of comparing version caps between two hosts, we return `true` only
+if the two host platforms are in fact identical.
+"""
+function compare_version_cap(a::String, b::String, a_requested::Bool, b_requested::Bool)
+    a = VersionNumber(a)
+    b = VersionNumber(b)
+
+    # If both b and a requested, then we fall back to equality:
+    if a_requested && b_requested
+        return a == b
+    end
+
+    # Otherwise, do the comparison between the the single version cap and the single version:
+    if a_requested
+        return b <= a
+    else
+        return a <= b
+    end
+end
+
+
+
+"""
+    HostPlatform(p::AbstractPlatform)
+
+Convert a `Platform` to act like a "host"; e.g. if it has a version-bound tag such as
+`"libstdcxx_version" => "3.4.26"`, it will treat that value as an upper bound, rather
+than a characteristic.  `Platform` objects that define artifacts generally denote the
+SDK or version that the artifact was built with, but for platforms, these versions are
+generally the maximal version the platform can support.  The way this transformation
+is implemented is to change the appropriate comparison strategies to treat these pieces
+of data as bounds rather than points in any comparison.
+"""
+function HostPlatform(p::AbstractPlatform)
+    if haskey(p, "os_version")
+        set_compare_strategy!(p, "os_version", compare_version_cap)
+    end
+    if haskey(p, "libstdcxx_version")
+        set_compare_strategy!(p, "libstdcxx_version", compare_version_cap)
+    end
+    return p
+end
+
+"""
+    arch(p::AbstractPlatform)
+
+Get the architecture for the given `Platform` object as a `String`.
+
+# Examples
+```jldoctest
+julia> arch(Platform("aarch64", "Linux"))
+"aarch64"
+
+julia> arch(Platform("amd64", "freebsd"))
+"x86_64"
+```
+"""
+arch(p::AbstractPlatform) = get(tags(p), "arch", nothing)
+
+"""
+    os(p::AbstractPlatform)
+
+Get the operating system for the given `Platform` object as a `String`.
+
+# Examples
+```jldoctest
+julia> os(Platform("armv7l", "Linux"))
+"linux"
+
+julia> os(Platform("aarch64", "macos"))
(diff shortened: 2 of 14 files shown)


base/binaryplatforms.jl
```diff
@@ -0,0 +1,999 @@
+module BinaryPlatforms
+
+export AbstractPlatform, Platform, HostPlatform, platform_dlext, tags, arch, os,
```
vtjnash, line 3:
```suggestion
export Platform, HostPlatform, platform_dlext, tags, arch, os,
```

Doesn't make sense to export a deprecated type. I'm also not sure other small names like `os` make sense to be exported either, or helper functions with internal-sounding names like `parse_dl_name_version`.

staticfloat: While I agree with you, I think we need to keep backward-compatibility.  Should I instead only export these from `Pkg.BinaryPlatforms`?

KristofferC: `BinaryPlatforms` was never an exposed user interface was it?

base/binaryplatforms.jl
```diff
@@ -0,0 +1,999 @@
...
+        elseif arch ∈ ("ppc64le",)
+            arch = "powerpc64le"
+        end
+
+        tags = Dict{String,String}(
+            "arch" => arch,
+            "os" => os,
+        )
+        for (tag, value) in kwargs
+            tag = lowercase(string(tag))
```
KristofferC, line 59:
```suggestion
            tag = lowercase(string(tag)::String)
```

am I getting this right @timholy?

base/binaryplatforms.jl
```diff
@@ -0,0 +1,999 @@
...
+                continue
+            end
+
+            # Normalize things that are known to be version numbers so that comparisons are easy.
+            # Note that in our effort to be extremely compatible, we actually allow something that
+            # doesn't parse nicely into a VersionNumber to persist, but if `validate_strict` is
+            # set to `true`, it will cause an error later on.
+            if tag ∈ ("libgfortran_version", "libstdcxx_version", "os_version")
+                normver(x::VersionNumber) = string(x)
+                function normver(str::AbstractString)
```
KristofferC, line 75: Perhaps make this

```suggestion
                function normver(str::String)
```

and then call it with `normver(string(value)::String)` below?

base/binaryplatforms.jl
```diff
@@ -0,0 +1,999 @@
...
+            # our support in the future while remaining as backwards-compatible as possible.  The
+            # only characters that are absolutely disallowed right now are `-`, `+`, ` ` and things
+            # that are illegal in filenames:
+            nonos = raw"""+- /<>:"'\|?*"""
+            if any(occursin(nono, tag) for nono in nonos)
+                throw(ArgumentError("Invalid character in tag name \"$(tag)\"!"))
+            end
+
+            # Normalize and reject nonos
+            value = lowercase(string(value))
```
KristofferC, line 97:
```suggestion
            value = lowercase(string(value)::String)
```

base/binaryplatforms.jl
```diff
@@ -0,0 +1,999 @@
...
+        if os == "linux" && !haskey(tags, "libc")
+            # Default to `glibc` on Linux
+            tags["libc"] = "glibc"
+        end
+        if os == "linux" && arch ∈ ("armv7l", "armv6l") && "call_abi" ∉ keys(tags)
+            # default `call_abi` to `eabihf` on 32-bit ARM
+            tags["call_abi"] = "eabihf"
+        end
+
+        # If the user is asking for strict validation, do so.
```
KristofferC, line 114: I mean, that's what the code just below it says 🤷

base/binaryplatforms.jl
```diff
@@ -0,0 +1,999 @@
...
+values will be lowercased when stored to reduce variation.
+
+Example:
+
+    Platform("x86_64", "windows"; cuda = "10.1")
+"""
+struct Platform <: AbstractPlatform
+    tags::Dict{String,String}
+    # The "compare strategy" allows selective overriding on how a tag is compared
+    compare_strategies::Dict{String,Any}
```
KristofferC, line 35:
```suggestion
    compare_strategies::Dict{String,Function}
```

or?

base/binaryplatforms.jl
```diff
@@ -0,0 +1,999 @@
...
+        ")",
+    )
+end
+
+# Simple equality definition; for compatibility testing, use `platforms_match()`
+Base.:(==)(a::AbstractPlatform, b::AbstractPlatform) = tags(a) == tags(b)
+
+function validate_tags(tags::Dict)
+    throw_invalid_key(k) = throw(ArgumentError("Key \"$(k)\" cannot have value \"$(tags[k])\""))
+    ARCHITECTURE_FLAGS = Dict(
```
KristofferC, line 154: make this a const and move it outside?

base/binaryplatforms.jl
```diff
@@ -0,0 +1,999 @@
...
+        "; ",
+        join(("$(k) = $(repr(v))" for (k, v) in tags(p) if k ∉ ("arch", "os")), ", "),
+        ")",
+    )
+end
+
+# Simple equality definition; for compatibility testing, use `platforms_match()`
+Base.:(==)(a::AbstractPlatform, b::AbstractPlatform) = tags(a) == tags(b)
+
+function validate_tags(tags::Dict)
```
KristofferC, line 152:
```suggestion
function validate_tags(tags::Dict{String, String})
```

I think?

base/binaryplatforms.jl
staticfloat: It is as exposed now as it was in `Pkg`.  Packages can `using Pkg.BinaryPlatforms` and they get a bunch of stuff, just like they can `using Base.BinaryPlatforms`.  But neither are re-exported from `Base` or from `Pkg`.

base/binaryplatforms.jl
```diff
@@ -0,0 +1,999 @@
...
+"""
+    Platform()
+
+Return the `Platform` object that corresponds to the current host system, with all
+relevant comparison strategies set to host platform mode.  This is equivalent to:
+
+    HostPlatform(parse(Platform, Base.BinaryPlatforms.host_triplet()))
+"""
+function Platform()
+    global default_host_platform
```
KristofferC, line 905:
```suggestion
```

I think?

It seems a bit odd for me that this is called `Platform` but it isn't a constructor.

stdlib/Artifacts/src/Artifacts.jl
```diff
@@ -0,0 +1,634 @@
...
+import Base: get, SHA1
+using Base.BinaryPlatforms, Base.TOML
+
+export artifact_exists, artifact_path, remove_artifact, verify_artifact,
+       artifact_meta, artifact_hash, find_artifacts_toml, @artifact_str
+
+"""
+    parse_toml(path::String)
+
+Uses Base.TOML to parse a TOML file
```
KristofferC, line 12: Feels like this doesn't really need a docstring? The reason to not add a dependency to the TOML stdlib is to keep this module as lean as possible? Could be added as a comment.

stdlib/Artifacts/src/Artifacts.jl
```diff
@@ -0,0 +1,634 @@
...
+Helper function to allow temporarily changing the artifact installation and search
+directory.  When this is set, no other directory will be searched for artifacts, and new
+artifacts will be installed within this directory.  Similarly, removing an artifact will
+only effect the given artifact directory.  To layer artifact installation locations, use
+the typical Julia depot path mechanism.
+"""
+function with_artifacts_directory(f::Function, artifacts_dir::String)
+    try
+        ARTIFACTS_DIR_OVERRIDE[] = artifacts_dir
+        f()
```
KristofferC, line 36:
```suggestion
        string(f())::String
```

maybe

stdlib/Artifacts/src/Artifacts.jl
```diff
@@ -0,0 +1,634 @@
...
+end
+
+"""
+    artifacts_dirs(args...)
+
+Return a list of paths joined into all possible artifacts directories, as dictated by the
+current set of depot paths and the current artifact directory override via the method
+`with_artifacts_dir()`.
+"""
+function artifacts_dirs(args...)
```
KristofferC, line 49: Looks like this is only called with a `String` so far?

stdlib/Artifacts/src/Artifacts.jl
```diff
@@ -0,0 +1,634 @@
...
+    if ARTIFACTS_DIR_OVERRIDE[] === nothing
+        return [abspath(depot, "artifacts", args...) for depot in Base.DEPOT_PATH]
```
KristofferC, line 51:
```suggestion
        return String[abspath(depot, "artifacts", args...) for depot in Base.DEPOT_PATH]
```

stdlib/Artifacts/src/Artifacts.jl
```diff
@@ -0,0 +1,634 @@
...
+    else
+        # If we've been given an override, use _only_ that directory.
+        return [abspath(ARTIFACTS_DIR_OVERRIDE[], args...)]
```
KristofferC, line 54:
```suggestion
        return String[abspath(ARTIFACTS_DIR_OVERRIDE[], args...)]
```

stdlib/Artifacts/src/Artifacts.jl
```diff
@@ -0,0 +1,634 @@
...
+    for override_file in reverse(artifacts_dirs("Overrides.toml"))
+        !isfile(override_file) && continue
+
+        # Load the toml file
+        depot_override_dict = parse_toml(override_file)
+
+        function parse_mapping(mapping::String, name::String)
+            if !isabspath(mapping) && !isempty(mapping)
+                try
+                    mapping = Base.SHA1(mapping)
```
KristofferC, line 108: there's no `tryparse` for `Base.SHA1`? Perhaps should be added?

stdlib/Artifacts/src/Artifacts.jl
```diff
@@ -0,0 +1,634 @@
...
+                end
+            end
+            return mapping
+        end
+        function parse_mapping(mapping::Dict, name::String)
+            return Dict(k => parse_mapping(v, name) for (k, v) in mapping)
+        end
+
+        for (k, mapping) in depot_override_dict
+            # First, parse the mapping. Is it an absolute path, a valid SHA1-hash, or neither?
```
KristofferC, line 121:
```suggestion
            # First, parse the mapping. Is it an absolute path, a valid SHA1-hash, or neither?
            mapping::Union{Dict{String, Any}, String}
```

I think?

stdlib/Artifacts/src/Artifacts.jl
```diff
@@ -0,0 +1,634 @@
...
+                    continue
+                end
+
+                # If this mapping is the empty string, un-override it
+                if mapping == ""
+                    delete!(overrides[:hash], hash)
+                else
+                    overrides[:hash][hash] = mapping
+                end
+            elseif isa(mapping, Dict)
```
KristofferC, line 145:
```suggestion
            elseif isa(mapping, Dict{String, Any})
```

stdlib/Artifacts/src/Artifacts.jl
```diff
@@ -0,0 +1,634 @@
...
+                # If this mapping is itself a dict, store it as a set of UUID/artifact name overrides
+                ovruuid = overrides[:UUID]::Dict{Base.UUID,Dict{String,Union{String,SHA1}}}
+                if !haskey(ovruuid, uuid)
+                    ovruuid[uuid] = Dict{String,Union{String,SHA1}}()
+                end
+
+                # For each name in the mapping, update appropriately
+                for name in keys(mapping)
+                    # If the mapping for this name is the empty string, un-override it
+                    if mapping[name] == ""
```
KristofferC, line 163: Why loop over `keys(mapping)` and then immediately lookup `mapping[name]`? Maybe just loop directly over the dict?

stdlib/Artifacts/src/Artifacts.jl
```diff
@@ -0,0 +1,634 @@
...
+                # If this mapping is the empty string, un-override it
+                if mapping == ""
+                    delete!(overrides[:hash], hash)
+                else
+                    overrides[:hash][hash] = mapping
+                end
+            elseif isa(mapping, Dict)
+                # Convert `k` into a uuid
+                uuid = try
+                    Base.UUID(k)
```
KristofferC, line 148: We need a `tryparse(UUID, k)`..

NEWS.md
```diff
@@ -84,6 +84,12 @@ Standard library changes
 * `push!(c::Channel, v)` now returns channel `c`. Previously, it returned the pushed value `v` ([#34202]).
 * `RegexMatch` objects can now be probed for whether a named capture group exists within it through `haskey()` ([#36717]).
 * A new standard library `TOML` has been added for parsing and printing [TOML files](https://toml.io) ([#37034]).
+* The `Pkg.BinaryPlatforms` module has been moved into `Base` as `Base.BinaryPlatforms` and heavily reworked.
```
KristofferC, line 87: Was this ever documented / exposed to users? There is nothing in the Pkg docs about it I think.

base/binaryplatforms.jl
KristofferC: Ok, but it wasn't documented / exposed from Pkg either, right?

base/binaryplatforms.jl
```diff
@@ -0,0 +1,999 @@
...
+                return a.major == b.major && a.minor == b.minor
+            end
+        end
+
+        return new(tags, compare_strategies)
+    end
+end
+
+# Other `Platform` types can override this (I'm looking at you, `AnyPlatform`)
+tags(p::Platform) = p.tags
```
KristofferC, line 133: Might want to use @timholy s "interface" trick like in 

https://github.com/JuliaLang/julia/commit/48fc8b4e5312eb9d815c7bfe665c55625a225654

here. Basically

```jl
function Base.getproperty(platform::AbstractPlatform, fieldname::Symbol)
    if fieldname === :tags
        return getfield(platform, :state)::Dict{String, String}
    else
        getfield(platform, fieldname)
    end
end
```

and then use `.tags` everywhere. That ensures inference know that `.tags` has to be a `Dict{String, String}`.

stdlib/Artifacts/src/Artifacts.jl
```diff
@@ -0,0 +1,634 @@
...
+a `~/.julia/artifacts/Override.toml` file with the following contents:
+
+    78f35e74ff113f02274ce60dab6e92b4546ef806 = "/path/to/replacement"
+    c76f8cda85f83a06d17de6c57aabf9e294eb2537 = "fb886e813a4aed4147d5979fcdf27457d20aa35d"
+
+    [d57dbccd-ca19-4d82-b9b8-9d660942965b]
+    c_simple = "/path/to/c_simple_dir"
+    libfoo = "fb886e813a4aed4147d5979fcdf27457d20aa35d""
+
+This file defines four overrides; two which override specific artifacts identified
```
KristofferC, line 72: Referring to "this file" in a docstring is a bit non-ideal maybe. The function could be moved and the docstring (as opposed to a comment) is mostly supposed to be read in some other medium like in the terminal or html rendering where the concept of the file is just an implementation detail.

KristofferC:
One general comment is that it is quite hard for me to understand what is the public API and what is internal.

For example, there is the function `split_artifact_slash` which has full documentation with docstrings and everything but it seems to just be called internally in another function. Is that really something that is going to be a public API and subject to non-breakage etc?

base/binaryplatforms.jl
KristofferC: I guess the jll files gotta use it though? Is that what we want to be backwards compatible with?

staticfloat: Right; older JLLs, BB and BP are the main users of it I think.

staticfloat:
> One general comment is that it is quite hard for me to understand what is the public API and what is internals.
> 
> For example, there is the function split_artifact_slash which has full documentation with docstrings and everything but it seems to just be called internally in another function. Is that really something that is going to be a public API and subject to non-breakage etc?

In my opinion, anything that is not completely trivial should have docstrings to aid others (and, honestly, myself) when editing this file in the future.  The demarcator for public/private isn't how well documented it is, it's whether it's exported or not.  Since both of these modules had certain exports in `Pkg`, I've defaulted to mostly exporting the same things from them (So that e.g. users can switch from `Pkg.Artifacts` to `Artifacts` with as little disruption as possible) but especially in the case of `BinaryPlatforms` there's not much need for that.

NEWS.md
staticfloat: There are a few other references to it, in `Tectonic`, `BugReporting`, `PlotlyBase`, `NNlib`, and `MethodAnalysis`, but I agree it's not too important.  I mostly just wanted to have something to show for weeks of effort.  :P

base/binaryplatforms.jl
staticfloat: I thought this kind of overriding of `getproperty()` was frowned upon?  Is that an outdated Julia style?

base/binaryplatforms.jl
staticfloat: Is that preferable?  I thought it was better to have it embedded so that it doesn't show up in as many scopes.

base/binaryplatforms.jl
staticfloat: Yes, I struggled with this.  I think we should perhaps instead call this `HostPlatform()`.

stdlib/Artifacts/src/Artifacts.jl
staticfloat: Why would I use a comment instead of a docstring?  That seems like a strictly worse developer experience.  I'd much rather be able to look up this docstring in editors and the REPL.

stdlib/Artifacts/src/Artifacts.jl
staticfloat: Why is `f()` returning a `String`?  I don't think that's true.

stdlib/Artifacts/src/Artifacts.jl
staticfloat: Within this file, yes.  I could probably annotate `args` as `::AbstractString...`.

stdlib/Artifacts/src/Artifacts.jl
staticfloat, line 51:
```suggestion
        return String[abspath(depot, "artifacts", args...) for depot in Base.DEPOT_PATH]
```

stdlib/Artifacts/src/Artifacts.jl
staticfloat: It seems very straightforward to me what file it's referring to: the ``~/.julia/artifacts/Override.toml` file it was talking about in the paragraph before.

stdlib/Artifacts/src/Artifacts.jl
staticfloat: Agreed.

base/binaryplatforms.jl
timholy: This isn't about adding OOP-like behavior; it's really about fixing inference problems you have abstract types but can't infer the specific concrete subtype. (E.g., if you extract `dict[key]` when `dict::Dict{String,Platform}`, you know the result is a `Platform` but not the concrete subtype.) Inference will generally return `::Any` as the inferred type of field accesses to abstract types. Defining a `getproperty` lets you add the type-asserts to fix it. See https://timholy.github.io/SnoopCompile.jl/stable/snoopr/#Inferrable-field-access-for-abstract-types-1 for more explanation.

base/binaryplatforms.jl
KristofferC: Otherwise, it will be created at every call, no? But maybe not performance sensitive so it doesn't matter? If you think about scope I think you would usually do it as


```
let
const FOO = ...
global function f(...)
    # use FOO
end
end
```

stdlib/Artifacts/src/Artifacts.jl
KristofferC: I think I mixed this up with some other function...

KristofferC:
>  The demarcator for public/private isn't how well documented it is, it's whether it's exported or not.

This is not really true. There are many things that are public that are not exported (and cannot be exported due to name clashes). For example `Profile.print`, `TOML.parse` etc. With few exceptions, only public-facing API in the Julia repo has docstrings.

stdlib/Artifacts/src/Artifacts.jl
KristofferC: So `args...` is really a path here? I don't get why it is accepting a vararg though, is it to not have the caller have to do a `joinpath`? Seems easier to read if it was just one argument and also likely better from the compiler point of view that you don't compile different versions for different number of arguments.

base/binaryplatforms.jl
KristofferC: In my opinion, it would be good to know, what are the exact things that need to be exported to make old JLLs work and what are the exact features that the new JLLs need? These are the things we should not break in the future Julias. With the exception of that, I think almost everything else should be considered internal.

NEWS.md
KristofferC: Show them `@time import Plots` instead or `@time import Gtk`?  :P Will likely impress move.

base/binaryplatforms.jl
KristofferC, line 59:
```suggestion
            tag = lowercase(String(tag)::String)
```

would make the compiler infer that `tag` is a `String` even though we don't know what `tag` is.

stdlib/Artifacts/src/Artifacts.jl
KristofferC: I understand now.

stdlib/Artifacts/src/Artifacts.jl
```diff
@@ -0,0 +1,634 @@
...
+        end
+        function parse_mapping(mapping::Dict, name::String)
+            return Dict(k => parse_mapping(v, name) for (k, v) in mapping)
+        end
+
+        for (k, mapping) in depot_override_dict
+            # First, parse the mapping. Is it an absolute path, a valid SHA1-hash, or neither?
+            try
+                mapping = parse_mapping(mapping, k)
+            catch
```
KristofferC, line 124: What are the exceptions that can be thrown here? If it is only about the `Base.SHA1` call then I think it makes sense to return `nothing` from `parse_mapping` and switch on that instead of a catch all `catch` that would just give you a log message even if someone later made a typo inside `parse_mapping`. This exact thing happened at https://github.com/JuliaLang/Pkg.jl/pull/1993.

stdlib/Artifacts/src/Artifacts.jl
```diff
@@ -0,0 +1,634 @@
...
+                    continue
+                end
+
+                # If this mapping is the empty string, un-override it
+                if mapping == ""
+                    delete!(overrides[:hash], hash)
+                else
+                    overrides[:hash][hash] = mapping
+                end
+            elseif isa(mapping, Dict{String, Any})
```
KristofferC, line 145: maybe make this into an `else` and type assert or an `else` and error. It just feels a little bit scary that we maybe missed that `mapping` is some other type and that it would then just go through and do nothing?

base/binaryplatforms.jl
staticfloat: I'm a little confused then because there are no abstract types here.  `Platform` is concrete, and `tags` is also concrete, it's a `Dict{String,String}`.  I don't want to define a `tags()` for other `AbstractPlatform` types because they will likely do something completely different (e.g. `AnyPlatform` won't have _any_ tags, it will just return a constant empty set, and other objects like the compatibility `Linux` type will sub out to a `tags()` on a wrapped `Platform` object).

staticfloat:
> This is not really true. There are many things that are public that are not exported (and cannot be exported due to e.g. name clashes). For example Profile.print, TOML.parse etc.

These are exceptions due to name clashes, it's not good practice to have public APIs that are unexported.  In fact, it is then even more important to have documentation stating that `TOML.parse` is a public API.

> Also, in packages, you might now want to export something that has a very usual name that is likely to be too generic to be understandable on its own.

I'm not sure what you're saying here; but naming of documented, non-exported symbols can be changed before exporting.

> What matters for public/private is documentation. With few exceptions, only public-facing API in the Julia repo has docstrings.

I strongly disagree.  While it may be the case that developers document public APIs more than internal ones, I argue that's a bad situation we accept because of limited developer time, not an ideal to be strived for.  We should never look down on more comments and documentation to help understand code.  Documentation is a tool to help people understand code, it is never desirable to make code less easy to understand.

stdlib/Artifacts/src/Artifacts.jl
staticfloat: Yeah, it was meant to mimic `abspath()` in that regard.  It does appear that I never ended up using that functionality, so I'll just restrict it to a single string.

stdlib/Artifacts/src/Artifacts.jl
staticfloat: Considering that `mapping` is coming from `parse_mapping`, I think the types are very tightly controlled.  But I believe in myself less than the compiler, so I still think it's a good idea.  :)

base/binaryplatforms.jl
staticfloat: Huh, I didn't realize that that was necessary.  Are there `String()` methods that don't return `String`?

base/binaryplatforms.jl
staticfloat: Yeah, I kind of thought the compiler would notice it was constant at compile-time and would store the completed object.  It looks like it doesn't do that though.

base/binaryplatforms.jl
timholy: Then it doesn't matter. I was going by the current implementation for which
```julia
julia> isabstracttype(Pkg.BinaryPlatforms.Platform)
true
```

base/binaryplatforms.jl
timholy:
```julia
julia> Base.return_types(String, (Any,))
9-element Vector{Any}:
 String
 String
 String
 String
 Any
 String
 String
 String
 String

julia> f(c) = String(c[1])

julia> @code_typed f(Any[1])
CodeInfo(
1 ─ %1 = Base.arrayref(true, c, 1)::Any
│   %2 = Main.String(%1)::Any
└──      return %2
) => Any
```

timholy: Though here I wonder if the compiler knows `tag` is a `Symbol` since it's a kwarg name.

timholy:
FWIW I like docstrings on internal methods too.

giordano:
I heard a few times the claim that public methods are those that have the docstring.  Don't we really have a better way to identify public and internal methods than having or not a docstring?

timholy:
Yeah, it seems silly to distinguish them that way. When I write Documenter docs, I only "publish" the ones I expect people to call, and I ignore the warnings about those internal functions that I've written a docstring for but which are really the internal interface.

base/binaryplatforms.jl
KristofferC: > so only users that run using Base.BinaryPlatforms are going to see any of these names.

Ok, and `Base` doesn't export `BinaryPlatforms` so everything is there is subject to change anyway. Then you can indeed export whatever.

base/binaryplatforms.jl
KristofferC: >  I don't want to define a tags() for other AbstractPlatform types because they will likely do something completely different 

I assumed you would since there is a bunch of code that calls `tags` on `AbstractPlatform` e.g.

```
 Base.haskey(p::AbstractPlatform, k::String) = haskey(tags(p), k)
```

and

```
# Simple equality definition; for compatibility testing, use `platforms_match()`
Base.:(==)(a::AbstractPlatform, b::AbstractPlatform) = tags(a) == tags(b)
```

etc. Should these be restricted to `Platform` then?

KristofferC:
> These are exceptions due to name clashes, it's not good practice to have public APIs that are unexported. 

Where does this notion come from? `export` is used to put something in the global namespace which is a shared resource and as you use more packages it will inherently start erroring. Many packages break as soon as we export something more from Base.
Also, for code readers, it makes it impossible to locally see where a symbol is coming from without help from an IDE which is awful when you are trying to get familiar with a new codebase. Anyway, this is a side discussion.


> I strongly disagree. While it may be the case that developers document public APIs more than internal ones, I argue that's a bad situation we accept because of limited developer time, not an ideal to be strived for. We should never look down on more comments and documentation to help understand code. Documentation is a tool to help people understand code, it is never desirable to make code less easy to understand.

When I said "documentation determines public/private", I should have been more specific and said, "public-facing documentation determines public/private". Adding a docstring to something indeed doesn't make it public, putting it into the manual does.

KristofferC:
I mean, the reason why I bring this up is that I see quite a few places which e.g. could have problems with invalidations (and one of the main reasons this code movement is being done is AFAIU due to compilation time). However, I don't want to nitpick every single place when it might not be a problem in practice. Therefore, if we have the freedom to change things, then it is probably better to merge, and to these fixes later.

But now e.g. `parse_dl_name_version(path::AbstractString, platform::AbstractPlatform)` is exported and documented so if we find that there are a bunch of invalidations from that and want to tightent it to `Union{String, SubString{String}}` are we then breaking the interface? It's exported so maybe? But the module itself is not so maybe not? But the jll packages will use `BinaryPlatforms` so I guess it is effectively public. But the jll packages might only call this with `String`. So is it ok to change? Etc.

stdlib/Artifacts/src/Artifacts.jl
```diff
@@ -0,0 +1,642 @@
...
+        return quote
+            local artifact_name, artifact_path_tail, hash = artifact_slash_lookup($(esc(name)), $(artifact_dict), $(artifacts_toml))
+            Base.invokelatest(_artifact_str, $(__module__), $(artifacts_toml), artifact_name, artifact_path_tail, $(artifact_dict), hash)
+        end
+    end
+end
+
+# Support `AbstractString`s, but avoid compilers needing to track backedges for callers
+# of these functions in case a user defines a new type that is `<: AbstractString`
+parse_toml(path::AbstractString) = parse_toml(string(path))
```
KristofferC, line 620:
```suggestion
parse_toml(path::AbstractString) = parse_toml(string(path)::String)
```

are needed.

But perhaps it is enough to restrict to `Union{String, SubString{String}}`?

base/binaryplatforms.jl
```diff
@@ -0,0 +1,999 @@
...
+```jldoctest
+julia> call_abi(Platform("armv7l", "Linux"))
+"eabihf"
+
+julia> call_Abi(Platform("x86_64", "macos"))
+```
+"""
+call_abi(p::AbstractPlatform) = get(tags(p), "call_abi", nothing)
+
+const platform_namess = Dict(
```
KristofferC, line 396: Typo I think?

Assuming that there is no intention to dynamically add platforms, `platform_name` below could be written as

```
function platform_name(p::AbstractPlatform)
    n = os(p)
    n === nothing && return "Unknown"
    n == "linux"   ? "Linux"   :
    n == "macos"   ? "macOS"   :
    n == "windows" ? "Windows" :
    n == "freebsd" ? "FreeBSD" :
    error("unhandled os $(repr(n))")
end
```

which avoids having to have the dictionary at all, makes it clear that `platform_namess` is not something that is extended and showes that there is a possible unrecoverable error when something unexpected is returned from `os`. The real hardcore version would be to define with an `@enum PlatformType` the different platforms listed here and make `os` return a `Union{Nothing, PlatformType}`  thus making the error impossible by the type system.

KristofferC:
Anyway, I didn't mean for my attempt at a review to start this type of a "heated" side discussion about stuff that is not directly related to the code in the PR. I just commented on some of my thoughts during reading the code and it was probably mostly a reflection of me being unfamiliar with the code and how it is being used in combination with that it didn't really use the standard documentation style as the rest of Base / stdlibs. In hindsight, I should probably have taken a bit different strategy.

The best way forward is probably to try to get stuff merged without too much nitpicking, get stuff in place, regenerate some jlls that can be used to test things out on, and then do eventual adjustments based on that.

base/binaryplatforms.jl
staticfloat: > Then it doesn't matter. I was going by the current implementation for which

Gotcha!  Makes sense.

> I assumed you would since there is a bunch of code that calls `tags` on AbstractPlatform

So the semantics of what `tags` does is stable; `tags(p::AbstractPlatform)` can be assumed to return a `Dict{String,String}`, but how it generates that dict is completely up to the `AbstractPlatform` object.  For an `AnyPlatform` object it will likely return an empty Dict.  For a `Linux` object (a compatibility wrapper for older BinaryProvider releases and such) it will sub out to an internal `Platform` object, etc...  So I still want to call `tags()` on other `AbstractPlatforms`, but I'm leaving it up to those other `AbstractPlatform` types to define just what needs to happen inside of `tags()`.

base/binaryplatforms.jl
staticfloat: > Typo I think?

Yes, thank you!

> Assuming that there is no intention to dynamically add platforms

I think having this escape hatch open is prudent.  While I don't expect to add any more future platforms in the near future, who knows, perhaps we'll add OpenBSD support or something and then we'll need to modify this.  It would be a lot easier if our JLL packages can push something into this dictionary via a Pkg hook or something.  :)

stdlib/Artifacts/src/Artifacts.jl
staticfloat: Aha, I figured out why I had this; it was actually as a convenient definition for `artifacts_dirs()`, which I use in `Pkg`.

staticfloat:
Alright, I've dropped the commit that bumps `Pkg`; this is now ready to merge.  Once this is merged, I'll merge [the PR against `Pkg`](https://github.com/JuliaLang/Pkg.jl/pull/2001), then we can bump `Pkg` in a future PR which will switch Julia over to a `Pkg` version that drops its duplicate `Artifacts` and `BinaryPlatforms` code.

stevengj:
a `git bisect` seems to indicate that this PR is triggering JuliaLang/IJulia.jl#968 … any ideas?

PR #33753, opened 2019-11-03
Only compute effect-free once

Keno:
At the moment, whenever a statement becomes unused the compiler
tries to determine whether that statement is effect free and
may thus be removed. This is wasteful, because effect-freeness
should be computable only from type information. This PR moves
that computation to the start of the IRCode construction phase
and caches the result in the flags array. Ideally, I'd like to
just propagate these kinds of flags interprocedurally during
inference, but I'd like to do this one step at a time.

PR #37333, opened 2020-09-01
inference: attach "Const.actual" flag to call instead of result type

vtjnash:
Been wanting to fix this for a long time, but finally it has become convenient.

"Const.actual" was not a property of the type-lattice but of the call, so it was a bit odd to track in this type (since it should not propagate). Instead, we can attach this to the stmt_info field now, where inlining will find it when needed.

(Possibly somewhat related to https://github.com/JuliaLang/julia/pull/33753 also.)

### DIFF
--- a/base/compiler/abstractinterpretation.jl
+++ b/base/compiler/abstractinterpretation.jl
@@ -95,7 +95,8 @@ function abstract_call_gf_by_type(interp::AbstractInterpreter, @nospecialize(f),
     if f !== nothing && napplicable == 1 && is_method_pure(applicable[1]::MethodMatch)
         val = pure_eval_call(f, argtypes)
         if val !== false
-            return CallMeta(val, info)
+            # TODO: add some sort of edge(s)
+            return CallMeta(val, MethodResultPure())
         end
     end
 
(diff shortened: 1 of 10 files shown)


base/compiler/abstractinterpretation.jl
```diff
@@ -95,7 +95,8 @@ function abstract_call_gf_by_type(interp::AbstractInterpreter, @nospecialize(f),
     if f !== nothing && napplicable == 1 && is_method_pure(applicable[1]::MethodMatch)
         val = pure_eval_call(f, argtypes)
         if val !== false
-            return CallMeta(val, info)
+            # TODO: add some sort of edge(s)
+            return CallMeta(val, MethodResultPure())
```
Keno, line 99: Other future consumers of stmt info (e.g. Cthulhu) may want to look at the `info` here, probably, but since that's not ready yet, I think we can cross that bridge when we get there.

Keno (approved):

base/compiler/abstractinterpretation.jl
vtjnash: Sure, but when we hit a call we think might be pure because of `info`, we're then ignoring the actual `info` when we do the `pure_eval_call`, so I think this is actually a bit more accurate.

PR #37340, opened 2020-09-02
add Downloads stdlib

StefanKarpinski:
This PR adds a Downloads stdlib and replaces the `Base.download` function which relies on a miscellany of external programs (`curl`, `wget`, `PowerShell`) which may or may not happen to be present in order to download files, with `Downloads.download` which uses `libcurl` to download files in the same way everywhere. In the process, we also gain support for passing headers and more (to be exposed and used in Pkg in future PRs).


stdlib/LibCURL.version
```diff
@@ -0,0 +1,2 @@
+LIBCURL_BRANCH = master
+LIBCURL_SHA1 = d018fe70928510b552600473e136de5d9bbe65b3
```
KristofferC, line 2: Is this just the same as https://github.com/JuliaWeb/LibCURL.jl?

I think you need to at least change the UUID otherwise when someone uses an existing Manifest with that package they will immidiately get the one in the sysimage instead of the version they have in their manifest.

StefanKarpinski: That's what's inevitably going to happen with this. I've been saying that we need a way to have private snapshots of external packages for a while now. However, I'm first going to get this PR working and passing tests by any means necessary and then we can figure out what to do about all the f*¢&ing dependencies.

KristofferC: > That's what's inevitably going to happen with this PR

Why inevitably? You could just change the UUID so it is not considered the same package. Manifests using the old package will use it, manifests using the new stdlib LibCURL will keep using that. Also, if we use the same UUID then we are effectively "hijacking" that package since it will no longer be updatable by its authors, post 1.6. The authors of LibCURL.jl might very well be fine with it but has there been any discussion about that?

> and others that we need to merge in 1.6

are there even more external packages that are going into stdlib for 1.6?

> we can figure out what to do about all the f*¢&ing dependencies.

Yeah, writing stdlibs sucks because you can't really use all the nice dependencies you want. But the strategy can't really be to work on a package externally, just like a normal package, decide that it should be an stdlib and then just "oh well, guess we have to put all its deps into stdlib as well"?

StefanKarpinski: What I mean by inevitably is that one of the following has to happen:

1. Rewrite `Downloader` to not use the full `LibCURL` wrapper, copy used methods
2. Add `LibCURL` as a stdlib, which prevents users from using it as a normal package
3. Add `LibCURL′` as a stdlib with a different UUID and a snapshot of the code (your suggestion)
4. Add a "private dependency" mechanism for doing 3 automatically and systematically

For now I'm just going with 2 but I agree that it's not a viable approach for a release, I just want to get this working with minimal fuss. Doing 3 is fine but it's pretty ugly and I think something like 4 would be better.

There's a general concept of a "private dependency" that can be extracted: if you want to use a package as a dependency but you promise that no objects of any types defined by that package will leak from your usage of it, then you can have a private copy of it. The general concept of public versus private dependencies is that there can only be a single public instance of a package but arbitrarily many private instances. All packages that have a public dependency on a package have to be able to agree on a version they can use, but if a package with a private dependency has a version conflict with another package with a public or private dependency on it, then they can just use a different copy with a different version.

This concept solves the issue of stdlibs wanting to use non-stdlib packages: as long as they use something as a private dependency, then it's fine. If something else has a public dependency on it that's compatible with the private copy that we've baked into the sysimg, then they can just use that one. If they need a different version, then we give them a different copy.

This also solves the very similar problem that VS Code, JuliaPro, etc. have: they need to have some packages (e.g. JSON) pre-loaded in the client process in order to function properly, but then they also want user code to be able to use whatever version of those packages they want. Currently this just doesn't work and if you're running code that wants to use certain versions of packages, they simply can't and their application might breaak. If they use these packages as private dependencies, then everything works fine: the IDE code uses the packages as private dependencies and the user code gets whatever version it wants.

I'm not sure we can get a feature like this into 1.6, but it seems like a really important concept that would solve many problems in the ecosystem in one fell swoop.

StefanKarpinski:
Aside from passing CI, I think this PR is mergeable. @KristofferC, you had marked it as "needs docs" but I'm not entirely sure why—was that because the Downloads stdlib doesn't have docs? Currently, I'm just viewing this as a replacement for existing functionality and I've updated the `Base.download` docs in this PR.

StefanKarpinski:
The Win64 tester keeps hanging, I have no idea why. It was passing previously, so I'm not sure if this is related to this PR or just because there's something up with the infrastructure. @staticfloat, any idea what's up?

base/download.jl
```diff
@@ -55,50 +9,22 @@ function download_url(url::AbstractString)
...
-        error("No download agent available; install curl, wget, busybox or fetch.")
-    end
-    return filename
-end
-
-function download(url::AbstractString)
-    filename = tempname()
-    download(url, filename)
-end
+Downloads() = require(PkgId(
```
KristofferC, line 12: Should `Downloads` just be in Base then (doesn't have to be done now but later)? So far, we've enabled optional features via the presence of standard libraries but now anything that doesn't include the `Downloads` stdlib is arguably a broken build of Julia.

StefanKarpinski: It seems better to continue to allow `Downloads` to be a separate body of code which can be used independently. For example, it can be loaded and used on Julia 1.3 as a normal package. If someone is using a Julia build that doesn't have `Downloads` in the sysimg, as long as they don't call `download` everything should be fine. The main improvement here seems like it would be to throw a clearer error if the `Downloads` package isn't a stdlib and hasn't been loaded.

KristofferC: I guess I can specialize PackageCompiler to never remove the Downloads stdlib so it is always considered required. It's starting to blur the difference between Julia and the stdlibs a bit uncomfortably though. We already have some of that blur with the type piracy in LinearAlgebra and Random but ideally, it is good to try not to introduce more.

musm:
This is awesome, and closes many long standing issues. 

If this is now a stdlib, shouldn't there now be a depwarn for   `Base.download`?

StefanKarpinski:
Yes, I could add one since deprecation warnings default to off these days, right?

base/download.jl
StefanKarpinski: I don't really understand the issue. If `Downloads` isn't included but `Base.download` is never called, isn't it fine?

StefanKarpinski:
> This is awesome, and closes many long standing issues.

Yep, and the next step is to use it for downloads in Pkg as well, so even more problems solved there. Figuring out how to use libcurl correctly from Julia with libuv was pretty tricky, but now that it works, libcurl is great: it's very portable and has every feature we might ever want for downloading/uploading stuff—all the protocols, progress support, authentication, etc. Since I have a good understanding of the API now, if we need any additional features in the future, I'm pretty confident that libcurl has them and that we can add them easily.

KristofferC:
Regarding the docs question, I just meant an entry in the manual so it is listed as an stdlib with the API it has, like the other stdlibs.

StefanKarpinski:
Right, I can add that once I've got it working.

vtjnash:
Marked as a draft while we figure out the win64 issue, so you can mark it ready-for-review when it's working

StefanKarpinski:
Ok, well, this finally works. Turns out the Win64 hang was a libcurl bug. Upgrading libcurl fixes it, but requires adding nghttp2 as a dependency. That's fine, since we want to support HTTP/2 anyway. For the record fde145d2b90292114c8ec01507fddfc5425c53b5 was the commit that worked.

StefanKarpinski:
Anything else needed here besides news?

StefanKarpinski:
Deprecating Base.download is non-trivial since we run tests with `depwarn=error`. As these things so often seem to be, it's more involved than you'd imagine. Having `Base.download` be an error during testing can be worked around and I have a branch that does that, but LibCURL uses `Base.download` in its tests, which presents a particular challenge. The best fix for that is to just change it to use `Downloads.download` instead of `Base.download`, but in order for that to work on older Julia's, it requires `Downloads` to be registered as an official package, a process that I've [started](https://github.com/JuliaRegistries/General/pull/21493) but which won't go through for another three days. Once that happens, I can fix LibCURL's tests to use `Downloads` as a normal package on 1.3-1.5 and as a stdlib on 1.6+, after which it will pass tests as a stdlib with `depwarn=error` and we can actually deprecate `Base.download` without breaking CI.

ViralBShah:
This is pretty amazing! 

Off-topic, we really need to move away from having a file for each download hash.

haampie:
@StefanKarpinski can you try `] add JuMP` on this PR, seems like it's broken:

```
(Pkg) pkg> add JuMP
   Updating registry at `/dev/shm/.julia/registries/General`
  Resolving package versions...
ERROR: Dependency `Test` in target `test` not listed in `deps` or `extras` section.
```

stdlib/LibCURL_jll/Project.toml
```diff
@@ -0,0 +1,11 @@
+name = "LibCURL_jll"
+uuid = "deac9b47-8bc7-5906-a0fe-35ac56dc84c0"
+
+[deps]
+Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
+
+[extra]
```
fredrikekre, line 7: s

stdlib/MozillaCACerts_jll/Project.toml
```diff
@@ -0,0 +1,8 @@
+name = "MozillaCACerts_jll"
+uuid = "14a3606d-f60d-562e-9121-12d972cd8159"
+
+[extra]
```
haampie, line 4: Same here

fredrikekre:
> ERROR: Dependency `Test` in target `test` not listed in `deps` or `extras` section.

Where is this thrown from?

haampie:
I've already compiled an older commit, so I cannot easily go back to see :(

Issue #37342, opened 2020-09-02
Type check in callee does not propagate to caller


yuyichao:
Somewhat similar to https://github.com/JuliaLang/julia/issues/25632 but more general and maybe more relavant as well.

```julia
julia> isType(@nospecialize t) = isa(t, DataType) && t.name === Core.Compiler._TYPE_NAME
isType (generic function with 1 method)

julia> function f(r)
           t = r[]
           if isType(t)
               return t.parameters[1]
           end
       end
f (generic function with 1 method)

julia> @code_typed f(Ref{Any}())
CodeInfo(
1 ─ %1  = Base.getfield(r, :x)::Any
│   %2  = (%1 isa Main.DataType)::Bool
└──       goto #3 if not %2
2 ─ %4  = π (%1, DataType)
│   %5  = Base.getfield(%4, :name)::Core.TypeName
│   %6  = Core.Compiler::Module
│   %7  = Base.getfield(%6, :_TYPE_NAME)::Core.TypeName
│   %8  = (%5 === %7)::Bool
└──       goto #4
3 ─       goto #4
4 ┄ %11 = φ (#2 => %8, #3 => false)::Bool
└──       goto #6 if not %11
5 ─ %13 = Base.getproperty(%1, :parameters)::Any
│   %14 = Base.getindex(%13, 1)::Any
└──       return %14
6 ─       return nothing
) => Any
```

Note that the access of the `parameters` field and subsequent calls are not inferred. If `isType` is manually inlined then the function works just fine. (The `@nospecialize` is due to copy-and-paste from type inference and does not have any effect on the issue.)

-------------

Noticed this when testing https://github.com/JuliaLang/julia/pull/37303 a while ago. Dynamic dispatch like this happens **everywhere** in inference/optimizer code (including most callers of `typeutils.jl` functions) and I've verified from LLVM IR dump that this is actually happening in the final code. I don't know how much performance issue it costs but could be significant... (note that it won't be as bad as the code above since `Core.Compiler.getproperty === getfield`). Adding type assertions back in to inference code could be a temporary solution but fixing the inference itself would be better...

PR #37343, opened 2020-09-02
Make Base.ifelse a generic function

### LINKED ISSUE #32844: Request: export ifelse() as normal (extensible) function
[As suggested on discourse](https://discourse.julialang.org/t/extending-builtin-functions-like-ifelse/27319/4), I'd like to propose exporting `ifelse()` as a normal function rather than a builtin.

We have a [usecase in NumberIntervals.jl](https://github.com/gwater/NumberIntervals.jl/issues/14) for extending `ifelse()` to do some three-value logic which is currently impossible for builtins.

CC @yuyichao

c42f:
Fixes #32844, in particular, to provide a way for packages to extend `ifelse` without needing a new package for it, such as in https://github.com/SciML/ModelingToolkit.jl/pull/568.  CC @ChrisRackauckas 

This is a WIP because it regresses some inference support for `ifelse` (likely undoing some of the work from #27068), in particular, the following test case now fails:

```julia
@noinline _f_ifelse_isa_() = rand(Bool) ? 1 : nothing
function _g_ifelse_isa_()
    x = _f_ifelse_isa_()
    ifelse(isa(x, Nothing), 1, x)
end
@test Base.return_types(_g_ifelse_isa_, ()) == [Int]
```

**Inference question**: `Core.ifelse` is handled in `abstract_call_builtin()` but it seems that the necessary information doesn't propagate from `Base.ifelse` to `Core.ifelse`. Why? Is this something we can expect to repair?

### DIFF
--- a/base/essentials.jl
+++ b/base/essentials.jl
@@ -477,6 +477,22 @@ Stacktrace:
 """
 sizeof(x) = Core.sizeof(x)
 
+"""
+    ifelse(condition::Bool, x, y)
+
+Return `x` if `condition` is `true`, otherwise return `y`. This differs from `?` or `if` in
+that it is an ordinary function, so all the arguments are evaluated first. In some cases,
+using `ifelse` instead of an `if` statement can eliminate the branch in generated code and
+provide higher performance in tight loops.
+
+# Examples
+```jldoctest
+julia> ifelse(1 > 2, 1, 2)
+2
+```
+"""
+ifelse(condition::Bool, x, y) = Core.ifelse(condition, x, y)
+
 # simple Array{Any} operations needed for bootstrap
 @eval setindex!(A::Array{Any}, @nospecialize(x), i::Int) = arrayset($(Expr(:boundscheck)), A, x, i)
 
--- a/base/boot.jl
+++ b/base/boot.jl
@@ -192,8 +192,8 @@ export
     Expr, QuoteNode, LineNumberNode, GlobalRef,
     # object model functions
     fieldtype, getfield, setfield!, swapfield!, modifyfield!, replacefield!,
-    nfields, throw, tuple, ===, isdefined, eval, ifelse,
-    # sizeof    # not exported, to avoid conflicting with Base.sizeof
+    nfields, throw, tuple, ===, isdefined, eval,
+    # ifelse, sizeof    # not exported, to avoid conflicting with Base
     # type reflection
     <:, typeof, isa, typeassert,
     # method reflection
--- a/base/compiler/abstractinterpretation.jl
+++ b/base/compiler/abstractinterpretation.jl
@@ -1082,7 +1082,7 @@ function abstract_call_builtin(interp::AbstractInterpreter, f::Builtin, (; fargs
                                sv::InferenceState, max_methods::Int)
     @nospecialize f
     la = length(argtypes)
-    if f === ifelse && fargs isa Vector{Any} && la == 4
+    if f === Core.ifelse && fargs isa Vector{Any} && la == 4
         cnd = argtypes[2]
         if isa(cnd, Conditional)
             newcnd = widenconditional(cnd)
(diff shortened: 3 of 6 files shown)


ChrisRackauckas:
As an upgrade path, I created https://github.com/SciML/IfElse.jl for me and @chriselrod (and whoever else) to start usinga common form, and when something gets added to Base we can deprecate that.

vtjnash:
That question is the same as https://github.com/JuliaLang/julia/issues/37342

base/essentials.jl
```diff
@@ -456,6 +456,22 @@ Stacktrace:
...
+using `ifelse` instead of an `if` statement can eliminate the branch in generated code and
+provide higher performance in tight loops.
+
+# Examples
+```jldoctest
+julia> ifelse(1 > 2, 1, 2)
+2
+```
+"""
+ifelse(condition, x, y) = Core.ifelse(condition, x, y)
```
JeffBezanson, line 473:
```suggestion
ifelse(condition::Bool, x, y) = Core.ifelse(condition, x, y)
```

vtjnash:
#37342 is fixed now. So, do we want to merge this or close it now?

chriselrod:
I'm in favor of merging it.

aviatesk:
Now I think the inference improvement is better to be done separately from this PR. It will be a requirement for this PR, but just a general improvement. xref: https://github.com/JuliaLang/julia/pull/42529.

chriselrod:
Now that https://github.com/JuliaLang/julia/pull/42529#event-5499541187 has merged, rebase?

c42f:
Some of those performance regressions look kind of bad. Should we revert this for now?

c42f:
@aviatesk do you have any thoughts about the performance regressions from the nanosolider run and whether there's anything that could be done about them in the short term?

aviatesk:
I don't think there is much performance difference.
AFAIU the benchmark time isn't that reliable measure of performance (where any increase in memory allocation usually indicates performance regression).

I compared the performance difference of `["sparse", "matmul", ("At_mul_B", "dense 50x5, sparse 50x500 -> dense 5x500")]` as an example ​with the following script:
```julia
include(normpath(ENV["JULIA_PKG_DEVDIR"], "BaseBenchmarks", "src", "utils", "RandUtils.jl"))

using .RandUtils
using BenchmarkTools
using SparseArrays
using LinearAlgebra

using LinearAlgebra: *, mul!

function allocmats_ds(om, ok, on, s, nnzc, T)
   ​m, k, n = map(x -> Int(s*x), (om, ok, on))
   ​densemat, sparsemat = samerand(T, m, k), samesprand(T, k, n, nnzc/k)
   ​tdensemat = transpose!(similar(densemat, reverse(size(densemat))), densemat)
   ​tsparsemat = transpose!(similar(sparsemat, reverse(size(sparsemat))), sparsemat)
   ​destmat = similar(densemat, m, n)
   ​return m, k, n, destmat,
       ​densemat, sparsemat,
       ​tdensemat, tsparsemat
end

for (om, ok, on) in (# order of matmul dimensions m, k, and n
   ​(10^2, 10^2, 10^2),  # dense square * sparse square -> dense square
   ​(10^1, 10^1, 10^3),  # dense square * sparse short -> dense short
   ​(10^2, 10^2, 10^1),  # dense square * sparse tall -> dense tall
   ​(10^1, 10^3, 10^3),  # dense short * sparse square -> dense short
   ​(10^1, 10^2, 10^3),  # dense short * sparse short -> dense short
   ​(10^1, 10^3, 10^2),  # dense short * sparse tall -> dense short
   ​(10^3, 10^1, 10^1),  # dense tall * sparse square -> dense tall
   ​(10^2, 10^1, 10^2),  # dense tall * sparse short -> dense square
   ​) # the preceding descriptions apply to dense-sparse matmul without
     ​# any transpositions. other cases are described below
   ​m, k, n, destmat, densemat, sparsemat, tdensemat, tsparsemat = allocmats_ds(om, ok, on, 1/2, 4, Float64)
   ​display(@benchmark *($densemat, $sparsemat))
   ​display(@benchmark *($densemat, $(Transpose(tsparsemat))))
end
```
But I couldn't find any effective performance difference on `master` and `master with this PR reverted`.

PR #37352, opened 2020-09-02
added namedtuple convenience function for constructing NamedTuple

ExpandingMan:
I find constructing named tuples to be somewhat confusing, and when I mentioned this on slack today I got 6 check mark emojis in what felt like 5 minutes, so I'm apparently not the only one.

This pull adds the `namedtuple` constructor function, in analogy to `tuple`, which adds some convenient constructors.  An argument in favor of adding these is that, even though the method bodies here are quite short, I often have a hard time remembering what is the "right way" to construct these, and I suspect I'm not alone.

From some quick benchmarking, it looks to me like `NamedTuple` is normally *much* faster than splatting, so having this constructor function might have people reaching for splatting less often, which seems to be a good thing.  It also seems more intuitive in a lot of cases, again, I find myself very often fumbling with things like `(;zip(K, V)...)` not necessarily immediately thinking that's a way to do it, since it's not really a common Julia pattern for anything else.

### DIFF
--- a/base/namedtuple.jl
+++ b/base/namedtuple.jl
@@ -394,7 +394,7 @@ keys must be of type `Symbol`.
 # Examples
 ```jldoctest
 julia> namedtuple([:a, :b, :c], [1, 2, 3])
- (a = 1, b = 2, c = 3)
+(a = 1, b = 2, c = 3)
 ```
 """
 namedtuple(K, V) = NamedTuple{Tuple(K)}(V)
--- a/base/namedtuple.jl
+++ b/base/namedtuple.jl
@@ -384,3 +384,33 @@ macro NamedTuple(ex)
     types = [esc(e isa Symbol ? :Any : e.args[2]) for e in decls]
     return :(NamedTuple{($(vars...),), Tuple{$(types...)}})
 end
+
+"""
+    namedtuple(K, V)
+
+Creates a `NamedTuple` object out of `K`, an iterator of keys and `V`, an iterator of values.  All
+keys must be of type `Symbol`.
+
+# Examples
+```jldoctest
+julia> namedtuple([:a, :b, :c], [1, 2, 3])
+ (a = 1, b = 2, c = 3)
+```
+"""
+namedtuple(K, V) = NamedTuple{Tuple(K)}(V)
+
+"""
+    namedtuple(d)
+
+Create a `NamedTuple` object from any iterable of `Pair`s `d` (for example, any `AbstractDict`).
+The first value in each pair must be of type `Symbol`.
+"""
+namedtuple(d::AbstractDict) = namedtuple(keys(d), values(d))
+function namedtuple(d)
+    NamedTuple{Tuple(first(p) for p ∈ d)}(last(p) for p ∈ d)
+end
+
+namedtuple(nt::NamedTuple) = nt
+
+
+export namedtuple
--- a/base/namedtuple.jl
+++ b/base/namedtuple.jl
@@ -407,7 +407,8 @@ The first value in each pair must be of type `Symbol`.
 """
 namedtuple(d::AbstractDict) = namedtuple(keys(d), values(d))
 function namedtuple(d)
-    NamedTuple{Tuple(first(p) for p ∈ d)}(last(p) for p ∈ d)
+    t = Tuple(d)
+    NamedTuple{map(first, t)}(map(last, t))
 end
 
 namedtuple(nt::NamedTuple) = nt
--- a/test/namedtuple.jl
+++ b/test/namedtuple.jl
@@ -304,3 +304,10 @@ let x = 1, y = 2
     @test Meta.lower(Main, Meta.parse("(; a.y, y)")) == Expr(:error, "field name \"y\" repeated in named tuple")
     @test (; a.y, x) === (y=2, x=1)
 end
+
+@testset "namedtuple" begin
+    @test namedtuple([:a, :b, :c], [1, 2, 3]) == (a=1, b=2, c=3)
+    @test namedtuple((:kirk, :spock, :bones), (im, ℯ, π)) == (kirk=im, spock=ℯ, bones=π)
+    @test namedtuple(Dict(:a=>1, :b=>2, :c=>3)) == (a=1, b=2, c=3)
+    @test namedtuple([:kirk=>im, :spock=>ℯ, :bones=>π]) == (kirk=im, spock=ℯ, bones=π)
+end


base/namedtuple.jl
```diff
@@ -384,3 +384,33 @@ macro NamedTuple(ex)
...
+
+"""
+    namedtuple(d)
+
+Create a `NamedTuple` object from any iterable of `Pair`s `d` (for example, any `AbstractDict`).
+The first value in each pair must be of type `Symbol`.
+"""
+namedtuple(d::AbstractDict) = namedtuple(keys(d), values(d))
+function namedtuple(d)
+    NamedTuple{Tuple(first(p) for p ∈ d)}(last(p) for p ∈ d)
```
KristofferC, line 410: Should probably not iterate `d` twice.

ExpandingMan: agreed, but it seems hard to avoid this without allocating some additional object, since in particular `Tuple` creates an immutable object.  Suggestions?

simeonschaub: You can just iterate `d` in a for loop and build up two tuples for the names and values by splatting.

ExpandingMan: I strongly suspect that will be slow, but I will try it

simeonschaub: An alternative might be to call `Tuple` on the whole dict first and then map `first` and `last` over that. That might also make constant propagation easier in some cases.

ExpandingMan: actually wait.. if I do it that way, wouldn't I still have to iterate twice to create the separate tuples?  Can you share an example; sorry if I'm being really thick here... either I'm just being an idiot or this is actually kind of hard to do...

simeonschaub: Something like:
```julia
keys, vals = (), ()
for (k, v) in d
    keys, vals = (keys..., k), (vals..., v)
end
```
Although I suspect that just `Tuple` and the `map` could be faster.

ExpandingMan: yeah I really can't imagine this being faster, but let me check...

KristofferC: It's not about being faster but about being correct for e.g. stateful iterators.

ExpandingMan: yeah, this seems a little bit slower... I'm getting between 800 ns and 900 ns for two elements everything I've tried so far which is undoubtedly incredibly slow.

It seems to me that the problem is that it's just generally hard to create multiple immutable objects in a single loop without creating extra allocations.

ExpandingMan: So far my original version of this scales better with size than anything else I've tried.

simeonschaub: For me, my second proposal is similarly fast than your original version (on master):
```julia
julia> using BenchmarkTools

julia> function f1(d)
           keys, vals = (), ()
           for (k, v) in d
               keys, vals = (keys..., k), (vals..., v)
           end
           NamedTuple{keys}(vals)
       end
f1 (generic function with 1 method)

julia> function f2(d)
           t = Tuple(d)
           NamedTuple{map(first, t)}(map(last, t))
       end
f2 (generic function with 1 method)

julia> function f3(d)
           NamedTuple{Tuple(first(i) for i in d)}(last(i) for i in d)
       end
f3 (generic function with 1 method)

julia> @btime f1([:a => 1, :b => 2])
  1.407 μs (13 allocations: 464 bytes)
(a = 1, b = 2)

julia> @btime f2([:a => 1, :b => 2])
  1.167 μs (11 allocations: 592 bytes)
(a = 1, b = 2)

julia> @btime f3([:a => 1, :b => 2])
  1.374 μs (10 allocations: 576 bytes)
(a = 1, b = 2)

julia> @btime f1([Symbol(k) => v for (k, v) in zip('a':'z', 1:26)])
  24.818 μs (135 allocations: 8.58 KiB)
(a = 1, b = 2, c = 3, d = 4, e = 5, f = 6, g = 7, h = 8, i = 9, j = 10, k = 11, l = 12, m = 13, n = 14, o = 15, p = 16, q = 17, r = 18, s = 19, t = 20, u = 21, v = 22, w = 23, x = 24, y = 25, z = 26)

julia> @btime f2([Symbol(k) => v for (k, v) in zip('a':'z', 1:26)])
  5.494 μs (63 allocations: 4.41 KiB)
(a = 1, b = 2, c = 3, d = 4, e = 5, f = 6, g = 7, h = 8, i = 9, j = 10, k = 11, l = 12, m = 13, n = 14, o = 15, p = 16, q = 17, r = 18, s = 19, t = 20, u = 21, v = 22, w = 23, x = 24, y = 25, z = 26)

julia> @btime f3([Symbol(k) => v for (k, v) in zip('a':'z', 1:26)])
  4.768 μs (36 allocations: 2.69 KiB)
(a = 1, b = 2, c = 3, d = 4, e = 5, f = 6, g = 7, h = 8, i = 9, j = 10, k = 11, l = 12, m = 13, n = 14, o = 15, p = 16, q = 17, r = 18, s = 19, t = 20, u = 21, v = 22, w = 23, x = 24, y = 25, z = 26)
```

simeonschaub: (interestingly, 1.5 is quite a bit faster here:
```julia
julia> @btime f1([:a => 1, :b => 2])
  975.231 ns (13 allocations: 464 bytes)
(a = 1, b = 2)

julia> @btime f2([:a => 1, :b => 2])
  691.838 ns (11 allocations: 592 bytes)
(a = 1, b = 2)

julia> @btime f3([:a => 1, :b => 2])
  867.345 ns (10 allocations: 576 bytes)
(a = 1, b = 2)
```
)

goretkin:
Is this the same as https://github.com/JeffreySarnoff/NamedTupleTools.jl#construction-from-names-and-values ? That's not intended to be an argument against including this in Base, but I wanted to reference `NamedTupleTools.jl`.

base/namedtuple.jl
ericphanson: I wonder if it makes more sense to benchmark bigger dictionaries rather than very small ones? Since it's an inherently type-unstable operation, so you'll be hit by dynamic dispatch for the next operation you do with the result, which means the speed here only matters when it's comparable to the cost of that dispatch

simeonschaub: I think with much more than 26 entries, you shouldn't really be using a `NamedTuple` though.

ericphanson: true, I hadn't seen those at the end of your benchmarks

ExpandingMan: Yeah, sorry, I'm actually on 1.5, I just never bothered building master.

Alright, seems like we have uncovered a significant performance regression.  I'm seeing similar results to what @simeonschaub is showing in that `f1`.

It looks like `f2` is the best option right now, so I'm changing it to that, commit incoming...

ExpandingMan:
Ok, I marked the previous inline conversation as resolved since I changed the method, but this is not an assertion that what I have here now is the best possible method.  This version seems to work well, but suggestions are welcome.

@goretkin, yes it seems NamedTupleTools.jl has similar methods.  Those all appear to use splatting, so the methods here should be faster in many cases.

JeffBezanson:
Instead I think we should add a method `merge(::NamedTuple{()}, ::Zip)`, which should make splatting just as fast.

> since it's not really a common Julia pattern for anything else.

If you have separate collections of keys and values, this is also how you construct a Dict from them (`Dict(zip(k,v))`).

We could probably add
```
NamedTuple(itr) = (; itr...)
```

This might be slightly inconsistent with `NamedTuple{names}(value_iterator)`, but I'm starting to think it's totally fine --- if the type already specifies the names, then the constructor only accepts values, but if you've specified neither then we expect key-value pairs. Makes sense.

ExpandingMan:
If we could make splatting just as fast (which if we can, of course we should anyway) and also have a function (perhaps just `NamedTuple`) which lets users create named tuples without explicitly splatting (because again, it's kind of a weird pattern, not least because historically there've been lots of performance issues with splatting), I'd be perfectly happy with that result.

base/namedtuple.jl
```diff
@@ -384,3 +384,34 @@ macro NamedTuple(ex)
...
+namedtuple(d::AbstractDict) = namedtuple(keys(d), values(d))
+function namedtuple(d)
+    t = Tuple(d)
+    NamedTuple{map(first, t)}(map(last, t))
+end
+
+namedtuple(nt::NamedTuple) = nt
+
+
+export namedtuple
```
ararslan, line 417: This should go in base/exports.jl.

ExpandingMan: Ah, I was wondering about that, will change

PR #37355, opened 2020-09-02
Emit more direct pointer comparisons when possible

yuyichao:
* `isa(..., ::Type{<concrete type>})`
* `isa(..., ::Type{<type with no type parameter>})`
* `... === <type with no type parameter>`
* `isa(..., ::<singleton type>)`
* Make sure `Union{}` is treated as singleton
* `===` when both sides are/can trivially be boxed and one side is
  a immutable type with guaranteed box cache.

Also fix `jl_pointer_egal` by blacklisting `Core.TypeofBottom` just like codegen/typeinference.

### DIFF
--- a/src/cgutils.cpp
+++ b/src/cgutils.cpp
@@ -1106,6 +1106,12 @@ static std::pair<Value*, bool> emit_isa(jl_codectx_t &ctx, const jl_cgval_t &x, 
         return std::make_pair(ConstantInt::get(T_int1, *known_isa), true);
     }
 
+    if (jl_is_type_type(intersected_type) && jl_pointer_egal(intersected_type)) {
+        // Use the check in `jl_pointer_egal` to see if the type enclosed
+        // has unique pointer value.
+        auto ptr = track_pjlvalue(ctx, literal_pointer_val(ctx, jl_tparam0(intersected_type)));
+        return {ctx.builder.CreateICmpEQ(boxed(ctx, x), ptr), false};
+    }
     // intersection with Type needs to be handled specially
     if (jl_has_intersect_type_not_kind(type) || jl_has_intersect_type_not_kind(intersected_type)) {
         Value *vx = boxed(ctx, x);
@@ -1148,6 +1154,10 @@ static std::pair<Value*, bool> emit_isa(jl_codectx_t &ctx, const jl_cgval_t &x, 
                 // handle the case where we know that `x` is unboxed (but of unknown type), but that concrete type `type` cannot be unboxed
                 return std::make_pair(ConstantInt::get(T_int1, 0), false);
             }
+        }
+        if (auto val = ((jl_datatype_t*)intersected_type)->instance) {
+            auto ptr = track_pjlvalue(ctx, literal_pointer_val(ctx, val));
+            return {ctx.builder.CreateICmpEQ(boxed(ctx, x), ptr), false};
         }
         return std::make_pair(ctx.builder.CreateICmpEQ(emit_typeof_boxed(ctx, x),
             track_pjlvalue(ctx, literal_pointer_val(ctx, intersected_type))), false);
--- a/src/codegen.cpp
+++ b/src/codegen.cpp
@@ -2504,6 +2504,10 @@ static Value *emit_bits_compare(jl_codectx_t &ctx, jl_cgval_t arg1, jl_cgval_t a
 // emit code for is (===).
 static Value *emit_f_is(jl_codectx_t &ctx, const jl_cgval_t &arg1, const jl_cgval_t &arg2)
 {
+    // handle simple static expressions with no side-effects
+    if (arg1.constant && arg2.constant)
+        return ConstantInt::get(T_int1, jl_egal(arg1.constant, arg2.constant));
+
     jl_value_t *rt1 = arg1.typ;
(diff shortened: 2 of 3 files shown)


src/cgutils.cpp
```diff
@@ -1106,6 +1106,12 @@ static std::pair<Value*, bool> emit_isa(jl_codectx_t &ctx, const jl_cgval_t &x,
         return std::make_pair(ConstantInt::get(T_int1, *known_isa), true);
     }
 
+    if (jl_is_type_type(intersected_type) && jl_pointer_egal(intersected_type)) {
+        // Use the check in `jl_pointer_egal` to see if the type enclosed
+        // has unique pointer value.
```
vtjnash, line 1111: Ah, nice, so this is the stronger version of `is_uniquerep_Type` apparently!

src/codegen.cpp
```diff
@@ -2523,6 +2529,21 @@ static Value *emit_f_is(jl_codectx_t &ctx, const jl_cgval_t &arg1, const jl_cgva
     if (jl_type_intersection(rt1, rt2) == (jl_value_t*)jl_bottom_type) // types are disjoint (exhaustive test)
         return ConstantInt::get(T_int1, 0);
 
+    if ((arg1.V || arg1.constant) && (arg2.V || arg2.constant) &&
+        (rt1 == (jl_value_t*)jl_bool_type || rt2 == (jl_value_t*)jl_bool_type ||
+         rt1 == (jl_value_t*)jl_int8_type || rt2 == (jl_value_t*)jl_int8_type ||
+         rt1 == (jl_value_t*)jl_uint8_type || rt2 == (jl_value_t*)jl_uint8_type)) {
+        // Both sides are boxed or can be trivially boxed
+        // also at least one side has known stable boxed address,
+        // we can simply compare the pointer.
```
vtjnash, line 2538: Only jl_true and jl_false have stable addresses guaranteed. The others are allowed to get copies (for example, the system image and runtime intrinsics don't care if they make copies, and they'd also be even allowed to appear on the stack, for example)

src/codegen.cpp
```diff
@@ -2559,18 +2580,8 @@ static bool emit_builtin_call(jl_codectx_t &ctx, jl_cgval_t *ret, jl_value_t *f,
...
-        if (v1.constant) {
-            if (v2.constant) {
-                *ret = mark_julia_type(ctx, ConstantInt::get(T_int8, jl_egal(v1.constant, v2.constant)), false, jl_bool_type);
-                return true;
-            }
-        }
         // emit comparison test
-        Value *ans = emit_f_is(ctx, v1, v2);
+        Value *ans = emit_f_is(ctx, argv[1], argv[2]);
         *ret = mark_julia_type(ctx, ctx.builder.CreateZExt(ans, T_int8), false, jl_bool_type);
```
vtjnash, line 2585:
```suggestion
        *ret = mark_julia_type(ctx, ans, false, jl_bool_type);
```
It'd be nice to go through and remove these extra conversions someday, assuming all the various accessors know how to convert correctly now.

vtjnash (approved):

src/codegen.cpp
yuyichao: Stack should not be an issue, since this uses `boxed` that should always return the right thing. I'm pretty sure no one else is allocating them so the only issue is the system image storing extra copies of them? Seems like it might be enough to add the `Int8` and `UInt8` boxes to the `_tags` list and fix the initialization?

src/cgutils.cpp
yuyichao: So I should/can update and use that instead? Seems that this doesn't handle `Type{Union{}}` correctly...

src/cgutils.cpp
```diff
@@ -1106,6 +1106,10 @@ static std::pair<Value*, bool> emit_isa(jl_codectx_t &ctx, const jl_cgval_t &x,
         return std::make_pair(ConstantInt::get(T_int1, *known_isa), true);
     }
 
+    if (is_uniquerep_Type(intersected_type)) {
```
vtjnash, line 1109: This is all wrong now. This says the types are egal, while you used to confirm pointer equivalence.

yuyichao: Hmm, ok, so `jl_pointer_egal` used to return true for `Type{Core.TypeofBottom}` whereas this returns false so I thought this is more accurate. I guess that's the wrong way around.... = = .... I guess I'll need to revert to the previous version...

OTOH, what should be the value returned for `Type{Core.TypeofBottom}`? It should match both `Core.TypeofBottom` and `Type{Union{}}` though whenever things hit the compiler it almost always get normalized to `Core.TypeofBottom`

vtjnash: Likely neither should return true, since neither is true

yuyichao:
Should be fixed now.
I left out `UInt8` and `Int8` since they are causing the test failures and the case that I want to fix are mainly the `=== true` and `=== false` in inference. OTOH, clearly the code path does get triggerred for `Int8` and `UInt8` as well so it might worth come back to.
`jl_pointer_egal` should not return `true` to a strict subset of `is_uniquerep_Type` when called on a `Type`.
I didn't touch the the zext, partially because I forgot and partially because it can potentially cause bisect problem....

PR #37365, opened 2020-09-03
inference: fix precision regression

vtjnash:
While in 1d08d70fb07cfcfc9e89ae1a2521411b1e1724e5 we wanted to avoid combining two types, we did not intend to prevent a type from being just a pass-through value.

### DIFF
--- a/base/compiler/typelimits.jl
+++ b/base/compiler/typelimits.jl
@@ -286,6 +286,8 @@ end
 # but without losing too much precision in common cases
 # and also trying to be mostly associative and commutative
 function tmerge(@nospecialize(typea), @nospecialize(typeb))
+    typea === Union{} && return typeb
+    typeb === Union{} && return typea
     suba = typea ⊑ typeb
     suba && issimpleenoughtype(typeb) && return typeb
     subb = typeb ⊑ typea
--- a/test/compiler/inference.jl
+++ b/test/compiler/inference.jl
@@ -99,6 +99,14 @@ tmerge_test(Tuple{}, Tuple{Complex, Vararg{Union{ComplexF32, ComplexF64}}},
 @test Core.Compiler.tmerge(Union{Nothing, Tuple{ComplexF32}}, Union{Nothing, Tuple{ComplexF32, ComplexF32}}) ==
     Union{Nothing, Tuple{Vararg{ComplexF32}}}
 @test Core.Compiler.tmerge(Vector{Int}, Core.Compiler.tmerge(Vector{String}, Vector{Bool})) == Vector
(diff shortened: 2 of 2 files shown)


Keno (approved):

martinholters (approved):

KristofferC:
Backporting this doesn't pass the added test https://build.julialang.org/#/builders/71/builds/3148/steps/5/logs/stdio. Should we just not backport it then?

martinholters:
Should rather try to figure out how to make them pass, as they used to on 1.4, so this fixes a real regression.

KristofferC:
Help appreciated :P

martinholters:
On 1.5, we're getting a `NOT_FOUND` there. So using
```julia
    (typea === Union{} || typea === NOT_FOUND) && return typeb
    (typeb === Union{} || typeb === NOT_FOUND) && return typea
```
makes the test pass. But I'm not sure that is valid. I'm also not sure the `NOT_FOUND` is even supposed to propagate up to here. I have no idea what it is supposed to signify here and where it comes from.

martinholters:
Prior to 1d08d70 we had `tmerge(NOT_FOUND, T) == tmerge(T, NOT_FOUND) == T` for any `T` (including `Union{}`). So that would mean the corresponding change to backport would be
```julia
typea === NOT_FOUND && return typeb
typeb === NOT_FOUND && return typea
typea === Union{} && return typeb
typeb === Union{} && return typea
```
On master, we seem to no longer see a `NOT_FOUND` in `tmerge`. So alternatively, we could consider backporting whichever change is responsible for that.

martinholters:
Aha, alternatively I guess we could backport this part of dcc0696971a100dbc63f90b8789632631dd1bfef:
```diff
--- a/base/compiler/inferencestate.jl
+++ b/base/compiler/inferencestate.jl
@@ -210,7 +210,10 @@ update_valid_age!(edge::InferenceState, sv::InferenceState) = update_valid_age!(
 function record_ssa_assign(ssa_id::Int, @nospecialize(new), frame::InferenceState)
     old = frame.src.ssavaluetypes[ssa_id]
     if old === NOT_FOUND || !(new ⊑ old)
-        frame.src.ssavaluetypes[ssa_id] = tmerge(old, new)
+        # typically, we expect that old ⊑ new (that output information only
+        # gets less precise with worse input information), but to actually
+        # guarantee convergence we need to use tmerge here to ensure that is true
+        frame.src.ssavaluetypes[ssa_id] = old === NOT_FOUND ? new : tmerge(old, new)
         W = frame.ip
         s = frame.stmt_types
         for r in frame.ssavalue_uses[ssa_id]
```

KristofferC:
Okay, I backported that part. Let's see how the tests do after that.

PR #37366, opened 2020-09-03
clangsa: Check that a value is rooted on return

### LINKED ISSUE #37364: Return value for non-static functions not treated as use by static analyser
The following C code added in any clangsa scanned source file should trigger an error but doesn't. The issue seems to be that the return value isn't treated as a use. Replacing the return by `jl_typeof` triggers an error correctly.

Since the analyser is smart enough about control flow, not error about unused return value from static functions should be fine (though still worth a warning unless suppressed with some marking). However, all return values from non-static function, especially exported function should certainly be treated as used...

```julia
JL_DLLEXPORT jl_value_t *jl_totally_used_function(int i)
{
    jl_value_t *v = jl_box_int32(i);
    jl_safepoint();
    return v; // jl_typeof(v);
}
```

Keno:
Fixes #37364
Also reverts part of 4ed484fe7ac591da67bfe384c369430ed658606d that
broke the clangsa self tests.

### DIFF
--- a/src/clangsa/GCChecker.cpp
+++ b/src/clangsa/GCChecker.cpp
@@ -705,9 +705,6 @@ void GCChecker::checkEndFunction(const clang::ReturnStmt *RS,
     C.addTransition(State);
   if (!C.inTopFrame())
     return;
-  // If `RS` is NULL, the function did not return normally.
-  // This could be either an abort/assertion failure or an exception throw.
-  // Do not require the GC frame to match in such case.
   if (C.getState()->get<GCDepth>() > 0)
     report_error(C, "Non-popped GC frame present at end of function");
 }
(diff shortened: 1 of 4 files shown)


Keno:
> Also reverts part of [4ed484f](https://github.com/JuliaLang/julia/commit/4ed484fe7ac591da67bfe384c369430ed658606d) that
> broke the clangsa self tests.

@staticfloat can we run the clangsa self tests just before anlyzegc on the bots. They're quite quick to run, so shouldn't be an issue to run them everytime.

src/clangsa/GCChecker.cpp
```diff
@@ -698,7 +708,7 @@ void GCChecker::checkEndFunction(const clang::ReturnStmt *RS,
   // If `RS` is NULL, the function did not return normally.
   // This could be either an abort/assertion failure or an exception throw.
   // Do not require the GC frame to match in such case.
-  if (RS && C.getState()->get<GCDepth>() > 0)
```
yuyichao, line 701: Also remove the comment or update it to clarify what `NULL` `RS` actually mean? AFAICT it does corresponds to ending the function from non-return but I guess somehow the core analysis deal with it so that the control flow is handled correctly?

vtjnash (approved):

src/clangsa/GCChecker.cpp
Keno: Null `RS` can also mean that there was no return statement in this function, which is a fine thing to do in C.

PR #37367, opened 2020-09-03
more powerful reverse(A; dims)

### LINKED ISSUE #37358: support multiple dimensions in reverse(A, dims=...)
Would be nice if you could pass a tuple for the `dims` keyword:
```
julia> A = rand(2,3)
2×3 Array{Float64,2}:
 0.909488  0.0318373  0.218921
 0.703028  0.361451   0.814981

julia> reverse(A, dims=(1,2)) # would be nice if this worked like reverse(reverse(A, dims=1), dims=2)
ERROR: TypeError: in keyword argument dims, expected Integer, got a value of type Tuple{Int64,Int64}
Stacktrace:
```
You could also support `dims=:` to reverse all of the dimensions, and maybe make this the default.   (Currently it is an error if you don't specify `dims`)

stevengj:
Fixes #37358:

* Supports reversing multiple dimensions simultaneously in `reverse(A; dims)` by passing a tuple for `dims`
* Makes `dims=:` (reverse all dimensions) the default for multidimensional `A`.  (Previously it was an error not to specify `dims`) — similar to [`numpy.flip`](https://numpy.org/doc/stable/reference/generated/numpy.flip.html).
* Adds multidimensional `reverse!(A; dims)`
* Consolidates the `Array` and `AbstractArray` implementations
* Seems to be mostly faster than the old `reverse(A; dims)` for flipping a single dimension.

Less code, does more, faster, backwards compatible — what's not to like?

To do:
- [x] Fix `OffsetArrays` test failures.
- [x] More tests
- [x] News

### DIFF
--- a/base/arraymath.jl
+++ b/base/arraymath.jl
@@ -103,25 +103,20 @@ _reverse!(A::AbstractArray{<:Any,N}, ::Colon) where {N} = _reverse!(A, ntuple(id
 _reverse!(A, dim::Integer) = _reverse!(A, (Int(dim),))
 _reverse!(A, dims::NTuple{M,Integer}) where {M} = _reverse!(A, Int.(dims))
 function _reverse!(A::AbstractArray{<:Any,N}, dims::NTuple{M,Int}) where {N,M}
-    if N < M || !allunique(dims) || !all(d -> 1 ≤ d ≤ N, dims)
+    dimrev = ntuple(k -> k in dims, Val{N}()) # boolean tuple indicating reversed dims
+
+    if N < M || M != sum(dimrev)
         throw(ArgumentError("invalid dimensions $dims in reverse!"))
     end
     M == 0 && return A # nothing to reverse
 
-    idx = Vector{Int}(undef, N) # temporary array for index calculations
+    # swapping loop only needs to traverse ≈half of the array
+    halfsz = ntuple(k -> k == dims[1] ? size(A,k) >> 1 : size(A,k), Val{N}())
 
-    # compute sizes for half of reversed array
-    ntuple(k -> @inbounds(idx[k] = size(A,k)), Val{N}())
-    @inbounds idx[dims[1]] = idx[dims[1]] >> 1
-
-    last1 = ntuple(k -> lastindex(A,k)+firstindex(A,k), Val{N}())
-    for i in CartesianIndices(ntuple(k -> firstindex(A,k):firstindex(A,k)-1+@inbounds(idx[k]), Val{N}()))
-        ntuple(k -> @inbounds(idx[k] = i[k]), Val{N}())
-        ntuple(Val{M}()) do k
-            @inbounds j = dims[k]
-            @inbounds idx[j] = last1[j] - idx[j]
-        end
-        iᵣ = CartesianIndex(ntuple(k -> @inbounds(idx[k]), Val{N}()))
+    last1 = ntuple(k -> lastindex(A,k)+firstindex(A,k), Val{N}()) # offset for reversed index
+    @inbounds for i in CartesianIndices(ntuple(k -> firstindex(A,k):firstindex(A,k)-1+@inbounds(halfsz[k]), Val{N}()))
+        iₜ = Tuple(i)
+        iᵣ = CartesianIndex(ifelse.(dimrev, last1 .- iₜ, iₜ))
         @inbounds A[iᵣ], A[i] = A[i], A[iᵣ]
     end
     if M > 1 && isodd(size(A, dims[1]))
--- a/base/arraymath.jl
+++ b/base/arraymath.jl
@@ -122,7 +122,7 @@ function _reverse!(A::AbstractArray{<:Any,N}, dims::NTuple{M,Int}) where {N,M}
         # middle slice for odd dimensions must be recursively flipped
         mid = firstindex(A, dims[1]) + (size(A, dims[1]) >> 1)
         midslice = CartesianIndices(ntuple(k -> k == dims[1] ? (mid:mid) : axes(A, k), Val{N}()))
-        _reverse!(@view(A[midslice]), dims[2:end])
+        _reverse!(view(A, midslice), dims[2:end])
     end
     return A
 end
--- a/base/arraymath.jl
+++ b/base/arraymath.jl
@@ -79,7 +79,7 @@ julia> reverse(b, dims=2)
  4  3
 
 julia> reverse(b)
-2×2 Array{Int64,2}:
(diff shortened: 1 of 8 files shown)


base/arraymath.jl
```diff
@@ -58,69 +58,77 @@ end
...
-                        B[boffs + k] = A[offs + k]
-                    end
-                end
-            end
+    M == 0 && return A # nothing to reverse
+
+    idx = Vector{Int}(undef, N) # temporary array for index calculations
+
+    # compute sizes for half of reversed array
+    ntuple(k -> @inbounds(idx[k] = size(A,k)), Val{N}())
```
martinholters, line 108: Using `ntuple` here is to get an unrolled loop, right? That's a cute trick; I wonder whether we should make that available under a different name to make the intention more obvious. Maybe `foreach(f, ::Val)`? But that's worth its own PR, probably.

stevengj: Yes.   I thought about using `@nexpr`, but that seems to require a generated function.

martinholters: Yes, that's how `ntuple` is doing it.

stevengj: The point is that I didn't not want to make `_reverse!` a generated function.  I'm perfectly happy to call other generated functions.

martinholters: Sure, got that, and makes perfect sense to me and conceptually, I'm totally fine with it. It's just that it gave me a small pause when reading it, as calling `ntuple` for side effects only is a bit usual. As said, a dedicated function could improve on that point, but is probably beyond the scope of this PR.

But to put the idea down somewhere:
```julia
@inline function nexprs(f::F, ::Val{N}) where {F,N}
    N::Int
    if @generated
        quote
            @nexprs $N i -> f(i)
        end
    else
        for i = 1:N
            f(i)
        end
    end
    return nothing
end
```
As the function version of `@nexprs`.

stevengj:
(There are certainly ways to speed this up further.  Working out of place, for example, just doing `r(X::AbstractMatrix) = X[reverse(axes(X,1)), reverse(axes(X,2))]` is a few times faster than `reverse(X)`.  But I'm not sure how much effort it is worth to performance-optimize `reverse`.)

base/arraymath.jl
```diff
@@ -58,69 +58,83 @@ end
...
+    @inbounds idx[dims[1]] = idx[dims[1]] >> 1
+
+    last1 = ntuple(k -> lastindex(A,k)+firstindex(A,k), Val{N}())
+    for i in CartesianIndices(ntuple(k -> firstindex(A,k):firstindex(A,k)-1+@inbounds(idx[k]), Val{N}()))
+        ntuple(k -> @inbounds(idx[k] = i[k]), Val{N}())
+        ntuple(Val{M}()) do k
+            @inbounds j = dims[k]
+            @inbounds idx[j] = last1[j] - idx[j]
         end
+        iᵣ = CartesianIndex(ntuple(k -> @inbounds(idx[k]), Val{N}()))
```
martinholters, line 124: Would it be worthwhile to set up an `NTuple{N,Bool}` or `Vector{Bool}` of length `N`, say `dimrev`, that encodes for every dimension of `A` whether it should be reversed? Then this would become
```julia
CartesianIndex(ntuple(k -> @inbounds(dimrev[k] ? last1[k] - i[k] : i[k]), Val{N}()))
```
and the two `ntuple` constructions above could be removed (at the expense of a similar construct before the loop to set up `dimrev`, of course).

I think to me, that would make this a bit more readable, but that may just be my personal taste, so please feel free to leave it as is if you prefer.

stevengj: Yes, that's reasonable.  Might even be faster (maybe with `ifelse`).

stevengj: One could even do `ifelse.(dimrev, last1 .- i, i)`, since tuple broadcasts are inlined/unrolled.

stevengj: Yup, the `ifelse` approach is indeed faster as well as simpler, and it lets me cut down the number of allocations to 0 for `reverse!`

base/arraymath.jl
stevengj: the new version no longer uses `ntuple` purely for iteration

base/arraymath.jl
martinholters: Yes, much nicer this way.

base/arraymath.jl
```diff
@@ -58,69 +58,83 @@ end
...
-                    end
-                end
-            end
+    M == 0 && return A # nothing to reverse
+
+    idx = Vector{Int}(undef, N) # temporary array for index calculations
+
+    # compute sizes for half of reversed array
+    ntuple(k -> @inbounds(idx[k] = size(A,k)), Val{N}())
+    @inbounds idx[dims[1]] = idx[dims[1]] >> 1
```
musm: might as well just use `÷ 2` since the compiler is smart enough that they are equivalent   for clarity sake

stevengj: They aren't actually equivalent — `>> 1` and `÷ 2` generate different instructions in LLVM.  But they might have similar performance, and in any case this computation occurs outside the loop so it isn't performance-sensitive.

musm: You right these are not 100% equivalent instruction wise,  but the timings wise  they are effectively identical.

Last I checked most c compilers were smart enough to take care of such micro-optimizations making them unnecessary, but I wonder why LLVM doesn't do it in this case (if I recall correctly, on older Julia versions they compiled down to the same instructions, but I might be misremembering).

musm (approved):

martinholters (approved):

stevengj:
`analyzegc` test failure seems unrelated; should be good to merge.

stevengj:
PS. The inspiration for this PR was watching @3b1b doing awkward nested `reverse` calls in [his 18.S191 lecture 1](https://youtu.be/DGojI9xcCfg?t=492).

stevengj:
A quick benchmark: for `A = rand(1026,2102)`, doing `reverse(reverse(A, dims=1), dims=2)` with Julia 1.5 takes **228ms**, doing the same call with the new version takes **11ms**, doing `reverse(A)` (which flips both dimensions at once) takes **5.9ms**, and doing `reverse!(A)` takes **3ms**.

`numpy.flip` actually returns a view so it is not comparale, but doing `numpy.flip(A).copy()` with the same matrix (which is big enough that Python overhead, which is < 1µs, is negligible) takes **15ms** on my machine.

stevengj:
A thought: should something like `@view reverse(A)` or `@views reverse(A)` change it to return a view?

PR #37369, opened 2020-09-03
Add Mmap.madvise!

### LINKED ISSUE #37353: Adding madvise to the Mmap stdlib
I've got a use case where I _think_ being able to `madvise` a memory-mapped array with the `MADV_WILLNEED` flag will help me with a computation/program I'm working on.

I've prototyped an implementation of a new `Mmap.madvise!` function, but before making a PR for it, I'd like to ask for feedback on whether it is desired. Right now, the `Mmap.mmap` and `Mmap.sync!` functions work on *nix & Windows, but as far as I can tell, Windows doesn't have any good equivalent of `madvise`. Therefore, my prototype `Mmap.madvise!` is only exposed on Unix system.

Is that an OK direction for a function in the standard library to take, or should this function live elsewhere?

jmert:
This PR adds `Mmap.madvise!` to give access to the `madvise` system calls on unix OSs.

I've done my best to fill in the list of OS-specific constants, but I only have a Linux machine locally from which I could extract the values from appropriate header files — for everything else, I relied on finding headers through Google searches.

Resolves #37353.

### DIFF
--- a/stdlib/Mmap/src/Mmap.jl
+++ b/stdlib/Mmap/src/Mmap.jl
@@ -349,4 +349,64 @@ function sync!(m::Array{T}, flags::Integer=MS_SYNC) where T
 end
 sync!(B::BitArray, flags::Integer=MS_SYNC) = sync!(B.chunks, flags)
 
+@static if Sys.isunix()
+const MADV_NORMAL = 0
+const MADV_RANDOM = 1
+const MADV_SEQUENTIAL = 2
+const MADV_WILLNEED = 3
+const MADV_DONTNEED = 4
+if Sys.islinux()
+    const MADV_FREE = 8
+    const MADV_REMOVE = 9
+    const MADV_DONTFORK = 10
+    const MADV_DOFORK = 11
+    const MADV_MERGEABLE = 12
+    const MADV_UNMERGEABLE = 13
+    const MADV_HUGEPAGE = 14
+    const MADV_NOHUGEPAGE = 15
+    const MADV_DONTDUMP = 16
+    const MADV_DODUMP = 17
+    const MADV_WIPEONFORK = 18
+    const MADV_KEEPONFORK = 19
+    const MADV_COLD = 20
+    const MADV_PAGEOUT = 21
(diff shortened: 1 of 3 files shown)


stdlib/Mmap/src/Mmap.jl
```diff
@@ -349,4 +349,64 @@ function sync!(m::Array{T}, flags::Integer=MS_SYNC) where T
...
+"""
+    Mmap.madvise!(array, flag::Integer = Mmap.MADV_NORMAL)
+
+Advises the kernel on the intended usage of the memory-mapped `array`, with the intent
+`flag` being one of the available `MADV_*` constants.
+"""
+function madvise!(m::Array{T}, flag::Integer=MADV_NORMAL) where T
+    offset = rem(UInt(pointer(m)), PAGESIZE)
+    ptr = pointer(m) - offset
+    mmaplen = length(m) * sizeof(T) + offset
```
JeffBezanson, line 403: This should use `sizeof(m)` instead since there might be alignment padding. (Then the `T` static parameter is also unnecessary.)

jmert: I assume you mean `length(m) * sizeof(T)` → `sizeof(m)` and that the `+ offset` part is kept? If so, I can make the same adjustment in the `sync!` function which uses the same calculation (which admittedly I just wrote into it in #37311).

jmert:
I've been testing this out, and while there aren't any issues, it also doesn't do what I actually want. It appears the `MADV_WILLNEED` flag is not akin to a late-advisory equivalent of `MAP_POPULATE` during the initial `mmap` call.[1] So while I'm still willing to do any revisions reviewers may want on this PR, it's not personally urgent to my work. (Instead it looks like I need to look into a nice way to get more flags through to `mmap` itself.)

[1] There was proposal from 2013 to the [Linux kernel mailing list](https://lkml.org/lkml/2013/6/27/698) to add a `MADV_POPULATE` which would have done what I wanted, but that appears to have long been abandoned so is no help to me.

PR #37374, opened 2020-09-03
Show recursive dep precomp info during precomp

IanButterworth:
An idea for making the precompillation process a bit less opaque

Demo with loading Plots
https://www.dropbox.com/s/z9n5bwle6wd125o/Julia%20dep%20precomp%20printing.mov?dl=0

Todo: 
- [x] Figure out how disable printing when the parent precompillation wasn't started in interactive mode
- [x] Maybe there's a better glyph than `>` to use

### DIFF
--- a/base/loading.jl
+++ b/base/loading.jl
@@ -1146,10 +1146,10 @@ function create_expr_cache(input::String, output::String, concrete_deps::typeof(
                        --eval 'eval(Meta.parse(read(stdin,String)))'`, stderr=stderr),
               "w", stdout)
     # write data over stdin to avoid the (unlikely) case of exceeding max command line size
-    # pass precompiling_interactively state recursively
+    # pass PRECOMPILING_INTERACTIVELY state recursively
     write(io.in, """
         begin
-            Base.precompiling_interactively!($(is_precompiling_interactively()))
+            Base.PRECOMPILING_INTERACTIVELY[] = $(PRECOMPILING_INTERACTIVELY[])
             Base.include_package_for_output($(repr(abspath(input))), $(repr(depot_path)), $(repr(dl_load_path)),
             $(repr(load_path)), $deps, $(repr(uuid_tuple)), $(repr(source_path(nothing))))
         end
@@ -1214,10 +1214,10 @@ function compilecache(pkg::PkgId, path::String)
     end
     # run the expression and cache the result
     verbosity = isinteractive() ? CoreLogging.Info : CoreLogging.Debug
-    isinteractive() && precompiling_interactively!(true)
+    isinteractive() && (PRECOMPILING_INTERACTIVELY[] = true)
     @logmsg verbosity "Precompiling $pkg"
-    should_log_deps = is_precompiling_interactively() && !isinteractive() && (CoreLogging.current_logstate().min_enabled_level > CoreLogging.Debug) #don't print for main package or when in debug
     precomp_msg = " > $(pkg.name)"
+    should_log_deps = PRECOMPILING_INTERACTIVELY[] && !isinteractive() && (CoreLogging.current_logstate().min_enabled_level > CoreLogging.Debug) #don't print for main package or when in debug
     should_log_deps && print(precomp_msg)
 
     # create a temporary file in `cachepath` directory, write the cache in it,
@@ -1251,8 +1251,6 @@ function compilecache(pkg::PkgId, path::String)
 end
 
 const PRECOMPILING_INTERACTIVELY = Ref{Bool}(false)
-precompiling_interactively!(b::Bool) = PRECOMPILING_INTERACTIVELY[] = b
-is_precompiling_interactively()::Bool = PRECOMPILING_INTERACTIVELY[]::Bool
 
 module_build_id(m::Module) = ccall(:jl_module_build_id, UInt64, (Any,), m)
 
--- a/base/loading.jl
+++ b/base/loading.jl
@@ -1146,10 +1146,10 @@ function create_expr_cache(input::String, output::String, concrete_deps::typeof(
                        --eval 'eval(Meta.parse(read(stdin,String)))'`, stderr=stderr),
               "w", stdout)
     # write data over stdin to avoid the (unlikely) case of exceeding max command line size
-    # pass isprecompilinginteractively state recursively
+    # pass precompiling_interactively state recursively
     write(io.in, """
         begin
-            Base.isprecompilinginteractively!($(isprecompilinginteractively()))
+            Base.precompiling_interactively!($(is_precompiling_interactively()))
             Base.include_package_for_output($(repr(abspath(input))), $(repr(depot_path)), $(repr(dl_load_path)),
             $(repr(load_path)), $deps, $(repr(uuid_tuple)), $(repr(source_path(nothing))))
         end
@@ -1214,10 +1214,10 @@ function compilecache(pkg::PkgId, path::String)
     end
     # run the expression and cache the result
     verbosity = isinteractive() ? CoreLogging.Info : CoreLogging.Debug
-    isinteractive() && isprecompilinginteractively!(true)
+    isinteractive() && precompiling_interactively!(true)
     @logmsg verbosity "Precompiling $pkg"
     precomp_msg = " ► $(pkg.name)"
-    should_log_deps = isprecompilinginteractively() && !isinteractive() && (CoreLogging.current_logstate().min_enabled_level > CoreLogging.Debug) #don't print for main package or when in debug
+    should_log_deps = is_precompiling_interactively() && !isinteractive() && (CoreLogging.current_logstate().min_enabled_level > CoreLogging.Debug) #don't print for main package or when in debug
     should_log_deps && print(precomp_msg)
 
     # create a temporary file in `cachepath` directory, write the cache in it,
@@ -1250,9 +1250,9 @@ function compilecache(pkg::PkgId, path::String)
     end
 end
 
-const IS_PRECOMPILING_INTERACTIVELY = Ref{Bool}(false)
-isprecompilinginteractively!(b::Bool) = IS_PRECOMPILING_INTERACTIVELY[] = b
-isprecompilinginteractively()::Bool = IS_PRECOMPILING_INTERACTIVELY[]::Bool
+const PRECOMPILING_INTERACTIVELY = Ref{Bool}(false)
+precompiling_interactively!(b::Bool) = PRECOMPILING_INTERACTIVELY[] = b
+is_precompiling_interactively()::Bool = PRECOMPILING_INTERACTIVELY[]::Bool
 
 module_build_id(m::Module) = ccall(:jl_module_build_id, UInt64, (Any,), m)
 
--- a/base/loading.jl
+++ b/base/loading.jl
@@ -1241,7 +1241,7 @@ function compilecache(pkg::PkgId, path::String)
         end
     finally
         rm(tmppath, force=true)
-        should_log_deps && print("\e[$(length(precomp_msg))D\e[0K") #clear last dep print
+        should_log_deps && print("\e[$(length(precomp_msg))D\e[0J") #clear last dep print
     end
     if p.exitcode == 125
         return PrecompilableError()
--- a/base/loading.jl
+++ b/base/loading.jl
@@ -1211,6 +1211,9 @@ function compilecache(pkg::PkgId, path::String)
     # run the expression and cache the result
     verbosity = isinteractive() ? CoreLogging.Info : CoreLogging.Debug
     @logmsg verbosity "Precompiling $pkg"
+    precomp_msg = " > $(pkg.name)"
+    should_log_deps = (Base.CoreLogging.current_logstate().min_enabled_level >= verbosity) && !isinteractive()
+    should_log_deps && print(precomp_msg)
 
     # create a temporary file in `cachepath` directory, write the cache in it,
     # write the checksum, _and then_ atomically move the file to `cachefile`.
@@ -1233,6 +1236,11 @@ function compilecache(pkg::PkgId, path::String)
         end
     finally
         rm(tmppath, force=true)
+        if should_log_deps
+            print("\e[$(length(precomp_msg))D") 
+            print(repeat(" ", length(precomp_msg))) #overwrite dep print
+            print("\e[$(length(precomp_msg))D") 
+        end
     end
     if p.exitcode == 125
         return PrecompilableError()
--- a/base/loading.jl
+++ b/base/loading.jl
@@ -1146,9 +1146,13 @@ function create_expr_cache(input::String, output::String, concrete_deps::typeof(
                        --eval 'eval(Meta.parse(read(stdin,String)))'`, stderr=stderr),
               "w", stdout)
     # write data over stdin to avoid the (unlikely) case of exceeding max command line size
+    # pass isprecompilinginteractively state recursively
     write(io.in, """
-        Base.include_package_for_output($(repr(abspath(input))), $(repr(depot_path)), $(repr(dl_load_path)),
+        begin
+            Base.isprecompilinginteractively!($(isprecompilinginteractively()))
+            Base.include_package_for_output($(repr(abspath(input))), $(repr(depot_path)), $(repr(dl_load_path)),
             $(repr(load_path)), $deps, $(repr(uuid_tuple)), $(repr(source_path(nothing))))
+        end
         """)
     close(io.in)
     return io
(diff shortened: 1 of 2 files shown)


ianshmean:
Also, there was the idea to add a progress indicator too, but I'm not sure that's possible accurately, because the loading happens recursively triggered by using/import, and I don't think there's a way to know for certain how many packages will be loaded in advance of fully loading the code.

Also, it's not mandatory that a dep listed in the Project.toml is actually loaded, so that can't really be used as an accurate total.

But perhaps it is true that `number of loaded packages` <= `number of deps in Project.toml`.. so if finishing the progress indicator early is ok, that could suffice.

However, given the recursive nature of precomp, counting the number of 1st-level deps loaded seems non-trivial..

Octogonapus:
Nice; my vote is for `>` instead of `►`, though, because it is consistent with other build tools I use (like in the JVM world).

ianshmean:
Any other opinions on `>` vs. `►` ?

mkitti:
I would go with `>` since this is Base. Stick with ASCII characters since it may not render on some terminals properly.

ianshmean:
Yeah, I think I agree. I'll revert to `>` but I'll hold off in case there are any other changes. Seems wasteful to run CI just for that alone

base/loading.jl
```diff
@@ -1210,7 +1214,11 @@ function compilecache(pkg::PkgId, path::String)
     end
     # run the expression and cache the result
     verbosity = isinteractive() ? CoreLogging.Info : CoreLogging.Debug
+    isinteractive() && precompiling_interactively!(true)
     @logmsg verbosity "Precompiling $pkg"
+    precomp_msg = " ► $(pkg.name)"
+    should_log_deps = is_precompiling_interactively() && !isinteractive() && (CoreLogging.current_logstate().min_enabled_level > CoreLogging.Debug) #don't print for main package or when in debug
```
c42f, line 1220: If you want to use the logging system rather than `print` that's great.  But an ad-hoc mixing some use of `CoreLogging` internals with the use of `print` is not the way to go!

What does this use of `min_enabled_level` achieve in practice?  Precompilation always happens in a separate process, so I don't think you can easily affect `min_enabled_level` in that process.

c42f (commented):
I really like the general idea of this, but we should really either use `print` or the logging system (don't mix both!)

Actually this would be an ideal use case for progress logging, because it would allow UIs which are not text terminals to benefit from this information in a user friendly way.  For example, progress reporting in Pluto, VSCode, (and even JuliaHub when the compute functionality is released).

Problem is, we need to import some parts of `ProgressLogging.jl` and `TerminalLoggers.jl` into `Logging` so that progress logging works smoothly by default in the REPL. So there's a fair amount of infrastructural background work on that side to do this cleanly.

I started writing an issue about having proper progress logging in the stdlib. I guess I should finish that off and we can assess which way to go about this.

base/loading.jl
ianshmean: I assumed that under the current behavior, the precomp subprocesses would adopt the main `JULIA_DEBUG` state, so if the user requests debug logging, the deps would print debug precomp messages.. but I didn't check that. 

I intended to preserve that behavior without messing up the messages with overprinting.

ianshmean:
Yeah, conforming to the logging system tooling sounds like the right way forward.

I can't figure out a way to get a sense of _progress_ though. Given the dep and sub-dep package loading and precomp happens during the code loading, it didn't seem possible to identify all the deps in advance.

Perhaps a manifest could be used, but not all those packages will be loaded for certain.. but perhaps a _progress_ indicator could terminate early if only a subset of packages in the manifest actually get loaded.

c42f:
> I can't figure out a way to get a sense of progress though.

Yes it's inherently difficult: we can't *really* know which packages will be used without running the top level code of the package (though a quick pass over the AST without evaluating it would give us a very good guess for a lot of packages). Worse, we don't know how much time each of the dependencies will take to precompile.

But when I talk about "progress logging" I'm thinking quite generally; I'd like progress reporting to support things like iterative solvers, which have a similar problem of open endedness.

So in that sense, I think the tree you're already displaying is a "sense of progress" and should be communicated via the logging system. Abstractly, as a "progress message".

ianshmean:
In terms of porting the current idea to the Logging tooling, I can't immediately see a way to do the kind of overprinting in this PR via `Logging` methods, so it seems that even a basic move over to `Logging` will need to wait for the imports you mentioned. Is that a fair assessment?

mkitti:
How does this work together with IJulia or Pluto.jl?

ianshmean:
Pluto seems to not execute in interactive mode, so no printing.

IJulia is interactive, but doesn't seem to support the escape code-based overwriting approach, so they just collect

image

c42f:
> I can't immediately see a way to do the kind of overprinting in this PR via Logging methods

Right. The logging backend needs to handle the overprinting UI concerns — you can't do it by plain text. There's a logging backend which does this properly - check out TerminalLoggers.jl (and its use with ProgressLogging.jl).

In the longer term, this is a good thing! Implementing this with overprinting assumes you're talking to something that behaves like a text terminal and not all user interfaces are terminals. But it's definitely more machinery to get working together.

> How does this work together with IJulia or Pluto.jl?

Not sure about IJulia; I think it all just goes into the stderr at the moment which is not ideal. And Pluto doesn't support capturing logs yet but I've been chatting to the author and I think it will. The point is not that these have first class integration right now, but the logging system provides a lot more structure to log records than text. Different UIs can take this structure and use it to pretty print GUI progress bars, or whatever.

ianshmean:
In the short-term, what if we took the approach in this PR and limited to just terminals? Specifically terminals that support escape codes. 

It would still be a mix of logging and printing, but it would retain the current logging behavior, as long as the terminal handled it properly. 

Is there a heuristic for identifying such terminals?

KristofferC:
For IJulia you have to (AFAIU) call `IJulia.clear_output(true)` between the updates.

base/loading.jl
```diff
@@ -1241,6 +1250,10 @@ function compilecache(pkg::PkgId, path::String)
     end
 end
 
+const PRECOMPILING_INTERACTIVELY = Ref{Bool}(false)
+precompiling_interactively!(b::Bool) = PRECOMPILING_INTERACTIVELY[] = b
```
KristofferC, line 1254: Imo just use the global variable. Makes it more clear from looking at the code what is happening.

base/loading.jl
```diff
@@ -1241,6 +1250,10 @@ function compilecache(pkg::PkgId, path::String)
...
+is_precompiling_interactively()::Bool = PRECOMPILING_INTERACTIVELY[]::Bool
```
KristofferC, line 1255: No need for type asserts here right since the variable is `Ref{Bool}`?

ianshmean:
> `IJulia.clear_output(true)`

I believe that's a bit too heavy-handed. We just need to remove the most recently printed characters, otherwise if the user prints something before loading a module, that would be wiped, along with the parent deps and `[ info: Precompiling...` message

halleysfifthinc:
> But when I talk about "progress logging" I'm thinking quite generally; I'd like progress reporting to support things like iterative solvers, which have a similar problem of open endedness.

From a UX perspective, this kind of progress reporting is better suited for "spinner" type progress notifications (in a terminal, something like [sindresorhus/ora](https://github.com/sindresorhus/ora)).

For precomp progress reporting, is maintaining the recursive/tree-like structure in the presentation of precomp progress ever expected to be useful for debugging? If not, a simpler form (that would also be more compatible with the `IJulia.clear_output(true)`) would be to use a spinner type animation and only output the package currently being precompiled (similar to the GIF in the README for sindresorhus/ora).

Additionally, consider the situation with a package with many/deep intermediate dependencies which all need precompiling (and/or including packages with long names), the current progress reporting would probably wrap in the terminal and result in a messy 1-line/2-line flickering; displaying a single package name at a time would effectively prevent that.

base/loading.jl
```diff
@@ -1146,9 +1146,13 @@ function create_expr_cache(input::String, output::String, concrete_deps::typeof(
...
               "w", stdout)
     # write data over stdin to avoid the (unlikely) case of exceeding max command line size
+    # pass PRECOMPILING_INTERACTIVELY state recursively
     write(io.in, """
-        Base.include_package_for_output($(repr(abspath(input))), $(repr(depot_path)), $(repr(dl_load_path)),
+        begin
+            Base.PRECOMPILING_INTERACTIVELY[] = $(PRECOMPILING_INTERACTIVELY[])
+            Base.include_package_for_output($(repr(abspath(input))), $(repr(depot_path)), $(repr(dl_load_path)),
             $(repr(load_path)), $deps, $(repr(uuid_tuple)), $(repr(source_path(nothing))))
+        end
```
tkf, line 1155: Can this be moved into `include_package_for_output`? It looks like that the motivation for introducing `include_package_for_output` was that it's a better style to put code in a function rather than in a function #36028. There was some precompilation speedup with the function approach.

KristofferC:
> For precomp progress reporting, is maintaining the recursive/tree-like structure in the presentation of precomp progress ever expected to be useful for debugging? If not, a simpler form (that would also be more compatible with the IJulia.clear_output(true)) would be to use a spinner type animation and only output the package currently being precompiled (similar to the GIF in the README for sindresorhus/ora). My recollection is that npm employs a similar strategy for some interfaces.

A spinner could be added, but the tree form is also much more fun to look at.

base/loading.jl
ianshmean: Makes sense. I've switched it over

ianshmean:
I agree that the tree form is fun, and I think it's necessary as well. A package hasn't finished being precompiled until all its packages have finished. The process is recursive not sequential

base/loading.jl
ianshmean: thanks btw

tkf:
I agree with what @c42f already said.  Just adding a few comments:

* I think coupling status printing too much with the precompilation core mechanism might complicate possible parallelism. But, since this is implemented with rather succinct code, maybe we don't need to worry about it for now.
* For "transient" message that is reprinted many times and disappears at some point, I think a dedicated protocol in logging framework makes sense. Something like https://github.com/JunoLab/ProgressLogging.jl/issues/23#issuecomment-595623683
* Regarding IJulia support, I think we can use [`update_display_data` messaging protocol](https://jupyter-client.readthedocs.io/en/stable/messaging.html#update-display-data) for cleanly updating a part of the output.  I'm not sure if IJulia supports it already.

Having said that, I don't think it's too crazy to do it just by writing things directly to `stderr` (maybe check if `isatty`) at least for now. It'd be great if we can decouple the rendering of the UI from actual message creation. But there are many things to do to support printing interfaces as rich as, e.g., Pkg.jl.

KristofferC:
If you have two packages with the same dependency, it would it only show when the dependency is precompiled for the first package? In that case, maybe showing the whole tree is a bit meaningless? But we would still want to show things "depth first" if possible. So if the dep graph is

```
A -> B -> C
 |> D - > C
```

We want to show it in the order `C, B, D, A` where the order of B and D doesn't matter.

ianshmean:
It seems like this PR has turned into a bigger discussion about precomp ordering, and I've got 1.6 being an LTS on my mind..

Is it instead worth trying to get parallelized precomp working sooner rather than later..?

Perhaps, as others have said above, doing: 
1) a recursive AST parse for used packages and their used packages..
2) spawning one package precomp per thread depth-first

tkf:
> 1. a recursive AST parse for used packages and their used packages.

Don't Manifest.toml have enough information to construct the dependency tree? (OK, I guess technically the user can deceive Manifest.toml by using `Base.require` instead of `using`...) Actually, I tried parallel precompilation a while ago with this idea <https://gist.github.com/f24da5a4d1593e7bcd484634699dc3e3>.  But it didn't go too far since I didn't know how to

> 2\. figure out which of those need precompiling

(or more like realized that I had to change `Base` to do it.)

c42f:
> a recursive AST parse for used packages and their used packages.

This won't entirely tell you what you need to know, because packages can have top level code which looks like

```julia
if some_condition()
    using SomePackage
end

if other_condition()
    include("some_file.jl")  # may also contain `using`
end
```

There's even wild stuff like
```julia
if VERSION < v"..."
    include("new_code.jl")
end
```
where `new_code.jl` can contain syntax which isn't even parsable on older Julia versions. (Perhaps not so much of a problem in 1.x. But was a thing in the past.)

So the list of `using`s isn't entirely available via static analysis. Normally you could get an accurate guess with some simple rules, but not in general.

Regarding the Manifest/Project — these are pkg concepts. I can see there's a bit of knowledge of the manifest available in loading.jl, but does this cover what's required here?

tkf:
I guess a pitfall of Manifest.toml-based approach is something like

```julia
if Sys.iswindows()
    using SomeWindowsOnlyPackage
...
```

Manifest.toml can only tell us the "upper bound" of the set of packages that have to be compiled.

ianshmean:
How about moving it to Pkg time? adding an opt-out precompile step to `Pkg.add`, `Pkg.update` etc. and leave the current tooling as-is to catch `dev`-ed packages or anything that gets through?

With the resolver information at-hand after an `add` or `update`, a depth-first parallelized precompillation step could be structured easily at that time for any added package. 

One could opt-out with something like `]precompillation disable` for people wanting to just test for resolver issues etc. 

Imagine julia precompiling all your newly added packages on all the threads you can throw at it..

tkf:
I think post-add precompilation would be a much more hairy task if, e.g., you want it to happen it in the background (so that you can keep using REPL; or maybe `Pkg.add` itself can happen in the background?). That's why I think it's a better idea to just expose generic post-add hook from Pkg.jl and develop such interface outside stdlib: https://github.com/JuliaLang/Pkg.jl/issues/973

But I'm not sure if that's related to parallel precompilation. Don't you need a parallel precompilation function in `Base` anyway, to invoke it via Pkg operations?

KristofferC:
> I think coupling status printing too much with the precompilation core mechanism might complicate possible parallelism. But, since this is implemented with rather succinct code, maybe we don't need to worry about it for now.

This seems like an unnecessary worry to me. If it stands between the pretty-printing and parallel compilation of packages at some later time we could just get rid of the pretty-printing.

tkf:
Yes, I agree it's not a problem as long as it's easy to disentangle the status printing part.

ianshmean:
Here's a candidate for parallel `Pkg.precompile` that works surprisingly well

https://github.com/JuliaLang/Pkg.jl/pull/2018

ianshmean:
I checked and this PR plays nicely with JuliaLang/Pkg.jl#2018. Rightly, none of the info displayed here prints during the new parallel `Pkg.precompile()` method because that parallel process isn't recursive, and already has one info print per dep.

@tkf:
> Having said that, I don't think it's too crazy to do it just by writing things directly to stderr (maybe check if isatty) at least for now

I could update this PR to `print(stderr, ...`, but I couldn't see a julia function for `isatty` to disable for IJulia etc. Does one exist?

tkf:
I'm a bit puzzled about how it would be compatible with JuliaLang/Pkg.jl#2018. Naively, `Pkg.precompile` with this patch would now run `print(precomp_msg)` and the corresponding escape sequence for deleting from multiple tasks. These prints can be arbitrarily interleaved.  Intuitively, we need a dedicated task or locking mechanism to correctly render the precompilation state.

> I couldn't see a julia function for `isatty` to disable for IJulia etc.

I guess you can use `ccall(:isatty, Cint, (Cint,), Base._fd(stdout))` in \*nix. But it looks like Base currently uses `tput` for this

https://github.com/JuliaLang/julia/blob/8bdf5695e25e77d53f8934d9f135858d174e30d1/base/ttyhascolor.jl#L4-L15

I guess you can do a similar trick but it'd mean that IJulia has to set `ENV["TERM"] = "dumb"` or something during its startup. But then, it'd be tricky to support IJulia in Windows.

---

Anyway, now that JuliaLang/Pkg.jl#2018 is merged and considering the possible trouble in Windows, I think structured logging direction as @c42f explained is much more appealing.

ianshmean:
The reason it's compatible is that in this PR `compilecache` only prints the new tree-based messages when `compilecache` was directly invoked from the interactive repl, not the recursive sub-calls during loading. Note how the main package name isn't printed in the tree, just the sub-dependencies. Given that in JuliaLang/Pkg.jl#2018 each dep is called directly from the interactive repl session, the normal `[ Info: Precompiling ..` message prints, but not any of the tree printing, including the overprinting codes given no recursive precomp happens

tkf:
Ah, thanks for the explanation. How `Pkg.precompile` and `PRECOMPILING_INTERACTIVELY` interact is trickier than I thought. So, `> DependencyPackageName` is not printed via `Pkg.precompile` because, as it invokes `compilecache` in a depth-first manner, `compilecache` is not called in the sub-processes? <strike>Is my understanding correct?</strike> (Now that I re-read your earlier comment, that's exactly what you said.)

ianshmean:
I actually flipped the start of that explanation accidentally. It's now corrected

evetion:
This has been very useful. :+1: 

Small nitpick, in my terminal (Terminal app on Fedora 32) wrap around to a newline isn't handled properly:
compile_cut

ianshmean:
To make wrap-around work properly with this implementation, there would need to be an ansi escape code that moves the cursor back a number of chars, not a number of cols. I.e. wrap around onto the previous line

I'm not aware of a code that does that, so I think this recursive overprinting approach can't handle this properly. 


I guess one fix could be to print each dep on a new line, but that would probably look messier

ianshmean:
Vertical tree printing actually doesn't look too bad: https://www.dropbox.com/s/44jtzujw2zr3828/julia%20Plots%20-%20vertical%20tree%20print.mov?dl=0

StefanKarpinski:
That does look quite nice. How about starting the printing right under the info message about precompiling?

halleysfifthinc:
How difficult would it be to indent the recursive dependencies? The caret and the lack of indents prevent the tree structure from being immediately obvious until several iterations of precomp dependencies have flashed. I think `└` instead of the caret would help make the tree structure more obvious even without indenting. A tree format for the printing could also be extended to represent parallel precompilation fairly clearly.

IanButterworth:
Closing as the new parallel precomp in 1.6 has reduced how helpful this would be

PR #37378, opened 2020-09-03
RFC: inference: remove union-split limit for linear signatures

vtjnash:
This size limit should be already be imposed elsewhere (tmerge), and
should not actually add cost to perform the union/tuple-switching when
there is no cartesian product to consider. This permits users to
explicitly demand larger union products (for example, with type-asserts
or field types) and still expect to get reliable union-splitting at any
size in single-dispatch sites.

### DIFF
--- a/base/compiler/abstractinterpretation.jl
+++ b/base/compiler/abstractinterpretation.jl
@@ -24,7 +24,7 @@ function abstract_call_gf_by_type(interp::AbstractInterpreter, @nospecialize(f),
     end
     valid_worlds = WorldRange()
     atype_params = unwrap_unionall(atype).parameters
-    splitunions = 1 < countunionsplit(atype_params) <= InferenceParams(interp).MAX_UNION_SPLITTING
+    splitunions = 1 < unionsplitcost(atype_params) <= InferenceParams(interp).MAX_UNION_SPLITTING
     mts = Core.MethodTable[]
     fullmatch = Bool[]
     if splitunions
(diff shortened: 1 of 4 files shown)


JeffBezanson (approved):

JeffBezanson:
Yeah, this definitely looks worth trying.

martinholters:
Absolutely worth trying :+1: 

IMHO, the implementation might be a bit too tricky, though. IIUC, `unionsplitcost`effectively computes `any(ti -> isa(ti, Union), atypes) ? 2*prod(sort(unionlen.(atypes))[1:end-1]) : 1` (but more efficient and saturating when overflowing). From the comment, I would have expected just `prod(sort(unionlen.(atypes))[1:end-1])`. The advantage of the more complicated expression is that the presence of a union can be seen immediately: Only if there are no unions, a 1 is returned. Anything above 1 means union-splitting can be applied. I understand how that is useful, but it's a bit non-obvious, not made any better by the comment "cost function over the effort to switch a tuple-union representation as a cartesian product, relative to the size of the original representation". If there is just a single union, then why does the cartesian product double the size? Maybe this just needs a slightly better comment. But in my experience, if something is tricky to explain, it may often be worthwhile to try to simplify the concept rather than to work on the explanation. So maybe it's time to take a step back and re-think what the cost measure should be. Not that I have a better idea, though; maybe this is fine as is.

vtjnash:
> If there is just a single union, then why does the cartesian product double the size

Expanding a single union such as `Tuple{typeof(+), Union{Int32, Int64}}` gets slightly bigger as `Union{Tuple{typeof(+), Int32}, Tuple{typeof(+), Int64}}`, before we consider the cartesian product of the other union arguments. I could have picked many other reasonable factors for this cost estimate—such as `sqrt(length(atypes))`—but that's just what this is: a cost measure estimate. It had no meaning except as a relation to the `MAX_UNION_SPLITTING` parameter, so I picked a number that behaved so that it behaved as I wanted it to with respect to that threshold.

martinholters:
Makes sense. Might be worth briefly mentioning that in the comment in case you happen to update the PR.

PR #37385, opened 2020-09-03
Make sure to build clangsa before running clangsa self tests

Keno:


vtjnash (approved):

Keno:
Should backport this, otherwise the analyzegc buildbot will start failing once it runs the clangsa tests.

staticfloat (approved):

PR #37386, opened 2020-09-03
fix a possible segfault during static compilation

JeffBezanson:
Note this only applies to 1.5. I'll make a separate patch for master.


KristofferC:
Maybe better to make the pr against the backport branch so it is together with everything else going into 1.5.2?

PR #37319, opened 2020-09-01
codegen: phic/upsilon nodes can be set to undefined

vtjnash:
Unlike other nodes, Upsilon nodes can be potentially undefined and need
special handling to ensure they aren't literally undef, as we might try
to union-copy or gc-root the data later.

Fix #37262

PR #37388, opened 2020-09-03
Fix use of CreateAlignedStore

yuyichao:
This is a compile error on 10+ and is probably triggering implicity conversion on LLVM 8 and 9.


yuyichao:
Introduced in https://github.com/JuliaLang/julia/pull/37319 so if that one get backported this one should be as well...

vtjnash (approved):

PR #37389, opened 2020-09-03
Bump `Pkg` version

staticfloat:
This is important as we need to fix `Pkg.BinaryPlatforms.platform_key_abi(::SubString)` for BB on Yggdrasil


fredrikekre:
shasums should be updated too.

> This is important as we need to fix Pkg.BinaryPlatforms.platform_key_abi(::SubString) for BB on Yggdrasil

Why not just patch Yggdrasil to use `Pkg.BinaryPlatforms.platform_key_abi(String(substr))` instead of waiting for new nightlies with this fix?

staticfloat:
> Why not just patch Yggdrasil to use Pkg.BinaryPlatforms.platform_key_abi(String(substr)) instead of waiting for new nightlies with this fix?

Because there's a lot of stuff deep in BB that doesn't do that.  All of BB is broken on master right now, I think.

PR #37391, opened 2020-09-04
Added canonicalize for Periods

cfitz25:
When i was using canonicalize i had crashes sometimes when doing Period addition as sometimes a Period would be returned instead of a CompoundPeriod, this is a fix for that


stevengj:
Could use a test…

cfitz25:
Oops, sorry ill go create one now, thanks for the heads up

stevengj (approved):

stevengj:
Maybe a NEWS item as well since this is a new method?

cfitz25:
Sorry i am new here, how would i make a NEWS item?

stevengj:
edit NEWS.md

PR #37395, opened 2020-09-04
Add `parse()` and `tryparse()` for `SHA1` and `UUID`

staticfloat:
Inspired by Kristoffer's comments in https://github.com/JuliaLang/julia/pull/37320/files

### DIFF
--- a/base/loading.jl
+++ b/base/loading.jl
@@ -91,6 +91,17 @@ struct SHA1
     end
 end
 SHA1(s::AbstractString) = SHA1(hex2bytes(s))
+parse(::Type{SHA1}, s::AbstractString) = SHA1(s)
+function tryparse(::Type{SHA1}, s::AbstractString)
+    try
+        return parse(SHA1, s)
+    catch e
+        if isa(e, ArgumentError)
+            return nothing
+        end
+        rethrow(e)
+    end
+end
 
 string(hash::SHA1) = bytes2hex(hash.bytes)
 print(io::IO, hash::SHA1) = bytes2hex(io, hash.bytes)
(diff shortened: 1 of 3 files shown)


base/loading.jl
```diff
@@ -91,6 +91,17 @@ struct SHA1
     end
 end
 SHA1(s::AbstractString) = SHA1(hex2bytes(s))
+parse(::Type{SHA1}, s::AbstractString) = SHA1(s)
+function tryparse(::Type{SHA1}, s::AbstractString)
+    try
+        return parse(SHA1, s)
+    catch e
+        if isa(e, InterruptException)
```
stevengj, line 99: Wouldn't it be better to specifically return `nothing` for an `ArgumentError` and `rethrow` everything else?

staticfloat: Yes, that does seem better to me.

KristofferC:
This is the wrong way around (in my opinion). There should not be any throws in the code when you do a `tryparse`. `tryparse` is the inner one and `parse` calls `tryparse` and does all the throwing depending on the return value of `tryparse`.

KristofferC (approved):
The stuff in my comment can be tweaked in a follow up PR though, now when the API is in place.

staticfloat:
I do think it's important to continue to get the rich information from the `parse()` function failing.  I don't think swallowing errors in `tryparse()` is that big of a deal; the contract of `tryparse()` seems like it's valid for it to swallow all errors greedily.  It would be nice if we can narrow down the source of errors, (As done in this PR to only filter out `ArgumentError`s) but if I am to `tryparse(Foo, x)` I really don't ever want that to throw for any reason that is attributable to `x`.

PR #37396, opened 2020-09-04
add `import ... as ...` syntax

### LINKED ISSUE #1255: module and import aliasing
For the module system, we can import and use a module with dot notation:

``` julia
import ArgParse
... 
    ArgParse.add_argument(p, "--opt1") 
...
```

This can be useful to prevent namespace pollution.  Because of the verbosity of module names, however, it would be nice to have module aliases:

``` julia
import ArgParse as ap 
... 
    ap.add_argument(p, "--opt1") 
...
```

JeffBezanson:
closes #1255

### DIFF
--- a/doc/src/manual/modules.md
+++ b/doc/src/manual/modules.md
@@ -13,9 +13,9 @@ is shown for illustrative purposes:
 module MyModule
 using Lib
 
-using BigLib: thing1, thing2
+using BigLib: thing1, thing2, thing3 as t3
 
-import Base.show
+import Base.show, Base.print as pr
 
 export MyType, foo
 
@@ -26,7 +26,7 @@ end
 bar(x) = 2x
 foo(a::MyType) = bar(a.x) + 1
 
-show(io::IO, a::MyType) = print(io, "MyType $(a.x)")
+show(io::IO, a::MyType) = pr(io, "MyType $(a.x)")
 end
 ```
 
@@ -85,6 +85,37 @@ functions into the current workspace:
 | `import MyModule`               | `MyModule.x`, `MyModule.y` and `MyModule.p`                                     | `MyModule.x`, `MyModule.y` and `MyModule.p` |
 | `import MyModule.x, MyModule.p` | `x` and `p`                                                                     | `x` and `p`                                 |
 | `import MyModule: x, p`         | `x` and `p`                                                                     | `x` and `p`                                 |
+
+### Import renaming
+
+An identifier brought into scope by `import` or `using` can be renamed with the keyword `as`.
+This is useful for working around name conflicts as well as for shortening names.
+For example, `Base` exports the function name `read`, but the CSV.jl package also provides `CSV.read`.
+If we are going to invoke CSV reading many times, it would be convenient to drop the `CSV.` qualifier.
+But then it is ambiguous whether we are referring to `Base.read` or `CSV.read`:
+
+```julia
+julia> read;
+
+julia> import CSV.read
+WARNING: ignoring conflicting import of CSV.read into Main
+```
+
+Renaming provides a solution:
+
+```julia
+julia> import CSV.read as rd
+```
+
+Imported packages themselves can also be renamed:
+
+```julia
+import BenchmarkTools as BT
+```
+
+`as` works with `using` only when a single identifier is brought into scope.
+For example `using CSV: read as rd` works, but `using CSV as C` does not, since it operates
+on all of the exported names in `CSV`.
 
 ### Modules and files
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -21,6 +21,8 @@ New language features
   `(x...) -> (op).(x...)`, which can be useful for passing the broadcasted version of an
   operator to higher-order functions, like for example `map(.*, A, B)` for an elementwise
   product of two arrays of arrays. ([#37583])
+* The syntax `import A as B` (plus `import A: x as y`, `import A.x as y`, and `using A: x as y`)
+  can now be used to rename imported modules and identifiers ([#1255]).
 
 Language changes
 ----------------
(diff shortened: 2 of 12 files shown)


quinnj:
Now that's an old issue number!

tkf:
Why not `using ... as ...`?  I think it's reasonable to expect that

```julia
import CSV.read as rd
rd(x::Int) = x
```

adds a method to `CSV.read`.  I think it is a good practice (or at least some people prefer) to use `using M: f` instead of `import M: f` to avoid accidentally overloading `f` and also to clarify that this function is imported just for calling it.  So, isn't it better to support `using CSV.read as rd` instead?

Also, isn't `import M.x as y` redundant if `import M: x as y` is supported?  Why not just support `import M: x as y`?  Tools like JuliaFormatter.jl may need to implement an extra transformation if there are multiple equivalent forms.  I think it might be a good idea to not introduce a redundant syntax in the first place.

JeffBezanson:
Ok, we can allow `using` as well. It's just a bit different because `using X as Y` makes less sense to me (since it doesn't clearly operate on a single identifier). But `using M: x as y` can work.

It might be ok to remove `import M.x` in the future, but I think as long as we have it `as` might as well work with it.

tkf:
Thanks for the consideration.  I think I can see the motivation behind `import M.x as y` as the syntax is more symmetry this way.  I'll probably avoid it personally but it's much lower priority than having `using M: x as y`.

> `using X as Y` makes less sense

I agree. I guess it has to be written as `import X as Y` or `using X: X as Y`.

JeffBezanson:
Added `using`.

JeffBezanson:
Triage is ok with this. The potential for abuse is very limited, and we think this will most likely only be used on the rare occasion where it's really convenient. The downvoters are free to make their case however :)

tkf:
Does it make sense to unsupport `import M.x as y` and `import M: x as y` (while keeping `using M.x as y` and `using M: x as y`)? It means that, for `import`, only `import M as N` would be supported. I'm bringing this up because it's unclear to me when you'd want to overload functions via aliases. Also, it's much harder to remove syntax than adding one. If people complain about the asymmetry between `import` and `using`, it's easy to add them later.

jmert:
Two questions about this:

1. Is it safe to assume there's a way to implement `@compat import M.T as S` in Compat.jl?
2. Does a `using` of a renamed module still work if the parent module and original module have the same name? For instance, I recently wanted to name a module `FileIO`, but I also wanted to use the `FileIO` package — would this now work
   ```julia
   module FileIO
       import FileIO as FIO    # or `@compat import FileIO as FIO`
       using .FIO
       # ...
   end
   ```
   without there being confusion due to the fact that normally the self-name is exported (i.e. `:FileIO in names(FIO) == true`).

fredrikekre:
> Is it safe to assume there's a way to implement `@compat import M.T as S` in Compat.jl?

It already exists in https://github.com/fredrikekre/ImportMacros.jl (but with the syntax `@import M.T as S` -- should be trivial to update that to `@compat import M.T as S` and move it to `Compat`.)

ararslan:
> Is it safe to assume there's a way to implement `@compat import M.T as S` in Compat.jl?

Syntactically yes but functionally not as far as I know. With this PR, the identifier `T` wouldn't come into scope at all, you would only have `S`. I believe the closest you can currently get to this (and thus how it would have to be implemented in Compat) is to do `import M.T; const S = T`, which does bring `T` into scope, or `import M; const S = M.T`, which brings `M` into scope. Whether that matters is situation-dependent, of course.

fredrikekre:
> Syntactically yes but functionally not as far as I know.

You underestimate my creativity:
```
julia> using ImportMacros

julia> @import Example as ex

julia> Example
ERROR: UndefVarError: Example not defined

julia> ex
Example
```

JeffBezanson:
I think we should keep this as-is. It's just too surprising for `import a.b as c` not to work, especially since it is so commonly written in python.

tkf:
There are other "surprising" APIs in Julia like `*` for string concatenation (for some people) and `**` even has special error presumably targeting Python programmers.  Furthermore, I think "it's familiar for Python programmers" is actually a reason to *not* add `import a.b as c` syntax because it'd mean that there is a higher chance they are using `import` without knowing that it's easy to accidentally add methods and modify the whole system.

JeffBezanson:
It just seems silly to me. ImportMacros.jl already allows it, so why would we add this syntax but then make you use that package only for one very obvious case of renaming a single identifier? Another possibility might be to disallow method extension for renamed identifiers, instead of disallowing the import syntax.

tkf:
I guess it boils down to whether or not Julia wants to be opinionated about code organization. Though I guess Julia has the lispy vibe of allowing almost everything so maybe I'm arguing something nonidiomatic.

> so why would we add this syntax but then make you use that package only for one very obvious case of renaming a single identifier?

Why not `using a: b as c`? Or do you mean to rename and extend methods? My assumption has been that allowing adding method to renamed function is not beneficial overall and mostly a footgun for new users.

> Another possibility might be to disallow method extension for renamed identifiers

I think the downside is that it might introduce confusion on the difference between `import` and `using` without `as`.

JeffBezanson:
Triage is ok with this as-is. `import a.b as c` is the single most obvious thing to try, so it should work. Adding methods to renamed functions is not bad enough to justify interfering with that.

Note: need to mention `using` in the NEWS item.

doc/src/manual/modules.md
```diff
@@ -86,6 +86,37 @@ functions into the current workspace:
...
+Renaming provides a solution:
+
+```julia
+julia> import CSV.read as rd
+```
+
+Imported packages themselves can also be renamed:
+
+```julia
+import BenchmarkTools as bt
```
ararslan, line 113:
```suggestion
import BenchmarkTools as BT
```
Can we continue the convention of upper camel case module names, at least in the documentation?

KristofferC:
Finnaly I can do `import LinearAlgebra as la`. 🎉

PR #37402, opened 2020-09-04
fix #37393: interpolation in for outer

### LINKED ISSUE #37393: Quasiquotation in `for outer $var...` expressions does not work
Julia appears not to recognise `for outer` constructions in Expressions when performing substitution. This has been tested using Julia 1.2 and 1.5.1.
  
The following snippet without `outer` works as expected, substituting the value `:p` into the `for` loop:
```
julia> g() = begin
       q = :p
       :(for $q = 1:2:3 end)
       end

julia> g()
:(for p = 1:2:3
  end)
```
However, when `outer` is added the variable substitution is not recognised:
```
julia> f() = begin
       q = :p
       :(for outer $q = 1:2:3 end)
       end

julia> f()
:(for outer $ q = 1:2:3
  end)
```

Julia version info:
```
Julia Version 1.5.1
Commit 697e782ab8 (2020-08-25 20:08 UTC)
Platform Info:
  CPU: Intel(R) Core(TM) i5-5300U CPU @ 2.30GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-9.0.1 (ORCJIT, broadwell)
Environment:
  JULIA_PATH = /usr/local/julia
  JULIA_GPG = 3673DF529D9049477F76B37566E3C7DC03D6E495
  JULIA_VERSION = 1.5.1
```

simeonschaub:
This also simplifies the logic a bit. This does now throw an error during lowering for `for outer +i = 1:3; end`, which was technically valid before, but I don't think that should be a problem?
fixes #37393


JeffBezanson:
The simplification is nice but a bit too good to be true --- we allowed `for outer in x` so that `outer` still generally works as a variable (not stealing it as a keyword).

simeonschaub:
I also can't quite figure out what this line is for:
https://github.com/JuliaLang/julia/blob/36473b1e8456238a17e1e86e22082d98b41d5a08/src/julia-parser.scm#L1668

Is this even needed anymore?

JeffBezanson:
I think the `:` line is for the discontinued comprehension syntax:
```
julia> [i for i = 1:10, :]
ERROR: syntax: comprehension syntax with `:` ranges has been removed around REPL[4]:1
```

`parse-eq*` might work, hard to say without trying it :)

simeonschaub:
> `parse-eq*` might work, hard to say without trying it :)

It almost works. Unfortunately, the current `parse-iteration-spec` allows 
```julia
[i for i
in 1:3]
```
as valid syntax, which I guess was just an oversight, but since it's actually used in https://github.com/JuliaLang/julia/blob/f90258498dfe55d8fc22d0e8a24c83d90c119135/base/compiler/ssair/inlining.jl#L1326, I guess we still need to allow that?

simeonschaub:
Ok, this should now be ready to review. It keeps the (somewhat questionable) current behavior regarding newlines, while still fixing #37393.

simeonschaub:
@JeffBezanson Would you mind taking another look at this?

JeffBezanson (approved):

simeonschaub:
I am going to rebase once #37247 is merged

PR #37403, opened 2020-09-04
Precompile `watch_file` and a few other functions used by Revise

timholy:
Together with some internal changes coming in Revise v3.0, on my machine this drops the overhead of using Revise from 3.5s to 0.5s. The change in this PR only contributes about 0.1s of that improvement, but compared to a final baseline of 0.5s it's a noticeable enhancement. The only one of these that makes a sizable impact is `watch_file`, but this does ensure that Revise doesn't trigger any additional "generic" compilation.

Because `precompile` fails silently, there's a risk that useless precompile statements bitrot, so I threw in a bunch of `@assert`s to make sure these don't go stale. I'm happy to remove them if folks don't like that. With the `@assert`s, what it means is that if someone makes changes so that these signatures aren't resolved anymore, they'll get an error during the precompile phase of building Julia. They can then either remove or modify the offending statement(s), preferably "modify" if there is a new version that is relevant.

### DIFF
--- a/contrib/generate_precompile.jl
+++ b/contrib/generate_precompile.jl
@@ -15,12 +15,14 @@ UP_ARROW = "\e[A"
 DOWN_ARROW = "\e[B"
 
 hardcoded_precompile_statements = """
-precompile(Tuple{typeof(Base.stale_cachefile), String, String})
-precompile(Tuple{typeof(push!), Set{Module}, Module})
-precompile(Tuple{typeof(push!), Set{Method}, Method})
-precompile(Tuple{typeof(push!), Set{Base.PkgId}, Base.PkgId})
-precompile(Tuple{typeof(setindex!), Dict{String,Base.PkgId}, Base.PkgId, String})
-precompile(Tuple{typeof(get!), Type{Vector{Function}}, Dict{Base.PkgId,Vector{Function}}, Base.PkgId})
+@assert precompile(Tuple{typeof(Base.stale_cachefile), String, String})
+@assert precompile(Tuple{typeof(push!), Set{Module}, Module})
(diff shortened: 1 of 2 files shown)


contrib/generate_precompile.jl
```diff
@@ -15,12 +15,17 @@ UP_ARROW = "\e[A"
...
-precompile(Tuple{typeof(push!), Set{Base.PkgId}, Base.PkgId})
-precompile(Tuple{typeof(setindex!), Dict{String,Base.PkgId}, Base.PkgId, String})
-precompile(Tuple{typeof(get!), Type{Vector{Function}}, Dict{Base.PkgId,Vector{Function}}, Base.PkgId})
+@assert precompile(Tuple{typeof(Base.stale_cachefile), String, String})
+@assert precompile(Tuple{typeof(push!), Set{Module}, Module})
+@assert precompile(Tuple{typeof(push!), Set{Method}, Method})
+@assert precompile(Tuple{typeof(push!), Set{Base.PkgId}, Base.PkgId})
+@assert precompile(Tuple{typeof(setindex!), Dict{String,Base.PkgId}, Base.PkgId, String})
+@assert precompile(Tuple{typeof(get!), Type{Vector{Function}}, Dict{Base.PkgId,Vector{Function}}, Base.PkgId})
+@assert precompile(Tuple{typeof(FileWatching.watch_file), String, Float64})
```
fredrikekre, line 24: What happens with these if you compile without FileWatching in the sysimg?

KristofferC: Yes, the ones that are dependent on a certain stdlib existing should go into conditionals similar to line ~50.

PR #37406, opened 2020-09-04
fix a possible segfault during static compilation

JeffBezanson:
Similar to #37386, for 1.6-dev


Sacha0 (approved):
Superficially LGTM! Thanks Jeff! :)

PR #37407, opened 2020-09-05
Bump libuv to latest commit

musm:
Includes 2 commits:
- security patch to fix buffer overruns in realpath/reallink
- Allows utf-16 surrogate pairs to pass through on windows (emoji support on windows)


vtjnash (approved):

musm:
I don't think we need a test for the buffer overrun issue in julia since it's already tested upstream.

PR #37408, opened 2020-09-05
Bump MbedTLS

musm:
What's with these two files:
https://github.com/JuliaLang/julia/tree/master/deps/checksums/mbedtls-2.16.0-apache.tgz
https://github.com/JuliaLang/julia/tree/master/deps/checksums/mbedtls-2.16.0-gpl.tgz

Is there a reason they are not handled by BB? I don't know if I need to do anything with them or if they are stragglers and should be removed.


ararslan:
Bump @staticfloat, can you review this?

musm:
@ararslan looks like you modified these files ages ago, so you probably don't remember, but in the off chance, do you know anything about:


> What's with these two files:
> https://github.com/JuliaLang/julia/tree/master/deps/checksums/mbedtls-2.16.0-apache.tgz
> https://github.com/JuliaLang/julia/tree/master/deps/checksums/mbedtls-2.16.0-gpl.tgz
> 
> Is there a reason they are not handled by BB? I don't know if I need to do anything with them or if they are stragglers and should be removed.

ararslan:
Oh, I see. Yeah those are the checksums for the source tarballs, i.e. what's downloaded when you build with `USE_BINARYBUILDER=0`.

staticfloat:
Yeah, I think you need to delete the checksums for those source archives, and then generate the checksums for the new source archives by building mbedtls with `USE_BINARYBUILDER=0`

musm:
Makes sense, since this is only touching the BB version, we should be ok with the PR as it currently stands.

musm:
NVM, fixed the non-BB version as well.

mbedtls moved over fully to the Apache license starting with this version and started to serve their downloads from Github directly, so I had to tweak the make file download location, as well.

deps/mbedtls.mk
```diff
@@ -1,12 +1,8 @@
 ## mbedtls
 
 ifneq ($(USE_BINARYBUILDER_MBEDTLS), 1)
-ifeq ($(USE_GPL_LIBS), 1)
-MBEDTLS_SRC = mbedtls-$(MBEDTLS_VER)-gpl
-else
-MBEDTLS_SRC = mbedtls-$(MBEDTLS_VER)-apache
-endif
-MBEDTLS_URL = https://tls.mbed.org/download/$(MBEDTLS_SRC).tgz
+MBEDTLS_SRC = mbedtls-$(MBEDTLS_VER)
```
musm, line 4: @staticfloat could you take another look to confirm this is all good with you?

staticfloat (approved):

ararslan:
Should generally wait for CI to finish before merging, FYI.

musm:
Yep, I usually do. First commit is the one that passed anyways, second commit modifies the non BB version which the buildbots don't test, so I figured it was fine.

PR #37410, opened 2020-09-05
allow slurping in lhs of assignment

### LINKED ISSUE #2626: a, b... = [1,2,3]

simeonschaub:
~~This is currently branched off #37268. I will rebase once that is merged, but for now, only the last commit is relevant.~~ This implements @StefanKarpinski's comment in https://github.com/JuliaLang/julia/issues/2626#issuecomment-587098181. I am not sure, `slurp_rest` is the best name and API for this, perhaps it would make more sense to extend `iterate_and_index` to return a third function for slurping, since `slurp_rest` currently relies on what `iterate` does, which could lead to gotchas if users overload `iterate_and_index`.
fixes #2626

### DIFF
--- a/base/abstractarray.jl
+++ b/base/abstractarray.jl
@@ -2399,3 +2399,11 @@ function hash(A::AbstractArray, h::UInt)
 
     return h
 end
+
+# The semantics of `collect` are weird. Better to write our own
+function rest(a::AbstractArray{T}, state...) where {T}
+    v = Vector{T}(undef, 0)
+    # assume only very few items are taken from the front
+    sizehint!(v, length(a))
+    return foldl(push!, Iterators.rest(a, state...), init=v)
+end
--- a/base/tuple.jl
+++ b/base/tuple.jl
@@ -97,7 +97,7 @@ function indexed_iterate(I, i, state)
 end
 
 rest(t::Tuple) = t
-rest(t::NTuple{N}, i::Int) where {N} = ntuple(x -> getfield(t, x+i-1), N-i+1)
+rest(t::Tuple, i::Int) = ntuple(x -> getfield(t, x+i-1), length(t)-i+1)
 rest(a::Array, i::Int=1) = a[i:end]
 rest(itr, state...) = Iterators.rest(itr, state...)
 
--- a/base/tuple.jl
+++ b/base/tuple.jl
@@ -96,6 +96,11 @@ function indexed_iterate(I, i, state)
     x
 end
 
+rest(t::Tuple) = t
+rest(t::NTuple{N}, i::Int) where {N} = ntuple(x -> getfield(t, x+i-1), N-i+1)
+rest(a::Array, i::Int=1) = a[i:end]
+rest(itr, state...) = Iterators.rest(itr, state...)
+
 # Use dispatch to avoid a branch in first
 first(::Tuple{}) = throw(ArgumentError("tuple must be non-empty"))
 first(t::Tuple) = t[1]
--- a/base/tuple.jl
+++ b/base/tuple.jl
@@ -96,6 +96,33 @@ function indexed_iterate(I, i, state)
     x
 end
 
+"""
+    Base.rest(collection[, itr_state])
+
+Generic function for taking the tail of `collection`, starting from a specific iteration
+state `itr_state`. Return a `Tuple`, if `collection` itself is a `Tuple`, a `Vector`, if
+`collection` is an `AbstractArray` and `Iterators.rest(collection[, itr_state])` otherwise.
+Can be overloaded for user-defined collection types to customize the behavior of slurping
+in assignments, like `a, b... = collection`.
+
+!!! compat "Julia 1.6"
+    `Base.rest` requires at least Julia 1.6.
+
+# Examples
+```jldoctest
+julia> a = [1 2; 3 4]
+2×2 Matrix{Int64}:
+ 1  2
+ 3  4
+
+julia> first, state = iterate(a)
+(1, 2)
+
+julia> first, Base.rest(a, state)
+(1, [3, 2, 4])
+```
+"""
+function rest end
 rest(t::Tuple) = t
 rest(t::Tuple, i::Int) = ntuple(x -> getfield(t, x+i-1), length(t)-i+1)
 rest(a::Array, i::Int=1) = a[i:end]
--- a/src/julia-syntax.scm
+++ b/src/julia-syntax.scm
@@ -1430,7 +1430,8 @@
                 ,@(reverse after)
                 (unnecessary (tuple ,@(reverse elts))))
         (let ((L (car lhss))
-              (R (car rhss)))
+              ;; rhss can be null iff L is a vararg
+              (R (if (null? rhss) '() (car rhss))))
           (cond ((and (symbol-like? L)
                       (or (not (pair? R)) (quoted? R) (equal? R '(null)))
                       ;; overwrite var immediately if it doesn't occur elsewhere
@@ -1442,6 +1443,16 @@
                        (cons (make-assignment L R) stmts)
                        after
                        (cons R elts)))
+                ((vararg? L)
+                 (if (null? (cdr lhss))
+                     (let ((temp (make-ssavalue)))
+                       `(block ,@(reverse stmts)
+                               (= ,temp (tuple ,@rhss))
+                               ,@(reverse after)
+                               (= ,(cadr L) ,temp)
+                               (unnecessary (tuple ,@(reverse elts) (... ,temp)))))
+                     (error (string "invalid \"...\" on non-final assignment location \""
+                                    (cadr L) "\""))))
                 ((vararg? R)
                  (let ((temp (make-ssavalue)))
                    `(block ,@(reverse stmts)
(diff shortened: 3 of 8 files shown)


JeffBezanson:
Cool! My current thinking is that if we do this, it should use `Base.rest` (or `tail`?), which can be overloaded appropriately --- arrays can return arrays, if you had a linked list (:rofl: ) it could return the list tail, tuples can return tuples, and infinite or lazy iterators can return `Rest` iterators. I don't think this is necessarily inconsistent with varargs, since arguments are always implicitly a tuple

simeonschaub:
Ok, that can certainly be discussed, I was just under the impression that Stefan's comment concluded the discussion in #2626. I personally would still favor just always returning tuples, because I think it's going to be difficult to find a rule for which container types should slurp to which type for the rest and it's simpler to remember that it's just always a tuple. For example, what should `a..., = 1` do? `a, b... = 1`? I see that it's not necessarily inconsistent with varargs, but to me, it still feels a bit weird for such a language feature to change meaning based on the input type instead  of returning a tuple.

JeffBezanson:
It's not changing meaning, it's using a different implementation of that meaning. Multiple data structures support the concept of `rest`, and each one implements it by returning a smaller version of the same structure. Using only tuples prevents e.g. infinite iterators from working, when they could work. Lots of syntax features like this lower to overloadable functions.

simeonschaub:
I mostly want this just for tuples, so what to do for other iterators is not a hill I'm willing to die on. What do you reckon would be the best way forward here? Should this just be triaged, or would it make sense to reopen the discussion in #2626, so that people that participated there can chime in?

JeffBezanson:
Ok, let's have this call `Base.rest(itr, state)`, and it can be overloaded to be eager for e.g. arrays and tuples.

simeonschaub:
So I'm thinking as the default return types of `Base.rest`:
* `Tuple` -> `Tuple`
* `AbstractArray` -> `Vector`
* everything else -> `Iterators.rest`
Does that make sense, or is there anything else that would make sense to special-case? Perhaps `UnitRange`?

simeonschaub:
Ah, this doesn't work with just passing the iteration state to `Base.rest`, because with #37268, `iterate_and_index(x)[1]`, won't tell us anything about the current iteration state for things like tuples or vectors, since it just returns the object itself and indexes with the index generated by lowering. We either need to change how `iterate_and_index` works for those or need to pass the index generated by lowering to `Base.rest` as well.

base/tuple.jl
```diff
@@ -81,21 +81,60 @@ function _maxlength(t::Tuple, t2::Tuple, t3::Tuple...)
...
+rest(t::NTuple{N}, i::Int) where {N} = ntuple(x -> getfield(t, x+i-1), N-i+1)
+rest(a::Array, i::Int=1) = a[i:end]
+# The semantics of `collect` are weird. Better to write our own
+function rest(a::AbstractArray{T}, state...) where {T}
+    v = Vector{T}(undef, 0)
+    # assume only very few items are taken from the front
+    sizehint!(v, length(a))
+    return foldl(push!, Iterators.rest(a, state...), init=v)
+end
+rest(itr, state...) = Iterators.rest(itr, state...)
```
simeonschaub, line 136: @JeffBezanson I would love to have some feedback on this design. It's a bit uglier than I initially imagined, but I don't think we can really avoid that with the way `iterate_and_index` currently works. `base/tuple` is probably also the wrong place for most of this, but I am still not sure, what the best home for `Base.rest` is.

base/tuple.jl
```diff
@@ -116,6 +116,25 @@ iterate_and_index(x) = (destruct_iterate, select_first)
 # to make inference's life easier.
 iterate_and_index(::Nothing) = throw(MethodError(iterate, (nothing,)))
 
+function _rest(itr, state, _)
+    @_inline_meta
+    state === () && return rest(itr)
```
JeffBezanson, line 121: I think this case can be handled entirely in the front end; it only occurs for `xs... = itr`?

simeonschaub: Hmm, so are you saying we skip the whole `iterate_and_index` entirely in favor of plain `iterate` in this case?

simeonschaub: Ah sorry, I misunderstood! So we should just produce a call to `Base.rest(itr)` during lowering in that case?

base/tuple.jl
JeffBezanson: The methods can go with their types, like iterate and getindex tend to.

simeonschaub: Yes, that's what I planned on doing. I just didn't really know where to put the generic `Base.rest` function. `base/tuple.jl` seems like a bit of a weird place for it.

JeffBezanson:
So the interesting thing is that this feature is slightly incompatible with #37268, since that change means we don't always have an iteration state available. To get elements, you don't need the iteration state, but for `...` you do. Hence the need for `_rest`. But it probably doesn't matter much since you only have to implement `_rest` if you implemented `indexed_iterate` before (right?), and nobody was supposed to do that :)

JeffBezanson:
I think we are not going to do the `indexed_iterate` lowering change for now, so this can be rebased on master and we can merge it.

src/julia-syntax.scm
```diff
@@ -2075,26 +2087,43 @@
...
+                      ,.(map (lambda (i lhs)
+                               (expand-forms
+                                 (if (and (pair? lhs) (eq? (car lhs) '|...|))
+                                     `(= ,(cadr lhs) (call (top rest) ,xx ,.(if (eq? i 0) '() `(,st))))
+                                     (lower-tuple-assignment
+                                      (if (= i (- n 1))
+                                          (list lhs)
+                                          (list lhs st))
+                                      `(call (top indexed_iterate)
+                                             ,xx ,(+ i 1) ,.(if (eq? i 0) '() `(,st)))))))
```
simeonschaub, line 2123: This stumped me quite a bit. Shouldn't `` `(,st) `` be exactly the same as `st`? Somehow Julia breaks if I replace this though.

simeonschaub:
So we decided to make `Base.rest` part of the public API, but not export it, right? Should probably have a docstring and some separate tests then. Also still needs news. The actual implementation should pretty much be done from my side now.

src/julia-syntax.scm
JeffBezanson: ``` `(,st)``` is the same as `(list st)`, and `,.` splats lists. It's the mutating version of `,@`, and so a kind of old-fashioned micro optimization. `,.` should probably not be used anymore.

simeonschaub: Ah, that makes total sense! I think I was just confusing it with the way Julia does unquoting, but of course in Lisp, parentheses are actually relevant. Thanks!

base/abstractarray.jl
```diff
@@ -2399,3 +2399,11 @@ function hash(A::AbstractArray, h::UInt)
 
     return h
 end
+
+# The semantics of `collect` are weird. Better to write our own
```
JeffBezanson, line 2403: How about `collect(T, Iterators.rest(a, state))`? Then for the 1-argument case `Vector{T}(a)`. Better to avoid repeated `push!` since it's a bit slow.

base/tuple.jl
```diff
@@ -96,6 +96,11 @@ function indexed_iterate(I, i, state)
     x
 end
 
+rest(t::Tuple) = t
+rest(t::NTuple{N}, i::Int) where {N} = ntuple(x -> getfield(t, x+i-1), N-i+1)
```
JeffBezanson, line 100:
```suggestion
rest(t::Tuple, i::Int) = ntuple(x -> getfield(t, x+i-1), length(t)-i+1)
```

JeffBezanson:
Looking good.

base/abstractarray.jl
simeonschaub: The problem with that is this:
```
julia> collect(Int, Iterators.rest([1 2; 3 4]))
2×2 Array{Int64,2}:
 1  2
 3  4

julia> collect(Int, Iterators.rest([1 2; 3 4], 1))
4-element Array{Int64,1}:
 1
 3
 2
 4
```
Perhaps we can solve that with doing either a `reshape` or just a single-slice `view` of `a` first though.

base/tuple.jl
simeonschaub: Oh, that's a good point! Didn't realize that:
```
julia> (1, 1.) isa NTuple{2}
false
```

base/abstractarray.jl
simeonschaub: > Then for the 1-argument case Vector{T}(a).

That won't work for the multi-dimensional case, unfortunately:
```
julia> Vector{Int}([1 2; 3 4])
ERROR: MethodError: no method matching Array{Int64,1}(::Array{Int64,2})
Closest candidates are:
  Array{Int64,1}(::AbstractArray{S,N}) where {T, N, S} at array.jl:562
  Array{Int64,1}() where T at boot.jl:425
  Array{Int64,1}(::UndefInitializer, ::Int64) where T at boot.jl:406
  ...
Stacktrace:
 [1] top-level scope at REPL[25]:1
```
`collect` also doesn't always return a `Vector` and I think it would be weird if `Base.rest(::A)` and `Base.rest(::A, st)` returned different types.

simeonschaub: Also, note that:
```
julia> Base.IteratorSize(Iterators.rest([1 2; 3 4], 1))
Base.SizeUnknown()
```
so `collect` will end up just using `push!` anyways.

StefanKarpinski:
OMG, I did not think this was happening!

PR #37411, opened 2020-09-05
Fix unnecessary use of closure boxing in type inference

yuyichao:
This should fix all appearance of `Core.Box` in the `Core` module with the exception of
`simple_walk_constraint` where it is actually necessary.

Most of the cases are the normal readonly closures fixed by adding a `let` around it
or moving assignments.
There are also two cases of recursive closures fixed by moving the recursion
out of the closure.


timholy:
Any noticeable improvements in compile times with this? Or are these not hot-spots?

yuyichao:
No I don’t have time measurements. These are just from scanning the code...

Issue #37405, opened 2020-09-04
pointer(A, i) doesn't work for reinterpret(view(reinterpret(view(A))


BioTurboNick:
For a `ReinterpretArray` which has nested subarrays and reshapes, the `pointer(A, i)` method breaks, while `pointer` itself works fine.

e.g. 
```
Base.ReinterpretArray{UInt32,1,UInt16,SubArray{UInt16,1,Base.ReinterpretArray{UInt16,1,Int64,SubArray{Int64,1,Array{Int64,1},Tuple{UnitRange{Int64}},true}},Tuple{UnitRange{Int64}},true}}
```

Specifically, `pointer(A, i)` relies on `_memory_offset(A, i)`, which relies on `strides(A)`. There is no `strides` method that handles this type.

I briefly looked into how the StridedArray types were defined and tried to add a type to make this work, but doing so broke some tests by defining some arrays as strided which shouldn't be. (e.g. `reshape(1:32, (4, 8))`) It also started making some of the StridedArray type definitions get really complex.

Is there a way to solve this that I haven't seen?

BioTurboNick:
I think I found a solution. Submitting a PR #37414

BioTurboNick:
Closing because the PR was merged.

PR #37414, opened 2020-09-05
Enable stride/strides to work on a ReinterpretArray with complex Strided parents.

BioTurboNick:
As described in issue #37405, a complex nested array type like reinterpret(view(reinterpret(view(A)) is not defined as a `StridedArray`. I tried implementing a type-based solution but it caused things to break and made the type signatures very long.

So instead I just implemented `stride`/`strides` methods that check if the parent is a `StridedArray`, erroring otherwise.

### DIFF
--- a/base/reinterpretarray.jl
+++ b/base/reinterpretarray.jl
@@ -48,9 +48,10 @@ end
 reinterpret(::Type{T}, a::ReinterpretArray) where {T} = reinterpret(T, a.parent)
 
 # Definition of StridedArray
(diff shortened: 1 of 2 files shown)


BioTurboNick:
Why are the 32-bit tests failing? Int16 and Int32 should behave the same regardless of architecture, right?

simeonschaub:
Int literals are `Int32`s on 32 bit platforms, so `1:20` is actually a `UnitRange{Int32}` there.

BioTurboNick:
Is the analyzegc failure related to this PR?

timholy:
The analyzegc test looks mostly at Julia's C code, so it shouldn't be related.

timholy:
Thanks!

PR #37417, opened 2020-09-05
Optimize codegen of `isa` Union type

yuyichao:


timholy:
So if I understand correctly, when `A` and `B` are concrete this optimizes `if x isa Union{A,B}` to `if x isa A || x isa B`? Nice, I've wanted that for a while now.

JeffBezanson (approved):

PR #27126, opened 2018-05-16
Introduce improved SROA pass

Keno:
This is a rebased and fixed version of the improved SROA pass from #26778.
There's a decent piece of new infrastructure wrapped up in this: The ability
to insert new nodes during compaction. This is a bit tricky because it requires
tracking which version of the statements buffer a given SSAValue belongs to.
At the moment this is done mostly manually, but I'm hoping to clean that up
in the future. The idea of the new SROA pass is fairly straightforward:
Given a use of an interesting value, it traces through all phi nodes, finding
all leaves, applies whatever transformation to those leaves and then re-inserts
a phi nest corresponding to the phi nest of the original value.

PR #37418, opened 2020-09-05
Fix Dict key check in type inference

yuyichao:
This has been there since #27126 (9100329d189b3d2ac6d1e03bfc7431cbe8d18c00)

Judging by the use of `reverse_mapping[stmt_val]` below
this should either be a branch condition or an assert.


KristofferC:
Can be merged? cc @keno

Keno (approved):

Issue #33210, opened 2019-09-10
pathof(::Module) doesn't show the path where the module actually got imported from


KristofferC:
```
julia> using Example

julia> pathof(Example)
"/Users/kristoffer/.julia/packages/Example/aqsx3/src/Example.jl"

(v1.2) pkg> add Example@0.5.1
 Resolving package versions...
  Updating `~/.julia/environments/v1.2/Project.toml`
  [7876af07] ↓ Example v0.5.3 ⇒ v0.5.1
  Updating `~/.julia/environments/v1.2/Manifest.toml`
  [7876af07] ↓ Example v0.5.3 ⇒ v0.5.1

julia> pathof(Example)
"/Users/kristoffer/.julia/packages/Example/kH44X/src/Example.jl"
```

I would have thought `pathof` should point to the path as it was when the module was imported. Not the file that would have been used if the module was not imported but got imported right now.

vtjnash:
Are you opening this as a self-assignment, or hoping someone else will pick it up? I guess we need to keep around a `Dict{Module => String}` giving the location `require` decided to get it from?

I wonder if we should also deprecate this in favor of just having `rootof(Example) == "/Users/kristoffer/.julia/packages/Example/aqsx3"`?  (although needs the same implementation fix either way)

KristofferC:
> Are you opening this as a self-assignment, or hoping someone else will pick it up?

Both :)

> I guess we need to keep around a Dict{Module => String} giving the location require decided to get it from?

Yeah, that seems simple enough

PR #37421, opened 2020-09-05
Code loading: track chosen cachefile and load the module path

timholy:
These changes allow Revise to leverage precompiled Base machinery, dropping about 150ms from Revise's extra latency penalty for the first package load. Especially for the upcoming Revise3, that's a pretty large fraction of its total latency (ballpark 750ms for both loading Revise and the penalty for the subsequent package load).

Revise had an `mtime` hack to try to guess which precompile file was being used, so this should make things more robust too since it keeps track of which one actually got loaded.

### DIFF
--- a/base/loading.jl
+++ b/base/loading.jl
@@ -824,9 +824,19 @@ function require(into::Module, mod::Symbol)
     return require(uuidkey, cache)
 end
 
+struct PkgOrigin
+    # version::VersionNumber
+    # path::String
+    cachepath::Union{String,Nothing}
+end
+const pkgorigins = Dict{PkgId,PkgOrigin}()
+
 function require(uuidkey::PkgId, cache::TOMLCache=TOMLCache())
     if !root_module_exists(uuidkey)
-        _require(uuidkey, cache)
+        cachefile = _require(uuidkey, cache)
+        if cachefile !== nothing
+            pkgorigins[uuidkey] = PkgOrigin(cachefile)
+        end
         # After successfully loading, notify downstream consumers
         for callback in package_callbacks
             invokelatest(callback, uuidkey)
@@ -881,6 +891,7 @@ function unreference_module(key::PkgId)
     end
 end
 
+# Returns `nothing` or the name of the newly-created cachefile
 function _require(pkg::PkgId, cache::TOMLCache)
     # handle recursive calls to require
     loading = get(package_locks, pkg, false)
@@ -942,7 +953,7 @@ function _require(pkg::PkgId, cache::TOMLCache)
                     if isa(m, Exception)
                         @warn "The call to compilecache failed to create a usable precompiled cache file for $pkg" exception=m
                     else
-                        return
+                        return cachefile
                     end
                 end
             end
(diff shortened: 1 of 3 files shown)


base/loading.jl
```diff
@@ -1246,6 +1252,13 @@ module_build_id(m::Module) = ccall(:jl_module_build_id, UInt64, (Any,), m)
 isvalid_cache_header(f::IOStream) = (0 != ccall(:jl_read_verify_header, Cint, (Ptr{Cvoid},), f.ios))
 isvalid_file_crc(f::IOStream) = (_crc32c(seekstart(f), filesize(f) - 4) == read(f, UInt32))
 
+struct CacheHeaderIncludes
```
KristofferC, line 1255: For the future, it would be good to include the path in which we load the package from here as well. Then issue https://github.com/JuliaLang/julia/issues/33210 could be fixed really easily. It (of course) doesn't have to be done here but I bring it up because maybe the name of the struct is a bit too narrow if we later include the load path?

timholy: I'm not sure this is the right place, though. How about I generalize `pkgcachefiles` to be a `Dict(id => (pkgpath, cachefilepath))`?

KristofferC: Sure, at some point I think I also want to put a version numer in there. Anyway, just wanted to to wanted to point out that in the future we might want even more meta data available from loading.

timholy: This implementation should future-proof against additional info.

base/loading.jl
```diff
@@ -813,13 +813,18 @@ function require(into::Module, mod::Symbol)
     return require(uuidkey, cache)
 end
 
-const pkgcachefiles = Dict{PkgId,String}()
+struct PkgOrigin
```
KristofferC, line 816: <3

timholy:
An alternative to `PkgOrigin` would be to move Revise's core data structures (or suitably-modified versions of them) to Base: [`PkgData`](https://github.com/timholy/Revise.jl/blob/a1a4020f84f5ef8c45678a6682d7fa1619d36c4c/src/types.jl#L120-L133) and its dependencies [`FileInfo`](https://github.com/timholy/Revise.jl/blob/a1a4020f84f5ef8c45678a6682d7fa1619d36c4c/src/types.jl#L91-L94), [`PkgFiles`](https://github.com/timholy/CodeTracking.jl/blob/78ce8a44a12f138a21615ce457762556c6b93adf/src/pkgfiles.jl#L10-L14), and their dependencies. I'm a bit hesitant to suggest this because (1) it's several types that aren't immediately useful in Base, and (2) some of these struct definitions have recently changed, and if they move into Base then the opportunity for quick-fixes diminishes. OTOH, a major part of Revise's remaining startup time is compiling various Dict and OrderedDict methods that operate on these containers, and moving them to Base would allow another considerable reduction in startup latency.

With Revise3's aggregate latency (time to load package + extra overhead for loading the *next* package while compiling Revise's internals) approaching 0.75s, my overall sense is that now is not the time to contemplate this, but that perhaps around the 1.7 time frame (once Revise3 has had a chance to stabilize) it might be time to consider this option more seriously. But of course, by that point maybe we'll have better precompilation, in which case the benefits diminish unless we think that a lot of the information might be useful anyway.

Thoughts?

KristofferC:
> Thoughts?

Maybe a bit overkill to move the Revise stuff into Base. As you say, perhaps reevaluate for 1.7?

vtjnash:
Ah, nice. IIUC, this will allow us to trivially fix `pathof` and `pkgdir` to now return the correct answers (#33210)?

timholy:
Yeah, someone will have to figure out the right place to set those values, but presuming that's fairly straightforward this should indeed make it pretty easy.

fredrikekre:
See #37586

PR #26948, opened 2018-05-01
[NewOptimizer] Make stmt_effect_free less aggressive

Keno:
Previously, the new optimizer used the pre-existing `effect_free` function
to determine whether it is safe to remove an unreferenced statement. However,
this function ignores many builtin's error cases, causing them to be
removed when that is not legal to do (because that would potentially remove
an exception that would otherwise be thrown). Start fixing this, by
introducing a version of the function that is correct for a subset of
intresting functions. We will likely need to expand this when we look
at the benchmarks, but this should be correct for now.

PR #37423, opened 2020-09-06
Some `getfield` related fixes in inference

yuyichao:
* Use the field index passed in in `lift_leaves`

    The caller has already done all the computation including bound checking.
    The `field` computed in this function is also affecting all the following iterations
    which is almost certainly wrong.

* Remove unnecessary type check on `field` in `lift_leaves` since it is always `Int`

* Move a branch disabling `return nothing` higher up

* Remove some duplicated calculation on field index in `getfield_elim_pass!`

* Fix `try_compute_fieldidx` to return `nothing` for non-`Int` `Integer` field index.

    This can cause `getfield_nothrow` to return incorrect result.
    It also gives the caller worse type info about the return value.

* Teach `getfield_nothrow` that `isbits` field cannot be undefined and getfield on such field cannot throw.

    This is already handled in `isdefined_tfunc`.

* Fix a few wrong use of `isbits` in dead branches

----

Ref #26948 (fa02d34496eb2bed9a55e4f77cb135a6b4180edc)
Ref #27126 (9100329d189b3d2ac6d1e03bfc7431cbe8d18c00)

### DIFF
--- a/base/compiler/ssair/passes.jl
+++ b/base/compiler/ssair/passes.jl
@@ -288,7 +288,7 @@ function lift_leaves(compact::IncrementalCompact, @nospecialize(stmt),
             else
                 def = compact[leaf]
             end
-            if is_tuple_call(compact, def) && isa(field, Int) && 1 <= field < length(def.args)
+            if is_tuple_call(compact, def) && 1 <= field < length(def.args)
                 lifted = def.args[1+field]
                 if is_old(compact, leaf) && isa(lifted, SSAValue)
                     lifted = OldSSAValue(lifted.id)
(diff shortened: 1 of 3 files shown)


base/compiler/ssair/passes.jl
```diff
@@ -342,22 +340,22 @@ function lift_leaves(compact::IncrementalCompact, @nospecialize(stmt),
             else
                 typ = compact_exprtype(compact, leaf)
                 if !isa(typ, Const)
+                    # Disabled since #27126
+                    return nothing
```
Keno, line 344: I guess I should probably get back to figuring out why this broke.

Keno (approved):

Keno:
Merged in #39036.

PR #37429, opened 2020-09-06
fix #37312 fast extrema computation on sparse arrays

### LINKED ISSUE #37312: extrema on SparseArray doesn't exploit sparsity
Calling the extrema function on a SparseArray falls back to iterating over all elements on the Array (including zeros).
This makes it extremely slow for large sparse matrices. It would be better to compute the extrema of the nonzeros and then account for the zeros (if there are zeros).

Kind regards,
Tom

cfauchereau:
Fix #37312
Exploit sparsity to speed up extrema computation. It iterates over non-zeros values and then account for the zero value if necessary.
It is consistently faster than the generic algorithm.

### DIFF
--- a/stdlib/SparseArrays/src/higherorderfns.jl
+++ b/stdlib/SparseArrays/src/higherorderfns.jl
@@ -4,7 +4,7 @@ module HigherOrderFns
 
 # This module provides higher order functions specialized for sparse arrays,
 # particularly map[!]/broadcast[!] for SparseVectors and SparseMatrixCSCs at present.
-import Base: map, map!, broadcast, copy, copyto!
+import Base: map, map!, broadcast, copy, copyto!, _extrema_dims, _extrema_itr
 
 using Base: front, tail, to_shape
 using ..SparseArrays: SparseVector, SparseMatrixCSC, AbstractSparseVector, AbstractSparseMatrixCSC,
@@ -29,6 +29,7 @@ using LinearAlgebra
 # (11) Define broadcast[!] methods handling combinations of scalars, sparse vectors/matrices,
 #       structured matrices, and one- and two-dimensional Arrays.
 # (12) Define map[!] methods handling combinations of sparse and structured matrices.
+# (13) Define extrema methods optimized for sparse vectors/matrices.
 
 
 # (0) BroadcastStyle rules and convenience types for dispatch
@@ -1153,5 +1154,61 @@ map(f::Tf, A::SparseOrStructuredMatrix, Bs::Vararg{SparseOrStructuredMatrix,N}) 
     (_checksameshape(A, Bs...); _noshapecheck_map(f, _sparsifystructured(A), map(_sparsifystructured, Bs)...))
 map!(f::Tf, C::AbstractSparseMatrixCSC, A::SparseOrStructuredMatrix, Bs::Vararg{SparseOrStructuredMatrix,N}) where {Tf,N} =
     (_checksameshape(C, A, Bs...); _noshapecheck_map!(f, C, _sparsifystructured(A), map(_sparsifystructured, Bs)...))
+
+
+# (13) extrema methods optimized for sparse vectors/matrices.
+function _extrema_itr(f, A::SparseVecOrMat)
+    M = length(A)
+    iszero(M) && throw(ArgumentError("Sparse array must have at least one element."))
+    N = nnz(A)
+    iszero(N) && return f(zero(eltype(A))), f(zero(eltype(A)))
+    vmin, vmax = _extrema_itr(f, nonzeros(A))
+    if N != M
+        f0 = f(zero(eltype(A)))
+        vmin = min(f0, vmin)
+        vmax = max(f0, vmax)
+    end
+    vmin, vmax
+end
+
+function _extrema_dims(f, x::SparseVector, dims)
+    sz = zeros(1)
+    for d in dims
+        sz[d] = 1
+    end
+    if sz == [1] && !iszero(length(x))
+        return [_extrema_itr(f, x)]
+    end
+    invoke(_extrema_dims, Tuple{Any, AbstractArray, Any}, f, x, dims)
+end
+
+function _extrema_dims(f, A::AbstractSparseMatrix, dims)
+    sz = zeros(2)
+    for d in dims
+        sz[d] = 1
+    end
+    if sz == [1, 0] && !iszero(length(A))
+        T = eltype(A)
+        B = Array{Tuple{T,T}}(undef, 1, size(A, 2))
+        @inbounds for col_idx in 1:size(A, 2)
+            col = @view A[:,col_idx]
+            fx = (nnz(col) == size(A, 1)) ? f(A[1,col_idx]) : f(zero(T))
+            B[col_idx] = (fx, fx)
+            for x in nonzeros(col)
+                fx = f(x)
+                if fx < B[col_idx][1]
+                    B[col_idx] = (fx, B[col_idx][2])
+                elseif fx > B[col_idx][2]
+                    B[col_idx] = (B[col_idx][1], fx)
+                end
+            end
+        end
+        return B
+    end
+    invoke(_extrema_dims, Tuple{Any, AbstractArray, Any}, f, A, dims)
+end
+
+_extrema_dims(f, A::SparseVector, ::Colon) = _extrema_itr(f, A)
+_extrema_dims(f, A::AbstractSparseMatrix, ::Colon) = _extrema_itr(f, A)
 
 end
--- a/stdlib/SparseArrays/test/higherorderfns.jl
+++ b/stdlib/SparseArrays/test/higherorderfns.jl
@@ -687,4 +687,37 @@ end
     @test SparseMatStyle(Val(3)) == Broadcast.DefaultArrayStyle{3}()
 end
 
+@testset "extrema" begin
+    n = 10
+    A = sprand(n, n, 0.2)
+    B = Array(A)
+    C = Array{Real}(undef, 0, 0)
+    x = sprand(n, 0.2)
+    y = Array(x)
+    z = Array{Real}(undef, 0)
+    f(x) = x^3
+    @test extrema(A) == extrema(B)
+    @test extrema(x) == extrema(y)
+    @test extrema(f, A) == extrema(f, B)
+    @test extrema(f, x) == extrema(f, y)
+    @test extrema(spzeros(n, n)) == (0.0, 0.0)
+    @test extrema(spzeros(n)) == (0.0, 0.0)
+    @test_throws ArgumentError extrema(spzeros(0, 0))
+    @test_throws ArgumentError extrema(spzeros(0))
+    @test extrema(sparse(ones(n, n))) == (1.0, 1.0)
+    @test extrema(sparse(ones(n))) == (1.0, 1.0)
+    @test extrema(A; dims=:) == extrema(B; dims=:)
+    @test extrema(A; dims=1) == extrema(B; dims=1)
+    @test extrema(A; dims=2) == extrema(B; dims=2)
+    @test extrema(A; dims=(1,2)) == extrema(B; dims=(1,2))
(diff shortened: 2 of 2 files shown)


jebej:
Maybe merge the two methods?

clement-f:
> Maybe merge the two methods?

I agree. I merged them in higherorderfns.jl which is made for these kind of functions if i understand it right.
minimum and maximum are still ten times faster on large arrays so there is room for improvements.

clement-f:
Actually the performance issue remains with Arrays so this is extrema in Base which is slow. This is out of the scope of this pull request.
The reason may be that minimum and maximum use mapreduce but extrema don't.

dkarrasch:
Yes, the generic case is tracked by https://github.com/JuliaLang/julia/issues/31442 and there is some promising work in https://github.com/JuliaLang/julia/pull/36265.

dkarrasch (approved):
LGTM!

dkarrasch:
I wonder whether we should do the same for `minimum` and `maximum`. I just quickly benchmarked the following:
```julia
julia> using SparseArrays, BenchmarkTools

julia> x = sprandn(1000, 1, 0.2);

julia> @btime minimum($x);
  617.713 ns (5 allocations: 144 bytes)

julia> x = sparse(vec(x));

julia> @btime minimum($x);
  505.575 ns (0 allocations: 0 bytes)
```

stdlib/SparseArrays/src/higherorderfns.jl
```diff
@@ -1154,4 +1155,23 @@ map(f::Tf, A::SparseOrStructuredMatrix, Bs::Vararg{SparseOrStructuredMatrix,N})
 map!(f::Tf, C::AbstractSparseMatrixCSC, A::SparseOrStructuredMatrix, Bs::Vararg{SparseOrStructuredMatrix,N}) where {Tf,N} =
     (_checksameshape(C, A, Bs...); _noshapecheck_map!(f, C, _sparsifystructured(A), map(_sparsifystructured, Bs)...))
 
+
+# (13) extrema methods optimized for sparse vectors/matrices.
+function extrema(f, A::SparseVecOrMat)
+    if iszero(length(A))
```
dkarrasch, line 1161: In the spirit of

https://github.com/JuliaLang/julia/blob/dbd3d4cd07e54cc665eae80485fb07394266ad57/stdlib/SparseArrays/src/sparsevector.jl#L1360-L1367

you may want to introduce a `M = length(A)` and reuse it below. As I said, I wonder whether we should move those here and extend them to `SparseVecOrMat`. Hopefully, this doesn't introduce method ambiguities, but can be done in a separate PR.

stdlib/SparseArrays/src/higherorderfns.jl
```diff
@@ -1154,4 +1155,23 @@ map(f::Tf, A::SparseOrStructuredMatrix, Bs::Vararg{SparseOrStructuredMatrix,N})
 map!(f::Tf, C::AbstractSparseMatrixCSC, A::SparseOrStructuredMatrix, Bs::Vararg{SparseOrStructuredMatrix,N}) where {Tf,N} =
     (_checksameshape(C, A, Bs...); _noshapecheck_map!(f, C, _sparsifystructured(A), map(_sparsifystructured, Bs)...))
 
+
+# (13) extrema methods optimized for sparse vectors/matrices.
+function extrema(f, A::SparseVecOrMat)
```
JeffBezanson, line 1160: This doesn't handle `dims` or `dims=:`, so I believe those will still call the slower generic code. We might want to at least handle `dims=:`, and use `invoke` to call the AbstractArray method for general dims.

simeonschaub: I think it's worse because `dims` is a keyword argument, so this will error now if you pass a `dims` argument. This will probably need to be renamed to a hidden function `_extrema` and `extrema` needs to check for a `dims` argument and fall back to the slow code if `dims` is not `:`. An alternative of course would be to handle the `dims` argument here directly.

stdlib/SparseArrays/src/higherorderfns.jl
```diff
@@ -1154,4 +1155,23 @@ map(f::Tf, A::SparseOrStructuredMatrix, Bs::Vararg{SparseOrStructuredMatrix,N})
...
+    vmin, vmax = extrema(f, nonzeros(A))
+    if N != length(A)
+        f0 = f(zero(eltype(A)))
+        vmin = min(f0, vmin)
+        vmax = max(f0, vmax)
+    end
+    vmin, vmax
+end
+
+extrema(A::SparseVecOrMat) = extrema(identity, A)
```
simeonschaub, line 1175: I believe this fallback is already defined in Base.

clement-f: The more specific definition is
`extrema(A::AbstractArray; dims = :) = _extrema_dims(identity, A, dims)`
in base/multidimensional.jl line 1599.
So it seems this is needed.
I benchmarked `extrema` with:
```
julia> S = sprand(10000, 10000, 0.00001);
julia> @btime extrema(S)
  3.328 μs (1 allocation: 32 bytes)
(0.0, 0.9993973695872556)
```
and without this line:
```
julia> @btime extrema(S)
  643.939 ms (1 allocation: 32 bytes)
(0.0, 0.9993973695872556)
```

stdlib/SparseArrays/src/higherorderfns.jl
clement-f: It doesn't throw errors. It only use the slow generic code.
There are only thew cases.
`extrema(f, A::SparseVector, dims=1)`, `extrema(f, A::AbstractSparseMatrix, dims=[1,2])` and `extrema(f, SparseVecOrMat, ::Colon)`
which are equivalent to `extrema(f, A::SparseVecOrMat)`.
We still need to implement `extrema(f, A::SparseVecOrMat, dims=1)` and `extrema(f, A::SparseVecOrMat, dims=2)`.
In my opinion there should be two separate functions because the optimized algorithms are quite different.
I believe it covers every cases.

simeonschaub: The problem is that `dims` is not a positional argument and keyword arts don't participate in dispatch. The correct function signature should probably be `Base._extrema_dims(f, A::SparseVecOrMat, ::Colon)`, so this only gets called for the `dims=:` case (which is also the default).

stdlib/SparseArrays/src/higherorderfns.jl
clement-f: Done.
Concerning `minimum` and `maximum`, they definitely need to be moved and applied on `SparseVecOrMat`. There are no optimizations for sparse matrices on either of these functions for now.

stdlib/SparseArrays/src/higherorderfns.jl
clement-f: I implemented the @JeffBezanson's idea of using `invoke`. Actually reimplementing the functions that handles dims would greatly improve speed but it is a bit tricky to get them consistent with the Base implementation.

simeonschaub:
Why do you define separate methods for `extrema_` for `SparseVector` and `AbstractSparseMatrixCSC`? I personally think just overloading `Base._extrema_dims` would be a much cleaner solution, as I would generally consider using `invoke` to be a bit of an antipattern, but I will leave the final judgement here to @JeffBezanson and @dkarrasch.

stdlib/SparseArrays/test/higherorderfns.jl
```diff
@@ -687,4 +687,23 @@ end
...
+    @test extrema(A) == extrema(B)
+    @test extrema(x) == extrema(y)
+    @test extrema(f, A) == extrema(f, B)
+    @test extrema(f, x) == extrema(f, y)
+    @test extrema(spzeros(n, n)) == (0.0, 0.0)
+    @test extrema(spzeros(n)) == (0.0, 0.0)
+    @test_throws ArgumentError extrema(spzeros(0, 0))
+    @test_throws ArgumentError extrema(spzeros(0))
+    @test extrema(sparse(ones(n, n))) == (1.0, 1.0)
+    @test extrema(sparse(ones(n))) == (1.0, 1.0)
```
simeonschaub, line 706: Can you add tests with `dims=:` and `dims=2` to test that the fallback still works correctly?

dkarrasch:
I think I agree with @simeonschaub. If the generic `extrema` function (with the `dims` keyword) is the single entry point which calls `Base._extrema_dims`, then we can specialize the latter for `SparseVecOrMat` first, and depending on vector vs matrix and the dims argument, call
* (for `dims = :` [or `dims = (1,2)` in the matrix case and `dims = 1` in the vector case]) another "helper function" whose body is the fast, main routine added here,
* (for `dims = 1` and matrices) call `extrema` on column views (which should be not difficult to write and fast)
* (for `dims = 2`) the `invoke` thing.

I wonder if it's acceptable for sparse matrices(!) and `dims=2` to form the transpose and then compute `extrema(...; dims=1)`. It can be a trap if the matrix isn't really sparse due to memory allocation, but it may be much faster for truly sparse matrices... But I guess allocating potentially much memory goes against the very concept of a reduction method.

stdlib/SparseArrays/test/higherorderfns.jl
clement-f: I added some tests. It should cover most cases.

clement-f:
> Why do you define separate methods for `extrema_` for `SparseVector` and `AbstractSparseMatrixCSC`? I personally think just overloading `Base._extrema_dims` would be a much cleaner solution, as I would generally consider using `invoke` to be a bit of an antipattern, but I will leave the final judgement here to @JeffBezanson and @dkarrasch.

It is indeed quite cleaner to extend '_extrema_dims'. In the same idea, I rename 'extrema' so it extends '_extrema_itr'. The new tests are passing and it works well in manual experimentations. However some other tests are failing. It seems to be an ambiguity with '_extrema_dims'. I haven't figure out why yet.

stdlib/SparseArrays/src/higherorderfns.jl
```diff
@@ -1154,4 +1155,58 @@ map(f::Tf, A::SparseOrStructuredMatrix, Bs::Vararg{SparseOrStructuredMatrix,N})
...
+                    B[col_idx] = (B[col_idx][1], fx)
+                end
+            end
+        end
+        return B
+    end
+    invoke(_extrema_dims, Tuple{Any, AbstractArray, Any}, f, A, dims)
+end
+
+_extrema_dims(f, A::SparseVecOrMat, ::Colon) = _extrema_itr(f, A)
```
dkarrasch, line 1210: I'm not sure, but maybe splitting this into two helps with the ambiguity?
```suggestion
_extrema_dims(f, A::AbstractSparseVector, ::Colon) = _extrema_itr(f, A)
_extrema_dims(f, A::AbstractSparseMatrix, ::Colon) = _extrema_itr(f, A)
```
I believe you could generalize the method on line 1174 towards `AbstractSparseVector`... well, or restrict the matrix methods to `SparseMatrixCSC`. I'm not sure what the API for the abstract ones is.

clement-f: Unfortunately it doesn't work.
The method on line 1174 can't be generalized because nnz is not defined for `AbstractSparseVector`. Almost every function in `Sparsearrays` is defined for `SparseVector` anf `AbstractSparseMatric`, actually.

dkarrasch: No problem, that's fine. But does splitting the methods into `SparseVector` and `AbstractSparseMatrix` help? I think there might be an ambiguity because the method here is more abstract in the second argument but more concrete in the third (but I forgot compared to which other method).

dkarrasch:
This is certainly "ready for review", if not for merging. 😉

clement-f:
It seems that the freebsd test is failing in test/stress.jl. I can't see how it is related to this pull request.

dkarrasch:
Thank you @clement-f for this great work.

PR #37430, opened 2020-09-06
[LateGCLowering] Look through Phi node

vchuravy:
@yhls looked at this a while back. I am running on a LLVM master from ~July and the gep for an alloca got sunk, and then duplicated, which materializes as a Phi-Node. I don't have a good reproducer and I haven't noticed this on LLVM11rc2, but it probably makes sense to be a bit more defensive here. At least we should `assert` that `SRet_gc` is valid after the `dyn_cast`.


vtjnash (approved):
It’d be nice if LLVM didn’t complicate this accidentally, but I didn’t write this code very defensively, and should probably fix that at some point.

PR #37437, opened 2020-09-06
optimize the REPL history reader

KristofferC:
My history file from the beginning of the year took ~0.15 seconds to read. It now takes ~0.02 seconds. So now I can chill for 7 years until I need to optimize this function further.

### DIFF
--- a/base/strings/substring.jl
+++ b/base/strings/substring.jl
@@ -177,6 +177,10 @@ end
 string(a::String)            = String(a)
 string(a::SubString{String}) = String(a)
 
+function Symbol(s::SubString{String})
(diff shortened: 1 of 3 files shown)


base/strings/substring.jl
```diff
@@ -177,6 +177,10 @@ end
 string(a::String)            = String(a)
 string(a::SubString{String}) = String(a)
 
+function Symbol(s::SubString{String})
+    return ccall(:jl_symbol_n, Ref{Symbol}, (Ptr{UInt8}, Int), s, sizeof(s))
```
KristofferC, line 181: Is this ok? @JeffBezanson

JeffBezanson: Yes looks good to me.

PR #37443, opened 2020-09-07
Fix `getfield_tfunc` for constant `TypeName`

yuyichao:
The field should be either `Symbol` or `Int`. Ref #37423

Unlike #37423, in additional to worse type info in inference,
the missing type check here can actually cause type inference error
due to errors in user code.

### DIFF
--- a/base/compiler/tfuncs.jl
+++ b/base/compiler/tfuncs.jl
@@ -745,6 +745,9 @@ function getfield_tfunc(@nospecialize(s00), @nospecialize(name))
         end
         if isa(name, Const)
             nv = name.val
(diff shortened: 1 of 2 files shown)


martinholters:
In #37293, I've just moved the check
```julia
            if !(isa(nv,Symbol) || isa(nv,Int))
                return Bottom
            end
```
further up which subsumes this. But #37293 is more than just a bugfix, so we probably want to merge and backport this, even if #37293 can then safely revert it.

yuyichao:
Updated the fix to be similar to https://github.com/JuliaLang/julia/pull/37293

martinholters (approved):

PR #37461, opened 2020-09-08
add init argument to count

simeonschaub:
Inspired by [this](https://stackoverflow.com/a/59125248) StackOverflow answer. Unfortunately, `count!` also already takes an `init` argument, which has a different meaning, so this will be the same situation as we currently have with `sum`/`sum!`, see #36266.


base/reduce.jl
```diff
@@ -954,30 +953,30 @@ julia> count([true, false, true, true])
...
+    n = init
     for x in itr
         n += pred(x)::Bool
     end
     return n
 end
 
-function count(::typeof(identity), x::Array{Bool})
-    n = 0
+function _simple_count(::typeof(identity), x::Array{Bool}, init::T=0) where {T}
```
simeonschaub, line 968: This is actually a bugfix (although it doesn't occur in 1.5), because `count(::Array{Bool})` currently errors when passed a `dims` argument.

nalimilan:
Interesting idea. Though I wonder whether the speedup obtained by using a smaller integer type isn't just due to a compiler bug in this particular case: see https://github.com/JuliaLang/julia/issues/37473. I don't see any speedup when counting *non-missing* values or the number of zeros:
```julia
julia> function count_nonmissing(x, T)
           c = zero(T)
           for i in 1:length(x)
               c += @inbounds x[i] !== missing
           end
           return Int(c)
       end
count_nonmissing (generic function with 1 method)

julia> x = rand([missing, rand(Int, 100)...], 1_000_000);

julia> @btime count_nonmissing(x, Int);
  25.500 ms (999485 allocations: 15.25 MiB)

julia> @btime count_nonmissing(x, Int16);
  27.453 ms (983978 allocations: 15.01 MiB)

julia> x = rand(1_000_000);

julia> @btime count_zero(x, Int);
  21.660 ms (0 allocations: 0 bytes)

julia> @btime count_zero(x, Int16);
  19.200 ms (0 allocations: 0 bytes)
```

simeonschaub:
Hmm, interesting! I would still argue that this is worth it for the consistency alone. It still seems useful to be able to control the accumulation type, perhaps someone wants to count using finite fields, or explicitely needs 64 bits of accuracy on 32-bit platforms.

simeonschaub:
I don't know who would be best to review this. @tkf, would you be willing to take a quick look at this, since I think you worked a lot on our reduce functions lately?

base/reduce.jl
tkf, line 968: I think we need to restrict `T` to `Integer` if we keep the return value type guarantee in the docstring signature (`-> Integer`):

```suggestion
function _simple_count(::typeof(identity), x::Array{Bool}, init::T=0) where {T <: Integer}
```

tkf: (Removing `-> Integer` from the docstring and calling this PR a minor change is, technically, another option.)

base/bitarray.jl
```diff
@@ -1386,15 +1386,15 @@ circshift!(B::BitVector, i::Integer) = circshift!(B, B, i)
 
 ## count & find ##
 
-function bitcount(Bc::Vector{UInt64})
-    n = 0
+function bitcount(Bc::Vector{UInt64}; init::T=0) where {T}
```
tkf, line 1389: ditto

```suggestion
function bitcount(Bc::Vector{UInt64}; init::T=0) where {T <: Integer}
```

tkf:
For now, I guess you can just use `sum` instead of `count`?

TBH, after reading https://github.com/JuliaLang/julia/pull/35947#issuecomment-634425865, I've been wondering how much we should be supporting `count`.  Should it even be a target for deprecation?  Likely not, but it _could_ be, depending on what goes in Julia 2.0.

base/reduce.jl
simeonschaub: It feels a bit arbitrary to restrict `T` to only `Integer` types, since there's really nothing preventing this from working with any other accumulation types. The docstring wasn't even that accurate in the first place, since `count` always returned `Int`, so I don't know why it's documented as `Integer`. I always thought "minor change" meant it was slightly breaking for some edge cases, which I don't think applies here, since the previous behavior should be completely maintained, just extended. But you are right, the docstring still needs to include the `dims` argument and have a compat annotation.

simeonschaub:
I guess currently `count` still has the advantage of being optimized for `BitArray`s and `Array{Bool}`, but I don't see why we couldn't apply this optimization to the more general `mapreduce` case. I guess `count` still has some merit over `sum` purely semantically, because "counting" logical values feels more natural than "summing" logical values, but I agree that it otherwise really doesn't add much over `sum`. A hard deprecation would probably still be a bit out of scope for 1.x, since I'd imagined it's used quite a lot.

base/reduce.jl
tkf: > there's really nothing preventing this from working with any other accumulation types

If you need other accumulation types, I think you should use `sum`.

> The docstring wasn't even that accurate in the first place, since `count` always returned `Int`

`Int <: Integer` so the docstring in Julia <= 1.5 is OK in the sense `c = count(...); c::Integer` never throws if `count` returns.

But I do agree that docstring is not accurate in the sense that what `->` means is not defined anywhere (i.e., we don't have a specification of the signature part of the docstring).

> I always thought "minor change" meant it was slightly breaking for some edge cases

"Minor change" means to violate [SemVer](https://semver.org) guarantee (= "technically breaking"). See also https://julialang.org/blog/2019/08/release-process/#minor_releases

> since the previous behavior should be completely maintained

Breaking or not is decided with respect to the public API _specification_ (in `julia`, it's documentation). We then check if the behavior follows the API; if not, it's a bug.

So, increasing the return value type bound (`Integer` to `Any`) in the documentation is breaking.

tkf:
> being optimized for `BitArray`s and `Array{Bool}`

Isn't it better to use these methods from `sum` as well?

> `count` still has some merit over `sum`

I feel this is, in principle, minor than assigning different semantics to `count` as discussed in #35947. Having said that, now that we are in the post-1.0 era, I still don't think such minor renaming should happen, even for 2.0.

base/reduce.jl
simeonschaub: Since it won't break any code in practice, I don't think the breakage here would be any kind of problem, but I'm ok with restricting it to `Integer` as well. I guess that restriction should be in the definition of `count` though, not the internal `_simple_count`.

martinholters: IIUC, when not specifying `init`, the result type is `Integer`, in line with the old docs. I.e. old signatures:
```
  count(p, itr) -> Integer
  count(itr) -> Integer
```
New signatures:
```
  count(p, itr) -> Integer
  count(itr) -> Integer
  count(p, itr; init::T) -> T
  count(itr; init::T) -> T
```
Where the latter can be condensed to the equivalent
```
  count(p, itr; init::T=0) -> T
  count(itr; init::T=0) -> T
```
So this seems to be purely an extension; I don't see anything technically breaking here. (For simplicity, this assumes `+(::T,::T)::T`, but that holds for `T==Int`.)

simeonschaub: So should I leave this as is, or does it make more sense to constrain this to `Integer`?

simeonschaub: @tkf I'm fine with either option, should I restrict `init` to `Integer or not?

StefanKarpinski:
Bump?

simeonschaub:
@tkf Are you fine with this as-is, or do you still think, `init` should be restricted to `Integer`? Other than this, I think this should be good to go.

test/reduce.jl
```diff
@@ -520,6 +520,12 @@ struct NonFunctionIsZero end
 @test count(NonFunctionIsZero(), [0]) == 1
 @test count(NonFunctionIsZero(), [1]) == 0
 
+@test count(Iterators.repeated(true, 3), init=0x00) isa UInt8
+@test count(!=(2), Iterators.take(1:7, 3), init=Int32(0)) isa Int32
+@test count(identity, [true, false], init=Int8(0)) isa Int8
+@test count(!, [true false; false true], dims=:, init=Int16(0)) isa Int16
+@test count(identity, [true false; false true], dims=2, init=UInt(0)) isa Matrix{UInt}
```
rfourquet, line 527: Would be good to add tests with nonzero `init`, and also which actually test the return value (vs. only the type).

base/reduce.jl
```diff
@@ -938,11 +938,10 @@ end
...
 
 """
-    count(p, itr) -> Integer
-    count(itr) -> Integer
+    count([f=identity,] itr; init=0) -> Integer
 
-Count the number of elements in `itr` for which predicate `p` returns `true`.
-If `p` is omitted, counts the number of `true` elements in `itr` (which
+Count the number of elements in `itr` for which the function `f` returns `true`.
+If `f` is omitted, counts the number of `true` elements in `itr` (which
```
rfourquet, line 944: The `init` keyword should be documented :) Also, what was wrong with "predicate"?

simeonschaub: Honestly, I don't quite remember, why I changed this... :laughing: I think I changed the name from `p` to `f` to be more consistent with oder `mapreduce`-like functions, so calling it "function" ended up making more sense? `function` does seem clearer to me here, because that's the terminology we use throughout the manual, and it seems a bit unnecessarily confusing to call it something else here. I don't feel very strongly about this though, so I can also leave it as-is.

simeonschaub:
Bump.

simeonschaub:
@rfourquet Would you mind taking another look at this before the branching for 1.6 happens, because `count(::Array{Bool}, dims=...)` is still broken currently?

rfourquet:
> Would you mind taking another look

Oh sorry, I had missed your previous review request. I will have a look later today.

base/reduce.jl
```diff
@@ -954,30 +956,30 @@ julia> count([true, false, true, true])
...
-    n = 0
+function _simple_count(::typeof(identity), x::Array{Bool}, init::T=0) where {T}
+    n::T = init
     chunks = length(x) ÷ sizeof(UInt)
     mask = 0x0101010101010101 % UInt
     GC.@preserve x begin
         ptr = Ptr{UInt}(pointer(x))
         for i in 1:chunks
-            n += count_ones(unsafe_load(ptr, i) & mask)
+            n += count_ones(unsafe_load(ptr, i) & mask) % T
```
rfourquet, line 978: I think I would rather write 
```suggestion
            n = (n + count_ones(unsafe_load(ptr, i) & mask)) % T
```
for the case where `T` is more heavy to construct, e.g. `BigInt`. Conversely, if `T` is smaller than `Int`, it probably doesn't matter to promote the addition to `Int` before the `%`-reduction. Similarly for `BitArray` and below.
But this is just a suggestion, which is maybe too defensive.

rfourquet (commented):
LGTM! Just the docstring needs to document `init`.

base/reduce.jl
rfourquet: I don't have a strong opinion here, but I think not forcing `init isa Integer` is fine.

simeonschaub:
Thank you very much for the thorough review! Hope I addressed it all.

base/reduce.jl
rfourquet: Oh, one last small bit:
```suggestion
If `f` is omitted, count the number of `true` elements in `itr` (which
```

base/reduce.jl
```diff
@@ -952,32 +955,35 @@ julia> count(i->(4<=i<=6), [2,3,4,5,6])
 
 julia> count([true, false, true, true])
 3
+
+julia> count(>(3), 1:7, init=0x00)
```
rfourquet, line 959: Maybe `init=0x03` or something to show how the behavior differs besides changing the return type.

base/reduce.jl
```diff
@@ -952,32 +955,35 @@ julia> count(i->(4<=i<=6), [2,3,4,5,6])
...
     GC.@preserve x begin
         ptr = Ptr{UInt}(pointer(x))
         for i in 1:chunks
-            n += count_ones(unsafe_load(ptr, i) & mask)
+            n = (n + count_ones(unsafe_load(ptr, i) & mask)) % T
         end
     end
     for i in sizeof(UInt)*chunks+1:length(x)
-        n += x[i]
+        n += x[i] % T
```
rfourquet, line 986: As you changed it above, maybe 
```suggestion
        n = (n + x[i]) % T
```
Or simply 
```suggestion
        n += x[i]
```
like in the general case (with the reasonable assumption that `typeof(n) == typeof(n+x[i]::Bool)`

base/bitarray.jl
```diff
@@ -1386,15 +1386,15 @@ circshift!(B::BitVector, i::Integer) = circshift!(B, B, i)
...
+    n::T = init
     @inbounds for i = 1:length(Bc)
-        n += count_ones(Bc[i])
+        n += count_ones(Bc[i]) % T
```
rfourquet, line 1392:
```suggestion
        n = (n + count_ones(Bc[i])) % T
```

rfourquet (commented):
Just a couple more nitpicks :)

base/reduce.jl
```diff
@@ -952,32 +955,35 @@ julia> count(i->(4<=i<=6), [2,3,4,5,6])
 
 julia> count([true, false, true, true])
 3
+
+julia> count(>(3), 1:7, init=0x00)
+0x04
```
simeonschaub, line 960:
```suggestion
julia> count(>(3), 1:7, init=0x03)
0x07
```

rfourquet (approved):

simeonschaub:
Is this good to go?

rfourquet:
I will merge in a couple of days if no objection :)

PR #37463, opened 2020-09-08
Note that `&&` and `||` are `and` and `or`.

henriquebecker91:
Suggested by this post in Discourse: https://discourse.julialang.org/t/and-or-bitwise-or-shortcircuit-what-we-get-searching-docs/46263

Only adds the phrase "Some languages (like Python) refer to them as `and` (`&&`) and `or` (`||`)." to the documentation.


kimikage:
I think the important thing is not what they are called, but what they are.
Some novices may know that Python's `and` is a logical operator, not bitwise, but they won't understand how it does short-circuit evaluation. Advanced programmers will know that the definition is never universal across various languages.

musm (approved):

PR #37467, opened 2020-09-08
Improve performance of `==(::AbstractString, ::AbstractString)`

omus:
Originally I was aiming to optimized `==(::SubString, ::String)` but that was already addressed as part of https://github.com/JuliaLang/julia/pull/35973. Most of the performance increase is probably due to removing allocations.

Current master (594ac899ff)

```julia
julia> @btime invoke(==, Tuple{AbstractString, AbstractString}, $"ABC", $"ABC")
  3.905 ns (0 allocations: 0 bytes)
true

julia> @btime invoke(==, Tuple{AbstractString, AbstractString}, $(SubString("ABC")), $"ABCD")
  23.621 ns (1 allocation: 64 bytes)
false

julia> @btime invoke(==, Tuple{AbstractString, AbstractString}, $(SubString("ABC")), $"ABC")
  24.395 ns (1 allocation: 64 bytes)
true

julia> @btime invoke(==, Tuple{AbstractString, AbstractString}, $(SubString("ABC")), $"DEF")
  14.759 ns (1 allocation: 64 bytes)
false

julia> str = "A" ^ 1000;

julia> @btime invoke(==, Tuple{AbstractString, AbstractString}, $(SubString(str)), $str)
  4.141 μs (1 allocation: 64 bytes)
true
```

Current PR:

```julia
julia> @btime invoke(==, Tuple{AbstractString, AbstractString}, $"ABC", $"ABC")
  4.364 ns (0 allocations: 0 bytes)
true

julia> @btime invoke(==, Tuple{AbstractString, AbstractString}, $(SubString("ABC")), $"ABCD")
  5.533 ns (0 allocations: 0 bytes)
false

julia> @btime invoke(==, Tuple{AbstractString, AbstractString}, $(SubString("ABC")), $"ABC")
  13.041 ns (0 allocations: 0 bytes)
true

julia> @btime invoke(==, Tuple{AbstractString, AbstractString}, $(SubString("ABC")), $"DEF")
  7.283 ns (0 allocations: 0 bytes)
false

julia> str = "A" ^ 1000;

julia> @btime invoke(==, Tuple{AbstractString, AbstractString}, $(SubString(str)), $str)
  2.524 μs (0 allocations: 0 bytes)
true
```

### DIFF
--- a/base/strings/basic.jl
+++ b/base/strings/basic.jl
@@ -323,10 +323,10 @@ false
 """
 function ==(a::AbstractString, b::AbstractString)
     a === b && return true
-    lastindex(a) != lastindex(b) && return false
     for (c::AbstractChar, d::AbstractChar) in zip(a, b)
         c != d && return false
     end
+    length(a) != length(b) && return false
     return true
 end
 
--- a/base/strings/basic.jl
+++ b/base/strings/basic.jl
@@ -321,7 +321,14 @@ julia> "abc" == "αβγ"
 false
 ```
 """
-==(a::AbstractString, b::AbstractString) = cmp(a, b) == 0
+function ==(a::AbstractString, b::AbstractString)
+    a === b && return true
+    ncodeunits(a) != ncodeunits(b) && return false
+    for (c::AbstractChar, d::AbstractChar) in zip(a, b)
+        c != d && return false
+    end
+    return true
+end
 
 """
     isless(a::AbstractString, b::AbstractString) -> Bool
--- a/base/strings/basic.jl
+++ b/base/strings/basic.jl
@@ -323,7 +323,7 @@ false
 """
 function ==(a::AbstractString, b::AbstractString)
     a === b && return true
-    ncodeunits(a) != ncodeunits(b) && return false
+    lastindex(a) != lastindex(b) && return false
     for (c::AbstractChar, d::AbstractChar) in zip(a, b)
         c != d && return false
     end
--- a/test/strings/basic.jl
+++ b/test/strings/basic.jl
@@ -680,7 +680,6 @@ end
 Base.iterate(x::CharStr) = iterate(x.chars)
 Base.iterate(x::CharStr, i::Int) = iterate(x.chars, i)
 Base.lastindex(x::CharStr) = lastindex(x.chars)
-Base.ncodeunits(x::CharStr) = sum(ncodeunits, x.chars)
 
 @testset "cmp without UTF-8 indexing" begin
     # Simple case, with just ANSI Latin 1 characters
--- a/test/strings/basic.jl
+++ b/test/strings/basic.jl
@@ -680,6 +680,8 @@ end
 Base.iterate(x::CharStr) = iterate(x.chars)
 Base.iterate(x::CharStr, i::Int) = iterate(x.chars, i)
 Base.lastindex(x::CharStr) = lastindex(x.chars)
+Base.ncodeunits(x::CharStr) = sum(ncodeunits, x.chars)
+
 @testset "cmp without UTF-8 indexing" begin
     # Simple case, with just ANSI Latin 1 characters
     @test "áB" != CharStr("áá") # returns false with bug


omus:
Added benchmarks in: https://github.com/JuliaCI/BaseBenchmarks.jl/pull/260

JeffBezanson:
I'm not sure all strings are required to implement `ncodeunits`. E.g. for `CharStr` it doesn't fully make sense. The numbers are also incomparable if the strings use different encodings.

omus:
> I'm not sure all strings are required to implement `ncodeunits`

I also wasn't quite sure of that myself. As `zip` requires `length` I should be able to use that safely

JeffBezanson:
> As `zip` requires `length`

It doesn't, but strings do.

omus:
> It doesn't, but strings do.

I misread the code. Need to be a little less hasty. I can switch it to `length`

omus:
As `length` checks that each character `isvalid` we lose most of the performance benefit. I'm looking into using `lastindex` instead.

omus:
Revised benchmarks using `lastindex` instead of `ncodeunits`:

Current master (594ac89)

```julia
julia> str = String('A':'Z') ^ 100;

julia> @btime invoke(==, Tuple{AbstractString, AbstractString}, $str, $str)
  93.236 ns (0 allocations: 0 bytes)
true

julia> @btime invoke(==, Tuple{AbstractString, AbstractString}, $(SubString(str)), $str)
  11.885 μs (1 allocation: 64 bytes)
true

julia> @btime invoke(==, Tuple{AbstractString, AbstractString}, $(SubString(str)), $(str * 'Z'))
  9.803 μs (1 allocation: 64 bytes)
false

julia> @btime invoke(==, Tuple{AbstractString, AbstractString}, $(SubString(str)), $(reverse(str)))
  15.075 ns (1 allocation: 64 bytes)
false
```

Current PR:

```julia
julia> str = String('A':'Z') ^ 100;

julia> @btime invoke(==, Tuple{AbstractString, AbstractString}, $str, $str)
  4.032 ns (0 allocations: 0 bytes)
true

julia> @btime invoke(==, Tuple{AbstractString, AbstractString}, $(SubString(str)), $str)
  8.229 μs (0 allocations: 0 bytes)
true

julia> @btime invoke(==, Tuple{AbstractString, AbstractString}, $(SubString(str)), $(str * 'Z'))
  12.026 ns (0 allocations: 0 bytes)
false

julia> @btime invoke(==, Tuple{AbstractString, AbstractString}, $(SubString(str)), $(reverse(str)))
  15.002 ns (0 allocations: 0 bytes)
false
```

JeffBezanson:
Unfortunately I think `lastindex` is also encoding-specific.

omus:
> Unfortunately I think `lastindex` is also encoding-specific.

Ah yes. I see where was going wrong.

```julia
julia> str = String('A':'Z') ^ 100;

julia> @btime invoke(==, Tuple{AbstractString, AbstractString}, $str, $str)
  4.030 ns (0 allocations: 0 bytes)
true

julia> @btime invoke(==, Tuple{AbstractString, AbstractString}, $(SubString(str)), $str)
  8.518 μs (0 allocations: 0 bytes)
true

julia> @btime invoke(==, Tuple{AbstractString, AbstractString}, $(SubString(str)), $(str * 'Z'))
  8.514 μs (0 allocations: 0 bytes)
false

julia> @btime invoke(==, Tuple{AbstractString, AbstractString}, $(SubString(str)), $(reverse(str)))
  7.535 ns (0 allocations: 0 bytes)
false
```

omus:
I experimented with using iteration directly to avoid using `Iterators.Stateful` and calling `length`

```julia
function ==(a::AbstractString, b::AbstractString)
    a === b && return true
    next_a = iterate(a)
    next_b = iterate(b)
    while next_a !== nothing && next_b !== nothing
        c, state_a = next_a
        d, state_b = next_b
        c != d && return false
        next_a = iterate(a, state_a)
        next_b = iterate(b, state_b)
    end
    return next_a === nothing && next_b === nothing
end
```

The results have similar performance to what is currently implemented:

```julia
julia> str = String('A':'Z') ^ 100;

julia> @btime invoke(==, Tuple{AbstractString, AbstractString}, $str, $str)
  3.784 ns (0 allocations: 0 bytes)
true

julia> @btime invoke(==, Tuple{AbstractString, AbstractString}, $(SubString(str)), $str)
  8.497 μs (0 allocations: 0 bytes)
true

julia> @btime invoke(==, Tuple{AbstractString, AbstractString}, $(SubString(str)), $(str * 'Z'))
  8.494 μs (0 allocations: 0 bytes)
false

julia> @btime invoke(==, Tuple{AbstractString, AbstractString}, $(SubString(str)), $(reverse(str)))
  8.313 ns (0 allocations: 0 bytes)
false
```
As this code is less readable I'll stay with the current implementation.

omus:
At this point I don't have any more ideas for ways to tweak this so we just need to decide if we want to go ahead with this change.

base/strings/basic.jl
```diff
@@ -321,7 +321,14 @@ julia> "abc" == "αβγ"
 false
 ```
 """
-==(a::AbstractString, b::AbstractString) = cmp(a, b) == 0
+function ==(a::AbstractString, b::AbstractString)
+    a === b && return true
+    for (c::AbstractChar, d::AbstractChar) in zip(a, b)
+        c != d && return false
+    end
+    length(a) != length(b) && return false
```
Keno, line 329: These are quite likely O(n) again. We should be able to get away with only one O(N) iteration.

KristofferC: @omus, maybe change back to the iterator based approach to fix this?

jakobnissen:
Is this issue fixed already with  #47125?

jakobnissen:
Closing this as fixed in #47125 which uses a similar iterator based approach

PR #37471, opened 2020-09-08
add  `and`  and `or` docs

### LINKED ISSUE #37469: add docstrings for `and` and `or`
https://discourse.julialang.org/t/and-or-bitwise-or-shortcircuit-what-we-get-searching-docs/46263/25

Since people coming from some languages expect `and` and `or` to be keywords, we could add docstrings for these as keywords so that `?and` and `?or` in the REPL give help strings indicating that the correct short-circuit boolean operators in Julia are `&&` and `||`.

Moelf:
fix  #37469

### DIFF
--- a/base/docs/basedocs.jl
+++ b/base/docs/basedocs.jl
@@ -942,7 +942,7 @@ kw"||"
 """
     `or` is not used in Julia
 
-Use `|| for short-circuiting boolean OR.
+Use `||` for short-circuiting boolean OR.
 """
 kw"or"
 
--- a/base/docs/basedocs.jl
+++ b/base/docs/basedocs.jl
@@ -926,11 +926,25 @@ Short-circuiting boolean AND.
 kw"&&"
 
 """
+    `and` is not used in Julia
+
+Use `&&` for short-circuiting boolean AND.
+"""
+kw"and"
+
+"""
     x || y
 
 Short-circuiting boolean OR.
 """
 kw"||"
+
+"""
+    `or` is not used in Julia
+
+Use `|| for short-circuiting boolean OR.
+"""
+kw"or"
 
 """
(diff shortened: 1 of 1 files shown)


base/docs/basedocs.jl
```diff
@@ -925,13 +925,27 @@ Short-circuiting boolean AND.
...
     x || y
 
 Short-circuiting boolean OR.
 """
 kw"||"
 
+"""
+    `or` is not used in Julia
+
+Use `|| for short-circuiting boolean OR.
```
cormullion, line 945: is there a backtick missing here?

Moelf: can't believe I just did that.... thx for catching it

simeonschaub:
One potential problem with this approach is that if a package defined the functions `and` or `or` with a docstring, those docstrings won't be visible from the REPL anymore. The REPL help mode can probably be changed so that it shows both of them though.

JeffBezanson:
From triage: we should not use the existing keyword doc mechanism for this. At the very least, the REPL would need to display both the keyword docstring and any found docstrings for the identifier. Ideally, we could have some new doc category like a "term", for attaching text to words regardless of where they come from.

StefanKarpinski:
The problem here is that this prevents defining something named `and` or `or` with a doc string and then seeing that documentation. If showing the doc string for a keyword were made to keep searching after showing the keyword documentation, that would avoid this problem. However, it's a little weird to call that functionality `kw"..."` since it's now really just a way of saying "if someone does `?text` show the following message". So perhaps doing that and also having a more semantically appropriate name for `kw"..."` would be good.

StefanKarpinski:
Another thought... doing something like this could be handy:
```jl
"""
Doc string.
"""
r"help"i
```
This would indicate that the given doc string should be shown when the help query matches the regex.

PR #37472, opened 2020-09-08
Fix memory leak in umfpack_numeric!

### LINKED ISSUE #37118: lu! leaks memory
Not sure if upstream bug, but the new lu! function seems to leak memory.

The following permanently consumes ~8GiB on my machine (until I decide to exit Julia):
```Julia
using LinearAlgebra, SparseArrays
function f()
    S = I + sprand(10000,10000,0.00001)
    F = lu(S)
    lu!(F,S)
end

for i = 1:10000
    f()
end
```

cfauchereau:
Fix #37118
If umfpack_numeric! is called on a UmfpackLU struct that has already been factorized, its pointer "numeric" is overwritten without freeing the pointed Numeric object.

This simple fix does:

- Check for an existing Numeric object before writing to UmfpackLU.numeric.
- If it exists, free the object.

### DIFF
--- a/stdlib/SuiteSparse/src/umfpack.jl
+++ b/stdlib/SuiteSparse/src/umfpack.jl
@@ -333,6 +333,7 @@ for itype in UmfpackIndexTypes
             if status != UMFPACK_WARNING_singular_matrix
                 umferror(status)
             end
+            U.numeric != C_NULL && umfpack_free_numeric(U)
             U.numeric = tmp[1]
             return U
         end
@@ -349,6 +350,7 @@ for itype in UmfpackIndexTypes
             if status != UMFPACK_WARNING_singular_matrix
                 umferror(status)
             end
+            U.numeric != C_NULL && umfpack_free_numeric(U)
             U.numeric = tmp[1]
             return U
         end


dkarrasch (commented):
Just a quick style comment: indention is slightly off (by two characters?). These one-line branches are often written as
```julia
U.numeric != C_NULL && umfpack_free_numeric(U)
```

clement-f:
> Just a quick style comment: indention is slightly off (by two characters?). These one-line branches are often written as
> 
> ```julia
> U.numeric != C_NULL && umfpack_free_numeric(U)
> ```

I made the change. It was a tab, my mistake.

dkarrasch:
Just to make sure: did you check that the performance benefits (of `lu!` over `lu`) are still preserved?

clement-f:
> Just to make sure: did you check that the performance benefits (of `lu!` over `lu`) are still preserved?

lu! is still faster than lu. The benefit comes from the reuse of the Symbolic object, the Numeric object is recreated each time. `reuse_numeric` is a misleading variable name.

andreasnoack (approved):
Thanks. I wonder if we ever manage to wrap these functions correctly.

andreasnoack:
Do we just ignore FreeBSD failures for the time being?

PR #37479, opened 2020-09-08
Add CPUID submodule

### LINKED ISSUE #36367: A CPUID library in Base or stdlib
As @giordano and I are working on microarchitecture-specific tarballs in BinaryBuilder, we are currently importing a modified version of [`CpuId.jl`](https://github.com/m-j-w/CpuId.jl), however if `Base` is going to be selecting artifacts based on the current CPU type, it makes sense that something with similar capabilities to this `CpuId` package would become a stdlib at the least, if not a part of `Base`.

We already have some CPUID code in `src/processor*.cpp`; should we export `jl_test_cpu_feature()` and simply maintain a Julia mapping of the flags?  The way I see it, we have three options:

* Export the C function, create a simple wrapper stdlib that has various flag definitions and invokes a `ccall()` to probe CPUID bits.
* Leave the C code alone, duplicate the functionality in pure Julia (similar to [`CpuId.jl`](https://github.com/m-j-w/CpuId.jl)), but have it shipped by default with Julia.
* Force all packages that want to use microarchitecture-specific artifacts to do the artifact selection themselves (this will only work for lazy artifacts; no `Pkg.add()`-time artifacts allowed)

I am in favor of 1 or 2, and between the two of them I'm in favor of (2) because it is most maintainable.

giordano:
This is a small internal module which defines the API to query the Instruction
Set Architecture (ISA) of the current CPU.  This can be used for example to
select an artifact in a JLL package which is compatible with the current CPU.

This is related to #37320, and might change to accommodate some needs of that PR.

Fix #36367.  CC: @staticfloat.

### DIFF
--- a/base/Makefile
+++ b/base/Makefile
@@ -12,7 +12,7 @@ else
 CPP_STDOUT := $(CPP) -E
 endif
 
-all: $(addprefix $(BUILDDIR)/,pcre_h.jl errno_h.jl build_h.jl.phony file_constants.jl uv_constants.jl version_git.jl.phony)
+all: $(addprefix $(BUILDDIR)/,pcre_h.jl errno_h.jl build_h.jl.phony features_h.jl file_constants.jl uv_constants.jl version_git.jl.phony)
 
 PCRE_CONST := 0x[0-9a-fA-F]+|[0-9]+|\([\-0-9]+\)
 ifeq ($(USE_SYSTEM_PCRE), 1)
@@ -20,6 +20,18 @@ ifeq ($(USE_SYSTEM_PCRE), 1)
 else
   PCRE_INCL_PATH := $(build_includedir)/pcre2.h
 endif
+
+define parse_features
+@echo "# $(2) features" >> $@
+@$(call PRINT_PERL, cat ../src/features_$(1).h | perl -lne 'print "const JL_$(2)_$$1 = UInt32($$2)" if /^\s*JL_FEATURE_DEF(?:_NAME)?\(\s*(\w+)\s*,\s*([^,]+)\s*,.*\)\s*(?:\/\/.*)?$$/' >> $@)
+@echo >> $@
+endef
+
+$(BUILDDIR)/features_h.jl: ../src/features_x86.h ../src/features_aarch32.h ../src/features_aarch64.h
+	@-rm -f $@
+	@$(call parse_features,x86,X86)
+	@$(call parse_features,aarch32,AArch32)
+	@$(call parse_features,aarch64,AArch64)
 
 $(BUILDDIR)/pcre_h.jl: $(PCRE_INCL_PATH)
 	@$(call PRINT_PERL, $(CPP) -D PCRE2_CODE_UNIT_WIDTH=8 -dM $< | perl -nle '/^\s*#define\s+PCRE2_(\w*)\s*\(?($(PCRE_CONST))\)?u?\s*$$/ and print index($$1, "ERROR_") == 0 ? "const $$1 = Cint($$2)" : "const $$1 = UInt32($$2)"' | LC_ALL=C sort > $@)
@@ -236,6 +248,7 @@ clean:
 	-rm -f $(BUILDDIR)/errno_h.jl
 	-rm -f $(BUILDDIR)/build_h.jl
 	-rm -f $(BUILDDIR)/build_h.jl.phony
+	-rm -f $(BUILDDIR)/features_h.jl
 	-rm -f $(BUILDDIR)/uv_constants.jl
 	-rm -f $(BUILDDIR)/file_constants.jl
 	-rm -f $(BUILDDIR)/version_git.jl
--- a/base/binaryplatforms.jl
+++ b/base/binaryplatforms.jl
@@ -7,6 +7,74 @@ export AbstractPlatform, Platform, HostPlatform, platform_dlext, tags, arch, os,
        select_platform, platforms_match, platform_name
 import .Libc.Libdl
 
+### Submodule with information about CPU features
+module CPUID
+
+export cpu_isa
+
+"""
+    ISA(features::Set{UInt32})
+
+A structure which represents the Instruction Set Architecture (ISA) of a
+computer.  It holds the `Set` of features of the CPU.
+
+The numerical values of the features are automatically generated from the C
+source code of Julia and stored in the `features_h.jl` Julia file.
+"""
+struct ISA
+    features::Set{UInt32}
+end
+
+Base.:<=(a::ISA, b::ISA) = a.features <= b.features
+Base.:<(a::ISA,  b::ISA) = a.features <  b.features
+Base.isless(a::ISA,  b::ISA) = a < b
+
+include("features_h.jl")
+
+# Keep in sync with `arch_march_isa_mapping`.
+const ISAs_by_family = Dict(
+    "x86_64" => (
+        # Source: https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html.
+        # Implicit in all sets, because always required: mmx, sse, sse2
+        "x86_64" => ISA(Set{UInt32}()),
+        "core2" => ISA(Set((JL_X86_sse3, JL_X86_ssse3))),
+        "nehalem" => ISA(Set((JL_X86_sse3, JL_X86_ssse3, JL_X86_sse41, JL_X86_sse42, JL_X86_popcnt))),
+        "sandybridge" => ISA(Set((JL_X86_sse3, JL_X86_ssse3, JL_X86_sse41, JL_X86_sse42, JL_X86_popcnt, JL_X86_avx, JL_X86_aes, JL_X86_pclmul))),
+        "haswell" => ISA(Set((JL_X86_movbe, JL_X86_sse3, JL_X86_ssse3, JL_X86_sse41, JL_X86_sse42, JL_X86_popcnt, JL_X86_avx, JL_X86_avx2, JL_X86_aes, JL_X86_pclmul, JL_X86_fsgsbase, JL_X86_rdrnd, JL_X86_fma, JL_X86_bmi, JL_X86_bmi2, JL_X86_f16c))),
+        "skylake" => ISA(Set((JL_X86_movbe, JL_X86_sse3, JL_X86_ssse3, JL_X86_sse41, JL_X86_sse42, JL_X86_popcnt, JL_X86_avx, JL_X86_avx2, JL_X86_aes, JL_X86_pclmul, JL_X86_fsgsbase, JL_X86_rdrnd, JL_X86_fma, JL_X86_bmi, JL_X86_bmi2, JL_X86_f16c, JL_X86_rdseed, JL_X86_adx, JL_X86_prfchw, JL_X86_clflushopt, JL_X86_xsavec, JL_X86_xsaves))),
+        "skylake_avx512" => ISA(Set((JL_X86_movbe, JL_X86_sse3, JL_X86_ssse3, JL_X86_sse41, JL_X86_sse42, JL_X86_popcnt, JL_X86_pku, JL_X86_avx, JL_X86_avx2, JL_X86_aes, JL_X86_pclmul, JL_X86_fsgsbase, JL_X86_rdrnd, JL_X86_fma, JL_X86_bmi, JL_X86_bmi2, JL_X86_f16c, JL_X86_rdseed, JL_X86_adx, JL_X86_prfchw, JL_X86_clflushopt, JL_X86_xsavec, JL_X86_xsaves, JL_X86_avx512f, JL_X86_clwb, JL_X86_avx512vl, JL_X86_avx512bw, JL_X86_avx512dq, JL_X86_avx512cd))),
+    ),
+    "arm" => (
+        "armv7l" => ISA(Set{UInt32}()),
+        "armv7l_neon" => ISA(Set((JL_AArch32_neon,))),
+        "armv7l_neon_vfp4" => ISA(Set((JL_AArch32_neon, JL_AArch32_vfp4))),
+    ),
+    "aarch64" => (
+        # Implicit in all sets, because always required: fp, asimd
+        "armv8.0_a" => ISA(Set{UInt32}()),
+        "armv8.1_a" => ISA(Set((JL_AArch64_lse, JL_AArch64_crc, JL_AArch64_rdm))),
+        "armv8.2_a_crypto" => ISA(Set((JL_AArch64_lse, JL_AArch64_crc, JL_AArch64_rdm, JL_AArch64_aes, JL_AArch64_sha2))),
+        "armv8.4_a_crypto_sve" => ISA(Set((JL_AArch64_lse, JL_AArch64_crc, JL_AArch64_rdm, JL_AArch64_fp16fml, JL_AArch64_dotprod, JL_AArch64_aes, JL_AArch64_sha2, JL_AArch64_dotprod, JL_AArch64_sve))),
+    ),
+)
+
+test_cpu_feature(feature::UInt32) = ccall(:jl_test_cpu_feature, Bool, (UInt32,), feature)
+cpu_family() = String(Sys.ARCH)
+
+"""
+    cpu_isa()
+
+Return the [`ISA`](@ref) (instruction set architecture) of the current CPU.
+"""
+function cpu_isa()
+    all_features = last(last(get(ISAs_by_family, cpu_family(), "" => [ISA(Set{UInt32}())]))).features
+    return ISA(Set{UInt32}(feat for feat in all_features if test_cpu_feature(feat)))
+end
+
+end # module CPUID
+
+using .CPUID
+
 # This exists to ease compatibility with old-style Platform objects
 abstract type AbstractPlatform; end
 
@@ -567,6 +635,31 @@ const arch_mapping = Dict(
     "armv6l" => "armv6l",
     "powerpc64le" => "p(ower)?pc64le",
 )
+# Keep this in sync with `CPUID.ISAs_by_family`
+const arch_march_isa_mapping = let
+    function get_set(arch, name)
+        all = CPUID.ISAs_by_family[arch]
+        return all[findfirst(x -> x.first == name, all)].second
+    end
+    Dict(
+        "x86_64" => Dict{String,CPUID.ISA}(
+            "x86_64" => get_set("x86_64", "x86_64"),
+            "avx" => get_set("x86_64", "sandybridge"),
+            "avx2" => get_set("x86_64", "haswell"),
+            "avx512" => get_set("x86_64", "skylake_avx512"),
+        ),
+        "armv7l" => Dict{String,CPUID.ISA}(
+            "armv7l" => get_set("arm", "armv7l"),
+            "neon" => get_set("arm", "armv7l_neon"),
+            "vfp4" => get_set("arm", "armv7l_neon_vfp4"),
+        ),
+        "aarch64" => Dict{String,CPUID.ISA}(
+            "armv8" => get_set("aarch64", "armv8.0_a"),
+            "thunderx2" => get_set("aarch64", "armv8.1_a"),
+            "carmel" => get_set("aarch64", "armv8.2_a_crypto"),
+        ),
+    )
+end
 const os_mapping = Dict(
     "macos" => "-apple-darwin[\\d\\.]*",
     "freebsd" => "-(.*-)?freebsd[\\d\\.]*",
--- a/base/.gitignore
+++ b/base/.gitignore
@@ -1,3 +1,4 @@
+/features_h.jl
 /pcre_h.jl
 /errno_h.jl
 /build_h.jl
--- a/src/processor.h
+++ b/src/processor.h
@@ -123,7 +123,7 @@ typedef enum {
 } jl_cpu_feature_t;
 #undef JL_FEATURE_DEF_NAME
 
-int jl_test_cpu_feature(jl_cpu_feature_t feature);
+JL_DLLEXPORT int jl_test_cpu_feature(jl_cpu_feature_t feature);
 
 static const uint32_t jl_sysimg_tag_mask = 0x80000000u;
 static const uint32_t jl_sysimg_val_mask = ~((uint32_t)0x80000000u);
--- a/test/binaryplatforms.jl
+++ b/test/binaryplatforms.jl
@@ -1,4 +1,16 @@
-using Test, Base.BinaryPlatforms
+using Test, Base.BinaryPlatforms, Base.BinaryPlatforms.CPUID
+
+@testset "CPUID" begin
+    @test CPUID.cpu_isa() isa CPUID.ISA
+
+    get_x86_64(n) = (CPUID.ISAs_by_family["x86_64"][n].second)
+    @test get_x86_64(2) <  get_x86_64(4)
+    @test get_x86_64(5) <= get_x86_64(5)
+    @test get_x86_64(3) >= get_x86_64(3)
+    @test get_x86_64(7) >= get_x86_64(1)
+    @test sort([get_x86_64(6), get_x86_64(4), get_x86_64(2), get_x86_64(4)]) ==
+        [get_x86_64(2), get_x86_64(4), get_x86_64(4), get_x86_64(6)]
+end
 
 # Helper constructor to create a Platform with `validate_strict` set to `true`.
 P(args...; kwargs...) = Platform(args...; validate_strict=true, kwargs...)


yuyichao:
As I said in https://github.com/JuliaLang/julia/issues/36367 this should not be a public API since using this to generate code is wrong. A strictly private API would be fine and should just use the existing C function instead.

yuyichao:
Also, this is capturing the build time environment which is wrong. Note that the selection of external binary also must be done at runtime and not package installation time or precompilation time. This is the problem shared by a lot of packages that must be fixed before they can be accepted to base/stdlib.

yuyichao (changes requested):
Fix detection time, do not export julia API.

staticfloat:
> As I said in #36367 this should not be a public API since using this to generate code is wrong. 

Sorry, I read through again and it's just not connecting for me.  How would this be used for code generation?

yuyichao:
For example, LoopVectorization.jl must not (and as I understand it currently is) using anything like this to determine what code to generate.
The target used for julia code and the actual hardware target are usually the same. However, whenever they disagree the hardware one is almost always the wrong one to use.

And again, having some dirty hack that should work for code loading is fine, as long as it's clear that no one outside of base should use it and the API may be deleted at any time without notice.

giordano:
Fair point about detecting the features at runtime.  Regarding the API, my understanding is that something is considered "public" if its documentation/docstring is exposed in the manual, which isn't the case in this pull request and wasn't my intention.

Regarding the use of `jl_test_cpu_feature`, should we manually copy the numeric values of _all_ features in `features_*.h` to this module and manually have a mapping between the value and the feature it represent, keeping this always in sync with the features defined on the C side?

yuyichao:
> Regarding the API, my understanding is that something is considered "public" if its documentation/docstring is exposed in the manual, which isn't the case in this pull request and wasn't my intention.

For now this can just go in the module that does the loading. That'll lower the chance of someone using it.

> Regarding the use of jl_test_cpu_feature, should we manually copy the numeric values of all features in features_*.h to this module and manually have a mapping between the value and the feature it represent, keeping this always in sync with the features defined on the C side?

[No.](https://github.com/JuliaLang/julia/issues/36367#issuecomment-646899712)

> you can define them by generating a jl file from the features_*.h headers.

You can just run a julia script through preprocessor. We do that for quite a few other headers.

giordano:
> We do that for quite a few other headers.

Any example to look at?

yuyichao:
https://github.com/JuliaLang/julia/blob/4a412e3cac1dbfaa88e76ac582a9855a575f5de4/base/Makefile#L24-L34

giordano:
Ok, I exported `jl_test_cpu_feature` and a Perl script kindly provided by Elliot generates a Julia file with the list of the numerical features in Julia.  Now there is a function to get the list of features.

This module is not going to stay in `Sys` but I'll move it later and I marked this PR as draft, also because some details of the API need to be ironed out, depending on exactly how we need to use the microarchitectures in the Artifacts module.

base/sysinfo.jl
```diff
@@ -526,4 +526,50 @@ function which(program_name::String)
...
+    features::Set{UInt32}
+end
+
+Base.:<=(a::ISA, b::ISA) = a.features <= b.features
+
+include("features_h.jl")
+
+# Sets of features for the x86-64 microarchitectures come from
+# https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html
+const X86_64_GENERIC = ISA(Set{UInt32}()) # Missing: mmx, sse, sse2
```
giordano, line 548: [`src/features_x86.h`](https://github.com/JuliaLang/julia/blob/08486888baad7584fe25ac4c6aee97730890968c/src/features_x86.h) doesn't have `mmx`, `sse`, nor `sse2`, do we just assume they're always there in x86-64 chips (which should be a safe assumption)?

yuyichao: Yes those are required

giordano:
I've moved the module under the new `BinaryPlatforms` module, so I believe I've now addressed all original comments

base/binaryplatforms.jl
```diff
@@ -7,6 +7,60 @@ export AbstractPlatform, Platform, HostPlatform, platform_dlext, tags, arch, os,
...
+### Submodule with information about CPU features
+module CPUID
+
+export cpu_isa
+
+"""
+    ISA(features::Set{String})
+
+A structure which represents the Instruction Set Architecture (ISA) of a
+computer.  It holds the `Set` of features of the CPU.
```
staticfloat, line 19: Can you document where the features are stored?  E.g. `These features are autogenerated from the `features_*.h` headers in the Julia C source code...` or something like that?

base/binaryplatforms.jl
```diff
@@ -7,6 +7,60 @@ export AbstractPlatform, Platform, HostPlatform, platform_dlext, tags, arch, os,
...
+# Sets of features for the x86-64 microarchitectures come from
+# https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html
+const X86_64_GENERIC = ISA(Set{UInt32}())
+const CORE2 = ISA(Set((X86_64_GENERIC.features..., JL_X86_sse3, JL_X86_ssse3)))
+const NEHALEM = ISA(Set((CORE2.features..., JL_X86_sse41, JL_X86_sse42, JL_X86_popcnt)))
+const SANDYBRIDGE = ISA(Set((NEHALEM.features..., JL_X86_avx, JL_X86_aes, JL_X86_pclmul)))
+const HASWELL = ISA(Set((SANDYBRIDGE.features..., JL_X86_movbe, JL_X86_avx2, JL_X86_fsgsbase, JL_X86_rdrnd, JL_X86_fma, JL_X86_bmi, JL_X86_bmi2, JL_X86_f16c)))
+const SKYLAKE = ISA(Set((HASWELL.features..., JL_X86_clflushopt, JL_X86_xsavec, JL_X86_xsaves)))
+const SKYLAKE_AVX512 = ISA(Set((HASWELL.features..., JL_X86_pku, JL_X86_rdseed, JL_X86_adx, JL_X86_prfchw, JL_X86_avx512f, JL_X86_clwb, JL_X86_avx512vl, JL_X86_avx512bw, JL_X86_avx512dq, JL_X86_avx512cd)))
+const X86_64_ISAS = [SKYLAKE_AVX512, SKYLAKE, HASWELL, SANDYBRIDGE, NEHALEM, CORE2, X86_64_GENERIC]
```
staticfloat, line 39: I think it would be nice to rephrase this as a big `Dict`, so that it can be easily explored by code.  E.g. something like:

```
ISAs_by_family = Dict(
    "x86_64" => Dict(
         "CORE2" => ISA( ... ),
         ...,
     ),
     "aarch64" => Dict(
         "armv8.1" => ISA( ... ),
         ...,
     ),
     ...
)
```

Then, we can have a `cpu_family()` method that returns something like `"x86_64"`, and `cpu_isa()` becomes just:

```
cpu_isa() = ISA(Set(last(get(ISAs_by_family, cpu_family(), [Set{UInt32}()])))
```

staticfloat (changes requested):
I think the largest piece missing from this is any kind of mapping for aarch64 and armv7l.  Once we have those pieces, this looks exactly like what I'm looking for.

giordano:
I addressed the two last comments

giordano:
On the `aarch64` build machine I get the following:
```julia
julia> using Base.BinaryPlatforms.CPUID

julia> for f in filter(n -> startswith(String(n), "JL_AArch64"), names(CPUID; all=true))
           CPUID.test_cpu_feature(getfield(CPUID, f)) && println(f)
       end
JL_AArch64_aes
JL_AArch64_crc
JL_AArch64_sha2
```
@yuyichao any clue why only these features are set?  In particular, the machine is armv8.1-a, shouldn't [`JL_AArch64_v8_1a`](https://github.com/JuliaLang/julia/blob/8bdf5695e25e77d53f8934d9f135858d174e30d1/src/features_aarch64.h#L65) be set as well?
```julia
julia> CPUID.test_cpu_feature(CPUID.JL_AArch64_v8_1a)
false
```

yuyichao:
ISA version is impossible to detect (it's not a thing, basically). The only way to set it is through guessing. What is the CPU? Looking at the list, the only v8.1a I've seen is thunderx2....

base/binaryplatforms.jl
```diff
@@ -7,6 +7,62 @@ export AbstractPlatform, Platform, HostPlatform, platform_dlext, tags, arch, os,
...
+Base.:<(a::ISA,  b::ISA) = a.features <  b.features
+Base.isless(a::ISA,  b::ISA) = a < b
+
+include("features_h.jl")
+
+# Sets of features for the x86-64 microarchitectures come from
+# https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html.  Keep in sync with
+# `arch_march_isa_mapping`.
+const ISAs_by_family = Dict(
+    "x86_64" => (
```
KristofferC, line 38: Does it make sense to make these enums so you know from the type what possible the possible values are?

staticfloat: I like Dict's better so that we can add to them via Pkg hooks in the future.

KristofferC: I didn't mean to change the Dict, just the entries, `"core2", `"nehalem"` etc. So you know that the keys are arches, and not say, the life work of Shakespeare and you can reflect on the instances of the enum.

staticfloat: If we change the keys like `"core2"` from a string to an enum type, how do we add a new value to that enum type at runtime?  What I'm considering is that Julia 1.6 is around for 3+ years, and in that time we want to add new ISAs, which we will be able to do from within a package by running something like `Base.BinaryPlatforms.ISAs_by_family["cannonlake"] = ...`.

KristofferC: If we want to be able to interactively add new ISA's at runtime from packages then it clearly can't be an enum. I didn't know that was desireable, especially when it seems like there are multiple data structures here that need to be kept in sync and that the features are generated from the C-code. But if that's the case, then yeah, can't use an enum.

yuyichao: > which we will be able to do from within a package by running something like `Base.BinaryPlatforms.ISAs_by_family["x86_64"]["cannonlake"] = ....`

Note that nothing from this PR will be a stable API.

giordano:
> What is the CPU?

The features shown in `/proc/cpuinfo` are 
```
fp asimd evtstrm aes pmull sha1 sha2 crc32
```

I'm a bit confused by the comments in https://github.com/JuliaLang/julia/blob/8bdf5695e25e77d53f8934d9f135858d174e30d1/src/features_aarch64.h#L14-L15 What does it mean that `aes` is implied by `aes`?  The feature at line 15 is called `aes` (like the one in line 14) but the comment mentions `PMULL`, is this correct?

yuyichao:
> The features shown in `/proc/cpuinfo` are

.... but what CPU is it...

> What does it mean that aes is implied by aes? The feature at line 15 is called aes (like the one in line 14) but the comment mentions PMULL, is this correct?

Yes it is correct. It says `` HWCAP_AES. Implied by `aes` ``. The exported name is to match the LLVM one and LLVM's `aes` actually means `HWCAP_PMULL` (`ID_AA64ISAR0_EL1.AES == 2`) rather than `HWCAP_AES` (`ID_AA64ISAR0_EL1.AES == 1 || ID_AA64ISAR0_EL1.AES == 2`). The commented out name also says `aes` simply because there's nothing else I can think of to put there....

giordano:
> .... but what CPU is it...

The list of features was the only information I could find in `/proc/cpuinfo`, no reference to the model.  `lspcu` says `ThunderX 88XX`

yuyichao:
> The list of features was the only information I could find in /proc/cpuinfo, no reference to the model. lspcu says ThunderX 88XX

What's the full output in `/proc/cpuinfo`.

> lspcu says ThunderX 88XX

ThunderX should be v8

giordano:
> What's the full content of /proc/cpuinfo.

```
processor       : 0
BogoMIPS        : 200.00
Features        : fp asimd evtstrm aes pmull sha1 sha2 crc32
CPU implementer : 0x43
CPU architecture: 8
CPU variant     : 0x1
CPU part        : 0x0a1
CPU revision    : 1
```
repeated a bunch of times for all CPUs

yuyichao:
> ```
> CPU implementer : 0x43
> CPU architecture: 8
> CPU variant     : 0x1
> CPU part        : 0x0a1
> CPU revision    : 1
> ```

Yes that's a thunderx88 and we should be [detecting it correctly](https://github.com/JuliaLang/julia/blob/8bdf5695e25e77d53f8934d9f135858d174e30d1/src/processor_arm.cpp#L871). Why do you believe it's v8.1a?

giordano:
> Why do you believe it's v8.1a?

I was told so :disappointed:

yuyichao:
... Well, it is v8 according to https://en.wikichip.org/wiki/cavium/thunderx unlike [thunderx2](https://en.wikichip.org/wiki/cavium/thunderx2) which is the only v8.1a that I've seen. (most jumps straight to v8.2). It's also annouced before v8.1a was announced according to wikichip and was released the same month v8.1 is announced so I don't think it'll be v8.1...

staticfloat:
I was told the machine was a Thunder X2. Heh. Well this is good progress!

staticfloat:
Confirmed that on a Graviton2, this successfully identifies it as `v8.2-a`!

yuyichao:
Since this detection is unreliable (it's based on known name) I would not recommend relying on it. You should check for the features you need instead.

staticfloat:
We get to set the features we need; e.g. we're generating binaries in BB by using compiler flags like `"-march=armv8.2-a"`.  Can you point me to the right table to use to translate from the JL_AArch64_XXX flags to ensure that any binary generated with `-march=armv8.2-a` can run on that CPU?

yuyichao:
Well, that's the part that needs to be fixed. v8.2a comes with features (don't remember which one off the top of my head) that cannot be reliably detected.

staticfloat:
Fixed by whom? Is it the LLVM detection that is unreliable?

yuyichao:
> Fixed by whom?

Fixed by you, to change the compilation flags.

> Is it the LLVM detection that is unreliable?

No, there's simply no way at the hardware level to detect which version it is. The only thing available is the feature set. And the linux kernel also does not expose all the features that LLVM may use.

staticfloat:
So is what you're saying that we should avoid the `-march=armv8.X-a` flags completely, and instead use `armv8a+feature1+feature2+feature3`, and only use features that can be reliably detected from Julia?

yuyichao:
Yes. At least until someone can find a way to reliably detect all features in all `8.*-a` versions. It would be OK to give them aliases like `v8.1-a` but that shouldn't be the thing passed to the compiler.

staticfloat:
Okay, this we can work with.  So you suggest that we look through what is defined in https://github.com/JuliaLang/julia/blob/master/src/features_aarch64.h, take those as the official list of flags that can be reliably detected on AArch64, and map those to what is implied by each `armv8.X-a` level.  We can then setup BB to define each "named" `armv8.x-a` level to that set of features, then we can test them here in the CPUID module.  And we'll completely ignore the `JL_AArch64_v8_Xa` flags?

yuyichao:
Yes.

yuyichao:
All of those maps to kernel ABI so that's about as stable as it can be. It's possible that more could be added later for earlier versions but doing that shouldn't break anything.

staticfloat:
As a datapoint, I looked at what [`archspec`](https://github.com/archspec/archspec) does (the microarchitecture management library used by `spack`) and it looks like they just use what you can get from `/proc/cpuinfo`, so they're in the same position as us, and they also only use families of feature flags.

Example mappings for [`thunderx2`](https://github.com/archspec/archspec-json/blob/a800d749e7997a72d1da777b4f4cb02f13331696/cpu/microarchitectures.json#L1275-L1289) and [`graviton2`](https://github.com/archspec/archspec-json/blob/a800d749e7997a72d1da777b4f4cb02f13331696/cpu/microarchitectures.json#L1378-L1399).

I'm going to collect here a list of features that we should map to: I'm going to edit this comment until I'm happy with it.  :)

First off, some useful resources:
* GCC aarch64 core definitions: https://github.com/gcc-mirror/gcc/blob/master/gcc/config/aarch64/aarch64-cores.def
* GCC base aarch64 architecture flags: https://github.com/gcc-mirror/gcc/blob/40d00b7aab9be046a98b282a149ff3829887ba79/gcc/config/aarch64/aarch64.h#L170-L181

* `armv8.0-a`: The base level, no optional features enabled.  Should run on every armv8 processor under the sun.  GCC says that `fp` and `asimd` should always be on from this level.
* `armv8.1-a`: GCC says: [`fp`, `asimd`], `lse`, `crc` and `rdma` (we call this `rdm`?).
* `armv8.2-a`: GCC says: [`fp16`] t give us any extra required extensions here.
* `armv8.3-a`: GCC doesn't give us any extra required extensions here.
* `armv8.4-a`: GCC says: [`fp`, `asimd`, `lse`, `crc`, `rdma`], `fp16fml`, `dotprod`
* `armv8.5-a`: GCC says: [`fp`, `asimd`, `lse`, `crc`, `rdma`, `fp16fml`, `dotprod`], `sb`, `ssbs`, `predres` (this is not available in LLVM, apparently?)

Notably, I see that there are no base `armv8.X-a` levels that require common extensions such as the crypto extensions (e.g. (`aes`, `sha2`) on armv8.1- and (`aes`, `sha2`, `sha3`, `sm4`) on armv8.2+), or the `sve` extension, which we may want to support.


Questions:

* `fp` and `asimd` are both [commented out](https://github.com/JuliaLang/julia/blob/a7ce2a2de7417a63d5df5e9cfb15f480997bcfb0/src/features_aarch64.h#L11-L12) in the header, presumably because they'll always be on.  Can we un-comment them so that we have less implicit knowledge and it's easier to do exhaustive tests?

* Given your domain knowledge of cores, do you think there are useful levels that we should compile for?  For x86_64 it seems clear that the `avx`, `avx2`, etc... feature boundaries are useful stepping stones for performance.  In this case, I can only guess that we should support some base levels but I'm not sure what optional extensions we should build for, if any.  Right now, I'm leaning toward us supporting the following:

  - `armv8.0-a` -> [`fp`, `asimd`] - matches RPi4 in aarch64 mode, and thunderx88
  - `armv8.1-a` -> [`fp`,  `asimd`, `lse`, `crc`, `rdma`] - matches thunderx2
  - `armv8.2-a_crypto` -> [`fp`,  `asimd`, `lse`, `crc`, `rdma`, `aes`, `sha2`] - matches thunderx2, carmel and graviton2
  - `armv8.4-a_crypto_sve` -> [`fp`, `asimd`, `lse`, `crc`, `rdma`, `fp16fml`, `dotprod`, `aes`, `sha2`, `sve`] - matches `a64fx`

yuyichao:
> First off, some useful resources:

You can just use our definition. https://github.com/JuliaLang/julia/blob/4864161c3c18734670ce79165d361d3b0f1297e2/src/processor_arm.cpp#L230-L341 I believe this is the most complete list you can get from anywhere.

> fp and asimd are both commented out in the header, presumably because they'll always be on. Can we un-comment them so that we have less implicit knowledge and it's easier to do exhaustive tests?

No, they are just required feature. Soft floating point is not supported. I'm not sure what can be tested about something that just isn't supposed to work.

> Given your domain knowledge of cores, do you think there are useful levels that we should compile for?

It was "clear" for x86 only because there are changes on the vector registers that somewhat overshadows other changes. There are actually also a few other extensions like bit manipulation and cache (prefetch, invalidate, flush) manipulations, half precisions, that doesn't really fit into the sse/avx evolution timeline. Of course since there are only two major players there aren't too many lines of succession. Even then, AMD is known for deprecating old and unused extensions in newer processors and actually removing them so it's not actually not as simple as the illusion one might get from just looking at the vector registers.

On AArch64, the only extra SIMD level exists is SVE (and SVE2 I guess) so there just isn't the same evolution line that overshadow everything else. Also, SVE may or may not actually work with our compiler and unless someone get a chip/server from fujitsu or is working on fugaku you'll not be able to actually use the instruction. There is at least still another useful enough line of succession on aarch64 though, which is the memory model. 8.1-atomic (i.e. `lse`) and later ones like `rcpc` are interesting enough (and according to Keno good enough to run `rr`) so that's one feature that's worth supporting, though probably only for libraries that does their own locking etc.

Other than the atomics, there are basically about the same number of other random interesting features. They are just not burried among a million avx512 flavors........ The usuable hardwares with these extensions are scarce enough that I don't really know what are the ones that people uses. (if anyone want to teach me how to hack a mid-high end phone/tablet so that I can run gdb on it I'll be very interested.............) OTOH, despite the dozens of cores and hundreds of SoC names, the actual unique features are actually pretty limited. If you have a look at the feature lists we have, most cores are basically either v8-a or v8.2-a and potentially with extensions that agrees with the cortex-a line fairly well. Apple seems to be the most agressive player and is bumping their ISA version on every new chip since a11 but unless someone can figure out how to run linux on there it won't be very relavent for linux libraries.

So in the end I'll say just go with what known cores have is probably fine. This is also how I decided to suggest thunderx originally as an additional target since at the time it seems like it'll be the most immediate "server" platform that people can use. Feature wise this shouldn't give too many variance to support (it seems that `v8` and `v8.2,rcpc,fullfp16` would be good, maybe adding carmel if people care a lot since it seems to be missing `rcpc`). I'm not sure if there's anything that's worth optimizing for a particular core but I kind of doubt it. With more and more players (apple and marvell being the main exceptions) using straight or tweaked arm cortex lines the performance model is probably not going to diverge too much....

staticfloat:
Thanks, that's really helpful Yichao!

> No, they are just required feature. Soft floating point is not supported. I'm not sure what can be tested about something that just isn't supposed to work.

It's not about testing with a platform that doesn't have `fp`, it's about having our feature lists be consistent the rest of the world.  It's annoying to have to carry the cognitive load of remembering that even though `/proc/cpuinfo` specifies `fp` and `asmid`, the fact that Julia doesn't advertise support for those flags doesn't mean it's not compatible, we just have to remember that it is automatically included.

> I'm not sure if there's anything that's worth optimizing for a particular core but I kind of doubt it. With more and more players (apple and marvell being the main exceptions) using straight or tweaked arm cortex lines the performance model is probably not going to diverge too much....

So it sounds like just two marches for now may be okay:
* `armv8 => armv8.0-a+fp+asimd` (yes, I know `fp` and `asimd` are already on by default, I just want to be explicit)
* `armv8_2_crypto => armv8.0-a+fp+asimd+lse+crc+rdma+aes+sha2`

I'm not sure how important `rcpc` or `fullfp16` is; is it worthwhile adding an `armv8_2_crypto_rcpc`?  If you don't think it's that big of a deal, I vote that we just go with these two for now and we can add more later.

yuyichao:
> It's annoying to have to carry the cognitive load of remembering that even though /proc/cpuinfo specifies fp and asmid, the fact that Julia doesn't advertise support for those flags doesn't mean it's not compatible, we just have to remember that it is automatically included.

Well, you are free to ignore these anywhere in the stack if you want copy-paste from `/proc/cpuinfo` etc. However, note that they are **not** the LLVM name so if anything if they are added here they will not be called these. (LLVM names are `neon` and `fp-armv8` I believe) and these aren't the only one either so you always have to know the translation. That's why I added document to map between llvm name, HWCAP name and feature register bits.

Also, the two are always exactly the same according to the arm architecture manual so allowing both will be even more confusing. It's the same reason many other features are commented out, (e.g. `asimdhp`, `sm3`, `svei8mm`, `svebf16`)

yuyichao:
> I'm not sure how important rcpc or fullfp16 is; is it worthwhile adding an armv8_2_crypto_rcpc? If you don't think it's that big of a deal, I vote that we just go with these two for now and we can add more later.

According to Keno, rcpc is suspected to be required by `rr`. I don't think it'll be important for many external libraries (and still, I believe very few of these are important for external libraries and the rest can and should do internal dispatch.........................)

base/binaryplatforms.jl
giordano: I changed how `arch_march_isa_mapping` is defined below, which should make it easier to keep that dictionary in sync with this one.

staticfloat (approved):
LGTM, I say ready to merge.

Thanks @yuyichao for his careful attention, and @giordano for his work putting this together!

staticfloat:
> According to Keno, rcpc is suspected to be required by `rr`.

He's not 100% certain, so we're going to do some experiments to figure out exactly what is required by building `rr` with different `march`'es and figuring out what we need.

chriselrod:
> For example, LoopVectorization.jl must not (and as I understand it currently is) using anything like this to determine what code to generate.
> The target used for julia code and the actual hardware target are usually the same. However, whenever they disagree the hardware one is almost always the wrong one to use.
> 
> And again, having some dirty hack that should work for library loading most of the time is fine, as long as it's clear that no one outside of base should use it and the API may be deleted at any time without notice.

@yuyichao I didn't see this. It would be great to have an API that provides features that are safe to really on.
LoopVectorization hasn't relied on CpuId for a couple months, instead queering LLVM with getHostCpuFeaures. I made the change from CpuId because some people experienced crashes when running on avx512 hardware while LLVM not listing it among the features (the recent bug on mac's).

A more abstracted api would be generally useful. It's be great to have a reliable FMA_NATIVE, for example.

yuyichao:
The answer from LLVM was wrong as well.

chriselrod:
It solved the issue on Macs, but it is incorrect when starting Julia with `-Civybridge`. I haven't noticed any issues outside of specifying arch on startup yet.
Is there any way to do this correctly?

Aside from "don't generate machine specific code from Julia at all".
Maybe someday I can try implementing it in LLVM, but experimenting is much easier in Julia -- at least for me -- and the library is still immature, lacking many capabilities I intend it to have. Meaning there is still a lot more experimenting to be done.

base/Makefile
```diff
@@ -21,6 +21,18 @@ else
   PCRE_INCL_PATH := $(build_includedir)/pcre2.h
 endif
 
+define parse_features
+@echo "# $(2) features" >> $@
+@$(call PRINT_PERL, cat ../src/features_$(1).h | perl -lne 'print "const JL_$(2)_$$1 = UInt32($$2)" if /^\s*JL_FEATURE_DEF(?:_NAME)?\(\s*(\w+)\s*,\s*([^,]+)\s*,.*\)\s*(?:\/\/.*)?$$/' >> $@)
+@echo >> $@
+endef
+
+$(BUILDDIR)/features_h.jl: ../src/features_x86.h ../src/features_aarch32.h ../src/features_aarch64.h
```
vchuravy, line 30: Broke out of dir build":

```
vchuravy@odin ~/b/julia-replloader [2]> make -j
make[1]: *** No rule to make target '../src/features_x86.h', needed by 'features_h.jl'.  Stop.
make[1]: *** Waiting for unfinished jobs....
make: *** [/home/vchuravy/src/julia/Makefile:66: julia-base] Error 2
make: *** Waiting for unfinished jobs....
```

PR #37491, opened 2020-09-09
change display of type aliases to avoid `=`

JeffBezanson:
```
julia> Vector
Vector{T} where T (alias for Array{T, 1} where T)
```
instead of
```
julia> Vector
Vector{T} where T = Array{T, 1} where T
```
The `where T =` is probably confusing for new users.


vtjnash (approved):
Just needs fixes to doctests

PR #37496, opened 2020-09-09
Docs: Plus and star operators, for numbers and strings

PallHaraldsson:
Trying to be friendly to beginners, mostly thinking of for strings here, rather than numbers.

### DIFF
--- a/base/docs/basedocs.jl
+++ b/base/docs/basedocs.jl
@@ -928,7 +928,7 @@ kw"&&"
 """
     x || y
 
-Short-circuiting boolean OR.
+Short-circuiting boolean OR. Note, to concatenate strings in Julia use the `*` operator.
 """
 kw"||"
 
@@ -2107,9 +2107,15 @@ Creates a null pointer to type `T`.
 Ptr{T}()
 
 """
+    `plus` is not used in Julia.
+Use `+` for adding numbers together, but note, `*` is used in Julia to concatenate strings.
+"""
+kw"plus"
+
+"""
     +(x, y...)
 
-Addition operator. `x+y+z+...` calls this function with all arguments, i.e. `+(x, y, z, ...)`.
+Addition operator. `x+y+z+...` calls this function with all arguments, i.e. `+(x, y, z, ...)`. Note, to concatenate strings in Julia, use the `*` operator.
 
 # Examples
 ```jldoctest


PallHaraldsson:
@Moelf at https://github.com/JuliaLang/julia/issues/37469#issuecomment-689567021
>can you show an example where we document non-existing stuff? I can work on this

You can see it here, and I wasn't sure either and copied from some other PR. [I think we should also target beginners, and e.g. "
"Java/JavaScript/C#" converts.]

base/docs/basedocs.jl
```diff
@@ -2106,10 +2106,16 @@ Creates a null pointer to type `T`.
...
+    `plus` is not used in Julia.
+Use `+` for adding numbers together, but note, `*` is used in Julia to concatenate strings.
+"""
+kw"plus"
+
 """
     +(x, y...)
 
-Addition operator. `x+y+z+...` calls this function with all arguments, i.e. `+(x, y, z, ...)`.
+Addition operator. `x+y+z+...` calls this function with all arguments, i.e. `+(x, y, z, ...)`. Note, to concatenate strings in Julia, use the `*` operator.
```
simeonschaub, line 2118: This does seem like a useful hint, because `+` for string concatenation is used in a majority of languages, most notably Python, which is probably the most common language Julia users switch from. The others seem to only be used in more obscure languages, so I'm a bit more skeptical about them. It also seems like the discussion in #37469 was generally against adding `plus`.

PallHaraldsson: SQL isn't really an obscure language, if you mean that, but `||` is unusual and I know it only used there. That however is just an addition to an already existing help item (for completeness, more or less, I know ++ is used in some language, I just wanted to cover the popular SQL).

dkarrasch:
I'm not sure these additions are helpful. `+(::String,::String)` currently throws an error, which forces you to do research on how to do string concatenation in Julia. This is likely the moment when any (including new ones) user finds https://docs.julialang.org/en/v1/manual/strings/index.html#man-concatenation, which seems to answer more questions than most users will actually have. Also, why is there a reference to string concatenation in the `||` docstring? Is that because this is used in some other language? To me, this looks like a very random place to add such a reference.

tpapp:
I don't doubt that you mean well, but making the docs longer with redundant/unnecessary information is not helpful to beginners.

String concatenation is explained in the [Strings chapter](https://docs.julialang.org/en/v1/manual/strings/#man-concatenation), and mentined in the [Noteworthy differences](https://docs.julialang.org/en/v1/manual/noteworthy-differences/) chapter. Further repetition is not warranted.

Generally the Julia docs should document Julia, not what Julia isn't.

PallHaraldsson:
I would like more feedback, e.g. from @StefanKarpinski (it was suggested privately I should join his (related?) "and", "or" discussion issue). I could trim out stuff in this PR, if people really want that.

I'm just trying to put myself in the shoes of beginners (it's been a while since I was one), and I could imagine this question (to google): beginner: "how do you plus together to strings".

>This is likely the moment when any (including new ones) user finds https://docs.julialang.org/en/v1/manual/strings/index.html#man-concatenation

In the manual, looking up "+" or "plus", is rather useless, you have to know already to look for the technical term "concatenation".

Googling "julia + together two strings" gets me the long string section (not the relevant subsection directly), and to be fair "Julia: concatenate string and integer - Stack Overflow". I just like this also to be looked up in the REPL.

We must realize, Julia is the obscure language, to many, at least using the * operator, so I think warranted to go out of our way. I see no harm in adding a few bytes to already existing help texts.

The very/only popular SQL has || for concatenation (the only standardized way; MySQL had only + then both later, e.g. PostgreSQL only still has the standard way), and related maybe in PL/SQL has it and a few other, also  I could see people learning SQL first or at least concurrently in the same first semester (then likely with Python, Java or other language, at least not where * is used, is Julia the only language?).

KristofferC:
Could use https://github.com/JuliaLang/julia/blob/10f5a05c21af3ca66c05654e738d0dbfc482c5fb/base/experimental.jl#L165-L183 to provide a hint when a method error for `"foo" + "bar"` is run.

PallHaraldsson:
I'm all for rather adding to the error message itself (or both), I wasn't just aware of how.

tpapp:
> In the manual, looking up "+" or "plus", is rather useless, you have to know already to look for the technical term "concatenation".

Not really. You just have to realize that you are dealing with strings, which have *their own chapter*, and you will find concatenation there (among other things; most of them indispensable if you are dealing with strings, eg how UTF8 is handled in indexing/iteration).

People are of course free to ignore the manual and google up every single step they want to perform. But this is a very unproductive and frustrating way of programming (or doing anything, really), not something we should support or encourage.

PallHaraldsson:
I agree you need to read the manual for how e.g. "UTF8 is handled in indexing/iteration", concatenation is however much more common, and what some people only do. I have my own ideas however on how to simplify the unfortunate UTF8 indexing situation (@ScottPJones' packages has similar ideas, while not exactly the same). Strings are in some sense, or should be a simple thing, maybe later... and I can see how beginners wouldn't want to read a whole chapter on it.

>and mentined in the Noteworthy differences chapter. Further repetition is not warranted.

People might overlook, intentionally, not coming from those few languages. I just disagree about the repetition not needed.

PR #37497, opened 2020-09-09
Improve precision of nfields_tfunc on `Union`

Keno:
`nfields_tfunc(Tuple{Int, Union{Int, Float64}})` gave `Const(2)`,
but `nfields_tfunc(Union{Tuple{Int, Float64}, Tuple{Int, Int}})` gave
`Int`. In general, it's not great for the tfuncs to give different
answers when given `==` types, because other parts of the system
will happily substitute such types without changes to the user code,
which can change the inference result and lead to instability.

### DIFF
--- a/base/compiler/tfuncs.jl
+++ b/base/compiler/tfuncs.jl
@@ -381,6 +381,11 @@ function nfields_tfunc(@nospecialize(x))
             return Const(isdefined(x, :types) ? length(x.types) : length(x.name.names))
         end
     end
+    if isa(x, Union)
+        na = nfields_tfunc(x.a)
+        na === Int && return Int
+        return tmerge(na, nfields_tfunc(x.b))
+    end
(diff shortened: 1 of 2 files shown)


base/compiler/tfuncs.jl
```diff
@@ -381,6 +381,9 @@ function nfields_tfunc(@nospecialize(x))
             return Const(isdefined(x, :types) ? length(x.types) : length(x.name.names))
         end
     end
+    if isa(x, Union)
+        return tmerge(nfields_tfunc(x.a), nfields_tfunc(x.b))
```
martinholters, line 385: Could optimize a bit for larger unions by returning immediately if `nfields_tfunc(x.a)==Int` and only recursing down `nfields_tfunc(x.b)`otherwise (or vice versa). Not sure if worth it.

PR #32765, opened 2019-08-02
Issue 32690 : SSHManager with csh shell

DcfGH:
Dear Julia team,

A pull request from issue 32690
Best regards

PR #37501, opened 2020-09-10
Support `addprocs()` when remote ssh shell is `csh`

FelipeLema:
Re-write of https://github.com/JuliaLang/julia/pull/32765 considering the following:
- [diff against newer code](https://github.com/JuliaLang/julia/pull/32765#issuecomment-612549856)
- [use `csh` as remote shell for tests](https://github.com/JuliaLang/julia/pull/32765#issuecomment-688963441)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,6 @@ matrix:
           sources:
             - ubuntu-toolchain-r-test
           packages:
-            - csh
             - ccache
             - libssl1.0.0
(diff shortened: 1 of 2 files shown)


FelipeLema:
I was expecting tests to fail in travis... am I modifying the proper file? is travis still being used?

fredrikekre:
> is travis still being used?

No.

PR #37505, opened 2020-09-10
Fix broadcasted assignment for Cartesian indexing where the dimension of the `CartesianIndex` is greater than one

DilumAluthge:
Briefly, #35926 breaks the following code:
```julia
a = [1, 2, 3, 4]
a[CartesianIndex(1,1)] .= 0
```

This PR fixes the behavior for `CartesianIndex` of dimension greater than one, but preserves the behavior for one-dimensional `CartesianIndex`.


DilumAluthge:
cc: @andreasnoack 
cc: @mbauman 
cc: @stev47

fredrikekre:
Why should `a[CartesianIndex(1,1)] .= 0` work? `a[1,1] .= 0` doesn't for example.

DilumAluthge:
Yep. It's looking like the error is the correct behavior here.

PR #37508, opened 2020-09-10
Replace some comparisons with `===` and `!==`

yuyichao:
When one side is one of

* Concrete `DataType`
* `Symbol`
* `TypeName`
* `Module`
* Singleton object
* Types with no type parameters
* Other types with unique representations

    `Union{}`, `Type{Union{}}`, `Type{Module}`


martinholters:
Note that `T == typeof(x)` is not always equivalent to `T === typeof(x)`, e.g.:
```julia
julia> Vector{Tuple{>:Int}} == typeof(Tuple{Any}[])
true

julia> Vector{Tuple{>:Int}} === typeof(Tuple{Any}[])
false
```
I'm not sure that poses any real problems with the changes proposed here; from a _brief_ glance everything looked harmless enough, but we might want to double-check.

base/compiler/tfuncs.jl
```diff
@@ -949,7 +949,7 @@ function fieldtype_tfunc(@nospecialize(s0), @nospecialize(name))
     if isa(s0, Const) && (!(isa(s0.val, DataType) || isa(s0.val, UnionAll) || isa(s0.val, Union)) || s0.val === Module)
         return Bottom
     end
-    if (s0 isa Type && (s0 == Type{Module} || s0 == Type{Union{}})) || isa(s0, Conditional)
```
vtjnash, line 952: This isn't true:
```
julia> Type{Union{}} == typeof(Union{}), Type{Union{}} === typeof(Union{})
(true, false)
```

stdlib/Distributed/test/distributed_exec.jl
```diff
@@ -428,10 +428,10 @@ try
     end
     error("unexpected")
 catch ex
-    @test typeof(ex) == CompositeException
+    @test typeof(ex) === CompositeException
     @test length(ex) == 5
-    @test typeof(ex.exceptions[1]) == TaskFailedException
-    @test typeof(ex.exceptions[1].task.exception) == ErrorException
+    @test typeof(ex.exceptions[1]) === TaskFailedException
+    @test typeof(ex.exceptions[1].task.exception) === ErrorException
```
vtjnash, line 434: I think these should be `isa` checks, since they theoretically could become abstract types without the API (and thus the test) being violated

stdlib/LibGit2/test/libgit2.jl
```diff
@@ -763,7 +763,7 @@ mktempdir() do dir
                 LibGit2.GitRepo(path)
                 error("unexpected")
             catch e
-                @test typeof(e) == LibGit2.GitError
+                @test typeof(e) === LibGit2.GitError
```
vtjnash, line 766: Also should be isa tests in this file?

stdlib/LibGit2/src/repository.jl
```diff
@@ -142,7 +142,7 @@ function (::Type{T})(repo::GitRepo, spec::AbstractString) where T<:GitObject
     @check ccall((:git_revparse_single, :libgit2), Cint,
                  (Ptr{Ptr{Cvoid}}, Ptr{Cvoid}, Cstring), obj_ptr_ptr, repo.ptr, spec)
     # check object is of correct type
-    if T != GitObject && T != GitUnknownObject
+    if T !== GitObject && T !== GitUnknownObject
```
vtjnash, line 145:
```suggestion
    if !(T <: GitObject || T <: GitUnknownObject)
```
might convey the intention better?

vtjnash (commented):
With a slight nod towards https://docs.julialang.org/en/v1/manual/style-guide/#Be-careful-with-type-equality, this lgtm

vtjnash:
rather stale. do you plan to finish?

fingolfin:
This has a gazillion conflicts and unanswered review comments, and no further reactions in four years. I think it is fair to close this at this point. That's not a rejection of the changes, though, of course they could be resubmitted!

PR #37512, opened 2020-09-10
Fix coercion of floats to integer for %d printf specifier

### LINKED ISSUE #37507: Regression in @sprintf
Version 1.5
```
using Printf
@sprintf("%d", 25.5)
"26"
```

Version 1.6
```
using Printf
@sprintf("%d", 25.5)
"25"
```

```
versioninfo()
Julia Version 1.6.0-DEV.882
Commit c7b17fe3a6 (2020-09-09 22:41 UTC)
Platform Info:
  CPU: Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
```

quinnj:
Fixes #37507. The issue here was the calling of `trunc` instead of
`round` when coercing `AbstractFloat` arguments to `Integer`. Note this
throws a scary warning and has undefined behavior in C, but it looks
like legacy Printf code supported this, so we can try to do the right
thing here. Previously, Printf called the grisu equivalent of
`Ryu.writefixed(x, precision=0)` for float arguments, but as far as I
can tell, calling `round(x)` should accomplish the same thing.


quinnj:
>  but as far as I can tell, calling round(x) should accomplish the same thing.

As noted in [this comment](https://github.com/JuliaLang/julia/issues/37507#issuecomment-690405879), there is a subtle difference between the Grisu & Ryu algorithms in that the former rounds `.5` upwards (`RoundNearestTiesUp`), while the latter rounds ties to even (`RoundNearest`).

PR #37515, opened 2020-09-10
Doc !== and similar operators to avoid gotchas

PallHaraldsson:
Might be helpful for dyslexics.
[skip ci]

### DIFF
--- a/base/operators.jl
+++ b/base/operators.jl
@@ -234,7 +234,7 @@ const ≡ = ===
     !==(x, y)
     ≢(x,y)
 
-Always gives the opposite answer as [`===`](@ref). Beware, it's not te opposite of [`==`](@ref).
+Always gives the opposite answer as [`===`](@ref). Beware, it's not the opposite of [`==`](@ref).
 
 # Examples
 ```jldoctest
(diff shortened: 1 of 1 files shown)


base/operators.jl
```diff
@@ -233,7 +234,7 @@ const ≡ = ===
     !==(x, y)
     ≢(x,y)
 
-Always gives the opposite answer as [`===`](@ref).
+Always gives the opposite answer as [`===`](@ref). Beware, it's not the opposite of [`==`](@ref).
```
dkarrasch, line 237:
```suggestion
Always gives the opposite answer to [`===`](@ref). Beware, it's not the opposite of [`==`](@ref).
```
? I'm not a native speaker, but you made the corresponding change above.

PallHaraldsson: I'm neither a native speaker, so I might be wrong. The "to"/"as" was a bit of a distraction. What do you think about the rest?

PallHaraldsson: https://english.stackexchange.com/questions/136940/opposite-to-or-opposite

Issue #37518, opened 2020-09-10
Can we extend `Base.@kwdef` to override default `show()` to print via keyword argument construction?


NHDaly:
```julia
julia> Base.@kwdef struct Statistics
           seen_count::Int = 0
           recomputed_count::Int = 0
           invalidated_count::Int = 0
       end
Statistics

julia> stats = Statistics(seen_count=3, recomputed_count=2)
Statistics(3, 2, 0)
```

People use `Base.@kwdef` for structs where they want users to set the fields by name. Can we extend this macro to also provide a definition for `Base.show()` that prints the struct with all of its args by name, something like this?:
```julia
julia> stats = Statistics(seen_count=3, recomputed_count=2)
Statistics(seen_count = 3, recomputed_count = 2, invalidated_count = 0)
```

------

EDIT: oh, duh, I guess this isn't possible because if the user then wants to provide their own `Base.show`, there would be a double-definition error. :/

Is there any clever way around that?

Or could we maybe have something like `@kwdef_and_show`? Or another macro `@kwshow` that could be composed with this macro? :)

NHDaly:
Or maybe via a parameter, like `@kwdef show=true struct S ... end`?

yuyichao:
A much cleaner way is to simply have a function that does the printing using keyword argument syntax and the user can simply use it with a single line of code.

NHDaly:
Mm that's a good idea, @yuyichao.

Hehe I realized I could just do this instead of asking, so i've implemented the `kwshow()` argument based approach here:
https://github.com/JuliaLang/julia/pull/37519

But now that i'm here and see your comment, you're probably right.. Open to changing it per others' feedback! :)

nickrobinson251:
I think [Parameters.jl](https://github.com/mauro3/Parameters.jl) defines a `show` method [when using `@with_kw`](https://mauro3.github.io/Parameters.jl/latest/manual/#Types-with-default-values-and-keyword-constructors-1)
e.g.
```julia
julia> using Parameters

julia> @with_kw struct Statistics
              seen_count::Int = 0
              recomputed_count::Int = 0
              invalidated_count::Int = 0
          end
Statistics

julia> stats = Statistics(seen_count=3, recomputed_count=2)
Statistics
  seen_count: Int64 3
  recomputed_count: Int64 2
  invalidated_count: Int64 0
```
Maybe other packages do too :)
[...]

PR #37519, opened 2020-09-10
Add optional `kwshow=true` argument to `@kwdef`

NHDaly:
Adds an optional `kwshow=true` argument to `Base.@kwdef` that will
auto-generate a `Base.show()` method that prints arguments as keyword
arguments, matching the way the type is constructed.

For example:
```julia
julia> Base.@kwdef kwshow=true struct Bar
    a::Int = 1         # specified default
    b::String          # required keyword
end
Bar

julia> Bar(a = 1, b = "hi")
Bar(a = 1, b = "hi")
```

### DIFF
--- a/base/show.jl
+++ b/base/show.jl
@@ -388,9 +388,10 @@ show(io::IO, @nospecialize(x)) = show_default(io, x)
 show(x) = show(stdout::IO, x)
 
 # avoid inferring show_default on the type of `x`
-show_default(io::IO, @nospecialize(x)) = _show_default(io, inferencebarrier(x))
+show_default(io::IO, @nospecialize(x); kwshow::Bool=false) =
+    _show_default(io, inferencebarrier(x); kwshow=kwshow)
 
-function _show_default(io::IO, @nospecialize(x))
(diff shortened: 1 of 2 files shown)


NHDaly:
Propagating over @yuyichao's comment from  https://github.com/JuliaLang/julia/issues/37518#issuecomment-690417116:

> A much cleaner way is to simply have a function that does the printing using keyword argument syntax and the user can simply use it with a single line of code.

vtjnash:
Yes, like Yichao, I'd suggest adding a parameter to `Base.show_default` and then this macro would only need to generate a call to it.

NHDaly:
> Yes, like Yichao, I'd suggest adding a parameter to `Base.show_default` and then this macro would only need to generate a call to it.

Sorry for my delay. I've done this now! :) Thanks for the suggestion.

It doesn't simplify the code too much, but I do agree it's an improvement. 👍
I only generate the overload for `Base.show(...) = Base.show_default(...; kwshow=true)` if the macro has the `kwshow=true` flag set, since otherwise the user may of course want to provide their own `show()` overload, and we don't want to cause a multiple definition warning. 

Thanks!

PR #37520, opened 2020-09-10
Base64: add isreadable and iswritable for Base64EncodePipe/Base64DecodePipe

johnnychen94:
This mimics the `isreadable` and `iswritable` methods of `IOStream` for Base64 pipelines and fixes the following error:

```julia
julia> using Base64

julia> Base64EncodePipe <: IO
true

julia> isreadable(Base64EncodePipe(IOBuffer()))
ERROR: MethodError: no method matching isreadable(::Base64EncodePipe)
```

### DIFF
--- a/stdlib/Base64/src/decode.jl
+++ b/stdlib/Base64/src/decode.jl
@@ -43,7 +43,7 @@ struct Base64DecodePipe <: IO
     end
 end
 
-Base.isreadable(pipe::Base64DecodePipe) = isreadable(pipe.io) && true
+Base.isreadable(pipe::Base64DecodePipe) = !isempty(pipe.rest) || isreadable(pipe.io)
 Base.iswritable(::Base64DecodePipe) = false
 
 function Base.unsafe_read(pipe::Base64DecodePipe, ptr::Ptr{UInt8}, n::UInt)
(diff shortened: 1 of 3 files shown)


stdlib/Base64/src/decode.jl
```diff
@@ -43,6 +43,9 @@ struct Base64DecodePipe <: IO
     end
 end
 
+Base.isreadable(pipe::Base64DecodePipe) = isreadable(pipe.io) && true
```
vtjnash, line 46:
```suggestion
Base.isreadable(pipe::Base64DecodePipe) = isreadable(pipe.io)
```

stdlib/Base64/src/encode.jl
```diff
@@ -44,6 +44,9 @@ struct Base64EncodePipe <: IO
     end
 end
 
+Base.isreadable(::Base64EncodePipe) = false
+Base.iswritable(pipe::Base64EncodePipe) = iswritable(pipe.io) && true
```
vtjnash, line 48:
```suggestion
Base.iswritable(pipe::Base64EncodePipe) = iswritable(pipe.io)
```

vtjnash (approved):
I'm not sure I understand the reason for `&& true`, though I suppose it may have come from `&& return true`, which is a common idiom for long form functions.

PR #37525, opened 2020-09-11
Make uniformscaling inequal to matrices

### LINKED ISSUE #716: isequal and hash are incompatible for UniformScaling
```julia
julia> isequal(diagm(ones(3)), I)
true

julia> isequal(diagm(ones(4)), I)
true

julia> hash(I)
0x6bd2535d045989b2

julia> hash(diagm(ones(4)))
0x6ba2635b5ca237d8

julia> hash(diagm(ones(3)))
0xf728e5a95d43274d
```

This is because `isequal` on `UniformScaling` is not defined and dispatches to the generic method

```julia
julia> @which isequal(diagm(ones(3)), I)
isequal(x, y) in Base at operators.jl:123
```

which calls `==`:

https://github.com/JuliaLang/julia/blob/b37cc3d7c03d0496c65431d621f5d6b7939f8089/base/operators.jl#L123

It seems `isequal` on `UniformScaling` has to be defined manually.

A more fundamental solution may be to define `isequal` in terms of the identity `===` by default.  It seems brittle to define a function with strict semantics (`isequal`) in terms of the one with weaker semantics (`==`).  (Though this is probably impossible to do within 1.x.)

### LINKED ISSUE #35485: isequal and hash are incompatible for UniformScaling

jakobnissen:
See https://github.com/JuliaLang/julia/pull/36918 . With this PR, instances of `UniformScaling` is now not `isequal` to matrices.

EDIT: Closes JuliaLang/LinearAlgebra.jl#716

### DIFF
--- a/stdlib/LinearAlgebra/test/uniformscaling.jl
+++ b/stdlib/LinearAlgebra/test/uniformscaling.jl
@@ -400,6 +400,11 @@ end
     @test 0denseI != 2I != 0denseI # test generic path / inequality on diag
     @test alltwos != 2I != alltwos # test generic path / inequality off diag
     @test rdenseI !=  I != rdenseI # test square matrix check
+
+    # isequal
+    @test !isequal(I, I(3))
+    @test !isequal(I(1), I)
+    @test !isequal([1], I)
 end
 
 @testset "operations involving I should preserve eltype" begin
--- a/stdlib/LinearAlgebra/test/uniformscaling.jl
+++ b/stdlib/LinearAlgebra/test/uniformscaling.jl
@@ -405,6 +405,8 @@ end
     @test !isequal(I, I(3))
     @test !isequal(I(1), I)
     @test !isequal([1], I)
+    @test isequal(I, 1I)
+    @test !isequal(2I, 3I)
 end
 
 @testset "operations involving I should preserve eltype" begin
--- a/NEWS.md
+++ b/NEWS.md
@@ -33,6 +33,8 @@ Language changes
   a function like other operators. The dotted version `.-->` is now parsed as well.
   For backwards compatibility, `-->` still parses using its own expression head
   instead of `:call`.
+* Instances of `UniformScaling` are no longer `isequal` to matrices. Previous
+  behaviour violated the rule that `isequal(x, y)` implies `hash(x) == hash(y)`.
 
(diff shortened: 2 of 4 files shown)


stdlib/LinearAlgebra/test/uniformscaling.jl
```diff
@@ -400,6 +400,11 @@ end
     @test 0denseI != 2I != 0denseI # test generic path / inequality on diag
     @test alltwos != 2I != alltwos # test generic path / inequality off diag
     @test rdenseI !=  I != rdenseI # test square matrix check
+
+    # isequal
+    @test !isequal(I, I(3))
+    @test !isequal(I(1), I)
+    @test !isequal([1], I)
```
simeonschaub, line 407: I first thought your approach would cause an ambiguity error for `isequal(I, I)`, but fortunately it doesn't because `UniformScaling` doesn't subtype `AbstractMatrix`. Nethertheless, would you mind adding a test for `isequal(I, 1I)` just to be sure this doesn't break in the future?

jakobnissen: Of course, added now.

vtjnash (approved):
Needs a rebase to pick up some CI fixes, but otherwise is good to merge (and knock an item off the v1.6 milestone label)

PR #37528, opened 2020-09-11
Use a fixed DateTime in rounding tests instead of now()

quinnj:
Using now led to inconsistent answers depending on when `now()` is; make
tests more reliable by using a fixed `DateTime`.

### DIFF
--- a/stdlib/Dates/test/rounding.jl
+++ b/stdlib/Dates/test/rounding.jl
@@ -190,7 +190,7 @@ end
 end
 
 @testset "Rouding DateTime to Date" begin
-    now_ = now()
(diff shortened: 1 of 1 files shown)


Moelf:
I agree but in this case it means there exist a single `Time` where `round(Date, now_, p, r)` did something different than first convert `now_` to `Date(now_)` and then do rounding -- which is strictly what user would expect.

quinnj:
Valid concerns; I dug in a bit further and discovered we were passing the wrong default rounding mode to `round`, which is `RoundNearestTiesUp`, but we were passing `RoundUp`, which led to the test failures. I've pushed the fix here.

Moelf:
thanks! I must missed that bit

PR #37531, opened 2020-09-11
performance tip: consider StaticArrays for small fixed-size arrays

stevengj:
I find myself recommending StaticArrays over and over again to people working with small fixed-size arrays.  Can we suggest this in the performance tips even though it is an external package?

### DIFF
--- a/doc/src/manual/performance-tips.md
+++ b/doc/src/manual/performance-tips.md
@@ -1110,7 +1110,7 @@ specialize code for the *size* of the array, e.g. by completely unrolling vector
 
 For example, if you are doing computations with 2d geometries, you might have many computations with 2-component vectors.  By
 using the `SVector` type from StaticArrays.jl, you can use convenient vector notation and operations like `norm(3v - w)` on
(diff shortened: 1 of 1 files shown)


stevengj:
I'm getting an unrelated CI error:
```
Error in testset Dates/rounding:
Test Failed at /buildworker/worker/tester_linux64/build/share/julia/stdlib/v1.6/Dates/test/rounding.jl:198
  Expression: round(Date, now_, p) == round(Date, now_, p, RoundUp)
   Evaluated: Dates.Date("2020-09-01") == Dates.Date("2020-10-01")
```

doc/src/manual/performance-tips.md
```diff
@@ -1101,6 +1101,17 @@ julia> @time begin
 Provided there is enough memory for the copies, the cost of copying the view to an array is
 far outweighed by the speed boost from doing the matrix multiplication on a contiguous array.
 
+## Consider StaticArrays for small fixed-size vector/matrix operations
```
KristofferC, line 1104: Maybe it makes sense to put out the `.jl` even in the title to highlight that we will talk about a package?

PR #37533, opened 2020-09-11
Add punctuation for pipes (|>) and composition (∘)

nathanrboyer:
I will need the syntax and accuracy checked but thought these should be included in the list.


StefanKarpinski (approved):

vtjnash:
Looks like `|` needs to be escaped in that table (with `&#124;` or `\|`?)

vtjnash:
It was fixed in https://github.com/JuliaLang/julia/pull/41167. Should we backport that?

PR #37535, opened 2020-09-11
Expand SnoopCompileCore.@snoopi to report _all MethodInstances_ that are inferred during a single inference call.

NHDaly:
Before this PR, it reported only the entry MethodInstance:
```julia
julia> module M
           h(a::Array) = a[1]::Integer + 1
           g(y::Integer, x) = h(Any[y]) + Int(x)
       end;

julia> SnoopCompileCore.@snoopi begin
           M.g(2,3.0)
       end
1-element Array{Tuple{Float64,Core.MethodInstance},1}:
 (0.0011429786682128906, MethodInstance for g(::Int64, ::Float64))
```

After this PR, we want to be able to get _all_ the MethodInstances that are (recursively) inferred during the inference to the top method instance.

Currently, it now looks like this:
```julia
julia> SnoopCompileCore.@snoopi begin
           M.g(2,3.0)
       end
1-element Vector{Tuple{Float64,Vector{Core.MethodInstance}}}:
 (0.001068115234375, [MethodInstance for g(::Int64, ::Float64), MethodInstance for h(::Vector{Any})])
```
But we're very open to other ideas for how best to represent this! :)
I think it might be super cool to return this as a _DAG_ rather than a vector, so that we could maybe then work our way up from the leaves timing `code_typed` on each of them one at a time to try to understand which function had the most _exclusive_ time?

----

This PR is basically applying the same trick we did in https://github.com/JuliaLang/julia/pull/37136 to get the full set of MethodInstances being compiled, rather than just the entry method instance.

### DIFF
--- a/base/compiler/abstractinterpretation.jl
+++ b/base/compiler/abstractinterpretation.jl
@@ -1257,8 +1257,15 @@ function abstract_eval_ssavalue(s::SSAValue, src::CodeInfo)
     return typ
 end
 
+#const __collect_inference_callees__ = Ref{Bool}(false)
+const __inference_callees__ = Vector{MethodInstance}()
+
 # make as much progress on `frame` as possible (without handling cycles)
 function typeinf_local(interp::AbstractInterpreter, frame::InferenceState)
+    #if __collect_inference_callees__[]
+        push!(__inference_callees__, frame.linfo)
+    #end
+
     @assert !frame.inferred
     frame.dont_work_on_me = true # mark that this function is currently on the stack
     W = frame.ip
--- a/base/compiler/abstractinterpretation.jl
+++ b/base/compiler/abstractinterpretation.jl
@@ -1257,14 +1257,14 @@ function abstract_eval_ssavalue(s::SSAValue, src::CodeInfo)
     return typ
 end
 
-#const __collect_inference_callees__ = Ref{Bool}(false)
+const __collect_inference_callees__ = fill(false)  # Singleton Unary Array since Ref isn't available yet
 const __inference_callees__ = Vector{MethodInstance}()
 
 # make as much progress on `frame` as possible (without handling cycles)
 function typeinf_local(interp::AbstractInterpreter, frame::InferenceState)
-    #if __collect_inference_callees__[]
+    if __collect_inference_callees__[]
         push!(__inference_callees__, frame.linfo)
-    #end
+    end
 
     @assert !frame.inferred
     frame.dont_work_on_me = true # mark that this function is currently on the stack
--- a/base/compiler/abstractinterpretation.jl
+++ b/base/compiler/abstractinterpretation.jl
@@ -1258,13 +1258,13 @@ function abstract_eval_ssavalue(s::SSAValue, src::CodeInfo)
 end
 
 const __collect_inference_callees__ = fill(false)  # Singleton Unary Array since Ref isn't available yet
-const __inference_callees__ = Vector{MethodInstance}()
+#const __inference_callees__ = Vector{MethodInstance}()
 
 # make as much progress on `frame` as possible (without handling cycles)
 function typeinf_local(interp::AbstractInterpreter, frame::InferenceState)
-    if __collect_inference_callees__[]
-        push!(__inference_callees__, frame.linfo)
-    end
+    #if __collect_inference_callees__[]
+    #    push!(__inference_callees__, frame.linfo)
+    #end
 
     @assert !frame.inferred
     frame.dont_work_on_me = true # mark that this function is currently on the stack
--- a/base/compiler/typeinfer.jl
+++ b/base/compiler/typeinfer.jl
@@ -497,6 +497,8 @@ function resolve_call_cycle!(interp::AbstractInterpreter, linfo::MethodInstance,
     return false
 end
 
+const __inference_callee_edges4__ = Vector{Tuple{MethodInstance, Vector{Tuple{MethodInstance,MethodInstance}}}}()
+
 # compute (and cache) an inferred AST and return the current best estimate of the result type
 function typeinf_edge(interp::AbstractInterpreter, method::Method, @nospecialize(atypes), sparams::SimpleVector, caller::InferenceState)
     mi = specialize_method(method, atypes, sparams)::MethodInstance
@@ -532,6 +534,11 @@ function typeinf_edge(interp::AbstractInterpreter, method::Method, @nospecialize
         end
         if caller.cached || caller.limited # don't involve uncached functions in cycle resolution
             frame.parent = caller
+        end
+        if __collect_inference_callees__[]
+            #Core.println("edge!")
+            #Core.println(__inference_callee_edges4__)
+            push!(__inference_callee_edges4__[end][2], (caller.linfo, frame.linfo))
         end
         typeinf(interp, frame)
         update_valid_age!(frame, caller)


NHDaly:
I'm super not sure where the best places actually are to do the logging from within Julia. I just poked around putting print statements in a bunch of places until I found a spot that seems(?) reasonable.

I'd super appreciate help / advice / suggestions on how best to do this! ❤️! Thanks in advance ~~ hopefully this is useful to others! :)

NHDaly:
This PR is basically applying the same trick we did in https://github.com/JuliaLang/julia/pull/37136 to get the full set of MethodInstances being compiled, rather than just the entry method instance.

timholy:
I haven't had a chance to think about this in detail, but you're not the first person to wish for this. :+1:

Keno:
This might be better done with a custom `AbstractInterpreter`, so you can also control caching. Otherwise the results of this will be wildly unstable, depending on what happens to be in the global cache.

NHDaly:
> This might be better done with a custom `AbstractInterpreter`, so you can also control caching. Otherwise the results of this will be wildly unstable, depending on what happens to be in the global cache.

Thanks @Keno. If I'm not mistaken, isn't what you're describing already true of the behavior of `@snoopi`? I think `@snoopi` is already going to return different answers depending on what's in the global cache, and I think this PR will behave the same way. Is that what you're referring to?
e.g. this?:
```julia
julia> f(x) = x+3

julia> SnoopCompileCore.@snoopi begin
           f(2)
       end
1-element Array{Tuple{Float64,Core.MethodInstance},1}:
 (0.0002541542053222656, MethodInstance for f(::Int64))

julia> SnoopCompileCore.@snoopi begin
           f(2)
       end
0-element Array{Tuple{Float64,Core.MethodInstance},1}

julia> 
```

If so, I think this is actually what we _want,_ because we want to know for a given run _exactly what ended up taking time in the compiler,_ not necessarily what functions are most expensive to compile from scratch (though that would _also_ be useful to know).

Is the above what you were referring to? Or were you referring to something else? Thanks!

NHDaly:
Question:
Is it expected that julia would be calling `infer()` on the same target function (even with the same `caller`) multiple times, inside `typeinf_edge()`?

Unless I've messed up my data collection, which is of course possible, in the example I just recorded it seems like there are hundreds of exact duplicate edges. I've added the logging inside `typeinf_edge`, on line 541, right before the call to `infer()`:
https://github.com/JuliaLang/julia/blob/be69119e4700da54f59e5d80731f0f2b10276283/base/compiler/typeinfer.jl#L525-L545

That's inside the body of `typeinf_edge`, whose signature is here:
https://github.com/JuliaLang/julia/blob/be69119e4700da54f59e5d80731f0f2b10276283/base/compiler/typeinfer.jl#L503

And when I collected the data, I found that within a single top-level call to `typeinf_ext_toplevel()`, we hit duplicate edges many times. For example, this edge shows up 144 times:
```julia
julia> [e for e in edges if last(e) == sorted_targets[1]]
144-element Vector{Pair{Core.MethodInstance,Core.MethodInstance}}:
 MethodInstance for (::Base.var"#printstyled##kw")(::NamedTuple{(:color,),Tuple{Symbol}}, ::typeof(printstyled), ::String) => MethodInstance for #printstyled#779(::Bool, ::Symbol, ::typeof(printstyled), ::String)
 MethodInstance for (::Base.var"#printstyled##kw")(::NamedTuple{(:color,),Tuple{Symbol}}, ::typeof(printstyled), ::String) => MethodInstance for #printstyled#779(::Bool, ::Symbol, ::typeof(printstyled), ::String)
 MethodInstance for (::Base.var"#printstyled##kw")(::NamedTuple{(:color,),Tuple{Symbol}}, ::typeof(printstyled), ::String) => MethodInstance for #printstyled#779(::Bool, ::Symbol, ::typeof(printstyled), ::String)
 MethodInstance for (::Base.var"#printstyled##kw")(::NamedTuple{(:color,),Tuple{Symbol}}, ::typeof(printstyled), ::String) => MethodInstance for #printstyled#779(::Bool, ::Symbol, ::typeof(printstyled), ::String)
 MethodInstance for (::Base.var"#printstyled##kw")(::NamedTuple{(:color,),Tuple{Symbol}}, ::typeof(printstyled), ::String) => MethodInstance for #printstyled#779(::Bool, ::Symbol, ::typeof(printstyled), ::String)
 MethodInstance for (::Base.var"#printstyled##kw")(::NamedTuple{(:color,),Tuple{Symbol}}, ::typeof(printstyled), ::String) => MethodInstance for #printstyled#779(::Bool, ::Symbol, ::typeof(printstyled), ::String)
 MethodInstance for (::Base.var"#printstyled##kw")(::NamedTuple{(:color,),Tuple{Symbol}}, ::typeof(printstyled), ::String) => MethodInstance for #printstyled#779(::Bool, ::Symbol, ::typeof(printstyled), ::String)
 MethodInstance for (::Base.var"#printstyled##kw")(::NamedTuple{(:color,),Tuple{Symbol}}, ::typeof(printstyled), ::String) => MethodInstance for #printstyled#779(::Bool, ::Symbol, ::typeof(printstyled), ::String)
 MethodInstance for (::Base.var"#printstyled##kw")(::NamedTuple{(:color,),Tuple{Symbol}}, ::typeof(printstyled), ::String) => MethodInstance for #printstyled#779(::Bool, ::Symbol, ::typeof(printstyled), ::String)
 MethodInstance for (::Base.var"#printstyled##kw")(::NamedTuple{(:color,),Tuple{Symbol}}, ::typeof(printstyled), ::String) => MethodInstance for #printstyled#779(::Bool, ::Symbol, ::typeof(printstyled), ::String)
 MethodInstance for (::Base.var"#printstyled##kw")(::NamedTuple{(:color,),Tuple{Symbol}}, ::typeof(printstyled), ::String) => MethodInstance for #printstyled#779(::Bool, ::Symbol, ::typeof(printstyled), ::String)
...

julia> unique([e for e in edges if last(e) == sorted_targets[1]])
1-element Vector{Pair{Core.MethodInstance,Core.MethodInstance}}:
 MethodInstance for (::Base.var"#printstyled##kw")(::NamedTuple{(:color,),Tuple{Symbol}}, ::typeof(printstyled), ::String) => MethodInstance for #printstyled#779(::Bool, ::Symbol, ::typeof(printstyled), ::String)
```

Out of the 6,212 edges I collected for one top-level function, there are only 145 _unique_ edges, and only 118 unique MethodInstances.

This also means that there are some MethodInstances being inferred from multiple parents. (There are 21 such cases, mostly with 2 callers, four with 3, and one with 5.)

----------

Is this expected? Based on reading the comments and the code, I thought that this should only be called if `mi` had not yet been inferred _and_ wasn't part of a cycle, which to me means that there shouldn't be duplicates?

Am I misunderstanding this?

Thanks! :) I figured I'd mention it here in case this is a bug / performance regression that could easily be fixed! :)

NHDaly:
And, actually, more evidence for the above:

For the method instance in question, `@snoopi` reported 2.6 seconds to infer in a cold process, but when I instead summed the time for `code_typed()` on _each inner MethodInstance encountered above_ (also on a cold process), starting from the leaves working all the way back up in a sorted DAG, the total time was only 0.42 seconds.

So I'm thinking that there's maybe a problem in the inference code, where it's not detecting duplicate nodes correctly, and it's redundantly inferring the same MethodInstances multiple times! 

@JeffBezanson: Maybe this is also something we can take a look at together tomorrow when we meet up? :)

PR #37547, opened 2020-09-12
fix sinc(Inf+Inf*im)

stevengj:
Fixes a regression caused by a typo in ~~37217~~ #37273: if the imaginary part of `x` is infinite, `sinc(x)` should not be zero and is generally just NaN.

(In principle, we could return a non-NaN complex infinity for `sinc(finite ± Inf*im)`, but I'm not sure it is worth the logic for that additional case.  In any case, that could be a separate PR since it is not a regression.)

### DIFF
--- a/base/special/trig.jl
+++ b/base/special/trig.jl
@@ -1079,10 +1079,8 @@ sinc(x::Integer) = iszero(x) ? one(x) : zero(x)
 _sinc(x::Number) = iszero(x) ? one(x) : isinf_real(x) ? zero(x) : sinpi(x)/(pi*x)
 _sinc_threshold(::Type{Float64}) = 0.001
 _sinc_threshold(::Type{Float32}) = 0.05f0
-@inline function _sinc(x::Union{T,Complex{T}}) where {T<:Union{Float32,Float64}}
-    a = fastabs(x)
-    return a < _sinc_threshold(T) ? evalpoly(x^2, (T(1), -T(pi)^2/6, T(pi)^4/120)) : isinf_real(a) ? zero(x) : sinpi(x)/(pi*x)
-end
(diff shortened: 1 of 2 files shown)


giordano:
> Fixes a regression caused by a typo in #37217

Reference to wrong pull request?

base/special/trig.jl
```diff
@@ -1079,10 +1079,8 @@ sinc(x::Integer) = iszero(x) ? one(x) : zero(x)
 _sinc(x::Number) = iszero(x) ? one(x) : isinf_real(x) ? zero(x) : sinpi(x)/(pi*x)
 _sinc_threshold(::Type{Float64}) = 0.001
 _sinc_threshold(::Type{Float32}) = 0.05f0
-@inline function _sinc(x::Union{T,Complex{T}}) where {T<:Union{Float32,Float64}}
-    a = fastabs(x)
-    return a < _sinc_threshold(T) ? evalpoly(x^2, (T(1), -T(pi)^2/6, T(pi)^4/120)) : isinf_real(a) ? zero(x) : sinpi(x)/(pi*x)
-end
+@inline _sinc(x::Union{T,Complex{T}}) where {T<:Union{Float32,Float64}} =
+    fastabs(x) < _sinc_threshold(T) ? evalpoly(x^2, (T(1), -T(pi)^2/6, T(pi)^4/120)) : isinf_real(x) ? zero(x) : sinpi(x)/(pi*x)
```
simeonschaub, line 1083: Am I overlooking something, or aren't these definitions exactly the same?

stevengj: `isinf_real(a)` was changed to `isinf_real(x)`

simeonschaub: Ah, I see! That was well hidden :smile:

PR #37548, opened 2020-09-12
Allow `clamp(n, 1:10)`

mcabbott:
This might sometimes be convenient. Matches `mod(n, 1:10)` but without wrapping around.

### DIFF
--- a/base/range.jl
+++ b/base/range.jl
@@ -1147,7 +1147,7 @@ mod(i::Integer, r::AbstractUnitRange{<:Integer}) = mod(i-first(r), length(r)) + 
 
 Clamp `x` to lie within range `r`.
 
-!!! compat "Julia 1.5"
-     This method requires at least Julia 1.5.
+!!! compat "Julia 1.6"
+     This method requires at least Julia 1.6.
 """
-clamp(x::Integer, r::AbstractUnitRange) = clamp(x, first(r), last(r))
+clamp(x::Integer, r::AbstractUnitRange{<:Integer}) = clamp(x, first(r), last(r))
--- a/base/range.jl
+++ b/base/range.jl
@@ -1141,13 +1141,3 @@ julia> mod(3, 0:2)
 """
 mod(i::Integer, r::OneTo) = mod1(i, last(r))
 mod(i::Integer, r::AbstractUnitRange{<:Integer}) = mod(i-first(r), length(r)) + first(r)
-
-"""
-    clamp(x::Integer, r::AbstractUnitRange)
-
-Clamp `x` to lie within range `r`.
-
-!!! compat "Julia 1.6"
-     This method requires at least Julia 1.6.
-"""
-clamp(x::Integer, r::AbstractUnitRange{<:Integer}) = clamp(x, first(r), last(r))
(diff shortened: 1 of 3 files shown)


base/range.jl
```diff
@@ -1087,3 +1087,13 @@ julia> mod(3, 0:2)
...
+
+"""
+    clamp(x::Integer, r::AbstractUnitRange)
+
+Clamp `x` to lie within range `r`.
+
+!!! compat "Julia 1.6"
+     This method requires at least Julia 1.6.
+"""
+clamp(x::Integer, r::AbstractUnitRange{<:Integer}) = clamp(x, first(r), last(r))
```
simeonschaub, line 1099: You need to add this inside the `Math` module in `base/math.jl`, since Base just imports `clamp` from `Math` and reexports it, so this just shadows `clamp` and doesn't overload it.

johnnychen94, line 1099: It's a bit weird to me that the convenient version of `clamp` and `mod` doesn't support `x::Real`.

```julia
julia> clamp(1.5, -1, 3)
1.5
```

simeonschaub: I don't really think it's weird, since `1.5` is clearly not in `-1:3`. I think it would be much more awkward if `UnitRange`s behaved like floating point intervals only in some contexts.

mcabbott: Oh that may explain all the bizarre test failures, thanks.

mcabbott:
November bump?

simeonschaub (approved):
Seems worthwhile to me.

StefanKarpinski:
Let me see if I have the correct mental model here. I'm picturing the unit range being thought of as tiling the integer line. So the integers are split into equivalence classes modulo `length(r)` and since the range contains one representative from each equivalence class, that's the element we map each integer to for this operation. Is that correct?

mcabbott:
I think that's the correct picture for `mod(n, 1:10)`. Thus `mod1(n, m) == mod(n, OneTo(m))`.

`clamp` is much simpler. But the motivation for this method is I guess similar, things like `clamp(i, axes(A,1))`.

StefanKarpinski:
Oh yeah, I'm not sure why I was thinking of `mod` here; this is dead obvious for `clamp`.

StefanKarpinski:
Rerunning CI to make sure.

StefanKarpinski:
This looks good to me. When CI passes, someone should squash merge this.

mcabbott:
Failure on "buildbot/tester_macos64" is:
```
Worker 2 failed running test InteractiveUtils:
Some tests did not pass: 124 passed, 0 failed, 4 errored, 0 broken.
InteractiveUtils: Error During Test at /Users/julia/buildbot/worker-tabularasa/tester_macos64/build/share/julia/stdlib/v1.5/InteractiveUtils/test/runtests.jl:447
  Test threw exception
  Expression: clipboard() == str
  failed process: Process(`pbpaste`, ProcessExited(1)) [1]
```

simeonschaub:
That failure looks unrelated to this change, but the analyzegc one is a bit odd. I rebased this, to see if that helps.

PR #37550, opened 2020-09-12
constrain private argument of propertynames to Bool

bkamins:
Currently you can write:
```
propertynames(x, "anything you want")
```
and it does not throw an error.

I propose a breaking (but I fell consistent with the docstring) change that `private` positional argument should be forced to be `Bool` by default.

The rationale is that users can be confused that the second positional argument of `propertynames` is allowed but is ignored if it is not `Bool`

### DIFF
--- a/base/reflection.jl
+++ b/base/reflection.jl
@@ -1465,7 +1465,7 @@ REPL tab completion on `x.` shows only the `private=false` properties.
 """
 propertynames(x) = fieldnames(typeof(x))
 propertynames(m::Module) = names(m)
-propertynames(x, private) = propertynames(x) # ignore private flag by default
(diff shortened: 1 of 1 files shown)


martinholters:
This will cause ambiguity errors for `propertynames(x::MyType, private)` methods (with unconstrained `private`) which might be out there.

bkamins:
This is a good point, but:
1. really the second argument should be `Bool` so I think these methods should be updated
2. before opening this I have checked and all `propertynames` functions in Julia Base and StdLib actually put `::Bool` constraint on the second argument.

(still - I agree this proposal is breaking)

JeffBezanson (approved):

bkamins:
Thank you!

PR #37557, opened 2020-09-13
Fix === and objectid of object with potentially undef inline immutable field

yuyichao:
An undef field should always be treated equal to another undef field of the same type
since there's no other way for the user to tell the difference between these.
These could previously cause inconsistent comparison results or crashes.

* Mark these types as `haspadding` so that they'll not hit the `memcmp` fast path.
* Make sure `jl_egal` and `jl_object_id_` doesn't read bits fields in undef inline immutable field
* Use `emit_getfield_knownidx` in `emit_bits_compare` so that the check can be done more easily

    Handle union fields of the same type in `emit_f_isa` to avoid regression.

* Allow input to `emit_f_isa` to be NULL and lazily emit NULL check if necessary

### DIFF
--- a/src/codegen.cpp
+++ b/src/codegen.cpp
@@ -2311,7 +2311,51 @@ static jl_cgval_t emit_getfield(jl_codectx_t &ctx, const jl_cgval_t &strct, jl_s
     return mark_julia_type(ctx, result, true, jl_any_type);
 }
 
-static Value *emit_box_compare(jl_codectx_t &ctx, const jl_cgval_t &arg1, const jl_cgval_t &arg2)
+template<typename Func>
+static Value *emit_guarded_test(jl_codectx_t &ctx, Value *ifnot, bool defval, Func &&func)
+{
+    BasicBlock *currBB = ctx.builder.GetInsertBlock();
+    BasicBlock *passBB = BasicBlock::Create(jl_LLVMContext, "guard_pass", ctx.f);
+    BasicBlock *exitBB = BasicBlock::Create(jl_LLVMContext, "guard_exit", ctx.f);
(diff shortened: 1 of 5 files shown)


yuyichao:
This is what led to https://github.com/JuliaLang/julia/pull/37511 and depends on https://github.com/JuliaLang/julia/pull/37511 . It came out of an attempt to reuse `getfield` in codegen for emitting `===` to make other optimizations easier which is why the fix is like this. It's probably possible to create a more backportable version without this part of the change (i.e. only touch `emit_bits_compare`) but it'll be more annoy to do. We'll see if this is too hard to backport = = ...

src/codegen.cpp
```diff
@@ -2564,31 +2567,38 @@ static Value *emit_f_is(jl_codectx_t &ctx, const jl_cgval_t &arg1, const jl_cgva
...
+            PHINode *cmp = ctx.builder.CreatePHI(T_int1, 2);
+            cmp->addIncoming(ConstantInt::get(T_int1, 0), currBB);
+            cmp->addIncoming(bitcmp, isaBB);
+            return cmp;
+        });
+    }
+
+    // TODO: handle the case where arg1.typ != arg2.typ, or when one of these isn't union,
+    //       or when the union can be pointer
+    if (arg1.TIndex && arg2.TIndex && arg1.typ == arg2.typ &&
```
vtjnash, line 2595:
```suggestion
    if (arg1.TIndex && arg2.TIndex && jl_egal(arg1.typ, arg2.typ) &&
```

vtjnash (approved):

src/codegen.cpp
yuyichao: I assume so is the assert in the callee.

PR #37559, opened 2020-09-13
Fix the performance of `reinterpretarray` with simultaneous reshaping

### LINKED ISSUE #28980: Performance of ReinterpretArray, continued
Although the problem described in #25014 was handled by #28707, there is more to be done.
To start, can the pattern of the latter PR be extended to `setindex`?

Example (based on [a regression in QuartzImageIO](https://github.com/JuliaImages/Images.jl/issues/743))

```julia
using BenchmarkTools, ImageCore, Colors, FixedPointNumbers

const old = VERSION < v"0.7"

function fillcolor1!(buffer::AbstractArray{T, 3}, imgsrc, nc) where T
    imwidth, imheight = size(buffer, 2), size(buffer, 3)
    pixelptr = pointer(imgsrc)
    if old
        imbuffer = unsafe_wrap(Array, pixelptr, (nc, imwidth, imheight), false)
    else
        imbuffer = unsafe_wrap(Array, pixelptr, (nc, imwidth, imheight), own=false)
    end
    buffer[:, :, :] = imbuffer
end

function fillcolor2!(buffer::AbstractArray{T, 3}, imgsrc, nc) where T
    imwidth, imheight = size(buffer, 2), size(buffer, 3)
    pixelptr = pointer(imgsrc)
    if old
        imbuffer = unsafe_wrap(Array, pixelptr, (nc, imwidth, imheight), false)
    else
        imbuffer = unsafe_wrap(Array, pixelptr, (nc, imwidth, imheight), own=false)
    end
    buffer .= imbuffer
end

function runme()
    T = N0f8
    sz = (512,512)
    sz3 = (3,sz...)
    imgsrc = rand(T,sz3)
    if old
        buf1 = Array{RGB{T}}(sz)
    else
        buf1 = Array{RGB{T}}(undef,sz)
    end
    buf2 = similar(buf1)
    buf1r = reshape(reinterpret(T, buf1), (3, sz...))
    buf2r = reshape(reinterpret(T, buf1), (3, sz...))
    println("buf2r is a ",typeof(buf2r))
    @btime fillcolor1!($buf1r, $imgsrc, 3)
    @btime fillcolor2!($buf2r, $imgsrc, 3)
    buf2r == buf1r
end
```

v0.6.4:
```
buf2r is a Array{FixedPointNumbers.Normed{UInt8,8},3}
  908.177 μs (8 allocations: 224 bytes)
  47.144 μs (2 allocations: 112 bytes)
true

```

v0.7.0:
```
buf2r is a Base.ReshapedArray{Normed{UInt8,8},3,Base.ReinterpretArray{Normed{UInt8,8},2,RGB{Normed{UInt8,8}},Array{RGB{Normed{UInt8,8}},2}},Tuple{}}
  505.656 ms (7863300 allocations: 275.98 MiB)
  3.866 ms (2 allocations: 112 bytes)
true
```

v1.1.0-DEV.140:
```julia
buf2r is a Base.ReshapedArray{FixedPointNumbers.Normed{UInt8,8},3,Base.ReinterpretArray{FixedPointNumbers.Normed{UInt8,8},2,ColorTypes.RGB{FixedPointNumbers.Normed{UInt8,8}},Array{ColorTypes.RGB{FixedPointNumbers.Normed{UInt8,8}},2}},Tuple{}}
  1.031 s (7863300 allocations: 275.98 MiB)
  2.497 ms (2 allocations: 112 bytes)
true
```

Are the two versions of `fillarray` guaranteed to have equivalent results? If so, could the first now be made to throw?

timholy:
This introduces `reinterpret(reshape, T, A)`. This adds a dimension when reinterpreting to a smaller element size (e.g., `Tuple{Int,Int}->Int`), removes a dimension when reinterpreting to a bigger element size, and keeps the dimensionality constant when the element size is unchanged.  Here's a demo:
```julia
julia> a = [(1,2), (3,4)]
2-element Vector{Tuple{Int64, Int64}}:
 (1, 2)
 (3, 4)

julia> reinterpret(reshape, Int, a)       # bigger elsize to smaller elsize, adds a dimension
2×2 reinterpret(reshape, Int64, ::Vector{Tuple{Int64, Int64}}):
 1  3
 2  4

julia> b = [1 2; 3 4]
2×2 Matrix{Int64}:
 1  2
 3  4

julia> reinterpret(reshape, Tuple{Int,Int}, b)    # smaller elsize to bigger elsize, removes a dimension
2-element reinterpret(reshape, Tuple{Int64, Int64}, ::Matrix{Int64}):
 (1, 3)
 (2, 4)

julia> c = [1, -1]
2-element Vector{Int64}:
  1
 -1

julia> reinterpret(reshape, UInt, c) # same elsize, no change in dimensionality
2-element reinterpret(reshape, UInt64, ::Vector{Int64}):
 0x0000000000000001
 0xffffffffffffffff
```
Obviously, the first dimension must have size commensurate with the difference in element size.

This scenario is widely used:
- it's used when transitioning between a `Vector` of `SVector`s and a matrix where each vector is on a column
- it's used in splitting and combining color channels in image processing

Unfortunately, since Julia 1.0 the performance of ReinterpretArray has been quite poor in these circumstances. For those who want benchmarks, see https://github.com/JuliaImages/ImageCore.jl/pull/142; the short answer is that the improvements are dramatic, and come close to erasing any difference between the performance of a copy and the view itself (in some cases, even providing performance improvements). The trick here is that LLVM knows the size of the first dimension and so can often unroll the inner loop.

This replaces the strategy I was aiming for in #37290 by reshaping the array. I definitely think this is a better solution (thanks as always @mbauman, who has a gift for brief but incredibly useful suggestions). This is split into two commits: the first is pretty straightforward and, I think, relatively risk-free. The second is much scarier: it attempts to retain the performance advantages of linear indexing when the parent array supports it, but introduces a new `AbstractCartesianIndex` subtype. I *think* I've prevented packages from getting borked by not supporting this index type (see the new `WrappedArray` tests), but there's no doubt that this is the part that deserves the closest scrutiny.

Fixes #28980
Closes #37290

### DIFF
--- a/base/reinterpretarray.jl
+++ b/base/reinterpretarray.jl
@@ -175,8 +175,96 @@ function check_writable(a::ReinterpretArray{T, N, S} where N) where {T,S}
     end
 end
 
-IndexStyle(a::ReinterpretArray) = IndexStyle(a.parent)
-IndexStyle(a::ReshapedReinterpretArray{T, N, S}) where {T, N, S} = sizeof(T) < sizeof(S) ? IndexCartesian() : IndexStyle(a.parent)
+## IndexStyle specializations
+
+# For `reinterpret(reshape, T, a)` where we're adding a channel dimension and with
+# `IndexStyle(a) == IndexLinear()`, it's advantageous to retain pseudo-linear indexing.
+struct IndexSCartesian2{K,N} <: IndexStyle end   # K = sizeof(S) ÷ sizeof(T), a static-sized 2d cartesian iterator
+
+IndexStyle(::Type{ReinterpretArray{T,N,S,A,false}}) where {T,N,S,A<:AbstractArray{S,N}} = IndexStyle(A)
+function IndexStyle(::Type{ReinterpretArray{T,N,S,A,true}}) where {T,N,S,A<:AbstractArray{S}}
+    if sizeof(T) < sizeof(S)
+        IndexStyle(A) === IndexLinear() && return IndexSCartesian2{sizeof(S) ÷ sizeof(T),N}()
+        return IndexCartesian()
+    end
+    return IndexStyle(A)
+end
+IndexStyle(::IndexSCartesian2{K,N}, ::IndexSCartesian2{K,N}) where {K,N} = IndexSCartesian2{K,N}()
+
+struct SCartesianIndex2{K,N} <: AbstractCartesianIndex{N}
+    i::Int
+    j::Int
+end
+to_index(i::SCartesianIndex2) = i
+
+struct SCartesianIndices2{K,N,R<:AbstractUnitRange{Int}} <: AbstractMatrix{SCartesianIndex2{K}}
+    indices2::R
+end
+SCartesianIndices2{K,N}(indices2::AbstractUnitRange{Int}) where {K,N} = (@assert K::Int > 1; SCartesianIndices2{K,N,typeof(indices2)}(indices2))
+
+eachindex(::IndexSCartesian2{K,N}, A::AbstractArray) where {K,N} = SCartesianIndices2{K,N}(eachindex(IndexLinear(), parent(A)))
+
+size(iter::SCartesianIndices2{K}) where K = (K, length(iter.indices2))
+axes(iter::SCartesianIndices2{K}) where K = (Base.OneTo(K), iter.indices2)
+
+first(iter::SCartesianIndices2{K,N}) where {K,N} = SCartesianIndex2{K,N}(1, first(iter.indices2))
+last(iter::SCartesianIndices2{K,N}) where {K,N}  = SCartesianIndex2{K,N}(K, last(iter.indices2))
+
+@inline function getindex(iter::SCartesianIndices2{K,N}, i::Int, j::Int) where {K,N}
+    @boundscheck checkbounds(iter, i, j)
+    return SCartesianIndex2{K,N}(i, iter.indices2[j])
+end
+
+function iterate(iter::SCartesianIndices2{K,N}) where {K,N}
+    ret = iterate(iter.indices2)
+    ret === nothing && return nothing
+    item2, state2 = ret
+    return SCartesianIndex2{K,N}(1, item2), (1, item2, state2)
+end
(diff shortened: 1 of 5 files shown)


timholy:
CC @johnnychen94, @RalphAS

base/reinterpretarray.jl
```diff
@@ -105,31 +175,132 @@ function check_writable(a::ReinterpretArray{T, N, S} where N) where {T,S}
...
+function IndexStyle(::Type{ReinterpretArray{T,N,S,A,true}}) where {T,N,S,A<:AbstractArray{S}}
+    if sizeof(T) < sizeof(S)
+        IndexStyle(A) === IndexLinear() && return IndexSCartesian2{sizeof(S) ÷ sizeof(T),N}()
+        return IndexCartesian()
+    end
+    return IndexStyle(A)
+end
+IndexStyle(::IndexSCartesian2{K,N}, ::IndexSCartesian2{K,N}) where {K,N} = IndexSCartesian2{K,N}()
+
+struct SCartesianIndex2{K,N} <: AbstractCartesianIndex{N}
```
timholy, line 194: I'm unsure about the `N` parameter here. It felt slightly weird to have it be a subtype of generic `AbstractCartesianIndex`, and this does make it clear how many dimensions the particular `SCartesianIndex2` is standing in for. The downside is that this will force a lot of completely useless specialization.

timholy:
I'll probably merge this Monday. Any thoughts? I'm especially interested in feedback on https://github.com/JuliaLang/julia/pull/37559#discussion_r488570153.

timholy:
I decided to get rid of the non-functional type parameter. I'll merge this once it gets through CI.

timholy:
I decided to get rid of the non-functional type parameter. When we merge, though, let's not squash, so we can simply revert the 3rd commit if folks think that was a bad idea.

That's a lot of failures, but as far as I can tell they have nothing to do with this PR since several of my PRs are showing somewhat similar errors. OK to merge?

PR #37562, opened 2020-09-14
Revert "Improve typesubtract for tuples (#35600)"

### LINKED ISSUE #37448: Exploding inference time with `map(identity, zip(x...))`
In:
```
function main(N)
    t = (nothing, 1)
    arr = fill(t, N)
    map(identity, zip(arr...))
    return nothing
end

for i in 6:9
    println(i)
    @time main(i)
end
```
Out:
```
6
  0.264842 seconds (518.52 k allocations: 30.021 MiB, 4.09% gc time)
7
  0.543801 seconds (561.86 k allocations: 31.749 MiB, 1.60% gc time)
8
  2.800034 seconds (681.19 k allocations: 35.867 MiB)
9
 21.998138 seconds (1.10 M allocations: 49.225 MiB, 0.08% gc time)
```

This happens at least on the latest master and 1.5.

All that time goes into inference, it seems there's a huge stack of recursive calls to `typeof_tfunc`. More than that I haven't figured out.

vtjnash:
Closes #37448, reopens #35566

Uses of `switchtupleunion` need to be guarded in cost (e.g. with `countunionsplit`), while there's currently no practical way to limit this. The inference algorithm assumes that the canonical form is that of `tuplemerge`, and dislikes when we form the switched version implicitly, as it has a higher lattice cost, so it's expected to only appear explicitly.

### DIFF
--- a/base/compiler/typeutils.jl
+++ b/base/compiler/typeutils.jl
@@ -70,15 +70,6 @@ function typesubtract(@nospecialize(a), @nospecialize(b))
     if isa(a, Union)
         return Union{typesubtract(a.a, b),
                      typesubtract(a.b, b)}
-    elseif a isa DataType
-        if b isa DataType
-            if a.name === b.name === Tuple.name && length(a.types) == length(b.types)
-                ta = switchtupleunion(a)
-                if length(ta) > 1
-                    return typesubtract(Union{ta...}, b)
-                end
-            end
-        end
     end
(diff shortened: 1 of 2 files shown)


vtjnash:
The before (green) / after (blue) times here are pretty cool to see:
image

The #37714 PR will make them a bit worse than the blue again, but not much (not like the green line).

PR #37566, opened 2020-09-14
Update types.md

jonas-schulze:


jonas-schulze:
The majority of hyphenations in that file were of the form `\w -- \w` (inclusing spaces), so I changed the outliers. That's also the main form in the whole repo. If there is interest, I can update the others as well.

```bash
$ git grep "\w---\w" | wc -l # 6
$ git grep "\w--\w" | wc -l # 42
$ git grep "\w -- \w" | wc -l # 99
```

PR #37567, opened 2020-09-14
pass correct typeinfo IOContext when showing keys and values in dict

### LINKED ISSUE #37565: Vector element type not hidden in Dict printing for custom structs
Consider the difference in printing a dictionary with values as `Vector{Int}` or `Vector{Foo}` where `Foo` is a custom struct.

```jl
julia> Dict("a" => [1,2,3])
Dict{String, Vector{Int64}} with 1 entry:
  "a" => [1, 2, 3]

julia> struct Foo
          a::Int
       end

julia> Dict("a" => [Foo(1), Foo(2), Foo(3)])
Dict{String, Vector{Foo}} with 1 entry:
  "a" => Foo[Foo(1), Foo(2), Foo(3)]
```

In the latter case, there is an extra `Foo` in front of the vector, even though we know from the summary that we have `Vector{Foo}`. Perhaps this should be limited in the same way?

KristofferC:
Fixes https://github.com/JuliaLang/julia/issues/37565.

I'm not really familiar with the display system but seems to work?

### DIFF
--- a/test/show.jl
+++ b/test/show.jl
@@ -1387,6 +1387,19 @@ end
 
     d = Dict("+"=>1)
     @test showstr(d) == "Dict(\"+\" => 1)"
+
+    struct Foo
+        a::Int
+    end
+    struct Bar
+        a::Int
+    end
+    d = Dict([Bar(1), Bar(2)] => [Foo(1), Foo(2)])
+    m = string(@__MODULE__)
+    @test showstr(d) == "Dict{Vector{$m.Bar}, Vector{$m.Foo}}([$m.Bar(1), $m.Bar(2)] => [$m.Foo(1), $m.Foo(2)])"
+    @test sprint(show, MIME("text/plain"), d) == """
+        Dict{Vector{$m.Bar}, Vector{$m.Foo}} with 1 entry:
+          [Bar(1), Bar(2)] => [Foo(1), Foo(2)]"""
 end
 
 @testset "alignment for pairs" begin  # (#22899)
--- a/test/show.jl
+++ b/test/show.jl
@@ -1395,10 +1395,9 @@ end
         a::Int
     end
     d = Dict([Bar(1), Bar(2)] => [Foo(1), Foo(2)])
-    m = string(@__MODULE__)
-    @test showstr(d) == "Dict{Vector{$m.Bar}, Vector{$m.Foo}}([$m.Bar(1), $m.Bar(2)] => [$m.Foo(1), $m.Foo(2)])"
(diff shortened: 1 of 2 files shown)


test/show.jl
```diff
@@ -1387,6 +1387,19 @@ end
...
+        a::Int
+    end
+    struct Bar
+        a::Int
+    end
+    d = Dict([Bar(1), Bar(2)] => [Foo(1), Foo(2)])
+    m = string(@__MODULE__)
+    @test showstr(d) == "Dict{Vector{$m.Bar}, Vector{$m.Foo}}([$m.Bar(1), $m.Bar(2)] => [$m.Foo(1), $m.Foo(2)])"
+    @test sprint(show, MIME("text/plain"), d) == """
+        Dict{Vector{$m.Bar}, Vector{$m.Foo}} with 1 entry:
```
vtjnash, line 1401:
```suggestion
    @test showstr(d) == "Dict{Vector{$(curmod_prefix)Bar}, Vector{$(curmod_prefix)Foo}}([$(curmod_prefix)Bar(1), $(curmod_prefix)Bar(2)] => [$(curmod_prefix)Foo(1), $(curmod_prefix)Foo(2)])"
    @test sprint(show, MIME("text/plain"), d) == """
        Dict{Vector{$(curmod_prefix)Bar}, Vector{$(curmod_prefix)Foo}} with 1 entry:
```

vtjnash (approved):

test/show.jl
```diff
@@ -1387,6 +1387,18 @@ end
...
+
+    struct Foo
+        a::Int
+    end
+    struct Bar
+        a::Int
+    end
+    d = Dict([Bar(1), Bar(2)] => [Foo(1), Foo(2)])
+    @test showstr(d) == "Dict{Vector{$(curmod_prefix)Bar}, Vector{$(curmod_prefix)Foo}}([$(curmod_prefix)Bar(1), $(curmod_prefix)Bar(2)] => [$(curmod_prefix)Foo(1), $(curmod_prefix)Foo(2)])"
+    @test sprint(show, MIME("text/plain"), d) == """
```
rfourquet, line 1399: I think you can use the `replstr` function (defined at the top of "test/show.jl") for that

rfourquet (approved):

test/show.jl
KristofferC: Thanks, rerunning CI for it felt a bit inefficient so I merged as-is. But I can fix it up in the next I add in this file.

rfourquet: Sure, agreed, I will also try to remember about it when I touch this file.

PR #37568, opened 2020-09-14
RFC: handle colors in Dict limit printing

### LINKED ISSUE #44718: Showing `Dict`s with `printstyled` elements can color-bleed when truncated
The problem is best illustrated by example: define a type and overload `Base.show` for it, using `printstyled` to get some color in the print out:
```jl
struct StyledType
    x::Vector{Int}
end

function Base.show(io::IO, ct::StyledType)
    print(io, ct.x)
    printstyled(io, " +color", color=:red)
end

ct = StyledType(1:3)
```
where `ct` then prints as:
image

But if we now put `ct` elements into a `Dict`, we get issues with the color when the `Dict` entries' display are truncated:
```jl
d = Dict(n => fill(StyledType(1:n), n) for n = 1:5)
```
image

KristofferC:
When a dictionary prints values compactly, it cuts off the values after a certain number of characters and writes a couple of dots. In presence of colors, this doesn't work properly, because the computation of the width of the line will be wrong, and the printer might cut off the printing in the middle of the color sequence, corrupting the terminal color state. As an example, the following code

```jl
using Random

struct RainBowString
    s::String
end

function Base.show(io::IO, rbs::RainBowString)
    for s in rbs.s
        _, color = rand(Base.text_colors)
        print(io, color, s, "\e[0m")
    end
end


d = Dict([randstring(8) => [RainBowString(randstring(8)) for i in 1:10] for j in 1:5]...)
```

now shows in the REPL as

Screenshot-20200914144516-606x195

with this PR, it shows as:

Screenshot-20200914144438-784x148

The original reason for making this PR is that I made some struct print with color but then it printed as:

Screenshot-20200914144907-881x613

With this PR, it prints as:
Screenshot-20200914144854-917x617

Marking as RFC because I don't know if we want to deal with colors inside containers in the display system outside of `MIME("text/plain")`.
Haven't added tests yet because they are annoying to write and want to see that people are ok with this first.

Fixes https://github.com/JuliaLang/julia/issues/44718

### DIFF
--- a/base/show.jl
+++ b/base/show.jl
@@ -47,6 +47,58 @@ end
 show(io::IO, ::MIME"text/plain", c::ComposedFunction) = show(io, c)
 show(io::IO, ::MIME"text/plain", c::Returns) = show(io, c)
 
+const ansi_regex = r"\x1B(?:[@-Z\\-_]|\[[0-?]*[ -/]*[@-~])"
+# An iterator similar to `pairs` but skips over "tokens" corresponding to
+# ansi sequences
+struct IgnoreAnsiIterator
+    captures::Base.RegexMatchIterator
+end
+IgnoreAnsiIterator(s::AbstractString) =
+    IgnoreAnsiIterator(eachmatch(ansi_regex, s))
+
+Base.IteratorSize(::Type{IgnoreAnsiIterator}) = Base.SizeUnknown()
+function iterate(I::IgnoreAnsiIterator, (i, m_st)=(1, iterate(I.captures)))
+    # Advance until the next non ansi sequence
+    if m_st !== nothing
+        m, j = m_st
+        if m.offset == i
+            i += sizeof(m.match)
+            return iterate(I, (i, iterate(I.captures, j)))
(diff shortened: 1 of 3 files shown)


base/show.jl
```diff
@@ -47,6 +47,58 @@ end
...
+        end
+    end
+    ci = iterate(I.captures.string, i)
+    ci === nothing && return nothing
+    i_prev = i
+    (c, i) = ci
+    return (i_prev => c), (i, m_st)
+end
+
+function _truncate_at_width_or_chars(io::IO, str, width, chars="", truncmark="…")
```
vtjnash, line 76: I am not sure it is worthwhile to compile this function for every `io` type it might encounter (which is usually on the order of 5 with various IOContext/TTY/IOBuffer combinations)
```suggestion
function _truncate_at_width_or_chars(ignore_ansi::Bool, str, width, chars="", truncmark="…")
```

KristofferC: Good point. Updated.

PR #37572, opened 2020-09-14
fix #37558, allow optimizing calls to functions that throw

### LINKED ISSUE #37558: performance gain when kernel operation is wrapped with a function
This originates from https://github.com/JuliaArrays/OffsetArrays.jl/issues/136#issuecomment-691699933

```julia
function overflow_check(r, offset::T) where T
    if offset > 0 && last(r) > typemax(T) - offset
        throw(ArgumentError("Boundary overflow detected: offset $offset should be equal or less than $(typemax(T) - last(r))"))
    end
end

function overflow_check_with_func(r, offset::T) where T
    throw_overflow_error() = throw(ArgumentError("Boundary overflow detected: offset $offset should be equal or less than $(typemax(T) - last(r))"))
    if offset > 0 && last(r) > typemax(T) - offset
        throw_overflow_error()
    end
end
```

These two works equivalently, but their performance is slightly different:

```julia
julia> r, offset = -3:3, -4;

julia> @btime overflow_check($r, $offset)
  4.801 ns (0 allocations: 0 bytes) # 1.6.0-DEV.883
  4.078 ns (0 allocations: 0 bytes) # 1.4.0, 1.5.1

julia> @btime overflow_check_with_func($r, $offset)
  3.765 ns (0 allocations: 0 bytes) # 1.6.0-DEV.883
  1.696 ns (0 allocations: 0 bytes) # 1.5.1
  1.919 ns (0 allocations: 0 bytes) # 1.4.0
```

I'm really unsure what an appropriate title should be, please rename it if that's unclear.

cc: @jishnub @stevengj

JeffBezanson:
This gives up a tiny bit of latency, but I suspect it's worth it to avoid inserting avoidable GC frames. Comparing against  8320fcc7d90ad2523a8f3a3410e7e2d6ac1aa1bf (the commit that introduced `find_throw_blocks`) as a baseline, this change adds ~2% to the TTFP. On master, it actually reduces the time to precompile Plots.jl by ~2% (don't know TTFP since it doesn't work currently). So overall the latency cost looks minimal.

### DIFF
--- a/base/compiler/utilities.jl
+++ b/base/compiler/utilities.jl
@@ -269,7 +269,7 @@ function find_throw_blocks(code::Vector{Any}, ir = RefValue{IRCode}())
                     push!(stmts, i)
                 end
             elseif s.head === :return
-            elseif is_throw_call(s) || s.head === :unreachable
+            elseif is_throw_call(s)
                 if try_depth == 0
                     push!(stmts, i)
                 end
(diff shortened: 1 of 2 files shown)


vtjnash (approved):
I suppose we should remember to revert this when we implement gc-frame minimization?

Needs a fix to the doctests

PR #37573, opened 2020-09-14
fix #37540, interpolation in quoted exprs returned from macros

### LINKED ISSUE #37540: Macro hygiene broken for interpolation into quoted expressions
MWE:
```julia
julia> macro foo()
           quote
               x = 1
               :($x)
           end
       end
@foo (macro with 1 method)

julia> @macroexpand @foo
quote
    #= REPL[40]:3 =#
    var"#58#x" = 1
    #= REPL[40]:4 =#
    x
end
```

JeffBezanson:
I suspect we added this extra rule to macro hygiene to be nice to macros like our own `@eval` that failed to wrap their output with `esc`. So this could be considered a bug fix, but like all such things, at the end of the day it just changes how macros work and so we might not be able to do it. PkgEval will be interesting here.


base/essentials.jl
```diff
@@ -216,10 +216,10 @@ Evaluate an expression with values interpolated into it using `eval`.
 If two arguments are provided, the first is the module to evaluate in.
 """
 macro eval(ex)
-    :(Core.eval($__module__, $(Expr(:quote,ex))))
+    :(Core.eval($__module__, $(Expr(:escape, Expr(:quote,ex)))))
```
vtjnash, line 219: Would this be simpler:
```suggestion
    return esc(Expr(:call, GlobalRef(Core, :eval), __module__, Expr(:quote, ex)))
```

base/essentials.jl
```diff
@@ -216,10 +216,10 @@ Evaluate an expression with values interpolated into it using `eval`.
...
 end
 macro eval(mod, ex)
-    :(Core.eval($(esc(mod)), $(Expr(:quote,ex))))
+    :(Core.eval($(esc(mod)), $(Expr(:escape, Expr(:quote,ex)))))
```
vtjnash, line 222:
```suggestion
     return esc(Expr(:call, GlobalRef(Core, :eval), mod, Expr(:quote, ex)))
```

vtjnash (approved):
I agree PkgEval may be necessary here, but hopefully this can be just a bug fix

JeffBezanson:

nanosoldier:
[Your package evaluation job](https://github.com/JuliaLang/julia/pull/37573#issuecomment-693133968) has completed - possible new issues were detected. A full report can be found [here](https://github.com/JuliaCI/NanosoldierReports/blob/c90e333603b5c02ebbd5fae167f939ca004c1e0e/pkgeval/by_hash/264b2a4_vs_8bdf569/report.md). cc @maleadt

simeonschaub:
I haven't gone through all the failures, but the issues mostly just seem to be in MacroTools.

vtjnash:
Marking this for triage, to decide if this is v2 or acceptable to break now

JeffBezanson:
Yeah, this seems quite breaking. I'm pretty sure the change is "the right thing" but it will likely have to wait for 2.0 (where we will want to do a significant redesign of macros for other reasons as well).

simeonschaub:
I stole @KristofferC's PkgEvalAnalysis notebook to get a better understanding of why packages were failing tests. Here is what I found (Notebook can be found [here](https://github.com/simeonschaub/PkgEvalAnalysis/blob/analyze_37540/analysis.ipynb)):
```
 (223, MacroTools)
 (9, Unknown)
 (5, TestStdlib)
 (5, CanonicalTraits)
 (2, QuickTypes)
 (2, PairwiseListMatrices)
 (1, TensorKit)
 (1, PDDL)
 (1, JuliaInterpreter)
 (1, AstroTime)
 (1, ApproxError)
```
So it's really only 8 packages and the `Test` stdlib that are broken because of this, the rest are all just dependencies of these packages. I have looked into all the `Unknown` ones, and am pretty sure those were just either caused by RNG noise or were broken the last time already. The biggest one of them by far is MacroTools, for which I should already have [a fix](https://github.com/FluxML/MacroTools.jl/pull/147). #37641 should fix the 5 failures caused by the `Test` stdlib. It shouldn't be too hard to fix the 7 remaining packages and I would be willing to look into them. The question is of course still whether this would then qualify it as a minor change, or if we're too afraid of how much private code this could break.

JeffBezanson:
Excellent work, many thanks!!

JeffBezanson:
Should add a news item: macros that return `:quote` expressions previously did not need to escape them, but now they do. This is backwards-compatible as well so triage is ok with it.

vtjnash:
I think that's forward-compatible, not backwards? For complete accuracy, this is not completely whatever-compatible, but is likely not impactful (the variation in behavior only matters for macros returned from macros that expand to include interpolation)

JeffBezanson:
I mean that macros rewritten to work with this change will also work in older julia versions.

timholy:
The analysis in https://github.com/JuliaLang/julia/pull/37573#issuecomment-694795596 seems like it must have missed some packages. I see that JuliaInterpreter was listed as an unknown breakage, and that is correct, but there are independent breakages in LoweredCodeUtils and Revise too. I'm in the process of fixing these, but there's a chance we underestimated the impact of this change. (Still, I like the change!)

aviatesk:
maybe, the dependency from MacroTools.jl to Lazy.jl is another example; this slightly changes the behaviour of `MacroTools.@capture` and it causes precompilation failure in Lazy.jl, which uses `@capture` macro (well, I've already fixed by this PR: https://github.com/FluxML/MacroTools.jl/pull/149)

So I guess the analysis somehow missed to track the influence in dependencies during test ?

> (Still, I like the change!)

totally agree with this though ;)

Issue #37532, opened 2020-09-11
ERROR: type UnionAll has no field size


jagot:
New exciting error, when [upgrading CompactBases](https://github.com/JuliaApproximation/CompactBases.jl/pull/22) for new versions of QuasiArrays/ContinuumArrays (sorry I could not make the MWE smaller):

```julia
using CompactBases

function crash_inference(fun::Function, Ns)
    errors = map(Ns) do N
        R,μ = fun(N)
        [1.0*μ 1.0*μ 1.0*μ]
    end
end

k = 3
a,b = -1.0, 1.0
Ns = ceil.(Int, 10 .^ range(2, stop=3, length=3))

crash_inference(Ns) do N
    t = LinearKnotSet(k, a, b, N)
    BSpline(t), 1
end
```

TOMLs attached: [crash-inference.zip](https://github.com/JuliaLang/julia/files/5210550/crash-inference.zip)

```
Internal error: encountered unexpected error in runtime:
ErrorException("type UnionAll has no field size")
jl_errorf at /Users/julia/buildbot/worker/package_macos64/build/src/rtutils.c:77
jl_field_index at /Users/julia/buildbot/worker/package_macos64/build/src/datatype.c:1001
jl_f_getfield at /Users/julia/buildbot/worker/package_macos64/build/src/builtins.c:780
intrinsic_nothrow at ./compiler/tfuncs.jl:1483
stmt_effect_free at ./compiler/ssair/queries.jl:30
maybe_erase_unused! at ./compiler/ssair/ir.jl:1175
maybe_erase_unused! at ./compiler/ssair/ir.jl:1170 [inlined]
simple_dce! at ./compiler/ssair/ir.jl:1262
finish at ./compiler/ssair/ir.jl:1285 [inlined]
compact! at ./compiler/ssair/ir.jl:1299
compact! at ./compiler/ssair/ir.jl:1296 [inlined]
run_passes at ./compiler/ssair/driver.jl:140
optimize at ./compiler/optimize.jl:174
typeinf at ./compiler/typeinfer.jl:33
typeinf_edge at ./compiler/typeinfer.jl:484
abstract_call_method at ./compiler/abstractinterpretation.jl:419
abstract_call_gf_by_type at ./compiler/abstractinterpretation.jl:111
abstract_call_known at ./compiler/abstractinterpretation.jl:904
abstract_call at ./compiler/abstractinterpretation.jl:926
abstract_call at ./compiler/abstractinterpretation.jl:911
[...]

PR #37575, opened 2020-09-15
inference: fix bitcast nothrow computation

vtjnash:
Attempted fix for #37532

This was suggested in one of the issue comments there, but also fails one of my added tests. There's also a bit of a problem with 7466d1168ea that may be relevant here: most users of this function aren't actually enforcing the constraint at runtime that's assumed by that PR. We should fix that too, so that the PR there is only relevant to the call to this function where that applies. I mention this here, since it may change how we want to define `isexact` for the different sorts of callers.


test/compiler/inference.jl
```diff
@@ -2776,3 +2776,10 @@ for badf in [getfield_const_typename_bad1, getfield_const_typename_bad2]
     @test code[end] === Core.ReturnNode()
     @test_throws TypeError badf()
 end
+
+# issue #37532
+@test Core.Compiler.intrinsic_nothrow(Core.bitcast, Any[Type{Ptr{Int}}, Int])
+@test Core.Compiler.intrinsic_nothrow(Core.bitcast, Any[Type{Ptr{T}} where T, Ptr])
+@test_broken !Core.Compiler.intrinsic_nothrow(Core.bitcast, Any[Type{Ptr}, Ptr])
```
martinholters, line 2783: Do you have a fix for that in the pipeline? Otherwise, just adding `isconcretetype(ty)` to the condition might be like a viable band-aid? It pessimizes cases like the one above, but it used to throw for those before, so still an improvement.

JeffBezanson (approved):

PR #37578, opened 2020-09-15
Remove unnecessary lines for reduce size

ghost:
Hi JuliaDev,

I reduce a base file with differences 0,9%

### DIFF
--- a/base/abstractarray.jl
+++ b/base/abstractarray.jl
@@ -457,7 +457,6 @@ end
 size_to_strides(s, d) = (s,)
 size_to_strides(s) = ()
 
-
 function isassigned(a::AbstractArray, i::Integer...)
     try
         a[i...]
@@ -870,7 +869,6 @@ end
 """
     copyto!(dest::AbstractArray, src) -> dest
 
-
 Copy all elements from collection `src` to array `dest`, whose length must be greater than
 or equal to the length `n` of `src`. The first `n` elements of `dest` are overwritten,
 the other elements are left untouched.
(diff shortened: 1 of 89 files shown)


vtjnash:
These blank lines are present for improved visual readability, and so removing them doesn't improve anything.

bahasalien:
in CS a whitespace in code actually reducing speed for intrepreting a code which one unicode taking 4 bytes. a dev must taking a note for julia motto as we know "speed" just changed to a "readable code*

a julia dev choice not mine.

oscardssmith:
Parsing is `O(n)` compared to everything else which is way quicker. This is in no way a bottleneck.

bahasalien:
xaxaxa okay whatever u said buddy. idk, which method you are used as basic. the publisher must checked twice on your book.

timholy:
@bahasalien, please maintain a respectful tone. You might also consider taking the comments you're receiving seriously:
```julia
julia> function parseall(str)
           i, n = 1, length(str)
           while i < n
               _, i = Meta.parse(str, i; raise=false)
           end
           return i
       end
parseall (generic function with 1 method)

julia> @btime read("abstractarray.jl", String);
  11.216 μs (15 allocations: 74.05 KiB)

julia> str = read("abstractarray.jl", String);

julia> @btime parseall(str);
  58.525 ms (13432 allocations: 716.36 KiB)
```
Once parsing is complete, all spaces are removed. Your change might shave a couple of microseconds off Julia's own build time, but no one will notice such a small change. Conversely, humans read the code and will notice the loss of readability.

PR #33021, opened 2019-08-22
rework basename (fixes #33000)

johnnychen94:
fixes #33000

The current implementation of `basename` isn't consistent to [POSIX basename](http://pubs.opengroup.org/onlinepubs/9699919799/functions/basename.html)

cc: @rfourquet @sjoelund

PR #37580, opened 2020-09-15
Add documentation clarifying trailing slashes for basename

### LINKED ISSUE #33000: consistent result of `basename` for dir path
Although `basename(path)` is designed to split file name out from `path`, it doesn't check if the path is a valid file. That's to say, we can still use it to split folder name out from `path` as well.

Since there're two valid usages of dir path, it's expected to get the same results:

```julia
julia> basename("~/.ssh/")
"" # should be ".ssh"

julia> basename("~/.ssh")
".ssh"
```

This is tested on `1.0.4`, `1.1.1`, and `1.2.0`

musm:
ref comments here: https://github.com/JuliaLang/julia/issues/33000#issuecomment-692485584

close https://github.com/JuliaLang/julia/issues/33000

alternative to https://github.com/JuliaLang/julia/pull/33021

### DIFF
--- a/base/path.jl
+++ b/base/path.jl
@@ -165,10 +165,17 @@ See also: [`basename`](@ref)
 
 Get the file name part of a path.
 
+Note that the result of this function is different from the Unix `basename` program, where trailing
+slashes are ignored, i.e. `basename` for `/foo/bar/` returns `bar`, whereas  the `basename` function returns
+an empty string `""`.
+
 # Examples
 ```jldoctest
 julia> basename("/home/myuser/example.jl")
 "example.jl"
+
+julia> basename("/home/myuser/")
+""
 ```
 
 See also: [`dirname`](@ref)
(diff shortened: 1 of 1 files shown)


musm:
bump

base/path.jl
```diff
@@ -165,10 +165,17 @@ See also: [`basename`](@ref)
 
 Get the file name part of a path.
 
+Note that the result of this function is different from the Unix `basename` program, where trailing
+slashes are ignored, i.e. `basename` for `/foo/bar/` returns `bar`, whereas  the `basename` function returns
+an empty string `""`.
```
fredrikekre, line 170:
```suggestion
!!! note
    This function differs slightly from the Unix `basename` program, where trailing
    slashes are ignored, i.e. `$ basename /foo/bar/` returns `bar`, whereas  Julia returns an empty string `""`.
```

musm: I guess that didn't work because it thinks it's interpolation syntax? 
https://build.julialang.org/#/builders/39/builds/3391/steps/3/logs/stdio

musm:
merging sans objections in a few days, in other words we are not going to rework basename to ignore trailing slashes (out behavior matches Python, but not Posix)

PR #37581, opened 2020-09-15
Fixed link to external dependencies

rashidrafeek:


StefanKarpinski:
Thanks! This should be a relative link, not an absolute link to the github copy of the doc—github is not the primary place where docs are viewed.

rashidrafeek:
Oh. I have made it a relative link now.

PR #37584, opened 2020-09-15
Add a couple more REPL signature types

timholy:
These prevent some invalidations from poor inference on `keys`.

### DIFF
--- a/stdlib/REPL/src/LineEdit.jl
+++ b/stdlib/REPL/src/LineEdit.jl
@@ -1512,9 +1512,10 @@ function fixup_keymaps!(dict::Dict{Char,Any}, level, s, subkeymap)
     nothing
 end
 
-function add_specialisations(dict::Dict{Char,Any}, subdict, level)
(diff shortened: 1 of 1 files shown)


stdlib/REPL/src/LineEdit.jl
```diff
@@ -1512,9 +1512,10 @@ function fixup_keymaps!(dict::Dict{Char,Any}, level, s, subkeymap)
     nothing
 end
 
-function add_specialisations(dict::Dict{Char,Any}, subdict, level)
+function add_specialisations(dict::Dict{Char,Any}, subdict::Dict{Char,Any}, level::Int)
     default_branch = subdict[wildcard]
     if isa(default_branch, Dict)
+        default_branch = default_branch::Dict{Char,Any}
```
timholy, line 1518: It's a little weird to do it this way, but I was worried that putting the concrete type in the `isa` might cause a behavior change for any package callers there might be out there. So this is an intermediate step that will force them to toe the line and then we can do it "right" later.

KristofferC (approved):

Issue #35663, opened 2020-04-30
Conflicting environments are handled incorrectly


timholy:
I'm reporting this to julia itself (rather than in Pkg.jl) because key components of the fix seemingly must be made in `base/loading.jl`. No objections, though, if folks want to transfer this to Pkg.jl.

This bug comes in two closely-related flavors.

### Flavor 1: switching between incompatible environments should be disallowed

Suppose I have two projects with incompatible `[compat]` requirements:

`Proj1.toml`:
```toml
name = "Proj1"
uuid = "61c9ae6d-79d8-47c4-9a39-97bb6365db78"
authors = ["Tim Holy <tim.holy@gmail.com>"]
version = "0.1.0"

[deps]
ColorTypes = "3da002f7-5984-5a60-b8a6-cbb66c0b333f"

[compat]
ColorTypes = "0.8"
```

`Proj2.toml`:
```toml
name = "Proj2"
uuid = "7bf079ca-24d6-42d5-9ea6-67ba343487f2"
authors = ["Tim Holy <tim.holy@gmail.com>"]
version = "0.1.0"

[deps]
ColorTypes = "3da002f7-5984-5a60-b8a6-cbb66c0b333f"
Colors = "5ae59095-9a9b-59fe-a467-6f913c188581"

[compat]
ColorTypes = "0.10"
Colors = "0.12"
```

Now let's use these in a Julia session:
```julia
tim@diva:/tmp/envs/Proj1$ julia --project
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.4.2-pre.0 (2020-04-15)
 _/ |\__'_|_|_|\__'_|  |  release-1.4/ef4fe83698* (fork: 122 commits, 121 days)
|__/                   |

julia> using ColorTypes

shell> cd ../Proj2/
/tmp/envs/Proj2

(Proj1) pkg> activate .
 Activating environment at `/tmp/envs/Proj2/Project.toml`

julia> using Colors
[ Info: Precompiling Colors [5ae59095-9a9b-59fe-a467-6f913c188581]
ERROR: LoadError: UndefVarError: XRGB not defined
Stacktrace:
[...]

PR #37586, opened 2020-09-15
store the path to where a package got loaded

### LINKED ISSUE #33210: pathof(::Module) doesn't show the path where the module actually got imported from

KristofferC:
And use it in `pkgdir`

Fixes https://github.com/JuliaLang/julia/issues/33210

### DIFF
--- a/base/loading.jl
+++ b/base/loading.jl
@@ -250,7 +250,9 @@ to get the file name part of the path.
 function pathof(m::Module)
     pkgid = get(Base.module_keys, m, nothing)
     pkgid === nothing && return nothing
-    return Base.locate_package(pkgid)
+    origin = get(Base.pkgorigins, pkgid, nothing)
+    origin === nothing && return nothing
+    return origin.path
 end
 
 """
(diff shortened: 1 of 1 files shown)


timholy (approved):

timholy:
So cool! Is #35663 next?

KristofferC:
What behavior do you want there? A warning when the package would have gotten loaded from a different place than if it was not already loaded?

timholy:
Personally, I want an error---don't let the user change environments.

KristofferC:
Storing the path here is an improvement but it isn't really perfect. For example, if I make a sysimage, send it to a friend and he uses it, loads the same version of a package that is in the sysimage, that will have a different path, but it is stil the same package version. What we should store is probably more akin to "how this package got loaded". And there are a few ways a package can get loaded:

- by `git-tree-sha1`
- by explicit path
- by implicit path due to the folder being in `LOAD_PATH` (how stdlibs are loaded).

In the case above, the same `git-tree-sha1` would be used so even if the absolute paths is resolved differently, it is still not a problem.

timholy:
Seems like this must have missed a case:
```julia
julia> using JuliaInterpreter

julia> pathof(JuliaInterpreter)
"/home/tim/.julia/packages/JuliaInterpreter/muncv/src/JuliaInterpreter.jl"

julia> exit()
tim@diva:~/.julia/dev/Revise$ julia-master --startup-file=no -q
julia> using Revise

julia> using JuliaInterpreter

julia> pathof(JuliaInterpreter)

julia>
```

KristofferC:
Thanks, will look at it.

PR #37591, opened 2020-09-15
avoid type-specialization in show-default

vtjnash:
Fixes the data pointer for conversion of Ref{Any} to Ptr{Cvoid} to point
at the data, instead of the pointer. Then use that in show_default to
avoid specializing the Ref(Value) types of each input (causing poor
inference of unsafe_convert later).

Refs #37582

### DIFF
--- a/base/refpointer.jl
+++ b/base/refpointer.jl
@@ -24,6 +24,15 @@ If `T` is a bitstype, `isassigned(Ref{T}())` will always be true.
 
 When passed as a `ccall` argument (either as a `Ptr` or `Ref` type), a `Ref`
 object will be converted to a native pointer to the data it references.
+For most `T`, or when converted to a `Ptr{Cvoid}`, this is a pointer to the
+object data. When `T` is an `isbits` type, this value may be safely mutated,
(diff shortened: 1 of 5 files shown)


JeffBezanson:
Seems like this could potentially break some C interop corner cases? Might need a news entry or more documentation.

vtjnash:
I don't think it was intended for `unsafe_convert(Ptr{Cvoid}, Ref{Any}(1))` to give you a pointer to the pointer inside the Ref instead of the data pointer itself? That feels like an oddly specific thing to depend upon (usually our pun on `Cvoid` here is to give you a pointer to the data, ignoring the type—though it does lead to awkward edge cases like this).

KristofferC:
Just confirming that this does reduce a bunch of invalidations.

KristofferC:
What tests are needed here? @JeffBezanson

JeffBezanson:
Apparently `unsafe_convert` previously returned the wrong pointer in some cases, so there is a bugfix here that should get a test.

vtjnash:
updated docs and test, as I realized some other things had gotten a bit out of sync

PR #37595, opened 2020-09-15
Add code loading support for `Preferences` package

staticfloat:
**EDIT**: This PR has now changed to add only the code-loading portion.

This commit adds the `Preferences` standard library; a way to store a
TOML-serializable dictionary into top-level `Project.toml` files, then
force recompilation of child projects when the preferences are modified.

This pull request adds the `Preferences` standard library, which does
the actual writing to `Project.toml` files, as well as modifies the
loading code to check whether the preferences have changed.

### DIFF
--- a/base/loading.jl
+++ b/base/loading.jl
@@ -322,6 +322,31 @@ function manifest_deps_get(env::String, where::PkgId, name::String, cache::TOMLC
     return nothing
 end
 
+function uuid_in_environment(project_file::String, uuid::UUID, cache::TOMLCache)
+    # First, check to see if we're looking for the environment itself
+    proj_uuid = get(parsed_toml(cache, project_file), "uuid", nothing)
+    if proj_uuid !== nothing && UUID(proj_uuid) == uuid
+        return true
+    end
+
+    # Check to see if there's a Manifest.toml associated with this project
+    manifest_file = project_file_manifest_path(project_file, cache)
+    if manifest_file === nothing
+        return false
+    end
+    manifest = parsed_toml(cache, manifest_file)
+    for (dep_name, entries) in manifest
+        for entry in entries
+            entry_uuid = get(entry, "uuid", nothing)::Union{String, Nothing}
+            if uuid !== nothing && UUID(entry_uuid) == uuid
+                return true
+            end
+        end
+    end
+    # If all else fails, return `false`
+    return false
+end
+
 function manifest_uuid_path(env::String, pkg::PkgId, cache::TOMLCache)::Union{Nothing,String}
     project_file = env_project_file(env)
     if project_file isa String
@@ -950,7 +975,7 @@ function _require(pkg::PkgId, cache::TOMLCache)
             if (0 == ccall(:jl_generating_output, Cint, ())) || (JLOptions().incremental != 0)
                 # spawn off a new incremental pre-compile task for recursive `require` calls
                 # or if the require search declared it was pre-compiled before (and therefore is expected to still be pre-compilable)
-                cachefile = compilecache(pkg, path)
+                cachefile = compilecache(pkg, path, cache)
                 if isa(cachefile, Exception)
                     if precompilableerror(cachefile)
                         verbosity = isinteractive() ? CoreLogging.Info : CoreLogging.Debug
@@ -1195,7 +1220,7 @@ end
 @assert precompile(create_expr_cache, (PkgId, String, String, typeof(_concrete_dependencies), Bool))
 @assert precompile(create_expr_cache, (PkgId, String, String, typeof(_concrete_dependencies), Bool))
 
-function compilecache_path(pkg::PkgId)::String
+function compilecache_path(pkg::PkgId, cache::TOMLCache)::String
     entrypath, entryfile = cache_file_entry(pkg)
     cachepath = joinpath(DEPOT_PATH[1], entrypath)
     isdir(cachepath) || mkpath(cachepath)
@@ -1205,6 +1230,7 @@ function compilecache_path(pkg::PkgId)::String
         crc = _crc32c(something(Base.active_project(), ""))
         crc = _crc32c(unsafe_string(JLOptions().image_file), crc)
         crc = _crc32c(unsafe_string(JLOptions().julia_bin), crc)
+        crc = _crc32c(get_preferences_hash(pkg.uuid, cache), crc)
         project_precompile_slug = slug(crc, 5)
         abspath(cachepath, string(entryfile, "_", project_precompile_slug, ".ji"))
     end
@@ -1218,18 +1244,17 @@ This can be used to reduce package load times. Cache files are stored in
 `DEPOT_PATH[1]/compiled`. See [Module initialization and precompilation](@ref)
 for important notes.
 """
-function compilecache(pkg::PkgId, cache::TOMLCache = TOMLCache())
+function compilecache(pkg::PkgId, cache::TOMLCache = TOMLCache(), show_errors::Bool = true)
     path = locate_package(pkg, cache)
     path === nothing && throw(ArgumentError("$pkg not found during precompilation"))
-    return compilecache(pkg, path)
+    return compilecache(pkg, path, cache, show_errors)
 end
 
 const MAX_NUM_PRECOMPILE_FILES = 10
 
-# `show_errors` is an "internal" interface for Pkg.precompile
-function compilecache(pkg::PkgId, path::String, show_errors::Bool = true)
+function compilecache(pkg::PkgId, path::String, cache::TOMLCache = TOMLCache(), show_errors::Bool = true)
     # decide where to put the resulting cache file
-    cachefile = compilecache_path(pkg)
+    cachefile = compilecache_path(pkg, cache)
     cachepath = dirname(cachefile)
     # prune the directory with cache files
     if pkg.uuid !== nothing
@@ -1333,6 +1358,8 @@ function parse_cache_header(f::IO)
         end
         totbytes -= 4 + 4 + n2 + 8
     end
+    prefs_hash = read(f, UInt64)
+    totbytes -= 8
     @assert totbytes == 12 "header of cache file appears to be corrupt"
     srctextpos = read(f, Int64)
     # read the list of modules that are required to be present during loading
@@ -1345,7 +1372,7 @@ function parse_cache_header(f::IO)
         build_id = read(f, UInt64) # build id
         push!(required_modules, PkgId(uuid, sym) => build_id)
     end
-    return modules, (includes, requires), required_modules, srctextpos
+    return modules, (includes, requires), required_modules, srctextpos, prefs_hash
 end
 
 function parse_cache_header(cachefile::String; srcfiles_only::Bool=false)
@@ -1354,21 +1381,21 @@ function parse_cache_header(cachefile::String; srcfiles_only::Bool=false)
         !isvalid_cache_header(io) && throw(ArgumentError("Invalid header in cache file $cachefile."))
         ret = parse_cache_header(io)
         srcfiles_only || return ret
-        modules, (includes, requires), required_modules, srctextpos = ret
+        modules, (includes, requires), required_modules, srctextpos, prefs_hash = ret
         srcfiles = srctext_files(io, srctextpos)
         delidx = Int[]
         for (i, chi) in enumerate(includes)
             chi.filename ∈ srcfiles || push!(delidx, i)
         end
         deleteat!(includes, delidx)
-        return modules, (includes, requires), required_modules, srctextpos
+        return modules, (includes, requires), required_modules, srctextpos, prefs_hash
     finally
         close(io)
     end
 end
 
 function cache_dependencies(f::IO)
-    defs, (includes, requires), modules = parse_cache_header(f)
+    defs, (includes, requires), modules, srctextpos, prefs_hash = parse_cache_header(f)
     return modules, map(chi -> (chi.filename, chi.mtime), includes)  # return just filename and mtime
 end
 
@@ -1383,7 +1410,7 @@ function cache_dependencies(cachefile::String)
 end
 
 function read_dependency_src(io::IO, filename::AbstractString)
-    modules, (includes, requires), required_modules, srctextpos = parse_cache_header(io)
+    modules, (includes, requires), required_modules, srctextpos, prefs_hash = parse_cache_header(io)
     srctextpos == 0 && error("no source-text stored in cache file")
     seek(io, srctextpos)
     return _read_dependency_src(io, filename)
@@ -1428,6 +1455,37 @@ function srctext_files(f::IO, srctextpos::Int64)
     return files
 end
 
+# Find the Project.toml that we should load/store to for Preferences
+function get_preferences_project_path(uuid::UUID, cache::TOMLCache = TOMLCache())
+    for env in load_path()
+        project_file = env_project_file(env)
+        if !isa(project_file, String)
+            continue
+        end
+        if uuid_in_environment(project_file, uuid, cache)
+            return project_file
+        end
+    end
+    return nothing
+end
+
+function get_preferences(uuid::UUID, cache::TOMLCache = TOMLCache();
+                         prefs_key::String = "compile-preferences")
+    project_path = get_preferences_project_path(uuid, cache)
+    if project_path !== nothing
+        preferences = get(parsed_toml(cache, project_path), prefs_key, Dict{String,Any}())
+        if haskey(preferences, string(uuid))
+            return preferences[string(uuid)]
+        end
+    end
+    # Fall back to default value of "no preferences".
+    return Dict{String,Any}()
+end
+get_preferences_hash(uuid::UUID, cache::TOMLCache = TOMLCache()) = UInt64(hash(get_preferences(uuid, cache)))
+get_preferences_hash(m::Module, cache::TOMLCache = TOMLCache()) = get_preferences_hash(PkgId(m).uuid, cache)
+get_preferences_hash(::Nothing, cache::TOMLCache = TOMLCache()) = UInt64(hash(Dict{String,Any}()))
+
+
 # returns true if it "cachefile.ji" is stale relative to "modpath.jl"
 # otherwise returns the list of dependencies to also check
 stale_cachefile(modpath::String, cachefile::String) = stale_cachefile(modpath, cachefile, TOMLCache())
@@ -1438,7 +1496,7 @@ function stale_cachefile(modpath::String, cachefile::String, cache::TOMLCache)
             @debug "Rejecting cache file $cachefile due to it containing an invalid cache header"
             return true # invalid cache file
         end
-        (modules, (includes, requires), required_modules) = parse_cache_header(io)
+        modules, (includes, requires), required_modules, srctextpos, prefs_hash = parse_cache_header(io)
         id = isempty(modules) ? nothing : first(modules).first
         modules = Dict{PkgId, UInt64}(modules)
 
@@ -1514,6 +1572,12 @@ function stale_cachefile(modpath::String, cachefile::String, cache::TOMLCache)
         end
 
         if isa(id, PkgId)
+            curr_prefs_hash = get_preferences_hash(id.uuid, cache)
+            if prefs_hash != curr_prefs_hash
+                @debug "Rejecting cache file $cachefile because preferences hash does not match 0x$(string(prefs_hash, base=16)) != 0x$(string(curr_prefs_hash, base=16))"
+                return true
+            end
+
             get!(PkgOrigin, pkgorigins, id).cachepath = cachefile
         end
 
--- a/base/util.jl
+++ b/base/util.jl
@@ -385,6 +385,8 @@ _crc32c(io::IO, crc::UInt32=0x00000000) = _crc32c(io, typemax(Int64), crc)
 _crc32c(io::IOStream, crc::UInt32=0x00000000) = _crc32c(io, filesize(io)-position(io), crc)
 _crc32c(uuid::UUID, crc::UInt32=0x00000000) =
     ccall(:jl_crc32c, UInt32, (UInt32, Ref{UInt128}, Csize_t), crc, uuid.value, 16)
+_crc32c(x::Integer, crc::UInt32=0x00000000) =
+    ccall(:jl_crc32c, UInt32, (UInt32, Vector{UInt8}, Csize_t), crc, reinterpret(UInt8, [x]), sizeof(x))
 
 """
     @kwdef typedef
--- a/src/dump.c
+++ b/src/dump.c
@@ -1123,6 +1123,31 @@ static int64_t write_dependency_list(ios_t *s, jl_array_t **udepsp, jl_array_t *
             write_int32(s, 0);
         }
         write_int32(s, 0); // terminator, for ease of reading
+
+        // Calculate Preferences hash for current package.
+        jl_value_t *prefs_hash = NULL;
+        if (jl_base_module) {
+            // Toplevel module is the module we're currently compiling, use it to get our preferences hash
+            jl_value_t * toplevel = (jl_value_t*)jl_get_global(jl_base_module, jl_symbol("__toplevel__"));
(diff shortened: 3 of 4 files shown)


base/loading.jl
```diff
@@ -1411,6 +1413,20 @@ function srctext_files(f::IO, srctextpos::Int64)
     return files
 end
 
+function get_preferences_hash(uuid::UUID, cache::TOMLCache = TOMLCache())
+    # check that project preferences match by first loading the Project.toml
+    active_project_file = Base.active_project()
+    if isfile(active_project_file)
+        preferences = get(parsed_toml(cache, active_project_file), "preferences", Dict{String,Any}())
```
tkf, line 1420: Sorry if I missed earlier discussion on this. But wouldn't this behave a bit confusingly when you stack environments?

Suppose I have `LOAD_PATH = ["@", "@v#.#", "@stdlib"]` and some preferences for SomeREPLUtilities.jl in `~/.julia/environments/v1.6/Project.toml`. If I activate `$PWD/Project.toml` and then evaluate `using SomeREPLUtilities`, does it mean the preferences in `~/.julia/environments/v1.6/Project.toml` would be ignored?  How about looking at all project files in `Base.load_path()`?

fredrikekre: It should *only* check the project file where `SomeREPLUtilities` is found I guess. If you need local preferences for that package you should probably add it to the local project.

tkf: Ah, yes, good point. I agree.

staticfloat: It's purposeful that it only checks the top-level project file.  We decided against any kind of merging for a few reasons:

1) Reproducibility; if a Project.toml somewhere high up in a chain is effecting things, then simply saving the `Project.toml` and `Manifest.toml` that you normally do would not be enough to guarantee identical results later on; we want all the configuration stored in one place.

2) Performance; we don't want to have to touch a bunch of files.  This can be ameliorated with caches, of course, but it's nice that we only have to read in one file.

3) Merging complexity; since we're storing full-on `Dict`'s here, what happens if a higher-up and lower-down Projects have conflicts around key mappings?

I'm torn on adding preferences for packages not in the project; it makes sense to disallow it, but it's also harmless.

JeffreySarnoff: I favor disallowing the inclusion of preferences for packages not in the project simply because allowing it allows a[two] non-involved package[s] to carry preferences that may conflict were that[those] package[s] to appear in the project at a future time. It works against clear communication in collaboration.

fredrikekre: 1. But if you use a package not in the current project you can't reproduce anyway so saving preferences for a non-existing package does not really help anything?
2. I don't think this will have any impact at all, we still go to that same file to find the UUID of the package (and with the new TOML caching we  have alraedy parsed that file).
3. That is a good point, you mean if some pakcage X has config in multiple places? I would say just use the first one then.

tkf: @staticfloat Thanks for explaining the context. Just to be explicit, I'm not suggesting to look into package's project file (in the sense `joinpath(pkgdir(MyPkg), "Project.toml")`) and also I'm not suggesting to do any "deep merge." Concretely, I'm suggesting something like

```julia
function get_preferences_hash(uuid::UUID, cache::TOMLCache = TOMLCache())
    # check that project preferences match by first loading the Project.toml
    for path in Base.load_path()
        if isfile(path)
            project = parsed_toml(cache, path)
            # `uuid` is a dependency:
            any(values(get(project, "deps", Dict{String,Any}()))) do uuid
                tryparse(UUID, uuid) == uuid
            end || continue
            # Now we are looking at correct Project.toml; let's load `preferences`
            preferences = get(project, "preferences", Dict{String,Any}())
            if haskey(preferences, string(uuid))
                return UInt64(hash(preferences[string(uuid)]))
            end
            break  # should we?
        end
    end
    return UInt64(hash(Dict{String,Any}()))
end
```

It'd only look up just a few files in `Base.load_path()` (typically one or two, I think). So, I'm not sure if there is too much performance overhead.

I think my implementation addresses the points raised by @fredrikekre.

> 2\. (and with the new TOML caching we  have alraedy parsed that file)

@fredrikekre I guess caching is not useful during precompilation since it'd happen in a separate process? Though I agree that parsing a few files instead of just one file does not seem to be a big impact on the performance.

fredrikekre: > I think my implementation addresses the points raised by @fredrikekre.

Yea, that is what I meant -- find the file where the package would be `using`d from.

> I guess caching is not useful during precompilation since it'd happen in a separate process?

True, I was only thinking of performance for getting/setting prefs.

StefanKarpinski: It might be fine to allow preferences for any package as long as you take preferences from the environment in whose manifest the package first appears. The reason that's ok is because that guarantees that preferences later in the load path can only affect packages that nothing earlier in the load path depends on. To spell that out, if A is a top-level package in the first environment in the load path and it depends on B and B has no preferences set in the first environment (because, e.g it is not a top-level package), then we need to load B with no preferences set. If, on the other hand, C is in the second environment and not depended on by anything in the first environment, then we can apply preferences for it from the second environment.

fredrikekre: Another point: If we *only* look in the top project file, then you need to configure the package in _every project you use it in_. E.g. if you have Revise in the global env, for every project you want to use Revise with you have to configure it again.

fredrikekre: > To spell that out, if A is a top-level package in the first environment in the load path and it depends on B and B has no preferences set in the first environment (because, e.g it is not a top-level package), then we need to load B with no preferences set. If, on the other hand, C is in the second environment and not depended on by anything in the first environment, then we can apply preferences for it from the second environment.

This is consistent with the idea that you load preferences from the project where the package is `using`d from, right? Since `B` would be found in the top manifest.

staticfloat: > Concretely, I'm suggesting something like

Ahhh, I understand what you're saying now.  Thanks, that was really helpful.  Yes, I think such an approach makes sense!  And I believe that Stefan's point is that the `break` in your `for` loop should exist, correct?

staticfloat:
There's another small design point I want to bring up; right now, _all_ preferences stored will force recompilation.  That may not be desirable in all cases, perhaps a package wants to store something that is only used at runtime (and hence does not need recompilation) but still wants it bundled into the `Project.toml`.  I'm not sure what the best API for this is.

One possibility is that we store all the compilation-time preferences under the `build_preferences` key and the non-compilation-time preferences under the `preferences` key in the TOML, then `loading.jl` only pays attention to `build_preferences`.  We would then, in `Preferences`, add a keyword argument to all the getters and setters that switches whether you're saving a build-time preference or not.  The part of this that is slightly awkward to me is that it feels strange to have two separate dictionaries that you have to manage, but perhaps that's not that awkward and is fine.  It's certainly better (from an implementation perspective) than tagging every key in a dictionary that is marked as `require-recompilation`, which is one of the the single-dict options.

fredrikekre:
IMO some system like that seem unnecessary complicated -- is it really so bad to reprecompile even though it will not change the result?

base/loading.jl
tkf: IIUC, there may be a subtle difference between what I suggested and what Stefan suggested. Suppose, as in Stefan's setting, I have:

* Packages: A, B
* Dependencies: A depends on B
* Environments (`LOAD_PATH`):
  * X: It contains A as a dependency (i.e., `deps` has `A = "u-u-i-d"`).
  * Y: It contains B as a dependency but A is not in the manifest. It has preferences for B.

With my suggestion, `using A` would load B with the preference stored in Y.  I cannot tell if B's preference in Y would be used in Stefan's approach.

The reason why I like my approach is that I think `using B` in the stacked environment X+Y should behaves as close as possible to single environment Y if B is installed explicitly only in Y (although it's not possible to do it always). Typically, I think Y is `@v#.#` and X is `@`. For example, if you have Revise.jl in `@#.#` and it has some preferences, you'd want to use the preferences recorded in `@#.#` even if you have a package that depends on Revise.jl in `@`.

(In the above setting `break`-or-not does not matter. So I'm not sure how it interacts with Stefan's suggestion.)

JeffreySarnoff:
Only if there is an undesired cost to that reprecompilation .. and the run-time only preferences are changing at a rate that exacerbates the disutility of that cost.  Is that something that is likely to occur now and again?

tkf:
How about adding an interface like `@get_preference(key, default)` (and recommend using it over `@load_preferences`)? This way, we can get a list of `key`s used during precompilation time. Maybe this is enough for optimizing unnecessary precompilation in the future?

staticfloat:
> is it really so bad to reprecompile even though it will not change the result?

Well, I can imagine there are users that want to do things like store ML model hyperparameters, or plotting backend (I'm assuming changing that can be done completely at runtime), or whatever.  There are all sorts of useful things to store, (and even ones that you might want to communicate to others with your `Project.toml`) and right now this is the closest thing to what a user might use to store preferences in their packages.

An alternative approach would be to write a second package that just gives a nice interface to `TOML` and `Scratch` to store non-compile-time preferences in a `Scratch` space.

KristofferC:
> Well, I can imagine there are users that want to do things like store ML model hyperparameters, or plotting backend (I'm assuming changing that can be done completely at runtime), or whatever. 

I tought one of the major points of this system is that you use it when the feature is required to be queried at compilation time. If that is not the case, you should probably use something else. E.g., storing hyper parameters in the Project files feels "wrong", that could go into a separate TOML file or something that you just load at runtime and pass via normal function arguments.

staticfloat:
I just got another question from a user that wanted to know if they should use `Preferences` to store things like whether a package had been run before, so they could give a one-time welcome message.  I think it's clear users want a key-value store and that they're not thinking about the fundamental capabilities of this package properly.

I suggest re-naming this to `CompiletimePreferences`, and we can create a separate `Preferences` package that is not a stdlib, that just stores preferences to a TOML file in a Scratch space.  How does that sound to everybody?

vchuravy:
Maybe instead of `CompiletimePreferences`, `ProjectPreferences`? E.g. noting where the preferences live? I don't like the moniker `Compiletime` since it is a very "in-group" name, and most users shouldn't worry about compiletime vs runtime. Other ideas are `Settings` or `CachedPreferences`.

tkf:
For storing software states like a welcome message, I think it's not ideal to contaminate user's Project.toml file. Isn't Scratch.jl the best way to do it?

But I think I agree separating compile-time and run-time key-value store is a clean design.

> I suggest re-naming this to `CompiletimePreferences`

Brain-storming other names: `CompileOptions`/`CompileConfigs`.

fredrikekre:
Some other ideas: PkgConfig, PackageConfiguration, ProjectConfig[uration], ProjectSettings

staticfloat:
> I don't like the moniker Compiletime since it is a very "in-group" name, and most users shouldn't worry about compiletime vs runtime.

I argue the only reason anyone should ever use this stdlib is because they (as a package developer) need to worry about compiletime vs runtime.  Anyone who doesn't care about the difference between runtime and compiletime either should use a different package (like `TOML`) or actually does need to care about the difference.

I think keeping the term `Compile` in there or maybe `Build` still communicates clearly that it's got a specific purpose.

* `BuildPreferences`/`BuildConfig`/`BuildOptions`
* `CompilePreferences`/`CompileConfig`/`CompileOptions`

I'm not so hot on `Package*` because that again feels a little too general.

tkf:
I think `Compile*` is better than `Build*` unless we can call `@load_preferences` in `deps/build.jl`.

KristofferC:
>  unless we can call @load_preferences in deps/build.jl.

We don't talk about that anymore.

KristofferC:
I also like to focus this package on compile-time preferences. However, some questions: For non-compile-time preferences, how does a user pass the preferences to the package? It's not via the project file so does the package have to write such an API then? Also, perhaps there are non-compile time preferences that you want to persist when you move your project to another machine. How would those be transferred?

tkf:
Sorry to mention the file-must-not-be-named :)

If we don't want to talk about it, I think `Compile*` is a better prefix.

> For non-compile-time preferences, how does a user pass the preferences to the package? It's not via the project file so does the package have to write such an API then?

I thought `Project.toml` would have two tables for two kinds of preferences as in

```toml
[deps]
PkgA = "342fba16-3e17-4664-b1bb-a60ccdbe268d"

[compile-preferences.342fba16-3e17-4664-b1bb-a60ccdbe268d]
compile_time_preference = "value"

[runtime-preferences.342fba16-3e17-4664-b1bb-a60ccdbe268d]
runtime_preference = "value"
```

staticfloat:
> unless we can call `@load_preferences` in `**************`.



> I thought `Project.toml` would have two tables for two kinds of preferences as in

Yes, I suggested this idea earlier and I think it's still my favorite idea.  I've moved this PR to store things under the key `compile-preferences`, and then `Preferences` can just be a normal package that stores things in `Project.toml` and that will _not_ trigger recompilation.

base/loading.jl
staticfloat: Ah, in that case, I think Stefan's approach makes more sense.

The reason being that we try hard not to allow things "lower down" in the stack to change the behavior of things at the top of the stack.  For instance, if I execute something where only `X` is in my load path, I will get the "no preferences" behavior, but if I execute the same project but with `Y` inserted underneath `X` in the stack, suddenly the behavior changes.  I think this is generally undesirable.

In the example of `Revise`, I think it makes sense to load the preferences for `Revise` from your global environment if `Revise` is not installed in your local package, but if it is, I do not think we should use the global environment's values.  One potential reason why is that if I were to set some custom options in my global environment, it then becomes difficult to have a "clean" environment where I get only the defaults without setting those defaults explicitly.

staticfloat: @KristofferC if I wanted to walk the `load_path()` and test at each point as to whether or not a particular UUID is contained within the manifest there, what's the best way to do that?  I don't want to check the `deps` section of the `Project.toml` because that only contains direct dependencies, I want to check everything in the Manifest.  Is there an already-loaded cache of that somewhere in `Base`, or should I just search for `Manifest.toml` files sitting within/next to the elements of `load_path()` and parse them directly?

staticfloat:
Alright, I think I've successfully implemented the desired functionality here; the preferences are now loaded from the first environment that knows what the given UUID is, and saving preferences goes to that same location, defaulting to the currently-active project in the event that you're using a UUID that you've never add'ed before.  This seemed safer to me than throwing an error, as it means that if users start add'ing/removing packages they're less likely to get strange errors, and it enables the usage of things like preferences with `@requires`, on the off-chance that someone sets a preference in a package before checking to see if it's actually a part of their project.

musm:
Wouldn't a name like `PkgPreferences` or `PkgConfig` be a lot more appropriate? `CompilePreferences` sounds to me  like it's related to static compilation  or sorts.

base/loading.jl
tkf: Hmm... My intuition is still that preference for B in Y should be loaded even if B is in X. This is because the fact that A depends on B is an implementation detail. For example, A may stop depending on B when changing from 1.1 to 1.2. Then `Pkg.up` on X (that removes B from X) suddenly changes the behavior of `using B` even when Y is completely untouched and the user doesn't know if B was in X. I think this behavior is counter-intuitive.

> it then becomes difficult to have a "clean" environment where I get only the defaults without setting those defaults explicitly

One solution may be to introduce a "dummy" environment (say) `@nopref` that "resets" preferences.  For example, `LOAD_PATH = ["@", "@nopref", "@#.#", "@stdlib"]` would only load preferences from `@`.  It's an additional complexity w.r.t implementation but I think it's much easier to control preferences this way.  Even with the current implementation, it's a bit hard to fully predict what preferences are used unless you know all the entries in the manifest files.

JeffreySarnoff:
> Wouldn't a name like `PkgPreferences` ..

Yes. Re-reading this thread, `ProjectPreferences` is even more accurate; although people are more familiar with Packages than Projects in Julia.

base/loading.jl
staticfloat: > This is because the fact that A depends on B is an implementation detail.

That implementation detail has big consequences though; think about it from a versioning perspective.  If `A` restricts package `B` to `v1.1` in environment `Y`, then even though environment `X` includes `B` as a direct dependency and would normally allow it to be any `v1.X` version, it will be restricted to `v1.1`.  In this manner, the behavior of `using B` is already changing based on transitive dependencies.

I'm not 100% convinced I'm right, but I'm still mostly convinced that apply preferences based on a UUID's presence in `Manifest.toml` is better than doing this based on presence in `Project.toml` (e.g. only for direct dependencies).  @StefanKarpinski are you still convinced?

fredrikekre: I agree with @tkf: If you need preferences on `B` in `Y` you can just add `B` such that it exists in the `Project.toml` file (https://github.com/JuliaLang/julia/pull/37595#discussion_r488979943).

I agree that it will be difficult to get a clean env though. But perhaps it is simple enough to play around with the LOAD_PATH then? I for one quite often use `export JULIA_LOAD_PATH=$(mktemp -d)` when I need to play around with a clean env.

fredrikekre: However, it does become a bit messy; when you load `A` and thus `B` is loaded implicitly, should you really go look for deps for `B` lower in the load path in a project that might have a different version of `B`?

tkf:
There will be another package named `Preferences` or `PkgPreferences` or `ProjectPreferences` or something else. As discussed above, we need two packages. One of which can be used during precompilation so I think it's very natural to have "Compile" in its name.

Even if the name `CompilePreferences` is not intuitive, I don't think that's a big problem. Presumably, people would then just use `Preferences`. If they use `Preferences` when they should be using `CompilePreferences`, it's pretty easy to detect this and emit a friendly warning to nudge the users to switch to `CompilePreferences`.

base/loading.jl
tkf: > In this manner, the behavior of `using B` is already changing based on transitive dependencies.

Yeah, I agree that's an issue. But isn't it an issue somewhat orthogonal to preferences? I think it'd be better to only support the case where the manifests are compatible and explicitly state that otherwise, the behavior is undefined. I think it's better to solve the issue with incompatible manifests with something like #32906.

But I can see that combining two manifests can change the version of a package in a manner incompatible with the stored preferences.

_If_ we can change how the package resolution works in a stacked environment, how about

1. When looking for a package to load, first try stacked *`Project.toml`* files. This way, the package version explicitly added by the user takes precedence over the indirect dependencies in the environments above.
2. Allow preferences to be set only when the package is added by the user (i.e., in `deps`).
3. For a given package, use the first `Project.toml` that contains the package as a dependency. If there are no preferences, use the default. (This is my implementation in https://github.com/JuliaLang/julia/pull/37595#discussion_r489018264)

I think the combination of these three ingredients ensure that preferences and package version are compatible while allowing the scenario I described in https://github.com/JuliaLang/julia/pull/37595#discussion_r489077392

tkf:
I started playing with this branch to see if I can do what I had in mind for PyCall.jl. So far I think it's working as expected:
https://github.com/JuliaPy/PyCall.jl/pull/835
https://github.com/tkf/PyPreferences.jl

KristofferC:
> There will be another package named Preferences or PkgPreferences or ProjectPreferences or something else. As discussed above, we need two packages. 

I didn't really understand why two packages are needed. In my opinion (and as was discussed on the triage call), all preference stuff can live in the same package, it doesn't have to be a stdlib, code loading just know that it has to recompile if things under `[compile-preferences]` change. It is the same way as how Pkg doesn't technically need to be a stdlib, or that other things that write project and manifests don't have to be a stdlib. They communicate via a "serialized format" (a file) and the file format has docs. `[compile-preferences]` is just one extra thing.

tkf:
I think many preference setup currently done with environment variables can be handled by "runtime-Preferences.jl" package. I don't think it's a good idea to re-compile packages when changing such runtime preferences. Grepping `julia`, some candidates are

```
JULIA_ERROR_COLOR
JULIA_WARN_COLOR
JULIA_INFO_COLOR
JULIA_STACKTRACE_EXPAND_BASEPATHS
JULIA_STACKTRACE_CONTRACT_HOMEDIR
JULIA_STACKTRACE_LINEBREAKS
JULIA_WORKER_TIMEOUT
JULIA_PKG_SERVER
JULIA_PKG_OFFLINE
JULIA_PKG_DEVDIR
JULIA_PKGRESOLVE_ACCURACY
JULIA_PKG_CONCURRENCY
JULIA_PKG_PRECOMPILE_AUTO
JULIA_NUM_PRECOMPILE_TASKS
```

(But depending on how the merging of runtime-Preferences.jl work, some of them may not be appropriate.)

A potential benefit of using runtime-Preferences.jl is that we can provide much better UI for configuring these values (e.g., using TerminalMenus.jl).

> They communicate via a "serialized format" (a file) and the file format has docs.

I agree that CompilePreferences package itself does not have to be in stdlib.

KristofferC:
>  I don't think it's a good idea to re-compile packages when changing such runtime preferences. 

Totally, agree but why does that require two packages?

tkf:
Ah, I think I see what you are getting at. Yeah, it should be fine as long as you can determine whether a given preference in the project is for compile-time or not. It's more about toml file schema than the fronted package(s).

KristofferC:
> Yeah, it should be fine as long as you can determine whether a given preference in the project is for compile-time or not. 

Yes, having two different tables for that seems pretty reasonable.

KristofferC:
The only slight annoyance is that the user has to know whether the preference is a compile-time or not. It would be nice if that is up to the package.

fredrikekre:
> The only slight annoyance is that the user has to know whether the preference is a compile-time or not. It would be nice if that is up to the package.

Isn't the idea that you use API functions for setting the variables? E.g. the examples here: https://github.com/JuliaLang/julia/pull/37595/files#diff-153066f1b4b0aa6caf26959fc7f1ce57

JeffreySarnoff:
keep it simple, so it easier for many to use well

tkf:
> Isn't the idea that you use API functions for setting the variables?

Yeah, I think the API functions would be the primary way to tweak the preferences. If you have something like

```toml
[compile-preferences.342fba16-3e17-4664-b1bb-a60ccdbe268d]
compile_time_preference = "value"
```

figuring out the UUID by hand is tedious.

KristofferC:
Being super lazy and not checking myself, are "global" preferences also available, for example

```
[compile-preferences]
debug = true
```

staticfloat:
> are "global" preferences also available, for example

I originally had those stored within a "global" UUID of `UUID(UInt128(0))`, and we could make that a convention.  You don't want to be able to get a top-level dict, as then you'll get the preferences for your sibling preferences as keys within the global dict.... easy to screw something up then.

base/loading.jl
tkf: Any update on this? I think falling back to the lower environment stack is a key feature. I think one of the important use-cases for CompilePreferences.jl is to capture machine-specific configuration such as system libraries and hardware properties. For this, `~/.julia/environments/v1.6/Project.toml` sounds like a good place to store such information. It doesn't make sense to share `Project.toml` with machine-specific values.

I think I need this for PyPreferences.jl https://github.com/JuliaPy/PyCall.jl/pull/835. If it is decided that it is not appropriate to add this to CompilePreferences.jl, I think I'll create SharedPreferences.jl to handle this use-case (e.g., by storing the preferences in `~/.julia/sharedpreferences/$UUID/v$major.toml` and call `include_dependency` on it).  I'll look up per-project preferences first via CompilePreferences.jl and then fallback to SharedPreferences.jl. The fact that it is implementable outside `Base` *could* mean that not adding this in `Base`/CompilePreferences.jl is a good idea.  But having many ways to store preferences may be confusing.  I'm not sure what's the best approach here.

staticfloat:
I've modified this PR to only include the `Project.toml` parsing and hashing parts in `loading.jl` and `dump.c`; the user-facing API is now in the new [`Preferences.jl` package](https://github.com/JuliaPackaging/Preferences.jl), let's move all API discussion to issues and pull requests to that package.

As a quick taste, it has two sets of bindings, one through `Preferences` and one through `Preferences.CompileTime`.

base/loading.jl
tkf: I opened an issue for this discussion https://github.com/JuliaLang/julia/issues/37791

PR #37596, opened 2020-09-16
Allow hiding stderr during compilecache

IanButterworth:
Required for the error handling in https://github.com/JuliaLang/Pkg.jl/pull/2021

cc. @tkf

### DIFF
--- a/base/loading.jl
+++ b/base/loading.jl
@@ -1152,7 +1152,7 @@ end
 @assert precompile(include_package_for_output, (PkgId,String,Vector{String},Vector{String},Vector{String},typeof(_concrete_dependencies),Nothing))
 @assert precompile(include_package_for_output, (PkgId,String,Vector{String},Vector{String},Vector{String},typeof(_concrete_dependencies),String))
 
-function create_expr_cache(pkg::PkgId, input::String, output::String, concrete_deps::typeof(_concrete_dependencies))
+function create_expr_cache(pkg::PkgId, input::String, output::String, concrete_deps::typeof(_concrete_dependencies), show_errors::Bool = true)
     rm(output, force=true)   # Remove file if it exists
     depot_path = map(abspath, DEPOT_PATH)
     dl_load_path = map(abspath, DL_LOAD_PATH)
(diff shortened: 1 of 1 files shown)


tkf:
Context (just for record): Current implementation of `Pkg.precompile` as of https://github.com/JuliaLang/Pkg.jl/pull/2018 tries to call `Base.compilecache` for all packages in Manifest.toml file. This has a problem because some packages may only be required in some specific platform/environment and the package author advises the downstream packages to hide `using` behind, e.g., `if Sys.iswindows()` https://github.com/JuliaLang/Pkg.jl/pull/2018#discussion_r488853897. So, I suggested https://github.com/JuliaLang/Pkg.jl/pull/2018#discussion_r489029683 to ignore precompilation error from indirect dependencies (which is implemented in https://github.com/JuliaLang/Pkg.jl/pull/2021 by @ianshmean). Doing this in a user-friendly manner requires us to silence messages from the subprocesses.  Note that https://github.com/JuliaLang/Pkg.jl/pull/2021 does not ignore precompilation error from direct dependencies (which may be caused by precompilation of indirect dependencies). So, the errors relevant to the user will be correctly surfaced.

(It means that `Pkg.precompile` has quadratic time-complexity when an error happens. I think it's reasonable to optimize for the happy path for now given that doing this "correctly" is presumably much more challenging https://github.com/JuliaLang/Pkg.jl/pull/2018#discussion_r489029683. Also, we can still use `using TheProblematicPackage` to get precompilation error in linear time.)

base/loading.jl
```diff
@@ -1210,7 +1210,7 @@ end
 
 const MAX_NUM_PRECOMPILE_FILES = 10
 
-function compilecache(pkg::PkgId, path::String)
+function compilecache(pkg::PkgId, path::String, show_errors::Bool = true)
```
tkf, line 1213: I think it'd be better to comment that `show_errors` is an "internal" interface for `Pkg.precompile`. It's a bit unclear why we have this from just looking at the code in Base.

tkf:
@c42f I think you have opinions/thoughts on nested error message "handling" (in this case just ignoring). I think it's an OK implementation (though obviously not super elegant) but I wonder if you have some alternative ideas.

c42f:
I haven't followed the larger story very closely but this seems fine to me, for now. Other options would probably be a lot more complicated.

tkf:
> Other options would probably be a lot more complicated.

Yeah, that's my thought, too. Thanks for taking a look at it.

tkf (approved):
LGTM. I think it's pretty safe to merge but I guess I'll wait for a few days just in case someone (ping @vtjnash and @JeffBezanson) wants to take a look at it.

staticfloat (approved):
Looks good to me! Rebase and I say merge.

ianshmean:
@staticfloat Rebased. Is the aarch64 issue relevant? `LoadError: LoadError: "hard kill repl test"`

Issue #37599, opened 2020-09-16
Typo in the "Workflow Tips" section of the manual


kernelmethod:
There are [some parts of the "workflow tips" section of the Julia manual](https://docs.julialang.org/en/v1.6-dev/manual/workflow-tips/#Revise-based-workflows) where there was clearly meant to be a code block, but the way the page is rendered, it ends up being inlined next to the Markdown on the page (see attached screenshot.)

![Screenshot_20200915_195035](https://user-images.githubusercontent.com/17100608/93282956-c2913600-f78c-11ea-8714-49949b0af71e.png)

PR #37600, opened 2020-09-16
Fix typo in "workflow tips" section of the manual (#37599)

kernelmethod:
Change section numbering so that code blocks render correctly, instead of appearing inline next to other parts of the documentation.

### DIFF
--- a/doc/src/manual/workflow-tips.md
+++ b/doc/src/manual/workflow-tips.md
@@ -83,35 +83,36 @@ the following modifications:
 1. Put your code in a module somewhere on your load path. There are
    several options for achieving this, of which two recommended choices are:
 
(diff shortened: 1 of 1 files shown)


kernelmethod:
For reference, here's how it looks before/after the change.

Before:

Screenshot_20200915_195035

After:

Screenshot_debdev_2020-09-15_21:18:53

fredrikekre:
Perhaps better with just a bullet list? Looks a little bit strange with two nested numbered lists.

kernelmethod:
Yeah, I suppose you're right. I've modified my branch to use bullets instead. How does this look?

Screenshot_debdev_2020-09-17_18:13:49

fredrikekre (approved):

PR #37601, opened 2020-09-16
For %d printf specifier and AbstractFloat, print as %0f

quinnj:
Alternative fix to #37554. As Jameson noted on that PR, the type
instability made the code pretty ugly and slow. An alternative solution
is to switch the `%d` format specifier to `%0f`, i.e. a fixed precision
of 0 for `AbstractFloat` arguments. The code here is indeed much cleaner
and more performant: several times faster than old Printf.

### DIFF
--- a/stdlib/Printf/src/Printf.jl
+++ b/stdlib/Printf/src/Printf.jl
@@ -38,8 +38,8 @@ Base.string(f::Spec{T}; modifier::String="") where {T} =
         f.precision == 0 ? ".0" : f.precision > 0 ? ".$(f.precision)" : "", modifier, char(T))
 Base.show(io::IO, f::Spec) = print(io, string(f))
 
-nonfinitefmt(s::Spec{T}) where {T} =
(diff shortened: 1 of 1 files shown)


quinnj:
FreeBSD failure looks unrelated

StefanKarpinski:
FWIW, the concept in the original code was that the format specifier only act as directions for how to format the value and have no effect on how to interpret it as a type since, unlike C, we already know that. Thus, writing `%d` means "print this number with no trailing digits" i.e. what `%0f` does. So 👍.

quinnj:
Yeah, once I went back to the old Printf code and realized they were just doing `%0f`, I realized that was a way simpler solution than trying to do the covert/coerce dance to integer.

StefanKarpinski:
It's also why I ultimately think that printf is a bad UI for a dynamic language: it's designed for C where the specifier has to encode both what the type of the argument is _and_ how to format it. In Julia, we only need to express the formatting part, which can be done more easily than what printf requires.

StefanKarpinski:
In 2.0 I'd very much like to kick Printf out of stdlibs and replace it with something better and simpler.

PR #37603, opened 2020-09-16
[BinaryPlatforms]: Add dict-like adapter for `Platform` objects

staticfloat:
This will make it a little bit easier to modify `Platform` objects

### DIFF
--- a/base/binaryplatforms.jl
+++ b/base/binaryplatforms.jl
@@ -135,7 +135,7 @@ tags(p::Platform) = p.tags
 # Make it act more like a dict
 Base.getindex(p::AbstractPlatform, k::String) = getindex(tags(p), k)
 Base.haskey(p::AbstractPlatform, k::String) = haskey(tags(p), k)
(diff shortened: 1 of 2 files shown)


base/binaryplatforms.jl
```diff
@@ -131,7 +131,11 @@ end
 
 # Other `Platform` types can override this (I'm looking at you, `AnyPlatform`)
 tags(p::Platform) = p.tags
+
+# Make it act more like a dict
+Base.getindex(p::AbstractPlatform, k::String) = getindex(tags(p), k)
 Base.haskey(p::AbstractPlatform, k::String) = haskey(tags(p), k)
+Base.setindex!(p::AbstractPlatform, v::String, k::String) = setindex!(tags(p), v, k)
```
fredrikekre, line 138: Should return `p`:
```suggestion
Base.setindex!(p::AbstractPlatform, v::String, k::String) = (setindex!(tags(p), v, k); p)
```

PR #37605, opened 2020-09-16
get rid of some invalidation sources in Artifacts

KristofferC:
Found with SnoopCompile and loading Plots.

Not sure if this is the best way to resolve them but it does the job. @timholy?

### DIFF
--- a/stdlib/Artifacts/src/Artifacts.jl
+++ b/stdlib/Artifacts/src/Artifacts.jl
@@ -172,6 +172,7 @@ function load_overrides(;force::Bool = false)
     end
 
     ARTIFACT_OVERRIDES[] = overrides
+    return overrides
 end
 
 # Helpers to map an override to an actual path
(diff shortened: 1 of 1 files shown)


stdlib/Artifacts/src/Artifacts.jl
```diff
@@ -270,7 +271,12 @@ function unpack_platform(entry::Dict, name::String, artifacts_toml::String)::Uni
     end
 
     # Collect all String-valued mappings in `entry` and use them as tags
-    tags = Dict(Symbol(k) => v for (k, v) in entry if isa(v, String))
```
timholy, line 273: `Symbol(k)::Symbol => v::String` might also work, but your version is good.

stdlib/Artifacts/src/Artifacts.jl
```diff
@@ -610,24 +620,24 @@ end
 # Support `AbstractString`s, but avoid compilers needing to track backedges for callers
 # of these functions in case a user defines a new type that is `<: AbstractString`
 with_artifacts_directory(f::Function, artifacts_dir::AbstractString) =
-    with_artifacts_directory(f, string(artifacts_dir))
+    with_artifacts_directory(f, String(artifacts_dir)::String)
```
KristofferC, line 623: These are just fixes so that the code adheres to what the comment claims it does.

timholy (approved):

staticfloat (approved):
Thanks!

staticfloat:
Looks like the keys to the dictionary passed to unpack_platform are not Strings, but possibly substrings.

KristofferC:
> Looks like the keys to the dictionary passed to unpack_platform are not Strings, but possibly substrings.

I tweaked things a bit.

Basically, we have a "funnel" where the top methods accept wide types and convert them to known concrete types for which the core implementations are defined on. This means that the core parts will not be invalidated since they have concrete type information while things should still accept generic types.

PR #37606, opened 2020-09-16
fix some invalidations in rpad/lpad

KristofferC:
These are often called with imprecise type information. Prevent inference from going down the rabbit hole too much.

### DIFF
--- a/base/strings/util.jl
+++ b/base/strings/util.jl
@@ -304,14 +304,15 @@ julia> lpad("March", 10)
 "     March"
 ```
 """
-lpad(s, n::Integer, p::Union{AbstractChar,AbstractString}=' ') = lpad(string(s), n, string(p))
+lpad(s, n::Integer, p::Union{AbstractChar,AbstractString}=' ') = lpad(string(s)::AbstractString, n, string(p))
 
 function lpad(
     s::Union{AbstractChar,AbstractString},
     n::Integer,
     p::Union{AbstractChar,AbstractString}=' ',
 ) :: String
-    m = signed(n) - length(s)
+    n = Int(n)::Int
+    m = signed(n) - Int(length(s))::Int
     m ≤ 0 && return string(s)
     l = length(p)
     q, r = divrem(m, l)
@@ -331,14 +332,15 @@ julia> rpad("March", 20)
 "March               "
 ```
 """
-rpad(s, n::Integer, p::Union{AbstractChar,AbstractString}=' ') = rpad(string(s), n, string(p))
+rpad(s, n::Integer, p::Union{AbstractChar,AbstractString}=' ') = rpad(string(s)::AbstractString, n, string(p))
 
 function rpad(
     s::Union{AbstractChar,AbstractString},
     n::Integer,
     p::Union{AbstractChar,AbstractString}=' ',
 ) :: String
-    m = signed(n) - length(s)
+    n = Int(n)::Int
+    m = signed(n) - Int(length(s))::Int
     m ≤ 0 && return string(s)
     l = length(p)
     q, r = divrem(m, l)


base/strings/util.jl
```diff
@@ -304,14 +304,15 @@ julia> lpad("March", 10)
...
-lpad(s, n::Integer, p::Union{AbstractChar,AbstractString}=' ') = lpad(string(s), n, string(p))
+lpad(s, n::Integer, p::Union{AbstractChar,AbstractString}=' ') = lpad(string(s)::AbstractString, n, string(p))
 
 function lpad(
     s::Union{AbstractChar,AbstractString},
     n::Integer,
     p::Union{AbstractChar,AbstractString}=' ',
 ) :: String
-    m = signed(n) - length(s)
+    n = Int(n)::Int
```
timholy, line 314: I don't think you need the assert at the end:
```julia
julia> f(n) = Int(n)

1-element Vector{Any}:
 CodeInfo(
1 ─ %1 = Main.Int(n)::Int64
└──      return %1
) => Int64
```
You do though, for the `length` call below.

You might alternatively write the method so it only takes `n::Int` and then do the conversion in a tiny stub method, but no big deal either way---this is not a huge function.

timholy (approved):

base/strings/util.jl
KristofferC: After loading `Plots`:

```
1-element Vector{Any}:
 CodeInfo(
1 ─ %1 = Main.Int(n)::Any
└──      return %1
) => Any
```

timholy: Ah yes, `MAX_METHODS` strikes again!

dkarrasch: For the mortals among us: is `Int(::Integer)` a pattern that should be turned into `Int(n)::Int` generally, or is there something specific to this place? I'm seeing `Int(n)` in `SparseArrays`, for instance.

KristofferC: This is my understanding:

If Julia infers a method with an imprecise type (for example `n::Integer`) then a call to `Int(n)` will infer differenly depending on the number of methods (and their return type) that are applicable. For example:

```
julia> f(n) = Int(n)

1-element Vector{Any}:
 CodeInfo(
1 ─ %1 = Main.Int(n)::Int64
└──      return %1
) => Int64

julia> using Plots

1-element Vector{Any}:
 CodeInfo(
1 ─ %1 = Main.Int(n)::Any
└──      return %1
) => Any

julia> methods(Int, (Integer,))
# 5 methods for type constructor:
[1] (::Type{T})(x::BigInt) where T<:Union{Int128, Int16, Int32, Int64, Int8} in Base.GMP at gmp.jl:356
[2] (::Type{IT})(x::GeometryTypes.OffsetInteger{O, T}) where {IT<:Integer, O, T<:Integer} in GeometryTypes at /Users/kristoffercarlsson/.julia/packages/GeometryTypes/gpBFp/src/faces.jl:15
[3] (::Type{IT})(x::GeometryBasics.OffsetInteger) where IT<:Integer in GeometryBasics at /Users/kristoffercarlsson/.julia/packages/GeometryBasics/CMxB4/src/offsetintegers.jl:37
[4] Int64(x::Union{Bool, Int32, Int64, UInt32, UInt64, UInt8, Int128, Int16, Int8, UInt128, UInt16}) in Core at boot.jl:737
[5] (::Type{T})(x::T) where T<:Number in Core at boot.jl:745
```

the additional two matching methods provided by GeometryTypes and GeometryBasics changed the result. This means that precompiled code that assumed that `Int(n::Integer)` gave an `Int` is possible no longer correct and has to be recompiled. So this is only a problem if both:

1. Methods are compiled with imprecise type information.
2. New methods are added that invalidate the assumptions made by the imprecise type information.

I don't think there needs to be any real changes in SparseArrays, I think most of that should be well inferred.

timholy: @dkarrasch this is only relevant when inference quality is poor. In SparseArrays I doubt it's that much of a problem. A lot of string-processing gets called on `d[key]` where `valtype(d) === Any`, and that means it has to be robust wrt minimal information from inference.

PR #37611, opened 2020-09-16
deprecate Base.download in favor of Downloads.download

StefanKarpinski:
Based on https://github.com/JuliaLang/julia/pull/37340, this PR deprecates `Base.download` while keeping tests working. The remaining work to do here is to change [LibCURL](https://github.com/JuliaWeb/LibCURL.jl) to use [Downloads](https://github.com/JuliaLang/Downloads.jl) for its tests, once the [registration](https://github.com/JuliaRegistries/General/pull/21493) process is complete.


DilumAluthge:
> The remaining work to do here is to change [LibCURL](https://github.com/JuliaWeb/LibCURL.jl) to use [Downloads](https://github.com/JuliaLang/Downloads.jl) for its tests, once the [registration](https://github.com/JuliaRegistries/General/pull/21493) process is complete.

Downloads is now registered.

StefanKarpinski:
Downloads is registered & LibCURL is updated to use it for its tests instead of `Base.download`, so this is now ready to go if it passes CI.

PR #37613, opened 2020-09-16
increase `RELOC_TAG_OFFSET` to handle larger system images

JeffBezanson:
Currently very large system images can crash during loading due to overflowing the allowed range of relocation offsets. We currently use 7 RefTags, so we might as well just increase the number of bits available for now.

### DIFF
--- a/src/staticdata.c
+++ b/src/staticdata.c
@@ -168,9 +168,9 @@ typedef enum {
 } jl_callingconv_t;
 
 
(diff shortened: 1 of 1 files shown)


yuyichao:
Any idea how much does a compile all sysimg increase the size/required offset by?

Keno:
Can we get an explicit error when a relocation would overflow the new limit, so we don't have to diagnose it by rr trace?

PR #37615, opened 2020-09-16
remove call to `find_throw_blocks` in inlining

JeffBezanson:
This should now be handled by statement info. I can confirm removing this has a negligible effect. We will now do the process_simple cases for these, but I doubt that has much cost.


Keno (approved):
LGTM, but has a failing doctest.

JeffBezanson:
Aha, the doctest is interesting, and tells us this is not a no-op. After inference and ssa conversion we add unreachable nodes, growing the set of statements to ignore. But after #37572 it should indeed be a no-op, so these PRs go well together I suppose :)

PR #31916, opened 2019-05-03
gf: support more dispatch on abstract types

vtjnash:
This removes the restriction on defining dispatch over user-defined abstract types. This fixes #14919 for 99% of cases.
```julia
abstract type A14919; end
struct B14919 <: A14919; end
struct C14919 <: A14919; end
struct D14919 <: Function; end
(::A14919)() = "It's a brand new world"
(::Union{C14919,D14919})() = "Boo."
@test B14919()() == "It's a brand new world"
@test C14919()() == D14919()() == "Boo."
```

The "cannot add methods to an abstract type" error is now only applicable to a couple types (`Any`, `Function`, and functions), and instead now gives a "not implemented yet" message (that's perhaps not entirely right, since in several instances of this, I think they would be ambiguous with calling a Builtin function—and we don't need to ever enable that particular foot-gun).

This PR actually hilariously similar to my branch https://github.com/JuliaLang/julia/issues/14919#issuecomment-221631911 from 3 years ago, but adds a few more error checks so that performance is unaffected.

Issue #37458, opened 2020-09-08
New StackOverflow in jl_has_free_typevars, intersect_all, intersect at, 


KristofferC:
In the latest PkgEval (https://github.com/JuliaCI/NanosoldierReports/blob/5401256da7d60e8b1ea5ec7c7a1a3becc543201b/pkgeval/by_hash/d97ccd7_vs_c4acbf9/report.md) a lot of packages started failing with logs like

```
Internal error: encountered unexpected error in runtime:
StackOverflowError()
jl_has_free_typevars at /workspace/srcdir/src/jltypes.c:190
intersect_union at /workspace/srcdir/src/subtype.c:2156
intersect at /workspace/srcdir/src/subtype.c:2979
intersect_all at /workspace/srcdir/src/subtype.c:3110
intersect_aside at /workspace/srcdir/src/subtype.c:2146
intersect_var at /workspace/srcdir/src/subtype.c:2337
intersect at /workspace/srcdir/src/subtype.c:2963
intersect_var at /workspace/srcdir/src/subtype.c:2257
```

The packages failing are:

```
AnyMOD
BifurcationKit
BilevelJuMP
BilevelOptimization
BirkhoffDecomposition
BundleMethod
CDDLib
COSMO
Cbc
ClimateMARGO
Clp
ConstraintSolver
Convex
CutPruners
DataEnvelopmentAnalysis
DensityRatioEstimation
DiffEqBayes
DiffEqCallbacks
DiffEqJump
DifferentialEquations
DiscreteFunctions
DualDecomposition
ECOS
ErdosExtras
Fri
GLPK
GasModels
GasPowerModels
GraphicalModelLearning
GuessworkQuantumSideInfo
InfiniteOpt
InfrastructureModels
Ipopt
JuliaGrid
Juniper
LatinSquares
LazySets
LightGraphsFlows
LightGraphsMatching
MINLPTests
MathOptInterface
ModelingToolkit
MosekTools
MultiScaleArrays
NaiveGAflux
NaiveNASflux
NaiveNASlib
OPFSampler
OpSel
ParameterJuMP
PiecewiseLinearOpt
PolyhedralRelaxations
PowerDynamics
PowerModelsACDC
PowerModelsAnnex
PowerModelsDistribution
PowerModelsRestoration
PowerModelsSecurityConstrained
PowerSimulations
PowerWaterModels
ProxSDP
QHull
QuadraticToBinary
RandomizedProgressiveHedging
SCS
SDDP
[...]

PR #37616, opened 2020-09-16
improve accuracy of ambiguity checking (take 2)

vtjnash:
There was a small typo in the original PR that caused this to disable a useful fast path in a few cases. Users were also unhappy to discover that poor type-intersection results that steal away performance were now being reported. Hide those again to calm the irate masses. Also finishes fixing detect_unbound_args and detect_ambiguities to correctly support #31916 (which only implemented a subset of the required corrections). This should now be faster and simpler to use, as it can avoid substantial excess work when used by a package to check itself.


KristofferC:
And I guess https://github.com/JuliaLang/julia/issues/37458 has been fixed not to happen again?

vtjnash:
Yes that’s the expectation, as I mentioned above, this continues to explicitly skip reporting ambiguities that might occur when type intersection Information seems poor. It remains a probable performance drain, but we’ll continue to avoid reporting them.

KristofferC:
For me this still seems to slow down Plots.jl time:

Before

```
  2.771853 seconds (6.19 M allocations: 442.641 MiB, 3.36% gc time)
```

After

```
julia> @time using Plots
  3.606244 seconds (7.18 M allocations: 497.321 MiB, 3.48% gc time)
```

Is that just for me?

vtjnash:
That's way below the variability I see for just building master. It can be somewhat more accurately measured by only rebuilding the src files. There's a small additional cost (I'd say about 10%), but that is just reflective of the bad job type-intersection is already doing on a range of important queries and so should see a more substantial speedup from fixing that (https://github.com/JuliaLang/julia/issues/36951).

KristofferC:
It seems pretty consistent to me:

PR:

```
~/julia master* ⇡
❯ ./julia -e '@time using Plots'
  3.528268 seconds (7.18 M allocations: 497.487 MiB, 2.88% gc time)

~/julia master* ⇡
❯ ./julia -e '@time using Plots'
  3.538384 seconds (7.18 M allocations: 497.479 MiB, 2.99% gc time)

~/julia master* ⇡
❯ ./julia -e '@time using Plots'
  3.539718 seconds (7.18 M allocations: 497.496 MiB, 3.17% gc time)
```

Master:

```
  ~/julia master*
❯ ./julia -e '@time using Plots'
  3.044624 seconds (6.15 M allocations: 441.648 MiB, 4.69% gc time)

~/julia master*
❯ ./julia -e '@time using Plots'
  2.919994 seconds (6.15 M allocations: 441.648 MiB, 4.92% gc time)

~/julia master*
❯ ./julia -e '@time using Plots'
  3.000036 seconds (6.15 M allocations: 441.650 MiB, 4.53% gc time)
```

A blanket 17% regression seems significant. Are you saying there is other work in the pipeline that brings this back or that we just have to eat this regression? Just looking at it from an outsider's p.o.v it seems unfortunate but I realize I don't have all the context for what this change does and what's upcoming.

vtjnash:
I can't tell exactly what you're measuring just from that, but it looks like you're not actually rebuilding, so the repeated trials aren't providing any useful additional information

vtjnash:
> Are you saying there is other work in the pipeline

I don't know if the work is in the pipeline, but that is why I keep pointing to the issue issue I created to document why this is an existing problem with master https://github.com/JuliaLang/julia/issues/36951

> A blanket 17% regression seems significant

Such a number is absolutely devoid of meaning without context. The context I'm attempting to give is that #36951 appears to cause significant extra work (disabling just some of the effect of it seems to take the ambiguity test here from about 80 seconds to about 14). On balance, this is intended to help stabilize https://github.com/JuliaLang/julia/pull/36733 to ensure we don't get odd ordering-related dependencies where latency times suddenly spike in certain circumstances (rebuilds).

PR #37627, opened 2020-09-17
Use logo in readme from within the repository

oxinabox:
We were loading the logo from the www.julialang.org website.
Apparently that logo moved or something tas the link is now broken.

We have a copy of the logo inside the repo already.
For sake of being self-contained and making this less likely to occur in future
maybe it is better to just refer to that?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 <a name="logo"/>
 <div align="center">
 <a href="https://julialang.org/" target="_blank">
-<img src="https://julialang.org/images/logo_hires.png" alt="Julia Logo" width="210" height="142"></img>
+<img src="doc/src/assets/logo.svg" alt="Julia Logo" width="210" height="142"></img>
 </a>
 </div>
 


simeonschaub:
Is it safe to assume that all browsers can render SVGs?

fredrikekre (approved):

oxinabox:
> Is it safe to assume that all browsers can render SVGs?

In short: yes.

Internet Explorer 8 was the last browser that supported images that couldn't.
It was the default browser that came with Windows 7.
But i think it gets replaced as soon as you run windows update.
Internet explorer 9 was released in March 2011, and fully supports SVGs.

simeonschaub:
But IE8 is my favorite browser! :laughing:

ViralBShah:
I think we should load the logos from https://github.com/JuliaLang/julia-logo-graphics

Those are the official source of logos. Pulling from there means you always get the latest one, and those URLs and filenames will be the most stable since they rarely change.

fredrikekre:
The one in this repo needs to be updated for the docs anyway, so might as well reuse it in the readme.

giordano:
I don't see the logo rendered on my phone (chrome on android)

Issue #28781, opened 2018-08-20
Workers should inherit Pkg environment


rened:
(Moved over from https://github.com/JuliaLang/Pkg.jl/issues/675)

If I have a project where I installed `MyPackage` and I can say

```julia --project -e "using MyPackage"```

the following does only work when I manually activate the environment:

```jl
using Distributed
using MyPackage   # works
addprocs(3)
@everywhere
  using Pkg; Pkg.activate(".")  # required
  using MyPackage
  ...
end
```

I would suggest that workers inherit the environment that is active when they are added with `addprocs`. Should there be any calls to `Pkg.activate` after `addprocs` they have to be manually performed by the user on all workers.

affans:
Can you explain a little bit more to whats happening? I have a large codebase I want to port to 1.0 that uses HPC/addprocs quite a bit. If I know better what is happening, maybe I can submit a PR

Element-126:
I just got bitten by this issue while I was porting a small project from 0.6 to 1.0. As soon as I started using a separate environment for the project, worker processes started to throw `Package ... is required but does not seem to be installed` exceptions.

In addition to @rened's suggestion, I would also suggest that `(MyProject) activate .` activates the environment on worker processes as well, or at least issues a warning instead of silently ignoring them. This would be useful when starting Julia with the `-p <N>` option and then manually activating the environment.

Element-126:
@affans It seems that `(MyProject) activate .` and `Pkg.activate(".")` only activate the environment on the master process. To activate it on all processes, you would need to use the following code:
```julia
using Distributed
@everywhere using Pkg
@everywhere Pkg.activate(".")
```
You can then use `@everywhere using MyPackage` as usual.

StefanKarpinski:
This requires some careful thinking about how best to do this. One possibility is to send a manifest from the master node to the workers and insist that the manifest be usable on all nodes. That's fine for non-dev packages and even for dev packages with relative paths but dev packages with absolute paths may be a bit of an issue. We could rewrite dev paths to use a commit instead and then send that over; of course, it requires that the workers know about the tree hash that's being used...

matteoacrossi:
I ran into the same problem and it took me a while to understand that the problem was with `activate`. So I believe there should at least be a warning when using activate with worker processes.

simonbyrne:
Another work around is to use the `JULIA_PROJECT` environment variable instead of `--project`, as that will be passed to subprocesses.

[...]

PR #37632, opened 2020-09-17
loading: provide code for reifying a load-path lookup result

vtjnash:
This allows asking (and storing) queries about what total set of modules might need to be loaded, before starting any load work. This is base work for later working on https://github.com/JuliaLang/julia/issues/28781, so we might add additional information later (such as folding `concrete_deps` into it?). This is intended to allow us to pass around the dependency list explicitly, so that worker processes (such as during module loading, precompile, Distributed, and so on) can be isolated from the implicit and mutable LOAD_PATH and Project file parts of the system, once a load is triggered by a toplevel `using` or `cachecompile` call.

### DIFF
--- a/base/loading.jl
+++ b/base/loading.jl
@@ -234,11 +234,13 @@ end
 ## package identification: determine unique identity of package to be loaded ##
 
 # Used by Pkg but not used in loading itself
-function find_package(arg)
-    pkg = identify_package(arg)
+find_package(name::String) = find_package(PkgId(""), name) # `where` without a uuid will be ignored
+function find_package(where::Union{Module,PkgId}, name::String)
+    pkg = identify_package(where, name)
     pkg === nothing && return nothing
     return locate_package(pkg)
 end
+
 
 ## package identity: given a package name and a context, try to return its identity ##
 identify_package(where::Module, name::String) = identify_package(PkgId(where), name)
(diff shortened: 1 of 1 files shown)


base/loading.jl
```diff
@@ -227,7 +251,7 @@ function locate_package(pkg::PkgId, cache::TOMLCache=TOMLCache())::Union{Nothing
                 return implicit_manifest_uuid_path(env, pkg, cache)
             end
             @assert found.uuid !== nothing
-            return locate_package(found, cache) # restart search now that we know the uuid for pkg
+            return locate_package(found, cache) # restart search now that we know the uuid for pkg (TODO: the existence of this line of code is probably a bug)
```
KristofferC, line 254: Yeah, this line feels strange to me too.

base/loading.jl
```diff
@@ -371,6 +425,7 @@ end
 
 # given a directory (implicit env from LOAD_PATH) and a name,
 # check if it is an implicit package
+# TODO: aren't we supposed to first check for the Project file first and see if it declares a path?
```
KristofferC, line 428: the path entry in `Project` the project file seems to me like a bit of an experiment that turned out to be unused. Pkg doesn't know about it for example.

base/loading.jl
```diff
@@ -459,18 +541,54 @@ function explicit_manifest_deps_get(project_file::String, where::UUID, name::Str
...
+    found_where = false
+    entries = get(d, where.name, nothing)::Union{Vector{Any}, Nothing}
+    if entries !== nothing
+        for entry in entries
+            entry::Dict{String, Any}
+            uuid = get(entry, "uuid", nothing)::Union{String, Nothing}
+            uuid === nothing && continue
+            if UUID(uuid) === where.uuid
+                found_where = true
+                # deps is either a list of names (deps = ["DepA", "DepB"]) or
```
KristofferC, line 561: Could some of this code be shared with `explicit_manifest_deps_get`?

KristofferC (commented):
This does indeed seem useful to have!

KristofferC:
> This is intended to allow us to pass around the dependency list explicitly, so that worker processes (such as during module loading, precompile, Distributed, and so on) can be isolated from the implicit and mutable LOAD_PATH and Project file parts of the system, once a load is triggered by a toplevel using or cachecompile call.

In what way did you imagine passing this information to worker. `repr` feels a bit too slow when large environment data has to be passed (#37989), and we don't really have access to serialization in Base.

vtjnash:
This doesn’t need either form. It’s just a trivial key/value store with a tiny, well-defined set of pairs. We actually already deal with these, just only sending a large subset of them instead of all.

KristofferC:
Rebased

vtjnash:
@KristofferC is this PR still of any interest?

PR #37635, opened 2020-09-17
ispow2(x) for non-Integer x

stevengj:
Seemed like there is no reason to limit `ispow2(x)` to `x::Integer`.

### DIFF
--- a/base/intfuncs.jl
+++ b/base/intfuncs.jl
@@ -368,9 +368,9 @@ _prevpow2(x::Unsigned) = one(x) << unsigned((sizeof(x)<<3)-leading_zeros(x)-1)
 _prevpow2(x::Integer) = reinterpret(typeof(x),x < 0 ? -_prevpow2(unsigned(-x)) : _prevpow2(unsigned(x)))
 
 """
-    ispow2(n::Integer) -> Bool
+    ispow2(n::Number) -> Bool
 
-Test whether `n` is a power of two.
+Test whether `n` is an integer power of two.
 
 # Examples
 ```jldoctest
@@ -379,8 +379,19 @@ true
 
 julia> ispow2(5)
 false
+
+julia> ispow2(4.5)
+false
+
+julia> ispow2(0.25)
+true
+
+julia> ispow2(1//8)
+true
 ```
 """
+ispow2(x::Number) = isreal(x) && ispow2(real(x))
+
 ispow2(x::Integer) = x > 0 && count_ones(x) == 1
 
 """
--- a/base/intfuncs.jl
+++ b/base/intfuncs.jl
@@ -389,6 +389,9 @@ true
 julia> ispow2(1//8)
 true
 ```
+
+!!! compat "Julia 1.6"
+    Support for non-`Integer` arguments was added in Julia 1.6.
 """
 ispow2(x::Number) = isreal(x) && ispow2(real(x))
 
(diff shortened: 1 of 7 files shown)


base/intfuncs.jl
```diff
@@ -368,9 +368,9 @@ _prevpow2(x::Unsigned) = one(x) << unsigned((sizeof(x)<<3)-leading_zeros(x)-1)
 _prevpow2(x::Integer) = reinterpret(typeof(x),x < 0 ? -_prevpow2(unsigned(-x)) : _prevpow2(unsigned(x)))
 
 """
-    ispow2(n::Integer) -> Bool
+    ispow2(n::Number) -> Bool
 
-Test whether `n` is a power of two.
+Test whether `n` is an integer power of two.
```
simeonschaub, line 373: Should this have a compat annotation?

stevengj: yes, I always forget those; will do.

stevengj:
CI errors seem unrelated:
```
command timed out: 7320 seconds without output running [b'bin/julia', b'-e', b'include(joinpath(Sys.BINDIR, Base.DATAROOTDIR, "julia", "test", "choosetests.jl")); Base.runtests(append!(choosetests()[1], ["LibGit2/online", "download"]); ncores=min(Sys.CPU_THREADS, 8, 5))'], attempting to kill
```
on tester_macos64 and
```
command timed out: 3600 seconds without output running ['sh', '-c', 'make VERBOSE=1 TAGGED_RELEASE_BANNER="Official https://julialang.org/ release" SRCCACHE=/tmp/srccache USECCACHE=1 JULIA_CPU_TARGET="generic;cortex-a57;thunderx2t99;armv8.2-a,crypto,fullfp16,lse,rdm" MARCH=armv8-a  LLVM_ASSERTIONS=1 FORCE_ASSERTIONS=1 USE_BINARYBUILDER=1  binary-dist'], attempting to kill
```
on package_linuxaarch64.

base/intfuncs.jl
```diff
@@ -379,8 +379,22 @@ true
...
+true
+
+julia> ispow2(1//8)
+true
 ```
+
+!!! compat "Julia 1.6"
+    Support for non-`Integer` arguments was added in Julia 1.6.
 """
+ispow2(x::Number) = isreal(x) && ispow2(real(x))
```
JeffBezanson, line 396: Maybe this should be restricted to `Complex`, since there isn't a fallback covering all Real values.

stevengj: I feel like there should be a way to throw a `MethodError` for common patterns like this, rather than `StackOverflowError`.  I've bumped into similar issues many times.

Could we implement something like
```jl
@norecurse ispow2(x::Number) = isreal(x) && ispow2(real(x))
```

stevengj: (For example, this `Number` method would work perfectly well for quaternion types.  It seems a shame to lose genericity due to dispatch limitations.)

JeffreySarnoff: I agree.

JeffreySarnoff: With @stevengj's approach ^, the work of making types consistent inside and collaboratively correct outside is a touch simpler. 
```
@norecurse fn(x::Real) = fn(float(x))
```
_dispatch resolves, giving attentive safety_

PR #37636, opened 2020-09-17
Attach docstring for 'abstract type' to 'abstract' too.

fredrikekre:
(And no, if anyone thinks I had a brain fart and forgot what was after `abstract` in `abstract type` and had to look it up and failed since `?abstract` in the REPL gave nothing -- that is *not* what is happening :upside_down_face: )


simeonschaub:
Since `abstract` is still a valid identifier, we eventually might want to figure out whether we can get the REPL help mode to still look for docs for the identifier `abstract`, even if we add a docstring for the keyword. (Ref: https://github.com/JuliaLang/julia/pull/37471#issuecomment-689563861) That doesn't have to block this PR though.

PR #37637, opened 2020-09-18
fix #37610, allow constant prop on signatures with unions

### LINKED ISSUE #37610: Imprecise inference when iterating a Pair with constant first type but varying second type.
```jl
julia> a = ("foo" => "bar", "baz" => nothing);

julia> function f(a, i)
           y = iterate(a, i)
           if y !== nothing
               (k, v), st = y
               return k, v
           end
           return y
       end

julia> @code_warntype f(a, 1)
```

infers to `Union{Nothing, Tuple{Union{Nothing, String},Union{Nothing, String}}}` In particular, `k` is inferred as `  k::Union{Nothing, String}` but it could maybe be inferrable as a `String`.

JeffBezanson:
Constant prop looks at `nonbot` (# of non-Bottom-returning method signatures) to determine whether there is only one match. If we split Union types though, there appears to be more than one, so we don't do constant prop, which is the source of the issue. This fixes it by also splitting the `argtypes` array that includes constant information, and still constant prop'ing as long as there is only one actual method match. I have not seen any measurable latency impact so far.

I added inline declarations to some indexed_iterate methods for good measure, though it is not really necessary to fix this.

While working on this I noticed that the two methods of `alloc_buf_hook` returned spuriously different types, so I changed them to consistently return `Ptr{Cvoid}, Int`. I'm totally indifferent as to whether we return `Ptr{UInt8}, UInt` instead though if anybody prefers that :joy: 

fix #37610

### DIFF
--- a/base/compiler/abstractinterpretation.jl
+++ b/base/compiler/abstractinterpretation.jl
@@ -29,9 +29,13 @@ function abstract_call_gf_by_type(interp::AbstractInterpreter, @nospecialize(f),
     fullmatch = Bool[]
     if splitunions
         splitsigs = switchtupleunion(atype)
+        split_argtypes = switchtupleunion(argtypes)
         applicable = Any[]
+        # arrays like `argtypes`, including constants, for each match
+        applicable_argtypes = Vector{Any}[]
         infos = MethodMatchInfo[]
-        for sig_n in splitsigs
+        for j in 1:length(splitsigs)
+            sig_n = splitsigs[j]
             mt = ccall(:jl_method_table_for, Any, (Any,), sig_n)
             if mt === nothing
                 add_remark!(interp, sv, "Could not identify method table for call")
@@ -45,6 +49,9 @@ function abstract_call_gf_by_type(interp::AbstractInterpreter, @nospecialize(f),
             end
             push!(infos, MethodMatchInfo(matches))
             append!(applicable, matches)
+            for _ in 1:length(matches)
+                push!(applicable_argtypes, split_argtypes[j])
+            end
             valid_worlds = intersect(valid_worlds, matches.valid_worlds)
             thisfullmatch = _any(match->(match::MethodMatch).fully_covers, matches)
             found = false
@@ -80,6 +87,7 @@ function abstract_call_gf_by_type(interp::AbstractInterpreter, @nospecialize(f),
         info = MethodMatchInfo(matches)
         applicable = matches.matches
         valid_worlds = matches.valid_worlds
+        applicable_argtypes = nothing
     end
     update_valid_age!(sv, valid_worlds)
     applicable = applicable::Array{Any,1}
@@ -136,6 +144,9 @@ function abstract_call_gf_by_type(interp::AbstractInterpreter, @nospecialize(f),
         if this_rt !== Bottom
             if nonbot === 0
                 nonbot = i
+            elseif nonbot === -1
+            elseif method === (applicable[nonbot]::MethodMatch).method
+                # another entry from the same method, due to union splitting
             else
                 nonbot = -1
             end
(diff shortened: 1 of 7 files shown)


base/compiler/abstractinterpretation.jl
```diff
@@ -151,7 +162,15 @@ function abstract_call_gf_by_type(interp::AbstractInterpreter, @nospecialize(f),
...
         # TODO: it feels like this could be better integrated into abstract_call_method / typeinf_edge
-        const_rettype = abstract_call_method_with_const_args(interp, rettype, f, argtypes, applicable[nonbot]::MethodMatch, sv, edgecycle)
+        const_rettype = Bottom
+        for i in 1:napplicable
+            mm = applicable[i]::MethodMatch
+            if i === nonbot || mm.method === (applicable[nonbot]::MethodMatch).method
+                this_argtypes = applicable_argtypes === nothing ? argtypes : applicable_argtypes[i]
+                one_rettype = abstract_call_method_with_const_args(interp, rettype, f, this_argtypes, mm, sv, edgecycle)
+                const_rettype = tmerge(const_rettype, one_rettype)
+            end
```
vtjnash, line 172:
```suggestion
                const_rettype ⊑ rettype || (const_rettype = Any)
                const_rettype === Any && break
            end
```

base/compiler/abstractinterpretation.jl
```diff
@@ -151,7 +162,15 @@ function abstract_call_gf_by_type(interp::AbstractInterpreter, @nospecialize(f),
         # if there's a possibility we could constant-propagate a better result
```
vtjnash, line 162:
```suggestion
            || isa(rettype, PartialStruct)
        # if there's a possibility we could constant-propagate a better result
```

vtjnash (commented):
I'm inclined to think we should majorly refactor this part of the code, so that we can make more intelligent decisions here, ~~but this LGTM.~~

In particular, `abstract_call_method_with_const_args` could be split into the policy and mechanism, that way we can check whether, for each split call, whether there is value in attempting constant splitting for that particular call site, before attempting it, (and then also remove the restriction heuristics that there's only one method match).

EDIT: as I wrote the above, I realized there's possibly an issue with this currently. The `abstract_call_method_with_const_args` call signals failures by returning `Any` (no improved information). That worked because it assumed we only would use it once. But that'll significantly degrade the reliability of the information that this is supposed to provide, because of the immediate `tmerge`.

JeffBezanson:
If `abstract_call_method_with_const_args` returns `Any` for any of the split signatures, shouldn't we just bail out of constant prop though (which I believe the current version of the code does)?

vtjnash:
No, that seems like it'd be highly unreliable, since that assumes all useful type information must have been derived from constant propagation, but that's expected to be nearly always false. Perhaps uncommon it'll matter, but that feels more related to the unlikeliness that the extra effort in this PR to constant propagate harder will be able to improve the inferred result than the unlikeliness that the function may result in the same answer for different input constants.

aviatesk:
@JeffBezanson @vtjnash 
Can I take the initiative on this ? I've opened #39305 that has the same goal as this without noticing this..
I agree with [@vtjnash 's comment](https://github.com/JuliaLang/julia/pull/37637#pullrequestreview-491618899), and I'd like to refactor abstract_call_method_with_const_args so that we can make more sensible decision earlier about whether to do constant prop' before we ends up `Any` return result on 2nd or subsequent constant prop' trial.

base/compiler/typeutils.jl
```diff
@@ -133,10 +133,16 @@ function switchtupleunion(@nospecialize(ty))
...
 end
 
+switchtupleunion(t::Vector{Any}) = _switchtupleunion(t, length(t), [], nothing)
+
 function _switchtupleunion(t::Vector{Any}, i::Int, tunion::Vector{Any}, @nospecialize(origt))
     if i == 0
-        tpl = rewrap_unionall(Tuple{t...}, origt)
-        push!(tunion, tpl)
+        if origt === nothing
+            push!(tunion, t)
```
aviatesk, line 141:
```suggestion
            push!(tunion, copy(t))
```

I found we need to copy `t` here, otherwise it will be mutated at `t[i] = ti` in recursion.

vtjnash:
@JeffBezanson do you want to make a separate PR for https://github.com/JuliaLang/julia/pull/37637/commits/2e01da6f3588f893ccd67db9ded938e57470aacc

PR #37643, opened 2020-09-18
upgrade OffsetArrays to v1.3.0

johnnychen94:
Separated from #37629

cc: @timholy


StefanKarpinski:
@timholy, I think this would just need a quick once-over from you.

stdlib/Distributed/test/splitrange.jl
```diff
@@ -22,7 +22,7 @@ using Distributed: splitrange
 @test splitrange(-1, 1, 4) == Array{UnitRange{Int64},1}([-1:-1,0:0,1:1])
 
 const BASE_TEST_PATH = joinpath(Sys.BINDIR, "..", "share", "julia", "test")
-isdefined(Main, :OffsetArrays) || @eval Main include(joinpath($(BASE_TEST_PATH), "testhelpers", "OffsetArrays.jl"))
+isdefined(Main, :OffsetArrays) || @eval Main @everywhere include(joinpath($(BASE_TEST_PATH), "testhelpers", "OffsetArrays.jl"))
```
mbauman, line 25: It surprises me that this wasn't needed before

test/testhelpers/OffsetArrays.jl
```diff
@@ -5,141 +5,348 @@
...
-@inline function Base.setindex!(A::OffsetArray, val, i::Int)
-    checkbounds(A, i)
-    @inbounds parent(A)[i] = val
-    val
+@propagate_inbounds function Base.setindex!(A::OffsetArray, val, i::Int)
+    parent(A)[i] = val
+    A
 end
 
-@inline function Base.deleteat!(A::OffsetArray, i::Int)
```
mbauman, line 106: Interesting — is there a reason why this was removed from OffsetArrays?  Or has it never made it there?

stdlib/Distributed/test/splitrange.jl
johnnychen94: I think this is because `eachindex(oa)` requires `axes(oa)`, which now returns `OffsetArrays.IdOffsetRange`.

test/testhelpers/OffsetArrays.jl
johnnychen94: Hmmm... I'm surprised that this seems to never appeared in OffsetArrays.jl codebase. My best guess of this not going into OffsetArrays.jl is to make clear the conceptual confusion between an array and a list.

There are tendencies to add support for OffsetList, though (e.g., https://github.com/JuliaArrays/OffsetArrays.jl/pull/137, #37629)

mbauman: Makes sense — and the old definition for `deleteat!` doesn't really test any other base functionality, so I think this is fine to remove.

The definition for `push!` is also changing, though, and in doing so I think we lose the test coverage on `resize!`.

johnnychen94:
The test failure seems unrelated.

test/testhelpers/OffsetArrays.jl
johnnychen94: The `resize!` tests are added in 1f9b290 (copied from `OffsetArrays.jl`)

timholy:
Thanks @johnnychen94!

PR #37644, opened 2020-09-18
Recursively check wrapper arrays for issparse

oxinabox:
Do we want this?
I don't know.
What does it mean to be `issparse`? 
I haven't seen a defination of an interface that must be furfilled.

Possibly all people want this for is to know if they have some algorthm that really won't run great on a sparse array, then they could issue a warning or run `collect` to make it dense.

`issparse` compiles away.

```
julia> @code_llvm debuginfo=:none issparse(Transpose(Base.ReshapedArray(Symmetric(m), (20, 5), ())))

define i8 @julia_issparse_2137([1 x { { { i64, i64, {}*, {}*, {}* }, i32 }, [2 x i64] }]* nocapture nonnull readonly align 8 dereferenceable(64)) {
top:
  ret i8 1
}

julia> @code_llvm debuginfo=:none issparse(Transpose(Base.ReshapedArray(Symmetric(Array(m)), (20, 5), ())))

define i8 @julia_issparse_2139([1 x { { {}*, i32 }, [2 x i64] }]* nocapture nonnull readonly align 8 dereferenceable(32)) {
top:
  ret i8 0
}
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -130,6 +130,7 @@ Standard library changes
 #### SparseArrays
 
 * Display large sparse matrices with a Unicode "spy" plot of their nonzero patterns, and display small sparse matrices by an `Matrix`-like 2d layout of their contents.
+* `issparse` now applies consistently to all wrapper arrays, including nested, by checking `issparse` on the wrapped parent array ([#37644]).
 
 #### Dates
 * `Quarter` period is defined ([#35519]).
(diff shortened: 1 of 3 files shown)


fredrikekre:
#34308 ?

oxinabox:
Huh, apparently i already made this PR.
Go past me.

PR #22489, opened 2017-06-23
WIP: add IteratorAccess trait: for indexing capabilities

rfourquet:
I made this change to allow e.g. `sort!(zip(a, b))`, and have `a` and `b` sorted in-place according to the values in `a` (which I needed in a context similar to where `a` and `b` are the keys and values of a `Dict`).

I guess this is a way to circumevent the temporary lack of `AbstractArray` trait.
I will wait to see if this receives support before continuing this work.

PR #37648, opened 2020-09-18
Support indexing with generators

omus:
Indexing into generators is something I found myself wanting as part of https://github.com/JuliaTime/TimeZones.jl/pull/291. Basically I had function which contained code similar to:
```julia
x = [expensive_call(i) for i in range]
```
In some cases I would need access to all of the elements but in the most performance critical case I only needed access to a single element for which the index was already known. Using an index-able generator was a good solution for this.

I realize that not all generators will be index-able but this allows a generator to use indexing if the underlying iterator used by the generator also supports indexing.

### DIFF
--- a/base/generator.jl
+++ b/base/generator.jl
@@ -52,6 +52,20 @@ size(g::Generator) = size(g.iter)
 axes(g::Generator) = axes(g.iter)
 ndims(g::Generator) = ndims(g.iter)
 
+function getindex(g::Generator{<:AbstractArray}, I...)
+    I′ = to_indices(g.iter, I)
+    subset = g.iter[I′...]
+    if isempty(index_shape(I′...))
+        g.f(subset)
+    else
+        map(g.f, subset)
+    end
+end
+
+getindex(g::Generator, I...) = collect(g)[I...]
+
+firstindex(g::Generator) = firstindex(g.iter)
+lastindex(g::Generator) = lastindex(g.iter)
 
 ## iterator traits
 
--- a/base/generator.jl
+++ b/base/generator.jl
@@ -62,8 +62,6 @@ function getindex(g::Generator{<:AbstractArray}, I...)
     end
 end
 
-getindex(g::Generator, I...) = collect(g)[I...]
-
 firstindex(g::Generator) = firstindex(g.iter)
 lastindex(g::Generator) = lastindex(g.iter)
 
--- a/base/generator.jl
+++ b/base/generator.jl
@@ -62,6 +62,16 @@ function getindex(g::Generator{<:AbstractArray}, I...)
     end
 end
 
+function getindex(g::Generator, I...)
+    A = collect(g.iter)
+    subset = A[I...]
+    if typeof(subset) == eltype(A)
+        g.f(subset)
+    else
+        map(g.f, subset)
+    end
+end
+
 firstindex(g::Generator) = firstindex(g.iter)
 lastindex(g::Generator) = lastindex(g.iter)
 
--- a/test/functional.jl
+++ b/test/functional.jl
@@ -219,3 +219,16 @@ end
 let (:)(a,b) = (i for i in Base.:(:)(1,10) if i%2==0)
     @test Int8[ i for i = 1:2 ] == [2,4,6,8,10]
 end
+
+@testset "indexing" begin
+    g = (i % 2 == 0 ? error("invalid") : i for i in 1:3)
+
+    @test g[3] == 3
+    @test g[1:2:3] == [1, 3]
+    @test firstindex(g) == 1
+    @test lastindex(g) == 3
+
+    @test (tuple(x) for x in ["a"])[1] == ("a",)
+    @test (tuple(x) for x in [[0]])[1] == ([0],)
+    @test (x * y for (x, y) in Dict(3 => 4))[1] == 12
+end
--- a/test/functional.jl
+++ b/test/functional.jl
@@ -230,5 +230,5 @@ end
 
     @test (tuple(x) for x in ["a"])[1] == ("a",)
     @test (tuple(x) for x in [[0]])[1] == ([0],)
-    @test_throws MethodError (x * y for (x, y) in Dict(3 => 4))[1]
+    @test (x * y for (x, y) in Dict(3 => 4))[1] == 12
 end
--- a/test/functional.jl
+++ b/test/functional.jl
@@ -230,5 +230,5 @@ end
 
     @test (tuple(x) for x in ["a"])[1] == ("a",)
     @test (tuple(x) for x in [[0]])[1] == ([0],)
-    @test (x * y for (x, y) in Dict(3 => 4))[1] == 12
+    @test_throws MethodError (x * y for (x, y) in Dict(3 => 4))[1]
 end
--- a/NEWS.md
+++ b/NEWS.md
@@ -92,7 +92,7 @@ New library features
 * New constructor `NamedTuple(iterator)` that constructs a named tuple from a key-value pair iterator.
 * A new `reinterpret(reshape, T, a::AbstractArray{S})` reinterprets `a` to have eltype `T` while potentially
   inserting or consuming the first dimension depending on the ratio of `sizeof(T)` and `sizeof(S)`.
-* Generators support `getindex`, `firstindex`, and `lastindex` if indexing is supported by the underlying iterator ([#37648]).
+* Generators support `getindex` when the underlying iterator is a subtype of `AbstractArray` ([#37648]).
 
 Standard library changes
 ------------------------
--- a/NEWS.md
+++ b/NEWS.md
@@ -92,7 +92,7 @@ New library features
 * New constructor `NamedTuple(iterator)` that constructs a named tuple from a key-value pair iterator.
 * A new `reinterpret(reshape, T, a::AbstractArray{S})` reinterprets `a` to have eltype `T` while potentially
   inserting or consuming the first dimension depending on the ratio of `sizeof(T)` and `sizeof(S)`.
-* Generators support `getindex` when the underlying iterator is a subtype of `AbstractArray` ([#37648]).
+* Generators support `getindex`, `firstindex`, and `lastindex` ([#37648]).
 
 Standard library changes
 ------------------------
--- a/NEWS.md
+++ b/NEWS.md
@@ -92,6 +92,7 @@ New library features
 * New constructor `NamedTuple(iterator)` that constructs a named tuple from a key-value pair iterator.
 * A new `reinterpret(reshape, T, a::AbstractArray{S})` reinterprets `a` to have eltype `T` while potentially
   inserting or consuming the first dimension depending on the ratio of `sizeof(T)` and `sizeof(S)`.
+* Generators support `getindex`, `firstindex`, and `lastindex` if indexing is supported by the underlying iterator ([#37648]).
 
 Standard library changes
 ------------------------


mbauman:
I have a branch that begins work on adding indexing to the ProductIterator — I've been out of the loop for a while now but if that's still not supported I think it should be before doing this.

omus:
>  have a branch that begins work on adding indexing to the ProductIterator — I've been out of the loop for a while now but if that's still not supported I think it should be before doing this.

Adding indexing to `ProductIterator` would be great! Although having indexing with `ProductIterator` would definitely be useful here I don't think this PR requires that functionality as we can always add it later.

rfourquet:
And for reference, I also have an old PR (#22489) which adds indexing/setindexing to a bunch of iterators from `Iterators` (but not finished for products ;-) ). but which failed to get traction. Maybe one reason was that it was trying to add also traits to classify indexing capabilities (I could easily remove the traits aspect and keep only indexing methods). Mentioning thishere so that it gets also considered when making a decision for this PR...

N5N3:
Should this be included in `LazyArray.jl`
(Both Iterator and Generator)

base/generator.jl
```diff
@@ -52,6 +52,9 @@ size(g::Generator) = size(g.iter)
 axes(g::Generator) = axes(g.iter)
 ndims(g::Generator) = ndims(g.iter)
 
+getindex(g::Generator, I...) = map(g.f, g.iter[I...])
```
tkf, line 55: IIUC, this is broken for something like `(tuple(x) for x in ["a"])[1]` and `(tuple(x) for x in [[0]])[1]`. I think it's better to avoid defining "vectorized" `getindex` for `Generator` (and let something like #30845 handle it).

test/functional.jl
```diff
@@ -219,3 +219,12 @@ end
 let (:)(a,b) = (i for i in Base.:(:)(1,10) if i%2==0)
     @test Int8[ i for i = 1:2 ] == [2,4,6,8,10]
 end
+
+@testset "indexing" begin
```
nalimilan, line 223: While you're at it, it could make sense to test `find*` functions too, as they rely on indexing.

base/generator.jl
```diff
@@ -52,6 +52,9 @@ size(g::Generator) = size(g.iter)
...
+firstindex(g::Generator) = firstindex(g.iter)
+lastindex(g::Generator) = lastindex(g.iter)
```
nalimilan, line 58: How about also defining `eachindex`, `keys`, `pairs`, `IndexStyle`, etc.?

omus: I was trying to stick to just implementing the indexing interface but it should be fine to extend into the abstract array interface. Note only `keys` is needed for `eachindex` and `pairs`.

base/generator.jl
omus: Thanks for pointing this out. Using broacasting would fix your first example but not the second. I think we'll need to address this in some way as the differences can occur with linear indexing as shown in your examples. Ideally we want `g[i...] == collect(g)[i...]`.

base/generator.jl
nalimilan: `eachindex` can return something different from  (more efficient than) `keys`, so I think both are needed.

omus:
Some benchmarks with comparing indexing into generators versus having to collect the generator to index into them:

```julia
julia> g = (string(x) for x in [1 3 5; 2 4 6])
Base.Generator{Matrix{Int64}, typeof(string)}(string, [1 3 5; 2 4 6])

julia> @btime g[end]
  67.197 ns (2 allocations: 96 bytes)
"6"

julia> @btime collect(g)[end]
  286.318 ns (13 allocations: 704 bytes)
"6"

julia> @btime g[1:6]
  292.162 ns (14 allocations: 832 bytes)
6-element Vector{String}:
 "1"
 "2"
 "3"
 "4"
 "5"
 "6"

julia> @btime collect(g)[1:6]
  311.695 ns (14 allocations: 832 bytes)
6-element Vector{String}:
 "1"
 "2"
 "3"
 "4"
 "5"
 "6"
```

However, I did notice worse performance when handling colon:

```julia
julia> @btime g[:]
  471.658 ns (16 allocations: 864 bytes)
6-element Vector{String}:
 "1"
 "2"
 "3"
 "4"
 "5"
 "6"

julia> @btime collect(g)[:]
  307.751 ns (14 allocations: 832 bytes)
6-element Vector{String}:
 "1"
 "2"
 "3"
 "4"
 "5"
 "6"
```

test/functional.jl
omus: I'll leave this out of this PR as it appears there have been other efforts to support this which have stalled: https://github.com/JuliaLang/julia/pull/34678

base/generator.jl
omus: I'm going to limit the scope of this PR just to the [indexing interface](https://docs.julialang.org/en/v1/manual/interfaces/#Indexing) as I suspect scope creep will result in this PR stalling out.

base/generator.jl
omus: I've revised the code to take into account the index shape

base/generator.jl
```diff
@@ -52,7 +52,16 @@ size(g::Generator) = size(g.iter)
 axes(g::Generator) = axes(g.iter)
 ndims(g::Generator) = ndims(g.iter)
 
-getindex(g::Generator, I...) = map(g.f, g.iter[I...])
+function getindex(g::Generator, I...)
+    I′ = to_indices(g.iter, I)
```
tkf, line 56: `to_indices` doesn't work on non-array indexables like `NamedTuple` and `Dict`. As I commented earlier, I don't think we should support vectorized indexing.

You can use LazyArrays.jl or MappedArrays.jl to get a full array API. If we were to add vectorized indexing for lazy mapping in `Base`, `Broadcasted` is probably a better interface to do it since it's already very array-like.

omus: > As I commented earlier, I don't think we should support vectorized indexing.

The tricky part of your statement is how would we limit `getindex` to just use scalar indexing?

omus:
As pointed out supporting `getindex` when the iterator is not array-like gets difficult. We could define a fallback of `getindex(g::Generator, I...) = collect(g)[I...]` which is functional but not very performant. Instead I've limited this PR to only support `getindex(::Generator{<:AbstractArray}, I...)` which should provide the functionality most people care about while ensuring that users don't use `getindex` on generators where they should be doing `collect(g)[I...]` themselves.

base/generator.jl
mbauman: Just define `getindex(::Generator, ::Integer...)` and nothing else.  I also think this is the right thing to do.

tkf: > how would we limit `getindex` to just use scalar indexing?

Ah, good point. How about introducing

```julia
function to_scalar_indices(A, I)
    J = to_indices(A, I)
    J isa Tuple{Vararg{Integer}} || error("expected scalar indices. got: ", I)
    return J
end

scalar_getindex(A::AbstractArray, I...) = getindex(A, to_scalar_indices(A, I)...)
scalar_getindex(A, I...) = getindex(A, I...)
```

and use it for `Generator`?

> Just define `getindex(::Generator, ::Integer...)`

Isn't it too restrictive and a half-way solution if it were to restrict the API? For example, it'd mean `x[1]` for `x = Iterators.map(string, (a=1, b=2))` works but `x[:a]` doesn't.

Also, I guess we need a different implementation for `AbstractDict` or error out? Applying `f` after `getindex` is not correct for `AbstractDict`.

omus: > Just define `getindex(::Generator, ::Integer...)`

Mostly that works but we get into some strange corner cases with `AbstractDict`:

```julia
julia> function Base.getindex(g::Base.Generator, I::Integer...)
           g.f(g.iter[I...])
       end

julia> g = (x * y for (x, y) in Dict(3 => 4))
Base.Generator{Dict{Int64,Int64},var"#1#2"}(var"#1#2"(),
Dict(3 => 4))

julia> collect(g)[1]
12

julia> g[1]
ERROR: KeyError: key 1 not found
Stacktrace:
 [1] getindex at ./dict.jl:467 [inlined]
 [2] getindex(::Base.Generator{Dict{Int64,Int64},var"#1#2"}, ::Int64) at ./REPL[1]:2
 [3] top-level scope at REPL[3]:1
```
This was the main reason I ended up restricting the logic to `AbstractArray`

omus: > I guess we need a different implementation for `AbstractDict` or error out

That was also my conclusion. A first pass at the `AbstractDict` version looks like:
```julia
function Base.getindex(g::Base.Generator{<:AbstractDict}, I...)
    subset = collect(pairs(g.iter))[I...]
    if subset isa Pair
        g.f(subset)
    else
        map(g.f, subset)
    end
end
```

mcabbott:
Should this also define `keys`? Today I wanted something like this to work:
```
 findall(c->c[1]>0, eachcol(randn(2,20)))
```
Adding `Base.keys(g::Base.Generator) = keys(g.iter)` would make it so, and might fit well with being able to index.

omus:
> Should this also define `keys`?

See: https://github.com/JuliaLang/julia/pull/37648#discussion_r492705746

omus:
I found a reasonable fallback for "handling" index with iterators. As `getindex` and iteration don't have to return the same values I need index on the iteration results. There's still a good performance here for cases where `f` is an expensive call:

```julia
julia> g = (string(p) for p in Dict(Pair.('a':'z', 1:26)));

julia> @btime g[2:5]
  1.481 μs (64 allocations: 3.03 KiB)
4-element Vector{String}:
 "'f' => 6"
 "'w' => 23"
 "'d' => 4"
 "'e' => 5"

julia> @btime collect(g)[2:5]
  8.895 μs (409 allocations: 15.81 KiB)
4-element Vector{String}:
 "'f' => 6"
 "'w' => 23"
 "'d' => 4"
 "'e' => 5"

julia> g = (p for p in Dict(Pair.('a':'z', 1:26)));

julia> @btime g[2:5]
  187.209 ns (3 allocations: 784 bytes)
4-element Vector{Pair{Char, Int64}}:
 'f' => 6
 'w' => 23
 'd' => 4
 'e' => 5

julia> @btime collect(g)[2:5]
  172.204 ns (2 allocations: 640 bytes)
4-element Vector{Pair{Char, Int64}}:
 'f' => 6
 'w' => 23
 'd' => 4
 'e' => 5
```

rapus95:
Suggestion: Change lines 61 and 71 to their lazy equivalents, i.e. `Generator(g.f, subset)` instead of `map(g.f, subset)`

### Reasoning 1 ###

We shouldn't mix lazy and eager evaluation if the user already decided on lazy evaluation by supplying a generator in the first place.

### Reasoning 2 ###

There are two things that `Generator(identity, iter)[:]` could mean:

a) `collect(iter)`
b) `iter`

To me b) seems to be favourable. In that case we must not call `map(g.f, iter)`.

### Example ###

`(1:8)[2:3] == 2:3` (not `==[2,3]`)
If we collect in those cases we would mix up lazy and eager evaluation and probably introduce some unexpected allocations.

### Reasoning 3 ###

Given that `Generator` is the lazy equivalent of map, I would expect to stay in the lazy operation mode as long as possible (i.e. without collecting). 
And as such the result of indexing into a generator should also be the lazy, non-collected, equivalent of indexing into an allocated structure.
Ultimately the decision about collection should be made by indexing into the iterator (for which the Generator is just a lazy relay) and thus shouldn't be made by the relay type by explicitly collecting the iterator.

This wouldn't cost us any possibilities because anybody who wants could still apply a manual explicit `collect`.

### Overall question ###

I know, I'm reasking what already has been asked, but I'm not convinced by the arguments.
Why do we want the lazy map variant to be more capable than the eager variant? I would consider that to be unexpected. Especially I wouldn't want the lazy variant to start collecting just to be more capable than the eager variant (which it wouldn't be anyway since it is itself eager then).

Why not just stick to
`getindex(g::Generator, I...) = Generator(g.f, g.iter[I...])`
to make it behave exactly like the eager variant and make it an effective pass-through call. It'd also perfectly align with https://github.com/JuliaLang/julia/pull/34678

base/generator.jl
rapus95: As in my new comment to this whole PR, why do you want to support AbstractDict in general?
IMO Generator should behave like a lazy map. And for map we have:
```julia
julia> map(x->println(x), Dict(:a=>2, :c=>"h"))
ERROR: map is not defined on dictionaries
Stacktrace:
 [1] error(::String) at .\error.jl:33
 [2] map(::Function, ::Dict{Symbol,Any}) at .\abstractarray.jl:2190
 [3] top-level scope at REPL[10]:1
```

PR #37650, opened 2020-09-18
restore type cache pruning in sysimage saving

JeffBezanson:
The code for this had bitrotted --- dump.c initializes builtin_typenames but doesn't use it, and staticdata.c uses it but doesn't initialize it :joy: This makes it do something again. I noticed this when looking into tree shaking, and in the future it will be very useful for that, to avoid spurious references to things that could otherwise be removed.

### DIFF
--- a/src/staticdata.c
+++ b/src/staticdata.c
@@ -99,7 +99,6 @@ static arraylist_t deser_sym;
 static htable_t backref_table;
 static int backref_table_numel;
 static arraylist_t layout_table;
-static arraylist_t builtin_typenames;
 
 // list of (size_t pos, (void *f)(jl_value_t*)) entries
 // for the serializer to mark values in need of rework by function f
@@ -259,8 +258,8 @@ static uintptr_t jl_fptr_id(void *fptr)
         return *(uintptr_t*)pbp;
 }
 
-#define jl_serialize_value(s, v) jl_serialize_value_(s,(jl_value_t*)(v))
-static void jl_serialize_value_(jl_serializer_state *s, jl_value_t *v);
+#define jl_serialize_value(s, v) jl_serialize_value_(s,(jl_value_t*)(v),1)
+static void jl_serialize_value_(jl_serializer_state *s, jl_value_t *v, int recursive);
 
 
 static void jl_serialize_module(jl_serializer_state *s, jl_module_t *m)
@@ -287,7 +286,7 @@ static void jl_serialize_module(jl_serializer_state *s, jl_module_t *m)
 
 #define NBOX_C 1024
 
-static void jl_serialize_value_(jl_serializer_state *s, jl_value_t *v)
+static void jl_serialize_value_(jl_serializer_state *s, jl_value_t *v, int recursive)
 {
     // ignore items that are given a special representation
     if (v == NULL || jl_is_symbol(v)) {
@@ -331,6 +330,8 @@ static void jl_serialize_value_(jl_serializer_state *s, jl_value_t *v)
         // skip it
     }
     else if (jl_is_svec(v)) {
+        if (!recursive)
+            return;
         size_t i, l = jl_svec_len(v);
         jl_value_t **data = jl_svec_data(v);
         for (i = 0; i < l; i++) {
@@ -365,6 +366,17 @@ static void jl_serialize_value_(jl_serializer_state *s, jl_value_t *v)
     }
     else if (jl_typeis(v, jl_module_type)) {
         jl_serialize_module(s, (jl_module_t*)v);
+    }
+    else if (jl_is_typename(v)) {
+        jl_typename_t *tn = (jl_typename_t*)v;
+        jl_serialize_value(s, tn->name);
+        jl_serialize_value(s, tn->module);
+        jl_serialize_value(s, tn->names);
+        jl_serialize_value(s, tn->wrapper);
+        jl_serialize_value_(s, (jl_value_t*)tn->cache, 0);
+        jl_serialize_value_(s, (jl_value_t*)tn->linearcache, 0);
+        jl_serialize_value(s, tn->mt);
+        jl_serialize_value(s, tn->partial);
     }
     else if (t->layout->nfields > 0) {
         char *data = (char*)jl_data_ptr(v);
@@ -1235,7 +1247,7 @@ static void jl_finalize_serializer(jl_serializer_state *s, arraylist_t *list)
 }
 
 
-static void jl_reinit_item(jl_value_t *v, int how)
+static void jl_reinit_item(jl_value_t *v, int how) JL_GC_DISABLED
 {
     switch (how) {
         case 1: { // rehash IdDict
@@ -1280,7 +1292,7 @@ static void jl_reinit_item(jl_value_t *v, int how)
 }
 
 
-static void jl_finalize_deserializer(jl_serializer_state *s)
+static void jl_finalize_deserializer(jl_serializer_state *s) JL_GC_DISABLED
 {
     // run reinitialization functions
     uintptr_t base = (uintptr_t)&s->s->buf[0];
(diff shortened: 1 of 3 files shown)


JeffBezanson:
Hmm, how do we convince the gc analyzer that the system image shenanigans are ok? The errors don't obviously point to changes in this PR so it makes me wonder how it worked before.

yuyichao:
Aren't all of these running with GC disabled?

JeffBezanson:
Yes.

yuyichao:
So it seems that `JL_GC_DISABLED` should fix it, though I have no idea why this wasn't detected as an error before. The control flow and the arguments leading to the error also seems unchanged...

src/staticdata.c
```diff
@@ -1316,6 +1319,9 @@ static void jl_prune_type_cache(jl_svec_t *cache)
     }
     if (i > ins) {
         memset(&jl_svec_data(cache)[ins], 0, (i - ins) * sizeof(jl_value_t*));
+        while (l > 0 && l/2 >= ins)
+            l = l/2;
+        jl_svec_len(cache) = l == 0 ? 1 : l;
```
vtjnash, line 1324:
```suggestion
        jl_svec_len(cache) = l == 0 ? 1 : l; // XXX: this'll cause the GC to abort later since we've corrupted the object metadata
        jl_rehash_type_cache(tn); // might as well do this now, and not on every restart?
```

vtjnash (commented):
Might be simpler to just NULL the references now (in a copy of the svec) then save/restore the `->cache` fields, instead of removing them and needing to mess with them on startup too?

JeffBezanson:
How exactly would that work? How do we know which entries to NULL if the ->cache fields are also saved?

vtjnash:
This is a mark-and-sweep moving GC, so you'd handle it the same way as the GC handles WeakRef, in between the record and copy steps.

JeffBezanson:
Oh, I think I tried implementing it that way first --- skipping those fields during the serialize_value step. It seemed ugly to me to add a special case there, but maybe it's no worse than the alternative.

JeffBezanson:
Ok, this version does seem nicer to me.

vtjnash (commented):
Better, yes, but it's still corrupting the current runtime by sorting the `tn->cache` field at the wrong time

src/staticdata.c
```diff
@@ -1597,6 +1618,11 @@ static void jl_restore_system_image_from_stream(ios_t *f)
     s.s = NULL;
 
     s.s = f;
+    init_builtin_typenames();
+    for (i = 0; i < builtin_typenames.len; i++) {
+        jl_typename_t *tn = (jl_typename_t*)builtin_typenames.items[i];
+        jl_rehash_type_cache(tn);
+    }
```
vtjnash, line 1625:
```suggestion
```

src/staticdata.c
```diff
@@ -1376,16 +1401,12 @@ static void jl_save_system_image_to_stream(ios_t *f)
             jl_value_t *tag = *tags[i];
             jl_serialize_value(&s, tag);
         }
+        // prune unused entries from built-in type caches
         for (i = 0; i < builtin_typenames.len; i++) {
             jl_typename_t *tn = (jl_typename_t*)builtin_typenames.items[i];
             jl_prune_type_cache(tn->cache);
             jl_prune_type_cache(tn->linearcache);
```
vtjnash, line 1408:
```suggestion
            jl_prune_type_cache(tn->linearcache);
            jl_rehash_type_cache(tn); // put the values in ->cache back to their old locations from before we called jl_prune_type_cache and messed up the order
```

src/staticdata.c
```diff
@@ -366,6 +371,25 @@ static void jl_serialize_value_(jl_serializer_state *s, jl_value_t *v)
...
+    else if (jl_is_typename(v)) {
+        jl_typename_t *tn = (jl_typename_t*)v;
+        size_t i;
+        int is_builtin = 0;
+        for (i = 0; i < builtin_typenames.len; i++) {
+            if (tn == (jl_typename_t*)builtin_typenames.items[i]) {
+                is_builtin = 1;
+                break;
+            }
+        }
```
vtjnash, line 383:
```suggestion
```

src/staticdata.c
```diff
@@ -366,6 +371,25 @@ static void jl_serialize_value_(jl_serializer_state *s, jl_value_t *v)
...
+        jl_serialize_value(s, tn->name);
+        jl_serialize_value(s, tn->module);
+        jl_serialize_value(s, tn->names);
+        jl_serialize_value(s, tn->wrapper);
+        jl_serialize_value_(s, (jl_value_t*)tn->cache, !is_builtin);
+        jl_serialize_value_(s, (jl_value_t*)tn->linearcache, !is_builtin);
```
vtjnash, line 389:
```suggestion
        jl_serialize_value_(s, (jl_value_t*)tn->cache, 0);
        jl_serialize_value_(s, (jl_value_t*)tn->linearcache, 0);
```

src/staticdata.c
```diff
@@ -1680,13 +1706,28 @@ JL_DLLEXPORT void jl_restore_system_image_data(const char *buf, size_t len)
...
+    arraylist_push(&builtin_typenames, jl_pointer_typename);
+    arraylist_push(&builtin_typenames, jl_type_typename);
+    arraylist_push(&builtin_typenames, ((jl_datatype_t*)jl_unwrap_unionall((jl_value_t*)jl_abstractarray_type))->name);
+    arraylist_push(&builtin_typenames, ((jl_datatype_t*)jl_unwrap_unionall((jl_value_t*)jl_densearray_type))->name);
+    arraylist_push(&builtin_typenames, jl_tuple_typename);
+    arraylist_push(&builtin_typenames, jl_vararg_typename);
+    arraylist_push(&builtin_typenames, jl_namedtuple_typename);
+    assert(builtin_typenames.len == NUM_BUILTIN_TYPENAMES);
+}
+
```
vtjnash, line 1723:
```suggestion
```

src/staticdata.c
```diff
@@ -1680,13 +1706,28 @@ JL_DLLEXPORT void jl_restore_system_image_data(const char *buf, size_t len)
...
 static void jl_init_serializer2(int for_serialize)
 {
     if (for_serialize) {
         htable_new(&symbol_table, 0);
         htable_new(&fptr_to_id, sizeof(id_to_fptrs) / sizeof(*id_to_fptrs));
         htable_new(&backref_table, 0);
-        arraylist_new(&builtin_typenames, 0);
+        init_builtin_typenames();
```
vtjnash, line 1730:
```suggestion
```

src/staticdata.c
```diff
@@ -101,6 +101,9 @@ static int backref_table_numel;
 static arraylist_t layout_table;
 static arraylist_t builtin_typenames;
 
+// # of typenames whose caches are treated as weak references
+#define NUM_BUILTIN_TYPENAMES 9
+
```
vtjnash, line 106:
```suggestion

```

src/staticdata.c
```diff
@@ -1376,16 +1401,12 @@ static void jl_save_system_image_to_stream(ios_t *f)
             jl_value_t *tag = *tags[i];
             jl_serialize_value(&s, tag);
         }
+        // prune unused entries from built-in type caches
         for (i = 0; i < builtin_typenames.len; i++) {
             jl_typename_t *tn = (jl_typename_t*)builtin_typenames.items[i];
```
vtjnash, line 1406:
```suggestion
        for (i = 0; i < backref_table.size; i += 2) {
            jl_typename_t *tn = (jl_typename_t*)backref_table.table[i];
            if (tn == HT_NOTFOUND || !jl_is_typename(tn))
                continue;
```

vtjnash (commented):
I don't think these types need to be a special case

vtjnash (approved):

JeffBezanson:
> Assertion failed: (jl_svec_len(newa) == sz), function jl_rehash_type_cache, file /usr/home/julia/buildbot/worker/package_freebsd64/build/src/jltypes.c, line 815.

Well, that's unexpected. Gotta love assertions.

JeffBezanson:
Hmm, I suppose the observed max probe distance is insertion-order-dependent, so rehashing to the same size is not guaranteed to work.

vtjnash:
yeah, we might want to do some sort of "replacements" htable, then we can put the correctly rehashed value in there? or just put the sentinel `nothing` in the table, instead of attempting to erase them entirely?

JeffBezanson:
I'll try that.

JeffBezanson:
Seems to work. What do you think?

vtjnash (approved):

vtjnash:
Cool, looks like this might save 1-2 MB from the image

PR #37652, opened 2020-09-18
Add in_sysimage(pkgid::PkgId) to check if a package is in the sysimage

IanButterworth:
Adds `in_sysimage(pkgid::PkgId)` for checking if a package is in the sysimage.

Proposed as the correct-est way to filter out packages that definitely won't need precompiling, in https://github.com/JuliaLang/Pkg.jl/pull/2021

Discussed here https://github.com/JuliaLang/Pkg.jl/pull/2018#discussion_r490576496

cc. @tkf @fredrikekre

### DIFF
--- a/base/Base.jl
+++ b/base/Base.jl
@@ -381,6 +381,9 @@ include(mapexpr::Function, mod::Module, _path::AbstractString) = _include(mapexp
 
 end_base_include = time_ns()
 
+const _sysimage_modules = PkgId[]
+in_sysimage(pkgid::PkgId) = pkgid in _sysimage_modules
+
 if is_primary_base_module
 function __init__()
     # try to ensuremake sure OpenBLAS does not set CPU affinity (#1070, #9639)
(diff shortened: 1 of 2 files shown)


base/Base.jl
```diff
@@ -405,6 +408,7 @@ function __init__()
     init_depot_path()
     init_load_path()
     init_active_project()
+    append!(_sysimage_modules, keys(loaded_modules))
```
tkf, line 411: I think we need `empty!` it for the incremental builds.

```suggestion
    append!(empty!(_sysimage_modules), keys(loaded_modules))
```

tkf:
I think this is good to go once we have some tests (in and not in stdlib).

tkf:
@KristofferC Do you think it works (i.e., `in_sysimage` returns `true`) with non-stdlib packages added via PackageCompiler.jl?

KristofferC:
> @KristofferC Do you think it works (i.e., in_sysimage returns true) with non-stdlib packages added via PackageCompiler.jl?

Probably, yes.

In theory, it should be possible to set this variable once and for all before the sysimage is written, so there is no work at runtime, but that might be more trouble than it is worth.

PR #37653, opened 2020-09-18
[BinaryPlatforms] Do not cache host platform

giordano:
Caching the host platform can lead to some nasty bugs, like libgfortran not
being detected because openblas -> libgfortran wasn't loaded when the variable
is set.

### DIFF
--- a/base/binaryplatforms.jl
+++ b/base/binaryplatforms.jl
@@ -1004,8 +1004,6 @@ function host_triplet()
     return str
 end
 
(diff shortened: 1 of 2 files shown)


staticfloat:
This should fix the failures seen when using the latest `Pkg` on `PackageEval`.  Our benchmarks show constructing it from scratch each time takes hundreds of microseconds, so not a performance issue at all.

staticfloat (approved):

Issue #7232, opened 2014-06-12
make `f (a,b)` an error


StefanKarpinski:
See https://groups.google.com/forum/#!topic/julia-users/ysc9loSduN0

We would want to make it a warning first, then an error later on.

stevengj:
Yeah, we should probably go ahead and disallow this.

ScottPJones:
An earlier comment asked if other languages disallowed spaces between the function name and (, and many of the older languages definitely did...and still do, in fact... I seem to recall even some early C compilers that did not allow them either [things like BDS C for the 8080 processor].  :+1: to disallow this...

tkelman:
Close due to #11891, or wait until the deprecation is removed?

ninjin:
@tkelman: I say we keep it open until we turn it into an error on the v0.5 branch.

ivarne:
Open a new issue to collect a list of 0.5 deprecations that should be removed?

stevengj:
I thought that all deprecations for version `0.$n` are generally turned into errors for `0.$(n+1)`?

tkelman:
Yeah, I don't think we need to remove deprecations that were added during 0.4-dev until we move from 0.5-dev to 0.5-pre.

swt30:
Maybe related: today I did

```julia
julia> using PyPlot. PyCall
```

What I expected: PyPlot and PyCall to be loaded
What I mistyped: A period instead of a comma 
What I got: The same as typing `using PyPlot.PyCall`
What I thought would happen if I made this mistake: A syntax error.

`a .b` is already deprecated but this is still permitted. I can imagine it being a little tricky to debug if you don't spot there's a dot instead of a comma there.

stevengj:
@swt30, this is unrelated; please file a separate issue if needed.

Keno:
Am I understanding correctly that the deprecation was introduced newly on 0.5? In that case this issue should be moved into the 0.6 milestone.

tkelman:
https://github.com/JuliaLang/julia/commit/3aed3f5489ab3d12155ba112644b3d86d7b42658 was included in 0.4.0, so no.

Keno:
I see, so this'll be closed when we remove deprecations.

tkelman:
The deprecation is entirely in the parser I think, so it depends whether those are as clearly organized as the ones in `base/deprecated.jl`.

PR #37657, opened 2020-09-19
fix #37656: don't always error on space before '

### LINKED ISSUE #37656: How is this conditional a syntax error?
This behavior seems odd to me:

```julia
if true 'a' else 1 end # syntax error
if true "a" else 1 end # returns "a"
if true 1 else 'a' end # returns 1
```

The backtrace is:

```
ERROR: syntax: space not allowed before "'"
Stacktrace:
 [1] top-level scope at none:1
```

simeonschaub:
This slightly changes the error message for things like `a 'b'` in non-space-sensitive contexts from `syntax: space not allowed before "'"` to `syntax: extra token "'" after end of expression`, which I don't think is really a big loss. It just doesn't make much sense to me to explicitly disallow whitespace before `'` in general. I wonder whether we should do the same for `(`, `[` and `{`, since I don't see any reason we should generally disallow whitespace before them as well. Was there a reason this was done in the first place?
fixes #37656


src/julia-parser.scm
```diff
@@ -1162,7 +1162,7 @@
   (let loop ((ex ex))
     (let ((t (peek-token s)))
       (if (or (and space-sensitive (ts:space? s)
-                   (memv t '(#\( #\[ #\{ |'| #\" #\`)))
+                   (memv t '(#\( #\[ #\{ #\' #\" #\`)))
```
JeffBezanson, line 1165: I'm pretty sure the lexer returns ' as a symbol.

simeonschaub: Ah, ok, I was a bit confused by that. So the lexer returns `(`, `[`, ... as characters and `'` as a token? At first I thought this was causing the issue, but apparently both work. I can change this back.

simeonschaub:
```suggestion
                   (memv t '(#\( #\[ #\{ |'| #\" #\`)))
```

JeffBezanson (approved):

JeffBezanson:
On `(`, see #7232. I think parens are a bit different. `if f 'x' end` is pretty clearly the condition `f` followed by a character literal. But, `if f (x) end` is ambiguous since in some languages people do put spaces between function names and parenthesized arguments. So I think we should continue to disallow that.

PR #37678, opened 2020-09-21
Show any compile time spent in time and timev macros

IanButterworth:
Adds a report of any compilation time to the `@time` and `@timev` macros, as a % of total time

```
julia> x = rand(10,10);

julia> @time x * x;
  0.606588 seconds (2.19 M allocations: 116.555 MiB, 3.75% gc time, 99.94% compilation time)

julia> @time x * x;
  0.000009 seconds (1 allocation: 896 bytes)
```

I was wondering if `compiler_start_time` should be set earlier than it is here, to catch more compiler-related time?

### DIFF
--- a/base/timing.jl
+++ b/base/timing.jl
@@ -55,6 +55,9 @@ function gc_alloc_count(diff::GC_Diff)
     diff.malloc + diff.realloc + diff.poolalloc + diff.bigalloc
 end
 
+# cumulative total time spent on compilation
+cumulative_compile_time_ns() = ccall(:jl_cumulative_compile_time_ns, UInt64, ())
+reset_cumulative_compile_time() = ccall(:jl_reset_cumulative_compile_time, Cvoid, ())
 
 # total time spend in garbage collection, in nanoseconds
 gc_time_ns() = ccall(:jl_gc_total_hrtime, UInt64, ())
@@ -102,7 +105,7 @@ function format_bytes(bytes) # also used by InteractiveUtils
     end
 end
 
-function time_print(elapsedtime, bytes=0, gctime=0, allocs=0)
+function time_print(elapsedtime, bytes=0, gctime=0, allocs=0, compile_time=0)
     timestr = Ryu.writefixed(Float64(elapsedtime/1e9), 6)
     length(timestr) < 10 && print(" "^(10 - length(timestr)))
     print(timestr, " seconds")
@@ -117,6 +120,9 @@ function time_print(elapsedtime, bytes=0, gctime=0, allocs=0)
     end
     if gctime > 0
         print(", ", Ryu.writefixed(Float64(100*gctime/elapsedtime), 2), "% gc time")
+    end
+    if compile_time > 0
+        print(", ", Ryu.writefixed(Float64(100*compile_time/elapsedtime), 2), "% compilation time")
     end
     if bytes != 0 || allocs != 0
         print(")")
@@ -170,12 +176,15 @@ macro time(ex)
     quote
         while false; end # compiler heuristic: compile this block (alter this if the heuristic changes)
         local stats = gc_num()
+        local compile_elapsedtime = cumulative_compile_time_ns()
         local elapsedtime = time_ns()
         local val = $(esc(ex))
         elapsedtime = time_ns() - elapsedtime
+        compile_elapsedtime = cumulative_compile_time_ns() - compile_elapsedtime
         local diff = GC_Diff(gc_num(), stats)
+        reset_cumulative_compile_time() # reduce risk of overflow on cumulative counter
         time_print(elapsedtime, diff.allocd, diff.total_time,
-                   gc_alloc_count(diff))
+                   gc_alloc_count(diff), compile_elapsedtime)
         println()
         val
     end
--- a/base/timing.jl
+++ b/base/timing.jl
@@ -57,7 +57,6 @@ end
 
 # cumulative total time spent on compilation
 cumulative_compile_time_ns() = ccall(:jl_cumulative_compile_time_ns, UInt64, ())
-reset_cumulative_compile_time() = ccall(:jl_reset_cumulative_compile_time, Cvoid, ())
 
 # total time spend in garbage collection, in nanoseconds
 gc_time_ns() = ccall(:jl_gc_total_hrtime, UInt64, ())
@@ -175,7 +174,6 @@ julia> @time begin
 macro time(ex)
     quote
         while false; end # compiler heuristic: compile this block (alter this if the heuristic changes)
-        reset_cumulative_compile_time() # reduce risk of overflow on cumulative counter
         local stats = gc_num()
         local compile_elapsedtime = cumulative_compile_time_ns()
         local elapsedtime = time_ns()
@@ -212,7 +210,6 @@ malloc() calls:    1
 macro timev(ex)
     quote
         while false; end # compiler heuristic: compile this block (alter this if the heuristic changes)
-        reset_cumulative_compile_time() # reduce risk of overflow on cumulative counter
         local stats = gc_num()
         local compile_elapsedtime = cumulative_compile_time_ns()
         local elapsedtime = time_ns()
--- a/base/timing.jl
+++ b/base/timing.jl
@@ -149,8 +149,8 @@ end
 
 A macro to execute an expression, printing the time it took to execute, the number of
 allocations, and the total number of bytes its execution caused to be allocated, before
-returning the value of the expression. If the time reported includes time spent doing 
-garbage collection (gc) or compilation, that will also be highlighted as a %.
+returning the value of the expression. Any time spent garbage collecting (gc) or
+compiling is shown as a percentage.
 
 See also [`@timev`](@ref), [`@timed`](@ref), [`@elapsed`](@ref), and
 [`@allocated`](@ref).
--- a/base/timing.jl
+++ b/base/timing.jl
@@ -280,6 +280,31 @@ macro allocated(ex)
 end
 
 """
+    @compiletime
+
+A macro to evaluate an expression, discarding the resulting value, instead returning the
+total time in seconds of code generation during evaluation of the expression.
+
+See also [`@time`](@ref), [`@timev`](@ref), [`@timed`](@ref),
+and [`@elapsed`](@ref).
+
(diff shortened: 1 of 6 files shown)


base/timing.jl
```diff
@@ -169,13 +175,16 @@ julia> @time begin
 macro time(ex)
     quote
         while false; end # compiler heuristic: compile this block (alter this if the heuristic changes)
+        reset_cumulative_compile_time() # reduce risk of overflow on cumulative counter
```
yuyichao, line 178: This breaks nesting.

yuyichao (changes requested):

base/timing.jl
ianshmean: Ah yes. Perhaps it could go at parse-time

yuyichao: It'll just cause more strange corner cases. There's no need for this. It won't overflow before time_ns does.

DilumAluthge:
This will be a great feature to have!

base/timing.jl
ianshmean: Good point. Removed 👍🏻

quinnj:
I agree; this seems like a great idea to help understand where time is being spent.

DilumAluthge:
Question: for the cases in which there was no compilation, does it make sense to have `0% compilation time` just for completeness and consistency? Or would that just be noise?

My instinct is that it would be unnecessary. But I just wanted to throw the idea out there.

ianshmean:
I have an uncommitted macro to return the actual value in seconds, if it's useful? I wasn't sure given the dedicated more detailed timing information that's available as a build option
```
julia> @compiletime x * x
0.54208595

julia> @compiletime x * x
0.0
```

DilumAluthge:
Also, is it too late to get this feature into 1.6? I have lost track of time and I have no idea when the feature freeze deadline is for 1.6.

Personally I hope we can get this in to 1.6.

DilumAluthge:
> I have an uncommitted macro to return the actual value in seconds, if it's useful? I wasn't sure given the dedicated more detailed timing information that's available as a build option
> 
> ```
> julia> @compiletime x * x
> 0.54208595
> 
> julia> @compiletime x * x
> 0.0
> ```

Nah... I don't think I can come up with a use case in which I would ask for this.

I think the approach you currently have in this PR is more useful.

ericphanson:
> I have an uncommitted macro to return the actual value in seconds, if it's useful? I wasn't sure given the dedicated more detailed timing information that's available as a build option

I think it's useful to be able to easily access that info, but maybe it could be part of `@timed` instead of a new macro?

ianshmean:
I thought about adding it to the end of the output of `@timed` but it would increase the length of the output tuple, which might be a breaking change?

ericphanson:
True, good question. In https://github.com/JuliaLang/julia/pull/34149 it was made a `NamedTuple` instead of a `Tuple` and @StefanKarpinski commented in https://github.com/JuliaLang/julia/issues/34147#issuecomment-567514637 that this was OK since it iterated the values the same. That reasoning would still hold here for unpacking like `value, time, bytes, gctime, gcdiff = @timed(...)` (since Julia will just drop the last element if you don't iterate it) but maybe this is a bigger change still since if you did something like `(1.0, @timed(...)...)` for some reason then you'd get a different result now whereas https://github.com/JuliaLang/julia/pull/34149 wouldn't have broken that .

ianshmean:
Perhaps better to play safe and add a TODO for 2.0? Until then, the non-exported underlying function could be used:
```
julia> begin
           before = Base.cumulative_compile_time_ns()
           x - x    # thing to test
           comptime_s =  (Base.cumulative_compile_time_ns() - before) / 1e9
       end
0.072066702
...
0.0
```

ianshmean:
I guess needs review by @vtjnash / @JeffBezanson ?

yuyichao (dismissed):

ianshmean:
I've added the `@compiletime` macro so all ideas here can be reviewed

base/timing.jl
```diff
@@ -279,6 +279,31 @@ macro allocated(ex)
...
+julia> x = rand(10,10);
+
+julia> @compiletime x * x
+0.54208595
+```
+"""
+macro compiletime(ex)
+    quote
+        while false; end # compiler heuristic: compile this block (alter this if the heuristic changes)
+        local t0 = cumulative_compile_time_ns()
```
yuyichao, line 301: Note that this number is inaccurate since it ignores the compile time of the macro itself.

It was OK as an additional info in `@time` output since the number is to be compared to the other outputs none of which includes the actual compile time. However, this is not the case when the compile time is the only output. People will think `@compiletime function_that_take_a_minute_to_compile()` has no compile time. At a minimum, the `while false; end` should be removed and even that is not accurate since the user code can still trigger a compilation.

ianshmean: Thanks for the additional review. I'm happy to revert the last commit. Just wanted to check that reviewers had considered it, given it came up in discussion.

For the sake of clarity though, I'm not sure I understand a few of your points. 

> inaccurate since it excludes the compile time of the macro itself

Why would we want to include the compile time of the macro in this timing? It's this way by design. 

> none of which includes the actual compile time. 

The total of `@time` includes any compile time, that's the motivation for this PR. To highlight when compile time is present, as it usually dominates. 

> people will think ... has no compile time

If the function you describe does take a minute to compile, the first time it's `@compiletime`-ed, it would return 60 seconds. Though, from then onwards it would return 0, which could be confusing. Is that what you mean?


Again, I'm not pushing for this additional macro too hard. I think the main benefit is the addition to `@time`

yuyichao: > Why would we want to include the compile time of the macro in this timing? It's this way by design.

Because that compilation might include the compilation of the function in question. See below.

> The total of `@time` includes any compile time, that's the motivation for this PR. To highlight when compile time is present, as it usually dominates.

No it doesn't. It sometimes include the compilation of the function on the first use which is why adding compilation time is useful. However, even on the first execution, the compile time may not be included, and that's exactly what that comment on the first line of the macro expansion means. Again, see below.

> If the function you describe does take a minute to compile, the first time it's `@compiletime`-ed, it would return 60 seconds. Though, from then onwards it would return 0, which could be confusing. Is that what you mean?

Nope. See the example below (and change the 1000 to 10000 if you dare.... that gives a longer than 1min compilation time for me for sure, though to be fare, it was still inferring after 5 min so I'm not sure how long each phase takes).

```julia
julia> @eval function f(a)
           $(quote $([:(a = a + a * 10 + (a > 0 ? $i : 0.2)) for i in 1:1000]...) end)
           $(quote $([:(a = a + a * 10 + (a > 0 ? $i : 0.2)) for i in 1:1000]...) end)
           return a
       end

julia> @time f(1)
  0.000000 seconds
Inf
```

The point I used a no argument function is that that function is guaranteed to be inferred and compiled based on the current implementation.

ianshmean: I see. So there may be more compilation going on related to the function under test that none of this timing tooling is catching, so it's only relevant to report when the time returned by `@time` contains any compiled time. (and in that case it doesn't make sense to report an absolute time given it's not a total, so the % is adequate as a highlight)

Makes sense, thanks for the explanation. I'll revert

ianshmean:
Friendly bump for review/merge 🙏

DilumAluthge:
@yuyichao Is this ready to go?

ianshmean:
I can’t assign reviewers, so if this needs more review can someone please assign the appropriate people

yuyichao (approved):
This can probably use a bit more mention in the doc and the C++ code can also use `+=` to be a bit more concise. Neither of these are blocking issues for the PR.

ianshmean:
Good point. The docstring now reads:
___
    @time

A macro to execute an expression, printing the time it took to execute, the number of
allocations, and the total number of bytes its execution caused to be allocated, before
returning the value of the expression. If the time reported includes time spent doing 
garbage collection (gc) or compilation, that will also be highlighted as a %.

See also [`@timev`](@ref), [`@timed`](@ref), [`@elapsed`](@ref), and [`@allocated`](@ref).

!!! note
    For more serious benchmarking, consider the `@btime` macro from the BenchmarkTools.jl
    package which among other things evaluates the function multiple times in order to
    reduce noise.

```
julia> x = rand(10,10);

julia> @time x * x;
  0.606588 seconds (2.19 M allocations: 116.555 MiB, 3.75% gc time, 99.94% compilation time)

julia> @time x * x;
  0.000009 seconds (1 allocation: 896 bytes)

julia> @time begin
           sleep(0.3)
           1+1
       end
  0.301395 seconds (8 allocations: 336 bytes)
2
```
___

DilumAluthge:
@ianshmean Perhaps you could also add a compilation time example to the docstring for the `@timev` macro?

ianshmean:
Should be good to go now

base/timing.jl
```diff
@@ -144,7 +149,8 @@ end
 
 A macro to execute an expression, printing the time it took to execute, the number of
 allocations, and the total number of bytes its execution caused to be allocated, before
-returning the value of the expression.
+returning the value of the expression. If the time reported includes time spent doing
```
KristofferC, line 152: Maybe just

```suggestion
returning the value of the expression. Eventual time spent garbage collecting (gc) or compiling is shown as a percentage.
```

The literal `%` feels a bit surprising in running text.

KristofferC:
Needs whitespace fix.

base/timing.jl
ianshmean: Agreed. "Eventual" didn't feel quite right so went with "Any"

ianshmean:
Failures seem unrelated?

KristofferC:
:tada:

base/timing.jl
NHDaly: Thanks both for the above conversation! I was really concerned about this, since even _before_ this change I was recently misled to find that code I thought took 30 seconds to compile actually too _100 seconds_ to compile, but most of the compilation time was spent in the top-level inference, exactly as in @yuyichao's example here. 😮 

I was worried about this new `% compilation time` metric (which i'm super thrilled about!!!) making it even easier to be confused.

So @vchuravy and I took a stab at reporting the top-level compilation time as well, here:
https://github.com/JuliaLang/julia/pull/37938

PR #37692, opened 2020-09-22
[BinaryPlatforms]: Impose an ordering on `arch_march_isa_mapping`

staticfloat:
We want to be able to easily walk these in "compatibility order", so we
arrange them in order.  It may be that we eventually have multiple
"branches", btu since there's a nice linearization right now, let's see
how far we can walk with this.

I've also switched to vectors rather than tuples, to make it easier to
modify these in the future if we must.


staticfloat:
@giordano does this look good to you?

giordano:
Yes, looks good :+1:

PR #37705, opened 2020-09-23
[BinaryPlatforms]: Remove overzealous verification

staticfloat:
Some of these verifications are a little over-zealous.  We don't
actually need `cuda` or `march` verification in here, since these will
be primarily externally-driven tags.


vtjnash:
Why is it `March`ing here anyways? In the Makefile, there's a gcc flag name combined from `-m arch` that sources this name, but thus that name doesn't make sense outside of assembly a gcc command line.

staticfloat:
These `march` tags are what allow us to encode different microarchitectures into Artifacts.tomls, so we can use `gcc -march=haswell -mtune=haswell` to build an AVX2 version of a library, for instance.

PR #37707, opened 2020-09-23
RFC: add explicit precompile statements for package precompilation and JLL loading

KristofferC:
The new JLLs are very quick to load, however, their precompile time is still quite bad. For example

```
julia> @time using GTK3_jll
[ Info: Precompiling GTK3_jll [77ec8976-b24b-556a-a1bf-49a033a670a6]
26.414237 seconds (1.07 M allocations: 64.685 MiB, 0.02% gc time)


# Restart and load (so fast, much wow)
julia> @time using GTK3_jll
  0.075747 seconds (198.91 k allocations: 13.934 MiB)
```

One reason for this is that JLL packages have a lot of dependencies and parts of the Artifact and BinaryPlatform code have to be compiled for every dependency when the precompilation process happens. This PR recorded what needed to be compiled and just adds that to the list of precompile statements.

With this PR:

```
julia> @time using GTK3_jll
[ Info: Precompiling GTK3_jll [77ec8976-b24b-556a-a1bf-49a033a670a6]
14.635319 seconds (983.36 k allocations: 59.006 MiB, 0.04% gc time)
```

so about a reduction of 50%.

It would probably be better to record this during build time like most of the other precompile statements but that requires us to have a "trial" JLL package to load and JLLPackages depend on JLLWrappers so it is slightly non-trivial.

cc @staticfloat


KristofferC:
Should redo this on top of https://github.com/JuliaLang/julia/pull/37710.

KristofferC:
With https://github.com/JuliaLang/julia/pull/37715 and https://github.com/JuliaLang/julia/pull/37710

```
julia> @time using GTK3_jll
[ Info: Precompiling GTK3_jll [77ec8976-b24b-556a-a1bf-49a033a670a6]
 11.819610 seconds (693.43 k allocations: 40.500 MiB)
```

KristofferC:
Doesn't feel as important anymore.

PR #37710, opened 2020-09-23
more effectively record precompile signatures from package precompilation

KristofferC:
This allows us to trace the precompile statements emitted in the process that is doing the precompilation. Use that to gather statements in the precompile process from executing `compilecache` on a module.

Also gets rid of the ugly `__PackagePrecompilationStatementModule` from `Base.loaded_modules`. The `using` of that module was in fact completely pointless because apparently precompilation doesn't happen when you load a package from a Julia instance that is creating a sysimage.

Before:
```
julia> @time using GTK3_jll
 [ Info: Precompiling GTK3_jll [77ec8976-b24b-556a-a1bf-49a033a670a6]
 25.203725 seconds (799.52 k allocations: 47.121 MiB, 0.02% gc time
```

After:

```
julia> @time using GTK3_jll
[ Info: Precompiling GTK3_jll [77ec8976-b24b-556a-a1bf-49a033a670a6]
 22.947980 seconds (701.36 k allocations: 40.960 MiB)
```

### DIFF
--- a/contrib/generate_precompile.jl
+++ b/contrib/generate_precompile.jl
@@ -124,18 +124,22 @@ function generate_precompile_statements()
     debug_output = devnull # or stdout
 
     # Precompile a package
+    global hardcoded_precompile_statements
     mktempdir() do prec_path
         push!(DEPOT_PATH, prec_path)
-        push!(LOAD_PATH, prec_path)
         pkgname = "__PackagePrecompilationStatementModule"
         mkpath(joinpath(prec_path, pkgname, "src"))
-        write(joinpath(prec_path, pkgname, "src", "$pkgname.jl"),
+        path = joinpath(prec_path, pkgname, "src", "$pkgname.jl")
+        write(path,
               """
               module $pkgname
               end
               """)
-        @eval using __PackagePrecompilationStatementModule
-        empty!(LOAD_PATH)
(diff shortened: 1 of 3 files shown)


KristofferC:
Interesting

```
ErrorException("More than one command line CPU targets specified without a `--output-` flag specified")
```

does the CPU target matter at all for the precompilation process?

contrib/generate_precompile.jl
```diff
@@ -122,21 +122,32 @@ end
...
+        # Running compilecache on buildbots fails with
+        # `More than one command line CPU targets specified without a `--output-` flag specified`
+        # so start a new process without a CPU target specified
+        s = """
+            push!(DEPOT_PATH, $(repr(prec_path)));
+            Base.PRECOMPILE_TRACE_COMPILE[] = $(repr(tmp));
+            Base.compilecache(Base.PkgId($(repr(pkgname))), $(repr(path)))
+            """
+        run(`$(julia_exepath()) -O0 --sysimage $sysimg -Cnative -e $s`)
+        t = read(tmp, String)
```
JeffBezanson, line 149: delete this line?

KristofferC: Ops, yes. Thanks.

PR #37714, opened 2020-09-23
Improve typesubtract for tuples (repeat #35600)

vtjnash:
This re-introduces the improvements #35600—recently reverted—to sharpen the information on Tuple objects through an isa-split (with some improvements of my own towards handling UnionAll over Union), but now limited by the MAX_UNION_SPLITTING parameter to prevent exponential blowup.

### DIFF
--- a/base/compiler/abstractinterpretation.jl
+++ b/base/compiler/abstractinterpretation.jl
@@ -621,7 +621,7 @@ function abstract_iteration(interp::AbstractInterpreter, @nospecialize(itft), @n
     while valtype !== Any
         stateordonet = abstract_call_known(interp, iteratef, nothing, Any[Const(iteratef), itertype, statetype], sv).rt
         stateordonet = widenconst(stateordonet)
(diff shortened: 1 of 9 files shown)


JeffBezanson (approved):

mhauru:
Would there be a downside to making the `MAX_UNION_SPLITTING::Int` argument of `typesubtract` optional, with a sensible default, or maybe with one that reproduces the old behavior? I noticed the change because commit https://github.com/JuliaLang/julia/commit/5a56ecdd767709c844f65d96be4d35c35698cbc3 broke a custom version of `@inferred` in https://github.com/Jutho/TestExtras.jl, that's calling `typesubtract`. I guess these are pretty deep internals, and the change is very easy to adapt to, but it still seems like unnecessary breakage to me.

vtjnash:
See af6542e259f688950004e19f2685ceb4511de774. As you guessed, packages should not rely upon access to `Core.Compiler`

PR #37720, opened 2020-09-23
[LLVM] stage git checkout through a bare repository

vchuravy:
I find myself these days switching between LLVM versions quite often
and I use out-of-tree to manage the different variants. 

This stages the checkout of the LLVM git repository through a local
cache (a bare repository), and then uses a local clone to create a
cheap copy that is shared in `LLVM_GIT_VER`.


deps/llvm.mk
```diff
@@ -372,8 +372,12 @@ ifneq ($(LLVM_VER),svn)
 	mkdir -p $(LLVM_SRC_DIR)
 	$(TAR) -C $(LLVM_SRC_DIR) --strip-components 1 -xf $(LLVM_TAR)
 else
+	([ ! -d $(LLVM_BARESRC_DIR) ] && \
+		git clone --bare $(LLVM_GIT_URL) $(LLVM_BARESRC_DIR) ) || \
+		(cd $(LLVM_BARESRC_DIR) && \
+		git fetch)
 	([ ! -d $(LLVM_MONOSRC_DIR) ] && \
-		git clone $(LLVM_GIT_URL) $(LLVM_MONOSRC_DIR) ) || \
+		git clone --local $(LLVM_BARESRC_DIR) $(LLVM_MONOSRC_DIR) ) || \
```
vtjnash, line 380:
```suggestion
		git clone --dissociate --reference $(LLVM_BARESRC_DIR) $(LLVM_GIT_URL) $(LLVM_MONOSRC_DIR) ) || \
```

deps/llvm.mk
```diff
@@ -574,8 +578,10 @@ check-llvm: $(LLVM_BUILDDIR_withtype)/build-checked
 
 ifeq ($(LLVM_VER),svn)
 update-llvm:
-	cd $(LLVM_MONOSRC_DIR) && \
-		git pull --ff-only
+	(cd $(LLVM_BARESRC_DIR) && \
+		git fetch)
+	(cd $(LLVM_MONOSRC_DIR) && \
+		git pull --ff-only)
```
vtjnash, line 584:
```suggestion
		git fetch $(LLVM_BARESRC_DIR) +refs/remotes/*:refs/remotes/* && \
		git pull --ff-only)
```

vtjnash:
WDYT?

vchuravy:
Why `--disassociate`? I mean I am on btrfs so I probably get the space savings anyway ;), but I used `--local` to not create a copy.

vtjnash:
That also creates a copy (that flag is a no-op)?

vchuravy:
Urgh this should have been a squash merge, sorry about that.

PR #37722, opened 2020-09-23
[BinaryPlatforms]: Flesh out CPUID support

staticfloat:
This PR fleshes out (and slightly renames) some pieces of the CPUID
machinery to better support BB's multi-march system.  It declares at
least one valid microarchitecture for all architectures, and splits the
CPUID module out into its own source file for readability.

### DIFF
--- a/base/cpuid.jl
+++ b/base/cpuid.jl
@@ -1,0 +1,79 @@
+module CPUID
+
+export cpu_isa
+
+"""
+    ISA(features::Set{UInt32})
+
+A structure which represents the Instruction Set Architecture (ISA) of a
+computer.  It holds the `Set` of features of the CPU.
+
+The numerical values of the features are automatically generated from the C
+source code of Julia and stored in the `features_h.jl` Julia file.
+"""
+struct ISA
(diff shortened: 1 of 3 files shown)


staticfloat:
@giordano just FYI

@vchuravy I guess we have no C CPUID code on power yet?  Too bad, we'll just compile everything for power8 for now then, until Julia gains the ability to select higher forms of computation.

base/cpuid.jl
```diff
@@ -63,8 +63,30 @@ const ISAs_by_family = Dict(
...
+    arch = lowercase(arch)
+    if arch ∈ ("amd64",)
+        arch = "x86_64"
+    elseif arch ∈ ("i386", "i486", "i586")
+        arch = "i686"
+    elseif arch ∈ ("armv6",)
+        arch = "armv6l"
+    elseif arch ∈ ("arm", "armv7", "armv8", "armv8l")
+        arch = "armv7l"
+    elseif arch ∈ ("arm64")
```
giordano, line 80:
```suggestion
    elseif arch ∈ ("arm64",)
```

PR #37729, opened 2020-09-24
fix #37664: treat comments as whitespace

### LINKED ISSUE #37664: Block comments between code are parsed to unexpected expression
```
julia> a = 3
3

julia> a a
ERROR: syntax: extra token "a" after end of expression
Stacktrace:
 [1] top-level scope at none:1

julia> a#==#a
9

julia> Meta.parse("a#==#a")
:(a * a)
```

should probably be parsed as `:(a a)` instead

belamenso:
Issue #37664 arises because of incorrect handling of juxtaposition. Function `juxtapose?` returns `#f` on `a b` because of the space between them (checked by `ts:space?`) and (incorrectly) `#t` on `a#==#b` because `next-token` doesn't treat comments as whitespace.
Alternative fix would be to add a (perhaps optional) parameter to `next-token`, but since it's called nonrecursively from two other places in the file I decided on adding an internal function `loop` in order to preserve its interface.


JeffBezanson:
This is great, thanks and welcome!

This could be done with an optional argument, `(define (next-token port s (comment-induced-whitespace #f))`, which has the advantage of not needing to re-indent the whole function.

Also please add a test in test/syntax.jl.

belamenso:
I added the tests and discovered that not only `a#==#a` was broken but also `a #==#a`.

test/syntax.jl
```diff
@@ -2345,3 +2345,9 @@ end
 
 # issue #37656
 @test :(if true 'a' else 1 end) == Expr(:if, true, quote 'a' end, quote 1 end)
+
+# issue #37664
+@test_throws ParseError("extra token \"b\" after end of expression") Meta.parse("a b")
```
simeonschaub, line 2350: Could you add a test in a space-sensitive context as well, like a macro call or an hcat expression?

belamenso:
I added further tests.

belamenso:
Is this pull request moving forward?

PR #37734, opened 2020-09-24
Ensure `hash(::Platform)` is stable

staticfloat:
This allows us to key dictionaries with `Platform` objects more reliably

### DIFF
--- a/base/binaryplatforms.jl
+++ b/base/binaryplatforms.jl
@@ -142,6 +142,20 @@ function Base.setindex!(p::AbstractPlatform, v::String, k::String)
     return p
 end
 
+# Hash definitino to ensure that it's stable
+function Base.hash(p::Platform, h::UInt)
+    h += 0x506c6174666f726d % UInt
+    h = hash(p.tags, h)
+    h = hash(p.compare_strategies, h)
+    return h
+end
+
+# Simple equality definition; for compatibility testing, use `platforms_match()`
+function Base.:(==)(a::Platform, b::Platform)
+    return a.tags == b.tags && a.compare_strategies == b.compare_strategies
+end
+
+
 # Allow us to easily serialize Platform objects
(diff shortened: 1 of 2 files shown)


base/binaryplatforms.jl
```diff
@@ -141,6 +141,7 @@ function Base.setindex!(p::AbstractPlatform, v::String, k::String)
     add_tag!(tags(p), k, v)
     return p
 end
+Base.hash(p::Platform) = hash(hash(a.tags), hash(a.comparison_strategies))
```
mbauman, line 144: Should this include `comparison_stratiges` if `==` does not check them?  Or perhaps `==` should include them?

https://github.com/JuliaLang/julia/pull/37734/files#diff-d985676b7739e0f47a51d1f028ec6fedR171

This should also probably include a magic constant so it doesn't hash the same as other data structures.

staticfloat: Yeah, `==` should include them.  Thanks, good catch!

staticfloat: Actually, my `==` definition doesn't really make sense.  :P

staticfloat: Not to mention the fact that I used `p` in the arguments and `a` in the definition.  🙄

base/binaryplatforms.jl
```diff
@@ -144,7 +144,7 @@ end
 
 # Hash definitino to ensure that it's stable
 function Base.hash(p::Platform)
```
rfourquet, line 146: I don't know the use-case here, but shouldn't this be `Base.hash(::Platform, ::UInt)` as recommended by the `hash` docs?

staticfloat: Thanks!  I clearly didn't read the docs properly.  ;)

PR #37736, opened 2020-09-25
fix #37690, bug in `lambda-optimize-vars!` on outer variables assigned in loops

### LINKED ISSUE #37690: Possible bug in scope behavior
```
function foo()
	local f
	local x
	for k = 1 : 2
		x = k
		if k == 1
			f = () -> x
		end
	end
	f
end
```
```
julia> foo()()
1
```
I'm explicitly declaring `x` as a local variable outside the for loop, so the result of `foo()()` should be `2` (the value of `x` at the end of the function). Indeed, if I replace `local x` by `local x = 0`, the result of `foo()()` is `2`.

```
julia> versioninfo()
Julia Version 1.5.0
Commit 96786e22cc (2020-08-01 23:44 UTC)
Platform Info:
  WORD_SIZE: 64
  LIBM: libopenlibm
```

JeffBezanson:
fix #37690

### DIFF
--- a/src/julia-syntax.scm
+++ b/src/julia-syntax.scm
@@ -3228,7 +3228,7 @@ f(x) = yt(x)
                    (lambda (x) (and (pair? x) (not (eq? (car x) 'lambda)))))))
 
 (define lambda-opt-ignored-exprs
-  (Set '(quote top core line inert local local-def unnecessary copyast
+  (Set '(quote top core line inert local-def unnecessary copyast
          meta inbounds boundscheck loopinfo decl aliasscope popaliasscope
          thunk with-static-parameters toplevel-only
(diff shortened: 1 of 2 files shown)


JeffBezanson:

nanosoldier:
[Your package evaluation job](https://github.com/JuliaLang/julia/pull/37736#issuecomment-704407254) has completed - possible new issues were detected. A full report can be found [here](https://github.com/JuliaCI/NanosoldierReports/blob/ff7bcc76370965a2875d78118392f63ae16216c6/pkgeval/by_hash/63c6892_vs_08a233d/report.md). cc @maleadt

JeffBezanson:
Ok, all the new package test failures are not reproducible (mostly numerical issues).

PR #37739, opened 2020-09-25
fix not loading startup.jl in generate_precompile.jl

### LINKED ISSUE #37737: ~/.julia/config/startup.jl is accessed while building `julia`
Since somewhere between 2a36f83ce9bdfd17c572472f60b830aedcf181f0 and 3b55daef61c728f2dab2973db953111d0a6a6ca9, it seems that `generate_precompile.jl` started to accidentally load `~/.julia/config/startup.jl`. Here is an MWE:

```sh
rm ~/.julia/config/startup.jl &&
echo oops > ~/.julia/config/startup.jl &&
rm -v usr/lib/julia/sys* &&
make
```

then I get

```
....
Sysimage built. Summary:
Total ───────  51.493338 seconds
Base: ───────  21.010576 seconds 40.8025%
Stdlibs: ────  30.481181 seconds 59.1944%
    JULIA usr/lib/julia/sys-o.a
ERROR: LoadError: UndefVarError: oops not defined
in expression starting at /home/takafumi/.julia/config/startup.jl:1
...
```

I'm running a brute-force search here: https://github.com/tkf/julia/actions/runs/271800886

@staticfloat Can we have something like

```julia
error("startup.jl is accessed while building `julia`")
```

in `~/.julia/config/startup.jl` during the build process to avoid regression like this?

KristofferC:
Fixes https://github.com/JuliaLang/julia/issues/37737

cc @tkf

### DIFF
--- a/contrib/generate_precompile.jl
+++ b/contrib/generate_precompile.jl
@@ -145,7 +145,7 @@ function generate_precompile_statements()
             Base.PRECOMPILE_TRACE_COMPILE[] = $(repr(tmp));
             Base.compilecache(Base.PkgId($(repr(pkgname))), $(repr(path)))
             """
-        run(`$(julia_exepath()) -O0 --sysimage $sysimg -Cnative -e $s`)
+        run(`$(julia_exepath()) -O0 --sysimage $sysimg --startup-file=no -Cnative -e $s`)
         hardcoded_precompile_statements *= "\n" * read(tmp, String)
     end
 


tkf:
Thanks! Re-running MWE: https://github.com/tkf/julia/runs/1164434873?check_suite_focus=true

tkf:
It works! https://github.com/tkf/julia/runs/1164434873?check_suite_focus=true#step:5:2200

